PR #116, opened 2019-07-03
[GOAL2-720] Make agreement handle network failures/delays during FPR

tsachiherman:
## Summary

The existing FPR is sending a bundle, followed by a series a votes. That series could be of a substantial size. We want to send the votes to the network library, but not to overflow it. If the network library cannot handle any more messages, we want to stop.

## High Level Changes

* Have the Broadcast and Relay commands return the underlying error ( if any )
* If we had an error while sending the FPR messages, log that and stop sending the remainder of the messages ( since they are likely to fail for the same reason ).
* Since we're sending many votes, we also need to check the context, as it may expire over time.


agreement/gossip/network.go
```diff
@@ -48,7 +48,8 @@ type networkImpl struct {
 	proposalCh chan agreement.Message
 	bundleCh   chan agreement.Message
 
-	net network.GossipNode
+	net           network.GossipNode
+	backgroundCtx context.Context
```
derbear: What's the purpose of this `backgroundCtx` field?

tsachiherman: Our usage of the network library is limited; we never use the blocking version of it ( where the provided context would make a difference ).

So, instead of calling the context.Background() on every iteration, I cached that value.

I don't expect to see any notable gains here; this object is static anyway.

derbear: It seems like something that would just be optimized out during compilation (see https://golang.org/src/context/context.go?s=7302:7327#L196). I think it makes more sense to just call it on every iteration so that we can get rid of this field.

agreement/abstractions.go
```diff
@@ -251,7 +251,7 @@ type Network interface {
 	//
 	// Calls to Broadcast by the agreement package are currently guaranteed
 	// to be serialized.
-	Broadcast(protocol.Tag, []byte)
+	Broadcast(protocol.Tag, []byte) error
```
derbear: Can we document what happens here when an error is returned?

agreement/actions.go
```diff
@@ -119,9 +119,21 @@ func (a networkAction) String() string {
...
 	if a.T == broadcastVotes {
 		tag := protocol.AgreementVoteTag
-		for _, uv := range a.UnauthenticatedVotes {
+		for i, uv := range a.UnauthenticatedVotes {
 			data := protocol.Encode(uv)
-			s.Network.Broadcast(tag, data)
+			sendErr := s.Network.Broadcast(tag, data)
+			if sendErr != nil {
+				s.log.Warnf("Network was unable to queue votes for broadcast(%v). %d / %d votes for round %d period %d step %d were dropped.",
+					len(a.UnauthenticatedVotes)-i, len(a.UnauthenticatedVotes),
```
derbear: I believe the arguments here are in the wrong order: `sendErr` should come before the number of votes.

Vervious:
Looks good to me (except I agree with Derek's documentation comment)

Vervious:
Saw another `db.TestDBConcurrencyRW timed out`, restarted build. see GOAL2-783

PR #118, opened 2019-07-05
Swagger changed --base-path to --work-dir

Karmastic:
Fix build

### DIFF
--- a/daemon/algod/api/client/models/models.go
+++ b/daemon/algod/api/client/models/models.go
@@ -27,6 +27,8 @@
 // sticking with a hard-coded internal approach.
 package models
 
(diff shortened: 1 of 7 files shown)


Vervious:
We shouldn't need to change any of the lib.Bytes, what problem does that fix? In particular we shouldn't be generating a swagger from client/models/model.go - that's for the go-client only and copied over from v1/handlers/model.go.

Karmastic:
This was field surgery to get the build working -- which failed.  Ben will fix this the proper way shortly.
See GOAL2-844: https://algorand.atlassian.net/browse/GOAL2-844

PR #121, opened 2019-07-05
Eric/goal2 569 multiple srv servers

egieseke:
## Summary

Support retrieving peer addresses using multiple SRV records

## Test Plan

Ran the existing set of unit tests.


network/wsNetwork.go
```diff
@@ -1162,16 +1161,24 @@ func (wn *WebsocketNetwork) meshThread() {
...
+				multiPhonebook.AddPhonebook(dnsPhonebook)
+				dnsStatus = true
+			} else {
+				wn.log.Debugf("got no DNS addrs for network %#v", wn.NetworkID)
 			}
-		} else {
-			wn.log.Debugf("got no DNS addrs for network %#v", wn.NetworkID)
+		}
+		if dnsStatus {
+			wn.phonebook = multiPhonebook
```
Karmastic: wn.phonebook needs to be set only once.  There are several sources that contribute to this multi-phonebook and DNSBoostrap results are only one.
I would rewrite to store a map of boostrapIDs to Phonebook.  Then a refresh would be: add/update dnsPhonebooks, mp.AddPhonebook(addedOrUpdated).  We can assume for now we'll never have to REMOVE since thie DNSBootstrapArray is based on a fixed string (from config.json).  The key is you can't just replace wn.phonebook as that wipes out other entries (like the -p set from the cmdline)

egieseke: Added map of phonebooks to MultiPhonebook.  Replace the map entries when updating DNS Bootstrap entries for SRV records.

network/phonebook.go
```diff
@@ -114,24 +114,36 @@ func (p *ThreadsafePhonebook) ReplacePeerList(they []string) {
...
-		return mp.phonebooks[0].GetAddresses(n)
+
+	phonebookList := []Phonebook{}
+
+	for _, phonebook := range mp.phonebookMap {
+		phonebookList = append(phonebookList, *phonebook)
 	}
-	sizes := make([]int, len(mp.phonebooks))
+
+	if len(phonebookList) == 1 {
```
Karmastic: Change for loop: `if len(mp.phonebookMap) == 0 { return phonebook.GetAddresses(n) }`

egieseke: There is no longer a phonebook variable to get addresses from, only the map which includes a "default" map.

config/config.go
```diff
@@ -681,6 +681,23 @@ func loadConfig(reader io.Reader, config *Local) error {
 	return dec.Decode(config)
 }
 
+// DNSBootstrapArray returns an array of one or more DNS Bootstrap identifiers
+func (cfg Local) DNSBootstrapArray(networkID protocol.NetworkID) (bootstrapArray []string) {
+	dnsBootstrapString := cfg.DNSBootstrap(networkID)
```
Karmastic: Why not just return strings.Split(bootstrapString, ";") ?

egieseke: It does not handle all cases, for example, if there is a leading or trailing semicolon or whitespace.  https://medium.com/@mlowicki/strings-fieldsfunc-vs-strings-split-96c667912f78

tsachiherman: you can always use ` strings.Split(bootstrapString, "; ")`, right ?
it's much more intuitive than the code below.
We don't have to support any user input here. The config file is assumed to be crafted correctly.

network/phonebook.go
```diff
@@ -163,14 +175,9 @@ func (mp *MultiPhonebook) GetAddresses(n int) []string {
 	return out
 }
 
-// AddPhonebook adds a Phonebook if it is new
-func (mp *MultiPhonebook) AddPhonebook(p Phonebook) {
+// AddOrUpdatePhonebook adds or updates Phonebook in Pnonebook map
```
tsachiherman: nit: typo Pnonebook -> PhonebookMap

network/phonebook.go
```diff
@@ -163,14 +175,9 @@ func (mp *MultiPhonebook) GetAddresses(n int) []string {
...
+func (mp *MultiPhonebook) AddOrUpdatePhonebook(name string, p Phonebook) {
```
tsachiherman: Could you please give the variable "name" into some more descriptive name ? maybe bootstrapNetworkName ?

network/wsNetwork.go
```diff
@@ -1162,16 +1161,19 @@ func (wn *WebsocketNetwork) meshThread() {
...
+		for _, dnsBootstrap := range dnsBootstrapArray {
+			dnsAddrs := wn.getDNSAddrs(dnsBootstrap)
+			if len(dnsAddrs) > 0 {
+				wn.log.Debugf("got %d dns addrs, %#v", len(dnsAddrs), dnsAddrs[:imin(5, len(dnsAddrs))])
+				dnsPhonebook := &ThreadsafePhonebook{
+					addrs: dnsAddrs,
+				}
+				multiPhonebook.AddOrUpdatePhonebook(dnsBootstrap, dnsPhonebook)
+			} else {
+				wn.log.Debugf("got no DNS addrs for network %#v", wn.NetworkID)
```
tsachiherman: 1. Change into "Infof"
2. don't use `%#v` everywhere. in this case, I think that `%s` would do.

network/phonebook.go
algobolson: Why wasn't AddPhonebook() enough? It filtered out duplicates and updates to the underlying data were visible because Phonebook is an interface reference type.

network/wsNetwork.go
```diff
@@ -1162,16 +1161,19 @@ func (wn *WebsocketNetwork) meshThread() {
...
 		// TODO: only do DNS fetch every N seconds? Honor DNS TTL? Trust DNS library we're using to handle caching and TTL?
-		dnsAddrs := wn.getDNSAddrs()
-		if len(dnsAddrs) > 0 {
-			wn.log.Debugf("got %d dns addrs, %#v", len(dnsAddrs), dnsAddrs[:imin(5, len(dnsAddrs))])
-			wn.dnsPhonebook.ReplacePeerList(dnsAddrs)
-			mp, ok := wn.phonebook.(*MultiPhonebook)
-			if ok {
-				mp.AddPhonebook(&wn.dnsPhonebook)
+		dnsBootstrapArray := wn.config.DNSBootstrapArray(wn.NetworkID)
+		multiPhonebook := MakeMultiPhonebook()
```
algobolson: it looks like nothing consumes any data gathered into multiPhonebook

config/config.go
Karmastic: and if it's a concern, consumer can trim and ignore empties while iterating (or we could do the same and return a sanitized array).  This approach is overly complicated.

network/phonebook.go
Karmastic: The for loop has a phonebook iterator - that's what I meant.

network/phonebook.go
```diff
@@ -114,24 +114,36 @@ func (p *ThreadsafePhonebook) ReplacePeerList(they []string) {
...
+
+	if len(phonebookList) == 1 {
+		return phonebookList[0].GetAddresses(n)
+	}
+	sizes := make([]int, len(phonebookList))
 	total := 0
-	addrs := make([][]string, len(mp.phonebooks))
-	for pi, p := range mp.phonebooks {
+	addrs := make([][]string, len(phonebookList))
+	for pi, p := range phonebookList {
```
Karmastic: We don't need a phonebookList array at all - just iterate over phonebookMap, and put the `len(map) == 0` check in this one loop.

network/wsNetwork.go
Karmastic: Yeah - i was just noticing that too.  I think this should be mp.phonebookMap -- we shouldn't have a local multiPhonebook instance.

network/phonebook.go
egieseke: Updated to iterate over the map.

config/config.go
egieseke: I started with Split() but updated to use the FieldsFunc() with white space removal so that the test cases would work.  I will revert to split and remove some of the test cases from config_test.TestLocal_DNSBootstrapArray().

network/phonebook.go
egieseke: removed phonebookList array.

network/phonebook.go
egieseke: fixed

network/phonebook.go
egieseke: done

egieseke: the underlying map of phonebooks requires key-value pairs.

algobolson: but it doesn't need to be a map. we never need to 'update(phonebookName, phonebookReference)' because it's already a reference and the value is never stale. it already de-duplicated phonebooks added.

network/wsNetwork.go
egieseke: good catch, fixed.

network/wsNetwork.go
egieseke: updated

network/phonebook.go
egieseke: That is how I started, but it was suggested to manage the DNS bootstrap entries as a map utilizing the MutliPhonebook.

network/phonebook.go
```diff
@@ -175,6 +182,12 @@ func (mp *MultiPhonebook) GetAddresses(n int) []string {
 	return out
 }
 
+func (mp *MultiPhonebook) addAddressArrayToAdressSet(addrMap *map[string]bool, addrArray *[]string) {
```
algobolson: doesn't actually need to be a member of MultiPhonebook since it never uses the `mp` reference. generically it might be `stringMapSetUpdate(...)`

PR #124, opened 2019-07-08
[GOAL2-558] Update S3 Usage

Karmastic:
Builds on Eric's fork to address rough edges after testing all of the scenarios.
Tested everything I could.  Algonet not tested but I'm working on a PR that updates it to set appropriate environment variables for nodecfg and log uploads (there's a new `algonet` IAM user).

There will be some manual work required to transition to the new buckets (need releases to exist in old and new buckets for the first new release, and need to update the bootstrappers).  It should be transparent to most people.

Note that `goal logging send` is now secured -- aws doesn't support anonymous multi-part uploads so we'll have to distribute upload creds as-needed.  This should almost eliminate our risk of S3 / billing abuse, with a slight inconvenience introduced when we need logs uploaded.


Karmastic:
See Confluence for details / usage scenarios: https://algorand.atlassian.net/wiki/spaces/DEVOPS/pages/420118549/S3+Buckets+and+Policies

.travis.yml
```diff
@@ -16,9 +16,9 @@ stages:
   - name: build_release
     if: branch =~ /^rel\// AND type != pull_request
   - name: deploy
-    if: branch =~ /^rel\/nightly/ AND type != pull_request
+    if: (branch =~ /^rel\/nightly/ OR branch =~ /^rel\/dev/) AND type != pull_request
   - name: release
-    if: branch =~ /^rel\/nightly/ AND type != pull_request
+    if: (branch =~ /^rel\/nightly/ OR branch =~ /^rel\/dev/) AND type != pull_request
```
winder: I left this comment on https://github.com/algorand/go-algorand-internal/pull/3
> To support more release branches than just rel/nightly, we should specifically exclude rel/stable rather than only include nightly:
> 
> This seems to work: branch != rel/stable AND branch =~ /^rel\//
> ```
> $ travis-conditions eval "branch != rel/stable AND branch =~ /^rel\//" --data '{"branch": "rel/stable"}'
> false
> $ travis-conditions eval "branch != rel/stable AND branch =~ /^rel\//" --data '{"branch": "rel/nightly"}'
> true
> $ travis-conditions eval "branch != rel/stable AND branch =~ /^rel\//" --data '{"branch": "master"}'
> false
> ```

winder: Additionally, we could use the travis deployments for this rather than tasks. I'm  using the github releases provider for SHD but we could use the script or s3 provider as well:
https://docs.travis-ci.com/user/deployment/script/
https://docs.travis-ci.com/user/deployment/s3/

scripts/create_and_deploy_private_version.sh
```diff
@@ -87,6 +87,12 @@ if [[ "${CHANNEL}" = "" || "${NETWORK}" = "" || "${CONFIGFILE}" = "" || "${TEMPL
     exit 1
 fi
 
+# Don't use environment variable for S3_RELEASE_BUCKET - default to algorand-internal for private deployments
+if [[ ! -z "${S3_RELEASE_BUCKET}" && -z "${BUCKET}" ]]; then
+    echo "Ignoring S3_RELEASE_BUCKET setting - defaulting to algorand-internal.  Use -b to override."
+fi
+S3_RELEASE_BUCKET="${BUCKET:-algorand-internal}"
```
winder: nit: Could you set the `algorand-internal` default at the top of the file when the rest are initialized to empty? I think it will let you simplify the code a bit:

```
if [[ ! -z "${S3_RELEASE_BUCKET}" ]]; then
    echo "Ignoring S3_RELEASE_BUCKET setting - defaulting to algorand-internal.  Use -b to override."
fi

# If we really need this it would always be.
export S3_RELEASE_BUCKET="${BUCKET}"
```

Then just use `-b "${BUCKET}"` at the end.

scripts/create_and_deploy_recipe.sh
```diff
@@ -77,6 +82,12 @@ if [[ -z "${CHANNEL}" || -z "${RECIPEFILE}" || -z "${ROOTDIR}" ]]; then
     exit 1
 fi
 
+# Don't use environment variable for S3_RELEASE_BUCKET - default to algorand-internal for private deployments
+if [[ ! -z "${S3_RELEASE_BUCKET}" && -z "${BUCKET}" ]]; then
+    echo "Ignoring S3_RELEASE_BUCKET setting - defaulting to algorand-internal.  Use -b to override."
+fi
+S3_RELEASE_BUCKET="${BUCKET:-algorand-internal}"
```
winder: Same as above

scripts/deploy_private_version.sh
```diff
@@ -58,11 +58,17 @@ while [ "$1" != "" ]; do
...
+    echo "Syntax: deploy_private_version -c <channel> [ -g <genesis-network> | -f <genesis-file> ] -n <network> [ -b <bucket> ]"
     echo "e.g. deploy_private_version.sh -c TestCatchup -g testnet -n testnetwork"
     exit 1
 fi
 
+# Don't use environment variable for S3_RELEASE_BUCKET - default to algorand-internal for private deployments
+if [[ ! -z "${S3_RELEASE_BUCKET}" && -z "${BUCKET}" ]]; then
+    echo "Ignoring S3_RELEASE_BUCKET setting - defaulting to algorand-internal.  Use -b to override."
+fi
+S3_RELEASE_BUCKET="${BUCKET:-algorand-internal}"
```
winder: Same as above

scripts/travis/release_packages.sh
```diff
@@ -13,10 +13,12 @@ set -e
 CHANNEL=""
 FULLVERSION=""
 S3CMD="s3cmd"
+BUILD_BUCKET=${S3_RELEASE_BUCKET}
+RELEASE_BUCKET="algorand-releases"
```
winder: This is the pattern I like :)

scripts/travis/release_packages.sh
```diff
@@ -38,9 +40,9 @@ function promote_nightly() {
     init_s3cmd
 
     # Rename the _CHANNEL_ and _CHANNEL-VARIANT_ pending files
```
winder: This comment is no longer correct

scripts/travis/release_packages.sh
```diff
@@ -50,9 +52,10 @@ function promote_stable() {
     init_s3cmd
 
     # Copy the _CHANNEL_ pending 'node' files to _CHANNEL-canary_
```
winder: Update comment

scripts/upload_config.sh
```diff
@@ -31,5 +31,5 @@ TARFILE=${TEMPDIR}/config_${CHANNEL}_${FULLVERSION}.tar.gz
 cd $1
 tar -zcf ${TARFILE} * >/dev/null 2>&1
 
-${GOPATH}/bin/updater send -s ${TEMPDIR} -c ${CHANNEL}
+${GOPATH}/bin/updater send -s ${TEMPDIR} -c ${CHANNEL} -b "${S3_RELEASE_BUCKET}"
```
winder: Has this environment variable been validated by another script which calls this one?

util/s3/fileIterator.go
```diff
@@ -26,17 +26,19 @@ import (
...
+	subFolder string
 	next      struct {
 		path string
 		f    *os.File
 	}
 	err error
 }
 
-func makeFileIterator(files []string, bucket string) s3manager.BatchUploadIterator {
+func makeFileIterator(files []string, bucket string, subFolder string) s3manager.BatchUploadIterator {
```
winder: The other function named this `targetFolder`, might as well keep them consistent.

util/s3/s3Helper.go
```diff
@@ -40,65 +56,138 @@ type Helper struct {
...
 		return
 	}
-	creds := credentials.NewStaticCredentials(awsID, awsKey, "")
-	helper, err = makeS3Session(creds, awsBucket)
+	return
+}
+
+func checkCredentialsRequired(action string, bucketName string) (required bool) {
+	required = true
+    if action == downloadAction && bucketName == s3DefaultReleaseBucket {
```
winder: spaces -> tab

util/s3/s3Helper_test.go
```diff
@@ -0,0 +1,182 @@
...
+			} else {
+				os.Setenv("S3_REGION", tt.wantRegion)
+			}
+			if gotRegion := getS3Region(); gotRegion != tt.wantRegion {
+				t.Errorf("getS3Region() = %v, want %v", gotRegion, tt.wantRegion)
+			}
+		})
+	}
+}
+
```
winder: Is it possible to check the session credentials in `gotHelper.session`? Would be nice to check directly that the anonymous credentials work. I guess it's implicitly tested in test5/test9 - maybe give those tests more specific names?

scripts/create_and_deploy_private_version.sh
Karmastic: Initializing prevents the ability to report this warning.

scripts/travis/release_packages.sh
Karmastic: What comment update is needed?

scripts/create_and_deploy_private_version.sh
winder: Given the users of this script I don't think it's a major point, but with the current logic `S3_RELEASE_BUCKET` could be silently overwritten with the default value.

scripts/travis/release_packages.sh
winder: The new comment here would be something like
> # Copy all files for a given channel to the release bucket.

Also, because it's a copy I think that for each new release this logic will copy all preceding releases as well. Maybe we need a second line to delete the files from `BUILD_BUCKET` (or move them to a separate `BUILD_BUCKET/RELEASED/CHANNEL` directory).

scripts/travis/release_packages.sh
winder: It's also copying the files to a new bucket
> Remove pending_ from 'node' files, rename to _CHANNEL-canary_, and copy to the RELEASE_BUCKET

Issue #92, opened 2019-06-25
Add a goal ledger command to dump a block or range of blocks


algoradam:
Right now there's no convenient way to dump and examine a raw block. It'd be nice if there were a command like `goal ledger dumpblock 400` to dump block 400 as base64 encoded msgpack to stdout. Perhaps it could have a `--json` flag to translate it to json, `msgpacktool`-style.

This would be useful for debugging (what's going on in this block header?) and also just a fun way to see what's going on in the blocks.

Vervious:
What's the rationale for keeping it encoded as canonical msgpack?

tsachiherman:
> What's the rationale for keeping it encoded as canonical msgpack?

You mean in the sqlite database ? We keep it encoded so that it would be faster when requested by the catchup. That should also align with the way that we received the block from the network.

algobolson:
#621 implements the algod side of this by adding a local API for fetching raw block bytes

PR #125, opened 2019-07-08
a dumpblock command for goal ledger

zacharyestep:
## Summary

#92 
https://github.com/algorand/go-algorand/issues/92

## Test Plan

Manual tests


cmd/goal/ledger.go
```diff
@@ -17,13 +17,25 @@
 package main
 
 import (
+	"bytes"
+	"encoding/json"
 	"fmt"
-
+	msgpackcore "github.com/algorand/go-algorand/cmd/msgpacktool/core"
```
Karmastic: Not commenting on the merit of this change, but:
1) imports should be sorted properly
2) We do not import from source under ./cmd.  If you want to reuse functionality, it should be broken out into a separate package NOT under ./cmd.

cmd/goal/ledger.go
```diff
@@ -52,3 +64,32 @@ var supplyCmd = &cobra.Command{
 		fmt.Printf("Round: %v\nTotal Money: %v microAlgos\nOnline Money: %v microAlgos\n", response.Round, response.TotalMoney, response.OnlineMoney)
 	},
 }
+
+var blockCmd = &cobra.Command{
+	Use:   "dumpblock <round>",
+	Short: "Show ledger block.",
+	Long:  "Show ledger block. In msgpack format or json format.",
```
Karmastic: Remove periods, especially the one after 'block'.  `Dump` instead of `show` a is better term.
Perhaps:
"Dump block from the ledger"
"Dump a specific block from from the ledger, in msgpack or json format"

cmd/msgpacktool/main.go
```diff
@@ -45,6 +45,7 @@ import (
 	"flag"
 	"fmt"
 	"os"
+	"github.com/algorand/go-algorand/cmd/msgpacktool/core"
```
Karmastic: Add space between groups of imports

zeldovich:
Why do you want to use msgpacktool here?  Why not just print the block with `protocol.EncodeJSON`?

zacharyestep:
Quoting issue --- > Right now there's no convenient way to dump and examine a raw block. It'd be nice if there were a command like goal ledger dumpblock 400 to dump block 400 as base64 encoded msgpack to stdout. Perhaps it could have a --json flag to translate it to json, msgpacktool-style.

This would be useful for debugging (what's going on in this block header?) and also just a fun way to see what's going on in the blocks.

---- End quote --- 
Dont really want to use msgpacktool at all, but it was described in the issue :-) 

V/r,

ZE

zacharyestep:
Hi all, did I put the msgpack internals in the right place here?

zeldovich:
The block you are printing is not a literal copy of the block from the blockchain (instead, it's the REST API model of a block from `daemon/algod/api/spec/v1/model.go`).  So I don't think there's any value printing it in a msgpack encoding -- that just seems to be more likely to lead to confusion about this being the real block.

PR #126, opened 2019-07-08
support for assets

zeldovich:


data/basics/userBalance.go
```diff
@@ -180,6 +203,15 @@ func (u AccountData) KeyDilution(proto config.ConsensusParams) uint64 {
 	return proto.DefaultKeyDilution
 }
 
+// IsZero checks if an AccountData value is the same as its zero value.
+func (ad AccountData) IsZero() bool {
+	if ad.Currencies != nil && len(ad.Currencies) == 0 {
+		ad.Currencies = nil
+	}
+
+	return reflect.DeepEqual(ad, AccountData{})
```
tsachiherman: Can we use a non-reflect way to do that ? It doesn't look like it would too complicated to consider other solutions here.

zeldovich: The main benefit of reflect that I was going for was that it makes it impossible to miss checking a field.

I agree it doesn't really require reflection; just checking that every field is zero.

data/basics/userBalance.go
```diff
@@ -103,6 +105,27 @@ type AccountData struct {
...
+	// added and removed from an account by special transactions.
+	// Each currency bumps the required MinBalance in this account.
+	// An account that created a currency (i.e., ThisCurrencyTotal>0)
+	// must have its own currency in the Currencies map until that
+	// currency is destroyed.
+	//
+	// NOTE: do not modify this value in-place in existing AccountData
+	// structs; allocate a copy and modify that instead.  AccountData
+	// is expected to have copy-by-value semantics.
+	Currencies map[Address]uint64 `codec:"cur"`
```
tsachiherman: The key is the CurrencyID, right ? can you add this to the field description ?

ledger/accountdb.go
```diff
@@ -230,7 +230,7 @@ func accountsNewRound(tx *sql.Tx, rnd basics.Round, updates map[basics.Address]a
 	defer replaceStmt.Close()
 
 	for addr, data := range updates {
-		if (data.new == basics.AccountData{}) {
+		if data.new.IsZero() {
```
tsachiherman: Yes, this is much better !

derbear:
We can get rid of the first transaction (where the user sends 0 Xs to herself to allocate space in the Currencies map) by requiring that any transaction which allocates new space in the map is cosigned by the transaction receiver.

zeldovich:
Yeah, in principle that should suffice.  The main challenge there is that we currently don't have a convenient way to attach multiple signatures to the same transaction (except for multisig, but then we don't have a way of attaching multiple multisig signatures)..

data/ledger.go
```diff
@@ -226,6 +226,9 @@ func (l *Ledger) BalanceAndStatus(addr basics.Address) (money basics.MicroAlgos,
 	}
 	moneyWithoutPendingRewards = dataWithoutRewards.MicroAlgos
 
+	totalCurrency = dataWithoutRewards.ThisCurrencyTotal
+	currencies = dataWithoutRewards.Currencies
```
tsachiherman: You're returning a map ( i.e. a pointer ); it's likely to be ok in most cases, as long as we use it for read-only.
If this is not the case, we'll need to either copy that or to lock that.

zeldovich: Sure, might as well be safe and clone the map.

The intent was that the caller is read-only, which is true of current callers, and elsewhere in the code we explicitly copy the map when modifying -- like clone() in data/transactions/currency.go -- but seems like a reasonable idea to guard against buggy callers here.

tsachiherman: thanks!

daemon/algod/api/server/v1/handlers/handlers.go
```diff
@@ -314,6 +314,14 @@ func AccountInformation(ctx lib.ReqContext, w http.ResponseWriter, r *http.Reque
 		return
 	}
 
+	var currenciesStr map[string]uint64
+	if len(currencies) > 0 {
+		currenciesStr = make(map[string]uint64)
```
tsachiherman: nit : initialize with the len(currencies)

data/pools/transactionPool.go
```diff
@@ -244,7 +244,7 @@ func (pool *TransactionPool) computeDeductions(t transactions.SignedTxn) (accoun
 	if err != nil {
 		return algosPendingSpend, fmt.Errorf("TransactionPool.Remember: failed to compute balance - %v", err)
 	}
-	senderBalance, _, _, _, _, err := pool.ledger.BalanceAndStatus(t.Txn.Src())
+	senderBalance, _, _, _, _, _, _, err := pool.ledger.BalanceAndStatus(t.Txn.Src())
```
tsachiherman: looks like we could refactor the BalanceAndStatus; all these ignored returned values aren't super-readable, and are a good indication of a mismatching interface.

tsachiherman: ( note that it's not really related to this PR; issue was there before )

libgoal/transactions.go
```diff
@@ -220,3 +220,64 @@ func (c *Client) MakeUnsignedGoOfflineTx(address string, round, txValidRounds, f
...
+		copy(genHash[:], params.GenesisHash)
+		tx.GenesisHash = genHash
+	}
+
+	// Default to the suggested fee, if the caller didn't supply it
+	// Fee is tricky, should taken care last. We encode the final
+	// transaction to get the size post signing and encoding.
+	// Then, we multiply it by the suggested fee per byte.
+	if fee == 0 {
+		tx.Fee = basics.MulAIntSaturate(basics.MicroAlgos{Raw: params.Fee}, tx.EstimateEncodedSize())
```
tsachiherman: how does msgpack encodes the fee ? I'm asking since if it's not a fixed byte-length, we might need to adjust the `tx.EstimateEncodedSize` according to the fee.

data/pools/transactionPool.go
derbear: We're planning to deprecate `data/ledger` at some point. Hopefully at that point we can fix up this interface.

libgoal/transactions.go
zeldovich: The encoding is variable-length: it's zero extra bytes for a zero value, map key length + 1 byte for a value up to 127, map key length + 2 bytes for a value up to 255, map key length + 3 bytes for a value up to 65535, map key length + 5 bytes for up to 2^32-2, and otherwise map key length + 9 bytes.  There might also be an extra byte added if the entire transaction object grows from 15 to 16 non-zero-valued keys (though I think that's unlikely).

That's one reason why the method is called `EstimateEncodedSize`.

data/transactions/currency.go
```diff
@@ -0,0 +1,360 @@
...
+
+			params, err := getParams(balances, ct.XferCurrency)
+			if err != nil {
+				return err
+			}
+
+			sndHolding.Frozen = params.DefaultFrozen
+		}
+
+		if sndHolding.Amount < ct.CurrencyAmount {
```
yossigi: nit: might be simpler to use OSub (maybe of a new type)

data/transactions/currency.go
```diff
@@ -0,0 +1,360 @@
...
+		rcvHolding, ok := rcv.Currencies[ct.XferCurrency]
+		if !ok {
+			return fmt.Errorf("currency not present in receiver account")
+		}
+
+		if ct.CurrencyAmount > 0 && rcvHolding.Frozen {
+			return fmt.Errorf("currency frozen in recipient")
+		}
+
+		rcvHolding.Amount += ct.CurrencyAmount
```
yossigi: there's no risk of overflow here, but might be good to have OAdd as a sanity check

data/transactions/currency.go
```diff
@@ -0,0 +1,360 @@
...
+		}
+
+		record.CurrencyParams[cc.ConfigCurrency.Index] = params
+	}
+
+	return balances.Put(record)
+}
+
+func (ct CurrencyTransferTxnFields) apply(header Header, balances Balances, spec SpecialAddresses, ad *ApplyData) error {
+	// Default to sending from the transaction sender's account.
```
yossigi: Maybe break this function into smaller pieces of code?
It took me about 45 minutes to read through it and figure out all the cases it covers.

cmd/goal/currency.go
```diff
@@ -0,0 +1,555 @@
...
+		cmd.HelpFunc()(cmd, args)
+	},
+}
+
+func lookupCurrencyID(cmd *cobra.Command, creator string, client libgoal.Client) {
+	if cmd.Flags().Changed("currencyid") && cmd.Flags().Changed("currency") {
+		reportErrorf("Only one of -currencyid and -currency can be specified")
+	}
+
+	if cmd.Flags().Changed("currencyid") {
```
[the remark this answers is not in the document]
zeldovich: In case a creator makes two currencies with the same unit name (e.g., multiple currencies where the unit name isn't specified at all, and is thus blank).  Good to have this option because currencyid is the thing that's really needed to identify a currency, and the `-currency` option does an inverse lookup on the unit name as a convenience.

daemon/algod/api/server/v1/handlers/handlers.go
```diff
@@ -119,17 +129,81 @@ func keyregTxEncode(tx transactions.Transaction, ad transactions.ApplyData) v1.T
...
+		Params: currencyParams(tx.CurrencyConfigTxnFields.ConfigCurrency.Creator,
+			tx.CurrencyConfigTxnFields.CurrencyParams),
+	}
+
+	return v1.Transaction{
+		CurrencyConfig: &config,
+	}
+}
+
+func currencyTransferTxEncode(tx transactions.Transaction, ad transactions.ApplyData) v1.Transaction {
```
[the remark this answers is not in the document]
zeldovich: Yeah, mostly for symmetry with the other encoding functions.

Though this made me realize that my code was implicitly applying pending rewards when accessing accounts in a currency transaction.  This wasn't a problem in terms of leaking rewards, but it meant that `ApplyData` was incomplete, so clients that were trying to track all of the places where rewards were being "applied" would get out of sync.  Fixed and pushed.

data/transactions/currency.go
zeldovich: Sure, done.

data/transactions/currency.go
zeldovich: Sure, done.

data/transactions/currency.go
zeldovich: I tried to refactor it a bit (see pushed code).

data/transactions/payment.go
```diff
@@ -84,11 +84,16 @@ func (payment PaymentTxnFields) apply(header Header, balances Balances, spec Spe
...
 
 		// Confirm that we have no balance left
-		rec, err = balances.Get(header.Sender)
+		rec, err = balances.Get(header.Sender, true)
 		if !rec.AccountData.MicroAlgos.IsZero() {
 			return fmt.Errorf("balance %d still not zero after CloseRemainderTo", rec.AccountData.MicroAlgos.Raw)
 		}
 
+		// Confirm that there is no asset-related state in the account
+		if len(rec.Assets) > 0 {
```
justicz: This appears to be sufficient since I don't think there's a way to make `len(rec.AssetParams) > 0` while `len(rec.Assets) == 0`, but it might be worth checking that `rec.AssetParams` is empty as well.

zeldovich: Yeah, that was my thinking too (should be impossible), but happy to add a check just in case.

zeldovich:
Thanks!

PR #127, opened 2019-07-08
remove swagger-dependent build (#119)

Karmastic:
temporary stop gap while we investigate how to make sure swagger does not break our build.

This is an infrastructure fix and should not be used to generate a new release.  Note the resultant binary will differ in that it won't have a swagger.json file to serve.


Karmastic:
This is specifically to unblock 3rd parties that want to build rel/stable releases directly, which requires a working unmodified enlistment.

PR #128, opened 2019-07-08
[GOAL2-815] Improve logging for "bad-handshake"

tsachiherman:
## Summary

Existing logic for "bad-handshake" abruptly disconnect when we have headers mismatch. The client is not informed of the disconnection reason. This behavior is neither HTTP compliant nor it allows us to log these use cases correctly.

## Test Plan

I have manually tested the changed using a deployed network.


network/wsNetwork.go
```diff
@@ -736,24 +736,57 @@ func (wn *WebsocketNetwork) getForwardedConnectionAddress(header http.Header) (i
...
 	otherPublicAddr = ""
 	otherVersion := header.Get(ProtocolVersionHeader)
 	if otherVersion != ProtocolVersion {
-		wn.log.Warnf("new peer %#v version mismatch, mine=%#v theirs=%#v, headers %#v", addr, ProtocolVersion, otherVersion, header)
-		return
-	}
-	otherGenesisID := header.Get(GenesisHeader)
-	if len(otherGenesisID) > 0 && wn.GenesisID != otherGenesisID {
-		wn.log.Warnf("new peer %#v genesis mismatch, mine=%#v theirs=%#v, headers %#v", addr, wn.GenesisID, otherGenesisID, header)
+		wn.log.Warnf("new peer %s version mismatch, mine=%s theirs=%s, headers %#v", addr, ProtocolVersion, otherVersion, header)
```
Karmastic: I think this should be an Info level log entry.  This is going to happen when someone doesn't upgrade their node in a timely fashion.

tsachiherman: make sense; I'll update the code.

network/wsNetwork.go
```diff
@@ -843,15 +875,39 @@ func (wn *WebsocketNetwork) ServeHTTP(response http.ResponseWriter, request *htt
 		return
 	}
 
+	// check to see that the genesisID in the request URI is valid and matches the supported one.
```
Karmastic: This section looks like it belongs in wn.checkHeaders().  These won't get tracked in networkConnectionsDroppedTotal metric.

network/wsNetwork.go
```diff
@@ -843,15 +875,39 @@ func (wn *WebsocketNetwork) ServeHTTP(response http.ResponseWriter, request *htt
...
+		}
+		n, err := response.Write([]byte(checkErr.Error()))
+		if err != nil {
+			wn.log.Warnf("ws failed to write response '%s' : n = %d err = %v", checkErr.Error(), n, err)
+		}
 		return
 	}
 	requestHeader := make(http.Header)
 	wn.setHeaders(requestHeader)
+	requestHeader.Set(GenesisHeader, wn.GenesisID)
```
Karmastic: This is already done in wn.setHeaders()

network/wsNetwork.go
tsachiherman: The networkConnectionsDroppedTotal metric is a good point. However, I disagree that it should be tested in the checkHeaders; it's being tested differently on the client and on the server, which is why it was removed from the checkHeaders function.

network/wsNetwork.go
```diff
@@ -1488,14 +1544,44 @@ func (wn *WebsocketNetwork) tryConnect(addr, gossipAddr string) {
...
 	conn, response, err := websocketDialer.DialContext(wn.ctx, gossipAddr, requestHeader)
 	if err != nil {
-		wn.log.Warnf("ws connect(%s) fail: %s", gossipAddr, err)
+		if err == websocket.ErrBadHandshake {
+			// reading here from ioutil is safe only because it came from DialContext above, which alredy finsihed reading all the data from the network
+			// and placed it all in a ioutil.NopCloser reader.
+			bodyBytes, _ := ioutil.ReadAll(response.Body)
+			errString := string(bodyBytes)
+
+			// we're gurenteed to have a valid response object.
```
Karmastic: sp: gurenteed

network/wsNetwork.go
tsachiherman: it was removed from the setHeaders, since the setHeaders only sets the common headers.
This header is used only on the response now to avoid conflicting with the genesis-id passed as past of the url.

network/wsNetwork.go
```diff
@@ -1488,14 +1544,44 @@ func (wn *WebsocketNetwork) tryConnect(addr, gossipAddr string) {
...
+	// checking the headers here is abit redundent; the server has already verified that the headers match. But we will need this in the future -
+	// once our server would support multiple protocols, we would need to verify here that we use the correct protocol, out of the "proposed" protocols we have provided in the
+	// request headers.
+	otherTelemetryGUID, _, _, headersErr := wn.checkHeaders(response.Header, gossipAddr, nil)
+	if headersErr != nil {
 		return
 	}
+	otherGenesisID := response.Header.Get(GenesisHeader)
+	if wn.GenesisID != otherGenesisID {
+		if len(otherGenesisID) > 0 {
```
Karmastic: otherGenesisID != ""

network/wsNetwork.go
```diff
@@ -1488,14 +1544,44 @@ func (wn *WebsocketNetwork) tryConnect(addr, gossipAddr string) {
 	requestHeader.Set(InstanceNameHeader, myInstanceName)
 	conn, response, err := websocketDialer.DialContext(wn.ctx, gossipAddr, requestHeader)
 	if err != nil {
-		wn.log.Warnf("ws connect(%s) fail: %s", gossipAddr, err)
+		if err == websocket.ErrBadHandshake {
+			// reading here from ioutil is safe only because it came from DialContext above, which alredy finsihed reading all the data from the network
+			// and placed it all in a ioutil.NopCloser reader.
+			bodyBytes, _ := ioutil.ReadAll(response.Body)
```
Karmastic: Do we really want to readall here?  We should limit its length.

tsachiherman: I had the same thought; but it doesn't matter. We don't read anything from the network here. It's already in a memory buffer at that point.

network/wsNetwork.go
```diff
@@ -1488,14 +1544,44 @@ func (wn *WebsocketNetwork) tryConnect(addr, gossipAddr string) {
...
 		return
 	}
 	// no need to test the response.StatusCode since we know it's going to be http.StatusSwitchingProtocols, as it's already being tested inside websocketDialer.DialContext.
-	ok, otherTelemetryGUID, _, _ := wn.checkHeaders(response.Header, gossipAddr, nil)
-	if !ok {
+	// checking the headers here is abit redundent; the server has already verified that the headers match. But we will need this in the future -
+	// once our server would support multiple protocols, we would need to verify here that we use the correct protocol, out of the "proposed" protocols we have provided in the
+	// request headers.
+	otherTelemetryGUID, _, _, headersErr := wn.checkHeaders(response.Header, gossipAddr, nil)
+	if headersErr != nil {
```
Karmastic: We're silently ignoring this error - should we do something more with it?

network/wsNetwork.go
```diff
@@ -1488,14 +1544,44 @@ func (wn *WebsocketNetwork) tryConnect(addr, gossipAddr string) {
...
 		return
 	}
+	otherGenesisID := response.Header.Get(GenesisHeader)
```
Karmastic: Any reason this isn't inside checkHeaders()?

tsachiherman: Yes; it's being checked differently on the client vs. server.
On the server, we check the URI; on the client we check the response header.

network/wsNetwork.go
tsachiherman: What would you like to do with it ? errors were already logged.

tsachiherman: we can add this to the do-not-call list. but that's a bit beyond the scope of this change.

network/wsNetwork.go
Karmastic: ok thanks - I didn't pick up that change.

network/wsNetwork.go
Karmastic: it matters in as much as it's also being logged to the log file and sent to telemetry.

tsachiherman: I see; so you don't truly care about the reading. You're concern is that we could end up logging large quantities of information. I'll limit that to 128 characters.

network/wsNetwork.go
Karmastic: Just feels wrong.  I guess it's fine.

network/wsNetwork.go
Karmastic: Not required but maybe having context-specific check functions would be clearer (eg checkOugoingHeaders() and checkIncomingHeaders())

tsachiherman: Agree; the existing code need to be slightly refactored to make it bit more structured.

tsachiherman:
> Looks like lots more logging. My main concern with this is - how many of these Warn calls are going to be common? Warnings generate telemetry events with log history.

Let me review the exact use cases where we add new warning. Adding warnings wasn't my goal here. The goal was to provide precise fail logging. ( i.e. replacing some warnings with others )

tsachiherman:
> > Looks like lots more logging. My main concern with this is - how many of these Warn calls are going to be common? Warnings generate telemetry events with log history.
> 
> Let me review the exact use cases where we add new warning. Adding warnings wasn't my goal here. The goal was to provide precise fail logging. ( i.e. replacing some warnings with others )

I don't see any case where we're creating more warnings; we're just replacing existing warnings with more detailed ones on the tryConnect side.

network/wsNetwork.go
```diff
@@ -1488,24 +1564,49 @@ func (wn *WebsocketNetwork) tryConnect(addr, gossipAddr string) {
...
+	// once our server would support multiple protocols, we would need to verify here that we use the correct protocol, out of the "proposed" protocols we have provided in the
+	// request headers.
+	if !wn.checkOutgoingConnectionVariables(response.Header, gossipAddr) {
+		// The error was already logged, so no need to log again.
 		return
 	}
+
 	peer := &wsPeer{wsPeerCore: wsPeerCore{net: wn, rootURL: addr}, conn: conn, outgoing: true, incomingMsgFilter: wn.incomingMsgFilter}
-	peer.TelemetryGUID = otherTelemetryGUID
+	peer.TelemetryGUID, _, _ = getCommonHeaders(response.Header)
```
algobolson: should we just get the one header instead of getting 3 and throwing away 2?

network/wsNetwork.go
```diff
@@ -1488,24 +1564,49 @@ func (wn *WebsocketNetwork) tryConnect(addr, gossipAddr string) {
...
 		return
 	}
+
 	// no need to test the response.StatusCode since we know it's going to be http.StatusSwitchingProtocols, as it's already being tested inside websocketDialer.DialContext.
-	ok, otherTelemetryGUID, _, _ := wn.checkHeaders(response.Header, gossipAddr, nil)
-	if !ok {
+	// checking the headers here is abit redundent; the server has already verified that the headers match. But we will need this in the future -
+	// once our server would support multiple protocols, we would need to verify here that we use the correct protocol, out of the "proposed" protocols we have provided in the
+	// request headers.
+	if !wn.checkOutgoingConnectionVariables(response.Header, gossipAddr) {
```
algobolson: the name of this function confused me when I first read it because checking 'outgoing' variables, the stuff I send, doesn't make sense. I get it now. how does 'check server response' or 'check server response variables' sound to you?

PR #130, opened 2019-07-09
[GOAL2-844] Re-enable go-swagger in build

Vervious:
We previously removed swagger build step due to change in swagger upstream master. Now, we pin build to specific swagger release version using brew and debian package. Reenables swagger build step.

I preferred this over go/vendoring because we use swagger as a build tool, not as an import dependency.

### DIFF
--- a/scripts/configure_dev.sh
+++ b/scripts/configure_dev.sh
@@ -16,6 +16,9 @@ function install_or_upgrade {
 if [ "${OS}" = "linux" ]; then
     sudo apt-get update
     sudo apt-get -y install libboost-all-dev expect jq
+    echo "deb [trusted=yes] https://dl.bintray.com/go-swagger/goswagger-debian ubuntu main" | sudo tee -a /etc/apt/sources.list
+    sudo apt-get update
+    sudo apt-get -y install swagger
 elif [ "${OS}" = "darwin" ]; then
     brew update
     brew tap caskroom/cask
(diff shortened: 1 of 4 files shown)


scripts/configure_dev.sh
```diff
@@ -16,6 +16,9 @@ function install_or_upgrade {
 if [ "${OS}" = "linux" ]; then
     sudo apt-get update
     sudo apt-get -y install libboost-all-dev expect jq
+    echo "deb [trusted=yes] https://dl.bintray.com/go-swagger/goswagger-debian ubuntu main" | sudo tee -a /etc/apt/sources.list
```
Karmastic: This is going to append every time someone runs configure_dev.sh.  Can we please ensure it's idempotent?

Vervious:
A note that existing dev environments may need a `go clean -i github.com/go-swagger/go-swagger/cmd/swagger` and then `cd go-algorand; sh ./scripts/configure_dev.sh`

scripts/configure_dev.sh
Vervious: good point

Karmastic: And do this as part of the above block - update sources.list if necessary, then update, then one install set.

Vervious:
`make deps` and `configure_dev.sh` now automatically run `go clean -i github.com/go-swagger/go-swagger/cmd/swagger` to remove the "broken" version, in case it is on the user's `PATH`. Devs still need to run `configure_dev.sh` again to build, after this PR is merged.

PR #131, opened 2019-07-09
[GOAL2-762] Add goal node create, for quickly creating nodes in desired data directories

EvanJRichard:
## Summary

This PR proposes to add command `goal node create`, which will place the correct genesis json in a desired extant data directory, along with a config file if required by the passed options. `goal node create` supports options such as `--archival` and `--hosted`. 

## Test Plan

Used `goal node create` to create nodes myself.


cmd/goal/node.go
```diff
@@ -374,3 +390,89 @@ var waitCmd = &cobra.Command{
...
+		if err != nil {
+			reportErrorf(errorNodeCreation, err)
+		}
+		destPath := filepath.Join(newNodeDestination, "genesis.json")
+		err = ioutil.WriteFile(destPath, genesisInput, 0666)
+		if err != nil {
+			reportErrorf(errorNodeCreation, err)
+		}
+
+		var configBuilder strings.Builder
```
Karmastic: Populate a Local config struct from the inputs and call "codec.SaveNonDefaultValuesToFile(filename, config, config.GetDefaultLocal(), nil, true)

cmd/goal/node.go
```diff
@@ -374,3 +390,89 @@ var waitCmd = &cobra.Command{
...
+			}
+		}
+		if correctPath == "" {
+			reportErrorf(errorNodeCreation, "no genesis file found")
+		}
+		genesisInput, err := ioutil.ReadFile(correctPath)
+		if err != nil {
+			reportErrorf(errorNodeCreation, err)
+		}
+		destPath := filepath.Join(newNodeDestination, "genesis.json")
```
Karmastic: You should check to see if the destination folder exists and fail if it does.  Then ensure you can create it before trying to save the genesis.json file there.

cmd/goal/node.go
```diff
@@ -64,8 +73,15 @@ func init() {
 	startCmd.Flags().StringVarP(&telemetryOverride, "telemetry", "t", "", `Enable telemetry if supported (Use "true", "false", "0" or "1")`)
 	restartCmd.Flags().StringVarP(&telemetryOverride, "telemetry", "t", "", `Enable telemetry if supported (Use "true", "false", "0" or "1")`)
 	pendingTxnsCmd.Flags().Uint64VarP(&maxPendingTransactions, "maxPendingTxn", "m", 0, "Cap the number of txns to fetch")
-
 	waitCmd.Flags().Uint32VarP(&waitSec, "waittime", "w", 5, "Time (in seconds) to wait for node to make progress")
+	createCmd.Flags().StringVarP(&newNodeNetwork, "network", "n", "testnet", "Network the new node should point to")
```
Karmastic: don't default to anything here.

cmd/goal/node.go
```diff
@@ -64,8 +73,15 @@ func init() {
...
+	createCmd.Flags().StringVar(&newNodeDestination, "destination", "", "Destination path for the new node")
```
Karmastic: Initialize these defaults to the actual current defaults (use config.GetDefaultLocal() settings)

cmd/goal/node.go
```diff
@@ -64,8 +73,15 @@ func init() {
...
+	createCmd.Flags().BoolVarP(&newNodeArchival, "archival", "a", false, "Make the new node archival, storing all blocks")
+	createCmd.Flags().BoolVarP(&runUnderHost, "hosted", "H", false, "Configure the new node to run hosted by algoh")
+	createCmd.Flags().BoolVarP(&newNodeIndexer, "indexer", "i", false, "Configure the new node to enable the indexer feature (implies --archival)")
+	createCmd.Flags().StringVarP(&peerDial, "peer", "p", "", "Peer address to dial for initial connection")
```
Karmastic: This option does not apply to 'create'

cmd/goal/node.go
```diff
@@ -374,3 +390,75 @@ var waitCmd = &cobra.Command{
...
+			reportErrorf(errorNodeCreation, err)
+		}
+		err = ioutil.WriteFile(destPath, genesisInput, 0666)
+		if err != nil {
+			reportErrorf(errorNodeCreation, err)
+		}
+
+		// build config and save to destination
+		localConfig := config.GetDefaultLocal()
+		localConfig.Archival = newNodeArchival
```
Karmastic: Archival == newNodeArchival || relay != "" || newNodeIndexer

cmd/goal/node.go
```diff
@@ -64,8 +73,15 @@ func init() {
...
-
 	waitCmd.Flags().Uint32VarP(&waitSec, "waittime", "w", 5, "Time (in seconds) to wait for node to make progress")
+	createCmd.Flags().StringVarP(&newNodeNetwork, "network", "n", "testnet", "Network the new node should point to")
+	createCmd.Flags().StringVar(&newNodeDestination, "destination", "", "Destination path for the new node")
+	createCmd.Flags().BoolVarP(&newNodeArchival, "archival", "a", false, "Make the new node archival, storing all blocks")
+	createCmd.Flags().BoolVarP(&runUnderHost, "hosted", "H", false, "Configure the new node to run hosted by algoh")
+	createCmd.Flags().BoolVarP(&newNodeIndexer, "indexer", "i", false, "Configure the new node to enable the indexer feature (implies --archival)")
+	createCmd.Flags().StringVarP(&peerDial, "peer", "p", "", "Peer address to dial for initial connection")
+	createCmd.Flags().StringVarP(&listenIP, "listen", "l", "", "REST API Endpoint (defaults to \"127.0.0.1:0\")")
+	createCmd.MarkFlagRequired("destination")
```
Karmastic: make network required as well

cmd/goal/node.go
```diff
@@ -74,14 +75,16 @@ func init() {
 	restartCmd.Flags().StringVarP(&telemetryOverride, "telemetry", "t", "", `Enable telemetry if supported (Use "true", "false", "0" or "1")`)
 	pendingTxnsCmd.Flags().Uint64VarP(&maxPendingTransactions, "maxPendingTxn", "m", 0, "Cap the number of txns to fetch")
 	waitCmd.Flags().Uint32VarP(&waitSec, "waittime", "w", 5, "Time (in seconds) to wait for node to make progress")
-	createCmd.Flags().StringVarP(&newNodeNetwork, "network", "n", "testnet", "Network the new node should point to")
+	createCmd.Flags().StringVar(&newNodeNetwork, "network", "testnet", "Network the new node should point to")
```
Karmastic: Do not default to any network!

cmd/goal/node.go
```diff
@@ -74,14 +75,16 @@ func init() {
...
-	createCmd.Flags().BoolVarP(&runUnderHost, "hosted", "H", false, "Configure the new node to run hosted by algoh")
-	createCmd.Flags().BoolVarP(&newNodeIndexer, "indexer", "i", false, "Configure the new node to enable the indexer feature (implies --archival)")
-	createCmd.Flags().StringVarP(&peerDial, "peer", "p", "", "Peer address to dial for initial connection")
-	createCmd.Flags().StringVarP(&listenIP, "listen", "l", "", "REST API Endpoint (defaults to \"127.0.0.1:0\")")
+	localDefaults := config.GetDefaultLocal()
+	createCmd.Flags().BoolVarP(&newNodeArchival, "archival", "a", localDefaults.Archival, "Make the new node archival, storing all blocks")
+	createCmd.Flags().BoolVarP(&runUnderHost, "hosted", "H", localDefaults.RunHosted, "Configure the new node to run hosted by algoh")
+	createCmd.Flags().BoolVarP(&newNodeIndexer, "indexer", "i", localDefaults.IsIndexerActive, "Configure the new node to enable the indexer feature (implies --archival)")
+	createCmd.Flags().StringVar(&newNodeRelay, "relay", localDefaults.NetAddress, "IP address that will be used for incoming connections as a relay")
```
Karmastic: Configure as a relay with specified listening address (NetAddress)

cmd/goal/node.go
```diff
@@ -446,11 +449,11 @@ var createCmd = &cobra.Command{
 
 		// build config and save to destination
 		localConfig := config.GetDefaultLocal()
-		localConfig.Archival = newNodeArchival
+		localConfig.Archival = newNodeArchival || newNodeRelay != "" || newNodeIndexer
 		localConfig.IsIndexerActive = newNodeIndexer
 		localConfig.RunHosted = runUnderHost
-		if peerDial != "" {
-			localConfig.NetAddress = peerDial
+		if newNodeRelay != "" {
```
Karmastic: Sorry - one more thing here.  We should probably validate the listenIP and newNodeRelay IP addresses.  They should be run through an ip parser to ensure they're valid IPv4 and/or IPv6 addresses and must specify a port #.  This is bonus credit though, so if it looks too complicated we can punt for now.  If we were Microsoft and this was going to 1e7 people, I would be more careful.

Vervious:
Looks good to me, but I think we want @Karmastic 's final review

cmd/goal/messages.go
```diff
@@ -58,11 +58,13 @@ const (
 	infoTryingToStopNode             = "Trying to stop the node..."
 	infoNodeSuccessfullyStopped      = "The node was successfully stopped."
 	infoNodeStatus                   = "Last committed block: %d\nTime since last block: %s\nSync Time: %s\nLast consensus protocol: %s\nNext consensus protocol: %s\nRound for next consensus protocol: %d\nNext consensus protocol supported: %v"
+	warnNodeCreationIPFailure        = "Parsing passed IP %v failed: need a valid IPv4 or IPv6 address with a specified port number. Acting as though IP was not supplied."
```
Karmastic: If parsing fails, don't act as no IP specified.  Error out.

cmd/goal/node.go
```diff
@@ -374,3 +394,97 @@ var waitCmd = &cobra.Command{
...
+
+var createCmd = &cobra.Command{
+	Use:   "create",
+	Short: "create a node at the desired data directory for the desired network",
+	Long:  "create a node at the desired data directory for the desired network",
+	Args:  validateNoPosArgsFn,
+	Run: func(cmd *cobra.Command, _ []string) {
+
+		// validate network input
+		validNetworks := map[string]bool{"mainnet": true, "testnet": true, "devnet": true}
```
Karmastic: Fine for now - but realized this should really allow anything and just fail if we don't find the genesis.json file for that network.

cmd/goal/node.go
```diff
@@ -374,3 +394,97 @@ var waitCmd = &cobra.Command{
...
+			reportErrorf(errorNodeCreation, "destination folder already exists")
+		}
+		destPath := filepath.Join(newNodeDestination, "genesis.json")
+		err = os.MkdirAll(newNodeDestination, 0666)
+		if err != nil {
+			reportErrorf(errorNodeCreation, "could not create destination folder")
+		}
+
+		// copy genesis block to destination
+		genesisInput, err := ioutil.ReadFile(correctPath)
```
Karmastic: io.CopyFile()

cmd/goal/node.go
```diff
@@ -374,3 +394,97 @@ var waitCmd = &cobra.Command{
...
+			reportErrorf(errorNodeCreation, err)
+		}
+
+		// build config and save to destination
+		localConfig := config.GetDefaultLocal()
+		localConfig.Archival = newNodeArchival || newNodeRelay != "" || newNodeIndexer
+		localConfig.IsIndexerActive = newNodeIndexer
+		localConfig.RunHosted = runUnderHost
+		if newNodeRelay != "" {
+			if isValidIP(newNodeRelay) {
```
Karmastic: Validate this before creating destination directory so you can error out without any side-effects

cmd/goal/node.go
```diff
@@ -374,3 +394,97 @@ var waitCmd = &cobra.Command{
...
+				localConfig.NetAddress = newNodeRelay
+			} else {
+				reportWarnf(warnNodeCreationIPFailure, newNodeRelay)
+			}
+		}
+		if listenIP != "" {
+			if isValidIP(listenIP) {
```
Karmastic: Same as above.

PR #132, opened 2019-07-09
goal: factor out common logic in txn creation

zeldovich:
Define shared functions that wait for transaction commit and that write
out a transaction into a file.

### DIFF
--- a/cmd/goal/clerk.go
+++ b/cmd/goal/clerk.go
@@ -26,6 +26,7 @@ import (
 	"github.com/algorand/go-algorand/crypto"
 	"github.com/algorand/go-algorand/data/basics"
 	"github.com/algorand/go-algorand/data/transactions"
+	"github.com/algorand/go-algorand/libgoal"
 	"github.com/algorand/go-algorand/protocol"
 
 	"github.com/spf13/cobra"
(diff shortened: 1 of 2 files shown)


cmd/goal/clerk.go
```diff
@@ -97,6 +98,69 @@ var clerkCmd = &cobra.Command{
...
+			reportErrorf(errorRequestFail, err)
+		}
+
+		if txn.ConfirmedRound > 0 {
+			reportInfof(infoTxCommitted, txid, txn.ConfirmedRound)
+			break
+		}
+
+		if txn.PoolError != "" {
+			reportErrorf(txPoolError, txid, txn.PoolError)
```
Karmastic: You changed the contact of changeAccountOnlineStatus() -- you need to return the error and propagate it.  generateAndRegisterPartKey() for example needs this to clean up in case of error.

cmd/goal/clerk.go
```diff
@@ -97,6 +98,69 @@ var clerkCmd = &cobra.Command{
...
+
+func writeTxnToFile(client libgoal.Client, dataDir string, walletName string, tx transactions.Transaction, filename string) {
+	var err error
+	var stxn transactions.SignedTxn
+	if sign {
+		// Sign the transaction
+		wh, pw := ensureWalletHandleMaybePassword(dataDir, walletName, true)
+		stxn, err = client.SignTransactionWithWallet(wh, pw, tx)
+		if err != nil {
+			reportErrorf(errorConstructingTX, err)
```
Karmastic: Same as above - return the error - don't reportErrorf() here.

cmd/goal/clerk.go
```diff
@@ -97,6 +98,69 @@ var clerkCmd = &cobra.Command{
...
+		if err != nil {
+			reportErrorf(errorRequestFail, err)
+		}
+	}
+}
+
+func writeTxnToFile(client libgoal.Client, dataDir string, walletName string, tx transactions.Transaction, filename string) {
+	var err error
+	var stxn transactions.SignedTxn
+	if sign {
```
Karmastic: helper functions like this are easier to follow if you pass in the globals.  can you pass 'sign' in as an arg?

cmd/goal/clerk.go
```diff
@@ -97,6 +98,69 @@ var clerkCmd = &cobra.Command{
 	},
 }
 
+func waitForCommit(client libgoal.Client, txid string) {
+	if noWaitAfterSend {
```
Karmastic: helper functions like this are easier to follow if you pass in the globals.  can you pass 'noWaitAfterSend' in as an arg?

cmd/goal/clerk.go
zeldovich: Oops, missed this difference.  Thanks!

PR #134, opened 2019-07-09
[GOAL2-613] Fix timesincelastround reset

egieseke:
## Summary

Updated to initialize node.lastRoundTimestampto the current time at node startup.  This will allow detecting a stall on startup.  

## Test Plan

Run unit and integration tests.


tsachiherman:
I think this is the wrong fix; The existing code was already doing the right thing.
I think that we should not have reported the TimeSinceLastRound in this case since we don't know what it's value. Currently we're reporting zero, which is obviously incorrect.

Given that it's being used in the goal/api, maybe @rotemh can give some feedback here ? ( adding him to the review )

tsachiherman:
see in -
```func nodeStatus(node node.Full) (res NodeStatus, err error) {
	stat, err := node.Status()
	if err != nil {
		return NodeStatus{}, err
	}

	return NodeStatus{
		LastRound:            uint64(stat.LastRound),
		LastVersion:          string(stat.LastVersion),
		NextVersion:          string(stat.NextVersion),
		NextVersionRound:     uint64(stat.NextVersionRound),
		NextVersionSupported: stat.NextVersionSupported,
		TimeSinceLastRound:   stat.TimeSinceLastRound().Nanoseconds(),
		CatchupTime:          stat.CatchupTime.Nanoseconds(),
	}, nil
}
```
and
```
func makeStatusString(stat models.NodeStatus) string {
	lastRoundTime := fmt.Sprintf("%.1fs", time.Duration(stat.TimeSinceLastRound).Seconds())
	catchupTime := fmt.Sprintf("%.1fs", time.Duration(stat.CatchupTime).Seconds())
	return fmt.Sprintf(infoNodeStatus, stat.LastRound, lastRoundTime, catchupTime, stat.LastVersion, stat.NextVersion, stat.NextVersionRound, stat.NextVersionSupported)
}
```

rotemh:
I'm with Tsachi here. While both `0` and `time.Now()` providing the wrong solution, `0` has the benefit of being "obviously wrong". As Tsachi said, since we really don't know the time, I don't think we have a good solution here. For better UI in goal (API i'd leave as 0 and make clear in the documentation), we can simply print the string `Unknown`.

Karmastic:
So, this is a change that I requested for a reason.  We know the time is
not zero, so all we know is it has at LEAST been since we started up.  This
allows us to detect stalls at startup, whereas if we never see a block
we'll keep reporting ZERO.

The TimeSinceLast is not expected to be a super-precise value - it's
inconsistent when we're catching up.  This change makes it reasonable for
tracking startup state.

zeldovich:
Yet another unsatisfactory option is to take the timestamp from the last block.  Unsatisfactory because it involves comparing timestamps from different machines, which might not be necessarily in sync.

I agree that it would be good to display "unknown" after startup until the agreement protocol confirms a block.

I'd rather not see the current change merged, because it makes it difficult for the user to distinguish "not yet synced after startup" (currently the time displays as 0 in this situation, but that wouldn't be the case with this change) from "was synced and stalled".

Karmastic:
SO this piece of information was added for the specific purpose of
indicating stalls to a lambda function. The current behavior does not work
for startup.  If you’re not happy with this solution, please suggest one
that addresses the bug. I don’t care what goal displays. I care what the
api returns, which must be a number.  And I want it to indicate when it has
been too long since we’ve seen a new block.

egieseke:
What if we changed the name of the variable from "TimeSinceLastRound" to be "StartTimeWaitingForNewRound"?  This would help clarify the semantics and also provide a means of detecting a stall on startup.

Karmastic:
I would avoid changing the name - the status API is probably the most used API.  The appropriate name would be TimeWaitingForNextRound, this fix would just need to be accompanied by documentation indicating the TimeSinceLastRound is actually TimeWaitingForNextRound or TimeWaitingForRoundChange.

tsachiherman:
> SO this piece of information was added for the specific purpose of indicating stalls to a lambda function. The current behavior does not work for startup. If you’re not happy with this solution, please suggest one that addresses the bug. I don’t care what goal displays. I care what the api returns, which must be a number. And I want it to indicate when it has been too long since we’ve seen a new block.
> […](#)

We can add a field named `Uptime`. Whoever is currently was to use the above logic should use the `Uptime` in case `TimeWaitingForNextRound` is zero. Moreover, we could also make `TimeWaitingForNextRound` marked with `omitempty` so that we don't omit incorrect fields. ( it will get deserialized to zero anyway).

tsachiherman:
btw, when looking on the goal code 👍 
```
func makeStatusString(stat models.NodeStatus) string {
	lastRoundTime := fmt.Sprintf("%.1fs", time.Duration(stat.TimeSinceLastRound).Seconds())
	catchupTime := fmt.Sprintf("%.1fs", time.Duration(stat.CatchupTime).Seconds())
	return fmt.Sprintf(infoNodeStatus, stat.LastRound, lastRoundTime, catchupTime, stat.LastVersion, stat.NextVersion, stat.NextVersionRound, stat.NextVersionSupported)
}
```
it looks like we're *rounding* the TimeSinceLastRound value; that means that small number can be mixed up with "not-yet-synced". we might want to patch this code so that it would return the minimal value ( i.e. 0.1 ) if the stat.TimeSinceLastRound < 0.1.

Karmastic:
Eric - can you please address the 'know if we've never synced' concerns here by adding another field to the status that is a simple HasSyncedSinceStartup boolean?

egieseke:
I will add a new HasSyncedSinceStartup boolean field.

tsachiherman:
Note that we need to have a unit test that covers the new (and existing!) fields on both the GOAL and the API level.

egieseke:
OK, I am adding unit tests, good thinking.

node/node.go
```diff
@@ -300,6 +302,9 @@ func (node *AlgorandFullNode) Start() {
 	node.mu.Lock()
 	defer node.mu.Unlock()
 
+	// initialize last round start time to current time
+	node.lastRoundTimestamp = time.Now()
```
rotemh: Please consider removing this line. It implies incorrect data.

egieseke: Removed.

node/node_test.go
```diff
@@ -40,6 +38,7 @@ import (
 	"github.com/algorand/go-algorand/util"
 	"github.com/algorand/go-algorand/util/db"
 	"github.com/algorand/go-algorand/util/execpool"
+	"github.com/stretchr/testify/require"
```
tsachiherman: could you please move this back ? our code is always having the vendored code separated.

egieseke: done.

egieseke:
All suggested changes have been applied.  Please approve.  Thanks!

PR #138, opened 2019-07-10
[GOAL2-761] WsFetcherService::handleNetworkMsg telemetry event becomes level Info, was level Warn

EvanJRichard:
## Summary

I noticed this task in the emailed list of pre-sprint-resumption tasks: downgrade the event in question to `Info` from `Warn`.

## Test Plan

Automated tests.


Vervious:
I'll merge for now but we shouldn't close GOAL2-761 as a more informative error is certainly preferable.

PR #139, opened 2019-07-10
Log a description of context in db.Atomic.

tsachiherman:
## Summary

Add a description field to the db.Atomic function so that we could log that in case of a warning/error.
Without any context, it's hard to understand which operation was consuming that timespan.


derbear:
> What do you think of having `Atomic()` instead compute the caller, something like:
> 
> ```
> _, file, line, ok := runtime.Caller(1)
> if ok {
>     fnDescription = fmt.Sprintf("%s:%d", file, line)
> }
> ```
> 
> It's a bit cumbersome to pass this string at every callsite, especially if there isn't a deep reason to require it (i.e., using `runtime.Callers()` like above).

I was thinking that in certain cases it might be nice to pass in more than just the function that's being called into the description.

derbear:
For example, maybe you'd like to pass in the round number for a block you're writing to the block database.

zeldovich:
Maybe make it an optional argument to db.Atomic then, using `...string` or `...interface{}`?

tsachiherman:
> For example, maybe you'd like to pass in the round number for a block you're writing to the block database.

Yes, that's a good example. I like it. In this case, I think the following makes more sense:
```
type idemFn func(tx *sql.Tx, args ...interface{}) error
func (db Accessor) Atomic(fn idemFn, args ...interface{} ) (err error)
```
where the caller would make the following call:
```
accessor.Atomic(storeBlockFunction, encodedBlockData, roundNumber)
```

The `Accessor.Atomic` could then figure out the function name using reflection and log it along with all the arguments ( as needed ).

as a bonus, it decouples the data from the implementation.

zeldovich:
It seems like the current version just makes the caller's life slightly more annoying, compared to using `runtime.Callers()`.  It seems like none of the callers do what @derek was describing (passing in additional information as part of this argument).

zeldovich:
Consider https://github.com/zeldovich/go-algorand/commit/b7a40dbc268fe4f823cdac797651b5a6dbf745d6

tsachiherman:
> Consider [zeldovich@b7a40db](https://github.com/zeldovich/go-algorand/commit/b7a40dbc268fe4f823cdac797651b5a6dbf745d6)

I like this, with the exception of the performance implications; it's ok to have all that code in case we need to log, but otherwise, we don't want to have that; I'll refactor that into a working PR.

tsachiherman:
example for output:
```
{"callee":"github.com/algorand/go-algorand/ledger.(*blockQueue).getBlockHdr.func1","caller":"/Users/tsachi/go/src/github.com/algorand/go-algorand/ledger/blockqueue.go:257","file":"dbutil.go","function":"github.com/algorand/go-algorand/util/db.Accessor.Atomic","level":"warning","line":218,"msg":"dbatomic: tx took 1.102593075s","readonly":true,"time":"2019-07-15T11:34:28.958451-04:00"}
{"callee":"github.com/algorand/go-algorand/ledger.(*blockQueue).getBlockAux.func1","caller":"/Users/tsachi/go/src/github.com/algorand/go-algorand/ledger/blockqueue.go:318","extra(0)":[569678],"file":"dbutil.go","function":"github.com/algorand/go-algorand/util/db.Accessor.Atomic","level":"warning","line":218,"msg":"dbatomic: tx took 1.107065333s","readonly":true,"time":"2019-07-15T11:34:30.065986-04:00"}
```

util/db/dbutil.go
```diff
@@ -125,24 +127,30 @@ func Retry(fn func() error) (err error) {
...
+	log := logging.Base().With("readonly", db.readOnly)
+	_, file, line, ok := runtime.Caller(2)
+	if ok {
+		log = log.With("caller", fmt.Sprintf("%s:%d", file, line))
+	}
+	log = log.With("callee", runtime.FuncForPC(reflect.ValueOf(fn).Pointer()).Name())
+	if len(extras) > 0 {
+		for i, e := range extras {
+			if e == nil || reflect.ValueOf(e).IsNil() {
+				continue
```
zeldovich: Does our logging package panic on a nil `With()` value?

util/db/dbutil.go
```diff
@@ -125,24 +127,30 @@ func Retry(fn func() error) (err error) {
...
-		descr = "r"
+// getDecoratedLogger retruns a decorated logger that includes the readonly true/false, caller and extra fields.
+func (db *Accessor) getDecoratedLogger(fn idemFn, extras ...interface{}) logging.Logger {
+	log := logging.Base().With("readonly", db.readOnly)
+	_, file, line, ok := runtime.Caller(2)
+	if ok {
+		log = log.With("caller", fmt.Sprintf("%s:%d", file, line))
+	}
+	log = log.With("callee", runtime.FuncForPC(reflect.ValueOf(fn).Pointer()).Name())
+	if len(extras) > 0 {
```
zeldovich: I think you can drop this `len()` check, since the inner `for` loop will be a noop when `len() == 0` anyway.

util/db/dbutil.go
tsachiherman: no, but it seems to printing out [null] when we're not passing any arguments, so I figured that I'll drop these non-useful fields. ( while retaining the index, so the n-th `nil` could be figured out )

ledger/blockqueue.go
```diff
@@ -319,7 +319,7 @@ func (bq *blockQueue) getBlockAux(r basics.Round) (blk bookkeeping.Block, aux ev
 		var err0 error
 		blk, aux, err0 = blockGetAux(tx, r)
 		return err0
-	})
+	}, r)
```
zeldovich: Seems like a good test for this PR, but it would be cleaner to leave this code consistent: either all `blockQueue.getBlock*()` methods should pass this `r` as extras, or none should.

util/db/dbutil.go
zeldovich: Fair enough, it does seem like the log message has enough context to figure out what the caller extras args were, even for nil values.

PR #145, opened 2019-07-11
[GOAL2-664] Add optional minTxnFee to api spec

### LINKED ISSUE #143: algod API for ConsensusParams
(copied)
It would be nice to make ConsensusParams available by API so that client libraries don't have to maintain a parallel copy to what is maintained in Go code.

`/v1/consensusparams` could return current version
`/v1/consensusparams/{escaped version`} a specific version

OR

The precise need that prompted this was wanting to see MinTxnFee in APIs related to transaction fees where suggested fee was 1 which was well below the actual MinTxnFee. So maybe we just need that.

Vervious:
This is useful for clients who want to lower bound the fee on their transactions.

(Optional, because we probably don't want to break the spec and force a v2 just yet. So clients that want to use this field should first test if it is zero)

fixes #143


daemon/algod/api/server/v1/handlers/handlers.go
```diff
@@ -603,6 +603,10 @@ func SuggestedParams(ctx lib.ReqContext, w http.ResponseWriter, r *http.Request)
 	params.GenesisHash = gh[:]
 	params.LastRound = uint64(stat.LastRound)
 	params.ConsensusVersion = string(stat.LastVersion)
+
+	proto := config.Consensus[stat.LastVersion]
```
zeldovich: Consider looking up the last block and using that block's consensus params.

Vervious: I was under the impression that `node.Status()` does exactly that - or is it better to do it explicitly in the handler?

zeldovich: Oops, you're right, so this does the right thing already.

Vervious:
> Do you have a unit test for this API that need to be updated as well ?

Good catch. Looks like no one has touched this test suite for a while. I will also rope in an update to client/models at the same time.

Vervious:
Sorry for the re-review. I'm piggybacking some tech debt cleanup onto this change.

Adding `MinTxnFee` means updating the client-facing `client/models/models.go`, which was entirely out of date with `v1/handlers/model.go`, so I decided to consolidate the two (and clean up dependencies), into `spec/v1/model.go` (and `spec/common/model.go`).

A lot of import statements got renamed.

Also added a test to `e2e-go/restAPI`

PR #146, opened 2019-07-11
Systemd service template

algobolson:
## Summary

Add algorand@.service unit template into .deb and .rpm packages.

## Test Plan

For deb I hacked the algorand@.service file into an older install, enabled both the base algorand.service for mainnet and an algorand@... for devnet. Installed a newly build .deb and it restarted both daemons. (Tested on Ubuntu 18)
A couple rounds of installing new rpm packages on Centos7 demonstrated restarting both the main `algorand` service and an `algorand@...` instance.

### DIFF
--- a/installer/algorand@.service
+++ b/installer/algorand@.service
@@ -1,13 +1,13 @@
 ## For some data directory $ALGORAND_DATA
 ##
 ## Setup:
-## ALGORAND_DATA=/var/lib/algorand_devnet2
+## ALGORAND_DATA=/var/lib/algorand_devnet
 ## sudo mkdir -p ${ALGORAND_DATA}
 ## sudo cp -p /var/lib/algorand/genesis/devnet/genesis.json ${ALGORAND_DATA}/genesis.json
 ## sudo cp -p /var/lib/algorand/system.json ${ALGORAND_DATA}/system.json
 ## sudo chown -R algorand ${ALGORAND_DATA}
 ##
-## systemd allows for config overrides is a .conf file under a directory named for the service. e.g.:
+## systemd allows for config overrides in a .conf file under a directory named for the service. e.g.:
 ##
 ## sudo mkdir -p /lib/systemd/system/algorand@$(systemd-escape ${ALGORAND_DATA}).service.d
 ## cat <<EOF>/lib/systemd/system/algorand@$(systemd-escape ${ALGORAND_DATA}).service.d/override.conf
--- a/installer/algorand@.service
+++ b/installer/algorand@.service
@@ -1,0 +1,38 @@
+## For some data directory $ALGORAND_DATA
+##
+## Setup:
+## ALGORAND_DATA=/var/lib/algorand_devnet2
+## sudo mkdir -p ${ALGORAND_DATA}
+## sudo cp -p /var/lib/algorand/genesis/devnet/genesis.json ${ALGORAND_DATA}/genesis.json
+## sudo cp -p /var/lib/algorand/system.json ${ALGORAND_DATA}/system.json
+## sudo chown -R algorand ${ALGORAND_DATA}
+##
+## systemd allows for config overrides is a .conf file under a directory named for the service. e.g.:
+##
+## sudo mkdir -p /lib/systemd/system/algorand@$(systemd-escape ${ALGORAND_DATA}).service.d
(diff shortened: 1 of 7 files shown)


installer/algorand@.service
```diff
@@ -0,0 +1,38 @@
...
+## sudo systemctl start algorand@$(systemd-escape ${ALGORAND_DATA})
+
+
+[Unit]
+Description=Algorand daemon for %I
+After=network.target
+
+[Service]
+ExecStart=/usr/bin/algod -d %I
+PIDFile=%I/algod.pid
```
tsachiherman: I'm not sure if the pid file would get generated by doing this;
maybe we should call `/usr/bin/goal node start -d %I` ?

algobolson: `algod` creates its own pid file

installer/rpm/algorand.spec
```diff
@@ -131,10 +102,12 @@ getent group nogroup >/dev/null || \
 
 %post
 chown -R algorand /var/lib/algorand
-%systemd_post algorand.service
+%systemd_post algorand
```
zeldovich: should we do `systemd_post algorand@*` too?

installer/debian/postinst
```diff
@@ -26,6 +26,7 @@ if [ "$1" = "configure" ] || [ "$1" = "abort-upgrade" ] || [ "$1" = "abort-decon
 		systemctl --system daemon-reload >/dev/null || true
 		if [ -n "$2" ]; then
 			_dh_action=restart
+			deb-systemd-invoke restart algorand@\* >/dev/null || true
```
zeldovich: Does this work?  I don't really know offhand.

installer/algorand@.service
```diff
@@ -0,0 +1,38 @@
...
+User=algorand
+Group=nogroup
+Restart=always
+RestartSec=5s
+ProtectSystem=full
+ProtectHome=true
+
+[Install]
+WantedBy=multi-user.target
+DefaultInstance=-var-lib-algorand
```
zeldovich: It seems like we don't want two services (algorand and algorand@) fighting over /var/lib/algorand.  What's the reason for having a default instance here?

installer/debian/postinst
zeldovich: Separately: it seems like something about `algorand@*` should go into prerm to stop all the algorand@* services

algobolson: It worked on one Ubuntu 18.04 system, I installed and upgraded and watched the algod's restart (`algorand` and `algorand@/path/to/devdir`)

installer/algorand@.service
algobolson: I was hoping that a default instance would allow for `algorand@` or `algorand` to exist based on the `algorand@.service` but I couldn't get those to work. Could probably drop the default instance.

installer/rpm/algorand.spec
algobolson: I don't think that will have any effect. I think plain 'post' only runs after the initial install but the preun-postun run on both uninstall and upgrade's stop-start. I wouldn't expect anyone to have created any systemd algorand@... instances before the initial install

installer/algorand@.service
Karmastic: Yeah - this is how we currently run as a service (starting algod directly)

installer/algorand@.service
```diff
@@ -0,0 +1,37 @@
+## For some data directory $ALGORAND_DATA
+##
+## Setup:
+## ALGORAND_DATA=/var/lib/algorand_devnet
```
Karmastic: Nit: Can we change the example to testnet instead of devnet?  We don't want to push people into connecting to devnet.

algobolson: swapped

installer/debian/postinst
zeldovich: In case you missed my other comment here: what about prerm?

algobolson: added parallel clauses in prerm and postrm; I'll do some testing of uninstall today.

algobolson: confirmed `dpkg -P algorand` kills both and removes their systemd cruft

algobolson:
posted fixes for .deb uninstall

Issue #84, opened 2019-06-24
Fix inconsistencies in `goal` long arguments.


derbear:
### Subject of the issue
Some goal commands accept `addr` while others accept `address` as long-form arguments.

### Your environment
* Software version: `algod -v`: 1.0.39516.dev [master] (commit #db5c6d56+)

* Operating System: Any 

### Steps to reproduce
`goal account balance` shows `  -a, --address string   Account address to retrieve balance (required)`
while
`goal account delete` shows `  -a, --addr string   Address of account to delete`

### Expected behaviour
The long-form options should either both be `--address` or `--addr`.

### Actual behaviour
One of the options is `--address` and the other is `--addr`.

rotemh:
Fixed by #147

PR #147, opened 2019-07-12
[GOAL2-794] In goal, change all instances of --addr to --address

EvanJRichard:
## Summary

A simple change: when specifying addresses to `goal`, sometimes the shorthand the argument was `addr`, and sometimes `address`. I searched for instances of `addr` and replaced them with `address`.

## Test Plan

Automated test suite.


Vervious:
Fixes #84

PR #149, opened 2019-07-12
Logging config search path

algobolson:
## Summary

Allow logging.config telemetry configuration file to live in $ALGORAND_DATA
If not found there, fall back to ~/.algorand/logging.config

## Test Plan

Manually tested `diagcfg telemetry {status,enable,disable}` and `goal logging {enable,disable}` to work with both ${ALGORAND_DATA}/logging.config and if that is not present fall back to ~/.algorand/logging.config
Updated unit tests for EnsureTelemetryConfig()


cmd/diagcfg/telemetry.go
```diff
@@ -36,11 +38,19 @@ func init() {
...
 	telemetryCmd.AddCommand(telemetryEndpointCmd)
 
+	telemetryCmd.PersistentFlags().StringVarP(&dataDir, "datadir", "d", "", "Data directory for the node")
 	// Enable Logging : node name
 	telemetryNameCmd.Flags().StringVarP(&nodeName, "name", "n", "", "Friendly-name to use for node")
 	telemetryEndpointCmd.Flags().StringVarP(&uri, "endpoint", "e", "", "Endpoint's URI")
 }
 
+// If we didn't get a value from -d, try $ALGORAND_DATA
+func maybeUpdateDataDirFromEnv() {
```
Karmastic: Nit: how about `resolveDataDir() string` which returns dataDir || os.Getenv()?  Updating the global var is ugly.  It looks like nil isn't really meaningful in calls to EnsureTelemetryConfig(), so you can just call EnsureTelemetryConfig(resolveDataDir(), ...).

algobolson:
ok, I think this should be:
* Read behavior (algod, goal logging status, diagcfg telemetry status): $ALGORAND_DATA/logging.config then $HOME/.algorand/logging.config
* Write behavior (goal logging enable/disable, diagcfg telemetry enable/disable): read as above, always write to $ALGORAND_DATA/logging.config (except in diagcfg where -d/$ALGORAND_DATA is optional, if that is not set then write to $HOME/.algorand/logging.config)

Karmastic:
I'm actually fine migrating to per-datadir, but gracefully handling when datadir not specified AND ~/.algorand/logging.config exists.  So you logic sounds correct; i would have one exception which is to also write the GUID/Host values to the ~/.algorand/logging.config file too (don't write enable flag or URI).  The other change would be to use an appropriate shared location instead of ~/.algorand when datadir is SHARED.  Actually, it would be even better if we did this in all cases instead of storing it in ~/.algorand/.  Let me know if this makes sense and we can work out details if needed.

algobolson:
If I `diagcfg telemetry enable` and I can't write to $ALGORAND_DATA then it should try to write to that file and error out so that I know it didn't work and I can `sudo` to the correct user or something.
If I as a common user run `diagcfg telemetry enable` without `-d` or `$ALGORAND_DATA` that will silently be useless in the deb/rpm installed /var/lib/algorand data dir case; it will write to $HOME/.algorand/logging.config and no daemon will check that. We could push the migration harder by refusing to create $HOME/.algorand/logging.config if it doesn't exist (maybe prompt them to `touch` the file if they want it).
I'm not sure what you meant about GUID/Host values; I haven't looked into the code about what is being generated into logging.config; I'm hoping that doesn't need to touch that and if we read a file or read nothing and initialize, modify, and write to a file the only thing that needs to change is the path of the file we read and the path of the file we write.

algobolson:
update from talking elsewhere:
Even if writing to $ALGORAND_DATA/logging.config, keep $HOME/.algorand/logging.config with at least {"GUID":""}  so that we can read from it to initialize a system-wide GUID and copy that into other $ALGORAND_DATA/logging.config we might make on a system with multiple algod data dirs.

Karmastic:
GUID and Name please.

algobolson:
There's no default value for "Name" in `logging/telemetryConfig.go` `createTelemetryConfig()`. Should there be? Were you thinking of a different field?

Karmastic:
No there is no default name.  My expectation was that you would update the shared copy with the current name - so it tracks the last name specified.  Basically, any time you would write the logging.config to the datadir, you would also persist the minimized config to the shared file.

algobolson:
ok, I think the latest round of code in `cmd/diagcfg/telemetry.go` `saveTelemetryConfig()` achieves this

cmd/goal/logging.go
```diff
@@ -46,7 +46,8 @@ var loggingCmd = &cobra.Command{
 	Long:  `Enable/disable and configure Algorand remote logging`,
 	Args:  validateNoPosArgsFn,
 	Run: func(cmd *cobra.Command, _ []string) {
-		cfg, err := logging.EnsureTelemetryConfig(nil, "")
+		dataDir := ensureSingleDataDir()
```
Karmastic: This exits if `-d` not specified which isn't exactly what we want, right?

logging/telemetry.go
```diff
@@ -71,28 +72,64 @@ func makeTelemetryState(cfg TelemetryConfig, hookFactory hookFactory) (*telemetr
...
+	if err != nil && os.IsNotExist(err) {
+		var configPath string
+		configPath, err = config.GetConfigFilePath(TelemetryConfigFilename)
+		if err != nil {
+			cfg = createTelemetryConfig()
+			return
+		}
+		cfg, err = LoadTelemetryConfig(configPath)
+	}
+	if err != nil {
```
Karmastic: Why are we swallowing the error here?

cmd/goal/logging.go
algobolson: duplicating logic between `goal logging` and `diagcfg telemetry` is bugging me. Can we just delete or deprecate `goal logging` and if we don't delete it let it be the slightly less capable version? Most of `goal` doesn't make sense without a data directory, so requiring a data directory here doesn't feel out of character.

Karmastic: Yes - `goal logging` should be deprecated

algobolson:
deprecation warning messages on `goal logging ...` (except `goal logging send`)
tighter error swallowing (no-file is okay, other errors get through)

algobolson:
ping. merge broke approval

PR #150, opened 2019-07-12
use ledger.BlockEvaluator to validate transactions in txpool

zeldovich:


ledger/eval.go
```diff
@@ -274,6 +274,13 @@ func (eval *BlockEvaluator) Round() basics.Round {
 	return eval.block.Round()
 }
 
+// ResetTxnBytes resets the number of bytes tracked by the BlockEvaluator to
+// zero.  This is a specialized operation used by the transaction pool to
+// simulate the effect of putting pending transactions in multiple blocks.
+func (eval *BlockEvaluator) ResetTxnBytes(validateTxnBytes bool) {
```
Karmastic: I don't see this used in this PR

data/pools/transactionPool.go
```diff
@@ -435,6 +301,92 @@ func (pool *TransactionPool) OnNewBlock(block bookkeeping.Block) {
...
+// transaction as verified.
+type alwaysVerifiedPool struct{}
+
+func (*alwaysVerifiedPool) Verified(txn transactions.SignedTxn) bool {
+	return true
+}
+
+func (pool *TransactionPool) addToPendingBlockEvaluatorOnce(tx *transactions.SignedTxn) error {
+	if tx.LastValid < pool.pendingBlockEvaluatorRound {
+		return fmt.Errorf("Transaction valid for %d..%d which is before %d", tx.FirstValid, tx.LastValid, simulatedNextRound)
```
Karmastic: You added a specific error for this - can you use that?

data/pools/transactionPool.go
```diff
@@ -435,6 +301,92 @@ func (pool *TransactionPool) OnNewBlock(block bookkeeping.Block) {
...
+// recomputeBlockEvaluator constructs a new BlockEvaluator and feeds all
+// in-pool transactions to it (removing any transactions that are rejected
+// by the BlockEvaluator).
+func (pool *TransactionPool) recomputeBlockEvaluator() (stats telemetryspec.ProcessBlockMetrics) {
+	pool.pendingBlockEvaluator = nil
+
+	latest := pool.ledger.Latest()
+	prev, err := pool.ledger.BlockHdr(latest)
+	if err != nil {
+		logging.Base().Warnf("TransactionPool.recomputeBlockEvaluator: cannot get prev header for %d: %v",
```
Karmastic: Any reason this isn't an error?

data/pools/transactionPool.go
```diff
@@ -435,6 +301,92 @@ func (pool *TransactionPool) OnNewBlock(block bookkeeping.Block) {
...
+			latest, err)
+		return
+	}
+
+	next := bookkeeping.MakeBlock(prev)
+	pool.pendingBlockEvaluatorRound = next.Round()
+	pool.pendingBlockEvaluator, err = pool.ledger.StartEvaluator(next.BlockHeader, &alwaysVerifiedPool{}, nil)
+	if err != nil {
+		logging.Base().Warnf("TransactionPool.recomputeBlockEvaluator: cannot start evaluator: %v", err)
```
Karmastic: Likewise - should this be an error?

data/pools/transactionPool.go
```diff
@@ -435,6 +301,92 @@ func (pool *TransactionPool) OnNewBlock(block bookkeeping.Block) {
...
 
+// alwaysVerifiedPool implements ledger.VerifiedTxnCache and returns every
+// transaction as verified.
+type alwaysVerifiedPool struct{}
+
+func (*alwaysVerifiedPool) Verified(txn transactions.SignedTxn) bool {
+	return true
+}
+
+func (pool *TransactionPool) addToPendingBlockEvaluatorOnce(tx *transactions.SignedTxn) error {
```
Karmastic: Curious - what does 'once' mean here?

data/pools/transactionPool.go
```diff
@@ -351,79 +265,31 @@ func (pool *TransactionPool) OnNewBlock(block bookkeeping.Block) {
...
+			} else {
+				unknownCommitted++
+			}
 		}
 	}
 
-	// remove all collected transactions
-	for txid, txErr := range remove {
-		pool.remove(txid, txErr)
+	if pool.pendingBlockEvaluator == nil || block.Round() >= pool.pendingBlockEvaluator.Round() {
```
Karmastic: Is this the correct Round() value?  There seems to be a potential for confusion between `pool.pendingBlockEvalutaor.Round()` and `pool.pendingBlockEvalutaorRound`.  It's not clear what the distinction is supposed to be between the two.

ledger/eval.go
```diff
@@ -274,6 +274,13 @@ func (eval *BlockEvaluator) Round() basics.Round {
...
+	eval.totalTxBytes = 0
```
Karmastic: should this increment eval.Round somehow?

data/pools/transactionPool.go
zeldovich: Hah, you're right, forgot about it.  Thanks.

data/pools/transactionPool.go
zeldovich: It might legitimately happen during catchup.

data/pools/transactionPool.go
zeldovich: Similarly, can legitimately happen during catchup.

data/pools/transactionPool.go
zeldovich: There's an `addToPendingBlockEvaluator` that retries adding to the block evaluator if it's out of space in the current "simulated" block, and it calls this `...Once()` function again.

data/pools/transactionPool.go
zeldovich: Yeah, the terminology here isn't ideal.  I'll think about how to clean it up.

`pendingBlockEvaluator.Round()` is the round in which we started "testing" the transactions in the pool (i.e., tentatively adding them to a block to see if they can commit).  `pendingBlockEvaluatorRound` is the round in which a particular transaction might eventually get committed, if we have more than a block's worth of transactions in the pool.  The reason for tracking the latter value is so that we can check that those future transactions are still going to be valid in those later rounds (by their LastValid, in particular).

ledger/eval.go
zeldovich: I considered that (that would be a better API for the transaction pool, to be able to simulate several future blocks, instead of the `pendingBlockEvaluatorRound` kludge that I currently have there).  But it turns out that changing `eval.Round()` leads to other complications in the block evaluator that didn't seem worth it; the block evaluator more-or-less assumes it executes all for a single round right now.

data/pools/transactionPool.go
```diff
@@ -160,45 +116,58 @@ func (pool *TransactionPool) Test(t transactions.SignedTxn) error {
...
+
+	// The threshold is multiplied by the feeThresholdMultiplier that
+	// tracks the load on the transaction pool over time.  If the pool
+	// is mostly idle, feeThresholdMultiplier will be 0, and all txns
+	// are accepted (assuming the BlockEvaluator approves them, which
+	// requires a flat MinTxnFee).
+	feePerByte = feePerByte * pool.feeThresholdMultiplier
+
+	// The threshold grows exponentially if there are multiple blocks
+	// pending in the pool.
```
derbear: What's the rationale here behind an exponential growth in threshold? I understand that it prevents spam, but isn't this a relatively aggressive strategy? Why exponential and not, say, quadratic?

data/pools/transactionPool.go
derbear: Is that because of the sqlite bug?

data/pools/transactionPool.go
```diff
@@ -352,81 +251,55 @@ func (pool *TransactionPool) OnNewBlock(block bookkeeping.Block) {
...
-			stats.KnownCommittedCount++
+	// Adjust the pool fee threshold.  The rules are:
+	// - If there was less than one full block in the pool, reduce
+	//   the multiplier by 2x.  It will eventually go to 0, so that
+	//   only the flat MinTxnFee matters if the pool is idle.
+	// - If there were less than two full blocks in the pool, keep
+	//   the multiplier as-is.
+	// - If there were two or more full blocks in the pool, grow
+	//   the multiplier by 2x (or increment by 1, if 0).
+	switch pool.numPendingWholeBlocks {
```
derbear: Shouldn't this happen after `recomputeBlockEvaluator`?

derbear: Also, should this be guarded by the conditional `if pool.pendingBlockEvaluator == nil || block.Round() >= pool.pendingBlockEvaluator.Round() {`?

data/pools/transactionPool.go
zeldovich: No, it might happen in the unlikely case that, by the time we call `BlockHdr()`, the latest block has moved well past `latest`, and that block has been removed from the local non-archival block store.

data/pools/transactionPool.go
zeldovich: The benefit of an exponential strategy is that we don't have to pick any magic fee.  With a quadratic approach, it would have to be quadratic relative to some starting point (say, MinTxnFee?), but that seems less satisfactory, and our starting point might not be a good guess for the steady-state operation of the txpool (say, if the txpool is operating in a regime when lots of users want to pay well over MinTxnFee), in which case the threshold would ramp up quite slowly to deal with a DoS attack.

zeldovich:
> To clarify, this changes the behavior of the transaction pool, correct? Instead of sorting transactions downward by priority order and then returning the first block of transactions, it
> 
> * accepts transactions in order of receipt and remembers them forever
> * as transactions are accepted, exponentially increases required fee for subsequent transactions to enter after we have two whole blocks' worth of transactions
> * on each new block, refreshes the set of pending transactions with the block evaluator

Indeed.  The big benefit of this approach is that we are quite sure that transactions in the pool can be committed, and as a result, transactions for which we expend effort (in-memory storage and bandwidth for gossip) will be charged a fee.

Previously, this wasn't true for two main reasons:

- The pool used different code than the block evaluator, so there were subtle differences where the pool thought a transaction could commit, but the block evaluator would never have accepted it.

- Transactions could be added to the block in a different order than the one in which they arrived, and that order might invalidate some transactions.  The simplest example would be that a close transaction arrived late, but has high priority, so it goes first into the block, preventing any other spends from that account from being added afterwards.  That's, in principle, fixable, but it becomes increasingly less tenable to fix such issues as we get more complex transactions, like the sub-currency support, multi-txns / grouped txns, eventually smart contracts, etc.

data/pools/transactionPool.go
zeldovich: We want to measure the size of the pool before the next block's worth of transactions get subtracted away.  Moving this code to after `recomputeBlockEvaluator` would mean we have no way to distinguish the case when the pool is under-utilized (`numPendingWholeBlocks == 0`, where we want to reduce the threshold) from the case when the pool is about in steady-state (`numPendingWholeBlocks == 1`, where we want to keep the threshold as-is).

zeldovich: Moving it into the guarded block sounds reasonable if it seems clearer.  I'll do that.

The reason I don't think it matters is that it would only affect the computation of `feeThresholdMultiplier` during catchup (when that guarded block might not fire), and the threshold is sort-of meaningless during catchup anyway.

PR #151, opened 2019-07-12
[GOAL2-744] Rate limit gossip connections

tsachiherman:
## Summary

Add rate limiting for the gossip connections.

## Config
```
	// ConnectionsRateLimitingWindowSeconds is being used in conjunction with ConnectionsRateLimitingCount;
	// see ConnectionsRateLimitingCount description for further information. Providing a zero value
	// in this variable disables the connection rate limiting.
	ConnectionsRateLimitingWindowSeconds uint

	// ConnectionsRateLimitingCount is being used along with ConnectionsRateLimitingWindowSeconds to determine if
	// a connection request should be accepted or not. The gossip network examine all the incoming requests in the past
	// ConnectionsRateLimitingWindowSeconds seconds that share the same origin. If the total count exceed the ConnectionsRateLimitingCount
	// value, the connection is refused.
	ConnectionsRateLimitingCount uint
```

## Implementation Overview

Each phonebook entry would include an additional field called retryAfter. When selecting a server to connect, we're going to exclude servers who's retryAfter field is in the future.
When a server receives a request, it looks to see how many connections we already have from the the same origin, and within the rate limiting window.
If we already have more connections than we'd like to have within that window, we reply with 429, requesting the client to hold off for a while.
On the client side, we detect this status and update the phonebook to indicate that we're currently busy.

## Implementation Details

To track the request that are coming from each host, a new object was created - requestTracker; It hooks up to the endpoint listener as well as to the ServeHTTP as a proxy. Whenever a request is coming, it follows it from the TCP accept point and until the moment where it's being sent to the respective ServeHTTP handler.


tsachiherman:
these are all good points. the last one basically tells me that we need
some better peer history statistics object. just placing it in the active
connection won't be enough.

network/wsNetwork.go
```diff
@@ -511,12 +512,15 @@ func (wn *WebsocketNetwork) GetPeers(options ...PeerOption) []Peer {
 }
 
 func (wn *WebsocketNetwork) setup() {
+
 	wn.upgrader.ReadBufferSize = 4096
 	wn.upgrader.WriteBufferSize = 4096
 	wn.upgrader.EnableCompression = false
+	wn.dnsPhonebooks = make(map[string]*ThreadsafePhonebook, 1)
```
egieseke: Why not use multi phonebook here?

network/requestTracker.go
```diff
@@ -0,0 +1,399 @@
...
+		log:                 log,
+		config:              config,
+		hostRequests:        make(map[string]*hostIncomingRequests, 0),
+		acceptedConnections: make(map[string]*TrackerRequest, 0),
+	}
+}
+
+// Accept waits for and returns the next connection to the listener.
+func (rt *RequestTracker) Accept() (conn net.Conn, err error) {
+	for {
```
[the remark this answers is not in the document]
tsachiherman: I'm not a big fan of using the goto as a language construct; naturally, the for loop is implemented with a goto, but that shouldn't be a factor.
do you think that it would be cleaner if I would move the inner logic into a separate function ( i.e. "accept" ) and leave the for just for the rejected incoming requests ?

network/requestTracker.go
```diff
@@ -0,0 +1,399 @@
...
+	rt.acceptedConnectionsMu.Lock()
+	defer rt.acceptedConnectionsMu.Unlock()
+	// now that we're done with it, we can remove the trackedRequest from the acceptedConnections.
+	delete(rt.acceptedConnections, trackedRequest.remoteAddr)
+
+}
+
+func (rt *RequestTracker) updateRequestRemoteAddr(request *http.Request, rateLimitingWindowStartTime time.Time) (trackedRequest *TrackerRequest) {
+	rt.acceptedConnectionsMu.Lock()
+	trackedRequest = rt.acceptedConnections[request.RemoteAddr]
```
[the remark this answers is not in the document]
tsachiherman: I think that it should be pretty safe; a request.RemoteAddr is composed of "ip:port". that's unlikely to be changed. but I might want to review this again and consider the LocalAddrContextKey instead.

network/requestTracker.go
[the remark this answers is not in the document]
algobolson: no need to separate the function, all of this stuff is one chunk that makes sense. probably just leave this as is.

network/phonebook.go
```diff
@@ -17,28 +17,76 @@
...
+	"math"
 	"math/rand"
+	"time"
 
 	"github.com/algorand/go-deadlock"
 )
 
+// when using GetAddresses with getAllAddresses, all the addresses will be retrieved, regardless
+// of how many addresses the phonebook actually has. ( with the retry-after logic applied )
+const getAllAddresses = math.MaxInt32
```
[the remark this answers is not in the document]
tsachiherman: I'm not opposing to change the variable name, but `maxAddresses` isn't the best one. It implies that this the highest number of supported addresses, which is not the rationale here.
( in practice, I would have set the maxAddresses to around 1024, but that's beyond the scope of this PR )

config/local_defaults.go
```diff
@@ -288,6 +290,12 @@ func migrate(cfg Local) (newCfg Local, err error) {
 		if newCfg.PriorityPeers == nil {
 			newCfg.PriorityPeers = map[string]bool{}
 		}
+		if newCfg.ConnectionsRateLimitingWindowSeconds == defaultLocalV3.ConnectionsRateLimitingWindowSeconds {
```
Karmastic: migration not required for new parameters.

installer/config.json.example
```diff
@@ -39,5 +39,7 @@
     "TxSyncIntervalSeconds": 60,
     "TxSyncServeResponseSize": 1000000,
     "TxSyncTimeoutSeconds": 30,
-    "ForceRelayMessages": false
+    "ForceRelayMessages": false,
```
Karmastic: Can you please re-alphabetize this file -- I would like to keep it alphabetized for easy reference.

tsachiherman: yes

tsachiherman:
I temporary closed this PR, as it require some additional work which I'd like to take offline.

network/requestTracker.go
```diff
@@ -0,0 +1,385 @@
...
+
+	"github.com/algorand/go-algorand/config"
+	"github.com/algorand/go-algorand/logging"
+	"github.com/algorand/go-algorand/logging/telemetryspec"
+)
+
+const (
+	// maxHeaderReadTimeout is the time limit where items would remain in the acceptedConnections cache before being pruned.
+	// certain malicious connections would never get to the http handler, and therefore must be pruned every so often.
+	maxHeaderReadTimeout = 30 * time.Second
```
egieseke: should the maxHeaderReadTimeout be configurable?

network/requestTracker.go
```diff
@@ -0,0 +1,385 @@
...
+		remoteAddr: remoteAddr,
+		remoteHost: remoteHost,
+		remotePort: remotePort,
+	}
+}
+
+// hostIncomingRequests holds all the requests that are originating from a single host.
+type hostIncomingRequests struct {
+	remoteHost string
+	requests   []*TrackerRequest // this is an ordered list, according to the requestsHistory.created
```
egieseke: could this create a memory problem if there are too many requests from a single host?

network/requestTracker.go
tsachiherman: not really. If we "fail" with this timeout, it just mean that the passed TrackerRequest structure would be generated on our http handler. This is sub-optimal, because we loose track of the original TrackerRequest creation time, but has no practical negative implications otherwise.

network/requestTracker.go
tsachiherman: It's not being stored for a long period of time. In fact, it's being stored only for the duration of the window size - which defaults to 1 second.

network/phonebook.go
```diff
@@ -86,43 +153,66 @@ func (p *ThreadsafePhonebook) ExtendPeerList(more []string) {
...
 // ReplacePeerList replaces set of addresses with that passed in.
 func (p *ThreadsafePhonebook) ReplacePeerList(they []string) {
 	p.lock.Lock()
 	defer p.lock.Unlock()
-	p.addrs = make([]string, len(they))
-	copy(p.addrs, they)
+	p.entries.ReplacePeerList(they)
+}
+
+// MergePeerList merges a set of addresses with that passed in.
```
algobolson: This is what I wish ReplacePeerList did. Why should ReplacePeerList forget the metadata associated with an address that was there before and is there in the new list?

network/phonebook.go
```diff
@@ -133,46 +223,33 @@ func (mp *MultiPhonebook) GetAddresses(n int) []string {
...
-	}
-
-	addrSet := make(map[string]bool, total)
-	for pi, size := range sizes {
-		if addrs[pi] != nil {
-			mp.addAddressArrayToAdressSet(&addrSet, &(addrs[pi]))
-		} else {
-			xa := (*mp.phonebookMap[names[pi]]).GetAddresses(size)
-			mp.addAddressArrayToAdressSet(&addrSet, &xa)
+	uniqueEntries := make(map[string]bool, 0)
```
algobolson: simplification: since we're gathering everything into uniqueEntries, we could at the same time gather into one `addrs []string` and shuffle-select from that, and we don't need `sizes` anymore.
The original intent was not to ever gather a copy of all the addresses into one structure thus doubling the memory impact of all the addresses; but I guess that was probably premature optimization and all the addresses will probably never be over 1000 or 10000 and those are trivial numbers to deal with.

network/phonebook.go
tsachiherman: Good point; I'll merge the two.

network/phonebook.go
tsachiherman: sounds good; I'll make the update.

network/phonebook.go
```diff
@@ -17,28 +17,76 @@
...
+		if t.After(entry.retryAfter) {
+			o = append(o, addr)
+		}
+	}
+	return o
+}
+
+// ReplacePeerList replaces set of addresses with that passed in.
+func (e *phonebookEntries) ReplacePeerList(they []string) {
+	// clear current map.
```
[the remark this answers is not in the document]
tsachiherman: Yes; note that we haven't used it in a way that it would make a difference. But it is cleaner that way.

PR #152, opened 2019-07-12
GOAL2-653 deprecate IsRelay field in NodeConfig

algobolson:
## Summary

Simplify NodeConfig struct. IsRelay is redundant, IsRelay() = nc.NetAddress != ""

## Test Plan

compilation plus unit tests

### DIFF
--- a/cmd/netgoal/generate.go
+++ b/cmd/netgoal/generate.go
@@ -199,7 +199,10 @@ func generateNetworkTemplate(templateFilename string, wallets, relays, nodeHosts
 		newNode.NodeNameMatchRegex = ""
 		newNode.FractionApply = 0.0
 		newNode.Name = name
-		newNode.IsRelay = true
+		if newNode.NetAddress == "" {
+			// if not set by relayTemplate ensure that it is a relay
+			newNode.NetAddress = "0.0.0.0:4160"
+		}
 		newNode.Wallets = nil
 		host.Nodes = append(host.Nodes, newNode)
 		network.Hosts = append(network.Hosts, host)
--- a/cmd/netgoal/generate.go
+++ b/cmd/netgoal/generate.go
@@ -18,6 +18,7 @@ package main
 
 import (
 	"encoding/json"
+	"errors"
 	"math/rand"
 	"os"
 	"regexp"
@@ -200,8 +201,7 @@ func generateNetworkTemplate(templateFilename string, wallets, relays, nodeHosts
 		newNode.FractionApply = 0.0
 		newNode.Name = name
 		if newNode.NetAddress == "" {
-			// if not set by relayTemplate ensure that it is a relay
-			newNode.NetAddress = "0.0.0.0:4160"
+			return errors.New("relay template did not set NetAddress")
 		}
 		newNode.Wallets = nil
 		host.Nodes = append(host.Nodes, newNode)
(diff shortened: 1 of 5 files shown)


cmd/netgoal/generate.go
```diff
@@ -199,7 +199,10 @@ func generateNetworkTemplate(templateFilename string, wallets, relays, nodeHosts
 		newNode.NodeNameMatchRegex = ""
 		newNode.FractionApply = 0.0
 		newNode.Name = name
-		newNode.IsRelay = true
+		if newNode.NetAddress == "" {
```
tsachiherman: I think that you don't want to have these workarounds; you want to make sure that the relayTemplates is configured correctly. If not, we need to fix it there.

algobolson: yeah, the other option would to be to fail with error message if a "relay" template failed to set NetAddress. I'll work that up.

algobolson:
looks like test/testdata/nettemplates/*.json need fixing; didn't show up when grepping the .go sources.

algobolson:
There's a separate `IsRelay bool` in a different template system in `netdeploy/networkTemplate.go`
That's what interprets the templates in `test/testdata/nettemplates/*.json`
That structure doesn't even have a 'NetAddress' field because it's a simple ephemeral system I think only used by `goal network create` which makes local private networks, where all the NetAddress values are effectively "127.0.0.1:0" so that they only listen locally and get a random dynamic port each time they start up.
I'm tempted to leave that system alone.

cmd/netgoal/generate.go
Karmastic: Yeah - fail if netaddress not specified for relays - don't default.

netdeploy/remote/nodeConfig.go
```diff
@@ -19,7 +19,7 @@ package remote
 // NodeConfig represents the configuration settings to apply to a single node running on a host
 type NodeConfig struct {
 	Name               string `json:",omitempty"`
-	IsRelay            bool
+	IsRelayDEPRECATED  bool   `json:"IsRelay,omitempty"` // The field still exists to parse any old serialized NodeConfig objects, but "IsRelay" should not be used.
```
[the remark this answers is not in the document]
Karmastic: Yeah - for most things we don't fail on unknown properties (we do on things like genesis.json)

netdeploy/remote/nodecfg/nodeDir.go
```diff
@@ -143,7 +143,7 @@ func (nd *nodeDir) configureNetAddress(address string) (err error) {
 	}
 	fmt.Fprintf(os.Stdout, " - Assigning NetAddress: %s\n", address)
 	nd.config.NetAddress = address
-	if address != "" && address[0] == ':' && nd.IsRelay {
+	if address != "" && address[0] == ':' && nd.IsRelay() {
```
Karmastic: You can remove the call to nd.IsRelay() - it's redundant now.

netdeploy/remote/nodeConfig.go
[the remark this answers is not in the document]
algobolson: field is deleted now

cmd/netgoal/generate.go
algobolson: returns an error now

netdeploy/remote/nodecfg/nodeDir.go
algobolson: redundancy eliminated, but kept nd.IsRelay() because that has some meaning

Issue #83, opened 2019-06-24
Improve goal support for shell pipelines.


derbear:
Certain `goal` commands accept filenames as input and output parameters. However, they do not accept `stdin` or `stdout`, which are convenient for use in shell pipelines. For instance, `goal clerk sign` accepts a `-i` input file and a `-o` output file but not stdin/stdout.

Adding support to goal for accepting stdin/stdout will simplify scripting.

algoradam:
It's worth noting that `-i /dev/stdin` and `-o /dev/stdout` should work, but it would be nicer and more idiomatic to accept `-i -` and `-o -`

algoradam:
Fixed by #153

Vervious:
Thanks @jsign!

PR #153, opened 2019-07-12
cmd/goal: allow stdin/stdout shell pipes

jsign:
## Summary
Propose a solution for #83 

Notes:
- File reads and writes were abstracted to new functions. Since they're quite generic for `cmd`, they live on commands.go
- Depending on `filename`, it distinguishes using two different functions (Write vs WriteFile, ReadAll vs ReadFile). The functions WriteFile and ReadFile functions close the file (not sure would be quite right for std*, so read/wrote directly to os.Std*).
- Further checks can be added to see if stdin is being piped or attached to a prompt. Worth it?

## Test Plan
This can be tested doing something like the following:
`./goal clerk send -a 1 -t 52ZOFGOJR7A33SIJGOCSD7LTHJGB3JW26Y44D5THR25WIN2P5VGW2IUQMQ -o - -w unsafeWallet | ./goal clerk sign -i - -o - -w unsafeWallet | ./goal clerk inspect -`
Output:
```
-[0]
{
  "sig": "PFPpUVfFG83xyKBcBz4qqDgg/AaN6Me7TBE9in1JbYvME5eslgmM/loqEfGbs4vhyUYrpLD6bpFr2N/vGcbxDQ==",
  "txn": {
    "amt": 1,
    "fee": 1000,
    "fv": 609159,
    "gen": "mainnet-v1.0",
    "gh": "wGHE2Pwdvd7S12BL5FaOP20EGYesN73ktiC1qzkkit8=",
    "lv": 610159,
    "note": "QIY+2fDk+6I=",
    "rcv": "52ZOFGOJR7A33SIJGOCSD7LTHJGB3JW26Y44D5THR25WIN2P5VGW2IUQMQ",
    "snd": "52ZOFGOJR7A33SIJGOCSD7LTHJGB3JW26Y44D5THR25WIN2P5VGW2IUQMQ",
    "type": "pay"
  }
}
```
Where `unsafeWallet` is an unencrypted wallet (to avoid the password prompt). The `-t addr` is dummy in this example. All affected commands use the same underlying new `readFile` and `writeFile` functions.

Curiously (not sure why), while doing this feature I ended up submitting a [tiny patch](https://github.com/golang/go/issues/33064) to the Go project for `ioutil.WriteFile()`.

### DIFF
--- a/cmd/goal/commands.go
+++ b/cmd/goal/commands.go
@@ -452,3 +452,26 @@ func reportErrorf(format string, args ...interface{}) {
 	// log.Warnf(format, args...)
 	os.Exit(1)
 }
+
+// writeFile is a wrapper of ioutil.WriteFile which considers the special
+// case of stdout filename
+func writeFile(filename string, data []byte, perm os.FileMode) error {
+	var err error
+	if filename == stdoutFilenameValue {
+		// Write to Stdout
+		if _, err = os.Stdout.Write(data); err != nil {
+			return err
+		}
+		return nil
+	}
+	return ioutil.WriteFile(filename, data, 0600)
+}
+
+// readFile is a wrapper of ioutil.ReadFile which consniders the
+// special case of stdin filename
+func readFile(filename string) ([]byte, error) {
+	if filename == stdinFileNameValue {
+		return ioutil.ReadAll(os.Stdin)
+	}
+	return ioutil.ReadFile(filename)
+}
--- a/cmd/goal/commands.go
+++ b/cmd/goal/commands.go
@@ -464,7 +464,7 @@ func writeFile(filename string, data []byte, perm os.FileMode) error {
 		}
 		return nil
 	}
-	return ioutil.WriteFile(filename, data, 0600)
+	return ioutil.WriteFile(filename, data, perm)
 }
 
 // readFile is a wrapper of ioutil.ReadFile which consniders the
--- a/cmd/goal/commands.go
+++ b/cmd/goal/commands.go
@@ -442,13 +442,13 @@ func reportWarnf(format string, args ...interface{}) {
 }
 
 func reportErrorln(args ...interface{}) {
-	fmt.Println(args...)
+	fmt.Fprintln(os.Stderr, args...)
 	// log.Warnln(args...)
 	os.Exit(1)
 }
 
 func reportErrorf(format string, args ...interface{}) {
-	fmt.Printf(format+"\n", args...)
+	fmt.Fprintf(os.Stderr, format+"\n", args...)
 	// log.Warnf(format, args...)
 	os.Exit(1)
 }
--- a/cmd/goal/clerk.go
+++ b/cmd/goal/clerk.go
@@ -20,7 +20,6 @@ import (
 	"encoding/base64"
 	"fmt"
 	"io"
-	"io/ioutil"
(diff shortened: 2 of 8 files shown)


algoradam:
Thanks for contributing!
Probably we should also make `goal`'s error messages go to stderr instead of stdout so we don't e.g. accidentally try to sign an error message. (`reportErrorf` and related functions in `cmd/goal/commands.go` are the functions to change)

jsign:
> Thanks for contributing!
> Probably we should also make `goal`'s error messages go to stderr instead of stdout so we don't e.g. accidentally try to sign an error message. (`reportErrorf` and related functions in `cmd/goal/commands.go` are the functions to change)
Yes, you're right. 
Done.

cmd/goal/commands.go
```diff
@@ -442,13 +442,36 @@ func reportWarnf(format string, args ...interface{}) {
...
+func writeFile(filename string, data []byte, perm os.FileMode) error {
+	var err error
+	if filename == stdoutFilenameValue {
+		// Write to Stdout
+		if _, err = os.Stdout.Write(data); err != nil {
+			return err
+		}
+		return nil
+	}
+	return ioutil.WriteFile(filename, data, 0600)
```
algoradam: `0600` should probably be `perm` here

jsign: Ups, agree.

algoradam:
```
--- FAIL: TestAccountNewDuplicateFails (1.07s)
    require.go:157: 
        	Error Trace:	account_test.go:67
        	Error:      	Not equal: 
        	            	expected: *errors.errorString(&errors.errorString{s:"account name already taken"})
        	            	actual  : *exec.ExitError(&exec.ExitError{ProcessState:(*os.ProcessState)(0xc00023fd80), Stderr:[]uint8(nil)})
        	Test:       	TestAccountNewDuplicateFails
```

Looks like one of our end-to-end tests expected the error message to be on stdout. The `goal` e2e test fixture in `test/framework/fixtures/goalFixture.go` likely needs to be updated to look at stderr.

Alternatively, if you prefer, feel free to revert the stdout/stderr change and I'll merge this PR and create a separate issue for the stdout/stderr thing.

jsign:
> ```
> --- FAIL: TestAccountNewDuplicateFails (1.07s)
>     require.go:157: 
>         	Error Trace:	account_test.go:67
>         	Error:      	Not equal: 
>         	            	expected: *errors.errorString(&errors.errorString{s:"account name already taken"})
>         	            	actual  : *exec.ExitError(&exec.ExitError{ProcessState:(*os.ProcessState)(0xc00023fd80), Stderr:[]uint8(nil)})
>         	Test:       	TestAccountNewDuplicateFails
> ```
> 
> Looks like one of our end-to-end tests expected the error message to be on stdout. The `goal` e2e test fixture in `test/framework/fixtures/goalFixture.go` likely needs to be updated to look at stderr.
> 
> Alternatively, if you prefer, feel free to revert the stdout/stderr change and I'll merge this PR and create a separate issue for the stdout/stderr thing.

Give me some minutes to take a look.

jsign:
I think these new changes make sense, second eyes will be important.
Couldn't ran e2e on this machine since I'm on low-battery. 
Let's take the shot and see what CI has to say, and see what we do.

Issue #148, opened 2019-07-12
Update `goal listpartkeys` to indicate which part key is currently registered


Vervious:
(Copied)
Or some mechanism for reporting this so we can tell which of possibly many part keys are actually being used.

Vervious:
When offline:
```
$ goal account listpartkeys
Registered	Filename                                                                        	Parent address                                              	 First round	  Last round	   First key
unknown   	BH55E5RMBD4GYWXGX5W5PJ5JAHPGM5OXKDQH5DC4O2MGI7NW4H6Q.10000.10111.partkey        	BH55E5RMBD4GYWXGX5W5PJ5JAHPGM5OXKDQH5DC4O2MGI7NW4H6VOE4CP4  	       10000	       10111	     29323.0
unknown   	BH55E5RMBD4GYWXGX5W5PJ5JAHPGM5OXKDQH5DC4O2MGI7NW4H6VOE4CP4.666888.1000000.partkey	BH55E5RMBD4GYWXGX5W5PJ5JAHPGM5OXKDQH5DC4O2MGI7NW4H6VOE4CP4  	      666888	     1000000	     69.7765
```

When online:
```
$ goal account listpartkeys
Registered	Filename                                                                        	Parent address                                              	 First round	  Last round	   First key
no        	BH55E5RMBD4GYWXGX5W5PJ5JAHPGM5OXKDQH5DC4O2MGI7NW4H6Q.10000.10111.partkey        	BH55E5RMBD4GYWXGX5W5PJ5JAHPGM5OXKDQH5DC4O2MGI7NW4H6VOE4CP4  	       10000	       10111	     29323.0
yes       	BH55E5RMBD4GYWXGX5W5PJ5JAHPGM5OXKDQH5DC4O2MGI7NW4H6VOE4CP4.666888.1000000.partkey	BH55E5RMBD4GYWXGX5W5PJ5JAHPGM5OXKDQH5DC4O2MGI7NW4H6VOE4CP4  	      666888	     1000000	     69.7779
```

PR #155, opened 2019-07-12
[GOAL2-770] refactor `Node.GetBalanceAndStatus`

Vervious:
This is in preparation for adding an API endpoint (or modifying `account/addr`) to fetch the current registered participation key for an account. This enables `goal` to look up whether a `.partkey` is currently active.

This PR does some prerequisite work:
- Removes `GetBalanceAndStatus` function and replaces with `GetBalanceRecord` and `GetBalanceRecordWithoutPendingRewards`. This simplifies the ledger interface and makes the future change more natural/readable.
- Removes redundant rewards application code (would appreciate a detailed review). 

Partial work for #148

### DIFF
--- a/cmd/goal/clerk.go
+++ b/cmd/goal/clerk.go
@@ -20,7 +20,6 @@ import (
 	"encoding/base64"
 	"fmt"
 	"io"
-	"io/ioutil"
 	"os"
 
 	"github.com/algorand/go-algorand/crypto"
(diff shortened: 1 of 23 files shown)


Vervious:
I somewhat agree. But consider the counterexample where two functions both return integers, where those two integers have completely distinct semantics (e.g. a sum vs a multiplication). To some extent you could argue that the two balance records (with and without rewards) have the same structure.

But yes, maybe it can be better. Not entirely sure what to do instead, let me think about it.

Vervious:
(work in progress)

- Removes `node.GetBalanceAndStatus` function, replaces with direct calls to `ledger.Lookup` and `ledger.LookupWithoutRewards`
    - this implicitly removes a redundant call to `data.Money(proto, totals.RewardsLevel)`. Merging #158 for better testing should happen first so we don't get this wrong
- Removes `node.GetSupply`, replaces with direct calls to `ledger`
- Removes `node.LatestRound`, replaces with direct calls to `ledger.Latest`
- Removes `node.GetBlock`, replaces with direct call to `ledger.BlockCert`

let me know if this is not a wise course of action.

PR #156, opened 2019-07-15
benchmark txpool steady-state operation

zeldovich:
Add a benchmark for steady-state operation of the txpool (adding transactions to the pool and then committing them in the ledger).  Helpful in understanding the performance implications of PR #150.

### DIFF
--- a/data/pools/transactionPool_test.go
+++ b/data/pools/transactionPool_test.go
@@ -19,15 +19,19 @@ package pools
 import (
 	"fmt"
 	"math/rand"
+	"strings"
 	"testing"
 
 	"github.com/stretchr/testify/require"
 
+	"github.com/algorand/go-algorand/agreement"
 	"github.com/algorand/go-algorand/config"
 	"github.com/algorand/go-algorand/crypto"
 	"github.com/algorand/go-algorand/data/basics"
 	"github.com/algorand/go-algorand/data/bookkeeping"
 	"github.com/algorand/go-algorand/data/transactions"
+	"github.com/algorand/go-algorand/ledger"
+	"github.com/algorand/go-algorand/logging"
 	"github.com/algorand/go-algorand/protocol"
 )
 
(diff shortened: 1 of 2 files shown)


zeldovich:
In this PR, yes, `alwaysVerifiedPool` is used only in test code.  It's in `transactionPool.go` because, in a subsequent dependent PR (#150), it's also used in `transactionPool.go`.

data/pools/transactionPool_test.go
```diff
@@ -40,6 +44,12 @@ func keypair() *crypto.SignatureSecrets {
 	return s
 }
 
+type TestingT interface {
```
Karmastic: Is this used anywhere for anything?

zeldovich: It allows passing in either a `testing.T` or a `testing.B`.  I think in this PR, only the benchmark code calls it; it's a leftover from the larger PR #150 that also uses this code from tests.

data/pools/transactionPool_test.go
```diff
@@ -988,3 +1116,96 @@ func BenchmarkTransactionPoolPending(b *testing.B) {
...
+
+	poolTxnQueue := signedTransactions
+	var ledgerTxnQueue []transactions.SignedTxn
+
+	for len(poolTxnQueue) > 0 || len(ledgerTxnQueue) > 0 {
+		// Fill up txpool
+		for len(poolTxnQueue) > 0 {
+			stx := poolTxnQueue[0]
+			err := transactionPool.Remember(stx)
+			if err == nil {
```
yossigi: nit: maybe you could avoid the continue/break statements using an else case.
(slightly more easy to follow in my opinion)

data/pools/transactionPool_test.go
```diff
@@ -988,3 +1116,96 @@ func BenchmarkTransactionPoolPending(b *testing.B) {
...
+				ledgerTxnQueue = append(ledgerTxnQueue, stx)
+				continue
+			}
+			if strings.Contains(err.Error(), "transaction pool is full") {
+				break
+			}
+			require.NoError(b, err)
+		}
+
+		// Commit a block
```
yossigi: it seems to benchmark a few things that the pool is doing together:
-remember transactions
-remove pending txes when a block gets committed 
-creating a proposed block.

Maybe that's okay for this use case, I'm not sure if more detailed benchmarks (splitting the 3 operations) would be more insightful

PR #158, opened 2019-07-15
[GOAL2-770] rework an e2e rewards test to test more API functionality

Vervious:
Also removes some extraneous account poking, clarifying the test.
This enables/qualifies PR #155 (refactor `Node.GetBalanceAndStatus`) which is in turn related to #148 (Update `goal listpartkeys` to indicate which part key is registered)


Vervious:
Forgot that I should have merged this before #155 ...

PR #161, opened 2019-07-16
[GOAL2-751] No omitempty fields when GETing a Transaction

EvanJRichard:
## Summary

Previously, unused fields of the Transaction object would be omitted "on the wire". This made it hard to develop against the Transaction object as an external developer.

This PR proposes to always return all fields of the Transaction object.

## Test Plan

Plan is to use automated test suite.


daemon/algod/api/spec/v1/model.go
```diff
@@ -150,25 +150,25 @@ type Transaction struct {
...
 	//
 	// required: false
-	PoolError string `json:"poolerror,omitempty"`
+	PoolError string `json:"poolerror"`
 
 	// This is a list of all supported transactions.
 	// To add another one, create a struct with XXXTransactionType and embed it here.
 	// To prevent extraneous fields, all must have the "omitempty" tag.
-	Payment *PaymentTransactionType `json:"payment,omitempty"`
+	Payment *PaymentTransactionType `json:"payment"`
```
rotemh: This should have `omitempty`. If it's a keyReg txn, there is no reason that `Payment` fields will be there.

Vervious:
Are we worried about the upgradeability of our API? (let's say we want to add a field, a non-breaking change).

PR #163, opened 2019-07-16
[GOAL2-770] Expose on-chain account participation information API

Vervious:
Add a field to `/v1/account/{address}` exposing registered participation root key, selection key, first valid, last valid, and key dilution. Consider splitting this into a separate API if motivated by the use case.

- adds two e2e tests for `/v1/account/{address}`


Vervious:
(Part of a series of PRs for #148)

PR #166, opened 2019-07-16
add config "LogArchiveName" which can be templatized

algobolson:
## Summary

Templatize "node.archive.log" filename so that we can keep indefinite old algod "node.log" files.

## Test Plan

A node wast tested with a small "LogSizeLimit" to cause frequent rollover and templated "LogArchiveName" with no suffix or with .gz or .bz2 compression was show to work. Unconfigured "LogArchiveName" fell back to old behavior and "LogArchiveName" with a static name (no template parts) works.
Test value: `"LogArchiveName": "node.log.{{.Year}}{{.Month}}{{.Day}}_{{.Hour}}{{.Minute}}{{.Second}}-{{.EndYear}}{{.EndMonth}}{{.EndDay}}_{{.EndHour}}{{.EndMinute}}{{.EndSecond}}.bz2",`


config/config_test.go
```diff
@@ -343,6 +343,11 @@ func TestConfigInvariant(t *testing.T) {
 	err = codecs.LoadObjectFromFile(filepath.Join(configsPath, "config-v4.json"), &c4)
 	a.NoError(err)
 	a.Equal(defaultLocalV4, c4)
+
+	c5 := Local{}
```
tsachiherman: I think that you should following the configuration patten that I've used for the rate limiting : https://github.com/algorand/go-algorand/pull/151

Instead of defining a new configuration version, add it to the recent version and provide upgrade support.

tsachiherman:
(I'm not approving this just because I'd like to have @Karmastic's feedback first). It looks good otherwise.

config/config_test.go
algobolson: I'll only do that if I can rewrite all the loud comments in config/local_defaults.go to describe an actual process we actually do

algobolson:
I didn't do the compression in algod because that would be another thing we should wait on at shutdown time, and I didn't want to figure out how to plumb through another goroutine into a WaitGroup. Starting a subprocess should keep going if the algod exits. Also I think `gzip` and `bzip2` are supposed to be well behaved in a couple subtle ways I don't want to reproduce.
I'd be in favor of _only_ writing compressed logs and teaching `carpenter` and other tools to read from compressed logs.

config/config_test.go
Karmastic: OK I heard you.  Yes the comments are in-your-face and no longer apply.  You now do NOT need to create a new version or bother with migration UNLESS YOU ARE CHANGING THE DEFAULTS FOR AN EXISTING PARAMETER.

config/local_defaults.go
```diff
@@ -46,6 +46,53 @@ const configVersion = uint32(4)
 //
 // !!! WARNING !!!
 
+var defaultLocalV5 = Local{
+	// DO NOT MODIFY THIS STRUCTURE IN ANY WAY - See WARNING at top of file
+	Version:                               4,
```
Karmastic: For future reference, if you're going to create a v5 config (which you shouldn't do here!), you need to change the version to 5 :)

config/local_defaults.go
```diff
@@ -290,6 +337,13 @@ func migrate(cfg Local) (newCfg Local, err error) {
 		}
 		newCfg.Version = 4
 	}
+	// Migrate 4 -> 5
```
Karmastic: no needed

installer/config.json.example
```diff
@@ -20,6 +20,7 @@
     "IncomingMessageFilterBucketSize": 512,
     "IsIndexerActive": false,
     "LogSizeLimit": 1073741824,
+    "LogArchiveName": "node.archive.log",
```
Karmastic: So close to being alphabetized.  :)  Please fix this.

logging/cyclicWriter.go
```diff
@@ -51,6 +64,45 @@ func MakeCyclicFileWriter(liveLogFilePath string, archiveFilePath string, sizeLi
...
+		Minute    string
+		Second    string
+		EndYear   string
+		EndMonth  string
+		EndDay    string
+		EndHour   string
+		EndMinute string
+		EndSecond string
+	}{
+		fmt.Sprintf("%04d", cyclic.logStart.Year()),
```
Karmastic: So this is misleading as it's the start time for the current session, not when the log file was created.  Do you see a use-case for the logStart value?

algobolson: it probably won't be more than a couple seconds in advance of the first log line and it will accurately be the start of subsequent log chunks.

config/local_defaults.go
```diff
@@ -30,24 +30,17 @@ const configVersion = uint32(4)
...
 //
-// Add fields or change defaults in a new defaultLocalV# instance,
-// bump the version number (configVersion), and add appropriate migration and tests.
-//
-// EXCEPTION: If you are adding a new value whose default value is it's implicit default value
-//            it is safe / acceptable to add it without bumping the Config version.
-//
-// CLARIFICATION:
+// New fields may be added to the current defaultLocalV#. If they have
+// a value other than the type default zero value add them to the
```
Karmastic: Migration not required.

logging/cyclicWriter.go
Karmastic: Sure but its value as the start of the last log chunk seems dubious.  Don't really care since i would just use the NOW time and assume it starts at the previous archive's NOW.

config/local_defaults.go
algobolson: Someone could have a V(N-1) config file in place and I think they would need `migrate()` to get an actual value for a non-zero config in V(N)

Karmastic: So after several iterations of getting this to work as desired, that is not true.

The current consumption of config.json looks like this:

- Start with LatestDefaults
- Reset Version to 0
- Load config.json (which overwrites explicitly-defined values)
- Process migrations

This does exactly what everyone assumed would happen when config.json was modified in the past.

algobolson:
implemented LogArchiveMaxAge

config/local_defaults.go
```diff
@@ -288,6 +283,9 @@ func migrate(cfg Local) (newCfg Local, err error) {
 		if newCfg.PriorityPeers == nil {
 			newCfg.PriorityPeers = map[string]bool{}
 		}
+		if newCfg.LogArchiveName == defaultLocalV3.LogArchiveName {
```
Karmastic: This can be removed - migration not required for new values.

daemon/algod/server.go
```diff
@@ -65,10 +65,19 @@ func (s *Server) Initialize(cfg config.Local) error {
...
+	liveLog := filepath.Join(s.RootPath, "node.log")
+	archive := filepath.Join(s.RootPath, cfg.LogArchiveName)
 	fmt.Println("Logging to: ", liveLog)
-	logWriter := logging.MakeCyclicFileWriter(liveLog, archive, cfg.LogSizeLimit)
+	var maxLogAge time.Duration
+	var err error
+	if cfg.LogArchiveMaxAge != "" {
+		maxLogAge, err = time.ParseDuration(cfg.LogArchiveMaxAge)
+		if err != nil {
+			s.log.Errorf("invalid config LogArchiveMaxAge: %s", err)
```
Karmastic: Can we s.log.Fatal() here since this should not be silently ignored - it's an ill-formed config.json file.  We otherwise log.Fatal() in algod/main.go if we encounter other errors loading the config file (which this would do is were were decoding it as a duration when loading the json).
> log.Fatalf("Cannot load config: %v", err)

PR #170, opened 2019-07-19
fix TestDBConcurrencyRW

algobolson:
## Summary

Use a channel instead of an atomic busy-wait to sync test thread starts.
Clean up sqlite db files. (repeated testing eventually failed due to out-of-disk error)

## Test Plan

Ran the test repeatedly for a while, it used to hit the intermittent crash after a few minutes but now I haven't seen a crash in much longer.
`while go test -count=50 github.com/algorand/go-algorand/util/db; do date; done`

### DIFF
--- a/util/db/dbutil_test.go
+++ b/util/db/dbutil_test.go
@@ -26,6 +26,7 @@ import (
 	"sync"
 	"sync/atomic"
 	"testing"
+	"time"
 
 	"github.com/stretchr/testify/require"
 
@@ -233,6 +234,7 @@ func TestDBConcurrencyRW(t *testing.T) {
 	})
 	require.NoError(t, err)
 
+	started := make(chan struct{})
 	var lastInsert int64
 	var wg sync.WaitGroup
 	wg.Add(1)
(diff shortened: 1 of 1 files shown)


zeldovich:
The cleanup thing is definitely a good idea -- thanks!

Do you have an example output for this test failing or crashing?  I can't seem to reproduce it on my laptop.

algobolson:
It might only reproduce on underpowered Linux hosts (2 cores or fewer) like what we were using for our test host and a small VM I reproduced this on. I think the test design would rely on multiple cores or for the Go runtime to preempt threads and I guess sometimes the Go runtime wasn't good enough at preempting the busy-wait threads.

zeldovich:
Ah, makes sense.

Can you just check that the test still catches the intended bug?  The test should fail if the following change is applied:

    --- a/util/db/dbutil.go
    +++ b/util/db/dbutil.go
    @@ -227,6 +227,8 @@ func URI(filename string, readOnly bool, memory bool) string {
            }
            if memory {
                    uri += "&mode=memory"
    +       }
    +       if readOnly || memory {
                    uri += "&cache=shared"
            }
            return uri

algobolson:
yup. that promptly fails.

PR #172, opened 2019-07-22
transaction grouping support

zeldovich:


data/pools/transactionPool.go
```diff
@@ -187,24 +201,34 @@ func (pool *TransactionPool) test(t transactions.SignedTxn) error {
...
 
-// Remember stores the provided transaction
+// RememberOne stores the provided transaction
+// Precondition: Only RememberOne() properly-signed and well-formed transactions (i.e., ensure t.WellFormed())
+func (pool *TransactionPool) RememberOne(t transactions.SignedTxn) error {
+	return pool.Remember([]transactions.SignedTxn{t})
+}
+
+// Remember stores the provided transaction group
 // Precondition: Only Remember() properly-signed and well-formed transactions (i.e., ensure t.WellFormed())
```
algoradam: Not related to this PR's changes, but do you know if this precondition (check WellFormed before calling Remember) is still necessary?

zeldovich: That comment is certainly misleading.  The well-formedness doesn't matter much.  It's critical that the transaction signature has been checked, though.  The pool assumes the signature is already valid (doesn't do its own check), and the block evaluator subsequently assumes any transactions coming from the pool have good signatures.

algoradam:
To be sure I understand correctly, would the following be an accurate addition to the spec describing this change?

A transaction may include a "Group" field (msgpack tag "grp"), a 32-byte hash that specifies what _transaction group_ the transaction belongs to.
Informally, a transaction group is an ordered list of transactions that, in order to be confirmed at all, must all be confirmed together, in order, in the same block. The "Group" field in each transaction is set to what is essentially the hash of the list of transaction hashes in the group, except to avoid circularity, when hashing a transaction it is hashed with its "Group" field omitted. In this way a user wanting to require transaction A to confirm if and only if transactions B and C confirm can take the hashes of transactions A, B, and C (without the "Group" field set), hash them together, and set the "Group" field of all three transactions to that hash before signing them.

More formally, when evaluating a block, consider the $i$th and $(i+1)$th transaction in the payset to belong to the same _transaction group_ if the "Group" fields of the two transactions are equal and nonzero.
The block may now be viewed as an ordered sequence of transaction groups, where each transaction group is a contiguous sublist of the payset consisting of one or more transactions with equal "Group" field.
For each transaction group where the transactions have nonzero "Group", compute the _TxGroup hash_ as follows:
	1. Take the hash of each transaction in the group but with its "Group" field omitted.
	2. Hash this ordered list of hashes -- more precisely, hash the canonical msgpack encoding of a struct with a field "txlist" containing the list of hashes, using "TG" as domain separation prefix.
If the TxGroup hash of any transaction group in a block does not match the "Group" field of the transactions in that group (and that "Group" field is nonzero), then the block is invalid.

zeldovich:
> To be sure I understand correctly, would the following be an accurate addition to the spec describing this change?

Yes, that's accurate -- thank you for reverse-engineering that out of the code!

algoradam:
What's the intended flow for signing a group of transactions where e.g., one of the transactions must be signed by person A and another must be signed by person B? `goal clerk sign` will abort if you give it a file with multiple transactions and it can't sign them all; `algokey` will sign all the transactions with whichever key you give it (including the transactions for which that's the wrong key). There doesn't seem to be a `goal` command to split apart a file with multiple transactions in it.

zeldovich:
> What's the intended flow for signing a group of transactions where e.g., one of the transactions must be signed by person A and another must be signed by person B? `goal clerk sign` will abort if you give it a file with multiple transactions and it can't sign them all; `algokey` will sign all the transactions with whichever key you give it (including the transactions for which that's the wrong key). There doesn't seem to be a `goal` command to split apart a file with multiple transactions in it.

Good point -- I'll add a goal command to split a transactions file into one file per txn.

(I was using msgpacktool to do this, but that's not something we should be suggesting to all of our users.)

data/txHandler.go
```diff
@@ -198,17 +204,36 @@ func (handler *TxHandler) asyncVerifySignature(arg interface{}) interface{} {
...
-	var unverifiedTxn transactions.SignedTxn
-	err := protocol.Decode(rawmsg.Data, &unverifiedTxn)
-	if err != nil {
-		logging.Base().Warnf("Received a non-decodable txn: %v", err)
+	dec := protocol.NewDecoderBytes(rawmsg.Data)
+	ntx := 0
+	unverifiedTxGroup := make([]transactions.SignedTxn, 1)
+	for {
+		if len(unverifiedTxGroup) == ntx {
+			n := make([]transactions.SignedTxn, len(unverifiedTxGroup)*2)
```
algoradam: To be sure I'm not missing something subtle here: Instead of decoding and calling `append()`, you're growing the array yourself so that you can decode directly into `&unverifiedTxGroup[ntx]`, saving a memcpy, right? Or am I missing something?

zeldovich: Yes, that's all that's going on here.

I believe the code was originally carefully optimized to the level where an extra copy of the `SignedTxn` mattered, so I tried to preserve that property.

algoradam:
It looks like there's currently no way to use goal to send a transaction group -- `goal clerk rawsend` tries to broadcast transactions individually.

zeldovich:
> It looks like there's currently no way to use goal to send a transaction group -- `goal clerk rawsend` tries to broadcast transactions individually.

Good catch (thought I did this, but clearly not).  I fixed rawsend to respect groups.

PR #178, opened 2019-07-24
switch from dep to go modules

zeldovich:
This allows checking out and building the go-algorand repo in any directory (doesn't have to be under $GOPATH).  This also means that it's possible to have multiple checkouts, and it's possible to check out a forked repo of go-algorand and have it build as expected.


Vervious:
The swagger refactor on `go-swagger:master` must have broken/fixed the (indirect) `strfmt` annotation... It looks like the following works:

```
...
	// Note is a free form data
	//
	// required: true
	// swagger:strfmt binary
	Note string `json:"noteb64,omitempty"`
...
```

generating:
```
        "noteb64": {
          "description": "Note is a free form data",
          "type": "string",
          "format": "binary",
          "x-go-name": "Note"
        },
```

(for some reason, the ordering of `required:` and `swagger:strfmt` matter here. In addition, if Note is of type `[]byte` the new swagger parser also complains and throws an error. I haven't found any other workaround).

perhaps this means we now need to manually base64-encode byte slices when sticking them into swagger models...

Vervious:
Perhaps the right approach here is to manually maintain the `swagger.json` files... We don't modify them much, and the upkeep cost of `go-swagger` has been extraordinary.

Vervious:
Let me dig into `go-swagger`, see what is broken with `strfmt` and perhaps make a PR

zeldovich:
Thank you for tracking it down!

To the extent the swagger file is used (and I think the Java SDK does actually use it), it seems worth generating it.  Otherwise we will surely forget to revise it when we change the Go code.

If we can't find a better workaround, then switching the binary types into Go type `string` and manually annotating them as `// swagger:strfmt base64` seems not too bad.  This means we'll have to also manually do the base64 encoding/decoding, since the JSON codec will not base64-encode the Go `string` type (but does base64-encode the `[]byte` type).

Vervious:
Doing the base64 conversion isn't that bad - there are only a couple of places.

Speaking of `// swagger:strfmt base64` - I recall intentionally choosing `//swagger:strfmt binary` over `// swagger:strfmt base64` and `// swagger:strfmt byte`... but never documented it (and have since forgotten). Let me also look into that again. probably has to do with `swagger-codegen` support in java for the java sdk.

Makefile
```diff
@@ -26,136 +26,107 @@ GOLDFLAGS_BASE  := -X github.com/algorand/go-algorand/config.BuildNumber=$(BUILD
...
+E2E_TEST_SOURCES := $(shell cd test/e2e-go && go list ./...)
 
 default: build
 
 # tools
 
 fmt:
-	cd $(SRCPATH) && \
-		go fmt `go list ./... | grep -v /vendor/`
+	go fmt `go list ./... | grep -v /vendor/`
```
winder: We don't need `grep -v /vendor/` here: https://github.com/algorand/go-algorand/pull/178/files

zeldovich: Hah, yeah, thanks!  That's cleaner.

Vervious:
We should be able to push this forward now, https://github.com/zeldovich/go-algorand/pull/1

PR #179, opened 2019-07-24
pass -trimpath for more deterministic builds (go 1.13+)

zeldovich:
Avoids including the full pathname of the build directory in the resulting
binaries, which makes it more likely that users can reproduce our builds
bit-for-bit.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,6 +16,7 @@ EXTLDFLAGS := -static-libstdc++ -static-libgcc
 endif
 
 GOTAGS          := --tags "sqlite_unlock_notify sqlite_omit_load_extension"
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -16,6 +16,7 @@ EXTLDFLAGS := -static-libstdc++ -static-libgcc
 endif
 
 GOTAGS          := --tags "sqlite_unlock_notify sqlite_omit_load_extension"
+GOTRIMPATH	:= $(shell go help build | grep -q .-trimpath && echo -trimpath)
```
tsachiherman: it's an interesting way to detect if `trimpath` is supported by the current go compiler..

PR #182, opened 2019-07-26
Fix missing connection metadata

winder:
# Summary

When telemetry is disabled connection information is incomplete when communicating with relay nodes.

## Test Plan

Before and after the fix, started a node with telemetry disabled and see that the error is reproduced and then fixed.

Missing metadata before fix:
before_fix

Metadata after fix:
after_fix

### DIFF
--- a/logging/telemetry.go
+++ b/logging/telemetry.go
@@ -85,7 +85,7 @@ func EnsureTelemetryConfigCreated(configDir *string, genesisID string) (Telemetr
 		configPath, err = config.GetConfigFilePath(loggingFilename)
 		if err != nil {
 			cfg := createTelemetryConfig()
-			initializeConfig(&cfg)
+			initializeConfig(cfg)
 			return createTelemetryConfig(), true, err
 		}
 	} else {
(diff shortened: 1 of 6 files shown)


winder:
@tsachiherman updated based on your feedback. I think using a global is fine because there is a single config file. When we start the `externalized telemetry` project that may be revisited.

logging/telemetry.go
```diff
@@ -87,6 +84,8 @@ func EnsureTelemetryConfigCreated(configDir *string, genesisID string) (Telemetr
 		var err error
 		configPath, err = config.GetConfigFilePath(loggingFilename)
 		if err != nil {
+			cfg := createTelemetryConfig()
+			initializeConfig(cfg)
 			return createTelemetryConfig(), true, err
```
tsachiherman: I think that you meant:
`return cfg, true, err`

winder: Yes, thank you.

algobolson:
Is there any privacy implication in disabling telemetry that implies, "I don't want to be identifiable in other nodes' telemetry" ? Should a telemetry-disabled node send "" as its GUID?

winder:
@algobolson I'm not sure. Telemetry is disabled by default, so we can't read into that fact as a user taking any specific action. I think that because the GUID is a random UUID, this is a good balance of respecting privacy and useful network diagnostic data.

Karmastic:
> @algobolson I'm not sure. Telemetry is disabled by default, so we can't read into that fact as a user taking any specific action. I think that because the GUID is a random UUID, this is a good balance of respecting privacy and useful network diagnostic data.

Yeah this is a squishy area.  We're only using this GUID to correlate network connections from the same host (the whole point of this PR).  This doesn't leak anything that is PII.

PR #184, opened 2019-07-26
sanitize telemetry headers

winder:
# Summary

Add some basic sanitization to a few telemetry headers which are passed directly to the telemetry DB. I don't think there's any real exploit here, but David made the suggestion and it made sense to check just in case.

## Test Plan

Added some tests to make sure the sanitize logic worked the way I expected, and ran algod to see it working in action.


logging/telemetryConfig.go
```diff
@@ -120,6 +130,22 @@ func (cfg TelemetryConfig) getInstanceName() string {
 	return fmt.Sprintf("%s:", pathHashStr[:16])
 }
 
+func SanitizeTelemetryString(input string, maxParts int) string {
+	parts := strings.SplitN(input, ":", maxParts)
```
tsachiherman: Add a limit on the length of the input string.

logging/telemetryConfig.go
```diff
@@ -120,6 +130,22 @@ func (cfg TelemetryConfig) getInstanceName() string {
...
+		if reg != nil {
+			parts[i] = reg.ReplaceAllString(parts[i], "")
+		}
+
+		// Limit length to support up to an ipv6 address
+		if len(parts[i]) > 39 {
+			parts[i] = parts[i][:39]
+		}
+	}
+	return strings.Join(parts, ":")
```
tsachiherman: add a limit on the length of the total combined output string.

logging/telemetryConfig.go
```diff
@@ -120,6 +130,22 @@ func (cfg TelemetryConfig) getInstanceName() string {
...
+func SanitizeTelemetryString(input string, maxParts int) string {
+	parts := strings.SplitN(input, ":", maxParts)
+	for i := range parts {
+		// Limit to alphanumeric, and _-
+		if reg != nil {
+			parts[i] = reg.ReplaceAllString(parts[i], "")
+		}
+
+		// Limit length to support up to an ipv6 address
+		if len(parts[i]) > 39 {
```
tsachiherman: instead of using the magic number 39 ( which is well described the line above ), use the following
```const ipv6AddressLength := 39```
instead.

File: logging/telemetryConfig.go [hunk 4]
```diff
@@ -120,6 +130,22 @@ func (cfg TelemetryConfig) getInstanceName() string {
 	return fmt.Sprintf("%s:", pathHashStr[:16])
 }
 
+func SanitizeTelemetryString(input string, maxParts int) string {
+	parts := strings.SplitN(input, ":", maxParts)
+	for i := range parts {
+		// Limit to alphanumeric, and _-
+		if reg != nil {
```
tsachiherman: nit: move the `reg != nil` outside the for loop. In practice, it will never happen, but we have to do the check.

winder: I'm not sure it's worth it, or what the error handling would even be. The current logic just skips the regex if it doesn't exist, but also supports removing colons from last part of the string split, so `guid-pattern:removes-the-:colon` -> `guid-pattern:removes-the-colon`.

logging/telemetryConfig.go
winder: There's an implicit limit of `39 * parts + parts - 1` which I think is good since we don't have any strict sanitization requirements at the moment.

logging/telemetryConfig.go
winder: Actually, we definitely can't remove colons, the IPv6 address format will have lots of them!
I'll apply the regex replace all outside the loop.

winder:
@tsachiherman thanks, updated PR based on your feedback. I agree that we may want something better / more specific to the fields, but I'm not sure what that would be.

logging/telemetryConfig.go
```diff
@@ -120,6 +131,26 @@ func (cfg TelemetryConfig) getInstanceName() string {
...
+func SanitizeTelemetryString(input string, maxParts int) string {
+	// Truncate to a reasonable size.
+	input = input[:maxParts * 100]
+
+	// Limit to alphanumeric, and _-
+	if reg != nil {
+		input = reg.ReplaceAllString(input, "")
+	}
+
+	parts := strings.SplitN(input, ":", maxParts)
```
algobolson: are you trying to split `host:port` ? IPv6 addresses have ':' in them. maybe `net.SplitHostPort()` ?

winder: The main reason for this is the "TelemetryHost" field is in a `GUID:Name` format, and I wanted to apply a length threshold to each separately. For sanitizing `AddressHeader` maxParts is 1, so  it should be unchanged.

winder: Added an ipv6 test case

logging/telemetryConfig_test.go
```diff
@@ -90,3 +90,23 @@ func Test_CreateSaveLoadTelemetryConfig(t *testing.T) {
...
+		expected string
+		parts    int
+	}
+
+	tests := []testcase{
+		{"no:change", "no:change", 2},
+		{"@@@@@@@remove(-some*&!-garbage", "remove-some-garbage", 1},
+		{"@@@@@@@remove(-some*&!-garbage:times-%two", "remove-some-garbage:times-two", 2},
+		{"this.needs.to.be.truncated.because.it.is.way.too.long", "this.needs.to.be.truncated.because.it.i", 1},
+		{"2001:0db8:85a3:0000:0000:8a2e:0370:7334", "2001:0db8:85a3:0000:0000:8a2e:0370:7334", 1},
```
algobolson: this works as it passes through strings.Join(["2001","0db8:85a3:0000:0000:8a2e:0370:7334"], ":")
if the intention was to split("{IPv6 addr}:{something else}") and make sure neither side was too long, it isn't that.
Does this even matter? Is the left hand side likely to be an IPv6 addr?
e.g. this test case should fail:
`{"2001:0db8:85a3:0000:0000:8a2e:0370:7334:whaarrrrrrgarrrrbleeeeeeeeeefoobarbazbearquxzapzowyow", "2001:0db8:85a3:0000:0000:8a2e:0370:7334:whaarrrrrrgarrrrbleeeeeeeeeefoobarbazbe", 1},`

winder: I think it matters because this is where I'm using the sanitizer:
```diff
func getCommonHeaders(headers http.Header) (otherTelemetryGUID, otherInstanceName, otherPublicAddr string) {
- 	otherTelemetryGUID = headers.Get(TelemetryIDHeader)
- 	otherInstanceName = headers.Get(InstanceNameHeader)
- 	otherPublicAddr = headers.Get(AddressHeader)
+ 	otherTelemetryGUID = logging.SanitizeTelemetryString(headers.Get(TelemetryIDHeader), 1)
+ 	otherInstanceName = logging.SanitizeTelemetryString(headers.Get(InstanceNameHeader), 2)
+ 	otherPublicAddr = logging.SanitizeTelemetryString(headers.Get(AddressHeader), 1)
	return
}
```

You're right that it is a little curious that these two tests pass: 
`{"2001:0db8:85a3:0000:0000:8a2e:0370:7334", "2001:0db8:85a3:0000:0000:8a2e:0370:7334", 1}`
and
`{"2001:0db8:85a3:0000:0000:8a2e:0370:7334", "2001:0db8:85a3:0000:0000:8a2e:0370:7334", 8}`

One thing to note is that in the case where max parts is 1, the result is an array whose single element is the input. With these tests you can validate that the ipv6 address truncates anything > 39 characters, i.e. this test passes:
`{"2001:0db8:85a3:0000:0000:8a2e:0370:7334:9999", "2001:0db8:85a3:0000:0000:8a2e:0370:7334", 1}`

winder: This is all N/A now. No longer trying to parse the inputs, just truncate them if they're big.

PR #192, opened 2019-07-29
[GOAL2-761] Provide context for WsFetcherService log messages

tsachiherman:
## Summary

The log files contains multiple entries for
`WsFetcherService: no fetcher registered for tag (UT)
`

This is believed to be the result of stale fetcher requests, but we don't have the data to back it up. This change augment the existing logging with the decoded message information, which would allow us to reconstruct the scenario.

## Test Plan

* Run the existing unit test. These already contains a functionality testing.
* Attempt to reproduce the hypothesized issue on a private deployed network.


rpcs/wsFetcher.go
```diff
@@ -139,66 +98,48 @@ func (wsf *WsFetcher) NumPeers() int {
...
 
-// GetBlockBytes implements FetcherClient
-func (w *wsFetcherClient) GetBlockBytes(ctx context.Context, r basics.Round) ([]byte, error) {
-	listen := w.getChForRound(uint64(r))
-	if listen == nil {
-		return nil, fmt.Errorf("wsFetcherClient(%d): preconnection closed", r)
-	}
+	defer func() {
+		cancelFunc()
+		w.mu.Lock()
```
Karmastic: Is this an unbalanced Lock() call?

tsachiherman: no. see `defer w.mu.Unlock()` above it.

Karmastic: This results in valid execution but is awkward to read.  Not worth rewriting to make it less awkward - how about a comment in the defer func() pointing out there's a defer unlock on the stack already?

rpcs/wsFetcherService.go
```diff
@@ -27,50 +31,108 @@ import (
...
+
+	waitKey := makePendingRequestKey(uniPeer, basics.Round(resp.Round), msg.Tag.Complement())
 	fs.mu.RLock()
-	f := fs.fetchers[msg.Tag]
+	f, hasWaitCh := fs.pendingRequests[waitKey]
 	fs.mu.RUnlock()
-	if f == nil {
-		fs.log.Infof("WsFetcherService: no fetcher registered for tag (%v)", msg.Tag)
+	if !hasWaitCh {
+		if resp.Error == "" || len(resp.BlockBytes) == 0 {
```
Vervious: why do we have this conditional?

rpcs/wsFetcher.go
tsachiherman: sure. I'll add that!

rpcs/wsFetcherService.go
tsachiherman: The difference is the error messages. One contains the error and len(resp.BlockBytes) while the other does not.

Vervious: The better question is why do we print the length and error if either are empty... I'm missing something here. The else only triggers if both are non-empty which seems like the best time to print it, but instead it's the opposite?

tsachiherman: yes, fixed.

PR #196, opened 2019-07-30
Rel beta build fix

algobolson:
updates to build_release.sh and friends for `beta` channel


algobolson:
Yeah, I was running lots of builds, and sometimes they would error and hang, and while I was at it I figured out how to have them error and exit.

PR #167, opened 2019-07-17
Improved keyreg output on backend API

mxmauro:
## Summary

This patch outputs the correct information of keyreg transactions.

Currently, keyreg tx's outputs a bogus `payment` sub-object.

## Test Plan

We compiled the code and launched a dev network. After executing:

`curl http://127.0.0.1:8080/v1/block/94154 -H "X-Algo-API-Token: #put-token-here#"` we get the expected values:

```
...
{
    "_id" : ObjectId("5d2f74fa26104a591b5fbc04"),
    "index" : 3225366,
    "txid" : "A2OMCBM7JTGVVXDJWYEP6OTBJCFGTGKRL7RYWVAOA2EPIDXGGZCA",
    "type" : "keyreg",
    "address" : "VJ6NE2QAWYS2JP7GFO3BMX5THTBZKT7GEHIG6GFZCMYSET6QU27G2PKOFU",
    "from_index" : 3,
    "fee" : 1000,
    "round" : 94154,
    "first_round" : 94153,
    "last_round" : 95153,
    "keyreg" : {
        "votekey" : "okryFvieL+YNRjJwSnxzl8BdL9rajZe6csjrhWNWCtc=",
        "selkey" : "xF1VTDoyWTqZKHpIkxDlLVjKaQnXuUerK3J7aniMHts=",
        "votefst" : 94150,
        "votelst" : 1500150,
        "votekd" : 10000
    },
    "timestamp" : 1560798075
}
...
```

With this modification, anyone can know if an account becomes online or offline too.

PR #206, opened 2019-07-31
Encode key registration transactions over REST API

tsachiherman:
## Summary

Encode key registration transactions over REST API.

## Notes

This is a modified implementation of the logging mechanism used in 
https://github.com/algorand/go-algorand/pull/167

If approved, the goal is to accept this one and close both.


daemon/algod/api/server/v1/handlers/errors.go
```diff
@@ -17,17 +17,21 @@
 package handlers
 
 var (
```
tsachiherman: The changes here aren't that substantial; I just reorder the errors alphabetically while adding three.

tsachiherman:
I'm not sure if @mxmauro has a reviewer permissions here, but would be happy to have his feedback.

Vervious:
Looks like we want to either merge this PR, or merge #167 with Tsachi's changes pulled in. Up to @mxmauro which way is preferable

Vervious:
Hi, I accidentally squashed your commits so lost authorship recognition. I will rectify this somehow @mxmauro @scnale

mxmauro:
Don't worry @Vervious . I'll wait for the release to index those txs too :)

Vervious:
#210

PR #209, opened 2019-08-01
Revert "Encode key registration transactions over REST API"

Vervious:
Reverts algorand/go-algorand#206


Vervious:
Squashing loses the commit history

PR #213, opened 2019-08-02
[GOAL2-793] HTTP request logging

tsachiherman:
## Summary

Add HTTP request logging to algod. The logging here would allow us to generate a CLF style logs, which could be processed by any standard web log analysis tool.
Common Log Format [https://en.wikipedia.org/wiki/Common_Log_Format]


network/requestLogger.go
```diff
@@ -0,0 +1,126 @@
...
+	"net"
+	"net/http"
+	"strings"
+	"sync"
+
+	"github.com/algorand/go-algorand/logging"
+	"github.com/algorand/go-algorand/logging/telemetryspec"
+)
+
+// RequestLogger is a middleware helps logging all the incoming requests.
```
egieseke: Please specify that the Request logger is for all incoming "http" requests.

tsachiherman: sure.

tsachiherman:
> Is there a unit test for the requestLogger?

hmm. I'm not sure exactly what need to be tested.
I have manually verified that the rl.log.EventWithDetails method is being called as needed by adding a "rl.log.Infof" and running the existing unit tests ( which invokes the functionality ).

Can you think of any particular aspects of this implementation which you'd like to be tested ?

tsachiherman:
> Anything you can think of to add in the area of tests?
I can add a unit test that shows that we get a telemetry event on a http call.

> Any chance we can make this optional?
Yes, we can. But it would defeat the purpose if we won't have that information from our relays. For non-relays, this code doesn't get executed anyway. I'll add an option to turn off this functionality.

> Do we really want these sent as telemetry (they won't go the log file - just telemetry)?
Adding these to the log file is great for local debugging ( of this feature.. ), but not so useful otherwise. It's the trends analysis of the relay farm that I'm after.. and log files won't help me to get there.

test/testdata/configs/config-v4.json
```diff
@@ -15,6 +15,7 @@
     "EnableIncomingMessageFilter": false,
     "EnableMetricReporting": false,
     "EnableOutgoingNetworkMessageFiltering": true,
+    "EnableRequestLogger": false,
```
Karmastic: This should be added to config.json.example

network/requestLogger.go
```diff
@@ -0,0 +1,130 @@
...
+		},
+	}
+	return rl
+}
+
+// this is the http entry point for the request logger.
+func (rl *RequestLogger) ServeHTTP(writer http.ResponseWriter, request *http.Request) {
+	trackingWriter := rl.trackingWritersPool.Get().(*trackingResponseWriter)
+	trackingWriter.Reset(writer)
+	defer func() {
```
Karmastic: Is there a reason we need to defer this rather than just execute it after the call to ServeHTTP()?  In practice, we don't recover from panics.

test/testdata/configs/config-v4.json
tsachiherman: Yup. It was added to config.json.example. See 3rd file from the top.

network/requestLogger.go
tsachiherman: Interesting point. I'm not sure it's correct. if you'll look into server.go in the go codebase, you'll see the following:
```
// Serve a new connection.
func (c *conn) serve(ctx context.Context) {
	c.remoteAddr = c.rwc.RemoteAddr().String()
	ctx = context.WithValue(ctx, LocalAddrContextKey, c.rwc.LocalAddr())
	defer func() {
		if err := recover(); err != nil && err != ErrAbortHandler {
			const size = 64 << 10
			buf := make([]byte, size)
			buf = buf[:runtime.Stack(buf, false)]
			c.server.logf("http: panic serving %v: %v\n%s", c.remoteAddr, err, buf)
		}
		if !c.hijacked() {
			c.close()
			c.setState(c.rwc, StateClosed)
		}
	}()
...
```

it looks to me as if it would print the error and close the connection.

Karmastic: What I meant is that _we_ (above ServeHTTP on the stack) do not recover, so the defer() will only be useful if downsteamHandler.ServeHTTP() panics; but we'll end up exiting anyway because we don't recover.  Since the panic() is simply logged, it means our defer() isn't required to ensure that block of code runs.  Not a big concern, just wondering.

test/testdata/configs/config-v4.json
Karmastic: Yup, sorry, not sure how I missed that!

PR #214, opened 2019-08-02
Relay pinned value in partitionPolicy()

### LINKED ISSUE #50: Next bundles need to include blocks
Summary: Next bundles which don’t have  blocks can negatively impact liveness. Agreement code needs to be fixed for these bundles to contain these blocks. Since this affects liveness, resolving this will reduce stall risk.

- [ ] First task is to implement this solution (sending blocks) ~ 3 points
- [ ] Second is to QA and deploy ~ 2 points

Vervious:
Fixes #50 
Previously, if there was no staged value in the same period as the freshest bundle, in `partitionPolicy` we would only relay the bundle. This is a very minor deviation from the spec.

This PR makes it so that in this case, we relay the pinned block (if it exists). This:
- aligns us with the spec
- is a minor liveness optimization
- if we make cert thresholds the freshest bundle, makes it easier to ensure that everyone gets the block


agreement/actions.go
```diff
@@ -113,6 +113,9 @@ func (a networkAction) String() string {
 	if a.t() == ignore || a.t() == disconnect {
 		return fmt.Sprintf("%v: %5v", a.t().String(), a.Err)
 	}
+	if a.Tag == protocol.ProposalPayloadTag {
+		return fmt.Sprintf("%v: %2v: %5v", a.t().String(), a.Tag, a.CompoundMessage.Proposal.value())
```
tsachiherman: nit: use %s for strings and %v for objects.

Vervious: done! There's a lot of cases in `actions.go` where this nit would apply, but let's leave that for another PR.

agreement/player.go
```diff
@@ -421,14 +421,23 @@ func (p *player) partitionPolicy(r routerHandle) (actions []action) {
 		bundlePeriod = b.Period
```
algoradam: This PR seems to make the comment a few lines above this no longer true, right? (lines 402 through 414)

agreement/proposalStore.go
```diff
@@ -333,6 +334,13 @@ func (store *proposalStore) handle(r routerHandle, p player, e event) event {
 		se.Committable = ea.Assembled
 		se.Payload = ea.Payload
 		return se
+	case readPinned:
```
algoradam: Add this event to the comment that documents `handle` (lines 138 - 181)

algoradam:
Let me confirm that I understand correctly:
On resynchronization, we send our freshest bundle and whatever block goes with it, if we have the block. There isn't a convenient way for the player to reach inside the proposalStore and directly ask it for the payload corresponding to a given proposal-value, but we can ask it for the staged payload and (as of this PR) the pinned payload.

- If our freshest bundle is a soft bundle, then the staged payload (if we have one) will be the block that corresponds to this soft bundle
- If our freshest bundle is a next-bundle (i.e., step > cert), then the pinned payload (if we have one) will be the block that corresponds to this next-bundle
- If our freshest bundle is a cert-bundle, then it doesn't matter what we do, because we either don't have the block or aren't on this round

So this PR sends the pinned payload on resynchronization if we don't have a staged payload. In all cases the payload that gets sent, if any, will correspond to the freshest bundle.

The case that this PR fixes: If we move into a new period via a next-bundle for a value and we don't see a soft quorum for this value in the new period but we do have the payload, then we'll have a pinned value but no staged value and before this PR we would fail to include the payload when resynchronizing.

Is my understanding correct?

Vervious:
Your understanding is correct!
I will update the PR with appropriate comments.

agreement/router.go
```diff
@@ -264,3 +264,10 @@ func stagedValue(p0 player, h routerHandle, r round, p period) stagingValueEvent
 	e := h.dispatch(p0, qe, proposalMachineRound, r, p, 0)
 	return e.(stagingValueEvent)
 }
+
+// pinnedValue gets the current pinned value for some (r, p)
```
derbear: Nit: change the comment (since no period is involved here).

agreement/events.go
```diff
@@ -591,6 +594,30 @@ func (e stagingValueEvent) ComparableStr() string {
 	return e.String()
 }
 
+type pinnedValueEvent struct {
+	// Round is the round for which to query the current pinned value
+	Round round
+
+	// Proposal holds the staging value itself.
```
derbear: Nit: holds the *pinned* value itself.

agreement/events.go
```diff
@@ -591,6 +594,30 @@ func (e stagingValueEvent) ComparableStr() string {
...
+	Proposal proposalValue
+	// Payload holds the payload, if one exists (which is the case if Committable is set).
```
derbear: Nit: if *PayloadOK* is set

agreement/events.go
```diff
@@ -591,6 +594,30 @@ func (e stagingValueEvent) ComparableStr() string {
...
+	Payload proposal
+	// PayloadOK is set if and only if the staging value is committable.
```
derbear: Nit: is set if and only if *a proposal was received for the pinned value*.

agreement/player_test.go
```diff
@@ -2260,4 +2359,100 @@ func TestPlayerRegression_EnsuresCertThreshFromOldPeriod_8ba23942(t *testing.T)
...
 }
 
+func TestPlayerAlwaysResynchsPinnedValue(t *testing.T) {
+	// a white box test that checks the pinned value is relayed even it is not staged in the period corresponding to the freshest bundle
+	const r = round(209)
+	const p = period(12)
+	pWhite, pM, helper := setupP(t, r, p-2, soft)
+	payload, pV := helper.MakeRandomProposalPayload(t, r)
+
+	// store a payload for p - 1
```
algoradam: Nit: "p - 1" should say "10"

algoradam:
Looks good to me once the already-mentioned documentation bugs are fixed

Vervious:
@derbear can I get one more sanity check sign off

agreement/proposalStore.go
```diff
@@ -333,6 +339,13 @@ func (store *proposalStore) handle(r routerHandle, p player, e event) event {
...
+		se := e.(pinnedValueEvent)
+		ea := store.Assemblers[store.Pinned] // If pinned is bottom, commitable = false, payload = bottom
```
derbear: By committable, you mean PayloadOK?

agreement/player.go
```diff
@@ -398,20 +398,17 @@ func (p *player) partitionPolicy(r routerHandle) (actions []action) {
...
+	// On resynchronization, first try relaying the staged proposal from the same period as
+	// the freshest bundle. If that does not exist, for instance if we saw two next quorums in a row,
+	// then we fall back to relaying the pinned value, for liveness.
+	// One specific scenario where this is essential, assuming we handle ensure digest asynchronously:
+	// - Let the majority of honest nodes cert vote, and then see a next value quorum, and enter p + 1.
+	// - They see another next value quorum, and enter p + 2.
+	// - The minority of honest nodes see a certThreshold (but without a block), in period p. Assume that
+	//   they are partitioned from the majority of honest nodes, until the majority reach p + 2.
+	// - The minority already has the freshest bundle, so will not advance to period p + 2. However, the
+	//   majority will also filter out the cert threshold (due to a stale period). The only way to resolve
```
derbear: I think this is an unrelated bug, but perhaps we should not filter out cert thresholds from stale periods. Perhaps we should use the same logic for determining the freshest bundle as we do in `bundleFresh` (inside of `voteAggregator`); that way, we precisely filter out the bundles that we know to be stale. (There is no change in denial-of-service characteristics here, as it is always trivial to generate a bundle with a malformed vote for a next period, and that bundle will always be fresh.) It might be possible to handle bundle filtering at the `voteTrackerRound`, since that's the machine that knows about the freshest bundle.

This is unrelated to this bug so probably the thing to do is to make an issue for it.

Vervious: I will probably rope this into my ensure digest PR

derbear: Sounds good to me.

PR #216, opened 2019-08-05
[GOAL2-901] Fix algoh

Karmastic:
## Summary

Running algod hosted (`goal node start -H` or setting `RunHosted` in config file) failed.
The recent `goal node create` feature caused this by having the `-l <listen-ip>` argument passed to `algoh`, which doesn't support the same arguments as `algod`.

## Test Plan

Verified that it works correctly


Karmastic:
Need to recreate this PR without the makefile change.

PR #217, opened 2019-08-05
[GOAL2-901] Fixes regression caused by unexpected  argument

Karmastic:
## Summary
Running algod hosted (goal node start -H or setting RunHosted in config file) failed.
The recent goal node create feature caused this by having the -l <listen-ip> argument passed to algoh, which doesn't support the same arguments as algod.
Also removed extraneous deps-related files.

## Test Plan
Verified that it works correctly


Karmastic:
@zeldovich Can you confirm we don't need the gopkg.* files anymore (they were confusing goland)?

zeldovich:
Huh, yeah, the `Gopkg.*` files should be deleted.  I definitely deleted them, but they must have re-appeared as a result of some merge, and I didn't notice.  Thanks for removing them.

PR #220, opened 2019-08-05
txpool: avoid rejecting transactions that race with OnNewBlock()

zeldovich:
When a new block for round R shows up in the ledger, it may take a while
until the TransactionPool's OnNewBlock() is called.  Until that happens,
the pool rejects new transactions that are valid starting from round
R+1, because according to the pool, the next round will be R, and the
transaction isn't valid for that round yet.  However, the client that
issued the transaction may have already seen that the latest round is R
(if the client queried the ledger's Latest() round, through the REST API),
and it's just the txpool that hasn't gotten the OnNewBlock() callback yet.

This commit deals with this potential race condition, which might be
the reason for some non-deterministic test failures.

### DIFF
--- a/data/pools/transactionPool.go
+++ b/data/pools/transactionPool.go
@@ -18,6 +18,8 @@ package pools
 
 import (
 	"fmt"
+	"sync"
+	"time"
 
 	"github.com/algorand/go-deadlock"
 
@@ -28,12 +30,14 @@ import (
 	"github.com/algorand/go-algorand/ledger"
 	"github.com/algorand/go-algorand/logging"
 	"github.com/algorand/go-algorand/logging/telemetryspec"
+	"github.com/algorand/go-algorand/util/condvar"
 )
 
 // TransactionPool is a struct maintaining a sanitized pool of transactions that are available for inclusion in
 // a Block.  We sanitize it by preventing duplicates and limiting the number of transactions retained for each account
 type TransactionPool struct {
 	mu                     deadlock.RWMutex
+	cond                   sync.Cond
 	pendingTxns            []transactions.SignedTxn
 	pendingTxids           map[transactions.Txid]transactions.SignedTxn
 	expiredTxCount         map[basics.Round]int
@@ -57,12 +61,17 @@ func MakeTransactionPool(ledger *ledger.Ledger, transactionPoolStatusSize int, l
 		statusCache:    makeStatusCache(transactionPoolStatusSize),
 		logStats:       logStats,
 	}
+	pool.cond.L = &pool.mu
 	pool.recomputeBlockEvaluator()
 	return &pool
 }
 
 // TODO I moved this number to be a constant in the module, we should consider putting it in the local config
 const expiredHistory = 10
+
+// timeoutOnNewBlock determines how long Test() and Remember() wait for
+// OnNewBlock() to process a new block that appears to be in the ledger.
+const timeoutOnNewBlock = time.Second
 
 // NumExpired returns the number of transactions that expired at the end of a round (only meaningful if cleanup has
 // been called for that round)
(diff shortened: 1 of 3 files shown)


data/pools/transactionPool.go
```diff
@@ -179,6 +179,44 @@ func (pool *TransactionPool) Remember(t transactions.SignedTxn) error {
...
+		if errt.Round >= errt.FirstValid {
+			// This is not an error due to the round being too old.
+			break
+		}
+
+		if pool.pendingBlockEvaluator == nil {
+			// We don't have a block evaluator at all; something
+			// else went wrong, and this isn't the place to fix
+			// it.  (The next OnNewBlock callback might recover.)
+			break
```
tsachiherman: You might want to change the returned err object here; it's no longer a dead transaction ( i.e. error 400 Bad Request) but rather ( 500 Internal Server Error )

data/pools/transactionPool.go
```diff
@@ -179,6 +179,44 @@ func (pool *TransactionPool) Remember(t transactions.SignedTxn) error {
...
+		}
+
+		latest := pool.ledger.Latest()
+		if pool.pendingBlockEvaluator.Round() > latest {
```
tsachiherman: this is all evaluated within the same lock ( pool.mu ). Is it really possible ? i.e. it should have been evaluated in test().

data/pools/transactionPool.go
zeldovich: I think this is unreachable anyway, because `TransactionPool.test()` never returns `TxnDeadError` if `pool.pendingBlockEvaluator == nil`.  It just didn't seem right to dereference the possibly-nil `pool.pendingBlockEvaluator` in this block of code without explicitly checking for nil, even if right now the surrounding code is such that it can't happen.

data/pools/transactionPool.go
zeldovich: You're suggesting to move this new blob of code (retrying, etc) into `test()`?  That's probably a good idea; thanks!

data/pools/transactionPool.go
```diff
@@ -179,6 +179,44 @@ func (pool *TransactionPool) Remember(t transactions.SignedTxn) error {
...
+			// Our block evaluator is already on a block after
+			// the latest block in the ledger.  No race to fix.
+			break
+		}
+
+		blk, err0 := pool.ledger.Block(latest)
```
tsachiherman: I'ma bit hesitant about this section; re-doing what should be done by the OnNewBlock() doesn't seems to be correct. If we have an issue there, it should be fixed there. Instead, we should use synchronization primitives ( somehow ) to wait until the OnNewBlock() arrives and making the correct adjustment. ( with a timeout in case of a server failure ).

zeldovich: Indeed, another option would be to wait for the right `OnNewBlock()`.  I was thinking it's a bit more prone to deadlock if the `OnNewBlock()` never arrives for some reason, and the lack of a timed `Cond.Wait()` in Go pushed me over the edge to the code in this PR..

Let me sketch out what this condvar-based approach would look like and see if it's cleaner.

data/pools/transactionPool.go
tsachiherman: if that's the case, could you just add a comment explaining that ?

data/pools/transactionPool.go
tsachiherman: thanks!

zeldovich:
I force-pushed a different implementation of this race avoidance, where Remember() and Test() wait for OnNewBlock().  Seems cleaner.

tsachiherman:
thanks for the changes; I know it wasn't trivial, but the outcome is really nice ;-)

zeldovich:
> TimedWait has some functional constrains in that during the call to c.Broadcast it will release multiple Wait locks which could be an issue ( i.e. in the general use case; for this use case this would work just fine ).

I don't believe `c.Broadcast()` releases any locks -- for that matter, I don't believe that holding the lock is required to call `c.Signal()` or `c.Broadcast()`.

> Also, having TimedWait return whether a timeout has occurred or not could be beneficial for providing improved error handling.

I don't have a strong preference for whether `TimedWait` should or should not attempt to tell the caller whether a timeout occurred.  I'll keep the current behavior for now.

PR #221, opened 2019-08-06
[GOAL2-788] Improve TestDBConcurrencyRW test reliability

tsachiherman:
## Summary

The TestDBConcurrencyRW test is already pretty well written.
This change address the following:
- Close the accessors correctly to prevent goroutine leaks.
- Adding a timeout for the entire test
- Limiting the test expected execution for 15 seconds, as it expected to be enough on the happy path.
- Better error handling in few corner cases.

## Test Plan

Test the updated unit test on Travis.

## Ticket
https://algorand.atlassian.net/browse/GOAL2-788


util/db/dbutil_test.go
```diff
@@ -255,28 +257,41 @@ func TestDBConcurrencyRW(t *testing.T) {
...
-		defer wg.Done()
-		defer atomic.StoreInt64(&lastInsert, -1)
-		for i := int64(0); i < 10000; i++ {
-			errw := acc.Atomic(func(tx *sql.Tx) error {
+		defer func() {
+			atomic.StoreInt64(&lastInsert, -1)
+			testRoutineComplete <- struct{}{}
+		}()
+		var errw error
+		for i := int64(1); i < 10000 && errw == nil; i++ {
```
zeldovich: Now that you have `targetTestDurationTimer`, maybe it's reasonable to get rid of the `10000` constant altogether?

PR #226, opened 2019-08-07
Asynchronous EnsureDigest

Vervious:
In sum:
- `certThresholds` now stage values in the proposal hierarchy, and essentially act like `softThresholds` (for the event.period)
-- note: we can receive `certThresholds` for the previous period (but not softs, which aren't the freshest bundle). So now we can stage a value for the previous period, which is a side effect.
-- `certThreshold`s fast forward periods
- adds `stageDigestAction`, distinct from `ensureAction`, to signal the ledger that it should attempt to fetch the block given a certificate. It is not a blocking operation.
- `certThreshold` without payloads now trigger `stageDigestAction`
- If we receive a payload, check if `cert` is freshest bundle; if so, finish round.

Note that accepting `certThreshold`s as freshest bundles certainly impacts relay rules. I went through them and any failure to relay new votes/bundles is benign, but this logic will need a review.


agreement/common_test.go
```diff
@@ -348,15 +348,18 @@ func (l *testLedger) EnsureDigest(c Certificate, quit chan struct{}, verifier *A
...
 	}
 
-	select {
-	case <-quit:
-		return
-	case <-l.Wait(r):
-		if !consistencyCheck() {
-			err := fmt.Errorf("Wait channel fired without matching block in round %v", r)
-			panic(err)
+	go func() {
```
tsachiherman: I think that this function is doing beyond and above what it needs to do.
This is part of the agreement service, not the ledger service. I don't think that we need to verify the ledger is working correctly. In the context of your change, this function could become a no-op.

node/impls.go
```diff
@@ -148,52 +153,61 @@ func (l agreementLedger) EnsureValidatedBlock(ve agreement.ValidatedBlock, c agr
...
 	logging.Base().Debug("consensus was reached on a block we don't have yet: ", blockHash)
-	for {
-		// Ask the fetcher to get the block somehow
-		block, fetchedCert, err := l.FetchBlockByDigest(round, quit)
-		if err != nil {
-			select {
-			case <-quit:
-				logging.Base().Debugf("EnsureDigest was asked to quit before we could acquire the block")
+
+	go func() {
```
tsachiherman: nit: for that sake of readability, could you move this into its own function rather than making an anonymous function ?

node/impls.go
```diff
@@ -135,6 +137,9 @@ type agreementLedger struct {
 
 	ff rpcs.FetcherFactory
 	n  network.GossipNode
+
+	// concurrent fetchByDigest kludge (until we remove fetchByDigest entirely)
+	ongoingFetchByDigest uint32 // 0 if no, 1 if yes
```
tsachiherman: place that as the first/second member of this struct to ensure it's 64-bit aligned.
It's needed for supporting RasPI.

node/impls.go
```diff
@@ -219,13 +233,23 @@ func (l agreementLedger) innerFetch(fetcher rpcs.Fetcher, round basics.Round, qu
...
+// fetch operation to be a no-op, because we can only have one fetcher per tag.
+var errOngoingFetch = errors.New("Ongoing fetch; can only fetch one digest at a time")
+
 // FetchBlockByDigest is a helper for EnsureDigest.
 // TODO This is a kludge. Instead we should have a service that sees what the ledger is waiting on, fetches it, and calls EnsureBlock on it.
 // TODO this doesn't actually use the digest from cert!
 func (l agreementLedger) FetchBlockByDigest(round basics.Round, quit chan struct{}) (bookkeeping.Block, agreement.Certificate, error) {
+	// Check if there is another ongoing FetchBlockByDigest call. If there is, this call becomes a no-op.
+	if !atomic.CompareAndSwapUint32(&l.ongoingFetchByDigest, 0, 1) {
+		return bookkeeping.Block{}, agreement.Certificate{}, errOngoingFetch
```
tsachiherman: nit: could you log an info message here, so we'll know that it happened ?

node/impls.go
Vervious: https://golang.org/pkg/sync/atomic/#pkg-note-BUG

Seems that this only applies to 64 bit words? not uint32

Vervious:
Closing this PR and directing further work to #754

PR #229, opened 2019-08-08
Adding multisignature signing

sharonha99:
## Summary

Adding multisignature signing from a ledger nano s.

## Test Plan

Manually tested multisignature transaction signing using both authorized and unauthorized Ledger Nano S devices.

### DIFF
--- a/daemon/kmd/wallet/driver/ledger.go
+++ b/daemon/kmd/wallet/driver/ledger.go
@@ -210,20 +210,24 @@ func (lw *LedgerWallet) SignTransaction(tx transactions.Transaction, pw []byte) 
 // MultisigSignTransaction implements the Wallet interface.
 func (lw *LedgerWallet) MultisigSignTransaction(tx transactions.Transaction, pk crypto.PublicKey, partial crypto.MultisigSig, pw []byte) (crypto.MultisigSig, error) {
 
+	isValidKey := false
 	for i := 0; i < len(partial.Subsigs); i++ {
 		subsig := &partial.Subsigs[i]
-		if subsig.Key == pk{
+		if subsig.Key == pk {
+			isValidKey = true
 			sig, err := lw.signTransactionHelper(tx)
 			if err != nil {
 				return partial, err
 			}
 			subsig.Sig = sig
-			return partial, nil
 		}
 	}
 
-	err := fmt.Errorf("Invalid key for multisig transaction")
-	return partial, err
+	if !isValidKey {
+		return partial, errMsigWrongKey
+	}
+
+	return partial, nil
 }
 
 func uint64le(i uint64) []byte {
--- a/daemon/kmd/wallet/driver/ledger.go
+++ b/daemon/kmd/wallet/driver/ledger.go
@@ -210,20 +210,27 @@ func (lw *LedgerWallet) SignTransaction(tx transactions.Transaction, pw []byte) 
 // MultisigSignTransaction implements the Wallet interface.
 func (lw *LedgerWallet) MultisigSignTransaction(tx transactions.Transaction, pk crypto.PublicKey, partial crypto.MultisigSig, pw []byte) (crypto.MultisigSig, error) {
 	isValidKey := false
-	sig, err := lw.signTransactionHelper(tx)
-	if err != nil {
-		return partial, err
-	}
 	for i := 0; i < len(partial.Subsigs); i++ {
 		subsig := &partial.Subsigs[i]
 		if subsig.Key == pk {
 			isValidKey = true
-			subsig.Sig = sig
 		}
 	}
 
 	if !isValidKey {
 		return partial, errMsigWrongKey
+	}
+
+	sig, err := lw.signTransactionHelper(tx)
+	if err != nil {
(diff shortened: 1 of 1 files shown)


daemon/kmd/wallet/driver/ledger.go
```diff
@@ -209,7 +209,21 @@ func (lw *LedgerWallet) SignTransaction(tx transactions.Transaction, pw []byte)
 
 // MultisigSignTransaction implements the Wallet interface.
 func (lw *LedgerWallet) MultisigSignTransaction(tx transactions.Transaction, pk crypto.PublicKey, partial crypto.MultisigSig, pw []byte) (crypto.MultisigSig, error) {
-	return crypto.MultisigSig{}, errNotSupported
+
+	for i := 0; i < len(partial.Subsigs); i++ {
+		subsig := &partial.Subsigs[i]
+		if subsig.Key == pk{
```
zeldovich: Please run `go fmt` on this file

daemon/kmd/wallet/driver/ledger.go
```diff
@@ -209,7 +209,21 @@ func (lw *LedgerWallet) SignTransaction(tx transactions.Transaction, pw []byte)
...
+			sig, err := lw.signTransactionHelper(tx)
+			if err != nil {
+				return partial, err
+			}
+			subsig.Sig = sig
+			return partial, nil
```
zeldovich: It would be good to handle the case when `pk` appears multiple times in `partial.Subsigs`.

daemon/kmd/wallet/driver/ledger.go
```diff
@@ -209,7 +209,21 @@ func (lw *LedgerWallet) SignTransaction(tx transactions.Transaction, pw []byte)
...
+		}
+	}
+
+	err := fmt.Errorf("Invalid key for multisig transaction")
```
zeldovich: You should reuse `errMsigWrongKey` that `sqlite.go` returns for this case.

daemon/kmd/wallet/driver/ledger.go
sharonha99: Should it sign multiple times then? Or should that be an error?

zeldovich: The current code (in `sqlite.go`) puts the same signature in each slot.  We should not ask the user to approve the transaction on the Ledger multiple times (one per multisig slot).

daemon/kmd/wallet/driver/ledger.go
```diff
@@ -209,7 +209,25 @@ func (lw *LedgerWallet) SignTransaction(tx transactions.Transaction, pw []byte)
...
 // MultisigSignTransaction implements the Wallet interface.
 func (lw *LedgerWallet) MultisigSignTransaction(tx transactions.Transaction, pk crypto.PublicKey, partial crypto.MultisigSig, pw []byte) (crypto.MultisigSig, error) {
-	return crypto.MultisigSig{}, errNotSupported
+
+	isValidKey := false
+	for i := 0; i < len(partial.Subsigs); i++ {
+		subsig := &partial.Subsigs[i]
+		if subsig.Key == pk {
+			isValidKey = true
+			sig, err := lw.signTransactionHelper(tx)
```
zeldovich: If the user's key appears N times in the multisig array, this code will ask the user to sign the transaction N times on their ledger.  It would be good to avoid that, and ask just once.

daemon/kmd/wallet/driver/ledger.go
```diff
@@ -209,7 +209,25 @@ func (lw *LedgerWallet) SignTransaction(tx transactions.Transaction, pw []byte)
 
 // MultisigSignTransaction implements the Wallet interface.
 func (lw *LedgerWallet) MultisigSignTransaction(tx transactions.Transaction, pk crypto.PublicKey, partial crypto.MultisigSig, pw []byte) (crypto.MultisigSig, error) {
-	return crypto.MultisigSig{}, errNotSupported
+
```
zeldovich: Minor nit: most (all?) of the existing code doesn't have a blank line at the beginning of the function body.

daemon/kmd/wallet/driver/ledger.go
```diff
@@ -209,7 +209,24 @@ func (lw *LedgerWallet) SignTransaction(tx transactions.Transaction, pw []byte)
 
 // MultisigSignTransaction implements the Wallet interface.
 func (lw *LedgerWallet) MultisigSignTransaction(tx transactions.Transaction, pk crypto.PublicKey, partial crypto.MultisigSig, pw []byte) (crypto.MultisigSig, error) {
-	return crypto.MultisigSig{}, errNotSupported
+	isValidKey := false
+	sig, err := lw.signTransactionHelper(tx)
```
zeldovich: Should this code first check that the key is present in the array, before it asks the user to page through the transaction on the Ledger's small screen and approve it?  I think that means two loops over `partial.Subsigs`, one before `signTransactionHelper` and one after, but loops are negligible compared to asking the user to sign a transaction that turns out to be useless because it's the wrong key.

Issue #230, opened 2019-08-09
Impossible to compile with multiple `GOPATH`


fabrice102:
**Is your feature request related to a problem? Please describe.**

The environment variable `$GOPATH` may contain multiple paths (see https://golang.org/cmd/go/#hdr-GOPATH_environment_variable).
This is not supported by the compilation system: compilation will failed in that setting.

**Describe the solution you'd like**

The assumption that `$GOPATH` is a single folder is used extensively in the source code and it is not clear that there is a simple solution to fix it.
Maybe the simplest solution is to add a comment in the README explaining that `$GOPATH` must be a single folder.

zeldovich:
Thanks for pointing that out.  I had not realized that Go tools support a `$GOPATH` containing multiple paths, colon-separated.  The assumption that `$GOPATH` has just a single element is indeed fairly widespread in our code and scripts, though it shows up a bit less now that we've added support for Go modules.

https://github.com/algorand/go-algorand/pull/231 should be enough to at least get `make` to run fine with a multi-component `$GOPATH`.

onetechnical:
The fix in #231 just presumes the first path in GOPATH is correct for installing binaries. This may or may not be true, so we should just update the README to indicate the same for now.

ian-algorand:
Blocked on 1.14 upgrade, this will be resolved at that point.

bbroder-algo:
dead or worked around by now

PR #231, opened 2019-08-09
support multi-element $GOPATH enough to run "make"

zeldovich:
Partially helps with issue #230


Makefile
```diff
@@ -1,12 +1,13 @@
 GOPATH		:= $(shell go env GOPATH)
+GOPATH1		:= $(firstword $(subst :, ,$(GOPATH)))
```
tsachiherman: we need to pick the path which has our source code, not necessarily the first one.

zeldovich: Actually what we care about here is where `go get` installs binaries, which (I believe) is always the first one.

Issue #204, opened 2019-07-31
expose a function to sign the notefield in JS and Java SDK


ian-algorand:
In the JS and Java SDKs we allow signing a bid with one key and then signing the transaction with another. This is a very multipurpose type of function where one person's data is transferred by a hosting companies transaction. It would be good to expose a function to sign the notefield in general. Ie example another function that works almost identically to signbid.

Vervious:
I'm not convinced there's a compelling case for supporting this very specific behavior. Users can just sign the notes themselves which is also more flexible. @rotemh @JasonWeathersby did you ever resolve this conversation?

rotemh:
@Vervious I'm with you on the compelling case. But, product says that there is a demand for that. Also, he argument, which can be countered as well, that makes it reasonable to be part of the SDK is to let people sign and verify messages using their Algorand "identity" i.e. SK and checksummed Address (vs some random pk).

algoradam:
If we do end up exposing a function for signing arbitrary messages using the Algorand "identity", we need to be sure it enforces correct domain separation. Otherwise it'll be super easy to shoot yourself in the foot and accidentally sign a note field that actually contains a valid Algorand transaction, and now someone just took all your algos.

rotemh:
@algoradam indeed. We use the `MX` one.

PR #233, opened 2019-08-09
LogicSig

algobolson:
rough draft of LogicSig data and where it would integrate into code

### DIFF
--- a/data/transactions/logicsig.go
+++ b/data/transactions/logicsig.go
@@ -23,6 +23,16 @@ import (
 	"github.com/algorand/go-algorand/protocol"
 )
 
+// LogicSigLogic is a trivial wrapper object to organize hashing/signing the contained Logic expression
+type LogicSigLogic struct {
+	_struct struct{} `codec:",omitempty,omitemptyarray"`
+
+	// Logic is an expression in a TBD language.
+	Logic []byte `codec:"l"`
+
+	// Future implementations could be "l2":[]byte, etc.
+}
+
 // LogicSig contains logic for validating a transaction.
 // LogicSig is signed by an account, allowing delegation of operations.
 // OR
@@ -30,9 +40,8 @@ import (
 type LogicSig struct {
 	_struct struct{} `codec:",omitempty,omitemptyarray"`
 
-	// Logic is an expression in a TBD language.
-	// Logic signed by Sig or Msig, OR hashed to be the Address of a contract account.
-	Logic []byte `codec:"lgc"`
+	// LogicSigLogic signed by Sig or Msig, OR hashed to be the Address of a contract account.
+	Logic LogicSigLogic `codec:"l"`
 
 	Sig  crypto.Signature   `codec:"sig"`
 	Msig crypto.MultisigSig `codec:"msig"`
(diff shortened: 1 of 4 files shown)


data/transactions/logicsig.go
```diff
@@ -0,0 +1,93 @@
...
+	Args [][]byte `codec:"arg"`
+}
+
+// Blank returns true if there is no content in this LogicSig
+func (lsig *LogicSig) Blank() bool {
+	return len(lsig.Logic) == 0
+}
+
+// ToBeHashed implements our crypto.Hashable interface
+func (lsig *LogicSig) ToBeHashed() (protocol.HashID, []byte) {
```
zeldovich: Should this be a hash of `Logic` rather than the entire `LogicSig`?  (That is, this method receiver should be on type `*Logic`.)

protocol/hash.go
```diff
@@ -33,6 +33,7 @@ const (
 	BalanceRecord     HashID = "BR"
 	Credential        HashID = "CR"
 	Genesis           HashID = "GE"
+	Logic             HashID = "LO"
 	Message           HashID = "MX"
```
zeldovich: This isn't strictly part of this PR, but could you move `multiSigString` from `crypto/multisig.go` into this list of `HashID` values?  Now that we have other things being hashed into addresses (specifically, these new logic addresses), it's important that the hash input for multisig addrs is domain-separated from logic addresses.

algobolson: If multisig.go was generating the multisig Address values through the cypto.Hashable interface that would make sense, but I think because multisig.go is doing its own custom byte appending to build the value that gets hashed I'd rather leave that alone.

data/transactions/logicsig.go
algobolson: I was trying to avoid making nested structures right away, like Txn/SignedTxn, but that might be better in the long run.
It'd be
`SignedLogic {
Logic { Logic: []byte },
Sig, Msig}`

zeldovich: Doesn't need to be a nested structure, you can define `type LogicCode []byte` and then define a method receiver on `LogicCode`.

protocol/hash.go
zeldovich: It would be good to convert it to use `crypto.Hashable`; it's too bad it wasn't written using `crypto.Hashable` to begin with.  The reason it's related to this PR is that, with this PR, it becomes important that multisig and logicsig are domain-separated inputs to our hash function.

data/transactions/logicsig.go
zeldovich: To clarify why I think this is a good idea: our convention is that `Hashable` hashes the whole thing on which `Hashable` is called, but here, `Hashable` isn't returning the entire struct.

zeldovich: Thanks.  Just curious why you didn't like the type alias for `[]byte`?  It would avoid the inefficiency you mentioned in the commit.

protocol/hash.go
algoradam: Nitpick: Can we start making our domain separation prefixes longer/more meaningful than two bytes? e.g., "LogicAddress" instead of "LO". They're never transmitted over the wire so we don't need to make them small.
(See also my comment in https://github.com/algorand/go-algorand/issues/204)

data/transactions/logicsig.go
```diff
@@ -0,0 +1,102 @@
...
+		return errors.New("logic signature validation failed")
+	}
+	if hasMsig {
+		if ok, _ := crypto.MultisigVerify(&lsig.Logic, crypto.Digest(txn.Src()), lsig.Msig); ok {
+			return nil
+		}
+		return errors.New("logic multisig validation failed")
+	}
+
+	return errors.New("inconsistent internal state verifying LogicSig")
```
tsachiherman: consider reconstructing this so that there won't be a "inconsistent internal state verifying LogicSig".
i.e. it would never happen, so we shouldn't have this path.
instead,
use:
if hasSig {
...} else {
  // this is a hasMsig use case..
}

PR #236, opened 2019-08-13
Logic sig eval

algobolson:
bytecode interpreter and assembler implemented far enough to run a time-locked-hash-contract, the logic necessary for a cross-chain swap.
This is still a work in progress.
Lookup for account data and other transactions not implemented yet. (Lookup for the currently active transaction is implemented.)
TODO: `goal clerk` extensions for assembling programs into transactions. Make a standalone assembler/disassembler binary that could be called by SDK libraries. Do compile-time type checks where possible.


zeldovich:
Should the language include a way to check a signature over the transaction?

algobolson:
Where this would be in the validation flow, signatures would have been checked, so it wouldn't 'check' a signature but check that a signer was among the signatures. I was thinking of adding an opcode that popped a []byte address off the stack and pushed a {0,1} boolean that's true if the address is the sole signer or in the active signers of the multisig. Would that work for what you imagined?

zeldovich:
What I was thinking is that the logic might want to implement complicated multisig-like functionality, so that the args to the logic contain signatures by various signers (that don't fit into our regular-sig or multisig models), and the logic code checks that these signatures come from a set of signers that suffices to authorize this transaction.

algobolson:
I think checking signatures on arbitrary bytes should be out of scope for this round, we'd have to add a bunch of operations for []byte manipulation. Checking signatures of the transaction, and then having logic like `(A && B) || (C && D)`. I guess the op would take two []byte off the stack, the Address and the signature, with the address as a []byte immediate constant in the code and the signature would probably come from Lsig.Args.

data/transactions/logic/eval.go
```diff
@@ -0,0 +1,709 @@
...
+
+	"golang.org/x/crypto/sha3"
+
+	"github.com/algorand/go-algorand/config"
+	"github.com/algorand/go-algorand/data/bookkeeping"
+	"github.com/algorand/go-algorand/data/transactions"
+	"github.com/algorand/go-algorand/protocol"
+)
+
+type stackValue struct {
```
tsachiherman: could you add a comment here explaining the the `Uint` and `Bytes` fields are mutually exclusive, and the distinction would be by examining if the Bytes slice is non-nil ?

data/transactions/logic/eval.go
```diff
@@ -0,0 +1,709 @@
...
+	opArgN(cx, 3)
+}
+
+func opBnz(cx *evalContext) {
+	last := len(cx.stack) - 1
+	cx.nextpc = cx.pc + 2
+	isNonZero := cx.stack[last].Uint != 0
+	cx.stack = cx.stack[:last] // pop
+	if isNonZero {
+		cx.nextpc += int(cx.program[cx.pc+1])
```
tsachiherman: is that enough ? i.e. you're limiting the jump to 256 bytes. giving the the program could be 1K, that's quite a limitation.

algobolson: The longest program I've written has 40 bytes of actual content. If we extend that to 2 bytes and limit it to 32767 so that a future language extension could allow for _signed_ jumps, that'd probably good enough for a long while. Ok. probably add this tomorrow.

cmd/goal/clerk.go
```diff
@@ -425,3 +481,76 @@ var signCmd = &cobra.Command{
...
+	fin, err := os.Open(fname)
+	if err != nil {
+		reportErrorf("%s: %s\n", fname, err)
+	}
+	ops := logic.OpStream{}
+	err = ops.Assemble(fin)
+	if err != nil {
+		reportErrorf("%s: %s\n", fname, err)
+	}
+	fin.Close()
```
tsachiherman: I was going to write that I think that you should test the return code of the `Close` function.
But thinking about it, I think that you should use `ioutil.ReadFile` instead, since it would get you a string in one shot. than, you could use the existing AssembeString().

cmd/opdoc/opdoc.go
```diff
@@ -0,0 +1,250 @@
...
+import (
+	"fmt"
+	"io"
+	"os"
+	"strings"
+
+	"github.com/algorand/go-algorand/data/transactions/logic"
+)
+
+type stringString struct {
```
tsachiherman: maybe stringPair ?

cmd/opdoc/opdoc.go
```diff
@@ -0,0 +1,250 @@
...
+	a string
+	b string
+}
+
+func stringStringListToMap(they []stringString) map[string]string {
+	out := make(map[string]string)
```
tsachiherman: nit: it would be nice to initialize the map size.

data/transactions/logic/README.md
```diff
@@ -0,0 +1,171 @@
+# Transaction Execution Approval Language (TEAL)
```
tsachiherman: Do we want to include compilation products in our repo ?

data/transactions/logic/assembler.go
```diff
@@ -0,0 +1,878 @@
...
+	argType = ops.typeStack[last]
+	ops.typeStack = ops.typeStack[:last]
+	return
+}
+
+// Intc writes opcodes for loading a uint64 constant onto the stack.
+func (ops *OpStream) Intc(constIndex uint) error {
+	switch constIndex {
+	case 0:
+		ops.Out.WriteByte(0x22) // intc_0
```
tsachiherman: nit: it would be nice to have a const variable for these opcodes.

data/transactions/logic/assembler.go
```diff
@@ -0,0 +1,878 @@
...
+		ops.Out.WriteByte(0x28) // bytec_0
+	case 1:
+		ops.Out.WriteByte(0x29) // bytec_1
+	case 2:
+		ops.Out.WriteByte(0x2a) // bytec_2
+	case 3:
+		ops.Out.WriteByte(0x2b) // bytec_3
+	default:
+		if constIndex > 0xff {
+			return errors.New("cannot have more than 256 byte constants")
```
tsachiherman: `"cannot have more than 256 byte constants")` - but the user can work around that limitation using byte arrays, right ?

data/transactions/logic/eval.go
```diff
@@ -0,0 +1,710 @@
...
+}
+
+func opMul(cx *evalContext) {
+	last := len(cx.stack) - 1
+	prev := last - 1
+	a := cx.stack[prev].Uint
+	b := cx.stack[last].Uint
+	v := a * b
+	if v/a != b {
+		cx.err = errors.New("* overflowed")
```
gcolvin: `v` is never assigned.  Perhaps:
```
cx.stack[prev].Uint *= cx.stack[last].Uint
if cx.stack[prev].Uint < cx.stack[last].Uint {
```

tsachiherman: good catch ;-)
Although I would place the assignment *after* the if clause so it would remain consistent with *no op on error"

gcolvin: Sounds right.

data/transactions/logic/tlhc.py
```diff
@@ -0,0 +1,61 @@
+#!/usr/bin/env python3
+
+import argparse
+import base64
+import hashlib
+import os
+import sys
+
+template = '''txn CloseRemainderTo
```
derbear: Should there be a cap on the fee here? (Otherwise, as a counterparty, I can destructively tamper with the contract by setting the fee to be very high, which would drain the sender of funds.)

ledger/eval.go
```diff
@@ -385,7 +386,7 @@ func (eval *BlockEvaluator) transactionGroup(txgroup []transactions.SignedTxnWit
 	for _, txad := range txgroup {
 		var txib transactions.SignedTxnInBlock
 
-		err := eval.transaction(txad.SignedTxn, txad.ApplyData, cow, &txib)
+		err := eval.transaction(txad.SignedTxn, txad.ApplyData, txgroup, cow, &txib)
```
algoradam: We now have a bug in our transaction group code (that previously didn't matter but does now):
If someone submits a list of transactions all with zero "group" field to our transaction pool claiming it's all one transaction group, we'll give it to our transaction pool's block evaluator as a single group, and the evaluator will accept it. But when other people see the block, they'll (correctly) interpret each of these transactions as a size-1 group of its own. Previously, this made no difference. But now, a logicsig transaction's validity may depend on whether a transaction is part of a group.
In particular, suppose one of these transactions has logic that rejects if the transaction is not part of a group of size 2 or larger. The node this group is submitted to will propose an invalid block but consider it valid, but everyone else will reject the block.
Probably the fix is to have the `transactionGroup()` function return an error at the beginning if the length of the transaction group is >= 1 but the group field is zero

algobolson: I posted a fix to this (pretty sure). Added a check a few lines below that rejects something submitted as a txn group that contains txns not tagged as part of a group

data/transactions/logic/tlhc.py
```diff
@@ -0,0 +1,65 @@
...
+addr {to}
+==
+txn Receiver
+addr {to}
+==
+&&
+arg 0
+len
+int 32
```
stechu: Should be `intc` here?

cmd/opdoc/opdoc.go
```diff
@@ -0,0 +1,243 @@
...
+		fmt.Fprintf(out, "- Pops: *... stack*, {%s A}", op.Args[0].String())
+		for i, v := range op.Args[1:] {
+			fmt.Fprintf(out, ", {%s %c}", v.String(), rune(int('B')+i))
+		}
+		out.Write([]byte("\n"))
+	}
+	if op.Returns == nil {
+		fmt.Fprintf(out, "- Pushes: _None_\n")
+	} else if len(op.Returns) == 1 {
+		fmt.Fprintf(out, "- Pushes: %s\n", op.Returns[0].String())
```
zeldovich: Omitting this `if` seems like it wouldn't change the behavior at all: the general case below handles `len(op.Returns)==1` just fine

cmd/opdoc/opdoc.go
```diff
@@ -0,0 +1,243 @@
...
+
+	opextra := logic.OpImmediateNote(op.Name)
+	cost := logic.OpCost(op.Name)
+	fmt.Fprintf(out, "\n## %s\n\n- Opcode: 0x%02x %s\n", op.Name, op.Opcode, opextra)
+	if op.Args == nil {
+		fmt.Fprintf(out, "- Pops: _None_\n")
+	} else if len(op.Args) == 1 {
+		fmt.Fprintf(out, "- Pops: *... stack*, %s\n", op.Args[0].String())
+	} else if len(op.Args) == 2 {
+		fmt.Fprintf(out, "- Pops: *... stack*, {%s A}, {%s B}\n", op.Args[0].String(), op.Args[1].String())
```
zeldovich: The general case below seems to handle `len(op.Args)==2` in the same way that this special case handles it; any reason for having this extra code?

data/transactions/logic/README.md
```diff
@@ -0,0 +1,211 @@
...
+
+Constants are pushed onto the stack by `intc`, `intc_[0123]`, `bytec`, and `bytec_[0123]`. The assembler will typically handle converting `int N` or `byte N` into the appropriate constant-offset opcode or opcode followed by index byte.
+
+## Operations
+
+Most operations work with only one type of argument, uint64 or bytes, and panic if the wrong type value is on the stack.
+
+This summary is supplemented by more detail in the [opcodes document](TEAL_opcodes.md).
+
+Some operations 'panic' and immediately end execution of the program. A transaction checked a program that panics is not valid.
```
zeldovich: Typo: "checked by a program"?

data/transactions/logic/README.md
```diff
@@ -0,0 +1,211 @@
...
+| 5 | Receiver | []byte |
+| 6 | Amount | uint64 |
+| 7 | CloseRemainderTo | []byte |
+| 8 | VotePK | []byte |
+| 9 | SelectionPK | []byte |
+| 10 | VoteFirst | uint64 |
+| 11 | VoteLast | uint64 |
+| 12 | VoteKeyDilution | uint64 |
+| 13 | Type | []byte |
+| 14 | TypeEnum | uint64 |
```
zeldovich: Where are the TypeEnum values enumerated?

ledger/eval.go
```diff
@@ -411,6 +412,8 @@ func (eval *BlockEvaluator) transactionGroup(txgroup []transactions.SignedTxnWit
 			txWithoutGroup.ResetCaches()
 
 			group.TxGroupHashes = append(group.TxGroupHashes, crypto.HashObj(txWithoutGroup))
+		} else if len(txgroup) > 1 {
+			return fmt.Errorf("transactionGroup: [%d] had zero Group but was submitted in a group of %d", gi, len(txgroup))
```
zeldovich: I think `goal clerk rawsend` might run into trouble here: right now it uploads the entire file as one group to the REST API, regardless of whether it has group fields set or not.

cmd/opdoc/opdoc.go
algobolson: simplified

cmd/opdoc/opdoc.go
algobolson: simplified

data/transactions/logic/README.md
algobolson: under the op docs for `txn`, noted that

ledger/eval.go
algobolson: what's the right level for separating behavior between an atomic group of txns vs a bunch of unrelated txns? I think Adam raised the issue that we shouldn't have the situation where a proposer might treat a bunch of txns that arrived together as an atomic set but a verifier only pieces back together the atomicity group by the Group field. Though, I guess at that point the whole group is in the block, presumably correctly, and other txns around it are also in the block and it doesn't matter why.

zeldovich: Actually, I looked more closely at `goal clerk rawsend`, and it looks like it already does the right thing.  So, never mind!  I think it'll all just work with your change.

data/transactions/logic/eval.go
```diff
@@ -0,0 +1,1057 @@
...
+			for i, sv := range cx.stack {
+				fmt.Fprintf(cx.Trace, "[%d] %s\n", i, sv.String())
+			}
+		}
+		return false, fmt.Errorf("stack len is %d instead of 1", len(cx.stack))
+	}
+	if cx.stack[0].Bytes != nil {
+		return false, errors.New("stack finished with bytes not int")
+	}
+	return cx.stack[0].Uint != 0, nil
```
pzbitskiy: why do we expect non-zero "return code" on the stack? looks a bit counter-intuitive compare to process exit codes and go's err == nil approach

algobolson: I started thinking of this in terms of a logic expression, like `(A && B) || (C && !D)`, and when all that is evaluated it boils down to one true/false value for whether the transaction is approved.

derbear:
Is it possible to extract the current/maximum TEAL version number supported by the assembler from the command-line? (I know we can extract the current/next consensus protocol and use that to look up the _protocol's_ latest TEAL version number, but I was wondering if there's some way to do it for the current assembler.)

data/transactions/logic/eval.go
```diff
@@ -0,0 +1,1104 @@
...
+	case 16:
+		sv.Uint = txn.AssetAmount
+	case 17:
+		sv.Bytes = txn.AssetSender[:]
+	case 18:
+		sv.Bytes = txn.AssetReceiver[:]
+	case 19:
+		sv.Bytes = txn.AssetCloseTo[:]
+	case 20:
+		panic("GroupIndex should be handled outside txnFieldToStack")
```
derbear: Why is there a special flow for GroupIndex?

data/transactions/logic/doc.go
```diff
@@ -0,0 +1,176 @@
...
+	{"bytec_0", "push bytes constant 0 to stack"},
+	{"bytec_1", "push bytes constant 1 to stack"},
+	{"bytec_2", "push bytes constant 2 to stack"},
+	{"bytec_3", "push bytes constant 3 to stack"},
+	{"arg", "push LogicSig.Args[N] value to stack by index"},
+	{"arg_0", "push LogicSig.Args[0] to stack"},
+	{"arg_1", "push LogicSig.Args[1] to stack"},
+	{"arg_2", "push LogicSig.Args[2] to stack"},
+	{"arg_3", "push LogicSig.Args[3] to stack"},
+	{"txn", "push field from current transaction to stack"},
```
derbear: Suggestion: also autogenerate documentation for txn fields.

algobolson: https://github.com/algobolson/go-algorand/blob/logic-sig-eval/cmd/opdoc/opdoc.go#L209

data/transactions/logic/eval.go
algobolson: made that better for the `txn` case. still special for `gtxn`

derbear: Perhaps it makes sense to make GroupIndex a `global` then if the `gtxn` case is essentially meaningless and needs special handling.

data/transactions/logicsig.go
```diff
@@ -0,0 +1,122 @@
...
+		return errors.New("LogicSig.Logic bad version")
+	}
+	if version > proto.LogicSigVersion {
+		return errors.New("LogicSig.Logic version too new")
+	}
+	if uint64(lsig.Len()) > proto.LogicSigMaxSize {
+		return errors.New("LogicSig.Logic too long")
+	}
+
+	// TODO: figure out how to fix circularity and enable this
```
pzbitskiy: should be done before merging?

data/transactions/logic/eval_test.go
```diff
@@ -0,0 +1,2013 @@
...
+==
+&&`
+
+func TestTxn(t *testing.T) {
+	t.Parallel()
+	for _, txnField := range TxnFieldNames {
+		if !strings.Contains(testTxnProgramText, txnField) {
+			t.Errorf("TestTxn missing field %v", txnField)
+		}
+	}
```
derbear: It would be good to have this test in `TestGtxn` as well. (Almost forgot to add a gtxn implementation of something I added.)

zeldovich:
I wonder if we should document somewhere the analysis for the worst-case running time of logicsigs?

A single logic sig can trigger 10 ed25519verify calls (because LogicSigMaxCost = 20k and the cost of each ed25519verify call is 1900).  Each call costs about 100 usec on my laptop.  A block contains ~5k transactions, so if each txn has the pessimal 10 calls to ed25519verify, this means ~50,000 signatures checked per block, or 5 seconds of time spent validating the block.

This means that an adversary that gets to be a block proposer can stall agreement for an extra 5 seconds, which isn't a disaster.

The situation before logicsigs is that a malicious block proposer can force the checking of a bit under 15,000 signatures per block (I think the most efficient encoding is to use huge multisigs, which require a little over 64 bytes of encoded SignedTxn size to force a signature check).  That amounts to a ~1.5 second stall of consensus.  In theory, those would be pretty easy to check in parallel (though they don't right now), in contrast to the logicsig's, which I think would be tricky to check in parallel correctly.

In the common case, even with full 1MB-sized blocks, this stall of consensus doesn't happen with a non-malicious proposer, because blocks mostly contain transactions that everyone has already seen in their txpool, so those signatures have already been checked, and are not re-checked as part of validating the block.

derbear:
If we want to reduce the worst-case stall time to that of (unparallelized) huge multisigs, we can reduce max program cost, to say, 7000 (3 signature checks per logicsig + the rest of the program), and then if a logicsig wants to check more signatures, it can require itself to be part of a group txn.

Also, it's easy to raise the max program cost later on via protocol upgrade. It might make sense to undershoot for now and then raise the ceiling as we see more scripts that use >3 ed25519verify calls.

data/transactions/logic/eval.go
```diff
@@ -0,0 +1,1100 @@
...
+			cx.err = fmt.Errorf("branch target at %d not an aligned instruction", cx.branchTargets[0])
+			return 0
+		}
+		for len(cx.branchTargets) > 0 && cx.branchTargets[0] == cx.pc {
+			// checks okay
+			cx.branchTargets = cx.branchTargets[1:]
+		}
+	}
+	if opsByOpcode[opcode].Returns != nil {
+		cx.checkStack = append(cx.checkStack, opsByOpcode[opcode].Returns...)
```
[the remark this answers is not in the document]
algobolson: done

data/transactions/logic/chacha20rng.go
```diff
@@ -0,0 +1,45 @@
+package logic
+
+import (
+	"encoding/binary"
+	"errors"
+
+	"gitlab.com/yawning/chacha20.git"
```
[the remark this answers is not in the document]
algobolson: go.mod has a commit hash version recorded. author hasn't tag-versioned anything. I'm slightly tempted to fork their project just to add tag versions. Or we could dump this and use a C wrapper around the libsodium chacha20 implementation since we already link in libsodium.

data/transactions/logic/assembler.go
```diff
@@ -0,0 +1,1053 @@
...
+	string(protocol.UnknownTx),
+	string(protocol.PaymentTx),
+	string(protocol.KeyRegistrationTx),
+	string(protocol.AssetConfigTx),
+	string(protocol.AssetTransferTx),
+	string(protocol.AssetFreezeTx),
+}
+
+// map TxnTypeName to its enum index, for `txn TypeEnum`
+var txnTypeIndexes map[string]int
```
derbear: Minor suggestion: move to `eval.go` because this is also used to implement `txn TypeEnum`.

data/transactions/logic/eval.go
```diff
@@ -0,0 +1,1111 @@
...
+
+func opBtoi(cx *evalContext) {
+	last := len(cx.stack) - 1
+	ibytes := cx.stack[last].Bytes
+	if len(ibytes) > 8 {
+		cx.err = fmt.Errorf("btoi arg too long, got [%d]bytes", len(ibytes))
+		return
+	}
+	value := uint64(0)
+	for _, b := range ibytes {
```
derbear: Is there a reason that you didn't do a symmetric `binary.BigEndian.Uint64` here? Also, what's the rationale behind using big-endian encoding?

derbear: One thing that is possible but perhaps undesirable to support if we have little-endian encoding: `btoi` can accept slices longer than 8 bytes as long as bytes 9-end are all 0. This may be undesirable because `btoi(itob(data)) != data`.

data/transactions/logic/assembler.go
```diff
@@ -0,0 +1,1053 @@
...
+	constIndex, err := strconv.ParseUint(args[0], 0, 64)
+	if err != nil {
+		return err
+	}
+	return ops.Bytec(uint(constIndex))
+}
+
+func parseBinaryArgs(args []string) (val []byte, consumed int, err error) {
+	arg := args[0]
+	if strings.HasPrefix(arg, "base32(") || strings.HasPrefix(arg, "b32(") {
```
derbear: Feature request: `string "asdf"` puts `[]byte("asdf")` into the program.

data/transactions/logicsig.go
algobolson: Done

data/transactions/logic/eval.go
algobolson: I still think of big endian as 'network order' (Thanks, SUN!)
I think `binary.BigEndian.Uint64()` chokes if the source isn't a full 8 bytes, but I wanted to be more flexible.

data/transactions/logic/assembler.go
algobolson: Yes, except string escape sequences are annoying to parse and it's Friday and I'm tired. It's a good idea though. Maybe next week (-:

data/transactions/logic/assembler.go
algobolson: `TxnTypeNames` and `txnTypeIndexes` are _mostly_ used in assembler.go , and it's all the same package.

cmd/goal/clerk.go
```diff
@@ -548,3 +699,187 @@ var splitCmd = &cobra.Command{
...
+		if err != nil {
+			reportErrorf("%s: %s\n", outname, err)
+		}
+	}
+}
+
+var compileCmd = &cobra.Command{
+	Use:   "compile",
+	Short: "compile a contract program",
+	Long:  "compile a contract program, report it's address",
```
EvanJRichard: nit `it's` -> `its`

data/transactions/logic/chacha20rng.go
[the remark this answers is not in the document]
justicz: I agree that we should do one of those two options

data/transactions/verify/txn.go
```diff
@@ -0,0 +1,202 @@
...
+	if hasMsig {
+		if ok, _ := crypto.MultisigVerify(s.Txn, crypto.Digest(s.Txn.Src()), s.Msig); ok {
+			return nil
+		}
+		return errors.New("multisig validation failed")
+	}
+	if hasLogicSig {
+		return LogicSig(&s.Lsig, proto, s)
+	}
+	return errors.New("has one mystery sig. WAT?")
```
tsachiherman: There is no point in this return. It cannot logically happen.
I would suggest restructuring the last two if statements : 
```
if hasMsig {
 		if ok, _ := crypto.MultisigVerify(s.Txn, crypto.Digest(s.Txn.Src()), s.Msig); ok {
 			return nil
 		}
 		return errors.New("multisig validation failed")
 	} else  {
 		return LogicSig(&s.Lsig, proto, s)
 	}
```
The reason is that it would always generate unreachable code, which is undesired.

data/transactions/verify/txn.go
```diff
@@ -0,0 +1,202 @@
...
+	}
+	version, vlen := binary.Uvarint(lsig.Logic)
+	if vlen <= 0 {
+		return errors.New("LogicSig.Logic bad version")
+	}
+	if version > proto.LogicSigVersion {
+		return errors.New("LogicSig.Logic version too new")
+	}
+	if uint64(lsig.Len()) > proto.LogicSigMaxSize {
+		return errors.New("LogicSig.Logic too long")
```
tsachiherman: adding the byte could could be helpful here.

data/transactions/verify/txn.go
```diff
@@ -0,0 +1,202 @@
...
+	}
+	if hasMsig {
+		program := transactions.Program(lsig.Logic)
+		if ok, _ := crypto.MultisigVerify(&program, crypto.Digest(stxn.Txn.Src()), lsig.Msig); ok {
+			return nil
+		}
+		return errors.New("logic multisig validation failed")
+	}
+
+	return errors.New("inconsistent internal state verifying LogicSig")
```
tsachiherman: same issue as before; pointless unreachable code.

data/txHandler.go
```diff
@@ -188,7 +189,7 @@ func (handler *TxHandler) backlogWorker() {
 func (handler *TxHandler) asyncVerifySignature(arg interface{}) interface{} {
 	tx := arg.(*txBacklogMsg)
 	for _, txn := range tx.unverifiedTxGroup {
-		tx.verificationErr = txn.Verify(tx.spec, tx.proto)
+		tx.verificationErr = verify.Txn(&txn, tx.spec, tx.proto)
```
tsachiherman: does the variable content might get modified by the next for-loop iteration before the work on the verification is complete ? If so, it might be a data-race use case.

ledger/eval.go
```diff
@@ -480,6 +485,31 @@ func (eval *BlockEvaluator) transaction(txn transactions.SignedTxn, ad transacti
...
+
+		if !txn.Lsig.Blank() {
+			recs := make([]basics.BalanceRecord, len(txgroup))
+			for i := range recs {
+				var err error
+				recs[i], err = cow.Get(txgroup[i].Txn.Sender, true)
+				if err != nil {
+					return fmt.Errorf("transaction %v: cannot get sender record: %v", txn.ID(), err)
+				}
+			}
```
tsachiherman: This section would cause a series of database requests. These should be:
1. further optimized ( since the same sender could be involved in multiple txns )
2. bounded. The number of requests here should be limited. ( if we do it somehwhere else on the stack than I guess it's ok to have it as is. )

data/transactions/logic/eval.go
```diff
@@ -0,0 +1,1164 @@
...
+	case StackNone:
+		return "None"
+	case StackAny:
+		return "any"
+	case StackUint64:
+		return "uint64"
+	case StackBytes:
+		return "[]byte"
+	}
+	return "internal error, unknown type"
```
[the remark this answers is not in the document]
algobolson: I have an enum+stringer version of several of the eval.go constants on a sub branch. Even if we don't keep the rest of that experiment I'll probably work up pulling in that change.

data/transactions/logic/assembler.go
```diff
@@ -0,0 +1,1081 @@
...
+	}
+}
+
+type lineErrorWrapper struct {
+	Line int
+	Err  error
+}
+
+func (lew *lineErrorWrapper) Error() string {
+	return fmt.Sprintf(":%d %s", lew.Line, lew.Err.Error())
```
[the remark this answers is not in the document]
algobolson: someone would have to try to write bad code, so I'm not too worried
```
wat := &lineErrorWrappe{}
wat.Err = wat
```

data/transactions/logic/assembler.go
```diff
@@ -0,0 +1,1081 @@
...
+		ops.sourceLine++
+		line := scanner.Text()
+		if len(line) == 0 {
+			continue
+		}
+		if strings.HasPrefix(line, "//") {
+			continue
+		}
+		fields := strings.Fields(line)
+		fields = filterFieldsForLineComment(fields)
```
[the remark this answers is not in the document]
algobolson: yeah, it's not properly a recursive parser. but by most code style guide it will work.

go.sum
```diff
@@ -1,3 +1,5 @@
+dmitri.shuralyov.com/gpu/mtl v0.0.0-20190408044501-666a987793e9/go.mod h1:H6x//7gZCb22OMCxBHrMx7a5I7Hp++hsVxbQ4BYO7hU=
+github.com/BurntSushi/xgb v0.0.0-20160522181843-27f122750802/go.mod h1:IVnqGOEym/WlBOVXweHU+Q+/VP0lqqI8lqeDx9IjBqo=
```
justicz: why did these dependencies get added?

go.sum
```diff
@@ -24,6 +27,7 @@ github.com/fsnotify/fsnotify v1.4.7 h1:IXs+QLmnXW2CcXuY+8Mzv/fWEsPGWxqefPtCP5CnV
 github.com/fsnotify/fsnotify v1.4.7/go.mod h1:jwhsz4b93w/PPRr/qN1Yymfu8t87LnFCMoQvtojpjFo=
 github.com/gen2brain/beeep v0.0.0-20180718162406-4e430518395f h1:eyHMPp7tXlBMF8PZHdsL89G0ehuRNflu7zKUeoQjcJ0=
 github.com/gen2brain/beeep v0.0.0-20180718162406-4e430518395f/go.mod h1:GprdPCZglWh5OMcIDpeKBxuUJI+fEDOTVUfxZeda4zo=
+github.com/go-gl/glfw v0.0.0-20190409004039-e6da0acd62b1/go.mod h1:vR7hzQXu2zJy9AVAgeJqvqgH9Q5CA+iKCZ2gyEVpxRU=
```
justicz: same here

data/transactions/verify/txn.go
justicz: I like the explicitness of the current code, because if we ever add a new signature type, we would instantly notice if we forgot to update this function.

ledger/eval.go
```diff
@@ -480,6 +485,34 @@ func (eval *BlockEvaluator) transaction(txn transactions.SignedTxn, ad transacti
 		if !txn.Txn.Group.IsZero() && !eval.proto.SupportTxGroups {
 			return fmt.Errorf("transaction groups not supported")
 		}
+
+		if !txn.Lsig.Blank() {
```
justicz: Hm, where are we checking that `LogicSig` has been enabled in the current protocol version? There's a comment somewhere about how `LogicSigVersion == 0` means no support, but I'm not sure where that's enforced.

data/transactions/verify/txn.go
```diff
@@ -0,0 +1,202 @@
...
+	"github.com/algorand/go-algorand/util/execpool"
+)
+
+// TxnPool verifies that a SignedTxn has a good signature and that the underlying
+// transaction is properly constructed.
+// Note that this does not check whether a payset is valid against the ledger:
+// a SignedTxn may be well-formed, but a payset might contain an overspend.
+//
+// This version of verify is performing the verification over the provided execution pool.
+func TxnPool(s *transactions.SignedTxn, spec transactions.SpecialAddresses, proto config.ConsensusParams, verificationPool execpool.BacklogPool) error {
```
justicz: Can we name this something clearer? Maybe `SignedAndWellFormed` or something? `TxnPool` makes it sound like a constructor or something IMO

data/transactions/verify/txn.go
```diff
@@ -0,0 +1,202 @@
...
+	}
+	if !lsig.Msig.Blank() {
+		hasMsig = true
+		numSigs++
+	}
+	if numSigs == 0 {
+		// if the txn.Sender == hash(Logic) then this is a (potentially) valid operation on a contract-only account
+		program := transactions.Program(lsig.Logic)
+		lhash := crypto.HashObj(&program)
+		if crypto.Digest(stxn.Txn.Sender) == lhash {
```
justicz: nit: we use `Src()` below instead of `Sender`, though I just checked and they're the same.

cmd/goal/clerk.go
```diff
@@ -74,6 +92,10 @@ func init() {
 	sendCmd.Flags().BoolVarP(&sign, "sign", "s", false, "Use with -o to indicate that the dumped transaction should be signed")
 	sendCmd.Flags().StringVarP(&closeToAddress, "close-to", "c", "", "Close account and send remainder to this address")
 	sendCmd.Flags().BoolVarP(&noWaitAfterSend, "no-wait", "N", false, "Don't wait for transaction to commit")
+	sendCmd.Flags().StringVarP(&programSource, "from-program", "F", "", "Program source to use as account logic")
+	sendCmd.Flags().StringVarP(&progByteFile, "from-program-bytes", "P", "", "Program binary to use as account logic")
+	sendCmd.Flags().StringSliceVar(&argB64Strings, "argb64", nil, "base64 encoded args to pass to transaction logic")
```
derbear: Feature request: Have a `--to-program` as well for symmetry. It helps with sending money to escrows to initialize them.

data/transactions/logic/assembler.go
```diff
@@ -0,0 +1,1112 @@
...
+func init() {
+	disByName = make(map[string]disassembler)
+	for _, dis := range disassemblers {
+		disByName[dis.name] = dis
+	}
+}
+
+var errTooManyIntc = errors.New("intcblock with too many items")
+
+func parseIntcblock(program []byte, pc int) (intc []uint64, nextpc int, err error) {
```
justicz: As discussed offline, I think it would make sense to move these functions into another file. It feels weird when we fix a panic in `Check` and the fix ends up in `assembler.go`

data/transactions/logic/eval.go
```diff
@@ -186,6 +187,10 @@ func Eval(program []byte, params EvalParams) (pass bool, err error) {
 			err = PanicError{x, errstr}
 		}
 	}()
+	if (params.Proto != nil) && (params.Proto.LogicSigVersion == 0) {
```
tsachiherman: maybe I'm not fully understand the context, but didn't you mean
```
if (params.Proto == nil) || (params.Proto.LogicSigVersion == 0)
```

justicz:
fuzzer found two more crashes in `Eval` (not `Check`). they might be the same root cause, haven't checked yet: `00260200ffffffffffffffffff0130` and `002602ffffffffffffffffff0130`

data/transactions/logic/assembler.go
```diff
@@ -0,0 +1,1112 @@
...
+		err = errTooManyItems
+		return
+	}
+	bytec = make([][]byte, numItems)
+	for i := uint64(0); i < numItems; i++ {
+		if pos >= len(program) {
+			err = errShortBytecblock
+			return
+		}
+		itemLen, bytesUsed := binary.Uvarint(program[pos:])
```
justicz: `itemLen` can be negative when cast as `int`, which can cause an OOB slice below

cmd/goal/clerk.go
```diff
@@ -548,3 +706,187 @@ var splitCmd = &cobra.Command{
...
+		block.BlockHeader.TimeStamp = timeStamp
+		proto := config.Consensus[protocol.ConsensusCurrentVersion]
+		for i, txn := range txgroup {
+			if txn.Lsig.Blank() {
+				continue
+			}
+			ep := logic.EvalParams{Txn: &txn.SignedTxn, Proto: &proto}
+			cost, err := logic.Check(txn.Lsig.Logic, ep)
+			if err != nil {
+				reportErrorf("program failed Check: %s")
```
derbear: Nit: pass `err` as an argument to the `reportErrorf` function.

cmd/goal/clerk.go
```diff
@@ -548,3 +706,187 @@ var splitCmd = &cobra.Command{
...
+		for i, st := range stxns {
+			txgroup[i].SignedTxn = st
+		}
+		if timeStamp == 0 {
+			timeStamp = time.Now().Unix()
+		}
+		block := bookkeeping.Block{}
+		block.BlockHeader.Round = basics.Round(round)
+		block.BlockHeader.TimeStamp = timeStamp
+		proto := config.Consensus[protocol.ConsensusCurrentVersion]
```
derbear: If `ConsensusCurrentVersion` doesn't support TEAL contracts, then this command won't work.

justicz:
should we prevent teal-only accounts from sending key registration transactions? it doesn't make sense for them to be online since they can't vote

data/transactions/logic/eval.go
```diff
@@ -0,0 +1,1193 @@
...
+	cx.checkStack = make([]StackType, 0, 10)
+	cx.program = program
+	for (cx.err == nil) && (cx.pc < len(cx.program)) {
+		cost += cx.checkStep()
+	}
+	if cx.err != nil {
+		err = fmt.Errorf("%3d %s", cx.pc, cx.err)
+		return
+	}
+	if len(cx.checkStack) != 1 {
```
derbear: Following discussion with @pzbitskiy:

One counterintuitive behavior of `Check` is that because it doesn't account for branches, it rejects programs that are valid when executed, but which leaves the same number of things on the stack in both sides of the branch and are thus rejected as they leave too many things on the stack. 

For instance, consider the following statement in a hypothetical high-level language:
```
if <cond>:
   x = <then-value>
otherwise:
   x = <else-value>
return x > 5
```

A natural way to compile this would be into the following assembly:
```
<cond>
bnz then
<else-condition>
int 1
bnz end
then:
<then-condition>
end:
5
>
```

Even though this program should execute correctly, because `Check` does not know that both branches are mutually exclusive, it will reject this program as leaving too much data on the stack.

data/transactions/logicsig.go
```diff
@@ -0,0 +1,75 @@
...
+// Blank returns true if there is no content in this LogicSig
+func (lsig *LogicSig) Blank() bool {
+	return len(lsig.Logic) == 0
+}
+
+// Len returns the length of Logic plus the length of the Args
+// This is limited by config.ConsensusParams.LogicSigMaxSize
+func (lsig *LogicSig) Len() int {
+	lsiglen := len(lsig.Logic)
+	for _, arg := range lsig.Args {
```
justicz: we should cap the max number of arguments (also, probably we should reject any  logicsig with`nil` (not `[]byte{}`) arguments)

data/transactions/logic/assembler.go
```diff
@@ -0,0 +1,1112 @@
...
+		if err != nil {
+			return
+		}
+		consumed = 1
+	} else if arg == "base32" || arg == "b32" {
+		if len(args) < 2 {
+			err = fmt.Errorf("need literal after 'byte %s'", arg)
+			return
+		}
+		val, err = base32.StdEncoding.WithPadding(base32.NoPadding).DecodeString(args[1])
```
stechu: The assembler seems doesn't allow padding for b32 but allow padding for b64 input. Could you elaborate why?

data/transactions/logic/eval.go
```diff
@@ -0,0 +1,1193 @@
...
+	case 12:
+		sv.Uint = txn.VoteKeyDilution
+	case 13:
+		sv.Bytes = []byte(txn.Type)
+	case 14:
+		sv.Uint = uint64(txnTypeIndexes[string(txn.Type)])
+	case 15:
+		sv.Bytes = make([]byte, 40)
+		copy(sv.Bytes, txn.XferAsset.Creator[:])
+		binary.BigEndian.PutUint64(sv.Bytes[32:], txn.XferAsset.Index)
```
derbear: Instead of having a custom byte-encoding here, we should probably define one inside of `userBalance.go`.

derbear: Actually, it may be possible to just refer to `txn.XferAsset.Index` and have only 8 bytes on the stack.

data/transactions/logic/assembler.go
algobolson: our common text representation for addresses is unpadded base32

data/transactions/logicsig.go
algobolson: not all programs need args, so i'd like to be able to use the most compact msgpack encoding of omitting the field entirely

data/transactions/logic/assembler.go
derbear: For addresses (`addr`) this makes sense. However, arbitrary base 32-encoded data will be decoded incorrectly (wrong length). Do we want to clarify that we don't accept arbitrary encodings here? (Seems less misleading to me to just drop `base32` in this case.)

derbear: For addresses (`addr`) this is fine, but for arbitrary `base32` encoded data this will silently fail to decode correctly (wrong length).

derbear: If we want to avoid confusion, it makes sense to me to either (1) drop `base32` as an encoding format for TEAL assembly or (2) have `base32` behave differently from `addr` and take standard padding.

algobolson: unpadded base32 is clean. for 1 to 11 bytes I encoded to unpadded base32 string and decoded back and got the same bytes.

algobolson: assembler `byte ...` will now accept base32 strings with constant padding

data/transactions/logicsig.go
justicz: Sure -- I still think capping the max number of arguments would make sense though, just so the representation of a valid txn can't be arbitrarily long.

PR #237, opened 2019-08-13
Created algocfg to provide cli access to config.json

Karmastic:
Adds `algocfg get|set|reset` commands to manipulate config.json
in a consistent and typesafe way.
Modified config.json is persisted minimally - only non-default
values are written.  Versioning is taken into account and migration
is performed as required.

This also includes a fix to SaveNonDefaultValuesToFile() which was
writing config.json files with a trailing empty line for each default
value that was skipped.

### DIFF
--- a/cmd/algocfg/datadir.go
+++ b/cmd/algocfg/datadir.go
@@ -1,0 +1,72 @@
+// Copyright (C) 2019 Algorand, Inc.
+// This file is part of go-algorand
+//
+// go-algorand is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as
+// published by the Free Software Foundation, either version 3 of the
+// License, or (at your option) any later version.
+//
+// go-algorand is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with go-algorand.  If not, see <https://www.gnu.org/licenses/>.
+package main
+
+import "os"
+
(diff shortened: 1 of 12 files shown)


cmd/algocfg/datadir.go
```diff
@@ -0,0 +1,72 @@
...
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with go-algorand.  If not, see <https://www.gnu.org/licenses/>.
+package main
+
+import "os"
+
+var dataDirs []string
+
+func resolveDataDir() string {
```
algobolson: these functions exist elsewhere (cmd/goal/... at least) and maybe it's time to extract them? If not libgoal than some other library of algo-command-utils

cmd/algocfg/setCommand.go
```diff
@@ -0,0 +1,127 @@
...
+	}
+
+	err := setFieldValue(f, value)
+	return object, err
+}
+
+func setFieldValue(field reflect.Value, value string) error {
+	switch k := field.Kind(); k {
+	case reflect.Int, reflect.Int8, reflect.Int16, reflect.Int32, reflect.Int64:
+		val, err := strconv.ParseInt(value, 10, 64)
```
algobolson: we could ParseInt(value, 0, 64), 0-base allows for detecting 0777 octal or 0xabc hex. (Also ParseUint below)

cmd/algocfg/datadir.go
Karmastic: Yes they do.  I moved in the direction of extracting them in this PR - moving them to distinct files.  I didn't want to incur the extra cost right now.

cmd/algocfg/setCommand.go
Karmastic: We could, but we do not want to or need to.  I don't want it to be ambiguous.

PR #239, opened 2019-08-15
[GOAL2-921] support wildcard for kmd CORS

Karmastic:


daemon/kmd/api/cors.go
```diff
@@ -35,6 +35,10 @@ func corsMiddleware(allowedOrigins []string) func(http.Handler) http.Handler {
 			for _, allowedOrigin := range allowedOrigins {
 				if subtle.ConstantTimeCompare([]byte(origin), []byte(allowedOrigin)) == 1 {
 					found = allowedOrigin
+					break
+				} else if allowedOrigin == "*" {
+					found = origin
+					break
```
tsachiherman: nit: You might want to consider removing the second `break` so that explicit domains would supersede all domains.

Karmastic: Why?  That changes nothing.

tsachiherman: If you have the following allowedOrigins list:
"*",
"r1.algorand.network",
"r2.algorand.network"

Then it would prefer the last two over the "*".

tsachiherman: arguably, if someone already configuring it for "*", it doesn't really matter.

Karmastic:
The result is the same.  We allow it and return the origin host string
(just get it different ways).

PR #260, opened 2019-08-19
Kmd sign bytes

algobolson:
## Summary

Allow KMD to sign arbitrary bytes so that in the future it can sign program bytecode blobs.

## Test Plan

Needs unit test coverage comparable to other API tests. TODO.

### DIFF
--- a/daemon/kmd/api/v1/handlers.go
+++ b/daemon/kmd/api/v1/handlers.go
@@ -858,6 +858,64 @@ func postTransactionSignHandler(ctx reqContext, w http.ResponseWriter, r *http.R
 	successResponse(w, resp)
 }
 
+// postDataSignHandler handles `POST /v1/data/sign`
+func postDataSignHandler(ctx reqContext, w http.ResponseWriter, r *http.Request) {
+	// swagger:operation POST /v1/data/sign SignData
+	//---
+	//    Summary: Sign data
+	//    Description: >
+	//      Signs the passed data with a key from the wallet, determined
+	//      by the sender encoded in the transaction.
+	//    Produces:
+	//    - application/json
+	//    Parameters:
+	//      - name: Sign Data Request
+	//        in: body
+	//        required: true
+	//        schema:
+	//          "$ref": "#/definitions/SignDataRequest"
+	//    Responses:
+	//      "200":
+	//        "$ref": "#/responses/SignDataResponse"
+	var req kmdapi.APIV1POSTDataSignRequest
+
+	// Decode the request
+	decoder := protocol.NewJSONDecoder(r.Body)
+	err := decoder.Decode(&req)
+	if err != nil {
+		errorResponse(w, http.StatusBadRequest, errCouldNotDecode)
+		return
+	}
+
+	// Fetch the wallet from the WalletHandleToken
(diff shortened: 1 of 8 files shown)


algobolson:
ok, all the APIs are now explicitly about Program and there's a prefix for isolation

protocol/hash.go
```diff
@@ -39,6 +39,7 @@ const (
 	OneTimeSigKey2    HashID = "OT2"
 	PaysetFlat        HashID = "PF"
 	Payload           HashID = "PL"
+	Program           HashID = "Program"
```
tsachiherman: I haven't seen where you've been using this constant, but I might have missed it.

rotemh: Also, we use the `MX` for all arbitrary bytes. You may want to consider using it. If not, I'd consider changing the constant to follow the convention of 2-3 capital letters.

justicz: @rotemh I think it should definitely have a new `HashID` since a program is semantically different from a human-readable message.

rotemh: I'm good with either.

daemon/kmd/wallet/driver/ledger.go
```diff
@@ -297,3 +336,17 @@ func (lw *LedgerWallet) signTransactionHelper(tx transactions.Transaction) (sig
 	copy(sig[:], reply)
 	return
 }
+
+func (lw *LedgerWallet) signProgramHelper(data []byte) (sig crypto.Signature, err error) {
+	lw.mu.Lock()
+	defer lw.mu.Unlock()
+
+	// TODO: this is probably wrong and needs extending the client code on the device
+	reply, err := lw.dev.Exchange(data)
```
justicz: The ledger expects data in a particular format -- I think this method should just return an error for now if the ledger firmware isn't ready.

protocol/hash.go
algobolson: adam was suggesting that we should start to move to longer prefixes that might be more secure or give us more namespace or something, and it hardly matters how long they are since they're never stored anywhere

daemon/kmd/wallet/driver/ledger.go
algobolson: ok, errors out now instead of doing something nonsensical

algobolson:
I think this change is ready to go now that I added tests.

daemon/kmd/api/v1/handlers.go
```diff
@@ -858,6 +858,64 @@ func postTransactionSignHandler(ctx reqContext, w http.ResponseWriter, r *http.R
...
+	//    Responses:
+	//      "200":
+	//        "$ref": "#/responses/SignProgramResponse"
+	var req kmdapi.APIV1POSTProgramSignRequest
+
+	// Decode the request
+	decoder := protocol.NewJSONDecoder(r.Body)
+	err := decoder.Decode(&req)
+	if err != nil {
+		errorResponse(w, http.StatusBadRequest, errCouldNotDecode)
```
tsachiherman: nit: it would be nice to have "what we failed to decode" in our log files so that we'll have something to work with when we attempt to fix this issue.

PR #261, opened 2019-08-20
Add the ability to mark an account as Non-Participating.

EvanJRichard:
## Summary

This PR proposes to add the ability to mark an account as `not participating`/`excluded`/`nonparticipating`, per design discussed with @algoradam (which is why I've requested him as a reviewer).

A "become non-participating" transaction is a keyreg transaction that supplies no keys, as though it were a "go offline" transaction. However, in the "become non-participating" scenario, an additional bool flag, `nonparticipating`, is high. It is an error (enforced by `transaction.WellFormed`) to have a keyreg transaction that both supplies key information and has the `nonparticipating` flag high. 

Once an account is non-participating, it can never exit that state. So, the work flow for making a new non-participating account would be: make a new account; fund the account; issue a become-nonparticipating transaction from that account.

## Test Plan

I added some tests (see PR: `transaction_test`, `keyreg_test`, `onlineStatusChange_test`). Additionally, using Travis suite to detect regressions in other tests, and `make test` passed on my machine. Further, I manually spun up a private network and marked some accounts as nonparticipating myself, using `goal`.

### Documentation concerns
Reviewers or drive-by readers: what documentation are you aware of that will need updating once this is in? 
So far, I have: https://developer.algorand.org/docs/goal-account


EvanJRichard:
Change requested offline by @derbear : this needs to be implemented as a protocol upgrade.

Vervious:
@tsachiherman how do you mean by this?

> Looks to me that setting the VotePK and SelectionPK would be preferable solution , since it dodges the use case of mutually exclusive VotePK & SelectionPK vs Nonparticipation.

note that clearing votePK and selectionPK is equivalent to a key-deregistration operation (iirc).

data/transactions/transaction.go
```diff
@@ -227,7 +227,13 @@ func (tx Transaction) WellFormed(spec SpecialAddresses, proto config.ConsensusPa
 			return err
 		}
 	case protocol.KeyRegistrationTx:
-		// All OK
+		// check that, if this tx is marking an account nonparticipating, it supplies no key (as though it were trying to go offline)
+		if tx.KeyregTxnFields.Nonparticipation {
+			suppliesNullKeys := tx.KeyregTxnFields.VotePK == crypto.OneTimeSignatureVerifier{} || tx.KeyregTxnFields.SelectionPK == crypto.VRFVerifier{}
```
zeldovich: Shouldn't this be `&&`?

EvanJRichard: The idea is to mimic the logic in `KeyregTxnFields.apply`, where just having one of the two zeroed-out is sufficient - see `keyreg.go:59` or so.

zeldovich: Ah, I see.  Sure.

tsachiherman:
> @tsachiherman how do you mean by this?
> 
> > Looks to me that setting the VotePK and SelectionPK would be preferable solution , since it dodges the use case of mutually exclusive VotePK & SelectionPK vs Nonparticipation.
> 
> note that clearing votePK and selectionPK is equivalent to a key-deregistration operation (iirc).

Exactly. And isn't key-deregistration and non-participation is practically the same ? ( i.e. this account would not vote ).

Vervious:
@tsachiherman Participating accounts that go offline can still accumulate rewards. Non-participating accounts can never accumulate rewards

tsachiherman:
> @tsachiherman Participating accounts that go offline can still accumulate rewards. Non-participating accounts can never accumulate rewards

Hmm.. I see. yes, that make sense now. I'm not sue why someone would want to get no rewards, but that's ok.

data/transactions/keyreg.go
```diff
@@ -27,11 +26,12 @@ import (
...
-	SelectionPK     crypto.VRFVerifier              `codec:"selkey"`
-	VoteFirst       basics.Round                    `codec:"votefst"`
-	VoteLast        basics.Round                    `codec:"votelst"`
-	VoteKeyDilution uint64                          `codec:"votekd"`
+	VotePK           crypto.OneTimeSignatureVerifier `codec:"votekey"`
+	SelectionPK      crypto.VRFVerifier              `codec:"selkey"`
+	VoteFirst        basics.Round                    `codec:"votefst"`
+	VoteLast         basics.Round                    `codec:"votelst"`
+	VoteKeyDilution  uint64                          `codec:"votekd"`
+	Nonparticipation bool                            `codec:"nonpart"`
```
Karmastic: Really would have made sense to make this a separate txn type.  Overloading an almost entirely orthogonal txn for this is a bad idea.

algoradam: I don't think it's orthogonal -- keyreg transactions are for changing account status. Now they can  change status to NonParticipating in addition to Online and Offline.

protocol/consensus.go
```diff
@@ -94,13 +94,18 @@ const ConsensusV17 = ConsensusVersion(
 	"https://github.com/algorandfoundation/specs/tree/5615adc36bad610c7f165fa2967f4ecfa75125f0",
 )
 
+// ConsensusV18 adds the ability to issue a transaction that marks an account non-participating
+const ConsensusV18 = ConsensusVersion(
+	"---->!!!TODO!!!<----",
```
algoradam: We'll want to PR a spec change into github.com/algorandfoundation/spec.
(Side note: It might or might not make sense to combine this protocol upgrade with Tsachi's protocol upgrade for fixing the reward rate calculation.)

data/transactions/transaction_test.go
```diff
@@ -51,3 +51,66 @@ func TestTransaction_EstimateEncodedSize(t *testing.T) {
...
+	// Also generate a new VRF key
+	vrf := crypto.GenerateVRFSecrets()
+	tx.KeyregTxnFields = KeyregTxnFields{
+		VotePK:           v.OneTimeSignatureVerifier,
+		SelectionPK:      vrf.PK,
+		Nonparticipation: true,
+	}
+	// this tx tries to both register keys to go online, and mark an account as non-participating.
+	// it is not well-formed.
+	feeSink := basics.Address{0x7, 0xda, 0xcb, 0x4b, 0x6d, 0x9e, 0xd1, 0x41, 0xb1, 0x75, 0x76, 0xbd, 0x45, 0x9a, 0xe6, 0x42, 0x1d, 0x48, 0x6d, 0xa3, 0xd4, 0xef, 0x22, 0x47, 0xc4, 0x9, 0xa3, 0x96, 0xb8, 0x2e, 0xa2, 0x21}
```
algoradam: Out of curiosity, where did this address come from? Is this the mainnet fee sink?

cmd/goal/account.go
```diff
@@ -1020,3 +1028,45 @@ var partkeyInfoCmd = &cobra.Command{
 		})
 	},
 }
+
+var markNonparticipatingCmd = &cobra.Command{
+	Use:   "marknonparticipating",
+	Short: "Mark an account as not participating",
```
algoradam: Kind of a nitpick, but could we add something like "(i.e., offline and earns no rewards)" to the short description? Otherwise I could imagine people seeing this and marking themselves nonparticipating when they really mean to mark themselves offline.

data/transactions/keyreg.go
Vervious: This situation is similar to keyreg transactions that register an empty key, where we interpret it semantically as a different case. Whether that's good design or not, I'm ambivalent, but at least there is already a precedent in the code base.

tsachiherman:
@EvanJRichard - could you please see if you could make any progress with this PR ? ( or close it ).
It has been idling for quite a while now.

EvanJRichard:
_@tsachiherman wrote_ :
> @EvanJRichard - could you please see if you could make any progress with this PR ? ( or close it ).
> It has been idling for quite a while now.

Yeah, no problem - I'll spend time today to fix the failing tests, which should make it merge ready.

data/transactions/keyreg_test.go
```diff
@@ -85,8 +85,16 @@ func TestKeyregApply(t *testing.T) {
 	_, err = tx.Apply(mockBal, SpecialAddresses{FeeSink: feeSink}, 0)
 	require.NoError(t, err)
 
-	// Nonparticipatory accounts should not be able to change status
-	mockBal.addrs[src] = basics.BalanceRecord{Addr: src, AccountData: basics.AccountData{Status: basics.NotParticipating}}
-	_, err = tx.Apply(mockBal, SpecialAddresses{FeeSink: feeSink}, 0)
-	require.Error(t, err)
+	// Going from online to nonparticipatory should be okay, if the protocol supports that
+	if mockBal.ConsensusParams().SupportBecomeNonParticipatingTransactions {
```
algoradam: Probably we should change mockBal's version from ConsensusCurrentVersion to ConsensusFuture for this test -- otherwise we aren't testing anything.

algoradam:
We need to change the keyreg test to test against vFuture instead of vCurrent, otherwise we're not testing anything. If that passes tests, this looks good to me!

PR #262, opened 2019-08-21
[do not merge] travis experiment

Vervious:
do not review - I'm experimenting with a test on travis, for the ensure digest asynchronous pr.

### DIFF
--- a/agreement/service_test.go
+++ b/agreement/service_test.go
@@ -150,6 +150,7 @@ type testingNetwork struct {
 	compoundPocket    chan<- multicastParams
 	partitionedNodes  map[nodeID]bool
 	crownedNodes      map[nodeID]bool
+	relayNodes        map[nodeID]bool
 	interceptFn       multicastInterceptFn
 }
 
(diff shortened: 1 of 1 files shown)


Vervious:
note this test passes consistently locally.

agreement/service_test.go
```diff
@@ -404,7 +411,17 @@ func (n *testingNetwork) crown(prophets ...nodeID) {
...
+		n.crownedNodes[prophets[i]] = true
+	}
+}
+
+// Star topology with the given nodes at the center; to revert, call repairAll
+func (n *testingNetwork) makeRelays(relays ...nodeID) {
+	n.mu.Lock()
+	defer n.mu.Unlock()
+	n.relayNodes = make(map[nodeID]bool)
+	for i := 0; i < len(relays); i++ {
```
tsachiherman: you're aware of the 
```
for relay := range relays {
...
```
syntax, right ?

Vervious: or, here, `for _, relay := range relays`, since we want the nodeID, not the index.
But we conflate the two all over this file. But yes, you have a point.

agreement/service_test.go
```diff
@@ -307,6 +308,11 @@ func (n *testingNetwork) multicast(tag protocol.Tag, data []byte, source nodeID,
 				return
 			}
 		}
+		if n.relayNodes != nil {
+			if !n.relayNodes[source] && !n.relayNodes[peerid] {
+				return
```
Vervious: I figured it out. I think this should not be a return.

PR #265, opened 2019-08-23
Reorder protocol network tags into lexicographic order.

derbear:
## Summary

Reorder the tags in protocol/tags so that they stay in lexicographic order (to prevent duplicates).

### DIFF
--- a/protocol/tags.go
+++ b/protocol/tags.go
@@ -20,7 +20,6 @@ package protocol
 // e.g., the agreement service can register to handle agreements with the Agreement tag.
 type Tag string
 
-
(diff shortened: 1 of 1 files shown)


derbear:
> please do update the comment above if you'd like it to be kept ordered..

I think the intent of this wording, when we had inserted it in originally, was to keep the lexicographic ordering of the tag _values_ to detect duplicates in the values, since there's no point in detecting duplicates in the tag names (in Go they're just the same name and the compiler will refuse to compile).

What's a good way to improve the wording of the comment so that it's clearer?

tsachiherman:
maybe :

// Tags, in lexicographic sort order of tag values to avoid duplicates.

PR #266, opened 2019-08-23
Fix acctupdates refcount concurrency bug.

derbear:
## Summary

If the callback passed to the database Atomic function is rerun due to a retry
after at least one loop iteration, the flushcount variable will be too high, and
too many accounts may be evicted.  This commit fixes this bug by resetting
flushcount inside the callback.


ledger/acctupdates.go
```diff
@@ -258,10 +258,11 @@ func (au *accountUpdates) committedUpTo(rnd basics.Round) basics.Round {
 	// Keep track of how many changes to each account we flush to the
 	// account DB, so that we can drop the corresponding refcounts in
 	// au.accounts.
-	flushcount := make(map[basics.Address]int)
+	var flushcount map[basics.Address]int
 
 	offset := uint64(newBase - au.dbRound)
 	err := au.dbs.wdb.Atomic(func(tx *sql.Tx) error {
+		flushcount = make(map[basics.Address]int)
```
tsachiherman: nit : I think that it would be "cleaner" to set the `flushcount` to nil in case of an error

nice catch ;-)

PR #268, opened 2019-08-24
Add swagger to go mod tracking.

derbear:
See https://github.com/go-modules-by-example/index/blob/master/010_tools/README.md

Perhaps we should change scripts/check_deps.sh to use `go install` instead of `go get`--not sure if that makes a difference.


zeldovich:
Is this to enforce a specific version of swagger to be installed by `scripts/check_deps.sh`?

I'm not sure how successful we will be, because the user might install a different version afterwards.

But, I think it would suffice to change `GO111MODULE=off` to `GO111MODULE=on` in `scripts/check_deps.sh` and `scripts/configure_dev-deps.sh`.  And we shouldn't need this `tools/tools.go`.

winder:
@derbear should we add `GO111MODULE=on` as well?

winder:
Lets just close this ticket, we're moving away from `go-swagger`

derbear:
> Lets just close this ticket, we're moving away from go-swagger

Sounds good.

PR #280, opened 2019-08-28
Add to the README a note on making datadirs for other networks

algoradam:
In particular, mention where to find the genesis.json files


Karmastic:
We recently added `goal node create` to automate this and simplify pre-configuring aspects of it (custom config.json file).

winder:
This looks good to. `goal node create` also seems like a great idea, but is currently (at least on my machine) looking for genesis files in locations where my machine doesn't have any

Karmastic:
Yeah, it's not designed for dev machines - it's designed for user
installs.  It wouldn't 'just work' unless we modify `make install` to
install all genesis files under ~/go/bin/ like we do for normal installs.
This is intentional - we can't know where to find ./installer/genesis
because the root source path can be anywhere now and `goal` doesn't know
where to look.

PR #282, opened 2019-08-29
-- Indexer no longer returns duplicate results

rotemh:
## Summary

when there was more than one transaction in a block, Indexer returned duplicate results. This PR fixes the issue. 

## Test Plan

Added a test that verifies that no duplicate blocks are returned


rotemh:
> I'm abit supprised that the code use
> `created_at > $1 AND created_at < $2`
> rather than
> `created_at >= $1 AND created_at <= $2`
> but that distinct change looks good.

The `created_at` refers to a block and the user specifies a date. so it's meaningless.

PR #288, opened 2019-09-03
Better error message when indexer stuck in obtaining blocks

pzbitskiy:
## Summary

When indexer cannot obtain blocks from the ledger, it outputs brief report about attempt to wait.
Although if it stuck on a specific block, there is no clue what happens. So after a threshold indexer starts reporting underlying details.

Plus switch/case code changed a bit to decrease nesting.

## Test Plan

No new tests, relying on existing indexer test.

### DIFF
--- a/crypto/multisig.go
+++ b/crypto/multisig.go
@@ -78,7 +78,7 @@ func (msig MultisigSig) Preimage() (version, threshold uint8, pks []PublicKey) {
 const multiSigString = "MultisigAddr"
 const maxMultisig = 255
 
-// MultisigAddrGen identifes the exact group, version,
+// MultisigAddrGen identifies the exact group, version,
 // and devices (Public keys) that it requires to sign
 // Hash("MultisigAddr" || version uint8 || threshold uint8 || PK1 || PK2 || ...)
 func MultisigAddrGen(version, threshold uint8, pk []PublicKey) (addr Digest, err error) {
@@ -99,7 +99,7 @@ func MultisigAddrGen(version, threshold uint8, pk []PublicKey) (addr Digest, err
 	return Hash(buffer), nil
 }
 
-// MultisigAddrGenWithSubsigs is similiar to MultisigAddrGen
+// MultisigAddrGenWithSubsigs is similar to MultisigAddrGen
 // except the input is []Subsig rather than []PublicKey
 func MultisigAddrGenWithSubsigs(version uint8, threshold uint8,
 	subsigs []MultisigSubsig) (addr Digest, err error) {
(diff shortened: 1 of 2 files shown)


pzbitskiy:
After discussion with @tsachiherman decided to refactor the code a bit.

crypto/multisig.go
```diff
@@ -78,7 +78,7 @@ func (msig MultisigSig) Preimage() (version, threshold uint8, pks []PublicKey) {
 const multiSigString = "MultisigAddr"
 const maxMultisig = 255
 
-// MultisigAddrGen identifes the exact group, version,
+// MultisigAddrGen identifies the exact group, version,
```
rotemh: I believe that this file change needs to be in a separate PR.

rotemh:
The change looks good to me but it's a patch. We need to figure out where is the problem coming from. A try to index a new block happens only when `ledger.Wait` returns, which should happen *only* after a block is written to the ledger. In other words, this error should be never raised in theory. 

There is probably a race between the calls or maybe `Wait` doesn't actually provide high enough guarantee.

@derbear `ledger.Wait(B)` returns but then `ledger.Block(B)` sometimes returns an error. Any idea for why?

tsachiherman:
Yes. Consider the following:
A node starts as a non-archival and reaches rounds 1M.
Then, the node configuration changes to archival + indexer.
The indexer attempts to update starting round 1, and keep failing since the catchup will never retrieve block 1.

tsachiherman:
Also, when the Wait(r) is satisfied, it means that we're got r+n round, but it doesn't mean that we have round r.

pzbitskiy:
Yes, that is what happened. I'm working on a proper fix.

rotemh:
I see. 
Indexer is configured to not wake up when the node is not in archival mode. There is a ticket somewhere (I'll find it and refer to it) that specifies the fix. Practically speaking, we need to make sure that when someone puts a node into archival mode, the node catches up from the genesis and not from the current stamp. @pzbitskiy @tsachiherman

pzbitskiy:
I was thinking about detecting this situation and to warn the user instead
of changing catchup logic (discard current sub-chain and start over).

derbear:
> `ledger.Wait(B)` returns but then `ledger.Block(B)` sometimes returns an error. Any idea for why?

Suppose we are currently at round 1, and then we wait for round 1000. The network is actually at round 10000. At some point, catchup will download and validate block 1000, which will cause the channel to fire. There is a chance that before we wake up, catchup downloads block 2010, which will cause all blocks up to 1010 to be deleted if our node is not archival. When we wake up, we'll look for block 1000, but it'll already be deleted.

pzbitskiy:
Closing it since there is a better fix done on ledger side.

PR #295, opened 2019-09-05
Block/Tracker DB reset on switching to archival mode

pzbitskiy:
## Summary

The PR addresses an issue with node state when it is switched from non-archival to archival. In this case it never tries to get the beginning of the blockchain but instead attempts to operate in "semi-archival" mode by growing only known subsequence of blocks. There are two problems with this approach:

1. Not possible to requests blocks from the beginning of the chain.
2. Indexer (if enables) is stuck in attempts to obtain block 1 from ledger and never succeed.

Key points:

- min(rnd) != 0 is used as a sign of switch
- reset is implemented as DROP TABLE
- accounts cache also requires reset due to inconsistency with local blocks DB

Minor changes:

- LoadLedger and OpenLedger interface functions changed, archival flag is now a part of the signature; ledger.InitState is encapsulates genesis data for OpenLedger.
- SetArchival method gone
- inMemory and isArchival boolean flags are named in tests using LoadLedger/OpenLedger

## Test Plan

Two test added: ensuring that archival ledger survives after restart, and checking that non-archival to archival happens with resetting databases.


pzbitskiy:
Related to https://github.com/algorand/go-algorand/pull/288

data/txHandler_test.go
```diff
@@ -18,19 +18,18 @@ package data
...
-
+	"github.com/algorand/go-algorand/components/mocks"
 	"github.com/algorand/go-algorand/crypto"
 	"github.com/algorand/go-algorand/data/basics"
 	"github.com/algorand/go-algorand/data/pools"
 	"github.com/algorand/go-algorand/data/transactions"
 	"github.com/algorand/go-algorand/logging"
 	"github.com/algorand/go-algorand/protocol"
 	"github.com/algorand/go-algorand/util/execpool"
+	"github.com/stretchr/testify/require"
```
tsachiherman: this line should be moved back.. part of our coding conventions:
imports goes into three groups:
1. go imports
2. non-algorand imports
3. algorand imports

within each group, items are lexicographically ordered.

ledger/acctupdates.go
```diff
@@ -90,17 +90,38 @@ func (au *accountUpdates) loadFromDisk(l ledgerForTracker) error {
...
 		}
 
-		au.dbRound, err0 = accountsRound(tx)
+		au.dbRound, err0 = initAux()
 		if err0 != nil {
 			return err0
 		}
+		// Check for blocks DB and tracker DB un-sync
+		if au.dbRound > latest {
+			// TODO: add assertions about non-archival to archival switch?
```
tsachiherman: I don't know about assertions, but clearly you need to log some info messages.

ledger/blockdb.go
```diff
@@ -208,6 +222,22 @@ func blockLatest(tx *sql.Tx) (basics.Round, error) {
 	return 0, fmt.Errorf("no blocks present")
 }
 
+// Returns amount of blocks in Block DB
+// At least one row (genesis block) must present, empty table is threated as error
+func blockCount(tx *sql.Tx) (uint64, error) {
+	var count sql.NullInt64
+	err := tx.QueryRow("SELECT COUNT(rnd) FROM blocks").Scan(&count)
```
tsachiherman: Using `SELECT COUNT(1)` should be faster.

ledger/blockdb.go
```diff
@@ -208,6 +222,22 @@ func blockLatest(tx *sql.Tx) (basics.Round, error) {
...
+	if err != nil {
+		return 0, err
+	}
+
+	if !count.Valid || count.Int64 == 0 {
```
tsachiherman: Consider using coalesce to avoid having this test.
https://www.sqlitetutorial.net/sqlite-functions/sqlite-coalesce/

ledger/accountdb.go
```diff
@@ -52,6 +52,12 @@ var accountsSchema = []string{
 		data blob)`,
 }
 
+var accountsResetExprs = []string{
```
tsachiherman: You might want to change this into `DROP TABLE IF EXISTS [tablename]` so that it won't generate an error if one of the tables are missing while the other exists.

ledger/acctupdates.go
```diff
@@ -90,17 +90,38 @@ func (au *accountUpdates) loadFromDisk(l ledgerForTracker) error {
 		return fmt.Errorf("accountUpdates.loadFromDisk: initAccounts not set")
 	}
 
+	latest := l.Latest()
 	err := au.dbs.wdb.Atomic(func(tx *sql.Tx) error {
 		var err0 error
-		err0 = accountsInit(tx, au.initAccounts, au.initProto)
-		if err0 != nil {
-			return err0
+		initAux := func() (basics.Round, error) {
```
zeldovich: Can this be a separate function?  Would be easier to understand.

ledger/blockdb.go
```diff
@@ -208,6 +222,22 @@ func blockLatest(tx *sql.Tx) (basics.Round, error) {
 	return 0, fmt.Errorf("no blocks present")
 }
 
+// Returns amount of blocks in Block DB
+// At least one row (genesis block) must present, empty table is threated as error
```
zeldovich: This is pretty expensive, and will cause algod to take a very long time to start.

How about using the min round number to check for archival?  `SELECT MIN(rnd) FROM blocks` should be much faster, and if that's 0, you have all of the blocks.

ledger/ledger.go
```diff
@@ -69,16 +68,25 @@ type Ledger struct {
...
+
 // OpenLedger creates a Ledger object, using SQLite database filenames
-// based on dbPathPrefix (in-memory if dbMem is true).  initBlocks and
-// initAccounts specify the initial blocks and accounts to use if the
+// based on dbPathPrefix (in-memory if dbMem is true). seed.initBlocks and
+// seed.initAccounts specify the initial blocks and accounts to use if the
 // database wasn't initialized before.
-func OpenLedger(log logging.Logger, dbPathPrefix string, dbMem bool, initBlocks []bookkeeping.Block, initAccounts map[basics.Address]basics.AccountData, genesisHash crypto.Digest) (*Ledger, error) {
+func OpenLedger(
+	log logging.Logger, dbPathPrefix string, dbMem bool, seed InitState, isArchival bool,
```
tsachiherman: don't refer to the InitState as seed; the usage of the term `seed` is already overloaded as is.
genesisInitState would be much cleaner.

ledger/ledger.go
```diff
@@ -69,16 +68,25 @@ type Ledger struct {
 	headerCache heapLRUCache
 }
 
+// InitState structure defines blockchain init params
+type InitState struct {
+	InitBlocks   []bookkeeping.Block
```
tsachiherman: Drop the "Init" prefix from the variable, since it's already reflected in the data structure name.
No need for array of blocks.

ledger/archival_test.go
```diff
@@ -132,6 +147,134 @@ func TestArchival(t *testing.T) {
...
+	blk := seed.InitBlocks[0]
+
+	const maxBlocks = 2000
+	for i := 0; i < maxBlocks; i++ {
+		blk.BlockHeader.Round++
+		blk.BlockHeader.TimeStamp += int64(crypto.RandUint64() % 100 * 1000)
+		l.AddBlock(blk, agreement.Certificate{})
+
+		// Don't bother checking the trackers every round -- it's too slow..
+		if crypto.RandUint64()%23 > 0 {
```
tsachiherman: Can we avoid using randomness here ? rationale is that we want our tests to be repetitive ( if we can ).

ledger/blockdb.go
pzbitskiy: @tsachiherman 
Measured performance with 13565K blocks in the DB (roughly equals to blocks from 2 years of runtime).
COUNT(1):       13,564,192 - 3m34.966600652s
COUNT(rnd):   13,564,192 - 3m36.368089268s

Table without deletion:
MIN(rnd):  387,385 - 2.120839ms

Table after  ~5M records deleted:
MIN(rnd):  5,000,000 - 1.201572ms

ledger/ledger.go
pzbitskiy: Removed Init, switching from an array of blocks to a single block seemed to large, will do as a separate PR

ledger/ledger.go
pzbitskiy: Done

ledger/blockdb.go
pzbitskiy: Done

ledger/acctupdates.go
pzbitskiy: Done

ledger/accountdb.go
pzbitskiy: Done

ledger/blockdb.go
pzbitskiy: Switched to MIN, not actual anymore

ledger/blockdb.go
pzbitskiy: Switched to MIN instead

ledger/acctupdates.go
pzbitskiy: Done

ledger/archival_test.go
pzbitskiy: Removed the check, not needed anymore

ledger/ledger_test.go
```diff
@@ -196,8 +203,10 @@ func TestLedgerBlockHeaders(t *testing.T) {
 	backlogPool := execpool.MakeBacklog(nil, 0, execpool.LowPriority, nil)
 	defer backlogPool.Shutdown()
 
-	initBlocks, initAccounts, _ := testGenerateInitState(t, protocol.ConsensusCurrentVersion)
-	l, err := OpenLedger(logging.Base(), t.Name(), true, initBlocks, initAccounts, crypto.Hash([]byte(t.Name())))
+	genesisInitState, _ := testGenerateInitState(t, protocol.ConsensusCurrentVersion)
+	const inMem = true
```
tsachiherman: nit: Your intuition to pass in variables here ( for readability ), it correct.
However, I would use constants for that, so that we won't need to define variables over and over.

pzbitskiy:
Rebased on top of Nikolai's assets change.

PR #297, opened 2019-09-08
fix benchmarks, make txpool more concurrent

zeldovich:
Some simple fixes to benchmarks, and an improvement to the txpool that makes it more concurrent, which is important for making progress under load (like in the e2e throughput benchmark).


config/config.go
```diff
@@ -415,17 +415,9 @@ func initConsensusTestProtocols() {
...
-		MaxVersionStringLen: 64,
-
-		MaxTxnBytesPerBlock: 100000000,
-		DefaultKeyDilution:  10000,
-
-		ApprovedUpgrades: map[protocol.ConsensusVersion]bool{},
-	}
+	testBigBlocks := Consensus[protocol.ConsensusCurrentVersion]
+	testBigBlocks.ApprovedUpgrades = map[protocol.ConsensusVersion]bool{}
+	Consensus[protocol.ConsensusTestBigBlocks] = testBigBlocks
```
tsachiherman: I was going to write that you just broke the ConsensusTestBigBlocks configuration.
But then I was looking where it was used. I can't find anywhere in the codebase that we actually use this.
If we don't use it, shouldn't we just eliminate it ?

zeldovich: It's referenced by name from `test/testdata/nettemplates/perf/*.json`.  But indeed you're right that I forgot to actually give it big blocks. :-)

Let me merge this PR, and then do another PR to fix up the big blocks (it doesn't matter too much, since it's just the e2e perf test).

PR #319, opened 2019-09-09
Change IPOrDNSName -> Address for relay deployment tooling.

tsachiherman:
## Summary

Change field name IPOrDNSName to Address.

The foundation changed their field name from IPOrDNSName to Address.
This change corresponds to their internal tooling.

## Test Plan

I used to tool and it works flawlessly after the change was made.


cmd/algorelay/eb/eb.go
```diff
@@ -19,7 +19,7 @@ package eb
 // Relay represents the configuration data necessary for a single Relay
 type Relay struct {
 	ID             int64 // db key injected when loaded
-	IPOrDNSName    string
+	Address        string
```
winder: Address is an overloaded term in our system, could you add `// IP / DNS name` here?

tsachiherman: done.

PR #321, opened 2019-09-10
avoid race conditions in the fee tracker

zeldovich:
This one actually got caught by our e2e tests running with race detection
enabled.


zeldovich:
> Looks good to me.
> Although I'm a bit surprised. Where do we process blocks in parallel? Isn't it all sequential?

Calls to `EstimateFee()` can happen concurrently with calls to `ProcessBlock()`.

PR #323, opened 2019-09-10
Ledger closing

pzbitskiy:
## Summary

Prevent possible double closing of ledger on error in OpenLedger
Added ledger closing in tests

## Test Plan

Trivial change, no new test


ledger/ledger.go
```diff
@@ -244,9 +241,11 @@ func initBlocksDB(tx *sql.Tx, l *Ledger, initBlocks []bookkeeping.Block, isArchi
 // and goroutines used by trackers).
 func (l *Ledger) Close() {
 	l.trackerDBs.close()
-	l.blockQ.close()
 	l.blockDBs.close()
 	l.trackers.close()
+	if l.blockQ != nil {
+		l.blockQ.close()
```
tsachiherman: nit: if you're already testing for blockQ being nil, then you might consider clearing it out after ( or even better - before ) calling the close.

pzbitskiy: Done

PR #324, opened 2019-09-11
-- Add API end point Pending by address

rotemh:
## Summary

This PR add an end point for pending transaction by address

### DIFF
--- a/daemon/algod/api/server/v1/handlers/errors.go
+++ b/daemon/algod/api/server/v1/handlers/errors.go
@@ -25,6 +25,7 @@ var (
 	errFailedParsingRoundNumber            = "failed to parse the round number"
 	errFailedToParseAddress                = "failed to parse the address"
 	errFailedToParseTransaction            = "failed to parse transaction"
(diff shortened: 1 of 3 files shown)


tsachiherman:
Few nits -
1. I would have embed the max size trimming into the for _, twr := range txs
    That way, you would avoid encoding unneeded entries.
2. I was kind of surprised to see max defined as integer rather than
unsigned integer.
3. The max parameter logic is optional, but I'm not sure if the error case
is the one that we want.
    I agree that it should be an optional parameter. I'm not sure that I
like the fact that we default to 0 if we fail to process the value.
    Instead, we should error so that the user would have an indication that
an invalid value was passed.
4. We need some sort of unit testing for this functionality, plus support
in our API.

with all the above being said, it looks perfectly good change, and I'll be
glad to approve it tomorrow morning when I have a chance.

rotemh:
Valid points - Thanks for the great comments.  

1. Done. 
2. It is unsigned int, in the swagger spec it says int64 because of swagger limitations...
3. Updated. 
4. That will get done as part of a wider API testing support.

PR #330, opened 2019-09-13
Add support for external ARM64 compiler

tsachiherman:
## Summary

Add travis support for compiling ARM64 binaries using external build host server.

## TODO
Still need to extend this to ARM binaries ( Raspberry PI ), but it will be a separate PR.


tsachiherman:
Closed PR due to CLAassistant issues; re-created in https://github.com/algorand/go-algorand/pull/334

PR #331, opened 2019-09-13
Serialize concurrent access to data in the auction tracker.

derbear:
This fixes data races when calling auction tracker methods concurrently.

This does not enforce the immutability of Auctions and AuctionKey. Thus, a better design for the Tracker may be not to expose Auctions and AuctionKey as modifiable by external code.


derbear:
(This issue was caught by the race detector on Travis: https://travis-ci.com/algorand/go-algorand/jobs/227763780)

PR #332, opened 2019-09-16
Eric/goal2 990 pipeline staging area

egieseke:
## Summary

Updated create_and_deploy_recipe.sh to support skipping build or deploy steps.  

## Test Plan

Tested with CICD Jenkins pipeline.


scripts/deploy_private_version.sh
```diff
@@ -106,6 +106,7 @@ else
     cp ${GENESISFILE} gen/${DEFAULTNETWORK}/genesis.json
 fi
 
+
```
tsachiherman: unless you're intentionally modifying this file, please revert this formatting.

scripts/create_and_deploy_recipe.sh
```diff
@@ -97,17 +97,26 @@ if [[ "${NETWORK}" = "" ]]; then
     NETWORK=${CHANNEL}
 fi
 
+# Build and Deploy binaries
+if [[ ${SKIP_BUILD_DEPLOY} != "true" ]]; then
```
tsachiherman: this one has to be tighted to the netgoal execution, since the net goal execution would not be there unless you've built it first.

scripts/create_and_deploy_recipe.sh
```diff
@@ -107,7 +114,10 @@ ${GOPATH}/bin/netgoal build -r "${ROOTDIR}" -n "${NETWORK}" --recipe "${RECIPEFI
 export S3_RELEASE_BUCKET="${S3_RELEASE_BUCKET}"
 ${SRCPATH}/scripts/upload_config.sh "${ROOTDIR}" "${CHANNEL}"
 
-if [ "${NO_DEPLOY}" = "" ]; then
-    # Now generate a private build using our custom genesis.json and deploy it to S3 also
-    ${SRCPATH}/scripts/deploy_private_version.sh -c "${CHANNEL}" -f "${ROOTDIR}/genesisdata/genesis.json" -n "${NETWORK}" -b "${S3_RELEASE_BUCKET}"
+# Build and Deploy binaries
+if [[ ${SKIP_BUILD_DEPLOY} != "true" ]]; then
```
tsachiherman: when comparing bash variables, you need to ensure placing them in double quotes to avoid it being rendered empty causing the script to fail.

scripts/deploy_private_version.sh
egieseke: reverted

scripts/create_and_deploy_recipe.sh
egieseke: included check for existing netgoal binary.

scripts/create_and_deploy_recipe.sh
egieseke: added double quotes around variable.

PR #333, opened 2019-09-18
Rename TxGroup.Transactions to TxGroup.TxGroupHashes

pzbitskiy:
## Summary

Rename TxGroup.Transactions to TxGroup.TxGroupHashes that appears to be a better name.

## Test Plan

Rely on existing tests.


pzbitskiy:
@derbear - for the context see https://github.com/algorand/go-algorand-sdk/pull/49#discussion_r325437965

PR #334, opened 2019-09-18
Add support for external ARM64 compiler

tsachiherman:
## Summary

Add travis support for compiling ARM64 binaries using external build host server.

## TODO
Still need to extend this to ARM binaries ( Raspberry PI ), but it will be a separate PR.


scripts/travis/external_build.sh
```diff
@@ -0,0 +1,101 @@
...
+
+if [ "${BUILD_TARGET}" = "" ]; then
+    echo "error: missing build target argument"
+    exit 1
+fi
+
+sudo apt-get install awscli
+
+# create build request
+echo "{ \"TRAVIS_BRANCH\" : \"${TRAVIS_BRANCH}\", \"TRAVIS_COMMIT\" : \"${TRAVIS_COMMIT}\", \"TRAVIS_PULL_REQUEST\" : \"${TRAVIS_PULL_REQUEST}\", \"AWS_ACCESS_KEY_ID\" : \"${AWS_ACCESS_KEY_ID}\", \"AWS_SECRET_ACCESS_KEY\" : \"${AWS_SECRET_ACCESS_KEY}\", \"EXEC\":\"${BUILD_TARGET}\" }" > ${TRAVIS_BUILD_NUMBER}.json
```
pzbitskiy: makes sense to output params to travis log as well

scripts/buildhost/linux-arm64.sh
```diff
@@ -0,0 +1,75 @@
...
+
+AWS_REGION="us-west-2"
+AWS_LINUX_AMI="ami-0c579621aaac8bade"
+AWS_INSTANCE_TYPE="a1.2xlarge"
+INSTANCE_NUMBER=$RANDOM
+
+set +e
+
+${SCRIPTPATH}/start_ec2_instance.sh ${AWS_REGION} ${AWS_LINUX_AMI} ${AWS_INSTANCE_TYPE}
+if [ "$?" != "0" ]; then
```
pzbitskiy: shouldn't shutdown_ec2_instance be called here as linux-arm.sh does?

scripts/buildhost/start_ec2_instance.sh
```diff
@@ -0,0 +1,107 @@
+#!/usr/bin/env bash
```
pzbitskiy: may be to use terraform for provision? Easier/more maintainable than a bash script

scripts/buildhost/linux-arm64.sh
tsachiherman: yes; I'll handle this on subsequent PR.

scripts/buildhost/start_ec2_instance.sh
tsachiherman: Using terraform to deploy a single machine wouldn't give us that much. I'm not sure the overhead here would justify that. ( maybe I'm talking from position, as I've had lots of technical issues with terraform in the past )

scripts/travis/external_build.sh
tsachiherman: noted; I'll add this in a subsequent PR.

.travis.yml
```diff
@@ -21,6 +21,9 @@ stages:
     if: branch != rel/stable AND (branch =~ /^rel\//) AND type != pull_request
 
 jobs:
+  allow_failures:
+    - name: External ARM64 Build
+    - name: External ARM64 Deploy
```
winder: Presumably we want to remove this in the future?

scripts/buildhost/start_ec2_instance.sh
winder: I agree with Pavel, these sorts of things should be done with a higher level tool in the future. We use ansible/terraform elsewhere so hopefully this can be rolled up into our other infrastructure at some point. EC2 instances can also be stopped / resumed, rather than completely reprovisioned each time if the goal is cost saving.

scripts/travis/configure_dev.sh
```diff
@@ -1,5 +1,56 @@
...
 
+# keep script execution on errors
+set +e
+
 SCRIPTPATH="$( cd "$(dirname "$0")" ; pwd -P )"
 
+OS=$(${SCRIPTPATH}/../ostype.sh)
+ARCH=$(${SCRIPTPATH}/../archtype.sh)
+
+if [ "${OS}" = "linux" ]; then
```
winder: Seems like there should be a better way to do this. Can we set the `ARCH` to `armv61` and use that variable?

.travis.yml
tsachiherman: Once we feel that it's stable enough; I don't want the external build to break travis for now.

scripts/buildhost/start_ec2_instance.sh
tsachiherman: We need to run a *single* task at a time. Using ansible/terraform for that won't save us much.
Stopping/resuming instances is also an issue. We want to start from a "fresh" state every time.
( just like travis )

scripts/travis/configure_dev.sh
tsachiherman: I'm confused about the context of your comment. Please stop by, and let's discuss that.

PR #335, opened 2019-09-18
Add support for travis ARM64 integration tests

tsachiherman:
## Summary

This PR adds the following functionality:
1. Add ARM64 integration testing support for travis PRs and releases.
2. Improve ARM64 builds performance by avoid installing unneeded components on travis machine.
3. Improve error handling of buildhost on arm64 builds.
4. Pass the following variables from travis to the ephermal build machines:
    TRAVIS_BRANCH
    TRAVIS_COMMIT
    TRAVIS_PULL_REQUEST
5. Simplify the travis scripts by invoking a integration_test.sh or build_test.sh; The BUILD_TYPE="integration" would be set by these scripts instead of Travis. That allows us to invoke the same functionality from other sources. ( i.e. ephermal build machine )
6. Replace the usage of TRAVIS_BUILD_NUMBER with TRAVIS_JOB_NUMBER allowing multiple concurrent external builds for the same build.


.travis.yml
```diff
@@ -33,42 +34,66 @@ jobs:
...
-        - scripts/travis/build.sh || travis_terminate 1;
-        - travis_wait 90 scripts/travis/test.sh
+        - scripts/travis/build_test.sh
+    - # same stage, parallel job
+      name: External ARM64 Integration test
+      os: linux
+      env:
+        - BUILD_TYPE: "external_build"
+        - TARGET_PLATFORM: "linux-arm64"
+      addons:
```
pzbitskiy: this is nice, did not know travis has it

scripts/travis/external_build.sh
```diff
@@ -52,29 +52,31 @@ if [ "${BUILD_TARGET}" = "" ]; then
     exit 1
 fi
 
-sudo apt-get install awscli
+BUILDID=${TRAVIS_JOB_NUMBER/./000}
```
pzbitskiy: I'd like to see a comment example what is in TRAVIS_JOB_NUMBER and what is in BUILDID after replacement (that is not well-known feature)

tsachiherman: will do in subsequent PR.

scripts/travis/travis_wait.sh
```diff
@@ -0,0 +1,69 @@
+#!/bin/bash
+#
+# Copyright (c) 2016 Travis CI GmbH <contact@travis-ci.org>
+#
+# SPDX-License-Identifier: MIT
+#
+# Based on the travis-ci/travis-build
```
winder: I never knew about this. Never really thought about how the YML is actually executed. A link to the actual source files may be useful for future debugging:
https://github.com/travis-ci/travis-build/blob/d8f183b29700c3b98b6995ecc4b463f3c3bc552f/lib/travis/build/bash/travis_jigger.bash
https://github.com/travis-ci/travis-build/blob/d8f183b29700c3b98b6995ecc4b463f3c3bc552f/lib/travis/build/bash/travis_wait.bash

PR #341, opened 2019-09-20
Deploy rel/stable MacOS builds

pzbitskiy:
## Summary

* Enable macos build for PR
* Ignore empty swagger.json.validated files (artifacts addon fails to upload them)
* Set macos version to the latest available 10.14
* Enable deploy build for rel/stable

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -66,6 +66,12 @@ jobs:
             - awscli
       script:
         - scripts/travis/external_build.sh ./scripts/travis/integration_test.sh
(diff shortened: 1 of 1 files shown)


tsachiherman:
The empty swagger.json.validated files might be a bigger issue.
Let's talk about that on Monday.

tsachiherman:
Change looks good to me; I'll let @winder see if that would resolve the issue he had with MacOS builds.

PR #342, opened 2019-09-21
Pass on the S3_RELEASE_BUCKET environment variable.

tsachiherman:
## Summary

1. The `S3_RELEASE_BUCKET` variable wasn't passed to the ephemeral build machine, causing it to upload to the default bucket instead of the designated bucket. 
2. The invocation of the build is using pipes to send the output to a secondary script.
    When testing the exit code, we always received 0 since this was the error generated by the second script. To play it safe, we need to take the first non-zero error code between the two.
3. When we call `deploy_packages.sh`, we want to abort the script if we fail to build.

## Test Plan

Run it, test it.


pzbitskiy:
I personally think we need to rename/get rid of S3_RELEASE_BUCKET variable in build scripts and leave it only in update scripts. It is used differently in these cases.

PR #348, opened 2019-09-23
Revisit s3_release_bucket usage in build scripts

pzbitskiy:
## Summary

Make S3_RELEASE_BUCKET specification explicit in build scripts.

### DIFF
--- a/scripts/deploy_version.sh
+++ b/scripts/deploy_version.sh
@@ -12,6 +12,7 @@
 #
 # Usage:    Can be used locally or on Travis to generate and upload a new build for auto-updates.
 #           Currently called by travis/deploy_packages.sh
+#           Expects target S3 bucket to be set as S3_RELEASE_BUCKET environment variable.
 #
 # Examples: scripts/deploy_version.sh my/testbranch linux/amd64
 #           scripts/deploy_version.sh my/testbranch linux/amd64 darwin/amd64
@@ -23,6 +24,11 @@ set -e
 if [ "$#" -lt 2 ]; then
     echo "Syntax: deploy_version <branch> <os>/<arch> ..."
     echo "e.g. deploy_version master darwin/amd64 linux/amd64"
+    exit 1
+fi
+
+if [[ -z "${S3_RELEASE_BUCKET}" ]]; then
+    echo "Target S3 bucket must be set as S3_RELEASE_BUCKET env var"
     exit 1
 fi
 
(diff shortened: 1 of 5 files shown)


scripts/deploy_version.sh
```diff
@@ -23,6 +24,7 @@ set -e
 if [ "$#" -lt 2 ]; then
     echo "Syntax: deploy_version <branch> <os>/<arch> ..."
     echo "e.g. deploy_version master darwin/amd64 linux/amd64"
+    echo "Target S3 bucket must be set as S3_RELEASE_BUCKET env var"
```
tsachiherman: You'll need to add explicit testing for this env variable in this script.

scripts/upload_version.sh
```diff
@@ -12,18 +12,14 @@
 # Usage:    Should be used after build_packages.sh completes successfully
 #           Currently called by deploy_version.sh
 #
-# Examples: scripts/upload_version.sh nightly $HOME/node_pkg/
+# Examples: scripts/upload_version.sh nightly $HOME/node_pkg/ algorand-releases
 
 if [[ "$#" -ne 2 && "$#" -ne 3 ]]; then
-    echo "Syntax: upload_version <channel> <directory-with-packages> <optional: bucket>"
+    echo "Syntax: upload_version <channel> <directory-with-packages> <bucket>"
```
tsachiherman: The if clause above is no longer correct. the number of argument need to be exactly 3.

pzbitskiy: done

pzbitskiy:
Thanks for catching this.

winder:
This seems fine to me.

Could you update this document to reflect these changes:
https://algorand.atlassian.net/wiki/spaces/DEVOPS/pages/420118553/Usage+Scenarios

pzbitskiy:
> Looks like `scripts/travis/deploy_packages.sh` needs to be updated as well.

@winder what exactly needs to be changed there? deploy_packages.sh is travis script that relies on environment.

pzbitskiy:
> Could you update this document to reflect these changes:

Looks up to date.

pzbitskiy:
> I think that we'll need a corresponding change in upload_config.sh.

Done

pzbitskiy:
1. Checked that deploying private network works as before.
2. Fixed missed hexdump error in build in Docker (discovered when checked p.1)

PR #351, opened 2019-09-24
Correct the log message entry

tsachiherman:
## Summary

When being used with travis directly, the message is valid and correct.
However, when invoked from our scripts, we want to have it emit the actual timed-command, rather then the travis-invoked script.

### DIFF
--- a/scripts/travis/travis_wait.sh
+++ b/scripts/travis/travis_wait.sh
@@ -55,9 +55,9 @@ travis_wait() {
   } || return 1
 
   if [ $result -eq 0 ]; then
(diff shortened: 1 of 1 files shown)


tsachiherman:
> This looks fine. But while you're editing this file, maybe we should add the `ANSI_X` definitions (or stop using them):
> 
> ```
> ANSI_RED="\033[31;1m"
> ANSI_GREEN="\033[32;1m"
> ANSI_RESET="\033[0m"
> ANSI_CLEAR="\033[0K"
> ```

That's good feedback; I'll add that to a separate PR.

PR #353, opened 2019-09-26
Validate "go get" output

tsachiherman:
## Summary

Running `go get -u <package name>` won't always return a non-zero error code when an error occur ( in particular, when the compilation of the downloaded package fails ).

That behavior prevents us from providing correct handing for that error case, and as a result we would fail the build down the road.

## Proposed Solution

The expected output on the success use case is always an empty string. By capturing the output and comparing it to an empty string ( in addition to existing error code testing ), we can detect build errors of downloaded components.

### DIFF
--- a/scripts/configure_dev-deps.sh
+++ b/scripts/configure_dev-deps.sh
@@ -4,8 +4,7 @@ set -ex
 
 function install_go_module {
     local OUTPUT
-    GO111MODULE=off
-    OUTPUT=$(go get -u $1 2>&1)
+    OUTPUT=$(GO111MODULE=off go get -u $1 2>&1)
     if [ "${OUTPUT}" != "" ]; then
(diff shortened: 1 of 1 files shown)


scripts/configure_dev-deps.sh
```diff
@@ -2,6 +2,15 @@
 
 set -ex
 
-GO111MODULE=off go get -u golang.org/x/lint/golint
-GO111MODULE=off go get -u golang.org/x/tools/cmd/stringer
-GO111MODULE=off go get -u github.com/go-swagger/go-swagger/cmd/swagger
+function install_go_module {
+    local OUTPUT
+    OUTPUT=$(GO111MODULE=off go get -u $1 2>&1)
```
pzbitskiy: how about also checking return code for robustness?

PR #365, opened 2019-10-01
Add Raspberry PI build machine to travis

tsachiherman:
## Summary

This PR add the required support for compiling a 32-bit ARM ( arm6 ) binaries. 
The build is being sent to the build host and executed on a spawned EC2 instance that contains a QEMU VM.

Given that the build takes a long time to complete ( more than two hours ), it will always appear to be failing on travis, but the uploaded binaries should be available.
( which is why I've included that only in the deploy stage for now ).


scripts/travis/build.sh
```diff
@@ -5,12 +5,17 @@ set +e
...
+OS=$("${SCRIPTPATH}/../ostype.sh")
+ARCH=$("${SCRIPTPATH}/../archtype.sh")
+
 # travis sometimes fail to download a dependency. trying multiple times might help.
 for (( attempt=1; attempt<=5; attempt++ ))
 do
     scripts/travis/configure_dev.sh
     ERR=$?
-    if [ "$?" == "0" ]; then
+    if [ "${ERR}" == "0" ]; then
```
tsachiherman: This one looks trivial, but I've been looking for this for a while.

scripts/travis/configure_dev.sh
```diff
@@ -29,9 +28,12 @@ if [ "${OS}" = "linux" ]; then
                 exit 1
             fi
         fi
+        set -e
+        sudo apt-get update -y
```
pzbitskiy: update might be slow... maybe just install sqlite?

scripts/buildhost/linux-arm64.sh
```diff
@@ -79,9 +79,30 @@ fi
...
+
 set -o pipefail
 ssh -i key.pem -o "StrictHostKeyChecking no" ubuntu@$(cat instance) 'bash -s' < exescript 2>&1 | ${SCRIPTPATH}/s3streamup.sh s3://${BUCKET}/${LOGFILE} ${NO_SIGN}
 ERR=$?
-exitWithError ${ERR} ""
+ps -p$timeout_monitor_pid &>/dev/null && kill $timeout_monitor_pid
+if [ "${ERR}" = "0" ]; then
+    exitWithError ${ERR} ""
+else
+    exitWithError ${ERR} "Failed building with error code ${ERR} on $(cat instance)"
```
winder: Should you quote `${ERR}` here?

tsachiherman: Yes, `${ERR}` should be quoted here. I will do that on subsequent PR.

PR #373, opened 2019-10-04
Optimize travis compilation stages

tsachiherman:
## Summary

This change covers multiple aspects of our travis building logic, targeted specifically for making arm builds viable:
1. Compile source code only once during arm build.
2. Compile source code only once during deploy stage.
3. Avoid compiling the `race` variant for deploy stage. ( note : go doesn't support --race on linux/arm )
4. Improve logging while waiting for an external build to complete.
5. Prevent having travis killing an external build with an error message. Preemptively abort the operation with "success".
6. Improve log streamer responsiveness by deleting subsequent file *after* uploading log chuck to allow better parallelism.
7. Revert arm-specific compilation mode ( one package at a time ). This is no longer needed once I increased the swap file size.


scripts/travis/configure_dev.sh
```diff
@@ -54,7 +54,10 @@ if [ "${OS}" = "linux" ]; then
         fi
         set -e
         sudo apt-get update -y
-        sudo apt-get -y install sqlite3
+        sudo apt-get upgrade -y
+        sudo apt-get -y install sqlite3 libssl-dev git apt-transport-https openssl
+        git config --global http.postBuffer 524288000
+        git config --global http.sslVerify false
```
algoradam: Why do we want to disable ssl verification for git on travis?

tsachiherman: ( it's a test. it's never intended to be submitted as such )

scripts/buildhost/s3streamup.sh
```diff
@@ -13,31 +13,30 @@ DEST=$1
...
 
 SEQ=1
+aws s3 rm ${DEST}-${SEQ} ${NO_SIGN}
+aws s3 rm ${DEST}-$((SEQ+1)) ${NO_SIGN}
 SEND_NEXT=$((SECONDS+10))
 BUFFER=
 while read line; do
         BUFFER=${BUFFER}${line}$'\n'
         if [ ${#BUFFER} -gt 10240 ]; then
-                NEXT_SEQ=$((SEQ+1))
```
pzbitskiy: not sure what happens here

scripts/travis/test.sh
```diff
@@ -2,6 +2,15 @@
...
+OS=$("${SCRIPTPATH}/../ostype.sh")
+ARCH=$("${SCRIPTPATH}/../archtype.sh")
+
+if [ "${OS}-${ARCH}" = "linux-arm" ]; then
+    # for arm, no tests need to be invoked.
+    exit 0
+fi
+
 make fixcheck
-if [[ "$TRAVIS_OS_NAME" != "osx" ]]; then scripts/travis/run_tests.sh; fi
```
pzbitskiy: why is the change needed? to unify with other scripts?

scripts/travis/deploy_packages.sh
```diff
@@ -12,4 +12,6 @@ set -e
 scripts/travis/build.sh
 
 export RELEASE_GENESIS_PROCESS=true
+export NO_BUILD=true
+export SkipCleanCheck=1
```
pzbitskiy: assumption of clean build on travis?

scripts/travis/external_build_printlog.sh
```diff
@@ -44,7 +45,19 @@ while [ $SECONDS -lt $end ]; do
...
+    if [ "${TRAVIS}" = "true" ]; then
+        # under travis, if we have passed the 1h:45m mark, the build is going to likely fail due to timeout.
+        # instead of failing, we want to exit the build with success, indicating where the log could be retrieved later on.
+        # ( note that there migth be an issue with that build, but we don't want to cap it via travis timeouts. )
+        if [ $((SECONDS-LOG_WAITING_START)) -gt $((60*105)) ]; then
+            echo "Build is taking too long. Travis is going to timeout this build, so we'll tell travis that we're done for now."
+            echo "Once this build is complete, you could get a complete log by typing:"
+            echo "./scripts/travis/external_build_printlog.sh \"${BUILD_LOG_PATH}\" \"${BUILD_COMPLETE_PATH}\" ${NO_SIGN_REQUEST}"
+            exit 0
+        fi
```
pzbitskiy: This is counterintuitive. Timeout == something wrong => fail

scripts/travis/test.sh
tsachiherman: I've just added a query for OS and PLATFORM few lines above this one. It would be weird to have the TRAVIS variables in one place and the OS/PLATFORM in another, so I modified them.

scripts/travis/deploy_packages.sh
tsachiherman: No. the `scripts/travis/build.sh` before already perform a complete build, plus check the enlistment.
so I was avoiding doing the same work twice. ( which is meaningful on slow arm builds )

scripts/travis/external_build_printlog.sh
tsachiherman: True, but unfortunately, this is the success case. I would have increased the timeout, but travis doesn't have this feature. 
If the build is going to take 3.5 hours, I don't want travis to tell me that the build is broken all the time ( that won't help me. Instead, printing this message would.. )

scripts/buildhost/s3streamup.sh
tsachiherman: instead of "delete log X+1, write to log X", I changed the logic to "write to log X, delete log X+2".

PR #374, opened 2019-10-07
Move TelemetryURL to SRV Record

winder:
## Summary

Lookup telemetry URL via SRV records.

## Test Plan

Created telemetry.devnet.algodev.network SRV record and have been using that for testing.


cmd/algod/main.go
```diff
@@ -282,6 +256,43 @@ func main() {
 	}
 	fmt.Fprintf(os.Stdout, "Deadlock detection is set to: %s (Default state is '%s')\n", deadlockState, config.DefaultDeadlock)
 
+	if log.GetTelemetryEnabled() {
+		// Make a copy of config and reset DNSBootstrapID in case it was disabled.
+		cfgCopy := cfg
+		cfgCopy.DNSBootstrapID = telemetryDNSBootstrapID
+
+		// Periodically check SRV records for new telemetry URI
+		go srvUpdaterLoop(1*time.Minute, cfgCopy, s.Genesis.Network, log)
```
tsachiherman: I'm not opposing to periodically checking the SRV records, but why not to do that in the network library itself ?
This function `func (wn *WebsocketNetwork) meshThread()` is doing exactly that type of work.
Just add the time-check and make the SRV lookup there.

This code doesn't have the shutdown process. Doing it at the network library save you from implementing that "responsibility".

cmd/algod/main.go
```diff
@@ -282,6 +256,43 @@ func main() {
...
+			Branch:       currentVersion.Branch,
+			Channel:      currentVersion.Channel,
+			InstanceHash: crypto.Hash([]byte(absolutePath)).String(),
+		}
+
+		log.EventWithDetails(telemetryspec.ApplicationState, telemetryspec.StartupEvent, startupDetails)
+
+		// Send a heartbeat event every 10 minutes as a sign of life
+		ticker := time.NewTicker(10 * time.Minute)
+		go func() {
```
tsachiherman: This is another place where we need to shut down correctly. It's ok if you intend to do that in a subsequent change.

logging/telemetryhook.go
```diff
@@ -73,22 +80,31 @@ func createAsyncHook(wrappedHook logrus.Hook, channelDepth uint, maxQueueDepth i
...
-	select {
-	case <-hook.quit:
-		return false
-	case entry := <-hook.entries:
-		hook.appendEntry(entry)
-		return true
+func (hook *asyncTelemetryHook) waitForEventAndReady() bool {
+	ticker := time.NewTicker(time.Second)
+
+	for !hook.ready || len(hook.pending) == 0 {
```
tsachiherman: this would cause a data race.
The ready and pending should be gated behind the hook lock.

logging/telemetryhook.go
```diff
@@ -73,22 +80,31 @@ func createAsyncHook(wrappedHook logrus.Hook, channelDepth uint, maxQueueDepth i
...
 
-func (hook *asyncTelemetryHook) waitForEvent() bool {
-	select {
-	case <-hook.quit:
-		return false
-	case entry := <-hook.entries:
-		hook.appendEntry(entry)
-		return true
+func (hook *asyncTelemetryHook) waitForEventAndReady() bool {
+	ticker := time.NewTicker(time.Second)
```
tsachiherman: ticket leak : you need to defer ticker.Stop() so that you won't leak a go routine. ( which the ticker is using internally )

logging/telemetryhook.go
```diff
@@ -73,22 +80,31 @@ func createAsyncHook(wrappedHook logrus.Hook, channelDepth uint, maxQueueDepth i
...
+
+	for !hook.ready || len(hook.pending) == 0 {
+		select {
+		case <-hook.quit:
+			return false
+		case entry := <-hook.entries:
+			hook.appendEntry(entry)
+		case <-ticker.C:
```
tsachiherman: that's a really poor way to test that the initialization is complete. You'd better use a closed channel for that.

cmd/algod/main.go
winder: I had it implemented this way originally, but Brian preferred keeping it out of the networking code. In either implementation `algoh` doesn't have a `meshThread()` to leverage so it will have to do something like this.

cc: @algobolson

tsachiherman: So, if you go on that path, than create a trivial service that you'll be able to shutdown. you might want to create a single service for both this as well as for the heartbeat ticker.

winder: I tried to put this function in one of our utils/tools packages, but there it was pulling in just the right dependencies to create import cycles. Maybe a new package, I'll take another stab at it and find a better spot for a service.

winder:
@tsachiherman I think I addressed all of the feedback, added the following:
- signal to async telemetry hook event loop on telemetry uri change instead of ticker.
- service with a shutdown signal to manage the telemetry update loop, used by algod/algoh.
- shutdown signal added to the heartbeat event loop

logging/telemetryhook.go
```diff
@@ -165,6 +169,10 @@ func createElasticHook(cfg TelemetryConfig) (hook logrus.Hook, err error) {
 	}
 	hostName := cfg.getHostName()
 	hook, err = elogrus.NewElasticHook(client, hostName, cfg.MinLogLevel, cfg.ChainID)
+
+	if err == nil && hook == nil {
```
tsachiherman: That's pretty untypical for go; usually the object and the error are mutually exclusive.

tools/network/telemetryURIUpdateService.go
```diff
@@ -0,0 +1,54 @@
...
+			case <-ticker.C:
+				updateTelemetryURI()
+			case <-abort:
+				return
+			}
+		}
+	}()
+}
+
+func lookupTelemetryURI(cfg config.Local, genesisNetwork protocol.NetworkID, log logging.Logger) string {
```
tsachiherman: note that this function does not return a URI. it returns an endpoint.

logging/telemetryhook.go
winder: I think that this doesn't actually happen and I was confused initially because nil, nil is returned when initialized with an empty telemetry string. I'll remove this and make sure that is really the case.

logging/telemetryFilteredHook.go
```diff
@@ -19,30 +19,42 @@ package logging
 import (
 	"strings"
 
+	"github.com/pkg/errors"
```
tsachiherman: would you be able to use either "errors" package or `fmt.Errorf` ?
I don't see any other reference to this package, and I'd like to avoid adding new ones.
( unless there is a good reason for that, of course )

logging/telemetryFilteredHook.go
```diff
@@ -49,6 +49,11 @@ func newTelemetryFilteredHook(cfg TelemetryConfig, hook logrus.Hook, reportLogLe
 
 // Fire is required to implement logrus hook interface
 func (hook *telemetryFilteredHook) Fire(entry *logrus.Entry) error {
+	// Just in case
+	if hook.wrappedHook == nil {
+		return errors.New("The wrapped hook has not been initialized.")
```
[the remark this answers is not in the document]
winder: fixed

logging/telemetryFilteredHook.go
winder: fixed

PR #376, opened 2019-10-08
Ledger.eval Bugfix: minimum balance error does not account for user accepted assets when reporting minimum balance

EvanJRichard:
## Summary

A user's minimum balance depends on the protocol minimum balance and the number of assets they have chosen to accept. However, currently the error returned when this check fails only reports the protocol minimum balance, it does not account for the number of accepted assets.

This PR proposes to change the returned error to reflect the account's effective minimum balance.

Resolves https://github.com/algorand/go-algorand-sdk/issues/59 

## Test Plan

Automated Travis testing to detect regressions.


algobolson:
`go-algorand/data/pools/transactionPool_test.go` has a test `TestSenderGoesBelowMinBalance` and I _think_ the error message from ledger/eval.go will get passed back up to this point and you can check the error message text there for having the proper new form

PR #377, opened 2019-10-08
Add semantic version output flag to updater

winder:
## Summary

`updater ver check -c stable` outputs a machine readable number that can conveniently be used in scripts, like `4294967323`. It would be nice to have a human readable form as well, so this adds a `--semantic` flag for printing the semantic version. 

## Test Plan

Added some unit tests and ran the command.

```bash
$ updater ver check -c stable
Checking for files matching: 'channel/stable/node_stable_linux-amd64' in bucket algorand-releases
4294967323
$ updater ver check -c stable -s
Checking for files matching: 'channel/stable/node_stable_linux-amd64' in bucket algorand-releases
1.0.27
```

### DIFF
--- a/util/s3/s3Helper.go
+++ b/util/s3/s3Helper.go
@@ -323,15 +323,16 @@ func GetVersionFromName(name string) (version uint64, err error) {
 	return
 }
 
+// GetVersionPartsFromVersion converts the merged version number back into parts.
 func GetVersionPartsFromVersion(version uint64) (major uint64, minor uint64, patch uint64, err error) {
 	val := version
 
 	if val < 1<<32 {
-		err = errors.New("Versions below 1.0.0 not supported.")
+		err = errors.New("versions below 1.0.0 not supported")
 		return
 	}
 
-	modVal := uint64(1<<16)
+	modVal := uint64(1 << 16)
 
 	patch = val % modVal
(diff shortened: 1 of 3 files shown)


util/s3/s3Helper.go
```diff
@@ -322,3 +322,22 @@ func GetVersionFromName(name string) (version uint64, err error) {
...
+	val := version
+
+	if val < 1<<32 {
+		err = errors.New("versions below 1.0.0 not supported")
+		return
+	}
+
+	modVal := uint64(1 << 16)
+
+	patch = val % modVal
```
tsachiherman: nit: division (or modulo ) is a dangerous operation. it could lead to theoretical division by zero.
if you were to change it to 
```patch = val & 65536```
you'd eliminate that one.

winder: Good idea, but wouldn't it need to be `patch = val & 65535`? Or `val & (modVal - 1)`

And do you prefer using the actual constant or the shifted bit?

tsachiherman: yes. I like @algobolson proposal to use 0xffff.
If you want to merge it as is ( since it's already approved ), you can go ahead and do that.
If you'd like to use bit masks, I would be happy to approve it again.

PR #378, opened 2019-10-08
Add lease bindings to goal.

derbear:
This allows goal to send transactions with a given lease.

### DIFF
--- a/cmd/dispenser/server.go
+++ b/cmd/dispenser/server.go
@@ -189,7 +189,7 @@ func dispense(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	tx, err := c.SendPaymentFromUnencryptedWallet(cfg.Source, target, uint64(cfg.Fee), uint64(cfg.Amount), nil)
+	tx, err := c.SendPaymentFromUnencryptedWallet(cfg.Source, target, uint64(cfg.Fee), uint64(cfg.Amount), [32]byte{}, nil)
 	if err != nil {
 		http.Error(w, fmt.Sprintf("failed to dispense money - %v", err), http.StatusInternalServerError)
 		return
--- a/cmd/dispenser/server.go
+++ b/cmd/dispenser/server.go
@@ -189,7 +189,7 @@ func dispense(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	tx, err := c.SendPaymentFromUnencryptedWallet(cfg.Source, target, uint64(cfg.Fee), uint64(cfg.Amount), [32]byte{}, nil)
+	tx, err := c.SendPaymentFromUnencryptedWallet(cfg.Source, target, uint64(cfg.Fee), uint64(cfg.Amount), nil)
 	if err != nil {
 		http.Error(w, fmt.Sprintf("failed to dispense money - %v", err), http.StatusInternalServerError)
 		return
--- a/cmd/goal/clerk.go
+++ b/cmd/goal/clerk.go
@@ -45,6 +45,7 @@ var (
 	rejectsFilename string
 	noteBase64      string
 	noteText        string
+	lease           string
 	sign            bool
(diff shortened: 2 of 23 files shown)


cmd/dispenser/server.go
```diff
@@ -189,7 +189,7 @@ func dispense(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	tx, err := c.SendPaymentFromUnencryptedWallet(cfg.Source, target, uint64(cfg.Fee), uint64(cfg.Amount), nil)
+	tx, err := c.SendPaymentFromUnencryptedWallet(cfg.Source, target, uint64(cfg.Fee), uint64(cfg.Amount), [32]byte{}, nil)
```
EvanJRichard: do we foresee `32` ever changing? (as well as its use in error checking in `clerk.go`)
if "no" this is fine, if "yes" it's probably still fine

test/e2e-go/features/transactions/asset_test.go
```diff
@@ -539,6 +539,6 @@ func TestAssetSend(t *testing.T) {
 	_, err = helperFillSignBroadcast(client, wh, extra, tx, err)
 	a.NoError(err)
 
-	_, err = client.SendPaymentFromWallet(wh, nil, extra, "", 0, 0, nil, account0, 0, 0)
+	_, err = client.SendPaymentFromWallet(wh, nil, extra, "", 0, 0, nil, account0, [32]byte{}, 0, 0)
```
EvanJRichard: The fact that a caller that doesn't care about the `lease` has to supply `[32]byte{}` rather than `nil` seems to break existing patterns to me, but not a big deal.

cmd/goal/clerk.go
```diff
@@ -211,6 +213,17 @@ var sendCmd = &cobra.Command{
...
+		// Parse lease field
+		leaseBytesRaw, err := base64.StdEncoding.DecodeString(lease)
+		if err != nil {
+			reportErrorf(malformedLease, lease, err)
+		}
+		if len(leaseBytesRaw) != 32 {
+			reportErrorf(malformedLease, lease, fmt.Errorf("lease length %d != 32", len(leaseBytesRaw)))
+		}
+		var leaseBytes [32]byte
+		copy(leaseBytes[:], leaseBytesRaw)
```
tsachiherman: Given that the content of the `lease` variable is optional ( i.e. could be empty ), I think that you don't really want to fail here if the encoded size isn't 32 bytes long.
I think that only the expect tests overs this functionality.

test/e2e-go/features/transactions/asset_test.go
tsachiherman: I think that we should be able to pass nil into a []byte{} array. not so sure about a [32]byte{} array.

tsachiherman: (i.e. it would work for slice, but I'm not sure about an array )

cmd/dispenser/server.go
derbear: Hmm, probably in protocol upgrades only. Yeah, I'm considering having a typedef for it so that `32` can change all at once. Might go ahead and do this.

test/e2e-go/features/transactions/asset_test.go
derbear: Sounds good--are you thinking it's better to have `SendPaymentFromWallet` and `SendPaymentFromWalletWithLease`?

cmd/goal/clerk.go
derbear: Ah, you're right--the field will be empty if not set. Forgot about this--I'll go fix it.

test/e2e-go/features/transactions/asset_test.go
derbear: Just read your top-level comment--sounds good.

PR #383, opened 2019-10-14
linux/arm won't get updated via updater tool

tsachiherman:
## Summary

linux/arm won't bet updated via updater tool.
The wrong package was downloaded ( arm64 instead of arm ), which prevents us from running the deployed node ( obviously ).

## Test Plan

Deploy the fix. Iteratively attempt to update.

### DIFF
--- a/util/s3/s3Helper.go
+++ b/util/s3/s3Helper.go
@@ -253,7 +253,7 @@ func (helper *Helper) UploadFiles(subFolder string, files []string) error {
 func (helper *Helper) GetPackageVersion(channel string, pkg string, specificVersion uint64) (maxVersion uint64, maxVersionName string, err error) {
 	osName := runtime.GOOS
 	arch := runtime.GOARCH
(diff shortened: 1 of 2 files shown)


util/s3/s3Helper.go
```diff
@@ -253,7 +253,7 @@ func (helper *Helper) UploadFiles(subFolder string, files []string) error {
 func (helper *Helper) GetPackageVersion(channel string, pkg string, specificVersion uint64) (maxVersion uint64, maxVersionName string, err error) {
 	osName := runtime.GOOS
 	arch := runtime.GOARCH
-	prefix := fmt.Sprintf("%s_%s_%s-%s", pkg, channel, osName, arch)
+	prefix := fmt.Sprintf("%s_%s_%s-%s_", pkg, channel, osName, arch)
```
tsachiherman: This would make sure that no arm64 packages are being picked on a arm system.

PR #386, opened 2019-10-16
Update upgrade path

tsachiherman:
## Summary

Update the upgrade path to allow rewards, named assets, multiple transactions and teal.

## Details

Production code is currently running version 17.
BetaNet, had version 18 enabled with the new features for testing purposes.

Now, that we have a final spec, we can merge the two onto v19, which has both accurate spec and matching feature set.


tsachiherman:
We can wait with reviving devnet for now. I was hoping to have the master with all the changes that will go into it asap. But you're right, we should wait until we have the actual string. In the meanwhile, we'll use the "future" protocol.
( I'll be holding this PR open. Once we have the string I will update the PR and ask for your approval again )

PR #387, opened 2019-10-16
Reencode transactions in txhandler

justicz:
To avoid partitioning the network between nodes that have upgraded their software (but not necessarily their protocol version) and those that have not, it is important that only canonical transactions are rebroadcast.

### DIFF
--- a/data/txHandler.go
+++ b/data/txHandler.go
@@ -114,7 +114,7 @@ func (handler *TxHandler) Stop() {
 
 func reencode(stxns []transactions.SignedTxn) []byte {
 	var result [][]byte
-	for stxn := range stxns {
+	for _, stxn := range stxns {
 		result = append(result, protocol.Encode(stxn))
 	}
 	return bytes.Join(result, nil)
--- a/data/txHandler.go
+++ b/data/txHandler.go
@@ -111,15 +111,6 @@ func (handler *TxHandler) Stop() {
 	handler.backlogWg.Wait()
 }
 
-func reencode(raw []byte) ([]byte, error) {
-	var tmpTxnGroup []transactions.SignedTxn
-	err := protocol.Decode(raw, &tmpTxnGroup)
-	if err != nil {
-		return nil, err
-	}
-	return protocol.Encode(tmpTxnGroup), nil
-}
-
 // backlogWorker is the worker go routine that process the incoming messages from the postVerificationQueue and backlogQueue channels
 // and dispatches them further.
 func (handler *TxHandler) backlogWorker() {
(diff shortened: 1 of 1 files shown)


data/txHandler.go
```diff
@@ -137,7 +146,14 @@ func (handler *TxHandler) backlogWorker() {
 				logging.Base().Debugf("could not remember tx: %v", err)
 				continue
 			}
-			handler.net.Relay(handler.ctx, protocol.TxnTag, wi.rawmsg.Data, false, wi.rawmsg.Sender)
+
+			reencoded, err := reencode(wi.rawmsg.Data)
```
algoradam: Is this the same as `protocol.Encode(verifiedTxGroup)`? If so could we could do that and avoid decoding twice?

justicz: it should be, yes! fixed, thanks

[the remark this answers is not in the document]
justicz: done, thanks

data/txHandler.go
```diff
@@ -111,6 +112,14 @@ func (handler *TxHandler) Stop() {
 	handler.backlogWg.Wait()
 }
 
+func reencode(stxns []transactions.SignedTxn) []byte {
+	var result [][]byte
+	for stxn := range stxns {
```
algoradam: I think `for stxn := range` needs to be be `for _, stxn := range` or it'll just give you the index

justicz: oops, embarrassing! thanks adam

tsachiherman:
I don't like this approach since it looks like we're trying to put a bandage over something that shouldn't require it to start it.
Prior to having the group txn we never had this problem. The issue started when we created the group txn which had non-backward compatible transport representation ( i.e. old implementation would *accept* the first txn ). Instead of doing this, don't we want old implementations to *ignore* this txn ?

pzbitskiy:
what is the point of reencoding? in **processIncomingTxn** we convert `rawmsg.Data` to `unverifiedTxGroup`, and it also ensures that `rawmsg.Data` make a group. My understanding is `rawmsg.Data` matches to `reencode(unverifiedTxGroup)`.

justicz:
Unfortunately, we cannot rely on `rawmsg.Data` corresponding exactly to `reencode(unverifiedTxGroup)`, because we are willing to decode non-canonical encodings of transactions. Some non-canonical encodings will be rejected by old nodes but not new ones, potentially leading to a network partition since nodes disconnect from peers who send them invalid transactions.

derbear:
I wrote a summary in the squashed commit message. Reproducing here:

> A transaction with a field set to a zero value explicitly in its msgpack encoding is non-canonically encoded. If this field is present in a new version of the software but not the old version, the transaction will be accepted by an upgraded node but will seem corrupt to an un-upgraded node, causing it to disconnect from nodes which have upgraded. This behavior will cause un-upgraded nodes to partition themselves from upgraded nodes.
> 
> This patch fixes this problem by causing all (well-behaving) nodes to relay only canonical encodings of transactions to other nodes by explicitly re-encoding them.

tsachiherman:
I don't think that this fix suffice, Instead, we should do something like
this :
https://github.com/algorand/go-algorand/pull/393

justicz:
I think that this should be sufficient when combined with #390, which rejects all transactions with fields that are too new, right?

tsachiherman:
It's not exactly the same. A new node would add such a message to it's
transaction pool while an older node would reject it.
We want both to reject the transaction.

justicz:
With just this change, you could individually target a node with the zero-encoded-transaction-field attack, but it shouldn't propagate the transaction to other nodes and cause a partition. And I don't think you can cause a partition just by sending such a transaction to all nodes individually -- the reason the transaction becomes a problem is when peers disconnect from each other when they see a "malformed" relayed message.

tsachiherman:
But that's exactly what we want. We want *everyone* to behave the same way
before the protocol upgrade.
If an attacker is sending malicious transaction before the protocol upgrade
takes place, we want it to be disconnected right away.

PR #390, opened 2019-10-17
Reject transactions with not-yet-protocol-enabled fields more quickly

justicz:
Without this I believe there are similar partition issues to #387 if a user e.g. sends a transaction with an `LSig` prior to the protocol upgrade

### DIFF
--- a/data/transactions/verify/txn.go
+++ b/data/transactions/verify/txn.go
@@ -46,10 +46,6 @@ func TxnPool(s *transactions.SignedTxn, spec transactions.SpecialAddresses, prot
 		return errors.New("empty address")
 	}
 
-	if s.Sig != (crypto.Signature{}) && !s.Msig.Blank() {
-		return errors.New("signedtxn should only have one of Sig or Msig")
-	}
-
 	outCh := make(chan error, 1)
 	cx := asyncVerifyContext{s, outCh, &proto}
 	verificationPool.EnqueueBacklog(context.Background(), stxnAsyncVerify, &cx, nil)
--- a/data/transactions/verify/txn.go
+++ b/data/transactions/verify/txn.go
@@ -136,6 +136,9 @@ func stxnVerifyCore(s *transactions.SignedTxn, proto *config.ConsensusParams) er
 // LogicSig checks that the signature is valid and that the program is basically well formed.
 // It does not evaluate the logic.
 func LogicSig(lsig *transactions.LogicSig, proto *config.ConsensusParams, stxn *transactions.SignedTxn) error {
+	if proto.LogicSigVersion == 0 {
+		return errors.New("LogicSig not enabled")
+	}
 	if len(lsig.Logic) == 0 {
 		return errors.New("LogicSig.Logic empty")
 	}
--- a/data/transactions/verify/txn.go
+++ b/data/transactions/verify/txn.go
@@ -88,20 +88,6 @@ func stxnAsyncVerify(arg interface{}) interface{} {
 }
 
 func stxnVerifyCore(s *transactions.SignedTxn, proto *config.ConsensusParams) error {
-	// Check for non-empty fields that are not yet enabled
-	if !proto.Asset &&
-		((s.Txn.AssetConfigTxnFields != transactions.AssetConfigTxnFields{}) ||
-		(s.Txn.AssetTransferTxnFields != transactions.AssetTransferTxnFields{}) ||
-		(s.Txn.AssetFreezeTxnFields != transactions.AssetFreezeTxnFields{})) {
-		return errors.New("assets not yet enabled")
-	}
-	if !proto.SupportTxGroups && (s.Txn.Group != crypto.Digest{}) {
-		return errors.New("txn groups not yet enabled")
-	}
-	if !proto.SupportTransactionLeases && (s.Txn.Lease != [32]byte{}) {
-		return errors.New("leases not yet enabled")
-	}
-
 	numSigs := 0
 	hasSig := false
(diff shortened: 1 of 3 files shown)


justicz:
Ugh, actually, I think `stxnVerifyCore` may need to reject all transaction fields that have been set if those fields are invalid for the current `proto`. Otherwise old nodes will reject the transaction in `txHandler` and disconnect (due to not knowing about the new fields), and new nodes will accept (since `stxnVerifyCore` only checks high level well-formedness).

data/transactions/verify/txn.go
```diff
@@ -46,10 +46,6 @@ func TxnPool(s *transactions.SignedTxn, spec transactions.SpecialAddresses, prot
 		return errors.New("empty address")
 	}
 
-	if s.Sig != (crypto.Signature{}) && !s.Msig.Blank() {
-		return errors.New("signedtxn should only have one of Sig or Msig")
-	}
-
```
tsachiherman: Are these conditions covered elsewhere ?

data/transactions/verify/txn.go
```diff
@@ -92,6 +88,33 @@ func stxnAsyncVerify(arg interface{}) interface{} {
 }
 
 func stxnVerifyCore(s *transactions.SignedTxn, proto *config.ConsensusParams) error {
+	// Check for non-empty fields that are not yet enabled
+	if !proto.Asset {
```
tsachiherman: Looks good. Implementation is a bit verbosed, but the logic seems solid.
i.e.
```
if !proto.Asset && (
   (s.Txn.AssetConfigTxnFields != transactions.AssetConfigTxnFields{}) ||
   (s.Txn.AssetTransferTxnFields != transactions.AssetTransferTxnFields{}) ||
   (s.Txn.AssetFreezeTxnFields != transactions.AssetFreezeTxnFields{})) {
   return errors.New("assets not yet enabled")
}
```

data/transactions/verify/txn.go
algobolson: yes, this is covered inside stxnVerifyCore

algobolson:
_while we're at it_ (sorry) these lines from ledger/eval.go probably make more sense here
```
		// Verify that groups are supported.
		if !txn.Txn.Group.IsZero() && !eval.proto.SupportTxGroups {
			return fmt.Errorf("transaction groups not supported")
		}
```

justicz:
Oh, actually it looks like I should just call `stxn.Txn.WellFormed`. It does these checks

data/transactions/verify/txn.go
justicz: changed as suggested, thanks

justicz:
Interesting -- in addition to that check in eval.go, `txn.WellFormed` is checked, followed by `verify.TxnPool`, which *also* calls `txn.WellFormed`. So there's more to delete there too

ledger/eval.go
```diff
@@ -467,25 +467,14 @@ func (eval *BlockEvaluator) transaction(txn transactions.SignedTxn, ad transacti
 			return TransactionInLedgerError{txn.ID()}
 		}
 
-		// Well-formed on its own?
-		err = txn.Txn.WellFormed(spec, eval.proto)
```
justicz: I deleted this because `WellFormed` is immediately called by `verify.TxnPool` below. Can someone please double check this for me since it's... pretty important

ledger/eval.go
```diff
@@ -467,25 +467,14 @@ func (eval *BlockEvaluator) transaction(txn transactions.SignedTxn, ad transacti
...
+		// Well-formed and signed properly?
 		if eval.txcache == nil || !eval.txcache.Verified(txn) {
 			err = verify.TxnPool(&txn, spec, eval.proto, eval.verificationPool)
 			if err != nil {
 				return fmt.Errorf("transaction %v: failed to verify: %v", txn.ID(), err)
 			}
 		}
 
-		// Verify that groups are supported.
-		if !txn.Txn.Group.IsZero() && !eval.proto.SupportTxGroups {
```
justicz: Removing this check since it should be checked in `TxnPool` as well now.

ledger/eval.go
justicz: can't remove because the transaction cache might become invalid over a protocol upgrade

ledger/eval.go
justicz: can't remove because the transaction cache might become invalid over a protocol upgrade, and group is currently not checked in wellformed. Something to fix another time.

data/transactions/verify/txn.go
```diff
@@ -92,6 +88,20 @@ func stxnAsyncVerify(arg interface{}) interface{} {
 }
 
 func stxnVerifyCore(s *transactions.SignedTxn, proto *config.ConsensusParams) error {
+	// Check for non-empty fields that are not yet enabled
```
derbear: This is a nit, but shouldn't these checks go into `Transaction.WellFormed`?

ledger/eval.go
algobolson: Yup. Txn and TxnPool both call WellFormed immediately before inline or pool based execution of stxnVerifyCore. I'm guessing that's out of some idea to catch simple things before going through the overhead of pool enqueue/dequeue.

data/transactions/verify/txn.go
justicz: Actually, it turns out a lot of these checks were done in `WellFormed` (I missed that `WellFormed` was getting called in `Txn`). The only missing check there was the `Group` check. So I added that and removed the extra ones I added.

PR #394, opened 2019-10-18
Implement domain separation for LogicSig ed25519verify.

derbear:
Arguments passed to ed25519verify are now domain-separated with the
constant string "ProgData" and the hash of the running program.

### DIFF
--- a/data/transactions/logic/eval.go
+++ b/data/transactions/logic/eval.go
@@ -1154,16 +1154,16 @@ func opGlobal(cx *evalContext) {
 	cx.nextpc = cx.pc + 2
 }
 
-// LogicMsg is data meant to be signed and then verified
-// with the ed25519verify opcode.
-type LogicMsg struct {
+// Msg is data meant to be signed and then verified with the
+// ed25519verify opcode.
+type Msg struct {
 	_struct     struct{}      `codec:",omitempty,omitemptyarray"`
 	ProgramHash crypto.Digest `codec:"p"`
 	Data        []byte        `codec:"d"`
 }
 
 // ToBeHashed implements crypto.Hashable
-func (msg LogicMsg) ToBeHashed() (protocol.HashID, []byte) {
+func (msg Msg) ToBeHashed() (protocol.HashID, []byte) {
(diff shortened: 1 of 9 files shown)


data/transactions/logic/eval.go
```diff
@@ -1150,6 +1154,19 @@ func opGlobal(cx *evalContext) {
 	cx.nextpc = cx.pc + 2
 }
 
+// Msg is data meant to be signed and then verified with the
+// ed25519verify opcode.
+type Msg struct {
+	_struct     struct{}      `codec:",omitempty,omitemptyarray"`
+	ProgramHash crypto.Digest `codec:"p"`
```
algobolson: I don't like putting the program hash in here because then a signed fact can have only one consumer. An oracle can't just say "thing is true!" it has to say "thing is true, Bob" and possibly repeat that for many recipients.

derbear: Okay, offline discussion on this is that an oracle can set up an explicit escrow account _E_ and then produce statements of the form `Sign_oracle(H(E) || stmt)`. _E_ allows group transactions in TEAL that can read off this statement. (Optionally the group constraint can require an extra fee to be paid to the oracle to read off the statement.)

derbear: (The other outcome of the discussion is we'll maybe support either byte blob concatenation or substring slicing on the blobs, but that will be after the next release.)

data/transactions/logic/eval.go
```diff
@@ -214,6 +217,7 @@ func Eval(program []byte, params EvalParams) (pass bool, err error) {
 	cx.EvalParams = params
 	cx.stack = make([]stackValue, 0, 10)
 	cx.program = program
+	cx.programHash = crypto.HashObj(Program(program))
```
algobolson: I slightly wish this was lazy since I bet we won't need it for 90% of programs, but I'm not sure we need that optimization yet.

derbear: Sounds good--maybe we can pass it in? (Since we'll have computed upstream at some point, either when decommitting to an escrow account or when verifying some delegated LogicSig.)

PR #396, opened 2019-10-18
Teal stateless

algobolson:
make TEAL stateless and cacheable
no longer has direct access to Round or TimeStamp (instead has FirstValidTime)
`rand` removed until it can be done right
add LogicSig Eval caching to TransactionPool
better docs


ledger/eval.go
```diff
@@ -479,39 +484,32 @@ func (eval *BlockEvaluator) transaction(txn transactions.SignedTxn, ad transacti
...
+			err = eval.checkLogicSig(txn, txgroup, groupIndex)
 			if err != nil {
-				return fmt.Errorf("transaction %v: rejected by logic err=%s", txn.ID(), err)
-			}
-			if !pass {
-				return fmt.Errorf("transaction %v: rejected by logic", txn.ID())
+				eval.txcache.EvalRemember(txn, err.Error())
+				return err
 			}
+			eval.txcache.EvalRemember(txn, "")
```
pzbitskiy: eval.txcache.EvalRemember(txn, nil) ?

data/transactions/logic/globalfield_string.go
```diff
@@ -0,0 +1,28 @@
...
+	// An "invalid array index" compiler error signifies that the constant values have changed.
+	// Re-run the stringer command to generate them again.
+	var x [1]struct{}
+	_ = x[MinTxnFee-0]
+	_ = x[MinBalance-1]
+	_ = x[MaxTxnLife-2]
+	_ = x[ZeroAddress-3]
+	_ = x[GroupSize-4]
+	_ = x[invalidGlobalField-5]
+}
```
pzbitskiy: what kind of magic is it?

ledger/eval.go
pzbitskiy: Nevermind,  nil can't be used as a value for a string field.

data/transactions/logic/globalfield_string.go
algobolson: x has one element. x[0] is the only valid index. if any of the constants change, this generated code will break signaling that it needs to be re-generated. (I'd normally not check in generated code, but we're already doing that elsewhere with this same `stringer` generator)

data/transactions/logic/eval.go
```diff
@@ -1021,57 +996,59 @@ func opDup(cx *evalContext) {
...
+	case GroupIndex:
 		sv.Uint = uint64(cx.GroupIndex)
-	case 21:
+	case TxID:
 		txid := txn.ID()
 		sv.Bytes = txid[:]
-	case 22:
-		sv.Uint = cx.GroupSenders[cx.GroupIndex].MicroAlgos.Raw
-	case 23:
+	/*case 22: // SenderBalance
```
justicz: are you planning on leaving these comments in?

data/pools/transactionPool.go
```diff
@@ -72,6 +76,7 @@ func MakeTransactionPool(ledger *ledger.Ledger, transactionPoolStatusSize int, l
 		ledger:          ledger,
 		statusCache:     makeStatusCache(transactionPoolStatusSize),
```
tsachiherman: The `statusCache` contains only the *failed* transaction groups. Shouldn't we rename it ?

data/pools/transactionPool.go
```diff
@@ -337,6 +342,20 @@ func (pool *TransactionPool) Verified(txn transactions.SignedTxn) bool {
...
+func (pool *TransactionPool) EvalOk(txid transactions.Txid) (tx transactions.SignedTxn, txErr string, found bool) {
+	pool.lcmu.RLock()
+	defer pool.lcmu.RUnlock()
+	return pool.lsigCache.check(txid)
+}
+
+// EvalRemember sets an error string from LogicSig Eval for some SignedTxn
+func (pool *TransactionPool) EvalRemember(tx transactions.SignedTxn, errString string) {
+	pool.lcmu.Lock()
+	defer pool.lcmu.Unlock()
```
tsachiherman: I like the way you wrapped the lock around the object.

data/pools/transactionPool.go
```diff
@@ -337,6 +342,20 @@ func (pool *TransactionPool) Verified(txn transactions.SignedTxn) bool {
 	return pendingSigTxn.Sig == txn.Sig && pendingSigTxn.Msig.Equal(txn.Msig) && pendingSigTxn.Lsig.Equal(&txn.Lsig)
 }
 
+// EvalOk for LogicSig Eval of a txn by txid, returns the SignedTxn, error string, and found.
+func (pool *TransactionPool) EvalOk(txid transactions.Txid) (tx transactions.SignedTxn, txErr string, found bool) {
```
tsachiherman: nit: The first return value is not being used, and can be omitted from the function signature; also, the second argument need to return just whether it's != "" or not. ( i.e. error'ed or not ).

data/transactions/logic/eval.go
algobolson: cleaned up

data/pools/transactionPool.go
algobolson: Yeah, it is that.
```
// failedTxnStatuses records failure messages from recently failed transactions, indexed by Txid.
// This allows algod clients to find out why the transaction they submitted failed.
failedTxnStatuses: makeStatusCache(...)
```

data/pools/transactionPool.go
algobolson: dropped the unused argument, but there's one place we return the error string if it's not ""

tsachiherman:
@algobolson - could you take another look and update the code so it would compile without any errors ? thanks!

PR #398, opened 2019-10-19
Prepare for performance test

stechu:
## Summary

This PR contains 3 changes that will be useful for performance tests.

*  make transaction pool's fee increase factor tunable (default is still 2), this helps performance tests
*  add a logic program option `--program`  to pingpong and let pingpong issue delegated logic sig transactions
*  add a fire rate control option  `--tps` to pingpong so that pingpong can issue transactions exactly at the rate we want 

 
## Test Plan

No protocol change. Tested on local deployed network.


shared/pingpong/pingpong.go
```diff
@@ -150,6 +151,14 @@ func RunPingPong(ctx context.Context, ac libgoal.Client, accounts map[string]uin
...
 				refreshTime = refreshTime.Add(cfg.RefreshTime)
 			}
+
+			localTimeDelta := time.Now().Sub(startTime)
+			currentTps := float64(totalSent) / localTimeDelta.Seconds()
+			if currentTps > float64(cfg.TxnPerSec) {
+				sleepSec := float64(totalSent)/float64(cfg.TxnPerSec) - localTimeDelta.Seconds()
+				sleepTime := time.Duration(int64(math.Round(sleepSec*1000))) * time.Millisecond
+				time.Sleep(sleepTime)
+			}
```
tsachiherman: That code would only work to slow down a single txn time to match the desired txn/sec time.
Unfortunately, when we try to push it to the end, we get to uneven txn times, where this wouldn't work so well.
I suggest that you'll try to implement it using a moving window of size cfg.TxnPerSec*2.

stechu:
Will break it into several PRs. Closing this one.

PR #400, opened 2019-10-21
Add the ability to create non-participating nodes via 'netgoal' command.

tsachiherman:
## Summary

Add the ability to create non-participating nodes via 'netgoal' command.
Adds the -H option to allow specify the number of non-participating nodes that would get created.

## Test Plan

I deployed a network with these changes, and it looks good.

### DIFF
--- a/cmd/netgoal/generate.go
+++ b/cmd/netgoal/generate.go
@@ -37,8 +37,10 @@ var templateToGenerate string
 var relaysToGenerate int
 var nodesToGenerate int
 var nodeHostsToGenerate int
+var nonPartnodesHostsToGenerate int
 var walletsToGenerate int
 var nodeTemplatePath string
+var nonParticipatingNodeTemplatePath string
 var relayTemplatePath string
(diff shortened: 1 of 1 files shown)


cmd/netgoal/generate.go
```diff
@@ -125,13 +143,13 @@ template modes for -t:`,
 				reportErrorf("must specify number of relays with -R")
 			}
 
-			err = generateNetworkTemplate(outputFilename, walletsToGenerate, relaysToGenerate, nodeHostsToGenerate, nodesToGenerate, baseNode, baseRelay)
+			err = generateNetworkTemplate(outputFilename, walletsToGenerate, relaysToGenerate, nodeHostsToGenerate, nodesToGenerate, nonPartnodesHostsToGenerate, baseNode, baseNonParticipatingNode, baseRelay)
```
algobolson: at some point this growing parameter list might look better as a parameter object. maybe next time.

PR #403, opened 2019-10-21
Switch to short asset ids

justicz:
This PR switches asset identifiers from being tuples of `(asset creator, unique asset ID)` to just `(unique asset ID)`.

It does this by adding a new table to the `accountdb` called `assetcreators` with two columns: `asset`, which is the integer primary key, and `creator`, which is a blob storing the address of the creator.

In order to minimize changes to the `ledger` API, I added a function called `getChangedAssetIndices` that takes in an `accountDelta` and finds the asset index of any created or deleted assets. So the new lookup mechanism should work transparently with `Put`. [0]

I think the trickiest change is in the `acctupdates` ledger tracker. There I added a new method, `getAssetCreatorForRound`, which accepts a round and an asset index and returns the creator. I tried to copy the existing logic surrounding scanning backwards over account deltas.

[0] Another way I could have implemented this would be to explicitly add `RegisterAsset` and `UnregisterAsset` methods to the cow layer/ledger to read/write from the new table directly. My current solution seemed cleaner to me but if it becomes inefficient we can change it


justicz:
Working on some changes in the `acctupdates` tracker, will comment again when pushed

justicz:
ok, I feel better about this now. @zeldovich @algoradam could you please take a look?

daemon/algod/api/client/restClient.go
```diff
@@ -221,14 +221,20 @@ func (client RestClient) TransactionsByAddr(addr string, first, last, max uint64
 	return
 }
 
+// AssetInformation gets the AssetInformationResponse associated with the passed asset index
+func (client RestClient) AssetInformation(index uint64) (response v1.AssetParams, err error) {
+	err = client.get(&response, fmt.Sprintf("/assets/%d", index), nil)
```
rotemh: The convention is `asset/%d` (no `s` when it's a specific one)

daemon/algod/api/client/restClient.go
```diff
@@ -221,14 +221,20 @@ func (client RestClient) TransactionsByAddr(addr string, first, last, max uint64
...
+
 // AccountInformation also gets the AccountInformationResponse associated with the passed address
 func (client RestClient) AccountInformation(address string) (response v1.Account, err error) {
 	err = client.get(&response, fmt.Sprintf("/account/%s", address), nil)
 	return
 }
 
-// AssetInformation gets the AssetInformationResponse associated with the passed asset creator and index
-func (client RestClient) AssetInformation(creator string, index uint64) (response v1.AssetParams, err error) {
+// AssetInformationWithCreator gets the AssetInformationResponse associated with the passed asset creator and index
```
zeldovich: What's the point of this API if `AssetInformation()` exists?

daemon/algod/api/server/v1/handlers/handlers.go
```diff
@@ -135,16 +135,19 @@ func keyregTxEncode(tx transactions.Transaction, ad transactions.ApplyData) v1.T
 	}
 }
 
-func assetParams(creator basics.Address, params basics.AssetParams) v1.AssetParams {
+func assetParams(creator basics.Address, aidx basics.AssetIndex, params basics.AssetParams) v1.AssetParams {
```
zeldovich: Why pass in `aidx`?

data/transactions/asset.go
```diff
@@ -72,44 +72,49 @@ type AssetFreezeTxnFields struct {
...
+	res := make(map[basics.AssetIndex]basics.AssetParams)
 	for id, val := range m {
 		res[id] = val
 	}
 	return res
 }
 
-func getParams(balances Balances, cid basics.AssetID) (basics.AssetParams, error) {
-	creator, err := balances.Get(cid.Creator, false)
+func getParams(balances Balances, aidx basics.AssetIndex) (basics.AssetParams, basics.Address, error) {
```
zeldovich: Can you name the returned values here?  `AssetParams` was clear enough without a name, but figuring out this `Address` requires looking at the code of this function.

data/transactions/asset.go
```diff
@@ -321,8 +321,13 @@ func (ct AssetTransferTxnFields) apply(header Header, balances Balances, spec Sp
...
 
 	if ct.AssetCloseTo != (basics.Address{}) {
+		creatorAddr, err := balances.GetAssetCreator(ct.XferAsset)
+		if err != nil {
+			return fmt.Errorf("failed to find asset creator: %v", creatorAddr)
+		}
+
 		// Cannot close asset ID allocated by this account; must use destroy.
-		if ct.XferAsset.Creator == source {
+		if creatorAddr == source {
```
zeldovich: Perhaps a more direct check is looking for `AssetParams[ct.XferAsset]` in the source account.  I don't feel too strongly either way though.

ledger/accountdb.go
```diff
@@ -214,37 +231,80 @@ func accountsPutTotals(tx *sql.Tx, totals AccountTotals) error {
...
 	}
 	defer replaceStmt.Close()
 
+	insertAssetIdxStmt, err := tx.Prepare("INSERT INTO assetcreators (asset, creator) VALUES (?, ?)")
+	if err != nil {
+		return
+	}
+	defer insertAssetIdxStmt.Close()
+
+	deleteAssetIdxStmt, err := tx.Prepare("DELETE FROM assetcreators WHERE asset=?")
```
zeldovich: Did you intend to call these?

ledger/acctupdates.go
```diff
@@ -291,6 +347,10 @@ func (au *accountUpdates) committedUpTo(rnd basics.Round) basics.Round {
 				return err
 			}
 
+			for aidx := range au.assetDeltas[i] {
```
zeldovich: Reset `assetFlushcount` at the top of this `Atomic()` like `flushcount`.

ledger/cow.go
```diff
@@ -153,6 +178,9 @@ func (cb *roundCowState) commitToParent() {
 	for txl, expires := range cb.mods.txleases {
 		cb.commitParent.mods.txleases[txl] = expires
 	}
+	for aidx, delta := range cb.mods.assets {
+		cb.commitParent.mods.assets[aidx] = delta
```
zeldovich: If an asset is created and deleted within the same block, you might end up with a "deleted" delta for an asset that never had a "created" delta.  This might confuse accountsdb calling `deleteAssetIdxStmt`, modulo that call being altogether missing.

justicz: I did think about this case, but it didn't seem like it would cause an issue? Regardless, I will eliminate it from the delta entirely so we don't have to worry about it.

zeldovich: I was thinking that `deleteAssetIdxStmt` should never fail.

ledger/accountdb.go
justicz: Yes :man_facepalming:

ledger/cow.go
justicz: Right, `DELETE`s that don't match any rows don't fail.

zeldovich: Ah, indeed; that's a bit unfortunate (would have been a nice sanity check), but probably fine as-is.  Doesn't seem like we can screw up too badly since asset IDs are unique over time.

ledger/acctupdates.go
justicz: added, oops! since this would have panicked and was not caught by tests, we should definitely add a test that actually sends txn and waits more than lookback rounds...

daemon/algod/api/client/restClient.go
justicz: fixed, and removed asset by creator as discussed

daemon/algod/api/client/restClient.go
justicz: I didn't want it either, but was asked to leave it -- I just asked rotem again though and now we can remove it :upside_down_face:

data/transactions/asset.go
justicz: done

justicz:
Yes, we really need a test that actually flushes assets to disk. I am working on this (based on archival tests) and will make a PR

justicz:
added a test that creates assets, ensures they are flushed to disk, deletes an old asset and a new asset, and checks that deletions were flushed to disk

justicz:
It looks like the algod api had a race where it grabbed a balance record from the ledger, and then iterated over `record.AssetParams`, calling `GetAssetCreator` for each entry. But this would cause a failure if the asset was deleted in before/during this iteration, and a test was triggering this race. But the creator lookups were unnecessary to begin with since anything in `AssetParams` was created by the account we were looking up. So I removed the call

PR #404, opened 2019-10-21
Customizable fee increase factor

stechu:
## Summary

Make fee increase factor customizable. This will help us in performance tests.

## Test Plan

No protocol and changes on the default behavior.  Tested locally.


stechu:
> Since you're re-purposing `cfg.TxPoolExponentialIncreaseFactor` for this new purpose, can you update the comment in `config/config.go` describing this field?
> 
> Also, I don't really see why a user might want to set this value to anything other than 2 in production. It's not at the level of a constant that we need to agree on through consensus upgrades, but it seems like this should always be 2, except that we want to disable fee growth altogether for testing.

Yes, this is only for test. Updated the comment.

justicz:
LGTM

PR #407, opened 2019-10-21
Stricter logic sig check

stechu:
## Summary

This PR contains two security fixes (Thanks @justicz!) .

* Although the size of logic sig is checked, but the way we calculate the size didn't capture the case that a malicious user putting a large number of args of empty byte array (In such case, each arg will be counted 0, but there are still bytes needed to encode them on the wire). This could cause DoS attacks. This PR implements a check that limits the number of args to logic sig to be at most 255.

* Add a check that rejects logicSig eval if the protocol version is earlier.

## Test Plan

Added unit tests in `eval_test.go` for these two cases.

### DIFF
--- a/data/transactions/logic/eval.go
+++ b/data/transactions/logic/eval.go
@@ -198,7 +198,7 @@ func Eval(program []byte, params EvalParams) (pass bool, err error) {
 		err = errLogicSignNotSupported
 		return
 	}
-	if len(params.Txn.Lsig.Args) > EvalMaxArgs {
+	if params.Txn.Lsig.Args != nil && len(params.Txn.Lsig.Args) > EvalMaxArgs {
 		err = errTooManyArgs
 		return
 	}
@@ -232,7 +232,7 @@ func Eval(program []byte, params EvalParams) (pass bool, err error) {
 		if cx.stepCount > len(cx.program) {
 			return false, errLoopDetected
 		}
-		if params.Proto == nil || uint64(cx.cost) > params.Proto.LogicSigMaxCost {
+		if uint64(cx.cost) > params.Proto.LogicSigMaxCost {
 			return false, errCostTooHigh
 		}
 	}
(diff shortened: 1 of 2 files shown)


data/transactions/logic/eval.go
```diff
@@ -224,7 +232,7 @@ func Eval(program []byte, params EvalParams) (pass bool, err error) {
 		if cx.stepCount > len(cx.program) {
 			return false, errLoopDetected
 		}
-		if params.Proto != nil && uint64(cx.cost) > params.Proto.LogicSigMaxCost {
+		if params.Proto == nil || uint64(cx.cost) > params.Proto.LogicSigMaxCost {
```
algobolson: maybe make one `if params.Proto == nil { return errProgrammingError }` check at the top and then the rest of the function can assume that params.Proto is not nil and not check it each time

data/transactions/logic/eval_test.go
```diff
@@ -35,6 +35,44 @@ import (
 	"github.com/algorand/go-algorand/protocol"
 )
 
+func defaultEvalProto() config.ConsensusParams {
+	return config.ConsensusParams{LogicSigVersion: 1, LogicSigMaxCost: 20000}
+}
+
+func defaultEvalParams() EvalParams {
+	proto := defaultEvalProto()
+	return EvalParams{Proto: &proto, Trace: &strings.Builder{}, Txn: &transactions.SignedTxn{}}
```
algobolson: Trace should default to nil; OR defaultEvalParams could take a `sb *strings.Builder` passed in. But if that's not done, then everything that now calls `defaultEvalParams()` should drop its `sb strings.Builder{}` because that won't point to anything useful, but was used for error reporting, and we shouldn't confuse ourselves with error reporting that doesn't work.

data/transactions/logic/eval.go
stechu: addressed.

data/transactions/logic/eval_test.go
stechu: addressed.

PR #408, opened 2019-10-22
Allow closing frozen assets to the creator

justicz:
Even when an asset is frozen, we want to be able to close it out to the asset creator. This allows  the holder to lower their min balance if they accidentally end up with a frozen asset they don't want.

### DIFF
--- a/data/transactions/asset.go
+++ b/data/transactions/asset.go
@@ -305,26 +305,17 @@ func (ct AssetTransferTxnFields) apply(header Header, balances Balances, spec Sp
 		}
 	}
 
-	// Are we spending our whole balance to the asset creator?
-	var closeToCreator bool
-	if (ct.AssetReceiver == ct.XferAsset.Creator) && (ct.AssetCloseTo == ct.XferAsset.Creator) {
-		closeToCreator = true
-	}
-
-	// Freezes are bypassed during clawbacks and when we close out our whole
-	// asset holding to the creator
-	bypassFreeze := clawback || closeToCreator
-
 	// Actually move the asset.  Zero transfers return right away
 	// without looking up accounts, so it's fine to have a zero transfer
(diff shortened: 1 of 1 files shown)


data/transactions/asset.go
```diff
@@ -305,16 +305,26 @@ func (ct AssetTransferTxnFields) apply(header Header, balances Balances, spec Sp
 		}
 	}
 
+	// Are we spending our whole balance to the asset creator?
+	var closeToCreator bool
+	if (ct.AssetReceiver == ct.XferAsset.Creator) && (ct.AssetCloseTo == ct.XferAsset.Creator) {
```
zeldovich: If you put this inside the `ct.AssetCloseTo != (basics.Address{}) {` case below, you won't need to worry about `AssetReceiver == Creator`, which seems simpler and closer to what I think we'd want to enforce.  Why allow a separate `AssetReceiver` transfer along with the close?

justicz: Oh I see, and `takeOut` should still succeed even though we're frozen because `ct.AssetAmount` better be zero if we're frozen. Cool

justicz: Changed, also added a comment mentioning that zero transfers succeed on frozen accounts.

data/transactions/asset.go
```diff
@@ -342,12 +349,12 @@ func (ct AssetTransferTxnFields) apply(header Header, balances Balances, spec Sp
 		}
 
 		// Move the balance out.
-		err = takeOut(balances, source, ct.XferAsset, sndHolding.Amount, clawback)
+		err = takeOut(balances, source, ct.XferAsset, sndHolding.Amount, bypassFreeze)
```
justicz: (clawback was always false here because of above check)

PR #409, opened 2019-10-22
Bump minor version.

winder:


winder:
The patch version is generated by [`buildnumber.dat`](https://github.com/algorand/go-algorand/blob/rel/beta/buildnumber.dat) file in each of the release branches. They will need to be reset separately.

I believe that this is the only place we need to change it, I ran to build locally and it seems to work.

PR #410, opened 2019-10-22
Add extra asset description fields

justicz:
This PR adds two optional fields to an asset's parameters:

`URL` points to somewhere with more information about the asset.
`MetadataHash` is a commitment to some metadata about the asset, in some unspecified format.

Each of these fields is stored internally in `AssetParams` as a `[32]byte`.

### DIFF
--- a/test/e2e-go/features/transactions/asset_test.go
+++ b/test/e2e-go/features/transactions/asset_test.go
@@ -77,6 +77,9 @@ func TestAssetConfig(t *testing.T) {
 	clawback, err := client.GenerateAddress(wh)
 	a.NoError(err)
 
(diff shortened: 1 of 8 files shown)


test/e2e-go/features/transactions/asset_test.go
```diff
@@ -77,6 +77,9 @@ func TestAssetConfig(t *testing.T) {
 	clawback, err := client.GenerateAddress(wh)
 	a.NoError(err)
 
+	assetURL := "foo://bar"
+	assetMetadataHash := []byte("ISTHISTHEREALLIFEISTHISJUSTFANTA")
```
EvanJRichard: :grin:

PR #411, opened 2019-10-22
Pingpong logic sig

stechu:
## Summary

Allow pingpong to send logic sig transactions (including cherry-picked commits from @justicz). 

* add `--program` option to pingpong, which specifies the path to the compiled logic sig program.
* add `Program` and `LogicArgs` to pingpong config
* add `--teal` option to pingpong, which specifies test scenario of logic sig txns

## Test Plan
Changes are only on pingpong client. Tested locally.


stechu:
@tsachiherman  The are a few sample program here. https://github.com/algorand/algoperf/tree/master/benchmarks

PR #418, opened 2019-10-24
Improve asset archival test

justicz:
This PR adds a few more checks to the archival asset test. We now check for creating/deleting an asset in the same block.


justicz:
Yup, all `gofmt`ed

PR #420, opened 2019-10-24
Explicitly disallow txns from zero address

algoradam:
Also add a test to confirm we don't accept a zero signature from a zero public key.

Max noticed that some ed25519 implementations (e.g., the reference Python implementation) accept zero signatures from zero public keys on 1 in 4 messages. Add a test to confirm this isn't true for our implementation. Also explicitly reject transactions with the zero address as the sender for defense-in-depth.

### DIFF
--- a/crypto/curve25519_test.go
+++ b/crypto/curve25519_test.go
@@ -35,6 +35,16 @@ func TestSignVerifyEmptyMessage(t *testing.T) {
 	}
 }
 
+func TestVerifyZeros(t *testing.T) {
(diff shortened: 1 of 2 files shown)


justicz:
They're kind of related -- the `transactions.go` change checks that we never have a zero sender, even if we have a valid signature for a zero sender, and the `curve25519_test.go` change tests that we never have a regression in the crypto library where we start seeing signatures for the zero public key as valid. Together they ensure there are two ways we never screw up and accept a tx signed by an empty key.

PR #421, opened 2019-10-24
TEAL add `global LogicSigVersion`

algobolson:
add to TEAL:
`global LogicSigVersion`
Primary use case being allowing escrow accounts to have a close-out-to-originator condition based on LogicSig/TEAL changes (you might want to re-write the contract with new features).


algobolson:
@zeldovich  that is a good point and an unresolved problem in the current TransactionPool lsigCache that needs to be fixed: it needs to be invalidated when there's a protocol upgrade. I'll make a separate patch for that.

zeldovich:
I thought that (fixing txpool signature cache validity) was addressed by stateless TEAL (PR #396), which merged, but perhaps I've lost track of what's going on there..

derbear:
"Stateless" TEAL is not truly stateless in that it still allows you to refer to protocol constants such as `MinTxnFee`. In principle this is okay if we invalidate the transaction pool's validated signature cache on protocol upgrade, but I think we haven't gotten around to it yet.

algobolson:
The logic sig cache fix is in #428

derbear:
Should you regenerate docs first?

algobolson:
see #637

PR #422, opened 2019-10-24
Ensure asset transaction correctness in case of zero sender

justicz:
Until now, it hasn't really mattered for security if a transaction could be sent from a zero public key. It seems like our crypto library does check for low order public keys, but because asset creators always intend for e.g. `Clawback == basics.Address{}` to signify that there is no way to clawback funds, we should explicitly check for that case.

@algoradam is adding a check for zero public key sender in `WellFormed` in a separate PR.

### DIFF
--- a/data/transactions/asset.go
+++ b/data/transactions/asset.go
@@ -151,8 +151,8 @@ func (cc AssetConfigTxnFields) apply(header Header, balances Balances, spec Spec
 		return err
 	}
 
-	if header.Sender != params.Manager {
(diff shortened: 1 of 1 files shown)


justicz:
@tsachiherman or @winder could this PR and #420 please make it into the code freeze branch once they're merged?

tsachiherman:
> @tsachiherman or @winder could this PR and #420 please make it into the code freeze branch once they're merged?

I can't but once it's merged in, ping @will with the commit id.
It would also be helpful if you create a unit test for this change.

will:
No please do not ping me with any blockchain garbage I think it’s all dumb please don’t ping me

PR #425, opened 2019-10-25
Skip failing Arm64 and Darwin tests until fixed

algonautshant:
## Summary
Skipping tests which are failing sporadically on arm64 and mac.
Adding a utility method to skip a tests on selected platform(s).

The utility will be disabled/removed once the test failures are resolved.


algonautshant:
> Could you provide some information explaining this fix ?
> i.e. your change drops the StdErr output. why would that fix the arm testing 

This PR is for testing the change on Travis. I am still evaluating this situation: calling expect script from inside go is taking a very long time. Causing tests to timeout and fail.

tsachiherman:
I think that we no longer need this PR; am I right here ?

algonautshant:
Yes, we no longer need this.

PR #427, opened 2019-10-25
Improve goal asset

justicz:
Don't make `--creator` a required flag, ensure that `--creator` is set if referring to asset only by `--asset`, make `goal asset info` functional with only `--assetid`.


cmd/goal/asset.go
```diff
@@ -68,8 +68,8 @@ func init() {
 	createAssetCmd.Flags().StringVar(&noteBase64, "noteb64", "", "Note (URL-base64 encoded)")
 	createAssetCmd.Flags().StringVarP(&noteText, "note", "n", "", "Note text (ignored if --noteb64 used also)")
 	createAssetCmd.Flags().BoolVarP(&noWaitAfterSend, "no-wait", "N", false, "Don't wait for transaction to commit")
-	createAssetCmd.MarkFlagRequired("creator")
 	createAssetCmd.MarkFlagRequired("total")
+	createAssetCmd.MarkFlagRequired("creator")
```
rotemh: Seems like it's required.

justicz: for create it is required

PR #428, opened 2019-10-25
Logicsig cache fix

algobolson:
I'm pretty sure this fixes the logic sig cache in the TransactionPool.
My best guess at a test for this is to build a pool, submit a lsig txn to it, cause the block to roll over, and check that logic.Eval gets called again to re-validate the lsig txn. But this sounds like an annoyingly large amount of setup. Not sure if there's a better way.

### DIFF
--- a/data/pools/transactionPool.go
+++ b/data/pools/transactionPool.go
@@ -30,6 +30,7 @@ import (
 	"github.com/algorand/go-algorand/ledger"
 	"github.com/algorand/go-algorand/logging"
 	"github.com/algorand/go-algorand/logging/telemetryspec"
+	"github.com/algorand/go-algorand/protocol"
 	"github.com/algorand/go-algorand/util/condvar"
 )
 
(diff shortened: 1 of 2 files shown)


data/pools/transactionPool.go
```diff
@@ -366,6 +375,14 @@ func (pool *TransactionPool) OnNewBlock(block bookkeeping.Block) {
 	var knownCommitted uint
 	var unknownCommitted uint
 
+	if block.CurrentProtocol != pool.currentProtocol {
+		// invalidate caches that could depend on protocol
+		pool.lcmu.Lock()
+		pool.lsigCache.clear()
+		pool.lcmu.Unlock()
+		pool.currentProtocol = block.CurrentProtocol
```
justicz: should this be inside the locked region? I think it might race with the check on 378 if `OnNewBlock` is called twice

justicz:
Hmm... so it looks like this is flushing the cache when it sees a validated block for the new protocol (in `OnNewBlock`). But if the node running the pool is itself proposing that block, the pool will assume we're still on the previous protocol version (and be caching `Lsigs` that might be invalid for this round).

If the above is correct, should we be checking if a protocol upgrade just got enabled in `OnNewBlock`?

algobolson:
see #437 for new version that makes lookup keyed on (protocol version, txid)

PR #429, opened 2019-10-25
Log check/eval panics for telemetry

justicz:
We will want to know if/when these panics happen. Test coverage in `eval_test.go`'s `TestPanic`.

### DIFF
--- a/data/transactions/logic/eval.go
+++ b/data/transactions/logic/eval.go
@@ -181,8 +181,8 @@ func Eval(program []byte, params EvalParams) (pass bool, err error) {
 					errstr += sb.String()
 				}
 			}
-			logging.Base().Errorf("recovered panic in Eval: %s", errstr)
 			err = PanicError{x, errstr}
+			logging.Base().Errorf("recovered panic in Eval: %s", err)
 		}
 	}()
 	if (params.Proto == nil) || (params.Proto.LogicSigVersion == 0) {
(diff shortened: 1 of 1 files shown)


winder:
How often do these happen? We don't want to flood telemetry with too many events. Also, does it need to be an error instead of a warning? I think we attach a log history with errors, and maybe that isn't necessary.

justicz:
The goal is that these recovered panics never happen -- they're what the [TEAL fuzzer](https://github.com/algorand/tealfuzz) is trying to find. In terms of severity we should definitely consider it more of an "error" than a "warning", but we probably don't need to include logs (also, the error message itself includes a stack trace)

Should I make it a warning?

algobolson:
@winder it's a "this should never happen" level of error that means we had a programming bug. pretty sure it's exactly the thing we should be reporting to ourselves.

winder:
Ok, error sounds right

data/transactions/logic/eval.go
```diff
@@ -180,6 +181,7 @@ func Eval(program []byte, params EvalParams) (pass bool, err error) {
 					errstr += sb.String()
 				}
 			}
+			logging.Base().Errorf("recovered panic in Eval: %s", errstr)
```
algobolson: you should include the panic value x in these. see `func (pe PanicError) Error() string`

justicz: thanks, I'm just `Errorf`ing the `PanicError` now

justicz:
An error here indicates a bug in _our_ code, not just in the TEAL script's code. A recovered panic indicates a programming mistake in our interpreter/checker.

PR #431, opened 2019-10-25
Fix goal asset transactions 

stechu:
## Summary

The current implementation of `FillUnsignedTxTemplate` will overwrite the txn's header, this will unset some fields of txn header such as `Note`. As a result, all the setters of the `Note` will be ineffective:
* https://github.com/algorand/go-algorand/blob/master/cmd/goal/asset.go#L221
* https://github.com/algorand/go-algorand/blob/master/cmd/goal/asset.go#L281
* https://github.com/algorand/go-algorand/blob/master/cmd/goal/asset.go#L362
*  https://github.com/algorand/go-algorand/blob/master/cmd/goal/asset.go#L436
* https://github.com/algorand/go-algorand/blob/master/cmd/goal/asset.go#L493

This PR change the implementation of `FillUnsignedTxTemplate` so that it didn't overwrite the entire txn header.

## Test Plan

All test passed.

### DIFF
--- a/libgoal/transactions.go
+++ b/libgoal/transactions.go
@@ -365,12 +365,10 @@ func (c *Client) FillUnsignedTxTemplate(sender string, firstValid, numValidRound
 	parsedLastValid := basics.Round(firstValid + numValidRounds)
 	parsedFee := basics.MicroAlgos{Raw: fee}
 
-	tx.Header = transactions.Header{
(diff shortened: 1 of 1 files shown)


tsachiherman:
The change overall looks good. I don't think it would trigger any test failures for the reason that we don't have any. ( but we should! )

PR #433, opened 2019-10-26
Adjust scenario2 for high transaction volume

tsachiherman:
## Summary

Adjust the scenario2 network transaction cost throttling so that we could achieve high tx volume without increasing the fees.


stechu:
LGTM.

PR #435, opened 2019-10-26
Pingpong Asset Support

stechu:
## Summary

Add asset transaction support in Pingpong:

* add `--numasset` option to pingpong. When `numasset` is assigned (1 - 1000), pingpong will create `numasset` assets and init these assets in each account. Pingpong will then send asset transactions instead of normal payment transactions.

## Test Plan

Tested locally.

### DIFF
--- a/shared/pingpong/pingpong.go
+++ b/shared/pingpong/pingpong.go
@@ -19,6 +19,7 @@ package pingpong
 import (
 	"context"
 	"fmt"
+	v1 "github.com/algorand/go-algorand/daemon/algod/api/spec/v1"
 	"math"
 	"math/rand"
 	"os"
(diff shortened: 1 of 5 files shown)


shared/pingpong/pingpong.go
```diff
@@ -320,7 +319,27 @@ func constructTxn(from, to string, fee, amt uint64, client libgoal.Client, cfg P
...
+		txn, err = client.ConstructPayment(from, to, fee, amt, noteField[:], "", [32]byte{}, 0, 0)
+		if !cfg.Quiet {
+			fmt.Fprintf(os.Stdout, "Sending %d : %s -> %s\n", amt, from, to)
+		}
+	} else {
+		var assetID uint64
+		for k := range assetParams {
+			assetID = k
+			break
+		}
```
tsachiherman: We want to pick a random assetID. this code would produce "undefined" assetID, but that id will remain the same across multiple iterations.

stechu:
After reading the asset code, I realized that sending 0 asset during the test cannot properly test the ledger behavior. Will update the implementation.

PR #436, opened 2019-10-27
ledger wallet API: support sending full tx msgpack to device

zeldovich:
This allows us to add support for additional transaction features (like
note fields or new asset transaction types) without having to keep
modifying kmd.  Fallback to older format of transaction encoding if the
Ledger device is running an old version of the Algorand app.

### DIFF
--- a/daemon/kmd/wallet/driver/ledger_hid.go
+++ b/daemon/kmd/wallet/driver/ledger_hid.go
@@ -36,7 +36,7 @@ type LedgerUSBError uint16
 
 // Error satisfies builtin interface `error`
 func (err LedgerUSBError) Error() string {
-	return fmt.Sprintf("unexpected status %x", err)
+	return fmt.Sprintf("Exchange: unexpected status 0x%x", err)
 }
 
 // Protocol reference:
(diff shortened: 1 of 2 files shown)


daemon/kmd/wallet/driver/ledger_hid.go
```diff
@@ -30,6 +30,15 @@ type LedgerUSB struct {
 	hiddev *hid.Device
 }
 
+// LedgerUSBError is a wrapper around the two-byte error code that the Ledger
+// protocol returns.
+type LedgerUSBError uint16
+
+// Error satisfies builtin interface `error`
+func (err LedgerUSBError) Error() string {
+	return fmt.Sprintf("unexpected status %x", err)
```
tsachiherman: nit: change it to `Exchange: unexpected status 0x%x` so that we'll know where this issue came from as well as making it easier to decode.

zeldovich: Good point.  Pushed; can you re-approve?

zeldovich:
> I would generally be against adding new features, but this one doesn't really interact with any of the others, so I guess there is no reason to block it from going to the upcoming release, right ?

Indeed, this code seems highly orthogonal to anything else going on in the codebase, so even in the worst case, it should at most break support for the Ledger wallet.

PR #437, opened 2019-10-28
LogicSig Eval caching explicitly based on (protocol version, txid)

algobolson:
alternative to #428 where protocol version is part of the cache key. I think this is probably simpler and more correct.

### DIFF
--- a/data/pools/lsigEvalCache.go
+++ b/data/pools/lsigEvalCache.go
@@ -1,0 +1,93 @@
+// Copyright (C) 2019 Algorand, Inc.
+// This file is part of go-algorand
+//
+// go-algorand is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as
+// published by the Free Software Foundation, either version 3 of the
+// License, or (at your option) any later version.
+//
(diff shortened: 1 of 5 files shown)


data/pools/lsigEvalCache.go
```diff
@@ -0,0 +1,93 @@
...
+import (
+	"github.com/algorand/go-algorand/data/transactions"
+	"github.com/algorand/go-algorand/protocol"
+)
+
+type lsigEvalSubcache struct {
+	they map[protocol.ConsensusVersion]map[transactions.Txid]error
+}
+
+func (les *lsigEvalSubcache) get(cvers protocol.ConsensusVersion, txid transactions.Txid) (err error, found bool) {
```
zeldovich: A common idiom is to have the error be the last return value when returning multiple results.

zeldovich: Ah, sorry, I didn't fully understand what's going on here.  I wonder if it warrants a comment saying this err is the cache lookup result..

algobolson:
passes checks now, re-approve?

PR #440, opened 2019-10-28
update config to allow upgrade to latest features.

tsachiherman:


tsachiherman:
> It seems rather dangerous to redefine an existing protocol..

Correct. It is.
However, this is a special case where we never upgraded to this protocol, and we have complete control of the nodes on the network. This change will not got into mainnet. Instead, for mainnet, we're going to jump from v17 to v19, which will have corrected spec. ( while supporting a dummy v18->v19 on betanet )

PR #441, opened 2019-10-28
remove unused EvalParams.Block field

algobolson:
delete dead code


tsachiherman:
Deleting dead-code is great. I think it should be ok after you'll handle all the build failures.

PR #443, opened 2019-10-28
Add lease argument to changing online status in goal

derbear:


cmd/goal/account.go
```diff
@@ -181,6 +184,20 @@ func init() {
 	markNonparticipatingCmd.Flags().BoolVarP(&noWaitAfterSend, "no-wait", "N", false, "Don't wait for transaction to commit")
 }
 
+func scLeaseBytes(cmd *cobra.Command) (leaseBytes [32]byte) {
```
tsachiherman: Looks to me that we should use the new function that you created for clerk.go:306-316 as well.

derbear: I would, but that uses the `lease` variable and not `statusChangeLease`. Maybe we should figure out what to do with all the duplicate variables we have across the different goal commands.

PR #444, opened 2019-10-29
Two fixes to possible infinite recursion in ::AlgorandGoal::StopNetwork

algonautshant:
Fix 1:
::AlgorandGoal::Abort calls ::AlgorandGoal::StopNetwork.
::AlgorandGoal::StopNetwork calls ::AlgorandGoal::Abort when fails to close the network.
This may cause infinite recursion when the conditions are right (e.g. is failing to close the network).
Fix: StopNetwork does not call Abort. 

Fix 2:
In ::AlgorandGoal::StopNetwork, when the spawn command output does not match the "Network Stopped under*", then NETWORK_STOP_MESSAGE is not set. If the spawn command did not timeout, then the "puts $NETWORK_STOP_MESSAGE" statement is executed. This is an expect script exception.
When ::AlgorandGoal::StopNetwork is called from try/catch statement, the exception is caught, and ::AlgorandGoal::Abort is called, which calls ::AlgorandGoal::StopNetwork.
Fix: initialize the variable to empty string. 

## Summary

## Test Plan

### DIFF
--- a/test/e2e-go/cli/goal/expect/goalExpectCommon.exp
+++ b/test/e2e-go/cli/goal/expect/goalExpectCommon.exp
@@ -97,10 +97,18 @@ proc ::AlgorandGoal::StartNetwork { NETWORK_NAME NETWORK_TEMPLATE TEST_ALGO_DIR 
 # Stop the network
 proc ::AlgorandGoal::StopNetwork { NETWORK_NAME TEST_ALGO_DIR TEST_ROOT_DIR } {
     set timeout 60
+    set NETWORK_STOP_MESSAGE ""
     puts "Stopping network: $NETWORK_NAME"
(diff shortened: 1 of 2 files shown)


test/e2e-go/cli/goal/expect/goalExpectCommon.exp
```diff
@@ -99,8 +99,16 @@ proc ::AlgorandGoal::StopNetwork { NETWORK_NAME TEST_ALGO_DIR TEST_ROOT_DIR } {
     set timeout 60
     puts "Stopping network: $NETWORK_NAME"
     spawn goal network stop -d $TEST_ALGO_DIR -r $TEST_ROOT_DIR
+    set NETWORK_STOP_MESSAGE "";
```
tsachiherman: I think that the semi-colon at the end of line is redundant. you might want to remove it to match the coding style here.

PR #445, opened 2019-10-29
Make asset name, URL, unit name fields variable length

justicz:
Byte arrays don't make a ton of sense for these fields since we have to trim null bytes everywhere we want to display them. @rotemh

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -192,6 +192,15 @@ type ConsensusParams struct {
 	// max number of assets per account
 	MaxAssetsPerAccount int
 
+	// max length of asset name
+	MaxAssetNameBytes int
+
+	// max length of asset unit name
+	MaxAssetUnitNameBytes int
+
+	// max length of asset url
+	MaxAssetURLBytes int
+
 	// support sequential transaction counter TxnCounter
 	TxnCounter bool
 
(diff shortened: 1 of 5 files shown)


config/config.go
```diff
@@ -425,6 +434,9 @@ func initConsensusProtocols() {
 	vFuture.LogicSigMaxSize = 1000
 	vFuture.LogicSigMaxCost = 20000
 	vFuture.MaxAssetsPerAccount = 1000
+	vFuture.MaxAssetNameBytes = 32
+	vFuture.MaxAssetUnitNameBytes = 8
+	vFuture.MaxAssetURLBytes = 32
```
tsachiherman: We can no longer make these changes. However, since these are all new parameters, you could add them to v7 as the default ?

tsachiherman:
Could you verify if the encoding of [32]byte and string are the same in msg pack ?

justicz:
Unfortunately, it does encode differently. Here is a test that fails:

```
type A struct {
        Foo [32]byte `codec:"algo"`
}

type B struct {
        Foo string `codec:"algo"`
}

func TestMsgPackAssetNewOld(t *testing.T) {
        var f [32]byte
        for i := 0; i < 32; i++ {
                // A\x00\x00...
                // AA\x00\x00...
                // AAA\x00\x00...
                s := strings.Repeat("A", i + 1)
                s = s + strings.Repeat("\x00", 31 - i)

                copy(f[:], s[:])

                a := protocol.Encode(A{Foo: f})
                b := protocol.Encode(B{Foo: s})
                require.Equal(t, a, b)
        }
}
```

PR #447, opened 2019-10-29
Fix version in config file generated by goal cmd

pzbitskiy:
## Summary

Before the fix 'goal network create' generated config.json
without 'Version' field. That forces config loader to assume Version=0
and start migrations.
If BroadcastConnectionsLimit set to 0 (special value) it is overwritten by -1
that is a new parameter introduced in config v4.
BroadcastConnectionsLimit does not exist in v3 with assumed value 0 in migration.

## Test Plan

No tests

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -731,6 +731,8 @@ func loadConfigFromFile(configFile string) (c Local, err error) {
 	}
 
 	// Migrate in case defaults were changed
+	// If a config file does not have version, it is assumed to be zero.
+	// All fields listed in migrate() might be changed if an actual value matches to default value from a previous version.
 	c, err = migrate(c)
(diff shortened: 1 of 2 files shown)


config/config.go
```diff
@@ -731,6 +731,8 @@ func loadConfigFromFile(configFile string) (c Local, err error) {
 	}
 
 	// Migrate in case defaults were changed
+	// If a config file does not have version, it is assumed to be zero.
+	// All fields listed in migrate() might be changed if an actual value matches to default value from a previous version.
```
pzbitskiy: 1. No logger exists to produce a message
2. Can't assume the latest version because missed version means 0.

PR #448, opened 2019-10-29
Fix fee increase in transaction pool

stechu:
## Summary

My previous PR #404  did a wrong change in transaction pool mechanism. The fee need to increase whenever the number of whole blocks in transaction pool increases (not only on receiving a new block). Otherwise, the transaction pool will leave DDoS opportunities.

### DIFF
--- a/data/pools/transactionPool.go
+++ b/data/pools/transactionPool.go
@@ -233,6 +233,16 @@ func (pool *TransactionPool) test(txgroup []transactions.SignedTxn) error {
 	// requires a flat MinTxnFee).
 	feePerByte = feePerByte * pool.feeThresholdMultiplier
 
+	// The threshold grows exponentially if there are multiple blocks
+	// pending in the pool.
+	if pool.numPendingWholeBlocks > 1 {
+		// golang has no convenient integer exponentiation, so we just
+		// do this in a loop
+		for i := 0; i < int(pool.numPendingWholeBlocks)-1; i++ {
+			feePerByte *= pool.expFeeFactor
+		}
+	}
+
 	for _, t := range txgroup {
 		feeThreshold := feePerByte * uint64(t.GetEncodedLength())
 		if t.Txn.Fee.Raw < feeThreshold {
--- a/data/pools/transactionPool.go
+++ b/data/pools/transactionPool.go
@@ -233,15 +233,20 @@ func (pool *TransactionPool) test(txgroup []transactions.SignedTxn) error {
 	// requires a flat MinTxnFee).
 	feePerByte = feePerByte * pool.feeThresholdMultiplier
 
+	// The feePerByte should be bumped to 1 to make the exponentially
+	// threshold growing valid.
+	if (feePerByte == 0) && pool.numPendingWholeBlocks > 0 {
+		feePerByte = uint64(1)
+	}
+
 	// The threshold grows exponentially if there are multiple blocks
 	// pending in the pool.
-	if pool.numPendingWholeBlocks > 1 {
-		// golang has no convenient integer exponentiation, so we just
-		// do this in a loop
-		for i := 0; i < int(pool.numPendingWholeBlocks)-1; i++ {
-			feePerByte *= pool.expFeeFactor
-		}
+	// golang has no convenient integer exponentiation, so we just
+	// do this in a loop
+	for i := 0; i < int(pool.numPendingWholeBlocks)-1; i++ {
+		feePerByte *= pool.expFeeFactor
 	}
+
 
 	for _, t := range txgroup {
 		feeThreshold := feePerByte * uint64(t.GetEncodedLength())
--- a/data/pools/transactionPool.go
+++ b/data/pools/transactionPool.go
@@ -235,7 +235,7 @@ func (pool *TransactionPool) test(txgroup []transactions.SignedTxn) error {
 
 	// The feePerByte should be bumped to 1 to make the exponentially
 	// threshold growing valid.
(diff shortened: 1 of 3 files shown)


data/pools/transactionPool.go
```diff
@@ -233,6 +233,16 @@ func (pool *TransactionPool) test(txgroup []transactions.SignedTxn) error {
...
 
+	// The threshold grows exponentially if there are multiple blocks
+	// pending in the pool.
+	if pool.numPendingWholeBlocks > 1 {
+		// golang has no convenient integer exponentiation, so we just
+		// do this in a loop
+		for i := 0; i < int(pool.numPendingWholeBlocks)-1; i++ {
+			feePerByte *= pool.expFeeFactor
+		}
+	}
```
tsachiherman: nit : I think it would be better to have the multiplier being calculated along with `pool.numPendingWholeBlocks`, so that we could always multiply the feePerByte with that multiplier.
Not only that it would be faster, but it would also have constant time.

btw - Correct me if I'm wrong, but with this change, you're practically rolling back your previous change while making it `pool.expFeeFactor` adjustable, right ?

justicz:
My understanding of the fee code is now this:

- `feeThresholdMultiplier` begins at 0.
- `numPendingWholeBlocks` begins at 0 and represents how many _whole_ blocks worth of transactions we have in the pool.
- When testing if a transaction can be included into the pool, we do this (assuming this PR is merged):

```
	feePerByte := uint64(1)

	// feePerByte can go to zero since feeThresholdMultiplier can be zero
	feePerByte = feePerByte * pool.feeThresholdMultiplier

	// if there is MORE THAN ONE pending whole block worth of txns in memory, multiply the
	// minimum fee per byte by pool.expFeeFactor ** (pool.numPendingWholeBlocks - 1)
	if pool.numPendingWholeBlocks > 1 {
		for i := 0; i < int(pool.numPendingWholeBlocks)-1; i++ {
			feePerByte *= pool.expFeeFactor
		}
	}

	// Now, for each transaction, test if it meets the minimum fee per byte
	for _, t := range txgroup {
		feeThreshold := feePerByte * uint64(t.GetEncodedLength())
		if t.Txn.Fee.Raw < feeThreshold {
			return fmt.Errorf("fee %d below threshold %d (%d per byte * %d bytes)",
				t.Txn.Fee, feeThreshold, feePerByte, t.GetEncodedLength())
		}
	}
```
- But _also_, `feeThresholdMultiplier`, is separately increased in `OnNewBlock`:
```
		switch pool.numPendingWholeBlocks {
		case 0:
			pool.feeThresholdMultiplier = pool.feeThresholdMultiplier / pool.expFeeFactor

		case 1:
			// Keep the fee multiplier the same.

		default:
			if pool.feeThresholdMultiplier == 0 {
				pool.feeThresholdMultiplier = 1
			} else {
				pool.feeThresholdMultiplier = pool.feeThresholdMultiplier * pool.expFeeFactor
			}
		}
```

These two different fee increases happen for different reasons. The first happens as we get bombarded with transactions from the network. The second happens sort of like an "integral" component -- the fee will keep getting higher as we have sustained transactions in memory over time.

One weird thing is that the second increase happens _before_ we recompute `numPendingWholeBlocks`. This means that if `numPendingWholeBlocks > 1`, we will definitely, 100% of the time increase `numPendingWholeBlocks` in `OnNewBlock`. Should we be recomputing `numPendingWholeBlocks` before the second increase?

data/pools/transactionPool.go
```diff
@@ -233,6 +233,16 @@ func (pool *TransactionPool) test(txgroup []transactions.SignedTxn) error {
 	// requires a flat MinTxnFee).
 	feePerByte = feePerByte * pool.feeThresholdMultiplier
 
+	// The threshold grows exponentially if there are multiple blocks
+	// pending in the pool.
+	if pool.numPendingWholeBlocks > 1 {
```
zeldovich: This `if` statement is no longer needed because you replaced the bitshift with a for loop.

data/pools/transactionPool.go
stechu: That would not be the case. Since if `pool.expFeeFactor` set to 1, then there is still no fee increase.

justicz:
ok, actually I think there is a bug here that was here before as well. `feeThresholdMultiplier` can go to zero. so if I just send you tons of transactions when it's zero, we never increase feePerByte
because

```
    feePerByte := uint64(1)

    feePerByte = feePerByte * pool.feeThresholdMultiplier

    if pool.numPendingWholeBlocks > 1 {
        // golang has no convenient integer exponentiation, so we just
        // do this in a loop
        for i := 0; i < int(pool.numPendingWholeBlocks)-1; i++ {
            feePerByte *= pool.expFeeFactor
        }
    }
```
will be just raising 0 to larger and larger powers until `OnNewBlock` is called

data/pools/transactionPool.go
```diff
@@ -233,6 +233,20 @@ func (pool *TransactionPool) test(txgroup []transactions.SignedTxn) error {
 	// requires a flat MinTxnFee).
 	feePerByte = feePerByte * pool.feeThresholdMultiplier
 
+	// The feePerByte should be bumped to 1 to make the exponentially
+	// threshold growing valid.
+	if feePerByte == 0 && pool.numPendingWholeBlocks > 0 {
```
zeldovich: This should be >1 since if we have just one whole block, we are not saturated yet..

stechu: Make sense. Thanks!

PR #451, opened 2019-10-30
make `txn FirstValidTime` actually the time of FirstValid-1

algobolson:
close hole where FirstValidTime of the block currently being assembled is hard to pin down

### DIFF
--- a/ledger/eval.go
+++ b/ledger/eval.go
@@ -587,10 +587,11 @@ func (eval *BlockEvaluator) transaction(txn transactions.SignedTxn, ad transacti
 }
 
 func (eval *BlockEvaluator) checkLogicSig(txn transactions.SignedTxn, txgroup []transactions.SignedTxnWithAD, groupIndex int) (err error) {
-	firstValid := basics.Round(txn.Txn.FirstValid - 1)
-	var hdr bookkeeping.BlockHeader
+	if txn.Txn.FirstValid == 0 {
+		return errors.New("LogicSig does not work with FirstValid==0")
+	}
 	// TODO: move this into some lazy evaluator for the few scripts that actually use `txn FirstValidTime` ?
-	hdr, err = eval.l.BlockHdr(firstValid)
+	hdr, err := eval.l.BlockHdr(basics.Round(txn.Txn.FirstValid - 1))
 	if err != nil {
 		return fmt.Errorf("could not fetch BlockHdr for FirstValid-1=%d (current=%d): %s", txn.Txn.FirstValid-1, eval.block.BlockHeader.Round, err)
 	}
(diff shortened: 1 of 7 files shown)


ledger/eval.go
```diff
@@ -587,24 +587,23 @@ func (eval *BlockEvaluator) transaction(txn transactions.SignedTxn, ad transacti
...
-		Proto:               &eval.proto,
-		TxnGroup:            txgroup,
-		GroupIndex:          groupIndex,
-		FirstValidTimeStamp: uint64(hdr.TimeStamp),
+		Txn:        &txn,
+		Proto:      &eval.proto,
+		TxnGroup:   txgroup,
+		GroupIndex: groupIndex,
+	}
+	if hdr.TimeStamp < 0 {
```
tsachiherman: if we're already picking up a date, why not to pick a more recent one ?

ledger/eval.go
```diff
@@ -587,24 +587,23 @@ func (eval *BlockEvaluator) transaction(txn transactions.SignedTxn, ad transacti
 }
 
 func (eval *BlockEvaluator) checkLogicSig(txn transactions.SignedTxn, txgroup []transactions.SignedTxnWithAD, groupIndex int) (err error) {
-	firstValid := basics.Round(txn.Txn.FirstValid)
+	firstValid := basics.Round(txn.Txn.FirstValid - 1)
```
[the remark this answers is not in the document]
tsachiherman: nit: change the variable name into "preFirstValid" to avoid confusion.

ledger/eval.go
algobolson: this is just to check that the cast from int64 to uint64 is clean

algobolson:
this passes tests and I believe issues are addressed. re-check?

ledger/eval.go
```diff
@@ -587,24 +587,24 @@ func (eval *BlockEvaluator) transaction(txn transactions.SignedTxn, ad transacti
...
-		if err != nil {
-			return fmt.Errorf("could not fetch BlockHdr for FirstValid=%d (current=%d): %s", txn.Txn.FirstValid, eval.block.BlockHeader.Round, err)
-		}
+	if txn.Txn.FirstValid == 0 {
+		return errors.New("LogicSig does not work with FirstValid==0")
+	}
+	// TODO: move this into some lazy evaluator for the few scripts that actually use `txn FirstValidTime` ?
+	hdr, err := eval.l.BlockHdr(basics.Round(txn.Txn.FirstValid - 1))
+	if err != nil {
+		return fmt.Errorf("could not fetch BlockHdr for FirstValid-1=%d (current=%d): %s", txn.Txn.FirstValid-1, eval.block.BlockHeader.Round, err)
```
tsachiherman: nit: last formatting object should be `%v` rather than `%s`

PR #452, opened 2019-10-30
make TestPanic quietly swallow panic logs when ok

algobolson:
make `go test` super quiet when all OK


algobolson:
passing in via (optional) value in EvalParams should be cleaner

PR #454, opened 2019-10-30
Optimize block assembly of previously included transactions

tsachiherman:
## Summary

Block assembly often get executed before the transactions from the previous round are cleared from the transactions pool.
Our existing implementation does detect these transactions and excludes them from the newly assembled proposal correctly, however, it does that with some undesired overhead.

This PR eliminate most of that overhead by taking the ledger lock *once*, acquiring the list of the transactions from the previous round, and using that list to filter out pending transactions.

### DIFF
--- a/ledger/eval.go
+++ b/ledger/eval.go
@@ -339,6 +339,11 @@ func (eval *BlockEvaluator) workaroundOverspentRewards(rewardPoolBalance basics.
 	return
 }
 
+// IsDup returns true if the transaction was included in the specified round.
+func (eval *BlockEvaluator) IsDup(txn transactions.SignedTxn, round basics.Round) (bool, error) {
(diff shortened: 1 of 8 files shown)


ledger/eval.go
```diff
@@ -339,6 +340,11 @@ func (eval *BlockEvaluator) workaroundOverspentRewards(rewardPoolBalance basics.
 	return
 }
 
+// GetRoundTxIds returns a map of the transactions ids that we have for the given round
+func (eval *BlockEvaluator) GetRoundTxIds(round basics.Round) (txMap map[transactions.Txid]bool) {
```
zeldovich: Can you expose this directly through `Ledger`?  It seems to be unrelated to the block evaluator, and you have access to the ledger at the call site.

tsachiherman: make sense. will do!

PR #455, opened 2019-10-30
add --proto to `goal clerk sign` and `goal clerk dryrun`

algobolson:


cmd/goal/clerk.go
```diff
@@ -563,6 +564,22 @@ func lsigFromArgs(lsig *transactions.LogicSig) {
...
 }
 
+func getProto(versArg string) config.ConsensusParams {
+	cvers := protocol.ConsensusCurrentVersion
+	if protoVersion != "" {
+		cvers = protocol.ConsensusVersion(protoVersion)
+	}
+	proto, ok := config.Consensus[cvers]
+	if !ok {
+		fmt.Fprintf(os.Stderr, "invalid consensus version. possible versions:\n")
```
pzbitskiy: Upper case Possible?

cmd/goal/clerk.go
```diff
@@ -135,7 +136,7 @@ func init() {
 	dryrunCmd.Flags().StringVarP(&txFilename, "txfile", "t", "", "transaction or transaction-group to test")
 	dryrunCmd.Flags().Uint64VarP(&round, "round", "r", 1, "round number to simulate")
 	dryrunCmd.Flags().Int64VarP(&timeStamp, "time-stamp", "S", 0, "unix time stamp simulate (default now)")
-	dryrunCmd.Flags().StringVarP(&seedBase64, "seed", "R", "2JNZvkCE2KFv3wVbEIEsPPNR64ttpaWuD5NEIXqXc3g=", "base64 bytes to seed random number generator with")
+	dryrunCmd.Flags().StringVarP(&protoVersion, "proto", "P", "", "consensus protocol version id string")
```
pzbitskiy: maybe put a default `config.Consensus[protocol.ConsensusCurrentVersion]` here and get rid of `getProto` function at all?

algobolson: still gonna keep getProto for its error messaging

cmd/goal/clerk.go
```diff
@@ -563,6 +564,22 @@ func lsigFromArgs(lsig *transactions.LogicSig) {
...
+	if protoVersion != "" {
+		cvers = protocol.ConsensusVersion(protoVersion)
+	}
+	proto, ok := config.Consensus[cvers]
+	if !ok {
+		fmt.Fprintf(os.Stderr, "Invalid consensus version. Possible versions:\n")
+		for xvers := range config.Consensus {
+			fmt.Fprintf(os.Stderr, "\t%s\n", xvers)
+		}
+		os.Exit(1)
```
tsachiherman: nit: most of the goal code is using `reportErrorf` to print and exit.

algobolson: yes, but it does one print and then exit, and I needed many prints and then an exit

PR #461, opened 2019-10-31
Add domain separators to opcode description.

derbear:


algobolson:
see #462 for change to doc.go to get this at the source

PR #464, opened 2019-10-31
add e2e test for teal script

algobolson:
exercises several cases of TLHC pattern with goal commands

### DIFF
--- a/cmd/goal/clerk.go
+++ b/cmd/goal/clerk.go
@@ -332,7 +332,12 @@ var sendCmd = &cobra.Command{
 		}
 		var stx transactions.SignedTxn
 		if lsig.Logic != nil {
-			proto := config.Consensus[protocol.ConsensusCurrentVersion]
+			params, err := client.SuggestedParams()
+			if err != nil {
+				reportErrorf(errorNodeStatus, err)
+			}
+
+			proto := config.Consensus[protocol.ConsensusVersion(params.ConsensusVersion)]
 			unsignedTxn := transactions.SignedTxn{Txn: payment}
 			err = verify.LogicSig(&lsig, &proto, &unsignedTxn)
 			if err != nil {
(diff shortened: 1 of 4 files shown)


cmd/goal/clerk.go
```diff
@@ -566,6 +571,17 @@ func getProto(versArg string) config.ConsensusParams {
 	cvers := protocol.ConsensusCurrentVersion
 	if versArg != "" {
 		cvers = protocol.ConsensusVersion(versArg)
+	} else {
+		dataDir := maybeSingleDataDir()
+		if dataDir != "" {
+			client := ensureAlgodClient(dataDir)
+			params, err := client.SuggestedParams()
```
tsachiherman: I think that you don't want this one for a dryrun

algobolson: I think it doesn't hurt to have it. It will try to contact an algod but fall back to defaults if an algod can't be found.

tsachiherman: My concern is not that we would try and fail. My concern is that connecting to a live node is opposed to "dryrun" mode. Do you think we could skip this portion for when we specify dryrun ?

tsachiherman:
Could you add some more printing so that in case of an error, we will be able to associate this particular test with the error string ? ( or the 'OK' result ? )

cmd/goal/clerk.go
algobolson: If you have $ALGORAND_DATA defined and this quietly checks with your network for the current version, that sounds like a good thing to me. `--proto` preempts and overrides that so the check with algod is never done. I think this is a good way to guess a reasonable default.

algobolson:
the bash script starts with `set -x` so that it echos every command it runs. we'll know which one failed.

algobolson:
ping. I think this is stable and good to go.

PR #471, opened 2019-11-05
Optimize OnNewBlock low hanging fruits

tsachiherman:
## Summary

Optimize low hanging fruits:
- reorder node and txpool OnNewBlock listeners.
- reduce the lock time of the node's OnNewBlock

## Explain
node's OnNewBlock typically takes 30-60ms
txpool OnNewBlock typically takes 200-600ms

we don't want the txpool's test to get blocked by waiting for the OnNewBlock for too long.
Easy solution - swap the order of the two so we can get free 30-60ms gains.

also, some minor improvements on the node's OnNewBlock locking.

### DIFF
--- a/node/node.go
+++ b/node/node.go
@@ -180,9 +180,7 @@ func MakeFull(log logging.Logger, rootDir string, cfg config.Local, phonebookDir
 		return nil, err
 	}
 
-	blockListeners := []ledger.BlockListener{
-		node,
-	}
+	blockListeners := []ledger.BlockListener{}
 
 	if node.config.EnableTopAccountsReporting {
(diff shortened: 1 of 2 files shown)


node/node.go
```diff
@@ -180,9 +180,7 @@ func MakeFull(log logging.Logger, rootDir string, cfg config.Local, phonebookDir
 		return nil, err
 	}
 
-	blockListeners := []ledger.BlockListener{
-		node,
-	}
+	blockListeners := []ledger.BlockListener{}
 
 	if node.config.EnableTopAccountsReporting {
 		blockListeners = append(blockListeners, &accountListener)
```
zeldovich: Maybe move this one down too?

PR #472, opened 2019-11-05
Improve TEAL template documentation.

derbear:
- All TEAL template files now contain comments documenting their function.
    
     - driver.py is replaced with algotmpl, a tool which extracts documentation and
       prints it.


tsachiherman:
Does this tool need to be distributed as part of our installation package, or is that just for developers ? also, it would be nice to have a "-v" option to find what version this binary was compiled with.

derbear:
> Does this tool need to be distributed as part of our installation package, or is that just for developers ?

I think this is an optional tool, since it's not necessary for running the node. (It's basically a simple search-and-replace tool used to demonstrate how to use the templates.)

> also, it would be nice to have a "-v" option to find what version this binary was compiled with.

Do other similar tools (e.g., `msgpacktool`) have such flags?

tools/teal/templates/limit-order.teal.tmpl
```diff
@@ -45,9 +68,9 @@ gtxn 1 AssetAmount
 &&
 &&
 int 1
-bnz label7
+bnz label3
 pop // check bug workaround
```
stechu: Do we still need that?

tools/teal/examples/dynamicfee.sh
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 
 # produce TEAL assembly for a delegated logic signature on a dynamic-fee transaction and compile it (note the required lease value)
-python driver.py dynamic-fee --cls AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAY5HFKQ --rcv NBH2CNQCPV7S2ZOSWVQ7JLFO7W5JERULPSGYUP4ZAEOLSVMTYODHVY37VQ --fv 170500 --lv 171500 --amt 100000 --x uFVDhjBpkpKQ8sZaau0qsDsf0eW3oXFEn1Ar5o39vkk= > dynamic.teal
+algotmpl -d `git rev-parse --show-toplevel`/tools/teal/templates dynamic-fee --cls AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAY5HFKQ --to NBH2CNQCPV7S2ZOSWVQ7JLFO7W5JERULPSGYUP4ZAEOLSVMTYODHVY37VQ --fv 170500 --lv 171500 --amt 100000 --lease uFVDhjBpkpKQ8sZaau0qsDsf0eW3oXFEn1Ar5o39vkk= > dynamic.teal
 goal clerk compile -a EVADN3MAXLTUAZFLJNIN7RD7WCNEPTB37GJMD5OACMVBKTGFXYYFMS5IT4 -s -o dynamic.lsig dynamic.teal -d .
```
stechu: It would be nice to replace these concrete addresses with variables so that some users could actually use these bash scripts (this perhaps improves readability as well)

stechu:
On another note, we should have end to end tests that cover each every template (maybe in other PRs).

tools/teal/templates/limit-order.teal.tmpl
derbear: Ah, this PR does not remove the workaround. The PR that fixes that is here: 

https://github.com/algorand/go-algorand/pull/481

I'm thinking this PR can be merged independently from that one.

tools/teal/examples/dynamicfee.sh
derbear: Yeah, probably the right thing to do is to replace these with end-to-end tests. The things in the example directory are purely "documentation" (which also is not exactly right). Maybe we can change the extension on the example directory so people don't execute the scripts and think that they work correctly.

tools/teal/templates/limit-order.teal.tmpl
stechu: 👍

stechu:
I think we should merge this PR first.

PR #474, opened 2019-11-05
Eric/nov performance recipe updates

egieseke:
Updates to Scenario1 for November Release performance testing

## Summary

Modified Scenario1 to support updated test plan. 

## Test Plan

Ran performance test according to matrix located here: 
https://docs.google.com/spreadsheets/d/1X1n5DFGDzF1sdiRmuLRgK-RYd57MO-UnRx0D9TT90Hk/


shared/pingpong/accounts.go
```diff
@@ -147,21 +150,30 @@ func prepareAssets(accounts map[string]uint64, client libgoal.Client, cfg PpConf
...
 
-	// get these assets
+	// submit the asset transactions and validate that they have all been accepted
 	for {
+		fmt.Printf("Creating assets\n")
+		for i := 0; i < toCreate; i++ {
+			txid, broadcastErr := client.BroadcastTransaction(*signedTrxs[i])
+			if broadcastErr != nil {
+				fmt.Printf("Cannot broadcast asset creation txn error: %v\n", broadcastErr)
+				//err = broadcastErr
```
stechu: Delete unused code

shared/pingpong/accounts.go
```diff
@@ -147,21 +150,30 @@ func prepareAssets(accounts map[string]uint64, client libgoal.Client, cfg PpConf
...
-		if broadcastErr != nil {
-			fmt.Printf("Cannot broadcast asset creation txn\n")
-			err = broadcastErr
-			return
-		}
+		signedTrxs[i] = &signedTxn
 
 		if !cfg.Quiet {
-			fmt.Printf("Create a new asset: supply=%d, txid=%s\n", totalSupply, txid)
+			fmt.Printf("Create a new asset: supply=%d \n", totalSupply)
```
stechu: Since we are not sending asset creation txns here. How about "Prepare a new asset"?

shared/pingpong/accounts.go
```diff
@@ -147,21 +150,30 @@ func prepareAssets(accounts map[string]uint64, client libgoal.Client, cfg PpConf
...
 		}
 		accounts[cfg.SrcAccount] -= tx.Fee.Raw
 	}
 
-	// get these assets
+	// submit the asset transactions and validate that they have all been accepted
 	for {
```
stechu: It might be helpful to add a timeout for the outer for loop so that we don't stuck here forever.

shared/pingpong/accounts.go
egieseke: Removed ping-pong changes from this PR.

shared/pingpong/accounts.go
egieseke: Removed ping-pong changes from this PR.

shared/pingpong/accounts.go
egieseke: Removed ping-pong changes from this PR.

PR #476, opened 2019-11-05
Allow closing deleted assets

justicz:
I introduced a bug in #403 that makes it impossible to close an asset holding if the overall asset has been deleted. If an asset has been deleted, looking up its creator will fail. This is bad because now there is an unrecoverable minbalance, since the holding/account can never be closed.

This also uncovered a higher level issue: we want to allow people to close out assets to the creator, even if their asset holding is frozen. What does this mean for an asset that has been deleted, since that asset no longer has a creator? To resolve this, this PR assumes that it is OK to close out an asset holding if the amount is zero. I am sending an email to discuss this (also b/c it is protocol incompatible with current betanet).

I manually tested that this change allows closing out a zero asset holding for an asset that has been deleted.

Thanks to Sam Abbassi for noticing this.

### DIFF
--- a/data/transactions/asset.go
+++ b/data/transactions/asset.go
@@ -327,41 +327,39 @@ func (ct AssetTransferTxnFields) apply(header Header, balances Balances, spec Sp
 			return fmt.Errorf("cannot close asset by clawback")
 		}
 
-		// Fetch the sender balance record to 1. ensure they're not
-		// the creator and 2. figure out how much balance to move
+		// Fetch the sender balance record. We will use this to ensure
+		// that the sender is not the creator of the asset, and to
+		// figure out how much of the asset to move.
 		snd, err := balances.Get(source, false)
 		if err != nil {
 			return err
 		}
 
-		// Cannot close asset ID allocated by this account; must use destroy.
+		// The creator of the asset cannot close their holding of the
+		// asset. Check if we are the creator by seeing if there is an
+		// AssetParams entry for the asset index.
 		if _, ok := snd.AssetParams[ct.XferAsset]; ok {
 			return fmt.Errorf("cannot close asset ID in allocating account")
 		}
 
-		// Fetch the asset holding
+		// Fetch our asset holding, which should exist since we're
+		// closing it out
 		sndHolding, ok := snd.Assets[ct.XferAsset]
 		if !ok {
 			return fmt.Errorf("asset %v not present in account %v", ct.XferAsset, source)
 		}
 
-		var bypassFreeze bool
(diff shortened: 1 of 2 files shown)


data/transactions/asset.go
```diff
@@ -322,44 +322,54 @@ func (ct AssetTransferTxnFields) apply(header Header, balances Balances, spec Sp
...
+		}
+
+		// Fetch the asset holding
 		sndHolding, ok := snd.Assets[ct.XferAsset]
 		if !ok {
 			return fmt.Errorf("asset %v not present in account %v", ct.XferAsset, source)
 		}
 
+		var bypassFreeze bool
+		if sndHolding.Amount == 0 {
```
zeldovich: Why is this needed?  `takeOut` and `putIn` special-case a zero amount already.

data/transactions/asset.go
```diff
@@ -322,44 +322,54 @@ func (ct AssetTransferTxnFields) apply(header Header, balances Balances, spec Sp
...
+		} else {
+			// Amount is nonzero, so asset shouldn't have been deleted.
+			// Creator should exist.
+			creatorAddr, err := balances.GetAssetCreator(ct.XferAsset)
+			if err != nil {
+				return fmt.Errorf("failed to find asset creator: %v", creatorAddr)
+			}
+
+			// Allow closing out to the asset creator even when frozen
+			if ct.AssetCloseTo == creatorAddr {
```
zeldovich: Instead of special-casing this if statement, maybe look up if the destination address has the asset in its `AssetParams`?  I think that'll require no if statements, then.

justicz: That seems much better, yes. And the 0 case just works because of the short circuit in `putIn` and `takeOut`

justicz: That seems much better, yes. And the 0 case just works because of the short circuit in `putIn` and `takeOut`

ledger/accountdb.go
```diff
@@ -183,6 +183,11 @@ func (qs *accountsDbQueries) lookupAssetCreator(assetIdx basics.AssetIndex) (add
 	err = db.Retry(func() error {
 		var buf []byte
 		err := qs.lookupAssetCreatorStmt.QueryRow(assetIdx).Scan(&buf)
+
+		if err == sql.ErrNoRows {
+			err = fmt.Errorf("asset %v does not exist or has been deleted", assetIdx)
```
tsachiherman: nit: "asset index %d does not exist or has been deleted"

justicz: updated, thanks, can you please reapprove?

PR #478, opened 2019-11-06
Eric/194 fix transaction submit

egieseke:
## Summary

Allow node to accept a transaction for rebroadcast even if it is already in the nodes transaction pool.  This is to fix the problem where the initial submit did not make it to parent relays.  

## Test Plan

Tested using ping-pong test as part of November release performance testing.


node/node.go
```diff
@@ -412,7 +412,6 @@ func (node *AlgorandFullNode) BroadcastSignedTxGroup(txgroup []transactions.Sign
 	err = node.transactionPool.Remember(txgroup)
 	if err != nil {
 		node.log.Infof("rejected by local pool: %v - transaction group was %+v", err, txgroup)
```
derbear: We'll probably want to add to the log message here that we're attempting to broadcast anyway, even though we've rejected the transaction from the pool.

egieseke: Updated info message to warning and added text to indicate that the transaction will be broadcast even though it was rejected from the pool.

PR #483, opened 2019-11-07
Add lease e2e tests

justicz:
Add tests for two identical leases from the same sender (one should fail), and two identical leases from different senders (both should succeed)

### DIFF
--- a/test/e2e-go/features/transactions/lease_test.go
+++ b/test/e2e-go/features/transactions/lease_test.go
@@ -73,7 +73,6 @@ func TestLeaseTransactionsSameSender(t *testing.T) {
 	// wait for the txids and check balance
 	txids := make(map[string]string)
 	txids[stx1.Txn.ID().String()] = account0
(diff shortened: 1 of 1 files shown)


test/e2e-go/features/transactions/lease_test.go
```diff
@@ -0,0 +1,160 @@
...
+	txids := make(map[string]string)
+	txids[tx.ID().String()] = account0
+	_, curRound := fixture.GetBalanceAndRound(account0)
+	confirmed := fixture.WaitForAllTxnsToConfirm(curRound+5, txids)
+	a.True(confirmed, "funding txn confirmed")
+
+	lease := [32]byte{1, 2, 3, 4}
+
+	// construct transactions for sending money to account1 and account2
+	// from same sender with identical lease
```
algobolson: copy pasta: _different_ sender

justicz: thx

PR #490, opened 2019-11-07
Add created asset index to txn status

justicz:
This PR adds a new field to `v1.Transaction` called `TransactionResults`. This field is a struct containing `CreatedAssetIndex`, the index (if any) of an asset created by this transaction.

It computes the created asset index by looking up the block associated with the transaction if it was confirmed (and the correct type of transaction), computing the offset within that block, and adjusting based on `TxnCounter`.

Currently we're just returning this on a best-effort basis. If we don't have access to the block, we can't compute the created asset index based on just the transaction id.

### DIFF
--- a/daemon/algod/api/server/v1/handlers/handlers.go
+++ b/daemon/algod/api/server/v1/handlers/handlers.go
@@ -225,17 +225,15 @@ func txWithStatusEncode(tr node.TxnWithStatus, l *data.Ledger) (v1.Transaction, 
 	}
 	s.ConfirmedRound = uint64(tr.ConfirmedRound)
 	s.PoolError = tr.PoolError
-	s.CreatedAssetIndex = computeAssetIndexFromTxn(tr, l)
+	s.TransactionResults = &v1.TransactionResults{
+		CreatedAssetIndex: computeAssetIndexFromTxn(tr, l),
+	}
 	return s, nil
 }
 
-func computeAssetIndexInBlock(tx node.TxnWithStatus, blk bookkeeping.Block) (aidx uint64) {
+func computeAssetIndexInPayset(tx node.TxnWithStatus, txnCounter uint64, payset []transactions.SignedTxnWithAD) (aidx uint64) {
 	// Compute transaction index in block
 	offset := -1
-	payset, err := blk.DecodePaysetFlat()
-	if err != nil {
-		return 0
-	}
 	for idx, stxnib := range payset {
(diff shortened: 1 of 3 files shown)


daemon/algod/api/server/v1/handlers/handlers.go
```diff
@@ -216,17 +217,74 @@ func assetFreezeTxEncode(tx transactions.Transaction, ad transactions.ApplyData)
...
 	s, err := txEncode(tr.Txn.Txn, tr.ApplyData)
 	if err != nil {
 		err = decorateUnknownTransactionTypeError(err, tr)
 		return v1.Transaction{}, err
 	}
 	s.ConfirmedRound = uint64(tr.ConfirmedRound)
 	s.PoolError = tr.PoolError
+	s.TransactionResults = &v1.TransactionResults{
+		CreatedAssetIndex: computeAssetIndexFromTxn(tr, l),
+	}
```
rotemh: Seems like we're enforcing behavior for future callers. i.e. all future functions must return the empty value in order to be omitted. I believe that the addition of extra information should happen at the caller level.

rotemh:
one comment, otherwise, looks good to me

daemon/algod/api/server/v1/handlers/handlers.go
justicz: fixed :+1:

daemon/algod/api/server/v1/handlers/handlers.go
```diff
@@ -682,12 +738,17 @@ func PendingTransactionInformation(ctx lib.ReqContext, w http.ResponseWriter, r
...
 	if txn, ok := ctx.Node.GetPendingTransaction(txID); ok {
+		ledger := ctx.Node.Ledger()
 		responseTxs, err := txWithStatusEncode(txn)
 		if err != nil {
 			lib.ErrorResponse(w, http.StatusInternalServerError, err, errFailedToParseTransaction, ctx.Log)
 			return
 		}
 
+		responseTxs.TransactionResults = &v1.TransactionResults{
+			CreatedAssetIndex: computeAssetIndexFromTxn(txn, ledger),
```
rotemh: can you just add the comment that it will be omitted if there is no assetID?

justicz: done :+1:

algobolson:
This is probably good but I feel like it would be better with a test that submits a create-asset transaction, gets this response, then looks up the creator's account info to see that the same asset id is listed there.

PR #492, opened 2019-11-08
Make updater forward compatible a timestamp prefix scheme

winder:
Looking for feedback, this makes the updater forward compatible, but 10x slower since it no longer filters by package.

### DIFF
--- a/util/s3/s3Helper.go
+++ b/util/s3/s3Helper.go
@@ -21,10 +21,12 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"path"
 	"path/filepath"
 	"regexp"
 	"runtime"
 	"strconv"
+	"strings"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/awserr"
@@ -262,7 +264,7 @@ func (helper *Helper) GetPackageFilesVersion(channel string, pkgFiles string, sp
 	maxVersion = 0
 	maxVersionName = ""
 
-	prefix := fmt.Sprintf("channel/%s/%s", channel, pkgFiles)
+	prefix := fmt.Sprintf("channel/%s/", channel)
 	svc := s3.New(helper.session)
 	input := &s3.ListObjectsInput{
 		Bucket:  &helper.bucket,
@@ -281,8 +283,14 @@ func (helper *Helper) GetPackageFilesVersion(channel string, pkgFiles string, sp
 	}
 
 	for _, item := range result.Contents {
-		var version uint64
 		name := string(*item.Key)
+
+		baseName := path.Base(name)
+		if !strings.HasPrefix(baseName, pkgFiles) {
+			continue
+		}
+
+		var version uint64
 		version, err = GetVersionFromName(name)
 		if err != nil {
 			return


algobolson:
current schema:
s3://algorand-releases/channel/{channel:dev,nightly,beta,stable}/{install,node,tools}_{channel}_{os:darwin,linux}_{arch:amd64,arm,arm64}_{version}.tar.gz

my proposed schema:
s3://algorand-releases/latest/{channel:dev,nightly,beta,stable}/{RSTAMP}_{version}/{install,node,tools}_{channel}_{os:darwin,linux}_{arch:amd64,arm,arm64}_{version}.tar.gz

so, same filenames but the path prefix goes from:
s3://algorand-releases/channel/{channel:dev,nightly,beta,stable}/
to
s3://algorand-releases/latest/{channel:dev,nightly,beta,stable}/{RSTAMP}_{version}/

The old way makes a directory with all the versions and all the variations in it. (not sustainable)
The new way makes a directory per version with all the package/os/arch variants in it.

Transitional updater will do two queries. The first query will be on the new schema, if it gets nothing, we haven't pushed anything to the new schema yet and it falls back on old logic for the old schema. Once we start pushing releases on the new schema the updater will be faster by using the new query only. Eventually we delete code for the old schema.

RSTAMP generator in `scripts/reverse_hex_timestamp`

tsachiherman:
I think that we need to discuss what is the expected number of releases we're planning to have. That would help us to shape the most suitable directory hierarchy.

winder:
What is the benefit of `latest`?
Would we remove files from this path at some point?
Why isn't it sustainable to have all of the channels under the same path? I don't think they should ever be listed.

To simplify the migration, in this PR I was imagining the new schema could simply inject the RSTAMP into the path and be:
s3://algorand-releases/channel/{channel: dev,nightly,beta,stable}/{RSTAMP}/{install,node,tools}_{channel}_{os:darwin,linux}-{arch:amd64,arm,arm64}_{version}.tar.gz

Wouldn't that give us the sorting benefit?

re: @tsachiherman's comment, we currently prefix a pattern up to the architecture, so we could have daily releases on a channel for almost 3 years before updater would have to scroll to see everything.

algobolson:
I want to change the root of the path because I want to deprecate `/channel/**`
Basically I consider the files there as 'wrong' and we should abandon them and move to something better. (of course storage is pretty cheap and we can leave them there archivally for quite a while)
If you `aws s3 ls s3://algorand-releases/channel/nightly/` you get 656 entries right now. I want to leave those all behind and move to a clean schema, so my best guess at something that would make sense to someone looking at the top level was `/latest/`

algobolson:
also
/channel/nightly/{RSTAMP}/... will sort before /channel/nightly/install*.tar.gz but only barely and it seems like a hack I don't want to rely on.

winder:
Thanks for the clarification, I think I understand your proposal now. When you say "only barely" because it only works because _install, node, and tools_ happen to sort after any hex value that RSTAMP would use as a prefix?

I don't think there's anything special in the new buckets yet, I'd be happy to delete the old builds and keep `/channel/<channel>/`. If we did that, I could update the code in this PR to stop iterating after finding something with an RSTAMP, and we might be done.

Maybe the 3 of us can discuss next week and see if this is something that we can sneak into the release, or if it's too late. There might be some other cases, like `updater send` would need to create an RSTAMP as well.

tsachiherman:
If I understand you correctly @algobolson, you're proposing to abandon the existing scheme and move toward a 
`s3://algorand-releases/latest/{channel:dev,nightly,beta,stable}/{RSTAMP}_{version}/`
I think that is make sense for the updater to have a single lookup directory where it wouldn't need to have multiple options. ( i.e. there will be only one version, and it's also happen to be the latest released version ).
We could keep the existing location to accumulate the rest of the versions, as we don't frequently look in there.

algobolson:
In the steady state the updater would only query the new schema. It would only query the old schema if nothing is found in the new schema. If anything is found in the new schema then that is used. I think we _shouldn't_ use the current directory schema with the new directory schema overlaid on top of it. (it's not actually _cannot_ , we can use the current directory and put new schema things into it, but it's not good design, just an accident that it can work)

PR #493, opened 2019-11-08
Preallocate the payset to avoid dynamic resizing

tsachiherman:
## About

The payset is the list of transactions that are included in a block.
During the block evaluation, we add the transactions, one at a time to that list.
This works correctly, however, not very efficient. Since arrays in GO are immutable, each time the array has to grow, it require a new array to be allocated, and the content need to be copied over.

Given that we know in advance the number of transactions that are needed, we should pre-allocate that.

## Metrics
When running S2 under heavy load, the BlockEvaluator::transactionGroup used to take 52.80s, out of them 7.1s spent on the growslice.
With this change while running the same scenario, BlockEvaluator::transactionGroup takes 56.36s, out of them 1.17s spent on the glowslice.

### DIFF
--- a/data/bookkeeping/block.go
+++ b/data/bookkeeping/block.go
@@ -389,7 +389,7 @@ func MakeBlock(prev BlockHeader) Block {
 		logging.Base().Panicf("MakeBlock: next protocol %v not supported", upgradeState.CurrentProtocol)
 	}
 
-	emptyPayset := make(transactions.Payset, 0)
+	var emptyPayset transactions.Payset
 
 	timestamp := time.Now().Unix()
 	if prev.TimeStamp > 0 {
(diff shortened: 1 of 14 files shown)


tsachiherman:
> How much does this help?

Expectation is to get to 20% improved performance when transaction pool has more than a single full block worth of transactions. I'm still working on gathering metrics and would update this PR and/or slack channel when done.

derbear:
Another option, instead of passing around the size everywhere, is just to allocate a slice of constant capacity of ~6K. This would be a much smaller change and produces about the same worst-case performance.

tsachiherman:
> Another option, instead of passing around the size everywhere, is just to allocate a slice of constant capacity of ~6K. This would be a much smaller change and produces about the same worst-case performance.

How come ? The performance drops is not caused by the memory allocator, but rather by the copying operation. Even if we set it arbitrarily to 6K, it will increase at 8K, 12K, 14K and so forth ( let's say up until 4x block capacity = 16K ). The performance drop is primarily due to the higher sized pending txn copies, which is exactly what we want to work around.

I can see how smarter grow function ( instead of the existing append ), would do that, though.

derbear:
Blocks are ~5K, so no copying will happen there.

If we set it to 6k, only two calls to copy will happen because each call doubles slice size (6K to 12K to 24K). 

We could even get rid of all copying completely by setting slice size to max pool size.

PR #500, opened 2019-11-08
test for split contrract template

algobolson:
slight tweak to split contract math


test/scripts/teal-split-test.sh
```diff
@@ -0,0 +1,68 @@
...
+
+# replaces prior trap0
+trap "goal network stop -r ${NETDIR}; rm -rf ${TEMPDIR}" 0
+
+export ALGORAND_DATA=${NETDIR}/Node
+
+ACCOUNT=$(goal account list|awk '{ print $3 }')
+
+ACCOUNTB=$(goal account new|awk '{ print $6 }')
+ACCOUNTC=$(goal account new|awk '{ print $6 }')
```
derbear: Just want to confirm--you intended this to be the same as `$ACCOUNTB`, right?

test/scripts/teal-split-test.sh
```diff
@@ -0,0 +1,68 @@
...
+goal clerk dryrun -t ${TEMPDIR}/group.stx
+
+goal clerk send -a 110000000 -f ${ACCOUNT} -t ${ACCOUNT_SPLIT}
+
+goal clerk rawsend -f ${TEMPDIR}/group.stx
+
+BALANCEB=$(goal account balance -a ${ACCOUNTB}|awk '{ print $1 }')
+BALANCEC=$(goal account balance -a ${ACCOUNTC}|awk '{ print $1 }')
+
+if [ $BALANCEB -ne 60000000 ]; then
```
derbear: I'm sure this works, but note that this depends on how rewards are being distributed.

derbear: (In particular, if this executes after `$ACCOUNTB` receives some rewards, it will fail.)

test/scripts/teal-split-test.sh
algobolson: same action, but again, so as to make two accounts B and C

test/scripts/teal-split-test.sh
algobolson: yeah, I'm hoping that being on a brand new network with no other traffic and everything happening within 20 rounds of start will keep that to a minimum.

derbear: Sounds good. Maybe leave a comment here just noting that.

tools/teal/templates/split.teal.tmpl
```diff
@@ -5,8 +5,8 @@
 // split, or single transaction, for closing the account.
 //
 // Withdrawals from this account are allowed as a group transaction which
-// sends TMPL_RCV1 exactly TMPL_RATN/TMPL_RATD times the sum of amounts
-// across both transactions.  At least TMPL_MINPAY must be sent to TMPL_RCV1.
+// sends TMPL_RCV1 and TMPL_RCV2 amounts with exactly the ratio of
+// TMPL_RATN/TMPL_RATD.  At least TMPL_MINPAY must be sent to TMPL_RCV1.
```
derbear: Also it might be good to change the text in the `TMPL_RATD` parameter documentation line below.

algobolson:
had to make a trivial change to pass tests, but they pass now, re-approve?

PR #503, opened 2019-11-12
Add markdown docs for atomic swap, delegated keyreg, dynamic fee, split

justicz:
Adding some docs that go over the code of each contract line-by-line. Publishing what I have so far so we can update NCC.


tools/teal/templates/docs/atomic-swap.teal.md
```diff
@@ -0,0 +1,142 @@
+# Atomic Swap
+
+## Functionality
+
+This contract implements part of an "atomic swap". It is intended to be used as a "contract only" account, not as a "delegated contract" account.
+
+The contract will approve transactions spending algos from itself under two circumstances:
+
+  1. If an argument `arg_0` is passed to the script such that `TMPL_HASHFN(arg_0)` is equal to `TMPL_HASHIMG`, then funds may be closed out to  `TMPL_RCV`.
```
derbear: ` `

derbear: Nit: extra space before `TMPL_RCV`.

tools/teal/templates/docs/delegate-keyreg.teal.md
```diff
@@ -0,0 +1,97 @@
...
+```
+txn FirstValid
+int TMPL_PERIOD
+%
+int 0
+==
+&&
+```
+
+Next, check that the lease field is exactly `TMPL_LEASE`.
```
derbear: It may be useful to explain why this and the previous few checks are important for together limiting the spending power of the delegate key.

tools/teal/templates/docs/delegate-keyreg.teal.md
```diff
@@ -0,0 +1,97 @@
+# Delegate Key Registration
+
+## Functionality
```
derbear: Something useful to note for this script is that we have a custom signing script `dsign` for this template. (see `tools/teal` directory)

justicz: added a comment -- what do you think?

tools/teal/templates/docs/delegate-keyreg.teal.md
justicz: added a comment -- what do you think?

PR #504, opened 2019-11-12
optimize oxtail for single comparison

tsachiherman:
<wip>


tsachiherman:
I think that Derek implementation is better, so I'm closing this one.

PR #505, opened 2019-11-12
Use a transaction's LastValid to expedite duplicate detection.

derbear:
This is an alternative to #504.

### DIFF
--- a/ledger/txtail.go
+++ b/ledger/txtail.go
@@ -32,8 +32,13 @@ type roundTxMembers struct {
 }
 
 type txTail struct {
-	recent    map[basics.Round]roundTxMembers
+	recent map[basics.Round]roundTxMembers
+
 	lastValid map[basics.Round]map[transactions.Txid]struct{} // map tx.LastValid -> tx confirmed set
+
+	// duplicate detection queries with LastValid not before
+	// lowWaterMark are not guaranteed to succeed
+	lowWaterMark basics.Round // the last round known to be committed to disk
 }
 
 func (t *txTail) loadFromDisk(l ledgerForTracker) error {
@@ -44,17 +49,14 @@ func (t *txTail) loadFromDisk(l ledgerForTracker) error {
 	}
 	proto := config.Consensus[hdr.CurrentProtocol]
 
-	t.lastValid = make(map[basics.Round]map[transactions.Txid]struct{})
-	for i := uint64(0); i <= proto.MaxTxnLife+1; i++ {
-		lv := latest + basics.Round(i)
-		t.lastValid[lv] = make(map[transactions.Txid]struct{})
-	}
-
 	// If the latest round is R, then any transactions from blocks strictly older than
 	// R + 1 - proto.MaxTxnLife
 	// could not be valid in the next round (R+1), and so are irrelevant.
 	// Thus we load the txids from blocks R+1-maxTxnLife to R, inclusive
 	old := (latest + 1).SubSaturate(basics.Round(proto.MaxTxnLife))
+
+	t.lowWaterMark = latest
+	t.lastValid = make(map[basics.Round]map[transactions.Txid]struct{})
 
 	t.recent = make(map[basics.Round]roundTxMembers)
 	for ; old <= latest; old++ {
@@ -76,10 +78,8 @@ func (t *txTail) loadFromDisk(l ledgerForTracker) error {
 		for _, txad := range payset {
 			tx := txad.SignedTxn
 			t.recent[old].txids[tx.ID()] = tx.Txn.LastValid
-			if _, ok := t.lastValid[tx.Txn.LastValid]; ok {
-				t.lastValid[tx.Txn.LastValid][tx.ID()] = struct{}{}
-			}
 			t.recent[old].txleases[txlease{sender: tx.Txn.Sender, lease: tx.Txn.Lease}] = tx.Txn.LastValid
+			t.putLV(tx.Txn.LastValid, tx.ID())
 		}
 	}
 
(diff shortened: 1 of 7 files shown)


ledger/txtail.go
```diff
@@ -43,6 +44,12 @@ func (t *txTail) loadFromDisk(l ledgerForTracker) error {
 	}
 	proto := config.Consensus[hdr.CurrentProtocol]
 
+	t.lastValid = make(map[basics.Round]map[transactions.Txid]struct{})
+	for i := uint64(0); i <= proto.MaxTxnLife+1; i++ {
```
zeldovich: Why `+1` here and elsewhere in `txtail.go`?  If the latest round is `latest`, then transactions in block `latest` can have a `LastValid` of at most `latest+MaxTxnLife`.

ledger/txtail.go
```diff
@@ -101,32 +119,31 @@ func (t *txTail) committedUpTo(rnd basics.Round) basics.Round {
 			delete(t.recent, r)
 		}
 	}
+	for lv := range t.lastValid {
+		if lv < rnd {
+			delete(t.lastValid, lv)
+		}
+	}
```
tsachiherman: nit: you could avoid this loop by storing the lowest value in the lastValid map keys. If the difference is one, you're guaranteed that there is only a single item that need to be deleted.

ledger/txtail.go
derbear: Once we've confirmed the `latest` round, we will start calling `isDup` on transactions whose `FirstValid` can be the next round, `latest+1`, so these transactions can have a `LastValid` of `latest+1+MaxTxnLife`. This initialization is done to return the correct value in `isDup` (and not return an error due to a missing `LastValid` in the map).

I guess initializing the map isn't strictly necessary here, since Go maps return `false` for membership checks if the map is `nil`. If I use Tsachi's suggestion below of keeping the last written round in the tracker, then I can probably remove this extra initialization.

ledger/txtail.go
derbear: That's a good point, and keeping the last value committed to the ledger may be useful for other things as well.

ledger/txtail.go
tsachiherman: I think that the map initialization here is a good move. It ensures that on `newBlock` you will have exactly one new entry. ( it's relatively fast, predictable and have constant time ).
Note that loading from disk happens only on initialization, so performance in this function isn't an issue.

ledger/txtail.go
```diff
@@ -92,6 +102,14 @@ func (t *txTail) newBlock(blk bookkeeping.Block, delta StateDelta) {
 		txleases: delta.txleases,
 		proto:    config.Consensus[blk.CurrentProtocol],
 	}
+
+	proto := config.Consensus[blk.CurrentProtocol]
+	if _, ok := t.lastValid[rnd+basics.Round(proto.MaxTxnLife)+1]; !ok {
+		t.lastValid[rnd+basics.Round(proto.MaxTxnLife)+1] = make(map[transactions.Txid]struct{})
+	}
```
tsachiherman: That won't work correctly if proto.MaxTxnLife goes up. i.e. 1000->1500
in that case, you'll need a for loop in here to add all the missing elements.
( very unlikely to happen, but plausible )

ledger/txtail.go
```diff
@@ -101,32 +119,31 @@ func (t *txTail) committedUpTo(rnd basics.Round) basics.Round {
...
+	for lv := range t.lastValid {
+		if lv < rnd {
+			delete(t.lastValid, lv)
+		}
+	}
 
 	return (rnd + 1).SubSaturate(maxlife)
 }
 
 func (t *txTail) isDup(proto config.ConsensusParams, current basics.Round, firstValid basics.Round, lastValid basics.Round, txid transactions.Txid, txl txlease) (bool, error) {
```
justicz: I think this optimization makes sense, but I think that this function signature is a bit dangerous now. This will fail to find a duplicate `txid` if the `lastValid` argument isn't exactly equal to the `LastValid` of the transaction. I think that `isDup` should instead take in a proto and a `transactions.Transaction` to avoid this situation.

justicz: I think it would also be ok to just change the argument names to `txFirstValid` and `txLastValid`.

ledger/txtail.go
```diff
@@ -92,6 +102,14 @@ func (t *txTail) newBlock(blk bookkeeping.Block, delta StateDelta) {
 		txleases: delta.txleases,
 		proto:    config.Consensus[blk.CurrentProtocol],
 	}
+
+	proto := config.Consensus[blk.CurrentProtocol]
+	if _, ok := t.lastValid[rnd+basics.Round(proto.MaxTxnLife)+1]; !ok {
```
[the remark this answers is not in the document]
tsachiherman: no, it could happen if MaxTxnLife goes down on a new protocol version.

ledger/txtail.go
derbear: Good idea--I think I'll just have the entire transaction passed in.

derbear:
New plan is to use a low- and a high-watermark to clean up the code and make things a bit faster.

ledger/txtail.go
pzbitskiy: >  so these transactions can have a LastValid of latest+1+MaxTxnLife.

goal did that when lastValid == 0. If firstVaild == 0 then we had validity window [lastRound+1, lastRound+1 + MaxTxnLife] that is MaxTxnLife + 1. I fixed that in goal.

ledger/txtail.go
derbear: I can fix the interface in a separate commit--will leave it as-is for now.

PR #509, opened 2019-11-13
Reduce lease test flakiness

justicz:
The new lease test is flaky. I removed a check that is inherently racy with block confirmation, and increased the window for another check to reduce its flakiness.

### DIFF
--- a/test/e2e-go/features/transactions/lease_test.go
+++ b/test/e2e-go/features/transactions/lease_test.go
@@ -194,12 +194,15 @@ func TestOverlappingLeases(t *testing.T) {
 	// transaction isn't valid until the first transaction's lease
 	// has expired.
 
+	const firstTxLeaseLife = 20
+	const secondTxLeaseLife = 100
+
 	// construct transactions for sending money to account1 and account2
 	// from same sender with identical lease, but different, overlapping ranges
-	tx1, err := client.ConstructPayment(account0, account1, 0, 1000000, nil, "", lease, basics.Round(leaseStart), basics.Round(leaseStart+20))
+	tx1, err := client.ConstructPayment(account0, account1, 0, 1000000, nil, "", lease, basics.Round(leaseStart), basics.Round(leaseStart+firstTxLeaseLife))
 	a.NoError(err)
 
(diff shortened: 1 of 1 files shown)


test/e2e-go/features/transactions/lease_test.go
```diff
@@ -196,7 +196,7 @@ func TestOverlappingLeases(t *testing.T) {
 
 	// construct transactions for sending money to account1 and account2
 	// from same sender with identical lease, but different, overlapping ranges
```
algonautshant: Overlapping ranges yield several different scenarios. However, this test is not testing all of these generated scenarios. It is testing one, and when sporadically another one of the scenarios plays out, the test is failing. 

Changing the values in this commit, merely favors the tested scenario, but keeps the other scenarios less likely to play out, and more importantly, untested. 

To make sure the overlapping ranges is fully tested, we must enumerate all the possible scenarios (in the comments),  and show which test item is covering each of these possible scenarios.

test/e2e-go/features/transactions/lease_test.go
```diff
@@ -235,17 +235,9 @@ func TestOverlappingLeases(t *testing.T) {
...
-
-	// submitting the second transaction should still fail
-	_, err = client.BroadcastTransaction(stx2)
-	a.Error(err)
-
-	// wait for us to be building leaseStart + 11, where the first txn's
+	// wait for us to be building leaseStart + 21, where the first txn's
 	// lease should have expired
-	fixture.WaitForRoundWithTimeout(leaseStart + 10)
+	fixture.WaitForRoundWithTimeout(leaseStart + 20)
```
algonautshant: The value '20' here is chosen based on the '20' set earlier. 
It will be nicer to have a variable to hold this value, and use is in both location. 
Importantly, the comments and the name of the variable can be descriptive enough to understand what and why is the value 20 for.

test/e2e-go/features/transactions/lease_test.go
justicz: This is true, and we can add more tests, but the purpose of this PR is just to reduce the flakiness of the test we have now to stop breaking builds. We can add more tests in a separate PR

test/e2e-go/features/transactions/lease_test.go
justicz: made this a const

PR #511, opened 2019-11-13
Limit the size of the pending transaction groups on the transaction pool

tsachiherman:
## Summary

Limit the size of the pending transaction groups on the transaction pool.
We've removed this limit during the evaluator integration, which could lead to slower worst-case scenario. Setting this number to 15K seems to cap the worst case scenario we've been seeing.

# Metrics

Setting this cap to 15K and running pingpong at full blast using regular payset transactions resulted in full blocks and block rate of 4.85

### DIFF
--- a/config/local_defaults.go
+++ b/config/local_defaults.go
@@ -83,7 +83,7 @@ var defaultLocalV4 = Local{
 	SuggestedFeeBlockHistory:              3,
 	SuggestedFeeSlidingWindowSize:         50,
 	TxPoolExponentialIncreaseFactor:       2,
-	TxPoolSize:                            50000,
+	TxPoolSize:                            15000,
 	TxSyncIntervalSeconds:                 60,
 	TxSyncTimeoutSeconds:                  30,
 	TxSyncServeResponseSize:               1000000,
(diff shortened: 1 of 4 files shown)


config/local_defaults.go
```diff
@@ -83,7 +83,7 @@ var defaultLocalV4 = Local{
 	SuggestedFeeBlockHistory:              3,
 	SuggestedFeeSlidingWindowSize:         50,
 	TxPoolExponentialIncreaseFactor:       2,
-	TxPoolSize:                            50000,
+	TxPoolSize:                            15000,
```
derbear:
```
// Changing a default value requires creating a new defaultLocalV# instance,
// bump the version number (configVersion), and add appropriate migration and tests.
```

Should we do that before merging this into `master`?

data/pools/transactionPool.go
```diff
@@ -172,6 +174,11 @@ func (pool *TransactionPool) Test(txgroup []transactions.SignedTxn) error {
 		txgroup[i].InitCaches()
 	}
 
+	pendingSize := len(pool.Pending())
```
derbear: Nit: Do we want to move this into `test`? That way `Remember` also picks up this behavior.

config/local_defaults.go
tsachiherman: I'll fix this tonight is a separate PR. I don't want to delay the testing effort.

zeldovich:
We removed the pool size limit altogether when we changed to the exponential fee growth plan.  The rationale for the current plan is that, if there are users willing to pay a high fee for a transaction, we should take their transaction.  The exponential growth of the required fee prevents users from growing the pool too large.

What's the rationale for having a fixed limit to the pool size?

Separately, the proposed number in this PR (15K) is only 3 blocks worth, which seems rather low.  But, my main comment is still that I don't see a good reason to have a fixed cap.

data/pools/transactionPool.go
tsachiherman: I'm not strongly opinionated around that. If someone places too much load on his own node, that's his issue.
I'll PR that tonight and we can discuss this issue separately ( once it's not on the critical path )

PR #515, opened 2019-11-13
add missing CloseRemainderTo==0 check

algobolson:


derbear:
Is this check also missing on `limit-order-b`?

algobolson:
the equivalent check in b was already there

PR #519, opened 2019-11-14
Fix failed teal test on arm64

tsachiherman:
## Summary

teal testing was failing on arm64; had nothing to do with arm64. The GOPATH had to be configured.

### DIFF
--- a/test/scripts/e2e_teal.sh
+++ b/test/scripts/e2e_teal.sh
@@ -4,6 +4,7 @@ date '+e2e_teal start %Y%m%d_%H%M%S'
 
 set -e
 set -x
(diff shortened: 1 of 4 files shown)


tsachiherman:
> this is fine but we could also move it up a level to e2e.sh that calls these scripts, then it won't be forgotten when we add the next sub-script

- The fact that we re-declare this in multiple location is a bug; but I don't want to fix that bug for this release, which is why I'm attempting to fix it "locally".
- After release, we will spend some time fixing all our automated testing systems. Hopefully, this would force us to detect these types of errors before they are merged in.

PR #521, opened 2019-11-14
Clean up txHandler and transactionPool.

derbear:


data/pools/transactionPool.go
```diff
@@ -295,15 +304,15 @@ func (pool *TransactionPool) Remember(txgroup []transactions.SignedTxn) error {
 	pool.mu.Lock()
 	defer pool.mu.Unlock()
 
+	if pool.pendingBlockEvaluator == nil {
+		return fmt.Errorf("TransactionPool.Remember: no pending block evaluator")
+	}
+
```
tsachiherman: this wouldn't hurt, but it's a duplicate. it's the first thing being done in pool.test

data/txHandler.go
```diff
@@ -166,38 +166,44 @@ func (handler *TxHandler) backlogWorker() {
...
-				handler.net.Disconnect(wi.rawmsg.Sender)
-				continue
-			}
-
-			// we've processed this message, so increase the counter.
-			transactionMessagesHandled.Inc(nil)
-
-			// at this point, we've verified the transaction, so we can safely treat the transaction as a verified transaction.
-			verifiedTxGroup := wi.unverifiedTxGroup
+			handler.postprocessCheckedTxn(wi)
```
tsachiherman: I think that you meant to replace the upper switch statement with this call as well, right ? (postprocessCheckedTxn )

data/pools/transactionPool.go
derbear: I just moved a block of code a few lines. Maybe we should delete it.

data/txHandler.go
derbear: Ah, yeah, there was some git issue.

derbear: FYI There was also a metric that wasn't being logged correctly before. This should fix that.

PR #522, opened 2019-11-14
Add atomic swap e2e teal test

justicz:
Test that the atomic swap teal script releases funds when given the correct preimage


test/scripts/atomic-swap-teal-test.sh
```diff
@@ -0,0 +1,66 @@
+#!/bin/bash
+
+date '+atomic-swap-teal-test start %Y%m%d_%H%M%S'
+
+set -e
+set -x
+
+TEMPDIR=$(mktemp -d)
```
algobolson: see Tsachi's recent change to the similar scripts about setting GOPATH near the top of these scripts, and yes, we'll clean it up properly post release

justicz: done, thanks

test/scripts/e2e.sh
```diff
@@ -66,6 +66,8 @@ cd "${SCRIPT_PATH}"
 
 ./timeout 200 ./teal-split-test.sh
 
+./timeout 200 ./atomic-swap-teal-test.sh
+
 # TODO: use ./timeout script across these three scripts when it gets committed from another branch
 ./periodic-teal-test.sh
```
derbear: (While you're here, can you use `timeout` on the other scripts as well? Thanks!)

PR #524, opened 2019-11-14
[DRAFT DO NOT MERGE] Skip redundant crypto operations.

derbear:
Don't merge this! I'd like to do this with a cleaner interface.

(I briefly tried using different unverified and verified txn types but stopped
because that change would be too large.)


winder:
Rather than having a skipCrypto flag you could add `VerifiedTransaction` and `VerifiedTransactionGroup`, that would at least get rid of the true/false and be descriptive about why crypto is being skipped.

Or maybe you can add a verified flag directly to `transactions.SignedTxn` / `transactions.SignedTxnWithAD` and use that to skip crypto.

PR #525, opened 2019-11-14
Fix checks preceding cryptographic verification in txHandler.

derbear:
This change reduces the minimum number of calls to pool.pendingBlockEvaluator
from 2 to 1 by skipping an extra call to TestTransactionGroup when trying to
remember a new transaction.

TransactionPool's Test() is updated to perform basic well-formedness and
duplicate detection checks but not to perform other stateful checks.

### DIFF
--- a/data/pools/transactionPool.go
+++ b/data/pools/transactionPool.go
@@ -304,10 +304,6 @@ func (pool *TransactionPool) Remember(txgroup []transactions.SignedTxn) error {
 	pool.mu.Lock()
 	defer pool.mu.Unlock()
 
-	if pool.pendingBlockEvaluator == nil {
-		return fmt.Errorf("TransactionPool.Remember: no pending block evaluator")
-	}
-
 	err := pool.test(txgroup)
 	if err != nil {
 		return fmt.Errorf("TransactionPool.Remember: %v", err)
--- a/data/pools/transactionPool.go
+++ b/data/pools/transactionPool.go
@@ -237,17 +237,17 @@ func (pool *TransactionPool) Test(txgroup []transactions.SignedTxn) error {
 }
 
 type poolIngestParams struct {
-	remember bool // if set, remember txn
-	checkFee bool // if set, perform fee checks
-	mustSync bool // if set, wait until ledger is caught up
+	remember   bool // if set, remember txn
+	checkFee   bool // if set, perform fee checks
+	preferSync bool // if set, wait until ledger is caught up
 }
 
 // remember attempts to add a transaction group to the pool.
 func (pool *TransactionPool) remember(txgroup []transactions.SignedTxn) error {
 	params := poolIngestParams{
-		remember: true,
-		checkFee: true,
-		mustSync: true,
+		remember:   true,
+		checkFee:   true,
+		preferSync: true,
 	}
 	return pool.ingest(txgroup, params)
 }
@@ -256,9 +256,9 @@ func (pool *TransactionPool) remember(txgroup []transactions.SignedTxn) error {
 // priority checks.
 func (pool *TransactionPool) add(txgroup []transactions.SignedTxn) error {
 	params := poolIngestParams{
-		remember: true,
-		checkFee: false,
-		mustSync: false,
+		remember:   true,
+		checkFee:   false,
+		preferSync: false,
 	}
 	return pool.ingest(txgroup, params)
 }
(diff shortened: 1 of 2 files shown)


data/pools/transactionPool.go
```diff
@@ -295,35 +233,137 @@ func (pool *TransactionPool) Remember(txgroup []transactions.SignedTxn) error {
...
+
+// ingest checks whether a transaction group could be remembered in the pool,
+// and stores this transaction in the pool if remember is set.
+//
+// ingest assumes that pool.mu is locked.  It might release the lock
+// while it waits for OnNewBlock() to be called.
+func (pool *TransactionPool) ingest(txgroup []transactions.SignedTxn, params poolIngestParams) error {
 	if pool.pendingBlockEvaluator == nil {
-		return fmt.Errorf("TransactionPool.Remember: no pending block evaluator")
+		return fmt.Errorf("TransactionPool.test: no pending block evaluator")
```
zeldovich: Typo: `test` should be `ingest`

tsachiherman: change to `.ingest`

data/pools/transactionPool.go
```diff
@@ -295,35 +233,137 @@ func (pool *TransactionPool) Remember(txgroup []transactions.SignedTxn) error {
...
 	}
 
-	err = pool.remember(txgroup)
-	if err != nil {
-		return fmt.Errorf("TransactionPool.Remember: %v", err)
+	if params.mustSync {
```
zeldovich: Since this code doesn't actually ensure sync (it just waits a bit), maybe it's better to rename this parameter to `preferSync`

data/pools/transactionPool.go
```diff
@@ -295,35 +233,137 @@ func (pool *TransactionPool) Remember(txgroup []transactions.SignedTxn) error {
...
+		if err != nil {
+			return err
+		}
+	}
+
+	if !params.remember {
+		tentativeRound := pool.pendingBlockEvaluator.Round() + pool.numPendingWholeBlocks
+		err := pool.pendingBlockEvaluator.TestTransactionGroup(txgroup)
+		if err == ledger.ErrNoSpace {
+			tentativeRound++
```
zeldovich: I think you're missing retry logic here

data/pools/transactionPool.go
tsachiherman: this test is being made with every addition of element ( i.e. pool.add )
that's abit excessive. We really need it only after each pool.mu.lock.

data/pools/transactionPool.go
derbear: It looks like the old code is also missing retry logic? https://github.com/algorand/go-algorand/pull/525/files/cb8efe1d421fc22aec7b2475b35b00ef28174bee#diff-d61c312f7cb0cc88e1721926edafe92bL220

zeldovich: Indeed, looks like you are correct.  Maybe it was correct before we added txgroups, but became incorrect after txgroups were added?  In any case, yeah, looks like the current test code could return `true` when the pending block evaluator runs out of space for a part of the group's transactions, even if later transactions in the group are invalid..

derbear: I think if we try to fix this by calling `ResetTxnBytes`, `test` will no longer be read-only and will possibly cause us to prematurely increment the next round to be checked by the pool's block evaluator.

derbear: Perhaps I'll keep the buggy behavior for now--I think in the worst case, we do extra crypto work on a transaction group. I guess the right behavior would be to save the old `blockTxBytes` in the evaluator, test, and then restore it after the call?

zeldovich: One possible fix would be to change the API for `TestTransactionGroup` to allow the caller to inflate the max block size limit for the duration of that `TestTransactionGroup` call.

derbear: Chatted with Max--the only reason we want `Test` is to do duplicate detection before doing any crypto. We don't need `Test` to be secure against an attacker since any attacker can always give us a bad signature and force us to do crypto. I think I'll modify `Test` to do much less--just check for duplicates and well-formedness.

zeldovich: Trimming down `Test` sounds like a great plan!

ledger/eval.go
```diff
@@ -359,6 +359,94 @@ func (eval *BlockEvaluator) ResetTxnBytes() {
...
+		}
+	}
+
+	return nil
+}
+
+// testTransaction performs basic duplicate detection and well-formedness checks
+// on a single transaction, but does not actually add the transaction to the block
+// evaluator, or modify the block evaluator state in any other visible way.
+func (eval *BlockEvaluator) testTransaction(txn transactions.SignedTxn, txgroup []transactions.SignedTxn, groupIndex int, cow *roundCowState) error {
```
tsachiherman: nit: you're not using the groupIndex.

derbear: Sounds good--I'll place that in a separate PR.

PR #527, opened 2019-11-15
Add random lease configuration to pingpong.

derbear:


shared/pingpong/pingpong.go
```diff
@@ -345,8 +345,14 @@ func constructTxn(from, to string, fee, amt, assetID uint64, client libgoal.Clie
 	copy(noteField, pingpongTag)
 	crypto.RandBytes(noteField[tagLen:])
 
+	// if random lease flag set, fill the lease field with random bytes
+	var lease [32]byte
+	if cfg.RandomLease {
+		crypto.RandBytes(lease[:])
```
tsachiherman: if we want to have contesting values, we should pick smaller random range.
If we want to have non-contesting values, we should ensure that by using a global variable..

derbear: Regarding the second thing, 32 bytes of cryptographic randomness will ensure we have no collisions, with cryptographic certainty. (`crypto.RandBytes` is thread-safe)

Regarding the first thing, that's a good idea. I was thinking that no collisions would produce worst-case performance, but it might be worth tuning so that occasionally we'll pick some colliding lease. Implementing this will probably require us to hold a bit of state per sender, or alternatively use PRNG seeds to trigger collisions.

PR #532, opened 2019-11-15
Fix race in asset_test between SuggestedParams and FillUnsignedTxTempate calls

pzbitskiy:
* Test uses SuggestedParams.LastRound for verification.
* FillUnsignedTxTemplate uses SuggestedParams.LastRound for calculation.
* If round change happens between these calls test fails.
* To prevent this test checks LastRound before and after FillUnsignedTxTemplate
  and adjusts verification

### DIFF
--- a/test/e2e-go/features/transactions/asset_test.go
+++ b/test/e2e-go/features/transactions/asset_test.go
@@ -138,20 +138,35 @@ func TestAssetValidRounds(t *testing.T) {
 	a.NoError(err)
 
 	tx, err := client.MakeUnsignedAssetCreateTx(100, false, manager, reserve, freeze, clawback, "test1", "testname1", "foo://bar", nil)
+	a.NoError(err)
 
 	fee := uint64(1000)
 	firstValid = 0
 	lastValid = 0
 
+	params, err = client.SuggestedParams()
+	a.NoError(err)
+	lastRoundBefore := params.LastRound
+
 	tx, err = client.FillUnsignedTxTemplate(account0, firstValid, lastValid, fee, tx)
 	a.NoError(err)
-
 	// zeros are special cases
-	// first valid must be set to last round + 1 and never be zero
-	// last valid must be set to first + MaxTxnLife - 1
+	// first valid never should be zero
 	a.NotEqual(basics.Round(0), tx.FirstValid)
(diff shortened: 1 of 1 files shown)


test/e2e-go/features/transactions/asset_test.go
```diff
@@ -143,15 +143,31 @@ func TestAssetValidRounds(t *testing.T) {
 	firstValid = 0
```
derbear: Not related to this PR, but did you mean to drop `err` above (line 140)?

test/e2e-go/features/transactions/asset_test.go
```diff
@@ -143,15 +143,31 @@ func TestAssetValidRounds(t *testing.T) {
...
-	a.Equal(basics.Round(params.LastRound+1), tx.FirstValid)
-	a.Equal(basics.Round(params.LastRound+cparams.MaxTxnLife+1), tx.LastValid)
+
+	params, err = client.SuggestedParams()
+	a.NoError(err)
+	lastRoundAfter := params.LastRound
+
+	lastRound := lastRoundBefore
+	if lastRoundBefore != lastRoundAfter {
+		// round changed, not sure when, ask +-1
```
derbear: This would be unlikely, but it might be possible that the ledger advanced _two_ rounds in between the calls, and thus neither would be correct. Maybe you could get rid of the if-else clauses and check that `tx.FirstValid` is between `lastRoundBefore` and `lastRoundAfter`, inclusive, and then check that `tx.LastValid` is correctly offset?

pzbitskiy: I thought about this and this seems to be more relaxed condition than the current one. I can change it if you think this is ok

test/e2e-go/features/transactions/asset_test.go
pzbitskiy: good catch I'll fix

pzbitskiy: good catch I'll fix

test/e2e-go/features/transactions/asset_test.go
pzbitskiy: done

PR #538, opened 2019-11-18
Include reason with peer disconnect log message.

winder:


network/wsNetwork.go
```diff
@@ -1609,7 +1609,7 @@ func (wn *WebsocketNetwork) removePeer(peer *wsPeer, reason disconnectReason) {
 	// first logging, then take the lock and do the actual accounting.
 	// definitely don't change this to do the logging while holding the lock.
 	localAddr, _ := wn.Address()
-	wn.log.With("event", "Disconnected").With("remote", peer.rootURL).With("local", localAddr).Infof("Peer %v disconnected", peer.rootURL)
+	wn.log.With("event", "Disconnected").With("remote", peer.rootURL).With("local", localAddr).Infof("Peer %v disconnected: %v", peer.rootURL, reason)
```
derbear: Nit: We can use `%s` since `disconnectReason` is a string.

PR #542, opened 2019-11-18
Tests for grouped asset transactions

pzbitskiy:
* Group of create + send zero in the same block
* Group of create + destroy in the same block
* Asset ID is guessed: tx count + 1 for both tests

### DIFF
--- a/test/e2e-go/features/transactions/asset_test.go
+++ b/test/e2e-go/features/transactions/asset_test.go
@@ -491,6 +491,147 @@ func TestAssetInformation(t *testing.T) {
 	a.NoError(err)
 }
 
+func TestAssetGroupCreateSendDestroy(t *testing.T) {
+	t.Parallel()
+	a := require.New(t)
+
+	var fixture fixtures.RestClientFixture
+	fixture.Setup(t, filepath.Join("nettemplates", "TwoNodes50Each.json"))
+	defer fixture.Shutdown()
+
+	client0 := fixture.GetLibGoalClientForNamedNode("Primary")
+	client1 := fixture.GetLibGoalClientForNamedNode("Node")
+
+	wh0, err := client0.GetUnencryptedWalletHandle()
+	a.NoError(err)
+	wh1, err := client1.GetUnencryptedWalletHandle()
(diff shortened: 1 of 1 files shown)


test/e2e-go/features/transactions/asset_test.go
```diff
@@ -491,6 +491,108 @@ func TestAssetInformation(t *testing.T) {
...
+		a.NoError(err)
+		txCreate, err = client0.FillUnsignedTxTemplate(account0, 0, 0, fee, txCreate)
+		a.NoError(err)
+
+		txSend, err := client1.MakeUnsignedAssetSendTx(txCount+1, 0, account1, "", "")
+		a.NoError(err)
+		txSend, err = client1.FillUnsignedTxTemplate(account1, 0, 0, fee, txSend)
+		a.NoError(err)
+
+		gid, err := client0.GroupID([]transactions.Transaction{txCreate, txSend})
```
tsachiherman: Make sure no error is returned here.

test/e2e-go/features/transactions/asset_test.go
```diff
@@ -491,6 +491,108 @@ func TestAssetInformation(t *testing.T) {
...
+		txSend.Group = gid
+		stxn, err = client1.SignTransactionWithWallet(wh1, nil, txSend)
+		a.NoError(err)
+		stxns = append(stxns, stxn)
+
+		// broadcasting group should succeed
+		err = client0.BroadcastTransactionGroup(stxns)
+		a.NoError(err)
+
+		txCount += uint64(len(stxns))
```
tsachiherman: You need to wait until the transaction group is confirmed an included in the block.
Just broadcasting the transaction doesn't mean it will be included in the block.

test/e2e-go/features/transactions/asset_test.go
pzbitskiy: done

test/e2e-go/features/transactions/asset_test.go
pzbitskiy: done and added more check after txns confirmation.

PR #544, opened 2019-11-18
Update CatchupParallelBlocks to reduce the server load

tsachiherman:
## Summary

Recent catchup test shown that we're hitting the rate limit too early.

Reducing the number of parallel catchup blocks seems to be an effective way to work around that.

### DIFF
--- a/config/local_defaults.go
+++ b/config/local_defaults.go
@@ -49,7 +49,7 @@ var defaultLocalV5 = Local{
 	PriorityPeers:                         map[string]bool{},
 	CadaverSizeTarget:                     1073741824,
 	CatchupFailurePeerRefreshRate:         10,
-	CatchupParallelBlocks:                 50,
+	CatchupParallelBlocks:                 16,
 	ConnectionsRateLimitingCount:          60,
 	ConnectionsRateLimitingWindowSeconds:  1,
 	DeadlockDetection:                     0,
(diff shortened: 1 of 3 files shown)


justicz:
I do worry a bit about slowing down catchup in the average case with this change -- I'd be inclined to up the rate limit on the servers instead if we can get away with it

justicz:
For context: we were having weird issues during catchup where a node would catch up a few hundred blocks, pause for ~15 seconds, catch up another few hundred blocks, and so on. Turns out we were hitting our rate limit and disconnecting.

tsachiherman:
Using these settings, I've measured catchup rate of 1K rounds / 16 sec. It sounds sufficiently high for me. Also, keep in mind that we don't want to rate to go too high, since it would deplete the relay outgoing bandwidth.

Changing the rate on the server can be a little bit more tricky, since it might have security implications.

justicz:
I just want to make sure we don't dramatically slow down users who aren't hitting the rate limit. What's the catchup rate with the current value when you don't hit the rate limit?

tsachiherman:
I have measured 1K rounds / 14 sec when using the existing configuration without seeing any disruptions.

justicz:
(But, I do think we should make a better fix for this after release by rate limiting requests on the client)

tsachiherman:
Agree.

PR #547, opened 2019-11-19
Fix debian package copyright.

winder:


algobolson:
I'm 99% certain that the "Copyright" on the debian package has to match the name on the gpg key. I'll look into what it would take to adjust the name on the gpg key.

PR #548, opened 2019-11-19
Return error when consensus protocol is unknown to libgoal.

derbear:


libgoal/participation.go
```diff
@@ -214,14 +214,20 @@ func (c *Client) InstallParticipationKeys(inputfile string) (part account.Partic
 		return
 	}
 
+	proto, ok := config.Consensus[protocol.ConsensusCurrentVersion]
+	if !ok {
+		err = fmt.Errorf("Unknown consensus protocol %s", protocol.ConsensusCurrentVersion)
+		return
+	}
+
```
tsachiherman: I'm not really sure how this would happen. once the binary is already compiled, the config.Consensus should already have the entry for protocol.ConsensusCurrentVersion.

derbear: Yeah. I thought about that as well. I put the check there anyway for consistency.

PR #549, opened 2019-11-19
Add e2e test for same sender, different lease.

derbear:


test/e2e-go/features/transactions/lease_test.go
```diff
@@ -85,6 +85,67 @@ func TestLeaseTransactionsSameSender(t *testing.T) {
 	a.Equal(bal2, uint64(0))
 }
 
+func TestLeaseTransactionsSameSenderDifferentLease(t *testing.T) {
+	t.Parallel()
+	a := require.New(t)
+
+	var fixture fixtures.RestClientFixture
+	fixture.Setup(t, filepath.Join("nettemplates", "TwoNodes50EachFuture.json"))
```
pzbitskiy: TwoNodes50EachFuture.json -> TwoNodes50Each.json because we've upgdaded

derbear: Per discussion on Slack, I think we should have two versions of these tests: one for future, and one for current. But I'll punt on that for now.

PR #554, opened 2019-11-19
Replace SignedTxnWithAD with SignedTxn in logic.EvalParams.

derbear:


algobolson:
I don't understand why. ledger/eval.go can pass what it has without copying if we pass -WithAD; that should be faster. This adds a copy which will make it slower.

algobolson:
I'm guessing this was a precursor to #556 and it makes sense in that context but not as itself here.

derbear:
Duplicate of #556.

PR #556, opened 2019-11-19
Bugfix: Prevent protocol stall from inconsistent LogicSig validity.

derbear:
Suppose an attacker submits a signed transaction T with a valid LogicSig at some
point in time.  Later, before T is confirmed, the attacker proposes a block with
some signed transaction T' which has the same transaction ID as T but has a
LogicSig that fails to verify.  Then this block has the potential to stall the
network if some validators have seen T and other validators have not; the
validators who have seen T will incorrectly sign off on the proposal because of
an invalid transaction pool-cache read, while the validators who have not seen T
will reject this block.  This subjectivity can call the protocol to stall
permanently, since validators will continue to cache blocks that have previously
been validated by the consensus protocol.

This change fixes this problem by adding a verification context to stateless
LogicSig (and other) checks.  The transaction pool's cache now stores the
verification context which a valid signed transaction was checked with respect
to.  When the cache is read to determine transaction validity, the stored
verification context is checked against that of alleged already-validated
transaction.


data/transactions/logicsig.go
```diff
@@ -56,6 +56,7 @@ func (lsig *LogicSig) Len() int {
 
 // Equal returns true if both LogicSig are equivalent
 func (lsig *LogicSig) Equal(b *LogicSig) bool {
+	// TODO is this sufficient?
 	if len(lsig.Logic) == 0 && len(b.Logic) == 0 {
 		return true
```
algoradam: Two LogicSigs could both have 0 length Logic but have different Sig, Msig, or Args. Not sure if this causes problems or not.

justicz: we should save ourselves a future headache and check that all args are exactly the same (including nilness)

data/pools/transactionPool.go
```diff
@@ -50,17 +50,19 @@ type TransactionPool struct {
...
 	// changes in OnNewBlock() without preventing a concurrent call
 	// to Pending() or Verified().
-	rememberedTxGroups [][]transactions.SignedTxn
-	rememberedTxids    map[transactions.Txid]transactions.SignedTxn
+	rememberedTxGroups     [][]transactions.SignedTxn
+	rememberedVerifyParams [][]verify.Params
+	rememberedTxids        map[transactions.Txid]txPoolVerifyCacheVal
 
 	// result of logic.Eval()
 	lsigCache *lsigEvalCache
```
algoradam: I think lsigEvalCache is unused now and should be removed

tsachiherman:
The `Verified` is performing single SignedTxn verification test. however, I haven't seen that it's ensuring that the rest of the transactions in the same transaction group are the same. if this is not the case, it might have the same results as the args, wouldn't it ?

derbear:
@tsachiherman I'm not sure if I fully understand your question, but the `GroupHash` will cryptographically commit to the other transactions in the group. `GroupHash` does get checked for consistency with the other transactions before the stateless checks: `TestTransactionGroup`in the tx handler, and inside of `eval.transaction` in the block evaluator (though notably not inside of `BroadcastTxGroup` inside of `node`, but that doesn't fall under the threat model).

data/transactions/verify/txn.go
```diff
@@ -182,20 +226,49 @@ func LogicSig(lsig *transactions.LogicSig, proto *config.ConsensusParams, stxn *
...
+		Txn:                 txn,
+		Proto:               &proto,
+		TxnGroup:            ctx.Group,
+		GroupIndex:          ctx.GroupIndex,
+		FirstValidTimeStamp: ctx.FirstValidTime,
+	}
+	pass, err := logic.Eval(txn.Lsig.Logic, ep)
+	if err != nil {
+		logicErrTotal.Inc(nil)
+		return fmt.Errorf("transaction %v: rejected by logic err=%s", txn.ID(), err)
```
tsachiherman: nit: if `err` is of `error` interface, we should use `%v`

data/txHandler.go
```diff
@@ -175,9 +189,13 @@ func (handler *TxHandler) postprocessCheckedTxn(wi *txBacklogMsg) {
 
 	// at this point, we've verified the transaction, so we can safely treat the transaction as a verified transaction.
 	verifiedTxGroup := wi.unverifiedTxGroup
+	verifyParams := make([]verify.Params, len(wi.verifyContexts))
+	for i := range verifyParams {
+		verifyParams[i] = wi.verifyContexts[i].Params
+	}
```
tsachiherman: I'm a bit puzzled around this one. It looks like you're pulling the `Params` variable from the `verifyContexts` array, but it doesn't looks like you ever initialized it in `prepareVerifyContexts`. Is it being initialized elsewhere, or did you intended to pass on the empty structure.

data/txHandler.go
```diff
@@ -263,19 +281,41 @@ func (handler *TxHandler) checkAlreadyCommitted(tx *txBacklogMsg) (processingDon
...
+			}
+			fvTime = uint64(hdr.TimeStamp)
+		}
 
-	return false
+		var ctx verify.Context
+		ctx.CurrSpecAddrs.FeeSink = latestHdr.FeeSink
+		ctx.CurrSpecAddrs.RewardsPool = latestHdr.RewardsPool
+		ctx.CurrProto = latestHdr.CurrentProtocol
+		ctx.FirstValidTime = fvTime
```
tsachiherman: nit: Initializing the Params member like this is very misleading. ( while being semantically correct )
```
ctx := verify.Context{
  params : Params {
      CurrSpecAddrs = ...,
  },
  FirstValidTime: fvTime,
}
```
is much more readable ( at least for me )

data/pools/transactionPool.go
```diff
@@ -346,10 +357,11 @@ func (pool *TransactionPool) Lookup(txid transactions.Txid) (tx transactions.Sig
 	pool.pendingMu.RLock()
 	defer pool.pendingMu.RUnlock()
 
-	tx, inPool := pool.pendingTxids[txid]
+	cacheval, inPool := pool.pendingTxids[txid]
 	if inPool {
 		return tx, "", true
 	}
+	tx = cacheval.txn
```
zeldovich: I think this needs to go inside the if check above.

data/txHandler.go
```diff
@@ -147,6 +146,21 @@ func (handler *TxHandler) backlogWorker() {
...
 				continue
 			}
+
+			// build the transaction verification context
+			latest := handler.ledger.Latest()
+			latestHdr, err := handler.ledger.BlockHdr(latest)
+			if err != nil {
+				logging.Base().Warnf("Could not get header for previous block %v: %v", latest, err)
+				continue
+			}
```
zeldovich: This code is the same at both callers of `prepareVerifyContexts`, so perhaps it should go inside `prepareVerifyContexts`

node/node.go
```diff
@@ -395,21 +395,49 @@ func (node *AlgorandFullNode) BroadcastSignedTxGroup(txgroup []transactions.Sign
...
+
+		var ctx verify.Context
+		ctx.CurrSpecAddrs.FeeSink = b.FeeSink
+		ctx.CurrSpecAddrs.RewardsPool = b.RewardsPool
+		ctx.CurrProto = b.CurrentProtocol
+		ctx.FirstValidTime = fvTime
+		ctx.Group = txgroup
+		ctx.GroupIndex = i
+
+		contexts[i] = ctx
```
zeldovich: Is there some way to avoid dup of this code with `TxHandler.prepareVerifyContexts`?

data/txHandler.go
```diff
@@ -263,19 +281,41 @@ func (handler *TxHandler) checkAlreadyCommitted(tx *txBacklogMsg) (processingDon
...
-	if err != nil {
-		logging.Base().Warnf("Could not get header for previous block %v: %v", latest, err)
-		return true
-	}
-	tx.proto = config.Consensus[latestHdr.CurrentProtocol]
-	tx.spec.FeeSink = latestHdr.FeeSink
-	tx.spec.RewardsPool = latestHdr.RewardsPool
+// note: this modifies tx
+func (handler *TxHandler) prepareVerifyContexts(tx *txBacklogMsg, latestHdr bookkeeping.BlockHeader) error {
+	// TODO set verifyContexts correctly
```
zeldovich: What does this comment mean?

derbear: Ah, this comment is outdated--I forgot to remove it.

ledger/eval.go
```diff
@@ -572,38 +569,43 @@ func (eval *BlockEvaluator) transaction(txn transactions.SignedTxn, ad transacti
...
-			err = verify.TxnPool(&txn, spec, eval.proto, eval.verificationPool)
+		ctx := verify.Context{
+			Params: verify.Params{
+				CurrSpecAddrs:  spec,
+				CurrProto:      eval.block.CurrentProtocol,
+				FirstValidTime: fvTime,
+			},
+			Group:      txgroup,
+			GroupIndex: groupIndex,
+		}
```
zeldovich: Would be great to avoid this copy of `TxHandler.prepareVerifyContexts` too..

derbear:
One subtle issue I've noticed in this change is that it'll cause us to disconnect from other users if we aren't able to read off our ledger correctly in the `txHandler`. I'll go fix that (and factor out some code) in another commit.

data/transactions/logicsig.go
```diff
@@ -56,20 +56,31 @@ func (lsig *LogicSig) Len() int {
...
 	}
 	return true
 }
+
+func safeSliceCheck(a, b []byte) bool {
+	if a == nil {
+		return b == nil
+	}
+	if b == nil {
+		return a == nil
```
algobolson: we know that a is not nil, return false

data/transactions/logicsig.go
```diff
@@ -56,20 +56,31 @@ func (lsig *LogicSig) Len() int {
...
-	for i, a := range lsig.Args {
-		if !bytes.Equal(a, b.Args[i]) {
+	for i := range lsig.Args {
+		if !safeSliceCheck(lsig.Args[i], b.Args[i]) {
 			return false
 		}
 	}
 	return true
 }
+
```
algobolson: might be worth saying why we care that `nil != b''`

data/txHandler.go
derbear: I'll leave out fetching the latest header in the factored code, since the ledger evaluator does not fetch the latest header; instead, it uses the header of the block it is currently building/validating.

justicz:
I am still reviewing this change, and I don't want to vouch for its correctness/safety yet. If our goal is to release right away then I think we should come up with something more likely to be correct in the short term, perhaps at the expense of some performance (as Tsachi suggested).

justicz:
one observation (also true before this PR): `TestTransactionGroup` will not error on a group containing duplicate transactions

justicz:
For LogicSig transactions, we look up the block header for `FirstValid - 1`. Will the ability of this to succeed differ between archival nodes and non archival nodes if `FirstValid` is 1000 rounds ago? Will this cause subjectivity?

zeldovich:
Good point about firstvalid-1 look back.  Perhaps we should define a tracker and use it to get the time, and that tracker would keep firstvalid-1 in the ledger.

data/pools/transactionPool.go
```diff
@@ -360,34 +367,24 @@ func (pool *TransactionPool) Lookup(txid transactions.Txid) (tx transactions.Sig
...
+	cacheval, ok := pool.pendingTxids[txn.ID()]
 	if !ok {
 		return false
 	}
 
+	if cacheval.params != params {
+		return false
+	}
+	pendingSigTxn := cacheval.txn
 	return pendingSigTxn.Sig == txn.Sig && pendingSigTxn.Msig.Equal(txn.Msig) && pendingSigTxn.Lsig.Equal(&txn.Lsig)
```
justicz: Probably ok but others should probably think about it too:

We only compare the `verify.Params`, not the entire `Context`. This means that you could pass `Verified` a `verify.Params` from inside a `Context` for which the group index / group would cause a logic sig to now fail.

I think this is probably OK because we only add transactions via `transactionGroup`, which does a well formedness check on the group (so you shouldn't be able to change the other parts of `Context` without also changing the `txid`)

ledger/eval.go
```diff
@@ -572,38 +574,17 @@ func (eval *BlockEvaluator) transaction(txn transactions.SignedTxn, ad transacti
 			return fmt.Errorf("transaction %v: malformed: %v", txn.ID(), err)
 		}
 
-		// Properly signed?
-		if eval.txcache == nil || !eval.txcache.Verified(txn) {
-			err = verify.TxnPool(&txn, spec, eval.proto, eval.verificationPool)
+		if eval.txcache == nil || !eval.txcache.Verified(txn, ctx.Params) {
```
justicz: We need to be sure that `ctx.CurrProto` is equal to `eval.proto` above -- it might be nice if we constructed `ctx.Params` right here for clarity

justicz: Actually, if we're ditching `FirstValidTime`, we can eliminate `PrepareContexts` entirely since the `Params` should be the same for every txn

data/transactions/logicsig.go
```diff
@@ -54,22 +54,35 @@ func (lsig *LogicSig) Len() int {
...
 }
 
-// Equal returns true if both LogicSig are equivalent
+// Equal returns true if both LogicSig are equivalent.
+//
+// Out of paranoia, Equal distinguishes zero-length byte slices
+// from byte slice-typed nil values as they may have subtly
+// different behaviors within the evaluation of a LogicSig,
+// due to differences in msgpack encoding behavior.
 func (lsig *LogicSig) Equal(b *LogicSig) bool {
```
pzbitskiy: seems like a test that captures this behavior (programs are the same for escrow accounts) is needed

data/pools/transactionPool.go
```diff
@@ -50,21 +50,19 @@ type TransactionPool struct {
 	txPoolMaxSize          int
 
 	// pendingMu protects pendingTxGroups and pendingTxids
```
pzbitskiy: add pendingVerifyParams to comment here and below

protocol/consensus.go
```diff
@@ -146,3 +150,12 @@ const ConsensusTestShorterLookback = ConsensusVersion("test-shorter-lookback")
 func ConsensusTestFastUpgrade(proto ConsensusVersion) ConsensusVersion {
 	return "test-fast-upgrade-" + proto
 }
+
+// Error is used to indicate that an unsupported protocol has been detected.
+type Error ConsensusVersion
```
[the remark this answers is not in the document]
derbear: It's the other way around--the `Error` is of concrete type `ConsensusVersion`.

data/pools/transactionPool.go
derbear: I'll do this in a future change; good catch.

data/transactions/logicsig.go
```diff
@@ -54,7 +54,12 @@ func (lsig *LogicSig) Len() int {
...
```
algoradam: Why does Equal take pointers as arguments? This will panic if either lsig or b is nil. (I'm not sure whether this is a problem now, but it might cause problems later)

justicz:
Adam and I noticed that `Lsig.Blank` is pretty broken, it just checks if `len(lsig.Logic) == 0`. This means a transaction could have a non empty `lsig.Args` and `Signature` and still be valid

justicz:
(adam and I are just dumping notes here) We should be checking logic sig max args in `LogicSigSanityCheck`

PR #557, opened 2019-11-19
nil []byte means int, so convert nil []byte to empty []byte{}

algobolson:


data/transactions/logic/eval.go
```diff
@@ -982,7 +990,7 @@ func (cx *evalContext) txnFieldToStack(txn *transactions.Transaction, field uint
 	case LastValid:
 		sv.Uint = uint64(txn.LastValid)
 	case Note:
-		sv.Bytes = txn.Note
+		sv.Bytes = nilToEmpty(txn.Note)
```
justicz: this makes it a slightly larger change, but could we instead have a method `SetBytes` on `(sv *stackValue)` that internally calls `sv.Bytes = nilToEmpty(b)`? That way if we add new byte slice fields we can't fail to call `nilToEmpty`

justicz: (in place of doing `sv.Bytes =` everywhere in this function)

tsachiherman:
Would that be a problem is the serialized nil and []byte{} aren't the same, yet, they would produce the same result ?

algobolson:
@derbear the byte constant pool is safe because it's not msgpack. what it is can encode empty string but not nil.
@justicz I've written myself a private TODO about that for after the release
@tsachiherman the problem is specific to fetching []byte values (but not [32]byte) from structs unpacked from msgpack and pushing them onto the stack in logic/eval.go because the evaluator uses the presence of a non-nil []byte value in the stackValue{} struct to signal that the 'union' type is []byte and not uint64. Copying a nil []byte would cause a type-check failure, but `len` on a nil []byte should return 0. So, push an empty 0 length []byte as a type marker and valid empty value.

PR #560, opened 2019-11-20
Deprecate the FirstValidTime txn field inside LogicSig programs.

derbear:
The ledger currently guarantees storage of the last 1000 blocks.  Reads on
blocks with rounds older than this on non-archival nodes will intermittently
fail, which may potentially cause the consensus protocol to stall (by
introducing subjectivity into block evaluation).

As a temporary fix, this commit disables the FirstValidTime txn field within
LogicSig programs.

### DIFF
--- a/cmd/goal/clerk.go
+++ b/cmd/goal/clerk.go
@@ -132,7 +132,6 @@ func init() {
 	compileCmd.Flags().StringVarP(&account, "account", "a", "", "Account address to sign the program (If not specified, uses default account)")
 
 	dryrunCmd.Flags().StringVarP(&txFilename, "txfile", "t", "", "transaction or transaction-group to test")
-	// dryrunCmd.Flags().Int64VarP(&timeStamp, "time-stamp", "S", 0, "unix time value for txn FirstValidTime (default now)")
 	dryrunCmd.Flags().StringVarP(&protoVersion, "proto", "P", "", "consensus protocol version id string")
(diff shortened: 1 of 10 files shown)


cmd/goal/clerk.go
```diff
@@ -132,7 +132,7 @@ func init() {
 	compileCmd.Flags().StringVarP(&account, "account", "a", "", "Account address to sign the program (If not specified, uses default account)")
 
 	dryrunCmd.Flags().StringVarP(&txFilename, "txfile", "t", "", "transaction or transaction-group to test")
-	dryrunCmd.Flags().Int64VarP(&timeStamp, "time-stamp", "S", 0, "unix time value for txn FirstValidTime (default now)")
+	// dryrunCmd.Flags().Int64VarP(&timeStamp, "time-stamp", "S", 0, "unix time value for txn FirstValidTime (default now)")
```
justicz: imo we should just delete this and its friends instead of commenting them out. we can look back at this commit later to see what was removed

derbear:
To clarify, calling `ledger.BlockHdr` on blocks that are _1001_ rounds older than the current round, which can be triggered with a transaction with a lifetime of 1000 rounds that reads from FirstValidTime (which is one round before FirstValid), will fail intermittently if we are not on an archival node. This can cause inconsistency in how validators vote for blocks with these transactions.

PR #564, opened 2019-11-21
Switch back to xenial for wider glibc compatibility.

winder:


algobolson:
Ubuntu 16 was previously used because its libc is old enough that it serves as a least common denominator that produces binaries compatible with centos7 and ubuntu 16 and 18. If the rpm built natively on centos7 is enough to support that platform, we can keep ubuntu .deb and .tar.gz builds on ubuntu 18. Several tools are better on Ubuntu 18 and I'm in favor of making that the preferred platform and deprecating Ubuntu 16. (And probably a wide variety of platforms work if someone builds from source)

PR #565, opened 2019-11-21
Build fixes 20191121

algobolson:
Has been demonstrated to build something like v2.0.0 RPM and DEB packages on Linux x86_64


scripts/build_release.sh
```diff
@@ -132,18 +141,23 @@ allow-preset-passphrase
 default-cache-ttl 2592000
 max-cache-ttl 2592000
 EOF
-gpg --generate-key --batch ${HOME}/tkey/keygenscript
-gpg --export -a > "${HOME}/docker_test_resources/key.pub"
+gpg --gen-key --batch ${HOME}/tkey/keygenscript
+gpg --gen-key --batch ${HOME}/tkey/rpmkeygenscript
+gpg --export -a dev@algorand.com > "${HOME}/docker_test_resources/key.pub"
+gpg --export -a rpm@algorand.com > "${HOME}/docker_test_resources/rpm.pub"
```
derbear: Why two different keys?

scripts/build_release_centos_docker.sh
```diff
@@ -64,7 +60,10 @@ rm -f ${HOME}/.gnupg/S.gpg-agent
 (cd ~/.gnupg && ln -s /S.gpg-agent S.gpg-agent)
 
 gpg --import /stuff/key.pub
-gpg --import ${REPO_ROOT}/installer/rpm/RPM-GPG-KEY-Algorand
+gpg --import /stuff/rpm.pub
+gpg --import ${REPO_DIR}/installer/rpm/RPM-GPG-KEY-Algorand
+rpmkeys --import /stuff/rpm.pub
+echo "wat"|gpg -u rpm@algorand.com --clearsign
```
derbear: Out of curiosity, what is this command for?

algobolson: It's a quick check to make sure signing works at all, before doing some stuff that takes more time. it's a fail-fast check. It was useful for debugging and I don't think there's any reason to take it out. GPG is a little fragile, so leaving it in is probably good in case it catches things in the future.

scripts/build_release.sh
algobolson: dev@ has better security properties and we use it to sign a bunch of things .tar.gz and .deb
rpm is dumb and requires a simpler gpg key so it gets a simpler lesser key.

PR #569, opened 2019-11-22
Bump mainnet pregen to 1.0.

derbear:


tsachiherman:
I don't see any issue with the change, but I'm curious for the trigger.

derbear:
> I don't see any issue with the change, but I'm curious for the trigger.

Main reason for the change is informational (having the template for the mainnet genesis file match the actual genesis file inside of `installer`).

PR #572, opened 2019-12-02
Telemetry Tweaks

winder:
This event has generally caused more harm than good. Disable it for now.

### DIFF
--- a/catchup/service.go
+++ b/catchup/service.go
@@ -198,7 +198,14 @@ func (s *Service) fetchAndWrite(fetcher rpcs.Fetcher, r basics.Round, prevFetchC
 
 		// Check that the block's contents match the block header (necessary with an untrusted block because b.Hash() only hashes the header)
 		if !block.ContentsMatchHeader() {
-			s.log.Warnf("fetchAndWrite(%v): block contents do not match header (attempt %d)", r, i)
+			// Check if this mismatch is due to an unsupported protocol version
+			if _, ok := config.Consensus[block.BlockHeader.CurrentProtocol]; !ok {
+				if i == 10 {
+					s.log.Errorf("fetchAndWrite(%v): unsupported protocol version detected", )
+				}
+			} else {
+				s.log.Warnf("fetchAndWrite(%v): block contents do not match header (attempt %d)", r, i)
+			}
 			client.Close()
 			continue // retry the fetch
 		}
--- a/catchup/service.go
+++ b/catchup/service.go
@@ -168,6 +168,12 @@ func (s *Service) fetchAndWrite(fetcher rpcs.Fetcher, r basics.Round, prevFetchC
 		default:
 		}
 
+		// Stop retrying after a while.
+		if i > 20 {
+			s.log.Errorf("fetchAndWrite(%v): failed to fetch block many times", )
+			return false
+		}
+
 		// Try to fetch, timing out after retryInterval
 
 		block, cert, client, err := s.innerFetch(fetcher, r)
--- a/catchup/service.go
+++ b/catchup/service.go
@@ -36,6 +36,8 @@ import (
 )
 
 const catchupPeersForSync = 10
+// this should be at least the number of relays
+const catchupRetryLimit = 500
 
 // Ledger represents the interface of a block database which the
 // catchup server should interact with.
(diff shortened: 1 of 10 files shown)


catchup/service.go
```diff
@@ -198,7 +198,14 @@ func (s *Service) fetchAndWrite(fetcher rpcs.Fetcher, r basics.Round, prevFetchC
 
 		// Check that the block's contents match the block header (necessary with an untrusted block because b.Hash() only hashes the header)
 		if !block.ContentsMatchHeader() {
-			s.log.Warnf("fetchAndWrite(%v): block contents do not match header (attempt %d)", r, i)
+			// Check if this mismatch is due to an unsupported protocol version
+			if _, ok := config.Consensus[block.BlockHeader.CurrentProtocol]; !ok {
+				if i == 10 {
+					s.log.Errorf("fetchAndWrite(%v): unsupported protocol version detected", )
+				}
```
tsachiherman: If you're adding a special handling for mismatching protocol version - 
1. Given that we *agreed* upon this protocol upgrade, there is no reason to wait for the 10th attempt before reporting it. You can report it right away.
2. Once you're reported this case, call 'client.Close()' and return false. There is absolutely no reason to make another attempt for the naive case.
3. For the non-naive case where you've received this invalid response from a hostile relay, the existing logic would retry again anyway, but with some delay. That would protect us against draining network resources endlessly once we cannot move forward due to protocol mismatch.

tools/network/telemetryURIUpdateService.go
```diff
@@ -41,9 +41,9 @@ func lookupTelemetryEndpoint(cfg config.Local, genesisNetwork protocol.NetworkID
 	for _, bootstrapID := range bootstrapArray {
 		addrs, err := ReadFromSRV("telemetry", bootstrapID, cfg.FallbackDNSResolverAddress)
 		if err != nil {
-			log.Warnf("An issue occurred reading telemetry entry for: %s", bootstrapID)
+			log.Infof("An issue occurred reading telemetry entry for: %s", bootstrapID)
```
tsachiherman: nit:
1. when formatting input strings, make sure to place them in quotes so we could identify white space issues. i.e. '%s'
2. If there is untyped, non-nil error, you want to log the error string as well.

catchup/service.go
winder: great! thanks I'll make this change.

catchup/service.go
```diff
@@ -168,6 +168,12 @@ func (s *Service) fetchAndWrite(fetcher rpcs.Fetcher, r basics.Round, prevFetchC
 		default:
 		}
 
+		// Stop retrying after a while.
+		if i > 20 {
+			s.log.Errorf("fetchAndWrite(%v): failed to fetch block many times", )
+			return false
+		}
+
```
tsachiherman: I don't think that we want that, as it competes with the existing catchup logic.

catchup/service.go
```diff
@@ -198,6 +204,13 @@ func (s *Service) fetchAndWrite(fetcher rpcs.Fetcher, r basics.Round, prevFetchC
 
 		// Check that the block's contents match the block header (necessary with an untrusted block because b.Hash() only hashes the header)
 		if !block.ContentsMatchHeader() {
+			// Check if this mismatch is due to an unsupported protocol version
+			if _, ok := config.Consensus[block.BlockHeader.CurrentProtocol]; !ok {
+				s.log.Errorf("fetchAndWrite(%v): unsupported protocol version detected", )
```
tsachiherman: that won't compile; but beside that, we want to also log which protocol is not supported.

logging/telemetry.go
```diff
@@ -70,7 +70,7 @@ func makeLevels(min logrus.Level) []logrus.Level {
 }
 
 func makeTelemetryState(cfg TelemetryConfig, hookFactory hookFactory) (*telemetryState, error) {
-	history := createLogBuffer(cfg.LogHistoryDepth)
+	history := createLogBuffer(2)
```
tsachiherman: Could you make it a local constant for now ?

catchup/service.go
winder: Is there a number between 20 and 6 million that you think would be alright?
```
fetchAndWrite(3317176): block contents do not match header (attempt 6242741)
```

tsachiherman: yes; we want to have a number that is greater than the number of relays so that every relay is likely to be attempted before giving up.

PR #575, opened 2019-12-02
Add lease support to asset cmds

justicz:
Asset commands were missing lease support


cmd/goal/asset.go
```diff
@@ -532,6 +541,7 @@ var freezeAssetCmd = &cobra.Command{
 		}
 
 		tx.Note = parseNoteField(cmd)
+		tx.Lease = parseLease(cmd)
```
derbear: It's unfortunate that all this code needs to be duplicated...

derbear:
Also, are the existing templates for limit orders replay-protected? (Do they correctly specify a lease?) I think the old limit order templates were replay-protected because they didn't support partial orders, but the new ones do, and "Lease" doesn't show up anywhere in those templates...

justicz:
Since each limit order template includes a regular, ed25519-signed transaction in its group, I don't think it's replayable because that txn will be caught by the duplicate detection.

PR #576, opened 2019-12-02
Create a protocol upgrade for asset precision

tsachiherman:
## Summary

Create a protocol upgrade for asset precision.
< do not merge this change !! >

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -221,6 +221,9 @@ type ConsensusParams struct {
 
 	// sum of estimated op cost must be less than this
 	LogicSigMaxCost uint64
+
+	// a precision for assets
+	SupportAssetsPrecision bool
 }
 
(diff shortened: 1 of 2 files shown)


config/config.go
```diff
@@ -221,6 +221,9 @@ type ConsensusParams struct {
 
 	// sum of estimated op cost must be less than this
 	LogicSigMaxCost uint64
+
+	// a precision for assets
+	SupportAssetsPrecision bool
```
justicz: Instead of a `bool`, can we just call this `MaxAssetDecimals` and have it be a `uint8`/`uint16`/`uint32` whose value is 19 after the upgrade? We'll need to set a maximum value anyway, and that way there's only one new proto variable instead of two.

PR #577, opened 2019-12-02
Let dsign sign arbitrary bytes, not just txids

justicz:
dsign currently only works with signing txids and whole `LogicSig`s. this adds a mode that lets it sign arbitrary data given a key, the program bytes, and a data file.

it would be nice to make this a prettier program with cobra, I didn't do that because I wanted to get something working quickly for the hackathon

### DIFF
--- a/tools/teal/dkey/dsign/main.go
+++ b/tools/teal/dkey/dsign/main.go
@@ -21,6 +21,7 @@
 package main
 
 import (
+	"encoding/base64"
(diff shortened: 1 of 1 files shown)


derbear:
Out of curiosity, what would you like to sign with this program?

justicz:
I'm signing oracle statements for use with https://github.com/justicz/algorand-december-hackathon

PR #578, opened 2019-12-03
static codegen for msgpack encode/decode

zeldovich:
This PR sketches out static code generation for msgpack encoding and decoding of blocks and transactions.  A basic benchmark suggests this is about 4x faster than our current use of go-codec.

Having fast encoding/decoding is not only good for performance, but will allow us to remove complex and dubious optimizations (like caching txid values or encoding lengths), and will allow us to perform checks that we previously thought would be too expensive (like making sure that an encoding is canonical, by re-encoding).

Having explicitly generated code also makes it easier to understand performance and tweak it further; I expect we could get even higher performance by benchmarking and optimizing the static code.  I haven't done any optimizations on this at all, but now pprof should be much less opaque (no reflection) and more actionable.

The code generation is done using a modified version of [github.com/tinylib/msgp](https://github.com/tinylib/msgp), forked as [github.com/zeldovich/msgp](https://github.com/zeldovich/msgp).  We should move it into the algorand github project once this is closer to merging.

Remaining tasks here are to actually wire up some places in the code to call into the new encoders/decoders, and to write some randomized tests to ensure that go-codec and this new encoder agree, along the lines of what `cmd/msgpacktool/core_test.go` does.


justicz:
Nice! I have looked at msgp a little bit before (I [forked it](https://github.com/justicz/msgp/commits/master) last month to play around with). Two things I noticed back then:

1. It doesn't recover from panics. It could be that the way it's doing decoding is clearly correct enough that we don't have to worry about that, or perhaps panicking is even what we want to do... but it's something we should think about.

2. It uses `unsafe` a little bit (though it can be overridden with a build flag to use `msgp/purego.go` instead of `msgp/unsafe.go`). It would be worth ensuring that their use of `UnsafeString` and `UnsafeBytes` is actually safe.

I would also like to take a look and make sure it can't be tricked into allocating lots of memory on small inputs.

data/transactions/transactions_gen.go
```diff
@@ -0,0 +1,2652 @@
...
+	var zb0002 uint32
+	zb0002, bts, err = msgp.ReadArrayHeaderBytes(bts)
+	if err != nil {
+		err = msgp.WrapError(err)
+		return
+	}
+	if cap((*z)) >= int(zb0002) {
+		(*z) = (*z)[:zb0002]
+	} else {
+		(*z) = make(Payset, zb0002)
```
justicz: For example, it looks like you can potentially force it to allocate space for `2**32 - 1` `SignedTxnInBlock`s here with a short message? `ReadArrayHeaderBytes` is [here](https://github.com/zeldovich/msgp/blob/master/msgp/read_bytes.go#L220).

justicz: Also, I think this could panic on 32-bit systems when `zb0002` is > `2**31 - 1`. Check out https://play.golang.org/p/cvjxsE7N4s- (I think the go playground site is 32-bit, system int size is 4 bytes).

derbear:
> It uses unsafe a little bit (though it can be overridden with a build flag to use msgp/purego.go instead of msgp/unsafe.go). It would be worth ensuring that their use of UnsafeString and UnsafeBytes is actually safe.

I wonder how much performance the library gets out of using `unsafe`.

zeldovich:
> > It uses unsafe a little bit (though it can be overridden with a build flag to use msgp/purego.go instead of msgp/unsafe.go). It would be worth ensuring that their use of UnsafeString and UnsafeBytes is actually safe.
> 
> I wonder how much performance the library gets out of using `unsafe`.

This new codegen plan gets about 10% slower if it uses the pure Go versions (as opposed to the unsafe versions) of string and byte array conversions.

data/transactions/transactions_gen.go
zeldovich: What were you thinking of for bounding the allocation?  One possibility I'm considering is to ensure that all maps and arrays must be no larger than the remaining number of bytes available to decode.  So, if we cap the size of the message (elsewhere in the code), the decoder blowup will be somewhat bounded over that max message size.  Tighter bounds are probably possible, though trickier to get right.

justicz: "no larger than the remaining number of bytes" sounds reasonable to me. Though I guess that could still give a blow up of approximately `(number of bytes being decoded) * (maximum struct size we decode slices of)`. Which could still be a lot... I'll check around the places we decode messages.

justicz: In `rpcs/httpTxSync.go`, we `protocol.Decode` a slice of `transactions.SignedTxn`. On my machine, `unsafe.Sizeof(transactions.SignedTxn{})` is 1056. If I'm reading it correctly, the maximum length body that gets decoded is 1,000,016 bytes. So you could allocate about 1GB of space per sync response there.

justicz: In catchup, we use the `WsFetcherService` to fetch blocks + certs, which has a 4MB response limit. `unsafe.Sizeof(transactions.SignedTxnInBlock{})` is 1096. So during catchup you can cause a ~4GB allocation. Since we run catchup goroutines in parallel, that could add up pretty quickly if the responses are timed together.

justicz:
Here is a tool to differentially fuzz the two msgpack implementations using [go-fuzz](https://github.com/dvyukov/go-fuzz): https://github.com/algorand/msgfuzz

zeldovich:
Thanks to @justicz for writing the msgfuzz tool that found a lot of behaviors of go-codec that we previously didn't know about or fully appreciate (but now we faithfully replicate them in our msgp code).

As a separate task, we can later consider whether our protocol is resilient to certain differences in decoding behavior, and if so, we can clean up some of these decoding behaviors that we added for compatibility.

I think this is ready to go now.

tsachiherman:
Two questions:
1. Does this change optimizes certain encoded objects, or does it cover all of them ?
2. Given that it's quite a large change, do we have any metrics that would demonstrate the gains here ?

derbear:
It looks like it applies to all structs that are encoded under msgpack, which includes transactions, blocks, votes, and certificates.

I think this change is nice for more than just performance. As long as performance doesn't _decrease_ under this change, we should merge it in because it simplifies encoding/decoding--no reflection is required any more. We should be able to catch bugs of the form: a datatype was updated (e.g., a vote), and we neglected to apply a protocol upgrade. We should be able to see by inspection if we forgot to add struct tags because the default tags will now show up in the code diff.

derbear:
I haven't looked at this in detail, but does `allocbound` cap slice size? If so, it also removes another source of DoS possibilities where somewhere we forget to bound the size of some encoded bytestring and are forced to waste a lot of time decoding.

derbear:
There are some things which have `allocbound=-`; can we disable this and safely cap those slices?

zeldovich:
> 1. Does this change optimizes certain encoded objects, or does it cover all of them ?

It covers objects defined in certain directories (`MSGP_GENERATE` in the Makefile).  Notably, it omits `agreement` because there's some additional complications there (the bounds on the number of elements in various structs isn't obviously known).  I think we should bring agreement into this static codegen plan as a separate later PR.

> 2. Given that it's quite a large change, do we have any metrics that would demonstrate the gains here ?

Yes, `BenchmarkBlockEncoding` in `data/encoding_test.go` is 4-5x faster.

zeldovich:
> I haven't looked at this in detail, but does `allocbound` cap slice size? If so, it also removes another source of DoS possibilities where somewhere we forget to bound the size of some encoded bytestring and are forced to waste a lot of time decoding.

Yes, that's the intent.

data/transactions/transaction.go
```diff
@@ -143,43 +136,22 @@ type TxGroup struct {
 	// together, sequentially, in a block in order for the group to be
 	// valid.  Each hash in the list is a hash of a transaction with
 	// the `Group` field omitted.
-	TxGroupHashes []crypto.Digest `codec:"txlist"`
+	TxGroupHashes []crypto.Digest `codec:"txlist,allocbound=config.Consensus[protocol.ConsensusCurrentVersion].MaxTxGroupSize"`
```
derbear: I wonder how error-prone this will be in event of a protocol upgrade.

zeldovich:
> There are some things which have `allocbound=-`; can we disable this and safely cap those slices?

I put that annotation in places where I thought the data structure would come from a trusted source, or DoS attacks didn't matter for some other reason.

We could cap them, but I don't think it really matters.

All of the serious unbounded allocations we had before are capped.

data/transactions/transaction.go
zeldovich: This is indeed one of the more subtle bounds.  I think this works out OK as long as we are increasing the limit upward.  If we decrease the limit at some point, I think we should keep the allocation bound high for compatibility, for at least one round of upgrades.

tsachiherman:
> > 1. Does this change optimizes certain encoded objects, or does it cover all of them ?
> 
> It covers objects defined in certain directories (`MSGP_GENERATE` in the Makefile). Notably, it omits `agreement` because there's some additional complications there (the bounds on the number of elements in various structs isn't obviously known). I think we should bring agreement into this static codegen plan as a separate later PR.
> 
> > 1. Given that it's quite a large change, do we have any metrics that would demonstrate the gains here ?
> 
> Yes, `BenchmarkBlockEncoding` in `data/encoding_test.go` is 4-5x faster.

It's perfectly fine to have some of the work done in a separate PR; I would hope to schedule all that work to be targeting the same *release* though.

Performance gains of 4-5x are really great.

derbear:
> (the bounds on the number of elements in various structs isn't obviously known)

For which structs is this true?

data/transactions/transaction.go
derbear: It does make things more awkward in the cases where we're having multiple networks (betanet vs. mainnet/testnet) updating through different versions prior to a software release. I guess there we'd manually set it to the maximum of all those values.

zeldovich:
> > (the bounds on the number of elements in various structs isn't obviously known)
> 
> For which structs is this true?

The `Votes` and `EquivocationVotes` in a bundle.  I suspect there's some bound (like 7750), but I didn't try to carefully verify that, and instead decided to punt on agreement encoding for now.

derbear:
> The Votes and EquivocationVotes in a bundle. I suspect there's some bound (like 7750), but I didn't try to carefully verify that, and instead decided to punt on agreement encoding for now.

It's the largest committee threshold in this list of config values: https://github.com/algorand/go-algorand/blob/master/config/config.go#L147

derbear:
Note that these were upgraded in the v8 consensus upgrade: https://github.com/algorand/go-algorand/blob/master/config/config.go#L312

If you can compute the maximum statically, I think you read those values off `config.ConsensusCurrentVersion`.

data/transactions/transaction.go
zeldovich: Thinking about it more, this allocbound is one of the ones that should have been `allocbound=-` because we never decode these structs -- we only encode them to compute the txgroup hash..

zeldovich:
> Note that these were upgraded in the v8 consensus upgrade: https://github.com/algorand/go-algorand/blob/master/config/config.go#L312
> 
> If you can compute the maximum statically, I think you read those values off `config.ConsensusCurrentVersion`.

I tried to add msgp support to agreement here:

   https://github.com/zeldovich/go-algorand/commit/1dc6af9e0bee0420aa4f55443b731ad2ce96e76c

I think it found a few missing annotations -- would be good to get your thoughts on whether those are real or not.

I also re-discovered why agreement code was somewhat tricky.  In some places, it encodes interfaces (like the action that's encoded in `persistence.go:encode()`).  It encodes some external types that don't have generated encoders/decoders (like `time.Duration` in `player`).  Probably other issues too, but these were the first things I ran into.

Anyway, definitely a separate PR for getting agreement code to use msgp.

justicz:
It would be cool if this checked for duplicate field names at generation time

protocol/codec.go
```diff
@@ -132,13 +155,49 @@ func EncodeJSON(obj interface{}) []byte {
...
+	}()
+
+	var rem []byte
+	rem, err = objptr.UnmarshalMsg(b)
+	if err != nil {
+		return err
+	}
+
+	// go-codec compat: allow remaining bytes, because go-codec allows it too
+	if false {
```
justicz: dead code

zeldovich:
> ~It would be cool if this checked for duplicate field names at generation time~ actually, looks like it gets caught later on at build time as a duplicate case in a switch statement.

Correct, duplicate field names turn into duplicate cases in the decoder switch, and `go vet` catches that.

protocol/codec.go
zeldovich: Do you think we should remove it?  I have hopes that we'll eventually grep for these `go-codec compat` places and make our decoder more strict, and this is one such case.

justicz: alright, fine with me. just thought this might have been accidental

data/transactions/aggregates.go
```diff
@@ -24,6 +24,7 @@ import (
 
 type (
 	// A Payset represents a common, unforgeable, consistent, ordered set of SignedTxn objects.
+	//msgp:allocbound Payset 10000
```
justicz: I think this is too small. I believe something of the following shape could be a valid `SignedTxnInBlock` weighing 93 bytes (Brian says the smallest valid TEAL program he could think of is 3 bytes), meaning you could fit more than 10k transactions in a valid block:

```
{
  "hgi": true,
  "lsig": {"l": "XXX"},
  "txn": {
    "fee": 1000,
    "fv": 1058301,
    "lv": 1059301,
    "snd:b64": "O1VjROADgX0+8QLV8Z8eANbSMC9xQVpkjNqZ7o7vcr0=",
    "type": "pay"
  }
}
```

zeldovich: I suspect you're correct.  I'll bump this to 100K, which should be a safe upper bound given the current block size, as well as still pretty good in terms of preventing huge memory allocations.

zeldovich:
We should decide what to do with this PR.  @justicz @tsachiherman 

I think we should merge this, because it's a big performance boost that we'll have to bite off eventually, and might as well do it now since we have the code for it.  Having this PR sitting unmerged will be annoying because it'll keep incurring irrelevant conflicts that have to be resolved.

I think we should also let this run for a while on testnet (a month, say?) before releasing on mainnet.  It's a pretty substantial change, and despite extensive testing on encoder/decoder equivalence, it'd be good to let it bake for some time.

tsachiherman:
> We should decide what to do with this PR. @justicz @tsachiherman
> 
> I think we should merge this, because it's a big performance boost that we'll have to bite off eventually, and might as well do it now since we have the code for it. Having this PR sitting unmerged will be annoying because it'll keep incurring irrelevant conflicts that have to be resolved.
> 
> I think we should also let this run for a while on testnet (a month, say?) before releasing on mainnet. It's a pretty substantial change, and despite extensive testing on encoder/decoder equivalence, it'd be good to let it bake for some time.

I don't think that we're ready for deploying different binaries on testnet vs mainnet. We can do that pretty easily on betanet.

zeldovich:
Ah, didn't know.  I don't care particularly about testing it on testnet vs betanet; I expect betanet would be just fine too.

tsachiherman:
> Ah, didn't know. I don't care particularly about testing it on testnet vs betanet; I expect betanet would be just fine too.

I've just deployed to betanet. I'm ok merging this in and letting it roll onto betanet on the next cycle.

zeldovich:
Hey @tsachiherman , so what's the right workflow for this to end up in betanet?  Should we merge this into master, and then it'll roll over into betanet the next time you update betanet?  Or should we merge this into some betanet branch (and/or you manually cherry-pick this PR into betanet)?  Max and I realized we don't fully understand the workflow for PRs in terms of how master and betanet fit together.

tsachiherman:
workflow is pretty simple:
(PR -> master) -> betanet -> mainnet.
Currently we're preparing the next mainnet upgrade. We try to make one "shift" at a time.
The next shift would move betanet changes into mainnet. Following that, we will move master into betanet.
Do you feel there would be a value of having a node with the new code catchup up from master, to see if we decode everything so far correctly ? ( or maybe you've already done that ? )

zeldovich:
OK, makes sense.

I've already checked that this code can catch up from existing nodes.

Sounds like the next step is to merge this into master.  Is now a good time, or are you currently working on something with master (like merging it into betanet or whatnot) that would make it better to wait a bit on merging this PR?

tsachiherman:
I'm ok with merging this change, but please do get ok from @derbear as well. If it won't work out, we'll roll back this change so that we can rework it offline.

derbear:
Could you re-merge with `master`? It would make it easier to review the change.

auction/sort.go
```diff
@@ -0,0 +1,29 @@
...
+// You should have received a copy of the GNU Affero General Public License
+// along with go-algorand.  If not, see <https://www.gnu.org/licenses/>.
+
+package auction
+
+// SortUint64 implements sorting by uint64 keys for
+// canonical encoding of maps in msgpack format.
+//msgp:ignore SortUint64
+//msgp:sort uint64 SortUint64
+type SortUint64 []uint64
```
derbear: What's the purpose of this type?

derbear: As in, I don't think it shows up anywhere in the change. Is this type required?

derbear: Also, a more general question--will `msgp` complain if it tries to serialize map with keys whose type I have not defined a sort function for? (Does this apply to all non `msgp:ignore`d types in the package?)

zeldovich: The `SortUint64` type implements `sort.Interface` for sorting `uint64`s.  The `msgp:sort` annotation in the comment above tells the msgp code-generator to use this type to sort map keys of type `uint64`.

I don't believe `msgp` will complain, but the generated go code for serializing a map will not compile if the key type (or the `//msgp:sort` annotation for the key type) doesn't implement `sort.Interface`.

zeldovich: Oh, and the `//msgp:ignore` annotation tells msgp not to generate encoders/decoders for that type.  This seems appropriate here because I don't expect any encoded/decoded data structures to contain `SortUint64`.

derbear: To clarify, it seems that there are three `sort.go` files (inside of `auction/`, `crypto/` and `data/basics/`), which specify sort interfaces for `uint64`, `Digest`, and `AssetIndex`, respectively. `git grep SortAssetIndex` shows up, as expected,
```
derek@crassus:~/newgo/go-algorand$ git grep SortAssetIndex
data/basics/msgp_gen.go:                        sort.Sort(SortAssetIndex(zb0001_keys))
data/basics/msgp_gen.go:                        sort.Sort(SortAssetIndex(zb0003_keys))
data/basics/msgp_gen.go:                        sort.Sort(SortAssetIndex(zb0001_keys))
data/basics/msgp_gen.go:                        sort.Sort(SortAssetIndex(zb0003_keys))
...
```
but the other two do not show up in generated code:
```
derek@crassus:~/newgo/go-algorand$ git grep SortDigest | grep -v auction/sort.go | grep -v crypto/sort.go | wc -l
0
derek@crassus:~/newgo/go-algorand$ git grep SortUint64 | grep -v auction/sort.go | grep -v crypto/sort.go | wc -l
0
```

zeldovich: Ah, I see.  Good catch!  I think they are remnants of data structures that used to be there but went away.

In particular, `SortDigest` was originally used because `auction.RunningAuction` has a `map[crypto.Digest]BidderState`.  But then I added support for `//msgp:ignore` to avoid codegen for that type altogether, since it's never actually encoded.  And I never went back to clean up the `SortDigest` thing.

The `SortUint64` used to be required when we had assets named by `uin64`, but then the `AssetIndex` type name got introduced for it, and I didn't clean up the previous `SortUint64` code.

I'll remove both of those unused sorts.  Thanks!

auction/encoding_test.go
```diff
@@ -0,0 +1,35 @@
...
+func TestEncodings(t *testing.T) {
+	protocol.RunEncodingTest(t, &Bid{})
+	protocol.RunEncodingTest(t, &SignedBid{})
+	protocol.RunEncodingTest(t, &Deposit{})
+	protocol.RunEncodingTest(t, &SignedDeposit{})
+	protocol.RunEncodingTest(t, &Params{})
+	protocol.RunEncodingTest(t, &SignedParams{})
+	protocol.RunEncodingTest(t, &Settlement{})
+	protocol.RunEncodingTest(t, &SignedSettlement{})
+	protocol.RunEncodingTest(t, &NoteField{})
```
derbear: Would it be possible to code-generate these tests, as well?

data/transactions/signedtxn_test.go
```diff
@@ -70,7 +70,6 @@ func TestDecodeNil(t *testing.T) {
 	if err == nil {
 		// These two functions used to panic when run on a zero value of SignedTxn.
```
derbear: Nit: update the comment

auction/encoding_test.go
zeldovich: Good idea; done.

auction/msgp_gen.go
```diff
@@ -0,0 +1,3405 @@
+package auction
+
+// Code generated by github.com/algorand/msgp DO NOT EDIT.
+
+import (
+	"github.com/algorand/msgp/msgp"
+)
+
+// MarshalMsg implements msgp.Marshaler
+func (z *Bid) MarshalMsg(b []byte) (o []byte, err error) {
```
derbear: Out of curiosity, how do we determine whether the type has a pointer receiver or a value one?

zeldovich: For mutable methods (i.e., `UnmarshalMsg`), it's always pointer receiver.  For immutable methods, there are some heuristics for deciding what to do:

https://github.com/algorand/msgp/blob/master/gen/spec.go#L272

PR #601, opened 2019-12-03
Add asset decimals field

justicz:
Prerequisite PR: https://github.com/algorand/go-algorand/pull/576

- Added a new field `Decimals` to `AssetParams`. I made it a `uint32` because I know we've had weird byte-alignment issues on certain builds, but if we don't need to worry about that we can make it a `uint8`.
- Added checks to `WellFormed` to ensure this field is only nonzero after the protocol upgrade.
- When displaying asset balances in `goal`, respect the `Decimals` field if it can fetch the asset params (it can't do that on offline machines, for example).
- Note that when _sending_ assets, `goal` still takes the base unit of the asset (you cannot pass `-a 1.234` of an asset with `Decimals = 3`). This is because we will sometimes fail to get an asset's parameters, for example if the machine is generating txns offline, and it would be bad to have changing semantics for the `amount` argument.
- Added `Decimals` to the API model of `AssetParams`


config/config.go
```diff
@@ -445,9 +448,24 @@ func initConsensusProtocols() {
...
+	// we want to adjust the upgrade time to be roughly one week.
+	// one week, in term of rounds would be:
+	// 140651 = (7 * 24 * 60 * 60 / 4.3)
+	// for the sake of future manual calculations, we'll round that down
+	// a bit :
+	v20.UpgradeWaitRounds = 140000
+	Consensus[protocol.ConsensusV20] = v20
+
+	// v19 can be upgraded to v20.
+	v19.ApprovedUpgrades[protocol.ConsensusV20] = true
```
zeldovich: Let's not do this until we have a final version string for `ConsensusV20`.

justicz: These parts of the PR originate from https://github.com/algorand/go-algorand/pull/576, I will rebase onto that before this is merged

protocol/consensus.go
```diff
@@ -108,6 +108,11 @@ const ConsensusV19 = ConsensusVersion(
 	"https://github.com/algorandfoundation/specs/tree/0e196e82bfd6e327994bec373c4cc81bc878ef5c",
 )
 
+// ConsensusV20 points to adding the precision to the assets.
+const ConsensusV20 = ConsensusVersion(
+	"https://github.com/algorandfoundation/specs/tree/add_precision_to_assets",
```
tsachiherman: Please change this to
"https://github.com/algorandfoundation/specs/tree/4a9db6a25595c6fd097cf9cc137cc83027787eaa"

cmd/goal/asset.go
```diff
@@ -598,13 +614,15 @@ var infoAssetCmd = &cobra.Command{
...
 		fmt.Printf("Asset ID:         %d\n", assetID)
 		fmt.Printf("Creator:          %s\n", params.Creator)
 		fmt.Printf("Asset name:       %s\n", params.AssetName)
 		fmt.Printf("Unit name:        %s\n", params.UnitName)
-		fmt.Printf("Maximum issue:    %d %s\n", params.Total, params.UnitName)
-		fmt.Printf("Reserve amount:   %d %s\n", reserve.Assets[assetID].Amount, params.UnitName)
-		fmt.Printf("Issued:           %d %s\n", params.Total-reserve.Assets[assetID].Amount, params.UnitName)
+		fmt.Printf("Maximum issue:    %s %s\n", assetDecimalsFmt(params.Total, params.Decimals), params.UnitName)
+		fmt.Printf("Reserve amount:   %s %s\n", assetDecimalsFmt(res.Amount, params.Decimals), params.UnitName)
+		fmt.Printf("Issued:           %s %s\n", assetDecimalsFmt(params.Total-res.Amount, params.Decimals), params.UnitName)
```
tsachiherman: You might want to print the decimals here as well.

test/e2e-go/cli/goal/expect/goalTxValidityTest.exp
```diff
@@ -58,19 +58,19 @@ if { [catch {
     set TX_FILE $TEST_ROOT_DIR/tx-create-$FILE_COUNTER.tx
     incr FILE_COUNTER
     set LV_EXPECTED 2
-    TestLastValidInTx "goal asset create --validrounds 1 --firstvalid 2 --creator $PRIMARY_ACCOUNT_ADDRESS --total 100 -d $TEST_PRIMARY_NODE_DIR -o $TX_FILE" $TX_FILE $LV_EXPECTED
+    TestLastValidInTx "goal asset create --decimals 0 --validrounds 1 --firstvalid 2 --creator $PRIMARY_ACCOUNT_ADDRESS --total 100 -d $TEST_PRIMARY_NODE_DIR -o $TX_FILE" $TX_FILE $LV_EXPECTED
```
tsachiherman: It would be nice if you could modify one of the existing asset test to cover a use case where decimals != 0

cmd/goal/asset.go
rotemh: Consider either printing the decimals as a separate field or actually doing the math and presenting it as a decimal number. Printing the number next to the amount seems a bit cryptic to me.

data/basics/userBalance.go
```diff
@@ -186,6 +186,13 @@ type AssetParams struct {
 	// created.
 	Total uint64 `codec:"t"`
 
+	// Decimals specifies the number of digits to display after the decimal
+	// place when displaying this asset. A value of 0 represents an asset
+	// that is not divisible, a value of 1 represents an asset divisible
+	// into tenths, and so on. This value must be between 0 and 19
+	// (inclusive).
+	Decimals uint32 `codec:"dc"`
```
rotemh: not dramatic, but why 32bits if we allow a range that is smaller than 8?

test/e2e-go/cli/goal/expect/goalCmdFlagsTest.exp
```diff
@@ -22,8 +22,8 @@ proc TestGoalCommandLineFlags { CMD EXPECTED_RE } {
 if { [catch {
     puts "Part 1: Check --validrounds and --lastvalid options combination for 'goal asset' and 'goal account"
 
-    TestGoalCommandLineFlags "goal asset create --validrounds 0 --creator ABC --total 100" ".*can not be zero.*"
-    TestGoalCommandLineFlags "goal asset create --validrounds 1 --lastvalid 1 --creator ABC --total 100" "Only one of .* can be specified"
+    TestGoalCommandLineFlags "goal asset create --decimals 0 --validrounds 0 --creator ABC --total 100" ".*can not be zero.*"
```
rotemh: isn't 0 the default? Consider testing some other numbers. (and w/o the flag, to make sure the default is actually being set)

protocol/consensus.go
justicz: done, thanks!

cmd/goal/asset.go
justicz: done

test/e2e-go/cli/goal/expect/goalTxValidityTest.exp
justicz: done

test/e2e-go/cli/goal/expect/goalCmdFlagsTest.exp
justicz: done -- yeah the flags were left over from when I made the decimals field mandatory (it's not now)

data/basics/userBalance.go
justicz: I mentioned this in the PR, but I remembered us having issues on some platforms when there were non word-aligned fields in structs. I talked about it with Tsachi though and that seemed to only happen when there were locks in the struct.

PR #605, opened 2019-12-04
Enumerate all ledgers

justicz:
Ledgers have lots of different product IDs. This PR enumerates all devices that appear to be manufactured by ledger.

This PR also only makes it a warning when a device fails to open. Before, when that would happen, all of kmd would crash since the wallet driver initialization would return an error. Instead, we just log an error to the log file now.

This PR does _not_ fix the issue where a ledger shows up twice. I thought about trying to fix that by checking `info.Interface` from https://godoc.org/github.com/karalabe/hid#DeviceInfo and deduping on that, but the docs indicate that `Interface` has platform specific behavior that I can't test easily.

### DIFF
--- a/daemon/kmd/wallet/driver/ledger_hid.go
+++ b/daemon/kmd/wallet/driver/ledger_hid.go
@@ -207,7 +207,7 @@ func LedgerEnumerate(log logging.Logger) ([]LedgerUSB, error) {
 		// Try to open the device
 		dev, err := info.Open()
 		if err != nil {
-			log.Warnf("enumerated but failed to open ledger %x", info.ProductID)
+			log.Warnf("enumerated but failed to open ledger %x: %v", info.ProductID, err)
 			continue
 		}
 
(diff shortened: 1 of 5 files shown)


daemon/kmd/wallet/driver/ledger_hid.go
```diff
@@ -207,23 +197,23 @@ func (l *LedgerUSB) USBInfo() hid.DeviceInfo {
...
-			}
-
-			devs = append(devs, LedgerUSB{
-				hiddev: dev,
-			})
+	for _, info := range hid.Enumerate(ledgerVendorID, 0) {
+		// Try to open the device
+		dev, err := info.Open()
+		if err != nil {
+			log.Warnf("enumerated but failed to open ledger %x", info.ProductID)
```
tsachiherman: nit: You might want to log the `err` as well.

justicz: oh, good idea

justicz: done

PR #611, opened 2019-12-06
TEAL: allow bnz to the end of program

pzbitskiy:
## Summary

* The following TEAL code is valid now:
int 1
int 1
bnz done
done:

* This allows jumping to the end similar to 'return 1' statement
in high level languages

## Test Plan

Assembler and Eval tests added.

### DIFF
--- a/data/transactions/logic/assembler_test.go
+++ b/data/transactions/logic/assembler_test.go
@@ -248,6 +248,18 @@ bnz nowhere`
 	require.Nil(t, program)
 }
 
+func TestAssembleJumpToTheEnd(t *testing.T) {
+	text := `intcblock 1
+intc 0
+intc 0
+bnz done
+done:`
+	program, err := AssembleString(text)
+	require.NoError(t, err)
+	require.Equal(t, 9, len(program))
+	require.Equal(t, []byte("\x01\x20\x01\x01\x22\x22\x40\x00\x00"), program)
+}
+
 func TestAssembleDisassemble(t *testing.T) {
 	// Specifically constructed program text that should be recreated by Disassemble()
 	// TODO: disassemble to int/byte psuedo-ops instead of raw intcblock/bytecblock/intc/bytec
--- a/data/transactions/logic/eval.go
+++ b/data/transactions/logic/eval.go
@@ -940,7 +940,7 @@ func checkBnz(cx *evalContext) int {
 	}
 	cx.nextpc = cx.pc + 3
 	target := cx.nextpc + int(offset)
-	if target >= len(cx.program) {
+	if target > len(cx.program) {
 		cx.err = errors.New("bnz target beyond end of program")
 		return 1
 	}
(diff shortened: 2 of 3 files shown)


derbear:
I think this technically requires a protocol upgrade.

algobolson:
@derbear I'm pretty sure this counts as a bug-fix in the assembler and is not a change to specified behavior.

tsachiherman:
@algobolson , this is a bug-fix, no arguing here. The issue is that the same program would get interpreted differently depending on what binary version you have.
This might make some transactions that were already accepted in the past fail now.. which would prevent us from re-evaluating the entire block-chain.

derbear:
Maybe I'm misunderstanding something here:

`checkBnz` is called by `checkStep`, which is called by `logic.Check`, which is called by `verify.LogicSigSanityCheck`, which is called by `verify.LogicSig`, which is called by `verify.stxnVerifyCore`, which is called by `stxnAsyncVerify`, which is called by `verify.TxnPool` (indirectly through our execution pool), which is finally called by *`ledger.BlockEvaluator.transaction`*. 

This change would cause a previously invalid transaction containing a specific `bnz` target to pass the block evaluator. This will cause the network to stall if not all nodes apply this upgrade simultaneously--a user which submits a transaction with a TEAL script of that form could cause unupgraded nodes to disagree with upgraded nodes, causing the Byzantine agreement protocol to stall.

Is there something wrong with the above analysis?

justicz:
I think Derek is right

pzbitskiy:
Right, some previously invalid logicsig transactions might become valid after this fix.

tsachiherman:
@derbear I see the issue you're pointing here. This stall would cause to to have "too few" votes to reach a quorum, and as a result we'll move to the next period. Regretfully, we're going to have the same issue there as well ( maybe; depend on the committees ).

Do you think we need to have some transaction-dispute mechanism to work around these issues ? i.e. vote to discard the pending transaction pool, advance the round number with empty rounds to make pending transactions obsolete, etc ?

derbear:
I don't think this has anything to do with the transaction pool. The bad transaction can be inside a proposed block. Once we finish executing block validation, we remember that the particular block was valid. It is dangerous to forget the block if it is possible it were committed. Upgraded nodes will continue to see the block as valid; un-upgraded nodes will continue to see it as invalid (until they upgrade).

algobolson:
So do we add this as a comment in the spec to fix previously under-defined behavior? Make eval.go check for LogicSigVersion >= 2 and allow branch-to-end only then?

pzbitskiy:
+1 to Brian

algobolson:
correction: if the compiled program version (the varuint prefix at the start of the bytecode blob) is >= 2, and LogicSigVersion >= 2 (which will be checked at the start of Check() and Eval() ), then allow branching to the end of the program

derbear:
Yeah, using LogicSigVersion seems fine.

algobolson:
I'm gonna write this up into spec change https://github.com/algorandfoundation/specs/pull/23 and probably merge this branch into another branch that bundles up other implementation of that spec change.

algobolson:
added to implementation in #637

Issue #604, opened 2019-12-04
TEAL decompiler can produce code that jumps to nonexistent label


justicz:
Programs that have multiple jumps to the same label fail to decompile correctly. For example:

```
int 1
bnz done
int 2
bnz done
int 3
done:
&&
```

when compiled and then decompiled with `goal clerk compile -D` generates

```
intcblock 1 2 3
intc_0
bnz label1
intc_1
bnz label2
intc_2
label2:
&&
```

where both bnzs should go to the same label

@algobolson

PR #612, opened 2019-12-06
fix Disassemble when multiple bnz have the same target label

algobolson:
for issue #604 
add test, fix dissassemble


stechu:
LGTM. Thanks for the fix!

PR #617, opened 2019-12-09
Fix license errors, enable check_license in travis.

winder:
We forgot to enable this in travis


winder:
> I would have preferred that
> 
> ```
> echo "Running check_license..."
>  ./scripts/check_license.sh
> ```
> 
> would be in a separate PR, but it's good regardless.

That's fair, I'll split up changes better in the future.

PR #620, opened 2019-12-10
Execution request transactions

gcolvin:
Here is a sketch of the PR code.

First we need a new Transaction type:
```
  algorand/go-algorand/src/data/transactions
      /transaction.go
          add ExecTxnFields to Transaction type
      /exec.go
          define ExecTxnFields type
              four phases — init, request, commit, fail
              data for checking storage

```
An account is addressed at the hash of its code, and exists on the ledger once it is funded in the ordinary way.

Code (will be) stored under its hash by an init transaction.  Once a code account exists a user can submit an execution request transaction - signed by its code - to the network.  (When code can be stored by hash the LogicSig can be a hash.)

Nodes pull transactions from the network for processing:
```
    algorand/go-algorand/src/data/transactions
        /txn.go
            LogicSig
               short circuit LogicSig for request transactions
        /exec.go
            now in roundCow
                apply transaction
                    on init:
                        store code by hash
                    on request:
                        set old data to storage
                    on commit:
                        if old data not same as storage
                            create failure transaction
                        else
                            set storage to new data    
                    on failure
                        create failure transaction
                    on …
                        …
                     …
               …
               propose transaction …
               …
```
        
An init, request, commit or failure transaction should now be on the blockchain.  Once the commit transaction is on the blockchain all state and balance changes are final.

The exebot pulls request transactions from the blockchain for processing, sending a raw commit or failure transaction to the network:
```
    algorand/exebot
        /exebot.py
            Exebot:
                loop
                    pulls blocks off of chain
                    parses transactions into JSON
                        passes each execution request transaction to handler

                execution handler:
                    pipe input, error, and output to buffers
                    spawn VM process with logicSig code
                    wait for process to exit
                    if error
                        send failure transaction
                    else
                        create commit transaction, set note field to output
                        (soon: create commit AMPT from output JSON)
                        send commit transaction
```


data/transactions/exec.go
```diff
@@ -0,0 +1,37 @@
...
+	"github.com/algorand/go-algorand/config"
+	"github.com/algorand/go-algorand/data/basics"
+)
+
+// ExecTxnFields captures the fields used by exec transactions.
+type ExecTxnFields struct {
+	_struct struct{} `codec:",omitempty,omitemptyarray"`
+
+	Inputs []string `codec:"inp`
+	Code string `codec:"cod"`
```
algobolson: Is this source code? Should it be []byte?

gcolvin: Yes, soon enough.  Should be just string for now, to cheat with a file name.

gcolvin:
@algobolson Recently pushed commits return to the design of signing the transaction with the Wasm code.  (Of course the code will need to be replaced with a hash once we can store code indexed by hash.)  I'm just short-circuiting the execution of the code in stxnVerifyCore() when the code is Wasm, so the transaction will land on the blockchain with code attached.

tsachiherman:
@derbear what would you like to do with this one ?

jing-algorand:
Can you make a copy and send it to me?

Thanks,
Jing

PR #622, opened 2019-12-11
Add `goal ledger block`

justicz:
Building on top of #621...

This PR adds a `goal ledger block <round>` command that fetches the raw msgpack of a block. By default it dumps the block encoded as json to stdout, but also takes in an `-o` parameter to specify an output file. Passing `-r` will dump the raw msgpack.

Closes https://github.com/algorand/go-algorand/issues/92


cmd/goal/ledger.go
```diff
@@ -22,8 +22,18 @@ import (
...
+	round       uint64
+	rawFilename string
+)
+
 func init() {
 	ledgerCmd.AddCommand(supplyCmd)
+	ledgerCmd.AddCommand(rawBlockCmd)
+
+	rawBlockCmd.Flags().Uint64VarP(&round, "round", "r", 0, "The round to fetch the raw block for")
+	rawBlockCmd.Flags().StringVarP(&rawFilename, "out", "o", stdoutFilenameValue, "The filename to dump the raw block to (if not set, use stdout)")
```
algoradam: Might be nicer UX to have this as an argument rather than a flag (e.g., `goal ledger rawblock 500` instead of `goal ledger rawblock -r=500`.)

justicz:
Alright -- made it a positional argument

cmd/goal/ledger.go
justicz: done

cmd/goal/ledger.go
```diff
@@ -52,3 +60,29 @@ var supplyCmd = &cobra.Command{
 		fmt.Printf("Round: %v\nTotal Money: %v microAlgos\nOnline Money: %v microAlgos\n", response.Round, response.TotalMoney, response.OnlineMoney)
 	},
 }
+
+var rawBlockCmd = &cobra.Command{
+	Use:   "rawblock",
```
algoradam: nitpick: should say "rawblock [round number]"

justicz: thanks, fixed

justicz:
were you thinking the output would look like what you get with `goal rawblock 1234 | msgpacktool -d`?

tsachiherman:
> were you thinking the output would look like what you get with `goal rawblock 1234 | msgpacktool -d`?

I'm not strongly opinionated about the precise formatting. However, the rest of our goal "interface" is primarily around user interaction with fallback for encoded data. ( for example, add "-o" so the transaction can be outputted to a file ).

if the output of `goal rawblock 1234 | msgpacktool -d` would end up being a parsable output ( i.e. jq / grep), then yes, I think it would be a good choice.

justicz:
Ok, it now dumps the encoded block by default, and the raw block if you pass `-r`. `-o` works the same.

cmd/goal/ledger.go
```diff
@@ -17,13 +17,30 @@
...
+	base32Encoding bool
+	strictJSON     bool
 )
 
 func init() {
 	ledgerCmd.AddCommand(supplyCmd)
+	ledgerCmd.AddCommand(blockCmd)
+
+	blockCmd.Flags().StringVarP(&rawFilename, "out", "o", stdoutFilenameValue, "The filename to dump the raw block to (if not set, use stdout)")
+	blockCmd.Flags().BoolVarP(&rawBlock, "raw", "r", false, "Dump raw block as msgpack")
```
tsachiherman: nit: Change the description to "Format block as msgpack".
The fact that we're internally storing the block in it's msgpack encoded form isn't visible to the goal end-user.

cmd/goal/ledger.go
```diff
@@ -17,13 +17,30 @@
...
+	rawBlock       bool
+	base32Encoding bool
+	strictJSON     bool
 )
 
 func init() {
 	ledgerCmd.AddCommand(supplyCmd)
+	ledgerCmd.AddCommand(blockCmd)
+
+	blockCmd.Flags().StringVarP(&rawFilename, "out", "o", stdoutFilenameValue, "The filename to dump the raw block to (if not set, use stdout)")
```
tsachiherman: nit: drop the `raw` from the description here.

cmd/goal/ledger.go
justicz: done!

cmd/goal/ledger.go
justicz: done!

PR #623, opened 2019-12-11
Simplified the release scripts that build images to push to docker hub

btoll:
In pushing the updated images to docker hub, I noticed that the Dockerfiles and the shell scripts were only differentiated by the network name (stable|testnet).

The only file in the dir is now `build_releases.sh`.  It will default to "mainnet", so the for that image it's only necessary to run `./build_releases.sh` with no args.

For "testnet", simply call the script like this: `./build_releases.sh testnet`.

The Dockerfile will be automatically created and passed to the `docker build` command via `stdin`.


docker/releases/build_stable.sh
```diff
@@ -1,13 +1,60 @@
...
+            NAME="$1"
+            NETWORK="-g $1"
+            ;;
+        *)
+            echo "$RED_FG[$0]$END_FG_COLOR Unknown option $1."
+            exit 1
+            ;;
+    esac
+    shift
+done
```
winder: Can probably skip the parameters, the only valid parameter is "testnet" anyway. I think you can just tell this to build mainnet/testnet.

btoll: Ok, will change.

tsachiherman:
> > Looks mostly fine. While we're poking around here I think we should rename the `algorand/stable` container to `algorand/mainnet`.
> 
> Will change.

Could you hold off with that change for a short while ?

I think that it's a good idea, but it would also require to change the instructions that we had..

docker/releases/build_stable.sh
tsachiherman: You still need to check that this script received either mainnet or testnet parameter.. otherwise, just exit the script with some instructions about the correct usage.

btoll: Will check for only those two strings.

btoll:
> Could you hold off with that change for a short while ?

Yes.  I'll add what has been suggested and then go pencils down.

docker/releases/build_stable.sh
btoll: Will check for only those two strings.

btoll:
@tsachiherman @winder Ok, I'm done making changes on this unless you want me to do more.

PR #624, opened 2019-12-11
DRAFT: Standalone tealc

algobolson:
This is a kinda messy change that moves the assembler to its own package with minimal dependencies. Then new cmd/tealc produces a binary on Linux that is fully statically linked and should be very portable. A few things from `protocol` and `basics` got copied into the assembler package to minimize dependencies, in particular to not pull in `crypto` and C libsodium.a; and `basics` and `go-codec` which is `unsafe`. (`import "unsafe"` causes Linux binaries to link to libc and be less portable)


algobolson:
Sorry, this isn't actually a good idea and should still count as DRAFT. DO NOT COMMIT

algobolson:
not the way to go

PR #626, opened 2019-12-12
Bring `shellcheck` into the build process

btoll:
Only check for the presence of the dependencies.  If one or more are missing, exit 1 and echo the install script (`./scripts/configure-dev.sh`).

https://www.shellcheck.net/
https://github.com/koalaman/shellcheck

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -56,10 +56,8 @@ vet:
 check_license:
 	./scripts/check_license.sh
 
-# Turn on recursive globbing to find all shell scripts.
 check_shell:
-	bash -c "shopt -s globstar"
-	shellcheck */**/*.sh
+	find . -type f -name *.sh -exec shellcheck {} +
 
 sanity: vet fix lint fmt check_license
 
--- a/Makefile
+++ b/Makefile
@@ -57,7 +57,7 @@ check_license:
 	./scripts/check_license.sh
 
 check_shell:
-	find . -type f -name *.sh -exec shellcheck {} +
+	find . -type f -name "*.sh" -exec shellcheck {} +
 
 sanity: vet fix lint fmt check_license
 
--- a/Makefile
+++ b/Makefile
@@ -56,6 +56,11 @@ vet:
 check_license:
 	./scripts/check_license.sh
 
+# Turn on recursive globbing to find all shell scripts.
+check_shell:
+	bash -c "shopt -s globstar"
+	shellcheck */**/*.sh
+
 sanity: vet fix lint fmt check_license
 
 cover:
@@ -243,4 +248,4 @@ dump: $(addprefix gen/,$(addsuffix /genesis.dump, $(NETWORKS)))
 install: build
 	scripts/dev_install.sh -p $(GOPATH1)/bin
 
-.PHONY: default fmt vet lint check_license sanity cover prof deps build test fulltest shorttest clean cleango deploy node_exporter install %gen gen NONGO_BIN
+.PHONY: default fmt vet lint check_license check_shell sanity cover prof deps build test fulltest shorttest clean cleango deploy node_exporter install %gen gen NONGO_BIN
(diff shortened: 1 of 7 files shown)


Makefile
```diff
@@ -56,6 +56,11 @@ vet:
 check_license:
 	./scripts/check_license.sh
 
+# Turn on recursive globbing to find all shell scripts.
+check_shell:
+	bash -c "shopt -s globstar"
+	shellcheck */**/*.sh
```
tsachiherman: I think that you want to do:
```
shopt -s globstar && \
shellcheck */**/*.sh
```
right ?

scripts/configure_dev-deps.sh
```diff
@@ -14,3 +14,23 @@ function install_go_module {
...
+        apt-get update
+        apt-get install sudo -y
+    fi
+
+    sudo apt-get update
+    sudo apt-get install shellcheck -y
+elif [ "$OS" == "Darwin" ]
+then
+    brew install shellcheck
+fi
```
tsachiherman: All that logic need to go to `configure-dev.sh`

btoll: I agree that is where it should go, but I don't think I can move that there without some extra work first.  Here is a grep that shows where else that file is called:

```
~/projects/go-algorand:master$ ag -C 2 configure_dev-deps.sh
docker/build/Dockerfile-deploy
22-    NETWORK=${NETWORK}
23-WORKDIR $GOPATH/src/github.com/algorand/go-algorand
24:RUN scripts/configure_dev-deps.sh && make deps && make clean && find tmp && TMPDIR/deploy_linux_version_exec.sh
25-ENTRYPOINT ["/bin/bash"]
26-

docker/build/Dockerfile
18-    PKG_ROOT=${PKG_ROOT}
19-WORKDIR $GOPATH/src/github.com/algorand/go-algorand
20:RUN scripts/configure_dev-deps.sh && make deps && make clean && scripts/local_install.sh -n
21-ENTRYPOINT ["/bin/bash"]
22-

scripts/build_release_centos_docker.sh
23-REPO_DIR="$( cd "$(dirname "$0")" ; pwd -P )"/..
24-
25:${REPO_DIR}/scripts/configure_dev-deps.sh
26-
27-cd ${REPO_DIR}

scripts/configure_dev.sh
33-fi
34-
35:${SCRIPTPATH}/configure_dev-deps.sh
36-
```

Makefile
btoll: `make` must use the Bourne shell to execute commands:

```
~/projects/go-algorand:shellcheck$ make check_shell 
shopt -s globstar && \
        shellcheck */**/*.sh
/bin/sh: 1: shopt: not found
Makefile:63: recipe for target 'check_shell' failed
make: *** [check_shell] Error 127
```
Can we specify the shell in the `Makefile`?

scripts/configure_dev-deps.sh
tsachiherman: -> For the two docker files, you need to add the spellcheck to the `RUN apt` line.
-> for the build_release_centos_docker.sh, it need to be added to the `RUN yum install -y autoconf awscli git gnupg2 nfs-utils python36 sqlite3 boost-devel expect jq libtool gcc-c++ libstdc++-devel libstdc++-static rpmdevtools createrepo rpm-sign bzip2` line.
The last one is good and doesn't need to be modified.

Makefile
tsachiherman: I think that we should attempt to stay more posix compliant here.. 
what's about `find . -name "*.sh" | xargs shellcheck` ?

scripts/configure_dev-deps.sh
btoll: Sounds good.

Makefile
btoll: Ok

Makefile
```diff
@@ -56,6 +56,9 @@ vet:
 check_license:
 	./scripts/check_license.sh
 
+check_shell:
+	find . -type f -name *.sh -exec shellcheck {} +
```
tsachiherman: this is a good idea, but you'll need to have the `*.sh` surrounded by double-quotes. i.e. `"*.sh"`

btoll: Will change.

PR #628, opened 2019-12-12
Cleanup evalAux

algobolson:
Simplify ledger data storage by removing 'evalAux' field which hasn't been used since before 1.0

### DIFF
--- a/ledger/acctupdates.go
+++ b/ledger/acctupdates.go
@@ -162,12 +162,9 @@ func (au *accountUpdates) loadFromDisk(l ledgerForTracker) error {
 		next := loaded + 1
 
 		blk, err := l.Block(next)
-		/*
-			blk, aux, err := l.blockAux(next)
-			if err != nil {
-				return err
-			}
-		*/
+		if err != nil {
+			return err
+		}
 
 		delta, err := l.trackerEvalVerified(blk)
 		if err != nil {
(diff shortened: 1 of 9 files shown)


algobolson:
This doesn't change the database schema for existing nodes, it just stops writing an empty struct into a column we never use, and stops fetching it back from the database. For a new node the column wouldn't exist. For an existing node the column is full of empty structs we can ignore.

tsachiherman:
> This doesn't change the database schema for existing nodes, it just stops writing an empty struct into a column we never use, and stops fetching it back from the database. For a new node the column wouldn't exist. For an existing node the column is full of empty structs we can ignore.

Right. but what happen if we now try to add a column to the schema. And that column happen to have the same name. It will execute correctly on some node, and fail on others. And given that it's deleted from the source code, we have no "memory" of the existence of this column.

tsachiherman:
btw - I completely agree that we should remove this column from the `insert` statements.

algoradam:
I think this shouldn't require a protocol upgrade since it doesn't change block evaluation at all. If we're worried we might forget and accidentally add an "auxdata" column to the schema sometime before we next upgrade the protocol, we can leave a comment in the code as a reminder not to use that name.

algobolson:
added comment noting the removal or vestigial presence of the auxdata column

PR #633, opened 2019-12-13
Fix asset unit name display in goal account list

justicz:
default value should be set before if statement, not after


justicz:
@tsachiherman any way this can make it into the release?

PR #634, opened 2019-12-13
Ensure that the proper channel is passed to `test_release.sh`

btoll:


test/packages/test_release.sh
```diff
@@ -24,8 +24,8 @@ OS_LIST=(
 
 # These are default values which can be changed by the CLI args.
 BUCKET=algorand-builds
-CHANNEL=stable
-
+BRANCH=$(../../scripts/compute_branch.sh)
+CHANNEL=$(../../scripts/compute_branch_channel.sh "$BRANCH")
```
tsachiherman: I think that placing this very logic on the caller makes more sense, don't you think so ? i.e. `/scripts/travis/test_release.sh`

Also, when you do that, you could use the TRAVIS_BRANCH environment variable. ( so you won't need the `compute_branch.sh` )

PR #639, opened 2019-12-17
move to go 1.13

algobolson:
Go 1.14 is in beta; 1.13 has been out a while. Let's upgrade

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 dist: xenial
 go:
-  - "1.12"
+  - "1.13"
 go_import_path: github.com/algorand/go-algorand
 language: go
 sudo: required
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ about using and installing the algorand platform.
 ## Building from source ##
 
 Development is done using the [Go Programming Language](https://golang.org/)
-version 1.12.x, and this document assumes that you have a functioning
+version 1.13.x, and this document assumes that you have a functioning
 environment setup. If you need assistance setting up an environment please visit
 the [official Go documentation website](https://golang.org/doc/).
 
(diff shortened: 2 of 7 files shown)


winder:
There are also several dockerfiles which would need to be updated.

algobolson:
found these:
``` bash
$ find . -type f -iname \*docker\* |xargs grep 1.12
./docker/Dockerfile:ENV GOLANG_VERSION 1.12
./docker/build/Dockerfile-deploy:ENV GOLANG_VERSION 1.12
./docker/build/Dockerfile:ENV GOLANG_VERSION 1.12```

tsachiherman:
Do you see any good reason we're not installing the go during the configure_dev.sh ?
The configure_dev.sh step is expected to be called by all the machines prior to building, and we're already installing some pieces there.

algobolson:
Eventually there's a vision to make "make the build environment" a separate step from "make the thing". Maybe we can do this as is in our Makefile pretty much as is where we could have `make` do everything, build a `make setup` that just does all the environment stuff, and adjust `make build` to just do the compilation step assuming all the tools are installed.
I feel like automatically installing `go` for a developer is a weird step. Automatically dumping it into the default /usr/local/go/... feels intrusive on a machine I don't own. Doing it on systems we own like build servers should be a configuration step of the build server.

tsachiherman:
I don't see how that is any different than installing boost, expect, jq etc.
Given that it goes into the same category, I think that we should control it from a single location.
Keep in mind that I'm not suggesting to override the system installed go. I'm suggesting to install it only if it's missing ( which is the use for out build machine ).
Alternatively, we can do it in `./scripts/travis/configure_dev.sh` which is used only by our build machines. ( ignore the fact that it's called "travis"; that's only there because it's currently being invoked by travis build )

algobolson:
installing boost/expect/jq etc happen inside package managers, either `apt` or `brew`. Someone who already uses those system wide package managers shouldn't be badly surprised to have something new show up within that system.

winder:
This is superseded by #1185

PR #642, opened 2019-12-18
scripts: fix for multi-part GOPATH (with colons)

dncohen:
## Summary

Currently `make install` fails when $GOPATH has a colon.  the build_release_local.sh script assumes $GOPATH is simply a directory.  But GOPATH is allowed to contain many directories separated by colons.  When that is the case `make install` fails with an error when build_release_local.sh attempts to copy files from $GOPATH.

This patch uses ${GOPATH%%:*} (the portion before the first ":"), rather than the entire $GOPATH. 
 This fixes `make install` (the only problem I've run into so far).  It would not surprise me to learn the same problem exists in other scripts.

## Test Plan

Without the patch, my local `make install` fails.  With the patch, it succeeds.

### DIFF
--- a/scripts/build_package.sh
+++ b/scripts/build_package.sh
@@ -59,7 +59,7 @@ mkdir ${PKG_ROOT}/bin
 # If you modify this list, also update this list in ./cmd/updater/update.sh backup_binaries()
 bin_files=("algocfg" "algod" "algoh" "algokey" "carpenter" "catchupsrv" "ddconfig.sh" "diagcfg" "find-nodes.sh" "goal" "kmd" "msgpacktool" "node_exporter" "update.sh" "updater" "COPYING")
 for bin in "${bin_files[@]}"; do
-    cp ${GOPATH}/bin/${bin} ${PKG_ROOT}/bin
+    cp ${GOPATH%%:*}/bin/${bin} ${PKG_ROOT}/bin
     if [ $? -ne 0 ]; then exit 1; fi
 done
 
@@ -113,7 +113,7 @@ echo "Staging tools package files"
 bin_files=("algons" "auctionconsole" "auctionmaster" "auctionminion" "coroner" "dispenser" "netgoal" "nodecfg" "pingpong" "cc_service" "cc_agent" "cc_client" "COPYING")
 mkdir -p ${TOOLS_ROOT}
 for bin in "${bin_files[@]}"; do
-    cp ${GOPATH}/bin/${bin} ${TOOLS_ROOT}
+    cp ${GOPATH%%:*}/bin/${bin} ${TOOLS_ROOT}
     if [ $? -ne 0 ]; then exit 1; fi
 done
 


dncohen:
It's useful to have multiple directories in GOPATH.  Personally, I put packages and tools that are common to every project in ~/go, but project specific things in ~/gopath/<project>.  So my GOPATH working with algorand is "/home/user/gopath/algorand:/home/user/go".

Your suggestion would silently change it to just "/home/user/gopath/algorand" when type `make`; causing different versions of tools and packages to be installed in two places, and causing inconsistencies between my development environment vs the algorand build.

If you can support only GOPATHs with a single entry, it would be much nicer to fail with "algorand can only be built when GOPATH has one path".

tsachiherman:
> Your suggestion would silently change it to just "/home/user/gopath/algorand" when type `make`; causing different versions of tools and packages to be installed in two places, and causing inconsistencies between my development environment vs the algorand build.

Could you explain the portion about "causing different versions of tools and packages to be installed in two places" ? The Makefile is currently using the first portion of the GOPATH as well, so I would the build to pass if the first portion contain the go-algorand, or fail otherwise.

dncohen:
The way I read the Makefile, it defines $GOPATH1 to be the first path in $GOPATH.  It uses $GOPATH1 when looking for executables, but it doesn't modify my $GOPATH when building.  (Please let me know if I misunderstand.)  

The change I suggest in this PR tries to do the same.  It uses only the first path when installing build artifacts.  But uses the entire GOPATH when compiling go code.

If I understand your earlier suggestion, you would change $GOPATH before building.  I'm pushing back against changing user's GOPATH for them, because it could have unintended consequences.  Suppose I use GOPATH="A:B".  I have Algorand specific source code in directory A, and I have my build tools and several packages in B.  Some of A's dependencies are in B.  Maybe I've patched code in B.  Next suppose I unwittingly run "GOPATH=A make install"... the make could `go get` many of the same packages that are already in B; possibly different versions of those dependencies.  It might then take me a lot of troubleshooting to figure out why the build finds dependencies in A when my editor, IDE or shell scripts expected them to be found in B.

To be clear, if the makefiles only support a single-entry GOPATH, that is workable. But if that's the case, make should fail with an error when GOPATH is mult-entry.  I'm saying it should not change the user's GOPATH for them.  Failing is preferable; and the user would adjust GOPATH manually to work around the issue.

tsachiherman:
> If I understand your earlier suggestion, you would change $GOPATH before building. I'm pushing back against changing user's GOPATH for them, because it could have unintended consequences. Suppose I use GOPATH="A:B". I have Algorand specific source code in directory A, and I have my build tools and several packages in B. Some of A's dependencies are in B. Maybe I've patched code in B. Next suppose I unwittingly run "GOPATH=A make install"... the make could `go get` many of the same packages that are already in B; possibly different versions of those dependencies. It might then take me a lot of troubleshooting to figure out why the build finds dependencies in A when my editor, IDE or shell scripts expected them to be found in B.

Hmm.. I see your point. In that case, why don't you create another variable 
```
GOPATH1=${GOPATH%%:*}
```

and use this variable instead of the (incorrect) usage of GOPATH ?

tsachiherman:
I'm closing this PR as there was no activity here for over a month. If you still want to have you changes reviewed, please open a new PR.

thanks!

Issue #508, opened 2019-11-13
Improve pull test speed


algobolson:
3 day timebox - may result in additional tickets

It takes over an hour for testing system to validate a pull request. I get distracted and check in an hour or two later and this slows down daily development.

Tests should run faster. Maybe some slow system tests that run networks and wait for rounds could run in parallel. Walk through the sets of tests called in `scripts/travis/integration_test.sh` and `scripts/travis/build_test.sh` and see what's slow and what can be optimized.

pzbitskiy:
e2e_go_tests execution time improved by 10-12 minutes.
next stops are `expect` tests and `auction` tests.

pzbitskiy:
Fixed stderr polling in expect tests, +2-5 minutes.

PR #645, opened 2019-12-18
Reduce e2e_go_tests execution time twice

pzbitskiy:
There are seven major contributors to integration tests running time
TestOnlineOfflineRewards (1248.64s)
TestAssetConfig (364.71s)
TestRewardRateRecalculation (226.78s)
TestStartAndEndAuctionTenUsersOneBidEach (196.34s)
TestNoDepositAssociatedWithBid (189.74s)
TestDeadbeatBid (188.70s)
TestStartAndCancelAuctionNoBids (183.35s)

This commit considers only first three.

1. Fixing rewards interval in config for TestRewardRateRecalculation
from 25 to 10 reduces time twice:
TestRewardRateRecalculation (119.34s)

2. Fixing initialRound in TestOnlineOfflineRewards test
from 301 to 11 reduces time 15 times:
TestOnlineOfflineRewards (73.80s)

3. TestAssetConfig looks long by design - commits and waits max allowed assets

4. Address TODO in run_integration_tests.sh.
Now e2e_client_runner calls 'goal network delete' to reflect this removal

Refers #508

### DIFF
--- a/test/scripts/e2e_client_runner.py
+++ b/test/scripts/e2e_client_runner.py
@@ -264,12 +264,39 @@ class RunSet:
                 self.ok = False
                 self._terminate()
 
-def goal_network_stop(netdir):
+
+# 'network stop' and 'network delete' are also tested and used as cleanup procedures
+# so it re-raises exception in 'test' mode
+already_stopped = False
+already_deleted = False
+
+def goal_network_stop(netdir, normal_cleanup=False):
+    global already_stopped, already_deleted
+    if already_stopped or already_deleted:
+        return
+
     logger.info('stop network in %s', netdir)
     try:
         xrun(['goal', 'network', 'stop', '-r', netdir], timeout=10)
     except Exception as e:
         logger.error('error stopping network', exc_info=True)
+        if normal_cleanup:
+            raise e
+    already_stopped = True
+
+def goal_network_delete(netdir, normal_cleanup=False):
+    global already_deleted
+    if already_deleted:
+        return
+
+    logger.info('delete network in %s', netdir)
+    try:
+        xrun(['goal', 'network', 'delete', '-r', netdir], timeout=10)
+    except Exception as e:
+        logger.error('error deleting network', exc_info=True)
+        if normal_cleanup:
+            raise e
+    already_deleted = True
 
 def xrun(cmd, *args, **kwargs):
     timeout = kwargs.pop('timeout', None)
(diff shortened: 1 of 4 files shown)


test/scripts/e2e_client_runner.py
```diff
@@ -353,6 +360,8 @@ def main():
     retcode = 0
     xrun(['goal', 'network', 'create', '-r', netdir, '-n', 'tbd', '-t', os.path.join(gopath, 'src/github.com/algorand/go-algorand/test/testdata/nettemplates/TwoNodes50EachFuture.json')], timeout=30)
     xrun(['goal', 'network', 'start', '-r', netdir], timeout=30)
+    if not args.keep_temps:
```
algobolson: don't need to specifically delete these because the entire tmpdir they live in is deleted by the above
``` Python
if not args.keep_temps:
   atexit.register(shutil.rmtree, tempdir, onerror=logger.error)
```

pzbitskiy:
I understand, this is to justify TODO and create/start/stop/delete test
deletion.

algobolson:
Ohhhh, I get it. It's a _test_ of `goal network delete`.
Now I wish it was in-line with main execution so that if that step fails it fails the test. The `atexit` handlers won't do that. The current atexit handler for `goal network stop` should be treated the same way. Maybe register the atexit _and_ do them in main() and have the code in main() set a global flag so the atexit handler knows the step was done and doesn't need doing?

pzbitskiy:
Rebased and added explicit testing calls for goal_network_stop and goal_network_delete.

test/scripts/e2e_client_runner.py
```diff
@@ -264,12 +264,39 @@ def wait(self, timeout):
...
+        return
+
+    logger.info('delete network in %s', netdir)
+    try:
+        xrun(['goal', 'network', 'delete', '-r', netdir], timeout=10)
+    except Exception as e:
+        logger.error('error deleting network', exc_info=True)
+        if normal_cleanup:
+            raise e
+    already_deleted = False
```
algobolson: True?

test/scripts/e2e_client_runner.py
```diff
@@ -264,12 +264,39 @@ def wait(self, timeout):
...
     logger.info('stop network in %s', netdir)
     try:
         xrun(['goal', 'network', 'stop', '-r', netdir], timeout=10)
     except Exception as e:
         logger.error('error stopping network', exc_info=True)
+        if normal_cleanup:
+            raise e
+    already_stopped = True
+
+def goal_network_delete(netdir, normal_cleanup=False):
```
algobolson: add atexit registration of this after `goal network create` ?

test/scripts/e2e_client_runner.py
pzbitskiy: My bad

test/scripts/e2e_client_runner.py
pzbitskiy: This is probably not needed because there is a generic cleaner with `rmtree`.
`already_deleted` flag is still needed to prevent running `goal_network_stop` from atexit handler.

PR #648, opened 2019-12-18
Improve missing msig preimage error message

justicz:
:pencil2:


daemon/kmd/wallet/driver/sqlite_errors.go
```diff
@@ -23,6 +23,7 @@ import (
 var errDatabase = fmt.Errorf("database error")
 var errDatabaseConnect = fmt.Errorf("error connecting to database")
 var errKeyNotFound = fmt.Errorf("key does not exist in this wallet")
+var errPreimageNotFound = fmt.Errorf("multisig preimage does not exist in this wallet")
```
algobolson: Is there a better name than "multisig preimage"? I'm not sure what part or life-cycle-phase of a multisig address a "preimage" is. Is there a term in our documentation that fits? https://developer.algorand.org/docs/creating-multisignature-account

justicz: I just made it talk more broadly about "information"

PR #650, opened 2019-12-18
Allow upgrades to specify the delay before their execution.

derbear:
This replaces UpgradeWaitRounds with MinUpgradeWaitRounds and
MaxUpgradeWaitRounds.  Proposers specify an upgrade's delay given
their own ApprovedUpgrades, encoding the proposed delay in
the UpgradeVote.  Verifiers check that the delay sits between
MinUpgradeWaitRounds and MaxUpgradeWaitRounds (inclusive).

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -75,8 +75,10 @@ type ConsensusParams struct {
 	// acceptance and its execution.  This gives clients time to notify
 	// users.  This delay is specified by the upgrade proposer and must
 	// be between MinUpgradeWaitRounds and MaxUpgradeWaitRounds (inclusive)
-	// in the old protocol's parameters.  If the delay is zero, it is
-	// equivalent to DefaultUpgradeWaitRounds.
+	// in the old protocol's parameters.  Note that these parameters refer
+	// to the representation of the delay in a block rather than the actual
+	// delay: if the specified delay is zero, it is equivalent to
+	// DefaultUpgradeWaitRounds.
 	//
 	// The maximum length of a consensus version string is
 	// MaxVersionStringLen.
--- a/config/config.go
+++ b/config/config.go
@@ -75,15 +75,17 @@ type ConsensusParams struct {
 	// acceptance and its execution.  This gives clients time to notify
 	// users.  This delay is specified by the upgrade proposer and must
 	// be between MinUpgradeWaitRounds and MaxUpgradeWaitRounds (inclusive)
-	// in the old protocol's parameters.
+	// in the old protocol's parameters.  If the delay is zero, it is
+	// equivalent to DefaultUpgradeWaitRounds.
 	//
 	// The maximum length of a consensus version string is
 	// MaxVersionStringLen.
-	UpgradeVoteRounds    uint64
-	UpgradeThreshold     uint64
-	MinUpgradeWaitRounds uint64
-	MaxUpgradeWaitRounds uint64
-	MaxVersionStringLen  int
+	UpgradeVoteRounds        uint64
+	UpgradeThreshold         uint64
+	DefaultUpgradeWaitRounds uint64
+	MinUpgradeWaitRounds     uint64
+	MaxUpgradeWaitRounds     uint64
+	MaxVersionStringLen      int
 
 	// MaxTxnBytesPerBlock determines the maximum number of bytes
 	// that transactions can take up in a block.  Specifically,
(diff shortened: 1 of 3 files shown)


derbear:
This change itself should not require a protocol upgrade (since it changes no behavior)--a protocol upgrade can be executed later to modify MinUpgradeWaitRounds and MaxUpgradeWaitRounds.

config/config.go
```diff
@@ -257,10 +268,10 @@ func initConsensusProtocols() {
 
 	// Base consensus protocol version, v7.
 	v7 := ConsensusParams{
-		UpgradeVoteRounds:   10000,
-		UpgradeThreshold:    9000,
-		UpgradeWaitRounds:   10000,
-		MaxVersionStringLen: 64,
+		UpgradeVoteRounds:        10000,
+		UpgradeThreshold:         9000,
+		DefaultUpgradeWaitRounds: 10000,
```
zeldovich: I think you want to specify `MinUpgradeWaitRounds` and `MaxUpgradeWaitRounds` here.

config/config.go
```diff
@@ -439,83 +450,83 @@ func initConsensusProtocols() {
...
 func initConsensusTestProtocols() {
 	// Various test protocol versions
 	Consensus[protocol.ConsensusTest0] = ConsensusParams{
-		UpgradeVoteRounds:   2,
-		UpgradeThreshold:    1,
-		UpgradeWaitRounds:   2,
-		MaxVersionStringLen: 64,
+		UpgradeVoteRounds:        2,
+		UpgradeThreshold:         1,
+		DefaultUpgradeWaitRounds: 2,
```
zeldovich: `MinUpgradeWaitRounds` and `MaxUpgradeWaitRounds` here too

config/config.go
```diff
@@ -439,83 +450,83 @@ func initConsensusProtocols() {
...
 	}
 
 	Consensus[protocol.ConsensusTest1] = ConsensusParams{
-		UpgradeVoteRounds:   10,
-		UpgradeThreshold:    8,
-		UpgradeWaitRounds:   10,
-		MaxVersionStringLen: 64,
+		UpgradeVoteRounds:        10,
+		UpgradeThreshold:         8,
+		DefaultUpgradeWaitRounds: 10,
```
zeldovich: `MinUpgradeWaitRounds` and `MaxUpgradeWaitRounds` here too

config/config.go
derbear: I'm having a value of 0 denote `DefaultUpgradeWaitRounds`, which might be a bit confusing.

derbear: (See https://github.com/algorand/go-algorand/pull/650/files#diff-590486bd16364da6743388800e2ea504R312)

data/bookkeeping/block.go
```diff
@@ -364,18 +382,21 @@ func ProcessUpgradeParams(prev BlockHeader) (uv UpgradeVote, us UpgradeState, er
...
-	// If there is a proposal being voted on, see if we approve it
+	// If there is a proposal being voted on, see if we approve it and its delay
 	round := prev.Round + 1
 	if round < prev.NextProtocolVoteBefore {
-		if prevParams.ApprovedUpgrades[prev.NextProtocol] {
+		expectedDelay, ok := prevParams.ApprovedUpgrades[prev.NextProtocol]
+		if expectedDelay == 0 {
+			expectedDelay = prevParams.DefaultUpgradeWaitRounds
+		}
+		if ok && prev.NextProtocolVoteBefore+basics.Round(expectedDelay) == prev.NextProtocolSwitchOn {
```
zeldovich: That's somewhat weird behavior: a node won't vote yes unless the proposer picked the same UpgradeWaitRounds as this node?  What's the rationale?

config/config.go
zeldovich: Ah, I see, setting them to zero forces the block to have a zero value which is backwards-compatible..

It would be good to revise the description of these fields in the config type definition above, to make it clear that these apply to the block header representation, rather than the actual number of rounds to wait before upgrading.

data/bookkeeping/block.go
derbear: I was thinking it would be good to avoid the scenario where a proposer tries to schedule a more urgent upgrade with a large delay or a slow upgrade with a short delay. I don't know how much this matters.

zeldovich: Either way seems potentially reasonable, but if there's no flexibility in terms of the number of wait rounds, then it seems like this should just be a fixed value in the next protocol's `ConsensusParams`, and then we don't need any min/max values.

config/config.go
derbear: Sounds good.

The other thing I considered was to have the delay implicit in `NextProtocolSwitchOn`, so no field needs to be introduced in the block header. The drawback of this is that it makes `applyUpgradeVote` a bit messier (see the first commit on this PR). For consistency, I've decided to add the field to make it explicit.

derbear:
A following PR will place this change in `ConsensusFuture` and will add tests for the bounds.

PR #651, opened 2019-12-19
Test pre-packaged executable on variety of linux platforms

tsachiherman:
## Summary

Our generated binaries are expected to work on a variety of platforms. Before doing any packaging for these, we need to perform some sanity checking to see that the binary is capable of running on these platforms.

This is where this PR comes in. It generate series of temporary docker images ( one for each platform ), and attempt to run goal and algod. If the process is capable of loading into memory and executing, it will pass the test.

### DIFF
--- a/test/platform/test_linux_amd64_compatibility.sh
+++ b/test/platform/test_linux_amd64_compatibility.sh
@@ -21,7 +21,7 @@ PLATFORM=$($SCRIPTPATH/../../scripts/osarchtype.sh)
 
 if [ "${PLATFORM}" != "linux/amd64" ]
 then
-    echo "The test_linux_amd64_compatibility.sh script is intended to support local execution only on linux/x86-64 machines."
+    echo "$RED_FG[$0]$END_FG_COLOR The test_linux_amd64_compatibility.sh script is intended to support local execution only on linux/x86-64 machines."
     exit 0
 fi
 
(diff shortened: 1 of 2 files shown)


test/platform/test_linux_amd64_compatibility.sh
```diff
@@ -0,0 +1,98 @@
...
+
+WORKDIR /root
+CMD ["/bin/bash"]
+EOF
+
+    for item in ${OS_LIST[*]}
+    do
+        WITH_PACMAN=$(echo -e "${TOKENIZED}")
+
+        echo -e "$BLUE_FG[$0]$END_FG_COLOR Testing $item..."
```
btoll: I don't think you need the `-e` here.

test/platform/test_linux_amd64_compatibility.sh
```diff
@@ -0,0 +1,98 @@
...
+
+        echo -e "${WITH_PACMAN/\{\{OS\}\}/$item}" > Dockerfile
```
btoll: How about just giving echo `$TOKENIZED` here instead of creating the temp var `WITH_PACKMAN`?

test/platform/test_linux_amd64_compatibility.sh
```diff
@@ -0,0 +1,98 @@
...
+)
+
+FAILED=()
+
+SCRIPTPATH="$( cd "$(dirname "$0")" ; pwd -P )"
+PLATFORM=$($SCRIPTPATH/../../scripts/osarchtype.sh)
+
+if [ "${PLATFORM}" != "linux/amd64" ]
+then
+    echo "The test_linux_amd64_compatibility.sh script is intended to support local execution only on linux/x86-64 machines."
```
btoll: Do you mind using the color vars such as `RED_FG` here?  I find it makes it much easier to spot in any log files.

PR #653, opened 2019-12-19
Fixing arm64 environment issues

algonautshant:
## Summary
1) python3-venv libffi-dev libssl-dev
libffi-dev (and libssl-dev) are needed by the cryptography package builder for python in e2e_basic_start_stop.

2) exporting GOPATHBIN needed to run algotmpl in template e2e tests.

## Test Plan
Run in travis

### DIFF
--- a/scripts/travis/integration_test.sh
+++ b/scripts/travis/integration_test.sh
@@ -14,10 +14,14 @@ export BUILD_TYPE="integration"
 if [ "${USER}" = "travis" ]; then
     # we're running on a travis machine
     "${SCRIPTPATH}/build.sh" --make_debug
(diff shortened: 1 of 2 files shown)


algobolson:
LGTM

scripts/travis/integration_test.sh
```diff
@@ -14,11 +14,15 @@ export BUILD_TYPE="integration"
 if [ "${USER}" = "travis" ]; then
     # we're running on a travis machine
     "${SCRIPTPATH}/build.sh" --make_debug
+    GOPATHBIN=$(go env GOPATH)/bin
```
pzbitskiy: it is interesting why it is not set on ARM.

PR #654, opened 2019-12-19
Fix concurrent access to wallet handles cache in goal

pzbitskiy:
## Summary

* In rare cases (i.e. e2e tests run in parallel on the same network)
a race cond happens when accessing goal.cache/walletHandles.json file
* Introduce advisory locking on the mentioned file
* Implementation is extendable by implementing *locker* interface
  for specific platform and providing a new *newLockedFile* constructor.

## Test Plan

Low impact, rely on existing tests.

### DIFF
--- a/libgoal/lockedFile.go
+++ b/libgoal/lockedFile.go
@@ -59,7 +59,7 @@ func (f *unixLocker) unlock(fd *os.File) error {
 // and supposed to be platform-agnostic with appropriate locker implementation.
 // Each platform needs own specific `newLockedFile`
 const maxRepeats = 10
-const sleepInterval = 10 * time.Microsecond
+const sleepInterval = 10 * time.Millisecond
 
 type lockedFile struct {
 	path   string
@@ -67,10 +67,10 @@ type lockedFile struct {
 }
 
 func newLockedFile(path string) *lockedFile {
-	f := new(lockedFile)
-	f.path = path
-	f.locker = new(unixLocker)
-	return f
+	return &lockedFile{
+		path:   path,
+		locker: &unixLocker{},
+	}
 }
 
 func (f *lockedFile) read() ([]byte, error) {
@@ -91,7 +91,7 @@ func (f *lockedFile) read() ([]byte, error) {
 }
 
 func (f *lockedFile) write(data []byte, perm os.FileMode) error {
-	fd, err := os.OpenFile(f.path, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, perm)
+	fd, err := os.OpenFile(f.path, os.O_WRONLY|os.O_CREATE, perm)
 	if err != nil {
 		return err
 	}
@@ -104,6 +104,7 @@ func (f *lockedFile) write(data []byte, perm os.FileMode) error {
 	}
 	defer f.locker.unlock(fd)
 
+	fd.Truncate(0)
 	_, err = fd.Write(data)
 	return err
 }
--- a/libgoal/lockedFile.go
+++ b/libgoal/lockedFile.go
@@ -1,0 +1,120 @@
+// Copyright (C) 2019 Algorand, Inc.
+// This file is part of go-algorand
+//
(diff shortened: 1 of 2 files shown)


libgoal/lockedFile.go
```diff
@@ -0,0 +1,120 @@
...
+
+type lockedFile struct {
+	path   string
+	locker locker
+}
+
+func newLockedFile(path string) *lockedFile {
+	f := new(lockedFile)
+	f.path = path
+	f.locker = new(unixLocker)
```
tsachiherman: I think it could be written better..
maybe
```
return &lockedFile{
    path: path,
    locker: unixLocker{},
}
```

libgoal/lockedFile.go
```diff
@@ -0,0 +1,120 @@
...
+	if err != nil {
+		return fmt.Errorf("Can't acquire lock for %s: %s", f.path, err.Error())
+	}
+	defer f.locker.unlock(fd)
+
+	_, err = fd.Write(data)
+	return err
+}
+
+func attemptLock(lockFunc func() error) error {
```
tsachiherman: make it a member of lockedFile and avoid passing arguments.

libgoal/lockedFile.go
```diff
@@ -0,0 +1,120 @@
...
+func (f *unixLocker) unlock(fd *os.File) error {
+	return syscall.Flock(int(fd.Fd()), syscall.LOCK_UN)
+}
+
+// lockedFile implementation
+// It uses non-blocking acquisition with repeats
+// and supposed to be platform-agnostic with appropriate locker implementation.
+// Each platform needs own specific `newLockedFile`
+const maxRepeats = 10
+const sleepInterval = 10 * time.Microsecond
```
tsachiherman: you'll need much higher numbers to support network storage. maybe 10 ms, over 10 iterations ?

libgoal/lockedFile.go
```diff
@@ -0,0 +1,120 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("Can't acquire lock for %s: %s", f.path, err.Error())
+	}
+	defer f.locker.unlock(fd)
+
+	return ioutil.ReadAll(fd)
+}
+
+func (f *lockedFile) write(data []byte, perm os.FileMode) error {
+	fd, err := os.OpenFile(f.path, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, perm)
```
algobolson: Doesn't this truncate the file immediately, before checking/taking the lock?

libgoal/lockedFile.go
algobolson: maybe meant 10 Millisecond? It might be my old brain thinking in terms of spinning rust, but I don't think of filesystem operations as happening in much less than 1 millisecond.

pzbitskiy: Done, you are right, the delay is too small, SSD read/write take a bit longer than 100 mcs.

libgoal/lockedFile.go
pzbitskiy: Done, good catch!

libgoal/lockedFile.go
pzbitskiy: Done

libgoal/lockedFile.go
pzbitskiy: Done

libgoal/lockedFile.go
pzbitskiy: Disagree, this is temp runtime info, not a part of the object state (although it is discussable in this particular case)

pzbitskiy:
Rebased and fixed remarks.

algobolson:
Good enough for me. I'll let Tsachi have the last word on if he really wants that one little change.

libgoal/lockedFile.go
tsachiherman: I think that it makes it over-complicated and hurts readability of this code.. but no logical issues here, so I'll merge that in.

PR #660, opened 2019-12-22
Shant/catchup stop on unapproved

algonautshant:
## Summary
The catchup service will save the round when an an-approved protocol
update will take place. Then, before creating a task to fetch a round,
will check if the next round is when an an-approved protocol round
begins, and stops the catchup service.

The ledger should have the round with NextProtocolSwitchOn to stop
the un-approved round from getting fetched.

## Test Plan

The added test covers the edge cases which may or may not happen
when the service runs.


catchup/service.go
```diff
@@ -172,7 +173,7 @@ func (s *Service) fetchAndWrite(fetcher rpcs.Fetcher, r basics.Round, prevFetchC
 
 		// Stop retrying after a while.
 		if i > catchupRetryLimit {
-			s.log.Errorf("fetchAndWrite(%v): failed to fetch block many times", )
+			s.log.Errorf("fetchAndWrite(%v): failed to fetch block many times")
```
tsachiherman: The formatting here is wrong; what's about
"fetchAndWrite: block retrieval exceeded retry limit"

catchup/service.go
```diff
@@ -464,3 +491,55 @@ func (s *Service) sync() {
...
 }
+
+// nextRoundIsNotApproved returns true if the next round upgrades to a protocol version
+// which is not approved.
+// In case of an error, it returns false
+func (s *Service) nextRoundIsNotApproved(nextRound basics.Round) bool {
+	lastLedgerRound := s.ledger.LastRound()
+	proto, err := s.ledger.ConsensusParams(lastLedgerRound)
+	if err != nil {
+		s.log.Errorf("catchup: could not get consensus parameters for round %v: $%v", lastLedgerRound, err)
```
tsachiherman: few requests -
1. replace the `catchup` with `nextRoundIsNotApproved`
2. when formatting the round, use `%d`
3. don't prefix the error with `$`

catchup/service.go
```diff
@@ -464,3 +491,55 @@ func (s *Service) sync() {
...
+		s.log.Errorf("catchup: could not determine if next round is from an approved protocol.")
```
tsachiherman: I don't think that this error logging is needed. Generating multiple errors for the same issue isn't very helpful, and creates redundant verbosity.

catchup/service.go
```diff
@@ -464,3 +491,55 @@ func (s *Service) sync() {
...
+		return false
+	}
+	approvedUpgrades := proto.ApprovedUpgrades
+
+	block, error := s.ledger.Block(lastLedgerRound)
+	if error != nil {
+		s.log.Errorf("catchup: could not last block (%v) from the ledger.", lastLedgerRound)
```
tsachiherman: few notes:
1. replace the `catchup` with `nextRoundIsNotApproved`
2. fix the error message : "could not retrieve last block (%d) from ledger"

catchup/service.go
```diff
@@ -464,3 +491,55 @@ func (s *Service) sync() {
...
+		nextRound >= bh.NextProtocolSwitchOn &&
+		!isAnApprovedUpgrade {
+		return true
+	}
+	return false
+}
+
+// handleUnapprovedRound records the last approved round number.
+// Checks if the last approved round was added to the ledger, and stops the service.
+func (s *Service) handleUnapprovedRound(nextUnapprovedRound basics.Round) {
```
tsachiherman: The function is correct, but only if the following is true:
The function expect to receive a round number r which was verified to be unsupported, while the round r-1 was verified to be supported.
We should add this to the comment above the function, as this is a prerequisite for the functional correctness.

catchup/service.go
```diff
@@ -464,3 +491,55 @@ func (s *Service) sync() {
...
+
+	// Save the last approved round
+	if s.lastApprovedRound == 0 || (nextUnapprovedRound-1) < s.lastApprovedRound {
+		s.lastApprovedRound = nextUnapprovedRound - 1
+	}
+
+	s.log.Infof("Catchup Service: round %v is not approved. Service will stop once the last approved round is added to the ledger.",
```
tsachiherman: change '%v' -> '%d'

catchup/service.go
```diff
@@ -464,3 +491,55 @@ func (s *Service) sync() {
...
+		s.lastApprovedRound)
+
+	// If the next round is an unapproved round, need to stop the
+	// catchup service. Should stop after the last approved round
+	// is added to the ledger.
+	lr := s.ledger.LastRound()
+	if lr == s.lastApprovedRound {
+		s.log.Infof("Catchup Service: finished catching up to the last approved round %v. The subsequent rounds are not approved. Service is stopping.",
```
tsachiherman: '%v' -> '%d'

catchup/service.go
algonautshant: Will drop the second error log.

catchup/service.go
algonautshant: dropping the second error log to reduce verbosity: 
could not determine if next round is from an approved protocol.

catchup/service.go
algonautshant: This part of the code is confusing, and arguably wrong. 
I am moving the statement that sets s.lastApprovedRound to nextRoundIsNotApproved, and get the value directly from NextProtocolSwitchOn.  

> The function expect to receive a round number r which was verified to be unsupported, while the round r-1 was verified to be supported.

This is not true. It could be that round r-1 is not supported:
Hypothetically, when not supporting beginning at round 1, round 1 and round 2 will be fetched in parallel, and will know that round 1 and 2 are not supported after fetching round 1. Hence, only after round 1 is fetched, handleUnapprovedRound may be called on (round 3). In which case, round 2 (r-1) is not supported.

PR #662, opened 2019-12-26
Fail on algod startup if logging.config file has bad permissions

pzbitskiy:
* This should prevent telemetry event loses on systems with
  invalid permissions on ~/.algorand/logging.config file
* Another possible workaround is to relax default config path mask in
  **cmd/goal/commands.go:ensureCacheDir** from 700 to 744.
  This is not implemented because of possible security risk.

### DIFF
--- a/cmd/algod/main.go
+++ b/cmd/algod/main.go
@@ -161,6 +161,10 @@ func main() {
 		if err != nil {
 			fmt.Fprintln(os.Stdout, "error loading telemetry config", err)
 		}
+		if os.IsPermission(err) {
+			fmt.Fprintf(os.Stderr, "Permission error on accessing telemetry config: %v", err)
+			os.Exit(1)
+		}
 
 		// Apply telemetry override.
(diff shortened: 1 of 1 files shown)


cmd/algod/main.go
```diff
@@ -161,6 +161,10 @@ func main() {
 		if err != nil {
 			fmt.Fprintln(os.Stdout, "error loading telemetry config", err)
 		}
+		if os.IsPermission(err) {
+			fmt.Fprintf(os.Stdout, "Permission error on accessing telemetry config: %v", err)
```
tsachiherman: If we're going to exit the process here, than we should make it into a `os.Stderr`.

tsachiherman: I would also suggest that you'd add some more details here to explain the user which directories and/or files are attempted to be accessed, as I'm not sure if these would be included in the error message.

pzbitskiy:
It is included: `Permission error on accessing telemetry config: open
/home/username/.algorand/logging.config: permission denied`

PR #664, opened 2019-12-26
Local telemetry

algobolson:
log to local file by setting logging.config:
{"URI":"file:///tmp/foo"}


winder:
I'd rather not add any more options to the telemetry code. Instead we should always include telemetry in the logfile, this has been requested many times and I don't see any good reason not to go for it.

The change would be smaller too, I think you just need to add `entry.Info(message)` to the end of `logTelemetry`

PR #665, opened 2019-12-26
Reduce execution time of expect tests

pzbitskiy:
* CombinedOutput blocks on copying empty stderr stream from expect that
      causes at least 60 sec timeout for most of the tests
    * This implementation uses a temp time for stderr accumulation.
      In this case exec.Cmd does not run goroutines for reading child's actual stderr.
    * 655 sec (before) vs 205 sec (after)


egieseke:
This page provides some alternatives to CombinedOutput that also capture stderr and stdout.  https://blog.kowalczyk.info/article/wOYk/advanced-command-execution-in-go-with-osexec.html

pzbitskiy:
Dump **stderr** to a temp file so exec.Cmd does not run goroutine and does not block in `io.Copy`.

PR #666, opened 2019-12-27
telemetry recorded locally as info log

algobolson:
logging.config:
{"Enable":false,"SendToLog":true}
OR:
config.json:
{"TelemetryToLog":true}

### DIFF
--- a/cmd/algod/main.go
+++ b/cmd/algod/main.go
@@ -165,7 +165,7 @@ func main() {
 		// Apply telemetry override.
 		telemetryConfig.Enable = logging.TelemetryOverride(*telemetryOverride)
 
-		if telemetryConfig.Enable {
+		if telemetryConfig.Enable || telemetryConfig.SendToLog {
 			// If session GUID specified, use it.
 			if *sessionGUID != "" {
 				if len(*sessionGUID) == 36 {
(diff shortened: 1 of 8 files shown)


tsachiherman:
I like the idea here, and think it's adds a lot. My only push back here is that the `Enable` in logging.config applies to all the nodes on the host, whereas the log file is per node. That would make the more-suitable place for per-node configuration to be in the config.json.
Alternatively, I think that if we do want to log all our telemetry entries, it doesn't need to go into the same per-node log file.

algobolson:
added local config bool TelemetryToLog. Kept `SendToLog` in the telemetry config struct because that made it easier to get that value to the right place in code. algod main() handles loading both configs and setup.

winder:
This logic looks fine, but I think we should just assume TelemetryToLog=true rather than have it be configurable. I'm really not seeing a compelling reason to make this configurable.

tsachiherman:
I think that this change is good. By default, I wouldn't want to send all telemetry to log, as it might fill up the log file too quickly. The immediate value I see here is the ability to debug what would go to the telemetry server without looking on the telemetry server.

algobolson:
In some recent development using this I see several node.log files with about 1500 debug level statements, 1800 info level statements, and 200 telemetry statements (which are logged at info level). I think that potentially a reasonable percentage to turn on all the time.

algobolson:
I think this is ready to go now

PR #668, opened 2019-12-28
Update network disconnection logic

tsachiherman:
## Active Outgoing Network Peer Disconnection

Gossip networks connections in previous releases were randomly selected, stable connections.
That approach has its advantages, but is also susceptible to two major problems : inefficient connections and clique formation.

This PR attempt to address both of the above issues. In this PR, we add an "attribute" to each outgoing connection called "throttled connection". A throttled connection is a connection who's performance is being tested continuously. The performance comparison is always done over the set of outgoing connections.

Once we complete the performance test, we disconnect the throttled connection who performed the worse. That would allow us to continuously attempt to "improve" our outgoing connections pool for the fastest relays.

Unfortunately, the above would also expedite the formation of cliques. Prior to this PR, our response for cliques formation was to restart the node.
In this PR, we have two separate mechanisms to handle cliques:
1. Avoidance; When creating a relay node, we configure only half of it's outgoing connections to be throttled connections. That reduce the likelihood of clique formation. For a non-relay, that's not an issue.
2. Fix existing clique; This PR add a dedicated feedback between the agreement service and the network layer. The network layer is exposing a watchdog-style handler that is being invoked by the agreement service. That allows the network library to determine if it's unable to make progress. When that happens, a random outgoing connection is being disconnected.

## Testing
This change was tested using a s2 network, and no performance regression was observed.


network/wsNetwork.go
```diff
@@ -1264,6 +1295,123 @@ func (wn *WebsocketNetwork) meshThread() {
...
+	if len(outgoingPeers) == 0 {
+		return false
+	}
+	if wn.numOutgoingPending() > 0 {
+		// we're currently trying to extend the list of outgoing connections. no need to
+		// disconnect any existing connection to free up room for another connection.
+		return false
+	}
+	var peer *wsPeer
+	disconnectPeerIdx := crypto.RandUint63() % uint64(len(outgoingPeers))
```
pzbitskiy: GossipFanout = 4 by default. In case of sporadic disconnect of some other node we have quite high probability of waiting `2 x cliqueResolveInterval` on partitioned network

network/wsNetwork.go
```diff
@@ -1591,7 +1740,25 @@ func (wn *WebsocketNetwork) tryConnect(addr, gossipAddr string) {
 		return
 	}
 
-	peer := &wsPeer{wsPeerCore: wsPeerCore{net: wn, rootURL: addr}, conn: conn, outgoing: true, incomingMsgFilter: wn.incomingMsgFilter, createTime: time.Now()}
+	throttledConnection := false
```
pzbitskiy: wonder if there is a code pass when a connection is dropped but removePeer is not called so that throttledOutgoingConnections counter can be screwed

pzbitskiy: seems like we are clear

network/wsNetwork.go
```diff
@@ -643,6 +663,10 @@ func (wn *WebsocketNetwork) Start() {
 		// wrap the limited connection listener with a requests tracker listener
 		wn.listener = wn.requestsTracker.Listener(listener)
 		wn.log.Debugf("listening on %s", wn.listener.Addr().String())
+		wn.throttledOutgoingConnections = int32(wn.config.GossipFanout / 2)
+	} else {
```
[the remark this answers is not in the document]
tsachiherman: yes.

network/wsNetwork.go
tsachiherman: Relays are expected to be relatively reliable, so there shouldn't be too many sporadic disconnection. The waiting 2 x cliqueResolveInterval is the fallback for agreement protocol not making any progress. Otherwise, the number of outgoing connections would be examined once a minute.

network/wsNetwork.go
tsachiherman: It would be great if you can find such. If you do, please let me know, since it also means that we didn't remove the entry from the peers list.

network/connPerfMon.go
```diff
@@ -0,0 +1,351 @@
...
+// advanceStage set the stage variable and update the stage start time.
+func (pm *connectionPerformanceMonitor) advanceStage(newStage pmStage, now int64) {
+	pm.stage = newStage
+	pm.stageStartTime = now
+}
+
+// updateMessageIdlingInterval updates the last message received timestamps and determines how long it has been since
+// the last message was received on any of the incoming peers
+func (pm *connectionPerformanceMonitor) updateMessageIdlingInterval(now int64) (minMsgInterval int64) {
+	currentIncomingMsgTime := pm.lastIncomingMsgTime
```
pzbitskiy: should not lastIncomingMsgTime be atomic?

tsachiherman: All access to lastIncomingMsgTime is behind the connectionPerformanceMonitor.lock, so I don't think it's needed.

network/wsPeer.go
```diff
@@ -148,6 +150,17 @@ type wsPeer struct {
 
 	// createTime is the time at which the connection was established with the peer.
 	createTime time.Time
+	// connMonitor used to measure the relative performance of the connection
+	// compared to the other outgoing connections. Incoming connections would have this
+	// field set to nil.
+	connMonitor *connectionPerformanceMonitor
+
+	// peerMessageDelay is calculated by the connection monitor; it's the relative avarage per-message delay.
+	peerMessageDelay int64
```
algobolson: isn't the above `lastPingRoundTripTime` already what we want to measure?

network/wsNetwork.go
```diff
@@ -1180,6 +1217,7 @@ func (wn *WebsocketNetwork) connectedForIP(host string) (totalConnections int) {
 }
 
 const meshThreadInterval = time.Minute
+const cliqueResolveInterval = 5 * time.Minute
```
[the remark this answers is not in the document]
tsachiherman: I'm not strongly opinionated regarding "how to express this constant". Only that "it needs to be a substantial amount of time without making any agreement progress". 
Using the SmallLambda implies that it would change with protocol upgrades, which seems to me like over engineering. I wouldn't mind thought moving this one into the config file, so that we could tweak this value over there.

node/impls.go
```diff
@@ -153,17 +153,26 @@ type agreementLedger struct {
 // EnsureBlock implements agreement.LedgerWriter.EnsureBlock.
 func (l agreementLedger) EnsureBlock(e bookkeeping.Block, c agreement.Certificate) {
 	l.Ledger.EnsureBlock(&e, c)
+	// let the network know that we've made some progress.
+	l.n.OnNetworkAdvance()
```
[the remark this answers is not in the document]
tsachiherman: No. If the agreement can't work due to gossip network issue, we don't want the catchup service to become the source of the new blocks. Instead, we want to disconnect, and hopefully, it would help fixing the gossip network.

network/wsPeer.go
tsachiherman: They measure different "creatures". One is the pure pinging time and one is the average relative message latency. Both are meaningful and important, and I think that we should add the ping time as well. (maybe in the next PR, just to reduce the number of changes here ? )

node/impls.go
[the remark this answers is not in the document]
tsachiherman: If the OnNewBlock is called *only* as a result of the agreement protocol advancements ( all the three ), then yes, it would be a good think to do. ( but indirectly; The node.go should be node that bridge the gap )

network/wsNetwork.go
```diff
@@ -1272,27 +1315,20 @@ func (wn *WebsocketNetwork) meshThread() {
...
-					wn.wg.Add(1)
-					go wn.tryConnect(na, gossipAddr)
-					need--
-					if need == 0 {
-						break
-					}
-				}
+
+		// as long as the call to checkExistingConnectionsNeedDisconnecting is deleting existing connections, we want to
+		// kick off the creation fo new connections.
```
onetechnical:
```suggestion
		// kick off the creation of new connections.
```

tsachiherman:
> Only grammatical fixes; these aren't PR blocking, are they?

no; it's not blocking, but since you've already identified these, I'll fix these before merging it in.

thanks!

PR #669, opened 2019-12-30
Avoid upgrading boost on travis Mac builds

tsachiherman:
## Summary

Boost version has changed, causing a new python3 to be installed. While there is nothing wrong here, the python3 symbolic link was not pointing to the newly installed due to some cask installation issue.

This change pin the boost version to the one we used to have, so that we can take and figure this issue offline.


fabrice102:
It looks like the bug will appear for any `brew` dependency that depends on `python` (see #891). A longer term solution might be https://support.circleci.com/hc/en-us/articles/360043542933--Error-The-brew-link-step-did-not-complete-successfully-When-Installing-Python

PR #670, opened 2019-12-31
Standalone update.sh

tsachiherman:
## Summary

When installing a node using the update script, one need to bootstrap his system with the updater tool. Currently, we suggest that the user would go to either one of the following:
* https://github.com/algorand/go-algorand-doc/tree/master/downloads/installers/darwin_amd64
* https://github.com/algorand/go-algorand-doc/tree/master/downloads/installers/linux_amd64
* https://github.com/algorand/go-algorand-doc/tree/master/downloads/installers/linux_arm

Extract the tarball, and run the update.sh bash script.

While there is nothing wrong with doing so, we could do a bit better - 
Regardless of what platform we're on, download 
* https://github.com/algorand/go-algorand-doc/tree/master/downloads/installers/update.sh

and run it.
It will automatically detect the current platform and download the corresponding gz.tar from GitHub. Than, it would proceed as if it was with the previous call.

### DIFF
--- a/cmd/updater/update.sh
+++ b/cmd/updater/update.sh
@@ -138,8 +138,88 @@ function determine_current_version() {
     echo Current Version = ${CURRENTVER}
 }
 
+function get_updater_url() {
+    local UNAME
+    local OS
+    local ARCH
+    UNAME=$(uname)
+    if [[ "${UNAME}" = "Darwin" ]]; then
+        OS="darwin"
+        UNAME=$(uname -m)
+        if [[ "${UNAME}" = "x86_64" ]]; then
+            ARCH="amd64"
+        else
+            echo "This platform ${UNAME} is not supported by updater."
+            exit 1
+        fi
+    elif [[ "${UNAME}" = "Linux" ]]; then
+        OS="linux"
+        UNAME=$(uname -m)
+        if [[ "${UNAME}" = "x86_64" ]]; then
+            ARCH="amd64"
+        elif [[ "${UNAME}" = "armv6l" ]]; then
+            ARCH="arm"
+        elif [[ "${UNAME}" = "armv7l" ]]; then
+            ARCH="arm"
+        elif [[ "${UNAME}" = "aarch64" ]]; then
+            ARCH="arm64"
+        else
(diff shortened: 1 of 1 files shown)


winder:
I think it would be nice to move the installer binaries to S3. I think `go-algorand-docs` is deprecated at this point, so we should probably start transitioning away from it. Also serving binaries from github isn't a great practice.

tsachiherman:
> I think it would be nice to move the installer binaries to S3. I think `go-algorand-docs` is deprecated at this point, so we should probably start transitioning away from it. Also serving binaries from github isn't a great practice.

Not yet. There are few big changes that are coming to the documentation web site. It will force us to make some changes on our end as well. Once all these changes are over, and we're done with the process, we could revise the `go-algorand-docs` storage.

cmd/updater/update.sh
```diff
@@ -138,8 +138,88 @@ function determine_current_version() {
     echo Current Version = ${CURRENTVER}
 }
 
+function get_updater_url() {
```
pzbitskiy: Add a comment that the function sets UPDATER_FILENAME and UPDATER_URL

cmd/updater/update.sh
```diff
@@ -138,8 +138,88 @@ function determine_current_version() {
...
+        # no curl is installed.
+        echo "updater binary is missing and cannot be downloaded since curl is missing."
+        if [[ "$(uname)" = "Linux" ]]; then
+            echo "To install curl, run the following command:"
+            echo "apt-get update; apt-get install -y curl"
+        fi
+        exit 1
+    fi
+
+    CURL_OUT=$(curl -LJO --silent ${UPDATER_URL})
```
pzbitskiy: I confused. 
1. -J tells curl to use Content-Disposition for output file name instead of URL name. But after we refer the file as UPDATER_FILENAME from URL. Can't Content-Disposition and actual file name be different?
2. curl drops file into current directory but we looking the file at "${SCRIPTPATH}/${UPDATER_FILENAME}" I do not see where SCRIPTPATH is set as CWD.

PR #684, opened 2020-01-03
Fix - Indexer now shows received transactions

rotemh:


data/transactions/transaction.go
```diff
@@ -412,6 +412,18 @@ func (tx Transaction) TxAmount() basics.MicroAlgos {
 	}
 }
 
+// GetReceiverAddress returns the address of the receiver. If the transaction has no receiver, it returns the empty address.
+func (tx Transaction) GetReceiverAddress() basics.Address {
+	switch tx.Type {
+	case protocol.PaymentTx:
+		return tx.PaymentTxnFields.Receiver
+	case protocol.AssetTransferTx:
```
algobolson: should the 'Receiver' be the clawback address in case of clawback transaction?

rotemh: I think that only the sender changes. unless I'm missing something.

PR #685, opened 2020-01-03
Catchup service stop on unsupported and e2e test 

algonautshant:
## Summary
In this change:
Updated catchup service to stop on unsupported and not on un-upgradable.
Added e2e test for stopped catchup service on unsupported protocol.

goal changes related to this are in PR: https://github.com/algorand/go-algorand/pull/686

## Test Plan

The changes are tested with a unit test and e2e test.

PR #686, opened 2020-01-03
Updates to the goal node status

algonautshant:
## Summary
This change is splitting the goal section from PR: https://github.com/algorand/go-algorand/pull/685
Updated goal node status to inform when the catchup service is stopped.
Updated goal node status by removing "Synced Since Startup" field.
Added field StoppedAtUnsupportedRound to v1.NodeStatus in  daemon/algod/api/spec/v1/model.go 
Added test to check libgoal Client StoppedAtUnsupportedRound in v1.NodeStatus


cmd/goal/node.go
```diff
@@ -290,7 +290,22 @@ func getStatus(dataDir string) {
...
+		infoNodeStatus,
+		stat.LastRound,
+		lastRoundTime,
+		catchupTime,
+		stat.LastVersion,
+		stat.NextVersion,
+		stat.NextVersionRound,
+		stat.NextVersionSupported)
+
+	if stat.LastRound+1 == stat.NextVersionRound &&
```
tsachiherman: That logic should not be here; to be more specific, the predicate
`stat.LastRound+1 == stat.NextVersionRound && !stat.NextVersionSupported` should be a bool field in the `v1.NodeStatus` struct that indicates whether the node is capable of making a progress or not.
It's the `AlgorandFullNode::Status()` that should fill up this value, and have it propagated all the way upstream.

PR #687, opened 2020-01-04
Improve lockFile error handling

tsachiherman:
## Solution

The file locking used wasn't consistently working in our testing environments. There were 2 reasons why this didn't work:
1. We have used the non-blocking calls, and allocated certain time for the lock to be acquired. That time might have surpassed, causing us to fail.
2. The lock mechanism used would not support NFS. This has not proved to be the issue, but could cause issues down the road.

This PR is doing the following:
1. Replace the `Flock` with `FcntlFloc`, which supports NFS.
2. Remove non-blocking mode, as it's not required functionality and is prone to timeout issues.

### DIFF
--- a/libgoal/lockedFile.go
+++ b/libgoal/lockedFile.go
@@ -53,12 +53,12 @@ type unixLocker struct {
 func makeUnixLocker() *unixLocker {
 	locker := &unixLocker{}
 	getlk := syscall.Flock_t{Type: syscall.F_RDLCK}
-	if err := syscall.FcntlFlock(0, 37 /*F_OFD_GETLK*/, &getlk); err == nil {
+	if err := syscall.FcntlFlock(0, 36 /*F_OFD_GETLK*/, &getlk); err == nil {
 		locker.ofd = true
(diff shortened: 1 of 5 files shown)


libgoal/lockedFile.go
```diff
@@ -40,26 +40,68 @@ type locker interface {
 }
 
 type unixLocker struct {
+	ofd         bool
+	setLock     int
```
pzbitskiy: ofd, setLock and getLock are not used

libgoal/lockedFile.go
```diff
@@ -40,26 +40,68 @@ type locker interface {
...
+	setLockWait int
+	getLock     int
 }
 
+// makeUnixLocker create a unix file locker.
+// note that the desired way is to use the OFD locker, which locks on the file descriptor level.
+// falling back to the non-OFD lock would allow obtaining two locks by the same process. If this becomes
+// and issue, we might want to use flock, which wouldn't work across NFS.
+func makeUnixLocker() *unixLocker {
```
pzbitskiy: shouldn't we provide Linux and Mac specific implementations?

PR #690, opened 2020-01-06
Initial commit - move scripts into new release/ dir

btoll:
See the README


scripts/release/build.sh
```diff
@@ -0,0 +1,108 @@
...
+EOF
+# strip leading 'export ' for docker --env-file
+sed 's/^export //g' < "${HOME}"/build_env > "${HOME}"/build_env_docker
+
+export GOPATH=${HOME}/go
+export PATH=${HOME}/gpgbin:${GOPATH}/bin:/usr/local/go/bin:${PATH}
+
+# Build!
+scripts/configure_dev.sh
+make crypto/lib/libsodium.a
```
tsachiherman: why ? doesn't make build creates that automatically ?

scripts/release/package.sh
```diff
@@ -0,0 +1,78 @@
+#!/usr/bin/env bash
+# shellcheck disable=2164,2166
+
```
tsachiherman: I think that you're missing `set -e` here.

scripts/release/socket.sh
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env bash
+
```
tsachiherman: maybe add `set -e` ?

tsachiherman: also, in the case of this script you also want set the option so that and error on one pipe would break the entire command.

scripts/release/Jenkinsfile
```diff
@@ -0,0 +1,99 @@
...
+    stages {
+        stage("create ec2 instance") {
+            steps {
+                sh script: 'scripts/release/start_ec2_instance.sh us-west-1 ami-0dd655843c87b6930 t2.2xlarge'
+            }
+        }
+
+        stage("setup ec2 instance") {
+            steps {
+                sh 'aws s3 cp s3://algorand-devops-misc/tools/gnupg2.2.9_centos7_amd64.tar.bz2 .'
```
tsachiherman: it's ok for now, but I would prefer to move this away from S3 and into github.
( like the reminder of our build sources ).

scripts/release/setup.sh
```diff
@@ -0,0 +1,128 @@
...
+sudo apt-get update -q
+sudo apt-get upgrade -q -y
+
+# Some of these dirs aren't used until later scripts.
+#umask 0077
+#mkdir -p ~/.gnupg
+#mkdir -p "${HOME}"/{.gnupg,go,gpgbin,docker_test_resources,dummyaptly,dummyrepo,prodrepo}
+mkdir -p "${HOME}"/{.gnupg,go,gpgbin,dummyaptly,dummyrepo,prodrepo}
+
+export GOPATH=${HOME}/go
```
tsachiherman: is there any particular reason you're not using 
```
GOPATH=$(go env GOPATH)
export GOPATH
```
?

scripts/release/build.sh
btoll: I'll have to check with Brian to see if there's a reason to build this.

scripts/release/socket.sh
btoll: I'll add the `set -e`.  I'm afraid I didn't understand your second comment.

scripts/release/Jenkinsfile
btoll: Ok, I'll either move this iteration or make a note to do so.

scripts/release/setup.sh
btoll: I remember at one point that it couldn't find the go binary and so I manually set it to keep moving.  I'll revisit this.

scripts/release/package.sh
btoll: Will set.

scripts/release/Jenkinsfile
algobolson: I built that compilation of gpg for centos. Centos7 ships with a version that is too old. I statically compiled a gpg that we can unpack into Centos7 and run. S3 is the right place for that archive.

scripts/release/Jenkinsfile
```diff
@@ -0,0 +1,104 @@
...
+        AWS_ACCESS_KEY_ID = credentials("aws-access-key-id")
+        AWS_SECRET_ACCESS_KEY = credentials("aws-secret-access-key")
+    }
+
+    agent any
+
+    stages {
+        stage("create ec2 instance") {
+            steps {
+                sh script: 'scripts/release/start_ec2_instance.sh us-west-1 ami-0dd655843c87b6930 t2.2xlarge'
```
algobolson: what is that ami? comment how it was picked and when? Like, it's the "us-west-1 ubuntu 18.04 as of 2019-12-27" or so.

scripts/release/README.md
```diff
@@ -0,0 +1,61 @@
...
+
+In addition, the build logs will be placed into the AWS `algorand-devops-misc` S3 bucket under `buildlog`.
+
+## Notes
+
+All of the `aws ...` commands are now executed by Jenkins and are defined in the `Jenkinsfile`.  The reason for this is simple:  Jenkins already has the AWS auth credentials, and we don't want or need to be sending them anywhere else in the cloud.
+
+## TODO
+
+Create the git tag.
```
algobolson: Disagree. I think tagging should remain a manual process. Then the build system picks that up. We decide "this is 2.1.42" and then the build system builds it.

btoll: Sure, I have no problem with that.

scripts/release/Jenkinsfile
btoll: I'll add a comment.

scripts/release/Jenkinsfile
```diff
@@ -0,0 +1,104 @@
...
+                    sh 'ssh -i BuilderInstanceKey.pem -A ubuntu@$(cat scripts/release/tmp/instance) bash go/src/github.com/algorand/go-algorand/scripts/release/upload.sh'
+                    // sh "aws s3 cp --quiet ubuntu@\$(cat scripts/release/tmp/instance)/build_status_${CHANNEL}_${FULLVERSION}.asc.gz s3://algorand-devops-misc/buildlog/${RSTAMP}/"
+                    sh "scp -i BuilderInstanceKey.pem -o StrictHostKeyChecking=no ubuntu@\$(cat scripts/release/tmp/instance):~/build_status_${CHANNEL}_*.asc.gz node_pkg/"
+                    sh "aws s3 cp --quiet node_pkg/build_status_${CHANNEL}_*.asc.gz s3://algorand-devops-misc/buildlog/${RSTAMP}/"
+                }
+            }
+        }
+
+        /*
+        stage("tag") {
```
winder: If you want to have a "push button" tag, you can use an "input":
https://jenkins.io/doc/pipeline/steps/pipeline-input-step/

I think that we probably want to reverse this though, and only run the job for tagged commits.

btoll: That make sense.  The workflow can be manually create a tag, then input that to Jenkins to pass to the build agent.

Yes, I'm already using the input step for both pausing the build and getting user input.

scripts/release/build.sh
algobolson: this _shouldn't_ be necessary, here and a couple other places, `make build` _should_ take care of building libsodium, but I guess at one point in the past this worked around a finicky build. Probably we should try without it and see if we don't need it and maybe this time we can take the time to fix it right, or comment here why it's a good workaround

tsachiherman: Brian is correct. At some point in the past it *was* needed.
Today, assuming that you're building on a clean environment every time, there *shouldn't* be a need to do that. ( but please do correct me if I'm wrong )

algobolson:
For each file that was copied scripts/* to scripts/release/... could you delete the original? git will hopefully detect that as a move and show the diff from the old version to the new version along with the move.

btoll:
> For each file that was copied scripts/* to scripts/release/... could you delete the original? git will hopefully detect that as a move and show the diff from the old version to the new version along with the move.

Yes, I wanted to keep them around just in case.  Once everything looks good, I'll remove them.

scripts/release/build.sh
btoll: I've found that if I don't include this right before we make the rpm package that it won't build.

For example, if I comment out the code here:
```
# definitely rebuild libsodium which could link to external C libraries
#if [ -f ${REPO_DIR}/crypto/libsodium-fork/Makefile ]; then
#    (cd ${REPO_DIR}/crypto/libsodium-fork && make distclean)
#fi
#rm -rf ${REPO_DIR}/crypto/lib
#make crypto/lib/libsodium.a

make build

RPMTMP=$(mktemp -d 2>/dev/null || mktemp -d -t "rpmtmp")
trap 'rm -rf ${RPMTMP}' 0
"${REPO_DIR}/scripts/release/helper/build_rpm.sh" "${RPMTMP}"
```
I get the following build errors:
```
crypto/lib/libsodium.a(libsodium_la-utils.o): In function `explicit_bzero':
/usr/include/x86_64-linux-gnu/bits/string_fortified.h:83: undefined reference to `__explicit_bzero_chk'
/usr/include/x86_64-linux-gnu/bits/string_fortified.h:83: undefined reference to `__explicit_bzero_chk'
/usr/include/x86_64-linux-gnu/bits/string_fortified.h:83: undefined reference to `__explicit_bzero_chk'
collect2: error: ld returned 1 exit status
make: *** [buildsrc] Error 2
build_release done packaging centos 20200108_033715
```

algobolson: ok, inside the rpm build inside the centos docker container I know we need the cleanup lines. The outer ubuntu build and the inner centos build are using the same checkout to ensure they have exactly the same source version. (We might be able to get rid of the entire 'centos build' now, we should do testing that current builds built on ubuntu18 run on ubuntu16 and centos7; this should work because of Tsachi's changes to make the build fully static)

The question I had about the build scripts was can we just skip the line `make crypto/lib/libsodium.a` that shows up in a few places.

btoll:
Closing in favor of https://github.com/algorand/go-algorand/pull/698

PR #693, opened 2020-01-07
Disable pprof endpoints by default

justicz:
The `/debug/pprof` endpoints should not be enabled by default since they can leak potentially sensitive information. This PR adds a flag in `config.json` to enable these endpoints.


algobolson:
This seems reasonable to me as "should have always been off" and it's the correct default. Probably only internal developers actually use pprof, and it's okay if we need to turn it back on.

justicz:
Yeah, I'm not sure what a configuration version bump would achieve in this case since we want the default value to be false.

daemon/algod/api/server/router.go
```diff
@@ -117,12 +118,21 @@ func NewRouter(logger logging.Logger, node *node.AlgorandFullNode, shutdown <-ch
...
+		// Registers /debug/pprof handler under root path and under /urlAuth path
+		// to support header or url-provided token.
+		router.PathPrefix(pprofEndpointPrefix).Handler(http.DefaultServeMux)
+
+		urlAuthRouter := router.PathPrefix(urlAuthEndpointPrefix)
+		urlAuthRouter.PathPrefix(pprofEndpointPrefix).Handler(http.DefaultServeMux).Name(debugRouteName)
+	} else {
+		// pprof disabled, route to black hole
+		disabledHandler := func(w http.ResponseWriter, r *http.Request) {
+			http.Error(w, "Profiler disabled", http.StatusForbidden)
```
tsachiherman: I think that the desired behavior here is to return the same output as if "there is nothing there" while logging the incident.

justicz: Sure, we can just not route pprof requests at all, done.

justicz:
Thanks, made that change, let's see if that was the only place it needs to be enabled...

PR #697, opened 2020-01-08
 Two fixes to basicCatchup_test: cloned node not stopped and env var conflict

algonautshant:
## Summary
Fix 1: 
TestBasicCatchup and newly added TestStoppedCatchupOnUnsupported
create a new node by cloning one of the network nodes. When
fixture.Shutdown() stops the original network nodes, leaves the cloned
node running. This change adds function shutDownClonedNode to stop the
cloned nodes.

Fix 2:
In TestStoppedCatchupOnUnsupported, an env variable is used to
delete ConsensusCurrentVersion, so that the cloned node behaves as if
its binary does not support the consensus version. However, when the
TestBasicCatchup runs in parallel, it also picks up the env variable,
and consequently deletes ConsensusCurrentVersion from the Consensus
map. When this happens, TestBasicCatchup sporadically fails.

In this change, instead of having ConsensusTestUnupgradedProtocol
upgrade to ConsensusCurrentVersion, or deleting
ConsensusCurrentVersion so it cannot be upgraded, it sets up
ConsensusTestUnupgradedProtocol to upgrade to
ConsensusTestUnupgradedToProtocol. Hence, the env variable is used to
delete ConsensusTestUnupgradedToProtocol. This way the conflict with
other tests is eliminated.

## Test Plan
This change fixes the test.


test/e2e-go/features/catchup/basicCatchup_test.go
```diff
@@ -164,6 +166,7 @@ func TestStoppedCatchupOnUnsupported(t *testing.T) {
 	a.NoError(err)
 	cloneClient, err := fixture.StartNode(cloneDataDir)
 	a.NoError(err)
+	defer shutdownClonedNode(cloneDataDir, &fixture, t)
```
tsachiherman: looks to me that a better way to do that would be to have fixture.StartNode detect that the given node directory isn't any of the ones it's currently managing, and add it internally.
Then, when fixture.Shutdown is called, you would add it there as well.

The rationale here is that it would save us from duplicating the shutDownClonedNode you've added here.

test/e2e-go/features/catchup/basicCatchup_test.go
```diff
@@ -131,7 +132,7 @@ func TestStoppedCatchupOnUnsupported(t *testing.T) {
 	t.Parallel()
 	a := require.New(t)
 
-	os.Setenv("ALGORAND_TEST_UNUPGRADEDPROTOCOL_DELETE_CURRENT_VERSION", "0")
+	os.Setenv("ALGORAND_TEST_UNUPGRADEDPROTOCOL_DELETE_UPGRADE", "0")
```
tsachiherman: Could you remove this variable from the environment once this test is done ?

test/e2e-go/features/catchup/basicCatchup_test.go
algonautshant: The way things are organized right now makes it difficult to unify the stopping and cleanup of the nodes. 

libgoalFixture.Shutdown calls netdeploy.Network Stop to stop the network. 
netdeploy.Network uses the data structures populated from netdeploy.CreateNetworkFromTemplate to identity which nodes to stop. 

Clone is done from nodecontrol package, outside of netdeploy. Given that the nodes are generated using separate packages, makes unification of their cleanup difficult. 

As we discussed, the alternative is to not clone, but to start the network nodes individually. This is a more promising approach for later.

PR #700, opened 2020-01-08
ledger/eval refactor

algobolson:
Move transaction signature checking out of BlockEvaluator
Txn from gossip or local API are checked on receipt.
Block validation for progress or catchup does signature checking in parallel separately.
Clean up logging in tests to run quieter.


ledger/eval.go
```diff
@@ -697,14 +656,100 @@ func (eval *BlockEvaluator) GenerateBlock() (*ValidatedBlock, error) {
...
+	}
+	return nil
+}
+
+// used by Ledger.Validate() Ledger.AddBlock() Ledger.trackerEvalVerified()(accountUpdates.loadFromDisk())
+//
+// Validate: eval(ctx, blk, true, txcache, executionPool)
+// AddBlock: eval(context.Background(), blk, false, nil, nil)
+// tracker:  eval(context.Background(), blk, false, nil, nil)
 func (l *Ledger) eval(ctx context.Context, blk bookkeeping.Block, validate bool, txcache VerifiedTxnCache, executionPool execpool.BacklogPool) (StateDelta, error) {
```
algobolson: @tsachiherman when this is called from Ledger.Validate() it will use the execution pool passed into Validate() to process the signature checks, same as happened before this refactor.

derbear:
What's the rationale for this refactor?

algobolson:
This refactor is mainly about simplifying a complicated region of code. BlockEvaluator _sometimes_ doing signature checking inside it was confusing and hard to manage. I decided to do this along the way of a different bigger refactor and this will help me do that (AssembleBlock/TransactionPool consolidation of BlockEvaluator usage).

ledger/eval.go
```diff
@@ -697,14 +656,100 @@ func (eval *BlockEvaluator) GenerateBlock() (*ValidatedBlock, error) {
...
+		groupNoAD := make([]transactions.SignedTxn, len(txgroup))
+		for i := range txgroup {
+			groupNoAD[i] = txgroup[i].SignedTxn
+		}
+		ctxs := verify.PrepareContexts(groupNoAD, tv.block.BlockHeader)
+
+		for gi, tx := range txgroup {
+			err := validateTransaction(tx.SignedTxn, tv.block, tv.proto, tv.txcache, ctxs[gi], tv.verificationPool)
+			if err != nil {
+				//logging.Base().Errorf("evalTxValidator reject tx %s", tx.ID())
```
justicz: nit: remove this

ledger/eval.go
```diff
@@ -697,14 +656,100 @@ func (eval *BlockEvaluator) GenerateBlock() (*ValidatedBlock, error) {
...
+				tv.done <- err
+				tv.cf()
```
justicz: can you please name this something more clear? I had to look up the type to know what it was

ledger/eval.go
```diff
@@ -697,14 +656,100 @@ func (eval *BlockEvaluator) GenerateBlock() (*ValidatedBlock, error) {
...
+
+func validateTransaction(txn transactions.SignedTxn, block bookkeeping.Block, proto config.ConsensusParams, txcache VerifiedTxnCache, ctx verify.Context, verificationPool execpool.BacklogPool) error {
+	// Transaction valid (not expired)?
+	err := txn.Txn.Alive(block)
+	if err != nil {
+		return err
+	}
+
+	// Well-formed on its own?
+	err = txn.Txn.WellFormed(ctx.CurrSpecAddrs, proto)
```
justicz: worth noting this gets called in `TxnPool` below

justicz: `WellFormed` does get called in `verify.TxnPool` below. I am not 100% sure yet if that means it's redundant here

ledger/eval.go
```diff
@@ -697,14 +656,100 @@ func (eval *BlockEvaluator) GenerateBlock() (*ValidatedBlock, error) {
...
+		default:
+		}
+		groupNoAD := make([]transactions.SignedTxn, len(txgroup))
+		for i := range txgroup {
+			groupNoAD[i] = txgroup[i].SignedTxn
+		}
+		ctxs := verify.PrepareContexts(groupNoAD, tv.block.BlockHeader)
+
+		for gi, tx := range txgroup {
+			err := validateTransaction(tx.SignedTxn, tv.block, tv.proto, tv.txcache, ctxs[gi], tv.verificationPool)
```
justicz: is it intentional that this is not checking in parallel? (since it waits in lockstep for each transaction to verify)

ledger/eval.go
algobolson: now that you mention it, _I'm_ sure. it could go badly if someone stuck a txn in the cache as Verified without running WellFormed on it, so let's never do that. The other two paths into the TransactionPool call verify.TxnPool() or verify.Txn() and those call WellFormed(), so I think we're currently covered.

ledger/eval.go
algobolson: It's not intentional. It's marginally better than doing it inline in BlockEvaluator (which processes all txns serially). So at least validate and ledger accounting are running in two parallel threads. A future update could run more validates in parallel if that's a useful optimization?

derbear:
As I mentioned in person earlier today, the cleanup looks pretty good. However, it also makes it slightly harder to inspect the code and ensure that all untrusted transactions passed to the block evaluator have had their signatures verified correctly (and have had well-formedness checks pass). It would be nice to have some mechanism (perhaps using the Go type system as in `agreement`) to ensure that only checked transactions (and by transitivity, blocks) make it into proposals/approving votes.

ledger/eval.go
```diff
@@ -394,11 +389,6 @@ func (eval *BlockEvaluator) TestTransactionGroup(txgroup []transactions.SignedTx
 // on a single transaction, but does not actually add the transaction to the block
 // evaluator, or modify the block evaluator state in any other visible way.
 func (eval *BlockEvaluator) testTransaction(txn transactions.SignedTxn, cow *roundCowState) error {
-	// Verify that groups are supported.
```
derbear: What's the reason that we're dropping this check here?

algobolson: it's redundant with a check inside WellFormed()

derbear: Ah, for some reason I missed that.

PR #703, opened 2020-01-09
Avoid starting the Telemetry service when logging is disabled

algobolson:
recent logging changes missed this

### DIFF
--- a/cmd/algoh/main.go
+++ b/cmd/algoh/main.go
@@ -301,7 +301,7 @@ func initTelemetry(genesis bookkeeping.Genesis, log logging.Logger, dataDirector
 				}
 
 				// If the telemetry URI is not set, periodically check SRV records for new telemetry URI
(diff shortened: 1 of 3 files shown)


cmd/algoh/main.go
```diff
@@ -301,7 +301,7 @@ func initTelemetry(genesis bookkeeping.Genesis, log logging.Logger, dataDirector
 				}
 
 				// If the telemetry URI is not set, periodically check SRV records for new telemetry URI
-				if log.GetTelemetryURI() == "" {
+				if telemetryConfig.Enable && log.GetTelemetryURI() == "" {
```
winder: There may be another bug, it looks like we're already in a `if telemetryConfig.Enable` block at this point, maybe that block should be unnested?

algobolson: Yup, that's a redundant check. I suspect algoh was already fine.

PR #708, opened 2020-01-09
More reliable fetcher unit tests

tsachiherman:
## Summary

Existing implementation of fetcher unit tests is too time-sensitive. It might work correctly on a fast machine, but could fail on another.

## Solution

Relax the testing condition that are time-related.

### DIFF
--- a/rpcs/fetcher_test.go
+++ b/rpcs/fetcher_test.go
@@ -45,8 +45,6 @@ type MockRunner struct {
 	txgroups      [][]transactions.SignedTxn
 }
 
-const goExecTime = 100 * time.Millisecond
-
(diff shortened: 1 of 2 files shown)


tsachiherman:
> Looks good.
> One recommendation is to introduce a set of predefined timeout values.
> This PR is eliminating a pre-defined value: goExecTime.
> While goExecTime was not sufficient to handle the different timeout needs, a set of such values (e.g. short, long, very long) can be used in the test, instead of using raw time values at every instance.

I'm not opposing of doing that, but let's defer that to a separate PR. As is, it should eliminate some of the random failures around these tests.

PR #713, opened 2020-01-14
Update the makefile to skip the static linking when compiling on centos

tsachiherman:
## Summary

Our build system is using Centos to compile the release binaries. 
Unfortunately, the static linking doesn't work there out of the box.
This change "undo" the static linking on the centos platform, which doesn't have the same cross-platform issue we've had when compiling using Ubuntu 18.04.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,7 +20,10 @@ GOTAGSLIST          := sqlite_unlock_notify sqlite_omit_load_extension
 ifeq ($(UNAME), Linux)
 EXTLDFLAGS := -static-libstdc++ -static-libgcc
 ifeq ($(ARCH), amd64)
+# the following predicate is abit misleading; it tests if we're not in centos.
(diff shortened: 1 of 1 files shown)


algobolson:
Another option for the future is to install rpm packaging tools on ubuntu and package the static built binaries from there for deployment on RPM systems like Centos. We used to have this until we switched to actually building on Centos. The only reason we build on Centos is because of the libc dependency that should be gone now.

tsachiherman:
> Another option for the future is to install rpm packaging tools on ubuntu and package the static built binaries from there for deployment on RPM systems like Centos. We used to have this until we switched to actually building on Centos. The only reason we build on Centos is because of the libc dependency that should be gone now.

Yes, of course. And I hope to get there. The goal here was to allow Ben to bring his change to a mergeable milestone.

PR #714, opened 2020-01-14
Minor: Use service-wide logger instead of logging.Base() in agreement

pzbitskiy:
When I've been measuring votes processing performance I found usages of default `logging.Base()` near standard / preconfigured logger in some components of agreement service. This PR fixes it.

### DIFF
--- a/agreement/actions.go
+++ b/agreement/actions.go
@@ -20,7 +20,6 @@ import (
 	"context"
 	"fmt"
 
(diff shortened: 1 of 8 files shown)


tsachiherman:
Using the logger.Base() in general is incorrect. It's an anti-pattern, as the logger was designed to support decorators ( i.e. With )

PR #716, opened 2020-01-15
Fix a bug in Credential.lowestOutput caused by improper domain separation

algoradam:
The bug causes larger accounts to be block proposers more often than should happen based on their fraction of online stake. This patch will cause nodes to vote for a protocol upgrade that fixes the buggy behavior. After the protocol upgrade goes through, all the upgrade-related code in this commit should be removed, as it's not necessary to retain the old buggy behavior for catchup. (For convenience code to be removed is marked with a "TODO(upgrade)" comment.)

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -237,6 +237,10 @@ type ConsensusParams struct {
 
 	// max decimal precision for assets
 	MaxAssetDecimals uint32
+
+	// whether to use the old buggy Credential.lowestOutput function
+	// TODO(upgrade): Please remove as soon as the upgrade goes through
+	UseBuggyProposalLowestOutput bool
 }
 
 // Consensus tracks the protocol-level settings for different versions of the
@@ -311,6 +315,7 @@ func initConsensusProtocols() {
 		MaxBalLookback: 320,
 
 		MaxTxGroupSize: 1,
+		UseBuggyProposalLowestOutput: true, // TODO(upgrade): Please remove as soon as the upgrade goes through
 	}
 
 	v7.ApprovedUpgrades = map[protocol.ConsensusVersion]uint64{}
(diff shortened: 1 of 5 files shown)


data/committee/credential.go
```diff
@@ -155,11 +156,62 @@ func (cred Credential) Selected() bool {
...
 	var lowest big.Int
 
 	h1 := cred.VrfOut
-	for i := uint64(0); i < cred.Weight; i++ {
+	// It is important that i start at 1 rather than 0 because cred.Hashable
+	// was already hashed with iter = 0 earlier (in UnauthenticatedCredential.Verify)
+	// for determining the weight of the credential. A nonzero iter provides
+	// domain separation between lowestOutput and UnauthenticatedCredential.Verify
+	//
+	// It we reused the iter = 0 hash output here it would be nonuniformly
```
tsachiherman: typo : It -> If

config/config.go
```diff
@@ -237,6 +237,10 @@ type ConsensusParams struct {
 
 	// max decimal precision for assets
 	MaxAssetDecimals uint32
+
+	// whether to use the old buggy Credential.lowestOutput function
+	// TODO(upgrade): Please remove as soon as the upgrade goes through
+	UseBuggyProposalLowestOutput bool
```
tsachiherman: Do you think we can name this in a less ...negative.. language ? maybe `ProposalTieBreakerEqualizer`

data/committee/credential.go
```diff
@@ -155,11 +156,62 @@ func (cred Credential) Selected() bool {
...
+	// distributed (because lowestOutput can only get called if weight > 0).
+	// In particular if i starts at 0 then weight-1 credentials are at a
+	// significant disadvantage because UnauthenticatedCredential.Verify
+	// wants the hash to be large but tiebreaking between proposals wants
+	// the hash to be small.
+	for i := uint64(1); i <= cred.Weight; i++ {
```
tsachiherman: nit: for loop could be written better in terms of performance

config/config.go
algoradam: I'm actually a fan of negative language for these sorts of things -- makes us more likely to actually go back and delete the lines afterward! :)
Also having "false" mean "correct behavior" means that test consensus params that don't explicitly set this flag to true (e.g. testBigBlocks) will have correct behavior.

data/committee/credential.go
algoradam: In practice cred.Weight is never above 6 or 7, so performance isn't really an issue here -- computing the VRF is going to be way more expensive than computing the lowest output

data/committee/credential.go
algoradam: Fixed, thanks!

data/committee/credential.go
```diff
@@ -155,11 +156,62 @@ func (cred Credential) Selected() bool {
...
+
+	return &lowest
+}
+
+// TODO(upgrade): Please remove the entire lowestOutputBuggy function as soon as the corresponding protocol upgrade goes through.
+func (cred Credential) lowestOutputBuggy() *big.Int {
+	var lowest big.Int
+
+	h1 := cred.VrfOut
+	for i := uint64(0); i <= cred.Weight; i++ {
```
justicz: Shouldn't this just be a `<`?

data/committee/credential.go
```diff
@@ -184,6 +236,28 @@ func (cred Credential) lowestOutput() *big.Int {
...
+func (cred Credential) LessBuggy(otherCred Credential) bool {
+	i1 := cred.lowestOutputBuggy()
+	i2 := otherCred.lowestOutputBuggy()
+
+	return i1.Cmp(i2) < 0
+}
+
+// LowestOutputDigest gives the lowestOutput as a crypto.Digest, which allows
+// pretty-printing a proposal's lowest output.
+///This function is only used for debugging.
```
justicz: nit: typo

data/committee/credential.go
derbear: Yeah, this should be a `<`.

algoradam: Good catch, thanks!

PR #720, opened 2020-01-16
Add omitempty to OneTimeSignature struct

algoradam:
Right now we're encoding the empty PKSigOld, which wastes space

While we can't delete the (deprecated) PKSigOld field (because it's nonempty in old certs), we can add omitempty so that we don't keep transmitting it. We don't need to do this as part of a protocol upgrade or anything because OneTimeSignatures only appear in votes and are never hashed or signed, so it's okay if they're noncanonically encoded.

### DIFF
--- a/crypto/onetimesig.go
+++ b/crypto/onetimesig.go
@@ -36,13 +36,16 @@ import (
 // secret-holder from signing a contradictory message in the future in the event
 // of a secret-key compromise.
 type OneTimeSignature struct {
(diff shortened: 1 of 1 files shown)


pzbitskiy:
Adam, do we have a test that runs old and new certs through this code?

derbear:
It might be nice to add a node with `omitempty` in its struct and have it run against a network. Another test would have a relay with this change.

algoradam:
I'm not sure how this change interacts with the new encoder -- closing the PR for now and may reopen it later

PR #721, opened 2020-01-17
Move slow test to get executed on nightly builds

tsachiherman:
## Summary

Move some more test to be "slow tests", and modify short test condition so that we will run
the long tests on nightly builds only.

### DIFF
--- a/scripts/travis/run_tests.sh
+++ b/scripts/travis/run_tests.sh
@@ -9,7 +9,7 @@ if [ "${BUILD_TYPE}" = "integration" ]; then
     # Run short tests when doing pull requests; leave the long testing for nightly runs.
     if [[ "${TRAVIS_BRANCH}" =~ ^rel/nightly ]]; then
         SHORTTEST=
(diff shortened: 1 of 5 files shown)


scripts/travis/run_tests.sh
```diff
@@ -6,10 +6,13 @@ SCRIPTPATH="$( cd "$(dirname "$0")" ; pwd -P )"
 OS=$("${SCRIPTPATH}/../ostype.sh")
 
 if [ "${BUILD_TYPE}" = "integration" ]; then
-    # For now, disable long-running e2e tests on Travis
-    # (the ones that won't complete...)
-    SHORTTEST=-short
-    export SHORTTEST    
+    # Run short tests when doing pull requests; leave the long testing for nightly runs.
```
algonautshant: Is this needed here? 
export SHELLOPTS

tsachiherman: why SHELLOPTS here ?

PR #726, opened 2020-01-20
Move EnsureDigest logic into the catchup service

tsachiherman:
## Summary

Move EnsureDigest logic into the catchup service. That makes the fetcher logic being used by a single source ( i.e. catchup ) rather than shared between the node and catchup, and would allow us to unify the logic there.

This PR is intended to be the first in a series, which handles only the portion of "moving the ensure digest logic from node into the catchup service".
Follow up for this PR would be "move fetcher logic from rpcs into catchup" and "move fetcher service into catchup"

### DIFF
--- a/catchup/service.go
+++ b/catchup/service.go
@@ -79,7 +79,7 @@ type Service struct {
 	initialSyncNotified          uint32
 	protocolErrorLogged          bool
 	lastSupportedRound           basics.Round
-	unmatchedPendingCertificates chan PendingUnmatchedCertificate
+	unmatchedPendingCertificates <-chan PendingUnmatchedCertificate
 
 	latestRoundFetcherFactory rpcs.FetcherFactory
 }
@@ -98,7 +98,7 @@ type BlockAuthenticator interface {
 
 // MakeService creates a catchup service instance from its constituent components
 // If wsf is nil, then fetch over gossip is disabled.
-func MakeService(log logging.Logger, config config.Local, net network.GossipNode, ledger Ledger, wsf *rpcs.WsFetcherService, auth BlockAuthenticator, unmatchedPendingCertificates chan PendingUnmatchedCertificate) (s *Service) {
+func MakeService(log logging.Logger, config config.Local, net network.GossipNode, ledger Ledger, wsf *rpcs.WsFetcherService, auth BlockAuthenticator, unmatchedPendingCertificates <-chan PendingUnmatchedCertificate) (s *Service) {
 	s = &Service{}
 	s.ctx, s.cancel = context.WithCancel(context.Background())
 	s.cfg = config
--- a/catchup/service.go
+++ b/catchup/service.go
@@ -518,7 +518,7 @@ func (s *Service) fetchRound(cert agreement.Certificate, verifier *agreement.Asy
 	defer func() {
 		fetcher.Close()
 	}()
-	for {
+	for s.ledger.LastRound() < cert.Round {
 		if fetcher.OutOfPeers(cert.Round) {
 			fetcher.Close()
 			// refresh peers and try again
--- a/catchup/service.go
+++ b/catchup/service.go
@@ -460,7 +460,8 @@ func (s *Service) periodicSync() {
 }
 
 // Syncs the client with the network. sync asks the network for last known block and tries to sync the system
-// up the to the highest number it gets
+// up the to the highest number it gets. When a certificate is provided, the sync function attempts to keep trying
+// to fetch the matching block or abort when the catchup service exits.
 func (s *Service) sync(cert *PendingUnmatchedCertificate) {
 	// Only run sync once at a time
 	// Store start time of sync - in NS so we can compute time.Duration (which is based on NS)
--- a/catchup/service.go
+++ b/catchup/service.go
@@ -511,6 +511,7 @@ func (s *Service) sync(cert *PendingUnmatchedCertificate) {
 	s.log.Infof("Catchup Service: finished catching up, now at round %v (previously %v). Total time catching up %v.", s.ledger.LastRound(), pr, elapsedTime)
 }
 
+// TODO this doesn't actually use the digest from cert!
 func (s *Service) fetchRound(cert agreement.Certificate, verifier *agreement.AsyncVoteVerifier) {
 	blockHash := bookkeeping.BlockHash(cert.Proposal.BlockDigest) // semantic digest (i.e., hash of the block header), not byte-for-byte digest
 	fetcher := s.latestRoundFetcherFactory.NewOverGossip(protocol.UniEnsBlockReqTag)
@@ -527,17 +528,18 @@ func (s *Service) fetchRound(cert agreement.Certificate, verifier *agreement.Asy
 		}
 		// Ask the fetcher to get the block somehow
 		block, fetchedCert, rpcc, err := s.innerFetch(fetcher, cert.Round)
-		rpcc.Close()
 
 		if err != nil {
 			select {
 			case <-s.ctx.Done():
-				logging.Base().Debugf("EnsureDigest was asked to quit before we could acquire the block")
+				logging.Base().Debugf("fetchRound was asked to quit before we could acquire the block")
 				return
 			default:
 			}
-			logging.Base().Panicf("EnsureDigest could not acquire block, fetcher errored out: %v", err)
+			logging.Base().Warnf("fetchRound could not acquire block, fetcher errored out: %v", err)
+			continue
 		}
+		rpcc.Close()
 
 		if block.Hash() == blockHash && block.ContentsMatchHeader() {
 			s.ledger.EnsureBlock(block, cert)
@@ -553,7 +555,7 @@ func (s *Service) fetchRound(cert agreement.Certificate, verifier *agreement.Asy
 			s := "!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!\n"
 			s += "!!!!!!!!!! FORK DETECTED !!!!!!!!!!!\n"
 			s += "!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!\n"
-			s += "EnsureDigest called with a cert authenticating block with hash %v.\n"
+			s += "fetchRound called with a cert authenticating block with hash %v.\n"
 			s += "We fetched a valid cert authenticating a different block, %v. This indicates a fork.\n\n"
 			s += "Cert from our agreement service:\n%#v\n\n"
 			s += "Cert from the fetcher:\n%#v\n\n"
--- a/catchup/service.go
+++ b/catchup/service.go
@@ -42,7 +42,7 @@ const blockQueryPeerLimit = 10
 // this should be at least the number of relays
 const catchupRetryLimit = 500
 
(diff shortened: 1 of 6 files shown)


node/impls.go
```diff
@@ -161,108 +162,29 @@ func (l agreementLedger) EnsureValidatedBlock(ve agreement.ValidatedBlock, c agr
...
-			s += "Cert from the fetcher:\n%#v\n\n"
-			s += "Block from the fetcher:\n%#v\n\n"
-			s += "!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!\n"
-			s += "!!!!!!!!!! FORK DETECTED !!!!!!!!!!!\n"
-			s += "!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!\n"
-			s = fmt.Sprintf(s, cert.Proposal.BlockDigest, fetchedCert.Proposal.BlockDigest, cert, fetchedCert, block)
-			fmt.Println(s)
-			logging.Base().Error(s)
-		}
+	waitForRound := l.Wait(cert.Round)
```
derbear: Nit: naming is a bit misleading--perhaps the variable name should indicate that it's a channel.

catchup/service.go
```diff
@@ -70,10 +75,13 @@ type Service struct {
...
 	// the overhead of starting prematurely (before this node is caught-up and can validate messages for example).
-	InitialSyncDone     chan struct{}
-	initialSyncNotified uint32
-	protocolErrorLogged bool
-	lastSupportedRound  basics.Round
+	InitialSyncDone              chan struct{}
+	initialSyncNotified          uint32
+	protocolErrorLogged          bool
+	lastSupportedRound           basics.Round
+	unmatchedPendingCertificates chan PendingUnmatchedCertificate
```
derbear: Nit: have this be of type `<-chan PendingUnmatchedCertificate`

catchup/service.go
```diff
@@ -492,6 +511,65 @@ func (s *Service) sync() {
 	s.log.Infof("Catchup Service: finished catching up, now at round %v (previously %v). Total time catching up %v.", s.ledger.LastRound(), pr, elapsedTime)
 }
 
+// TODO this doesn't actually use the digest from cert!
+func (s *Service) fetchRound(cert agreement.Certificate, verifier *agreement.AsyncVoteVerifier) {
```
derbear: Instead of passing the `AsyncVoteVerifier`, perhaps you can use the `BlockAuthenticator` that you already have?

node/impls.go
```diff
@@ -161,108 +162,29 @@ func (l agreementLedger) EnsureValidatedBlock(ve agreement.ValidatedBlock, c agr
...
+	// clear out the pending certificates ( if any )
+	select {
+	case pendingCert := <-l.UnmatchedPendingCertificates:
```
derbear: Does this currently ever happen?

node/impls.go
```diff
@@ -161,108 +162,29 @@ func (l agreementLedger) EnsureValidatedBlock(ve agreement.ValidatedBlock, c agr
...
-	go func() {
-		block, cert, _, err := fetcher.FetchBlock(ctx, round)
-		localdone <- fbreturn{block, cert, err}
-	}()
 	select {
-	case ret := <-localdone:
-		return ret.block, ret.cert, ret.err
+	case l.UnmatchedPendingCertificates <- catchup.PendingUnmatchedCertificate{Cert: cert, VoteVerifier: verifier}:
+		// good, we've placed the cert in the queue.
 	case <-quit:
```
derbear: Given that this channel send will be non-blocking, do we need to handle this?

derbear: Also nit: you should comment that the channel send is guaranteed to be non-blocking.

catchup/service.go
```diff
@@ -36,21 +37,25 @@ import (
...
+	VoteVerifier *agreement.AsyncVoteVerifier
+}
+
 // Ledger represents the interface of a block database which the
 // catchup server should interact with.
 type Ledger interface {
-	NextRound() basics.Round
-	LastRound() basics.Round
-	Wait(basics.Round) chan struct{}
+	agreement.LedgerReader
```
derbear: Do you need the entire `LedgerReader` here as a dependency? (Or just `Wait`?)

node/impls.go
tsachiherman: sure.

catchup/service.go
tsachiherman: yes

catchup/service.go
tsachiherman: The BlockAuthenticator is using lowPriorityCryptoVerificationPool whereas the passed-in verifier is using a highPriorityCryptoVerificationPool.

I agree that it's a bit clunky, and should be passed only on initialization. Let's do this cleanup in a follow-up change ?

node/impls.go
tsachiherman: sure.

catchup/service.go
tsachiherman: Needed for the usage of Certificate.Authenticate() call; I'm not a big fan of having yet-another Ledger interface.
I'd rather to have an instance of the data/Ledger directly, but I don't want to make this into a bigger change in the context of this change. ( I guess it used to be important for preventing circular referencing ? )

node/impls.go
```diff
@@ -161,108 +162,29 @@ func (l agreementLedger) EnsureValidatedBlock(ve agreement.ValidatedBlock, c agr
...
-	}
-	localdone := make(chan fbreturn, 1)
-	go func() {
-		block, cert, _, err := fetcher.FetchBlock(ctx, round)
-		localdone <- fbreturn{block, cert, err}
-	}()
 	select {
-	case ret := <-localdone:
-		return ret.block, ret.cert, ret.err
+	case l.UnmatchedPendingCertificates <- catchup.PendingUnmatchedCertificate{Cert: cert, VoteVerifier: verifier}:
```
algonautshant: What if periodicSync is not running here? e.g. catchup stopped because the next round protocol is not supported?

node/impls.go
tsachiherman: It could. The is a very low probability that the catchup is at a stage where it would try to catch up after no new block was received by the agreement service. if that happens, we could have ended up the first unmatched pending certificate to be fulfilled by a "regular" catchup. at that point, we have a "stale" item in the "UnmatchedPendingCertificates". 
We'd probably do better by addressing this edge-case in both EnsureDigest as well as in catchup service.

catchup/service.go
algonautshant: The name unmatchedPendingCertificates is not very descriptive in the context of the catchup service. 
In the catchup context, this channel looks like a fast-track fetch block request channel. There is nothing specific to unmatched pending certificates in the catchup service. Right?

node/impls.go
tsachiherman: If catchup is no longer able to catch up, the agreement won't be able to make progress either, so this doesn't make it any worse. In fact, the agreement should not have attempted to call EnsureDigest on a block that cannot be agreed on due to unsupported protocol.

catchup/service.go
tsachiherman: Let me know if you have a better descriptive name for this channel and/or data structure.

algonautshant: How about: immediateFetchCertificate?

catchup/service.go
```diff
@@ -36,21 +37,25 @@ import (
 )
 
 const catchupPeersForSync = 10
+const blockQueryPeerLimit = 10
 
 // this should be at least the number of relays
 const catchupRetryLimit = 500
 
+// PendingUnmatchedCertificate is a single certificate that is being waited upon to have it's corresponding block fetched.
```
algoradam: nit: it's -> its

catchup/service.go
```diff
@@ -90,14 +98,17 @@ type BlockAuthenticator interface {
...
 	s = &Service{}
 	s.ctx, s.cancel = context.WithCancel(context.Background())
 	s.cfg = config
 	s.fetcherFactory = rpcs.MakeNetworkFetcherFactory(net, catchupPeersForSync, wsf)
 	s.ledger = ledger
 	s.net = net
 	s.auth = auth
+	s.unmatchedPendingCertificates = unmatchedPendingCertificates
+
+	s.latestRoundFetcherFactory = rpcs.MakeNetworkFetcherFactory(net, blockQueryPeerLimit, wsf)
```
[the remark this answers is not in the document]
tsachiherman: Right now, we don't. I chose to keep it with the same logic as it was before to minimize the amount of changes.

catchup/service.go
```diff
@@ -492,6 +512,65 @@ func (s *Service) sync() {
...
+			s += "We fetched a valid cert authenticating a different block, %v. This indicates a fork.\n\n"
+			s += "Cert from our agreement service:\n%#v\n\n"
+			s += "Cert from the fetcher:\n%#v\n\n"
+			s += "Block from the fetcher:\n%#v\n\n"
+			s += "!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!\n"
+			s += "!!!!!!!!!! FORK DETECTED !!!!!!!!!!!\n"
+			s += "!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!\n"
+			s = fmt.Sprintf(s, cert.Proposal.BlockDigest, fetchedCert.Proposal.BlockDigest, cert, fetchedCert, block)
+			fmt.Println(s)
+			logging.Base().Error(s)
```
algoradam: Shouldn't this be a Panic? We don't want to keep running after a fork

tsachiherman: Current code doesn't panic. I'm not sure if panicking would be the smartest idea.
I think that logging to telemetry and having some forever-sleeping here would be a better thing to do. Keep in mind that the agreement is already locked-step with this one, so if we go to sleep here, both the agreement & catchup are disabled.
Either way, I would break that off to a separate change since it's a behavioral change.

catchup/service.go
derbear: Sounds good.

node/impls.go
derbear: Given that the channel send is non-blocking, it might be worth ignoring the `quit` channel here since we'll `select` on it immediately afterwards anyway.

catchup/service.go
derbear: I guess the point of this interface was to decouple the `Ledger`-related requirements of the `catchup` package from the rest of the codebase, and importing `agreement.LedgerReader` seems to work against that objective. I don't have strong opinions here, though.

node/impls.go
tsachiherman: The intent here was to avoid writing the content to the channel is quit is closed. I believe it's still correct, although the implementation is incorrect. I'll fix it.

derbear: I'm confused--this doesn't seem to solve the problem? Since `s.quit` can always be closed after this `select` block, which races on the channel send below...

tsachiherman: The `quit` can get closed at any moment. that's true.
When the `quit` is closed after/during the send into the channel then the logic on the catchup would need to tackle with that ( I believe that it does ). However, in the case the `EnsureDigest` was called with a closed quit channel, there is no reason to push that onto the channel. We want to expedite the shutdown process.

derbear: I guess the check seems unnecessary to me, and it's unclear how much it matters in practice. This isn't a blocker for the PR though.

PR #729, opened 2020-01-21
Update main.go

pipaman:
Program bytes are needed for the Sign function. It doesn't work with 4 parameters in this way since it doesn't verify when you call ed25519verify from teal. With this change, it does verify.
There is another issue I've noticed: when you pass a file as the data to sign, it is likely to have a special character in it, so the final signature includes this special character. I suggest making a string parameter version to avoid this problem.

## Summary

## Test Plan

### DIFF
--- a/tools/teal/dkey/dsign/main.go
+++ b/tools/teal/dkey/dsign/main.go
@@ -62,7 +62,9 @@ func main() {
 		ddata, err := ioutil.ReadFile(os.Args[3])
 		failFast(err)
 
-		programbytes, err := logic.AssembleString(pdata)
+		programstr := fmt.Sprintf("%s", pdata)
+		
+		programbytes, err := logic.AssembleString(programstr)
 		failFast(err)
 
 		dsig := sec.Sign(logic.Msg{
--- a/tools/teal/dkey/dsign/main.go
+++ b/tools/teal/dkey/dsign/main.go
@@ -62,8 +62,11 @@ func main() {
 		ddata, err := ioutil.ReadFile(os.Args[3])
 		failFast(err)
 
+		programbytes, err := logic.AssembleString(pdata)
+		failFast(err)
+
 		dsig := sec.Sign(logic.Msg{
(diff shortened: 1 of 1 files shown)


justicz:
It looks like this change makes it so `dsign` takes in the *source code* of a TEAL program, whereas right now it takes in a *compiled* TEAL program (as produced by `goal clerk compile foo.logic`). I don't think the current behavior is wrong, but perhaps it should be better documented.

pipaman:
The problem is that you can't use the output on a TEAL code to call ed25519verify. Maybe, we'll need another option of the same tool that generates the signature of the compiled program that can be used in a TEAL code.

tools/teal/dkey/dsign/main.go
```diff
@@ -62,8 +62,13 @@ func main() {
 		ddata, err := ioutil.ReadFile(os.Args[3])
 		failFast(err)
 
+		programstr := fmt.Sprintf("%s", pdata)
```
robert-zaremba: Why using `fmt.Sprintf` where `fmt.Sprint` would do the work (notice no `f` in function name)?

pipaman: I'm using Sprint because I've got a byte[] containing a string (the code) and I want to create a string. The alternative is to use:
programstr := string(pdata)
 
Your proposed change doesn't work because it converts the byte[] to a string containing the binary codes (e.g.: '27 34 93 59') instead of a string with the TEAL code.

derbear:
I'm of the opinion that the `dsign` program should just sign bytes and not include an entire TEAL assembler, which is redundant with the `goal clerk compile` command. I do agree with @justicz that the documentation here is lacking.

derbear:
@pipaman, we decided to take your suggestion and integrate this functionality into `goal clerk tealsign`. How does #800 look?

pipaman:
It looks awesome, integrating it in goal is the correct way to do it. Thank
you!

PR #732, opened 2020-01-22
Shant/limit connections

algonautshant:
## Summary
This PR limits the number or connection requests from the fetcher service to the server such that no more than ConnectionsRateLimitingCount are mad in ConnectionsRateLimitingWindowSeconds.

This PR is for the fetcher and txSynch. gossip network will be fixed in a separate PR.

## Test Plan
Added a unit test for the main function computing the timers.


network/phonebook.go
```diff
@@ -82,7 +105,48 @@ func (e *phonebookEntries) ReplacePeerList(they []string) {
...
 func (e *phonebookEntries) updateRetryAfter(addr string, retryAfter time.Time) {
-	(*e)[addr] = phonebookData{retryAfter: retryAfter}
+	_, found := (*e)[addr]
+	if !found {
+		(*e)[addr] = phonebookData{retryAfter: retryAfter, recentConnectionTimes: make([]time.Time, 0)}
+	} else {
+		phbData := (*e)[addr]
+		phbData.retryAfter = retryAfter
+		(*e)[addr] = phbData
+	}
```
tsachiherman: nit: the code is correct, however, it could be more nicely written - 
```
  	entry := (*e)[addr]
  	entry.retryAfter = retryAfter
  	(*e)[addr] = entry
```
and it's functionally identical!

network/wsNetwork.go
```diff
@@ -1762,3 +1775,14 @@ func SetUserAgentHeader(header http.Header) {
 	ua := fmt.Sprintf("algod/%d.%d (%s; commit=%s; %d) %s(%s)", version.Major, version.Minor, version.Channel, version.CommitHash, version.BuildNumber, runtime.GOOS, runtime.GOARCH)
 	header.Set(UserAgentHeader, ua)
 }
+
+// DoHTTP will check call to WaitAndAddConnectionTime to
+// make sure connectionsRateLimitingCount is not violated, and register the
+// connection time of the request, before sending the request to the server. 
+func DoHTTP(client *http.Client,
```
tsachiherman: Add this method as a method of WebsocketNetwork; i.e. 
```
func (wn *WebsocketNetwork) MakeHttpRequest(client *http.Client, request *http.Request )
```

which would extract the address from the request, take care of the phonebook waiting and updating and call the client.Do with the proper arguments.

rpcs/httpFetcher.go
```diff
@@ -45,15 +45,16 @@ type FetcherClient interface {
 type HTTPFetcher struct {
 	peer    network.HTTPPeer
 	rootURL string
+	net     *network.GossipNode
```
tsachiherman: I'm pretty confident that you don't need the `*network.GossipNode` but rather the `network.GossipNode`. The latter is already an interface, so it's a struct pointer.

rpcs/ledgerService_test.go
```diff
@@ -53,6 +54,27 @@ type httpTestPeerSource struct {
 func (s *httpTestPeerSource) GetPeers(options ...network.PeerOption) []network.Peer {
 	return s.peers
 }
+func (s *httpTestPeerSource) Address() (string, bool) {
+	return "", false
+}
+func (s *httpTestPeerSource) Broadcast(ctx context.Context, tag protocol.Tag, data []byte, wait bool, except network.Peer) error {
+	return errors.New("mock MockClientAggregator does not implement Broadcast")
```
tsachiherman: If you already decided to have your own custom errors, you should at least make sure these would contain a meaning text..

network/phonebook.go
```diff
@@ -82,7 +105,48 @@ func (e *phonebookEntries) ReplacePeerList(they []string) {
...
+}
+
+// waitAndAddConnectionTime will wait to prevent exceeding connectionsRateLimitingCount.
+// Then it will register the next connection time.
+func (e *phonebookEntries) waitAndAddConnectionTime(addr string,
```
tsachiherman: Could you have a unit test to validate the correctness of this method ?

network/phonebook.go
```diff
@@ -82,7 +105,48 @@ func (e *phonebookEntries) ReplacePeerList(they []string) {
...
+	connectionRateLimitWindowDuration := time.Duration(connectionsRateLimitingWindowSeconds)*time.Second
+	_, found := (*e)[addr]
+	if !found {
+		(*e)[addr] = phonebookData{retryAfter: time.Now(), recentConnectionTimes: make([]time.Time, 0)}
+	}
+
+	var timeSince time.Duration
+	// Remove from recentConnectionTimes the times later than ConnectionsRateLimitingWindowSeconds
+	for len((*e)[addr].recentConnectionTimes) > 0 {
+		timeSince = time.Since(((*e)[addr].recentConnectionTimes)[0])
```
tsachiherman: Using time.Since is abit of an issue since the time is moving forward while the function is executing. You want to get a reference time ( i.e. `curTime := time.Now()` ) and use that all across the function.
Also, in this section you want to find the "tail" and cut it in a single slice rather than one item at a time.

network/wsNetwork.go
```diff
@@ -772,6 +777,14 @@ func (wn *WebsocketNetwork) checkServerResponseVariables(header http.Header, add
 	return true
 }
 
+// WaitAndAddConnectionTime will wait to prevent exceeding connectionsRateLimitingCount.
+// Then it will register the next connection time.
+func (wn *WebsocketNetwork) WaitAndAddConnectionTime(addr string) {
```
tsachiherman: I'm not sure if you really need this function at all. For sure there is no reason to export it.
Deleting this one would save you from implementing it across all the testing interfaces..

network/phonebook.go
```diff
@@ -35,14 +35,37 @@ type Phonebook interface {
...
+	retryAfter            time.Time
+	recentConnectionTimes []time.Time
 }
 
 type phonebookEntries map[string]phonebookData
 
+// PopEarliestTime removes the earliest time from recentConnectionTimes in
+// phonebookData for addr
+// It is expected to be later than ConnectionsRateLimitingWindowSeconds
+func (e *phonebookEntries) popEarliestTime(addr string) {
```
justicz: I know you didn't write the methods that were here before, but for what it's worth I don't think these methods need to be using pointer receivers. If it were just `(e phonebookEntries)` the map could still be modified from within the method, since maps are a reference type.

network/phonebook.go
```diff
@@ -256,3 +340,16 @@ func (mp *MultiPhonebook) UpdateRetryAfter(addr string, retryAfter time.Time) {
...
+// WaitAndAddConnectionTime will wait to prevent exceeding connectionsRateLimitingCount.
+// Then it will register the next connection time.
+func (mp *MultiPhonebook) WaitAndAddConnectionTime(addr string,
+	connectionsRateLimitingCount uint, connectionsRateLimitingWindowSeconds uint) {
+	mp.lock.Lock()
+	defer mp.lock.Unlock()
+	for _, op := range mp.phonebookMap {
+		op.WaitAndAddConnectionTime(addr,
+			connectionsRateLimitingCount, connectionsRateLimitingWindowSeconds)
+		break // If the first phonebook does not have the phonebookData, will add it there
```
tsachiherman: iteration over maps aren't guaranteed to be executed at the same order.
my suggestion is that if we can't find the address on any of the maps we would just exit.

network/phonebook_test.go
```diff
@@ -202,6 +203,69 @@ func TestMultiPhonebookDuplicateFiltering(t *testing.T) {
...
+	phBookData = entries[addr1]
+	assert.Equal(t, 2, len(phBookData.recentConnectionTimes))
+
+	// wait for the time the first element should be removed
+	time.Sleep(1 * time.Second)
+
+	// the first time should be removed and a new one added
+	entries.waitAndAddConnectionTime(addr1, 3, 2)
+	phBookData = entries[addr1]
+	assert.Equal(t, 2, len(phBookData.recentConnectionTimes))
```
tsachiherman: instead of assert, use require

tsachiherman: You also want to verify the content of the entries time.
i.e. save the content of the *second* element in the array *before* calling the waitAndAdd... and make sure it's the first element after.

network/phonebook_test.go
```diff
@@ -202,6 +203,69 @@ func TestMultiPhonebookDuplicateFiltering(t *testing.T) {
...
+
+	// try requestiong from another address, make sure
```
tsachiherman: nit: typo requestiong ->requesting

network/phonebook_test.go
```diff
@@ -202,6 +203,69 @@ func TestMultiPhonebookDuplicateFiltering(t *testing.T) {
...
+	// a separate array is used for these new requests
+
+	// add 3 values to another address. should not be blocked
+	t0 := time.Now()
+	entries.waitAndAddConnectionTime(addr2, 3, 2)
+	entries.waitAndAddConnectionTime(addr2, 3, 2)
+	entries.waitAndAddConnectionTime(addr2, 3, 2)
+	// expected no wait
+	assert.True(t, (time.Since(t0)) < 1*time.Second)
```
tsachiherman: this is a problematic statement; on a slow machine it might fail.
why don't you add a return value to entries.waitAndAddConnectionTime which indicates if the function went to sleep ?

network/phonebook.go
```diff
@@ -35,17 +35,40 @@ type Phonebook interface {
...
+		connectionsRateLimitingCount uint, connectionsRateLimitingWindowSeconds uint)
 }
 
 type phonebookData struct {
-	retryAfter time.Time
+	retryAfter            time.Time
+	recentConnectionTimes []time.Time
 }
 
 type phonebookEntries map[string]phonebookData
```
tsachiherman: I think that you'd like to change the phonebookEntries to be a struct, so that it could hold the connectionsRateLimitingCount  and connectionsRateLimitingWindowSeconds.

That would allow you to pass these during the construction rather than over and over.

network/phonebook_test.go
```diff
@@ -202,6 +203,69 @@ func TestMultiPhonebookDuplicateFiltering(t *testing.T) {
...
+	phBookData = entries[addr2]
+	// all three times should be queued
+	assert.Equal(t, 3, len(phBookData.recentConnectionTimes))
+
+	// add another element to trigger wait
+	t0 = time.Now()
+	entries.waitAndAddConnectionTime(addr2, 3, 2)
+	// there should be some wait the earliest element expires
+	assert.True(t, time.Since(t0) > 1*time.Second)
```
tsachiherman: I would hope that you can re-design this test to take less than one second..

rpcs/ledgerService_test.go
```diff
@@ -228,6 +274,25 @@ type BasicRPCNode struct {
 	peers    []network.Peer
 }
 
+func (b *BasicRPCNode) Address() (string, bool) {
```
tsachiherman: is it possible to embed the MockNetwork here instead of re-implementing all of these ?

network/phonebook.go
```diff
@@ -61,28 +101,76 @@ func (e *phonebookEntries) ReplacePeerList(they []string) {
...
+	if uint(numElts) >= e.connectionsRateLimitingCount {
+		// Wait until the earliest time expires
+		time.Sleep(e.connectionsRateLimitingWindow - timeSince)
+		waited = true
+		// Remove it from recentConnectionTimes
+		e.popNElements(1, addr)
+	}
+
+	// Append the time for the next connection request
+	e.appendTimeNow(addr)
```
tsachiherman: instead of letting this function determine what is "now", re-use the `curTime`

network/wsNetwork.go
```diff
@@ -772,6 +778,21 @@ func (wn *WebsocketNetwork) checkServerResponseVariables(header http.Header, add
 	return true
 }
 
+// waitAndAddConnectionTime will wait to prevent exceeding connectionsRateLimitingCount.
+// Then it will register the next connection time.
+func (wn *WebsocketNetwork) waitAndAddConnectionTime(addr string) {
+	wn.phonebook.WaitAndAddConnectionTime(addr)
+}
```
tsachiherman: I think it's abit pointless to create a wrapper function for a single line of code..

network/phonebook.go
algonautshant: Actually, this logic is wrong.

Consider the case where multiple parallel requests are made to the server. For some reason, it takes a long time to connect to the server. The recorded connected times will be much earlier than when the server actually received the requests. 

Then, the client makes more requests, since the recorded times are expired by now. However, this may exceed the limit for the server. 

To correct this, I am separating the wait for the connection from the add time. The add time should be made after the request was sent.

tsachiherman: I think that the code is correct - but we need to "append" before and "update" once the connection is established.

network/phonebook.go
```diff
@@ -61,28 +105,100 @@ func (e *phonebookEntries) ReplacePeerList(they []string) {
...
+	// Append the provisional time for the next connection request
+	e.appendTime(addr, curTime)
+	return found, waited, curTime
+}
+
+// UpdateConnectionTime will update the provisional connection time.
+// Returns true of the addr was in the phonebook
+func (e *phonebookEntries) updateConnectionTime(addr string, provisionalTime time.Time) bool {
+	entry, found := e.data[addr]
+	if found {
```
tsachiherman: cleanup: 
```
if !found {
     return false
}
...
```

network/phonebook.go
```diff
@@ -35,17 +35,61 @@ type Phonebook interface {
...
 
-type phonebookEntries map[string]phonebookData
+// makePhonebookEntries creates phonebookEntries with the passed configuration values
+func makePhonebookEntries(connectionsRateLimitingCount uint,
+	connectionsRateLimitingWindow time.Duration) phonebookEntries {
+	var pbEnteries phonebookEntries
+	pbEnteries.connectionsRateLimitingCount = connectionsRateLimitingCount
+	pbEnteries.connectionsRateLimitingWindow = connectionsRateLimitingWindow
+	pbEnteries.data = make(map[string]phonebookData, 0)
+	return pbEnteries
```
tsachiherman: nit: I think it would make it clear to use struct literal for that.

PR #735, opened 2020-01-22
Format numbers using number specifier

tsachiherman:
## Summary

Existing code was using `%v` to format numeric values. While that `%v` is a completely valid parameter, it could result in a recursive looping when a `fmt.Stringer` interface is being implemented on the argument.

Using `%d` eliminate this potential issue.

### DIFF
--- a/agreement/pseudonode.go
+++ b/agreement/pseudonode.go
@@ -476,7 +476,7 @@ func (t pseudonodeProposalsTask) execute(verifier *AsyncVoteVerifier, quit chan 
 			ObjectRound:  uint64(vote.R.Round),
 			ObjectPeriod: uint64(vote.R.Period),
 		}
-		t.node.log.with(logEvent).Infof("pseudonode.makeProposals: proposal created for (%v, %v)", vote.R.Round, vote.R.Period)
+		t.node.log.with(logEvent).Infof("pseudonode.makeProposals: proposal created for (%d, %d)", vote.R.Round, vote.R.Period)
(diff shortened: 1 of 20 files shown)


agreement/pseudonode.go
```diff
@@ -484,7 +484,7 @@ func (t pseudonodeProposalsTask) execute(verifier *AsyncVoteVerifier, quit chan
 			Period:  uint64(vote.R.Period),
 		})
 	}
-	t.node.log.Infof("pseudonode.makeProposals: %v proposals created for round %v, period %v", len(verifiedVotes), t.round, t.period)
+	t.node.log.Infof("pseudonode.makeProposals: %v proposals created for round %d, period %d", len(verifiedVotes), t.round, t.period)
```
derbear: The first `%v` here should be changed too, right?

agreement/selector.go
```diff
@@ -64,19 +64,19 @@ func membership(l LedgerReader, addr basics.Address, r basics.Round, p period, s
 
 	record, err := l.BalanceRecord(balanceRound, addr)
 	if err != nil {
-		err = fmt.Errorf("Service.initializeVote (r=%v): Failed to obtain balance record for address %v in round %v: %v", r, addr, balanceRound, err)
+		err = fmt.Errorf("Service.initializeVote (r=%v): Failed to obtain balance record for address %v in round %d: %v", r, addr, balanceRound, err)
```
derbear: The first %v here (and below) should be changed too, right?

agreement/pseudonode.go
tsachiherman: yes.

agreement/selector.go
tsachiherman: right.

PR #736, opened 2020-01-22
Faster stringer implementation for Address

tsachiherman:
## Summary

Just make it faster. Nothing exciting going here.

## Benchmark Plan
Address.String():  550 ns / op -> 350 ns / op
UnmarshalChecksumAddress():  1116 ns / op -> 1001 ns / op

### DIFF
--- a/data/basics/address.go
+++ b/data/basics/address.go
@@ -82,6 +82,7 @@ func UnmarshalChecksumAddress(address string) (Address, error) {
 func (addr Address) String() string {
 	addrWithChecksum := make([]byte, crypto.DigestSize+checksumLength)
 	copy(addrWithChecksum[:crypto.DigestSize], addr[:])
+	// calling addr.GetChecksum() here takes 20ns more than just rolling it out, so we'll just repeat that code.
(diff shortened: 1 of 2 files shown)


data/basics/address.go
```diff
@@ -77,9 +80,11 @@ func UnmarshalChecksumAddress(address string) (Address, error) {
 
 // String returns a string representation of Address
 func (addr Address) String() string {
-	var addrWithChecksum []byte
-	addrWithChecksum = append(addr[:], addr.GetChecksum()...)
-	return base32.StdEncoding.WithPadding(base32.NoPadding).EncodeToString(addrWithChecksum)
+	addrWithChecksum := make([]byte, crypto.DigestSize+checksumLength)
+	copy(addrWithChecksum[:crypto.DigestSize], addr[:])
+	shortAddressHash := crypto.Hash(addr[:])
```
pzbitskiy: why not using addr.GetChecksum() ? It would make the code more readable and seems like there is almost no performance penalty on this extra call

PR #741, opened 2020-01-23
Change account rest model to use a map of strings for assets.

### LINKED ISSUE #739: Account Assets and AssetParams properties aren't properly defined in swagger specfiles
### Subject of the issue
The algod swagger file isn't quite right, there is no **type** field, or object reference for `thisassettotal` and `assets`:

```
 828     "Account": {
 829       "description": "Account Description",
 830       "type": "object",
 831       "required": [
 832         "round",  
 833         "address",
 834         "amount",
 835         "pendingrewards",
 836         "amountwithoutpendingrewards",
 837         "rewards",
 838         "status"
 839       ],
 840       "properties": {
 841         "address": {
 842           "description": "Address indicates the account public key",
 843           "type": "string",
 844           "x-go-name": "Address"
 845         },
...
 858         "assets": {
 859           "description": "Assets specifies the holdings of assets by this account,\nindexed by the asset ID.",
 860           "x-go-name": "Assets"
 861         },
 862         "participation": {
 863           "$ref": "#/definitions/Participation"
 864         },
...
 888         "thisassettotal": {
 889           "description": "AssetParams specifies the parameters of assets created by this account.",
 890           "x-go-name": "AssetParams"
 891         }
 892       },
 893       "x-go-package": "github.com/algorand/go-algorand/daemon/algod/api/spec/v1"
 894     },
```

### Investigation
In `daemon/algod/api/spec/v1/model.go` I see these defined as maps of uint64:
```go
  // AssetParams specifies the parameters of assets created by this account.    
  //    
  // required: false    
  AssetParams map[uint64]AssetParams `json:"thisassettotal,omitempty"`    
    
  // Assets specifies the holdings of assets by this account,    
  // indexed by the asset ID.    
  //    
  // required: false    
  Assets map[uint64]AssetHolding `json:"assets,omitempty"`    
```

According to the go-swagger doc that isn't even supported by JSON, so it may be an invalid endpoint in general:
https://goswagger.io/faq/faq_spec.html#maps-as-swagger-parameters

The OpenAPI 2.0 spec agrees. When creating map properties the type is set to "object", with an additional parameter provided for the value. :
https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#model-with-mapdictionary-properties

### Next Steps

It looks like the correct way to fix this is to change `map[uint64]` to `map[string]`. At that point we would need to convert the map to `map[uint64]` before using it in the rest of the code.

winder:
## Summary

The `map[uint64]AssetParam` types are impossible to represent with swagger, they are most likely being serialized to `map[string]AssetParam` anyway, so this makes that more explicit.

I'm not sure what unintended side effects this may have...

This fixes #739

### DIFF
--- a/cmd/goal/account.go
+++ b/cmd/goal/account.go
@@ -508,7 +508,7 @@ var listCmd = &cobra.Command{
 					}
 				}
 
-				fmt.Printf("\t%20s %-8s%s (creator %s, ID %d%s%s%s%s)\n", assetDecimalsFmt(bal.Amount, assetDecimals), unitName, decimalInfo, bal.Creator, aid, assetName, assetURL, assetMetadata, frozen)
+				fmt.Printf("\t%20s %-8s%s (creator %s, ID %s%s%s%s%s)\n", assetDecimalsFmt(bal.Amount, assetDecimals), unitName, decimalInfo, bal.Creator, aid, assetName, assetURL, assetMetadata, frozen)
 			}
 		}
 	},
--- a/cmd/goal/accountsList.go
+++ b/cmd/goal/accountsList.go
@@ -227,7 +227,7 @@ func (accountList *AccountsList) outputAccount(addr string, acctInfo v1.Account,
 	if len(acctInfo.AssetParams) > 0 {
 		fmt.Printf("\t[created assets:")
 		for curid, params := range acctInfo.AssetParams {
-			fmt.Printf(" %d (%d %s)", curid, params.Total, params.UnitName)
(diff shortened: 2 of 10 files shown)


winder:
I'm not sure this is a swagger issue, other than the fact that it silently fails to generate a type. We're telling it that a type will be provided which would create invalid JSON, we should have been more considerate when designing the interface. And once this fix is in place, along with the rest of the doc automation, we will catch these design issues automatically when the documentation fails to regenerate.

PR #744, opened 2020-01-27
Set explicit IsRelay flag in netdeploy net to cloudspec converter

pzbitskiy:
* There no way to determine relay or node in a cloudspec, algonet
  uses IncomingPorts != "" heuristic to filter out relays and set host role.
  This does not work well because in additional to NetAddress out nodes
  expose APIEndpoint and MetricsURI so that IncomingPorts contains some values even for nodes.
* Having IsRelay flag would allow mostly precise separation in test scenarios
  since relays sit on dedicated hosts.
* Corresponding change for algonet is coming.

* Minor: fixed configure_dev.sh script name in check_deps.sh error message.


pzbitskiy:
Found a better solution.

PR #750, opened 2020-01-28
Configurable consensus protocol

tsachiherman:
## Summary

The algod binary comes packed with a built-in map `Consensus`. That map defines the supported consensus protocols as well as the upgrade path between them.
This change extend that by allowing an external configuration file (`consensus.json`) to override the existing consensus map entries.

The benefits of this change are - 
1. Released binaries no longer need to contain testing related consensus code.
2. The test-specific consensus configuration could be placed within the test itself.
3. Testing different nodes running different upgrades paths is now doable by configuration; binaries can remain identical.


gen/generate.go
```diff
@@ -51,7 +51,7 @@ type genesisAllocation struct {
 }
 
 // GenerateGenesisFiles generates the genesis.json file and wallet files for a give genesis configuration.
-func GenerateGenesisFiles(genesisData GenesisData, outDir string, verbose bool) error {
+func GenerateGenesisFiles(genesisData GenesisData, outDir string, verbose bool, consensus map[protocol.ConsensusVersion]config.ConsensusParams) error {
```
algonautshant: consensus is in addition to config.Consensus. It will be nice if the name reflects that. 
Suggestions: optionalConsensus or testConsensus

config/config.go
```diff
@@ -267,6 +263,86 @@ func init() {
...
+// PreloadConfigurableConsensusProtocols loads the configurable protocols from the data directroy
+// and merge it with a copy of the Consensus map. Then, it returns it to the caller.
+func PreloadConfigurableConsensusProtocols(dataDirectory string) (map[protocol.ConsensusVersion]ConsensusParams, error) {
+	consensusProtocolPath := filepath.Join(dataDirectory, ConfigurableConsensusProtocolsFilename)
+	file, err := os.Open(consensusProtocolPath)
+
+	if err != nil {
+		if os.IsNotExist(err) {
+			// this file is not required, only optional. if it's missing, no harm is done.
+			return nil, nil
```
pzbitskiy: a bit unnatural but I do not have better ideas

config/config.go
```diff
@@ -267,6 +263,86 @@ func init() {
...
+
+	staticConsensus := make(map[protocol.ConsensusVersion]ConsensusParams)
+	for consensusVersion, consensusParams := range Consensus {
+		// recreate the ApprovedUpgrades map since we don't want to modify the original one.
+		if consensusParams.ApprovedUpgrades != nil {
+			newApprovedUpgrades := make(map[protocol.ConsensusVersion]uint64)
+			for ver, when := range consensusParams.ApprovedUpgrades {
+				newApprovedUpgrades[ver] = when
+			}
+			consensusParams.ApprovedUpgrades = newApprovedUpgrades
```
pzbitskiy: consensusParams is a ref to the original/default entry in Consensus. what is the point of replacing its ApprovedUpgrades key by a new copy? Don't we simply want a new deep copy of Consensus map?

gen/generate.go
pzbitskiy: and another nit: a new param is kind of "src" param, and outDir is "dst". So it might make sense to put the new param as the second arg of the function.

node/node_test.go
```diff
@@ -118,7 +120,7 @@ func setupFullNodes(t *testing.T, proto protocol.ConsensusVersion, verificationP
 		if err != nil {
 			panic(err)
 		}
-
+		access.Close()
```
pzbitskiy: nice catch but shouldn't it also be done inside err branch after account.GenerateRoot?

config/config.go
tsachiherman: consensusParams is not a ref; it's a copy. And since the ApprovedUpgrades is a map, we don't want to modify the original map but rather create a copy of it.
Do you have another way to generate a deep copy of the Consensus map ?

config/config.go
algonautshant: Why not report error here? Silently discarding the load operation is confusing.

gen/generate.go
algonautshant: or better, configurableConsensus as is used elsewhere.

config/config.go
```diff
@@ -267,6 +267,86 @@ func init() {
...
+
+	encodedConsensusParams, err := json.Marshal(params)
+	if err != nil {
+		return err
+	}
+	err = ioutil.WriteFile(consensusProtocolPath, encodedConsensusParams, 0644)
+	return err
+}
+
+// LoadConfigurableConsensusProtocols loads the configurable protocols from the data directroy
```
algonautshant: It will be nice to add more details here: 
LoadConfigurableConsensusProtocols loads the configurable protocols and merges them with Consensus

nodecontrol/algodControl.go
```diff
@@ -375,3 +375,14 @@ func (nc NodeController) readGenesisJSON(genesisFile string) (genesisLedger book
...
 	return
 }
+
+// SetConsensus applies a new consensus settings which would get deployed before
+// any of the nodes starts
+func (nc NodeController) SetConsensus(consensus config.ConsensusProtocols) error {
+	return config.SaveConfigurableConsensus(nc.algodDataDir, consensus)
+}
+
+// GetConsensus rebuild the consensus version from the data directroy
```
algonautshant: rebuild: say loads and merges

config/config.go
tsachiherman: It's silently discard the error in one case only - there is no such file.
You could argue that we should perform that test at the caller, but there is no single point that would effectively use that value. ( i.e. other than applying the exact same logic. )

node/node_test.go
tsachiherman: I've moved it before the error checking.

gen/generate.go
tsachiherman: I've changes the function signature so it will receive a merged consensus as an input.

algoradam:
Just want to check that I understand the behavior correctly:
* `$DATADIR/consensus.json` contains a (json-encoded) map from consensus version to consensus params. It's fine if the file is missing.
* As long as you give ApprovedUpgrades in the params for a consensus version (even if ApprovedUpgrades is just an empty map), then we replace whatever ConsensusParams the code may have had for that ConsensusVersion with the params given in the json file. Any field in the ConsensusParams struct not specified in the json file is given the zero value.
* If you omit ApprovedUpgrades in the params for a consensus version, that means "Pretend this consensus version doesn't exist" -- it gets deleted from the map of known consensus versions (and the node will never vote to upgrade to this version)

Is this correct?

tsachiherman:
> Just want to check that I understand the behavior correctly:
> 
> * `$DATADIR/consensus.json` contains a (json-encoded) map from consensus version to consensus params. It's fine if the file is missing.
> * As long as you give ApprovedUpgrades in the params for a consensus version (even if ApprovedUpgrades is just an empty map), then we replace whatever ConsensusParams the code may have had for that ConsensusVersion with the params given in the json file. Any field in the ConsensusParams struct not specified in the json file is given the zero value.
> * If you omit ApprovedUpgrades in the params for a consensus version, that means "Pretend this consensus version doesn't exist" -- it gets deleted from the map of known consensus versions (and the node will never vote to upgrade to this version)
> 
> Is this correct?

Yes to all of these.

PR #754, opened 2020-01-28
Asynchronous EnsureDigest (updated)

derbear:
An updated version of #226. (original author is @Vervious, who is now at Cornell)

Original description:

In sum:

 - `certThresholds` now stage values in the proposal hierarchy, and essentially act like `softThresholds` (for the `event.period`)
    - note: we can receive `certThresholds` for the previous period (but not `soft`s, which aren't the freshest bundle). So now we can stage a value for the previous period, which is a side effect.
    - `certThreshold`s fast forward periods
 - adds `stageDigestAction`, distinct from `ensureAction`, to signal the ledger that it should attempt to fetch the block given a certificate. It is not a blocking operation.
 - `certThreshold` without payloads now trigger `stageDigestAction`
 - If we receive a payload, check if `cert` is freshest bundle; if so, finish round.

Note that accepting `certThreshold`s as freshest bundles certainly impacts relay rules. I went through them and any failure to relay new votes/bundles is benign, but this logic will need a review.


Vervious:
Amazing

derbear:
@Vervious 
In case you were wondering, I think the issue with the test before is that to reliably trigger a call to `EnsureDigest`, you needed to also drop the proposer's proposal-vote; otherwise, there is a small (1/total nodes) chance that the proposer's value wins leadership, and since the proposer already has his own block, giving him the certificate will cause him to move on to the next round and not trigger `EnsureDigest`.

algoradam:
I think we need to change bundleFresh() to consider cert bundles from any period fresh. (Or maybe it's fine and I'm misunderstanding the code somewhere.) Consider the following scenario:
- Since a cert bundle from any period is freshest, we might in period 1 see a cert bundle from period 0 where we don't have the block, and this becomes our freshest bundle that we'll transmit during recovery.
- Say the network advances to period 2 by seeing a next quorum. We don't advance because the nextThreshold gets eaten by our voteTrackerRound because it's less fresh than our cert bundle. Say the network then advances to period 3.
- We now ignore (and don't relay) all votes from period 3 and above because we're still on period 1. And our voteTrackerRound eats any bundle that isn't a cert-bundle because it's not fresh. So we're stuck in recovery on period 1. But we're periodically sending everyone our cert bundle, so it's fine, right?
- BUT everyone else's bundleFresh() considers our cert bundle stale because they're more than one period ahead of the bundle, so they ignore our freshest bundle. Now we're effectively partitioned from the network (and aren't relaying any votes) until we hear the block.

(I guess this is no worse than old EnsureDigest, which I think also makes us stop propagating any votes or bundles until we hear the block?)

tsachiherman:
> I think we need to change bundleFresh() to consider cert bundles from any period fresh. (Or maybe it's fine and I'm misunderstanding the code somewhere.) Consider the following scenario:
> 
> * Since a cert bundle from any period is freshest, we might in period 1 see a cert bundle from period 0 where we don't have the block, and this becomes our freshest bundle that we'll transmit during recovery.
> * Say the network advances to period 2 by seeing a next quorum. We don't advance because the nextThreshold gets eaten by our voteTrackerRound because it's less fresh than our cert bundle. Say the network then advances to period 3.
> * We now ignore (and don't relay) all votes from period 3 and above because we're still on period 1. And our voteTrackerRound eats any bundle that isn't a cert-bundle because it's not fresh. So we're stuck in recovery on period 1. But we're periodically sending everyone our cert bundle, so it's fine, right?
> * BUT everyone else's bundleFresh() considers our cert bundle stale because they're more than one period ahead of the bundle, so they ignore our freshest bundle. Now we're effectively partitioned from the network (and aren't relaying any votes) until we hear the block.
> 
> (I guess this is no worse than old EnsureDigest, which I think also makes us stop propagating any votes or bundles until we hear the block?)

Yes, it's not any worse than it currently is. A node can get partitioned from the network if it haven't heard the block, and we have no recovery mechanism for that within the agreement. However, you might have just found a bug in the catchup: we currently verify that the given certificate in ensure digest aligns with the block digest of the block header. That's all fine as long as we keep the same proposal ( i.e. including reproposing ). If we attempt to make a new proposal however, the block might still be legit, but have a different header.

derbear:
@algoradam 
1. I do agree that updating `bundleFresh` to actually transmit cert-bundles makes our treatment of "freshness" more consistent. It's a good idea--I'll go make that change.
2. I think this is slightly better than the old behavior in that a node which holds the certificate but not the block will now accept blocks via gossip (as opposed to only via catchup because the agreement main loop is blocked on `EnsureDigest`). If our safety guarantees have held (inductively), then the network will stay with the block referred to by the certificate. Then presumably one of the nodes whose cert-votes were seen by the relay will _eventually_ transmit the block to the certificate-holding node (maybe via a few intermediate nodes who have only seen next-quorums). 

@tsachiherman 
I believe that once we get a certificate, the agreement protocol guarantees that we will always repropose the same block. If that's ever not the case, then we have a safety bug and may fork.

Vervious:
+1 @algoradam, I agree with cert bundles deserving special relay status.

agreement/player.go
```diff
@@ -262,23 +266,39 @@ func (p *player) handleThresholdEvent(r routerHandle, e thresholdEvent) []action
 	var actions []action
 	if e.t() == certThreshold {
 		// this threshold must be for p.Round, and originates from the vote SM tree
-		cert := Certificate(e.Bundle)
+		// First, stage the threshold if not already. It is entirely fine to stage values for periods in the past.
+		r.dispatch(*p, e, proposalMachine, p.Round, p.Period, 0)
+		// Now, also check if we have the block.
 		res := stagedValue(*p, r, e.Round, e.Period)
```
algoradam: After discussion, should be p.Period so that we're reading the staged value from the current period

data/datatest/impls.go
```diff
@@ -159,13 +159,9 @@ func (i ledgerImpl) EnsureDigest(cert agreement.Certificate, quit chan struct{},
...
 
-	select {
-	case <-quit:
-		return
-	case <-i.Wait(r):
-		if !consistencyCheck() {
-			err := fmt.Errorf("Wait channel fired without matching block in round %v", r)
-			panic(err)
-		}
+	<-i.Wait(r)
```
algoradam: The actual EnsureDigest is nonblocking now, should this function also be made nonblocking?

algoradam:
Other than the p vs e thing in handleThresholdEvent, I haven't found any issues.
I would recommend adding way more comments though!

agreement/player.go
derbear: Actually, `e.Round` and `e.Period` are correct; the router sends the thresholds based on `e.Round` and `e.Period` to the right place, and `p.Round` and `p.Period` are misleading...however, the system must still ensure that proposals in old periods are not dropped.

PR #762, opened 2020-01-29
Implement local net template generation with netgoal

pzbitskiy:
* Usage:
  netgoal generate -n 1 -R 1 -w 10 -o mynettemplate.json -r . -t goalnet
  goal network create -t mynettemplate.json -r mynet -n mynet

* Remove duplicate definitions from netdeploy/networkTemplate

### DIFF
--- a/cmd/netgoal/generate.go
+++ b/cmd/netgoal/generate.go
@@ -29,6 +29,7 @@ import (
 	"github.com/spf13/cobra"
 
 	"github.com/algorand/go-algorand/gen"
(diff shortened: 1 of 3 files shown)


pzbitskiy:
`netgoal gererate` creates network templates by taking # hosts, # nodes, # relalys. Then this template might be used with `algonet` for cloud deployment. I re-used the same logic but added a new output format compatible with `goal network create`. It does not have per-host allocation and includes genesis.

tsachiherman:
Sounds good!

PR #763, opened 2020-01-29
Release build pipeline step 1: Build, package, sign, deploy to staging

btoll:
Sorry for the size of this :(

There is a lot to this but it can be broken down into the following basic steps:
1. build
1. package
1. sign
1. deploy to staging

The Jenkinsfile will call shell scripts in the `./release/ci/stage/` directory which will in turn call the "real" task a directory above it in the `./release/ci/` directory (well, this is true except for any `aws` cli commands).

I tried to isolate the logic of the building, packaging and signing of the deb and rpm packages to the `./release/deb` and `./release/rpm` directories, respectively.  This may a bit more redirection than necessary but I thought it would help from a conceptual point of view.

Importantly, the binaries are compiled in the centos docker container and then moved into `$GOPATH/bin` on the host machine.  This means that it gets its own cloned `go-algorand` repo and `make` is of course run on this source.

These binaries are then used for both the rpm and deb packages.

Note that the removed code is all the testing bits that will be done in a separate pipeline.


buildnumber.dat
```diff
@@ -0,0 +1 @@
+3
```
tsachiherman: Please undo this change. We don't want the build number.dat on the master.

btoll: Removed

bricerisingalgorand:
I wish a PR this dense could've sat for a couple more days. There are several things we need to chat about in here

PR #764, opened 2020-01-29
back out locking added in c78ada09f230a3c66cd934860700f93ff31a93eb

algobolson:
This seemed to be needed at the time, but now seems to be causing a major slowdown and removing it doesn't break any tests.

### DIFF
--- a/util/execpool/backlog.go
+++ b/util/execpool/backlog.go
@@ -19,15 +19,12 @@ package execpool
 import (
 	"context"
 	"sync"
-
-	"github.com/algorand/go-deadlock"
 )
 
 // A backlog for an execution pool. The typical usage of this is to
 // create non-blocking queue which would get executed once the execution pool is ready to accept new
 // tasks.
 type backlog struct {
-	mu        deadlock.Mutex
 	pool      ExecutionPool
 	wg        sync.WaitGroup
 	buffer    chan backlogItemTask
@@ -35,7 +32,6 @@ type backlog struct {
 	ctxCancel context.CancelFunc
 	owner     interface{}
 	priority  Priority
-	quit      bool
 }
 
 type backlogItemTask struct {
@@ -82,25 +78,11 @@ func (b *backlog) GetParallelism() int {
 
 // IsFull test to see if the input buffer is full.
 func (b *backlog) IsFull() bool {
-	b.mu.Lock()
-	defer b.mu.Unlock()
 	return len(b.buffer) == cap(b.buffer)
 }
 
 // Enqueue enqueues a single task into the backlog
 func (b *backlog) Enqueue(enqueueCtx context.Context, t ExecFunc, arg interface{}, priority Priority, out chan interface{}) error {
-	b.mu.Lock()
-	defer b.mu.Unlock()
-	if b.quit {
-		select {
-		case <-enqueueCtx.Done():
-			return enqueueCtx.Err()
-		case <-b.ctx.Done():
-			return b.ctx.Err()
-		default:
-			return nil
-		}
-	}
 	select {
 	case b.buffer <- backlogItemTask{
 		enqueuedTask: enqueuedTask{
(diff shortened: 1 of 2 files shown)


util/execpool/backlog.go
```diff
@@ -120,18 +102,6 @@ func (b *backlog) Enqueue(enqueueCtx context.Context, t ExecFunc, arg interface{
...
-			return enqueueCtx.Err()
-		case <-b.ctx.Done():
-			return b.ctx.Err()
-		default:
-			return nil
-		}
-	}
 	select {
 	case b.buffer <- backlogItemTask{
 		enqueuedTask: enqueuedTask{
```
justicz: github won't let me comment on lines not in this PR, but it looks like this doesn't totally undo the changes done to this file in #700 b/c of the `case <-b.ctx.Done()` case

justicz: (in both `Enqueue` and `EnqueueBacklog`)

algobolson: This may actually be the minimum change needed to fix things. I think without this Enqueue() and EnqueueBacklog() will panic if Shutdown() has been called because they try to send on a closed b.buffer . The internal Context b.ctx is the only thing that can signal that Enqueue shuoldn't actually try to append to b.buffer

algobolson:
@tsachiherman this fixes the performance regression introduced at #700 ; I have measured it so.

util/execpool/backlog.go
derbear: Is there a race between context cancellation and closing the buffer?

derbear:
I'm not sure how easy this is, but the backlog struct seems somewhat messy, and it's hard to show that it's correct. Maybe it's worth simplifying this part of the code.

algobolson:
the net diff for backlog.go from before #700 to this is
```
$ git diff 4f4a372f4c204d89457be0ea3a558e77f9f6726b..dd6ee4cf4d35aec8b28cd42cf1472f73cb58e37e .
diff --git a/util/execpool/backlog.go b/util/execpool/backlog.go
index bdce408a..62899fcb 100644
--- a/util/execpool/backlog.go
+++ b/util/execpool/backlog.go
@@ -95,6 +95,8 @@ func (b *backlog) Enqueue(enqueueCtx context.Context, t ExecFunc, arg interface{
                return nil
        case <-enqueueCtx.Done():
                return enqueueCtx.Err()
+       case <-b.ctx.Done():
+               return b.ctx.Err()
        }
 }
 
@@ -112,6 +114,8 @@ func (b *backlog) EnqueueBacklog(enqueueCtx context.Context, t ExecFunc, arg int
                return nil
        case <-enqueueCtx.Done():
                return enqueueCtx.Err()
+       case <-b.ctx.Done():
+               return b.ctx.Err()
        }
 }
```

util/execpool/backlog.go
```diff
@@ -82,25 +78,11 @@ func (b *backlog) GetParallelism() int {
 
 // IsFull test to see if the input buffer is full.
 func (b *backlog) IsFull() bool {
```
derbear: Is this method actually called anywhere?

tsachiherman:
Not anymore. I think we should remove it too ( as it seems to require the
lock which we just removed.. )

>

tsachiherman:
I think that the following line is why you're seeing this issue:
```txvalidator.txgroups = make(chan []transactions.SignedTxnWithAD, 10)```

if you were to resize this channel to the size of the groups, it would not make the blocking on the backlog, and could restore the usage profile on the execution pool.

Also, I'm not sure if the #700 change correctly drops expired transactions ( i.e. pre-validation, naturally ).

tsachiherman:
btw - don't take me wrong : This PR could still be making a great performance booster

algobolson:
I believe this change is strictly better than both pre- #700 and better than current master

algobolson:
Further experimentation shows this is still probably only to the level of "probably won't crash" and "slightly less likely to crash than 2 weeks ago"

derbear:
As discussed, though it triggers a panic, the execution race happens only on `goal node stop` and so it's probably not critical to remove it. I'd still like to fix the race condition at some point.

PR #765, opened 2020-01-30
Make dockerignore file

bricerisingalgorand:
This file will prevent docker build contexts from loading certain files when creating docker build contexts. I just made it a copy of .gitignore since those files don't seem to be necessary for any current Dockerfile for go-algorand.

## Summary

This is preventing docker from loading unnecessary files (primarily what's in /tmp) into docker build contexts. This will speed up docker builds using the root dir of this project as a context.

## Test Plan

I made sure our docker build process still worked


bricerisingalgorand:
Thanks @pzbitskiy, closing this one since this commit is part of that PR. I'll fix both together

PR #766, opened 2020-01-30
Improve net templates support

pzbitskiy:
While running a local test with 10k equal accounts I found few problems in netgoal and goal. This PR addresses them, special thanks to Max and Brian.

* Fix file descriptors leak in 'goal account'. Now goal can import
  more than maxfiles keys
* Fix uint overflow in stake distribution validation.
  Details: values 10 and -110 were casted to uint and sum up to 100 pct
  with 32 bits overflow
* Allow pct fraction of stake in goal net templates
* Fix stake distribution in netgoal.generate: it always produces pcts
  and not values in algos as was incorrectly thought before
* Add tests for netdeploy.Validate()


cmd/netgoal/generate.go
```diff
@@ -426,19 +426,13 @@ func saveGoalTemplateToDisk(template netdeploy.NetworkTemplate, filename string)
 	return err
 }
 
-func generateWalletGenesisData(wallets, npnHosts int, totalStake int64) gen.GenesisData {
+func generateWalletGenesisData(wallets, npnHosts int) gen.GenesisData {
 	data := gen.DefaultGenesis
 	if npnHosts > 0 {
 		wallets = 2 * wallets
 	}
-	// cloud templates use percentage of stake
```
pzbitskiy: it was my incorrect assumption after briefly looking into net template validation code in goal. This change simply reverts pct-based calculation.

netdeploy/networkTemplate.go
```diff
@@ -169,17 +170,21 @@ func loadTemplateFromReader(reader io.Reader, template *NetworkTemplate) error {
 func (t NetworkTemplate) Validate() error {
 	// Genesis wallet percentages must add up to 100
 	// Genesis account names must be unique
-	totalPct := uint(0)
+	totalPct := big.NewFloat(float64(0))
 	accounts := make(map[string]bool)
 	for _, wallet := range t.Genesis.Wallets {
-		totalPct += uint(wallet.Stake)
+		if wallet.Stake < 0 {
```
pzbitskiy: this fixes 10 + (-110) summation

netdeploy/networkTemplate.go
```diff
@@ -169,17 +170,21 @@ func loadTemplateFromReader(reader io.Reader, template *NetworkTemplate) error {
...
+			return fmt.Errorf("invalid template: negative stake on Genesis account %s", wallet.Name)
+		}
+		totalPct = totalPct.Add(totalPct, big.NewFloat(wallet.Stake))
```
pzbitskiy: this line enables validation for fractions of pct

PR #769, opened 2020-01-31
Make goal node state change commands systemd aware

bricerisingalgorand:
## Summary

I added a property to libgoal/system.go where we can set whether or not our algod process is managed by systemd. This adds the feature, but without automation for ensuring that the field is set to true on all systemd installs

## Test Plan

I ran it locally on OSX and linux and it appeared to work appropriately. Making a PR to ensure all integration tests are ran

### DIFF
--- a/libgoal/system.go
+++ b/libgoal/system.go
@@ -26,7 +26,8 @@ import (
 type SystemConfig struct {
 	// SharedServer is true if this is a daemon on a multiuser system.
 	// If not shared, kmd and other files are often stored under $ALGORAND_DATA when otherwise they might go under $HOME/.algorand/
-	SharedServer bool `json:"shared_server,omitempty"`
+	SharedServer   bool `json:"shared_server,omitempty"`
+	SystemdManaged bool `json:"systemd_managed,omitempty"`
 }
 
 // map data dir to loaded config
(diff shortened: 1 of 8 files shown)


libgoal/system.go
```diff
@@ -72,3 +73,16 @@ func AlgorandDataIsPrivate(dataDir string) bool {
...
+// if not, algod will be managed as an indivudal process for the dir
+func AlgorandDaemonSystemdManaged(dataDir string) bool {
+	if dataDir == "" {
+		return false
+	}
+	sc, err := ReadSystemConfig(dataDir)
+	if err != nil {
+		return false
+	}
+	return sc.SystemdManaged
```
pzbitskiy: where do we set `SystemdManaged` or `systemd_managed` ?

bricerisingalgorand: In the system.json file of the data dir

algobolson: go ahead and add this to `installer/system.json` which is included in RPM and APT packages which install on systemd distros?

bricerisingalgorand: Cool, I'm adding that with my next commit once I have the unit tests working

bricerisingalgorand:
I added tests using expect to try to make sure messages coming from goal on systemd enabled computers look appropriate. This tests that the goal cli is using the field in system.json appropriately. I also added systemd_managed: true as a default value in our system.json file that gets published with linux installs. I looked into wring a full e2e test to make sure that linux installs have that set and it's working appropriately, but our automation for creating linux packages needs a lot of improvement before it can be used conveniently for that purpose. I think we should hold off on writing that test once our automation is in a better state.

PR #771, opened 2020-01-31
Make util.ExecAndCaptureOutput able to process large output

pzbitskiy:
* In case of large amount of data written to stdout/stderr
  from the wrapped command the process is blocked until
  stdout/stderr buffers cleared.
* Old implementation waited until cmd return and then read stdout/stderr.
* New implementation reads stdout/stderr pipes in goroutines.

### DIFF
--- a/util/process.go
+++ b/util/process.go
@@ -17,9 +17,11 @@
 package util
 
 import (
-	"io/ioutil"
+	"io"
 	"os"
 	"os/exec"
+	"sync"
+	"time"
 )
 
 // ExecAndCaptureOutput runs the specified command and args and captures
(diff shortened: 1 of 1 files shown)


util/process.go
```diff
@@ -38,19 +39,40 @@ func ExecAndCaptureOutput(command string, args ...string) (string, string, error
...
+			default:
+			}
+			buf := make([]byte, 1024)
+			read, _ := input.Read(buf)
+			if read > 0 {
+				*output = append(*output, buf[0:read]...)
+			}
+			time.Sleep(time.Microsecond)
+		}
+		wg.Done()
```
tsachiherman: could you move this to the beginning of the function with a defer statement ?

util/process.go
```diff
@@ -38,19 +39,40 @@ func ExecAndCaptureOutput(command string, args ...string) (string, string, error
...
-		err = errIO
+	var wg sync.WaitGroup
+	wg.Add(2)
+
+	reader := func(input *os.File, output *[]byte) {
+		completed := false
+		for !completed {
+			select {
+			case <-done:
+				completed = true
```
tsachiherman: with the defer mentioned below, you should be able to return here instead of having this varaible.

util/process.go
```diff
@@ -38,19 +39,40 @@ func ExecAndCaptureOutput(command string, args ...string) (string, string, error
 	subcmd.Stdout = wStdout
 	subcmd.Stderr = wStderr
 
-	err = subcmd.Run()
+	outputStdout := make([]byte, 0, 10240)
+	outputStderr := make([]byte, 0, 10240)
```
tsachiherman: given that we don't expect too much errors ( hopefully ), I would set it to a lower number. it will grow dynamically anyway.

util/process.go
pzbitskiy: done

util/process.go
pzbitskiy: No, read still needs to be done to read rest of the data

util/process.go
pzbitskiy: done

PR #774, opened 2020-02-01
Move fetcher client into catchup

tsachiherman:
## Summary

In the previous PR, we moved the ensure digest fetching logic into the catchup. This is a followup PR. In this PR, we're moving the fetcher client into the catchup as well. ( since by now, the catchup is the only component that uses the fetcher ).

Note that the "ledger service" is still in the RPCS and would likely to remain there. The catchup contains all the "client-side" code.


rpcs/fetcher_test.go
```diff
@@ -1,414 +0,0 @@
-// Copyright (C) 2019-2020 Algorand, Inc.
```
algonautshant: Can this file be moved to /catchup/fetcher_test.go instead of deleting it?

catchup/fetcher.go
```diff
@@ -63,7 +64,7 @@ type FetcherFactory interface {
 type NetworkFetcherFactory struct {
 	net       network.GossipNode
 	peerLimit int
-	fs        *WsFetcherService
+	fs        *rpcs.WsFetcherService
```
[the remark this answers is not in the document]
tsachiherman: The `WsFetcherService` is the server side portion. I'd like to keep the separation between client-side and server-side components.

rpcs/fetcher_test.go
tsachiherman: While the file has been deleted, all the test that it included were transported into the the catchup library. ( some of these have been re-written to various degree )

PR #780, opened 2020-02-03
Tunnel outgoing connection via a rate limiting dialer

algonautshant:
## Summary

The Dialer now waits to make sure the connections comply with connectionsRateLimitingCount.
Special care is taken to prevent the deadlock detection to trigger on the phonebook lock:
The wait happens outside the phonebook lock.


network/wsNetwork.go
```diff
@@ -1544,10 +1538,13 @@ func (wn *WebsocketNetwork) numOutgoingPending() int {
...
 }
 
-var websocketDialer = websocket.Dialer{
-	Proxy:             http.ProxyFromEnvironment,
-	HandshakeTimeout:  45 * time.Second,
-	EnableCompression: false,
+// GetNetTransport returns an http.Transport that limits the number of connection
+// to comply with connectionsRateLimitingCount.
+func (wn *WebsocketNetwork) GetNetTransport() *http.Transport {
+	transport := http.DefaultTransport.(*http.Transport)
```
tsachiherman: Changing the "stock" transport isn't a good idea. ( as the data race suggests ).
instead, maybe something in the lines of 
```
func (wn *WebsocketNetwork) GetNetTransport() *http.Transport {
	// we want to create a copy of the stock transport object here.
	transport := *http.DefaultTransport.(*http.Transport)
	transport.DialContext = wn.GetDialer().DialContext
	return &transport
}
```

network/rateLimitedTransport.go
```diff
@@ -0,0 +1,61 @@
...
+import (
+	"context"
+	"net"
+	"net/http"
+	"time"
+)
+
+// RateLimitedTransport is a wrapper around the http.Transport that overrides the Dial and DialContext functions.
+// It limits the rate of the outgoing connections to comply with connectionsRateLimitingCount
+type RateLimitedTransport struct {
```
tsachiherman: you've tried to create some new "hybrid" creature for transport and dialer; Not so sure it's a good idea, since it will make it much harder down the road. Plus, you didn't really initialized the transport so I'm not even sue what this even mean.

network/phonebook.go
```diff
@@ -164,18 +170,17 @@ func (e *phonebookEntries) waitForConnectionTime(addr string) (addrInPhonebook,
 	// If there are max number of connections within the time window, wait
 	numElts := len(e.data[addr].recentConnectionTimes)
```
pzbitskiy: nit: move `numElts` above and reuse in `for` loop

network/wsNetwork.go
```diff
@@ -505,7 +507,8 @@ func (wn *WebsocketNetwork) GetPeers(options ...PeerOption) []Peer {
 			var addrs []string
 			addrs = wn.phonebook.GetAddresses(1000)
 			for _, addr := range addrs {
-				outPeers = append(outPeers, &wsPeerCore{net: wn, rootURL: addr})
+				peerCore := makePeerCore(wn, addr, wn.GetRoundTripper(), "" /*origin addres*/)
```
pzbitskiy: origin addres -> origin address

network/phonebook.go
```diff
@@ -136,24 +139,27 @@ func (e *phonebookEntries) updateRetryAfter(addr string, retryAfter time.Time) {
...
+		return addrInPhonebook, 0 /* not unsed */, curTime /* not unsed */
 	}
 
 	var timeSince time.Duration
 	var numElmtsToRemove int
 	// Remove from recentConnectionTimes the times later than ConnectionsRateLimitingWindowSeconds
 	for numElmtsToRemove < len(e.data[addr].recentConnectionTimes) {
 		timeSince = curTime.Sub((e.data[addr].recentConnectionTimes)[numElmtsToRemove])
-		if timeSince > e.connectionsRateLimitingWindow {
+		if timeSince >= e.connectionsRateLimitingWindow {
```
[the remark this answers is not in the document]
algonautshant: If the time window is 5 seconds, and time elapsed (timeSince) is 5 sec, then the connection time can be safely removed. No need to wait to go over 5 sec to qualify for greater than. Equal is sufficient.

network/phonebook.go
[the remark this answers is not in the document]
algonautshant: I did not want to use numElts in the loop, and again down here because the number of elements may change in
e.popNElements(numElmtsToRemove, addr)
between the loop and its usage here. In the event it is not updated again here, it will be a bug.

PR #782, opened 2020-02-04
Restructure release/ dir

btoll:
## Shared Directory Structure Pattern

The `release/` directory will have a structure that looks like the following:

        release/
            build/
            test/
            prod/
            ...
            <scripts common to all builds {.sh,.md,Dockerfile,etc}>

Each subdirectory of `release/` represents a build pipeline and is self-contained.  In other words, it will have its own `Jenkinsfile` and whatever else it needs to perform its job.

The pattern is the following (all subdirecties of `release/` will follow this pattern):

        test/                       // Maps to "test" build pipeline.
            Jenkinsfile
            stage/
                setup/              // Maps to "setup" stage in Jenkinsfile.
                    run.sh
                    task.sh
                test/               // Maps to "test" stage in Jenkinsfile.
                    run.sh
                    task.sh
            deb/
                *.sh
                testDebian.exp
            rpm/
                *.sh

The `Jenkinsfile` is self-explanatory, so I'll gloss over it.  `stage/` will contain directories that directly map to a defined stage in the `Jenkinsfile`.  Each directory will contain two files:

1. run.sh
1. task.sh

`run.sh` will be the script called in the `Jenkinsfile` and contains any `aws cli` commands.  `task.sh` is then initiated by `run.sh` and contains the logic for the particular build stage.

The `rpm` and `deb` directories will contain scripts that are relevant to their respective packaging formats.

You may rely on this pattern, and this makes any troubleshooting relatively straightforward as you know where to find the code that executes a particular task due to setting reasonable expectations.


scripts/release/README.md
```diff
@@ -1,24 +1,78 @@
...
+
+## Build Outcomes
+
+This section briefly describes the expected outcomes of the current build pipelines.
+
+1. build
+
+    The result of running this job will be to put the build artifacts and their detached signatures in the staging `algorand-builds` bucket.
+
+    In addition, the build logs will be placed into the AWS `algorand-devops-misc` S3 bucket under `buildlog`.
```
tsachiherman: I think that I'd prefer to have all the pipeline output goes into the Algorand-builds directory.
however, from that point and on, I'm not over picky. If you would make it under ( let's say ) `algorand-builds/build-logs/channel/beta/` that would be fine.

scripts/release/socket.sh
```diff
@@ -12,15 +12,17 @@ fi
 
 INSTANCE=$1
 gpgp=$(find /usr/lib/gnupg{2,,1} -type f -name gpg-preset-passphrase 2> /dev/null)
-KEYGRIP=$(gpg -K --with-keygrip --textmode dev@algorand.com | grep Keygrip | head -1 | awk '{ print $3 }')
+#KEYGRIP=$(gpg -K --with-keygrip --textmode dev@algorand.com | grep Keygrip | head -1 | awk '{ print $3 }')
```
tsachiherman: this file doesn't looks right.

scripts/release/test/stage/setup/run.sh
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env bash
+# shellcheck disable=2029
+
+# Path(s) are relative to the root of the Jenkins workspace.
+INSTANCE=$(cat scripts/release/tmp/instance)
+
+rm -rf pkg/* && mkdir -p pkg/"$FULLVERSION"
```
tsachiherman: I'm pretty sure `ben-test-2.0.3` is not where we want to end up with.

scripts/release/README.md
btoll: Sure, I can change that.  It's the old location.

scripts/release/socket.sh
btoll: Yeah, one of the problems I found is that the keygrip, which is generated from the public key:
```
To identify a key we use a thing called keygrip which is the SHA-1 hash of an canonical encoded S-Expression of the public key as used in Libgcrypt. For the purpose of this interface the keygrip is given as a hex string. The advantage of using this and not the hash of a certificate is that it will be possible to use the same keypair for different protocols, thereby saving space on the token used to keep the secret keys. 
```
https://www.gnupg.org/documentation/manuals/gnupg/Agent-Protocol.html

If a key has a subkey, then there will be more than one keygrip.  I've decided to hardcode the keygrips in for now, and the reason I didn't remove or change the code yet is that I would like to revisit this.

Since there is going to be at least one more iteration of the release build before pencils down, I thought it would be ok to leave as-is.

scripts/release/test/stage/setup/run.sh
btoll: :)  Will change.

PR #788, opened 2020-02-06
Release build pipeline step 3: Added "prod" pipeline to `release/`

btoll:
When Jenkins pauses the pipeline at the `snapshot` stage, manually forward the `gpg-agent` and invoke the following commands at the remote shell prompt:
```
export AWS_ACCESS_KEY_ID=
export AWS_SECRET_ACCESS_KEY=
go/src/github.com/algorand/go-algorand/scripts/release/prod/stage/snapshot.sh
```


scripts/release/prod/stage/sync/run.sh
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env bash
+# shellcheck disable=2029
+
+# Path(s) are relative to the root of the Jenkins workspace.
+STAGING="$1"
+PROD="$2"
+CHANNEL="$3"
+RELEASE="$4"
+RSTAMP=$(scripts/release/prod/reverse_hex_timestamp)
```
tsachiherman: I think that you need to write it as ./script rather than just script, as the current directory cannot be assume to be in the path, right ?

btoll: It should be fine, as the paths are relative to the root of the workspace.  I can change it to be preceded by `./`, though, that probably is safer.

PR #789, opened 2020-02-06
Allow creation of asset transactions while offline.

winder:
## Summary

Using the suggested params endpoint while creating an asset prevents users from creating the transaction while offline or with a node which has not completed catchup. When this scenario is detected initialize the parameters to V18.


winder:
Looks like there are other issues with creating these transactions using goal and an offline node, this may not be such a straight forward change.

PR #798, opened 2020-02-10
require Encode() and Decode() to take msgp'ed types

zeldovich:
This ensures that calls to protocol.Encode() and protocol.Decode()
are going to hit fast msgp-generated encoders and decoders.

There are some places where we can't use msgp-generated code yet, for
whatever reason, and those still invoke the reflection-based go-codec,
using protocol.EncodeRefect() and protocol.DecodeReflect().

The main intent of this commit is to clearly identify places where we
still invoke go-codec, and fix some trivial cases (like passing a struct
to protocol.Encode by value instead of by pointer).

Later on, we can go through the calls to protocol.EncodeReflect() and
protocol.DecodeReflect() to see if we can get rid of the harder cases,
to reduce or eliminate the use of go-codec altogether.

## Summary

## Test Plan


protocol/codec.go
```diff
@@ -190,10 +170,9 @@ func DecodeMsgp(b []byte, objptr msgp.Unmarshaler) (err error) {
...
 // into an object instance pointed to by objptr.
-func Decode(b []byte, objptr interface{}) error {
-	msgp, ok := objptr.(msgp.Unmarshaler)
-	if ok && msgp.CanUnmarshalMsg(msgp) {
-		return DecodeMsgp(b, msgp)
+func Decode(b []byte, objptr msgp.Unmarshaler) error {
+	if objptr.CanUnmarshalMsg(objptr) {
+		return DecodeMsgp(b, objptr)
 	}
 	return DecodeReflect(b, objptr)
```
justicz: Might be worth adding logging here since we should no longer ever hit this case

protocol/codec.go
```diff
@@ -109,33 +109,13 @@ func EncodeMsgp(obj msgp.Marshaler) []byte {
...
 // Encode returns a msgpack-encoded byte buffer for a given object.
-func Encode(obj interface{}) []byte {
-	msgp, ok := obj.(msgp.Marshaler)
-	if ok && msgp.CanMarshalMsg(msgp) {
-		return EncodeMsgp(msgp)
+func Encode(obj msgp.Marshaler) []byte {
+	if obj.CanMarshalMsg(obj) {
+		return EncodeMsgp(obj)
 	}
 	return EncodeReflect(obj)
```
justicz: same here

zeldovich:
Sorry for taking a while to get back to this..  I was mulling over how to avoid the undesirable inheritance of `MarshalMsg` through embedded fields, but couldn't think of anything nice.  I'll add a logging statement for when we encode a non-msgp'ed thing with an embedded msgp'ed field.

crypto/msgp_gen.go
```diff
@@ -3,6 +3,7 @@ package crypto
 // Code generated by github.com/algorand/msgp DO NOT EDIT.
 
 import (
+	"C"
 	_ "runtime/cgo"
 	_ "unsafe"
```
derbear: Out of curiosity, are these imports required for something important, or are they just an artifact of codegen?

zeldovich: I think it's just an artifact -- msgp runs `goimports` to clean up the import statements, and I don't really know what rules `goimports` applies, other than the fact that it makes sure that the compiler doesn't complain about unused or missing imports.

zeldovich:
Well, what do you know -- we actually had an instance of the embedded field problem!  `crypto.SignatureSecrets` embedded `crypto.SignatureVerifier`, and only the latter had msgp-generated marshal/unmarshal methods..

justicz:
Probably the test was working correctly and falling back to go-codec before, and in this PR we added logging?

zeldovich:
Max is exactly right: this was purely a performance bug, strictly speaking, because we did a reflection check to look for precisely these cases.

PR #799, opened 2020-02-10
Add support for multiple network protocol versions

tsachiherman:
## Summary

Having the ability for the network library to know which features available on the other peer could help us making more efficient communication choices. This PR use the customary http-like accept model to accomplish that. i.e. the client sends the `X-Algorand-Accept-Version` to denote which versions it supports. The server compares that list against it's internal supported versions and response with `X-Algorand-Version` which select the version which would be used for this session.


network/wsNetwork.go
```diff
@@ -826,6 +826,28 @@ func (wn *WebsocketNetwork) checkIncomingConnectionLimits(response http.Response
 	return http.StatusOK
 }
 
+// checkProtocolVersionMatch test ProtocolAcceptVersionHeader and ProtocolVersionHeader headers from the request/response and see if it can find a match.
+func (wn *WebsocketNetwork) checkProtocolVersionMatch(otherHeaders http.Header) (matchingVersion string, otherVersion string) {
```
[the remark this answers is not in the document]
tsachiherman: I think it's quite self-explanatory - matchingVersion is the version that matches between the request and the local node. otherVersion is returned if the other side only supports a single version, which isn't supported by the local node.

network/wsNetwork.go
```diff
@@ -826,6 +826,28 @@ func (wn *WebsocketNetwork) checkIncomingConnectionLimits(response http.Response
...
+	otherAcceptedVersions := otherHeaders[textproto.CanonicalMIMEHeaderKey(ProtocolAcceptVersionHeader)]
+	for _, otherAcceptedVersion := range otherAcceptedVersions {
```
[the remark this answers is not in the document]
tsachiherman: right; look on the comment we have for the SupportedProtocolVersions.

network/wsNetwork.go
```diff
@@ -826,6 +826,28 @@ func (wn *WebsocketNetwork) checkIncomingConnectionLimits(response http.Response
...
+		}
+	}
+
+	otherVersion = otherHeaders.Get(ProtocolVersionHeader)
+	for _, supportedProtocolVersion := range SupportedProtocolVersions {
+		if supportedProtocolVersion == otherVersion {
+			return supportedProtocolVersion, otherVersion
+		}
+	}
+	return matchingVersion, otherVersion
```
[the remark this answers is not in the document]
tsachiherman: sure.

network/wsNetwork.go
```diff
@@ -899,6 +908,19 @@ func (wn *WebsocketNetwork) ServeHTTP(response http.ResponseWriter, request *htt
...
+	if matchingVersion == "" {
+		wn.log.Infof("new peer %s version mismatch, mine=%v theirs=%s, headers %#v", request.RemoteAddr, SupportedProtocolVersions, otherVersion, request.Header)
+		networkConnectionsDroppedTotal.Inc(map[string]string{"reason": "mismatching protocol version"})
+		response.WriteHeader(http.StatusPreconditionFailed)
+		message := fmt.Sprintf("Requested version %s not in %v mismatches server version", otherVersion, SupportedProtocolVersions)
+		n, err := response.Write([]byte(message))
+		if err != nil {
+			wn.log.Warnf("ws failed to write response '%s' : n = %d err = %v", message, n, err)
+		}
+		return
```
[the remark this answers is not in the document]
tsachiherman: no; ServeHTTP have no return value. The response already contained the error.

network/wsNetwork.go
```diff
@@ -1426,9 +1450,15 @@ func (wn *WebsocketNetwork) getDNSAddrs(dnsBootstrap string) []string {
 	return srvPhonebook
 }
 
-// ProtocolVersionHeader HTTP header for protocol version. TODO: this may be unneeded redundance since we also have url versioning "/v1/..."
+// ProtocolVersionHeader HTTP header for protocol version.
 const ProtocolVersionHeader = "X-Algorand-Version"
 
+// ProtocolAcceptVersionHeader HTTP header for accept protocol version. Client use this to advertise supported protocol versions.
```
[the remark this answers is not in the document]
tsachiherman: this is the common accept "model" for http headers. Feel free to suggest alternative comment.

network/wsPeer.go
```diff
@@ -148,6 +148,9 @@ type wsPeer struct {
 
 	// createTime is the time at which the connection was established with the peer.
 	createTime time.Time
+
+	// peer version ( this is one of the version supported by the current node and listed in SupportedProtocolVersions )
```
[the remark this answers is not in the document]
tsachiherman: it's a global variable.

network/wsNetwork.go
```diff
@@ -826,6 +826,28 @@ func (wn *WebsocketNetwork) checkIncomingConnectionLimits(response http.Response
 	return http.StatusOK
 }
 
+// checkProtocolVersionMatch test ProtocolAcceptVersionHeader and ProtocolVersionHeader headers from the request/response and see if it can find a match.
+func (wn *WebsocketNetwork) checkProtocolVersionMatch(otherHeaders http.Header) (matchingVersion string, otherVersion string) {
+	otherAcceptedVersions := otherHeaders[textproto.CanonicalMIMEHeaderKey(ProtocolAcceptVersionHeader)]
+	for _, otherAcceptedVersion := range otherAcceptedVersions {
+		// do we have a matching version ?
+		for _, supportedProtocolVersion := range SupportedProtocolVersions {
```
pzbitskiy: nit: use map? (and have O(n) vs O(n^2). Understand that for small values there no difference but the code also would look better

network/wsNetwork.go
```diff
@@ -1553,6 +1583,11 @@ func (wn *WebsocketNetwork) tryConnect(addr, gossipAddr string) {
 	defer wn.wg.Done()
 	requestHeader := make(http.Header)
 	wn.setHeaders(requestHeader)
+	for _, supportedProtocolVersion := range SupportedProtocolVersions {
+		requestHeader.Add(ProtocolAcceptVersionHeader, supportedProtocolVersion)
+	}
+	// for backward compatability, include the ProtocolVersion header as well.
+	requestHeader.Set(ProtocolVersionHeader, ProtocolVersion)
```
pzbitskiy: Doesn't  ProtocolAcceptVersionHeader implies own version?
What kind of backward compatibility?

network/wsNetwork.go
tsachiherman: The order is meaninfull, which is why I used an array.

network/wsNetwork.go
tsachiherman: Existing relays expect to get the ProtocolVersionHeader.

tsachiherman: Existing relays expect to get the ProtocolVersionHeader.

network/wsNetwork.go
zeldovich: Why does this function return `otherVersion`?  It doesn't seem to be meaningfully used by its caller.

tsachiherman: The usage of `otherVersion` is for the logging/response text. This is the only indication we have that the mismatch was with the `ProtocolVersionHeader`. As an alternative, we can try to re-extract the value from the headers, but I'm not sure if that would improve the code readability.

PR #800, opened 2020-02-10
Move `dsign` functionality to goal

justicz:
This PR moves some of the `dsign` functionality into a new goal command, `goal clerk tealsign`.

Currently this command only supports signing using keyfiles as generated by `algokey`. The next thing to do is add support for signing with keys from `kmd`. I left that out of this PR, because I wanted to add a "sign data for logic" endpoint to `kmd` rather than export private keys just to make this work. That will come next.

### DIFF
--- a/cmd/goal/tealsign.go
+++ b/cmd/goal/tealsign.go
@@ -59,8 +59,14 @@ func init() {
 var tealsignCmd = &cobra.Command{
 	Use:   "tealsign",
 	Short: "Sign data to be verified in a TEAL program",
-	Long:  `Sign data to be verified in a TEAL program`,
-	Args:  validateNoPosArgsFn,
+	Long: `Sign data to be verified in a TEAL program.
+
+Data verified by the ed25519verify TEAL opcode must be domain separated. As part of this process, the signed payload includes the hash of the program logic. This hash must be specified. To do this, provide a transaction whose logic sig contains the program via --lsig-txn, or provide a contract address directly with --contract-addr. These options are mutually exclusive.
+
+Next, you must specify the data to be signed. When using --lsig-txn, you can use the --sign-txid flag to sign that transaction's txid. Alternatively, arbitrary data can be signed with the --data-file, --data-b64, or --data-b32 options. These options are mutually exclusive.
+
+The base64 encoding of the signature will always be printed to stdout. Optionally, when using --lsig-txn, you may specify that the signature be used as a TEAL argument for that transaction. Specify the argument index with the --set-lsig-arg-idx flag. The --lsig-txn file will be updated in place, and any existing argument at that index will be overwritten.`,
+	Args: validateNoPosArgsFn,
 	Run: func(cmd *cobra.Command, args []string) {
 		/*
 		 * First, fetch the key for signing
--- a/cmd/goal/tealsign.go
+++ b/cmd/goal/tealsign.go
@@ -45,15 +45,15 @@ var (
 func init() {
 	clerkCmd.AddCommand(tealsignCmd)
 
-	tealsignCmd.Flags().StringVarP(&keyFilename, "with-keyfile", "f", "", "algokey private key file to sign with")
-	tealsignCmd.Flags().StringVarP(&signerAcct, "with-account", "t", "", "address of account to sign with")
-	tealsignCmd.Flags().StringVarP(&lsigTxnFilename, "lsig-txn", "l", "", "transaction with logicsig to sign data for")
-	tealsignCmd.Flags().StringVarP(&contractAddr, "contract-addr", "c", "", "contract address to sign data for. not necessary if --lsig-txn is provided")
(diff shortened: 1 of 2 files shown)


cmd/goal/tealsign.go
```diff
@@ -0,0 +1,223 @@
...
+	setLsigArg      int
+)
+
+func init() {
+	clerkCmd.AddCommand(tealsignCmd)
+
+	tealsignCmd.Flags().StringVar(&keyFilename, "keyfile", "", "algokey private key file to sign with")
+	tealsignCmd.Flags().StringVar(&signerAcct, "account", "", "address of account to sign with")
+	tealsignCmd.Flags().StringVar(&lsigTxnFilename, "lsig-txn", "", "transaction with logicsig to sign data for")
+	tealsignCmd.Flags().StringVar(&contractAddr, "contract-addr", "", "contract address to sign data for. not necessary if --lsig-txn is provided")
```
derbear: Perhaps complain if both `lsig-txn` and `contract-addr` are provided?

cmd/goal/tealsign.go
```diff
@@ -0,0 +1,223 @@
...
+	tealsignCmd.Flags().BoolVar(&signTxID, "sign-txid", false, "use the txid of --lsig-txn as the data to sign")
+	tealsignCmd.Flags().StringVar(&dataFile, "data-file", "", "data file to sign")
+	tealsignCmd.Flags().StringVar(&datab64, "data-b64", "", "base64 data to sign")
+	tealsignCmd.Flags().StringVar(&datab32, "data-b32", "", "base32 data to sign")
```
derbear: Perhaps complain if more than one `data-` argument is provided?

cmd/goal/tealsign.go
```diff
@@ -0,0 +1,223 @@
...
+			dataArgs++
+		}
+
+		// Ensure there is one unambiguous source of data
+		if dataArgs != 1 {
+			reportErrorf(tealsignDataReq)
+		}
+
+		/*
+		 * Sign the payload and print signature to stdout
```
derbear: Nit: remove "and print signature to stdout"

cmd/goal/tealsign.go
justicz: I think we do -- see `tealsignMutLsigArgs` below

cmd/goal/tealsign.go
justicz: I think we do -- see `tealsignDataReq` below

cmd/goal/tealsign.go
```diff
@@ -0,0 +1,223 @@
...
+
+	tealsignCmd.Flags().StringVar(&keyFilename, "keyfile", "", "algokey private key file to sign with")
+	tealsignCmd.Flags().StringVar(&signerAcct, "account", "", "address of account to sign with")
+	tealsignCmd.Flags().StringVar(&lsigTxnFilename, "lsig-txn", "", "transaction with logicsig to sign data for")
+	tealsignCmd.Flags().StringVar(&contractAddr, "contract-addr", "", "contract address to sign data for. not necessary if --lsig-txn is provided")
+	tealsignCmd.Flags().BoolVar(&signTxID, "sign-txid", false, "use the txid of --lsig-txn as the data to sign")
+	tealsignCmd.Flags().StringVar(&dataFile, "data-file", "", "data file to sign")
+	tealsignCmd.Flags().StringVar(&datab64, "data-b64", "", "base64 data to sign")
+	tealsignCmd.Flags().StringVar(&datab32, "data-b32", "", "base32 data to sign")
+	tealsignCmd.Flags().IntVar(&setLsigArg, "set-lsig-arg", -1, "if --lsig-txn is also specified, set this lsig arg to the raw signature bytes")
```
derbear: Add a note saying that if this is provided, then in `lsigTxnFilename` will be overwritten such that the lsig arg at the position here will be overwritten with the signature.

derbear: It might be a good idea to rename this to `set-lsig-arg-index`

Might also want to note that `nil` lsig args will be appended until the index is valid.

cmd/goal/tealsign.go
derbear: Ah, I misread the program.

cmd/goal/tealsign.go
derbear: Ah, I misread the program.

cmd/goal/tealsign.go
justicz: done

cmd/goal/tealsign.go
justicz: done, thanks!

cmd/goal/tealsign.go
```diff
@@ -46,14 +46,14 @@ func init() {
 	clerkCmd.AddCommand(tealsignCmd)
```
derbear: It might be useful to say in the help text somewhere which arguments are mutually exclusive.

justicz: added better help text

PR #818, opened 2020-02-12
Some build fixes

btoll:
Most importantly, move the `fullversion.dat` file to the $HOME directory
and use it for the name of the upload directory on s3.  It should have
been doing this before, but it was copying it to the wrong location on
the ec2 instance.

### DIFF
--- a/scripts/release/build/stage/upload/run.sh
+++ b/scripts/release/build/stage/upload/run.sh
@@ -4,13 +4,17 @@
 
 CHANNEL="$1"
 BUCKET_LOCATION="$2"
-FULLVERSION="$4"
 INSTANCE=$(cat scripts/release/common/ec2/tmp/instance)
 
-rm -rf pkg/* && mkdir -p pkg/"$FULLVERSION"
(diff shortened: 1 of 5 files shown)


scripts/release/build/stage/upload/run.sh
```diff
@@ -4,13 +4,17 @@
 
 CHANNEL="$1"
 BUCKET_LOCATION="$2"
-FULLVERSION="$4"
 INSTANCE=$(cat scripts/release/common/ec2/tmp/instance)
 
-rm -rf pkg/* && mkdir -p pkg/"$FULLVERSION"
+scp -i ReleaseBuildInstanceKey.pem -o StrictHostKeyChecking=no -r ubuntu@"$INSTANCE":~/fullversion.dat .
```
tsachiherman: nit: instead of copying it locally and deleting it afterward, just ssh and have it `cat ~/fullversion.dat`

btoll: Good call, I'll change it.

PR #819, opened 2020-02-12
Added utils for testing release packages

btoll:
check_sig: Verify gpg signatures of build artifacts.

test_package: Verifies the packages were built from the correct branch
with the correct hash and verifies the test version release number.

### DIFF
--- a/scripts/release/util/check_sig.sh
+++ b/scripts/release/util/check_sig.sh
@@ -1,0 +1,42 @@
+#!/usr/bin/env bash
+# shellcheck disable=2045
+
+if [ $# -eq 0 ]
+then
+    echo "Usage: $0 <DIRECTORY>"
+    exit 1
+fi
+
+GREEN_FG=$(tput setaf 2 2>/dev/null)
+RED_FG=$(tput setaf 1 2>/dev/null)
+END_FG_COLOR=$(tput sgr0 2>/dev/null)
+RETVAL=0
+
+pushd "$1" > /dev/null
+
+for file in $(ls ./*.{gz,deb,rpm})
+do
+    key_id=dev@algorand.com
+
+    # Check the filename extension.
+    if [ "${file##*.}" == "rpm" ]
+    then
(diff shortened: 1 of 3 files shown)


scripts/release/util/check_sig.sh
```diff
@@ -0,0 +1,42 @@
+#!/usr/bin/env bash
+# shellcheck disable=2045
+
+if [ $# -eq 0 ]
```
tsachiherman: you want to check that you have a single argument, rather then checking that the user forgot to specify a directory.

scripts/release/util/smoke_test.sh
```diff
@@ -0,0 +1,61 @@
...
+            shift
+            RELEASE="$1"
+            ;;
+        *)
+            echo "Unknown option" "$1"
+            exit 1
+            ;;
+    esac
+    shift
+done
```
tsachiherman: could you please add verification that all the above parameters have been provided ( either via export or explicitly provided as command line arguments ? )

scripts/release/util/test_package.sh
```diff
@@ -0,0 +1,122 @@
...
+            shift
+            RELEASE="$1"
+            ;;
+        *)
+            echo "$RED_FG[$0]$END_FG_COLOR Unknown option $1"
+            exit 1
+            ;;
+    esac
+    shift
+done
```
tsachiherman: same as before; make sure that these arguments are set before proceeding further.

scripts/release/util/check_sig.sh
btoll: Will change.

scripts/release/util/smoke_test.sh
btoll: Will do.

scripts/release/util/test_package.sh
btoll: Will do.

PR #821, opened 2020-02-12
Deleting out-of-date wallet folder in go-algorand.

algonautshant:
## Summary

Delete go-algorand/wallet.
It is out-of-date and not used. 
It no longer works against the current algod REST API.


rotemh:
Feel free to kill them. We used them for auction testing back then but a lot have changed.

PR #824, opened 2020-02-13
Remove empty serialized crash recovery fields

tsachiherman:
## Summary

Before the consensus version makes a vote, it serialize it's internal state so that the voting would remain consistent if it get restarted.

When the `networkAction` structure is being persisted, all the three fields are being stored : `UnauthenticatedVote`, `UnauthenticatedBundle` and `CompoundMessage`. By the nature of this action, only one of them would get populated at a time. That results in storing more data then we really need to.

A simple workaround for that is to let the encoder know that it doesn't need to persist empty fields.


derbear:
Do we have an idea of the cost savings from this? It seems like this change reduces our disk footprint by an almost imperceptible amount (three msgpack'd keys per networkAction, which is probably much, much smaller than a single voteTracker) in exchange for more CPU time spent checking zero values (which also seems imperceptible).

tsachiherman:
I'm not particular about this one, so I'm going to close it down. I think that we should optimize the storage, but dropping this off as I can't justify the time spent vs. potential gains.

PR #827, opened 2020-02-13
netgoal: create accounts in parallel 

tsachiherman:
## Summary

Creating large number of accounts is pretty painful process. While the existing tooling are performing the task, they do that in serial.

This change move the wallets generation to be done in parallel. Tested with s3, it dropped the time it took to create the accounts from 389 seconds to 250 seconds.


gen/generate.go
```diff
@@ -115,94 +120,148 @@ func generateGenesisFiles(outDir string, protoVersion protocol.ConsensusVersion,
...
+		// create a listener for the verbosedOutput
+		go func() {
+			for textOut := range verbosedOutput {
+				fmt.Printf("%s\n", textOut)
+			}
+		}()
+	}
+
+	creatingWalletsWaitGroup.Add(concurrentWalletGenerators)
+	for routinesCounter := 0; routinesCounter < concurrentWalletGenerators; routinesCounter++ {
```
pzbitskiy: nit: short loop variable, might be i/c/something short

gen/generate.go
```diff
@@ -115,94 +120,148 @@ func generateGenesisFiles(outDir string, protoVersion protocol.ConsensusVersion,
...
-		genesisAddrs[wallet.Name] = root.Address()
+	close(verbosedOutput)
 
-		rootDB.Close()
-		if wallet.Online == basics.Online {
-			partDB.Close()
-		}
+	// check to see if we had any errors.
+	select {
+	case err := <-errorsChannel:
```
pzbitskiy: read only one error by might have up to concurrentWalletGenerators. It might be possible that one failure cascades others and because of concurrency we read here only secondary error but not the original one.

gen/generate.go
```diff
@@ -115,94 +120,148 @@ func generateGenesisFiles(outDir string, protoVersion protocol.ConsensusVersion,
...
+				data.VoteKeyDilution = part.KeyDilution
+			}
 
-				partDB, err = db.MakeErasableAccessor(pfilename)
-				if err != nil {
-					err = fmt.Errorf("couldn't open participation DB accessor %s: %v", pfilename, err)
-					os.Remove(pfilename)
-					return
-				}
+			writeMu.Lock()
```
pzbitskiy: side thoughts: interesting if it is possible in go pre-allocate map and assign ranges to every goroutine so that it can write to maps without locking

tsachiherman: not with a map; but it's doable with a preallocated array.

gen/generate.go
tsachiherman: I don't think that there is an elegant solution here, other than reducing concurrency to 1 and trying again.

pzbitskiy: why not reading until it's over? like for + select + break on default.

tsachiherman: I won't feel bad to have that handled separately by constructing an error object which lists all the errors that were encountered during the process.

PR #829, opened 2020-02-14
Avoid block assembly if no participating account is going to make a proposal

tsachiherman:
## Summary

Refactor existing proposals generation so that we won't generate proposals if none of the participation accounts are members of the relevant committee. In particular, `asyncPseudonode::makeProposals` was modified to use the same thread pool as the async verifier to evaluate each of the participating account memberships. Once a single account is found to be a member of the committee, we assemble the block.

From performance perspective, the changes are as follows:
1. Instead of generating the VRF.Prove and signature in `asyncPseudonode::makeProposals` ( single thread), we're using the thread pool to calculate the VRF.Prove and VRF.Verify for each account that has participation key.
1. Since we already have the credential, we no longer need to verify the vote; hence, we no longer need to use the async verifier to verify the self-generated votes.
1. The agreement `selector` was slightly optimized by storing it's internal encoded selector bytes. This would become useful when we need to hash it twice across the Prove and Verify functionalities.

## Performance Metrics

Running this test on S2 showed no gain in terms of round time. The round time seems to be dominated by vote count rather then local-vote generation. However, when it comes to transactions per second, there are notable gains:

|test | master | this change |  
-- | -- | -- | --
payment | 829 | 935 | 11.34%
teal_light | 782 | 835 | 6.35%
teal_heavy | 466 | 477 | 2.31%
teal_normal | 588 | 576 | -2.08%
atomic_transfer_small | 812 | 842 | 3.56%
atomic_transfer_heavy | 794 | 843 | 5.81%
mixed_light | 856 | 886 | 3.39%


agreement/vote.go
```diff
@@ -182,6 +182,44 @@ func makeVote(rv rawVote, voting crypto.OneTimeSigner, selection *crypto.VRFSecr
 	return unauthenticatedVote{R: rv, Cred: cred, Sig: sig}, nil
 }
 
+// signVote sign a given raw vote and uses the provided credential.
+//
+// signVote returns an error it it fails.
+func signVote(rv rawVote, voting crypto.OneTimeSigner, l Ledger, cred committee.Credential) (vote, error) {
```
derbear: This seems to duplicate a lot of code with `makeVote`. It would be nice to avoid this.

tsachiherman: I agree. I think that makeVote isn't needed at all ( eventually). But this change doesn't touch the non-proposal vote generation, which is why it's still there. ( i.e. the only other consumer for that in a non-test code is the asyncPseudonode::makeVotes.

tsachiherman:
> This change seems to modify several things at once:
> 
> * Creating votes in parallel, which benefits machines with many accounts
> * Drop the sanity check on created votes, which benefits all machines but also makes it harder to maintain vote verification code
> * Cache hashes in credential computation, which is nice, but hashing is 2-3 orders of magnitude cheaper than the VRF computation/signature
> * Skip block assembly for nodes without an account selected for proposal, which in practice does not benefit the system (because these nodes are not on the critical path) but shows up nicely in our synthetic benchmarks.
> 
> Could we separate out these into several changes? I'm afraid of introducing extra complexity in the agreement code, which will complicate maintenance and increase chances of a system fork/stall, for performance benefits which are not well-understood.

Sure, I have no problem breaking this apart. However, most of the gains wouldn't be there unless we skip the block assembly. ( and honestly, there is no real "justification" for most of these changes unless we could reduce the frequency of the block assembly ).

Note that this change could be beneficial also for low-powered devices where block assembly is an issue to start with.

tsachiherman:
I'm closing this one as we've concluded that we shouldn't keep pushing forward on performance optimizations.

PR #834, opened 2020-02-17
Limit msgp tool warning message scope

tsachiherman:
## Solution

The newly introduces message pack utility generates large amount of information messages when being invoked. This change limits the verbosity of these messages to the data types in go-algorand project.


tsachiherman:
> Looks reasonable to me. You might want to tag a new version of msgp (create and push a new tag to the algorand-msgp repo) and then specify this new version in `go.mod` of this repo (`go-algorand`).

Great idea ! done.

PR #835, opened 2020-02-17
Brice/refactor make

bricerisingalgorand:
## Summary

This is a PR to refactor our make file to make it more useful for our cicd process. Part of this was to refactor the libsodium-fork build so that it creates folders for different os/platforms so that our project's working directory can be shared between several different docker containers and systems at the same time. I also implemented two new targets to separate the concerns of gathering dependencies and building the project. This way our docker build process can execute the gathering deps target and our ci process can execute building artifacts.

## Test Plan

I tested that this works for a multi-platform build (it now creates artifacts for darwin/amd64, linux/amd64 and linux/arm64v8). I also attempted to verify that previous workflows of the related targets still work. I would like to have a lot of reviewers on this ticket verifying they can still use the scripts they've used previously, since the changes to libsodium could potentially affect some existing workflows that I missed.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -280,7 +280,3 @@ ci-deps:
 ci-build: buildsrc gen
 	mkdir -p $(SRCPATH)/tmp/node_pkgs/$(OS_TYPE)/$(ARCH) && \
 	PKG_ROOT=$(SRCPATH)/tmp/node_pkgs/$(OS_TYPE)/$(ARCH) NO_BUILD=True VARIATIONS=literally_anything scripts/build_packages.sh $(OS_TYPE)/$(ARCH)
-
-.EXPORT_ALL_VARIABLES:
-CGO_CFLAGS  = -I$(SRCPATH)/crypto/libs/$(OS_TYPE)/$(ARCH)/include
-CGO_LDFLAGS = $(SRCPATH)/crypto/libs/$(OS_TYPE)/$(ARCH)/lib/libsodium.a
--- a/Makefile
+++ b/Makefile
@@ -274,12 +274,16 @@ install: build
 ### TARGETS FOR CICD PROCESS
 
 ci-deps:
-	scripts/configure_dev.sh && \
+	scripts/configure_dev-deps.sh && \
 	scripts/check_deps.sh
 
 ci-buildsrc: crypto/lib/libsodium.a node_exporter NONGO_BIN $(ALGOD_API_SWAGGER_INJECT) $(KMD_API_SWAGGER_INJECT)
-	OS_TYPE=$(OS_TYPE) ARCH=$(ARCH) go install $(GOTRIMPATH) $(GOTAGS) $(GOBUILDMODE) -ldflags="$(GOLDFLAGS)" ./...
+	go install $(GOTRIMPATH) $(GOTAGS) $(GOBUILDMODE) -ldflags="$(GOLDFLAGS)" ./...
 
 ci-build: ci-buildsrc gen
 	mkdir -p $(SRCPATH)/tmp/$(OS_TYPE)/$(ARCH)/dev_pkg && \
 	PKG_ROOT=$(SRCPATH)/tmp/$(OS_TYPE)/$(ARCH)/dev_pkg scripts/dev_install.sh -p $(GOPATH1)/bin
+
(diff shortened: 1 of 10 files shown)


bricerisingalgorand:
If anyone wants to try using the cicd cli to do multi-platform builds, let me know and I can help you get it set up!

Makefile
```diff
@@ -90,9 +91,11 @@ ALWAYS:
 
 # build our fork of libsodium, placing artifacts into crypto/lib/ and crypto/include/
 crypto/lib/libsodium.a:
-	cd crypto/libsodium-fork && \
-		./autogen.sh && \
-		./configure --disable-shared --prefix="$(SRCPATH)/crypto/" && \
+	mkdir -p crypto/copies/$(OS_TYPE)/$(ARCH)
+	cp -R crypto/libsodium-fork crypto/copies/$(OS_TYPE)/$(ARCH)/libsodium-fork
```
zeldovich: A more common pattern is to check in the results of running `autogen.sh` into git, and then do builds by invoking `configure` from a different directory.  The `configure` script takes care of setting up symlinks, etc.

In particular, I'm thinking of something like:

```
crypto/libs/$(OS_TYPE)/$(ARCH)/Makefile:
  mkdir -p $(@D)
  cd $(@D) && ../../../libsodium-fork/configure --disable-shared --prefix="$(SRCPATH)/$(@D)"

crypto/libs/$(OS_TYPE)/$(ARCH)/src/libsodium/.libs/libsodium.a: crypto/libs/$(OS_TYPE)/$(ARCH)/Makefile
  cd crypto/libs/$(OS_TYPE)/$(ARCH) && make

crypto/libs/$(OS_TYPE)/$(ARCH)/libsodium.a: crypto/libs/$(OS_TYPE)/$(ARCH)/src/libsodium/.libs/libsodium.a
  cd $(@D) && make install
```

bricerisingalgorand: I've never used automake before, does this not produce files that are specific to the platform it's running on?

zeldovich: The results of autogen.sh should be platform-agnostic

pzbitskiy: +1 to Nikolai, it is common to generate platform-dependant build scripts to a dir and run the build from there without copying sources.

Makefile
```diff
@@ -265,3 +270,13 @@ install: build
...
+
+### TARGETS FOR CICD PROCESS
+
+ci-deps:
+	scripts/configure_dev-deps.sh && \
+	scripts/check_deps.sh
+
+ci-build: buildsrc gen
+	mkdir -p $(SRCPATH)/tmp/node_pkgs/$(OS_TYPE)/$(ARCH) && \
+	PKG_ROOT=$(SRCPATH)/tmp/node_pkgs/$(OS_TYPE)/$(ARCH) NO_BUILD=True VARIATIONS=literally_anything scripts/build_packages.sh $(OS_TYPE)/$(ARCH)
```
tsachiherman: I'm abit confused about the `VARIATIONS=literally_anything`.. could you provide some more context ?

Makefile
bricerisingalgorand: I've been thinking and this effort seems out of scope for the story I'm working on since I would need to implement a new feature that doesn't exist in our libsodium-fork. Also I think moving diverging from upstream libsodium's build automation could be a little dangerous for us over time.

zeldovich: What feature are you referring to?

bricerisingalgorand: The feature of supporting multi-platform concurrent builds of libsodium using a shared working directory.

bricerisingalgorand: I agree though that's something we would want if we don't have to diverge too much from upstream to implement it, I'm just not sure if that's something within the scope of this work effort

zeldovich: It's built into libsodium as-is.

bricerisingalgorand: I can't seem to figure out how to do that, since automake and configure seem to be making a lot of files for the host operating system in the current working directory. Would you be alright with me making an issue to target this feature in a future sprint?

zeldovich: Fine by me.

automake and configure are very different in these respects.  automake is platform-agnostic.  configure is what populates the build directory with platform-specific files.

PR #836, opened 2020-02-18
More release build fixes

btoll:
Most importantly, this improves the shared `scripts/compute_branch_channel.sh` script to be called from any directory in the project, no matter how deeply it's nested.

Also, tighten error reporting and increasing confidence by setting more shell scripts to return early on any failures `set -e`.


scripts/release/test/stage/setup/run.sh
```diff
@@ -1,26 +1,27 @@
...
-BUILD_ENV=$(ssh -i ReleaseBuildInstanceKey.pem -o -A ubuntu@"$INSTANCE" cat build_env)
+
+aws s3 cp s3://algorand-devops-misc/tools/gnupg2.2.9_centos7_amd64.tar.bz2 .
+scp -i ReleaseBuildInstanceKey.pem -o StrictHostKeyChecking=no -r gnupg2.2.9_centos7_amd64.tar.bz2 ubuntu@"$INSTANCE":
+
+scp -i ReleaseBuildInstanceKey.pem -o StrictHostKeyChecking=no -r scripts/release/common/setup.sh ubuntu@"$INSTANCE":
+ssh -i ReleaseBuildInstanceKey.pem -A ubuntu@"$INSTANCE" bash setup.sh "$BRANCH"
+
+BUILD_ENV=$(ssh -i ReleaseBuildInstanceKey.pem -A ubuntu@"$INSTANCE" cat build_env)
 CHANNEL=$(sed -n 's/.*CHANNEL=\(.*\)/\1/p' <<< "$BUILD_ENV")
```
btoll: The order of operations is essential here.  The `setup.sh` script *must* be copied to the instance and invoked before the remote call to `cat` the `build_env` file, which is created during `setup`.

scripts/release/test/deb/test_algorand.sh
```diff
@@ -8,7 +8,7 @@ export PATH=${GOPATH}/bin:/usr/local/go/bin:${PATH}
 apt-get update
 apt-get install -y gnupg2 curl software-properties-common python3
 
-apt install -y /root/subhome/node_pkg/*.deb
+apt-get install -y /root/subhome/node_pkg/*.deb
```
btoll: `apt-get` is better suited to scripting.

PR #840, opened 2020-02-20
Run build and test jobs in a docker container

btoll:


scripts/release/common/docker/setup.Dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM ubuntu:18.04
+RUN DEBIAN_FRONTEND=noninteractive apt-get update && apt-get install -y awscli jq ssh
+#RUN adduser --uid $(grep jenkins /etc/passwd | awk -F: '{ print $3 }') ubuntu
+RUN adduser --uid 111 ubuntu
```
btoll: This obviously isn't great, but it gets us moving forward.  I'll keep up with something better.

PR #841, opened 2020-02-20
Refactoring peer unicast implementation

algonautshant:
## Summary
In this change, an array of key/value pairs called Topics is added to facilitate the request/response communication between ws-peers. 

## Test Plan

Added tests for the request/response roundtrip, as well as tests for the marshall/unmarshall of topics, nonce generation.


network/wsNetwork_test.go
```diff
@@ -1384,3 +1384,89 @@ func TestCheckProtocolVersionMatch(t *testing.T) {
...
+		},
+	}
+	return OutgoingMessage{
+		Action: Respond,
+		Tag:    protocol.TopicMsgRespTag,
+		Topics: respTopics,
+	}
+}
+
+var TopicMsgReqTag Tag = "TQ"
```
tsachiherman: `TopicMsgReqTag` -> `topicMsgReqTag`

network/wsPeer.go
```diff
@@ -222,13 +237,48 @@ func (wp *wsPeer) Unicast(ctx context.Context, msg []byte, tag protocol.Tag) err
...
 }
 
+// Respond submits the request to the server, waits for a response
+func (wp *wsPeer) Respond(ctx context.Context, reqMsg IncomingMessage, respMsg OutgoingMessage) (e error) {
+
+	// Get the hash/key of the request message
+	requestHash := Hash(reqMsg.Data)
+
+	topics := respMsg.Topics
+	// Add the request key
```
tsachiherman: you mean `request hash`, right ?

network/wsPeer.go
```diff
@@ -222,13 +237,48 @@ func (wp *wsPeer) Unicast(ctx context.Context, msg []byte, tag protocol.Tag) err
...
+	requestHashData := make([]byte, binary.MaxVarintLen64)
+	binary.PutUvarint(requestHashData, requestHash)
+	topics = append(topics, Topic{key: "RequestHash", data: requestHashData})
+
+	// Add nonce as a topic
+	nonce := wp.getNonce()
+	topics = append(topics, Topic{key: "nonce", data: nonce})
+
```
tsachiherman: no need to add the nonce to the response.

network/wsPeer.go
```diff
@@ -525,3 +602,75 @@ func (wp *wsPeer) CheckSlowWritingPeer(now time.Time) bool {
 	timeSinceMessageCreated := now.Sub(time.Unix(0, ongoingMessageTime))
 	return timeSinceMessageCreated > maxMessageQueueDuration
 }
+
+// getNonce returns the byte representation of ever increasing uint64
+// The value is stored on wsPeer
+func (wp *wsPeer) getNonce() []byte {
+	atomic.AddUint64(&wp.nonce, 1)
```
tsachiherman: The function AddUint64 returns you a Uint64; use the returned value as the input to the binary.PutUvarint

network/wsPeer.go
```diff
@@ -525,3 +602,75 @@ func (wp *wsPeer) CheckSlowWritingPeer(now time.Time) bool {
...
+		return resp, ctx.Err()
+	}
+
+	// wait for the channel.
+	select {
+	case resp = <-wp.responseChannels[hash]:
+		return resp, nil
+	case <-wp.closing:
+		wp.net.log.Debugf("peer closing %s", wp.conn.RemoteAddr().String())
+		return
```
tsachiherman: you need to exit here with an error so that the caller won't attempt to use the `resp`. In go, when there is a `something, error` returned from a function, there is expectation that if the latter is nil then the former is valid.

network/wsPeer_test.go
```diff
@@ -37,3 +38,29 @@ func TestCheckSlowWritingPeer(t *testing.T) {
...
+				doneChannel <- true
+			}
+		}()
+	}
+	maxWait := time.After(2 * time.Second)
+	done := false
+	select {
+	case <-doneChannel:
+		done = true
+	case <-maxWait:
```
tsachiherman: I'm not sure if this test is the best way to verify the getNonce function.

network/wsPeer.go
algonautshant: Rewrote the test and it failed with this implementation of getNonce()
This suggestion fixed getNonce()

network/wsPeer_test.go
algonautshant: Rewrote the test to catch the bug in the getNonce() implementation.

network/wsPeer.go
```diff
@@ -222,13 +237,44 @@ func (wp *wsPeer) Unicast(ctx context.Context, msg []byte, tag protocol.Tag) err
 	return err
 }
 
+// Respond submits the request to the server, waits for a response
+func (wp *wsPeer) Respond(ctx context.Context, reqMsg IncomingMessage, respMsg OutgoingMessage) (e error) {
```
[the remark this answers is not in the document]
algonautshant: this task will be addressed in a future PR.

network/topics.go
```diff
@@ -0,0 +1,132 @@
...
+// Topics is an array of type Topic
+// The maximum number of topics allowed is 32
+// Each topic key can be 64 characters long and cannot be size 0
+type Topics []Topic
+
+// MarshallTopics serializes the topics into a byte array
+func (ts Topics) MarshallTopics() (b []byte) {
+
+	// Calculate the total buffer size required to store the topics
+	bufferSize := binary.MaxVarintLen32 // store topic array size
```
pzbitskiy: should not it be binary. MaxVarintLen64 since uint64(len(ts)) is written below?

network/wsNetwork.go
```diff
@@ -209,6 +209,7 @@ type OutgoingMessage struct {
 	Action  ForwardingPolicy
 	Tag     Tag
 	Payload []byte
+	Topics  Topics
```
pzbitskiy: why not to use Payload to store serialized topic?

network/wsPeer.go
```diff
@@ -222,13 +237,44 @@ func (wp *wsPeer) Unicast(ctx context.Context, msg []byte, tag protocol.Tag) err
 	return err
 }
 
+// Respond submits the request to the server, waits for a response
```
pzbitskiy: I can't see where does it wait for response, seems like outdated comment

network/topics.go
```diff
@@ -0,0 +1,132 @@
...
+
+import (
+	"encoding/binary"
+	"fmt"
+
+	"github.com/algorand/go-algorand/crypto"
+)
+
+// Topic is a key-value pair
+type Topic struct {
```
pzbitskiy: I'd rename Topics to something else to not confuse with MQ topics...
In fact it is used as a dict (** GetValue** the only method). Although I probably do not understand all the future use cases but it looks like map is more suitable then an array of struct unless we expect duplicate keys.

network/wsNetwork.go
tsachiherman: The reason for not having the topics paid over the payload is so that the encoding of the topics would be implemented within the network library.

network/topics.go
tsachiherman: I was thinking about map for Topics too; the issue is that we want to make sure that the encoding of the Topics would consistently generate the same hash every time. This isn't guaranteed with a map, where the order is arbitrary.

pzbitskiy: What do we expect in topics? if there are many keys then GetValue might be an issue. In this case (and if the keys are unique) an ordered map might work:
```
type OrderedMap struct {
  data map[string][]byte
  keys []string
}

func (om OrderedMap) Put(k string, v []byte) {
  om.keys = append(om.keys, k)
  om.data[k] = v
}

func (om OrderedMap) Serialize() (res []byte) {
  for _, k := range om.keys {
    res = append(res, []byte(k))
    res = append(res, om.data[k])
  }
}
```
But not sure if we buy much here. I'm fine with arrays for small amount of keys.

network/topics.go
algonautshant: bufferSize is estimating the buffer size, and on this line, it is adding the needed space for storing the topic array size. 
The maximum topic array size can be 32, and the estimate here is for storing a 32-bit integer. So, we are not under-estimating here. On the contrary, over-estimating.
Below, it is converted to uint64, because the function expects uint64, and not because it may need 64bits to store the value. 

The only situation were the code is not handling is when the number of topics passed is greater than or equal to 2^32. This will be an error when unmarshalling the topics.

network/wsPeer.go
algonautshant: Correcting it to:
// Respond sends the response of a request message

network/topics.go
algonautshant: The number of topics is limited to 32. So there will be no more than 32 keys. Doing anything fancier that an array will just overcomplicate things. 

As for duplication, nothing prevents the construction of multiple topics with the same key, and it will be marshalled/unmarshalled properly, but GetValue returns a single value, which will be the first topic with the key value.

tsachiherman: I expect to have around 3 keys.

network/topics.go
tsachiherman: If we ever get to a situation where we trying to send 2GB of data in a single "message", than we have a serious design issue here.. sure we can test it here, but the source for the issue wouldn't be here but rather in the application level.

tsachiherman: @algonautshant, while I don't think it would never be an issue, could you change this into  MaxVarintLen64, or change the `uint64(len(ts))` into `uint32(len(ts))` ?

pzbitskiy: My bad, I misread the code

PR #845, opened 2020-02-24
Improve periodic payment template

stechu:
## Summary

Improve the close case of Periodic Payment Template. Now the close case doesn't have to be "periodic"


winder:
This requires updating the SDKs as well cc: @rotemh

PR #853, opened 2020-02-26
Add MoI to network

tsachiherman:
## Solution
Adds the ability for a node in the gossip network to request the connected peer to send it only certain types  of messages ( i.e. tags ).

This feature is intended to allow us to create dedicated types of nodes on the network. Few types of such nodes includes node that are only sending messages, nodes that are only receiving messages of a certain type ( i.e. telemetry ), relay nodes that would exchange information about backbone structure layout, etc.

## Minor changes
* Add the `EnablePingHandler` configuration option ( default to true ), which controls whether the local node would process incoming ping requests.
* Rename `MsgSkipTag` to `MsgDigestSkipTag` to better reflect its role.


network/wsPeer.go
```diff
@@ -446,6 +513,17 @@ func (wp *wsPeer) writeLoopSend(msg sendMessage) (exit bool) {
 		// just drop it, don't break the connection
 		return false
 	}
+	if msg.msgTags != nil {
+		wp.sendMessageTag = msg.msgTags
+		return false
+	}
+	// the tags are always 2 char long; note that this is safe since it's only being used for messages that we have generated locally.
```
pzbitskiy: can you add a unit test asserting this?

network/wsPeer.go
```diff
@@ -384,42 +442,51 @@ func (wp *wsPeer) readLoop() {
...
+		select {
+		case wp.net.readBuffer <- msg:
+		case <-wp.closing:
+			wp.net.log.Debugf("peer closing %s", wp.conn.RemoteAddr().String())
+			return
 		}
 	}
 }
 
+func (wp *wsPeer) handleMessageOfInterest(msg IncomingMessage) bool {
```
pzbitskiy: a bit strange return code: false when processed and error, and true on close. Consider using `error` or naming that bool as *continuation* or something.

config/local_defaults.go
```diff
@@ -416,6 +417,9 @@ func migrate(cfg Local) (newCfg Local, err error) {
 		if newCfg.DNSSecurityFlags == 0 {
 			newCfg.DNSSecurityFlags = defaultLocalV6.DNSSecurityFlags
 		}
+		if newCfg.EnablePingHandler == defaultLocalV5.EnablePingHandler {
```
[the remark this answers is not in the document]
tsachiherman: Yes. As long as we haven't released it, it should be fine.

network/wsPeer.go
```diff
@@ -384,42 +442,51 @@ func (wp *wsPeer) readLoop() {
...
+	select {
+	case wp.sendBufferHighPrio <- sm:
+		return false
+	case <-wp.closing:
+		wp.net.log.Debugf("peer closing %s", wp.conn.RemoteAddr().String())
+		return true
+	default:
+	}
+
+	select {
```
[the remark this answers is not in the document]
tsachiherman: This is intentional. We want to try to push it to the high priority first. If we fail ( hence the "default" ), we want to try again with the low priority one as well.

network/wsPeer.go
```diff
@@ -446,6 +513,17 @@ func (wp *wsPeer) writeLoopSend(msg sendMessage) (exit bool) {
 		// just drop it, don't break the connection
 		return false
 	}
+	if msg.msgTags != nil {
```
algonautshant: is this the case that (if msg.msgTags != nil), then the writeLoopSend call is to register the peer's interests, and not actually send some message to the peer? 
A little comment about this can be helpful here if this is the case.

config/local_defaults.go
```diff
@@ -90,6 +90,7 @@ var defaultLocalV6 = Local{
 	TxSyncServeResponseSize:               1000000,
 	PeerConnectionsUpdateInterval:         3600,
 	DNSSecurityFlags:                      0x01, // New value with default 0x01
+	EnablePingHandler:                     true,
```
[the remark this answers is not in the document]
tsachiherman: not sure how that would help. The `migrate` function is usually the place to look for.

network/msgOfInterest.go
```diff
@@ -0,0 +1,65 @@
...
+	"github.com/algorand/go-algorand/protocol"
+)
+
+var errUnableUnmarshallMessage = errors.New("unmarshalMessageOfInterest: could not unmarshall message")
+var errInvalidMessageOfInterest = errors.New("unmarshalMessageOfInterest: message missing the tags key")
+var errInvalidMessageOfInterestLength = errors.New("unmarshalMessageOfInterest: message length is too long")
+
+const maxMessageOfInterestTags = 1024
+
+func unmarshallMessageOfInterest(data []byte) (map[protocol.Tag]bool, error) {
```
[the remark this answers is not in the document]
tsachiherman: sure.

network/wsPeer.go
tsachiherman: that is correct. I'll add a comment.

network/wsPeer.go
tsachiherman: done.

network/wsPeer.go
tsachiherman: done.

PR #854, opened 2020-02-26
Trap errors and remove ec2 instance

btoll:
There is some overlap with some changes between this PR and some outstanding ones.  If there are merge conflicts, I'll fix as needed.


scripts/release/common/setup.sh
```diff
@@ -7,17 +7,24 @@ if [ -z "${BUILDTIMESTAMP}" ]; then
     BUILDTIMESTAMP=$(cat "${HOME}/buildtimestamp")
     export BUILDTIMESTAMP
     echo run "${0}" with output to "${HOME}/buildlog_${BUILDTIMESTAMP}"
-    (bash "${0}" "${1}" 2>&1) | tee "${HOME}/buildlog_${BUILDTIMESTAMP}"
-    exit 0
+    bash "${0}" "${1}" 2>&1 | tee "${HOME}/buildlog_${BUILDTIMESTAMP}"
+    # http://tldp.org/LDP/abs/html/internalvariables.html#PIPESTATUSREF
+    exit "${PIPESTATUS[0]}"
```
btoll: This will get the exit code of the first command in the pipeline, which is what we want.

scripts/release/common/setup.sh
```diff
@@ -7,17 +7,24 @@ if [ -z "${BUILDTIMESTAMP}" ]; then
...
 fi
 
 echo
 date "+build_release begin SETUP stage %Y%m%d_%H%M%S"
 echo
 
+echo -e "deb http://us.archive.ubuntu.com/ubuntu/ bionic main universe multiverse\ndeb http://archive.ubuntu.com/ubuntu/ bionic main universe multiverse" | sudo tee /etc/apt/sources.list.d/ubuntu
+
```
btoll: This is a hack to ensure that `apt-get` will download all the packages will need.

scripts/release/common/setup.sh
```diff
@@ -7,17 +7,24 @@ if [ -z "${BUILDTIMESTAMP}" ]; then
...
 sudo apt-get update
 sudo apt-get upgrade -y
-sudo apt-get install -y build-essential automake autoconf awscli docker.io git gpg nfs-common python3 rpm sqlite3 python3-boto3 g++ libtool rng-tools
+
+sudo apt-get update
+sudo apt-get upgrade -y
+
```
btoll: Like the above comment, this is a needed hack so we don't get stuck.  I really think we should leave these in for now.

scripts/release/common/setup.sh
```diff
@@ -29,14 +36,24 @@ export BRANCH
...
 
 # Install latest Go
 cd "${HOME}"
-python3 "${HOME}/go/src/github.com/algorand/go-algorand/scripts/get_latest_go.py" --version-prefix=1.12
+
+if ! python3 "${HOME}/go/src/github.com/algorand/go-algorand/scripts/get_latest_go.py" --version-prefix=1.12
+then
+    echo Golang could not be installed!
+    exit 1
+fi
```
btoll: I'm having issues installing go1.12...did something change?

scripts/release/common/setup.sh
tsachiherman: nice, but that's not what you want. You want to use `set -o pipefail` which would set the error code you get from the pipe to the first failuire.

scripts/release/common/setup.sh
tsachiherman: I'm quite skeptical that the upgrade -y is needed. ( i.e. if you need specific upgrades for specific components, you should explicitly ask for that.

scripts/release/common/setup.sh
tsachiherman: not that I'm aware of.

scripts/release/common/setup.sh
btoll: Ok, I'll try removing it.

scripts/release/common/setup.sh
btoll: Ok, will change and test.

btoll: Oh, I remember now why I used `pipestatus` instead.  I agree that in most cases setting `pipefail` is the better choice, but we do need a return value here or the script will execute one more time than we need.

``` 
if [ -z "${BUILDTIMESTAMP}" ]; then
    date "+%Y%m%d_%H%M%S" > "${HOME}/buildtimestamp"
    BUILDTIMESTAMP=$(cat "${HOME}/buildtimestamp")
    export BUILDTIMESTAMP
    echo run "${0}" with output to "${HOME}/buildlog_${BUILDTIMESTAMP}"
    bash "${0}" "${1}" 2>&1 | tee "${HOME}/buildlog_${BUILDTIMESTAMP}"
    # http://tldp.org/LDP/abs/html/internalvariables.html#PIPESTATUSREF
    exit "${PIPESTATUS[0]}"
fi
``` 
1. When the script first runs, `BUILDTIMESTAMP` isn't set so we set it and then start capturing the logs.
2. We call the script again: `bash "${0}" "${1}" 2>&1`
3. Since `BUILDTIMESTAMP` is set skip this portion and parse the file.
4. Then, we need the return value and exit the script.

tsachiherman: yes, setting it to `pipefail` would set the error flag once the `bash "${0}" "${1}" 2>&1 | tee "${HOME}/buildlog_${BUILDTIMESTAMP}"` command is complete. Then, given that you already have a `set -ex`, it will exit right away.

tsachiherman: note that pipefail would not propagate to the child bash invocation since SHELLOPTS is not exported.

btoll: Yes, but the happy path is not to have an error.  So, if we don't explicitly return here, the script will be parsed again, which is not what we want.

PR #855, opened 2020-02-26
Explicitly curl go.1.12.9 and archive `get_latest_go.py`

btoll:
The golang download page was changed and our pinned version of golang is
no longer referenced on it.  This was breaking our build.  Instead, for
now we'll explicitly download the tarball via `curl`.

https://golang.org/dl/?mode=json

Let's use `curl` rather than `wget` since several distros (including centos) and OS like Darwin don't include it natively.  Also, we're already installing `curl`, so we can continue to use it.


algobolson:
The real problem is that we're not following Go releases fast enough. They released 1.14 and we should at least be on 1.13 by now. I think their policy is to only release security patches for current and previous minor version.

algobolson:
should do #639 instead

tsachiherman:
@algobolson , yes - once the build process is complete we will start implement reproducible builds. When doing so, we will be able to have the latest and greatest GO.

tsachiherman:
You'll need to resolve the conflicts.

scripts/release/build/rpm/build.sh
```diff
@@ -11,13 +11,21 @@ export HOME=/root
...
+if ! git clone --single-branch --branch "${BRANCH}" https://github.com/algorand/go-algorand go-algorand
+then
+    echo There has been a problem cloning the "$BRANCH" branch.
+    exit 1
+fi
+
+# Get golang 1.12.9 and build its own copy of go-algorand.
 cd "${HOME}"
-python3 "${HOME}/go/src/github.com/algorand/go-algorand/scripts/get_latest_go.py" --version-prefix=1.12
+if ! curl -O https://dl.google.com/go/go1.12.9.linux-amd64.tar.gz
```
pzbitskiy: How about setting 1.12.9 from a var? Brian already has somewhere a pending PR for go 1.13

tsachiherman: Let's hold off with that until after Brice finish merging his changes into the build process.
Once he'll be done with that, we should have a single location to update.

PR #857, opened 2020-02-26
Better error case handling in database utils

tsachiherman:
## Solution

Provide better handling for incorrect usage of the `Accessor` object.
In particular -  
1. The `func (db Accessor) Close()` function would now become a `func (db *Accessor) Close()`, so that the internal `Handle` variable would be cleared on the operating object.
1. In case of an initialization statement error, set the Handle to nil after closing the database object.

### DIFF
--- a/agreement/service_test.go
+++ b/agreement/service_test.go
@@ -797,7 +797,7 @@ func setupAgreementWithValidator(t *testing.T, numNodes int, traceLevel traceLev
 
 	cleanupFn := func() {
 		for _, accessor := range dbAccessors {
-			defer accessor.Close()
+			accessor.Close()
 		}
 
 		if r := recover(); r != nil {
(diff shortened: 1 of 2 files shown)


agreement/service_test.go
```diff
@@ -797,7 +797,7 @@ func setupAgreementWithValidator(t *testing.T, numNodes int, traceLevel traceLev
 
 	cleanupFn := func() {
 		for _, accessor := range dbAccessors {
-			defer accessor.Close()
+			accessor.Close()
```
tsachiherman: this one kept me puzzled for few seconds. The change I've made below was changing the close to work with a pointer receiver. That was causing this call to try and close the same accessor over and over since the variable `accessor` has a single memory location.

tsachiherman:
> Can you also change to pointer Accessor.Atomic?

Yes, that would be a good thing to do. I'll do that in a subsequent PR.

PR #870, opened 2020-02-27
Refactor are combine the phonebook implementations 

algonautshant:
## Summary

This PR consolidates all phonebook implementations into a single multi-network phonebook.
 

## Test Plan

No new functionality is added.
The existing tests are updated to reflect the new phonebook implementation.

### DIFF
--- a/network/phonebook.go
+++ b/network/phonebook.go
@@ -28,8 +28,6 @@ import (
 // of how many addresses the phonebook actually has. ( with the retry-after logic applied )
 const getAllAddresses = math.MaxInt32
 
-const defaultList = "default"
-
 // Phonebook stores or looks up addresses of nodes we might contact
 type Phonebook interface {
 	// GetAddresses(N) returns up to N addresses, but may return fewer
@@ -50,9 +48,9 @@ type Phonebook interface {
 	// Returns true of the addr was in the phonebook
 	UpdateConnectionTime(addr string, provisionalTime time.Time) bool
 
-	// ReplacePeerList merges a set of addresses with that passed in.
-	// new entries in they are being added
-	// existing items that aren't included in they are being removed
+	// ReplacePeerList merges a set of addresses with that passed in for networkName
+	// new entries in dnsAddresses are being added
+	// existing items that aren't included in dnsAddresses are being removed
 	// matching entries don't change
 	ReplacePeerList(dnsAddresses []string, networkName string)
 
@@ -76,7 +74,7 @@ func makePhonebookData(networkName string) phonebookData {
 }
 
 // phonebookImpl holds the server connection configuration values
-// and the list of request times withing the time window for each
+// and the list of request times within the time window for each
 // address.
 type phonebookImpl struct {
 	connectionsRateLimitingCount  uint
@@ -164,7 +162,6 @@ func (e *phonebookImpl) ReplacePeerList(dnsAddresses []string, networkName strin
 	}
 }
 
-// UpdateRetryAfter updates ...
 func (e *phonebookImpl) UpdateRetryAfter(addr string, retryAfter time.Time) {
 	e.lock.Lock()
 	defer e.lock.Unlock()
@@ -173,7 +170,7 @@ func (e *phonebookImpl) UpdateRetryAfter(addr string, retryAfter time.Time) {
 
 	_, found := e.data[addr]
 	if !found {
-		entry = makePhonebookData(defaultList)
+		entry = makePhonebookData("default")
 	} else {
 		entry = e.data[addr]
 	}
--- a/network/phonebook.go
+++ b/network/phonebook.go
@@ -111,6 +111,17 @@ func (e *PhonebookImpl) deletePhonebookEntry(entryName, networkName string) {
 	}
 }
 
+func (e *phonebookEntries) deletePhonebookEntry(entryName, phonebookName string) {
+	e.lock.Lock()
+	defer p.lock.Unlock()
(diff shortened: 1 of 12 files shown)


network/phonebook.go
```diff
@@ -104,48 +132,62 @@ func (e *phonebookEntries) filterRetryTime(t time.Time) []string {
...
-			e.data[addr] = phonebookData{}
+			e.data[addr] = makePhonebookData(networkName)
 		}
 	}
 
-	// remove items that were missing in they
+	// remove items that were missing in dnsAddresses
 	for k := range removeItems {
-		delete(e.data, k)
+		e.deletePhonebookEntry(k, networkName)
```
tsachiherman: Instead of this, shouldn't we avoid placing entries in `removeItems` that shouldn't be deleted ?
i.e. entries in e.data that aren't in networkNames[] shouldn't be added to start with.

network/phonebook.go
```diff
@@ -104,48 +132,62 @@ func (e *phonebookEntries) filterRetryTime(t time.Time) []string {
...
+func (e *phonebookImpl) UpdateRetryAfter(addr string, retryAfter time.Time) {
+	e.lock.Lock()
+	defer e.lock.Unlock()
+
+	var entry phonebookData
+
 	_, found := e.data[addr]
 	if !found {
-		e.data[addr] = phonebookData{retryAfter: retryAfter, recentConnectionTimes: make([]time.Time, 0)}
+		entry = makePhonebookData("default")
```
tsachiherman: UpdateRetryAfter does not need to create a new entry. You can just return.

tsachiherman: UpdateRetryAfter does not need to create a new entry. You can just return.

network/phonebook.go
```diff
@@ -47,50 +47,78 @@ type Phonebook interface {
...
 		connectionsRateLimitingWindow: connectionsRateLimitingWindow,
-		data:                          make(map[string]phonebookData, 0),
+		data:                          make(map[string]phonebookEntryData, 0),
+	}
+}
+
+func (e *phonebookImpl) deletePhonebookEntry(entryName, networkName string) {
+	pbEntry := e.data[entryName]
+	delete(pbEntry.networkNames, networkName)
+	if 0 == len(pbEntry.networkNames) {
```
pzbitskiy: 1. how pbEntry.networkNames can be non zero after deletion?
2. this is looks like the very first usage of "0 ==" order. I'd prefer to keep it consistent with all the rest code

network/phonebook.go
```diff
@@ -47,50 +47,78 @@ type Phonebook interface {
...
+	}
+	pbData.networkNames[networkName] = true
+	return pbData
+}
+
+// phonebookImpl holds the server connection configuration values
+// and the list of request times within the time window for each
 // address.
-type phonebookEntries struct {
+type phonebookImpl struct {
```
pzbitskiy: how about just a "phonebook"? lowercase and no Impl suffix?

network/phonebook.go
```diff
@@ -47,50 +47,78 @@ type Phonebook interface {
...
+	// new entries in dnsAddresses are being added
+	// existing items that aren't included in dnsAddresses are being removed
+	// matching entries don't change
+	ReplacePeerList(dnsAddresses []string, networkName string)
+
+	// ExtendPeerList adds unique addresses to this set of addresses
+	ExtendPeerList(more []string, networkName string)
 }
 
-type phonebookData struct {
```
pzbitskiy: could you please add a comment explaining the phonebook internals? How about renaming **phonebookEntryData** to addressData since it holds info about addresses - timing and inclusion to networks.

network/phonebook.go
```diff
@@ -104,48 +132,62 @@ func (e *phonebookEntries) filterRetryTime(t time.Time) []string {
 // new entries in they are being added
 // existing items that aren't included in they are being removed
 // matching entries don't change
-func (e *phonebookEntries) ReplacePeerList(they []string) {
+func (e *phonebookImpl) ReplacePeerList(dnsAddresses []string, networkName string) {
```
pzbitskiy: I like **they** more than **dnsAddresses** - the latter assumes addresses come from DNS/SRV but they can also come from a config file

network/phonebook.go
algonautshant: Can be non-zero after deletion because we are only deleting networkName from pbEntry.networkNames, and  there could be other elements besides the given networkName there. 

0 == comes from C/C++. avoiding an l-value to the left of a logical operator helps the compiler catch the = instead of == errors. This is not be an issue in go. I'll reverse it.

network/phonebook.go
algonautshant: @tsachiherman ?

tsachiherman: I think that after eliminating the phonebook from any external source ( i.e. node ), you'll find that the only consumer for the phonebook is the wsNetwork. When we get to that point, it would make sense to delete the interface and just use the rename `phonebookImpl` -> `phonebook`. But we're not there yet.

network/phonebook.go
algonautshant: changing to addressesThey

network/phonebook.go
algonautshant: name changed to addressData. 
Added comment to addressData.

network/phonebook.go
algonautshant: This point will be resolved in a future PR.

PR #877, opened 2020-02-29
Shoehorn `test_package.sh` into the test phase

btoll:
This tests the output of `algod -v` for several different distros to make sure it's correct.


scripts/release/common/setup.sh
```diff
@@ -44,6 +44,9 @@ then
     exit 1
 fi
 
+cd go-algorand
+COMMIT_HASH=$(git rev-parse "${BRANCH}")
+
```
btoll: Capture the commit hash which we'll use later in the test utility to check against the `stdout` of `algod -v`.

scripts/release/test/util/smoke_test.sh
```diff
@@ -39,9 +36,9 @@ while [ "$1" != "" ]; do
     shift
 done
 
-if [ -z "$BRANCH" ] || [ -z "$HASH" ] || [ -z "$RELEASE" ]
+if [ -z "$BRANCH" ] || [ -z "$COMMIT_HASH" ] || [ -z "$FULLVERSION" ]
 then
-    echo "$YELLOW_FG[Usage]$END_FG_COLOR $0 -b BRANCH -c CHANNEL -h HASH -r RELEASE"
+    echo "[ERROR] $0 -b BRANCH -c CHANNEL -h COMMIT_HASH -r FULLVERSION"
```
tsachiherman: bug: The variables on this line need to be inside ${...} to be printed correctly.

scripts/release/test/util/test_package.sh
```diff
@@ -24,41 +23,11 @@ OS_LIST=(
...
-            ;;
-    esac
-    shift
-done
-
-if [ -z "$BRANCH" ] || [ -z "$HASH" ] || [ -z "$RELEASE" ]
+if [ -z "$BRANCH" ] || [ -z "$COMMIT_HASH" ] || [ -z "$FULLVERSION" ]
 then
-    echo "$YELLOW_FG[Usage]$END_FG_COLOR $0 -b BRANCH -c CHANNEL -h HASH -r RELEASE"
+    echo "[ERROR] $0 was not provided with BRANCH, CHANNEL, COMMIT_HASH or FULLVERSION!"
```
tsachiherman: Is the `CHANNEL` still need to be here ?

btoll:
Closed by accident.

scripts/release/test/util/test_package.sh
btoll: Yes, it's needed to check the result of `algod -v`:
```
if [[ "$STR" =~ .*"$FULLVERSION.$CHANNEL [$BRANCH] (commit #$SHORT_HASH)".* ]]
then
    echo -e "[$0] The result of \`algod -v\` is a correct match.\n$STR"
    exit 0
fi
```

scripts/release/test/util/smoke_test.sh
btoll: I don't want to print the value of the variables here, I want the error message to print something like:
```
[ERROR] ./foo.sh -b BRANCH -c CHANNEL -h COMMIT_HASH -r FULLVERSION
```

tsachiherman: How would that error message help you to diagnose that the wrong FULLVERSION was passed in ?

scripts/release/test/util/test_package.sh
tsachiherman: if that's the case, why you're not testing for it's existence on the `if` line ?

btoll: I see what you mean.  I had a default value set before that I since removed.  I'll add the `CHANNEL` to the `if` conditional.

scripts/release/test/util/smoke_test.sh
btoll: I was originally thinking that it would act more like a "usage" message, but you're right that it would be better displaying the values.  I'll change it.

PR #879, opened 2020-03-02
Brice/refactor cicd stages to use persistent fields

bricerisingalgorand:
## Summary

This updates our cicd.yaml to use persistent fields to generate version numbers consistently through our build stages. This also includes a stage that builds for armv6

## Test Plan

I built the binaries for each os/platform and I was able to execute algod on each one.


Makefile
```diff
@@ -28,6 +28,11 @@ endif
 GOTAGSLIST  += osusergo netgo static_build
 GOBUILDMODE := -buildmode pie
 endif
+ifeq ($(ARCH), arm)
```
tsachiherman: could you make this more specific, so that raspbian based builds would not have these variables set ?
i.e. I'd like to see if we could transition to your docker build without modifying the existing system.

bricerisingalgorand: Sure, adding that now 👍

bricerisingalgorand: I attempted to resolve this

PR #882, opened 2020-03-03
shellchecked `build_deb.sh`

btoll:
I tested this by running `./build_deb.sh amd64 foo-dir` and then installing and running `algod -v`.


scripts/build_deb.sh
```diff
@@ -92,34 +94,34 @@ else
...
 debian_files=("control" "postinst" "prerm" "postrm" "conffiles")
 for ctl in "${debian_files[@]}"; do
     # Copy first, to preserve permissions, then overwrite to fill in template.
-    cp -a installer/debian/${ctl} ${PKG_ROOT}/DEBIAN/${ctl}
-    cat installer/debian/${ctl} \
-      | sed -e s,@ARCH@,${ARCH}, \
-            -e s,@VER@,${VER}, \
-      > ${PKG_ROOT}/DEBIAN/${ctl}
+    cp -a installer/debian/"${ctl}" "${PKG_ROOT}/DEBIAN/${ctl}"
+    < installer/debian/"${ctl}" \
```
pzbitskiy: cat version was obvious, this one is tricky

algobolson: wow, never seen `< in cmd > out` usually I've seen `cmd < in > out`. huh. Today I Learned about a new weird bashism.

scripts/build_deb.sh
```diff
@@ -44,45 +46,45 @@ DEFAULT_RELEASE_NETWORK=$(./scripts/compute_branch_release_network.sh "${DEFAULT
 PKG_ROOT=$(mktemp -d)
 trap "rm -rf $PKG_ROOT" 0
 
-mkdir -p ${PKG_ROOT}/usr/bin
+mkdir -p "${PKG_ROOT}"/usr/bin
```
tsachiherman: The double quotes need to end after the bin, right ? i.e.
`mkdir -p "${PKG_ROOT}/usr/bin"`

scripts/build_deb.sh
tsachiherman: it looks very cumbersome, and I'm not a big fan of that.. but it's better cross-platforming between Mac and ubuntu.

btoll: It was flagged as a UUOC:
https://github.com/koalaman/shellcheck/wiki/SC2002
https://en.wikipedia.org/wiki/Cat_%28Unix%29#Unix_culture

If everyone would rather opt for readable, I can disable this rule at the top of the file.

scripts/build_deb.sh
btoll: It doesn't need to be.

btoll:
> could you make sure that when you use double-quotes you include the entire argument rather then subsets ?
> i.e. don't `cat "${VAR}"/bin` but do this `cat "${VAR}/bin"`

I got in the habit of (usually) only quoting the variable itself rather than the entire string/argument after I had introduced some bugs.  This was because special characters to bash then needed to be prefaced with a backslash.

If you feel strongly about this I can quote the entire argument, I can go either way.

tsachiherman:
It sounds to me like placing the quotes on the variable itself would miss the purpose of doing this to start with. i.e. if you want to do it correctly, do it all the way.

But I'm not picky if you decide to defer that to a separate change, as I don't see any blockers here.

btoll:
@tsachiherman I just did you as you suggested.

PR #883, opened 2020-03-03
Update list of DNSSEC-aware resolvers

pzbitskiy:
Researched some well-known public DNS resolvers and updated defaults.


tsachiherman:
it looks like this change broke some of the unit tests. Could you take a look ?

PR #885, opened 2020-03-03
sketch teal mode checks

algobolson:
draft sketch of teal mode checks

demo disabling ed25519verify, which might not actually be what we want.

names for "Application Approval" and "Application State" modes might have better names possible.


pzbitskiy:
Integrated to https://github.com/justicz/go-algorand/pull/3

PR #888, opened 2020-03-04
telemetryhook gets stuck during shutdown

algonautshant:
## Summary
When the elasticHook returns an error, asyncTelemetryHook will not change to ready state.
Consequently, the elements will get stuck in the channel and the buffer.
The waitGroup associated with these elements, will prevent the telemetry from quitting.
This PR stops waiting for these elements.

### DIFF
--- a/logging/telemetryhook.go
+++ b/logging/telemetryhook.go
@@ -146,6 +146,7 @@ func (hook *asyncTelemetryHook) Fire(entry *logrus.Entry) error {
 	select {
 	case <-hook.quit:
 		// telemetry quit
+		hook.wg.Done()
 	case hook.entries <- entry:
 	default:
 		hook.wg.Done()
--- a/logging/telemetryhook.go
+++ b/logging/telemetryhook.go
@@ -48,7 +48,25 @@ func createAsyncHookLevels(wrappedHook logrus.Hook, channelDepth uint, maxQueueD
 	}
 
 	go func() {
-		defer hook.wg.Done()
+		defer func() {
+			// flush the channel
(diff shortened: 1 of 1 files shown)


logging/telemetryhook.go
```diff
@@ -126,6 +144,8 @@ func (hook *asyncTelemetryHook) waitForEventAndReady() bool {
 func (hook *asyncTelemetryHook) Fire(entry *logrus.Entry) error {
 	hook.wg.Add(1)
 	select {
+	case <-hook.quit:
```
algobolson: this case should have a hook.wg.Done()

algonautshant: Yes. thanks for catching this.

logging/telemetryhook.go
```diff
@@ -48,7 +48,25 @@ func createAsyncHookLevels(wrappedHook logrus.Hook, channelDepth uint, maxQueueD
...
 
 	go func() {
-		defer hook.wg.Done()
+		defer func() {
+			// flush the channel
+			moreEntries := true
+			for moreEntries {
+				select {
+				case entry := <-hook.entries:
+					hook.appendEntry(entry)
```
winder: I don't think you need to call `appendEntry`, call `hook.wg.Done()` and skip the range below

algonautshant: At this point, there may or may not be elements in hook.pending. 
For every element in hook.entries and hook.pending, there should be a hook.wg.Done()
So, there needs to be two range loops to achieve this.

PR #889, opened 2020-03-04
Merge pending changed from beta 2.0.11 to stable 2.0.5

tsachiherman:
Merge the accumulated changed from beta onto stable branch.


auction/messages.go
```diff
@@ -184,12 +184,12 @@ type BidOutcomes struct {
 
 	// Outcomes is a list of bid outcomes, one for every placed bid
 	// in the auction.
-	Outcomes []BidderOutcome `codec:"outcomes"`
+	Outcomes []BidderOutcome `codec:"outcomes,allocbound=-"`
```
bricerisingalgorand: is this setting some sort of default for allocbound?

bricerisingalgorand:
I don't see anything that would make me scared to merge this if all of the tests pass and we don't see a regression in performance.

auction/messages.go
tsachiherman: I think that "-" means that there is no allocation limitation on arrays.

tsachiherman:
We should do some performance testing on the stable branch, but I don't expect any performance regression here. ( I secretly tested this along with some other changes before; but we should re-test that again, just to be sure )

PR #890, opened 2020-03-04
Increase timeouts for agreement tests

bricerisingalgorand:
## Summary

Our timeouts were very low and this caused tests to inconsistently fail. I doubled the two timeouts that were causing us problems.

## Test Plan

I ran the unit tests and they still worked


bricerisingalgorand:
Eric is fixing this in PRs related to his story

PR #901, opened 2020-03-09
Brice/write install script for mule

bricerisingalgorand:
## Summary

This script ensures that mule is installed on the machine it is run on. This will be necessary for our travis job and likely useful for any devs wanting to use the mule cli.

## Test Plan

I tested the following scenarios:
* No pip installation
* Pip installed with wrong version of python
* git clone from ssh url of go-algorand-ci repo
* git clone from https url of go-algorand-ci repo
* install of default mule version (0.0.0)
* version change to 0.0.0 when version 0.0.1 is currently installed on the system
* desired version already installed on system (both for default and set to 0.0.1)

### DIFF
--- a/scripts/install_mule.sh
+++ b/scripts/install_mule.sh
@@ -26,8 +26,6 @@ while getopts ":v:uh" opt; do
   esac
 done
 
-
-
 if CURRENT_MULE_VERSION=$(mule -v 2> /dev/null); then
     if [[ ${CURRENT_MULE_VERSION} == ${MULE_VERSION} ]]; then
         echo "Mule version $(mule -v) is already installed"
--- a/scripts/install_mule.sh
+++ b/scripts/install_mule.sh
@@ -1,0 +1,29 @@
+#!/usr/bin/env bash
+
+set -e
+
+if mule -v &> /dev/null; then
+    echo "Mule version $(mule -v) is already installed"
+    exit 0
+fi
+
+if VERSION=$(pip3 --version) &> /dev/null; then
+    PIP="pip3"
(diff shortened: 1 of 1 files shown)


scripts/install_mule.sh
```diff
@@ -0,0 +1,67 @@
+#!/usr/bin/env bash
+
+set -e
+
+MULE_VERSION='0.0.0'
+
+HELP="Usage: cmd [-v] [-u]
```
onetechnical:
```suggestion
HELP="Usage: $0 [-v] [-u]
```

onetechnical: Might as well use the program name.

winder:
Do we need this in go-algorand? It references go-algorand-ci which isn't a public repo.

scripts/install_mule.sh
bricerisingalgorand: Thanks!

bricerisingalgorand:
> Do we need this in go-algorand? It references go-algorand-ci which isn't a public repo.

Hmm it needs to be somewhere easily accessible by the travis job. I believe putting it anywhere but a public repo would require some sort of ci process to make sure it's available.

bricerisingalgorand:
This is no longer needed since the mulecli is in pypi now

btoll:
Could you put this in an archive dir in the mule area of go-algorand-ci?  It'd be a shame to lose this, and who knows if it would come in handy.

bricerisingalgorand:
oh that's a great idea for each script we end up sunsetting. I'll make that PR now

PR #906, opened 2020-03-09
Eric/expect test timeout adjustment

egieseke:
## Summary

Extend timeouts for goal tests to allow running integration tests within a container.

## Test Plan

Running expect tests as part of CI Integration pipeline.


egieseke:
The default timeout is 30 seconds.  There is no guarantee that the goal operations will complete within a single round.  The expect tests are designed to wait until the operations complete.  Running within a docker container on MAC slows requires additional time then running outside of the container, extending the timeouts was required for the test to run successfully.

tsachiherman:
@egieseke - having the timeout set to 30 ( or 60 ) seconds might be reasonable for some operations and completely wrong for others. I'm not opposing for changing this to 60 because I don't think that this would work. I think that it would mask the true nature of the issue here, and that we have *something* that consumes large quantity of our CPU time. That "obstacle" is the one that need to be removed rather than adjusting the timeouts.
( i.e. I'm willing to bet that if you'll perform the very same action manually, it will take only 1-2 seconds, whereas the script waits for up to a minute.. why is that ? )

PR #908, opened 2020-03-10
REST API V2

winder:
## Summary

* Update algod rest server to use [echo](https://github.com/labstack/echo).
* Generate V2 endpoint interface and models from an open-api specfile (`daemon/algod/api/server/v2/generated`)
* Modified server to use echo, and updated the old endpoints to be echo compatible (removed gorilla mux from REST endpoints).
* Rewrote auth middleware to be echo compatible.


daemon/algod/api/server/v2/errors.go
```diff
@@ -0,0 +1,30 @@
+package v2
+
+var (
```
winder: I need to take a pass through the errors and make sure I copied them all over from v1 correctly, a lot of these aren't used and will either go away or the code will be updated to use them.

daemon/algod/api/server/v2/handlers.go
```diff
@@ -0,0 +1,544 @@
+package v2
```
winder: Most of this code is transcribed versions or the v1 endpoints.

daemon/algod/api/server/v2/handlers.go
```diff
@@ -0,0 +1,420 @@
...
+// WaitForBlock gets the node status after waiting for the given round.
+// (GET /v2/status/wait-for-block-after/{round}/)
+func (v2 *Handlers) WaitForBlock(ctx echo.Context, round uint64) error {
+	ledger := v2.Node.Ledger()
+	latestBlkHdr, err := ledger.BlockHdr(ledger.Latest())
+	if err != nil {
+		return returnError(ctx, http.StatusInternalServerError, err, errFailedRetrievingNodeStatus, v2.Log)
+	}
+
+	// TODO: Replace this with Node.Status().StoppedAtUnsupportedRound ?
```
rotemh: can be nice

daemon/algod/api/server/v2/handlers.go
```diff
@@ -0,0 +1,420 @@
...
+	if err != nil {
+		return returnError(ctx, http.StatusBadRequest, err, err.Error(), v2.Log)
+	}
+
+	// For backwards compatibility, return txid of first tx in group
+	txid := txgroup[0].ID()
+	return ctx.JSON(http.StatusOK, generated.PostTransactionsResponse{TxId: txid.String()})
+}
+
+// TransactionParams gets parameters for constructing a new transaction
```
rotemh: `returns the suggested parameters for a constructing a new transaction`

daemon/algod/api/server/v2/handlers.go
```diff
@@ -0,0 +1,420 @@
...
+	params.LastRound = uint64(stat.LastRound)
+	params.ConsensusVersion = string(stat.LastVersion)
+
+	proto := config.Consensus[stat.LastVersion]
+	params.MinFee = proto.MinTxnFee
+
+	return ctx.JSON(http.StatusOK, params)
+}
+
+// PendingTransactionInformation gets a specific pending transaction, or looks it up in the ledger if the recently
```
rotemh: nit - fix the description the match the function. Including the number of previous rounds it searches at (MaxTxnLife == 1000), etc.

daemon/algod/api/server/v2/handlers.go
```diff
@@ -0,0 +1,420 @@
...
+
+		return ctx.JSON(http.StatusOK, response)
+	}
+
+	// We didn't find it, return a failure
+	err := errors.New(errTransactionNotFound)
+	return returnError(ctx, http.StatusNotFound, err, err.Error(), v2.Log)
+}
+
+// getPendingTransactions is a generalized version of the get pending transactions endpoints for code reuse.
```
rotemh: seems like this function should not be a function of `v2 *Handlers` (since it's not a handler). Consider making it an independent one and handle the responses in the callers.

daemon/algod/api/server/v2/handlers.go
```diff
@@ -0,0 +1,420 @@
...
+// GetBlock gets the block for the given round.
+// (GET /v2/blocks/{round})
+func (v2 *Handlers) GetBlock(ctx echo.Context, round uint64, params generated.GetBlockParams) error {
+	handle, err := getCodecHandle(params.Format)
+	if err != nil {
+		return returnError(ctx, http.StatusBadRequest, err, errFailedParsingFormatOption, v2.Log)
+	}
+
+	// TODO: What is raw block bytes, should I use that instead?
+	//blockbytes, err := rpcs.RawBlockBytes(v2.Node.Ledger(), basics.Round(round))
```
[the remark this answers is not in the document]
algobolson: v1 has a mode `/block/{}?raw=1` that returns the msgpack blob for a block, the same bytes (using the same server side code, also "Content-Type": rpcs.LedgerResponseContentType) as the catchup interface. Indexer and Algobot need this. Without `?raw=1` it returns JSON per the rest of the API spec.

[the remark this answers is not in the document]
rotemh: This feature stays. The question here is about the certs. Do you use them?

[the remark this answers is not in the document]
algobolson: I don't use the certs, but we should keep them in by default to be true to the intent of "here's the raw data (all of it)". This endpoint was added because the original API block fetch was missing data. I don't think it's worth adding the complexity of optionally stripping them out because I expect most users of this endpoint to be talking to a local `algod` either within one server or at least within the same datacenter. No bandwidth problem. If you want to get this data remotely, then go to a public archival relay node and GET from the catchup endpoint.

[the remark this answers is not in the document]
rotemh: I was talking about optionally adding it in but do you expect the msgpack and json answers to return the same thing for this call? 

Also, how do you expect people to contact the catchup endpoint? it's not exactly publicized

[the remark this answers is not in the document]
algobolson: I'm sure the msgpack case should return everything. The only people who will bother with that mode want everything or can deal with everything. The JSON API can be more tailored to other app developers.

daemon/algod/api/server/v2/handlers.go
```diff
@@ -0,0 +1,420 @@
...
+	// TODO
+	return nil
+}
+
+// AccountInformation gets account information for a given account.
+// (GET /v2/accounts/{address})
+func (v2 *Handlers) AccountInformation(ctx echo.Context, address string) error {
+	addr, err := basics.UnmarshalChecksumAddress(address)
+	if err != nil {
+		return returnError(ctx, http.StatusBadRequest, err, errFailedToParseAddress, v2.Log)
```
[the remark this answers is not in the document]
winder: done

daemon/algod/api/server/v2/handlers.go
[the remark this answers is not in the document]
winder: Based on this conversation I'll include the cert when the user requests `format=msgp`, with the following change:
```
	response := struct {
		Block bookkeeping.Block     `json:"omitempty"`
		Cert  agreement.Certificate `json:"omitempty"`
	} {
		Block: block,
	}

	// If we're using the messagepack encoding, also include the certificates.
	if handle == protocol.CodecHandle {
		response.Certs = cert
	}
```

I'm not using `RawBlockBytes`

daemon/algod/api/server/v2/handlers.go
winder: Yes, thats why I made it private.

daemon/algod/api/server/common/handlers.go
```diff
@@ -20,13 +20,15 @@ import (
...
+	"github.com/labstack/echo/v4"
+
 	"github.com/algorand/go-algorand/config"
 	"github.com/algorand/go-algorand/daemon/algod/api/server/lib"
 	"github.com/algorand/go-algorand/daemon/algod/api/spec/common"
 )
 
 // SwaggerJSON is an httpHandler for route GET /swagger.json
-func SwaggerJSON(ctx lib.ReqContext, w http.ResponseWriter, r *http.Request) {
+func SwaggerJSON(ctx lib.ReqContext, context echo.Context) {
```
tsachiherman, line 31: per your change, you already included the echo.Context within the lib.ReqContext, so why do you need to pass it over again ?

daemon/algod/api/server/v1/routes/routes.go
```diff
@@ -38,7 +36,7 @@ var Routes = lib.Routes{
 	lib.Route{
 		Name:        "wait-for-block",
 		Method:      "GET",
-		Path:        "/status/wait-for-block-after/{round:[0-9]+}",
+		Path:        "/status/wait-for-block-after/:round",
```
tsachiherman, line 39: It looks as if we have lost the ability to do a regex sanitation over the input arguments, which could lead to undesired consequences.
I think that you'd like to retain that. Suggestion : When registering each of these entry points, direct that to a stub function which would perform the sanitation. if all good, it could be forwarded to the underlying implementation.

winder: We were already validating these in the code, so removing the regex didn't really effect anything.

daemon/algod/server.go
```diff
@@ -172,9 +173,11 @@ func (s *Server) Start() {
 
 	s.stopping = make(chan struct{})
 
+	e := echo.New()
+
 	// use the data dir as the static file dir (for our API server), there's
 	// no need to separate the two yet. This lets us serve the swagger.json file.
-	apiHandler := apiServer.NewRouter(s.log, s.node, s.stopping, apiToken)
+	apiServer.ConfigureRouter(s.log, s.node, s.stopping, apiToken, e)
```
tsachiherman, line 180: Could you avoid configuring the default http mux but rather creating your own mux ?
( it's generally bad idea to change the default mux as there is only one of them. especially when you serve http requests on multiple ports.. )

data/transactions/transaction.go
```diff
@@ -366,6 +366,9 @@ func (tx Transaction) RelevantAddrs(spec SpecialAddresses, proto config.Consensu
 		if tx.AssetTransferTxnFields.AssetCloseTo != (basics.Address{}) {
 			addrs = append(addrs, tx.AssetTransferTxnFields.AssetCloseTo)
 		}
+		if tx.AssetTransferTxnFields.AssetSender != (basics.Address{}) {
+			addrs = append(addrs, tx.AssetTransferTxnFields.AssetSender)
+		}
```
tsachiherman, line 371: looks like an unrelated bug fix to me.. ( not opposing, but would rather seeing in a separate PR )

go.mod
```diff
@@ -57,3 +56,5 @@ require (
 	gopkg.in/sohlich/elogrus.v3 v3.0.0-20180410122755-1fa29e2f2009
 	gopkg.in/toast.v1 v1.0.0-20180812000517-0a84660828b2 // indirect
 )
+
+replace github.com/deepmap/oapi-codegen v1.3.5 => github.com/winder/oapi-codegen v1.3.5-algorand
```
tsachiherman, line 60: it's very nice that you forked it.. but if it's still needed, it should go into algorand/oapi-codegen rather then winder/oapi-codegen

daemon/algod/api/server/v2/handlers.go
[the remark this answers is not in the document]
tsachiherman: @rotemh , we need to keep the certs as an option, especially for the raw format.
When we would need to automate the creation of blocks servers, it would be needed.

daemon/algod/api/server/v1/routes/routes.go
tsachiherman: We are validating strings which were previously pruned. Now we're not.
I don't know if the existing string validation is sufficiently "smart" to handle these use cases in the same way. ( do I want it to be ? yes. is it ? I'm not really sure )
for instance, the parser might be now capable of parsing round value of 0x60, whereas it didn't before.

daemon/algod/server.go
winder: I think that echo is the separate mux. I actually had to take special steps to forward pprof endpoints to the default mux.

Is your suggestion that I move `echo.New()` into `ConfigureRouter` and rename it back to `NewRouter`?

daemon/algod/api/server/v1/routes/routes.go
winder: I see, I think that the code we already had to verify these thins is stricter than the regex but maybe I'm incorrect.

My code is now depending on this for round:
```
	queryRound, err := strconv.ParseUint(context.Param("round"), 10, 64)
	if err != nil {
		lib.ErrorResponse(w, http.StatusBadRequest, err, errFailedParsingRoundNumber, ctx.Log)
		return
	}
```

this for address:
```
	queryAddr := context.Param("addr")
	addr, err := basics.UnmarshalChecksumAddress(queryAddr)
	if err != nil {
		lib.ErrorResponse(w, http.StatusBadRequest, err, errFailedToParseAddress, ctx.Log)
		return
	}
```

and this for txid:
```
	queryTxID := context.Param("txid")
	if queryTxID == "" {
		lib.ErrorResponse(w, http.StatusBadRequest, fmt.Errorf(errNoTxnSpecified), errNoTxnSpecified, ctx.Log)
		return
	}

	txID := transactions.Txid{}
	if txID.UnmarshalText([]byte(queryTxID)) != nil {
		lib.ErrorResponse(w, http.StatusBadRequest, fmt.Errorf(errNoTxnSpecified), errNoTxnSpecified, ctx.Log)
		return
	}
```

data/transactions/transaction.go
winder: Pulled these out into a new PR: https://github.com/algorand/go-algorand/pull/948

daemon/algod/server.go
tsachiherman: ok, so I think I got it. Echo replaces the handler in the provided server with it's own so it can process the incoming requests...
Yes, it would be nice to move it all into NewRouter, which would receive a tcpListener as well.
( I'm surprised there is no way to set the Handler on the http.Server to point to the echo handler.

daemon/algod/api/server/v2/handlers.go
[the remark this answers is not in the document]
rotemh: @tsachiherman is the option critical or the fact that it will be there?

daemon/algod/api/server/v1/routes/routes.go
rotemh: It is actually better to not have the regex and it something I wanted to fix in this version. The issue is from the user experience. Since the rejection happens in the routing layer, the user gets the wrong error. For example, if  the user hit `/v2/blocks/345khj`, the user would a get a `404` (because the regex prevented it from matching any address) rather than 400 with sensible "please correct your block number" error message.

daemon/algod/api/server/router.go
```diff
@@ -60,79 +60,75 @@
...
+	e.Use(echo.WrapMiddleware(middlewares.Logger(logger)))
+	// TODO: New auth middleware - https://github.com/algorand/go-algorand/issues/863
+	e.Use(middlewares.Auth(logger, apiToken))
+	e.Use(echo.WrapMiddleware(middlewares.CORS))
 
-	// Middleware
-	router.Use(middlewares.Logger(logger))
-	router.Use(middlewares.Auth(logger, apiToken))
-	router.Use(middlewares.CORS)
+	// Note: Echo has builtin middleware for logging / CORS that we should investigate:
```
rotemh, line 105: it's ok to use any of them, let's just stick to one. 
I'd woudn't change just for the sake of consistency, but don't spend more than 4 minutes on that.

daemon/algod/api/server/v2/errors.go
```diff
@@ -0,0 +1,33 @@
...
+	errFailedLookingUpTransactionPool   = "failed to retrieve information from the transaction pool"
+	errFailedRetrievingNodeStatus       = "failed retrieving node status"
+	errFailedParsingFormatOption        = "failed to parse the format option"
+	errFailedToParseAddress             = "failed to parse the address"
+	errFailedToParseTransaction         = "failed to parse transaction"
+	errFailedToParseBlock               = "failed to parse block"
+	errFailedToParseCert                = "failed to parse cert"
+	errInternalFailure                  = "internal failure"
+	errNoTxnSpecified                   = "no transaction ID was specified"
+	errTransactionNotFound              = "couldn't find the required transaction in the required range"
```
rotemh, line 30: The only place this error is used is when the required txid is not found in pending transactions endpoint. Since the user doesn't have the option to specify any range, this is probably not the right message to provide him with.

daemon/algod/api/server/v2/handlers.go
rotemh: something is still off about this comment. The handler doesn't get any specific pending transaction. the handler takes a TXID and returns the associated transaction's information if found in the Transaction Pool or the last 1000 valid round.

daemon/algod/api/server/v2/handlers.go
rotemh: but it's not. It's still part of the Handlers struct.

rotemh: but it's not. It's still part of `Handlers`.

daemon/algod/api/server/v2/handlers.go
```diff
@@ -0,0 +1,447 @@
...
+
+		return ctx.JSON(http.StatusOK, response)
+	}
+
+	// We didn't find it, return a failure
+	err := errors.New(errTransactionNotFound)
+	return notFound(ctx, err, err.Error(), v2.Log)
+}
+
+// getPendingTransactions is a generalized version of the get pending transactions endpoints for code reuse.
```
rotemh, line 382: consider changing the comment to describe the function by what it does and not by other functions.

daemon/algod/api/server/v2/handlers.go
```diff
@@ -0,0 +1,447 @@
...
+		encodedTxns = append(encodedTxns, encodedTxn)
+	}
+
+	return ctx.JSON(http.StatusOK, generated.PendingTransactionsResponse{
+		TopTransactions:   encodedTxns,
+		TotalTransactions: uint64(len(txns)),
+	})
+}
+
+// GetPendingTransactions gets a list of unconfirmed transactions currently in the transaction pool.
```
rotemh, line 437: // GetPendingTransactions returns the list of unconfirmed transactions currently in the transaction pool.

daemon/algod/api/server/v2/handlers.go
```diff
@@ -0,0 +1,447 @@
...
+// (GET /v2/transactions/pending)
+func (v2 *Handlers) GetPendingTransactions(ctx echo.Context, params generated.GetPendingTransactionsParams) error {
+	return v2.getPendingTransactions(ctx, params.Max, params.Format, nil)
+}
+
+// GetPendingTransactionsByAddress gets a list of unconfirmed transactions currently in the transaction pool by address.
```
rotemh, line 443: GetPendingTransactionsByAddress takes an Algorand address and returns its associated list of unconfirmed transactions currently in the transaction pool.

daemon/algod/api/server/v2/handlers.go
```diff
@@ -0,0 +1,447 @@
...
+
+		// continue if we have an address filter and the address doesn't match the transaction.
+		if addrPtr != nil && !txn.Txn.MatchAddress(*addrPtr, spec) {
+			continue
+		}
+
+		// Encode the transaction and added to the results
+		encodedTxn, err := encode(handle, txn)
+		if err != nil {
+			return internalError(ctx, err, errFailedLookingUpTransactionPool, v2.Log)
```
rotemh, line 426: why do we return this error for a failed parsing?

daemon/algod/api/server/v2/handlers.go
```diff
@@ -0,0 +1,447 @@
...
+
+	var params generated.TransactionParametersResponse
+	params.Fee = v2.Node.SuggestedFee().Raw
+	params.GenesisId = v2.Node.GenesisID()
+	params.GenesisHash = gh[:]
+	params.LastRound = uint64(stat.LastRound)
+	params.ConsensusVersion = string(stat.LastVersion)
+
+	proto := config.Consensus[stat.LastVersion]
+	params.MinFee = proto.MinTxnFee
```
rotemh, line 325: Consider changing it to initialization using a struct literal for consistency.

daemon/algod/api/server/v2/handlers.go
```diff
@@ -0,0 +1,447 @@
...
+		NextVersionSupported:      stat.NextVersionSupported,
+		TimeSinceLastRound:        uint64(stat.TimeSinceLastRound().Nanoseconds()),
+		CatchupTime:               uint64(stat.CatchupTime.Nanoseconds()),
+		StoppedAtUnsupportedRound: stat.StoppedAtUnsupportedRound,
+	}
+
+	return ctx.JSON(http.StatusOK, response)
+}
+
+// WaitForBlock gets the node status after waiting for the given round.
```
rotemh, line 250: nit - returns the node's status?

daemon/algod/api/server/v2/utils.go
```diff
@@ -0,0 +1,165 @@
...
+	} else if format == "msgpack" || format == "msgp" {
+		handle = protocol.CodecHandle
+	} else {
+		return nil, fmt.Errorf("invalid format: %s", format)
+	}
+
+	return handle, nil
+}
+
+func encode(handle codec.Handle, obj interface{}) (string, error) {
```
rotemh, line 156: what does this function do?
encode from what to what?

daemon/algod/api/server/v2/handlers.go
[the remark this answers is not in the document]
tsachiherman: @rotemh the option is not *currently* being used, but I was expecting to use this feature when implementing the block storage servers. If it's too big of a change for now, it's ok to loose that ability, and we could work down the road on bringing it back.

daemon/algod/api/server/v1/routes/routes.go
tsachiherman: Yes, that's a good point. My concern is that we need now to revalidate all the entry point parsers.

daemon/algod/api/server/v2/utils.go
winder: It encodes to JSON or MessagePack depending on which codec handle is passed in.

daemon/algod/api/server/v1/routes/routes.go
winder: @tsachiherman Great, sounds like we are all on the same page here. When making this change I went through the exercise of verifying that each spot where I replaced the regex validation was properly validated in the code.

daemon/algod/api/server/v2/handlers.go
winder: I guess I'm not sure what your concern is here. It isn't visible outside of the `v2` package, if there was a way to reduce the scope further I would have done that.

In addition to not being visible outside of the package, it's also not part of the interface being implemented by this class.

daemon/algod/api/server/v2/handlers.go
winder: The comment reflects what happens inside `v2.Node.GetPendingTransaction(txID)`, which looks for the specified txID in the transaction pool, and if not found it looks for it in the last `proto.MaxTxnLife` rounds.

daemon/algod/api/server/v2/handlers.go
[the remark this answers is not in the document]
winder: @tsachiherman could you give me some specific requirements for what you need?

With the current implementation `format=msgp` should provide the certs. The format is a little weird though:
```
{
    "block": "<json/msgp encoded block object>",
    "cert": "<msgp encoded cert object>"
}
```

So the same data is available but there is an extra step to decode the field.

daemon/algod/server.go
winder: I made this change.

Also, it turns out `StartServer` on the echo object is a thin wrapper around `http.Server.Serve`, and sets itself to be the handler. So it's actually working exactly the way that you thought it should:
```
// StartServer starts a custom http server.
func (e *Echo) StartServer(s *http.Server) (err error) {
	// Setup
	e.colorer.SetOutput(e.Logger.Output())
	s.ErrorLog = e.StdLogger
	s.Handler = e
```

daemon/algod/api/server/v2/handlers.go
rotemh: > looks for the specified txID in the transaction pool, and if not found it looks for it in the last proto.MaxTxnLife rounds

This is a much better comment for this handler.

data/transactions/common_test.go
```diff
@@ -78,7 +78,7 @@ func (b BalanceMap) Put(record basics.BalanceRecord) error {
 // set up a BalanceMap for a transaction containing only the transactions RelevantAddrs.
 func makeTestBalancesForTransaction(tx Transaction) BalanceMap {
 	bals := make(BalanceMap)
-	for _, addr := range tx.RelevantAddrs(SpecialAddresses{RewardsPool: poolAddr}, config.Consensus[protocol.ConsensusCurrentVersion]) {
+	for _, addr := range tx.RelevantAddrs(SpecialAddresses{RewardsPool: poolAddr}) {
```
tsachiherman, line 81: I think that you already did this on a separate PR. Could you rebase against master ?

go.mod
winder: yes, done.

daemon/algod/api/server/common/handlers.go
winder: I decided against making a copy of the `ReqContext` for each REST call, but forgot to remove the Context from it. Removed `echo.Context` from `ReqContext`.

winder:
@rotemh @tsachiherman I believe that all of your feedback has been addressed.

daemon/algod/api/server/v2/handlers.go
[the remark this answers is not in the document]
tsachiherman: Yes, I can. When we would want to have an external block server, we need a way to extract the blocks from the algod instance in a way where we can serve that to another algod which wouldn't be able to tell the difference.
In practice we want to make sure that whatever we have in the database, is being exported without any decoding/encoding. That "blob" would then be stored in S3 ( or others ) and served as a file.
Adding any msgpack decoding/encoding might modify the content, which is what we'd like to avoid.
( it's not just performance issue - it might also modify the correctness of the process )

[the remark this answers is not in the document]
algobolson: Can we just keep the current behavior where `/block/{blocknumber}?raw=1` returns the exact same data as the catchup endpoint?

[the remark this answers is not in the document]
algobolson: The catchup endpoint is definitionally all the data anyone could need. It's already done. It's simple. Let's do that?

daemon/algod/api/server/v2/handlers.go
```diff
@@ -0,0 +1,448 @@
...
+
+// GetBlock gets the block for the given round.
+// (GET /v2/blocks/{round})
+func (v2 *Handlers) GetBlock(ctx echo.Context, round uint64, params generated.GetBlockParams) error {
+	handle, err := getCodecHandle(params.Format)
+	if err != nil {
+		return badRequest(ctx, err, errFailedParsingFormatOption, v2.Log)
+	}
+
+	// TODO: What is raw block bytes, should I use that instead?
```
algobolson, line 172: RawBlockBytes needs to be _available_ but it doesn't have to be the default. (in v1 it's enabled by `?raw=1`). **Indexer depends on this data.**

daemon/algod/api/server/v2/handlers.go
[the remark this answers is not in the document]
rotemh: @algobolson correct me if I'm wrong but the new endpoint is just an extension of the old one. The endpoint still delivers the block in msgpack format just with the certs as well. Is there any information that indexer is dependent on that is not being returned now?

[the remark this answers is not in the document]
algobolson: that's not the current implementation and this sounded like a long thread about maybe doing something else so I wanted to reiterate that we should indeed make available exactly the raw data that the current block endpoint does

daemon/algod/api/server/v2/handlers.go
```diff
@@ -0,0 +1,449 @@
...
+	}
+
+	// msgpack format will return the raw block bytes and attach some custom headers.
+	if handle == protocol.CodecHandle {
+		blockbytes, err := rpcs.RawBlockBytes(v2.Node.Ledger(), basics.Round(round))
+		if err != nil {
+			return internalError(ctx, err, errFailedToLookupRawBlock, v2.Log)
+		}
+
+		ctx.Response().Writer.Header().Add("Content-Encoding", "application/x-algorand-block-v1")
```
algobolson, line 180: nonstandard use of Content-Encoding. I think we should drop this and just use the X- header
https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.11

daemon/algod/api/server/v2/handlers.go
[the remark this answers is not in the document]
winder: updates have been made based on this discussion

daemon/algod/api/server/router.go
```diff
@@ -60,79 +60,83 @@
...
```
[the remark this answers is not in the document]
winder: I left this for future-me, re-implementing the auth is in the backlog, and auth/logging/cors still needs to be added to indexer. After those tickets I expect that whatever work done there will be applied here as well.

daemon/algod/api/server/v2/handlers.go
```diff
@@ -0,0 +1,448 @@
...
+type Handlers struct {
+	Node     *node.AlgorandFullNode
+	Log      logging.Logger
+	Shutdown <-chan struct{}
+}
+
+// RegisterParticipationKeys registers participation keys.
+// (POST /v2/register-participation-keys/{address})
+func (v2 *Handlers) RegisterParticipationKeys(ctx echo.Context, address string, params generated.RegisterParticipationKeysParams) error {
+	// TODO: register participation keys endpoint
```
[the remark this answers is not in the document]
winder: Yes. My intention was to finish all TODOs except for where there is a prioritized ticket in the backlog.

PR #913, opened 2020-03-12
Catchup service through gossip connections using the topics

algonautshant:
* Introduce versioning: old: 1, new: 2.1
* Add keys to topics as const in topics.go
* Introduce MakeTopic to create topics without exporting the Topic fields
* Change wsPeer.Respond to accept Topics instead of OutgoingMessage
* Add methods: Version, Request, Respond to UnicastPeer interface (should rename UnicastPeer?)
* Make LedgerService handleCatchupReq support v2.1 (request/respond using topics)
* Make WsFetcherService RequestBlock support v2.1 (request/respond using topics)

Modifying TestGetBlockWS to test version 2.1

## Test Plan

Will add e2e test to test all cases of the two version combinations.

### DIFF
--- a/network/topics.go
+++ b/network/topics.go
@@ -23,15 +23,10 @@ import (
 	"github.com/algorand/go-algorand/crypto"
 )
 
-// Constant strings used as keys or values for topics
+// Constant strings used as keys for topics
 const (
-	RoundKey           = "ky"
-	RequestDataTypeKey = "typ"
-	RequestHashKey     = "RequestHash"
-	BlockDataKey       = "BDK"
-	CertDataKey        = "CDK"
-	ErrorKey           = "ERR"
-	BlockAndCertValue  = "BlockAndCertificate"
+	requestHashKey     = "RequestHash"
+	ErrorKey           = "ERR" // used for passing an error message
 )
 
 // Topic is a key-value pair
@@ -40,7 +35,7 @@ type Topic struct {
 	data []byte
 }
 
-// MakeTopic Create a Topic
+// MakeTopic Creates a Topic
 func MakeTopic(key string, data []byte) Topic {
 	return Topic{key: key, data: data}
 }
--- a/network/topics.go
+++ b/network/topics.go
@@ -23,15 +23,15 @@ import (
 	"github.com/algorand/go-algorand/crypto"
 )
 
-
+// Constant strings used as keys or values for topics
 const (
-	RoundKey  = "ky"
-	RequestDataTypeKey  = "typ"
-	RequestHashKey = "RequestHash"
-	BlockDataKey = "BDK"
-	CertDataKey = "CDK"
-	ErrorKey = "ERR"
-	BlockAndCertValue = "BlockAndCertificate"
+	RoundKey           = "ky"
+	RequestDataTypeKey = "typ"
+	RequestHashKey     = "RequestHash"
+	BlockDataKey       = "BDK"
+	CertDataKey        = "CDK"
+	ErrorKey           = "ERR"
+	BlockAndCertValue  = "BlockAndCertificate"
 )
 
 // Topic is a key-value pair
@@ -40,8 +40,9 @@ type Topic struct {
 	data []byte
 }
 
+// MakeTopic Create a Topic
 func MakeTopic(key string, data []byte) Topic {
-	return Topic {key: key, data:data}
+	return Topic{key: key, data: data}
 }
 
 // Topics is an array of type Topic
--- a/network/topics.go
+++ b/network/topics.go
@@ -25,8 +25,8 @@ import (
 
 // Constant strings used as keys for topics
 const (
-	requestHashKey     = "RequestHash"
-	ErrorKey           = "ERR" // used for passing an error message
+	requestHashKey = "RequestHash"
+	ErrorKey       = "ERR" // used for passing an error message
 )
 
 // Topic is a key-value pair
--- a/network/topics.go
+++ b/network/topics.go
@@ -23,10 +23,25 @@ import (
 	"github.com/algorand/go-algorand/crypto"
 )
 
+
+const (
+	RoundKey  = "ky"
+	RequestDataTypeKey  = "typ"
+	RequestHashKey = "RequestHash"
+	BlockDataKey = "BDK"
+	CertDataKey = "CDK"
+	ErrorKey = "ERR"
+	BlockAndCertValue = "BlockAndCertificate"
+)
+
 // Topic is a key-value pair
 type Topic struct {
 	key  string
 	data []byte
+}
+
+func MakeTopic(key string, data []byte) Topic {
+	return Topic {key: key, data:data}
 }
 
 // Topics is an array of type Topic
--- a/network/topics.go
+++ b/network/topics.go
@@ -26,7 +26,7 @@ import (
 // Constant strings used as keys for topics
 const (
 	requestHashKey = "RequestHash"
-	ErrorKey       = "ERR" // used for passing an error message
+	ErrorKey       = "Error" // used for passing an error message
 )
 
 // Topic is a key-value pair
--- a/network/wsPeer.go
+++ b/network/wsPeer.go
@@ -255,7 +255,7 @@ func (wp *wsPeerCore) PrepareURL(rawURL string) string {
 	return strings.Replace(rawURL, "{genesisID}", wp.net.GenesisID, -1)
 }
 
-func (wp *wsPeer) Verion() string {
+func (wp *wsPeer) Version() string {
 	return wp.version
 }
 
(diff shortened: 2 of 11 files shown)


network/wsPeer.go
```diff
@@ -238,6 +241,10 @@ func (wp *wsPeerCore) PrepareURL(rawURL string) string {
 	return strings.Replace(rawURL, "{genesisID}", wp.net.GenesisID, -1)
 }
 
+func (wp *wsPeer) Verion() string {
```
tsachiherman: add some description for what the function is doing.

network/wsPeer.go
```diff
@@ -397,9 +403,9 @@ func (wp *wsPeer) readLoop() {
 				wp.net.log.Warnf("wsPeer readLoop: could not read the message from: %s %s", wp.conn.RemoteAddr().String(), err)
 				continue
 			}
-			requestHash, found := topics.GetValue("RequestHash")
+			requestHash, found := topics.GetValue(RequestHashKey)
 			if !found {
-				wp.net.log.Warnf("wsPeer readLoop: message from %s is missing the RequestHash", wp.conn.RemoteAddr().String())
+				wp.net.log.Warnf("wsPeer readLoop: message from %s is missing the %v", wp.conn.RemoteAddr().String(), RequestHashKey)
```
tsachiherman: use %s instead of %v

rpcs/ledgerService.go
```diff
@@ -229,10 +230,37 @@ func (ls *LedgerService) ListenForCatchupReq(reqs <-chan network.IncomingMessage
 // a blocking function for handling a catchup request
 func (ls *LedgerService) handleCatchupReq(ctx context.Context, reqMsg network.IncomingMessage) {
 	var res WsGetBlockOut
+	target := reqMsg.Sender.(network.UnicastPeer)
+	var respTopics network.Topics
+
+	if target.Version() == "2.1" {
```
tsachiherman: I think that we want to flip this around. If we have version "1", then we need to use the old implementation, otherwise, we should use the new implementation.

rpcs/wsFetcherService.go
```diff
@@ -104,7 +105,48 @@ func (fs *WsFetcherService) RequestBlock(ctx context.Context, target network.Uni
...
+				network.RoundKey,
+				roundBin),
+		}
+		resp, err := target.Request(ctx, tag, topics)
+		if err != nil {
+			return WsGetBlockOut{}, fmt.Errorf("WsFetcherService.RequestBlock(%d): Request failed, %v", round, err)
+		}
+
+		if errMsg, found := resp.Topics.GetValue(network.ErrorKey); found {
+			return WsGetBlockOut{}, fmt.Errorf(string(errMsg))
```
tsachiherman: Avoid using `fmt.Errorf` if you're not going to add any information to the error. `errors.New` would do that for you.
In this case, you want to argument the error string with some context-related data, since the error is literally just the string that came from the other end.

rpcs/ledgerService.go
```diff
@@ -229,10 +230,37 @@ func (ls *LedgerService) ListenForCatchupReq(reqs <-chan network.IncomingMessage
...
+		}
+		roundBytes, found := topics.GetValue(network.RoundKey)
+		if !found {
+			topics = network.Topics{
+				network.MakeTopic(network.ErrorKey,
+					[]byte("LedgerService handleCatchupReq: round number topic is missing"))}
+			return
+		}
+		requestType, _ := topics.GetValue(network.RequestDataTypeKey)
+		round, _ := binary.Uvarint(roundBytes)
```
tsachiherman: You need to have error checking here to avoid potential attack vector.
( both requestType and round )

rpcs/ledgerService.go
```diff
@@ -259,6 +287,23 @@ func (ls *LedgerService) sendCatchupRes(ctx context.Context, target network.Unic
 	}
 }
 
+func topicBlockBytes(ledger *data.Ledger, round basics.Round, requestType string) network.Topics {
+	blk, cert, err := ledger.EncodedBlockCert(round)
+	if err != nil {
+		return network.Topics{
+			network.MakeTopic(network.ErrorKey, []byte(err.Error()))}
+	}
+	if requestType == network.BlockAndCertValue {
```
tsachiherman: nit: use switch-case here, and if the requestType is not supported, return an error code.

network/topics.go
```diff
@@ -23,12 +23,28 @@ import (
...
 
+// Constant strings used as keys or values for topics
+const (
+	RoundKey           = "ky"
+	RequestDataTypeKey = "typ"
+	RequestHashKey     = "RequestHash"
+	BlockDataKey       = "BDK"
+	CertDataKey        = "CDK"
+	ErrorKey           = "ERR"
+	BlockAndCertValue  = "BlockAndCertificate"
```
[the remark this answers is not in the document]
algonautshant: I would like to keep the ErrorKey here, just to have it uniform in the code.

rpcs/ledgerService.go
```diff
@@ -229,24 +230,65 @@ func (ls *LedgerService) ListenForCatchupReq(reqs <-chan network.IncomingMessage
...
+	if !found {
+		topics = network.Topics{
+			network.MakeTopic(network.ErrorKey,
+				[]byte("LedgerService handleCatchupReq: round-number topic is missing"))}
 		return
 	}
-	res.BlockBytes = encodedBlob
+	requestType, found := topics.GetValue(requestDataTypeKey)
+	if !found {
+		topics = network.Topics{
```
tsachiherman: changing the `topic` here won't help much. I think that you meant to change the `respTopics`.
( same goes for the rest of the errors )

rpcs/ledgerService.go
```diff
@@ -229,24 +230,65 @@ func (ls *LedgerService) ListenForCatchupReq(reqs <-chan network.IncomingMessage
...
+			network.MakeTopic(network.ErrorKey,
+				[]byte("LedgerService handleCatchupReq: request data-type is missing"))}
+		return
+	}
+	
+	round, read := binary.Uvarint(roundBytes)
+	if read <= 0 {
+		topics = network.Topics{
+			network.MakeTopic(network.ErrorKey,
+				[]byte("LedgerService handleCatchupReq: error reading the round number"))}
```
tsachiherman: while the error string here is valid, it's not really what the other party expects.
If you were to write this error to the local log, it would make sense. however, since you're sending it to the other side, you should omit the function name/service and focus on the reason this message was wrong. i.e. `Unable to parse round number`.

rpcs/wsFetcherService.go
```diff
@@ -36,6 +37,16 @@ type WsFetcherService struct {
 	pendingRequests map[string]chan WsGetBlockOut
 }
 
+
+// Constant strings used as keys for topics
+const (
+	roundKey           = "roundKey"
```
tsachiherman: nit: could you add description for the meaning of each of these constants ?

network/topics.go
```diff
@@ -23,12 +23,23 @@ import (
 	"github.com/algorand/go-algorand/crypto"
 )
 
+// Constant strings used as keys for topics
+const (
+	requestHashKey = "RequestHash"
+	ErrorKey       = "ERR" // used for passing an error message
```
tsachiherman: nit: all the reminder of the topics are nicely formatted string, with the exception for this one. shouldn't we make it into `Error` to align the formatting to be the same ?

rpcs/ledgerService.go
```diff
@@ -259,6 +308,28 @@ func (ls *LedgerService) sendCatchupRes(ctx context.Context, target network.Unic
 	}
 }
 
+func topicBlockBytes(ledger *data.Ledger, round basics.Round, requestType string) network.Topics {
+	blk, cert, err := ledger.EncodedBlockCert(round)
+	if err != nil {
+		errMsg := "LedgerService topicBlockBytes: error in EncodedBlockCert: " + err.Error()
+		return network.Topics{
+			network.MakeTopic(network.ErrorKey, []byte(errMsg))}
```
tsachiherman: I think that you would want better error message, specifically for the `ledger.ErrNoEntry` case.
The rationale here is that it's very common to ask for a block where the other party doesn't have it.
In this case, returning an error such as `requested block is not available` would be a better error message.
( there is no "error" here )

rpcs/ledgerService.go
```diff
@@ -259,6 +308,28 @@ func (ls *LedgerService) sendCatchupRes(ctx context.Context, target network.Unic
...
+	switch requestType {
+	case blockAndCertValue:
+		return network.Topics{
+			network.MakeTopic(
+				blockDataKey, blk),
+			network.MakeTopic(
+				certDataKey, cert),
+		}
+	default:
+		errMsg := "LedgerService topicBlockBytes: request type is unknown"
```
tsachiherman: nit: use the same "language" that the caller has used. i.e. `requestedDataType specifies unsupported data type`

PR #917, opened 2020-03-18
Applications

justicz:


ledger/eval.go
```diff
@@ -545,11 +579,11 @@ func (eval *BlockEvaluator) transaction(txn transactions.SignedTxn, ad transacti
 	// If we are validating and generating, we have no ApplyData yet.
 	if eval.validate && !eval.generate {
 		if eval.proto.ApplyData {
-			if ad != applyData {
+			if !ad.Equal(applyData) {
 				return fmt.Errorf("transaction %v: applyData mismatch: %v != %v", txn.ID(), ad, applyData)
 			}
 		} else {
-			if ad != (transactions.ApplyData{}) {
+			if !ad.Equal(transactions.ApplyData{}) {
```
pzbitskiy: Empty() to get rid of copying

data/basics/teal.go
```diff
@@ -0,0 +1,253 @@
...
+
+import (
+	"fmt"
+
+	"github.com/algorand/go-algorand/config"
+)
+
+// DeltaAction is an enum of actions that may be performed when applying a
+// delta to a TEAL key/value store
+type DeltaAction uint64
```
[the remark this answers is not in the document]
justicz: I don't think it matters to the codec as long as the numbers are small, it will pick the smallest msgpack integer type that fits.

data/transactions/application.go
```diff
@@ -0,0 +1,607 @@
...
+	OnCompletion    OnCompletion     `codec:"apan"`
+	ApplicationArgs []string         `codec:"apaa,allocbound=1024"`
+	Accounts        []basics.Address `codec:"apat,allocbound=1024"`
+
+	LocalStateSchema  basics.StateSchema `codec:"apls"`
+	GlobalStateSchema basics.StateSchema `codec:"apgs"`
+	ApprovalProgram   string             `codec:"apap"`
+	ClearStateProgram string             `codec:"apsu"`
+
+	// If you add any fields here, remember you MUST modify the Empty
```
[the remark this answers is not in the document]
justicz: We could use the reflect package (which internally uses unsafe, but is probably fine?)

data/transactions/application.go
```diff
@@ -0,0 +1,607 @@
...
+	}
+	return res
+}
+
+// getAppParams fetches the creator address and AppParams for the app index,
+// if they exist. It does NOT clone the AppParams, so the returned params must
+// not be modified directly.
+func getAppParams(balances Balances, aidx basics.AppIndex) (params basics.AppParams, creator basics.Address, doesNotExist bool, err error) {
+	creator, doesNotExist, err = balances.GetAppCreator(aidx)
+	if err != nil {
```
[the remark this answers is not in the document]
justicz: these are different cases -- we care about distinguishing between a database error and the app not existing.

data/basics/userBalance.go
```diff
@@ -136,6 +136,49 @@ type AccountData struct {
...
+
+	Schema   StateSchema  `codec:"hsch"`
+	KeyValue TealKeyValue `codec:"tkv"`
+}
+
+// AppParams stores the global information associated with an application
+type AppParams struct {
+	_struct struct{} `codec:",omitempty,omitemptyarray"`
+
+	ApprovalProgram   string      `codec:"approv,allocbound=-"`
```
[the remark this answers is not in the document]
justicz: I chose a string here because in general I like not having to worry about nil checking in functions like `Clone` below. But `[]byte` is probably more appropriate.

data/transactions/transaction.go
```diff
@@ -263,6 +299,66 @@ func (tx Transaction) WellFormed(spec SpecialAddresses, proto config.ConsensusPa
...
+
+		if tx.ApplicationID != 0 && tx.OnCompletion != UpdateApplicationOC {
+			// Ensure programs are only set for creation or update
+			if tx.ApprovalProgram != "" || tx.ClearStateProgram != "" {
+				return fmt.Errorf("scripts may only be specified during application creation or update")
+			}
+		}
+
+		if tx.ApplicationID != 0 {
+			if tx.LocalStateSchema != (basics.StateSchema{}) ||
```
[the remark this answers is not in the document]
justicz: I think we should compare with an empty struct like this wherever possible, because it's likely the compiler can optimize it well. I only added `Empty` methods when they are necessary because the struct contains a reference type.

data/basics/sort.go
```diff
@@ -25,3 +25,33 @@ type SortAssetIndex []AssetIndex
...
+func (a SortString) Swap(i, j int)      { a[i], a[j] = a[j], a[i] }
+
+// SortAddress implements sorting by Address keys for
+// canonical encoding of maps in msgpack format.
+//msgp:ignore SortAddress
+//msgp:sort Address SortAddress
+type SortAddress []Address
+
+func (a SortAddress) Len() int           { return len(a) }
+func (a SortAddress) Less(i, j int) bool { return string(a[i][:]) < string(a[j][:]) }
```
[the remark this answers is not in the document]
justicz: changed :+1:

daemon/algod/api/server/v1/handlers/handlers.go
```diff
@@ -791,6 +833,7 @@ func PendingTransactionInformation(ctx lib.ReqContext, w http.ResponseWriter, r
 			// it was created in), because computeAssetIndexFromTxn will
 			// return 0 in that case.
 			CreatedAssetIndex: computeAssetIndexFromTxn(txn, ledger),
+			CreatedAppIndex:   computeAppIndexFromTxn(txn, ledger),
```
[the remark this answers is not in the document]
justicz: yes -- fixed the comment above

ledger/accountdb.go
```diff
@@ -57,11 +57,18 @@ var accountsSchema = []string{
 		creator blob)`,
 }
 
+var creatablesMigration = []string{
+	`ALTER TABLE assetcreators RENAME TO creatables`,
+	`ALTER TABLE creatables RENAME COLUMN asset TO creatable`,
+	`ALTER TABLE creatables ADD COLUMN ctype INTEGER DEFAULT 0`,
```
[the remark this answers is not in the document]
justicz: I just tried this with the following index:

```
CREATE INDEX creatable_ctype_idx ON creatables (creatable, ctype)
```

I then asked sqlite to explain its query plan on the two lookup queries we do:

```
sqlite> EXPLAIN QUERY PLAN SELECT creatable, creator, ctype FROM creatables WHERE creatable <= 1000 AND ctype = 1 ORDER BY creatable desc LIMIT 10;
0|0|0|SEARCH TABLE creatables USING INTEGER PRIMARY KEY (rowid<?)
sqlite> EXPLAIN QUERY PLAN SELECT creator FROM creatables WHERE creatable = 1234 AND ctype = 1234;
0|0|0|SEARCH TABLE creatables USING INTEGER PRIMARY KEY (rowid=?)
```

It doesn't look like it wants to use the index in either case, I'm not sure why...

ledger/appeval.go
```diff
@@ -0,0 +1,208 @@
+// Copyright (C) 2019-2020 Algorand, Inc.
```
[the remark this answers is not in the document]
justicz: done :+1:

data/basics/teal.go
```diff
@@ -0,0 +1,253 @@
...
+
+// StateDelta is a map from key/value store keys to ValueDeltas, indicating
+// what should happen for that key
+//msgp:allocbound StateDelta config.MaxStateDeltaKeys
+type StateDelta map[string]ValueDelta
+
+// Equal checks whether two StateDeltas are equal. We don't check for nilness
+// equality because an empty map will encode/decode as nil. So if our generated
+// map is empty but not nil, we want to equal a decoded nil off the wire.
+func (sd StateDelta) Equal(o StateDelta) bool {
```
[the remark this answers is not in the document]
justicz: maps are already reference types, so I don't think making it a map pointer would make a difference, right?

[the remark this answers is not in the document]
pzbitskiy: Right, I confused it with larger `ValueDelta` type.

data/transactions/logic/eval.go
```diff
@@ -291,7 +423,12 @@ func Check(program []byte, params EvalParams) (cost int, err error) {
 	cx.EvalParams = params
 	cx.program = program
 	for (cx.err == nil) && (cx.pc < len(cx.program)) {
+		prevpc := cx.pc
 		cost += cx.checkStep()
+		if cx.pc == prevpc {
```
justicz: this is a good check! you might consider making it `<=` to catch even more potential problems

data/transactions/logic/eval.go
```diff
@@ -451,42 +456,82 @@ const MaxStackDepth = 1000
...
+		cx.err = fmt.Errorf("%s not allowed in current mode", spec.Name)
 		return
 	}
-	argsTypes := opsByOpcode[opcode].Args
+	if spec.Version > cx.version {
+		cx.err = fmt.Errorf("%s not allowed in program version %d", spec.Name, cx.version)
+		return
+	}
+	argsTypes := spec.Args
 	if len(argsTypes) >= 0 {
```
justicz: isn't this always true?

pzbitskiy: it is, @algobolson do you remember the rationale?

data/transactions/logic/eval.go
```diff
@@ -1549,81 +1554,70 @@ func opAppCheckOptedIn(cx *evalContext) {
...
-		var err error
-		tkv, err = cx.Ledger.AppLocalState(addr, basics.AppIndex(appID))
+func (cx *evalContext) getReadOnlyLocalState(appID uint64, addr basics.Address) (basics.TealKeyValue, error) {
+	kvIdx := ckey{appID, addr}
+	if cx.readOnlyLocalStates[kvIdx] == nil {
+		localKV, err := cx.Ledger.AppLocalState(addr, basics.AppIndex(appID))
+		if err != nil {
+			return nil, fmt.Errorf("failed to fetch app local state local state for acct %s, app %d: %v", addr, appID, err)
+		}
+		cx.readOnlyLocalStates[kvIdx] = localKV
```
pzbitskiy, line 1564: `return localKV, nil` to eliminate exceed lookup

data/transactions/application.go
```diff
@@ -0,0 +1,769 @@
...
+		return "update"
+	case DeleteApplicationOC:
+		return "delete"
+	}
+	return "?"
+}
+
+// ApplicationCallTxnFields captures the transaction fields used for all
+// interactions with applications
+type ApplicationCallTxnFields struct {
```
winder, line 77: Could you add comments here describing the different fields?

daemon/algod/api/spec/v1/model.go
```diff
@@ -542,6 +691,74 @@ type AssetFreezeTransactionType struct {
 	NewFreezeStatus bool `json:"freeze"`
 }
 
+// ApplicationCallTransactionType contains the additional fields for an ApplicationCall transaction
+// swagger:model ApplicationCallTransactionType
+type ApplicationCallTransactionType struct {
```
winder, line 696: All of these fields are marked as required, but that isn't really true is it? Could we make an OnCompletion / field matrix to define when they are required?

ledger/acctupdates.go
```diff
@@ -276,49 +279,61 @@ func (au *accountUpdates) allBalances(rnd basics.Round) (bals map[basics.Address
...
-	for aidx := range au.assets {
-		if aidx <= maxAssetIdx {
-			keys = append(keys, aidx)
+	keys := make([]basics.CreatableIndex, 0, len(au.creatables))
+	for cidx, delta := range au.creatables {
+		if delta.ctype != ctype {
+			continue
+		}
+		if cidx <= maxCreatableIdx {
+			keys = append(keys, cidx)
```
algorandskiy, line 300: should break here in `else` ?

algomaxj: I don't think so, since `au.creatables` is a map

config/consensus.go
```diff
@@ -217,6 +217,31 @@ type ConsensusParams struct {
...
+	AppFlatParamsMinBalance  uint64
+	AppFlatOptInMinBalance   uint64
+	SchemaMinBalancePerEntry uint64
+	SchemaUintMinBalance     uint64
+	SchemaBytesMinBalance    uint64
+	MaxLocalSchemaEntries    uint64
+	MaxGlobalSchemaEntries   uint64
+
+	MaximumMinimumBalance uint64
 }
```
algorandskiy, line 245: all these fields must be documented

ledger/eval.go
```diff
@@ -77,10 +86,18 @@ func (x *roundCowBase) txnCounter() uint64 {
...
-func (cs *roundCowState) GetAssetCreator(assetIdx basics.AssetIndex) (basics.Address, error) {
+func (cs *roundCowState) GetAssetCreator(assetIdx basics.AssetIndex) (creator basics.Address, ok bool, err error) {
 	return cs.getAssetCreator(assetIdx)
 }
 
+func (cs *roundCowState) GetAppCreator(appIdx basics.AppIndex) (creator basics.Address, ok bool, err error) {
+	return cs.getAppCreator(appIdx)
+}
+
+func (cs *roundCowState) Round() basics.Round {
```
algorandskiy, line 97: is it used somewhere? Seems like a leftover after introduction
```
			AppTealGlobals: AppTealGlobals{
				CurrentRound:    eval.prevHeader.Round + 1,
				LatestTimestamp: eval.prevHeader.TimeStamp,
```

daemon/algod/api/spec/v1/model.go
winder: Something similar to the Assets documentation on the developer site would be even better:

https://developer.algorand.org/docs/reference/transactions/#asset-transfer-transaction

ledger/acctupdates.go
```diff
@@ -1005,6 +1002,12 @@ func (au *accountUpdates) commitRound(offset uint64, dbRound basics.Round, lookb
 		}
 	}
 
+	for i := uint64(0); i < offset; i++ {
```
tsachiherman, line 1005: nit: you could have merged this for-loop with the one above it. ( but it's correct either way )

algomaxj: ah, good call :+1:

daemon/algod/api/server/v1/handlers/handlers.go
```diff
@@ -207,6 +257,40 @@ func assetConfigTxEncode(tx transactions.Transaction, ad transactions.ApplyData)
...
+	app := v1.ApplicationCallTransactionType{
+		ApplicationID:     uint64(tx.ApplicationID),
+		ApprovalProgram:   b64.EncodeToString([]byte(tx.ApprovalProgram)),
+		ClearStateProgram: b64.EncodeToString([]byte(tx.ClearStateProgram)),
+		LocalStateSchema:  modelSchema(tx.LocalStateSchema),
+		GlobalStateSchema: modelSchema(tx.GlobalStateSchema),
+		OnCompletion:      tx.OnCompletion.String(),
+	}
+
+	var encodedAccounts []string
```
tsachiherman, line 271: nit: preallocate `encodedAccounts`, `encodedForeignApps` and `encodedForeignApps`

daemon/algod/api/server/v1/handlers/handlers.go
```diff
@@ -648,11 +767,27 @@ func AccountInformation(ctx lib.ReqContext, context echo.Context) {
...
+		assetParams = make(map[uint64]v1.AssetParams)
 		for idx, params := range record.AssetParams {
-			thisAssetParams[uint64(idx)] = assetParams(addr, params)
+			assetParams[uint64(idx)] = modelAssetParams(addr, params)
+		}
+	}
+
+	var apps map[uint64]v1.AppLocalState
+	if len(record.AppLocalStates) > 0 {
+		apps = make(map[uint64]v1.AppLocalState)
```
tsachiherman, line 780: nit: preallocate map size.

daemon/algod/api/server/v2/account.go
```diff
@@ -0,0 +1,348 @@
...
+	}
+
+	createdAssets := make([]generated.Asset, 0, len(record.AssetParams))
+	for idx, params := range record.AssetParams {
+		asset := AssetParamsToAsset(address, idx, &params)
+		createdAssets = append(createdAssets, asset)
+	}
+
+	var apiParticipation *generated.AccountParticipation
+	if record.VoteID != (crypto.OneTimeSignatureVerifier{}) {
```
tsachiherman, line 55: nit: would have been nice if OneTimeSignatureVerifier could have a "IsZero" / "IsNotZero" methods so that you wouldn't be required to explicitly create a dummy instance.

daemon/algod/api/server/v2/dryrun.go
```diff
@@ -0,0 +1,512 @@
...
+	"github.com/algorand/go-algorand/protocol"
+)
+
+// DryrunRequest object uploaded to /v2/teal/dryrun
+// It is the same as generated.DryrunRequest but Txns deserialized properly.
+// Given the Transactions and simulated ledger state upload, run TEAL scripts and return debugging information.
+// This is also used for msgp-decoding
+type DryrunRequest struct {
+	// Txns is transactions to simulate
+	Txns []transactions.SignedTxn `codec:"txns"` // not supposed to be serialized
```
tsachiherman, line 38: if you don't want this to get serialized, make it small case or msgp ignore.

daemon/algod/api/server/v2/dryrun.go
```diff
@@ -0,0 +1,512 @@
...
+
+	// Optional, useful for testing Application Call txns.
+	Accounts []generated.Account `codec:"accounts"`
```
tsachiherman, line 41: could you limit size using allocbound ?

daemon/algod/api/server/v2/dryrun.go
```diff
@@ -0,0 +1,512 @@
...
+	// ProtocolVersion specifies a specific version string to operate under, otherwise whatever the current protocol of the network this algod is running in.
+	ProtocolVersion string `codec:"protocol-version"`
+
+	// Round is available to some TEAL scripts. Defaults to the current round on the network this algod is attached to.
+	Round uint64 `codec:"round"`
+
+	// LatestTimestamp is available to some TEAL scripts. Defaults to the latest confirmed timestamp this algod is attached to.
+	LatestTimestamp int64 `codec:"latest-timestamp"`
+
+	Sources []generated.DryrunSource `codec:"sources"`
```
tsachiherman, line 54: nit: add a comment to public data member

daemon/algod/api/server/v2/dryrun.go
```diff
@@ -0,0 +1,512 @@
...
+func DryrunRequestFromGenerated(gdr *generated.DryrunRequest) (dr DryrunRequest, err error) {
+	for _, raw := range gdr.Txns {
+		// no transactions.SignedTxn in OAS, map[string]interface{} is not good enough
+		// json.RawMessage does the job
+		var txn transactions.SignedTxn
+		err = protocol.DecodeJSON(raw, &txn)
+		if err != nil {
+			return
+		}
+		dr.Txns = append(dr.Txns, txn)
```
tsachiherman, line 68: nit: could you preallocate dr.Txns

daemon/algod/api/server/v2/dryrun.go
```diff
@@ -0,0 +1,512 @@
...
+		program, err := logic.AssembleString(s.Source)
+		if err != nil {
+			return fmt.Errorf("Dryrun Source[%d]: %v", i, err)
+		}
+		switch s.FieldName {
+		case "lsig":
+			dr.Txns[s.TxnIndex].Lsig.Logic = program
+		case "approv", "clearp":
+			for ai, app := range dr.Apps {
+				if app.AppIndex == s.AppIndex {
```
tsachiherman, line 90: nit: These two loops would require O(n*m); could you optimize it to O(n+m) ?

daemon/algod/api/server/v2/dryrun.go
```diff
@@ -0,0 +1,512 @@
...
+					switch s.FieldName {
```
tsachiherman, line 91: in the case of "other", please generate an error.

daemon/algod/api/server/v2/dryrun.go
```diff
@@ -0,0 +1,512 @@
...
+	history       []generated.DryrunState
+	scratchActive []bool
+}
+
+func (ddr *dryrunDebugReceiver) updateScratch() {
+	any := false
+	maxActive := -1
+	lasti := len(ddr.history) - 1
+
+	if ddr.history[lasti].Scratch != nil {
```
tsachiherman, line 118: nit: improve readability : 
```
if ddr.history[lasti].Scratch == nil {
    return
}

// do the rest...
```

daemon/algod/api/server/v2/dryrun.go
```diff
@@ -0,0 +1,512 @@
...
+	if dl.accounts == nil {
+		dl.accounts = make(map[basics.Address]basics.BalanceRecord)
+	}
+	dl.accounts[br.Addr] = br
+	return nil
+}
+
+// PutWithCreatables is like Put, but should be used when creating or deleting an asset or application.
+func (dl *dryrunLedger) PutWithCreatables(record basics.BalanceRecord, newCreatables []basics.CreatableLocator, deletedCreatables []basics.CreatableLocator) error {
+	return nil
```
tsachiherman, line 303: is that a placeholder ?

daemon/algod/api/server/v2/generated/private/routes.go
```diff
@@ -39,9 +39,7 @@ type ServerInterfaceWrapper struct {
 // AbortCatchup converts echo context to params.
 func (w *ServerInterfaceWrapper) AbortCatchup(ctx echo.Context) error {
 
-	validQueryParams := map[string]bool{
-		"pretty": true,
-	}
+	validQueryParams := map[string]bool{}
```
tsachiherman, line 42: could you please check why we "lost" the `pretty` option ?

daemon/algod/api/server/v2/handlers.go
```diff
@@ -364,6 +319,68 @@ func (v2 *Handlers) RawTransaction(ctx echo.Context) error {
...
+	}
+	req := ctx.Request()
+	buf := new(bytes.Buffer)
+	req.Body = http.MaxBytesReader(nil, req.Body, maxTealDryrunBytes)
+	buf.ReadFrom(req.Body)
+	data := buf.Bytes()
+
+	var dr DryrunRequest
+	var gdr generated.DryrunRequest
+	err := decode(protocol.JSONHandle, data, &gdr)
```
tsachiherman, line 336: The decoding here need to be determined by Content-Type and not by trying and failing.

daemon/algod/api/server/v2/handlers.go
```diff
@@ -364,6 +319,68 @@ func (v2 *Handlers) RawTransaction(ctx echo.Context) error {
...
+	}
+
+	var response generated.DryrunResponse
+
+	var proto config.ConsensusParams
+	if dr.ProtocolVersion != "" {
+		var ok bool
+		proto, ok = config.Consensus[protocol.ConsensusVersion(dr.ProtocolVersion)]
+		if !ok {
+			return badRequest(ctx, nil, "invalid protocol version", v2.Log)
```
tsachiherman, line 366: could you change it from `"invalid protocol version"` -> `"unsupported protocol version"` ? 
The protocol could be very much valid one, but just unsupported by current binary.

also, when issuing the response, add the provided protocol version that was sent in the request.

data/basics/userBalance_test.go
```diff
@@ -142,9 +151,52 @@ func TestEncodedAccountDataSize(t *testing.T) {
...
+			KeyValue: maxLocalState,
+			Schema:   maxStateSchema,
+		}
+		ad.AppLocalStates[AppIndex(0x1234123412341234-appHolderApps)] = ls
+	}
+
 	encoded, err := ad.MarshalMsg(nil)
 	require.NoError(t, err)
-	require.Equal(t, MaxEncodedAccountDataSize, len(encoded))
+	require.GreaterOrEqual(t, MaxEncodedAccountDataSize, len(encoded))
```
tsachiherman, line 199: cool! I didn't know that require has a `GreaterOrEqual` !

data/transactions/logic/doc_test.go
```diff
@@ -80,7 +80,7 @@ func TestOpImmediateNote(t *testing.T) {
 func TestOpDocExtra(t *testing.T) {
 	xd := OpDocExtra("bnz")
 	require.NotEmpty(t, xd)
-	xd = OpDocExtra("+")
+	xd = OpDocExtra("-")
```
tsachiherman, line 83: nit: shouldn't be part of this PR.

data/pools/transactionPool.go
```diff
@@ -102,13 +110,24 @@ type txPoolVerifyCacheVal struct {
...
 
+type poolAsmResults struct {
+	ok    bool
+	blk   *ledger.ValidatedBlock
+	stats telemetryspec.AssembleBlockMetrics
+	err   error
+}
+
 // TODO I moved this number to be a constant in the module, we should consider putting it in the local config
 const expiredHistory = 10
```
tsachiherman, line 121: now would be a good time to do so.

ledger/applications.go
```diff
@@ -0,0 +1,295 @@
...
+
+import (
+	"fmt"
+
+	"github.com/algorand/go-algorand/data/basics"
+	"github.com/algorand/go-algorand/data/transactions"
+	"github.com/algorand/go-algorand/data/transactions/logic"
+)
+
+// TODO remove round from balances
```
tsachiherman, line 27: can we delete the comment, or is there any action item here ?

ledger/ledger.go
```diff
@@ -323,29 +323,63 @@ func (l *Ledger) GetLastCatchpointLabel() string {
 // GetAssetCreatorForRound looks up the asset creator given the numerical asset
 // ID. This is necessary so that we can retrieve the AssetParams from the
 // creator's balance record.
-func (l *Ledger) GetAssetCreatorForRound(rnd basics.Round, assetIdx basics.AssetIndex) (basics.Address, error) {
+func (l *Ledger) GetAssetCreatorForRound(rnd basics.Round, aidx basics.AssetIndex) (creator basics.Address, ok bool, err error) {
```
tsachiherman, line 326: what are the gains of having 3 functions instead of just one ?
GetAssetCreatorForRound
GetAppCreatorForRound
GetCreatorForRound

it looks to me that it's a lot of pointless interface dups.

data/basics/teal.go
```diff
@@ -239,6 +239,14 @@ func (tt TealType) String() string {
 	return "?"
 }
 
+// TealTypeFromString converts "b"/"u" string back to TealType
+func TealTypeFromString(t string) TealType {
+	if t == "b" {
+		return TealBytesType
+	}
+	return TealUintType
```
derbear, line 247: Shouldn't this produce an error if the string is neither `"b"` or `"u"`?

data/basics/userBalance.go
```diff
@@ -76,6 +76,21 @@ func (s Status) String() string {
...
+func UnmarshalStatus(value string) (s Status) {
+	switch value {
+	case "Offline":
+		s = Offline
+	case "Online":
+		s = Online
+	case "Not Participating":
+		s = NotParticipating
+	default:
+		s = Offline
```
derbear, line 89: Shouldn't this case produce an error?

data/basics/teal.go
algorandskiy: This is similar to argType in `logic` package - if not byte then it, so it might be ok.

data/basics/userBalance.go
algorandskiy: It mirrors `Status.String()` and the fact `Offline` = 0 is default value

derbear: I see. I think it might make sense to return an error to prevent the case where someone accidentally calls this on `"online"` and then gets `Offline`.

data/transactions/logic/eval.go
```diff
@@ -386,6 +381,11 @@ func eval(program []byte, cx *evalContext) (pass bool, err error) {
 	cx.stack = make([]stackValue, 0, 10)
 	cx.program = program
 
+	err = cx.checkRekeyAllowed()
```
derbear, line 384: Is this check still required given that you already do it below in `check`?

ledger/accountdb.go
```diff
@@ -79,6 +79,13 @@ var accountsSchema = []string{
 		strval text)`,
 }
 
+// TODO: Post applications, rename assetcreators -> creatables and rename
+// 'asset' column -> 'creatable'
+var creatablesMigration = []string{
+	`ALTER TABLE assetcreators ADD COLUMN ctype INTEGER DEFAULT 0`,
+	`CREATE INDEX assetcreators_ctype_idx ON assetcreators (asset, ctype)`,
```
derbear, line 86: Out of curiosity, how much does the index help?

ledger/eval.go
```diff
@@ -479,10 +535,15 @@ func (eval *BlockEvaluator) transactionGroup(txgroup []transactions.SignedTxnWit
 
 	cow := eval.state.child()
 
+	// Prepare TEAL contexts for any ApplicationCall transactions in the group
+	appEvaluators := eval.prepareAppEvaluators(txgroup)
```
derbear, line 539: Here's my understanding of the current mechanisms to expose ledger state to `logic.Eval`:

1. From `cow.TransactionGroup`, prepare a `appTealEvaluator` for each transaction in the group. Pass this to `Transaction.Apply` via the `transactions.StateEvaluator` interface. `Transaction.Apply` calls `InitLedger`, which initializes the `StateEvaluator` so that it correctly implements `ledger.LedgerForLogic` (specifically by creating an `appLedger`), and then calls `StateEvaluator.Check` and `StateEvaluator.Apply`, which internally pass the `appLedger`, which implements `logic.LedgerForLogic` to `logic.Eval`.

2. From the TEAL debugger, call `MakeDebugAppLedger`, which returns an `appLedger`, and pass this directly to `logic.Eval`.

Is the above correct?

derbear: If so, have you considered instead preparing an `appLedger` instead of an `appTealEvaluator` for each transaction in the group, dropping `InitLedger` (so others don't need to remember to call it), and having `appLedger` implement both `StateEvaluator` and `LedgerForLogic`?

derbear: (Here's a mediocre diagram showing the proposed changes, which are illustrated with blue pen:)

[https://drive.google.com/file/d/1z11-UdQUXsns-f1gOuXs-nBY1qJYdlfj/view?usp=sharing](url)

ledger/eval.go
```diff
@@ -102,11 +111,15 @@ func (cs *roundCowState) Get(addr basics.Address, withPendingRewards bool) (basi
 }
 
 func (cs *roundCowState) Put(record basics.BalanceRecord) error {
+	return cs.PutWithCreatables(record, nil, nil)
+}
+
+func (cs *roundCowState) PutWithCreatables(record basics.BalanceRecord, newCreatables []basics.CreatableLocator, deletedCreatables []basics.CreatableLocator) error {
```
derbear, line 117: Out of curiosity, do we ever call `PutWithCreatables` with more than one new or deleted asset/app? Could we replace these slices with pointers? Or perhaps just directly pass in a `modifiedCreatable`?

ledger/cow.go
```diff
@@ -125,18 +133,28 @@ func (cb *roundCowState) txnCounter() uint64 {
...
 	}
 
-	// Get which asset indices were created and deleted, and update state
-	assetDeltas := getChangedAssetIndices(addr, accountDelta{old: old, new: new})
-	for aidx, delta := range assetDeltas {
-		cb.mods.assets[aidx] = delta
+	for _, cl := range newCreatables {
+		cb.mods.creatables[cl.Index] = modifiedCreatable{
+			ctype:   cl.Type,
+			creator: addr,
```
derbear, line 147: Nit: I don't think this ever happens, but the redundancy here assumes `addr` and `cl.Creator` are the same.

ledger/eval.go
algomaxj: I think your understanding of the code is correct.

The reason `InitLedger` exists is that if it didn't, the ledger would need to know more about application internals. In particular, `logic.Eval` needs to know about the application ID even in the event that the application is being created by this transaction (for use in the access whitelist), and it needs to know about the application's state schemas even in the event that the application is _not_ being created (which requires looking up the creator's balance record).

In practice, I think this would mean putting logic in the ledger that 1. allocates application IDs using the transaction counter and 2. looks up existing application parameters before calling `apply` OR uses the schemas mentioned in the transaction if this is a creation txn. If we're OK doing that, then we can make that change. I think it would make the interface cleaner in exchange for moving application specific details into the ledger. What do you think?

ledger/ledger.go
```diff
@@ -323,29 +323,63 @@ func (l *Ledger) GetLastCatchpointLabel() string {
...
+func (l *Ledger) GetAppCreatorForRound(rnd basics.Round, aidx basics.AppIndex) (creator basics.Address, ok bool, err error) {
+	l.trackerMu.RLock()
+	defer l.trackerMu.RUnlock()
+	return l.accts.getCreatorForRound(rnd, basics.CreatableIndex(aidx), basics.AppCreatable)
+}
+
+// GetCreatorForRound takes a CreatableIndex and a CreatableType and tries to
+// look up a creator address, setting ok to false if the query succeeded but no
+// creator was found.
+func (l *Ledger) GetCreatorForRound(rnd basics.Round, cidx basics.CreatableIndex, ctype basics.CreatableType) (creator basics.Address, ok bool, err error) {
```
derbear, line 343: Consider adding in `error.go` a new error (maybe with the name `NoCreatorFoundError`) and returning that instead of having an extra `ok bool` return value.

ledger/eval.go
derbear: Ah, okay, makes sense. Yeah, it might be better to keep the current design, in that case.

ledger/eval.go
algomaxj: good call, I replaced the slices with pointers.

ledger/eval.go
algomaxj: nope -- removed, thanks!

config/consensus.go
algomaxj: done

daemon/algod/api/server/v1/handlers/handlers.go
algomaxj: done, thanks!

daemon/algod/api/server/v1/handlers/handlers.go
algomaxj: done!

data/transactions/logic/eval.go
algomaxj: I don't think so, but there are several sanity checks done by both `check` and `eval`, and I don't want to break that pattern here.

ledger/ledger.go
algomaxj: Yeah, you're right. Removed `GetAssetCreator` and `GetAppCreator` in favor of a generic `GetCreator` that accepts the type as an additional argument. Thanks!

ledger/accountdb.go
algomaxj: I think this is just there for the `v1` `listAssets` endpoint. Any benefit is likely to be pretty marginal -- I would be in favor of getting rid of this entirely unless @algorandskiy objects. Easy to add back if some kind of perf issue pops up.

algomaxj: The specific query is `SELECT asset, creator, ctype FROM assetcreators WHERE asset <= ? AND ctype = ? ORDER BY asset desc LIMIT ?`

ledger/cow.go
algomaxj: yeah, this is a good point; resolved so that it uses the creatable's `Creator` field like it should.

ledger/applications.go
algomaxj: done!

data/transactions/application.go
algomaxj: done!

ledger/eval.go
algomaxj: I have simplified `InitLedger` substantially by removing unnecessary whitelisting here: https://github.com/justicz/go-algorand/pull/153 . This makes `ledger/application.go` a significantly thinner wrapper around balances.

daemon/algod/api/server/v2/dryrun.go
algomaxj: I think this is here because `dryrunLedger` needs to fulfill the `Balances` interface. @algorandskiy can you confirm?

daemon/algod/api/server/v2/generated/private/routes.go
algomaxj: resolved!

daemon/algod/api/server/v2/handlers.go
algomaxj: @algorandskiy

data/basics/teal.go
algomaxj: I was able to just delete this method entirely, it's unused.

data/basics/userBalance.go
algomaxj: I agree with @derbear, added an error here.

daemon/algod/api/server/v2/handlers.go
algorandskiy: I agree in general. But I also find annoying setting content-type from `curl`, and this endpoint more for developers so this is a bit of convenience.

daemon/algod/api/server/v2/dryrun.go
algomaxj: done

daemon/algod/api/server/v2/dryrun.go
algomaxj: This is already handled by the outer switch on `s.FieldName`

daemon/algod/api/server/v2/dryrun.go
algomaxj: done

daemon/algod/api/server/v2/dryrun.go
algomaxj: @algorandskiy should this directory be included in `MSGP_GENERATE` in the `Makefile`?

algorandskiy: I'm was trying to avoid adding these api struct to msgp, so it is only serialized in `libgoal.go:MakeDryrunStateBytes` with `protocol.EncodeReflect`

daemon/algod/api/server/v2/handlers.go
algorandskiy: done

ledger/accountdb.go
algomaxj: removed

cmd/tealdbg/localLedger.go
```diff
@@ -0,0 +1,218 @@
...
+
+	if groupIndex >= len(txnGroup) {
+		return nil, appState{}, fmt.Errorf("invalid groupIndex %d exceed txn group length %d", groupIndex, len(txnGroup))
+	}
+	txn := txnGroup[groupIndex]
+
+	accounts := []basics.Address{txn.Txn.Sender}
+	for _, addr := range txn.Txn.Accounts {
+		accounts = append(accounts, addr)
+	}
```
derbear, line 53: This is a very tiny nit, but I believe it should be possible to do
```
	accounts := []basics.Address{txn.Txn.Sender}
	accounts = append(accounts, txn.Txn.Accounts...)
```
instead here.

cmd/tealdbg/localLedger.go
```diff
@@ -0,0 +1,218 @@
...
+
+	apps := []basics.AppIndex{appIdx}
+	for _, aid := range txn.Txn.ForeignApps {
+		apps = append(apps, aid)
+	}
```
derbear, line 58: (see above)

cmd/tealdbg/localLedger.go
algomaxj: done!

cmd/tealdbg/localLedger.go
algomaxj: done!

ledger/acctupdates.go
```diff
@@ -348,8 +363,8 @@ func (au *accountUpdates) listAssets(maxAssetIdx basics.AssetIndex, maxResults u
 			return res, nil
 		}
 
-		// Asset was deleted
-		if _, ok := deletedAssets[basics.AssetIndex(loc.Index)]; ok {
+		// Creatable was deleted
+		if _, ok := deletedCreatables[loc.Index]; ok {
```
derbear, line 367: Nit (not introduced by this PR): Consider having
```
		if _, ok := deletedCreatables[loc.Index]; !ok {
			res = append(res, loc)
		}
```
instead.

ledger/accountdb.go
```diff
@@ -275,7 +297,7 @@ func accountsDbInit(r db.Queryable, w db.Queryable) (*accountsDbQueries, error)
 	var err error
 	qs := &accountsDbQueries{}
 
-	qs.listAssetsStmt, err = r.Prepare("SELECT asset, creator FROM assetcreators WHERE asset <= ? ORDER BY asset desc LIMIT ?")
+	qs.listCreatablesStmt, err = r.Prepare("SELECT asset, creator, ctype FROM assetcreators WHERE asset <= ? AND ctype = ? ORDER BY asset desc LIMIT ?")
```
derbear, line 300: Nit: you can probably drop `ctype` from the returned columns (e.g., `SELECT asset, creator FROM assetcreators ...`)

ledger/acctupdates.go
```diff
@@ -949,22 +965,19 @@ func (au *accountUpdates) commitRound(offset uint64, dbRound basics.Round, lookb
 
 	// create a copy of the deltas, round totals and protos for the range we're going to flush.
 	deltas := make([]map[basics.Address]accountDelta, offset, offset)
+	creatableDeltas := make([]map[basics.CreatableIndex]modifiedCreatable, offset, offset)
 	roundTotals := make([]AccountTotals, offset+1, offset+1)
 	protos := make([]config.ConsensusParams, offset+1, offset+1)
 	copy(deltas, au.deltas[:offset])
+	copy(creatableDeltas, au.creatableDeltas[:offset])
```
derbear, line 972: Nit (not introduced by this PR): It would probably be good to add a comment explaining why deep copies of maps in `creatableDeltas` (and `deltas`) are not required here.

ledger/acctupdates.go
```diff
@@ -292,51 +295,63 @@ func (au *accountUpdates) lookup(rnd basics.Round, addr basics.Address, withRewa
...
 	sort.Slice(keys, func(i, j int) bool { return keys[i] > keys[j] })
 
-	// Check for assets that haven't been synced to disk yet.
-	var unsyncedAssets []basics.CreatableLocator
-	deletedAssets := make(map[basics.AssetIndex]bool)
-	for _, aidx := range keys {
-		delta := au.assets[aidx]
+	// Check for creatables that haven't been synced to disk yet.
+	var unsyncedCreatables []basics.CreatableLocator
+	deletedCreatables := make(map[basics.CreatableIndex]bool)
```
derbear, line 325: Nit: You might be able to get away with having the map value be a struct instead as `map[basics.CreatableIndex]struct{}`.

data/transactions/application.go
```diff
@@ -665,6 +665,12 @@ func (ac *ApplicationCallTxnFields) apply(header Header, balances Balances, spec
 		return err
 	}
 
+	// Ensure that the only operation we can do is ClearState if the application
+	// does not exist
+	if !exists && ac.OnCompletion != ClearStateOC {
+		return fmt.Errorf("only clearing out is supported for applications that do not exist")
+	}
+
```
derbear, line 673: I wonder if you can just completely handle a deleted application at this point and instantly return (this might involve moving some chunks of code around).

Issue #915, opened 2020-03-17
Catchup requests fails due to locked database


tsachiherman:
## Issue

There are several variation of this issue, but the most common is the following logged message:
```
**{"file":"ledgerService.go","function":"github.com/algorand/go-algorand/rpcs.(*LedgerService).ServeHTTP","level":"warning","line":174,"msg":"ServeHTTP : failed to retrieve block 3735696 database is locked","time":"2020-03-15T16:21:05.316959Z"}
```

In all cases, the sqlite is returning `sqlite3.ErrLocked`, which we log on.

We do have support for retrying ( which seems to resolve the very same issue for `accountdb.go` ), however, this might not be the best approach.
Our existing database is using write-ahead-log ( WAL ), so that it should support multiple readers. clearly, we don't see this here, and we might want to try to figure out why instead of patching this will retrying mechanism ( which might hide the underlying issue ).

PR #918, opened 2020-03-18
Add retry for creating a DB connection

tsachiherman:
## Summary

Ledger service was generating the following error:
```
{"file":"ledgerService.go","function":"github.com/algorand/go-algorand/rpcs.(*LedgerService).ServeHTTP","level":"warning","line":174,"msg":"ServeHTTP : failed to retrieve block 3735696 database is locked","time":"2020-03-15T16:21:05.316959Z"}
```

The root cause was the sqlite driver returning a `database is locked` when a new connection attempt was made. Existing code was already testing extensively for temporary locked database conditions on transaction creation/execution, but not on connection creation.

resolves issue https://github.com/algorand/go-algorand/issues/915

### DIFF
--- a/util/db/dbutil.go
+++ b/util/db/dbutil.go
@@ -147,12 +147,12 @@ func LoggedRetry(fn func() error, log logging.Logger) (err error) {
 	for i := 0; ; i++ {
 		if i > 0 {
 			if i < infoTxRetries {
-				log.Infof("db.LoggedRetry: %d connection retries (last err: %v)", i, err)
+				log.Infof("db.LoggedRetry: %d retries (last err: %v)", i, err)
 			} else if i >= 1000 {
-				log.Errorf("db.LoggedRetry: %d connection retries (last err: %v)", i, err)
+				log.Errorf("db.LoggedRetry: %d retries (last err: %v)", i, err)
 				return
 			} else if i%warnTxRetriesInterval == 0 {
-				log.Warnf("db.LoggedRetry: %d connection retries (last err: %v)", i, err)
+				log.Warnf("db.LoggedRetry: %d retries (last err: %v)", i, err)
 			}
 		}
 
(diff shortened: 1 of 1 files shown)


util/db/dbutil.go
```diff
@@ -209,21 +212,42 @@ func (db *Accessor) Atomic(fn idemFn, extras ...interface{}) (err error) {
 	}
 
 	var tx *sql.Tx
-	ctx := context.Background()
 	var conn *sql.Conn
-	conn, err = db.Handle.Conn(ctx)
+	ctx := context.Background()
+
+	for i := 0; ; i++ {
```
pzbitskiy: use LoggedRetry?

util/db/dbutil.go
```diff
@@ -145,12 +145,15 @@ func (db *Accessor) Close() {
 // Sends warnings and errors to log.
 func LoggedRetry(fn func() error, log logging.Logger) (err error) {
 	for i := 0; ; i++ {
-		if i > 0 && i%warnTxRetries == 0 {
```
pzbitskiy: nit: reduce amount of branches with something like this:
```
conn, err = db.Handle.Conn(ctx)
for i := 1; dbretry(err); i++ {
  // logging
  conn, err = db.Handle.Conn(ctx)
}
```

justicz:
Just confirming, were you able to reproduce this locally?

util/db/dbutil.go
tsachiherman: good idea.

util/db/dbutil.go
tsachiherman: I want to have a different info/warn/err message to indicate that it's a connection error. If this becomes persistent, we need to have that information.

tsachiherman:
> Just confirming, were you able to reproduce this locally?

regretfully, no. And it's not due to lack of trying. I suspect it's an external issue, but the information that we currently have isn't enough to back this up. The added info/warn/error on the connection attempt should help us confirm that its a permanent issue caused by a misconfigured installation. ( i.e. maybe somehow the algod lost the ability to make changes to the database ? )

PR #923, opened 2020-03-21
DO NOT MERGE!! Prepare scripts to support multi-channel, multi-version distributions

btoll:
I'm wondering what different use cases we should support.

Currently, if you have `algorand` installed and want to install `algorand-beta` (or `algorand-nightly`), you'll get an error because it needs to overwrite `genesis.json` and two other files (of course, this isn't an issue if no algorand-* package has been installed yet).

This is easily solved, but does it make sense to, and are there technological reasons/limitations to not want to do this?


tsachiherman:
Next time, could you open this as a draft pr ? The diffrrence is that draft
cannot be merged unless being converted to regular pr first.

PR #924, opened 2020-03-22
WIP: Refactor travis job to use mule

bricerisingalgorand:
I'm making this PR to have travis kick off jobs to run jobs using mule to execute our builds/tests


.travis.yml
```diff
@@ -1,24 +1,19 @@
 dist: bionic
-go:
-  - "1.12"
-go_import_path: github.com/algorand/go-algorand
-language: go
 sudo: required
+language: minimal
 
 # Don't build tags
 if: tag IS blank
```
winder: If we're getting rid of the `deploy` stage we can get rid of this. The old deploy script had a dependency on the branch name (`rel/stable`, `rel/nightly`, etc) which is why we couldn't build tags.

.travis.yml
```diff
@@ -28,122 +23,131 @@ jobs:
...
+      name: MacOS AMD64 Unit Test
+      go:
+        - "1.12"
+      go_import_path: github.com/algorand/go-algorand
+      language: go
+      before_install:
+        - brew update
+        - brew upgrade pyenv
+        - pyenv install 3.7.6
+        - pyenv local  3.7.6
```
winder: Why do you need the go stuff, Isn't everything happening in a container now?

.travis.yml
```diff
@@ -28,122 +23,131 @@ jobs:
...
```
winder: One trick we used to avoid specifying all of this for different architectures in travis, was having a portable `configure_dev.sh` script. If you wrap `mule build-local-shorttest` in a script. If you add a `configure_mule.sh` script, and wrap the mule calls in a script which calls `configure_mule.sh` you can reduce some of this duplicate setup code.

.travis.yml
bricerisingalgorand: This may still be required because I'm actually just calling the script that the deploy stage used as a make target. Later on we're planning to write a mule task that handles the uploads on it's own, so maybe then will be a better time to remove this

```
archive:
	PATH=$(SRCPATH)/tmp/node_pkgs/$(OS_TYPE)/$(ARCH)/dev/$(OS_TYPE)-$(ARCH)/bin:$${PATH} \
	scripts/upload_version.sh $(BUILDCHANNEL) $(SRCPATH)/tmp/node_pkgs/$(OS_TYPE)/$(ARCH) $(S3_RELEASE_BUCKET)
```

.travis.yml
bricerisingalgorand: We're having to execute the OSX builds locally since there aren't darwin containers unfortunately

.travis.yml
bricerisingalgorand: Oh that could work. Since python is required for our integration tests anyway, I could probably go ahead and have the configure_dev script guarantee the correct version of python is installed for mule. Then it'll be a one liner in the travis to say pip install mulecli

PR #925, opened 2020-03-23
Prepare shell scripts to support multi-versioned, multi-channel APT repo distributions

btoll:
This supports installation of multiple versions of the same package for a distribution, i.e., `algorand`, `algorand-beta` or `algorand-nightly`.

Currently, multiple cross-distribution installations are not supported, i.e., having both `algorand` and `algorand-beta` installed at the same time.


algobolson:
would a separate data directory make sense? install data to /var/lib/algorand-beta ?
Fully supporting that would imply doing binaries like `algod-stable` and `algod-beta` with a symlink from `algod` to whichever was most recently installed. Or just document that installing both versions on one system will give weird results.

installer/debian/control
```diff
@@ -1,4 +1,4 @@
-Package: algorand
+Package: @PKG_NAME@
```
algobolson: this seems confusing, why is PKG_NAME a standin for 'stable' in 51algorand-updates but replacing 'algorand' here? Is it that 'stable' should have been 'algorand-stable' ?

btoll: Yeah, that's not right, I need to fix this.

btoll:
> would a separate data directory make sense? install data to /var/lib/algorand-beta ?
> Fully supporting that would imply doing binaries like `algod-stable` and `algod-beta` with a symlink from `algod` to whichever was most recently installed. Or just document that installing both versions on one system will give weird results.

Tsachi and I talked about this today, and we thought the best way moving forward, at least for now, is to inform the user that installing by writing to the shell that different packages aren't supported (i.e., you have `algorand-beta` installed and now want to install `algorand`).

installer/debian/preinst
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+
+PKG_NAME=@PKG_NAME@
+VERSION="$2"
+
+if test "$1" = install &&
+    dpkg-query --list 'algorand*' > /dev/null
+then
+    if PKG_INFO=$(dpkg-query --show --showformat='${Package} ${Status}\n' 'algorand*' | grep "install ok installed")
```
onetechnical: I think this will not catch partial installs / install failures, but that's okay for this iteration.

scripts/build_deb.sh
```diff
@@ -3,13 +3,13 @@
...
-# Syntax:   build_deb.sh <arch> <output directory>
+# Syntax:   build_deb.sh <arch> <output directory> <channel>
 #
 # Examples: scripts/build_deb.sh amd64
 
 set -e
-if [ ! "$#" -eq 2 ]; then
-    echo "Syntax: build_deb.sh <arch> <output directory>"
+if [ "$#" -lt 2 ]; then
+    echo "Syntax: build_deb.sh <arch> <output directory> <channel>"
```
bricerisingalgorand: you may want to consider using getopts here instead of this approach to do validation on each argument. Not a blocker for the PR, but approaching bash script parameters this way helps new users down the road understand how to use the script faster:


```bash
HELP="Usage: $0 [-v] [-u]
Script for installing the mule cli.
Requires:
    * pip
    * python 3
    * git (And access to algorand/go-algorand-ci repository on github)
Options:
    -v        Version of mule cli (default is '${MULE_VERSION}')
    -u        Authenticate to github (algorand/go-algorand-ci repository) with user credentials (default is over ssh)
"

while getopts ":v:uh" opt; do
  case ${opt} in
    v ) MULE_VERSION=$OPTARG
      ;;
    u ) USER_AUTH="True"
      ;;
    h ) echo "${HELP}"
        exit 0
      ;;
    \? ) echo "${HELP}"
        exit 2
      ;;
  esac
done
```

scripts/build_deb.sh
```diff
@@ -20,8 +20,10 @@ if [ "$EUID" != "0" ]; then
 fi
 
 OS=linux
-ARCH=$1
+ARCH="$1"
 OUTDIR="$2"
+CHANNEL=${CHANNEL:-$3}
+PKG_NAME=$(./scripts/compute_package_name.sh "${CHANNEL:-stable}")
```
bricerisingalgorand: Do we want to default to stable? It could be unfortunate to accidentally package a beta release as stable

scripts/build_deb.sh
btoll: I've used a similar construction for other scripts, but I didn't for this because other pre-existing scripts are using this and I wanted to leave as light a footprint as I could.

Also, I feel that depending on how the script is used it's better to use positional arguments, especially for automation.

scripts/build_deb.sh
btoll: It *shouldn't* ever have an empty value by the time line 26 is parsed.  Is this 12 of one half dozen of the other?

scripts/build_deb.sh
```diff
@@ -103,21 +105,24 @@ done
 unattended_upgrades_files=("51algorand-upgrades")
 mkdir -p "${PKG_ROOT}/etc/apt/apt.conf.d"
 for f in "${unattended_upgrades_files[@]}"; do
-    cp "installer/${f}" "${PKG_ROOT}/etc/apt/apt.conf.d"
+    < "installer/${f}" \
+      sed -e "s,@CHANNEL@,${CHANNEL}," \
```
bricerisingalgorand: This is a way I've enjoyed expanding bash variables in templates in the past that you may like better. With this approach you can use bash style syntax to include values.

```
eval "cat <<EOF
$(<file.tmpl.txt)
EOF
" > file.txt
```

scripts/compute_package_name.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+
+# ./compute_package_name <CHANNEL>
+
```
bricerisingalgorand: Should we have another name for testnet packages? Or are we using the same packages and suggesting users swap their genesis files?

winder: testnet/mainnet both run the same binaries (the ones from rel/stable)

scripts/build_deb.sh
btoll: I'm not sure how that would help here.

For context, this block was flagged for an unnecessary use of `cat`, which has since been removed, replaced by the statement that you see here.  I personally like that rule (UUOC) and  would also like to avoid `eval` unless necessary.

scripts/build_deb.sh
egieseke: I agree that including parameter names helps clarify and ensure proper use.

scripts/compute_package_name.sh
btoll: What names do you have in mind?  I'm afraid I don't understand your second question.

installer/debian/preinst
btoll: If the install fails or is aborted, I think that will be ok as this is only flagged if a different packages is attempted to be installed.

scripts/build_deb.sh
btoll: I should add that the intention of this PR was just to implement the changes needed to support the repos.  A larger change such as the one you're suggesting could be its own PR (in other words, I'm not against that, just don't want to do it here).

scripts/build_deb.sh
bricerisingalgorand: I tried googling that idiom and I'm still confused. But I like the confidence so I'm ok with this default

scripts/build_deb.sh
bricerisingalgorand: Hmm there may be a way to do this without cat. The reason this can be more powerful is that you don't have to do a sed individually for env vars you want to evaluate. A case could be made that in some cases defining fields you want to evaluate can be better. I'm usually more inclined to go with generalized templating strategies, but it's more of a preference thing.

scripts/build_deb.sh
btoll: This is not intended to be a cli tool.

scripts/compute_package_name.sh
bricerisingalgorand: They should be the same binaries, but should they be the same packages? Users may want the option to install a testnet package that comes with the correct genesis file already set up in their data dir

scripts/build_deb.sh
btoll: Yes, it's a valid point, but I still think that that would be a larger change not really related to the intent of this PR and should be its own PR.

scripts/build_deb.sh
btoll: https://idioms.thefreedictionary.com/six+of+one+and+half+a+dozen+of+the+other

I can change the default if you want.

scripts/compute_package_name.sh
btoll: I've talked with Tsachi about this and the feeling is that a user can do that with the tarballs, but different expectations come with downloading from a repo.

For instance, a more experienced user will download the tarballs and have the binaries point to different data directories, but a different kind of user will want to download from a repository and just have it work.

I'd also like to keep the existing implementation as is for now, and consider reviewing the outcome once the entire build system is complete.

PR #934, opened 2020-03-26
Re-keying: backend

algoradam:
Add backend support for rekeying: allowing transactions to change the spending key of an existing account. Such transactions will only be allowed once the protocol supports them -- this PR can be merged now without changing any behavior, and a future protocol upgrade can flip the switch to enable the feature.
This PR is the backend logic -- the ledger / block evaluator. Frontend work (goal / kmd support) will be in a subsequent PR.

### DIFF
--- a/data/transactions/logic/eval.go
+++ b/data/transactions/logic/eval.go
@@ -209,10 +209,15 @@ func Eval(program []byte, params EvalParams) (pass bool, err error) {
 		cx.err = fmt.Errorf("program version %d greater than protocol supported version %d", version, params.Proto.LogicSigVersion)
 		return false, cx.err
 	}
+	if (params.Txn.Txn.RekeyTo != basics.Address{}) { // Currently no TEAL version knows about the RekeyTo field, but this may change in a future TEAL version
+		cx.err = fmt.Errorf("program version %d doesn't allow transactions with nonzero RekeyTo field", version)
+		return false, cx.err
+	}
 	// TODO: if EvalMaxVersion > version, ensure that inaccessible
 	// fields as of the program's version are zero or other
 	// default value so that no one is hiding unexpected
 	// operations from an old program.
+
 	cx.version = version
 	cx.pc = vlen
 	cx.EvalParams = params
--- a/data/transactions/transaction.go
+++ b/data/transactions/transaction.go
@@ -93,6 +93,12 @@ type Header struct {
 	// the LastValid round passes.  While this transaction possesses the
 	// lease, no other transaction specifying this lease can be confirmed.
 	Lease [32]byte `codec:"lx"`
+
+	// RekeyTo, if nonzero, sets the sender's SpendingKey to the given address
+	// If the RekeyTo address is the sender's actual address, the SpendingKey is set to zero
+	// This allows "re-keying" a long-lived account -- rotating the signing key, changing
+	// membership of a multisig account, etc.
+	RekeyTo basics.Address `codec:"rekey"`
 }
 
 // Transaction describes a transaction that can appear in a block.
@@ -332,6 +338,7 @@ func (tx Transaction) WellFormed(spec SpecialAddresses, proto config.ConsensusPa
 	if !proto.SupportTxGroups && (tx.Group != crypto.Digest{}) {
 		return fmt.Errorf("transaction has group but groups not yet enabled")
 	}
+	// TODO(rekey) check that rekeyto is empty if rekeying is not supported
 	return nil
 }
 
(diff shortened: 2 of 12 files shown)


data/transactions/verify/txn.go
```diff
@@ -125,6 +128,9 @@ func Txn(s *transactions.SignedTxn, ctx Context) error {
 	if s.Txn.Src() == zeroAddress {
 		return errors.New("empty address")
 	}
+	if !proto.SupportRekeying && (s.AuthAddr != basics.Address{}) {
```
justicz: Hm, I guess that you're doing this here since `WellFormed` is on a `transactions.Transaction` and not a `transactions.SignedTxn`, but quickly grepping through our code, it looks like we always a `SignedTxn` around when calling `WellFormed` (except maybe some tests?)... this doesn't have to happen here, but maybe we should update that in a separate PR since this really feels like a `WellFormed` check.

data/transactions/transaction.go
```diff
@@ -93,6 +93,12 @@ type Header struct {
 	// the LastValid round passes.  While this transaction possesses the
 	// lease, no other transaction specifying this lease can be confirmed.
 	Lease [32]byte `codec:"lx"`
+
+	// RekeyTo, if nonzero, sets the sender's SpendingKey to the given address
+	// If the RekeyTo address is the sender's actual address, the SpendingKey is set to zero
+	// This allows "re-keying" a long-lived account -- rotating the signing key, changing
+	// membership of a multisig account, etc.
+	RekeyTo basics.Address `codec:"rekey"`
```
zeldovich: Overall, this PR looks great!

Minor suggestion/comment: rekeying is a major operation, but this allows any transaction to also rekey.  I wonder if it's worth defining a new txtype for rekey, or at least tacking it onto keyreg txtypes (so that, for example, a payment can never rekey)?

derbear: Yeah, one alternative design was to have a separate rekeying transaction, and then have transactions that wanted to do X and rekey atomically bundled in a group transaction.

The assumption here was that a small proportion of high-security accounts will want to rekey on every transaction and that for these accounts, it would be marginally more convenient to have everything done in one transaction. (A TEAL program can enforce that a particular transaction doesn't rekey by specifying that the `RekeyTo` field is zero.)

zeldovich: Ah, yeah, that seems like an interesting use case.  (Of course doable with tx groups too, but not obvious which plan is better overall.)

data/transactions/logic/eval.go
```diff
@@ -209,10 +209,15 @@ func Eval(program []byte, params EvalParams) (pass bool, err error) {
 		cx.err = fmt.Errorf("program version %d greater than protocol supported version %d", version, params.Proto.LogicSigVersion)
 		return false, cx.err
 	}
+	if (params.Txn.Txn.RekeyTo != basics.Address{}) { // Currently no TEAL version knows about the RekeyTo field, but this may change in a future TEAL version
```
winder, line 212: nit: can use `!params.Txn.Txn.RekeyTo.IsZero()`

winder:
LGTM, this should be pretty straight forward to support in the SDKs

algobolson:
LGTM

data/transactions/verify/txn.go
```diff
@@ -100,6 +100,9 @@ func TxnPool(s *transactions.SignedTxn, ctx Context, verificationPool execpool.B
...
```
derbear, line 103: Since `WellFormed` now does this check, and we call it a few lines above, can we now drop this check from here?

derbear: (I think we can take out the `if s.Txn.Src` check above, for that matter.)

data/transactions/verify/txn.go
derbear: (See comment above)

ledger/eval.go
```diff
@@ -542,6 +542,20 @@ func (eval *BlockEvaluator) transaction(txn transactions.SignedTxn, ad transacti
...
+		// Does the address that authorized the transaction actually match whatever address the sender has rekeyed to?
+		// i.e., the sig/lsig/msig was checked against the txn.Authorizer() address, but does this match the SpendingKey in the sender's balance record?
+		acctdata, err := cow.lookup(txn.Txn.Sender)
+		if err != nil {
+			return err
+		}
+		correctAuthorizer := acctdata.SpendingKey
+		if (correctAuthorizer == basics.Address{}) {
+			correctAuthorizer = txn.Txn.Sender
+		}
```
derbear, line 555: I'm not sure if this is a good idea, but consider adding something like `Authorizer()` to not just `SignedTxn` but also `AccountData`.

data/transactions/verify/txn.go
algoradam: WellFormed checks the `RekeyTo` but it doesn't (yet) check the `AuthAddr` -- in this PR `WellFormed` still takes in a `Transaction` rather than a `SignedTxn` so it can't see the `AuthAddr` field. (I'm going to change that in a separate PR as Max suggested earlier, as it'll involve changing a bunch of the transaction tests.)

derbear: Ah, right, I missed that. Looks good then.

PR #936, opened 2020-03-30
Add support for building packages using Mule

btoll:
This is a good start, but there are several things that still need to be done.
1.  It would be helpful if `mule` would support multiple bind mounts.  I worked around this by copying `$GOPATH/bin` to `go-algorand/go/bin`.
1.  The `build_packages.sh` script is also creating the deb package, which isn't great.  This means that it is called for both the `build` and `package-deb` tasks.  It would be nice to break this functionality out.
1.  My brain is furzed.  I'll finish this in the morning.

Please keep in mind that this is still a WIP.  And I would like to do the future work that I mentioned above.


Makefile.mule
```diff
@@ -0,0 +1,31 @@
+# This file is imported into go-algorand/Makefile.
+#
+.PHONY: ci-clean ci-deps ci-setup ci-build package package-deb package-rpm package-$(OS_TYPE)-$(ARCH) package-deb-$(OS_TYPE)-$(ARCH) package-rpm-$(OS_TYPE)-$(ARCH)
```
btoll, line 3: Since this Makefile is being included in the main Makefile, will defining `.PHONY` here collide with the one in the main Makefile?

docker/build/cicd.centos.Dockerfile
```diff
@@ -27,3 +27,4 @@ RUN rm -rf $GOPATH/src/github.com/algorand/go-algorand && \
     mkdir -p $GOPATH/src/github.com/algorand/go-algorand
 RUN echo "vm.max_map_count = 262144" >> /etc/sysctl.conf
 CMD ["/bin/bash"]
+
```
btoll, line 30: I like to end the file with a blank line since it makes the output of utils like `cat` more readable.

package.yaml
```diff
@@ -0,0 +1,55 @@
...
+      workDir: /projects/go-algorand
+    target: package-deb
+
+jobs:
+  package:
+    configs:
+      arch: amd64
+      docker:
+        env:
+          - FOO=foo
```
btoll, line 49: Just a placeholder, for now.

scripts/release/mule/package/linux/amd64/rpm/package.sh
```diff
@@ -0,0 +1,32 @@
...
+export DEFAULT_RELEASE_NETWORK
+
+RPMTMP=$(mktemp -d 2>/dev/null || mktemp -d -t "rpmtmp")
+trap 'rm -rf $RPMTMP' 0
+
+TEMPDIR=$(mktemp -d)
+trap 'rm -rf $TEMPDIR' 0
+< "$REPO_DIR/installer/rpm/algorand.spec" \
+    sed -e "s,@PKG_NAME@,${PKG_NAME:-algorand}," \
+        -e "s,@VER@,${FULLVERSION:-6.6.6}," \
```
btoll, line 22: These default values are placeholders.

scripts/release/mule/package/linux/amd64/deb/package.sh
```diff
@@ -0,0 +1,26 @@
...
+export NO_BUILD=1
+export REPO_DIR=/projects/go-algorand
+export GOPATH="$REPO_DIR/go"
+export PATH="$GOPATH:/usr/local/go/bin:$PATH"
+# TODO: Don't hardcode this path!
+# Will change to something like:
+# "$REPO_DIR/tmp/node_pkgs/$OS_TYPE/$ARCH/pkg"
+export PKG_ROOT="$REPO_DIR/tmp/node_pkgs/linux/amd64/deb"
+export VARIATIONS=linux/amd64
+"$REPO_DIR/scripts/build_packages.sh" "$VARIATIONS"
```
bricerisingalgorand, line 19: Can we add an option to pass in a PKG_ROOT with already built binaries instead of having this script always build the tarballs? Also it looks like the docker command is commented out

scripts/release/mule/package/linux/amd64/rpm/package.sh
bricerisingalgorand: 7.7.7 is much luckier

scripts/release/mule/package/linux/amd64/deb/package.sh
btoll: So, that *shouldn't* be rebuilding the tarballs again since the NO_BUILD flag is set, but I haven't verified that.  And yes, that could be an option.

I commented out the docker build b/c it should probably be built by another `mule` task.

btoll:
I'm going to close this and open up a new PR.

PR #937, opened 2020-03-30
WIP: e2e_basic_start_stop debugging

algobolson:
more logging for travis debugging


algobolson:
(using the normal pull-request build to debug this, doesn't necessarily need to be approved or merged yet)

algobolson:
It turns out waiting just a little bit longer makes it work. iirc there's somewhere in algod node code that waits for something for up to 5 seconds while shutting down gracefully. Maybe we're hitting that, but `algod` does shut down within 5 seconds if we wait that long.

PR #939, opened 2020-03-30
Fix timing bug in TestAgreementTime

tsachiherman:
## Summary

The TestAgreementTime was testing the timing measurement performed by `blockstats::onBlock`. To do so, it was calling it twice in a row, with a predetermined delay.
Then, it would compare the resulting event time measurement by comparing it to the expected value ( i.e. the delay ).

While the idea here is correct, the implementation wasn't that great. On slower machines, the `time.Sleep(XXX)` function might take longer than the desired time. As a result, the time measured by `blockstats::onBlock` could be skewed upward. To work around this limitation, the test was modified to use dynamic delay with self-validation for expected errors. Note that the change here would work correctly for this type of tests, but might not work well on others.

### DIFF
--- a/cmd/algoh/blockstats_test.go
+++ b/cmd/algoh/blockstats_test.go
@@ -57,17 +57,36 @@ func TestConsecutiveBlocks(t *testing.T) {
 }
 
 func TestAgreementTime(t *testing.T) {
-	sender := MockEventSender{}
-	bs := blockstats{log: &sender}
+	sleepTime := 50 * time.Millisecond
+	testAttempts := 0
+	const maxTestAttempts = 10
+	for {
+		sleepTimeHighWatermark := time.Duration(int64(sleepTime) * 105 / 100)
 
-	bs.onBlock(v1.Block{Round: 300})
-	time.Sleep(500 * time.Millisecond)
-	bs.onBlock(v1.Block{Round: 301})
+		sender := MockEventSender{}
+		bs := blockstats{log: &sender}
 
-	require.Equal(t, 1, len(sender.events))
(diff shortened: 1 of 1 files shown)


cmd/algoh/blockstats_test.go
```diff
@@ -57,17 +57,36 @@ func TestConsecutiveBlocks(t *testing.T) {
...
+		start := time.Now()
+		bs.onBlock(v1.Block{Round: 300})
+		time.Sleep(sleepTime)
+		bs.onBlock(v1.Block{Round: 301})
+		end := time.Now()
+
+		require.Equal(t, 1, len(sender.events))
+		details := sender.events[0].details.(telemetryspec.BlockStatsEventDetails)
+
+		// Make sure the duration is close to 500ms
```
algobolson: not actually 500ms but `sleepTime` which is now 50ms ?

tsachiherman: exactly. The goal of this test is to see that onBlock measure the time difference.
There is no reason to wait 500ms for that if the machine is reasonably fast. On the other hand, on slow machine, this is unlikely to be enough.
So, I changed the timeout to be +5% over the sleep time, and tested the worst case scenario by wrapping the two called between start/end timers.
That allows faster machines to test that quickly, and would take longer on slower machines.

algobolson: yeah, I was just commenting that cleaning up the comment to be more accurate would be nice

tsachiherman: oh, I completely missing that.. I'll update the comment to reflect what's being done.

thanks!

PR #940, opened 2020-03-31
Only compute TxID once during validated eval

justicz:
CPU pprof graphs here: [txid-graphs.tar.gz](https://github.com/algorand/go-algorand/files/4410682/txid-graphs.tar.gz)

The benchmark I used may be [found on the maxj/applications branch here](https://github.com/justicz/go-algorand/blob/maxj/applications/ledger/ledger_perf_test.go).

```
maxj@algog ~/P/a/g/ledger> go test -run=XXX -bench='BenchmarkPay.*5000' -benchtime=25x -cpuprofile perf-no-opt.prof
goos: linux
goarch: amd64
BenchmarkPaymentEvalPerf5000-8                25         240774911 ns/op
...
maxj@algog ~/P/a/g/ledger> git checkout maxj/txid_calc_eval_once
Switched to branch 'maxj/txid_calc_eval_once'
maxj@algog ~/P/a/g/ledger> go test -run=XXX -bench='BenchmarkPay.*5000' -benchtime=25x -cpuprofile perf-wi-opt.prof
goos: linux
goarch: amd64
BenchmarkPaymentEvalPerf5000-8                25         231878331 ns/op
...
```

Currently, about 13.0% of time spent in `(*BlockEvaluator).transaction` is spent in `(*roundCowState).addTx`, which keeps track of the transaction for duplicate detection.

When we're verifying a block, we have already computed the `TxID` when we were checking if it was a duplicate. If we pass that value down to `(*roundCowState).addTx`, then we only spend 6.54% of our time in `(*BlockEvaluator).transaction` there.

### DIFF
--- a/ledger/cow.go
+++ b/ledger/cow.go
@@ -140,8 +140,12 @@ func (cb *roundCowState) put(addr basics.Address, old basics.AccountData, new ba
 	}
 }
 
-func (cb *roundCowState) addTx(txn transactions.Transaction) {
-	cb.mods.Txids[txn.ID()] = txn.LastValid
+func (cb *roundCowState) addTx(txn transactions.Transaction, cachedTxid transactions.Txid) {
+	if cachedTxid == (transactions.Txid{}) {
+		cachedTxid = txn.ID()
+	}
+
+	cb.mods.Txids[cachedTxid] = txn.LastValid
 	cb.mods.txleases[txlease{sender: txn.Sender, lease: txn.Lease}] = txn.LastValid
 }
 
--- a/ledger/cow.go
+++ b/ledger/cow.go
@@ -140,12 +140,8 @@ func (cb *roundCowState) put(addr basics.Address, old basics.AccountData, new ba
 	}
 }
 
-func (cb *roundCowState) addTx(txn transactions.Transaction, cachedTxid transactions.Txid) {
-	if cachedTxid == (transactions.Txid{}) {
-		cachedTxid = txn.ID()
-	}
-
-	cb.mods.Txids[cachedTxid] = txn.LastValid
+func (cb *roundCowState) addTx(txn transactions.Transaction, txid transactions.Txid) {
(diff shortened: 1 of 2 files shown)


ledger/cow.go
```diff
@@ -140,8 +140,12 @@ func (cb *roundCowState) put(addr basics.Address, old basics.AccountData, new ba
 	}
 }
 
-func (cb *roundCowState) addTx(txn transactions.Transaction) {
-	cb.mods.Txids[txn.ID()] = txn.LastValid
+func (cb *roundCowState) addTx(txn transactions.Transaction, cachedTxid transactions.Txid) {
+	if cachedTxid == (transactions.Txid{}) {
```
tsachiherman, line 144: nit: you could eliminate this conditional check if you were to add the `cachedTxid = txn.ID()` on the `if !eval.validate` condition in the BlockEvaluator::transaction.

ledger/eval.go
```diff
@@ -520,8 +521,8 @@ func (eval *BlockEvaluator) transaction(txn transactions.SignedTxn, ad transacti
 		}
 
 		// Transaction already in the ledger?
-		txid := txn.ID()
-		dup, err := cow.isDup(txn.Txn.First(), txn.Txn.Last(), txid, txlease{sender: txn.Txn.Sender, lease: txn.Txn.Lease})
+		cachedTxid = txn.ID()
```
tsachiherman, line 524: nit: use the `cachedTxid` in `return TransactionInLedgerError{txn.ID()}`

ledger/cow.go
pzbitskiy: +1, no conditions => better chance for inlining

ledger/eval.go
justicz: done!

ledger/cow.go
justicz: good suggestion, done.

PR #941, opened 2020-04-01
Eric/fix beta release network

egieseke:
## Summary

Fix for "Betanet releases should point to betanet rather then devnet" https://github.com/algorand/go-algorand-internal/issues/294

Updated check_clean_enlistment.sh to ignore modifications to buildnumber.dat and removed the check for TRAVIS_BRANCH.  Also removed the use of check_clean_entitlement.sh script except for checkout_branch.sh, so that the channel and network names and deadlock default are always computed based on the branch. 

## Test Plan

Tested check_clean_enlistment.sh to verify that it ignores modifications to buildnumber.dat. 

Created ticket to check for correct default network after next beta release:  https://github.com/algorand/go-algorand-internal/issues/346


egieseke:
@btoll  check_clean_enlistment.sh script is there to prevent overwitting changes in a working directory, this is from the file header: # Usage:    Use before a forced checkout to ensure we don't clobber or lose any changes.  So it is still useful for that.

btoll:
That comment was relevant when David was doing multiple builds on his laptop.  Since we're no longer doing that, the only thing the script can do is cause problems.

Branching depending upon env vars set by external tools is a hack that we should avoid.

egieseke:
@tsachiherman , if enlistment check fails, which it was on Jenkins, then the network defaults to "devnet", per the script compute_branch_network.sh   I just removed the enlistment check from compute_branch_network.sh and compute_branch_channel.sh, so that the correct channel and network will be computed regardless of the state of enlistment, per Ben's suggestion.

btoll:
This is a good start, but we should remove the file entirely.

egieseke:
The enlistment script is still referenced in 2 other places.  
in checkout_branch.sh to protect the developer from losing local changes.  
Also, referenced in compute_branch_deadlock_default.sh.  
So we can't remove the enlistment script unless we remove those references too.

btoll:
I think the file and all the references should be removed.  They are relics from several iterations ago, before we started the automation.

egieseke:
It's not clean due to the version update.

tsachiherman:
> It's not clean due to the version update.

Then maybe the version update need to be excluded from this test by setting it in the .gitignore ?

tsachiherman:
As a sidenote, the proper location for invoking the `check_clean_enlistment.sh` is at the very beginning of the build ( and only once ).

btoll:
I think the questions to ask are:
1.  Are there legitimate reasons why the workspace could be dirty?
1.  Are there legitimate reasons to fail a build or change the network name due to a dirty workspace?

For #1, yes, there are legitimate reasons.  I know I had an issue before with the check enlistment script, although I forget the use case now since it's been several months.  But, it was a false positive, and it was annoying that the check enlistment script was incorrectly flagging the build.

We could add that name to `.gitignore`, but what if we change the file name or introduce another process that also dirties the workspace?  Then, this old bugaboo would bite us again.

egieseke:
Updated check_clean_enlistment.sh to ignore modifications to buildnumber.dat.

egieseke:
@tsachiherman , Removed enlistment check from compute_branch_deadlock_default.sh.

tsachiherman:
> compute_branch_deadlock_default.sh

Yes, making the enlistment checking there doesn't make any sense to me.

tsachiherman:
looks good, could you please update the issue title to correctly reflect your changes ?
Also, since this issue was originated in betanet, you'll need to create another ticket to validate the fix once we release the next betanet.

btoll:
I don't think it's a good idea to hardcode that name (or any other) in the conditional test.

What is the justification for keeping this script?

tsachiherman:
> I don't think it's a good idea to hardcode that name (or any other) in the conditional test.
> 
> What is the justification for keeping this script?

I think that the usage in deploy_version.sh is legit.
i.e. deploy_version.sh -> checkout_branch.sh -> check_clean_enlistment.sh

as it test for changes and abort if that's not the case. The is a ENV_VAR SkipCleanCheck that allow you to bypass that.

scripts/check_clean_enlistment.sh
```diff
@@ -12,11 +12,9 @@
 #
 # Examples: scripts/check_clean_enlistment.sh
 
-# Treat travis builds as 'clean'
-if [ -z "${TRAVIS_BRANCH}" ]; then
-    if [[ -n $(git status --porcelain) ]]; then
-        exit 1
-    fi
+# Check for clean enlistment ignoring the update to buildnumber.dat which is expected during a release build
+if [ $(git status --porcelain | grep -v "buildnumber.dat" | wc -l) == 0 ]; then
```
tsachiherman, line 16: I don't think that this would work. You need to have a single "=" rather than two.

btoll:
Then I would do something like this in `checkout_branch.sh` and remove the check enlistment script:
```
if git status --porcelain; then
      echo Enlistment has changes. Please clean your enlistment before trying again.
      exit 1
fi
```
Putting `buildnumber.dat` in `.gitignore` is much better than hardcoding the file name.

onetechnical:
> Then I would do something like this in `checkout_branch.sh` and remove the check enlistment script:
> 
> ```
> if git status --porcelain; then
>       echo Enlistment has changes. Please clean your enlistment before trying again.
>       exit 1
> fi
> ```
> 
> Putting `buildnumber.dat` in `.gitignore` is much better than hardcoding the file name.

If you do that, won't you break future updates to buildnumber.dat? (I'm not wild about this file in the first place.)

btoll:
Yes, that's right and wouldn't work for this file.  I was thinking of how my portion of it works, where that file is already versioned.

scripts/check_clean_enlistment.sh
egieseke: == works, however, I replaced the enclosing brackets with double parens so that it is a numeric comparison rather than a string comparison.

scripts/check_clean_enlistment.sh
```diff
@@ -12,11 +12,9 @@
 #
 # Examples: scripts/check_clean_enlistment.sh
 
-# Treat travis builds as 'clean'
-if [ -z "${TRAVIS_BRANCH}" ]; then
-    if [[ -n $(git status --porcelain) ]]; then
-        exit 1
-    fi
+# Check for clean enlistment ignoring the update to buildnumber.dat which is expected during a release build
+if (( $(git status --porcelain | grep -v "buildnumber.dat" | wc -l) == 0 )); then
```
btoll, line 16: The `wc` in the pipeline isn't necessary, and you can just check the exit code in the branch to make this much simpler.
But, please see my suggestion about moving this to `checkout_branch.sh`.  It makes it much simpler.

Also, you'll want to install shellcheck to lint the script files that you work on, as it stands I believe the scripts would be generating errors.

egieseke: No errors, but shellcheck suggested using grep -c in place of wc -l.  I made that change.

btoll:
I don't agree with this approach, but I removed my block.

scripts/check_clean_enlistment.sh
tsachiherman: yes, that's a good idea as on Mac `wc -l` is adding spaces.

tsachiherman:
I think that with the removal of `check_clean_enlistment.sh` from the three shell scripts, it's no longer required to make the changes to `check_clean_enlistment.sh`, isn't it ?
The reason I'm asking is that it's enough to fix the issue, and we can always do refactoring in the future.

egieseke:
The updates to check_clean_enlistment.sh will allow it to work correctly on Jenkins or other environments.  I recommend that we keep those changes, versus creating a bug and fixing again at some point in the future.

tsachiherman:
I'm going to merge this in for now, as it solves the issue. I agree that there is more work to be done around it, but these issues could be deferred to a separate PRs.

egieseke:
@btoll , thanks for reviewing and your comments, can you approve the PR so that it can be committed?

PR #943, opened 2020-04-02
Disable binary backup when running make install

winder:
## Summary

`make install` calls `updater.sh`, which backs up the old binaries before installing new ones. On my machine it takes 7 seconds to backup the files, which for small/no-op builds is 50% of the build time. Added `-s` option to updater.sh to skip the backup, and enable it by default in `dev_install.sh`.

## Test Plan

Run `make install` and see that it doesn't freeze for 5-10 seconds while backing up binaries

### DIFF
--- a/cmd/updater/update.sh
+++ b/cmd/updater/update.sh
@@ -590,7 +590,9 @@ fi
 # Shutdown node before backing up so data is consistent and files aren't locked / in-use.
 shutdown_node
 
-backup_current_version
+if [ ${ALGOD_UPDATER_SKIP_BACKUP:-0} -ne 1 ]; then
+    backup_current_version
+fi
 
 # We don't care about return code - doesn't matter if we failed to archive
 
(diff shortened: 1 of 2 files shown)


winder:
Mainly because that would require updating the Makefile as well. Would you be more comfortable adding a CLI argument and updating the Makefile to use it?

tsachiherman:
Yes; Adding this to the pipeline makes perfect sense to me.
Keep in mind that this is a script that is used by end-users as well, so we want to keep it polished.
Note that `local_install.sh` already supports ADDITIONALFLAGS, which you could use to pass this new parameter.
I wouldn't mind if the developer default would not include backing up these files. I have never found them useful for debug purposes.

winder:
@tsachiherman I updated the PR with your suggestion, now everyone will save 5-10 seconds when they use `make install`!

tsachiherman:
nit: update the PR title and description

PR #944, opened 2020-04-02
Hint payset size in block evaluator

justicz:
In go, as `append` is called over and over on a slice, its capacity grows like this:

```
0xc000014180 cap 1
0xc0000200f0 cap 2
0xc00001a0c0 cap 4
0xc00009c000 cap 8
0xc00009e000 cap 16
0xc0000a0000 cap 32
0xc00001c600 cap 64
0xc0000a2000 cap 128
0xc0000a8000 cap 256
0xc0000ae000 cap 512
0xc0000b4000 cap 1024
0xc0000ba000 cap 1365
0xc0000c2000 cap 1706
0xc0000cc000 cap 2389
0xc0000da000 cap 3072
0xc0000ec000 cap 4096
0xc000104000 cap 5120
0xc000122000 cap 6485
0xc000148000 cap 8192
0xc000178000 cap 10240
```

Currently, in the `BlockEvaluator`, the block's `Payset` is growing only by successive calls to `append`. However, everywhere we are currently using a `BlockEvaluator`, we have a reasonable hint for how big the evaluated `Payset` will be.

On my `maxj/applications` branch with this PR applied, I'm seeing about a 16% performance improvement in `BenchmarkPay`, which evaluates full blocks of payment transactions. However, it would be good to see `pingpong` data on this branch before merging.

Additionally, though I don't see a reason why this would hurt performance for empty/small blocks, it would be good to confirm that this is the case.

### DIFF
--- a/data/pools/transactionPool.go
+++ b/data/pools/transactionPool.go
@@ -335,6 +335,17 @@ func (pool *TransactionPool) Remember(txgroup []transactions.SignedTxn, verifyPa
 	return nil
 }
 
+// PaysetLength returns the length of the current block evaluator's payset,
+// to be used as memory allocation size hint during block assembly.
+func (pool *TransactionPool) PaysetLength() int {
+	pool.mu.Lock()
+	defer pool.mu.Unlock()
+	if pool.pendingBlockEvaluator != nil {
+		return pool.pendingBlockEvaluator.PaysetLength()
+	}
+	return 0
+}
+
 // Lookup returns the error associated with a transaction that used
 // to be in the pool.  If no status information is available (e.g., because
 // it was too long ago, or the transaction committed successfully), then
@@ -484,6 +495,12 @@ func (pool *TransactionPool) addToPendingBlockEvaluator(txgroup []transactions.S
 // in-pool transactions to it (removing any transactions that are rejected
 // by the BlockEvaluator).
 func (pool *TransactionPool) recomputeBlockEvaluator(committedTxIds map[transactions.Txid]basics.Round) (stats telemetryspec.ProcessBlockMetrics) {
+	// Remember the current payset length as a size hint for the new evaluator
+	var poolPaysetLength int
+	if pool.pendingBlockEvaluator != nil {
+		poolPaysetLength = pool.pendingBlockEvaluator.PaysetLength()
+	}
+
 	pool.pendingBlockEvaluator = nil
 
 	latest := pool.ledger.Latest()
@@ -511,7 +528,7 @@ func (pool *TransactionPool) recomputeBlockEvaluator(committedTxIds map[transact
 
 	next := bookkeeping.MakeBlock(prev)
 	pool.numPendingWholeBlocks = 0
-	pool.pendingBlockEvaluator, err = pool.ledger.StartEvaluator(next.BlockHeader)
+	pool.pendingBlockEvaluator, err = pool.ledger.StartEvaluator(next.BlockHeader, poolPaysetLength)
 	if err != nil {
 		logging.Base().Warnf("TransactionPool.recomputeBlockEvaluator: cannot start evaluator: %v", err)
 		return
--- a/data/pools/transactionPool.go
+++ b/data/pools/transactionPool.go
@@ -514,7 +514,7 @@ func (pool *TransactionPool) recomputeBlockEvaluator(committedTxIds map[transact
 		return
 	}
 
-	// Grab the trnasactions to be replayed through the new block evaluator
+	// Grab the transactions to be played through the new block evaluator
 	pool.pendingMu.RLock()
 	txgroups := pool.pendingTxGroups
 	verifyParams := pool.pendingVerifyParams
@@ -529,7 +529,7 @@ func (pool *TransactionPool) recomputeBlockEvaluator(committedTxIds map[transact
 		return
 	}
 
-	// Feed the transactions in order.
+	// Feed the transactions in order
 	for i, txgroup := range txgroups {
 		if len(txgroup) == 0 {
 			continue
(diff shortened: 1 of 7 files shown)


justicz:
@egieseke just started a ping pong test on this branch, will report back when that's done :+1:

data/pools/transactionPool.go
```diff
@@ -164,7 +164,12 @@ func (pool *TransactionPool) rememberCommit(flush bool) {
 func (pool *TransactionPool) PendingCount() int {
 	pool.pendingMu.RLock()
 	defer pool.pendingMu.RUnlock()
+	return pool.pendingCountLocked()
+}
 
+// pendingCountLocked is a helper for PendingCount that returns the number of
+// transactions pending in the pool
+func (pool *TransactionPool) pendingCountLocked() int {
```
pzbitskiy, line 172: the name confusing, please rename to `pendingCountNoLock` or similar

justicz:
Hm, actually, this optimization was broken for zero-length paysets because a `nil` payset encodes differently than a non-`nil` but empty `Payset`. Pushed a change so that `nil` paysets and zero-length paysets have the same `Commit()`.

data/transactions/aggregates.go
```diff
@@ -37,6 +37,10 @@ type (
 // depends on the order in which the Txids appear, and that Txids do NOT cover
 // transaction signatures.
 func (payset Payset) Commit(flat bool) crypto.Digest {
+	if len(payset) == 0 {
```
pzbitskiy, line 40: I think a comment explaining this needed

justicz: Yeah -- and I might not actually be able to do this, because it looks like `makeGenesisBlock` actually calls `Commit` on a non-`nil` but empty payset...

data/pools/transactionPool.go
justicz: updated, thanks!

data/transactions/aggregates.go
justicz: Added a comment and confirmed that nodes can still make progress from genesis

justicz:
(confirmed I was able to successfully start mainnet catchup with this branch, indicating that even though `makeGenesisBlock` calls `Commit` on a non-`nil` but zero-length payset, we can still catch up)

data/transactions/aggregates.go
justicz: (see comment regarding `makeGenesisBlock` below)

ledger/eval_test.go
```diff
@@ -52,7 +52,7 @@ func TestBlockEvaluator(t *testing.T) {
 	defer l.Close()
 
 	newBlock := bookkeeping.MakeBlock(genesisInitState.Block.BlockHeader)
-	eval, err := l.StartEvaluator(newBlock.BlockHeader)
+	eval, err := l.StartEvaluator(newBlock.BlockHeader, 0)
```
tsachiherman, line 55: I'm curious why you chose to override the payset allocation inside StartEvaluator rather then when it was originally allocated in MakeBlock ( as nil, if I recall correctly ).

data/pools/transactionPool.go
```diff
@@ -509,20 +514,22 @@ func (pool *TransactionPool) recomputeBlockEvaluator(committedTxIds map[transact
...
+	pool.pendingMu.RLock()
+	txgroups := pool.pendingTxGroups
+	verifyParams := pool.pendingVerifyParams
+	pendingCount := pool.pendingCountNoLock()
+	pool.pendingMu.RUnlock()
+
 	next := bookkeeping.MakeBlock(prev)
 	pool.numPendingWholeBlocks = 0
-	pool.pendingBlockEvaluator, err = pool.ledger.StartEvaluator(next.BlockHeader)
+	pool.pendingBlockEvaluator, err = pool.ledger.StartEvaluator(next.BlockHeader, pendingCount)
```
tsachiherman, line 526: I'm curious why you chose to override the payset allocation inside StartEvaluator rather then when it was originally allocated in MakeBlock ( as nil, if I recall correctly ).
It's also weird that we create a new block ( i.e. MakeBlock ) just to steal it's header. but I guesses it's just semantics.

justicz: Yeah, `StartEvaluator` takes in a block header and not a whole block, and changing that behavior seemed like a much bigger design change.

ledger/eval.go
```diff
@@ -36,6 +36,11 @@ import (
 // ErrNoSpace indicates insufficient space for transaction in block
 var ErrNoSpace = errors.New("block does not have space for transaction")
 
+// maxPaysetHint makes sure that we don't allocate too much memory up front
+// in the block evaluator, since there cannot reasonably be more than this
+// many transactions in a block.
+const maxPaysetHint = 20000
```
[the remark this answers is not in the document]
justicz: I'd be ok with changing it to a constant 15,000 since I think there can't be more than that many transactions in a block, but since `TxPoolSize` is a variable that may be changed in `config.json` to any value, I don't think it is appropriate to use it here -- `maxPaysetHint` should just be >= the maximum number of transactions that could ever appear in a block, and < some value that could cause problems due to using a lot of memory.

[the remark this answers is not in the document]
tsachiherman: `maxPaysetHint` is a hint; incorrectly setting it would not make anything harmful other than inefficient allocation.

I don't think that we need this to start with, as the hint should suffice ( and, in practice, it already provides a cap ).

When it comes from the eval path, the limit matching the number of transaction in the block.
When it comes from the TxPool, the limit is the number of transactions that are in the TxPool, which is already capped at `TxPoolSize`. ( i.e. the actual number is abit higher, as it also include dups )

Either way, I'm going to merge this as is. If you feel like changing this to TxPoolSize or some other number, I'll be happy to approve and merge it on a separate PR.

PR #945, opened 2020-04-02
Remove call to build_deb.sh from build_packages.sh

btoll:
This code is only needed in one place, so let's move it there.  Also, extracting it from `build_packages.sh` helps greatly when creating breaking the build pipeline up into individual operations that then become `mule` tasks.


btoll:
Yes, I tested it.

The env vars are being set here: https://github.com/algorand/go-algorand/blob/master/scripts/release/common/setup.sh#L133

Keep in mind that this is for the baseline pipeline.  For the one using `mule`, the approach is different and will have checks for anything that is needed at the top of the script.

PR #946, opened 2020-04-03
Make license check failure more obvious

winder:
## Summary

This was in the travis log, but I didn't realize it indicated a license check failure:
```
...
Running go vet...
Running gofmt...
Running golint...
Running check_license...
daemon/algod/api/server/v2/errors.go
daemon/algod/api/server/v2/handlers.go
daemon/algod/api/server/v2/utils.go
The command "scripts/travis/build_test.sh" exited with 1.
```

My first thought for improving this was to add a big error message from check_license.sh on a failure. Instead I made check_license.sh act more like `go fmt`, and updated the travis `before_build.sh` script to output a failure message in the same pattern that fmt/vet use.

Also to hopefully prevent this from happening to begin with I moved `check_license.sh -i` into `make fmt`.

The travis failure will now look like this:
```
Running go vet...
Running gofmt...
Running golint...
Running check_license...
agreement/message.go
check_license.sh must be clean.  Please run the following to fix issues:
 check_license.sh -i
The command "scripts/travis/build_test.sh" exited with 1.
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -60,6 +60,7 @@ default: build
 
 fmt:
 	go fmt ./...
+	./scripts/check_license.sh -i
 
 fix: build
 	$(GOPATH1)/bin/algofix */
@@ -73,13 +74,10 @@ lint: deps
 vet:
 	go vet ./...
 
-check_license:
-	./scripts/check_license.sh
-
 check_shell:
 	find . -type f -name "*.sh" -exec shellcheck {} +
 
-sanity: vet fix lint fmt check_license
+sanity: vet fix lint fmt
 
 cover:
 	go test $(GOTAGS) -coverprofile=cover.out $(UNIT_TEST_SOURCES)
--- a/Makefile
+++ b/Makefile
@@ -301,7 +301,7 @@ dump: $(addprefix gen/,$(addsuffix /genesis.dump, $(NETWORKS)))
 install: build
 	scripts/dev_install.sh -p $(GOPATH1)/bin
 
-.PHONY: default fmt vet lint check_license check_shell sanity cover prof deps build test fulltest shorttest clean cleango deploy node_exporter install %gen gen NONGO_BIN
+.PHONY: default fmt vet lint check_shell sanity cover prof deps build test fulltest shorttest clean cleango deploy node_exporter install %gen gen NONGO_BIN
 
 ### TARGETS FOR CICD PROCESS
 
--- a/scripts/check_license.sh
+++ b/scripts/check_license.sh
@@ -24,7 +24,7 @@ usage() {
     echo
     echo "Args:"
     echo "-i    Edit in-place."
-    echo "-v    Verbose, same as doing \`head -n 15\` on each file."
+    echo "-v    Verbose, same as doing \`head -n 16\` on each file."
     echo
 }
 
@@ -94,5 +94,6 @@ if [ $RETURN_VALUE -ne 0 ]; then
     fi
     echo ""
 fi
+
 exit $RETURN_VALUE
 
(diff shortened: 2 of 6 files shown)


Makefile
```diff
@@ -73,13 +74,10 @@ lint: deps
...
 	go vet ./...
 
-check_license:
-	./scripts/check_license.sh
-
 check_shell:
 	find . -type f -name "*.sh" -exec shellcheck {} +
 
-sanity: vet fix lint fmt check_license
+sanity: vet fix lint fmt
```
tsachiherman, line 80: you need to remove the check_license target from the .PHONY list as well.

scripts/travis/before_build.sh
```diff
@@ -39,6 +39,20 @@ function runGoLint() {
     return 1
 }
 
+function runCheckLicense() {
```
tsachiherman, line 42: I'm not sure why you would need this change here. The changes in `check_license.sh` seems to be enough. ( unless you refer to the writing to the stderr, where I'm not a big fan of.
( i.e. it prevents capturing the output of the script by just storing all the stdout into a file )

scripts/check_license.sh
```diff
@@ -87,5 +86,14 @@ if [ "$(<README.md grep "${READMECOPYRIGHT}" | wc -l | tr -d ' ')" = "0" ]; then
     RETURN_VALUE=1
     echo "README.md file need to have it's license date range updated."
 fi
+
+if [ $RETURN_VALUE -ne 0 ] && [ $VERBOSE == "true" ] ; then
```
tsachiherman, line 90: looks to me like we want to have this regardless of verbosity, aren't we ?

scripts/check_license.sh
```diff
@@ -87,5 +86,14 @@ if [ "$(<README.md grep "${READMECOPYRIGHT}" | wc -l | tr -d ' ')" = "0" ]; then
...
+    echo -e "\n${RED_FG}FAILED LICENSE CHECK.${END_FG_COLOR}"
+    if [ $INPLACE == "false" ]; then
```
tsachiherman, line 92: In the case where we already modified the files, you might to write something different, like "5 files were modified" or something like that.

scripts/check_license.sh
```diff
@@ -24,7 +24,7 @@ usage() {
     echo
     echo "Args:"
     echo "-i    Edit in-place."
-    echo "-v    Verbose, same as doing \`head -n 15\` on each file."
+    echo "-v    Verbose, same as doing \`head -n 16\` on each file."
```
tsachiherman, line 27: nit: this magical number should be automatically derived from the license file rather than hard coded.

winder:
@tsachiherman @btoll updated with feedback.

There is one remaining problem that I wasn't able to solve. When using the in place update mode (`-i`) in many cases it inserts the header in a way that causes linting errors. Specifically if the first line was `package packageName` then there will be an error like:
```
agreement/errors.go:1:1: package comment should be of the form "Package agreement ..."
```

I wasn't sure how to fix this in the shell script, from some testing it looks like this line doesn't preserve the trailing newline, I was testing by putting `echo -e "$LICENSE"` in the usage output:
```
LICENSE=$(sed "s/{DATE_Y}/$(date +"%Y")/" "$LICENSE_LOCATION")
```

scripts/check_license.sh
```diff
@@ -87,5 +83,16 @@ if [ "$(<README.md grep "${READMECOPYRIGHT}" | wc -l | tr -d ' ')" = "0" ]; then
     RETURN_VALUE=1
     echo "README.md file need to have it's license date range updated."
```
btoll, line 84: Typo:  `it's` -> `its'

PR #949, opened 2020-04-03
Add e2e test for upgrading from v21 to v22.

derbear:


derbear:
```
// this test checks that two accounts can send money to one another
// across a protocol upgrade.
```

I was thinking it's sufficient to check that the network doesn't just break, but I guess we can also add a check that we've landed in v22.

PR #955, opened 2020-04-08
Support packaging with mule

btoll:


package.yaml
```diff
@@ -0,0 +1,57 @@
...
+      workDir: /projects/go-algorand
+    target: ci-build
+
+  - task: docker.Make
+    name: rpm
+    docker:
+      image: algorand/go-algorand-ci-linux-centos
+      version: '{{ docker.Version.outputs.version }}'
+      workDir: /projects/go-algorand
+      volumes: [ /home/btoll/go/bin:/projects/go-algorand/go/bin ]
```
btoll, line 33: Obviously, this needs more work.

package.yaml
```diff
@@ -0,0 +1,57 @@
...
+    target: package-rpm
+
+  - task: docker.Make
+    name: deb
+    docker:
+      image: algorand/go-algorand-ci-linux-ubuntu
+      version: '{{ docker.Version.outputs.version }}'
+      workDir: /projects/go-algorand
+      volumes: [ /home/btoll/go/bin:/projects/go-algorand/go/bin ]
```
btoll, line 42: Obviously, this needs more work.

btoll:
I tested the current `package.yml` file and its env vars agains the branch here: https://github.com/algorand/mule/pull/16 and it worked as expected.

scripts/release/mule/package/linux/amd64/deb/package.sh
```diff
@@ -0,0 +1,137 @@
...
+trap "rm -rf $PKG_ROOT" 0
+
+mkdir -p "${PKG_ROOT}/usr/bin"
+
+if [ "${VARIATIONS}" = "" ]; then
+    # NOTE: keep in sync with installer/rpm/algorand.spec
+    bin_files=("algocfg" "algod" "algoh" "algokey" "carpenter" "catchupsrv" "ddconfig.sh" "diagcfg" "goal" "kmd" "msgpacktool" "node_exporter")
+fi
+
+for bin in "${bin_files[@]}"; do
```
bricerisingalgorand, line 52: Could this be written so that it assumes these files are in the folders produced by ci-build? That way we can push the responsibility of producing artifacts to the build tasks. Also this may not work as expected where build artifacts are produced in a docker container and therefore won't be in the gopath of the host.

docker/build/cicd.centos.Dockerfile
```diff
@@ -20,11 +20,11 @@ ENV PATH=$GOPATH/bin:$GOROOT/bin:$PATH \
     BUILDCHANNEL=${BUILDCHANNEL} \
     DEFAULTNETWORK=${DEFAULTNETWORK} \
     FULLVERSION=${FULLVERSION} \
-    PKG_ROOT=${PKG_ROOT} \
-    GOPROXY=https://gocenter.io
```
egieseke, line 24: Why is GOPROXY removed?

package.yaml
```diff
@@ -0,0 +1,54 @@
...
+
+jobs:
+  package:
+    configs:
+      arch: amd64
+    tasks:
+      - shell.docker.Ensure.rpm
+      - docker.Make.rpm
+      - shell.docker.Ensure.deb
+      - docker.Make.deb
```
egieseke, line 53: What about the packaging of the docker container?

btoll: My thought was to have the docker container packaged in a separate task.  I think it makes sense to have that removed from the packaging of the deb and rpm.

docker/build/cicd.centos.Dockerfile
btoll: I'm not sure.  Perhaps over-zealousness :)

I'll restore it.

scripts/release/mule/package/linux/amd64/deb/package.sh
btoll: I'll take a look at this.

btoll: I've been thinking about this, and I think in order for this to happen some things need to be changed:

- The packaging is doing its own compilation of the binaries in the centos 7 image.  This is because we need to still support ubuntu 16.04 (although I think that may be deprecated soon), and the GNU C toolchain is too old on that distro so to be safe we're building on centos.

- Of course, the previous point could be made moot in two ways: the deprecation of support for ubuntu 16.04 or compiling our binaries in a way that doesn't have the external dependency on GNU C compiler.  I know there has been talk of this for a while, and I'm not sure what the state of it is.

- Also, how does the packaging "know" that the build artifacts are safe to be packaged?  What if there was a build that had been done a week ago but is stale now (perhaps you've already worked this out).

Once these issues are sorted, we could remove the build step from the packaging step and just pick up the previously-built assets, as you've suggested.

bricerisingalgorand: The artifacts built by the mule build jobs also use centos 7. I think I actually started by copying the dockerfile you're using here

bricerisingalgorand: https://github.com/algorand/go-algorand/blob/master/docker/build/cicd.Dockerfile

btoll: Ok, that's good, I'll have that use those build artifacts then.

Issue #942, opened 2020-04-02
Agreement attempting to create a proposal for unsupported protocol


tsachiherman:
## Issue

The issue is randomly demonstrated by TestStoppedCatchupOnUnsupported, causing a panic.

The catchup code was testing it’s ability to stop catching up once a consensus protocol is reached which is not supported by the local node. It was working fine.
The agreement service, however, tried to make a proposal for a block that it could not support, and failed doing so. The obvious solution would be ( of course ) to avoid making proposals & votes for unsupported consensus protocol versions.

## Log

```
=== CONT  TestStoppedCatchupOnUnsupported
3652Created new rootkey: /home/travis/gopath/src/github.com/algorand/go-algorand/tmp/out/e2e/678011328/TestStoppedCatchupOnUnsupported/Wallet2.rootkey
3653Created new partkey: /home/travis/gopath/src/github.com/algorand/go-algorand/tmp/out/e2e/678011328/TestStoppedCatchupOnUnsupported/Wallet2.0.3000000.partkey
3654Created new rootkey: /home/travis/gopath/src/github.com/algorand/go-algorand/tmp/out/e2e/678011328/TestBasicCatchup/Wallet1.rootkey
3655Created new partkey: /home/travis/gopath/src/github.com/algorand/go-algorand/tmp/out/e2e/678011328/TestBasicCatchup/Wallet1.0.3000000.partkey
3656Created new rootkey: /home/travis/gopath/src/github.com/algorand/go-algorand/tmp/out/e2e/678011328/TestStoppedCatchupOnUnsupported/Wallet1.rootkey
3657Created new rootkey: /home/travis/gopath/src/github.com/algorand/go-algorand/tmp/out/e2e/678011328/TestBasicCatchup/Wallet2.rootkey
3658test-unupgraded-protocol 100000
3659Created new partkey: /home/travis/gopath/src/github.com/algorand/go-algorand/tmp/out/e2e/678011328/TestBasicCatchup/Wallet2.0.3000000.partkey
3660https://github.com/algorandfoundation/specs/tree/57016b942f6d97e6d4c0688b373bb0a2fc85a1a2 100000
3661Created new partkey: /home/travis/gopath/src/github.com/algorand/go-algorand/tmp/out/e2e/678011328/TestStoppedCatchupOnUnsupported/Wallet1.0.3000000.partkey
[...]

PR #956, opened 2020-04-08
Update catchup test to support adjustable upgrade window

tsachiherman:
## Summary

The catchup upgrade test was not setting up the MinUpgradeWaitRounds while using an upgrade window of 0. This caused the test to use an MinUpgradeWaitRounds=10000, which failed the applyUpgradeVote test. 

Fix https://github.com/algorand/go-algorand/issues/942


tsachiherman:
It was missed because this e2e test is run only on nightly builds.

algonautshant:
Is this related to DefaultUpgradeWaitRounds? If so, we may have similar situation in: 
node/node_test.go in TestSimpleUpgrade
and 
test/e2e-go/upgrades/send_receive_upgrade_test.go

PR #957, opened 2020-04-08
Remove `check_clean_enlistment.sh` script

btoll:
Note that I changed this because I ran into this again with the build pipeline work.  I've found the word "enlistment" to be confusing, so I wanted to change that message.

----------------------------

This file was only being referenced from one file, so I removed the
conditional that was in it to where it was being referenced.

- Improved warning message when the workspace is dirty.
- Fixed bug where files like `buildnumber.datdat` are not flagged.
- Fixed warnings discovered by shellcheck.


algobolson:
There are a lot of error checks. Might be simpler to run bash with `set -e` and just require every command to pass?

btoll:
Yes, good idea, I'll change that.

btoll:
Ok, will change.

btoll:
That was an oversight.  Thanks for catching that.

PR #958, opened 2020-04-09
Bugfix: Swap EnableAssembleStats with EnableProcessBlockStats.

derbear:
The documented intended behavior for EnableAssembleStats is to
generate AssembleBlockMetrics, while for EnableProcessBlockStats it is
to generate ProcessBlockMetrics.  The current implementation reverses
these behaviors.

### DIFF
--- a/data/pools/transactionPool.go
+++ b/data/pools/transactionPool.go
@@ -79,7 +79,7 @@ func MakeTransactionPool(ledger *ledger.Ledger, cfg config.Local) *TransactionPo
 		expiredTxCount:  make(map[basics.Round]int),
 		ledger:          ledger,
 		statusCache:     makeStatusCache(cfg.TxPoolSize),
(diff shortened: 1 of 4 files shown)


data/pools/transactionPool.go
```diff
@@ -79,7 +79,7 @@ func MakeTransactionPool(ledger *ledger.Ledger, cfg config.Local) *TransactionPo
 		expiredTxCount:  make(map[basics.Round]int),
 		ledger:          ledger,
 		statusCache:     makeStatusCache(cfg.TxPoolSize),
-		logStats:        cfg.EnableAssembleStats,
+		logStats:        cfg.EnableProcessBlockStats,
```
tsachiherman, line 82: nit: a better name for this variable would be enableLogStats, but it's beyond the scope of your change.

PR #959, opened 2020-04-09
Simplify block assembly.

derbear:
This change simplifies and slightly speeds up block assembly.

This change moves block assembly to the transaction pool in preparation for further optimizations, which involve coordinating between the block assembly goroutine and the transaction pool to reduce wasted work.

### DIFF
--- a/data/pools/transactionPool.go
+++ b/data/pools/transactionPool.go
@@ -555,3 +555,131 @@ func (pool *TransactionPool) recomputeBlockEvaluator(committedTxIds map[transact
 	pool.rememberCommit(true)
 	return
 }
+
+// AssembleBlock assembles a block for a given round, trying not to
+// take longer than deadline to finish.
+func (pool *TransactionPool) AssembleBlock(round basics.Round, deadline time.Time, logStats bool) (*ledger.ValidatedBlock, error) {
+	start := time.Now()
+	prev, err := pool.ledger.BlockHdr(round - 1)
+	if err != nil {
+		return nil, fmt.Errorf("could not make proposals at round %d: could not read block from ledger: %v", round, err)
+	}
+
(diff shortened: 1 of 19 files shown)


node/node.go
```diff
@@ -231,7 +230,7 @@ func MakeFull(log logging.Logger, rootDir string, cfg config.Local, phonebookAdd
 		Local:          node.config,
 		Network:        gossip.WrapNetwork(node.net, log),
 		Ledger:         agreementLedger,
-		BlockFactory:   blockFactory,
+		BlockFactory:   blockFactory{TransactionPool: node.transactionPool},
```
tsachiherman, line 233: If you're doing this, why don't you eliminate the blockFactoryImpl completely and move it's functionality to the node ?

derbear: Sounds good--removed.

data/pools/transactionPool.go
```diff
@@ -555,3 +559,133 @@ func (pool *TransactionPool) recomputeBlockEvaluator(committedTxIds map[transact
...
+	stats.Nanoseconds = dt.Nanoseconds()
+
+	if pool.logAssembleStats {
+		payset := lvb.Block().Payset
+		if len(payset) != 0 {
+			totalFees := uint64(0)
+
+			for i, txib := range payset {
+				fee := txib.Txn.Fee.Raw
+				encodedLen := txib.GetEncodedLength()
```
algobolson, line 655: I kinda bet that this is the slowest part of turning on logAssembleStats; maybe a TODO comment or a backlog issue for making a stats-lite mode that records the cheap stats?

derbear:
@egieseke ran one set of S1 tests on this branch tip, which showed a 2% decrease in throughput relative to master. Any opinions on whether this merits holding off on a merge?

PR #961, opened 2020-04-09
Remove unnecessary call to `getChangedAssetIndices` in acctupdates.go

justicz:
This code was needlessly recomputing deltas that already existed in the cow.


justicz:
(on my applications branch, I also completely removed the equivalent of `getChangedAssetIndices` by explicitly informing the cow when an app/asset is created)

PR #963, opened 2020-04-09
Avoid running block assembly twice per round.

derbear:


derbear:
It may be easier to read this diff: https://github.com/derbear/go-algorand/compare/clean-up-assembly...derbear:assemble-once

ledger/eval.go
```diff
@@ -684,6 +684,7 @@ func (eval *BlockEvaluator) GenerateBlock() (*ValidatedBlock, error) {
 		blk:   eval.block,
 		delta: eval.state.mods,
 	}
+	eval.state = makeRoundCowState(eval.state, eval.block.BlockHeader) // XXX fix this
```
tsachiherman, line 687: Do you want to fix this one ? ( or at least address the comment ? )

data/pools/transactionPool.go
```diff
@@ -540,159 +584,143 @@ func (pool *TransactionPool) recomputeBlockEvaluator(committedTxIds map[transact
...
-		var details struct {
-			Round uint64
-		}
-		details.Round = uint64(round)
-		logging.Base().Metrics(telemetryspec.Transaction, stats, details)
+	pool.assemblyMu.Lock()
+	defer pool.assemblyMu.Unlock()
+
+	pool.assemblyDeadline = deadline
+	deadline = deadline.Add(assemblyWaitEps)
```
tsachiherman, line 707: I'm not opposing to add 10ms to the 250ms delay, but could you explain why these 10ms are needed ?
i.e. isn't the transaction pool receives the OnNewBlock *always* before the agreement is attempting to make a proposal ?

derbear: I wanted to avoid the scenario where `recomputeBlockEvaluator` broadcasts on the condition variable at 1ms post-deadline, but `AssembleBlock` wakes up at exactly the deadline and so fails to receive the block. The other approach I was thinking of was to have `recomputeBlockEvaluator` attempt to broadcast at some epsilon before the deadline.

justicz:
From a performance perspective, I'm seeing significant TPS improvements on the stateful TEAL branch with this change applied.

ledger/eval.go
derbear: Yeah--this was a hack so that I could reuse the block evaluator. I'm wondering what a good way of explicitly allowing reuse is.

zeldovich: Maybe putting it into `GenerateBlock()` would make more sense.

Depending on whether you still have callers that want a "dead-end" `GenerateBlock()`, or it's only the txpool calling this function, you might consider adding an argument to `GenerateBlock()` to indicate whether the caller plans to add more transactions to this block evaluator later.  We should guard against repeated calls to `GenerateBlock()` on the same `BlockEvaluator`, just in case.

derbear:
Rendered obsolete with #917.

PR #964, opened 2020-04-09
Add optional pk argument to KMD's SignTransaction

algoradam:
Give KMD's SignTransaction an (optional) public key argument.
If provided, KMD signs the transaction using the corresponding private key and sets the AuthAddr if necessary.
goal / libgoal will use this KMD functionality in order to sign transactions from rekeyed accounts (which will come in a separate PR)

### DIFF
--- a/daemon/kmd/lib/kmdapi/requests.go
+++ b/daemon/kmd/lib/kmdapi/requests.go
@@ -163,6 +163,7 @@ type APIV1POSTTransactionSignRequest struct {
 	WalletHandleToken string `json:"wallet_handle_token"`
 	// swagger:strfmt byte
 	Transaction    []byte `json:"transaction"`
+	PublicKey      crypto.PublicKey `json:"public_key"`
 	WalletPassword string `json:"wallet_password"`
 }
 
--- a/daemon/kmd/lib/kmdapi/requests.go
+++ b/daemon/kmd/lib/kmdapi/requests.go
@@ -162,9 +162,9 @@ type APIV1POSTTransactionSignRequest struct {
 	APIV1RequestEnvelope
 	WalletHandleToken string `json:"wallet_handle_token"`
 	// swagger:strfmt byte
-	Transaction    []byte `json:"transaction"`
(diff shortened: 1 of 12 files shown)


daemon/kmd/wallet/driver/ledger.go
```diff
@@ -307,16 +308,39 @@ func (lw *LedgerWallet) DeleteMultisigAddr(addr crypto.Digest, pw []byte) error
...
+	if err != nil {
+		return nil, err
+	}
+	// Right now the device only supports one key
+	if len(pks) > 1 {
+		return nil, errors.New("LedgerWallet device only supports one key but ListKeys returned more than one")
+	}
+	if (pk != crypto.PublicKey{}) {
+		// A specific key was requested; return an error if it's not the one on the device.
+		if len(pks) < 1 || pks[0] != crypto.Digest(pk) {
```
justicz, line 322: you check `len(pks) < 1` here before dereferencing with  `pks[0]`, but don't do that check before dereferencing below on line 326. can you please be consistent one way or the other?

justicz:
Seems good, tested on my ledger to make sure I could still sign, other than that one code style comment. Will approve once build is passing, seems like there's a weird codec thing going on?

daemon/kmd/wallet/driver/ledger.go
algoradam: Good catch, thanks!

daemon/kmd/lib/kmdapi/requests.go
```diff
@@ -163,6 +163,7 @@ type APIV1POSTTransactionSignRequest struct {
 	WalletHandleToken string `json:"wallet_handle_token"`
 	// swagger:strfmt byte
 	Transaction    []byte `json:"transaction"`
+	PublicKey      crypto.PublicKey `json:"public_key"`
 	WalletPassword string `json:"wallet_password"`
```
justicz, line 167: gofmt everything pls

PR #965, opened 2020-04-10
Allow trailing slash in rest endpoints.

winder:
It seems that Gorilla Mux ignores trailing slashes while Echo is more strict about them. The Java SDK was appending a slash, so it isn't working properly without this fix.

### DIFF
--- a/daemon/algod/api/server/router.go
+++ b/daemon/algod/api/server/router.go
@@ -64,6 +64,7 @@ import (
 	"net/http"
 
 	"github.com/labstack/echo/v4"
+	"github.com/labstack/echo/v4/middleware"
 
 	"github.com/algorand/go-algorand/daemon/algod/api/server/common"
 	"github.com/algorand/go-algorand/daemon/algod/api/server/lib"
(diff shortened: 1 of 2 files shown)


daemon/algod/api/server/router.go
```diff
@@ -103,6 +104,9 @@ func NewRouter(logger logging.Logger, node *node.AlgorandFullNode, shutdown <-ch
 	e.Listener = listener
 	e.HideBanner = true
 
+	// Make sure '/v1/ledger/supply' and '/v1/ledger/supply/' are both accepted.
+	e.Pre(middleware.RemoveTrailingSlash())
+
```
tsachiherman, line 109: Which one of these is the correct one according to the REST API ? I'd guess the former, and so the Java SDK would need to be modified to emit the correct request.

Also, note that the comment suggest that it applies to `/v1/ledger/supply` command where is it affects all the v1 endpoints.

winder: The correct form is with no trailing slash. The Java SDK should also be fixed to stop appending the trailing slash.

The fix that I added here effects all endpoints, they'll now be accepted with and without the trailing slash.

daemon/algod/api/server/router.go
```diff
@@ -103,6 +104,9 @@ func NewRouter(logger logging.Logger, node *node.AlgorandFullNode, shutdown <-ch
 	e.Listener = listener
 	e.HideBanner = true
 
+	// Make sure '/v1/ledger/supply' and '/v1/ledger/supply/' are both accepted.
```
winder, line 107:
```suggestion
```

PR #967, opened 2020-04-14
Beef up baseline build release pipeline docs

btoll:


scripts/release/README.md
```diff
@@ -46,20 +50,55 @@ This section briefly describes the expected outcomes of the current build pipeli
...
 
-    The result of running this job will be to put the build artifacts and their detached signatures in the staging `algorand-builds` bucket.
+    1. Build (compile) the binaries in a Centos 7 docker container that will then be used by both `deb` and `rpm` packaging.
 
-    In addition, the build logs will be placed into the AWS S3 bucket`algorand-builds/build-logs/channel`.
+    1. Docker containers will package `deb` and `rpm` artifacts inside of Ubuntu 18.04 and Centos 7, respectively.
+
+    1. Jenkins will then pause to wait for [the only manual part of the build/package/test phase], which is to forward the `gpg-agent` that establishes a direct between the local machine that contains the signing keys and the remote ec2 instance.
+
+    1. Once the signatures have been verified, the all the build artificats (tarballs, `rpm` and `deb` packages, signatures) to an `s3` bucket.  Included in the bucket are the build logs.
```
onetechnical, line 59:
```suggestion
    1. Once the signatures have been verified, all the build artifacts (tarballs, `rpm` and `deb` packages, signatures) are uploaded to an `s3` bucket.  Included in the bucket are the build logs.
```

scripts/release/README.md
```diff
@@ -46,20 +50,55 @@ This section briefly describes the expected outcomes of the current build pipeli
...
+
+    Each package is downloaded in the respective environment and the following tests are done:
+
+    - The signatures are correct.
+    - The packages are built from the correct branch and channel and are the correct version.  This done by running `algod -v`.
+        + This is done for the following docker containers:
+            - centos:7
+            - fedora:28
+            - ubuntu:16.04
+            - ubuntu:18.04
```
onetechnical, line 75: Are we going to add ubuntu:20.04 and deprecate ubuntu 16:04 after 4/20?

scripts/release/README.md
```diff
@@ -46,20 +50,55 @@ This section briefly describes the expected outcomes of the current build pipeli
...
+
+        `aws s3 sync s3://{SOURCE} s3://{DESTINATION}`
+
+    In addition, the releases page (located on s3 in the root of the `algorand-releases` bucket), is updated automatically by [running a python script] (although, this does need some more work to be fully automated).
+
+    > Note that the [`reverse_hex_timestamp`] script is used to create a bucket name so that it will always be the first available folder in the `algorand-dev-deb-repo` bucket.  This is used by the just-mentioned python release page script to ensure that it always grabs the most-recent build.
+    >
+    > For example:
+    >
+    >   `fa16c3cb5_2.0.13`
```
onetechnical, line 96: Not for this PR, just a general question - why not just use a datetime stamp like YYYYMMDD-hhmm-VERSION? Because it's shorter?

onetechnical: Oh, first in the list...I see. Still seems weird.

scripts/release/README.md
```diff
@@ -46,20 +50,55 @@ This section briefly describes the expected outcomes of the current build pipeli
...
 
     In addition, local snapshots are used by Debian-based (`aptly`) and RHEL-based tooling to deploy the respective packages to `algorand-releases`.  These are the packages which can then by downloaded by `apt` and `yum`.
 
+    There are now three separate distributions: `stable`, `beta` and `nightly`.  All three distros contain multiple versions of the following packages, respectively: `algorand`, `algorand-beta` and `algorand-nightly`.
```
onetechnical, line 100: Does this mean we can have a 'nightly' distribution with an algorand-beta package? Is there a reason we'd want this?

scripts/release/README.md
btoll: Sounds good to me.

scripts/release/README.md
btoll: This may change after it's decided what to do with that releases page.

scripts/release/README.md
btoll: It wouldn't be from the beta channel, at least I don't think so.  I don't think that the process for issuing nightly builds has been fully fleshed out yet, so I'll remove the bit about the nightly for now.

PR #968, opened 2020-04-14
Encode `SignedTxnInBlock` to measure length, not `SignedTxn`

justicz:
`GetEncodedLength` is defined on `SignedTxn`, not `SignedTxnInBlock` (which embeds `SignedTxnWithAD` which embeds `SignedTxn`). So attempting to call it on a `SignedTxnInBlock` will produce an unexpected result since the encoding will omit e.g. `ApplyData`.

This PR matches is closer to how length is [actually computed in the block evaluator](https://github.com/algorand/go-algorand/blob/82e804e4c374876e3c8a2f45d2ce49cdbf82dc26/ledger/eval.go#L485).


algobolson:
on the plus side, this only affects reported stats, and block-full already detects this length. But yes, I believe this is more correct.

PR #970, opened 2020-04-15
Fast Catchup

tsachiherman:
## Fast Catchup

Fast catchup is an alternative, user-driven, complementary way to have a local node catchup in a realistic and (relatively) constant time.

## What's in this PR ?
1. Enable the goal commands to use the fast catchup
1. Add unit test TestLargeAccountCountCatchpointGeneration
1. Add telemetry event for catchpoint file generation
1. Fix bug when creating a catchpoint for a period that has no account changes
1. Documentation for watchdogStreamReader
1. Add unit test TestBasicCatchpointWriter


daemon/algod/api/server/v1/handlers/handlers.go
```diff
@@ -442,6 +451,17 @@ func WaitForBlock(ctx lib.ReqContext, context echo.Context) {
 		}
 	}
 
+	nodeStatus, err := getNodeStatus(ctx.Node)
```
winder, line 454: If you implement this check as a middleware we can ensure that it is done for every endpoint, and we don't have to remember to add it to future endpoints.

Which endpoints that we want to allow even when the node is catchup up? At least the `DELETE` catchpoint call, what else?

daemon/algod/api/server/v1/handlers/handlers.go
```diff
@@ -42,21 +42,27 @@ import (
...
 )
 
-func nodeStatus(node *node.AlgorandFullNode) (res v1.NodeStatus, err error) {
+func getNodeStatus(node *node.AlgorandFullNode) (res v1.NodeStatus, err error) {
 	stat, err := node.Status()
 	if err != nil {
 		return v1.NodeStatus{}, err
 	}
 
 	return v1.NodeStatus{
```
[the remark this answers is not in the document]
tsachiherman: That's a good point. I will be creating a separate PR for moving the goal to use v2 for the status requests, and then remove these from this one.

daemon/algod/api/server/v1/handlers/handlers.go
tsachiherman: We clearly want this in any API that touches the transaction pool. 
I'm not sure how easy it would be to add this as a middleware. I would imagine that we need to export the spec tags into the code so that we can make these decision in runtime.

daemon/algod/api/server/v1/handlers/handlers.go
```diff
@@ -570,6 +604,7 @@ func AccountInformation(ctx lib.ReqContext, context echo.Context) {
 		return
 	}
 
+	// the following has a bug : the request below is not atomic.
```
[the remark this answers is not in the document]
tsachiherman: I agree that it's not a very frequent issue, but it's a correctness issue - not likelihood issue.
To fix this, the entire logic here need to be ported down the stack and executed at a location where we can take a ledger lock. ( it's likely to be in accountupd.go ).
obviously, it has nothing to do with this PR, so I'll remove it from here.

daemon/algod/api/server/v2/handlers.go
```diff
@@ -248,6 +255,24 @@ func (v2 *Handlers) WaitForBlock(ctx echo.Context, round uint64) error {
 	if stat.StoppedAtUnsupportedRound {
 		return badRequest(ctx, err, errRequestedRoundInUnsupportedRound, v2.Log)
 	}
+	if stat.Catchpoint != "" {
+		// node is currently catching up to the requested catchpoint.
+		return serviceUnavailable(ctx, fmt.Errorf("WaitForBlock failed as the node was catchpoint catchuping"), errOperationNotAvailableDuringCatchup, v2.Log)
+	}
+
```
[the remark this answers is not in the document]
tsachiherman: not quite. The `stat.StoppedAtUnsupportedRound` is called once we already stopped, and it won't make any further progress. The other check verify the following:
- let's say we're at round 54
- let's say the node will switch to a new protocol (unsupported) on round 60
- let's say the WaitForBlock was asked to wait until round 65
In that case, there is no point to *start* waiting for round 65. It's guaranteed never to arrive.

daemon/algod/api/server/v1/handlers/handlers.go
winder: We could decide to be overly aggressive and disable all<sub>1</sub> v1/v2 endpoints will be disabled during catchup. This may be a safer approach, since we wouldn't be able to forget the catchup check in future endpoints.

1: exceptions to all:
* `DELETE` + path.contains(`/v2/catchup/`)
* `GET` + path.contains(`v[\d]/status`)

tsachiherman: Isn't that being abit draconic ? I was trying to make sure that we limit the functionality only where it makes a difference. 
btw - there are few more exceptions in the common api.

catchup/ledgerFetcher.go
```diff
@@ -0,0 +1,160 @@
...
+	var downloadProgress ledger.CatchpointCatchupAccessorProgress
+	for {
+		header, err := tarReader.Next()
+		if err != nil {
+			if err == io.EOF {
+				return nil
+			}
+			return err
+		}
+		balancesBlockBytes := make([]byte, header.Size)
```
pzbitskiy, line 131: how big the ledger?

catchup/service.go
```diff
@@ -130,7 +129,6 @@ func (s *Service) Stop() {
 	if atomic.CompareAndSwapUint32(&s.initialSyncNotified, 0, 1) {
 		close(s.InitialSyncDone)
 	}
-	s.auth.Quit()
```
pzbitskiy, line 133: why so?

catchup/service_test.go
```diff
@@ -619,6 +634,13 @@ func (avv *MockVoteVerifier) Parallelism() int {
 	return 1
 }
 
+// Start the catchup service, without starting the periodic sync.
+func (s *Service) testStart() {
```
pzbitskiy, line 638: hacky and nice :)

config/local_defaults.go
```diff
@@ -39,6 +39,65 @@ const configVersion = uint32(6)
...
+	TelemetryToLog:                        true,
+	TxPoolExponentialIncreaseFactor:       2,
+	TxPoolSize:                            15000,
+	TxSyncIntervalSeconds:                 60,
+	TxSyncTimeoutSeconds:                  30,
+	TxSyncServeResponseSize:               1000000,
+	PeerConnectionsUpdateInterval:         3600,
+	DNSSecurityFlags:                      0x01,
+	EnablePingHandler:                     true,
+	CatchpointInterval:                    20000, // added in V7
```
pzbitskiy, line 94: thank you

daemon/algod/api/algod.oas2.json
```diff
@@ -595,6 +631,106 @@
           "required": true
         }
       ]
+    },
+    "/v2/catchup/{catchpoint}": {
+      "put": {
```
pzbitskiy, line 636: post, as discussed?

daemon/algod/api/server/v1/handlers/handlers.go
winder: It might be, but I'm having trouble coming up with any legitimate uses for an out of date node.

ledger/catchupaccessor.go
```diff
@@ -0,0 +1,490 @@
...
+	catchpointCatchupStateLast = CatchpointCatchupStateSwitch
+)
+
+// MakeCatchpointCatchupAccessor creates a CatchpointCatchupAccessor given a ledger
+func MakeCatchpointCatchupAccessor(ledger *Ledger, log logging.Logger) *CatchpointCatchupAccessor {
+	rdb := ledger.trackerDB().rdb
+	wdb := ledger.trackerDB().wdb
+	accountsq, err := accountsDbInit(rdb.Handle, wdb.Handle)
+	if err != nil {
+		// todo : log error.
```
pzbitskiy, line 69: log error :)

ledger/catchupaccessor.go
```diff
@@ -0,0 +1,490 @@
...
+	}
+	if strings.HasPrefix(sectionName, "balances.") && strings.HasSuffix(sectionName, ".msgpack") {
+		return c.processStagingBalances(ctx, bytes, progress)
+	}
+	// we want to allow undefined sections to support backward compatibility.
+	c.log.Warnf("CatchpointCatchupAccessor::ProgressStagingBalances encountered unexpected section name '%s' of length %d, which would be ignored", sectionName, len(bytes))
+	return nil
+}
+
+// ProgressStagingBalances deserialize the given bytes as a temporary staging balances
```
pzbitskiy, line 184: ProgressStagingBalances  -> processStagingContent

ledger/catchupaccessor.go
```diff
@@ -0,0 +1,490 @@
...
+	})
+	if err == nil {
+		progress.SeenHeader = true
+		progress.TotalAccounts = fileHeader.TotalAccounts
+		progress.TotalChunks = fileHeader.TotalChunks
+	}
+	return err
+}
+
+// ProgressStagingBalances deserialize the given bytes as a temporary staging balances
```
pzbitskiy, line 226: processStagingBalances

ledger/catchupaccessor.go
```diff
@@ -0,0 +1,490 @@
...
+	_, err = c.accountsq.writeCatchpointStateUint64(ctx, "catchpointCatchupState", uint64(state))
+	if err != nil {
+		return fmt.Errorf("unable to write catchpoint catchup state 'catchpointCatchupState': %v", err)
+	}
+	return
+}
+
+// GetLabel returns the current catchpoint catchup label
+func (c *CatchpointCatchupAccessor) GetLabel(ctx context.Context) (label string, err error) {
+	label, _, err = c.accountsq.readCatchpointStateString(ctx, "catchpointCatchupLabel")
```
pzbitskiy, line 104: probably it is better to make these string literals named constants

ledger/ledger.go
```diff
@@ -249,13 +276,18 @@ func initBlocksDB(tx *sql.Tx, l *Ledger, initBlocks []bookkeeping.Block, isArchi
 // Close reclaims resources used by the ledger (namely, the database connection
 // and goroutines used by trackers).
 func (l *Ledger) Close() {
-	l.trackerDBs.close()
-	l.blockDBs.close()
-	l.trackers.close()
+	// we shut the the blockqueue first, since it's sync goroutine dispatches calls
+	// back to the trackers.
```
pzbitskiy, line 280: nice catch

ledger/ledger.go
```diff
@@ -472,11 +512,28 @@ func (l *Ledger) GenesisHash() crypto.Digest {
 	return l.genesisHash
 }
 
+// GetCatchpointCatchupState returns the current state of the catchpoint catchup.
+func (l *Ledger) GetCatchpointCatchupState(ctx context.Context) (state CatchpointCatchupState, err error) {
+	return MakeCatchpointCatchupAccessor(l, l.log).GetState(ctx)
+}
+
+// GetCatchpointStream - todo ..
```
pzbitskiy, line 520: comment

ledger/totals.go
```diff
@@ -116,3 +119,72 @@ func (at *AccountTotals) RewardUnits() uint64 {
...
+
+// CanMarshalMsg implements msgp.Marshaler
+func (AccountTotals) CanMarshalMsg(z interface{}) bool {
+	_, ok := (z).(AccountTotals)
+	return ok
+}
+
+// MarshalMsg implements msgp.Marshaler
+func (at AccountTotals) MarshalMsg(b []byte) (o []byte, err error) {
+	//o = msgp.Require(b, msgp.Uint64Size)
```
pzbitskiy, line 131: dead comments

ledger/totals.go
```diff
@@ -116,3 +119,72 @@ func (at *AccountTotals) RewardUnits() uint64 {
...
+}
+
+// CanUnmarshalMsg implements msgp.Unmarshaler
+func (*AccountTotals) CanUnmarshalMsg(z interface{}) bool {
+	_, ok := (z).(*AccountTotals)
+	return ok
+}
+
+// UnmarshalMsg implements msgp.Unmarshaler
+func (at *AccountTotals) UnmarshalMsg(bts []byte) (o []byte, err error) {
```
pzbitskiy, line 157: serialization/deserialization test?

network/multiplexer.go
```diff
@@ -85,6 +86,18 @@ func (m *Multiplexer) RegisterHandlers(dispatch []TaggedMessageHandler) {
 }
 
 // ClearHandlers deregisters all the existing message handlers.
-func (m *Multiplexer) ClearHandlers() {
-	m.msgHandlers.Store(make(map[Tag]MessageHandler))
+func (m *Multiplexer) ClearHandlers(excludeTags []Tag) {
+	excludeTagsMap := make(map[Tag]bool)
+	for _, tag := range excludeTags {
+		excludeTagsMap[tag] = true
+	}
```
pzbitskiy, line 93: move under `if` statement => one less alloc

network/multiplexer.go
```diff
@@ -85,6 +86,18 @@ func (m *Multiplexer) RegisterHandlers(dispatch []TaggedMessageHandler) {
...
+	newMap := make(map[Tag]MessageHandler)
```
pzbitskiy, line 94: prealloc len(m.getHandlersMap()) - len(excludeTags)

rpcs/wsFetcherService.go
```diff
@@ -176,15 +177,26 @@ func (fs *WsFetcherService) RequestBlock(ctx context.Context, target network.Uni
...
+func (fs *WsFetcherService) Start() {
+	handlers := []network.TaggedMessageHandler{
+		{Tag: protocol.UniCatchupResTag, MessageHandler: network.HandlerFunc(fs.handleNetworkMsg)},  // handles the response for a block catchup request
+		{Tag: protocol.UniEnsBlockResTag, MessageHandler: network.HandlerFunc(fs.handleNetworkMsg)}, // handles the response for a block ensure digest request
+	}
+	fs.net.RegisterHandlers(handlers)
+}
+
+// Stop stops the WsFetcherService
+func (fs *WsFetcherService) Stop() {
```
pzbitskiy, line 200: unregister ?

node/node.go
```diff
@@ -107,16 +118,15 @@ type AlgorandFullNode struct {
 	lastRoundTimestamp    time.Time
 	hasSyncedSinceStartup bool
 
-	txPoolSyncer *rpcs.TxSyncer
+	txPoolSyncerService *rpcs.TxSyncer
```
pzbitskiy, line 121: If you touch it... please move above to other services?

agreement/demux.go
```diff
@@ -69,21 +69,21 @@ type demux struct {
 // makeDemux initializes the goroutines needed to process external events, setting up the appropriate channels.
 //
 // It must be called before other methods are called.
-func makeDemux(net Network, ledger LedgerReader, validator BlockValidator, voteVerifier *AsyncVoteVerifier, processingMonitor EventsProcessingMonitor, log logging.Logger) (d *demux) {
+func makeDemux(net Network, ledger LedgerReader, validator BlockValidator, voteVerifier *AsyncVoteVerifier, processingMonitor EventsProcessingMonitor, log logging.Logger, monitor *coserviceMonitor) (d *demux) {
```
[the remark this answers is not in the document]
tsachiherman: I'm ok with packing this into a DemuxParams.

node/node.go
```diff
@@ -346,21 +390,29 @@ func (node *AlgorandFullNode) ListeningAddress() (string, bool) {
...
-	node.txHandler.Stop()
 	node.net.Stop()
-	node.algorandService.Shutdown()
-	node.syncer.Stop()
-	node.txPoolSyncer.Stop()
-	node.ledgerService.Stop()
+	if node.catchpointCatchupService != nil {
+		node.catchpointCatchupService.Stop()
+	} else {
+		node.txHandler.Stop()
```
pzbitskiy, line 403: does startup/shutdown order matter?

node/node.go
```diff
@@ -756,6 +825,104 @@ func (node *AlgorandFullNode) GetTransactionByID(txid transactions.Txid, rnd bas
...
+		} else {
+			node.log.Info("Indexer was started successfully")
+		}
+	} else {
+		node.log.Infof("Indexer is not available - %v", err)
+	}
+
+	node.cancelCtx()
+
+	node.startMonitoringRoutines()
```
pzbitskiy, line 916: this is similar to `MakeFull`, mb make a function like `startServices()`?

node/node.go
```diff
@@ -756,6 +825,104 @@ func (node *AlgorandFullNode) GetTransactionByID(txid transactions.Txid, rnd bas
...
+	stats := node.catchpointCatchupService.GetStatistics()
+	if stats.CatchpointLabel != catchpoint {
+		return fmt.Errorf("unable to abort catchpoint catchup for '%s' - already catching up '%s'", catchpoint, stats.CatchpointLabel)
+	}
+	node.catchpointCatchupService.Abort()
+	return nil
+}
+
+// SetCatchpointCatchupMode change the node's operational mode from catchpoint catchup mode and back
+func (node *AlgorandFullNode) SetCatchpointCatchupMode(catchpointCatchupMode bool) (newCtx context.Context) {
```
pzbitskiy, line 867: Is my understanding correct?
1. If prev catchupointing is detected in `MakeFull` then continue catching up
2. If one calls `StartCatchup` then `catchpointCatchupMode = true` and force all services to stop
3. When catchupointing is done then `catchpointCatchupMode = false` and start all the services on updated ledger.

agreement/fuzzer/networkFacade_test.go
```diff
@@ -258,7 +258,7 @@ func (n *NetworkFacade) RegisterHandlers(dispatch []network.TaggedMessageHandler
 
 // ClearHandlers
 func (n *NetworkFacade) ClearHandlers() {
-	n.mux.ClearHandlers()
+	n.mux.ClearHandlers([]network.Tag{})
```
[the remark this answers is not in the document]
tsachiherman: Hmm.. the counter argument is that there are only two that has the empty handers, and both are unit tests.
I'm not sure if adding two interface function to the multiplexer would simplify anything. Neither of the unit test doesn't really use the real "underlaying" network anyway.

agreement/pseudonode.go
```diff
@@ -118,7 +118,7 @@ type pseudonodeVerifier struct {
 //msgp:ignore verifiedCryptoResults
 type verifiedCryptoResults []asyncVerifyVoteResponse
 
-func makePseudonode(factory BlockFactory, validator BlockValidator, keys KeyManager, ledger Ledger, voteVerifier *AsyncVoteVerifier, log serviceLogger) pseudonode {
+func makePseudonode(factory BlockFactory, validator BlockValidator, keys KeyManager, ledger Ledger, voteVerifier *AsyncVoteVerifier, log serviceLogger, monitor *coserviceMonitor) pseudonode {
```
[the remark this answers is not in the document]
tsachiherman: As before, I'm ok with packing that into PseudonodeParams.

catchup/catchpointService.go
```diff
@@ -0,0 +1,508 @@
...
+	"github.com/algorand/go-algorand/data/basics"
+	"github.com/algorand/go-algorand/data/bookkeeping"
+	"github.com/algorand/go-algorand/ledger"
+	"github.com/algorand/go-algorand/logging"
+	"github.com/algorand/go-algorand/network"
+)
+
+// CatchpointCatchupNodeServices defines the extenal node support needed
+// for the catchpoint service to switch the node between "regular" operational mode and catchup mode.
+type CatchpointCatchupNodeServices interface {
```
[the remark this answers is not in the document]
tsachiherman: hmm. I tried to use the "Catchpoint" alone for all the server-side portion, and "CatchpointCatchup" for all the client-side. I'm definitely don't think it's the best name, but I'm afraid that using the "catchpoint" term too much would render the code unclear.

catchup/catchpointService.go
```diff
@@ -0,0 +1,508 @@
...
+	SetCatchpointCatchupMode(bool) (newCtx context.Context)
```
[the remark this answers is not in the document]
tsachiherman: I think that we need a method name which would be clear for the reader of the node.go source file that it's a callback from the catchpoint service. Yes, we can add as much documentation as needed, but I was hoping the name would be a good self-documenting one.
As for breaking this into two methods, I have no issue with that.

catchup/catchpointService.go
```diff
@@ -0,0 +1,508 @@
...
+	// maxLedgerDownloadAttempts is the number of attempts the catchpoint file will be downloaded and verified. Once we pass that number of attempts
+	// the catchpoint catchup is considered to be a failuire, and the node would resume to work as before.
+	maxLedgerDownloadAttempts = 50
+	// maxBlockDownloadAttempts is the number of block download attempts failuires allows while downloading the blocks. Once we pass that number of attempts
+	// the catchpoint catchup is considered to be a failuire, and the node would resume to work as before.
+	maxBlockDownloadAttempts = 50
+)
+
+// MakeCatchpointCatchupService creates a catchpoint catchup service for a node that is already in catchpoint catchup mode
+func MakeCatchpointCatchupService(ctx context.Context, node CatchpointCatchupNodeServices, log logging.Logger, net network.GossipNode, l *ledger.Ledger) (*CatchpointCatchupService, error) {
```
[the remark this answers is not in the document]
tsachiherman: sounds good.

catchup/catchpointService.go
```diff
@@ -0,0 +1,508 @@
...
+		},
+		node:           node,
+		ledgerAccessor: ledger.MakeCatchpointCatchupAccessor(l, log),
+		stage:          ledger.CatchpointCatchupStateInactive,
+		log:            log,
+		newService:     true,
+		net:            net,
+		ledger:         l,
+	}
+	if catchpoint == "" {
```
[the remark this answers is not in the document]
tsachiherman: yes

catchup/catchpointService.go
```diff
@@ -0,0 +1,508 @@
...
+			err = cs.processStageLastestBlockDownload()
+		case ledger.CatchpointCatchupStateBlocksDownload:
+			err = cs.processStageBlocksDownload()
+		case ledger.CatchpointCatchupStateSwitch:
+			err = cs.processStageSwitch()
+		default:
+			err = cs.abort(fmt.Errorf("unexpected catchpoint catchup stage encountered : %v", cs.stage))
+		}
+
+		if err != nil {
```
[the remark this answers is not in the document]
tsachiherman: One exception to what you wrote - 
In the case of err != nil and cs.ctx.Err() != nil, we still want to print out the error message from err.

catchup/catchpointService.go
```diff
@@ -0,0 +1,508 @@
...
+		attemptsCount++
+
+		err = cs.ledgerAccessor.ResetStagingBalances(cs.ctx, true)
+		if err != nil {
+			if err == cs.ctx.Err() {
+				return cs.abort(err) // we want to keep it with the context error.
+			}
+			return cs.abort(fmt.Errorf("processStageLedgerDownload failed to reset staging balances : %v", err))
+		}
+		err = ledgerFetcher.getLedger(cs.ctx, round)
```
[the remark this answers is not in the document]
tsachiherman: downloadLedger ?

catchup/catchpointService.go
```diff
@@ -0,0 +1,508 @@
...
+	for blocksFetched <= lookback {
+		attemptsCount++
+
+		if err := cs.ctx.Err(); err != nil {
+			return cs.abort(err)
+		}
+
+		blk = nil
+		// check to see if the current ledger might have this block. If so, we should try this first instead of downloading anything.
+		if ledgerBlock, err := cs.ledger.Block(topBlock.Round() - basics.Round(blocksFetched)); err == nil {
```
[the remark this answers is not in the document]
tsachiherman: `ledger.ErrNoEntry` can be ignored for sure. I'm a little bit more hesitant about other issues.
I'll add a warning if we fail due to error other than ledger.ErrNoEntry

catchup/ledgerFetcher.go
tsachiherman: it can be really big when we have millions of accounts. I'll be doing some more precise calculation later on.
The long term design is to use ranged http requests so that we can do resumes, but we need to figure out how to persist the tar/gzip decoder states first.

catchup/service.go
tsachiherman: so that we can Start/Stop multiple times. The above auth is being generated by the node, and there is no need to stop/resume it. we can just stop using it. The Quit was moved to the AlgorandFullNode.Stop()

cmd/algod/main.go
```diff
@@ -248,6 +249,17 @@ func main() {
 		if cfg.GossipFanout > len(peerOverrideArray) {
 			cfg.GossipFanout = len(peerOverrideArray)
 		}
+
```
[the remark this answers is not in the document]
tsachiherman: true, it was made during the testing of the catchup, after I learned that passing invalid parameters to algod was still *partially* working. I'll move this into a different PR.

ledger/catchupaccessor.go
tsachiherman: yes!

ledger/catchupaccessor.go
tsachiherman: yes.

ledger/catchupaccessor.go
tsachiherman: yes

node/node.go
tsachiherman: sure.

node/node.go
tsachiherman: that sounds correct.

rpcs/wsFetcherService.go
tsachiherman: We don't use "unregister". Instead, we call "ClearHandlers" and remove all of them.

daemon/algod/api/server/v2/handlers.go
```diff
@@ -443,6 +501,40 @@ func (v2 *Handlers) getPendingTransactions(ctx echo.Context, max *uint64, format
...
+	err = v2.Node.AbortCatchup(catchpoint)
+	if err != nil {
+		return internalError(ctx, err, fmt.Sprintf(errFailedToAbortCatchup, err), v2.Log)
+	}
+
+	return ctx.JSON(http.StatusOK, private.CatchpointAbortResponse{
+		CatchupMessage: catchpoint,
+	})
+}
```
rotemh, line 536: These two, including the Shutdown call are not RESTful endpoints. Given these private endpoints grow more than anticipated, we should consider switching to a separate thing Maybe gRPC.

tsachiherman: I would be glad to do so, but not in the context of this PR. Preferably, we would have a PR going in ahead of this one, which would introduce the gRPC framework and having this PR extends that functionality.

daemon/algod/api/algod.oas2.json
tsachiherman: Rotem think it needs to be moved into a gRPC like interface instead of REST. If we're going to do so, I'm going to leave it like that for now.  (since it's going to be deleted anyway )

ledger/blockqueue.go
```diff
@@ -67,14 +68,20 @@ func bqInit(l *Ledger) (*blockQueue, error) {
 
 func (bq *blockQueue) close() {
 	bq.mu.Lock()
-	defer bq.mu.Unlock()
+	defer func() {
+		bq.mu.Unlock()
+		<-bq.closed
```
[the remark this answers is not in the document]
tsachiherman: great question. I will add a comment!

ledger/catchpointwriter.go
```diff
@@ -0,0 +1,299 @@
...
+type encodedBalanceRecord struct {
+	_struct     struct{}  `codec:",omitempty,omitemptyarray"`
+	Address     []byte    `codec:"pk"`
+	AccountData codec.Raw `codec:"ad"`
+}
+
+type catchpointFileHeader struct {
+	_struct           struct{}      `codec:",omitempty,omitemptyarray"`
+	Version           uint64        `codec:"version"`
+	BalancesRound     basics.Round  `codec:"balancesRound"`
```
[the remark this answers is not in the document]
tsachiherman: Strictly speaking - No. We might be able to avoid storing both BalancesRound as well as BlocksRound. However, it helps to have all the data in one place. Given that the storage cost is not really meaningful, I have included few additional fields here which I think could become useful. For instance Catchpoint or BlockHeaderDigest.
These could be used by external tools to verify the correctness and metadata of this file.
Note that the file is designed to be binary *identical* across machines, which is why it doesn't contain any time signature or machine-specific metadata.

ledger/catchpointwriter.go
```diff
@@ -0,0 +1,299 @@
...
+				return
+			}
+			_, err = cw.tar.Write(encodedChunk)
+			if err != nil {
+				return
+			}
+
+			if len(cw.balancesChunk) < balancesChunkReadSize || cw.balancesChunkNum == cw.fileHeader.TotalChunks {
+				cw.tar.Close()
+				cw.gzip.Close()
```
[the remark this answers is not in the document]
tsachiherman: look at `cw.tar = tar.NewWriter(cw.gzip)`

ledger/ledger.go
tsachiherman: done, thanks.

ledger/totals.go
tsachiherman: thanks.

network/multiplexer.go
tsachiherman: done.

network/multiplexer.go
tsachiherman: done.

node/node.go
tsachiherman: Some of them might. For example, you don't want the network to send messages to the tx handler while it's being stopped.  (which is why we clear the handlers before ).
I'm sure there are other cases as well.
Generally speaking, I haven't modified the order unless there was a reason to do so.

crypto/merkletrie/node.go
```diff
@@ -0,0 +1,383 @@
...
+		for i := idiff - 1; i >= 0; i-- {
+			// create a parent node for pnode.
+			pnode2, nodeID2 := cache.allocateNewNode()
+			pnode2.leaf = false
+			pnode2.children = make([]storedNodeIdentifier, 256)
+			pnode2.childrenNext = make([]byte, 256)
+			pnode2.children[d[i]] = nodeID
+			pnode2.firstChild = d[i]
+			pnode2.childrenNext[d[i]] = d[i]
+			pnode2.hash = append(path, d[:i]...)
```
algoradam, line 125: If `path` has enough capacity, then all the `pnode2.hash` slices are going to have the same backing array. Is this intentional?

algoradam: (I don't see any immediate problems arising from this, but having multiple nodes whose `hash` fields all point to the same array seems likely to cause subtle problems if we ever write to that array or call `append(node.hash, x)` anywhere, so if it isn't intentional then we probably shouldn't do it)

crypto/merkletrie/node.go
```diff
@@ -0,0 +1,383 @@
...
+	for {
+		childNode, err := cache.getNode(n.children[i])
+		if err != nil {
+			return err
+		}
+		lenMod := 1
+		if !childNode.leaf {
+			lenMod = 128
+		}
+		hashAccumulator = append(hashAccumulator, byte(len(childNode.hash)+lenMod)) // we add this string length before the actual string so it could get "decoded"; in practice, it makes a good domain separator.
```
algoradam, line 206: Is it possible for `len(childNode.hash)` to be 128 or more? If so this can wrap around.

crypto/merkletrie/node.go
tsachiherman: That's a good observation. I think that in the current usage it's advantageous.

crypto/merkletrie/node.go
tsachiherman: That's true. We might want to target this implementation for "shorter" elements. ( i.e. right now we're using a 36 byte elements. ) having longer element also make the worst case node much deeper.

crypto/merkletrie/node.go
```diff
@@ -0,0 +1,383 @@
...
+
+import (
+	"bytes"
+	"encoding/binary"
+
+	"github.com/algorand/go-algorand/crypto"
+)
+
+type node struct {
+	hash         []byte
```
algoradam, line 27: Can you add a comment describing what node.hash is supposed to be? Is it the full path to that node before you call `calculateHash` and then the hash-of-child-hashes afterward?

tsachiherman: it's a tricky one for sure. It can be one of 3 things:
as a leaf, it's the residual element.
as a non-leaf, before calling commit ( which internally calls calculateHash), it's the path to that node.
as a non-leaf, after calling commit, it's the hash of the children plus the path as implemented in calculateHash.

crypto/merkletrie/node.go
```diff
@@ -0,0 +1,383 @@
...
+	pnode.hash = path
+	return nodeID, nil
+}
+
+// calculateHash calculate the hash of the non-leaf nodes
+// when this function is called, the hashes of all the child node are expected
+// to have been calculated already. This is achived by doing the following:
+// 1. all node id allocations are done in incremental monolitic order, from the bottom up.
+// 2. hash calculations are being doing in node id incremental ordering
+func (n *node) calculateHash(cache *merkleTrieCache) error {
```
algoradam, line 188: Is the following correct?
Hash of a leaf node is the suffix of the value it represents that isn't already determined by its position in the trie. (i.e., path to leaf node + hash of leaf node = value the leaf node represents)
Hash of a middle node is crypto.Hash of the following:
* the length of the path to this node, as a byte
* the path to this node
* for each child node:
    * the length of the hash of that child + 1 if the child is not a leaf, or the length of the hash of that child + 128 if it is a leaf
    * the index of that child, a byte (i.e, the next byte on the path to the child)
    * the hash of the child

Is my understanding correct?

crypto/merkletrie/node.go
```diff
@@ -0,0 +1,383 @@
...
+	if n.leaf {
+		return nil
+	}
+	path := n.hash
```
algoradam, line 192: Since `calculateHash` overwrites `n.hash`, this means we can never call `calculateHash` on the same (non-leaf) node twice, or the output will be wrong. Is this the intended behavior?

crypto/merkletrie/node.go
tsachiherman: almost.
two more things to note: 
1. In the for each child node, it's also according to the incremental order of the child node index.
2. the length of the element in the trie is always identical. It means that all leaf children would have the precise same length.

crypto/merkletrie/node.go
tsachiherman: yes. It's not an issue because all the nodes in the trie are immutable. If a non-leaf node need to be modified, a new node would get created to replace it. ( it's also required as part of the update-changes only )

rpcs/blockService.go
```diff
@@ -0,0 +1,361 @@
...
+
+// Stop servicing catchup requests over ws
+func (ls *BlockService) Stop() {
+	close(ls.stop)
+}
+
+// ServerHTTP returns blocks
+// Either /v{version}/block/{round} or ?b={round}&v={version}
+// Uses gorilla/mux for path argument parsing.
+func (ls *BlockService) ServeHTTP(response http.ResponseWriter, request *http.Request) {
```
algonautshant, line 114: Consider changing ls -> bs for blockService instead of ledgerService.

rpcs/blockService.go
```diff
@@ -0,0 +1,361 @@
...
+// MakeBlockService creates a BlockService around the provider Ledger and registers it for RPC with the provided Registrar
+func MakeBlockService(config config.Local, ledger *data.Ledger, net network.GossipNode, genesisID string) *BlockService {
+	service := &BlockService{
+		ledger:        ledger,
+		genesisID:     genesisID,
+		catchupReqs:   make(chan network.IncomingMessage, config.CatchupParallelBlocks*blockServerCatchupRequestBufferSize),
+		net:           net,
+		enableService: config.EnableBlockService,
+	}
+	if service.enableService {
```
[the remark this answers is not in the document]
tsachiherman: By avoiding registering the handler, we're preventing the HTTP entry point from working.
This is effectively disabling the service. We should have had that functionality, but it was never implemented. I expect that in the future we would want to deploy relays that doesn't support backing block storage, and it's going to be useful there.

network/multiplexer.go
```diff
@@ -85,19 +85,26 @@ func (m *Multiplexer) RegisterHandlers(dispatch []TaggedMessageHandler) {
 	m.msgHandlers.Store(mp)
 }
 
-// ClearHandlers deregisters all the existing message handlers.
+// ClearHandlers deregisters all the existing message handlers other then the one provided in the excludeTags list
```
algonautshant, line 88: typo: then->than

catchup/ledgerFetcher.go
```diff
@@ -0,0 +1,160 @@
...
+
+func (lf *ledgerFetcher) downloadLedger(ctx context.Context, round basics.Round) error {
+	if len(lf.peers) == 0 {
+		lf.peers = lf.net.GetPeers(network.PeersPhonebook)
+		if len(lf.peers) == 0 {
+			return fmt.Errorf("no peers are available")
+		}
+	}
+	// use the first one -
+	for {
```
algonautshant, line 67: Why is a for loop needed here?

tsachiherman: you're correct. This for loop is not needed.

catchup/ledgerFetcher.go
```diff
@@ -0,0 +1,160 @@
...
+
+	// at this point, we've already receieved the response headers. ensure that the
+	// response content type is what we'd like it to be.
+	contentTypes := response.Header["Content-Type"]
+	if len(contentTypes) != 1 {
+		err = fmt.Errorf("http ledger fetcher invalid content type count %d", len(contentTypes))
+		return err
+	}
+
+	if contentTypes[0] != rpcs.LedgerResponseContentType {
```
algonautshant, line 116: This is okay for now. 
We may need to de-couple it from the ledger service in case the ledger is upgraded and fetching from an older node.

crypto/merkletrie/node.go
algoradam: I don't think this gives us the security properties we want. Say you give me a trie with RootHash `h` where the root node is not a leaf. I can construct a different trie that also has RootHash `h` as follows: the root node is a leaf, and the value it represents is (the hash you get from calling `calculateHash` on the root node of the trie you gave me). If I've understood correctly these tries will then both have the same RootHash.

Is it required that `calculateHash` leave the hash of leaf nodes unchanged? If it's allowed to change the hash of leaf nodes you could instead do something like the following:
* Hash of a leaf node is `concat(0, value it represents)`
* Hash of a non-leaf node node is `concat(1, (what calculateHash already does but without lenMod))`
This way node.hash can't be the same for leaves and non-leaves.

If you're not allowed to modify the hash of leaf nodes, you can have trie.RootHash distinguish between when the root is a leaf and when it isn't (e.g. return `crypto.Hash(concat(pnode.leaf ? 0 : 1, pnode.hash))`). I think also in `calculateHash` instead of leaves adding 128 vs 1 to the length field, there needs to be a separate byte for whether it's a leaf or not. Otherwise I can cause wraparound by giving you a tree with leaves of length `127+(length of a crypto.Digest)`. Element length isn't committed to in the RootHash, and the wraparound will make the parents of my leaves hash as if my nodes were non-leaves, so I'm pretty sure this can lead to a collision.

crypto/merkletrie/trie.go
```diff
@@ -0,0 +1,282 @@
...
+		if mt.cache.modified {
+			return 0, ErrUnableToEvictPendingCommits
+		}
+	}
+	return mt.cache.evict(), nil
+}
+
+// serialize serializes the trie root
+func (mt *Trie) serialize() []byte {
+	serializedBuffer := make([]byte, (8+1)*4) // allocate the worst-case scenario for the trie header.
```
algoradam, line 236: I think this should be `(8+1)*5` or there might not be room for `pageSizeLength`.

ledger/catchupaccessor.go
```diff
@@ -0,0 +1,490 @@
...
+	}
+	// we want to allow undefined sections to support backward compatibility.
+	c.log.Warnf("CatchpointCatchupAccessor::ProgressStagingBalances encountered unexpected section name '%s' of length %d, which would be ignored", sectionName, len(bytes))
+	return nil
+}
+
+// processStagingContent deserialize the given bytes as a temporary staging balances content
+func (c *CatchpointCatchupAccessor) processStagingContent(ctx context.Context, bytes []byte, progress *CatchpointCatchupAccessorProgress) (err error) {
+	if progress.SeenHeader {
+		return fmt.Errorf("content chunk already seen")
```
algonautshant, line 187: Shall we add context to the error message?

rpcs/ledgerService.go
```diff
@@ -17,89 +17,62 @@
...
 
-// Start listening to catchup requests over ws
+// Start listening to catchup requests
 func (ls *LedgerService) Start() {
-	go ls.ListenForCatchupReq(ls.catchupReqs, ls.stop)
 }
 
-// Stop servicing catchup requests over ws
+// Stop servicing catchup requests
 func (ls *LedgerService) Stop() {
```
algonautshant, line 73: ledgerService Start/Stop is called from node., but start and stop are empty. 
Adding a comment in start and stop to explain why will help.

rpcs/ledgerService.go
```diff
@@ -110,19 +83,19 @@ func (ls *LedgerService) ServeHTTP(response http.ResponseWriter, request *http.R
 	genesisID, hasGenesisID := pathVars["genesisID"]
 	if hasVersionStr {
 		if versionStr != "1" {
-			logging.Base().Debug("http block bad version", versionStr)
+			logging.Base().Debug("http ledger bad version", versionStr)
 			response.WriteHeader(http.StatusBadRequest)
```
algonautshant, line 87: The StatusBadRequest and other failure codes do not provide much information. 
Can we return the error message with the response to help the client identify the issue?
We can we use something like ErrorResponse.

rpcs/ledgerService.go
```diff
@@ -17,89 +17,62 @@
...
+// MakeLedgerService creates a LedgerService around the provider Ledger and registers it with the HTTP router
+func MakeLedgerService(config config.Local, ledger *data.Ledger, net network.GossipNode, genesisID string) *LedgerService {
+	service := &LedgerService{
+		ledger:        ledger,
+		genesisID:     genesisID,
+		net:           net,
+		enableService: config.EnableLedgerService,
+	}
+	if service.enableService {
+		net.RegisterHTTPHandler(LedgerServiceLedgerPath, service)
```
algonautshant, line 63: Is it registered for: ?b={round}&v={version}

crypto/merkletrie/trie.go
tsachiherman: good point.

rpcs/ledgerService.go
tsachiherman: The header need to remain StatusBadRequest; however, we can return whatever information we want in the response body.

ledger/catchpointwriter.go
```diff
@@ -0,0 +1,299 @@
...
+	// have we timed-out / canceled by that point ?
+	select {
+	case <-ctx.Done():
+		err = ctx.Err()
+		if err == context.DeadlineExceeded {
+			return true, nil
+		}
+		return
+	default:
+	}
```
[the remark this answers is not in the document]
tsachiherman: good idea.

ledger/catchpointwriter.go
```diff
@@ -0,0 +1,299 @@
...
+		// have we timed-out / canceled by that point ?
+		select {
+		case <-ctx.Done():
+			err = ctx.Err()
+			if err == context.DeadlineExceeded {
+				return true, nil
+			}
+			return
+		default:
+		}
```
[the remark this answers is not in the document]
tsachiherman: true.

ledger/catchpointlabel.go
```diff
@@ -0,0 +1,97 @@
...
+	copy(buffer[:], l.ledgerRoundBlockHash[:])
+	copy(buffer[crypto.DigestSize:], l.balancesMerkleRoot[:])
+	encodedTotals := protocol.EncodeReflect(&l.totals)
+	copy(buffer[crypto.DigestSize*2:], encodedTotals)
+	return crypto.Hash(buffer[:crypto.DigestSize*2+len(encodedTotals)])
+}
+
+// ParseCatchpointLabel parse the given label and breaks it into the round and hash components. In case of a parsing failuire,
+// the returned err is non-nil.
+func ParseCatchpointLabel(label string) (round basics.Round, hash crypto.Digest, err error) {
```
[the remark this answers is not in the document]
tsachiherman: yes, added.

rpcs/ledgerService.go
algonautshant: That will be very useful.

ledger/acctupdates.go
```diff
@@ -469,74 +1079,147 @@ func (au *accountUpdates) committedUpTo(rnd basics.Round) basics.Round {
...
+			return
+		}
+	}
+	if au.catchpointFileHistoryLength == -1 {
 		return
 	}
-
-	totals = au.roundTotals[offset]
+	var filesToDelete map[basics.Round]string
+	filesToDelete, err = au.accountsq.getOldestCatchpointFiles(context.Background(), 2, au.catchpointFileHistoryLength)
```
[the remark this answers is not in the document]
tsachiherman: if sysadmin/user want to add new entries, they would just copy the files to the correct location. Once files are there, the getCatchpointStream would know how to get them ( and add them to the database ), and saveCatchpointFile would ensure to keep the number of stored files at the desired level.
If we shrink the catchpointFileHistoryLength, we need to dynamically delete *more* than a single file. Deleting 2 files is the minimum required to eventually get where we want to be.

ledger/acctupdates.go
```diff
@@ -469,74 +1079,147 @@ func (au *accountUpdates) committedUpTo(rnd basics.Round) basics.Round {
...
+		irnd = irnd / 256
+	}
+	outStr = filepath.Join(outStr, strconv.FormatInt(int64(rnd), 10)+".catchpoint")
+	return outStr
 }
 
-func (au *accountUpdates) totals(rnd basics.Round) (totals AccountTotals, err error) {
-	offset, err := au.roundOffset(rnd)
-	if err != nil {
+func (au *accountUpdates) saveCatchpointFile(round basics.Round, fileName string, fileSize int64, catchpoint string) (err error) {
```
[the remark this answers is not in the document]
tsachiherman: The filename is always generated by catchpointRoundToPath. There is no "guessing" there..

ledger/catchpointwriter.go
```diff
@@ -0,0 +1,299 @@
...
+
+// GetBalancesRound returns the round number of the balances to which this catchpoint is generated for.
+func (cw *catchpointWriter) GetBalancesRound() basics.Round {
+	if cw.fileHeader != nil {
+		return cw.fileHeader.BalancesRound
+	}
+	return basics.Round(0)
+}
+
+// GetRound returns the catchpoint string to which this catchpoint file was generated for.
```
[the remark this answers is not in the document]
tsachiherman: will fix.

rpcs/ledgerService.go
```diff
@@ -161,189 +134,37 @@ func (ls *LedgerService) ServeHTTP(response http.ResponseWriter, request *http.R
...
+		io.Copy(response, cs)
 		return
 	}
-	// Else, if version == 2.1
-	defer func() {
-		target.Respond(ctx, reqMsg, respTopics)
-	}()
-
-	topics, err := network.UnmarshallTopics(reqMsg.Data)
+	decompressedGzip, err := gzip.NewReader(cs)
```
algonautshant, line 163: Is this necessary? Why not restrict only to gzip streams?
I worry that this may be very expensive, and overwhelm the server even within the DOS prevention filter limits.

ledger/acctupdates.go
```diff
@@ -376,60 +520,526 @@ func (au *accountUpdates) getAssetCreatorForRound(rnd basics.Round, aidx basics.
...
+	// if the database doesn't know about that round, see if we have that file anyway:
+	fileName := filepath.Join("catchpoints", catchpointRoundToPath(round))
+	catchpointPath := filepath.Join(au.dbDirectory, fileName)
+	file, err := os.OpenFile(catchpointPath, os.O_RDONLY, 0666)
+	if err == nil && file != nil {
+		// great, if found that we should have had this in the database.. add this one now :
+		fileInfo, err := file.Stat()
+		if err != nil {
+			// we couldn't get the stat, so just return with the file.
+			return file, nil
```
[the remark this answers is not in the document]
tsachiherman: The file *should* be a valid one. Only local admins have access to that file, so if it was placed there, and we can read it, it should be good enough. ( note that I don't expect to get this situation frequently, as we typically won't be able to open the file at all )

rpcs/ledgerService.go
tsachiherman: It is necessary to support "classic" http streams so that we could tunnel the stream through proxies and other storage providers in a http compliant way. Not all the storage providers and/or proxy/firewall providers supports compressed streams. By supporting both on either ends we're covering all the cases.
I don't think that the DoS risk here is that great. The load on the server is incremental - i.e. it will perform the decompression only as long as the other side is receiving the bytes.
Also, the compression rate here is not malicious ( since we've created this file), so that won't be an issue.

ledger/catchupaccessor.go
tsachiherman: sure.

rpcs/ledgerService.go
tsachiherman: added support for start/stop.

crypto/merkletrie/node.go
tsachiherman: updated.

rpcs/ledgerService.go
tsachiherman: I don't think that it need to be registered for query arguments.

rpcs/ledgerService.go
```diff
@@ -110,19 +83,19 @@ func (ls *LedgerService) ServeHTTP(response http.ResponseWriter, request *http.R
...
-			logging.Base().Debugf("http block bad genesisID mine=%#v theirs=%#v", ls.genesisID, genesisID)
+			logging.Base().Debugf("http ledger bad genesisID mine=%#v theirs=%#v", ls.genesisID, genesisID)
 			response.WriteHeader(http.StatusBadRequest)
 			return
 		}
 	} else {
-		logging.Base().Debug("http block no genesisID")
+		logging.Base().Debug("http ledger no genesisID")
 		response.WriteHeader(http.StatusBadRequest)
 		return
```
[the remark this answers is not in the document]
tsachiherman: yes.

rpcs/ledgerService.go
```diff
@@ -146,12 +119,12 @@ func (ls *LedgerService) ServeHTTP(response http.ResponseWriter, request *http.R
 		if ok {
```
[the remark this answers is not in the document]
tsachiherman: yes

ledger/catchupaccessor.go
tsachiherman: done.

ledger/totals.go
tsachiherman: added!

catchup/ledgerFetcher.go
```diff
@@ -0,0 +1,155 @@
...
+	var downloadProgress ledger.CatchpointCatchupAccessorProgress
+	for {
+		header, err := tarReader.Next()
+		if err != nil {
+			if err == io.EOF {
+				return nil
+			}
+			return err
+		}
+		balancesBlockBytes := make([]byte, header.Size)
```
algomaxj, line 127: I believe `header.Size` is an untrusted value and could cause a panic here if it is too large/potentially negative (it's an `int64`). I think some validation needs to be done here.

tsachiherman: That's an excellent observation @algomaxj; I'll address that shortly.

tsachiherman: done.

ledger/catchpointlabel.go
```diff
@@ -0,0 +1,97 @@
...
+		return
+	}
+	round = basics.Round(uintRound)
+	var hashBytes []byte
+	hashBytes, err = base32Encoder.DecodeString(splitted[1])
+	if err != nil {
+		return
+	}
+	if len(hashBytes) > crypto.DigestSize {
+		return
```
algomaxj, line 92: this should be setting an error, right?

tsachiherman: yes, fixed.

ledger/catchupaccessor.go
```diff
@@ -0,0 +1,495 @@
...
+	}
+	err = c.FinishBlalances(ctx)
+	if err != nil {
+		return err
+	}
+
+	return c.ledger.reloadLedger()
+}
+
+// FinishBlalances concludes the catchup of the balances(tracker) database.
```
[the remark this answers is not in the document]
tsachiherman: typo fixed.

agreement/service.go
```diff
@@ -117,9 +111,33 @@ func (s *Service) SetTracerFilename(filename string) {
 
 // Start executing the agreement protocol.
 func (s *Service) Start() {
```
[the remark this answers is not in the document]
tsachiherman: The changes to the agreement package are focused around the ability to start/stop the service repeatedly. I'm not sure how easy it would be to extract these, as there are other dependencies which would required to be updated along. ( i.e. network and node are two examples.. )

catchup/ledgerFetcher.go
```diff
@@ -0,0 +1,165 @@
...
+		err = fmt.Errorf("getPeerLedger : http ledger fetcher invalid content type count %d", len(contentTypes))
+		return err
+	}
+
+	if contentTypes[0] != rpcs.LedgerResponseContentType {
+		err = fmt.Errorf("getPeerLedger : http ledger fetcher response has an invalid content type : %s", contentTypes[0])
+		return err
+	}
+
+	tarReader := tar.NewReader(response.Body)
```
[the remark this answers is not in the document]
tsachiherman: regarding the second one : https://github.com/golang/go/issues/13647

algomaxj:
I think `VerifyCatchpoint` and `processStageBlocksDownload` are doing insufficient validation of fetched blocks, allowing the insertion of invalid transactions into the block database.

They check `block.Digest()` and `blockHeader.Hash()`, but both of these values only commit to the block header -- not the transactions in the payset. I think you need to additionally call `block.ContentsMatchHeader` to avoid this.

ledger/acctupdates.go
```diff
@@ -376,60 +523,526 @@ func (au *accountUpdates) getAssetCreatorForRound(rnd basics.Round, aidx basics.
...
+		}
+
+		err = au.saveCatchpointFile(round, fileName, fileInfo.Size(), "")
+		if err != nil {
+			au.log.Warnf("accountUpdates: getCatchpointStream: unable to save missing catchpoint entry: %v", err)
+			err = nil
```
algomaxj, line 749: this assignment is not used

ledger/acctupdates.go
```diff
@@ -376,60 +523,526 @@ func (au *accountUpdates) getAssetCreatorForRound(rnd basics.Round, aidx basics.
...
+
 	if err != nil {
+		au.balancesTrie = nil
 		au.log.Warnf("unable to advance account snapshot: %v", err)
-		return au.dbRound
+		return
+	}
+	if isCatchpointRound {
+		committedRoundDigest = au.roundDigest[offset+uint64(lookback)-1]
+		catchpointLabel, err = au.accountsCreateCatchpointLabel(dbRound+basics.Round(offset)+lookback, dbRound+basics.Round(offset), committedRoundDigest)
```
algomaxj, line 1036: err is not being checked

ledger/acctupdates.go
tsachiherman: removed.

ledger/acctupdates.go
tsachiherman: added a logging around that.

tsachiherman:
> VerifyCatchpoint

good find. I've updated the code.

catchup/ledgerFetcher.go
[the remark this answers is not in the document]
tsachiherman: The mentioned LimitReader isn't what we're after, but the idea is correct -
we need a limiting reader with the following properties:
1. abort once a certain throughput has reached.
2. reset the throughput counter at the end of each chunk
3. time limit the amount per chunk

in addition, we also need to limit the time we would spend on retrieving the entire file ( by passing timeout context during the request creation )

catchup/catchpointService.go
```diff
@@ -0,0 +1,534 @@
...
+		if _, ok := config.Consensus[blk.BlockHeader.CurrentProtocol]; !ok {
+			cs.log.Warnf("processStageLastestBlockDownload: unsupported protocol version detected: '%v'", blk.BlockHeader.CurrentProtocol)
+			// try again.
+			blk = nil
+			continue
+		}
+
+		// check to see that the block header and the block payset aligns
+		if !blk.ContentsMatchHeader() {
+			cs.log.Warnf("processStageLastestBlockDownload: downloaded block content does not match downloaded block header")
```
algomaxj, line 315: Should this and the error check above be checking `attemptsCount`?

catchup/catchpointService.go
```diff
@@ -0,0 +1,534 @@
...
+			}
+			// success
+			client.Close()
+		}
+
+		cs.updateBlockRetrievalStatistics(1, 0)
+
+		// validate :
+		if prevBlock.BlockHeader.Branch != blk.Hash() {
+			// not identical, retry download.
```
algomaxj, line 424: shouldn't all of these error conditions be checking `attemptsCount`? otherwise this can loop forever

ledger/acctupdates.go
```diff
@@ -376,60 +523,537 @@ func (au *accountUpdates) getAssetCreatorForRound(rnd basics.Round, aidx basics.
...
+
+// functions below this line are all internal functions
+
+// accountHashBuilder calculates the hash key used for the trie by combining the account address and the account data
+func accountHashBuilder(addr basics.Address, accountData basics.AccountData, encodedAccountData []byte) []byte {
+	hash := make([]byte, 4+crypto.DigestSize)
+	// write out the lowest 32 bits of the reward base. This should improve the caching of the trie by allowing
+	// recent updated to be in-cache, and "older" nodes will be left alone.
+	for i, rewards := 3, accountData.RewardsBase; i >= 0; i, rewards = i-1, rewards>>8 {
+		hash[i] = byte(rewards & 0x255)
```
algomaxj, line 763: This should be `0xFF`

algomaxj: actually, you can probably just kill the bitwise and entirely: https://play.golang.org/p/7bgJ5VOYh_C

ledger/catchpointlabel.go
```diff
@@ -0,0 +1,98 @@
...
+func (l CatchpointLabel) String() string {
+	hash := l.Hash()
+	encodedHash := base32Encoder.EncodeToString(hash[:])
+	out := fmt.Sprintf("%d#%s", l.ledgerRound, encodedHash)
+	return out
+}
+
+// Hash return the hash portion of this catchpoint label
+func (l CatchpointLabel) Hash() crypto.Digest {
+	buffer := make([]byte, 2*crypto.DigestSize+256)
```
algomaxj, line 63: Instead of 256, this should be `len(encodedTotals)`. Also, the `AccountTotals` struct lacks codec tags, so if we ever rename fields in that struct, this will break.

ledger/acctupdates.go
```diff
@@ -376,60 +523,537 @@ func (au *accountUpdates) getAssetCreatorForRound(rnd basics.Round, aidx basics.
...
+	var added, deleted bool
+	for addr, delta := range accountsDeltas {
+		if !delta.old.IsZero() {
+			deleteHash := accountHashBuilder(addr, delta.old, protocol.Encode(&delta.old))
+			deleted, err = au.balancesTrie.Delete(deleteHash)
+			if err != nil {
+				return err
+			}
+			if !deleted {
+				au.log.Warnf("failed to delete hash '%v' from merkle trie", deleteHash)
```
algomaxj, line 869: here and in `if !added`, it seems like you should return an error, since it indicates that your data structure is corrupted.

ledger/catchpointwriter.go
```diff
@@ -0,0 +1,292 @@
...
+const (
+	// BalancesPerCatchpointFileChunk defines the number of accounts that would be stored in each chunk in the catchpoint file.
+	// note that the last chunk would typically be less than this number.
+	BalancesPerCatchpointFileChunk = 512
+
+	// catchpointFileVersion is the catchpoint file version
+	catchpointFileVersion = uint64(0200)
+
+	// MaxEncodedAccountDataSize is a rough estimate for the worst-case scenario we're going to have of the account data and address serialized.
+	MaxEncodedAccountDataSize = 64 * 1024
```
algomaxj, line 47: This is too small. An account with 1,000 created assets could use, just for asset params (looking at the `AssetParams` struct), `1000 * (8 + 4 + 1 + 8 + 32 + 32 + 32 + 32 + 32 + 32 + 32) == 245,000` bytes. I think 500k would be a safe number for today's system.

catchup/ledgerFetcher.go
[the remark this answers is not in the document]
tsachiherman: added watchdogStreamReader to handle the above needs.

catchup/catchpointService.go
tsachiherman: yes, fixed.

catchup/catchpointService.go
tsachiherman: yes, fixed.

ledger/acctupdates.go
tsachiherman: I have no idea how you managed to pick 12345678, where 12345678 & 255 is 78.
i.e. the  12345600 is divisible by 256. what are the chances for that ?!

ledger/catchpointlabel.go
tsachiherman: I can't go all the way here with MicroAlgos codec-ing, as it would break the current AccountData.

tsachiherman: done.

ledger/acctupdates.go
tsachiherman: I was thinking about it. I'm not sure if we want to.

ledger/catchpointwriter.go
tsachiherman: I will create a unit test to measure this number.

tsachiherman: The actual number is surprisingly high : 324205; I have added a unit test to verify that.

algomaxj:
Ah, something I just realized -- up until this PR `basics.AccountData` has only come from trusted sources. As such the `allocbound` for some fields in that struct is `-`. Those `allocbound`s need to be set properly now to avoid crashing a node on a malicious encoded `AccountData`.

catchup/catchpointService.go
```diff
@@ -0,0 +1,558 @@
...
+			case ledger.ErrNoEntry:
+				// this is expected, ignore this one.
+			default:
+				cs.log.Warnf("processStageBlocksDownload encountered the following error when attempting to retrieve the block for round %d : %v", topBlock.Round()-basics.Round(blocksFetched), err)
+			}
+		}
+
+		if blk == nil {
+			fetcher := fetcherFactory.New()
+			blk, _, client, err = fetcher.FetchBlock(cs.ctx, topBlock.Round()-basics.Round(blocksFetched))
```
algoradam, line 413: Would someone ever try to use a catchpoint to catch up to a really early round? If so, if we're trying to catch up to a round less than `lookback`, then after we've fetched block 0, `topBlock.Round() - basics.Round(blocksFetched)` will underflow and we'll end up aborting because we'll never fetch block 2^64-1.
(Not sure if this actually matters)

ledger/catchupaccessor.go
```diff
@@ -0,0 +1,493 @@
...
+	err = wdb.Atomic(func(tx *sql.Tx) (err error) {
+		sq, err := accountsDbInit(tx, tx)
+		if err != nil {
+			return fmt.Errorf("CatchpointCatchupAccessor::processStagingContent: unable to initialize accountsDbInit: %v", err)
+		}
+		_, err = sq.writeCatchpointStateUint64(ctx, catchpointStateCatchupBlockRound, uint64(fileHeader.BlocksRound))
+		if err != nil {
+			return fmt.Errorf("CatchpointCatchupAccessor::processStagingContent: unable to write catchpoint catchup state '%s': %v", catchpointStateCatchupBlockRound, err)
+		}
+		_, err = sq.writeCatchpointStateUint64(ctx, catchpointStateCatchupBalancesRound, uint64(fileHeader.BalancesRound))
```
algomaxj, line 211: I can't find anywhere that `fileHeader.BalancesRound` is validated/checked against the catchpoint

crypto/merkletrie/trie.go
```diff
@@ -0,0 +1,286 @@
...
+		if mt.cache.modified {
+			return 0, ErrUnableToEvictPendingCommits
+		}
+	}
+	return mt.cache.evict(), nil
+}
+
+// serialize serializes the trie root
+func (mt *Trie) serialize() []byte {
+	serializedBuffer := make([]byte, (8+1)*5) // allocate the worst-case scenario for the trie header.
```
algomaxj, line 240: In the worst case, `PutUvarint` can use 10 bytes, not 9. See https://play.golang.org/p/81WJFN4E30f

ledger/acctupdates.go
```diff
@@ -376,60 +523,535 @@ func (au *accountUpdates) getAssetCreatorForRound(rnd basics.Round, aidx basics.
...
+				break
+			}
+			accountIdx += trieRebuildAccountChunkSize
+		}
+	}
+	au.balancesTrie = trie
+	return rnd, nil
+}
+
+func (au *accountUpdates) accountsUpdateBalaces(accountsDeltas map[basics.Address]accountDelta) (err error) {
```
algomaxj, line 854: `accountsUpdateBalances`

crypto/merkletrie/trie.go
tsachiherman: todo : use binary.MaxVarintLen64

catchup/catchpointService.go
tsachiherman: I think that it would be a good practice to test that before going into the loop - not because it would cause an issue, but rather to early detect fail cases.

ledger/catchupaccessor.go
tsachiherman: you're correct. It should be validated against the BlockRound - lookback for the BlockRound consensus protocol params.

ledger/acctupdates.go
tsachiherman: done.

crypto/merkletrie/trie.go
tsachiherman: done

ledger/catchupaccessor.go
tsachiherman: I've avoided using the field altogether. instead, I have calculated the balancesRound and stored it, avoiding any external influence.

catchup/catchpointService.go
tsachiherman: yes, I've adjusted the range of blocksFetched so that it won't reach 0 or below.

ledger/catchpointwriter.go
```diff
@@ -0,0 +1,292 @@
...
+	balancesChunkNum  uint64
+	writtenBytes      int64
+	blocksRound       basics.Round
+	blockHeaderDigest crypto.Digest
+	label             string
+}
+
+type encodedBalanceRecord struct {
+	_struct     struct{}  `codec:",omitempty,omitemptyarray"`
+	Address     []byte    `codec:"pk,allocbound=crypto.DigestSize"`
```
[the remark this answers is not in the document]
tsachiherman: done.

ledger/catchupaccessor.go
```diff
@@ -0,0 +1,495 @@
...
+	return nil
+}
+
+// processStagingContent deserialize the given bytes as a temporary staging balances content
+func (c *CatchpointCatchupAccessor) processStagingContent(ctx context.Context, bytes []byte, progress *CatchpointCatchupAccessorProgress) (err error) {
+	if progress.SeenHeader {
+		return fmt.Errorf("CatchpointCatchupAccessor::processStagingContent: content chunk already seen")
+	}
+	var fileHeader catchpointFileHeader
+	err = protocol.DecodeReflect(bytes, &fileHeader)
```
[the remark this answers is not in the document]
tsachiherman: switched to new msgp codec.

ledger/acctupdates.go
tsachiherman: I'm closing this conversation, but please do let me know if you think it's an issue.

ledger/accountdb.go
```diff
@@ -55,20 +64,134 @@ var accountsSchema = []string{
...
 		asset integer primary key,
 		creator blob)`,
+	`CREATE TABLE IF NOT EXISTS storedcatchpoints (
+		round integer primary key,
+		filename text NOT NULL,
+		catchpoint text NOT NULL,
+		filesize size NOT NULL,
+		pinned integer NOT NULL)`,
+	`CREATE TABLE IF NOT EXISTS accounthashes (
+			id integer primary key,
```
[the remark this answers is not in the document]
tsachiherman: sure, fixed.

ledger/accountdb.go
```diff
@@ -216,6 +394,109 @@ func (qs *accountsDbQueries) lookup(addr basics.Address) (data basics.AccountDat
 	return
 }
 
+func (qs *accountsDbQueries) storeCatchpoint(ctx context.Context, round basics.Round, fileName string, catchpoint string, fileSize int64) (err error) {
+	err = db.Retry(func() (err error) {
+		_, err = qs.deleteStoredCatchpoint.ExecContext(ctx, round)
+
+		if err != nil || (fileName == "" && catchpoint == "" && fileSize == 0) {
```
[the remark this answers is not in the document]
tsachiherman: why ? the cpu time it would take to do that isn't substantial enough, and we need to delete the old stored catchpoint anyway.
I think that a better thing to do here is to figure out if we want to perform just delete vs update/insert statement. But I can't see how we get any practical performance gains here.

ledger/acctupdates.go
```diff
@@ -376,60 +523,535 @@ func (au *accountUpdates) getAssetCreatorForRound(rnd basics.Round, aidx basics.
...
+// wait until the catchpoint creation is done, so that the persistance of the catchpoint file would have an
+// uninterrupted view of the balances at a given point of time.
+func (au *accountUpdates) committedUpTo(committedRound basics.Round) (retRound basics.Round) {
+	var isCatchpointRound, hasMultipleIntermediateCatchpoint bool
+	var offset uint64
+	var dc deferedCommit
+	au.accountsMu.RLock()
+	defer func() {
+		au.accountsMu.RUnlock()
+		if dc.offset != 0 {
```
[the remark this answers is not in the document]
tsachiherman: exactly. But we also want to write to the channel only after releasing the lock, and I'd rather release the lock once instead of releasing it, writing to the channel and taking the lock again.

ledger/acctupdates_test.go
```diff
@@ -38,7 +38,7 @@ type mockLedgerForTracker struct {
 	log    logging.Logger
```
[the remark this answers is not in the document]
tsachiherman: We clearly could benefit from additional testings, however, be aware that the TestArchival and TestAcctUpdates already provide pretty good coverage of the functions.

ledger/acctupdates.go
```diff
@@ -88,25 +136,88 @@ type accountUpdates struct {
...
+	if cfg.CatchpointFileHistoryLength < -1 {
+		au.catchpointFileHistoryLength = -1
+	}
+}
+
+// loadFromDisk is the 2nd level initializtion, and is required before the accountUpdates becomes functional
+// The close function is expected to be call in pair with loadFromDisk
 func (au *accountUpdates) loadFromDisk(l ledgerForTracker) error {
+	var writingCatchpointRound uint64
+	au.accountsMu.Lock()
```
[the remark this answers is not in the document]
tsachiherman: Yes, that's a good idea.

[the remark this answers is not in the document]
tsachiherman: done.

ledger/acctupdates_test.go
```diff
@@ -507,3 +509,87 @@ func BenchmarkCalibrateCacheNodeSize(b *testing.B) {
 	}
 	trieCachedNodesCount = defaultTrieCachedNodesCount
 }
+
+func TestLargeAccountCountCatchpointGeneration(t *testing.T) {
```
algonautshant, line 513: Can you please add a summary describing what is tested here?

PR #971, opened 2020-04-15
Make algorelay handle domain names type conversions

tsachiherman:
## Summary

Algorelay needs to have the ability to convert a previously created A DNS record into a CNAME DNS record and vice versa.
This PR adds that ability as well as improve the error logging in case of an issue.


tsachiherman:
> Looks good.

thanks!

PR #977, opened 2020-04-17
Add `update_releases_page` to update releases.algorand.com

btoll:
A lot of this is from a script that bolson had written that is in the `go-algorand-ci` repo.

This builds the release `index.html` every time.


scripts/release/prod/update_releases_page
```diff
@@ -0,0 +1,176 @@
...
+                hv = hashes.get(hn)
+                if hv:
+                    ftext += '<div class="hash {}">{}</div>\n'.format(hn, hv)
+        if not hashes and not stext:
+            continue
+        tbody.append("<tr><td>{}</td><td>{}</td><td>{}</td></tr>\n".format(ftext, size, stext))
+    tbody.append("</tbody>\n\n")
+    tbody.append('<tbody><tr class="spacer"><td></td></tr></tbody>\n\n')
+
+    match_string = "<!--" + channel + "-->"
```
algobolson, line 150: this will probably work, but it could break in a bad way. If this were automated a runaway loop could leave a page full of garbage. There's nothing to prevent from adding the same record many times. I think a safer strategy would be to accumulate data to a separate structure that parses well (CSV file, JSON file, DynamoDB, ...) and then parse that and emit the whole index.html based on the source data.

scripts/release/prod/update_releases_page
```diff
@@ -0,0 +1,176 @@
...
+    tbody_string = "".join(tbody)
+
+    try:
+        with open("index.html", "r+") as fd:
+            contents = fd.readlines()
+            for index, line in enumerate(contents):
+                if match_string in line and tbody_string not in contents[index + 1]:
```
algobolson, line 157: `tbody_string not in contents` won't actually prevent duplicates becase tbody_string contains newlines and the file has been broken on newlines so the previous tbody_string will have been broken into many lines. Either put all of tbody on one line or do the big strategy change like in my comment above.

scripts/release/prod/update_releases_page
```diff
@@ -0,0 +1,176 @@
...
+                    contents.insert(index + 1, tbody_string)
```
algobolson, line 158: inserting into a list while it is being iterated over can break

scripts/release/prod/update_releases_page
btoll: Will change.

scripts/release/prod/update_releases_page
btoll: Will change.

scripts/release/prod/update_releases_page
btoll: Will change.

scripts/release/prod/releases_page/update_releases_page
```diff
@@ -0,0 +1,183 @@
...
+
+    for channel in ["stable", "beta"]:
+        staging_response = s3.list_objects_v2(Bucket=staging_bucket, Prefix="releases/" + channel + "/", MaxKeys=100)
+        release_sets = get_stage_release_set(staging_response)
+        releases_response = s3.list_objects_v2(Bucket=releases_bucket)
+        release_files = objects_by_fname(releases_response["Contents"])
+
+        table = []
+
+        for key in release_sets:
```
algobolson, line 138: `for key, rset in release_sets.items():`

scripts/release/prod/releases_page/update_releases_page
```diff
@@ -0,0 +1,183 @@
...
+            rset = release_sets[key]
+            hashftext = get_hashes_data(s3, rset)
+            fhashes = read_hashes(hashftext.splitlines())
+            files = release_set_files(rset)
+
+            def get_furl(fname, skey):
```
algobolson, line 144: either move this up to just after release_files is set, or move it up to file scope and add release_files as an argument

scripts/release/prod/releases_page/update_releases_page
btoll: Good to know, will change.

scripts/release/prod/releases_page/update_releases_page
btoll: Will change.

PR #979, opened 2020-04-17
Add v2 OpenAPI spec to the code and a driver script for regenerating the code.

winder:
## Summary

This is some leftover work from the recent API v2 endpoints. To allow others to update the API this adds spec definition file, a script, and some documentation about how things work.

I also ran the script to regenerate code, and fixed the compiler errors. The specfile had changed slightly since the last time.


daemon/algod/api/README.md
```diff
@@ -1,5 +1,31 @@
...
+The API is defined using [OpenAPI v2](https://swagger.io/specification/v2/) in **algod.oas2.yml**.
+
+## Updating the V2 REST API
+Prerequisit:
+You need to install our fork of **oapi-codegen**, available here: https://github.com/algorand/oapi-codegen/
+Run `go build` in the cmd/oapi-codegen. At this point I haven't been able to get `go get` to work.
+
+1. Document your changes by editing **algod.oas2.yml**
+2. Regenerate the endpoints by running **generate.sh**. The sources at **server/v2/generated/** will be updated.
+3. Update the implementation in **server/v2/handlers.go**. It is sometimes useful to consult **generated/routes.go** to make sure the handler properly implements **ServerInterface**.
```
tsachiherman, line 15: you might want to see if you can add some compile time test for that.

daemon/algod/api/generate.sh
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
+set -e
+
+rootdir=`dirname $0`
+pushd $rootdir
+
+# Convert v2 to v3
+curl -s -X POST "https://converter.swagger.io/api/convert" -H "accept: application/json" -H "Content-Type: application/json" -d @./algod.oas2.json  -o 3.json
```
tsachiherman, line 8: this would fail without internet connection. it might be acceptable compromise, as it's needed only when we regenerate the apis.

daemon/algod/api/server/v2/generated/types.go
```diff
@@ -27,9 +27,7 @@ type Account struct {
 	// \[apar\] parameters of assets created by this account.
 	//
 	// Note: the raw account uses `map[int] -> Asset` for this type.
-	CreatedAssets *[]Asset `json:"created-assets,omitempty"`
-
-	// AccountParticipation describes the parameters used by this account in consensus protocol.
+	CreatedAssets *[]Asset              `json:"created-assets,omitempty"`
```
tsachiherman, line 30: why did you deleted the description for AccountParticipation ?

winder: I'm not sure why some comments were deleted. They're still in the spec. I'll need to look into this more.

daemon/algod/api/README.md
winder: Good idea. The spot where the routes are registered requires a `ServerInterface` implementation, so there would already be a compiler error here:
https://github.com/algorand/go-algorand/blob/a566ae36962c1581316e9a5ae894094365a3a4df/daemon/algod/api/server/router.go#L142

daemon/algod/api/README.md
```diff
@@ -1,5 +1,31 @@
...
 
+# V2 Endpoint
+With the V2 REST API we started using a design driven process.
+
+The API is defined using [OpenAPI v2](https://swagger.io/specification/v2/) in **algod.oas2.yml**.
+
+## Updating the V2 REST API
+Prerequisit:
+You need to install our fork of **oapi-codegen**, available here: https://github.com/algorand/oapi-codegen/
+Run `go build` in the cmd/oapi-codegen. At this point I haven't been able to get `go get` to work.
```
rotemh, line 11: "At this point I haven't been able to get `go get` to work." What do you expect the user to do with this comment? :) Consider deleting it or getting `go get` to work.

pbennett: If there is a need to have 'tools' installed and go modules are used a common pattern is to define a file like tools.go in the module root with a build tag that prevents its use in package builds.
The fetching of packages by the module system will still honor dependencies listed within it though and can be used for automatic dependencies and fetching of tools.

tools.go
----
```
// +build tools

package tools

import (
        _ "github.com/algorand/oapi-codegen"
)
```

tsachiherman: some of our tools are currently being installed during configure-deps.sh; would that work for this tool as well ?

winder: Brian was able to fix go get on Friday. Thanks for all the feedback on this, I'll get it resolved tomorrow.

go.sum
```diff
@@ -38,9 +37,13 @@ github.com/gen2brain/beeep v0.0.0-20180718162406-4e430518395f h1:eyHMPp7tXlBMF8P
 github.com/gen2brain/beeep v0.0.0-20180718162406-4e430518395f/go.mod h1:GprdPCZglWh5OMcIDpeKBxuUJI+fEDOTVUfxZeda4zo=
 github.com/getkin/kin-openapi v0.2.0 h1:PbHHtYZpjKwZtGlIyELgA2DploRrsaXztoNNx9HjwNY=
 github.com/getkin/kin-openapi v0.2.0/go.mod h1:V1z9xl9oF5Wt7v32ne4FmiF1alpS4dM6mNzoywPOXlk=
+github.com/getkin/kin-openapi v0.3.0/go.mod h1:W8dhxZgpE84ciM+VIItFqkmZ4eHtuomrdIHtASQIqi0=
+github.com/getkin/kin-openapi v0.3.1 h1:9mLtayAmieqUnNACL0HqHbxkTc+z1+15sxXpLoJOGEQ=
+github.com/getkin/kin-openapi v0.3.1/go.mod h1:W8dhxZgpE84ciM+VIItFqkmZ4eHtuomrdIHtASQIqi0=
```
tsachiherman, line 42: Could you see why you need *all* these github.com/getkin/kin-openapi versions ?

daemon/algod/api/generate.sh
winder: The expectation is that the REST endpoints will change infrequently, and only during development. So this wont be a significant burden.

daemon/algod/api/server/v2/generated/types.go
winder: Found the issue, the codegen tool seems to need the file extension to be `yml` even if it's actually a `json` file. For now I've left the file extension as `yml` so that the comments get generated.

go.sum
winder: fixed

daemon/algod/api/README.md
winder: fixed.

winder:
@tsachiherman @rotemh I believe the problems you found have been addressed.

algobolson:
Rename `daemon/algod/api/algod.oas3.yml` it's actually json

winder:
@algobolson also there is a bug in the codegen tool if it has the correct extension

PR #981, opened 2020-04-19
Change account balance, rewards, and changeonlinestatus commands to support default account as well as human-friendly alias.

pbennett:
## Summary

It's a trivial thing but I noticed several commands I was using didn't support the 'human-friendly name' of accounts, nor default accounts.  Being able to use things like clerk send with an account alias, but not account commands where the alias was created was confusing to me as a new user.


JasonWeathersby:
Thanks for the PR. If you have other feedback on the new user experience we would love to hear about it.

pbennett:
Obviously they were trivial changes but they struck me as odd omissions given how the account CLi seemed intended to work.  If users are going to use the goal apis for account manipulation it seemed odd to allow aliasing yet to then require full account addresses for some commands that could trivially use the provided aliases.   Same argument for 'default' accounts that were ignored as defaults.
Beyond that, there are certainly a lot of things that could be enhanced in the CLI but other venues may be better for that discussion.

cmd/goal/account.go
```diff
@@ -539,7 +542,11 @@ var rewardsCmd = &cobra.Command{
 	Run: func(cmd *cobra.Command, args []string) {
 		dataDir := ensureSingleDataDir()
 		client := ensureAlgodClient(dataDir)
-		response, err := client.AccountInformation(accountAddress)
+		accountList := makeAccountsList(dataDir)
+		if accountAddress == "" {
+			accountAddress = accountList.getDefaultAccount()
```
pzbitskiy, line 547: I think we need to print the address used if default is used.

cmd/goal/account.go
```diff
@@ -522,7 +520,12 @@ var balanceCmd = &cobra.Command{
...
+		}
+
```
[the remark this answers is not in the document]
pbennett: I just tried to follow the existing code which honored the account names without changing it and keep it simple but if you think this is something that makes sense to apply across the board, then absolutely.

cmd/goal/account.go
pbennett: Since the existing rewards command just outputs the rewards amount and nothing else (unlike account list) I didn't want to alter any existing output behaviors.  People could be parsing it for all I knew.   
My (simple) initial intent was just to make some of the commands behave the same as some of the other account commands that honored account aliases (and defaults) as it seemed an odd omission at the time.
If members of the team/community think it makes sense to alter the output I'd love to change it.
Were you thinking just something simple like what goal account list shows for the default account (*Default as the last column?)

pzbitskiy: > Since the existing rewards command just outputs the rewards amount and nothing else (unlike account list) I didn't want to alter any existing output behaviors. People could be parsing it for all I knew.

This is a good point.

cmd/goal/account.go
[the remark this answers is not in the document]
pbennett: I'll go ahead and update the PR, create a helper and switch all obvious account lookups to the new helper.

pbennett:
Sorry about that - messed things up w/ a rebase.  I'm going to close and re-open this PR with a clean delta.

pbennett:
Ah, no, sorry - I kind of gave up on the PR and it's parked somewhere on my machine because at the time the unit/integ tests never worked for me (and it was asked that I had expect tests) due to the Algorand code assuming a very old GOPATH style environment (pre modules).  I need to try again at some point as the friendly names should really be usable in all commands.

PR #984, opened 2020-04-20
Rename socket.sh -> forward_gpg_agent.sh

btoll:
Now, instead of `./socket ec2-13-57-188-227.us-west-1.compute.amazonaws.com` just do `./forward_gpg_agent.sh`.

Initially, I looked into using a new key for devops only use instead of an ephemeral key created upon creation of ec2 instance.  However, after spending some time, I've decided against this for the following reasons:
 
- There is too much work that would need to be done to the baseline build pipeline to order to support this (for instance, moving many lines of code back into various Jenkinsfiles), time that I think is better served creating the mule tasks that will supersede this pipeline. In the meantime, I've created a workaround that I think is acceptable.
- The baseline build release pipeline is intended to be, well, a baseline, and it is not perfect.  It's sole intention is to function as expected.


btoll:
I agree that the workaround isn't the best, but since this is the baseline pipeline, I would rather concentrate my efforts on the `mule` tasks that would supersede this.  This works, and that's the most important (just to be clear, I wouldn't feel that this kind of effort would be fine for `mule`).

Issue #863, opened 2020-02-26
Middleware auth extension 


rotemh:
the auth middleware should support multiple tokens. 
Each token should be able to access a specific pre defined set of routes.
A route can be defined to as Private by default. (i.e. no token should have access to it unless it was explicitly defined)

PR #987, opened 2020-04-22
Add private endpoints to the algod REST interface

tsachiherman:
## Add private endpoints to the algod REST interface
On V2, we have two sets of commands - one for "classic" node access and one for administrative access ( i.e. private ).
This PR separates the security token used for each one of these;
The private endpoints would be using the administrative token ( available in algod.admin.token ), and the classic endpoints would be using the regular token ( available in algod.token )

Resolves https://github.com/algorand/go-algorand/issues/863

## What's included in this PR
* The Shutdown command was implemented on goal, libgoal and restClient. It's a first pass-thru to ensure the entire pipeline is ready for private endpoints. Nevertheless, the server-side shutdown command is not yet functional.
* The pprof authentication was changed to use the admin token.
* The auth middleware is mapping the v1/classic v2 into the regular token, and the private API to use the admin token. It doesn't change any of the handlers implementation.
* Few small modifications to the generation script, allowing it work on Mac.

## Few editorial notes
* Previous attempts to implement this authentication via a middleware were not successful. When using Echo's middleware, the context resolves the path, but doesn't provide any mean to extract the originating route. Since our security paths are defined in the context of the originating route - using a middleware isn't the way to go.
Example:
When we define the endpoint `/v2/accounts/:address/transactions/pending`, we want to approve it for a subset of security tokens. However, when the middleware is being invoked, it will provide a path of `/v2/accounts/01234567801234567/transactions/pending`, which need to be "re-matched" against the original routes to figure out which "original" route was specified, and what were the allowed tokens on that route.


daemon/algod/api/server/lib/common.go
```diff
@@ -37,6 +37,7 @@ type Route struct {
 	Method      string
 	Path        string
 	HandlerFunc HandlerFunc
+	AuthFree    bool // routes that has the AuthFree set to true doesn't require authentication
```
justicz, line 40: I think `NoAuth` is a better name

daemon/algod/api/server/lib/middlewares/auth.go
```diff
@@ -96,10 +114,19 @@ func Auth(log logging.Logger, apiToken string) func(echo.HandlerFunc) echo.Handl
...
+			for i, privAuth := range authTokens {
+				compareResult := subtle.ConstantTimeCompare(providedToken, privAuth.tokenBytes)
+				// if compareResult is 1, we want to update matchingAuthIndex with i. otherwise, we want to keep it as is.
+				matchingAuthIndex = compareResult*i + matchingAuthIndex*(1-compareResult)
+			}
+
+			if matchingAuthIndex >= 0 {
 				// Token was correct, keep serving request
-				return next(ctx)
+				if authTokens[matchingAuthIndex].paths[route] {
```
justicz, line 127: it looks like you'll now get an `"Invalid API Token"` instead of a 404/405 if you request a route that doesn't exist?

daemon/algod/api/server/router.go
```diff
@@ -97,8 +98,42 @@ func registerHandlers(router *echo.Echo, prefix string, routes lib.Routes, ctx l
...
+	}
+
+	// pick the "regular" endpoints
+	for path := range v2.GetRoutes(false) {
+		authRoutes[apiToken] = append(authRoutes[apiToken], path)
+	}
+
+	// pick the "admin" endpoints
+	for path := range v2.GetRoutes(true) {
+		authRoutes[adminAPIToken] = append(authRoutes[adminAPIToken], path)
```
justicz, line 126: is it intentional that the admin API token can only hit the admin endpoints, and not regular endpoints?

daemon/algod/api/server/lib/middlewares/auth.go
```diff
@@ -34,20 +34,38 @@ const TokenHeader = "X-Algo-API-Token"
 const urlAuthFormatter = "/urlAuth/%s"
 const debugRouteName = "debug"
 
-// Allowed auth bypass names
-var noneAuthRoutes = []string{"healthcheck", "swagger.json"}
+// AuthRoutes define the mapping of authenticaion key to allowed routes.
```
justicz, line 37: authenticaion -> authentication

daemon/algod/api/server/lib/middlewares/auth.go
tsachiherman: yes, it's an issue, but it's not there. let me revise that.

daemon/algod/api/server/router.go
tsachiherman: yes, that was my intension.

daemon/algod/api/server/lib/middlewares/auth.go
tsachiherman: np.

daemon/algod/api/server/router.go
```diff
@@ -97,8 +98,42 @@ func registerHandlers(router *echo.Echo, prefix string, routes lib.Routes, ctx l
 	}
 }
 
+func makeAuthRoutes(apiToken string, adminAPIToken string, enableProfiler bool) middlewares.AuthRoutes {
+	authRoutes := make(middlewares.AuthRoutes)
+
```
justicz, line 103: It would be good to call `tokens.ValidateAPIToken` on `apiToken` and `adminAPIToken` here, just to make absolutely sure we don't serve these endpoints without auth.

justicz: (I know we call it earlier elsewhere, but it's in a different package, and doing it here might prevent a mistake in the future)

daemon/algod/api/server/lib/middlewares/auth.go
```diff
@@ -96,10 +114,19 @@ func Auth(log logging.Logger, apiToken string) func(echo.HandlerFunc) echo.Handl
...
+
+			if matchingAuthIndex >= 0 {
 				// Token was correct, keep serving request
-				return next(ctx)
+				if authTokens[matchingAuthIndex].paths[route] {
+					return next(ctx)
+				}
 			}
 
 			return ctx.String(http.StatusUnauthorized, "Invalid API Token")
```
winder, line 132: To make the error format consistent, could you change this to: `return echo.NewHTTPError(http.StatusUnauthorized, "Invalid API Token")`

daemon/algod/api/generate.sh
```diff
@@ -2,14 +2,21 @@
...
+$(GO111MODULE=on go get -u "github.com/algorand/oapi-codegen/...@v1.3.5-algorand2")
+
+popd
 
 # Convert v2 to v3
 curl -s -X POST "https://converter.swagger.io/api/convert" -H "accept: application/json" -H "Content-Type: application/json" -d @./algod.oas2.json  -o 3.json
 cat 3.json | json_pp > algod.oas3.yml
 rm 3.json
-sed -i 's/\*.\*/application\/json/g' algod.oas3.yml
+sed -i '' 's/\*.\*/application\/json/g' algod.oas3.yml
```
winder, line 16: This breaks in on Linux, does `sed --posix -i 's/\*.\*/application\/json/g' algod.oas3.yml` work on the mac?

Alternatively we might need something like this:
```
if [[ "$OSTYPE" == "darwin"* ]]; then
  sed -i '' 's/\*.\*/application\/json/g' algod.oas3.yml
else
  sed -i 's/\*.\*/application\/json/g' algod.oas3.yml
fi
```

daemon/algod/api/server/router.go
```diff
@@ -107,8 +142,7 @@ func NewRouter(logger logging.Logger, node *node.AlgorandFullNode, shutdown <-ch
 	e.Pre(middleware.RemoveTrailingSlash())
 
 	e.Use(echo.WrapMiddleware(middlewares.Logger(logger)))
-	// TODO: New auth middleware - https://github.com/algorand/go-algorand/issues/863
-	e.Use(middlewares.Auth(logger, apiToken))
+	e.Use(middlewares.Auth(logger, makeAuthRoutes(apiToken, adminAPIToken, node.Config().EnableProfiler)))
```
winder, line 145: Rather than attaching the Auth middleware globally, it's possible to attach it to specific sub-routes. So for example we could skip adding it to the common routes.

I think that you're implementation is fine, but this is another option.

daemon/algod/api/server/v2/utils.go
```diff
@@ -186,3 +187,55 @@ func toCodecMap(input interface{}, output *map[string]interface{}) (err error) {
 	err = decode(protocol.CodecHandle, encoded, &output)
 	return
 }
+
+type pathColletingRouter struct {
```
winder, line 191: s/Colleting/Collecting

daemon/algod/api/generate.sh
algobolson: isn't this obsolete anyway? I don't think there are any `*.*` to be replaced in the current files.

daemon/algod/api/server/router.go
winder: The collector pattern is very clever, but I think we can avoid it. I could also update our `oapi-codgen` fork so that we could pass middleware to the handlers which could simplify some of the code:
```
generated.RegisterHandlers(router, handler, authMiddleware(generalKey))
private.RegisterHandlers(router, handler, authMiddleware(privatekey))
```

We could already do this with the pprof endpoints:
```
e.GET("/debug/pprof/*", echo.WrapHandler(http.DefaultServeMux), authMiddleware(private))
e.GET("/urlAuth/:token/debug/pprof/*", echo.WrapHandler(http.DefaultServeMux), authMiddleware(private)
```

daemon/algod/api/generate.sh
winder: I thought I had them fully defined in the oas2 file, but they were getting changed to `*.*` after the conversion. I haven't checked this since adding the fix originally.

tsachiherman: Since I can't see *.* in the file, I'm going to comment out this line for now.

daemon/algod/api/server/router.go
tsachiherman: Yes, I was thinking about refactoring a better code to start with, but figured out that we could push that for later. As long as it's behaves correctly, we can change the underlaying implementation later on.

daemon/algod/api/server/auth.go
```diff
@@ -0,0 +1,168 @@
...
+		}
+	}
+
+	if h.route.Path == "/urlAuth/:token/debug/pprof/*" {
+		return h.DebugHandler
+	}
+	return h.Handler
+}
+
+// DebugHandler is the pprof debug handler, and used only for the `/urlAuth/:token/debug/pprof/*` path
```
winder, line 107: Do you know why the pprof endpoint needs to have the token provided as part of the path?
Why not use the same header as everything else and skip the special case?

algobolson: `go tool pprof` can auto fetch from a URL, but can't set headers

winder:
Because of your editorial comment, I think that we should fix the `RegisterHandlers` code to allow setting different auth middleware to different routes. This should significantly simplify the implementation since route matching will no longer be relevant to the authenticator.

I took a look at what would be involved to do this and found that we would need to edit a template used by oapi-codegen. This seems to be a configurable thing already, but since we've already forked the tool I went ahead and modified our fork. If you upgrade to `v1.3.5-algorand3` it should allow injecting different versions of the authentication middleware to different routes.

tsachiherman:
> Because of your editorial comment, I think that we should fix the `RegisterHandlers` code to allow setting different auth middleware to different routes. This should significantly simplify the implementation since route matching will no longer be relevant to the authenticator.
> 
> I took a look at what would be involved to do this and found that we would need to edit a template used by oapi-codegen. This seems to be a configurable thing already, but since we've already forked the tool I went ahead and modified our fork. If you upgrade to `v1.3.5-algorand3` it should allow injecting different versions of the authentication middleware to different routes.

I'm completely fine if you'd like to make further implementation changes. I agree that the current output of oapi-codegen is sub-optimal for our usage, but I attempted to avoid making changes there as well. If the code here would work correctly, let's merge it in. We have other features that depends on this functionality to be available ( fast catchup, stateful teal ).

We can do the updates to oapi-codegen later on along with corresponding changes to the authentication piece. ( which would be transparent to the rest of the system )

daemon/algod/api/server/auth.go
```diff
@@ -0,0 +1,168 @@
...
+// RegisterHandlers registers the given authentication routes
+func (auth *Auth) RegisterHandlers(authRoutes AuthRoutes) {
+	routeTokens := make(map[echo.Route][]string)
+
+	// for each unique route, find the list of tokens.
+	for token, routes := range authRoutes {
+		for route := range routes {
+			routeTokens[route] = append(routeTokens[route], token)
+		}
+	}
```
rotemh, line 60: seems like it would simplify this code as well.

daemon/algod/api/server/router.go
```diff
@@ -71,13 +71,13 @@ import (
...
 	"github.com/algorand/go-algorand/daemon/algod/api/server/v2"
-	"github.com/algorand/go-algorand/daemon/algod/api/server/v2/generated"
 	"github.com/algorand/go-algorand/logging"
 	"github.com/algorand/go-algorand/node"
+	"github.com/algorand/go-algorand/util/tokens"
 )
 
 const (
-	apiV1Tag = "v1"
+	apiV1Tag = "/v1"
```
rotemh, line 80: It's a bit weird that the tag has a `/` in it

daemon/algod/api/client/restClient.go
```diff
@@ -88,16 +81,11 @@ func stripTransaction(tx string) string {
 }
 
 // submitForm is a helper used for submitting (ex.) GETs and POSTs to the server
-func (client RestClient) submitForm(response interface{}, path string, request interface{}, requestMethod string, encodeJSON bool, decodeJSON bool) error {
+func (client RestClient) submitForm(response interface{}, path string, request interface{}, requestMethod string, encodeJSON bool, decodeJSON bool, authToken string) error {
```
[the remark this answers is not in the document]
tsachiherman: I'll remove the parametrized authToken, but would keep the versioned path, as we will need it for v2 commands.

daemon/algod/api/server/auth.go
```diff
@@ -0,0 +1,168 @@
...
+
+	"github.com/algorand/go-algorand/daemon/algod/api/server/lib/middlewares"
+	"github.com/algorand/go-algorand/logging"
+)
+
+const urlAuthFormatter = "/urlAuth/%s"
+const debugRouteName = "debug"
+
+// AuthRoutes define the desired mapping of authentication token keys to allowed routes.
+type AuthRoutes map[string]map[echo.Route]echo.HandlerFunc
```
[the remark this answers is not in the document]
tsachiherman: The user-facing implementation has two keys, that's true. But the inner handler is capable to handle as many as defined. 
If we'll decide to extend the user facing interface to support more sophisticated schema ( i.e. json db ), we wouldn't need to change this file, but rather just the "configuration" part. 
I tried to follow the spirit of https://github.com/algorand/go-algorand/issues/863, supporting any number of tokens, rather than forcing it into two.

daemon/algod/api/server/auth.go
justicz: it's not incorrect, but from a design point of view, it seems weird to store every token on every route it can authenticate. It seems like there should be something that maps `token -> token type` and then something that maps `token type -> allowed routes`

justicz: (or `route -> allowed token types` as it may be)

tsachiherman: that's a nice optimization, for above.. I don't see it becoming a critical unless we're going to have either large number of tokens or large number of entry points.

daemon/algod/api/server/router.go
tsachiherman: yes, I did it to ensure the registration with Echo is symmetrical between the v1 and v2 entry points. I can eliminate it here, but need to make corresponding changes to the code generator ( which would be good, but abit outside the scope of this task )

daemon/algod/api/generate.sh
```diff
@@ -2,14 +2,22 @@
 set -e
 
 rootdir=`dirname $0`
-pushd $rootdir
+pushd ${rootdir}
+
+cd ${rootdir}/../../..
+$(GO111MODULE=on go get -u "github.com/algorand/oapi-codegen/...@v1.3.5-algorand2")
```
winder, line 8:
```suggestion
$(GO111MODULE=on go get -u "github.com/algorand/oapi-codegen/...@v1.3.5-algorand4")
```

PR #991, opened 2020-04-24
Rekeying: basic frontend

algoradam:
Add very basic rekeying support to goal, algokey, and libgoal.

`algokey` now sets the authaddr if necessary, inferring the address from the private key used for signing and checking whether it matches the transaction sender.

`libgoal` gets a new method `kmdClient.SignTransactionWithWalletAndSigner` that's like `SignTransactionWithWallet` but takes in an extra argument, the signer address. (If the signer address is the empty string, the behavior is the same as `SignTransactionWithWallet`.)

`goal clerk spend` gets a flag to set the rekey-to address. (It should be easy to add this flag to other goal commands that create transactions.) `goal clerk sign` gets a flag to specify the signer address; it uses the new libgoal method.

### DIFF
--- a/cmd/algokey/sign.go
+++ b/cmd/algokey/sign.go
@@ -34,7 +34,6 @@ var signKeyfile string
 var signTxfile string
 var signOutfile string
 var signMnemonic string
-var rekeyed bool
 
 func init() {
 	signCmd.Flags().StringVarP(&signKeyfile, "keyfile", "k", "", "Private key filename")
@@ -43,7 +42,6 @@ func init() {
 	signCmd.MarkFlagRequired("txfile")
 	signCmd.Flags().StringVarP(&signOutfile, "outfile", "o", "", "Transaction output filename")
 	signCmd.MarkFlagRequired("outfile")
-	signCmd.Flags().BoolVarP(&rekeyed, "rekeyed", "r", false, "Required if the sending account has been rekeyed and the signing key doesn't match the account address")
 }
 
 var signCmd = &cobra.Command{
@@ -74,7 +72,7 @@ var signCmd = &cobra.Command{
 			}
 
 			stxn.Sig = key.Sign(stxn.Txn)
-			if rekeyed {
+			if stxn.Txn.Sender != basics.Address(key.SignatureVerifier) {
 				stxn.AuthAddr = basics.Address(key.SignatureVerifier)
 			}
 			outBytes = append(outBytes, protocol.Encode(&stxn)...)
--- a/cmd/algokey/sign.go
+++ b/cmd/algokey/sign.go
@@ -25,6 +25,7 @@ import (
 	"github.com/spf13/cobra"
 
 	"github.com/algorand/go-algorand/crypto"
+	"github.com/algorand/go-algorand/data/basics"
 	"github.com/algorand/go-algorand/data/transactions"
 	"github.com/algorand/go-algorand/protocol"
 )
@@ -33,6 +34,7 @@ var signKeyfile string
 var signTxfile string
 var signOutfile string
 var signMnemonic string
+var rekeyed bool
 
 func init() {
 	signCmd.Flags().StringVarP(&signKeyfile, "keyfile", "k", "", "Private key filename")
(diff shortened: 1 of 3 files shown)


cmd/algokey/sign.go
```diff
@@ -71,6 +74,9 @@ var signCmd = &cobra.Command{
 			}
 
 			stxn.Sig = key.Sign(stxn.Txn)
+			if rekeyed {
+				stxn.AuthAddr = basics.Address(key.SignatureVerifier)
+			}
```
rotemh, line 79: To make sure I understand. You sign the Transaction and then modify the structure? Why don't we sign the rekeyed param?

cmd/goal/clerk.go
```diff
@@ -338,6 +353,10 @@ var sendCmd = &cobra.Command{
 		if err != nil {
 			reportErrorf(errorConstructingTX, err)
 		}
+		if (rekeyTo != basics.Address{}) {
```
tsachiherman, line 356: nit : There is a IsZero function that would do just that.

cmd/goal/clerk.go
```diff
@@ -663,7 +682,7 @@ var signCmd = &cobra.Command{
 				signedTxn = uncheckedTxn
 			} else {
 				// sign the usual way
-				signedTxn, err = client.SignTransactionWithWallet(wh, pw, uncheckedTxn.Txn)
+				signedTxn, err = client.SignTransactionWithWalletAndSigner(wh, pw, signerAddress, uncheckedTxn.Txn)
```
tsachiherman, line 685: Don't we need some type of input validation for `signerAddress` ?

cmd/algokey/sign.go
algoradam: The AuthAddr is part of the SignedTxn but not part of the Transaction -- it doesn't get signed. (The AuthAddr doesn't change what the transaction does, it just tells you what address to check the sig / msig / lsig against.)

cmd/goal/clerk.go
algoradam: Yeah, I wasn't sure whether the libgoal function should take in an Address or a string. I made it a string because that's what libgoal's `MultisigSignTransactionWithWallet` function takes in for its signer address, but I could switch it and have the address decoding in goal instead of libgoal if you think that would be better.

cmd/goal/clerk.go
algoradam: Fixed!

cmd/algokey/sign.go
tsachiherman: It sounds like the "AuthAddr" isn't even required here, as the node could see if it's defined in the account. If it does, it could use it against the signature anyway.

rotemh: @tsachiherman If I remember correctly, this is was a performance decision tradeoff.

algoradam: Yes, the AuthAddr is deliberately included in the SignedTxn explicitly because otherwise it makes signature checks stateful, and a lot of the optimizations we have for the block evaluator need signature checking to be stateless. (We went with stateless TEAL for similar reasons.) Explicitly including the AuthAddr does make SignedTxns a little bigger, but only if they're from accounts that have rekeyed, which are expected to be rare. The design doc discusses this in a little more detail.

cmd/algokey/sign.go
```diff
@@ -41,6 +43,7 @@ func init() {
 	signCmd.MarkFlagRequired("txfile")
 	signCmd.Flags().StringVarP(&signOutfile, "outfile", "o", "", "Transaction output filename")
 	signCmd.MarkFlagRequired("outfile")
+	signCmd.Flags().BoolVarP(&rekeyed, "rekeyed", "r", false, "Required if the sending account has been rekeyed and the signing key doesn't match the account address")
```
derbear, line 46: Why is this option required? Isn't it sufficient to (below) check if `key.SignatureVerifier = stxn.Sender` and set `stxn.AuthAddr` if not?

algoradam: I'll go ahead and make it automatic and get rid of the flag. (I originally put the flag there so that I wouldn't change any existing algokey behavior, because I was worried that there might be some tests or obscure use cases that depend on algokey's existing behavior. But this seems pretty unlikely since a transaction signed by the wrong key with no authaddr set is guaranteed to be invalid anyway.)

PR #992, opened 2020-04-24
Eric/cicd performance common telemetry

egieseke:
## Summary

Update performance test algonet recipe files to share a common telemetry server `cicd-performance`.  This will simplify running performance tests on arbitrary branches.  

## Test Plan

Tested by starting up a shared telemetry server and running scenario1 performance test.


egieseke:
> This would not play nicely once there is more than one concurrent performance test running.
> it's not just for the cicd which is currently limited to a single instance at a time, but rather conflicting with runs that I might be doing.
> 
> I think that for the purpose of the cicd, it's ok to have a single telemetry server for now, but instead of modifying the scenario1 recipe, clone it into scenario1-cicd, so that you could make the specialized changes there. another solution is to dynamically modify the JSON file to use the specific telemetry server.

The individual test will be writing to different ES indexes so they won't conflict even if running at the same time.

egieseke:
> You need to create new recipe files to hold the hard coded telemetry configuration with a hard coded telemetry server, put them along side the Jenkins task definition.
> 
> While you're at it you should probably fork the other config files as well so that the perf test cases don't change if the "master" scenarios are updated.
> 
> Additionally we probably don't want to advertise this server, it's an internal thing that isn't really useful to the public.

I prefer a simpler solution that does require managing forks.  Only the recipes for automated performance testing have been updated.  Maybe Tsachi's idea to use a playbook to override is the best solution.

tsachiherman:
> You need to create new recipe files to hold the hard coded telemetry configuration with a hard coded telemetry server, put them along side the Jenkins task definition.
> 
> While you're at it you should probably fork the other config files as well so that the perf test cases don't change if the "master" scenarios are updated.
> 
> Additionally we probably don't want to advertise this server, it's an internal thing that isn't really useful to the public.

From my perspective, the last one ( which I missed ! ) is the most important one. I can't imagine a concrete attack vector here, but no reason to publish it.

winder:
The algonet config files were designed for exactly this purpose. I think that creating `perftest-config.json` and `perftest-recipe.json` is the simple solution. The tool is already pretty complicated, so I think we should avoid adding new features if they can be accomplished using existing ones.

egieseke:
Updated cicd performance automation pipeline to use diagcfg utility to dynamically update the telemetry endpoint.  Closing this pull request which is no longer required.

PR #994, opened 2020-04-27
Skip V22 consensus upgrade

onetechnical:
To prevent a dual upgrade round, we need to jump directly from V21 to
V23.

## Summary

## Test Plan


onetechnical:
I don't think we need this; if we update master, it'll update in the next betanet release anyway.

PR #995, opened 2020-04-27
REST API v2: Regenerate endpoints with expected parameter checking and pluggable middleware

winder:
I updated oapi-codegen tool to generate an error when invalid parameters are provided.

This also brings in the new pluggable middleware option to `RegisterHandlers`.

It turns out that editing the oapi-codegen templates is pretty straight forward, it's using golangs built in templating engine:
* [unknown param checking](https://github.com/algorand/oapi-codegen/commit/9cd29e22fd2f9b48236718bc7c53b4d7a91a4c02#diff-b3d9327d388e20f7fa18dfe9c8b4e0c7)
* [pluggable middleware](https://github.com/algorand/oapi-codegen/commit/e7b6d6c35522a01a388e299c9669383ec4b62e2f#diff-af363b41d3a0933bd4b3863bf7ef4eb4)


daemon/algod/api/server/v2/generated/routes.go
```diff
@@ -244,6 +316,16 @@ func (w *ServerInterfaceWrapper) ShutdownNode(ctx echo.Context) error {
 
 // GetStatus converts echo context to params.
 func (w *ServerInterfaceWrapper) GetStatus(ctx echo.Context) error {
+
+	validQueryParams := map[string]bool{}
+
+	// Check for unknown query parameters.
+	for name, _ := range ctx.QueryParams() {
+		if _, ok := validQueryParams[name]; !ok {
```
tsachiherman, line 324: In the empty case, the test is not requited.

algobolson:
A warning that goes back to the caller would be useful. Logging it server side not so much. I ran into a case recently writing an API and testing it where it took me some debugging to figure out the filter wasn't working because I typo'd the query key. Failing the query would have been fine in that case too.

daemon/algod/api/server/v2/generated/routes.go
winder: I would also like to move the maps to the top level so that they aren't recomputed every time. Perhaps the compiler is smart enough to do this automatically, but it wouldn't be too much work to scope the variables with the operation ID.

Another improvement is to skip these:
```
	// ------------- Optional query parameter "format" -------------
	if paramValue := ctx.QueryParam("format"); paramValue != "" {
	}
```

It seems to be an artifact related to required query parameters (of which we have none). There is an `else` generated which returns an error if the required parameter is missing. Alone the check is meaningless.

At the moment I'm not planning to add these improvements.

winder:
This might be one of those REST design philosophies that people debate about. Personally I much prefer an error. If I add a parameter, I expect it to do something. If the server silently ignores it and the response doesn't reflect what I thought the parameter should do, then my first thought is that there is a bug in the server.

I think that returning an error will make the REST APIs easier to support because of this.

tsachiherman:
> This might be one of those REST design philosophies that people debate about. Personally I much prefer an error. If I add a parameter, I expect it to do something. If the server silently ignores a typo and the response doesn't reflect the parameter I thought was being set, then my first thought is that there is a bug in the server.
> 
> I think that returning an error will make the REST APIs easier to support because of this.

I think that given that we're going to have this only for v2, it would be ok : we won't be breaking any existing applications, and while 3rd parties would migrate their solutions, they would need to enforce these restrictions.

tsachiherman:
I think that you'll need to change the go.sum as well.

go.sum
```diff
@@ -34,8 +35,11 @@ github.com/gen2brain/beeep v0.0.0-20180718162406-4e430518395f h1:eyHMPp7tXlBMF8P
 github.com/gen2brain/beeep v0.0.0-20180718162406-4e430518395f/go.mod h1:GprdPCZglWh5OMcIDpeKBxuUJI+fEDOTVUfxZeda4zo=
 github.com/getkin/kin-openapi v0.3.1 h1:9mLtayAmieqUnNACL0HqHbxkTc+z1+15sxXpLoJOGEQ=
 github.com/getkin/kin-openapi v0.3.1/go.mod h1:W8dhxZgpE84ciM+VIItFqkmZ4eHtuomrdIHtASQIqi0=
+github.com/getkin/kin-openapi v0.8.0 h1:a6TQjTqwkyscC4/hShJX7WhCVE+4bi9lzw61XHQW5hE=
+github.com/getkin/kin-openapi v0.8.0/go.mod h1:zZQMFkVgRHCdhgb6ihCTIo9dyDZFvX0k/xAKqw1FhPw=
```
tsachiherman, line 39: The v0.3.1 version can be removed.

go.sum
```diff
@@ -129,13 +134,16 @@ github.com/stretchr/objx v0.1.0/go.mod h1:HFkY916IF+rwdDfMAkV7OtwuqBVzrE8GR6GFx+
 github.com/stretchr/testify v1.3.0/go.mod h1:M5WIy9Dh21IEIfnGCwXGc5bZfKNJtfHm1UVUgZn+9EI=
 github.com/stretchr/testify v1.4.0 h1:2E4SXV/wtOkTonXsotYi4li6zVWxYlZuYNCXe9XRJyk=
 github.com/stretchr/testify v1.4.0/go.mod h1:j7eGeouHqKxXV5pUuKE4zz7dFj8WfuZ+81PSLYec5m4=
+github.com/stretchr/testify v1.5.1 h1:nOGnQDM7FYENwehXlg/kFVnos3rEvtKTjRvOWSzb6H4=
+github.com/stretchr/testify v1.5.1/go.mod h1:5W2xD1RspED5o8YsWQXVCued0rvSQ+mT+I5cxcmMvtA=
```
tsachiherman, line 138: github.com/stretchr/testify 1.3 and 1.4 can be removed.

winder:
@tsachiherman is there a recommended way for doing this? I've been running `go mod tidy` and sometimes it adds things instead of removing them.

tsachiherman:
> @tsachiherman is there a recommended way for doing this? I've been running `go mod tidy` and sometimes it adds things instead of removing them.

not that I'm aware of. But I try to review it manually in GitHub to understand what I'm doing. When not sure, it's always safe to default with previously subset.

winder:
@tsachiherman I removed some of these but `make install` adds most back in.

tsachiherman:
> @tsachiherman I removed some of these but `make install` adds most back in.

We need to figure these out.. but we can do that another day.

pbennett:
go.sum will accrue hashes of anything ever fetched.  It's meant to be a backstop against referencing version X and then later when attempting to fetch version X again (different user, machine, etc.), the hash doesn't match.   This would imply the source has been altered.
So if you referenced 1.1.0 at one point, you'd have entries for its various files (its go.mod file as well as I believe a content hash).  If you then updated to reference 1.2.0, 1.1.0 will remain in go.sum.  If you go back to 1.1.0, you want to get the same version you had before and that it wasn't altered in some way from when you successfully used it/tested it.  I think it's really more an issue of verifying that automated builds only ever access 'verified' content but backing out is a valid use case as well.

Almost any go build command (even 'go list' !) will update go.mod to include all dependencies referenced by the packages within the current module.  This is why you'll see go.mod and go.sum change in seemingly random places.  This means your dependencies are not current.  Running go test may add new dependencies as well - for the dependencies needed for testing.
There are tricks for handling 'tools' as well - so hack makefile recipes to get tooling (like mock generation tools, etc.) can go away. (ie: configure_dev-deps.sh shouldn't be necessary)

go mod tidy will *remove* references from go.mod that aren't currently referenced.  Note, at no point will it remove items from go.sum.  It's meant as a go.mod 'cleanup'.  Note, it will likely remove things that will then be added right back if you then run tests.

winder:
@pbennett thank you for the explanation :)

PR #1002, opened 2020-04-28
Improve legacy pipeline

btoll:
- Tested that build artifacts can be pushed to s3 buckets.
- Improved script (`snapshot.sh`) that downloads the build packages and creates new snapshots, along with pushing them to s3 for `apt` and `yum` use.
- Remove ec2-specific stages from `prod/Jenkinsfile` as it's no longer needed.


scripts/release/prod/stage/snapshot.sh
```diff
@@ -1,13 +1,31 @@
...
+
 echo
 date "+build_release begin SNAPSHOT stage %Y%m%d_%H%M%S"
 echo
 
-set -ex
+DEBS_DIR="$HOME/packages/deb/$CHANNEL"
+DEB="algorand_${CHANNEL}_linux-amd64_${VERSION}.deb"
+
+aws s3 cp "s3://algorand-builds/channel/$CHANNEL/$DEB" .
```
winder, line 20: A lot of the spots where you're using `aws` could use `updater` instead. Or `updater` could be extended to do this. The key benefit to this approach is that the path pattern is a function of updater, and therefore if we want to do a migration to another format all of the related logic is encapsulated in one place (`utils/s3/s3Helper.go`).

For this call, I think you would use `updater send -c $CHANNEL $DEB`. In some cases it uses environment variables to switch the upload context (i.e. configure an upload to go to `algorand-builds`, `algorand-internal`, etc), but you could also use `-b algorand-builds`. The travis web-configuration lets us set environment variables, and we configure the bucket for legacy scripts there (then developer machines use a different bucket for their builds).

btoll: Good to know.  I'd rather not change this here, since this is the legacy pipeline and will quickly be replaced by `mule` tasks.  I'll look into this approach for those, however.

btoll: Now that I think of it, @egieseke wrote a `mule` task that allows for the uploading and downloading of files, and perhaps copying as well.  I haven't looked at it yet but need to soon, so I'm not sure how this is being handled.  I would assume it's using the `boto` library, so perhaps this is moot?

scripts/release/build/rpm/package.sh
```diff
@@ -1,30 +1,34 @@
...
-cd "$(dirname "$0")"/..
-export GOPATH=$HOME/subhome/go
-export REPO_DIR=${HOME}/subhome/go/src/github.com/algorand/go-algorand
-
 echo "Building RPM package"
 
+cd "$(dirname "$0")"/..
+
+REPO_DIR=${HOME}/subhome/go/src/github.com/algorand/go-algorand
+export REPO_DIR
```
egieseke, line 10: Consider consolidating assignment and export on a single line.

scripts/release/build/rpm/package.sh
```diff
@@ -1,30 +1,34 @@
-#!/bin/bash -e
-
-# build_rpm.sh - Build a .deb package for one platform.
-#
-# Syntax:   build_rpm.sh <output directory>
-#
```
egieseke, line 6: Consider specifying environment variables that are either required or optional.

scripts/release/prod/stage/generate_releases_page/run.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+
```
egieseke, line 2: Add file header

scripts/release/prod/stage/snapshot.sh
```diff
@@ -1,13 +1,35 @@
 #!/usr/bin/env bash
```
egieseke, line 2: Add file header.

scripts/release/build/rpm/package.sh
btoll: This is a shellcheck suggestion that I've been following since we have brought that tool into our process.

scripts/release/prod/stage/generate_releases_page/run.sh
btoll: I'm not sure I see the relevance here.  This is script that is is automated and shouldn't isn't invoked on the cli.  The `.generate_releases_page` itself, however, can be called from the cli and therefore does have a header that briefly describes its usage.

scripts/release/prod/stage/snapshot.sh
btoll: This isn't a cli tool.  Also, it's clear from the variable definitions and the conditional what the expected inputs are.

scripts/release/build/rpm/package.sh
btoll: There aren't any.

btoll: There are some shell variables that are inherited from parent processes.  One of the first tasks in `setup.sh` for almost every Jenkinsfile is creating a `build_env` file that is then often sourced by parent processes.

The `mule` tasks won't work like this.

PR #1004, opened 2020-04-28
REST API v2: Fix 'format' option in pending/blocks endpoints

winder:
`/v2/transactions/pending`, `/v2/transactions/pending/{txid}`, and `/v2/blocks/{round}` have a format option that toggles between JSON and MessagePack. This wasn't working properly before. I replaced `toCodecMap`, with proper go-codec encoding and real objects.

Talking with a few people it sounds like there have been related encoding issues in the past, so I opted to duplicate the generated model objects using real types (like `transactions.SignedTxn`) rather than trying to get it working with `interface{}` types.


daemon/algod/api/server/v2/handlers.go
```diff
@@ -181,17 +182,30 @@ func (v2 *Handlers) GetBlock(ctx echo.Context, round uint64, params generated.Ge
 		return ctx.Blob(http.StatusOK, "application/msgpack", blockbytes)
 	}
 
-	ledger := v2.Node.Ledger()
-	block, _, err := ledger.BlockCert(basics.Round(round))
-	if err != nil {
-		return internalError(ctx, err, errFailedLookingUpLedger, v2.Log)
-	}
+	if handle == protocol.JSONHandle {
```
tsachiherman, line 185: suggestion: instead of comparing the handle, add a switch statement above and switch by the params.Format instead. then, set the handle accordingly.

daemon/algod/api/server/v2/handlers.go
```diff
@@ -326,50 +335,51 @@ func (v2 *Handlers) PendingTransactionInformation(ctx echo.Context, txid string,
...
+	response := struct {
+		AssetIndex *uint64 `codec:"asset-index,omitempty"`
+		CloseRewards *uint64 `codec:"close-rewards,omitempty"`
+		ClosingAmount *uint64 `codec:"closing-amount,omitempty"`
+		ConfirmedRound *uint64 `codec:"confirmed-round,omitempty"`
+		PoolError string `codec:"pool-error"`
+		ReceiverRewards *uint64 `codec:"receiver-rewards,omitempty"`
+		SenderRewards *uint64 `codec:"sender-rewards,omitempty"`
+		Txn transactions.SignedTxn `codec:"txn"`
+	} {
```
tsachiherman, line 356: I think that the encoding wasn't working because of Txn map[string]interface{} instead of having a concrete data type.

looks to me like you need to define that in the spec so that the spec would align with that.

tsachiherman: doing it this was is a clear indication that the auto-generate types are wrong..

winder: The generated type was intended to express "map of objects", it's intentionally under-specifying so that we don't need to synchronize the backend type and the specification. We can always amend that later if we decide there's value to it.

I was able to get a `map[string]interface{}` by serializeing and deserializing the object, but had some trouble getting `map[string]interface{}` to work with `json/encoder`. I didn't realize we also use `go-codec` for JSON until I had already worked around the issue with the hack you see here. After talking with Derek I understand that there have been related encoding issues when we first designed the Transaction objects, so I may have rediscovering an old bug/limitations with the encoding.

With more time I would like to explore the following::
1) expand the oapi-codegen type-mapping so that I can plug in a real type + import it
2) See if the serialize/deserialize/serialize trick works using `go-codec` instead of `json/encoder`

For now I think that this hack has a pretty low surface area and would like to stick with it.

I created a tech debt ticket for this issue: https://github.com/algorand/go-algorand/issues/1015

tsachiherman: I think that extending the `oapi-codegen` ( or updating the spec by other means ) is the only way to go, for the long run.
since this corrects an issue, it's clearly a move to on the right direction.

PR #1007, opened 2020-04-28
Algod compile api

### LINKED ISSUE #1093: Add Teal Compile Rest Endpoint
Teal compile is used in many scenarios, many of which there isn't necessary `goal` installed. Adding the API end point will streamline smart contract deployment, especially in mobile.

stechu:
## Summary
New v2 API endpoint for compile teal code:
/v2/compile
POST text/plain utf-8 encoded source code
Return json if succeed
{
  “result”:base64 encoded bytes,
  “hash”:base32 sha512_256 of program bytes (Address style),
}
Otherwise 400 error (bad request) with actual error message
****
Resolve #1093 

## Test Plan

Tested locally.


daemon/algod/api/server/v2/handlers.go
```diff
@@ -447,3 +450,23 @@ func (v2 *Handlers) GetPendingTransactions(ctx echo.Context, params generated.Ge
 func (v2 *Handlers) GetPendingTransactionsByAddress(ctx echo.Context, addr string, params generated.GetPendingTransactionsByAddressParams) error {
 	return v2.getPendingTransactions(ctx, params.Max, params.Format, &addr)
 }
+
+// Compile TEAL code to binary, return both binary and hash
+// (POST /v2/compile)
+func (v2 *Handlers) TealCompile(ctx echo.Context) error {
+	buf := new(bytes.Buffer)
```
algobolson, line 457: In other places we carefully have a limited reader from Request.Body; it'd be good to do that too here with a max source size of 100kB or so. See net/http MaxBytesReader which we use a couple times.

stechu: added that in the new commit.

pzbitskiy:
I guess we need a simple `int 1` unit test for this.

stechu:
@pzbitskiy currently we don't have any unit test for rest API. The tests are done by SDK separately. I plan to add python sdk support as well as test after python sdk support v2 APIs.

daemon/algod/api/server/v2/handlers.go
```diff
@@ -447,3 +452,24 @@ func (v2 *Handlers) GetPendingTransactions(ctx echo.Context, params generated.Ge
...
+	source := buf.String()
+	program, err := logic.AssembleString(source)
+	if err != nil {
+		return badRequest(ctx, err, err.Error(), v2.Log)
+	}
+	
+	pd := logic.HashProgram(program)
+	addr := basics.Address(pd)
+	response := generated.PostCompileResponse {
+		Hash: addr.String(),
```
rotemh, line 471: it's spitting out the hash with a checksum - is that intentional?

algobolson: it should be Address style because if the program is for logic-owned stateless teal then the hash of the program is the account Address

daemon/algod/api/algod.oas3.yml
```diff
@@ -1,423 +1,643 @@
 {
```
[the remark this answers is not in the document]
algobolson: Yes, I have (in an unrelated pull) something that changes this to sorted keys for stable diffs.

daemon/algod/api/algod.oas2.json
```diff
@@ -595,7 +595,60 @@
           "required": true
         }
       ]
-    }
+    },
+      "/v2/compile": {
```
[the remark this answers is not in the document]
algobolson: I had the other as /v2/transactions/dryrun , but /v2/teal/... makes sense too

algobolson:
This endpoint _could_ unit test pretty easily since it doesn't rely on setting up any node or ledger state, but the endpoint is a pretty thin wrapper on logic.AssembleString() which is well tested. I don't think it needs a unit test. I kinda expect setting up a test to fake up an echo.Context() and call the handler would be annoying.

daemon/algod/api/algod.oas2.json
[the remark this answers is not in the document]
stechu: `/v2/teal/compile` makes sense. Will change to that.

stechu:
@algobolson @pzbitskiy @winder @rotemh I have fixed all the comments and this is ready to merge.

brianolson:
Switched primary accounts, this is the new me. LGTM. I guess we're waiting on @rotemh to see if his concerns have been addressed?

stechu:
ping  @rotemh

PR #1011, opened 2020-04-29
Logic signature return in transaction REST object

astudnev:
Added logic_signature optional field to transaction object in REST response.
This includes the code, arguments and disassembled output

Will be very usefull for explorers and data indexing


algobolson:
There's a v2 API in the works that returns the full signed transaction, with signature, transliterated from msgpack (as stored on the blockchain) into JSON. So, this data will be available via that API soon.
https://github.com/algorand/go-algorand/blob/master/daemon/algod/api/server/v2/generated/types.go#L281

astudnev:
Nice to hear. Please make sure this also returned on block request

PR #1012, opened 2020-04-29
Add `mule` task for testing packages

btoll:
All tests will either use the packages previously created in `$HOME/tmp/node_pkgs/` or will download them from staging and put them in the same location (the `USE_CACHE` env var controls this).

The initial docker container instanced from the host will create more docker containers to test the package(s) on multiple distros or versions, depending on the type of package.

```
  - task: docker.Make
    name: package-test-deb
    docker:
      image: algorand/mule-linux-ubuntu
      version: '{{ docker.Version.outputs.version }}'
      workDir: /projects/go-algorand
      env: [
        AWS_ACCESS_KEY_ID=$AWS_ACCESS_KEY_ID,
        AWS_SECRET_ACCESS_KEY=$AWS_SECRET_ACCESS_KEY,
        USE_CACHE=true
        CHANNEL=stable,
        BRANCH=rel/stable,
        SHA=8b0be452,
        VERSION=2.0.5
      ]   
      volumes: [ /var/run/docker.sock:/var/run/docker.sock ]
    target: package-test-deb WORKDIR=/projects/go-algorand
```


docker/build/mule.ubuntu.Dockerfile
```diff
@@ -0,0 +1,19 @@
...
+COPY *.yaml ./
+
+RUN pip install mulecli
+
+RUN apt-get update && apt-get install -y build-essential curl libboost-all-dev autoconf bsdmainutils
+RUN curl https://dl.google.com/go/go1.12.linux-amd64.tar.gz | tar -xzf -
+RUN mv go /usr/local
+ENV GOROOT=/usr/local/go \
+    GOPATH=$HOME/go
+ENV PATH=$GOPATH/bin:$GOROOT/bin:$PATH
```
btoll, line 14: This isn't great.  We should make the ticket to install go from only one location a priority.

scripts/release/mule/test/linux/amd64/deb/goal.exp
```diff
@@ -0,0 +1,105 @@
+#!/usr/bin/expect -f
+
```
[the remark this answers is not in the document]
btoll: Are we using this elsewhere?  If so, I'll remove it.

package-test.yaml
```diff
@@ -0,0 +1,80 @@
+tasks:
```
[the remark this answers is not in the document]
btoll: Aren't the names of the jobs and the file sufficient?

docker/build/mule.ubuntu.Dockerfile
```diff
@@ -0,0 +1,19 @@
+FROM docker:19 as docker
+FROM python:3.7
+
+COPY --from=docker /usr/local/bin/docker /usr/local/bin/docker
+COPY *.yaml ./
```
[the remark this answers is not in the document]
btoll: Yes, I'll look at changing it.

docker/build/mule.ubuntu.Dockerfile
btoll: That's **probably** all that is needed, but I'd rather keep it in sync with the version that we're using internally.

docker/build/mule.ubuntu.Dockerfile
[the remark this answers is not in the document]
btoll: Changed.

scripts/release/mule/test/linux/amd64/deb/goal.exp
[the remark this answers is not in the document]
egieseke: This is a copy of /Users/ericgieseke/GolandProjects/src/github.com/algorand/go-algorand/test/e2e-go/cli/goal/expect/basicGoalTest.exp, why?

package-test.yaml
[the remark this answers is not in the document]
egieseke: File header comments are generally a good idea to provide context.

package-test.yaml
```diff
@@ -0,0 +1,80 @@
...
+    name: package-test-deb
+    docker:
+      image: algorand/mule-linux-ubuntu
+      version: '{{ docker.Version.outputs.version }}'
+      workDir: /projects/go-algorand
+      env: [
+        AWS_ACCESS_KEY_ID=$AWS_ACCESS_KEY_ID,
+        AWS_SECRET_ACCESS_KEY=$AWS_SECRET_ACCESS_KEY
+      ]
+      volumes: [ /var/run/docker.sock:/var/run/docker.sock ]
```
bricerisingalgorand, line 22: I'm confused, is this make target starting it's own docker container? If so could this be split into two separate mule tasks, one using this docker image and another using the image of the container it's starting up? This way mule handles all the docker runs, so that way mule will know to clean them up afterward. Also this may work on our jenkins agents, but I'm not 100% sure it will

docker/build/mule.ubuntu.Dockerfile
bricerisingalgorand: I believe this actually installs 1.12.0. We probably want to use 1.12.17 instead. Why are we making an ubuntu image with go installed on it though? I thought we decided to do our builds on the centos images

btoll: This is just a placeholder so I could have a version of go and keep moving forward.  My point in calling it out is that it shouldn't be here at all, we should be referencing a common place that is doing the installation.

This isn't doing a build, it's testing, and I needed python3.7 and mule to do that.  I actually took most of this from your open PR.

I frankly don't care too much what dockerfile I use, as long as it gets the job done.

package-test.yaml
btoll: Yes, you must have missed my comments in the that describe the ticket description above.

I don't think that what you're proposing is a good strategy for this task.  The main docker container is spinning up at least four other docker images to do test the packages, and other distros may (probably?) be added.

If we could define our tasks using loops like in a vagrantfile or ansible scripts, that would make what I just said moot.  Otherwise, it's way too much boilerplate when it can be handled as I've implemented it here.

package-test.yaml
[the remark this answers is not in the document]
btoll: I fully agree when the code/markup is confusing as to its intent, but I don't think that's the case here.

package-test.yaml
onetechnical: I'm okay with this for now, but it may need to be re-tested on the Jenkins/EKS cluster, since the agents are not native hosts.

docker/build/mule.ubuntu.Dockerfile
btoll: Just to follow up here, I mis-labeled the dockerfile, it's actually a Debian distro:
```
root@f88dec6be829:~# cat /etc/*release
PRETTY_NAME="Debian GNU/Linux 10 (buster)"
NAME="Debian GNU/Linux"
VERSION_ID="10"
VERSION="10 (buster)"
VERSION_CODENAME=buster
ID=debian
HOME_URL="https://www.debian.org/"
SUPPORT_URL="https://www.debian.org/support"
BUG_REPORT_URL="https://bugs.debian.org/"
```
I like this dockerfile for tasks where I need `mule`, but again, I'm open to better proposals and am not particularly concerned as to the dockerfile.

PR #1016, opened 2020-05-01
Refactor Auth to use Middleware

winder:
Our routing library supports defining different middleware for different routes. It can be used to simplify the existing public/private auth implementation.

This PR re-introduces auth as a middleware object. This time it is stateless. Because it doesn't need to know about what endpoint it is authorizing it simply checks for one of the tokens it was initialized with and verifies that it exists in one of the valid locations.

I also went ahead and updated our CORS/Logging middleware to implement echo's middleware interface.


daemon/algod/api/server/router.go
```diff
@@ -60,6 +60,8 @@
 package server
 
 import (
+	"github.com/algorand/go-algorand/daemon/algod/api/server/v2/generated"
```
tsachiherman, line 63: move to a separate group

tsachiherman:
Note that this PR removes the ability to have a dynamic set of authenticated endpoints. I think it was one of Rotem's goals around the authentication piece.

daemon/algod/api/server/lib/middlewares/auth.go
```diff
@@ -0,0 +1,98 @@
...
+type AuthMiddleware struct {
+	// Header is the token header which needs to be provided. For example 'X-Algod-API-Token'.
+	header string
+
+	// Tokens is the set of tokens which can be set to allow access.
+	tokens [][]byte
+}
+
+// MakeAuth constructs the auth middleware function
+func MakeAuth(header string, tokens []string) echo.MiddlewareFunc {
```
rotemh, line 40: since we don't really have support in multiple tokens (well, exactly two). When we would, it will need to be more sophisticated than single list so I believe this can support only one token for now.

daemon/algod/api/server/router.go
```diff
@@ -88,44 +90,18 @@ func wrapCtx(ctx lib.ReqContext, handler func(lib.ReqContext, echo.Context)) ech
...
-	authRoutes[adminAPIToken] = adminAuthRoutes
-	for _, route := range common.Routes {
-		if route.NoAuth {
-			noAuthRoutes[echo.Route{Method: route.Method, Path: route.Path}] = wrapCtx(ctx, route.HandlerFunc)
-			continue
-		}
-		defaultAuthRoutes[echo.Route{Method: route.Method, Path: route.Path}] = wrapCtx(ctx, route.HandlerFunc)
-		adminAuthRoutes[echo.Route{Method: route.Method, Path: route.Path}] = wrapCtx(ctx, route.HandlerFunc)
+// registerHandler registers a set of Routes to [router]. if [prefix] is not empty, it
+// registers the routes to a new sub-router [prefix]
```
justicz, line 94: is this comment incorrect?

winder: It's worded a little strangely but I think it's accurate. I probably pulled this comment from the older version of this file.

PR #1018, opened 2020-05-01
Modify libgoal to use v2 status instead of v1 status

tsachiherman:
## Summary

Libgoal is currently using the v1 entry points exclusively.
This change modify the libgoal to start using the v2 status command, which will be expanded in subsequent PRs.

### DIFF
--- a/daemon/algod/api/algod.oas2.json
+++ b/daemon/algod/api/algod.oas2.json
@@ -358,7 +358,7 @@
         }
       }
     },
-    "/v2/status/wait-for-block-after/{round}/": {
+    "/v2/status/wait-for-block-after/{round}": {
       "get": {
         "description": "Waits for a block to appear after round {round} and returns the node's status at the time.",
         "produces": [
--- a/cmd/algoh/client.go
+++ b/cmd/algoh/client.go
@@ -19,12 +19,13 @@ package main
 import (
 	"context"
 
+	generatedV2 "github.com/algorand/go-algorand/daemon/algod/api/server/v2/generated"
 	"github.com/algorand/go-algorand/daemon/algod/api/spec/v1"
 )
 
 // Client is a minimal interface for the RestClient
 type Client interface {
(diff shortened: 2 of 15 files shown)


daemon/algod/api/algod.oas2.json
```diff
@@ -358,7 +358,7 @@
         }
       }
     },
-    "/v2/status/wait-for-block-after/{round}/": {
+    "/v2/status/wait-for-block-after/{round}": {
```
tsachiherman, line 361: When I attempted to make the call with extra "/", it would return me 404. It looks like something related to Echo, but I didn't spent too much time here, as I've figure that we want it to align with the rest of the REST commands. ( i.e. which doesn't have trailing "/" )

tsachiherman:
> LGTM, I was expecting there to me more changes though.
> 
> Do you know the difference between `client/algod/wrappers.go` and `libgoal.go? Do these things reference each other somehow?
> 
> One problem we'll have to deal with is that some of the end points (especially things like /assets) are going away with v2. I'm not sure how we deal with things like `pingpong` which depend on them.

Yes, I'm not sure what the best approach here. If we want to have our client infrastructure support and test both endpoints, setting "/v2/" isn't a good idea. We'll need to double the width of the infrastructure, which sounds like a really bad idea.

For the "status" alone, I think that the above is a reasonable solution.

tsachiherman:
I didn't really liked the first iteration, as it prevented us from testing the v1 APIs. I added this version affinity to allow us to test both variations. This should allow us to retain the pingpong scripts and extend that as we need.

daemon/algod/api/algod.oas2.json
winder: There is a "strip removing slash" feature in echo, implemented as a "Pre" middleware (as in before routing is done). I think whats happening here is the router requires the trailing slash, but the pre-router middleware removes it. So basically there's no way to actually reach the endpoint when there is a slash.

Nice catch.

PR #122, opened 2019-07-05
-- Fixes issue #120

rotemh:
## Summary

This PR fixes issue #120

PR #1032, opened 2020-05-05
Eric/update with tools

egieseke:
#122  Summary

Update the updater.sh script to support -gettools <outputdir> option to fetch tools in addition to algod binaries.   This option will be used on the dispensers or other installations where the tools are required so that the tools are kept in sync with the algod binaries. 

## Test Plan

Tested installation by following the installation guides on the developer portal, with and without the -gettools option.  

Link to issue:
https://github.com/algorand/go-algorand-internal/issues/419


cmd/updater/update.sh
```diff
@@ -249,7 +255,39 @@ TEMPDIR=""
...
+    export CLEANUP_UPDATE_TEMP_DIR=${TEMPDIR}
+    trap "rm -rf ${CLEANUP_UPDATE_TEMP_DIR}" 0
+
+    TARFILE=${TEMPDIR}/${LATEST}.tar.gz
+    UPDATESRCDIR=${TEMPDIR}/a
+    mkdir "${UPDATESRCDIR}"
+
+    if ! "${SCRIPTPATH}"/updater gettools -c "${CHANNEL}" -o "${TARFILE}" "${BUCKET}" "${SPECIFIC_VERSION}"
+    then
+        echo Error downloading tools tarfile
```
winder, line 270: missing quotes around these echo strings. Looks like the rest of this file is a bit inconsistent as well, but we may as well default to the convention we use in most of our scripts.

cmd/updater/update.sh
```diff
@@ -249,7 +255,39 @@ TEMPDIR=""
...
+}
+
 function download_update() {
+
+    SPECIFIC_VERSION=$1
+    if [ "${TOOLS_OUTPUT_DIR}" != "" ]; then
+        download_tools_update "${SPECIFIC_VERSION}"
+    fi
+
     SPECIFIC_VERSION=$1
```
winder, line 290: Don't define this twice

cmd/updater/update.sh
```diff
@@ -249,7 +254,39 @@ TEMPDIR=""
...
+function download_tools_update() {
+    SPECIFIC_VERSION=$1
+    echo "downloading tools update ${SPECIFIC_VERSION}"
+    TEMPDIR=$(mktemp -d 2>/dev/null || mktemp -d -t "tmp")
+    export CLEANUP_UPDATE_TEMP_DIR=${TEMPDIR}
+    trap "rm -rf ${CLEANUP_UPDATE_TEMP_DIR}" 0
+
+    TARFILE=${TEMPDIR}/${LATEST}.tar.gz
+    UPDATESRCDIR=${TEMPDIR}/a
+    mkdir "${UPDATESRCDIR}"
```
winder, line 266: The lines above looks wrong, `TARFILE` is used later in the script to extract the archive. I suspect that the tools will not be extracted as expected. It looks like you copied the code below without checking how all the variables are used later in the script.

cmd/updater/update.sh
egieseke: Added quotes around echo strings.  

> This looks almost correct, but I'm not sure how useful it is without also restarting the dispenser. You still have a manual process that you have to remember every time.
> 
> Why not put a simple script together to drop in the crontab ahead of `update.sh` which does exactly what is needed?
> 
> ./update_dispenser.sh beta algorand-releases /path/to/algod/binaries/ /path/to/dispenser/
> 
> ```
> CHANNEL=$1
> BUCKET=$2
> BINDIR=$3
> DISPENSER_DIR=$4
> 
> CURRENTVER="$(( ${BINDIR}/algod -v 2>/dev/null || echo 0 ) | head -n 1)"
> LATEST="$(${SCRIPTPATH}/updater ver check -c ${CHANNEL} ${BUCKET} | sed -n '2 p')"
> if [ ${CURRENTVER} -ge ${LATEST} ]; then
>     return 1
> fi
> 
> ${BINDIR}/updater gettools -c "${CHANNEL}" -o "/tmp/tools.tar.gz" "${BUCKET}"
> tar xvf /tmp/tools.tar.gz -o "${DISPENSER_DIR}"
> 
> pkill dispenser
> ${DISPENSER_DIR}/dispenser
> ```

This does not support all of the options of the updater.sh script, and fetching the tools may be generally useful for things besides the dispenser.   Also, the dispenser runs as a service, which needs to be stopped prior to updating the binary.  I am planning to put this logic in a separate crontab script which will check for a new dispenser binary in the tools directory.

cmd/updater/update.sh
egieseke: Removed.

cmd/updater/update.sh
egieseke: It does work correctly due to the order of operation, but I renamed the tools tar file var to avoid confusion.

cmd/updater/update.sh
```diff
@@ -245,12 +250,42 @@ function check_for_update() {
...
+}
+
 TEMPDIR=""
 TARFILE=""
 UPDATESRCDIR=""
 
 function download_update() {
     SPECIFIC_VERSION=$1
+
+    if [ "${TOOLS_OUTPUT_DIR}" != "" ]; then
```
btoll, line 285: You could check if it's non-empty by:
```
if [ -n "$TOOLS_OUTPUT_DIR" ]; then
...
```

cmd/updater/update.sh
```diff
@@ -245,12 +250,42 @@ function check_for_update() {
...
+        download_tools_update "${SPECIFIC_VERSION}"
+    fi
```
btoll, line 287: You're using two different versions of if/else (`then` on same line as `if` and other times on the next line).  Would you mind picking one for consistency?

cmd/updater/update.sh
egieseke: updated to use -n

cmd/updater/update.sh
egieseke: updated to use `; then` syntax

btoll:
Were you able to shellcheck this?

egieseke:
> Were you able to shellcheck this?

The newly added code passes spellcheck, the rest of the script needs work.

PR #1038, opened 2020-05-05
Fix finding dev@algorand.com's Keygrip for signing

onetechnical:
## Summary

The code to find the correct Keygrip for the signing key was not correct for dev@algorand.com. This fixes it by selecting the right key.

## Test Plan

Keygrip command tested on Ubuntu 18.04 Vagrant image.

### DIFF
--- a/scripts/release/forward_gpg_agent.sh
+++ b/scripts/release/forward_gpg_agent.sh
@@ -29,7 +29,7 @@ INSTANCE=$(ssh -i "$JENKINS_KEY" "$JENKINS" sudo cat /opt/jenkins/workspace/$BRA
 gpgp=$(find /usr/lib/gnupg{2,,1} -type f -name gpg-preset-passphrase 2> /dev/null)
 
 # Here we need to grab the signing subkey, hence `tail -1`.
-KEYGRIP=$(gpg -K --with-keygrip --textmode dev@algorand.com | grep Keygrip | tail -1 | awk '{ print $3 }')
+KEYGRIP=$(gpg -K --with-keygrip --textmode dev@algorand.com | egrep -A 1 '^ssb[^#]' | grep Keygrip | awk '{ print $3 }')
 echo "enter dev@ password"
 $gpgp --verbose --preset "$KEYGRIP"
 


scripts/release/forward_gpg_agent.sh
```diff
@@ -29,7 +29,7 @@ INSTANCE=$(ssh -i "$JENKINS_KEY" "$JENKINS" sudo cat /opt/jenkins/workspace/$BRA
 gpgp=$(find /usr/lib/gnupg{2,,1} -type f -name gpg-preset-passphrase 2> /dev/null)
 
 # Here we need to grab the signing subkey, hence `tail -1`.
-KEYGRIP=$(gpg -K --with-keygrip --textmode dev@algorand.com | grep Keygrip | tail -1 | awk '{ print $3 }')
+KEYGRIP=$(gpg -K --with-keygrip --textmode dev@algorand.com | egrep -A 1 '^ssb[^#]' | grep Keygrip | awk '{ print $3 }')
```
algobolson, line 32: I think would grab the wrong key for me. The gpg text output is inconvenient to parse.
```
$ gpg -K --with-keygrip brian.olson@algorand.com
sec   rsa2048 2020-04-30 [SC] [expires: 2022-04-30]
      09...
      Keygrip = A0...
uid           [ultimate] Brian Olson <brian.olson@algorand.com>
ssb   rsa2048 2020-04-30 [E] [expires: 2022-04-30]
      Keygrip = A1...
```
I think I need the "sec" key with "[SC]" flags for signing, not the "ssb" key with "[E]" key for encrypting.
I think keying on "ssb" or "sec" is closer to what we need, maybe actually the key-role part of "[SC]" (I think the S is for Sign)
Oh, just found the `--with-colons` option which is a bit more regular:
`gpg --list-secret-keys --with-keygrip --with-colons brian.olson@algorand.com`

onetechnical: Interesting. I needed to do this because I only had a signing subkey for dev@algorand.com:

```
johnlee@Algo-Admin-MBP final % gpg -K --with-keygrip --textmode rpm@algorand.com
sec   rsa2048 2019-06-12 [SC] [expires: 2021-06-11]
      1393C1B569ED363091FDBB6622999ED991A8BA41
      Keygrip = 151555B500ED5C7A6CF8117D4D4C49AD5194FD03
uid           [ unknown] Algorand RPM <rpm@algorand.com>
ssb   rsa2048 2019-06-12 [E] [expires: 2021-06-11]
      Keygrip = 0FA3F0FB9F5586E564D4B1CF5BE6A9C3BC9E66C1

johnlee@Algo-Admin-MBP final % gpg -K --with-keygrip --textmode dev@algorand.com
sec#  rsa3072 2019-06-11 [SC]
      611DE94A396F01359C72AF56EDACD29DA10A4EA6
      Keygrip = 59BEC781337B485284B9C255953D1809C30D9A4A
uid           [ unknown] Algorand developers <dev@algorand.com>
ssb   rsa3072 2019-11-20 [S] [expires: 2021-11-19]
      Keygrip = 54AAB31D7752B6103B40A2E007E6E473D98268E4
ssb#  rsa3072 2019-06-11 [E]
      Keygrip = 61A3E52A294D86F775D6E5A772A9955C188C7864
ssb#  rsa3072 2019-11-20 [E] [expires: 2021-11-19]
      Keygrip = 596A52110EF724A8504DFD3FC343A3A4F39962A0
```

So in this I can use the master key for signing, but in the second, I need the subkey. I suppose I need the master key for dev@algorand.com? 🤔

algobolson: riiight, I didn't have the dev@ key loaded, but it does have a special subkey for signing from. In the --with-colons output there's a column for key capabilities and if it has "s" for signing it's good to go (may be mixed with e for encrypt or c for certify). Or if we want to be fancy and can rely on ubuntu 18 or greater, `apt-get install -y python3-gpg` and there should be better scripting possibilities with better data structures in there. If we still need the script to run on centos7 then I'd go for --with-colons (and bash+awk+grep or python)

onetechnical: Because everything is more complicated than it needs to be, here is a blow out of the colon-formatted secret key for dev:

```
 1 sec: type of record
 2 -: calculated trust
 3 3072: length of key
 4 1: algorithm
 5 <redacted>: keyID
 6 1560257538: creation date
 7 : expiration date
 8 : serial number in crt records
 9 f: owner trust
10 : user-ID 
11 : signature class
12 scESC: Key capabilities (upper case is usable capabilities
13 : Used in FPR records for S/MIME keys for fingerprint of issuer cert
14 : Flag field used in the --edit menu output
15 #: sec number of token or a # if a simple stub
16 : hash algorithm
17 : curve name
18 23: compliance flags (23 is de-vs)
19 : last update
20 0: origin
21 (comment)
```
I can't use this key because of field 15 - it's only a stub. This is either an offline key or one generated via secure key like yubikey, which I obviously don't have. So I'd need to search out a subkey with signing capability.

btoll: > So in this I can use the master key for signing, but in the second, I need the subkey. I suppose I need the master key for [dev@algorand.com](mailto:dev@algorand.com)?

I only have the signing key for dev@ as well.  It's better to pass subkeys around because you can revoke them if you need to without jeopardizing your entire identity (i.e, the master key).

https://wiki.debian.org/Subkeys

onetechnical: Because Ben and I are currently the only signers, and we only have the signing key for dev@, I propose merging this as is. We need to have a larger conversation about GPG key management, because it's not clear how we want to handle this as an organization.

Issue #909, opened 2020-03-10
goal clerk compile without filesystem


jeapostrophe:
`algorand/sandbox` is really useful for working with Algorand without a complicated dev environment. However, it is currently not set up to do any sort of `$PWD`-forwarding, so it is not really useful to use `./sandbox goal clerk compile` since the compiler is based around files.

It might be nice to change the sandbox to do something like that, one trick is to use `docker-compose` like this:

```
    cat<<EOF | docker-compose -f - run -w /pwd goal $*
version: '3'
services:
  algorand-goal:
    image: algorand-sandbox
    volumes:
      - $PWD:/pwd
EOF
```

But I think it would be better for `goal clerk compile` to directly support getting its input from `STDIN` and giving its output to `STDOUT` as well. `solc` has a pretty good model for this where it accepts all of the input and produces all of the output as a JSON object with everything as some field/etc.

derbear:
I believe you can use `--` to read from `stdin`: https://github.com/algorand/go-algorand/blob/master/cmd/goal/commands.go#L528

jeapostrophe:
Two things...

First: While that doesn't cause a crash for me from `goal`, it doesn't work either. I get the error `the input device is not a TTY`.

Second: The usage of `--` for `stdin` is very non-standard. `--` is used in POSIX to terminate all options and `-` is used for STDIN/OUT. See [1] for a discussion. I actually tried `-` before submitted the issue.

1. https://www.gnu.org/software/libc/manual/html_node/Argument-Syntax.html

algobolson:
I was able to run `goal clerk compile -` to read from stdin, but if you also need output to stdout that appears to not be working at this time.

derbear:
Ah, you're right--I meant `-`, not `--`. My bad.

jeapostrophe:
FWIW, I found that just removing `-t` from the `sandbox goal` command so that I could send STDIN to the sandbox'd `goal`. It works well then.

jeapostrophe:
Here is a fix

https://github.com/justicz/go-algorand/pull/67

https://github.com/justicz/go-algorand/pull/67/commits/e4430cb40bcfd0e4f1276b5d505bbbd9e7727750

jeapostrophe:
I changed the push to the main repo

PR #1043, opened 2020-05-05
Fix PR909, compile without filesystem

jeapostrophe:
See pr

PR #1044, opened 2020-05-05
`goal clerk compile` assemble or disassemble to stdout

algobolson:
tested with
`goal clerk compile - < tlhc.teal|goal clerk compile --disassemble -|goal clerk compile -|goal clerk compile --disassemble -`
implements #909


algobolson:
Inspired by #1043

cmd/goal/clerk.go
```diff
@@ -872,20 +875,12 @@ var compileCmd = &cobra.Command{
...
-					reportErrorf("%s: %s\n", outname, err)
-				}
-				err = fout.Close()
+				err := writeFile(outname, outblob, 0666)
 				if err != nil {
 					reportErrorf("%s: %s\n", outname, err)
 				}
 			}
-			if !signProgram {
+			if !signProgram && outname != stdoutFilenameValue {
```
tsachiherman, line 883: If the filename is stdoutFilenameValue and the use specified signProgram, we need to exit with an error code.

algobolson: I think this is okay. I think this is just saying "if we are signing the program, we don't need to print its hash because it's not going to be used as a contract account because we're signing it for delegate useage"; and it's saying that we don't have a dedicated output "--hash-out-file foo" so if stdout is in use we don't print the program hash.

tsachiherman: If we were asked by the user to provide the signature of the program, than we should either do that, or completely fail. What about changing the output format so that it will be considered a TEAL comment ?

algobolson: I'm wondering if two things are getting mixed up on this thread.
1. the hash of the program, the sha512_256 of the program used as a 32 byte Address of a contract-account. printed on line 886 `fmt.Printf("%s: %s\n", fname, addr.String())`
2. a LogicSig where the program has been ed25519 signed by the private key of a standard account to make a delegated LogicSig, assembled in the `if signProgram{}` block 854-876.
The hash isn't needed if we're in signProgram/LogicSig mode; and is dropped if stdout is being used for the regular output.

jeapostrophe:
This is much more beautiful, I don't need thanks. I thanks you!

cmd/goal/clerk.go
tsachiherman: Ahh I see.
Yes, I missed the two `if signProgram` and `if !signProgram` sections. I thought that you've just ignoring the user request completely which seems incorrect.

PR #1048, opened 2020-05-06
Limit txn group size when broadcasting raw transactions

justicz:
## Summary

We should never need to decode more than `MaxTxGroupSize` transactions on transaction submission, because that group can never be valid.

## Test Plan

Covered by an existing test that I modified in this PR. The failure just happens higher up the stack now.

## Issue

https://github.com/algorand/go-algorand/issues/1090


tsachiherman:
Two requests before it can be merged -
1. Create a corresponding issue and link it to this PR.
2. Use the template for the PR header.

PR #1049, opened 2020-05-07
Add mule task for package signing

btoll:


scripts/release/mule/Makefile.mule
```diff
@@ -46,32 +46,30 @@ ci-build: ci-setup buildsrc gen
...
 
-package-docker: package-docker-$(OS_TYPE)-$(ARCH)
+package-test-deb:
+	scripts/release/mule/test/$(OS_TYPE)/$(ARCH)/test.sh $$WORKDIR $(OS_TYPE) $(ARCH) $(ARCH_BIT) $(VER) deb
 
-package-test: package-test-$(OS_TYPE)-$(ARCH)
-package-test-deb: package-test-deb-$(OS_TYPE)-$(ARCH)
-package-test-rpm: package-test-rpm-$(OS_TYPE)-$(ARCH)
+package-test-rpm:
+	scripts/release/mule/test/$(OS_TYPE)/$(ARCH)/test.sh $$WORKDIR $(OS_TYPE) $(ARCH) $(ARCH_BIT) $(VER) rpm
```
btoll, line 75: These targets are way too verbose, however, I have a commit that will simplify them and will create a PR when this code has been merged.

scripts/release/mule/sign/linux/amd64/sign.sh
```diff
@@ -0,0 +1,100 @@
...
+
+echo
+date "+build_release begin SIGN stage %Y%m%d_%H%M%S"
+echo
+
+WORKDIR="$1"
+
+if [ -z "$WORKDIR" ]
+then
+    echo "WORKDIR variable must be defined."
```
tsachiherman, line 13: It's not the WORKDIR variable that need to be defined.. it's that the first parameter is missing, right ?

scripts/release/mule/sign/linux/amd64/sign.sh
```diff
@@ -0,0 +1,100 @@
+#!/usr/bin/env bash
+
+set -ex
```
tsachiherman, line 3: add -pipefail, since you're piping error codes in 
```git archive --prefix="algorand-$FULLVERSION/" "$BRANCH" | gzip >| "$PKG_DIR/algorand_${CHANNEL}_source_${VERSION}.tar.gz"```

btw - could you try and spellcheck this one ?

File: scripts/release/mule/sign/linux/amd64/sign.sh (hunk 2 above)
Comment (btoll) [reply scripts/release/mule/sign/linux/amd64/sign.sh:13]: Hmm, I should change the order of those params.

Comment (btoll) [reply scripts/release/mule/sign/linux/amd64/sign.sh:13]: I'll correct the other targets in another PR.

File: scripts/release/mule/sign/linux/amd64/sign.sh (hunk 3 above)
Comment (btoll) [reply scripts/release/mule/sign/linux/amd64/sign.sh:3]: I'll add pipefail.  I have shellcheck integrated with vim, so it will always check on saving/closing the file.

It didn't flag the pipeline. Is there anything in particular that caught your attention?

File: scripts/release/mule/sign/linux/amd64/sign.sh (hunk 2 above)
Comment (btoll) [reply scripts/release/mule/sign/linux/amd64/sign.sh:13]: Changed.

File: docker/build/mule.go.debian.Dockerfile [hunk 4]
```diff
@@ -4,11 +4,11 @@ FROM python:3.7
 COPY --from=docker /usr/local/bin/docker /usr/local/bin/docker
 COPY *.yaml /root/
 
-RUN pip install mulecli
+RUN apt-get update && apt-get install -y autoconf bsdmainutils build-essential curl git libboost-all-dev && \
+    curl https://dl.google.com/go/go1.12.linux-amd64.tar.gz | tar -xzf - && \
+    mv go /usr/local && \
+    pip install mulecli
```
btoll, line 10: I'm only hardcoding the version of golang here since this will be changed very shortly to reference a single location that will download and install golang.

scripts/release/mule/Makefile.mule
```diff
@@ -46,32 +46,30 @@ ci-build: ci-setup buildsrc gen
...
-package-rpm-$(OS_TYPE)-$(ARCH):
-	scripts/release/mule/package/$(OS_TYPE)/$(ARCH)/rpm/package.sh $(OS_TYPE) $(ARCH) $$WORKDIR
-
-package-docker-$(OS_TYPE)-$(ARCH):
+package-docker:
 	scripts/release/mule/package/$(OS_TYPE)/$(ARCH)/docker/package.sh $(OS_TYPE) $(ARCH) $$WORKDIR
 
-package-test-$(OS_TYPE)-$(ARCH): package-test-deb-$(OS_TYPE)-$(ARCH) package-test-rpm-$(OS_TYPE)-$(ARCH)
+package-rpm:
+	scripts/release/mule/package/$(OS_TYPE)/$(ARCH)/rpm/package.sh $(OS_TYPE) $(ARCH) $$WORKDIR
```
egieseke, line 56: Why are the scripts specific to os type and arch, and also take those in as arguments?

btoll: I thought about hardcoding the OS_TYPE and ARCH in the shell scripts, but for now I'd like to leave that as generic as possible for possible re-use.

scripts/release/mule/sign/linux/amd64/sign.sh
tsachiherman: Yes, if the git command returns a non-zero exit code with no file names, the gzip would not fail, and set the error code to zero. that's not what we want, right ?

btoll: Yes, that makes sense and I've added the pipefail option.  I was curious about why you were asking for the shellcheck; is it because of the `>|` redirect?

scripts/release/mule/Makefile.mule
egieseke: rather remove the os_type and arch from the path?

btoll: We're going to want the flexibility of building packages for different operating systems and architectures.

docker/build/mule.go.debian.Dockerfile
bricerisingalgorand: I think this one is go 1.12.0, we may want to use g1.12.17 here instead. They're pretty consistent with their urls also so you could do something like this to make it configurable:

```
ARG GO_VERSION=1.12
RUN apt-get update && apt-get install -y autoconf bsdmainutils build-essential curl git libboost-all-dev && \
    curl https://dl.google.com/go/go${GO_VERSION}.linux-amd64.tar.gz | tar -xzf - && \
    mv go /usr/local && \
    pip install mulecli
```

scripts/release/mule/Makefile.mule
bricerisingalgorand: Could we make one package script that generically packages all builds in the tmp/node_pkg folder? The os/arch information is in the path to each bin folder under that directory so this information could be captured there right?

bricerisingalgorand: To minimize the amount of rewriting necessary, we could have a package script that references these os/arch specific scripts for each folder in the tmp/node_pkg dir. This could be better for us since then we can use one jenkins agent to do all of our package, rather than using extra compute to spin up an agent for each os/arch type

package-sign.yaml
```diff
@@ -0,0 +1,113 @@
...
+      workDir: /projects/go-algorand
+      env: [
+        AWS_ACCESS_KEY_ID=$AWS_ACCESS_KEY_ID,
+        AWS_SECRET_ACCESS_KEY=$AWS_SECRET_ACCESS_KEY
+      ]
+      volumes: [
+        $XDG_RUNTIME_DIR/gnupg/S.gpg-agent:/root/.gnupg/S.gpg-agent,
+        $HOME/.gnupg/pubring.kbx:/root/.gnupg/pubring.kbx
+      ]
+    target: package-sign-rpm WORKDIR=/projects/go-algorand
```
bricerisingalgorand, line 42: did the makefile SRCPATH var not work for workdir?

package-test.yaml
```diff
@@ -6,7 +6,7 @@ tasks:
     name: package
     image: algorand/mule-linux-ubuntu
     version: '{{ docker.Version.outputs.version }}'
-    dockerFilePath: docker/build/mule.ubuntu.Dockerfile
+    dockerFilePath: docker/build/mule.go.debian.Dockerfile
```
onetechnical, line 9: If we're tagging this as being sourced from Debian, then shouldn't we update the image names from mule-linux-ubuntu in this file?

PR #1053, opened 2020-05-07
Parameter support for dispenser

rfustino:
## Summary

This feature will allow for a more seemless integrtion with Mobile Apps and Websties that use TestNet and BetaNet Dispensers

### DIFF
--- a/cmd/dispenser/server.go
+++ b/cmd/dispenser/server.go
@@ -75,14 +75,9 @@ const topPageTemplate = `
 	  
 	 <script>
 	  function loadparm(){
-			const queryString = window.location.search;
-			const urlParams = new URLSearchParams(queryString);
-			const parmTarget = urlParams.get('account');
-			var target1 = document.getElementById('target');
-			target1.value = "";
-			target1.value = parmTarget;
-			var $target = $('target');
-			$target.html(target1.value);
+				const queryString = window.location.search;
+				const urlParams = new URLSearchParams(queryString);
+				$('#target').html(urlParams.get('account'));
 	  }  
       function onload() {
             $('#dispense').click(function (e) {
--- a/cmd/dispenser/server.go
+++ b/cmd/dispenser/server.go
@@ -75,9 +75,9 @@ const topPageTemplate = `
 	  
 	 <script>
 	  function loadparm(){
-			const queryString = window.location.search;
-			const urlParams = new URLSearchParams(queryString);
-			$('#target').val(urlParams.get('account'));
+            const queryString = window.location.search;
+            const urlParams = new URLSearchParams(queryString);
+            $('#target').val(urlParams.get('account'));
 	  }  
       function onload() {
             $('#dispense').click(function (e) {
@@ -93,9 +93,8 @@ const topPageTemplate = `
               }).fail(function () {
                 $('#status').html('Code ' + req.status + ' ' + req.statusText + ': ' + req.responseText);
               });
-		 });
-		 
-		loadparm();
+		 });		 
+            loadparm();
       }
     </script>
   </head>
--- a/cmd/dispenser/server.go
+++ b/cmd/dispenser/server.go
@@ -75,9 +75,9 @@ const topPageTemplate = `
 	  
 	 <script>
 	  function loadparm(){
-				const queryString = window.location.search;
-				const urlParams = new URLSearchParams(queryString);
-				$('#target').html(urlParams.get('account'));
+			const queryString = window.location.search;
+			const urlParams = new URLSearchParams(queryString);
+			$('#target').val(urlParams.get('account'));
 	  }  
       function onload() {
             $('#dispense').click(function (e) {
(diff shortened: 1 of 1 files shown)


rfustino:
This feature will allow for a more seemless integrtion with Mobile Apps and Websties that use TestNet and BetaNet Dispensers

cmd/dispenser/server.go
```diff
@@ -71,14 +71,24 @@ const topPageTemplate = `
...
-      crossorigin="anonymous"></script>
-
-    <script>
+	   crossorigin="anonymous"></script>
+	  
+	 <script>
+	  function loadparm(){
+			const queryString = window.location.search;
+			const urlParams = new URLSearchParams(queryString);
+			const parmTarget = urlParams.get('account');
```
tsachiherman, line 80: you need to check if urlParams.has('account'), since if it wasn't passed on, you don't want to attempt and assign the value.

cmd/dispenser/server.go
```diff
@@ -71,14 +71,24 @@ const topPageTemplate = `
...
+			var target1 = document.getElementById('target');
+			target1.value = "";
+			target1.value = parmTarget;
+			var $target = $('target');
+			$target.html(target1.value);
```
tsachiherman, line 85: what's the purpose of having `target1` ? I think that you can do
$('#target').html(urlParams.get('account'))

rfustino: Yes, that is a cleaner way to do it @tsachiherman . Also, if urlParams.get('account') returns null it will be now setting the target.html to null which should be ok. Do I need to re-submit this? 

    const queryString = window.location.search;
    const urlParams = new URLSearchParams(queryString);
    $('#target').html(urlParams.get('account'))

cmd/dispenser/server.go
rfustino: If it was not passed in, the value assigned will be null

cmd/dispenser/server.go
```diff
@@ -71,14 +71,19 @@ const topPageTemplate = `
...
-      crossorigin="anonymous"></script>
-
-    <script>
+	   crossorigin="anonymous"></script>
+	  
+	 <script>
+	  function loadparm(){
+				const queryString = window.location.search;
+				const urlParams = new URLSearchParams(queryString);
+				$('#target').html(urlParams.get('account'));
```
justicz, line 80: This introduces an [XSS vulnerability](https://portswigger.net/web-security/cross-site-scripting). Instead of calling `$('#target').html`, please use something like `$('#target').val`, which is safe to call with this untrusted parameter.

cmd/dispenser/server.go
```diff
@@ -71,14 +71,19 @@ const topPageTemplate = `
...
+	  }  
       function onload() {
             $('#dispense').click(function (e) {
-              var recaptcha = grecaptcha.getResponse();
-              var target = $('#target').val();
-
+			   var recaptcha = grecaptcha.getResponse();
```
justicz, line 84: nitpick, but please fix the indentation here and below -- the rest of this file uses spaces.

rfustino: OK. This has been fixed. Thanks! Better than new. :)

cmd/dispenser/server.go
rfustino: $('#target').val is working! and inserted. Thanks for the feedback.

cmd/dispenser/server.go
rfustino: changed to $('#target').val  and this works.

rfustino: changed this
$('#target').html(urlParams.get('account'));
to this
$('#target').val(urlParams.get('account'));

cmd/dispenser/server.go
rfustino: changed

cmd/dispenser/server.go
```diff
@@ -88,7 +93,8 @@ const topPageTemplate = `
               }).fail(function () {
                 $('#status').html('Code ' + req.status + ' ' + req.statusText + ': ' + req.responseText);
               });
-        });
+		 });
```
justicz, line 96: this is still tabs

cmd/dispenser/server.go
```diff
@@ -71,14 +71,19 @@ const topPageTemplate = `
...
+            $('#target').val(urlParams.get('account'));
+	  }  
       function onload() {
             $('#dispense').click(function (e) {
-              var recaptcha = grecaptcha.getResponse();
-              var target = $('#target').val();
-
+               var recaptcha = grecaptcha.getResponse();
+               var target = $('#target').val();
+
```
justicz, line 86: tabs

cmd/dispenser/server.go
```diff
@@ -71,14 +71,19 @@ const topPageTemplate = `
...
+            const queryString = window.location.search;
+            const urlParams = new URLSearchParams(queryString);
+            $('#target').val(urlParams.get('account'));
+	  }  
       function onload() {
             $('#dispense').click(function (e) {
-              var recaptcha = grecaptcha.getResponse();
-              var target = $('#target').val();
-
+               var recaptcha = grecaptcha.getResponse();
```
justicz, line 84: these two lines have an extra space now

PR #1061, opened 2020-05-08
add `goal protocols`

algobolson:
add a goal command to dump all known protocols as json. Useful for other tools to import (Indexer).

### DIFF
--- a/cmd/goal/commands.go
+++ b/cmd/goal/commands.go
@@ -224,7 +224,7 @@ var reportCmd = &cobra.Command{
 var protoCmd = &cobra.Command{
 	Use:   "protocols",
 	Short: "",
-	Long:  "Dump known consensus protocols as json to stdout.",
+	Long:  "Dump standard consensus protocols as json to stdout.",
 	Args:  validateNoPosArgsFn,
 	Run: func(cmd *cobra.Command, args []string) {
 		os.Stdout.Write(protocol.EncodeJSON(config.Consensus))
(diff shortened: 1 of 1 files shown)


cmd/goal/commands.go
```diff
@@ -220,6 +221,16 @@ var reportCmd = &cobra.Command{
 	},
 }
 
+var protoCmd = &cobra.Command{
+	Use:   "protocols",
+	Short: "",
+	Long:  "Dump known consensus protocols as json to stdout.",
```
tsachiherman, line 227: could you rephrase it to "Dump built-in consensus.." ?
We do support dynamically loaded consensus protocols, so saying "built-in" would better describe these.

algobolson: how about "standard",  "Dump standard consensus protocols ..."

tsachiherman: Sounds good

cmd/goal/commands.go
```diff
@@ -220,6 +221,16 @@ var reportCmd = &cobra.Command{
...
+	Args:  validateNoPosArgsFn,
+	Run: func(cmd *cobra.Command, args []string) {
+		os.Stdout.Write(protocol.EncodeJSON(config.Consensus))
```
pzbitskiy, line 230: mb also `-o` to dump as msgp

tsachiherman: I think that @pzbitskiy is a good idea, but let's finish this one first and consider to do the other in a separate PR.

PR #1070, opened 2020-05-12
Simplify Makefile.mule

btoll:
This isn't the final pass, but it's much better than what is currently there.

https://scene-si.org/2019/12/04/make-dynamic-makefile-targets/

### DIFF
--- a/package-test.yaml
+++ b/package-test.yaml
@@ -4,7 +4,7 @@ tasks:
 
   - task: shell.docker.Ensure
     name: package
-    image: algorand/mule-linux-ubuntu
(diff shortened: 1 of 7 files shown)


package-test.yaml
```diff
@@ -4,36 +4,36 @@ tasks:
 
   - task: shell.docker.Ensure
     name: package
-    image: algorand/mule-linux-ubuntu
+    image: algorand/mule-linux-debian
```
btoll, line 7: @onetechnical These are the changes you requested on the other PR.

PR #1072, opened 2020-05-13
Eric/use mule in jenkins

egieseke:
## Summary

Create a Jenkins mule pipeline to build and test go-algorand.
https://github.com/algorand/go-algorand-internal/issues/424

## Test Plan

Setup and executed Jenkins pipeline:

https://jenkins.algodev.network/job/JenkinsMulePipeline/


test/e2e-go/cli/goal/expect/goalExpectCommon.exp
```diff
@@ -142,14 +142,14 @@ proc ::AlgorandGoal::StartNetwork { NETWORK_NAME NETWORK_TEMPLATE TEST_ALGO_DIR
 
     # Running on ARM64, it seems that network creation is pretty slow.
     # 30 second won't be enough here, so I'm changing this to 90 seconds.
```
winder, line 144: Update comment. If you know why 90 seconds is no longer enough maybe add that in too.

test/e2e-go/restAPI/restClient_test.go
```diff
@@ -146,7 +146,7 @@ func waitForRoundOne(t *testing.T, testClient libgoal.Client) {
 	select {
 	case err := <-errchan:
 		require.NoError(t, err)
-	case <-time.After(1 * time.Minute): // Wait 1 minute (same as WaitForRound)
+	case <-time.After(2 * time.Minute): // Wait 1 minute (same as WaitForRound)
```
winder, line 149: Update (or remove) comment.

test/e2e-go/cli/goal/expect/goalExpectCommon.exp
egieseke: Updated comment.

test/e2e-go/restAPI/restClient_test.go
egieseke: Updated comment.

network/ping_test.go
```diff
@@ -60,7 +60,7 @@ func TestPing(t *testing.T) {
 		if lastPingRoundTripTime > 0 {
 			postPing := time.Now()
 			testTime := postPing.Sub(prePing)
-			if (lastPingRoundTripTime < testTime) && (lastPingRoundTripTime > (testTime - (2 * waitStep))) {
+			if (lastPingRoundTripTime < testTime) && (lastPingRoundTripTime > (testTime - ((time.Duration(i) * time.Second) * waitStep))) {
```
tsachiherman, line 63: Could you please explain why this was a problem to start with, and why running with mule requires this ?

scripts/upload_version.sh
```diff
@@ -24,4 +24,5 @@ DIRECTORY=$2
 BUCKET=$3
 
 export GOPATH=$(go env GOPATH)
-${GOPATH}/bin/updater send -s "${DIRECTORY}" -c "${CHANNEL}" -b "${BUCKET}"
+export PATH=${PATH}:${GOPATH}/bin
+updater send -s "${DIRECTORY}" -c "${CHANNEL}" -b "${BUCKET}"
```
tsachiherman, line 28: Could you please explain why the changes to this file are needed ?
I can see why this would work, but unless you need the PATH being set from within the updater, I don't see any benefit here.

test/e2e-go/features/transactions/group_test.go
```diff
@@ -156,7 +156,7 @@ func TestGroupTransactionsDifferentSizes(t *testing.T) {
 
 		// wait for the txids and check balances
 		_, curRound := fixture.GetBalanceAndRound(account0)
-		confirmed := fixture.WaitForAllTxnsToConfirm(curRound+5, txids)
+		confirmed := fixture.WaitForAllTxnsToConfirm(curRound+10, txids)
```
tsachiherman, line 159: I can't see any reason why this would fix a failure in `a.True(confirmed, "txgroup")`. We might have an issue elsewhere, but increasing the wait time wouldn't help.

egieseke: Increasing the max round allowed the test to pass.

tsachiherman: sure. (btw : deleting this test would also allow it to pass )
so you've just confirmed that we have a bug somewhere, maybe here, or maybe elsewhere.
we really need to find where the bug is and fix it rather that wait longer.
If you can't fix it, disable it for your configuration by testing the environment variables and/or platform at the beginning of the test.

test/e2e-go/restAPI/restClient_test.go
```diff
@@ -146,7 +146,7 @@ func waitForRoundOne(t *testing.T, testClient libgoal.Client) {
 	select {
 	case err := <-errchan:
 		require.NoError(t, err)
-	case <-time.After(1 * time.Minute): // Wait 1 minute (same as WaitForRound)
+	case <-time.After(2 * time.Minute): // Wait 2 minutes (same as WaitForRound)
```
tsachiherman, line 149: why ? The goal of this timeout was to be 100% aligned with the one in WaitForRound in the algod.
If a test you're running is failing, you should find out why it is failing.

Makefile
```diff
@@ -155,8 +156,14 @@ $(KMD_API_SWAGGER_INJECT): $(KMD_API_SWAGGER_SPEC) $(KMD_API_SWAGGER_SPEC).valid
...
 
+# We're making an empty file in the go-cache dir to
+# get around a bug in go build where it will fail
+# to cache binaries from time to time on empty NFS
+# dirs
 buildsrc: crypto/libs/$(OS_TYPE)/$(ARCH)/lib/libsodium.a node_exporter NONGO_BIN deps $(ALGOD_API_SWAGGER_INJECT) $(KMD_API_SWAGGER_INJECT)
-	go install $(GOTRIMPATH) $(GOTAGS) $(GOBUILDMODE) -ldflags="$(GOLDFLAGS)" ./...
+	mkdir -p tmp/go-cache && \
+	touch tmp/go-cache/file.txt && \
+	GOCACHE=$(SRCPATH)/tmp/go-cache go install $(GOTRIMPATH) $(GOTAGS) $(GOBUILDMODE) -ldflags="$(GOLDFLAGS)" ./...
```
tsachiherman, line 166: could you provide some background why the GOCACHE need to be modified ?

docker/build/Dockerfile
```diff
@@ -6,7 +6,7 @@ WORKDIR /root
 RUN wget --quiet https://dl.google.com/go/go${GOLANG_VERSION}.linux-amd64.tar.gz && tar -xvf go${GOLANG_VERSION}.linux-amd64.tar.gz && mv go /usr/local
 ENV GOROOT=/usr/local/go \
     GOPATH=$HOME/go \
-    GOPROXY=https://gocenter.io
+    GOPROXY=https://gocenter.io,https://goproxy.io,direct
```
onetechnical, line 9: This won't work until we upgrade to golang 1.13, will it?

docker/build/Dockerfile-deploy
```diff
@@ -6,7 +6,7 @@ WORKDIR /root
 RUN wget --quiet https://dl.google.com/go/go${GOLANG_VERSION}.linux-amd64.tar.gz && tar -xvf go${GOLANG_VERSION}.linux-amd64.tar.gz && mv go /usr/local
 ENV GOROOT=/usr/local/go \
     GOPATH=$HOME/go \
-    GOPROXY=https://gocenter.io
+    GOPROXY=https://gocenter.io,https://goproxy.io,direct
```
onetechnical, line 9: Same here

docker/build/cicd.alpine.Dockerfile
```diff
@@ -24,7 +24,7 @@ RUN apk add dpkg && \
 COPY . $GOPATH/src/github.com/algorand/go-algorand
 WORKDIR $GOPATH/src/github.com/algorand/go-algorand
 ENV GCC_CONFIG="--with-arch=armv6" \
-    GOPROXY=https://gocenter.io
+    GOPROXY=https://gocenter.io,https://goproxy.io,direct
```
onetechnical, line 27: ...and here. This will be a pain to update later, won't it.

docker/build/cicd.centos.Dockerfile
```diff
@@ -1,12 +1,15 @@
...
-ARG ARCH="amd64"
+
 RUN yum install -y https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm && \
     yum update -y && \
-    yum install -y autoconf wget awscli git gnupg2 nfs-utils python36 sqlite3 boost-devel expect jq libtool gcc-c++ libstdc++-devel libstdc++-static rpmdevtools createrepo rpm-sign bzip2 which ShellCheck
+    yum install -y autoconf wget awscli git gnupg2 nfs-utils python3-devel sqlite3 boost-devel expect jq \
+    libtool gcc-c++ libstdc++-devel libstdc++-static rpmdevtools createrepo rpm-sign bzip2 which ShellCheck \
+    libffi-devel openssl-devel
 WORKDIR /root
+ARG ARCH="amd64"
```
onetechnical, line 11:
```suggestion
ARG ARCH
```

scripts/configure_dev.sh
```diff
@@ -32,6 +58,11 @@ elif [ "${OS}" = "darwin" ]; then
     install_or_upgrade autoconf
     install_or_upgrade automake
     install_or_upgrade shellcheck
+    install_or_upgrade python3
+fi
+
+if [[ ${SKIP_GO_DEPS} == "True" ]]; then
```
btoll, line 64: Bash supports booleans, you can just do `if $SKIP_GO_DEPS`

test/scripts/e2e.sh
```diff
@@ -53,15 +62,19 @@ reset_dirs
 echo Killing all instances and installing current build
 
 pkill -u $(whoami) -x algod || true
-./scripts/local_install.sh -c ${CHANNEL} -p ${BINDIR} -d ${DATADIR}
 
+if [[ ! ${NO_BUILD} == "True" ]]; then
```
btoll, line 66: Same as my comment above, you can use a boolean here rather then checking a string.

scripts/build_packages.sh
```diff
@@ -55,8 +55,6 @@ if [ "${PKG_ROOT}" = "" ]; then
     PKG_ROOT=${HOME}/node_pkg
 fi
 
-rm -rf ${PKG_ROOT}
```
btoll, line 58: I think we may have some assumptions with our scripting that there is only one build version in `PKG_ROOT`.  I'm not sure we'd want to change this behavior without testing.

scripts/configure_dev.sh
```diff
@@ -1,15 +1,41 @@
...
+    h ) echo "${HELP}"
+        exit 0
+      ;;
+    \? ) echo "${HELP}"
+        exit 2
+      ;;
+  esac
+done
+
 SCRIPTPATH="$( cd "$(dirname "$0")" ; pwd -P )"
```
btoll, line 27: Is this needed?  Is this ever invoked directly from the cli rather than through scripting?

scripts/configure_dev.sh
```diff
@@ -1,15 +1,41 @@
...
+    f ) FORCE="True"
+      ;;
+    h ) echo "${HELP}"
+        exit 0
+      ;;
+    \? ) echo "${HELP}"
+        exit 2
+      ;;
+  esac
+done
```
btoll, line 25: Is this needed? Is this ever invoked directly from the cli rather than through scripting?

bricerisingalgorand: This was actually done for travis since we needed to run this script for each build. Since we removed the travis changes, this isn't 100% necessary anymore

Makefile
bricerisingalgorand: I'll need to do some digging to find out what the bug was, but we found a bug where go installs fail a small percentage of the time when the go-cache is empty on travis. Since we're not changing the travis job this may actually not be necessary anymore

docker/build/Dockerfile
bricerisingalgorand: Yea, but the makefile overwrites this with `GOPROXY=direct`. We figured since we already found all the places to set this goproxy setting, and it doesn't really affect our ci process at the moment, we could leave them changed for the future. An argument could be made that that we should remove them from everywhere except for the Makefile

docker/build/cicd.centos.Dockerfile
bricerisingalgorand: Do you not like having a default for our arch? That's ok, but it just means that whenever this Dockerfile is used, we'll have to use the --build-arg option

network/ping_test.go
bricerisingalgorand: This test failed a lot in lower compute environments because the testTime would be very long. I thought having a gradually increasing wait time would be better, but it's possible I'm missing exactly what this is trying to exercise

scripts/upload_version.sh
bricerisingalgorand: We made this change because in the pipeline we're storing built binaries, and then running the uploading on another host later in the process. We have our makefile set up to download those binaries and add them to the path. We left `export PATH=${PATH}:${GOPATH}/bin` in here as a failsafe in case someone calls this script not from the make targets so the go bin will still be part of the path.

scripts/build_packages.sh
egieseke: We do assume that there can be multiple builds under PKG_ROOT, so we don't want to delete what's already there.

scripts/configure_dev.sh
egieseke: OK, will convert to booleans.

test/e2e-go/features/transactions/group_test.go
egieseke: Why is 5 the correct number of rounds?

test/scripts/e2e.sh
egieseke: OK, will change to use booleans.

docker/build/cicd.centos.Dockerfile
onetechnical: If you pass without arg it'll just pull the value from before the FROM line.

bricerisingalgorand:
It looks like this test failed on the osx travis job

```
--- FAIL: TestAccountsCanClose (105.66s)
    restClientFixture.go:249: 
        	Error Trace:	restClientFixture.go:249
        	            				close_account_test.go:63
        	Error:      	Received unexpected error:
        	            	fixture.WaitForRound took 100.14 seconds between round 0 and 1
        	Test:       	TestAccountsCanClose
        	Messages:   	fixture should be able to wait for one round to pass
FAIL
FAIL	github.com/algorand/go-algorand/test/e2e-go/features/transactions	297.823s
```

tsachiherman:
> It looks like this test failed on the osx travis job
> 
> ```
> --- FAIL: TestAccountsCanClose (105.66s)
>     restClientFixture.go:249: 
>         	Error Trace:	restClientFixture.go:249
>         	            				close_account_test.go:63
>         	Error:      	Received unexpected error:
>         	            	fixture.WaitForRound took 100.14 seconds between round 0 and 1
>         	Test:       	TestAccountsCanClose
>         	Messages:   	fixture should be able to wait for one round to pass
> FAIL
> FAIL	github.com/algorand/go-algorand/test/e2e-go/features/transactions	297.823s
> ```

Yes, and I'm willing to bet it has nothing to do with OS X or travis, but rather a bug in either the node, test framework or test itself.

test/e2e-go/restAPI/restClient_test.go
bricerisingalgorand: Thanks, I changed it back

Makefile
bricerisingalgorand: I can't find where that solution was proposed, but here's an issue about that problem. Apparently this error happens a small percentage of the time on certain types of drive mounts

https://github.com/golang/go/issues/29667

scripts/upload_version.sh
tsachiherman: I see. so you basically saying that you want to give priority to the updater in the path before taking the updater from the GOPATH. In this case, I have only one request - make it use the GOPATH1 ( i.e. the first path in the GOPATH)

test/e2e-go/features/transactions/group_test.go
tsachiherman: I don't know that 5 was correct. But if you're changing this into another number, you'd better have a good reasoning why the new number is "correct" and the previous is "incorrect".

In particular, you need to ask yourself why you needed to make this change.

network/ping_test.go
tsachiherman: Could you please have @algobolson review this change ? He wrote this unit test, and could suggest what the best course of action would be.

brianolson: This is perhaps a slightly foolish timing assertion. The first half `(lastPingRoundTripTime < testTime)` says that the measured ping round trip time is less than the time observed at the test from before the ping is sent until after the reply is received. If this fails then ping time is being measured wrong or the test system clock is insane.
The second half `(lastPingRoundTripTime > (testTime - (2 * waitStep)))` says that the polling for ping reply didn't wait more than 2x the polling period to get the reply. _maybe this doesn't matter and we should just delete this half of the expression_ ?

Makefile
bricerisingalgorand: That makes sense. Is it alright for us to make a ticket to develop a better solution?

tsachiherman: sure.

test/scripts/e2e.sh
```diff
@@ -82,10 +95,9 @@ export TESTDIR=${TEMPDIR}
 export TESTDATADIR=${SRCROOT}/test/testdata
 export SRCROOT=${SRCROOT}
 
-./e2e_go_tests.sh
+./e2e_go_tests.sh ${GO_TEST_ARGS}
 
 rm -rf ${TEMPDIR}
-rm -rf ${PKG_ROOT}
```
tsachiherman, line 88: don't you need this one when 
`! ${NO_BUILD}` ?

test/scripts/e2e_subs/keyreg-teal-test.sh
```diff
@@ -39,8 +39,8 @@ cat<<EOF|python - > ${TEMPDIR}/pbound
 print(((${ROUND} // ${PERIOD}) * ${PERIOD}) + ${PERIOD})
 EOF
 PBOUND=$(cat ${TEMPDIR}/pbound)
-while [ ${ROUND} != ${PBOUND} ]; do
-    goal node wait
+while [ ${ROUND} -lt ${PBOUND} ]; do
+    goal node wait --waittime 30
     ROUND=$(goal node status | grep 'Last committed block:'|awk '{ print $4 }')
 done
```
tsachiherman, line 45: this is a very nice bug that you've fixed. it should have it's own PR.

egieseke: @tsachiherman, Moved this change to a separate PR.

test/scripts/e2e.sh
egieseke: @tsachiherman The build_packages script now cleans up the platform root before it builds.

tsachiherman: yes, build_packages is ok to clean up before creating new packages.
however, it would be nice if this script would leave a clean slate after it's execution - after all, the goal of e2e.sh is to conduct the e2e test, not to perform a build. Once the test is completed, there is no point in leaving these files around.

test/scripts/e2e_basic_start_stop.sh
```diff
@@ -55,17 +55,18 @@ verify_none_running
...
-${BINDIR}/goal node stop -d ${DATADIR}
+goal node stop -d ${DATADIR}
 verify_none_running
 
 #----------------------
 # Test that we can start a generic node straight with no overrides
 echo Verifying a generic node will start directly
-${BINDIR}/algod -d ${DATADIR} &
+algod -d ${DATADIR} &
+sleep 1
```
tsachiherman, line 69: If this does becomes an issue, this is not the correct place to fix it. you might want to try to sleep in case the pgrep  returns no pids in update_running_count

test/scripts/e2e_subs/keyreg-teal-test.sh
egieseke: moved to a separate PR.

test/scripts/e2e.sh
egieseke: @tsachiherman OK, added back the removal of PKG_ROOT with conditional: ! NO_BUILD.

test/scripts/e2e_basic_start_stop.sh
brianolson: I'm surprised the answer isn't to make verify_at_least_one_running have a retry loop like verify_one_running has

egieseke: I just updated verify_at_least_one_running to include the retry loop and removed the sleep statements. Good thinking.

test/e2e-go/cli/goal/expect/goalExpectCommon.exp
```diff
@@ -784,7 +795,7 @@ proc ::AlgorandGoal::AddParticipationKey { ADDRESS FIRST_ROUND LAST_ROUND TEST_P
 # Register online participation with a given account
 proc ::AlgorandGoal::TakeAccountOnline { ADDRESS FIRST_ROUND LAST_ROUND TEST_PRIMARY_NODE_DIR } {
    if { [ catch {
-       spawn goal account changeonlinestatus --address $ADDRESS --firstvalid $FIRST_ROUND --lastvalid $LAST_ROUND -d datadir
+       spawn goal account changeonlinestatus --address $ADDRESS --firstvalid $FIRST_ROUND --lastvalid $LAST_ROUND -d datadir --online
```
tsachiherman, line 798: this change is supposed to be a no-op. i.e. the default for `--online` is true. If this is not the case, then we have a bug, which is unrelated to this PR.

egieseke: `--online` is required by the `goal account changeonlinestatus` command.    Without it,  the command fails.  I will create a ticket to fix the command and update this test case to remove the `--online` option.

tsachiherman: The PR looks good to me, but I wonder how this test did not detect this issue before.
It's quite concerning that we have a test that is supposed to test exactly one thing - and the test doesn't alert us when that thing is broken.

egieseke: We found this issue by running the integration tests as part of the Jenkins/Mule pipeline testing.  The issue was causing the test test/e2e-go/cli/goal/expect/listExpiredParticipationKeyTest.exp to fail.  What is puzzling is why this error does not surface as part of the Travis build and testing.  I created this ticket to resolve: https://github.com/algorand/go-algorand-internal/issues/542

egieseke:
> Hmm, the mule-specific changes here look okay at a glance, but the test changes give me pause. Is it possible to add commentary around changes as to why they were made?

The resource-constrained platforms require more time to complete the steps of the tests. Especially for the expect tests that are running 2 node networks for goal to interact with.

scripts/configure_dev.sh
```diff
@@ -32,6 +60,11 @@ elif [ "${OS}" = "darwin" ]; then
     install_or_upgrade autoconf
     install_or_upgrade automake
     install_or_upgrade shellcheck
+    install_or_upgrade python3
```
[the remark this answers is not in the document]
egieseke: Currently installing 3.7.7, but subject to change at homebrew's discretion.

scripts/release/mule/Makefile.mule
```diff
@@ -26,25 +26,19 @@ endif
...
-	PKG_ROOT=$(SRCPATH)/tmp/node_pkgs/$(OS_TYPE)/$(ARCH) \
-		 NO_BUILD=True VARIATIONS=$(OS_TYPE)/$(ARCH) \
-		 scripts/build_packages.sh $(OS_TYPE)/$(ARCH)
+	test/scripts/e2e.sh -c $(BUILDCHANNEL) -n
+
+ci-build: buildsrc gen
+	mkdir -p $(SRCPATH)/tmp/node_pkgs && \
+	CHANNEL=$(BUILDCHANNEL) PKG_ROOT=$(SRCPATH)/tmp/node_pkgs NO_BUILD=True VARIATIONS=$(OS_TYPE)-$(ARCH) \
+	scripts/build_packages.sh $(OS_TYPE)/$(ARCH) && \
+	mkdir -p $(SRCPATH)/tmp/node_pkgs/$(BUILDCHANNEL)/$(OS_TYPE)-$(ARCH)/data && \
```
btoll, line 40: I'm pretty sure that this is going to break the existing mule tasks (see the `package*.yaml` files in the root dir) because this path is different.

If you look at the `PKG_ROOT` var in the deleted snippet above it's expecting the build assets to be put in a different path than the one you have now.

scripts/release/mule/Makefile.mule
```diff
@@ -26,25 +26,19 @@ endif
...
-	./test/scripts/e2e_go_tests.sh
-else
-	export NODEBINDIR=$(SRCPATH)/tmp/node_pkgs/$(OS_TYPE)/$(ARCH)/dev/$(OS_TYPE)-$(ARCH)/bin && \
-	export PATH=$(SRCPATH)/tmp/node_pkgs/$(OS_TYPE)/$(ARCH)/dev/$(OS_TYPE)-$(ARCH)/bin:$$PATH && \
-	export PATH=$(SRCPATH)/tmp/node_pkgs/$(OS_TYPE)/$(ARCH)/dev/$(OS_TYPE)-$(ARCH)/test-utils:$$PATH && \
+	export NODEBINDIR=$(SRCPATH)/tmp/node_pkgs/$(BUILDCHANNEL)/$(OS_TYPE)-$(ARCH)/bin && \
+	export PATH=$(SRCPATH)/tmp/node_pkgs/$(BUILDCHANNEL)/$(OS_TYPE)-$(ARCH)/bin:$$PATH && \
+	export PATH=$(SRCPATH)/tmp/node_pkgs/$(BUILDCHANNEL)/$(OS_TYPE)-$(ARCH)/tools:$$PATH && \
+	export PATH=$(SRCPATH)/tmp/node_pkgs/$(BUILDCHANNEL)/$(OS_TYPE)-$(ARCH)/test-utils:$$PATH && \
 	export SRCROOT=$(SRCPATH) && \
```
btoll, line 33: Looks like these env vars should be precede the call to the shell script rather than exporting them before the call to the script.

egieseke: Look below to see where the call to test/scripts/e2e.sh is appended to the series of exports.

btoll: The script needs the env vars, not the calling environment.  It's a common shell idiom to do:
```
FOO=foo BAR=bar my_script.sh
```
Also, the `ci-build` target is doing it that way and it would be great to keep that consistency.

I realize this was existing code, but it would be great if you could change it.

scripts/release/mule/Makefile.mule
egieseke: Please see updates.

scripts/release/mule/Makefile.mule
egieseke: @btoll please see updates.

scripts/release/mule/Makefile.mule
```diff
@@ -26,25 +26,18 @@ endif
...
-		 scripts/build_packages.sh $(OS_TYPE)/$(ARCH)
+	NODEBINDIR=$(SRCPATH)/tmp/node_pkgs/$(OS_TYPE)/$(ARCH)/$(BUILDCHANNEL)/$(OS_TYPE)-$(ARCH)/bin \
+	PATH=$(SRCPATH)/tmp/node_pkgs/$(OS_TYPE)/$(ARCH)/$(BUILDCHANNEL)/$(OS_TYPE)-$(ARCH)/bin:$$PATH \
+	PATH=$(SRCPATH)/tmp/node_pkgs/$(OS_TYPE)/$(ARCH)/$(BUILDCHANNEL)/$(OS_TYPE)-$(ARCH)/tools:$$PATH \
+	PATH=$(SRCPATH)/tmp/node_pkgs/$(OS_TYPE)/$(ARCH)/$(BUILDCHANNEL)/$(OS_TYPE)-$(ARCH)/test-utils:$$PATH \
+	SRCROOT=$(SRCPATH) \
+	test/scripts/e2e.sh -c $(BUILDCHANNEL) -n
+
+ci-build: buildsrc gen ci-setup
+	CHANNEL=$(BUILDCHANNEL) PKG_ROOT=$(SRCPATH)/tmp/node_pkgs/$(OS_TYPE)/$(ARCH) NO_BUILD=True VARIATIONS=$(OS_TYPE)-$(ARCH) \
```
btoll, line 37: Thanks for the changes.

What about version?  A local dev might want to have multiple versions built on their system.

btoll:
@egieseke Thanks for the changes, I added one more comment.

scripts/release/mule/Makefile.mule
egieseke: That's a good idea, but out of scope for this PR.

btoll:
Just approved.  I'm assuming the brittleness of the tests will be addressed in future tickets.

PR #1073, opened 2020-05-13
Change goal account and clerk commands to consistently accept account aliases for address

pbennett:
### Summary
A number of the goal account commands are inconsistent in their allowance of users specifying account aliases. Many only accept a full account address.
This change adds a simple helper to provide for commands (not all were changed) to use the default account if an account isn't specified, an account address, or an account alias.

### Test Plan
Unfortunately, there are currently no test abstractions in the codebase for the CLI and much of the logic is directly within the CLI run commands.

The individual pieces of logic are typically quite simple though so visual validation combined with test executions against testnet seem sufficient.

### DIFF
--- a/cmd/goal/account.go
+++ b/cmd/goal/account.go
@@ -114,15 +114,13 @@ func init() {
 	infoMultisigCmd.MarkFlagRequired("address")
 
 	// Balance flags
-	balanceCmd.Flags().StringVarP(&accountAddress, "address", "a", "", "Account address to retrieve balance (required)")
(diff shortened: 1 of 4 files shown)


justicz:
Thanks for the PR!

> Unfortunately, there are currently no test abstractions in the codebase for the CLI and much of the logic is directly within the CLI run commands.

If you're interested, there are some end-to-end tests for goal in this directory: https://github.com/algorand/go-algorand/tree/master/test/e2e-go/cli/goal/expect

PR #1074, opened 2020-05-13
Implement protocol.EncodeJSONStrict

### LINKED ISSUE #1076: Add EncodeJSONStrict
EncodeJSON emits integer keys for `map[int]interface{}` that is not a valid json.  EncodeJSONStrict fixes it.

pzbitskiy:
### Summary
EncodeJSONStrict is like EncodeJSON but converts integer keys to strings so that the output is valid json.

### Test Plan
Unit test ensures no integer keys and correct decoding.

Closes #1076

### DIFF
--- a/protocol/codec.go
+++ b/protocol/codec.go
@@ -34,6 +34,10 @@ var CodecHandle *codec.MsgpackHandle
 // with our settings (canonical, paranoid about decoding errors)
 var JSONHandle *codec.JsonHandle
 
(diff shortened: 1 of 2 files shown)


tsachiherman:
Pavel, could you please do the following:
1. use the template for the PR description
2. create a corresponding issue, and link this PR to the issue

I will follow up and merge it afterward.

algoskiy:
updated

PR #1088, opened 2020-05-18
Perform parameter validation on -p arguments on goal & algod

tsachiherman:
## Summary

Neither algod nor goal were validating the -p parameter ( i.e. the -p parameter is used to specify a peer list to connect to ).

The -p parameter is used extensively in our testing environment, but rarely on production code. Since the automated testing was always passing the "correct" and "valid" values, it never was an issue. However, while implementing the fast catchup, I attempted to use it incorrectly and ran into various issues as a result. Having the code verify the correctness upstream would have prevented these issues.

## Test Plan

This change is covered by an expect test.

## Issue

While not strictly required for the fast catchup, it was created during the fast catchup development, so I'll leave that as the originating issue. issue #1024

### DIFF
--- a/cmd/algod/main.go
+++ b/cmd/algod/main.go
@@ -36,8 +36,9 @@ import (
 	"github.com/algorand/go-algorand/data/bookkeeping"
 	"github.com/algorand/go-algorand/logging"
 	"github.com/algorand/go-algorand/logging/telemetryspec"
+	"github.com/algorand/go-algorand/network"
 	"github.com/algorand/go-algorand/protocol"
-	"github.com/algorand/go-algorand/tools/network"
+	toolsnet "github.com/algorand/go-algorand/tools/network"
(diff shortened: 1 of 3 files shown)


cmd/algod/main.go
```diff
@@ -248,6 +249,20 @@ func main() {
...
+		// make sure that the format of each entry is valid:
+		stagingArray := []string{}
+		for _, peer := range peerOverrideArray {
+			url, err := network.ParseHostOrURL(peer)
+			if err != nil {
+				fmt.Fprintf(os.Stderr, "Provided command line parameter '%s' is not a valid host:port pair\n", peer)
+				os.Exit(1)
+				return
+			}
+			stagingArray = append(stagingArray, url.Host)
```
algorandskiy, line 262: nit: `peerOverrideArray[idx] = url.Host` and no new vars

PR #1089, opened 2020-05-19
Rename balrecord.SpendingKey to AuthAddr

algoradam:
We're already using "spending key" to refer to private keys in various
user-facing places, so calling the rekeying-related address field a spending
key would be unnecessarily confusing.

### DIFF
--- a/data/basics/userBalance.go
+++ b/data/basics/userBalance.go
@@ -137,11 +137,11 @@ type AccountData struct {
 	// is expected to have copy-by-value semantics.
 	Assets map[AssetIndex]AssetHolding `codec:"asset,allocbound=-"`
 
-	// SpendingKey is the address against which signatures/multisigs/logicsigs should be checked.
+	// EffectiveAddr is the address against which signatures/multisigs/logicsigs should be checked.
 	// If empty, the address of the account whose AccountData this is is used.
-	// A transaction may change an account's SpendingKey to "re-key" the account.
+	// A transaction may change an account's EffectiveAddr to "re-key" the account.
 	// This allows key rotation, changing the members in a multisig, etc.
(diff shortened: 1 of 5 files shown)


data/basics/userBalance.go
```diff
@@ -137,11 +137,11 @@ type AccountData struct {
...
 	Assets map[AssetIndex]AssetHolding `codec:"asset,allocbound=-"`
 
-	// SpendingKey is the address against which signatures/multisigs/logicsigs should be checked.
+	// EffectiveAddr is the address against which signatures/multisigs/logicsigs should be checked.
 	// If empty, the address of the account whose AccountData this is is used.
-	// A transaction may change an account's SpendingKey to "re-key" the account.
+	// A transaction may change an account's EffectiveAddr to "re-key" the account.
 	// This allows key rotation, changing the members in a multisig, etc.
-	SpendingKey Address `codec:"spend"`
+	EffectiveAddr Address `codec:"spend"`
```
algomaxj, line 144: do you want to rename the codec tag here too?

tsachiherman: changing this here would break devnet, wouldn't it ?
( i.e. if someone ever sent such a transaction to devnet. it might not be the case )

rotemh: If there is no objection, I'd suggest calling it `AuthAddr` to keep consistent across.

brianolson: I like AuthAddr or more verbosely "AuthorizedAddress"
I'm a little worried "Effective Address" might confuse someone into trying to send to that addr, which I guess wouldn't be the worst thing since someone _does_ have the key they'd need to receive that transfer.

PR #1094, opened 2020-05-21
Remove old buggy version of proposal LowestOutput

algoradam:
Now that the protocol upgrade fixing the bug has gone through, we no longer need to keep the buggy version around for compatibility.


tsachiherman:
Looks good; I'd suggest we'll run this change via a complete catchup so that we can ensure it won't cause any unexpected catchup issues.

tsachiherman:
@derbear could you please review this and see if we could close this one ?

derbear:
The plan is to merge but only after the pending June mainnet release is set.

PR #1106, opened 2020-05-26
Add support for Raw msgp interfaces on unit tests

tsachiherman:
## Summary

Recently we added support for Raw structures in the msgp library. This PR makes the corresponding changes to the unit testing, ensuring that structures which contains Raw interfaces would not get compared to the go-codec. ( which has it's own Raw struct implementation )

### DIFF
--- a/protocol/codec_tester.go
+++ b/protocol/codec_tester.go
@@ -22,8 +22,9 @@ import (
 	"math/rand"
 	"reflect"
 	"strings"
-	"sync"
 	"testing"
+
+	"github.com/algorand/go-deadlock"
 
 	"github.com/algorand/msgp/msgp"
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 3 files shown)


protocol/codec_tester.go
```diff
@@ -205,6 +261,11 @@ func EncodingTest(template msgpMarshalUnmarshal) error {
 func RunEncodingTest(t *testing.T, template msgpMarshalUnmarshal) {
 	for i := 0; i < 1000; i++ {
 		err := EncodingTest(template)
+		if err == errSkipRawMsgpTesting {
```
zeldovich, line 264: Although it's probably too much effort to get consistency between msgp and go-codec for raw fields, it would be nice to make sure that `msgp` is at least self-consistent with respect to raw fields (i.e., forall x, decode(encode(x))=x).

tsachiherman: nice idea. I will add it in a subsequent PR.

tsachiherman:
> The change looks reasonable -- up to you whether you feel like implementing a test for raw field encoding/decoding now or later when you actually get a struct where it matters.

will do.

PR #1108, opened 2020-05-26
Added mule `deploy` task

btoll:
## Summary

Add mule `deploy` task.

## Test Plan

Run mule tasks in `package-deploy.yaml`.

### DIFF
--- a/docker/build/docker.ubuntu.Dockerfile
+++ b/docker/build/docker.ubuntu.Dockerfile
@@ -1,10 +1,21 @@
 ARG ARCH="amd64"
 
 FROM ${ARCH}/ubuntu:18.04
-RUN apt-get update && apt-get install curl python build-essential apt-transport-https ca-certificates software-properties-common -y && \
+RUN apt-get update && apt-get install curl python python3.7 python3-pip build-essential apt-transport-https ca-certificates software-properties-common -y && \
     curl -fsSL https://download.docker.com/linux/ubuntu/gpg | apt-key add - && \
     add-apt-repository "deb [arch=amd64] https://download.docker.com/linux/ubuntu $(lsb_release -cs) stable" && \
     apt-get update && apt-get install docker-ce -y
+
+# Mule needs >= python3.7 so set that as the default.
+RUN update-alternatives --install /usr/bin/python python /usr/bin/python2.7 1 && \
+    update-alternatives --install /usr/bin/python python /usr/bin/python3.7 2 && \
+    update-alternatives --set python /usr/bin/python3.7 && \
+    pip3 install mulecli
+
+RUN apt-get update && apt-get install -y autoconf bsdmainutils git libboost-all-dev && \
+    curl https://dl.google.com/go/go1.12.linux-amd64.tar.gz | tar -xzf - && \
+    mv go /usr/local
+
 WORKDIR /root
 CMD ["/bin/bash"]
 
--- a/docker/build/mule.go.centos.Dockerfile
+++ b/docker/build/mule.go.centos.Dockerfile
@@ -1,0 +1,32 @@
+ARG ARCH="amd64"
+
+FROM ${ARCH}/centos:7
+ENV GOLANG_VERSION 1.12
+ARG ARCH="amd64"
+RUN yum install -y https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm && \
+    yum update -y && \
+    yum install -y autoconf wget awscli git gnupg2 nfs-utils python36 python3.7 pip3-python sqlite3 boost-devel expect jq libtool gcc-c++ libstdc++-devel libstdc++-static rpmdevtools createrepo rpm-sign bzip2 which ShellCheck
+WORKDIR /root
+RUN wget https://dl.google.com/go/go${GOLANG_VERSION}.linux-${ARCH%v*}.tar.gz \
+    && tar -xvf go${GOLANG_VERSION}.linux-${ARCH%v*}.tar.gz && \
+    mv go /usr/local
+RUN pip3 install mulecli
+ENV GOROOT=/usr/local/go \
+    GOPATH=$HOME/go
+RUN mkdir -p $GOPATH/src/github.com/algorand
+COPY . $GOPATH/src/github.com/algorand/go-algorand
+ENV PATH=$GOPATH/bin:$GOROOT/bin:$PATH \
+    BRANCH=${BRANCH} \
+    CHANNEL=${CHANNEL} \
+    BUILDCHANNEL=${BUILDCHANNEL} \
+    DEFAULTNETWORK=${DEFAULTNETWORK} \
+    FULLVERSION=${FULLVERSION} \
+    GOPROXY=https://gocenter.io \
+    PKG_ROOT=${PKG_ROOT}
+WORKDIR $GOPATH/src/github.com/algorand/go-algorand
+RUN make ci-deps && make clean
+RUN rm -rf $GOPATH/src/github.com/algorand/go-algorand && \
+    mkdir -p $GOPATH/src/github.com/algorand/go-algorand
+RUN echo "vm.max_map_count = 262144" >> /etc/sysctl.conf
+CMD ["/bin/bash"]
+
--- a/package-deploy.yaml
+++ b/package-deploy.yaml
@@ -62,28 +62,25 @@ tasks:
       ]
     target: mule-deploy-rpm WORKDIR=/projects/go-algorand
 
-  - task: s3.DownloadFile
+  - task: s3.BucketCopy
     name: deb
-    bucketName: algorand-staging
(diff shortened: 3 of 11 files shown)


docker/build/mule.go.centos.Dockerfile
```diff
@@ -0,0 +1,32 @@
+ARG ARCH="amd64"
```
bricerisingalgorand, line 1: Why couldn't you use the centos dockerfile we're using for the build process?

docker/build/docker.ubuntu.Dockerfile
```diff
@@ -1,10 +1,21 @@
 ARG ARCH="amd64"
 
 FROM ${ARCH}/ubuntu:18.04
-RUN apt-get update && apt-get install curl python build-essential apt-transport-https ca-certificates software-properties-common -y && \
+RUN apt-get update && apt-get install curl python python3.7 python3-pip build-essential apt-transport-https ca-certificates software-properties-common -y && \
     curl -fsSL https://download.docker.com/linux/ubuntu/gpg | apt-key add - && \
     add-apt-repository "deb [arch=amd64] https://download.docker.com/linux/ubuntu $(lsb_release -cs) stable" && \
     apt-get update && apt-get install docker-ce -y
+
+# Mule needs >= python3.7 so set that as the default.
```
bricerisingalgorand, line 9: We shouldn't need to install mule in any of the dockerfiles we're maintaining in this repo

package-deploy.yaml
```diff
@@ -0,0 +1,129 @@
...
+    docker:
+      image: algorand/mule-linux-ubuntu
+      version: '{{ docker.Version.outputs.version }}'
+      workDir: /projects/go-algorand
+      env: [
+        AWS_ACCESS_KEY_ID=$AWS_ACCESS_KEY_ID,
+        AWS_SECRET_ACCESS_KEY=$AWS_SECRET_ACCESS_KEY
+      ]
+      volumes: [
+        $HOME/packages:/root/packages,
```
bricerisingalgorand, line 42: What is this packages folder for? We should only be writing the packages to the tmp folder inside of the project workspace so that we're not filling other directories on the host with files for this projects build process. The CI process will scale to more projects better that way

package-deploy.yaml
```diff
@@ -0,0 +1,129 @@
...
+        $XDG_RUNTIME_DIR/gnupg/S.gpg-agent:/root/.gnupg/S.gpg-agent,
+        $HOME/.gnupg/pubring.kbx:/root/.gnupg/pubring.kbx
+      ]
+    target: mule-deploy-rpm WORKDIR=/projects/go-algorand
+
+  - task: s3.BucketCopy
```
bricerisingalgorand, line 65: are the src and dest fields backwards here? It looks like we're intending to download files from s3 and put them on our host. Also we should be using relative paths on the host since we'll be running mule from the root directory of our project. This way it will work on any system we need to run it on

scripts/release/mule/deploy/rpm/deploy.sh
```diff
@@ -0,0 +1,55 @@
...
+do
+    python2 rpmsign.py "$rpm"
+    cp -p "$rpm" rpmrepo
+done
+
+createrepo --database rpmrepo
+rm -f rpmrepo/repodata/repomd.xml.asc
+gpg -u rpm@algorand.com --detach-sign --armor rpmrepo/repodata/repomd.xml
+
+mule -f package-deploy.yaml package-deploy-rpm-repo
```
bricerisingalgorand, line 54: It would be a better practice for us to have this mule task execute this script. Ideally we can use mule to execute any automated process we need to, instead of having shell scripts executing mule jobs. That way it's more consistent and easier for people onboarding to the project

package-deploy.yaml
```diff
@@ -0,0 +1,129 @@
...
+      image: algorand/mule-linux-ubuntu
+      version: '{{ docker.Version.outputs.version }}'
+      workDir: /projects/go-algorand
+      env: [
+        AWS_ACCESS_KEY_ID=$AWS_ACCESS_KEY_ID,
+        AWS_SECRET_ACCESS_KEY=$AWS_SECRET_ACCESS_KEY
+      ]
+      volumes: [
+        $HOME/packages:/root/packages,
+        $XDG_RUNTIME_DIR/gnupg/S.gpg-agent:/root/.gnupg/S.gpg-agent,
```
bricerisingalgorand, line 43: Is this expecting any host this task runs on to have some certain gpg configurations in this directory?

package-deploy.yaml
```diff
@@ -0,0 +1,129 @@
...
+    target: mule-deploy-generate_releases_page WORKDIR=/projects/go-algorand
+
+  - task: docker.Make
+    name: deb
+    docker:
+      image: algorand/mule-linux-ubuntu
+      version: '{{ docker.Version.outputs.version }}'
+      workDir: /projects/go-algorand
+      env: [
+        AWS_ACCESS_KEY_ID=$AWS_ACCESS_KEY_ID,
```
bricerisingalgorand, line 38: Just to confirm, will this task work on hosts where these env vars aren't available? In aws we're using host metadata to authenticate to aws, so these env vars won't exist

docker/build/docker.ubuntu.Dockerfile
btoll: Could you explain this more?  How would you suggest that we get mule?

docker/build/mule.go.centos.Dockerfile
btoll: I need mule.  See the comment above.  Please suggest an alternative.

package-deploy.yaml
btoll: Currently, the expectation is that the previous packages will be located in that folder.  The newest package is then copied to that directory, and then the snapshot is made.

This needs to be improved.  We'll need to determine from where we'll get all of those packages, and then this will be updated to that location.

package-deploy.yaml
btoll: Yes, the signing will need to be done before the task is run.

package-deploy.yaml
btoll: No, this package should already be on staging from running the previous build and package tasks.  We need to get it from staging and then put it in a location where we can add it to the previous packages and create a snapshot to then upload to `algorand-releases`.

package-deploy.yaml
btoll: It wouldn't work unless aws is able to determine from another source other than env vars that a machine has been authenticated.  I'm unfamiliar with the method you're describing.

scripts/release/mule/deploy/rpm/deploy.sh
btoll: I'm calling mule tasks from several locations in these shell scripts.  I'm not sure how that can be avoided, but I'm open to a better approach.

package-deploy.yaml
```diff
@@ -0,0 +1,129 @@
...
+    dest: /projects/go-algorand/tmp/node_pkgs/${OS_TYPE}/${ARCH_TYPE}
+
+  - task: s3.BucketCopy
+    name: gnupg
+    src: s3://algorand-devops-misc/tools/gnupg2.2.9_centos7_amd64.tar.bz2
+    dest: /root
+
+  - task: s3.BucketCopy
+    name: releases-page
+    src: /tmp/index.html
```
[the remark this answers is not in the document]
btoll: This uses those class helper methods under the cover.  I've found the BucketCopy to be much easier to use.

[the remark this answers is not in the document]
btoll: Maybe BucketCopy isn't the best name?

scripts/release/mule/deploy/generate_releases_page/generate_releases_page
```diff
@@ -0,0 +1,182 @@
+#!/usr/bin/env python3
```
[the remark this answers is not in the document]
btoll: No, I'll re-add it.

docker/build/mule.go.centos.Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+    GOPATH=$HOME/go
+RUN mkdir -p $GOPATH/src/github.com/algorand
+COPY . $GOPATH/src/github.com/algorand/go-algorand
+ENV PATH=$GOPATH/bin:$GOROOT/bin:$PATH \
+    BRANCH=${BRANCH} \
+    CHANNEL=${CHANNEL} \
+    BUILDCHANNEL=${BUILDCHANNEL} \
+    DEFAULTNETWORK=${DEFAULTNETWORK} \
+    FULLVERSION=${FULLVERSION} \
+    GOPROXY=https://gocenter.io \
```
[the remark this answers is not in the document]
btoll: This file will disappear soon.  Brice is making a ticket to add a feature to `mule` to support conditions, and when that happens I will delete this dockerfile.

PR #1114, opened 2020-05-29
Restructure config package to use reflection

tsachiherman:
## Summary
Existing config files are using hard-coded data structure to represent versioned configuration setting. The per-version structures need to be duplicated with every change to the configuration structure, which makes it error prone, especially when dealing with non-default field values.
( since these non-default field values require some explicit written code to handle the config version upgrades )

Replacing that with a configuration struct which is self-tagged with the default values, allow us to auto-generate the above logic and avoid programmatic mistakes. 

## Test Plan

Unit tests were updated

## Release Plan

This change is targeted for the July release


config/config.go
```diff
@@ -46,42 +46,55 @@ const Mainnet protocol.NetworkID = "mainnet"
...
+// denoting a new version. When doing so, also update the
+// installer/config.json.example and test/testdata/configs/config-v{n}.json
+//
+// !!! WARNING !!!
 type Local struct {
 	// Version tracks the current version of the defaults so we can migrate old -> new
 	// This is specifically important whenever we decide to change the default value
 	// for an existing parameter.
-	Version uint32
+	Version uint32 `version[0]:"0" version[1]:"1" version[2]:"2" version[3]:"3" version[4]:"4" version[5]:"5" version[6]:"6" version[7]:"7"`
```
derbear, line 63: Is this strictly necessary, or would it be possible to have use `getLatestConfigVersion` to give the latest version number and fill in the missing versions on initialization?

config/config.go
```diff
@@ -46,42 +46,55 @@ const Mainnet protocol.NetworkID = "mainnet"
...
+// !!! WARNING !!!
+//
+// These versioned struct tags need to be maintained CAREFULLY and treated
+// like UNIVERSAL CONSTANTS - they should not be modified once committed.
+//
+// New fields may be added to the Local struct, along with a version tag
+// denoting a new version. When doing so, also update the
+// installer/config.json.example and test/testdata/configs/config-v{n}.json
+//
+// !!! WARNING !!!
```
derbear, line 58: It might be good to add a comment here also saying that newer versions of the config will pick up the newest user-defined default value.

config/config.go
derbear: What was the rationale behind putting defaults in tags and not in a map which is used to fill out the configs at initialization? One unfortunate fact about this design is that it's harder to see what fields changed in a particular version. Was the intent here to make it easier to group up changes happening to a single field?

config/config.go
```diff
@@ -46,42 +46,55 @@ const Mainnet protocol.NetworkID = "mainnet"
...
-	ReconnectTime time.Duration
+	GossipFanout int    `version[0]:"4"`
+	NetAddress   string `version[0]:""`
+
+	// 1 * time.Minute = 60000000000 ns
+	ReconnectTime time.Duration `version[0]:"60" version[1]:"60000000000"`
+
 	// what we should tell people to connect to
-	PublicAddress string
+	PublicAddress string `version[0]:""`
```
derbear, line 79: Does it make sense to omit the tag if the Go zero value will be filled in?

config/config.go
```diff
@@ -46,42 +46,55 @@ const Mainnet protocol.NetworkID = "mainnet"
...
 
 // Local holds the per-node-instance configuration settings for the protocol.
+// !!! WARNING !!!
+//
+// These versioned struct tags need to be maintained CAREFULLY and treated
+// like UNIVERSAL CONSTANTS - they should not be modified once committed.
+//
+// New fields may be added to the Local struct, along with a version tag
+// denoting a new version. When doing so, also update the
+// installer/config.json.example and test/testdata/configs/config-v{n}.json
```
derbear, line 56: I wonder if these can be autogenerated based on the values defined here...

config/local_defaults.go
```diff
@@ -18,487 +18,178 @@ package config
...
-			newCfg.MaxConnectionsPerIP = defaultLocalV3.MaxConnectionsPerIP
-		}
-		if newCfg.CatchupParallelBlocks == defaultLocalV2.CatchupParallelBlocks {
-			newCfg.CatchupParallelBlocks = defaultLocalV3.CatchupParallelBlocks
-		}
-		newCfg.Version = 3
+	return
+}
+
+func getLatestConfigVersion() uint32 {
```
derbear, line 109: Does it also make sense to instead have an expected "latest config version" global variable, and have a unit test catch whether or not this is inconsistent with the tags?

config/config.go
tsachiherman: Yes, as you've figured, I have tried to place it all in "one" place.

config/config.go
tsachiherman: My plan was to ensure all field in the `Local` struct are versioned, and verify that in a unit test. That way, no one would add a new field and "forget" to associate it with a version.

config/config.go
tsachiherman: yes, I plan to do so. I will update this PR with these changes.

config/local_defaults.go
tsachiherman: I'm trying to ensure that when we make changes to the `Local`, it need to be reflected in less places, which would make it more simple to use and maintain. Adding a global variable that need to be updated with any `Local` change would defeat that, wouldn't it ?

tsachiherman:
I'm closing this PR for further work. I'll open this one again once the changes are done.

config/config.go
derbear: Ah, I see. That's a good idea.

config/local_defaults.go
derbear: Yeah, that's true. I did wonder whether this would be easier than adding another tag to `Local.Version`, i.e. `version[i]:"i"`, which is somewhat redundant as well; it seems like over time that the list of tags would get really long, and it doesn't seem to provide any extra information.

PR #1115, opened 2020-06-01
Improve network connectionPerformanceMonitor performance

tsachiherman:
## Summary

The `connectionPerformanceMonitor` is suspected to slow down critical communication paths.
I was able to optimize it further and it looks now quite decent. ( i.e. 170Kns->6Kns )
The reminder of the 5Kns is mainly attributed to the deadlock locking logic, which I'm inclined to change in the PR unless proven it's really needed.

## Change
The culprit was the `pruneOldMessages` function. For each message that was processed by the `connectionPerformanceMonitor` on stage `pmStageAccumulate` and `pmStageStopping`, it used to prune the old messages. The `pruneOldMessages` was comparing the current timestamp to the timestamps of all the pending messages, and prunes the ones that are too old. In terms of worst-case-scenario efficiency, where n is the number of messages, the projected efficiency is O(n^2). This change split the messages into time-based buckets, making it into O(n).

## Test Plan

Changes were tested using S1/S2 networks. No performance change were noted.
Benchmark was added, performance gains are as follows:

|Stage | Before | After |
|------|--------|------|
|pmStagePresync |5,259ns |5,327ns|
|pmStageSync | 4,957ns|4,967ns|
|pmStageAccumulate |178,698ns |6,160ns|
|pmStageStopping |142,656ns |5,944ns|

( this test check the average message delay imposed by the `connectionPerformanceMonitor` after feeding the monitor with 700K messages )


network/connPerfMon.go
```diff
@@ -66,21 +67,28 @@ type pmStatistics struct {
 	messageCount   int64              // the number of messages used to calculate the above statistics
 }
 
+// pmPendingMessageBucket is used to buffer messages in time ranges blocks.
+type pmPendingMessageBucket struct {
+	messages  map[crypto.Digest]*pmMessage // the pendingMessages map contains messages that haven't been received from all the peers within the pmMaxMessageWaitTime, and belong to the timerange of this bucket.
+	startTime int64                        // the start-range of the timestamp which bounds the messages ranges which would go into this bucket.
```
algorandskiy, line 73: nit: comment the duration unit?

network/connPerfMon.go
```diff
@@ -298,31 +305,58 @@ func (pm *connectionPerformanceMonitor) updateMessageIdlingInterval(now int64) (
...
+	}
 	if pendingMsg == nil {
 		if newMessages {
+			if msgBucket == nil {
+				// no bucket was found. create one.
+				msgBucket = &pmPendingMessageBucket{
+					messages:  make(map[crypto.Digest]*pmMessage),
+					startTime: msg.Received - (msg.Received % int64(pmMessageBucketDuration)), // align with pmMessageBucketDuration
+				}
+				msgBucket.endTime = msgBucket.startTime + int64(pmMessageBucketDuration) - 1
```
algorandskiy, line 355: nit: seems like endTime can be
```
msgBucket.endTime = msgBucket.startTime + int64(pmMessageBucketDuration
```
if use semi-open interval above:
```
if msg.Received >= currentMsgBucket.startTime && msg.Received < currentMsgBucket.endTime {
```

network/connPerfMon.go
```diff
@@ -298,31 +305,58 @@ func (pm *connectionPerformanceMonitor) updateMessageIdlingInterval(now int64) (
 
 func (pm *connectionPerformanceMonitor) pruneOldMessages(now int64) {
 	oldestMessage := now - int64(pmMaxMessageWaitTime)
-	for digest, pendingMsg := range pm.pendingMessages {
-		if oldestMessage < pendingMsg.firstPeerTime {
+	prunedBucketsCount := 0
+	for bucketIdx, currentMsgBucket := range pm.pendingMessagesBuckets {
+		if currentMsgBucket.endTime > oldestMessage {
+			pm.pendingMessagesBuckets[bucketIdx-prunedBucketsCount] = currentMsgBucket
```
algorandskiy, line 311: what is the point? if `prunedBucketsCount == 0` and `bucketIdx == 0` then assigning self. plus, `prunedBucketsCount` and `bucketIdx` and synchronized and incremented on each iteration, so the index is always zero?

network/connPerfMon.go
tsachiherman: maybe I'm missing something, but I don't see any particular "value" in that. i.e. the compiler would already perform the "-1" as part of the computation, and doing a comparison operation using < vs <= has the same performance characteristics. I can see that there might be some subtle value in dropping the endTime completely, as it can always be calculated off the startTime, but again - I don't think we will ever see any concrete benefits doing so.

network/connPerfMon.go
tsachiherman: no, the `prunedBucketsCount` is not always identical to the `bucketIdx`. That's why I added a `continue`.
( I created a unit test `TestBucketsPruning` to verify that )

network/connPerfMon.go
tsachiherman: done.

network/connPerfMon.go
algorandskiy: ah, right, my bad

PR #1120, opened 2020-06-01
Fix datarace when writing a catchpoint label

tsachiherman:
## Summary

This PR address a potential data race detected by our automation testing. The data race occurs when a catchpoint label is being written to disk and a new block is added to the ledger at the same time.

## Test Plan

Repurpose one of the accountUpdates unit test to verify the fix works as expected.

### DIFF
--- a/ledger/acctupdates.go
+++ b/ledger/acctupdates.go
@@ -900,28 +900,18 @@ func (au *accountUpdates) accountsUpdateBalances(accountsDeltas map[basics.Addre
 	return
 }
 
-func (au *accountUpdates) accountsCreateCatchpointLabel(committedRound, balancesRound basics.Round, ledgerBlockDigest crypto.Digest) (label string, err error) {
-	var totals AccountTotals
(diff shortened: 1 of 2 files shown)


ledger/acctupdates.go
```diff
@@ -1055,8 +1050,7 @@ func (au *accountUpdates) commitRound(offset uint64, dbRound basics.Round, lookb
 		return
 	}
 	if isCatchpointRound {
-		committedRoundDigest = au.roundDigest[offset+uint64(lookback)-1]
-		catchpointLabel, err = au.accountsCreateCatchpointLabel(dbRound+basics.Round(offset)+lookback, dbRound+basics.Round(offset), committedRoundDigest)
+		catchpointLabel, err = au.accountsCreateCatchpointLabel(dbRound+basics.Round(offset)+lookback, roundTotals[offset], committedRoundDigest)
```
tsachiherman, line 1053: At this point we don't have the lock taken, so we shouldn't access any of the struct's variables.

PR #1121, opened 2020-06-02
Fix two more dataraces during catchup mode switch

tsachiherman:
## Summary

The initializeFromDisk function is called a second ( or third..) time when we attempt to complete the fast catchup process and switch to the new accounts data. When that happen, we need to shut down the entire backend and reload it.

This PR addresses two such data races - 
The first in the account initialization, where the previously allocated communication channel with the commit goroutine need to be updated.
The second is on the node itself, where we need to preallocate the node context object prior to creating the node monitoring goroutines 

## Test Plan

These two issues were found using the fast catchup e2e test. I have used this test in order to verify the fix as well.

### DIFF
--- a/ledger/acctupdates.go
+++ b/ledger/acctupdates.go
@@ -764,7 +764,7 @@ func (au *accountUpdates) initializeFromDisk(l ledgerForTracker) (lastBalancesRo
 	close(au.catchpointWriting)
 	au.ctx, au.ctxCancel = context.WithCancel(context.Background())
 	au.committedOffset = make(chan deferedCommit, 1)
-	go au.commitSyncer()
+	go au.commitSyncer(au.committedOffset)
 
 	lastBalancesRound = au.dbRound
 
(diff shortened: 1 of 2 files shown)


ledger/acctupdates.go
```diff
@@ -930,10 +930,10 @@ func (au *accountUpdates) roundOffset(rnd basics.Round) (offset uint64, err erro
 	return off, nil
 }
 
-func (au *accountUpdates) commitSyncer() {
+func (au *accountUpdates) commitSyncer(deferedCommits chan deferedCommit) {
```
tsachiherman, line 933: passing the `deferedCommits` over the stack instead of the struct allows us to replace the one on the struct.

node/node.go
```diff
@@ -910,14 +910,14 @@ func (node *AlgorandFullNode) SetCatchpointCatchupMode(catchpointCatchupMode boo
 		node.log.Infof("Indexer is not available - %v", err)
 	}
 
-	node.cancelCtx()
+	// Set up a context we can use to cancel goroutines on Stop()
```
onetechnical, line 913: Wouldn't you still want to keep a `node.cancelCtx()` before re-initializing here since it's not explicitly called?

tsachiherman: This should not be required : the existing context at that time is only used by the catchpoint service itself. The catchpoint service is going to use the return value from this function to update it's context, so it won't have any affect.

PR #1122, opened 2020-06-02
Unit tests for v2 handler functions

EvanJRichard:
## Summary
This PR proposes to add unit tests for the various handlers in `algod/api/server/v2/handlers.go`. All use a mock node and the `echo` context to check that `200` is returned in response to valid input. Some additionally check error cases, such as invalid format strings. The test setup supports additional scrutiny into the recorded response using `httptest.ResponseRecorder`, this is currently only used to check the status code.

## Test Plan

`make test` and `make integration` in progress; `go test` on proposed new tests pass. Of note is that the test for `WaitForRound` lasts one minute. The catchpoint tests are not yet implemented, but compile teal is.

### See also
https://github.com/algorand/go-algorand/issues/1111


daemon/algod/api/server/v2/test/helpers.go
```diff
@@ -0,0 +1,285 @@
...
+	"github.com/algorand/go-algorand/data/transactions"
+	"github.com/algorand/go-algorand/ledger"
+	"github.com/algorand/go-algorand/logging"
+	"github.com/algorand/go-algorand/node"
+	"github.com/algorand/go-algorand/node/indexer"
+	"github.com/algorand/go-algorand/protocol"
+	"github.com/algorand/go-algorand/util/db"
+)
+
+type mockNode struct {
```
tsachiherman, line 42: would it be hard to move the `mockNode' into /components/mocks ?

node/node.go
```diff
@@ -83,6 +83,34 @@ func (status StatusReport) TimeSinceLastRound() time.Duration {
 	return time.Since(status.LastRoundTimestamp)
 }
 
+// Interface represents node fns.
+type Interface interface {
```
tsachiherman, line 87: this interface should include all the external calls into the node, not internal implementation; hence, whenever we're implementing an interface within the AlgorandFullNode, we don't need to export it via the `Interface` interface.
two obvious examples are `Uint64()` and `OnNewBlock()`

node/node.go
```diff
@@ -83,6 +83,34 @@ func (status StatusReport) TimeSinceLastRound() time.Duration {
...
+	GetTransaction(addr basics.Address, txID transactions.Txid, minRound basics.Round, maxRound basics.Round) (TxnWithStatus, bool)
+	PoolStats() PoolStats
+	IsArchival() bool
+	OnNewBlock(block bookkeeping.Block, delta ledger.StateDelta)
+	Uint64() uint64
+	Indexer() (*indexer.Indexer, error)
+	GetTransactionByID(txid transactions.Txid, rnd basics.Round) (TxnWithStatus, error)
+	AssembleBlock(round basics.Round, deadline time.Time) (agreement.ValidatedBlock, error)
+	StartCatchup(catchpoint string) error
+	AbortCatchup(catchpoint string) error
```
tsachiherman, line 111: The StartCatchup and AbortCatchup are here explicitly to be used by the handlers. Could you move that upward ?

node/node.go
algorandskiy: I think this must live not in `node.go` but in handers where it is used.

EvanJRichard:
Looks like moving `mockNode` to `mocks` causes an import cycle, investigating how to get around that 🤔

daemon/algod/api/server/v2/test/helpers.go
EvanJRichard: Harder than I thought. `mockNode` needs package `data` and package `node`, but both of those ultimately import `mocks`, so it creates a cycle when `mockNode` is moved there.

tsachiherman: ok, thanks for trying.

daemon/algod/api/server/v2/test/helpers.go
```diff
@@ -0,0 +1,319 @@
...
+	if err != nil {
+		panic(err)
+	}
+
+	tx := make([]transactions.SignedTxn, TXs)
+	latest := ledger.Latest()
+	if latest != 0 {
+		panic(fmt.Errorf("newly created ledger doesn't start on round 0"))
+	}
+	bal := bootstrap.Balances()
```
tsachiherman, line 260: isn't `bal` == `genesis` ? I don't think that the "Balances()" call is needed.

PR #1129, opened 2020-06-03
Eric/fix pingpong asset

egieseke:
## Summary

Updated ping-pong utility to improve support for asset testing.  Modified setup to use new accounts and each has its own set of assets, which are opted in and distributed to the other participant accounts.  

The GitHub issue is here:
https://github.com/algorand/go-algorand-internal/issues/247

## Test Plan

Tested locally using a local node, and also tested using the Algonet framework.  Small and large test cases are now close to 900 TPS using scenario 1 test with Algonet.


shared/pingpong/accounts.go
```diff
@@ -82,186 +85,232 @@ func ensureAccounts(ac libgoal.Client, initCfg PpConfig) (accounts map[string]ui
...
-		accounts = takeTopAccounts(accounts, cfg.NumPartAccounts, cfg.SrcAccount)
-	} else {
-		// Not enough accounts yet (or just enough).  Create more if needed
-		if len(accounts) != int(cfg.NumPartAccounts+1) {
-			fmt.Printf("Not enough accounts - creating %d more\n", int(cfg.NumPartAccounts+1)-len(accounts))
-		}
-		accounts, err = generateAccounts(ac, accounts, cfg.NumPartAccounts, wallet)
-		if err != nil {
-			return
+	if cfg.NumAsset == 0 {
```
stechu, line 88: Could you explain why this check is needed?

egieseke: Added comments explain this.

stechu: Thanks for the clarification!

shared/pingpong/accounts.go
```diff
@@ -82,186 +85,233 @@ func ensureAccounts(ac libgoal.Client, initCfg PpConfig) (accounts map[string]ui
...
-		fmt.Printf("Cannot lookup source account")
-		err = accountErr
-		return
+// throttle transaction rate
+func throttleTransactionRate(startTime time.Time, cfg PpConfig, totalSent uint64) {
+	localTimeDelta := time.Now().Sub(startTime)
+	currentTps := float64(totalSent) / localTimeDelta.Seconds()
+	if currentTps > float64(cfg.TxnPerSec) {
+		sleepSec := float64(totalSent)/float64(cfg.TxnPerSec) - localTimeDelta.Seconds()
+		sleepTime := time.Duration(int64(math.Round(sleepSec*1000))) * time.Millisecond
```
tsachiherman, line 117: `sleepTime := time.Duration(int64(math.Round(sleepSec*1000))) * time.Millisecond` ->
`sleepTime := time.Duration(int64(math.Round(sleepSec))) * time.Second`

shared/pingpong/accounts.go
```diff
@@ -82,186 +85,233 @@ func ensureAccounts(ac libgoal.Client, initCfg PpConfig) (accounts map[string]ui
...
+					tx, sendErr := constructTxn(addr, addr2, cfg.MaxFee, assetAmt, k, client, cfg)
+					if sendErr != nil {
+						fmt.Printf("Cannot transfer asset %v from account %v\n", k, addr)
+						err = sendErr
+						return
+					}
+
+					_, err = signAndBroadcastTransaction(accounts, addr, tx, client, cfg)
+					if err != nil {
+						_, _ = fmt.Fprintf(os.Stderr, "signing and broadcasting asset distribution failed with error %v\n", err)
```
derbear, line 282: What's the `_, _ = ` for?

egieseke: To satisfy the linter in GoLand.  Thanks for your review!

brianolson: Alternate pattern, since this is in a function that returns error, augent the error that happened with notes about where and why it happened:
`err = fmt.Errorf("message about what's up in this function, %v", err)`
then return `err`

brianolson: Stderr works in this case because it's inside pingpong which is a pretty user oriented command line util, but I use that error pattern a lot and it works whether the thing at the top is command line or httpd or unit test.

PR #1138, opened 2020-06-04
Start merging Applications types

algomaxj:
## Summary

This PR merges some of the the non-ledger pieces of Applications, crossing into `ledger` only to make minor changes to fix the build.


cmd/kmd/main.go
```diff
@@ -23,9 +23,9 @@ import (
 	"path/filepath"
 	"time"
 
-	"golang.org/x/sys/unix"
 	"github.com/spf13/cobra"
 	"github.com/spf13/cobra/doc"
+	"golang.org/x/sys/unix"
```
derbear, line 28: Did `make fmt` get run on master recently?

config/consensus.go
```diff
@@ -217,6 +217,31 @@ type ConsensusParams struct {
...
+
+	AppFlatParamsMinBalance  uint64
+	AppFlatOptInMinBalance   uint64
+	SchemaMinBalancePerEntry uint64
+	SchemaUintMinBalance     uint64
+	SchemaBytesMinBalance    uint64
+	MaxLocalSchemaEntries    uint64
+	MaxGlobalSchemaEntries   uint64
+
+	MaximumMinimumBalance uint64
```
derbear, line 244: As mentioned by @pzbitskiy, it'd be nice to add documentation to these fields.

data/basics/teal.go
```diff
@@ -0,0 +1,298 @@
...
+// equality because an empty map will encode/decode as nil. So if our generated
+// map is empty but not nil, we want to equal a decoded nil off the wire.
+func (sd StateDelta) Equal(o StateDelta) bool {
+	// Lengths should be the same
+	if len(sd) != len(o) {
+		return false
+	}
+	// All keys and deltas should be the same
+	for k, v := range sd {
+		if o[k] != v {
```
tsachiherman, line 87: this would work, but not for the reason you've mentioned.
If v happen to have the default value and o[k] doesn't exist, the function might still return true.
however, this could not happen in reality, since the DeltaAction is always non-zero.

data/basics/teal.go
```diff
@@ -0,0 +1,298 @@
...
+		}
+		switch delta.Action {
+		case SetBytesAction:
+			if len(delta.Bytes) > proto.MaxAppBytesValueLen {
+				return fmt.Errorf("cannot set value for key 0x%x, too long: length was %d, maximum is %d", key, len(delta.Bytes), proto.MaxAppBytesValueLen)
+			}
+		case SetUintAction:
+		case DeleteAction:
+		default:
+			return fmt.Errorf("unknown delta action")
```
tsachiherman, line 109: it would be nice if you could add to the Errorf the delta.Action value. I wouldn't expect anything but zero, but would rather not having the need to make that guess.

data/basics/teal.go
```diff
@@ -0,0 +1,298 @@
...
+	return nil
+}
+
+// EvalDelta stores StateDeltas for an application's global key/value store, as
+// well as StateDeltas for some number of accounts holding local state for that
+// application
+type EvalDelta struct {
+	_struct struct{} `codec:",omitempty,omitemptyarray"`
+
+	GlobalDelta StateDelta `codec:"gd"`
```
tsachiherman, line 121: StateDelta is a map, could you add an allocbound for that ?

data/basics/teal.go
```diff
@@ -0,0 +1,298 @@
...
+
+	// When decoding EvalDeltas, the integer key represents an offset into
+	// [txn.Sender, txn.Accounts[0], txn.Accounts[1], ...]
+	LocalDeltas map[uint64]StateDelta `codec:"ld,allocbound=config.MaxEvalDeltaAccounts"`
```
tsachiherman, line 125: I think that we need a way to limit the decoder allocation on the inner StateDelta here.
maybe 
//msgp:allocbound StateDelta 12345

data/basics/teal.go
```diff
@@ -0,0 +1,298 @@
...
+// because the msgpack codec will encode/decode an empty map as nil, and we want
+// an empty generated EvalDelta to equal an empty one we decode off the wire.
+func (ed EvalDelta) Equal(o EvalDelta) bool {
+	// GlobalDeltas must be equal
+	if !ed.GlobalDelta.Equal(o.GlobalDelta) {
+		return false
+	}
+
+	// LocalDeltas length should be the same
+	if len(ed.LocalDeltas) != len(o.LocalDeltas) {
```
tsachiherman, line 139: can you move this check first ? it's much faster than map items comparison.

data/basics/teal.go
```diff
@@ -0,0 +1,298 @@
...
+
+	// Cost for uints
+	uintCost := MulSaturate(proto.SchemaUintMinBalance, sm.NumUint)
+
+	// Cost for byte slices
+	bytesCost := MulSaturate(proto.SchemaBytesMinBalance, sm.NumByteSlice)
+
+	// Sum the separate costs
+	var min uint64
+	min = AddSaturate(min, flatCost)
```
tsachiherman, line 196: nit:
min := flatCost.Raw

data/basics/userBalance.go
```diff
@@ -153,6 +185,63 @@ type AccountData struct {
...
+
+	Schema   StateSchema  `codec:"hsch"`
+	KeyValue TealKeyValue `codec:"tkv"`
+}
+
+// AppParams stores the global information associated with an application
+type AppParams struct {
+	_struct struct{} `codec:",omitempty,omitemptyarray"`
+
+	ApprovalProgram   []byte      `codec:"approv"`
```
tsachiherman, line 219: this need to be allocbound. ( yes, I know that we currently don't support byte slices allocbounding )

data/basics/userBalance_test.go
```diff
@@ -147,11 +147,23 @@ func TestEncodedAccountDataSize(t *testing.T) {
 	require.Equal(t, MaxEncodedAccountDataSize, len(encoded))
 }
 
-func TestEncodedAccountAssetsAllocationBound(t *testing.T) {
+func TestEncodedAccountAllocationBounds(t *testing.T) {
```
tsachiherman, line 150: I wasn't thinking that this would be the evolution of my unit test.. but I guess it fits ;-)

data/transactions/application.go
```diff
@@ -0,0 +1,769 @@
...
+}
+
+// ApplicationCallTxnFields captures the transaction fields used for all
+// interactions with applications
+type ApplicationCallTxnFields struct {
+	_struct struct{} `codec:",omitempty,omitemptyarray"`
+
+	ApplicationID   basics.AppIndex   `codec:"apid"`
+	OnCompletion    OnCompletion      `codec:"apan"`
+	ApplicationArgs [][]byte          `codec:"apaa,allocbound=256"`
```
tsachiherman, line 82: shouldn't the 256 limit here be set as a max value of a consensus param value ?

data/transactions/application.go
```diff
@@ -0,0 +1,769 @@
...
+	Accounts        []basics.Address  `codec:"apat,allocbound=256"`
+	ForeignApps     []basics.AppIndex `codec:"apfa,allocbound=256"`
+
+	LocalStateSchema  basics.StateSchema `codec:"apls"`
+	GlobalStateSchema basics.StateSchema `codec:"apgs"`
+	ApprovalProgram   []byte             `codec:"apap"`
```
tsachiherman, line 88: Could you add allocbound to these as well ? even if it's not (yet) implemented ?

data/transactions/application.go
```diff
@@ -0,0 +1,769 @@
...
+		// This should never happen. If app exists then we should have
+		// found the creator successfully.
+		err = fmt.Errorf("app %d not found in account %s", aidx, creator.String())
+		return
+	}
+
+	return
+}
+
+func applyDelta(kv basics.TealKeyValue, stateDelta basics.StateDelta) error {
```
tsachiherman, line 180: The order of `stateDelta` entries applies to the `kv` could result in different results. Is that expected and handled by the caller ?

data/transactions/application.go
```diff
@@ -0,0 +1,769 @@
...
+	}
+
+	record, err := balances.Get(sender, false)
+	if err != nil {
+		return err
+	}
+
+	// Ensure sender actually has LocalState allocated for this app.
+	// Can't clear out if not currently opted in
+	_, ok := record.AppLocalStates[appIdx]
```
tsachiherman, line 492: don't you need to check if AppLocalStates == nil ?

data/transactions/application.go
```diff
@@ -0,0 +1,769 @@
...
+			// Fill in applyData, so that consumers don't have to implement a
+			// stateful TEAL interpreter to apply state changes
+			ad.EvalDelta = stateDeltas
+		} else {
+			// Ignore errors and rejections from the ClearStateProgram
+		}
+	}
+
+	// Fetch the (potentially updated) sender record
+	record, err = balances.Get(sender, false)
```
tsachiherman, line 524: would it make sense to move this into the if statement above to avoid re-fetching the same data twice ?

data/transactions/keyreg_test.go
```diff
@@ -54,6 +62,10 @@ func (balances keyregTestBalances) ConsensusParams() config.ConsensusParams {
 	return config.Consensus[balances.version]
 }
 
+func (balances keyregTestBalances) Round() basics.Round {
+	return basics.Round(8675309)
```
tsachiherman, line 66: in the spirit for testing against large numbers, how's about a number that won't fit into uint32 ?

data/transactions/logic/debugger.go
```diff
@@ -0,0 +1,272 @@
...
+
+func (dbg *WebDebuggerHook) postState(state *DebugState, endpoint string) error {
+	var body bytes.Buffer
+	enc := protocol.NewJSONEncoder(&body)
+	err := enc.Encode(state)
+	if err != nil {
+		return err
+	}
+
+	req, err := http.NewRequest("POST", fmt.Sprintf("%s/%s", dbg.URL, endpoint), &body)
```
tsachiherman, line 237: 1. "MethodPost" -> http.MethodPost
2. Use the url.URL to build the url here instead of formatting it yourself using Sprintf

data/transactions/logic/debugger.go
```diff
@@ -0,0 +1,272 @@
...
+	if err != nil {
+		return err
+	}
+
+	httpClient := &http.Client{}
+	r, err := httpClient.Do(req)
+	if err == nil && r.StatusCode != 200 {
+		err = fmt.Errorf("bad response: %d", r.StatusCode)
+	}
+	return err
```
tsachiherman, line 247: you need to drain the r.Body, or you'll have a resource leak, and eventually run out of file descriptors.

data/transactions/logic/debugger.go
```diff
@@ -0,0 +1,272 @@
...
+
+// Register sends state to remote debugger
+func (dbg *WebDebuggerHook) Register(state *DebugState) error {
+	u, err := url.Parse(dbg.URL)
+	if err != nil {
+		logging.Base().Errorf("Failed to parse url: %s", err.Error())
+	}
+	h := u.Hostname()
+	// check for 127.0.0/8 ?
+	if h != "localhost" && h != "127.0.0.1" && h != "::1" {
```
tsachiherman, line 258: look at https://golang.org/pkg/net/?m=all#isLocalhost

data/transactions/logic/doc.go
```diff
@@ -125,14 +155,27 @@ func OpImmediateNote(opName string) string {
...
 var opDocExtraList = []stringString{
-	{"ed25519verify", "The 32 byte public key is the last element on the stack, preceeded by the 64 byte signature at the second-to-last element on the stack, preceeded by the data which was signed at the third-to-last element on the stack."},
-	{"bnz", "The `bnz` instruction opcode 0x40 is followed by two immediate data bytes which are a high byte first and low byte second which together form a 16 bit offset which the instruction may branch to. For a bnz instruction at `pc`, if the last element of the stack is not zero then branch to instruction at `pc + 3 + N`, else proceed to next instruction at `pc + 3`. Branch targets must be well aligned instructions. (e.g. Branching to the second byte of a 2 byte op will be rejected.) Branch offsets are currently limited to forward branches only, 0-0x7fff. A future expansion might make this a signed 16 bit integer allowing for backward branches and looping."},
+	{"ed25519verify", "The 32 byte public key is the last element on the stack, preceded by the 64 byte signature at the second-to-last element on the stack, preceded by the data which was signed at the third-to-last element on the stack."},
```
tsachiherman, line 158: We need to be more consistent here: either we end a description with a "." or not.

data/transactions/logic/eval.go
```diff
@@ -205,39 +364,50 @@ func Eval(program []byte, params EvalParams) (pass bool, err error) {
...
 		return false, cx.err
 	}
-	if version > params.Proto.LogicSigVersion {
-		cx.err = fmt.Errorf("program version %d greater than protocol supported version %d", version, params.Proto.LogicSigVersion)
+	if version > cx.EvalParams.Proto.LogicSigVersion {
+		cx.err = fmt.Errorf("program version %d greater than protocol supported version %d", version, cx.EvalParams.Proto.LogicSigVersion)
 		return false, cx.err
 	}
-	if (params.Txn.Txn.RekeyTo != basics.Address{}) { // Currently no TEAL version knows about the RekeyTo field, but this may change in a future TEAL version
+	if cx.EvalParams.Txn.Txn.RekeyTo != (basics.Address{}) { // Currently no TEAL version knows about the RekeyTo field, but this may change in a future TEAL version
```
tsachiherman, line 371: I think that using !cx.EvalParams.Txn.Txn.RekeyTo.Zero() would be cleaner

data/transactions/logic/eval.go
```diff
@@ -296,7 +478,12 @@ func Check(program []byte, params EvalParams) (cost int, err error) {
 	cx.EvalParams = params
 	cx.program = program
 	for (cx.err == nil) && (cx.pc < len(cx.program)) {
+		prevpc := cx.pc
 		cost += cx.checkStep()
+		if cx.pc == prevpc {
```
tsachiherman, line 483: since we don't have backward jump, wouldn't it be a stronger predicate to use `cx.pc <= prevpc` ?

data/transactions/logic/eval.go
```diff
@@ -1148,3 +1537,541 @@ func opStore(cx *evalContext) {
...
+		return nil, err
+	}
+
+	kvIdx := ckey{appID, addr}
+	localKV, ok := cx.readOnlyLocalStates[kvIdx]
+	if !ok {
+		var err error
+		localKV, err = cx.Ledger.AppLocalState(addr, basics.AppIndex(appID))
+		if err != nil {
+			return nil, fmt.Errorf("failed to fetch app local state for acct %s, app %d: %v", addr, appID, err)
```
tsachiherman, line 1663: If the `addr` is of type `basics.Address`, could you use `%v` instead of `%s` when formatting it ?

data/transactions/logic/eval.go
```diff
@@ -1148,3 +1537,541 @@ func opStore(cx *evalContext) {
...
+	}
+
+	// We key localStateCows by address and not accountIdx, because
+	// multiple accountIdxs may refer to the same address
+	idxCow, ok := cx.localStateCows[addr]
+	if !ok {
+		// No cached cow for this address. Make one.
+		localKV, err := cx.Ledger.AppLocalState(addr, basics.AppIndex(cx.Txn.Txn.ApplicationID))
+		if err != nil {
+			return nil, fmt.Errorf("failed to fetch app local state for acct %s: %v", addr, err)
```
tsachiherman, line 1684: same comment as before regarding the `%s` -> `%v`

data/basics/teal.go
algomaxj: oh wow, good catch! that's terrible. fixing.

algomaxj: pushed a fix, thanks again!

data/basics/teal.go
```diff
@@ -0,0 +1,312 @@
...
+// Valid checks whether the keys and values in a StateDelta conform to the
+// consensus parameters' maximum lengths
+func (sd StateDelta) Valid(proto *config.ConsensusParams) error {
+	for key, delta := range sd {
+		if len(key) > proto.MaxAppKeyLen || proto.MaxAppKeyLen == 0 {
+			return fmt.Errorf("key too long: length was %d, maximum is %d", len(key), proto.MaxAppKeyLen)
+		}
+		switch delta.Action {
+		case SetBytesAction:
+			if len(delta.Bytes) > proto.MaxAppBytesValueLen {
```
derbear, line 110: Is there a reason why there's no `proto.MaxAppBytesValueLen=0` check here?

data/transactions/application.go
```diff
@@ -0,0 +1,769 @@
...
+	// update the ApprovalProgram and ClearStateProgram for the application
+	UpdateApplicationOC OnCompletion = 4
+
+	// DeleteApplicationOC indicates that an application transaction will
+	// delete the AppParams for the application from the creator's balance
+	// record
+	DeleteApplicationOC OnCompletion = 5
+)
+
+func (oc OnCompletion) String() string {
```
derbear, line 57: Consider using `stringer` here.

data/transactions/application.go
```diff
@@ -0,0 +1,769 @@
...
+	ApprovalProgram   []byte             `codec:"apap"`
+	ClearStateProgram []byte             `codec:"apsu"`
+
+	// If you add any fields here, remember you MUST modify the Empty
+	// method below!
+}
+
+// Empty indicates whether or not all the fields in the
+// ApplicationCallTxnFields are zeroed out
+func (ac *ApplicationCallTxnFields) Empty() bool {
```
derbear, line 97: Is there an easy way to use the `msgp` `.Zero` version of this?

derbear: Also, since this is pass-by-pointer, consider checking for `ac == nil` here.

data/basics/teal.go
algomaxj: The `allocbound` for StateDelta is set on the StateDelta type -- it is set to `config.MaxStateDeltaKeys` above (you can confirm it is used by searching for `MaxStateDeltaKeys` in `./data/basics/msgp_gen.go`)

data/basics/teal.go
algomaxj: (see above comment -- `StateDelta` has an allocbound set on its type to be == `config.MaxStateDeltaKeys`. Check `(z *EvalDelta) UnmarshalMsg` to see it used)

data/transactions/application.go
algomaxj: That would be very bad if it is possible -- but since there can be at most one value delta per key, I didn't think it was. Is there a scenario I am missing?

data/transactions/application.go
algomaxj: should be OK as is: https://play.golang.org/p/Mw6QlAey-HD

data/transactions/keyreg_test.go
algomaxj: sure, done.

data/transactions/logic/eval.go
algomaxj: sure, done :+1:

data/transactions/application.go
algorandskiy: > A nil map behaves like an empty map when reading, but attempts to write to a nil map will cause a runtime panic.

This one was kind of eye-opener on go nils vs c++ :)

data/transactions/logic/eval.go
algomaxj: done :+1:

data/basics/teal.go
algomaxj: done, thanks!

data/transactions/application.go
tsachiherman: Given that this is an assumption the function is making, could you please document it at the function header ?

data/basics/teal.go
algomaxj: done

data/basics/teal.go
algomaxj: Personally I find the way it is more clear, since each subsequent that adds to min is exactly the same. I can change it if you feel strongly

data/basics/teal.go
algorandskiy: I thought about adding it but in fact `proto.MaxAppKeyLen` and `proto.MaxAppBytesValueLen` are in sync and unsupported protocol will be filtered out above.

data/transactions/application.go
algomaxj: done; set constants and added a test like you did for user balance

data/transactions/application.go
algomaxj: Added a comment to the for loop to explain why order doesn't matter (I think it is impossible to construct a `stateDelta` that would apply non-deterministically)

data/transactions/logic/debugger.go
algomaxj: I think Pavel's implementation is better: https://github.com/algorand/go-algorand/pull/1138

data/basics/userBalance.go
algomaxj: I don't want to add an `allocbound` if `msgp` doesn't support it, otherwise I think it is confusing to someone reading the code.

data/transactions/application.go
algomaxj: Same as above, I don't want to add an allocbound if msgp doesn't support it, otherwise I think it is confusing to someone reading the code.

data/transactions/logic/debugger.go
algomaxj: done, thanks

data/transactions/logic/debugger.go
algomaxj: done (call `r.Body.Close()` on nil error)

data/transactions/application.go
algomaxj: nice, thanks. done.

data/transactions/logic/eval.go
algomaxj: done

data/transactions/application.go
algomaxj: I want to use it, but the `func (z *ApplicationCallTxnFields) MsgIsZero() bool` method doesn't check for nilness on the slices -- it just checks for zero length. I don't see a reason to not be strict about that, since we should always be decoding `nil` on the wire if those fields are supposed to be empty.

As for the `nil` check... I think I would rather panic in that situation than decide if a `nil` pointer is empty.

data/transactions/logic/eval.go
algomaxj: pushed pavel's change from https://github.com/justicz/go-algorand/pull/110

data/transactions/logic/doc.go
algomaxj: pushed pavel's change from https://github.com/justicz/go-algorand/pull/110

data/transactions/application.go
algomaxj: done, thanks!

algomaxj:
Alright, I think I have addressed the above comments

data/basics/userBalance.go
```diff
@@ -181,12 +270,35 @@ type BalanceDetail struct {
...
+// "creatables")
+type CreatableIndex uint64
+
+// CreatableType is an enum representing whether or not a given creatable is an
+// application or an asset
+type CreatableType uint64
+
+const (
+	// AssetCreatable is the CreatableType corresponding to assets
+	AssetCreatable CreatableType = 0
```
derbear, line 289: Consider using `iota` here

algomaxj: Ok, I'm curious about your opinion on this -- in this case when I upgrade the database I'm doing `ADD COLUMN ctype INTEGER DEFAULT 0` to the table containing asset IDs/app IDs. The table currently only contains assets. So I really need the `AssetCreatable` `CreatableType` to be 0, and not just distinct from the `AppCreatable` type. Would you still use `iota` in this case?

tsachiherman: Regardless of whether you choose to use `iota` or just zero, it would be a good idea to document the reason you've just mentioned.

algomaxj: For sure, added a comment, thanks :+1:

data/basics/teal.go
```diff
@@ -0,0 +1,312 @@
...
+		}
+	}
+	return true
+}
+
+// Valid checks whether the keys and values in a StateDelta conform to the
+// consensus parameters' maximum lengths
+func (sd StateDelta) Valid(proto *config.ConsensusParams) error {
+	for key, delta := range sd {
+		if len(key) > proto.MaxAppKeyLen || proto.MaxAppKeyLen == 0 {
```
tsachiherman, line 105: can we move the `proto.MaxAppKeyLen == 0` check outside the loop in case len(sd) > 0 ?
that would retain the logical behavior, but would conduct the test only once per function call.

data/transactions/logic/debugger.go
```diff
@@ -0,0 +1,281 @@
...
+	ds.Scratch = scratch
+
+	if (cx.runModeFlags & runModeApplication) != 0 {
+		if cx.globalStateCow != nil {
+			for k, v := range cx.globalStateCow.delta {
+				ds.GlobalStateChanges[k] = v
+			}
+		}
+		for addr, cow := range cx.localStateCows {
+			delta := make(basics.StateDelta)
```
tsachiherman, line 218: nit: `make(basics.StateDelta)` -> `make(basics.StateDelta, len(cow.cow.delta))`

algomaxj: done

data/basics/teal.go
algomaxj: done -- (also checked that `len(sd) > 0` to make it completely equivalent)

Makefile
```diff
@@ -190,13 +190,13 @@ test: build
 
 fulltest: build-race
 	for PACKAGE_DIRECTORY in $(UNIT_TEST_SOURCES) ; do \
-		go test $(GOTAGS) -timeout 2000s -race $$PACKAGE_DIRECTORY; \
+		go test $(GOTAGS) -timeout 2500s -race $$PACKAGE_DIRECTORY; \
```
tsachiherman, line 193: I'm not a big fan of extending the timeout, but it's not a problem with this change.

data/transactions/application.go
```diff
@@ -0,0 +1,776 @@
...
+			return fmt.Errorf("unknown delta action %d", valueDelta.Action)
+		}
+	}
+	return nil
+}
+
+// applyStateDeltas applies a basics.EvalDelta to the app's global key/value
+// store as well as a set of local key/value stores. If this function returns
+// an error, the transaction must not be committed.
+func (ac *ApplicationCallTxnFields) applyStateDeltas(evalDelta basics.EvalDelta, params basics.AppParams, creator, sender basics.Address, balances Balances, appIdx basics.AppIndex, errIfNotApplied bool) error {
```
derbear, line 215: Nit: It seems slightly more intuitive to me to call this `applyEvalDelta` and to call `applyDelta` instead `applyStateDelta`.

derbear: Also nit: it would be good to document `errIfNotApplied`.

derbear: Alternatively, instead of having `errIfNotApplied`, you could return different error types depending on whether you want your caller to filter out the error you returned and let the caller decide whether to drop the error.

data/transactions/application.go
```diff
@@ -0,0 +1,776 @@
...
+		// LocalDeltas are keyed by account index [sender, tx.Accounts[0], ...]
+		addr, err := ac.AddressByIndex(accountIdx, sender)
+		if err != nil {
+			return err
+		}
+
+		// Ensure this is not a duplicate address in case we were
+		// passed an invalid EvalDelta
+		_, ok := changes[addr]
+		if ok {
```
derbear, line 275: Is it okay to have duplicates of zero-deltas?

data/transactions/application.go
```diff
@@ -0,0 +1,776 @@
...
+			// stateful TEAL interpreter to apply state changes
+			ad.EvalDelta = stateDeltas
+		} else {
+			// Ignore errors and rejections from the ClearStateProgram
+		}
+
+		// Fetch the (potentially updated) sender record
+		record, err = balances.Get(sender, false)
+		if err != nil {
+			ad.EvalDelta = basics.EvalDelta{}
```
derbear, line 532: For my own understanding, this (and the clear at line 550) are unnecessary, right? If the intention is future-proofing, and the performance impact is not great, consider using `defer`.

data/transactions/application.go
```diff
@@ -0,0 +1,776 @@
...
+		if err != nil {
+			return
+		}
+	}
+
+	// Initialize our TEAL evaluation context. Internally, this manages
+	// access to balance records for Stateful TEAL programs. Stateful TEAL
+	// may only access
+	// - The sender's balance record
+	// - The balance records of accounts explicitly listed in ac.Accounts
```
derbear, line 610: Can a TEAL program use this mechanism to read global in additional to local state from foreign applications (without specifying it them in `ForeignApps`)?

config/consensus.go
```diff
@@ -217,6 +217,80 @@ type ConsensusParams struct {
...
+	// maximum number of total key/value pairs allowed by a given
+	// LocalStateSchema (and therefore allowed in LocalState)
+	MaxLocalSchemaEntries uint64
+
+	// maximum number of total key/value pairs allowed by a given
+	// GlobalStateSchema (and therefore allowed in GlobalState)
+	MaxGlobalSchemaEntries uint64
+
+	// maximum total minimum balance requirement for an account, used
+	// to limit the maximum size of a single balance record
```
derbear, line 292: Is `MaximumMinimumBalance` ever used?

derbear: Ah, it's planned for future use.

data/transactions/logic/doc.go
```diff
@@ -81,12 +83,33 @@ var opDocList = []stringString{
...
+	{"return", "use last value on stack as success value; end"},
 	{"pop", "discard value X from stack"},
 	{"dup", "duplicate last value on stack"},
+	{"dup2", "duplicate two last values on stack: A, B -> A, B, A, B"},
+	{"concat", "pop two byte strings A and B and join them, push the result"},
+	{"substring", "pop a byte string X. For immediate values in 0..255 N and M: extract a range of bytes from it starting at N up to but not including M, push the substring result"},
+	{"substring3", "pop a byte string A and two integers B and C. Extract a range of bytes from A starting at B up to but not including C, push the substring result"},
+	{"balance", "get balance for the requested account specified by Txn.Accounts[A] in microalgos. A is specified as an account index in the Accounts field of the ApplicationCall transaction"},
+	{"app_opted_in", "check if account specified by Txn.Accounts[A] opted in for the application B => {0 or 1}"},
+	{"app_local_get", "read from account specified by Txn.Accounts[A] from local state of the current application key B  => value"},
```
derbear, line 103: Nit: extra whitespace before the `=>` (and below)

data/transactions/logic/doc.go
```diff
@@ -125,14 +155,27 @@ func OpImmediateNote(opName string) string {
...
+	{"bnz", "The `bnz` instruction opcode 0x40 is followed by two immediate data bytes which are a high byte first and low byte second which together form a 16 bit offset which the instruction may branch to. For a bnz instruction at `pc`, if the last element of the stack is not zero then branch to instruction at `pc + 3 + N`, else proceed to next instruction at `pc + 3`. Branch targets must be well aligned instructions. (e.g. Branching to the second byte of a 2 byte op will be rejected.) Branch offsets are currently limited to forward branches only, 0-0x7fff. A future expansion might make this a signed 16 bit integer allowing for backward branches and looping.\n\nAt LogicSigVersion 2 it became allowed to branch to the end of the program exactly after the last instruction, removing the need for a last instruction or no-op as a branch target at the end. Branching beyond that may still fail the program."},
+	{"bz", "See `bnz` for details on how branches work."},
```
derbear, line 160: Might want to add here that `bz` inverts the behavior of `bnz`.

data/transactions/logic/doc.go
```diff
@@ -125,14 +155,27 @@ func OpImmediateNote(opName string) string {
...
 	{"bytecblock", "`bytecblock` loads the following program bytes into an array of byte string constants in the evaluator. These constants can be referred to by `bytec` and `bytec_*` which will push the value onto the stack. Subsequent calls to `bytecblock` reset and replace the bytes constants available to the script."},
 	{"*", "Overflow is an error condition which halts execution and fails the transaction. Full precision is available from `mulw`."},
 	{"txn", "FirstValidTime causes the program to fail. The field is reserved for future use."},
-	{"gtxn", "for notes on transaction fields available, see `txn`. If this transaction is _i_ in the group, `gtxn i field` is equivalent to `txn field`"},
-	{"btoi", "`btoi` panics if the input is longer than 8 bytes"},
+	{"gtxn", "for notes on transaction fields available, see `txn`. If this transaction is _i_ in the group, `gtxn i field` is equivalent to `txn field`."},
+	{"btoi", "`btoi` panics if the input is longer than 8 bytes."},
+	{"concat", "`concat` panics if the result would be greater than 4096 bytes."},
+	{"app_opted_in", "params: account index, application id (top of the stack on opcode entry)."},
```
derbear, line 169: Does a value of 1 here mean we've opted in, or the opposite?

data/transactions/logic/doc.go
```diff
@@ -125,14 +155,27 @@ func OpImmediateNote(opName string) string {
...
+	{"app_local_get", "params: account index, state key. Return: value. The value is zero if the key does ont exist."},
+	{"app_local_get_ex", "params: account index, application id, state key. Return: did_exist flag (top of the stack), value."},
+	{"app_global_get_ex", "params: application id, state key. Return: value."},
+	{"app_global_get", "params: state key. Return: value. The value is zero if the key does ont exist."},
```
derbear, line 173: Nit: `ont`

data/transactions/logic/doc.go
```diff
@@ -125,14 +155,27 @@ func OpImmediateNote(opName string) string {
...
+	{"app_local_put", "params: account index, state key, value."},
+	{"app_local_del", "params: account index, state key."},
+	{"app_global_del", "params: state key."},
+	{"asset_holding_get", "params: account index, asset id. Return: did_exist flag, value."},
+	{"asset_params_get", "params: account index, asset id. Return: did_exist flag, value."},
```
derbear, line 178: Does a value of 1 for `did_exist` mean the asset exists, or the opposite? (It might be a good idea to add a comment saying somewhere that by convention, a "1" returned by an opcode means "true" and a "0" means "false".

data/transactions/logic/doc.go
```diff
@@ -220,11 +255,31 @@ func TypeNameDescription(typeName string) string {
 	return "invalid type name"
 }
 
+// see assembler.go TxnTypeNames
+// also used to parse symbolic constants for `int`
+var onCompletionDescriptions = map[OnCompletionConstType]string{
+	NoOp:              "Application transaction will simply call its ApprovalProgram.",
+	OptIn:             "Application transaction will allocate some LocalState for the application in the sender's account.",
+	CloseOut:          "Application transaction will deallocate some LocalState for the application from the user's account.",
+	ClearState:        "Similar to CloseOutOC, but may never fail. This allows users to reclaim their minimum balance from an application they no longer wish to opt in to.",
```
derbear, line 264: Nit: `CloseOutOC` -> `CloseOut`

data/transactions/logic/assembler.go
```diff
@@ -368,23 +527,26 @@ func assembleByte(ops *OpStream, args []string) error {
...
+	for i, xs := range args {
 		cu, err := strconv.ParseUint(xs, 0, 64)
 		if err != nil {
 			return err
 		}
 		l = binary.PutUvarint(scratch[:], cu)
 		ops.Out.Write(scratch[:l])
+		ops.intc[i] = cu
 	}
+	ops.noIntcBlock = true
```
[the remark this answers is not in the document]
derbear: I see. The naming seems a little counterintuitive to me (it seems more natural to set it if and only if it were _not_ called).

data/transactions/application.go
algomaxj: I went ahead and just disallowed zero-length deltas. There is no reason to support them since they are never generated.

data/transactions/application.go
algomaxj: renamed methods and added docs for `errIfNotApplied`

PR #1139, opened 2020-06-05
Restructure config package to use reflection

tsachiherman:
## Summary
Existing config files are using hard-coded data structure to represent versioned configuration setting. The per-version structures need to be duplicated with every change to the configuration structure, which makes it error prone, especially when dealing with non-default field values.
( since these non-default field values require some explicit written code to handle the config version upgrades )

Replacing that with a configuration struct which is self-tagged with the default values, allow us to auto-generate the above logic and avoid programmatic mistakes. 

## Test Plan

Unit tests were updated

## Release Plan

This change is targeted for the July release

## Notes

This is a re-opened version of https://github.com/algorand/go-algorand/pull/1114
Compared to previous version of this PR, this one code-generate the defaultLocal variable in the local_defaults.go, as well as code-generate the ./installer/config.json.example file.
These auto-generated files gives us the ability to review current values more clearly, while keeping the "source" of the configuration schema & data in one location.


config/local_defaults.go
```diff
@@ -18,175 +18,57 @@ package config
...
-	// DO NOT MODIFY VALUES - New values may be added carefully - See WARNING at top of file
+// The following defaultLocal is auto-generated. !! DO NOT EDIT !!
+// if you want to make changes to this structure, make the changes to the Local structure in config.go and run "make generate" which would automatically
+// update this one.
+var defaultLocal = Local{
 	Version:                               8,
-	Archival:                              false,
-	BaseLoggerDebugLevel:                  4,
-	BroadcastConnectionsLimit:             -1,
 	AnnounceParticipationKey:              true,
```
derbear, line 33: I like having these in sorted order.

config/local_defaults.go
```diff
@@ -195,379 +77,198 @@ var defaultLocalV6 = Local{
...
-	// When we have more time, we can use reflection to migrate from initial
-	// version to latest version (progressively applying defaults)
-	// Migrate 0 -> 1
-	if newCfg.Version == 0 {
-		if newCfg.BaseLoggerDebugLevel == defaultLocalV0.BaseLoggerDebugLevel {
-			newCfg.BaseLoggerDebugLevel = defaultLocalV1.BaseLoggerDebugLevel
-		}
-		if newCfg.IncomingConnectionsLimit == defaultLocalV0.IncomingConnectionsLimit {
-			newCfg.IncomingConnectionsLimit = defaultLocalV1.IncomingConnectionsLimit
+	for {
```
derbear, line 115: I might be missing something here, but did you intend to write `for ; newCfg.Version != latestConfigVersion; newCfg.Version++` instead?

config/local_defaults.go
```diff
@@ -195,379 +77,198 @@ var defaultLocalV6 = Local{
...
-	if newCfg.Version == 6 {
-		if newCfg.CatchpointInterval == defaultLocalV6.CatchpointInterval {
-			newCfg.CatchpointInterval = defaultLocalV7.CatchpointInterval
+	// apply version specific changes.
+	localType := reflect.TypeOf(local)
+	for fieldNum := 0; fieldNum < localType.NumField(); fieldNum++ {
+		field := localType.Field(fieldNum)
+		versionDefaultValue, hasTag := reflect.StructTag(field.Tag).Lookup(fmt.Sprintf("version[%d]", version))
+		if !hasTag {
+			continue
```
derbear, line 217: What's the intended setting of the default value for `LogArchiveName` if `version = 5` was passed in? Is it `""` or `"node.archive.log"`? I think the former is inconsistent with old behavior.

algorandskiy:
>I think it's important to make sure that any file is either wholly autogenerated or wholly intended for user modification.

I second this

config/config.go
```diff
@@ -46,42 +46,55 @@ const Mainnet protocol.NetworkID = "mainnet"
...
+// test/testdata/configs/config-v{n}.json and call "make generate" to regenerate the constants.
+//
+// !!! WARNING !!!
 type Local struct {
 	// Version tracks the current version of the defaults so we can migrate old -> new
 	// This is specifically important whenever we decide to change the default value
-	// for an existing parameter.
-	Version uint32
+	// for an existing parameter. This field tag must be updated any time we add a new version.
+	Version uint32 `version[0]:"0" version[1]:"1" version[2]:"2" version[3]:"3" version[4]:"4" version[5]:"5" version[6]:"6" version[7]:"7" version[8]:"8"`
```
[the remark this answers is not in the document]
tsachiherman: We need to be able to read this value off the current config file, and therefore it would helpful to keep this field here. As for the fact that this field has many entries - yes, we could make this field to have an exception where it would have a different struct tag which would denote the latest supported version only. I think that this could be broken into a separate PR though.

config/defaults_gen/defaults_gen.go
```diff
@@ -0,0 +1,171 @@
...
+)
+
+const localDefaultsFileName = "./local_defaults.go"
+const exampleFileName = "../installer/config.json.example"
+const defaultLocalVariableDeclaration = "var defaultLocal"
+
+func main() {
+	localDefaultsBytes, err := ioutil.ReadFile(localDefaultsFileName)
+	if err != nil {
+		fmt.Printf("Unable to load file %s : %v", localDefaultsFileName, err)
```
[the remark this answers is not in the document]
tsachiherman: sounds good.

config/config.go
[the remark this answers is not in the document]
algorandskiy: Additionally `version[x]="0"` looks too verbose is case of multiple changes, maybe `v[x]="0"` ? Or even `0:"0"` ?

config/local_defaults.go
tsachiherman: no, I did not. The outcome is the same nevertheless.
The version field value is being "upgraded", from version x -> version x+1 in the inner for loop.

config/local_defaults.go
tsachiherman: The definition for LogArchiveName is
LogArchiveName string `version[4]:"node.archive.log"`
In the above case, calling the Lookup would return hasTag = false, which would ignore this variable for version 5.
this means that it would keep the value it had for version 4.

config/config.go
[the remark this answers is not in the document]
tsachiherman: With the exception of the version field, we rarely have more than two versions. I'd rather stick with code clarity over "compactness" here.

installer/config.json.example
```diff
@@ -33,26 +42,29 @@
...
     "SuggestedFeeSlidingWindowSize": 50,
+    "TLSCertFile": "",
+    "TLSKeyFile": "",
     "TelemetryToLog": true,
     "TxPoolExponentialIncreaseFactor": 2,
     "TxPoolSize": 15000,
     "TxSyncIntervalSeconds": 60,
     "TxSyncServeResponseSize": 1000000,
     "TxSyncTimeoutSeconds": 30,
-    "PeerConnectionsUpdateInterval": 3600,
```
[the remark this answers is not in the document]
tsachiherman: The sorted is just for the config.json, which is user facing. The internal data structure is using the go-struct ordering.

config/local_defaults.go
derbear: I see--does this cause an infinite loop, then? Since (I believe) `newCfg` never gets reassigned to in the body of the loop.

config/local_defaults.go
derbear: Ah, right--I missed the recursive call on line 209.

config/config_test.go
```diff
@@ -493,3 +427,68 @@ func TestLocal_DNSBootstrap(t *testing.T) {
...
+		versionTags = append(versionTags, fmt.Sprintf("version[%d]", ver))
+		ver++
+	}
+
+	for fieldNum := 0; fieldNum < localType.NumField(); fieldNum++ {
+		field := localType.Field(fieldNum)
+		if field.Tag == "" {
+			require.Failf(t, "Field is missing versioning information", "Field Name: %s", field.Name)
+		}
+		if field.Name == "Version" {
```
algorandskiy, line 452: nit: add a comment `Version` is tested in `TestLocalVersionField`

config/defaults_gen/defaults_gen.go
```diff
@@ -0,0 +1,179 @@
+// Copyright (C) 2019-2020 Algorand, Inc.
```
algorandskiy, line 1: the name looks confusing, I thought this file is auto-generated

config/migrate.go
```diff
@@ -0,0 +1,200 @@
...
+	"reflect"
+	"strconv"
+)
+
+//go:generate $GOROOT/bin/go run ./defaults_gen/defaults_gen.go -h ../scripts/LICENSE_HEADER -p config -o ./local_defaults.go -j ../installer/config.json.example
+//go:generate $GOROOT/bin/go fmt local_defaults.go
+
+// AutogenLocal - this variable is the "input" for the config default generator which automatically updates the above defaultLocal varaible.
+// it's implemented in ./config/defaults_gen.go, and should be the only "consumer" of this exported variable
+var AutogenLocal = getVersionedDefaultLocalConfig(getLatestConfigVersion())
```
algorandskiy, line 30: Why is it done this way? the generator might call `getVersionedDefaultLocalConfig(getLatestConfigVersion())` or `getVersionedDefaultLocalConfig(version)` with version from the command line.

config/config_test.go
```diff
@@ -314,68 +293,23 @@ func TestConfigMigrateFromDisk(t *testing.T) {
 func TestConfigInvariant(t *testing.T) {
 	a := require.New(t)
 
-	a.Equal(uint32(9), configVersion, "If you bump Config Version, please update this test (and consider if you should be adding more)")
```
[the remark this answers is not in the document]
tsachiherman: I want to reduce the number of changes needed with every config version change.. In the case that getLatestConfigVersion() is 10, and we don't have a `config-v10.json`, we would already fail this unit test few lines below that.

config/config_test.go
tsachiherman: sure.

config/migrate.go
tsachiherman: The generator is in a separate package, and cannot access these methods directly. Sure, we could expose the functions - but then we're other parties as well. Maybe this should be refactored using internal package ? I would rather finish this pass before refactoring it further, as the gained value diminish.

config/defaults_gen/defaults_gen.go
tsachiherman: do you think that defaults_generator or defaultsGenerator as the package name would be more appropriate ?

derbear: `defaultsGenerator` seems like a better name

config/defaults_gen/defaults_gen.go
```diff
@@ -0,0 +1,179 @@
...
+var autoGenHeader = `
+// This file was auto generated by ./config/defaults_gen/defaults_gen.go, and SHOULD NOT BE MODIFIED in any way
+// If you want to make changes to this file, make the corresponding changes to Local in config.go and run "go generate".
+`
+
+func main() {
+
+	flag.Parse()
+	if *outputfilename == "" || *packageName == "" || *headerFileName == "" || *jsonExampleFileName == "" {
+		fmt.Printf("one or more of the required input arguments was not provided\n")
```
derbear, line 48: (Repeating an earlier comment) I think adding `os.Args[0]` to the error string is good because if code generation fails, we'll be able to trace the failure to this tool or rule it out as cause of failure.

PR #1143, opened 2020-06-05
Add license headers to dbgen generated files

tsachiherman:
## Summary

Since we added the headers to all our go files, we had to maintain them to have these headers.
That didn't take into account auto-generated go files, such as agreeInstall.go.
This file is generated and when it does, the license header is missing. To correct this, I've added the license header to the generator.

## Notes

The trigger for this is the change in the config, which encourage developer to user the `make generate`, which is currently broken ( i.e. prior to this change ).


data/account/account.go
```diff
@@ -16,8 +16,7 @@
 
 package account
 
-//go:generate dbgen -i root.sql -p account -n root -o rootInstall.go
-//go:generate dbgen -i part.sql -p account -n part -o partInstall.go
```
tsachiherman, line 20: the partInstall.go isn't an auto-generated file, and the part.sql doesn't exists either.

agreement/persistence.go
```diff
@@ -16,7 +16,6 @@
 
 package agreement
 
-//go:generate dbgen -i agree.sql -p agreement -n agree -o agreeInstall.go
```
tsachiherman, line 19: This is a duplicate generation of the same file as below.

algorandskiy: nice catch

PR #1146, opened 2020-06-08
Onetechnical/relstable2.0.7

onetechnical:
## Summary

Algorand 2.0.7-stable

## Test Plan

N/A


scripts/configure_dev-deps.sh
```diff
@@ -2,9 +2,24 @@
 
 set -ex
 
+function get_go_version {
+    (
+	cd $(dirname "$0")
+	VERSION=$(cat ../go.mod | grep "$1" 2>/dev/null | awk -F " " '{print $2}')
```
tsachiherman, line 8: I think that if we were to use this for the long run, we need to add `pipefail` as well as `export SHELLOPTS`

PR #1147, opened 2020-06-09
FastCatchup - add unit testing for ledgerFetcher

tsachiherman:
## Summary

Add testing for the ledgerFetcher

## Notes

1. Refactor ledger.CatchpointCatchupAccessor into an Interface, so it could be mocked out and abstracted.
2. Create a mockCatchpointCatchupAccessor.go
3. Create a unit tests for ledgerFetcher which covers all the error execution paths.


ledger/catchupaccessor.go
```diff
@@ -62,23 +109,23 @@ const (
 )
 
 // MakeCatchpointCatchupAccessor creates a CatchpointCatchupAccessor given a ledger
-func MakeCatchpointCatchupAccessor(ledger *Ledger, log logging.Logger) *CatchpointCatchupAccessor {
+func MakeCatchpointCatchupAccessor(ledger *Ledger, log logging.Logger) CatchpointCatchupAccessor {
```
wjurayj-algo, line 112: Why is the output type no longer a pointer if it still returns the address of the CCA?

catchup/ledgerFetcher_test.go
```diff
@@ -0,0 +1,98 @@
...
+	lf := makeLedgerFetcher(&mocks.MockNetwork{}, &mocks.MockCatchpointCatchupAccessor{}, logging.TestingLog(t), &dummyLedgerFetcherReporter{})
+	peer := testHTTPPeer(listener.Addr().String())
+	err = lf.getPeerLedger(context.Background(), &peer, basics.Round(0))
+	require.Equal(t, errNoLedgerForRound, err)
+
+	httpServerResponse = http.StatusInternalServerError
+	err = lf.getPeerLedger(context.Background(), &peer, basics.Round(0))
+	require.Equal(t, fmt.Errorf("getPeerLedger error response status code %d", httpServerResponse), err)
+
+	httpServerResponse = http.StatusOK
```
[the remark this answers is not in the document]
tsachiherman: This assignment is updating the variable declared on line 69. it's being used on line 75.

ledger/catchupaccessor.go
tsachiherman: An interface in go is a pointer-type. Hence, there is no need to return a pointer to it.

wjurayj-algo: Ah, good to know!

tsachiherman:
> Would be nice to have a happy case test as well

definitely. I'm planning to get there. however, I'm trying to prioritize the use cases which aren't covered at all by other tests first. The success case is covered by the e2e test, so it "defaults" to have a lower priority.

PR #1151, opened 2020-06-10
Add byteslice checking to the msgp decoder

tsachiherman:
## Solution
Currently, we don't have a bounds on a byte slices in the msgp decoder. This change allows us to limit the size of a byte slice using the allocbound mechanism used for other slice/map types.

Specifically, this change add the allocbound checking to the LogicSig and Notes field.

## MSGP changes
The corresponding msgp changes are in https://github.com/algorand/msgp/pull/6


tsachiherman:
> A test needed for checking all Max constants are non-zero.

sure. I'll add it on the next PR.

PR #1153, opened 2020-06-10
Regenerate routes to allow 'pretty' parameter.

winder:
## Summary

I updated the oapi-codegen template to always include the 'pretty' parameter. This is something echo gives us for free, but was accidentally disabled with the strict parameter checking.

## Test Plan

Previously `pretty` was rejected by the REST API. So I made sure that it is no longer rejected, and that it works as expected:
```
will@will-algorand:~/go/src/github.com/algorand/go-algorand$ curl -H "Authorization: bearer $(cat ~/.algorand/algod.token)" $(cat ~/.algorand/algod.net)/v2/status
{"catchpoint":"","catchpoint-acquired-blocks":0,"catchpoint-processed-accounts":0,"catchpoint-total-accounts":0,"catchpoint-total-blocks":0,"catchup-time":106228045571,"last-catchpoint":"","last-round":5268737,"last-version":"https://github.com/algorandfoundation/specs/tree/e5f565421d720c6f75cdd186f7098495caf9101f","next-version":"https://github.com/algorandfoundation/specs/tree/e5f565421d720c6f75cdd186f7098495caf9101f","next-version-round":5268738,"next-version-supported":true,"stopped-at-unsupported-round":false,"time-since-last-round":8421718}


will@will-algorand:~/go/src/github.com/algorand/go-algorand$ curl -H "Authorization: bearer $(cat ~/.algorand/algod.token)" $(cat ~/.algorand/algod.net)/v2/status?pretty
{
  "catchpoint": "",
  "catchpoint-acquired-blocks": 0,
  "catchpoint-processed-accounts": 0,
  "catchpoint-total-accounts": 0,
  "catchpoint-total-blocks": 0,
  "catchup-time": 108446796064,
  "last-catchpoint": "",
  "last-round": 5268864,
  "last-version": "https://github.com/algorandfoundation/specs/tree/e5f565421d720c6f75cdd186f7098495caf9101f",
  "next-version": "https://github.com/algorandfoundation/specs/tree/e5f565421d720c6f75cdd186f7098495caf9101f",
  "next-version-round": 5268865,
  "next-version-supported": true,
  "stopped-at-unsupported-round": false,
  "time-since-last-round": 32976904
}
```


daemon/algod/api/Makefile
```diff
@@ -23,6 +23,6 @@ algod.oas3.yml:	.3tmp.json
 	curl -s -X POST "https://converter.swagger.io/api/convert" -H "accept: application/json" -H "Content-Type: application/json" -d @./algod.oas2.json -o .3tmp.json
 
 oapi-codegen:	.PHONY
-	GO111MODULE=on go get -u "github.com/algorand/oapi-codegen/...@v1.3.5-algorand4"
+	GO111MODULE=on go get -u "github.com/algorand/oapi-codegen/...@v1.3.5-algorand5"
```
tsachiherman, line 26: we need to find a way to derive that from the go.mod, so that upgrading the oapi-codegen would have a single place where we would make the change. but that's not an issue for this PR.

daemon/algod/api/server/v2/generated/routes.go
```diff
@@ -60,7 +60,9 @@ type ServerInterfaceWrapper struct {
 // AccountInformation converts echo context to params.
 func (w *ServerInterfaceWrapper) AccountInformation(ctx echo.Context) error {
 
-	validQueryParams := map[string]bool{}
+	validQueryParams := map[string]bool{
```
tsachiherman, line 63: given that this requires a memory allocation, you might want to refactor this to be a global map variable.

winder: Thanks for the reminder. Yes, this is an improvement that I'm aware of. It will require spending a bit more time ramping up on go's template engine.

daemon/algod/api/Makefile
winder: I'm open to suggestions for how to do this. Technically it's not a bug if they mismatch because it's a library and a cli tool. So far changes have been to one or the other, not both. It doesn't look right to have different versions though, so I've been trying to keep them synchronized.

brianolson: This gets me pretty close?
`grep oapi-codegen go.mod | awk '{ print $1 "/...@" $2 }'`

PR #1154, opened 2020-06-10
Algod telemetry config location logging

wjurayj-algo:
## Summary

Removed non-front facing flag, externalHostName, from diagcfg metric command.
Added logging for telemetry config file location in algod and algoh startup processes.

## Test Plan

Added e2e expect testing framework for algod, with a single test validating algod pulled telemetry config from the expected location.
Performed several manual tests of the diagcfg CLI to ensure that the removed flag didn't alter any functionality.


brianolson:
I think the expect test is just watching `algod` stderr for the log line you added? As such it doesn't need a full copy of testnet genesis.json (which is available elsewhere anyway) but just a local private network `algod` starting up should make the same log line, right? (and also be a more well contained test that doesn't even try for network access to the outside world)

tsachiherman:
Could you mark this PR as "Ready for review", followed by re-running this PR until it passes the Travis CI ?

PR #1156, opened 2020-06-10
Add a more restrictive timeout for block fetching

tsachiherman:
## Summary

The timeouts values used by the fast catchup service doesn't handle failure cases well. Since these are hard-coded, we can't really address any field-deployment issue by tweaking the configuration settings.

This PR moves the hard-coded values to the configuration file, along with providing more robust default values. These should allow us to adjust them dynamically if (and when) needed.


catchup/service.go
```diff
@@ -152,7 +152,7 @@ func (s *Service) SynchronizingTime() time.Duration {
 
 // function scope to make a bunch of defer statements better
 func (s *Service) innerFetch(fetcher Fetcher, r basics.Round) (blk *bookkeeping.Block, cert *agreement.Certificate, rpcc FetcherClient, err error) {
-	ctx, cf := context.WithTimeout(s.ctx, DefaultFetchTimeout)
+	ctx, cf := context.WithCancel(s.ctx)
```
algonautshant, line 155: Can you please explain why we don't need the timeout here?

catchup/catchpointService.go
```diff
@@ -283,7 +277,7 @@ func (cs *CatchpointCatchupService) processStageLastestBlockDownload() (err erro
 		return cs.abort(fmt.Errorf("processStageLastestBlockDownload failed to retrieve catchup block round : %v", err))
 	}
 
-	fetcherFactory := MakeNetworkFetcherFactory(cs.net, 10, nil)
+	fetcherFactory := MakeNetworkFetcherFactory(cs.net, 10, nil, &cs.config)
```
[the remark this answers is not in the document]
tsachiherman: as far as I can tell, it's an historical configuration option that used to limit the number of concurrent downloads. It's no longer being used.

catchup/service.go
tsachiherman: This timeout would automatically cancel the context. Content cancellation (naturally) is only meaningful when being examined. Concretely, it's being examined within the http and gossip fetchers themselves, where I have replaced it with a fetcher-specific timeout context. ( i.e. moving the global one into a fetcher-specific one, and also making it configurable )

Issue #1109, opened 2020-05-27
Make `/teal` end points disabled by default


stechu:
A node could be DDoSed by the compile API call. So we should me this feature disabled by default.

PR #1163, opened 2020-06-11
teal/compile config

stechu:
## Summary

Add a local config to disable `teal/compile` in algod by default. This PR solves #1109.


daemon/algod/api/server/v2/handlers.go
```diff
@@ -588,6 +589,11 @@ func (v2 *Handlers) AbortCatchup(ctx echo.Context, catchpoint string) error {
 // TealCompile compiles TEAL code to binary, return both binary and hash
 // (POST /v2/teal/compile)
 func (v2 *Handlers) TealCompile(ctx echo.Context) error {
+	// return early if teal compile is not allowed in node config
+	if (! v2.Node.Config().EnableTealCompile) {
+		return ctx.String(http.StatusLocked, "/teal/compile disabled by node config.")
+	}
+
```
tsachiherman, line 596: I'll let @rotemh and @winder make the call on this one, but I think it would be more appropriate to prevent this handler to be reached at all if the `EnableTealCompile` endpoint is not set. ( i.e. error 404 )

 Also, I think that we should bundle the TealCompile, TealSign and DryRun into a single toggle in the config file, or maybe extend the use of the EnableProfiler, as it seems to have similar usage patterns.

algorandskiy: > we should bundle the TealCompile, TealSign and DryRun into a single toggle in the config file

I second this

stechu: @tsachiherman the difference in terms of DDoS resilience would be minimal since function call is cheap. I use 403 here just to inform the user who want to enable `teal/compile` to change node config.

stechu: Bundling the config flag for multiple things make sense, will change the name of the flag to include other things.

tsachiherman: My concern here wasn't purely DDoS; we need to "stick" with our "flow". In the case of the Profiler, we're currently removing the handler. We should do the same for all other non-enabled use cases as well ( unless there is a difference between the use cases )

stechu: @tsachiherman that makes sense. I wrote this PR based on the assumption that how do we currently using handlers. Disable non-enabled use case seems to be a better design decision. Shall I close this PR?

tsachiherman: No please don't close this PR. I think that it's very important for us to have the ability to toggle these new features. Maybe @winder could provide some more concrete lead on this one ?

winder: Sorry for the delay.

I don't mind starting with a simple short-circuit to disable these in the handler. Unless there are additional restrictions we want to add (like an admin token used with the private endpoints), having another generated package seems like overkill.

daemon/algod/api/server/v2/handlers.go
```diff
@@ -588,6 +589,11 @@ func (v2 *Handlers) AbortCatchup(ctx echo.Context, catchpoint string) error {
 // TealCompile compiles TEAL code to binary, return both binary and hash
 // (POST /v2/teal/compile)
 func (v2 *Handlers) TealCompile(ctx echo.Context) error {
+	// return early if teal compile is not allowed in node config
+	if (! v2.Node.Config().EnableTealService) {
+		return ctx.String(http.StatusLocked, "/teal/compile disabled by node config.")
```
winder, line 594: I'm not sure the implications of "StatusLocked", a 404 is probably fine. It looks like there is a shortcut for this, so you can return `echo.ErrNotFound`

tsachiherman:
I'm closing this one after opening a clone of this one here : https://github.com/algorand/go-algorand/pull/1191

PR #1165, opened 2020-06-11
Eric/fix keyreg teal test

egieseke:
## Summary

Fixed bug in key registration teal test.

## Test Plan

make clean install, make test, make integration.


brianolson:
Looks good. I notice 4 other uses of `goal node wait` and I wonder if they should have a similar more generous timeout? (They got the `-lt` right though)

egieseke:
> Looks good. I notice 4 other uses of `goal node wait` and I wonder if they should have a similar more generous timeout? (They got the `-lt` right though)

@brianolson, I added timeouts to the 3 other goal node wait instances.

PR #1167, opened 2020-06-12
Update code to take advantage of byteslice allocbounding

tsachiherman:
## Summary

Add a allocbound for the byte slices definition in our code that gets message-packed.
The unit testing is verifying that we didn't miss any additional instance.

## Test Plan

Auto-generated tests were updated, and detection of missing allocbound tag testing was updated.

### DIFF
--- a/config/consensus.go
+++ b/config/consensus.go
@@ -321,6 +321,14 @@ var MaxLogicSigMaxSize int
 // of the consensus protocols. used for decoding purposes.
 var MaxTxnNoteBytes int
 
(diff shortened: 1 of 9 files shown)


zeldovich:
Tangentially related (prompted by the extent to which you are adding allocbounds): it would be good if we checked the allocbound on the encode path, to make sure that the ledger DB never contains data that we cannot decode.  If we hit the allocbound on the encode path, that's almost certainly a bug elsewhere, but finding it at encode time could help us track it down easier and faster than if we discover some items in the DB that we can't decode, and we don't quite know how they got there.

PR #1169, opened 2020-06-12
Handle stopping and aborting in catchpoint catchup service

tsachiherman:
## Summary

When the node is in fast catchup mode, either stopping the node or aborting the catchup does not work in fashionable manner.
The stop/abort operation is expected to happen right away, whereas we had some long delays before it would complete.

## Test Plan

Both node stopping during fast catchup as well as aborting the fast catchup were tested manually.
I was able to confirm the issue with both of them, as well as confirm that the issue was resolved after the change.

### DIFF
--- a/catchup/catchpointService.go
+++ b/catchup/catchpointService.go
@@ -568,7 +568,8 @@ func (cs *CatchpointCatchupService) abort(originatingErr error) error {
 	}
 	cs.updateNodeCatchupMode(false)
 	// we want to abort the catchpoint catchup process, and the node already reverted to normal operation.
-	// at this point, all we need to do is to abort the run function.
+	// as part of the returning to normal operation, we've re-created our context. This context need to be
+	// canceled so that when we go back to run(), we would exit from there right away.
 	cs.cancelCtxFunc()
 	return outError
 }
--- a/catchup/catchpointService.go
+++ b/catchup/catchpointService.go
@@ -146,6 +146,9 @@ func (cs *CatchpointCatchupService) Start(ctx context.Context) {
 
 // Abort aborts the catchpoint catchup process
 func (cs *CatchpointCatchupService) Abort() {
+	// In order to abort the catchpoint catchup process, we need to first set the flag of abortCtxFunc, and follow that by canceling the main context.
+	// The order of these calls is crucial : The various stages are blocked on the main context. When that one expires, it uses the abort context to determine
+	// if the cancelation meaning that we want to shut down the process, or aborting the catchpoint catchup completly.
 	cs.abortCtxFunc()
 	cs.cancelCtxFunc()
 }
--- a/catchup/catchpointService.go
+++ b/catchup/catchpointService.go
@@ -52,24 +52,36 @@ type CatchpointCatchupStats struct {
 
 // CatchpointCatchupService represents the catchpoint catchup service.
 type CatchpointCatchupService struct {
-	stats          CatchpointCatchupStats
-	statsMu        deadlock.Mutex
-	node           CatchpointCatchupNodeServices
-	ctx            context.Context
-	cancelCtxFunc  context.CancelFunc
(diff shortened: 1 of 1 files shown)


catchup/catchpointService.go
```diff
@@ -124,12 +139,14 @@ func MakeNewCatchpointCatchupService(catchpoint string, node CatchpointCatchupNo
...
 func (cs *CatchpointCatchupService) Start(ctx context.Context) {
 	cs.ctx, cs.cancelCtxFunc = context.WithCancel(ctx)
+	cs.abortCtx, cs.abortCtxFunc = context.WithCancel(context.Background())
 	cs.running.Add(1)
 	go cs.run()
 }
 
 // Abort aborts the catchpoint catchup process
 func (cs *CatchpointCatchupService) Abort() {
+	cs.abortCtxFunc()
```
algorandskiy, line 149: please add a comment saying about cancellation func must be called before cancel cancellation func

tsachiherman: done.

catchup/catchpointService.go
```diff
@@ -523,6 +547,15 @@ func (cs *CatchpointCatchupService) processStageSwitch() (err error) {
 	return nil
 }
 
+// stopOrAbort is called when any of the stage processing function sees that cs.ctx has been canceled. It can be
```
algonautshant, line 553: I find the following confusing: calling stopOrAbort will cancel cs.ctx which is already canceled. 

Let me describe how I understand it: 
- cs.abort does two things: 
       (A) fix internal state: ResetStagingBalances and updateNodeCatchupMode
       (B) cancel cs.ctx
- stopOrAbort will be called if cs.ctx has been canceled. Then, it calls cs.abort

In effect, stopOrAbort will, among other things (the (A) part above), will cancel cs.ctx which is already canceled. 

Will make it clearer, if the (A) part is removed from abort into a separate func, and called from abort and stopOrAbort.

catchup/catchpointService.go
```diff
@@ -523,6 +547,15 @@ func (cs *CatchpointCatchupService) processStageSwitch() (err error) {
...
+// due to the end user attempting to abort the current catchpoint catchup operation or due to a node shutdown.
+func (cs *CatchpointCatchupService) stopOrAbort() error {
```
algonautshant, line 555: I understand that  StopOrAbort is meaningful only if it is preceded by an Abort or Stop calls. 
If this is correct, then the name StopOrAbort is misleading.

catchup/catchpointService.go
tsachiherman: I think that the part that there is one point that you didn't mentioned here : 
The `abort` function is doing the following:
```
	cs.updateNodeCatchupMode(false)
	// we want to abort the catchpoint catchup process, and the node already reverted to normal operation.
	// at this point, all we need to do is to abort the run function.
	cs.cancelCtxFunc()
```
note that the second portion is canceling the context created in `updateNodeCatchupMode`. This is required, since it's a "brand new context".

The `abort` is called due to either external `Abort` call, or due to internal problem.

catchup/catchpointService.go
tsachiherman: I'm open to alternative naming for the function.

catchup/catchpointService.go
algonautshant: > note that the second portion is canceling the context created in `updateNodeCatchupMode`. This is required, since it's a "brand new context".

Okay, this makes sense now. 
Can you please add a comment about this when cs.cancelCtxFunc() is called in abort?

tsachiherman: done.

catchup/catchpointService.go
algonautshant: how about AbortAndRevertNode ?

PR #1170, opened 2020-06-14
Make account changeonlinestatus online flag optional

egieseke:
## Summary

Since the `online` option in the `goal account changeonlinestatus` defaults to true, updated the command so that the `online` option is not required. 

## Test Plan

Ran `make test` and `make integration`.


egieseke:
There is an integration test for this today, which uncovered the issue as part of the Jenkins/Mule integration work.  The test is:  test/e2e-go/cli/goal/expect/listExpiredParticipationKeyTest.exp

tsachiherman:
> @tsachiherman There is an integration test for this today, which uncovered the issue as part of the Jenkins/Mule integration work. The test is: test/e2e-go/cli/goal/expect/listExpiredParticipationKeyTest.exp

Yes, but this integration test is clearly broken. Otherwise, it would have failed before.
( i.e. it should have consistently failed without the changes you've made here. this clearly did not happen )

For this PR to be complete, it should demonstrate the not passing the "online" flag at all is equivalent to passing a "online true", as well as that the "online false" is working as expected.

PR #1171, opened 2020-06-15
Mark the in-body parameter as binary

algonautshant:
Mark the compile teal param as binary in the spec file: 
format": "binary"


daemon/algod/api/algod.oas2.json
```diff
@@ -726,7 +726,8 @@
             "in": "body",
             "required": true,
             "schema": {
-              "type": "string"
+              "type": "string",
+              "format": "binary"
```
algorandskiy, line 730: Why binary? it is text source.
And seems like `binary` format is [not supported by oapi-codegen](https://github.com/deepmap/oapi-codegen/blob/master/pkg/codegen/schema.go#L260).

algonautshant: It is handy to distinguish string (type) from in-body with extra tag (format) binary so that the getter will return a byte array and not string for writing to the client body. 

This is similar to: 
            "description": "The byte encoded signed transaction to broadcast to network",
            "name": "rawtxn",
            "in": "body",
            "required": true,
            "schema": {
              "type": "string",
              "format": "binary"
            }

algorandskiy: got it, thanks

PR #1173, opened 2020-06-16
Fix content of hashes file(s)

btoll:
## Summary

This removes "./" from the beginning of some lines in the hashes files.

## Test Plan

If the releases page generates the checksums correctly, then this works.


btoll:
For go-algorand: https://github.com/algorand/go-algorand/pull/1173
For indexer: https://github.com/algorand/indexer/pull/130

PR #1177, opened 2020-06-16
Reduce parallel testings on travis

tsachiherman:
## Summary

Reducing the amount of parallel testing on travis is likely to reduce the memory demand stress. On travis machines, where memory is limited, this could help us avoid OOM cases.

### DIFF
--- a/scripts/travis/run_tests.sh
+++ b/scripts/travis/run_tests.sh
@@ -16,10 +16,12 @@ if [ "${BUILD_TYPE}" = "integration" ]; then
     ./test/scripts/run_integration_tests.sh
 elif [ "${TRAVIS_EVENT_TYPE}" = "cron" ] || [[ "${TRAVIS_BRANCH}" =~ ^rel/ ]]; then
     if [[ "${OS}" != "darwin" ]]; then
-	make fulltest -j4
+	    make fulltest -j2
     fi
 else
     if [[ "${OS}" != "darwin" ]]; then
-	make shorttest -j4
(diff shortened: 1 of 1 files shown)


scripts/travis/run_tests.sh
```diff
@@ -16,10 +16,12 @@ if [ "${BUILD_TYPE}" = "integration" ]; then
...
     if [[ "${OS}" != "darwin" ]]; then
-	make fulltest -j4
+	    make fulltest -j2
     fi
 else
     if [[ "${OS}" != "darwin" ]]; then
-	make shorttest -j4
+        # setting it to 1 disable parallel making. This is done specicifically for travis, as travis seems to
+        # have memory limitations and setting this to 1 could reduce the likelihood of hitting these.
+	    make shorttest -j1
```
algorandskiy, line 25: too radical maybe we better measure mem usage of unit tests with bin-search?

tsachiherman: > too radical maybe we better measure mem usage of unit tests with bin-search?

The test is currently invoking 4 packages tester concurrently. The package ordering and the tests themselves are non-deterministic. If I knew that there is one unit test which consumes lots of memory whereas all the rest aren't, it would have been a worthy cause. But I don't.
There could be multiple unit tests that when executed in parallel would fail the build.

I hope that the above solution would : 1. work 2. buy us enough time before we could find a better solution.

My only other "card" is executing each of the tests in each of the packages separately, but that would result in a substantial increase of test time.

PR #1186, opened 2020-06-20
Expect test for fast catchup

tsachiherman:
## Summary

When switching to catchup mode, the `SetCatchpointCatchupMode` function is being called. The function would get blocked by the node mutex, which might be held by `Stop`. That could lead to a deadlock.

To resolve the above, we return a channel from `SetCatchpointCatchupMode` instead. The operation is carried by a goroutine, which at the ned of operation write the new node context to the channel. This allows the caller to select between the reception of the new context and the expiration of the service.

Second issue which is addressed in this PR is the timeout context handling. The existing context cancelation was comparing the returned propagated error with the context error `err == cs.ctx.Err()`. This would have worked correctly if the context errors would have been retained across all the handlers. However, the http client seems to wrap the underlying context error with it's own when aborting due to context expiration/cancelation. This means that in order to correctly detect context expiration, we need to explicitly test `cs.ctx.Err() != nil`.

## Test Plan

The provided expect test cover the above use case, and provides a good basis for testing the catchpoint catchup functionality.

## Reviewer Notes
Much of the files that were changed in this PR are due to splitting of the `StartNetwork` into `CreateNetwork` and `StartNetwork`. Doing this allows the test to modify the configuration files and/or consensus protocol of the network.


node/node.go
```diff
@@ -876,63 +877,83 @@ func (node *AlgorandFullNode) AbortCatchup(catchpoint string) error {
...
+	// one.
+	ctxCh := make(chan context.Context)
+	outCtxCh = ctxCh
+	go func() {
+		node.mu.Lock()
+		// check that the node wasn't canceled. If it have been canceled, it means that the node.Stop() was called, in which case
+		// we should close the channel.
+		if node.ctx.Err() == context.Canceled {
+			close(ctxCh)
+			return
```
algonautshant, line 896: node.mu is not released here?

tsachiherman: you're right! good catch.

test/e2e-go/cli/goal/expect/corsTest.exp
```diff
@@ -18,6 +18,9 @@ if { [catch {
     set NETWORK_TEMPLATE "$TEST_DATA_DIR/nettemplates/TwoNodes50Each.json"
 
     # Create network
+    ::AlgorandGoal::CreateNetwork $NETWORK_NAME $NETWORK_TEMPLATE $TEST_ALGO_DIR $TEST_ROOT_DIR
```
egieseke, line 21: Did you separate the creation of the network from the starting of the network, to make it easier to start and restart the same network?  Or was there another reason?

test/e2e-go/cli/goal/expect/goalExpectCommon.exp
```diff
@@ -85,7 +89,7 @@ proc ::AlgorandGoal::StartNode { TEST_ALGO_DIR {SYSTEMD_MANAGED ""} } {
 # Stop the node
 proc ::AlgorandGoal::StopNode { TEST_ALGO_DIR {SYSTEMD_MANAGED ""} } {
     set ::GLOBAL_TEST_ALGO_DIR $TEST_ALGO_DIR
-    set timeout 15
+    set timeout 5
```
egieseke, line 92: Reducing the timeout here may result in sporadic test failures on slower platforms.

test/e2e-go/cli/goal/expect/goalExpectCommon.exp
```diff
@@ -149,14 +151,26 @@ proc ::AlgorandGoal::StartNetwork { NETWORK_NAME NETWORK_TEMPLATE TEST_ALGO_DIR
         puts "network create $NETWORK_NAME"
         spawn goal network create --network $NETWORK_NAME --template $NETWORK_TEMPLATE --datadir $TEST_ALGO_DIR --rootdir $TEST_ROOT_DIR
         expect {
-            timeout { close; ::AlgorandGoal::Abort "Timed out creating network with timeout: $timeout" }
-            "^Network $NETWORK_NAME created under.*" { puts "Network $NETWORK_NAME created" ; close  }
-            close
+            timeout { close; ::AlgorandGoal::Abort "Timed out creating network" }
+            "^Network $NETWORK_NAME created under.*" { puts "Network $NETWORK_NAME created" ; exp_continue }
+            eof { ::AlgorandGoal::CheckEOF "Unable to create network" }
```
egieseke, line 156: What is the significance of this statement?

test/e2e-go/cli/goal/expect/goalExpectCommon.exp
```diff
@@ -804,3 +895,14 @@ proc ::AlgorandGoal::TakeAccountOnline { ADDRESS FIRST_ROUND LAST_ROUND TEST_PRI
...
+# CheckEOF checks if there was an error, and aborts if there was an error
+proc ::AlgorandGoal::CheckEOF { { ERROR_STRING "" } } {
+    upvar spawn_id spawn_id
+    catch { wait -i $spawn_id } result;
+    if { [lindex $result 3] != 0 } {
+        puts $ERROR_STRING
+        puts "returned error code is [lindex $result 3]"
+        exit 1
+    }
+}
```
egieseke, line 908: add a final line to script.

test/e2e-go/cli/goal/expect/goalNodeStatusTest.exp
```diff
@@ -18,6 +18,9 @@ if { [catch {
     set NETWORK_TEMPLATE "$TEST_DATA_DIR/nettemplates/TwoNodes50Each.json"
 
     # Create network
+    ::AlgorandGoal::CreateNetwork $NETWORK_NAME $NETWORK_TEMPLATE $TEST_ALGO_DIR $TEST_ROOT_DIR
```
egieseke, line 21: Consider adding CreateAndStartNetwork proc to combine these steps.

test/e2e-go/cli/goal/expect/corsTest.exp
tsachiherman: I explained that in the PR title. I broke that functionality into the goal-oriented functions because it's the only way to create the network first, modify some of it's parameters, and start the nodes on the network in a specific order.

test/e2e-go/cli/goal/expect/goalExpectCommon.exp
tsachiherman: This change is needed in order to verify the return code from the goal command.

test/e2e-go/cli/goal/expect/goalExpectCommon.exp
tsachiherman: sure.

test/e2e-go/cli/goal/expect/goalNodeStatusTest.exp
tsachiherman: I'm not opposing to having a utility function that would combine the two, but I don't feel that the "value" is there.
i.e. given that this is a goal-oriented test, we need to use the tests to invoke goal commands.. if the command structure is too complicated, it implies that our goal API isn't design well.

test/e2e-go/cli/goal/expect/goalExpectCommon.exp
tsachiherman: I'm ok setting it to 15; but please make sure not to extend it beyond 30. If the timeout takes more than 30, it means that the node internally failed to stop and we were forcing the process to start. ( which indicates of some locking issue )

test/e2e-go/cli/goal/expect/catchpointCatchupWebProxy/webproxy.go
```diff
@@ -0,0 +1,85 @@
...
+		time.Sleep(time.Duration(*webProxyRequestDelay) * time.Millisecond)
+		mu.Unlock()
+		// prevent requests for block #2 to go through.
+		if strings.HasSuffix(request.URL.String(), "/block/2") {
+			response.WriteHeader(http.StatusBadRequest)
+			return
+		}
+		if *webProxyLogFile != "" {
+			f, _ := os.OpenFile(*webProxyLogFile, os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0644)
+			f.Write([]byte(fmt.Sprintf("proxy saw request for %s\n", request.URL.String())))
```
algonautshant, line 63: Should this statement be guarded with the mutex?

tsachiherman: It would probably be a good idea, although at this time, it's not critical  - this functionality is not being used.
( I was using it for debugging the expect test, but now that I'm done with it, it's not really "required" ).

I'll do that in a separate PR, since it's a no-op.

algonautshant:
Looks good.

PR #1191, opened 2020-06-22
Add EnableDeveloperAPI configuration flag

tsachiherman:
## Summary

Add a local config to disable teal/compile in algod by default. This PR solves #1109.

The goal is to tie the TealCompile and DryRun in the algod REST endpoint with this new flag.
The DryRun functionality would be added with or after the application main PR, and this PR prepares the ground for that.

## Test Plan

Unit test was updated to use the new construct.
This is a spinoff of the original PR https://github.com/algorand/go-algorand/pull/1163 by Shumo.


tsachiherman:
> LGTM

thanks !

tsachiherman:
@pzbitskiy could you please verify that once you implement the dryrun api it would also be using the same `EnableDeveloperAPI` as above ?

PR #1195, opened 2020-06-23
Test archival

wjurayj-algo:
## Summary

Added some logging to ledger/archival_test.go to explore memory overload problem in Travis runtime.

## Test Plan

n/a, this PR should not be approved.


Makefile
```diff
@@ -200,10 +200,12 @@ fulltest: build-race
 		go test $(GOTAGS) -timeout 2500s -race $$PACKAGE_DIRECTORY; \
 	done
 
-shorttest: build-race $(addprefix short_test_target_, $(UNIT_TEST_SOURCES))
+shorttest: build-race 
+	go test $(GOTAGS) -timeout 2500s -race go-algorand/ledger
```
tsachiherman, line 204: maybe `github.com/algorand/go-algorand/ledger` would work better here ?

wjurayj-algo: You're right--I think the former is better, though, because it doesn't presume the make is being run from the go-algorand directory

PR #1202, opened 2020-06-25
ARM build test

bricerisingalgorand:
## Summary

## Test Plan


tsachiherman:
If you no longer need this one, could you close it ?

PR #1203, opened 2020-06-25
Refactor account update locking

tsachiherman:
## Summary

This change refactors the locking we've had on the `accountUpdates` to allow more structured locking between the ledger tracker's mutex and the `accountUpdates` balances mutex.

From the `Ledger` structure perspective, all the activity on the trackers can now be synchronized uniformly using the tracker's lock. If it's a read only operation, than a read lock is enough, otherwise a full write lock is required. This also allows the `reloadLedger` function to take the lock, ensuring proper synchronization when we switch to a new catchpoint.

From the `accountUpdates` perspective, the balances lock is used to synchronize access to the current balances state. In particular, it's needed as we might get updates to the balances internally as a result of the `commitRound` function.
The `loadFromDisk` function used to be an exception, as we couldn't take the lock thought out the function execution : the function was initializing the cache using the evaluator, which was using the parent ledger to call back into the `accountUpdates`. To resolve that, I've passed to the evaluator a "crafted"  `ledgerForEvaluator` called `accountUpdatesLedgerEvaluator`, which allow us to bypass the ledger's tracker lock as well as `accountUpdates` internal locks.

With that change, the `accountUpdates.loadFromDisk` no longer depends on the `Ledger`'s tracker lock, and allow the `loadFromDisk` to take the internal lock once. This simplify the reasoning of "what-does-the-lock-protect" much better and provides a better locking modularity.

## Reviewer notes
`accountUpdates.lookup` split into `Lookup` and `lookupImpl` where the former take the balances lock before passing the call to the latter.
`accountUpdates.getAssetCreatorForRound` split into `GetAssetCreatorForRound ` and `getAssetCreatorForRoundImpl` where the former take the balances lock before passing the call to the latter.
`accountUpdates.totals` split into `Totals` and `totalsImpl` where the former take the balances lock before passing the call to the latter.
`accountUpdates.newBlock` split into `newBlock` and `newBlockImpl` where the former take the balances lock before passing the call to the latter. ( I avoided making it a capital case as it's an interface function. We can refactor that separately )

## Tests
Existing tests already cover these use cases, and these tests were updated to work with the corresponding changes.
A load test was added for the lodger reload, to ensure we correctly taking locks.


ledger/acctupdates.go
```diff
@@ -246,51 +248,16 @@ func (au *accountUpdates) close() {
...
-		}
-	}
-
-	// No updates of this account in the in-memory deltas; use on-disk DB.
-	// The check in roundOffset() made sure the round is exactly the one
-	// present in the on-disk DB.  As an optimization, we avoid creating
-	// a separate transaction here, and directly use a prepared SQL query
-	// against the database.
-	return au.accountsq.lookup(addr)
+	return au.lookup(rnd, addr, withRewards)
```
algomaxj, line 257: previously we had other callers in the `ledger` package that were calling `lookup` externally. For these refactored, lockless methods, it might be good to change their names like you did with `newBlock` and `newBlockImpl` to make sure we don't accidentally call the wrong version.

pbennett: To add some hopefully helpful commentary, this isn't too different than patterns I've seen in the past like in the ACE C++ library, where xxxx_i methods represented 'internal' methods that assumed a lock was already held by the caller.
This way you knew that a public method, Foo could lock, but (assuming non-reentrant locks) could call no other non _i method.  It could only call _i methods (once obtaining a lock).  _i methods could only call other _i methods.  _i methods could always assume a lock was held.  This could allow more performant non-reentrant locking primitives and avoid deadlocks through a simple but effective stylistic convention.

I think Go's public/private capitalization for this same pattern is a good approach but might break down quickly as it's mixing a valuable, but firm and compiler-enforced visibility pattern with a hint about locking expectations.   Perhaps something similar to the _i)nternal 'hint?'

tsachiherman: Yes, I agree that we should find a better way to separate external and internal functions. What if `accountUpdates` would have been an interface, and internally, there was a struct which implements it.
Any external references to this interface would be limited ( by definition ) to the interface methods. Internally, we could pick whatever naming convention of our choice, knowing that it won't be exposed outside.
( we've already used this design patten in the agreement code somewhere )

ledger/acctupdates.go
```diff
@@ -193,6 +194,10 @@ type accountUpdates struct {
 
 	// accountsWriting provides syncronization around the background writing of account balances.
 	accountsWriting sync.WaitGroup
+
+	// commitSyncerWaitGroup is the blocking channel for syncronizing closing the commitSyncer goroutine. Once it's closed, the
```
algonautshant, line 198: where is commitSyncerWaitGroup?

ledger/acctupdates.go
```diff
@@ -212,11 +217,16 @@ func (au *accountUpdates) initialize(cfg config.Local, dbPathPrefix string, gene
 	if cfg.CatchpointFileHistoryLength < -1 {
 		au.catchpointFileHistoryLength = -1
 	}
+	// initialize the commitSyncerClosed with a closed channel ( since the commitSyncer go-routine is not active )
+	au.commitSyncerClosed = make(chan struct{})
+	close(au.commitSyncerClosed)
 }
 
 // loadFromDisk is the 2nd level initializtion, and is required before the accountUpdates becomes functional
 // The close function is expected to be call in pair with loadFromDisk
```
algonautshant, line 226: What is meant by: "in pair with"? 
should close be called after or before loadFromDisk?

ledger/acctupdates.go
tsachiherman: correct. Fixed.

ledger/acctupdates.go
tsachiherman: "in pair with" means that every call to `loadFromDisk` should be followed ( eventually ) by a `close`, and not, for instance two subsequent calls to `loadFromDisk`

ledger/ledger.go
```diff
@@ -137,14 +137,22 @@ func OpenLedger(
...
+	// blockQ is having a sync goroutine which indirectly calls other trackers. We want to eliminate that go-routine first,
+	// and follow up by taking the trackers lock.
 	if l.blockQ != nil {
 		l.blockQ.close()
 		l.blockQ = nil
 	}
 
-	// reload.
+	// take the trackers lock. This would ensure that no other goroutine is using the trackers.
+	l.trackerMu.Lock()
```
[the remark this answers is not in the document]
tsachiherman: I was thinking about it - the gains in doing this are pretty minor since it's only 5-6 lines of code. unifying the code path here is mostly cosmetic, and I think that proper in-code documentation would be just as beneficial.

ledger/acctupdates.go
```diff
@@ -246,66 +255,36 @@ func (au *accountUpdates) waitAccountsWriting() {
...
 }
 
+// close closes the accountUpdates, waiting for all the child go-routine to complete
 func (au *accountUpdates) close() {
 	if au.ctxCancel != nil {
 		au.ctxCancel()
 	}
 	au.waitAccountsWriting()
+	// this would block until the commitSyncerClosed channel get closed.
+	<-au.commitSyncerClosed
```
algonautshant, line 265: Should this be inside a switch?
In case this statement is waiting before 	

```
	// initialize the commitSyncerClosed with a closed channel ( since the commitSyncer go-routine is not active )
	au.commitSyncerClosed = make(chan struct{})
	close(au.commitSyncerClosed)
```

tsachiherman: > Should this be inside a switch?
You mean "inside a select", right ? Then the answer is "no". The `loadFromDisk` and `close` are expected to be synchronized by the caller, so there is no need to worry for thread safety between them.

algonautshant: Okay. It is also synchronized with `initialize` right?

algonautshant: So, `close` will run in parallel **only** with `commitSyncer` ?

tsachiherman: that's correct.

tsachiherman: The `initialize` is internal function called by `loadFromDisk`

ledger/acctupdates.go
```diff
@@ -246,66 +255,36 @@ func (au *accountUpdates) waitAccountsWriting() {
...
 }
 
-func (au *accountUpdates) lookup(rnd basics.Round, addr basics.Address, withRewards bool) (data basics.AccountData, err error) {
+// Lookup returns the accound data for a given address at a given round. The withRewards indicates whether the
+// rewards should be added to the AccountData before returning. Note that the function doesn't update the account with the rewards,
+// even while it could return the AccoutData which represent the "rewarded" account data.
```
algonautshant, line 270: > Note that the function doesn't update the account with the rewards, 

It is confusing here. Will the returned information be updated? And consequently differ from what is in the account?

tsachiherman: This function does not update the balance when retrieving a account data with rewards. It only does the calculation.

ledger/acctupdates.go
```diff
@@ -655,11 +552,87 @@ func (au *accountUpdates) getCatchpointStream(round basics.Round) (io.ReadCloser
...
+}
+
+// LookupWithoutRewards returns the account balance for a given address at a given round, without the reward
+func (aul *accountUpdatesLedgerEvaluator) LookupWithoutRewards(rnd basics.Round, addr basics.Address) (basics.AccountData, error) {
+	return aul.au.lookupImpl(rnd, addr, false)
+}
+
+// GetCreatorForRound returns the asset/app creator for a given asset/app index at a given round
+func (aul *accountUpdatesLedgerEvaluator) GetCreatorForRound(rnd basics.Round, cidx basics.CreatableIndex, ctype basics.CreatableType) (creator basics.Address, ok bool, err error) {
+	return aul.au.getCreatorForRoundImpl(rnd, cidx, ctype)
```
algonautshant, line 607: no need to acquire the lock here?

tsachiherman: the `accountUpdatesLedgerEvaluator` function are called *only* in the following path:

loadFromDisk-> initializeFromDisk -> <evaluator> -> accountUpdatesLedgerEvaluator  :: GetCreatorForRound/Lookup/Totals

you'll see that in `loadFromDisk` we already took the lock, which is why we want to bypass taking the lock on all the `accountUpdatesLedgerEvaluator` functions.

algonautshant: Now I understand better. TY.
Can we have it getCreatorForRound (not exported?)

tsachiherman: The function name need to match the one on the interface, which has a capital letter.
Note that *all* the accountUpdatesLedgerEvaluator belongs to the interface.

algonautshant: Can't we make the interface lower letter? It looks like all are only accesed from within the package.

tsachiherman: Changing `ledgerForEvaluator` interface, also means that we will need to change the `Ledger` implementation, which is beyond what we're trying to achieve here since it will also force creating private/public interface for all these functions in the Ledger struct itself.

ledger/ledger_test.go
```diff
@@ -887,3 +887,33 @@ func TestLedgerBlockHdrCaching(t *testing.T) {
...
+	defer l.Close()
+
+	blk := genesisInitState.Block
+	for i := 0; i < 128; i++ {
+		blk.BlockHeader.Round++
+		blk.BlockHeader.TimeStamp += int64(crypto.RandUint64() % 100 * 1000)
+		l.AddBlock(blk, agreement.Certificate{})
+
+		if i%7 == 0 {
+			l.reloadLedger()
```
algonautshant, line 909: why the error is not checked here?

tsachiherman: you're right; I should have tested the error code. Could you please review the changes?

ledger/acctupdates.go
```diff
@@ -246,66 +255,36 @@ func (au *accountUpdates) waitAccountsWriting() {
 	au.accountsWriting.Wait()
```
algomaxj, line 255: The error on line 244 is currently ignored, could you please add an `if err != nil` check there?

ledger/acctupdates.go
```diff
@@ -655,11 +552,87 @@ func (au *accountUpdates) getCatchpointStream(round basics.Round) (io.ReadCloser
 
 // functions below this line are all internal functions
 
+// accountUpdatesLedgerEvaluator is a "ledger emulator" which is used *only* by initializeCaches, as a way to shortcut
+// the locks taken by the real ledger object when making requests that are being served by the accountUpdates.
+// Using this struct allow us to take the tracker lock *before* calling the loadFromDisk, and having the operation complete
+// without taking any locks. Note that it's not only the locks performance that is gained : by having the loadFrom disk
+// not requiring any external locks, we can safely take a trackers lock on the ledger during reloadLedger, which ensures
+// that even during catchpoint catchup mode switch, we're still correctly protected by a mutex.
+type accountUpdatesLedgerEvaluator struct {
```
algomaxj, line 561: If I understand correctly, the whole point of this struct is so that `loadFromDisk` can take the `au.accountsMu` lock, right?

ledger/acctupdates.go
tsachiherman: yes! good find.

ledger/acctupdates.go
tsachiherman: That's true, but it extend further than that. The purpose of the struct is to allow bypassing the ledger's tracker mutex. Having the ability to cleanup the au.accountsMu is a bonus. ( which plays nicely ).

When using this struct, we can do the following:
1. make sure that on the Ledger's implementation, all code access to the `accountUpdates` are synchronized via the trackers mutex.
2. make sure that all the external method of  `accountUpdates` are all synchronized via the au.accountsMu ( including loadFromDisk )

#2 is particularly important during reloadLedger, since we want to take the Ledger's trackers mutex and then make changes to the `accountsUpdate` by calling `loadFromDisk`. If `loadFromDisk` requires to take the tracker lock during the eval, it would create a double-locking which won't work with current locking primitives.

PR #1207, opened 2020-06-26
Fix cicd perf test

egieseke:
## Summary

Don't override Channel if already set.  This is to fix the CICD performance test framework which depends on setting the channel name to a value that can be referenced as part of the Elastic Search index.  Also, removed use of BUILDCHANNEL which is redundant with CHANNEL.

## Test Plan

Make clean install test and integration.  I also ran CICD_perf tests.


Makefile
```diff
@@ -13,7 +13,8 @@ S3_RELEASE_BUCKET = $$S3_RELEASE_BUCKET
 BUILDNUMBER      ?= $(shell ./scripts/compute_build_number.sh)
 COMMITHASH       := $(shell ./scripts/compute_build_commit.sh)
 BUILDBRANCH      := $(shell ./scripts/compute_branch.sh)
-BUILDCHANNEL     := $(shell ./scripts/compute_branch_channel.sh $(BUILDBRANCH))
+CHANNEL          ?= $(shell ./scripts/compute_branch_channel.sh $(BUILDBRANCH))
+BUILDCHANNEL     :=  CHANNEL
```
tsachiherman, line 17: maybe I'm missing something, but what is the importance of having the BUILDCHANNEL at all at that point ?
it looks like you could have just used the CHANNEL variable all along, while declaring it as 
```CHANNEL ?= $(shell ./scripts/compute_branch_channel.sh $(BUILDBRANCH))```

egieseke: I considered removing BUILDCHANNEL, however, it is referenced in the scripts.  Cleaning that up is out of the scope of this PR which is focused on fixing the perf tests.

tsachiherman: so, can we use just BUILDCHANNEL instead ? having both BUILDCHANNEL and CHANNEL is confusing.

egieseke: CHANNEL is more widely used.

btoll: My preference is to just do the change now.  By introducing this here, it is bringing it into scope, and this kind of variable assignment is really confusing (especially when there's no context given in the form of an comment).

I just grepped for `BUILDCHANNEL` and it doesn't look at all like it would introduce any risk by doing a simple search and replace.

egieseke: Replaced all references to $BUILDCHANNEL with $CHANNEL.

btoll: Great, thanks

egieseke:
> looks good to me, but would appreciate if you could run few more tests on this one, including creating a betanet/stable releases.

@tsachiherman John ran a rel/beta build and except for unrelated test failure, it succeeded. Is this sufficient? https://github.com/algorand/go-algorand/pull/1213

tsachiherman:
> > looks good to me, but would appreciate if you could run few more tests on this one, including creating a betanet/stable releases.
> 
> @tsachiherman John ran a rel/beta build and except for unrelated test failure, it succeeded. Is this sufficient? #1213

If you have run tests on this one, and verified that it won't generate any regressions, I can't see any reason this shouldn't be merged in.

PR #1209, opened 2020-06-29
Fix random fails in TestCatchupOverGossip

tsachiherman:
## Summary

TestCatchupOverGossip was using an incorrect verification and shown random failures. The expression
```
a.True(nodeStatus.LastRound >= primaryStatus.LastRound)
```
Is incorrect, since the network does move forward atomically. The primary node, even when having no accounts, might end up moving to the next round before the other node, which "told" it how to do so.

### DIFF
--- a/test/e2e-go/features/catchup/basicCatchup_test.go
+++ b/test/e2e-go/features/catchup/basicCatchup_test.go
@@ -153,6 +153,7 @@ func runCatchupOverGossip(t *testing.T,
 	err = fixture.LibGoalFixture.ClientWaitForRoundWithTimeout(lg, waitForRound)
 	a.NoError(err)
 
+	waitStart := time.Now()
 	// wait until the round number on the secondary node matches the round number on the primary node.
 	for {
 		nodeLibGoalClient := fixture.LibGoalFixture.GetLibGoalClientFromDataDir(nc.GetDataDir())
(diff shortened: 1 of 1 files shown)


test/e2e-go/features/catchup/basicCatchup_test.go
```diff
@@ -161,11 +162,16 @@ func runCatchupOverGossip(t *testing.T,
 
 		primaryStatus, err := lg.Status()
 		a.NoError(err)
-		a.True(nodeStatus.LastRound >= primaryStatus.LastRound)
-		if nodeStatus.LastRound == primaryStatus.LastRound && waitForRound < nodeStatus.LastRound {
+		if nodeStatus.LastRound <= primaryStatus.LastRound && waitForRound < nodeStatus.LastRound {
```
algonautshant, line 165: Can't the secondary also advance and stay one round ahead of the primary and this condition never get satisfied?
 nodeStatus.LastRound <= primaryStatus.LastRound
Why not just keep one condition here: waitForRound < nodeStatus.LastRound

tsachiherman: The loop is checking the status every 50ms. We might momentarily be non-synchronized, but for the most of the time, the node's round would be synced. By writing `nodeStatus.LastRound <= primaryStatus.LastRound`, we'll be waiting until this condition is reached.

PR #1210, opened 2020-06-29
Add e2e teal/compile test

tsachiherman:
## Solution

Add an e2e teal/compile endpoint test


libgoal/teal.go
```diff
@@ -0,0 +1,36 @@
...
+)
+
+// Compile compiles the given program and returned the compiled program
+func (c *Client) Compile(program []byte) (compiledProgram []byte, compiledProgramHash crypto.Digest, err error) {
+	algod, err2 := c.ensureAlgodClient()
+	if err2 != nil {
+		return nil, crypto.Digest{}, err2
+	}
+	var hash []byte
+	compiledProgram, hash, err = algod.Compile(program)
```
algorandskiy, line 30: As I understand `compiledProgram` is base64 and `hash` is string hash, so conversion needed here or in the test

test/e2e-go/features/teal/compile_test.go
```diff
@@ -0,0 +1,74 @@
...
+	cfg.SaveToDisk(primaryNode.GetDataDir())
+
+	fixture.Start()
+
+	var hash crypto.Digest
+	compiledProgram, hash, err = libGoalClient.Compile([]byte("int 1"))
+	a.NotNil(compiledProgram)
+	a.NoError(err, "A valid program should result in a compilation success")
+	a.Equal([]byte{0x41, 0x69, 0x41, 0x42, 0x41, 0x53, 0x49, 0x3d}, compiledProgram)
+	a.Equal(crypto.Digest{0x59, 0x4f, 0x45, 0x36, 0x43, 0x32, 0x32, 0x47, 0x48, 0x43, 0x54, 0x4b, 0x41, 0x4e, 0x33, 0x48, 0x55, 0x34, 0x53, 0x45, 0x35, 0x50, 0x47, 0x49, 0x50, 0x4e, 0x35, 0x55, 0x4b, 0x58, 0x41, 0x4a}, hash)
```
algorandskiy, line 68: nit: string representation of the address would be more compact

libgoal/teal.go
algorandskiy: see the handler:
```
	response := generated.PostCompileResponse{
		Hash:   addr.String(),
		Result: base64.StdEncoding.EncodeToString(program),
	}
```

test/e2e-go/features/teal/compile_test.go
```diff
@@ -0,0 +1,74 @@
...
+	cfg.EnableDeveloperAPI = true
+	cfg.SaveToDisk(primaryNode.GetDataDir())
+
+	fixture.Start()
+
+	var hash crypto.Digest
+	compiledProgram, hash, err = libGoalClient.Compile([]byte("int 1"))
+	a.NotNil(compiledProgram)
+	a.NoError(err, "A valid program should result in a compilation success")
+	a.Equal([]byte{0x41, 0x69, 0x41, 0x42, 0x41, 0x53, 0x49, 0x3d}, compiledProgram)
```
algorandskiy, line 67:
```
base64 1.tok | hexdump -C
00000000  41 53 41 42 41 53 49 3d  0a                       |ASABASI=.|
00000009
```
So compiledProgram is a base-64 string

libgoal/teal.go
tsachiherman: I've added the conversion in the restClient.go, since this is where it leaves the "http tier". Both base64 and string formatting are meant for the way to tunnel these over http, so that's the best place to decode these as well.

test/e2e-go/features/teal/compile_test.go
tsachiherman: I'd like to avoid using hash as a string and use it as a crypto.Digest whenever possible. I agree for that comparison, hashes are easier, and I've updated the code to use that.
I also verified the output and correctness by comparing it with the goal clerk compile hash and Tok file.

PR #1216, opened 2020-06-30
Drop legacy ledger storage

tsachiherman:
## Summary

Remove the support of a single database file. At launch( i.e. a year ago ), we were creating two files ( `block.sqlite` and `tracker.sqlite` ), and it's time to drop the support for features that nobody uses.

As a sidenote, the trigger for this was the "database is locked" notification from sqlite when using in-memory databases. For in-memory database, we always used the same filename, which resulted in a locking issues between the block db and tracker db.

# Notes
1. Individuals that still have this one-database setup, would need to reinstall their node after upgrading. While it's not ideal, if these installations would remain as is, they would run into "database is lock" conditions which could negatively affect the node's performance. 
2. We shouldn't have any such installation, and I've verified with PureStake that none of their installation have that configuration.


ledger/ledger.go
```diff
@@ -205,23 +205,20 @@ func openLedgerDB(dbPathPrefix string, dbMem bool) (trackerDBs dbPair, blockDBs
...
 
-	commonDBFilename := dbPathPrefix + ".sqlite"
 	if !dbMem {
+		commonDBFilename := dbPathPrefix + ".sqlite"
 		_, err = os.Stat(commonDBFilename)
+		if !os.IsNotExist(err) {
+			// before launch, we used to have both blocks and tracker
+			// state in a single SQLite db file. We don't have that anymore,
+			// and we want to fail when that's the case.
+			err = fmt.Errorf("A single ledger database file '%s' was detected. This is no longer supported by current binary", commonDBFilename)
```
algorandskiy, line 215: can't it cause problems for existing users? How'd they migrate?

tsachiherman: Yes, it will cause problems. But I'm pretty sure we don't have any such users.
A user that has a single database file need to have created his data directory before March 2019. The first mainnet data directory was created at the beginning of June 2019. I would have developed an upgrade path, but I don't know how to bootstrap such a data directory : we don't have any binary which would produce it.
btw : testnet was created during May 2019, and betanet was created few month after that.

PR #1219, opened 2020-07-02
Automatic DNSBootstrapID for betanet

tsachiherman:
## Solution

This PR adds automatic detection of `betanet` genesis files and update the `DNSBootstrapID` if it has not been set by the user.
The implementation follow the same logic as with `devnet`, and would allow us to deploy `betanet` nodes without requiring the end user to provide a config.json file which sets the `DNSBootstrapID` explicitly.

## Test
I tested this change manually and confirm it's correctness.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -33,6 +33,9 @@ import (
 // Devnet identifies the 'development network' use for development and not generally accessible publicly
 const Devnet protocol.NetworkID = "devnet"
 
+// Betanet identifies the 'beta network' use for early releases of feature to the public prior to releasing these to mainnet/testnet
+const Betanet protocol.NetworkID = "betanet"
+
 // Devtestnet identifies the 'development network for tests' use for running tests against development and not generally accessible publicly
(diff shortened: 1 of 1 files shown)


config/config.go
```diff
@@ -33,6 +33,9 @@ import (
 // Devnet identifies the 'development network' use for development and not generally accessible publicly
 const Devnet protocol.NetworkID = "devnet"
 
+// Betanet identifies the 'beta network' use for early releases of feature to the public prior to releasing these to mainnet/testnet
+const Betanet protocol.NetworkID = "betanet"
+
 // Devtestnet identifies the 'development network for tests' use for running tests against development and not generally accessible publicly
 const Devtestnet protocol.NetworkID = "devtestnet"
```
egieseke, line 40: Is `devtestnet` obsolete?  should it be removed?

tsachiherman: No, as it's being used by multiple unit tests.

PR #1228, opened 2020-07-07
Implement an auction with applications.

derbear:
(This is a draft PR for referential purposes.)

Depends on #917.


jeapostrophe:
You should include a README that points interested parties to https://github.com/derbear/steal/blob/proto-sovauc/apps/sovauc.rkt so they can see how the TEAL was generated. The steal code is much easier to understand, IMHO.

yaovi-a:
@algorandskiy, Max was looking at this and #917 is closed. What's do you think is the next step here?

PR #1229, opened 2020-07-08
Test build packages on multiple architectures

btoll:
## Summary

This makes sure that the test build scripts work on `arm*` architectures.  The main change is to test the `ARCH` variable for some variation of `arm*` in `mule/test/test.sh` and will call the `run_tests` script directly.  This bypasses the path that `amd64` builds take, which is to further spin up more docker containers and run the `run_tests` in each of them.  I feel this is unnecessary for `arm*`, and further the `run_tests` script initiates all the tests that all architectures must run.

The rest (and the majority) of the changes were necessary to allow the `run_tests` shell script to be called directly.  These changes now have each shell script determine the values which it needs rather than having them passed from a Makefile.  This is better practice and is inline with what is being done in the `indexer` project, as well.

## Test Plan

Build a docker image and run the make targets.  Here is an example `Dockerfile`:
```
FROM arm64v8/ubuntu:18.04

ENV DEBIAN_FRONTEND noninteractive
RUN apt-get update && apt-get install curl python python3.7 python3-pip build-essential apt-transport-https ca-certificates software-properties-common autoconf bsdmainutils git -y && \
    curl https://dl.google.com/go/go1.12.linux-arm64.tar.gz | tar xzf - && \
    mv go /usr/local

ENV GOROOT=/usr/local/go GOPATH=$HOME/go
ENV PATH=$GOPATH/bin:$GOROOT/bin:$PATH

WORKDIR /root
COPY . .
CMD ["/bin/bash"]
```


tsachiherman:
be aware that arm32 compiler does not support `-race` flag, which makes it somewhat less effective in finding bugs.

btoll:
> be aware that arm32 compiler does not support `-race` flag, which makes it somewhat less effective in finding bugs.

Good to know, thanks.

PR #1231, opened 2020-07-08
Add CatchpointDump utility

tsachiherman:
## Summary

The catchpointdump utility allows to perform the following:
1. Given a catchpoint file, it knows how to read it's content and dump the output in a human readable format into a separate file.
2. Given a network name and a round number, it would fetch the catchpoint for that round from all the relays of that network, save the catchpoint file, and dump it's content.
3. Given an ledger tracker database, it can dump the database content into a file


cmd/catchpointdump/file.go
```diff
@@ -0,0 +1,225 @@
...
+				err = fmt.Errorf("Account DB address length mismatch: %d != %d", len(addrbuf), len(addr))
+				return
+			}
+			copy(addr[:], addrbuf)
+			jsonData, err := json.Marshal(data)
+			if err != nil {
+				return err
+			}
+
+			fmt.Fprintf(outFile, "%v : %s\n", addr, string(jsonData))
```
algonautshant, line 219: If outFile is blocked, the database is blocked. 
If this tool is used on a production database, and accidentally the output is blocked (e.g. printing to screen with paging), it can be a problem. 
Maybe put a timeout?

cmd/catchpointdump/file.go
```diff
@@ -0,0 +1,225 @@
...
+		if fileHeader.Version != 0 {
+			balancesTable = "catchpointbalances"
+		}
+		rows, err := tx.Query(fmt.Sprintf("SELECT address, data FROM %s order by address", balancesTable))
+		if err != nil {
+			return
+		}
+		defer rows.Close()
+
+		for rows.Next() {
```
algonautshant, line 194: Printing in json format prints too much data: slow, redundant, and difficult to read. 
Why not print a header row, and put only the values in the following rows? 
Is every jsonData of the same structure?

cmd/catchpointdump/database.go
```diff
@@ -0,0 +1,57 @@
...
+	"github.com/spf13/cobra"
+
+	"github.com/algorand/go-algorand/ledger"
+)
+
+var ledgerTrackerFilename string
+
+func init() {
+	databaseCmd.Flags().StringVarP(&ledgerTrackerFilename, "tracker", "t", "", "Specify the ledger tracker file name ( i.e. ./ledger.tracker.sqlite )")
+	databaseCmd.Flags().StringVarP(&outFileName, "output", "o", "", "Specify an outfile for the dump ( i.e. ./ledger.tracker.sqlite )")
```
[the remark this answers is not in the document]
tsachiherman: will change.

cmd/catchpointdump/file.go
```diff
@@ -0,0 +1,225 @@
...
+					err = fmt.Errorf("getPeerLedger received io.EOF while reading from tar file stream prior of reaching chunk size %d / %d", readComplete, header.Size)
+				}
+				return fileHeader, err
+			}
+		}
+		err = catchupAccessor.ProgressStagingBalances(ctx, header.Name, balancesBlockBytes, &downloadProgress)
+		if err != nil {
+			return fileHeader, err
+		}
+		if header.Name == "content.msgpack" {
```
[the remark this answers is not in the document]
tsachiherman: it means that it's one of the blocks.. which is expected and fine to be ignored.

cmd/catchpointdump/file.go
```diff
@@ -0,0 +1,225 @@
...
+		}
+	}
+}
+
+func printAccountsDatabase(databaseName string, fileHeader ledger.CatchpointFileHeader, outFile *os.File) error {
+	dbAccessor, err := db.MakeAccessor(databaseName, true, false)
+	if err != nil || dbAccessor.Handle == nil {
+		return err
+	}
+	if fileHeader.Version != 0 {
```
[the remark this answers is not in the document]
tsachiherman: This field cannot be zero when it comes from a catchpoint file. It's already being validated before. ( inside the catchup accessor )
it's zero when this function is called directly on a database file which isn't associated with any particular catchpoint. It allows us to extract the same information from a ledger file directly.

cmd/catchpointdump/file.go
tsachiherman: no, the jsonData might not be the same for all the rows.
I know that there is a lot more to be desired here, but keep in mind that it was developed as a debugging tool - we wanted to compare two catchpoints and review differences, if any. If it will become an ongoing used tool, we should definitely consider additional format options.

cmd/catchpointdump/file.go
tsachiherman: I don't expect to be using it directly on production servers.. the intention here is to allow offline database to be examined. Also, when dumping a catchpoint, it creates it's own database, so there are no locking contentions.

PR #1234, opened 2020-07-09
Eric/update single go version

egieseke:
## Summary

Updated specification of single go version to use what is specified in the `get_golang_version.sh` script.  This is to avoid an issue with go version `1.14` where the patch version is no longer allowed to be specified in `go.mod` file.  Added a check to make sure that the go version specified in the `get_golang_version.sh` script is compatible with the version specified `go.mod`, where the major and minor versions should match, and also patch version if specified in the `go.mod` file. 

This pull request closes https://github.com/algorand/go-algorand-internal/issues/650

## Test Plan

ran make clean install test and integration.


scripts/get_golang_version.sh
```diff
@@ -1,2 +1,16 @@
...
+# Required GOLANG Version for project specified here
+GOLANG_VERSION="1.12.17"
+
+# Check to make sure that it matches what is specified in go.mod
+
+GOMOD_VERSION=$(awk '/^go[ \t]+[0-9]+\.[0-9]+(\.[0-9]+)?[ \t]*$/{print $2}' go.mod)
+TRIMED_GOLANG_VERSION=$(echo ${GOLANG_VERSION} |awk -F. '{ print $1 "." $2 }')
+
+if ! [[ ${GOLANG_VERSION} == ${GOMOD_VERSION} || ${TRIMED_GOLANG_VERSION} == ${GOMOD_VERSION} ]]
+then
```
btoll, line 11: You could simplify lines 8 - 10 by doing something like this:
```
if ! [[ "$GOLANG_VERSION" =~ "$GOMOD_VERSION" ]]
...
```
Here are some simple tests:
```
~:$ [[ 1.12 =~ 1.12 ]]
~:$ echo $?
0
~:$ [[ 1.12.17 =~ 1.12 ]]
~:$ echo $?
0
~:$ [[ 1.12 =~ 1.13 ]]
~:$ echo $?
1
```

btoll: Also, `shellcheck` :)

egieseke: Quoted strings on the right side of the equality check to satisfy spellcheck.
Using a regular expression comparison does not work in all cases, for example:
`if [[ 123456 =~ 1.3 ]] ; then echo true; else echo false; fi
true`

btoll: I'm afraid I don't see the relevance of that example.  The golang version string would never look like that.

My reading of this is to match 1.12 == 1.12 or a trimmed version of 1.12.17.  If that is the case, the example I used above will work quite well.  It saves two processes and having to add pipefail as a shell option.

Also, please shellcheck the other shell scripts in the PR.

brianolson: I like the test as it is. It's explicit and precise.

btoll: Or even this using partial string match:
```
if ! [[ "$GOLANG_VERSION" = "$GOMOD_VERSION"* ]]
...

Simple and readable.

```
~:$ [[ 1.12.17 = 1.12* ]]
~:$ echo $?
0
~:$ [[ 1.12.17 = 1.12.17* ]]
~:$ echo $?
0
~:$ [[ 1.12.17 = 1.12.18* ]]
~:$ echo $?
1
~:$ [[ 1.12.17 = 1.13* ]]
~:$ echo $?
1
```

Just suggestions, do what you want :)

Comment (egieseke) [reply scripts/get_golang_version.sh:11]: Leaving the version check as is since as Brian points out, it is explicit and precise.  Cleaned up shellcheck warnings.

File: scripts/get_golang_version.sh [hunk 2]
```diff
@@ -1,2 +1,16 @@
 #!/usr/bin/env bash
-awk '/^go[ \t]+[0-9]+\.[0-9]+(\.[0-9]+)?[ \t]*$/{print $2}' go.mod
+# Required GOLANG Version for project specified here
+GOLANG_VERSION="1.12.17"
+
+# Check to make sure that it matches what is specified in go.mod
+
+GOMOD_VERSION=$(awk '/^go[ \t]+[0-9]+\.[0-9]+(\.[0-9]+)?[ \t]*$/{print $2}' go.mod)
+TRIMED_GOLANG_VERSION=$(echo ${GOLANG_VERSION} |awk -F. '{ print $1 "." $2 }')
```
btoll, line 8: You have a typo here `TRIMED` -> `TRIMMED`.

Once that's fixed and the merge conflict resolved I can approve it.

egieseke: Good catch, updated.

btoll: Thanks

PR #1242, opened 2020-07-13
Ledger db split

wjurayj-algo:
## Summary

Updated in-memory ledger DB to exist as two databases (block and tracker) rather than a single db to prevent memory overload caused by failed DB access due to already held locks.

## Test Plan

Add 10k blocks to the ledger, so that without the fix there will be many concurrent writes to the same DB, often causing a memory overload Segfault or failure to allocate memory in the Travis build.


ledger/ledger_test.go
```diff
@@ -862,3 +866,48 @@ func testLedgerRegressionFaultyLeaseFirstValidCheck2f3880f7(t *testing.T, versio
...
+
+func TestLedgerDBConcurrentAccess(t *testing.T) {
+	// This test ensures that trackers return the correct value from
+	// committedUpTo() -- that is, if they return round rnd, then they
+	// do not ask for any round before rnd on a subsequent call to
+	// loadFromDisk().
+	//
+	// We generate mostly empty blocks, with the exception of timestamps,
+	// which affect participationTracker.committedUpTo()'s return value.
+
```
tsachiherman, line 878: The comment is no longer relevant.

ledger/ledger_test.go
```diff
@@ -862,3 +866,48 @@ func testLedgerRegressionFaultyLeaseFirstValidCheck2f3880f7(t *testing.T, versio
 		a.NoError(l.appendUnvalidatedTx(t, initAccounts, initSecrets, correctPayLease, ad), "should allow leasing payment transaction with newer FirstValid")
 	}
 }
+
+func TestLedgerDBConcurrentAccess(t *testing.T) {
```
tsachiherman, line 870: please describe what this test is intend to test.

ledger/ledger_test.go
```diff
@@ -862,3 +866,48 @@ func testLedgerRegressionFaultyLeaseFirstValidCheck2f3880f7(t *testing.T, versio
...
+	wl := &wrappedLedger{
+		l: l,
+	}
+
+	blk := genesisInitState.Block
+
+	for i := 0; i < 10000; i++ {
+		blk.BlockHeader.Round++
+		blk.BlockHeader.TimeStamp += int64(crypto.RandUint64() % 100 * 1000)
+		wl.l.AddBlock(blk, agreement.Certificate{})
```
tsachiherman, line 897: I think that if you'll increase the block size, it would dramatically increase the likelihood for db contention.
You can do the by filling the `Payset` field with 5-10 thousand dummy transactions.
Make sure to create the transactions outside of the loop so that you can "push" them into the "AddBlock" quickly.

ledger/ledger_test.go
```diff
@@ -862,3 +866,48 @@ func testLedgerRegressionFaultyLeaseFirstValidCheck2f3880f7(t *testing.T, versio
...
+
+		// Don't bother checking the trackers every round -- it's too slow...
+		if crypto.RandUint64()%23 > 0 {
+			continue
+		}
```
tsachiherman, line 902: I think that you don't need this clause.

ledger/ledger_test.go
```diff
@@ -18,7 +18,11 @@ package ledger
 
 import (
 	"context"
+	"database/sql"
+//	"io/ioutil"
 	"fmt"
+//	"os"
+//	"path/filepath"
```
tsachiherman, line 25: It's ok that you had some trials here, but clean up commented out entries before you're done.

ledger/ledger_test.go
wjurayj-algo: Working on this, but my current version has a timeout caused by indefinite blocking.

PR #1243, opened 2020-07-14
Implement database versioning support

tsachiherman:
## Summary
This PR implements the core functionality for database versioning support.
In this PR, I haven't used it in any particular database. It's part of the support for reencoding the accounts.

- [x] Implement database versioning support
- [ ]  Extend Atomic callback to support context and long running operations
- [ ]  Implement reencoding of accounts

### DIFF
--- a/util/db/versioning.go
+++ b/util/db/versioning.go
@@ -44,23 +44,31 @@ func SetUserVersion(ctx context.Context, tx *sql.Tx, userVersion int32) (previou
 	if err != nil {
 		return
 	}
+	if previousUserVersion == userVersion {
+		return
+	}
 	var result sql.Result
 	var rowsAffected int64
 	result, err = tx.ExecContext(ctx, fmt.Sprintf("PRAGMA user_version = %d", userVersion))
 	if err != nil {
-		if err == ctx.Err() {
-			// if we're aborting due to context cancelation, just clear out the previous version.
-			// this is done so that the function result is consistant.
-			previousUserVersion = 0
-		}
+		// if we're aborting due to an error, clear the previousUserVersion so that
+		// on all error cases we'll be returning zero.
+		previousUserVersion = 0
 		return
 	}
 	rowsAffected, err = result.RowsAffected()
 	if err != nil {
+		// if we're aborting due to an error, clear the previousUserVersion so that
+		// on all error cases we'll be returning zero.
+		previousUserVersion = 0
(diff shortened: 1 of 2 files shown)


util/db/versioning.go
```diff
@@ -0,0 +1,66 @@
...
+		return
+	}
+	rowsAffected, err = result.RowsAffected()
+	if err != nil {
+		return
+	}
+	if rowsAffected != 0 {
+		err = fmt.Errorf("expected rows to be affected by updating user_version was 0, but %d rows got updated instead", rowsAffected)
+	}
+	return
```
algorandskiy, line 65: why is it here?

util/db/versioning.go
```diff
@@ -0,0 +1,66 @@
...
+		return
+	}
+	var result sql.Result
+	var rowsAffected int64
+	result, err = tx.ExecContext(ctx, fmt.Sprintf("PRAGMA user_version = %d", userVersion))
+	if err != nil {
+		if err == ctx.Err() {
+			// if we're aborting due to context cancelation, just clear out the previous version.
+			// this is done so that the function result is consistant.
+			previousUserVersion = 0
```
algorandskiy, line 54: shouldn't we clear `previousUserVersion` for all errors not only context error? Like it if failed for whatever reason return 0, err

tsachiherman: sounds like a good idea.

util/db/versioning.go
tsachiherman: go likes a return statement at the end of a function.

algorandskiy: not the return statement but the entire block of checking `RowsAffected`

tsachiherman: Do you feel that testing the rowsAffected is not required as long as we've verified the function doesn't error ?
I'm not overly particular about it, but was trying to ensure that I'm testing all the return values to avoid corner cases slip through.

algorandskiy: I mean in [the doc](https://www.sqlite.org/pragma.html#pragma_user_version):
```The user_version pragma will to get or set the value of the user-version integer at offset 60 in the database header. The user-version is an integer that is available to applications to use however they want. SQLite makes no use of the user-version itself.
```
Sounds like it does not performs any table modifications so there is no reason to look into `sqlite3_changes` or `RowsAffected`

PR #1246, opened 2020-07-15
Avoid warning messages on expected long transactions

tsachiherman:
## Summary
This PR extends the database accessor functionality by allowing long running transactions to extend their "legit" execution runtime. This would allow the database logging logic to suppress undesired warnings caused by long running transactions.

This PR is part of the reencoding accounts data effort.

- [x] Implement database versioning support (https://github.com/algorand/go-algorand/commit/88bc89a53c99e3ddcb7e6953d1358d1cfa3c266b)
- [x]  Extend Atomic callback to support context and long running operations (https://github.com/algorand/go-algorand/pull/1246)
- [ ]  Implement reencoding of accounts (https://github.com/algorand/go-algorand/pull/1251)

### DIFF
--- a/util/db/dbutil.go
+++ b/util/db/dbutil.go
@@ -60,7 +60,10 @@ type Accessor struct {
 	log      logging.Logger
 }
 
-type txContext struct {
+// txExecutionContext contains the data that is associated with every created transaction
+// before sending it to the user-defined callback. This allows the callback function to
+// make changes to the execution setting of an ongoing transaction.
+type txExecutionContext struct {
 	deadline time.Time
 }
 
@@ -301,7 +304,7 @@ func (db *Accessor) atomic(fn idemFn, commitLocker sync.Locker, extras ...interf
 		}
 
 		// create a transaction context data
-		txContextData := &txContext{
+		txContextData := &txExecutionContext{
 			deadline: atomicDeadline,
 		}
 
@@ -343,13 +346,15 @@ func (db *Accessor) atomic(fn idemFn, commitLocker sync.Locker, extras ...interf
 }
 
 // ResetTransactionWarnDeadline allow the atomic function to extend it's warn deadline by setting a new deadline.
-// the ctx and tx parameters were added to support recursive atomic calls.
+// The Accessor can be copied and therefore isn't suitable for multi-threading directly,
+// however, the transaction context and transaction object can be used to uniquely associate the request
+// with a particular deadline.
 // the function fails if the given transaction is not on the stack of the provided context.
 func ResetTransactionWarnDeadline(ctx context.Context, tx *sql.Tx, deadline time.Time) (prevDeadline time.Time, err error) {
-	txContextData, ok := ctx.Value(tx).(*txContext)
+	txContextData, ok := ctx.Value(tx).(*txExecutionContext)
 	if !ok {
 		// it's not a valid call. just return an error.
-		return time.Time{}, fmt.Errorf("the provided tx does not have a valid txContext object in it's context")
+		return time.Time{}, fmt.Errorf("the provided tx does not have a valid txExecutionContext object in it's context")
 	}
 	prevDeadline = txContextData.deadline
 	txContextData.deadline = deadline
--- a/util/db/dbutil.go
+++ b/util/db/dbutil.go
@@ -60,6 +60,10 @@ type Accessor struct {
 	log      logging.Logger
 }
 
+type txContext struct {
+	deadline time.Time
+}
+
 // MakeAccessor creates a new Accessor.
 func MakeAccessor(dbfilename string, readOnly bool, inMemory bool) (Accessor, error) {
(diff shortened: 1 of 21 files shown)


util/db/dbutil.go
```diff
@@ -202,10 +209,10 @@ func (db *Accessor) Atomic(fn idemFn, extras ...interface{}) (err error) {
 // Atomic executes a piece of code with respect to the database atomically.
 // For transactions where readOnly is false, sync determines whether or not to wait for the result.
 func (db *Accessor) atomic(fn idemFn, commitLocker sync.Locker, extras ...interface{}) (err error) {
-	start := time.Now()
+	atomicDeadline := time.Now().Add(time.Second)
```
algonautshant, line 212: Why does the clock starts ticking from here and not at the guardedFn call?

util/db/dbutil.go
```diff
@@ -296,7 +303,12 @@ func (db *Accessor) atomic(fn idemFn, commitLocker sync.Locker, extras ...interf
 			break
 		}
 
-		err = guardedFn(tx)
+		// create a transaction context data
+		txContextData := &txExecutionContext{
+			deadline: atomicDeadline,
```
algonautshant, line 308: It seems to me that the two variables: atomicDeadline and txContextData.deadline are serving the same purpose.
This is confusing me. 
Why not initialize txContextData at the top, where atomicDeadline is set, and not use the atomicDeadline variable?
This way, atomicDeadline does not need to be updated with txContextData.deadline.

util/db/dbutil.go
tsachiherman: The default clock is when the entire atomic function starts.
We want this to be the default, unless we're going to start being more picky around "how long does it take to establish a database connection?", etc.

util/db/dbutil.go
tsachiherman: The intention here was to avoid a failed transactions from modifying the execution duration warning deadline. 
It's for the last ( or successful ) transaction to update it, as it sees fit.
If the transaction fail, we will have an error message printed out, so that the warning is not an issue anyway.

util/db/dbutil.go
```diff
@@ -325,13 +339,25 @@ func (db *Accessor) atomic(fn idemFn, commitLocker sync.Locker, extras ...interf
...
+	}
+	return
+}
+
+// ResetTransactionWarnDeadline allow the atomic function to extend it's warn deadline by setting a new deadline.
+// The Accessor can be copied and therefore isn't suitable for multi-threading directly,
+// however, the transaction context and transaction object can be used to uniquely associate the request
+// with a particular deadline.
+// the function fails if the given transaction is not on the stack of the provided context.
+func ResetTransactionWarnDeadline(ctx context.Context, tx *sql.Tx, deadline time.Time) (prevDeadline time.Time, err error) {
```
algorandskiy, line 353: is it supposed to be used in subsequent PR?

util/db/dbutil.go
```diff
@@ -325,13 +339,25 @@ func (db *Accessor) atomic(fn idemFn, commitLocker sync.Locker, extras ...interf
 		commitLocker.Unlock()
 	}
 
-	end := time.Now()
-	delta := end.Sub(start)
-	if delta > time.Second {
-		db.getDecoratedLogger(fn, extras).Warnf("dbatomic: tx took %v", delta)
-	} else if delta > time.Millisecond {
```
algorandskiy, line 332: Removed because?...

util/db/dbutil_test.go
```diff
@@ -333,3 +335,41 @@ func TestDBConcurrencyRW(t *testing.T) {
...
+func TestResettingTransactionWarnDeadline(t *testing.T) {
+	acc, err := MakeAccessor("fn.db", false, true)
+	require.NoError(t, err)
+	defer acc.Close()
+	logger := WarningLogCounter{
+		Logger: logging.Base(),
+	}
+	acc.log = &logger
+	err = acc.Atomic(func(ctx context.Context, tx *sql.Tx) error {
+		time.Sleep(1001 * time.Millisecond)
```
algorandskiy, line 362: I'd prefer seeing timeouts configurable, just not to waste test running time. but we can leave with 2 seconds if you add `t.Parallel()`

util/db/dbutil.go
tsachiherman: Yes. it will be in the subsequent draft pr..

util/db/dbutil.go
tsachiherman: The printing in debug wasn't really useful. It was just spitting it in *some* unit tests, and we generally ignored it completely. As some of the queries getting longer than 1 millisecond consistently, we always were outputting this, which again, became pointless.
I'm hoping that with this change in place, we'll be able to craft more precise warn timeouts.

util/db/dbutil_test.go
tsachiherman: I'd rather avoid having it configurable in this change, but agree that we should add that in the future.
I have updated this one to use Parallel so that we can run both of the tests at the same time.

PR #1250, opened 2020-07-15
REST API: Merge ApplicationLocalStates and ApplicationLocalState.

winder:
There was some unnecessary encapsulation in the REST models for ApplicationLocalState/ApplicationLocalStates


winder:
> Changes looks good, not sure about updating gomod/gosum, so we have a policy?

I'm trying to revert these, in the past there have been issues. :crossed_fingers:

PR #1252, opened 2020-07-15
Fix tealdbg CDT url

algorandskiy:
## Summary

Newer versions of Google Chrome dropped chrome-devtools:// schema and use devtools:// instead

## Test Plan

Tested manually


tsachiherman:
The change itself looks ok, but i'm concerned about the solution logativity. Would it work with the chrome release that is planned for next Monday ? What about the one that comes after ?

algorandskiy:
I do not think this is problem, `chrome-devtools://` was around for 3-5 years.

PR #1255, opened 2020-07-16
Unattended deb upgrades fail because `origin` value is mixed case

btoll:
##  Summary

Currently, there is a mixed-case discrepancy among the shell scripts that create the deb package.  This will lower-case all areas where `algorand` is given for the package name or instance which references it.

## Test Plan

n/a

### DIFF
--- a/installer/51algorand-upgrades
+++ b/installer/51algorand-upgrades
@@ -2,7 +2,7 @@
 ## unattended upgrades for the Algorand node software.
 
 Unattended-Upgrade::Allowed-Origins {
-  "Algorand:@CHANNEL@";
+  "algorand:@CHANNEL@";
 };
 
 Dpkg::Options {
--- a/installer/rpm/algorand.repo
+++ b/installer/rpm/algorand.repo
@@ -1,5 +1,5 @@
 [algorand]
-name=Algorand
+name=algorand
 baseurl=https://releases.algorand.com/rpm/stable/
 enabled=1
 gpgcheck=1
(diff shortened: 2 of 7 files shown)


brianolson:
Won't this will break all currently deployed autoupdaters? It was currently deployed as "Algorand", so pushing a change will break the update, right?

btoll:
> Won't this will break all currently deployed autoupdaters? It was currently deployed as "Algorand", so pushing a change will break the update, right?

It's a good question.  When we temporarily removed and then restored the 2.0.7 and 2.0.8 releases, I had re-published rather than switching out snapshots.  When I did so, I lower-cased the `origin` field when publishing, which unfortunately broke the unattended upgrade.

This PR lower-cases everything to be consistent.

tsachiherman:
> > Won't this will break all currently deployed autoupdaters? It was currently deployed as "Algorand", so pushing a change will break the update, right?
> 
> It's a good question. When we temporarily removed and then restored the 2.0.7 and 2.0.8 releases, I had re-published rather than switching out snapshots. When I did so, I lower-cased the `origin` field when publishing, which unfortunately broke the unattended upgrade.
> 
> This PR lower-cases everything to be consistent.

Regretfully, the switching of 2.0.7/2.0.8 only increases the permutation matrix here. Is there any way to determine which users are expected to see an issue ? Is there any way we could draw a timeline here ?
The reason I'm saying that is that I'm trying to understand if we want to fix it in the "right way" and aligning it all with "lower case", keeping it "mixed case" to reduce end user issues or to change policy here ( i.e. would require some public communication as well )

brianolson:
It should have never changed. How many auto-updaters out there right now never updated past that change? (and still wouldn't change because they're still expecting "Algorand")

brianolson:
I have a t3.small Ubuntu leaf node to test unattended upgrade and I just checked its logs and it started at 2.0.2 and did every upgrade through 2.0.8 (I have a similar Centos node which is at 2.0.8 but I didn't check its logs). So, now I don't know what was actually deployed at what time in the past, but I still don't think we should make a possibly breaking change unless absolutely necessary.

btoll:
I'm off this week and will fix this when I get back on Monday.

btoll:
After chatting more with @brianolson, I'm going to close this and republish with the "Algorand" string.

PR #1257, opened 2020-07-17
Fetch balance from indexer for tealdbg

nicholasguoalgorand:
## Summary

Adding indexer support for the tealdbg. You can now pass the `-i` flag along with an indexer api endpoint to fetch balances from a specified round number. The debugger will only populate balances with indexer results if the balance is not specified to the debugger

## Test Plan

Manual testing, added unit test that mocks the indexer


cmd/tealdbg/localLedger.go
```diff
@@ -53,6 +69,25 @@ func makeAppLedger(
 	apps := []basics.AppIndex{appIdx}
 	apps = append(apps, txn.Txn.ForeignApps...)
 
+	// first populate balances from the indexer
```
derbear, line 72: Does this break existing behavior if `indexerURL` is not provided?

cmd/tealdbg/localLedger.go
```diff
@@ -53,6 +69,25 @@ func makeAppLedger(
...
+			fmt.Println(queryString)
+			if resp, err := http.Get(queryString); err == nil {
+				if responseData, err := ioutil.ReadAll(resp.Body); err == nil {
+					var accountResp AccountResponse
+					json.Unmarshal(responseData, &accountResp)
+					balances[acc], err = v2.AccountToAccountData(&accountResp.Account)
+					fmt.Println(balances[acc])
+				}
+			} else {
+				fmt.Println(err)
```
derbear, line 86: I know this is for debugging purposes, but you'll eventually want to return the error so your caller can handle it.
```
resp, err := expr
if err != nil {
  return err
}
etc
```
is one pattern that's used a lot.

cmd/tealdbg/localLedger.go
```diff
@@ -53,6 +69,25 @@ func makeAppLedger(
...
 	apps = append(apps, txn.Txn.ForeignApps...)
 
+	// first populate balances from the indexer
+	for _, acc := range accounts {
+		fmt.Println("debug hello")
+		if _, ok := balances[acc]; !ok {
+			queryString := fmt.Sprintf("%s/v2/accounts/%s?round=%v", indexerURL, acc, round)
+			fmt.Println(queryString)
+			if resp, err := http.Get(queryString); err == nil {
+				if responseData, err := ioutil.ReadAll(resp.Body); err == nil {
```
derbear, line 79: Make sure to handle the error here. 

Also consider using something other than `ioutil.ReadAll`, which will try to process huge responses. If someone mistypes a URL, limiting the amount of response data read over the network will prevent `tealdbg` from trying to allocate tons of memory and then crashing.

cmd/tealdbg/localLedger.go
```diff
@@ -53,6 +69,25 @@ func makeAppLedger(
...
+					var accountResp AccountResponse
+					json.Unmarshal(responseData, &accountResp)
+					balances[acc], err = v2.AccountToAccountData(&accountResp.Account)
```
derbear, line 82: Make sure to handle the error here.

cmd/tealdbg/localLedger.go
```diff
@@ -53,6 +69,25 @@ func makeAppLedger(
...
+	// first populate balances from the indexer
+	for _, acc := range accounts {
+		fmt.Println("debug hello")
+		if _, ok := balances[acc]; !ok {
+			queryString := fmt.Sprintf("%s/v2/accounts/%s?round=%v", indexerURL, acc, round)
+			fmt.Println(queryString)
+			if resp, err := http.Get(queryString); err == nil {
+				if responseData, err := ioutil.ReadAll(resp.Body); err == nil {
+					var accountResp AccountResponse
+					json.Unmarshal(responseData, &accountResp)
```
derbear, line 81: This function returns an error--make sure to handle it.

derbear:
Eventually, when the PR is ready, you'll want to edit its description.

cmd/tealdbg/localLedger.go
```diff
@@ -53,6 +69,25 @@ func makeAppLedger(
 	apps := []basics.AppIndex{appIdx}
 	apps = append(apps, txn.Txn.ForeignApps...)
 
+	// first populate balances from the indexer
+	for _, acc := range accounts {
+		fmt.Println("debug hello")
+		if _, ok := balances[acc]; !ok {
+			queryString := fmt.Sprintf("%s/v2/accounts/%s?round=%v", indexerURL, acc, round)
```
derbear, line 76: Consider using `path.Join` here (from https://golang.org/pkg/path/)

cmd/tealdbg/localLedger.go
nicholasguoalgorand: should not be executed at all, added support for that

cmd/tealdbg/localLedger.go
nicholasguoalgorand: can path.join handle the `?round=x` part?

cmd/tealdbg/localLedger.go
```diff
@@ -17,17 +17,33 @@
...
 	"github.com/algorand/go-algorand/config"
 	"github.com/algorand/go-algorand/crypto"
+	"github.com/algorand/go-algorand/daemon/algod/api/server/v2/generated"
 	"github.com/algorand/go-algorand/data/basics"
 	"github.com/algorand/go-algorand/data/transactions"
 	"github.com/algorand/go-algorand/data/transactions/logic"
 	"github.com/algorand/go-algorand/ledger"
 )
 
+type AccountResponse struct {
```
algorandskiy, line 36: suggest AccountIndexerResponse name to make it clear - AccountResponse is a thing for algod response

cmd/tealdbg/localLedger.go
algorandskiy, line 76: this can lead to confusion. If I specified accounts and some are fetched from indexer, some from command like args. I guess logging an info message if accounts are non-empty and we go to indexer, and no message if only indexer used

nicholasguoalgorand: Accounts from command args have priority over the indexer? So if accounts are non-empty we will not use the indexer.

cmd/tealdbg/localLedger.go
```diff
@@ -130,6 +161,26 @@ func makeAppLedger(
 	return ledger, states, err
 }
 
+func getBalanceFromIndexer(indexerURL string, account basics.Address, round int) (basics.AccountData, error){
+	queryString := fmt.Sprintf("%s/v2/accounts/%s?round=%v", indexerURL, account, round) // path.Join(indexerURL, "v2", "accounts", fmt.Sprintf("%s?round=%v", acc, round))
```
tsachiherman, line 165: round is an integer, would you mind using `%d` instead ?

cmd/tealdbg/localLedger.go
```diff
@@ -130,6 +161,26 @@ func makeAppLedger(
...
+	fmt.Println(queryString)
+	resp, err := http.Get(queryString)
+	if err != nil {
+		return basics.AccountData{}, err
+	}
+	var accountResp AccountIndexerResponse
```
tsachiherman, line 171: you need to check that the response code is 200. if it's not, there is no point in going forward.

cmd/tealdbg/localLedger.go
```diff
@@ -130,6 +161,26 @@ func makeAppLedger(
...
+	err = json.NewDecoder(resp.Body).Decode(&accountResp)
```
tsachiherman, line 172: the resp.Body need to be closed after you're done reading from it.

cmd/tealdbg/localLedger.go
```diff
@@ -130,6 +162,28 @@ func makeAppLedger(
...
 }
 
+func getBalanceFromIndexer(indexerURL string, account basics.Address, round int) (basics.AccountData, error){
+	queryString := fmt.Sprintf("%s/v2/accounts/%s?round=%d", indexerURL, account, round)
+	resp, err := http.Get(queryString)
+	if err != nil {
+		return basics.AccountData{}, err
+	}
+	if resp.StatusCode != 200 {
+		return basics.AccountData{}, fmt.Errorf("indexer error response status code %d", resp.StatusCode)
```
tsachiherman, line 172: If StatusCode is, let's say, 400, the resp.Body would not get closed..
The "classic" go design patten here is to call
`defer resp.Body.Close()` right after checking the error code from "Get".

algorandskiy: Plus, we want to see error text from the indexer if any

nicholasguoalgorand: Error text meaning the response right? If so, what would be the best way to fetch that? Would Ioutil.ReadAll() be a good method? This might not be the best approach if the endpoint returned a high amount of data?

algorandskiy: `Ioutil.ReadAll()` looks OK.

cmd/tealdbg/main.go
```diff
@@ -156,6 +157,7 @@ func init() {
 	debugCmd.Flags().StringVarP(&balanceFile, "balance", "b", "", "Balance records to evaluate stateful TEAL on in form of json or msgpack file")
 	debugCmd.Flags().StringVarP(&ddrFile, "dryrun-req", "d", "", "Program(s) and state(s) in dryrun REST request format")
 	debugCmd.Flags().IntVarP(&appID, "app-id", "a", 1380011588, "Application ID for stateful TEAL if not set in transaction(s)")
+	debugCmd.Flags().StringVarP(&indexerURL, "indexer-url", "i", "", "URL for indexer to fetch Balance records from to evaluate stateful TEAL")
```
algorandskiy, line 160: I'm not quite familiar with the indexer, but I suppose it has some authentication mechanism like API token or similar. If so, we need another parameter for it.

brianolson: It can take the same kind of header token as `algod`

cmd/tealdbg/local.go
```diff
@@ -321,7 +321,7 @@ func (r *LocalRunner) Setup(dp *DebugParams) (err error) {
 				ledger, states, err = makeAppLedger(
```
algorandskiy, line 321: sounds like it is time for appLedgerParams struct..

cmd/tealdbg/localLedger.go
```diff
@@ -39,7 +55,7 @@ type balancesAdapter struct {
 func makeAppLedger(
 	balances map[basics.Address]basics.AccountData, txnGroup []transactions.SignedTxn,
 	groupIndex int, proto config.ConsensusParams, round int, latestTimestamp int64,
-	appIdx basics.AppIndex, painless bool,
+	appIdx basics.AppIndex, painless bool, indexerURL string, indexerToke string,
```
algorandskiy, line 58: indexerToken

cmd/tealdbg/local_test.go
```diff
@@ -925,3 +935,116 @@ func TestLocalLedger(t *testing.T) {
...
+	// make balance records
+	appIdx := basics.AppIndex(100)
+	assetIdx := basics.AssetIndex(50)
+	brs := makeSampleBalanceRecord(sender, assetIdx, appIdx)
+	//balanceBlob := protocol.EncodeMsgp(&brs)
+	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		switch {
+		case regexp.MustCompile(`^/v2/accounts/`).MatchString(r.URL.Path):
+			w.WriteHeader(200)
+			if r.URL.Path[13:] == brs.Addr.String() {
```
nicholasguoalgorand, line 953: find a better way of doing this

cmd/tealdbg/localLedger.go
```diff
@@ -53,6 +80,31 @@ func makeAppLedger(
...
+			// only populate from indexer if balance record not specified
+			if _, ok := balances[acc]; !ok {
+				var err error
+				balances[acc], err = getBalanceFromIndexer(indexerURL, indexerToken, acc, round)
+				if err != nil {
+					return nil, appState{}, err
+				}
+			}
+		}
+		// TODO (nicholasguo): populate balances with creators of apps[]
```
algorandskiy, line 95: implement TODO

cmd/tealdbg/local_test.go
algorandskiy: see `strings.HasPrefix`

cmd/tealdbg/local_test.go
```diff
@@ -923,3 +934,125 @@ func TestLocalLedger(t *testing.T) {
...
+				account, err := v2.AccountDataToAccount(brs.Addr.String(), &brs.AccountData, map[basics.AssetIndex]string{}, 100, basics.MicroAlgos{Raw: 0})
+				a.NoError(err)
+				accountResponse := AccountIndexerResponse{Account: account, CurrentRound: 100}
+				response, err := json.Marshal(accountResponse)
+				a.NoError(err)
+				w.Write(response)
+			}
+		case regexp.MustCompile(`^/v2/applications/`).MatchString(r.URL.Path):
+			w.WriteHeader(200)
+			if r.URL.Path[17:] == strconv.FormatUint(uint64(appIdx), 10) {
```
algorandskiy, line 962: `strings.HasPrefix`

cmd/tealdbg/local_test.go
```diff
@@ -158,6 +163,11 @@ func makeSampleBalanceRecord(addr basics.Address, assetIdx basics.AssetIndex, ap
 			Total:     100,
 			UnitName:  "tok",
 			AssetName: "asset",
+			Manager: addr,
```
derbear, line 166: Can you run `go fmt` on all the changed files?

PR #1263, opened 2020-07-17
Fix `TxnRoot` for block 0

algomaxj:
## Summary

Back in https://github.com/algorand/go-algorand/pull/944, we started encoding nil paysets and empty paysets with the same commitment. In normal operation, blocks with no transactions create a `nil` payset. The genesis block, however, commits to a payset that is empty but not nil. I noted this in the PR at the time and didn't think it was a big deal because it didn't affect catchup, but in hindsight that was silly since it introduces version-based subjectivity into what block 0 is.

This PR computes the genesis `TxnRoot` as we originally did for the genesis block. For archival nodes, it replaces block 0 with the recomputed block 0 at startup. We should remove this code after the next protocol upgrade, when we're sure that all up-to-date archival nodes have run it once.

### DIFF
--- a/data/transactions/aggregates_test.go
+++ b/data/transactions/aggregates_test.go
@@ -61,6 +61,7 @@ func TestPaysetDoesNotCommitToSignatures(t *testing.T) {
 func TestEmptyPaysetCommitment(t *testing.T) {
 	const nilFlatPaysetHash = "WRS2VL2OQ5LPWBYLNBCZV3MEQ4DACSRDES6IUKHGOWYQERJRWC5A"
 	const emptyFlatPaysetHash = "E54GFMNS2LISPG5VUGOQ3B2RR7TRKAHRE24LUM3HOW6TJGQ6PNZQ"
+	const merklePaysetHash = "4OYMIQUY7QOBJGX36TEJS35ZEQT24QPEMSNZGTFESWMRW6CSXBKQ"
 
 	// Non-genesis blocks should encode empty paysets identically to nil paysets
 	var nilPayset Payset
@@ -70,4 +71,11 @@ func TestEmptyPaysetCommitment(t *testing.T) {
 	// Genesis block should encode the empty payset differently
 	require.Equal(t, emptyFlatPaysetHash, Payset{}.CommitGenesis(true).String())
 	require.Equal(t, nilFlatPaysetHash, nilPayset.CommitGenesis(true).String())
+
+	// Non-flat paysets (which we have dropped support for) should encode
+	// the same regardless of nilness or if this is a genesis block
+	require.Equal(t, merklePaysetHash, Payset{}.CommitGenesis(false).String())
+	require.Equal(t, merklePaysetHash, nilPayset.CommitGenesis(false).String())
+	require.Equal(t, merklePaysetHash, Payset{}.Commit(false).String())
+	require.Equal(t, merklePaysetHash, nilPayset.Commit(false).String())
 }
--- a/data/transactions/aggregates_test.go
+++ b/data/transactions/aggregates_test.go
@@ -57,3 +57,17 @@ func TestPaysetDoesNotCommitToSignatures(t *testing.T) {
 	commit2 := payset.Commit(false)
 	require.Equal(t, commit1, commit2)
 }
+
+func TestEmptyPaysetCommitment(t *testing.T) {
+	const nilFlatPaysetHash = "WRS2VL2OQ5LPWBYLNBCZV3MEQ4DACSRDES6IUKHGOWYQERJRWC5A"
+	const emptyFlatPaysetHash = "E54GFMNS2LISPG5VUGOQ3B2RR7TRKAHRE24LUM3HOW6TJGQ6PNZQ"
+
+	// Non-genesis blocks should encode empty paysets identically to nil paysets
+	var nilPayset Payset
+	require.Equal(t, nilFlatPaysetHash, Payset{}.Commit(true).String())
+	require.Equal(t, nilFlatPaysetHash, nilPayset.Commit(true).String())
+
+	// Genesis block should encode empty payset differently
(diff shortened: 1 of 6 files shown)


algomaxj:
(I will test this on a `mainnet` archival node before merging this by fetching block 0 before and after)

data/transactions/aggregates_test.go
```diff
@@ -57,3 +57,17 @@ func TestPaysetDoesNotCommitToSignatures(t *testing.T) {
 	commit2 := payset.Commit(false)
 	require.Equal(t, commit1, commit2)
 }
+
+func TestEmptyPaysetCommitment(t *testing.T) {
```
tsachiherman, line 61: The test is good.. but does not cover the non-flat Commits. I know that we stopped supporting this by version 10, but if we're to retain the support and add a unit test to validate that, than we should validate both cases.
I would not feel bad if we were to drop the support for non-flat Commits. ( since its not in use )

ledger/blockdb.go
```diff
@@ -144,6 +144,17 @@ func blockGetCert(tx *sql.Tx, rnd basics.Round) (blk bookkeeping.Block, cert agr
 	return
 }
 
+func blockReplaceIfExists(tx *sql.Tx, blk bookkeeping.Block, cert agreement.Certificate) error {
+	_, err := tx.Exec("UPDATE blocks SET proto=?, hdrdata=?, blkdata=?, certdata=? WHERE rnd=?",
```
tsachiherman, line 148: Can you examine the result and return bool which indicates if the row was updated or not ?
( and it case it was updated, write an info line to the log file )

ledger/blockdb.go
```diff
@@ -144,6 +144,17 @@ func blockGetCert(tx *sql.Tx, rnd basics.Round) (blk bookkeeping.Block, cert agr
 	return
 }
 
+func blockReplaceIfExists(tx *sql.Tx, blk bookkeeping.Block, cert agreement.Certificate) error {
```
tsachiherman, line 147: It would be a nice thing if we could add some validation at this point : 
The Proto and certdata should be identical to the old ones. If that's not the case, we want to write some warning since we didn't expect that.

ledger/ledger.go
```diff
@@ -269,7 +268,25 @@ func initBlocksDB(tx *sql.Tx, l *Ledger, initBlocks []bookkeeping.Block, isArchi
...
+		// Manually replace block 0, even if we already had it
+		// (necessary to normalize the payset commitment because of a
+		// bug that caused its value to change)
+		//
+		// Don't bother for non-archival nodes since they will toss
+		// block 0 almost immediately
+		//
+		// TODO remove this once a version containing this code has
+		// been deployed to archival nodes
+		if len(initBlocks) > 0 && initBlocks[0].Round() == basics.Round(0) {
```
tsachiherman, line 281: Ideally we wouldn't have this piece of code being executed upon every startup, however, given that we're going to remove this after two consecutive consensus upgrade releases, I think it can stay as is.

data/transactions/aggregates_test.go
algomaxj: added!

ledger/blockdb.go
algomaxj: done!

algomaxj:
Thanks -- will comment back here once I have tested on a mainnet node.

algomaxj:
Tested on an archival `mainnet` node. Block 0 before:

```
{
  "block": {
    "fees:b32": "Y76M3MSY6DKBRHBL7C3NNDXGS5IIMQVQVUAB6MP4XEMMGVF2QWNA====",
    "gen": "mainnet-v1.0",
    "gh:b32": "YBQ4JWH4DW655UWXMBF6IVUOH5WQIGMHVQ333ZFWEC22WOJERLPQ====",
    "proto": "https://github.com/algorandfoundation/specs/tree/5615adc36bad610c7f165fa2967f4ecfa75125f0",
    "rate": 20000000,
    "rwcalr": 500000,
    "rwd:b32": "737777777777777777777777777777777777777777777777777Q====",
    "seed:b32": "YBQ4JWH4DW655UWXMBF6IVUOH5WQIGMHVQ333ZFWEC22WOJERLPQ====",
    "ts": 1560211200,
    "txn:b32": "WRS2VL2OQ5LPWBYLNBCZV3MEQ4DACSRDES6IUKHGOWYQERJRWC5A===="
  },
  "cert": {}
}
```

After:

```
{
  "block": {
    "fees:b32": "Y76M3MSY6DKBRHBL7C3NNDXGS5IIMQVQVUAB6MP4XEMMGVF2QWNA====",
    "gen": "mainnet-v1.0",
    "gh:b32": "YBQ4JWH4DW655UWXMBF6IVUOH5WQIGMHVQ333ZFWEC22WOJERLPQ====",
    "proto": "https://github.com/algorandfoundation/specs/tree/5615adc36bad610c7f165fa2967f4ecfa75125f0",
    "rate": 20000000,
    "rwcalr": 500000,
    "rwd:b32": "737777777777777777777777777777777777777777777777777Q====",
    "seed:b32": "YBQ4JWH4DW655UWXMBF6IVUOH5WQIGMHVQ333ZFWEC22WOJERLPQ====",
    "ts": 1560211200,
    "txn:b32": "E54GFMNS2LISPG5VUGOQ3B2RR7TRKAHRE24LUM3HOW6TJGQ6PNZQ===="
  },
  "cert": {}
}
```

Block 2345 before (it's big):

```
$ goal ledger block 2345 --b32 -d ~/bigdisk/.catchup/ | sha256sum
01537f6a0510ec550ff36a7e5e24fa812ec708ec1791fe8189b36d2660955bd7  -
```

Block 2345 after:

```
ubuntu@ip-172-31-21-169:~/go-algorand$ goal ledger block 2345 --b32 -d ~/bigdisk/.catchup/ | sha256sum
01537f6a0510ec550ff36a7e5e24fa812ec708ec1791fe8189b36d2660955bd7  -
```

PR #1264, opened 2020-07-18
Improve dryrun and tealdbg compatibility

algorandskiy:
## Summary

* tealdbg now uses DryrunRequest.Sources if supplied
* Fix competing AppParams resolution
  if both present in accounts and DryrunRequest.Apps: account data
  take priority and only empty values are overwritten
* Document program search in both dryrun and tealdbg

## Test Plan

Existing autotest.
Manual test with Python SDK: craft a TEAL test, save DryrunRequest, use it to run both dryrun and tealdbg.

### DIFF
--- a/daemon/algod/api/server/v2/dryrun.go
+++ b/daemon/algod/api/server/v2/dryrun.go
@@ -78,7 +78,9 @@ func DryrunRequestFromGenerated(gdr *generated.DryrunRequest) (dr DryrunRequest,
 	return
 }
 
-func (dr *DryrunRequest) expandSources() error {
+// ExpandSources takes DryrunRequest.Source, compiles and
+// puts into appropriate DryrunRequest.Apps entry
+func (dr *DryrunRequest) ExpandSources() error {
 	for i, s := range dr.Sources {
 		program, err := logic.AssembleString(s.Source)
 		if err != nil {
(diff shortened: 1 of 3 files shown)


daemon/algod/api/server/v2/dryrun.go
```diff
@@ -287,8 +289,17 @@ func (dl *dryrunLedger) Get(addr basics.Address, withPendingRewards bool) (basic
...
-		out.AppParams = make(map[basics.AppIndex]basics.AppParams)
-		out.AppParams[basics.AppIndex(app.Id)] = ApplicationParamsToAppParams(&app.Params)
+		params := ApplicationParamsToAppParams(&app.Params)
+		if out.AppParams == nil {
+			out.AppParams = make(map[basics.AppIndex]basics.AppParams)
+			out.AppParams[basics.AppIndex(app.Id)] = params
+		} else {
+			ap, ok := out.AppParams[basics.AppIndex(app.Id)]
+			if ok {
+				MergeAppParams(&ap, &params)
```
algorandskiy, line 299: this fixes the situation when AccountData has AppParams with schema and predefined global state but a program to run is supplied as source with in DryrunRequest.Sources

daemon/algod/api/server/v2/dryrun.go
```diff
@@ -287,8 +289,17 @@ func (dl *dryrunLedger) Get(addr basics.Address, withPendingRewards bool) (basic
...
+				out.AppParams[basics.AppIndex(app.Id)] = ap
+			}
```
tsachiherman, line 301: it looks like the symmetrically case of !ok is missing. I can't tell if it's not needed or it's a genuine bug.

algorandskiy: Good catch, fixed

PR #1267, opened 2020-07-20
Ensure `ApplyData.EvalDelta` is clear if `applyEvalDelta` fails

algomaxj:
If `StatefulEval` returns an invalid `EvalDelta` for a `ClearStateProgram`, the current code would still fill in `ApplyData` with the `EvalDelta`, which it shouldn't do.

Fortunately, this case should be unreachable, because `StatefulEval` should never return an invalid `EvalDelta`. However, I used this as an opportunity to clean up the relatively confusing `errIfNotApplied` path in favor of a new error type.

This also uncovered a test that wasn't working properly -- it was assuming the `ClearStateProgram` was succeeding because of the non-empty `EvalDelta` in `ApplyData`, when in fact it was failing.


data/transactions/application_test.go
```diff
@@ -519,22 +512,21 @@ func TestAppCallApplyGlobalStateDeltas(t *testing.T) {
 	// now check errors with non-default values
 	b.SetProto(protocol.ConsensusV23)
 	ed.GlobalDelta["uint"] = basics.ValueDelta{Action: basics.SetUintAction, Uint: 1}
-	err = ac.applyEvalDelta(ed, params, creator, sender, &b, appIdx, errIfNotApplied)
-	a.NoError(err)
+	err = ac.applyEvalDelta(ed, params, creator, sender, &b, appIdx)
+	a.True(isApplyError(err))
```
algorandskiy, line 516: nit: add `a.Error(err)` above

data/transactions/application_test.go
```diff
@@ -583,18 +575,19 @@ func TestAppCallApplyLocalsStateDeltas(t *testing.T) {
 	var params basics.AppParams
 	var appIdx basics.AppIndex
 	var b testBalances
-	var errIfNotApplied = false
+	errIfNotApplied := false
```
algorandskiy, line 578: is it needed?

data/transactions/application_test.go
```diff
@@ -1012,6 +1004,7 @@ func TestAppCallClearState(t *testing.T) {
 	ad := &ApplyData{}
 	b.appCreators = make(map[basics.AppIndex]basics.Address)
 	b.balances = make(map[basics.Address]basics.AccountData, 2)
+	b.SetProto(protocol.ConsensusFuture)
```
algorandskiy, line 1007: good catch

data/transactions/application_test.go
algomaxj: done (here and below as well)

data/transactions/application_test.go
algomaxj: nope! removed

PR #1270, opened 2020-07-20
merklearray implementation

zeldovich:
This adds simple support for Merkle trees over dense arrays.

This will be useful in committing to our payset (so that we can have compact proofs of transactions), and for committing to online accounts (as will be needed for compact certificates).

Nothing in this PR is using this code yet, but I figured I'd ask for feedback on this, since it's reasonably small and might be useful for other things as well.

### DIFF
--- a/crypto/merklearray/layer.go
+++ b/crypto/merklearray/layer.go
@@ -27,7 +27,7 @@ func (p *pair) ToBeHashed() (protocol.HashID, []byte) {
 // and returns the next-higher level in the tree,
 // represented as a layer.
 func (l layer) up() layer {
-	res := make(layer, (len(l)+1)/2)
+	res := make(layer, (uint64(len(l))+1)/2)
 	for i := 0; i < len(l); i += 2 {
 		var p pair
 		p.l = l[i]
--- a/crypto/merklearray/layer.go
+++ b/crypto/merklearray/layer.go
@@ -1,0 +1,40 @@
+package merklearray
+
+import (
+	"github.com/algorand/go-algorand/crypto"
+	"github.com/algorand/go-algorand/protocol"
+)
+
+// A layer of the Merkle tree consists of a dense array of hashes at that
+// level of the tree.  Hashes beyond the end of the array (e.g., if the
+// number of leaves is not an exact power of 2) are implicitly zero.
+type layer []crypto.Digest
+
+// A pair represents an internal node in the Merkle tree.
+type pair struct {
+	l crypto.Digest
+	r crypto.Digest
+}
+
+func (p *pair) ToBeHashed() (protocol.HashID, []byte) {
+	var buf [2 * crypto.DigestSize]byte
+	copy(buf[:crypto.DigestSize], p.l[:])
+	copy(buf[crypto.DigestSize:], p.r[:])
+	return protocol.MerkleArrayNode, buf[:]
+}
+
+// up takes a layer representing some level in the tree,
+// and returns the next-higher level in the tree,
+// represented as a layer.
+func (l layer) up() layer {
+	res := make(layer, (len(l)+1)/2)
+	for i := 0; i < len(l); i += 2 {
+		var p pair
+		p.l = l[i]
+		if i+1 < len(l) {
+			p.r = l[i+1]
+		}
+		res[i/2] = crypto.HashObj(&p)
+	}
+	return res
+}
--- a/crypto/merklearray/layer.go
+++ b/crypto/merklearray/layer.go
@@ -1,3 +1,19 @@
+// Copyright (C) 2019-2020 Algorand, Inc.
+// This file is part of go-algorand
+//
+// go-algorand is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as
+// published by the Free Software Foundation, either version 3 of the
+// License, or (at your option) any later version.
+//
+// go-algorand is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with go-algorand.  If not, see <https://www.gnu.org/licenses/>.
+
 package merklearray
 
 import (
--- a/crypto/merklearray/merkle.go
+++ b/crypto/merklearray/merkle.go
@@ -122,6 +122,10 @@ func Verify(root crypto.Digest, elems map[uint64]crypto.Hashable, proof []crypto
 		if err != nil {
 			return err
 		}
+
+		if l > 64 {
+			return fmt.Errorf("Verify exceeded 64 levels, more than 2^64 leaves not supported")
+		}
 	}
 
 	computedroot, ok := pl[0]
--- a/crypto/merklearray/merkle.go
+++ b/crypto/merklearray/merkle.go
@@ -7,6 +7,8 @@ import (
 	"github.com/algorand/go-algorand/crypto"
 )
 
+// Tree is a Merkle tree, represented by layers of nodes (hashes) in the tree
+// at each height.
 type Tree struct {
 	// Level 0 is the leaves.
 	levels []layer
@@ -16,6 +18,7 @@ func (tree *Tree) topLayer() layer {
 	return tree.levels[len(tree.levels)-1]
 }
 
+// Build constructs a Merkle tree given an array.
 func Build(array Array) (*Tree, error) {
 	tree := &Tree{}
 
@@ -41,6 +44,7 @@ func Build(array Array) (*Tree, error) {
 	return tree, nil
 }
 
+// Root returns the root hash of the tree.
 func (tree *Tree) Root() crypto.Digest {
 	// Special case: commitment to zero-length array
 	if len(tree.levels) == 0 {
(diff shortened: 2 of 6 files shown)


crypto/merklearray/layer.go
```diff
@@ -0,0 +1,40 @@
...
+	copy(buf[:crypto.DigestSize], p.l[:])
+	copy(buf[crypto.DigestSize:], p.r[:])
+	return protocol.MerkleArrayNode, buf[:]
+}
+
+// up takes a layer representing some level in the tree,
+// and returns the next-higher level in the tree,
+// represented as a layer.
+func (l layer) up() layer {
+	res := make(layer, (len(l)+1)/2)
```
algomaxj, line 30: I am not sure how reasonable of a concern this is, but on a 32-bit platform then I think this could plausibly overflow if `l` has length `(2 << 31) - 1`

https://play.golang.org/p/W-PmprYmWEc

tsachiherman: Isn't that intended to contain the transactions within a block ? if so, I can't imagine that we'll reach more than 2^31 tx/block. or maybe I miss something ?

algomaxj: Yeah, I was just reviewing this as an abstract primitive. If something at a higher level is limiting the length then it's nothing to worry about.

zeldovich: Sure, might as well fix it!  Thanks!

crypto/merklearray/merkle.go
```diff
@@ -0,0 +1,133 @@
...
+
+	var leaves layer
+	arraylen := array.Length()
+	for i := uint64(0); i < arraylen; i++ {
+		data, err := array.Get(i)
+		if err != nil {
+			return nil, err
+		}
+
+		leaves = append(leaves, crypto.HashObj(data))
```
tsachiherman, line 29: If you're going to store transaction ids in the tree, maybe we can avoid hashing the leafs ? I know that from order of magnitude it would not make a difference, but we might be able to squeeze some more performance out by doing so. In case it has any security reasons, then scrap that.

crypto/merklearray/merkle.go
```diff
@@ -0,0 +1,133 @@
...
+	}
+
+	if arraylen > 0 {
+		tree.levels = []layer{leaves}
```
tsachiherman, line 33: you can pretty easily preallocate the levels array.

crypto/merklearray/merkle.go
zeldovich: We can just make the data be the transaction in that case.

crypto/merklearray/merkle.go
```diff
@@ -0,0 +1,133 @@
...
+
+	for l := uint64(0); len(s.hints) > 0 || len(pl) > 1; l++ {
+		var err error
+		pl, err = pl.up(s, l, true)
+		if err != nil {
+			return err
+		}
+	}
+
+	computedroot, ok := pl[0]
```
algomaxj, line 127: Is it OK that I could give you a length 0 proof, and an `elems` that maps `0` to the `pair` that hashes to `root`, and this function would succeed? In other words it seems like `Prove` can succeed for intermediate nodes?

zeldovich: Yeah, I was thinking about how to avoid this, but none of the options seemed terribly clean, and since the `pair` is internal to this package, this would require the caller to explicitly return the `HashID` of `pair` in their own object.

crypto/merklearray/merkle.go
zeldovich: I think it would need to be something like `1+floor(log2(len(leaves)-1)`, which seems messy enough, combined with the lack of `log2()` for integers in Go, to not bother with the preallocation..

crypto/merklearray/partial.go
```diff
@@ -0,0 +1,107 @@
...
+	for i := 0; i < len(positions); i++ {
+		pos := positions[i]
+		posHash := pl[pos]
+
+		siblingPos := pos ^ 1
+		siblingHash, ok := pl[siblingPos]
+		if ok {
+			// If our sibling is also in the partial layer, use its
+			// hash (and skip over its position).
+			i++
```
algomaxj, line 76: Just to confirm my understanding, we can only do this because the `positions` are sorted, right? So if we hit this case, we must be in the left sibling.

zeldovich: Exactly right

crypto/merklearray/merkle.go
```diff
@@ -0,0 +1,133 @@
...
+	pl := make(partialLayer)
+	for pos, elem := range elems {
+		pl[pos] = crypto.HashObj(elem)
+	}
+
+	s := &siblings{
+		hints: proof,
+	}
+
+	for l := uint64(0); len(s.hints) > 0 || len(pl) > 1; l++ {
```
algomaxj, line 119: Is there an easy way to validate that `s.hints` is not needlessly long? It would be cool to prevent a DoS due to a very long proof.

algomaxj: Also, suppose `elems` contained every leaf. Now we will never even look at `proof`. Is it OK that we have the property that proofs are not canonical?

zeldovich: Worst case it should be log(N)*len(pl), where N is the total number of leaves in the tree.

But I think the DoS you're thinking about is probably in terms of receiving the hints in the first place?  In my code (not part of this PR) that sends a proof over the wire, I added a `allocbound=` for `msgp` on the `[]crypto.Digest` proof.

algomaxj: If I send you `n` `hints`, then you will do at least `n` hashes because you will keep calling `top` in this loop, right? Worst case for a valid proof would be `log(N)*len(pl)`, but I could make a needlessly long proof. But if the `allocbound` is small enough then I guess that is not a problem.

I was just wondering if there was some easy way to be like "oh, we've already reached the root, if there are any hints left then we can stop anyway"

zeldovich: Ah, I see what you mean.  Yeah, it seems like we'll do at least 1 hash per element in `s.hints`, or perhaps more (if elements are close enough to each other).

Regarding canonical proofs: I was planning to check for `len(s.hints)==0` at the end, but forgot..  Thanks for catching that!  I think with that check, proofs should be canonical.

I believe the root cause for these complications is that we don't know the size of the tree.  If we grow the root commitment to include not just the `crypto.Digest` but also a `uint64` height, then this would let us compute the bound on `len(hints)`, and as a result also bound how many hashes are computed.

Perhaps a reasonable trade-off would be to define a function to compute the bound on `len(hints)` if the caller has a bound on the number of leaves?

zeldovich: Edit to the above comment: now I remember what happened to the `len(s.hints)==0` check that I was planning to add..  It got folded into the loop condition.  So if you have extra hints, the loop will keep hashing up, past the (expected) root of the tree, and the resulting `computedroot` will not match, so `Verify()` will fail.  So, I think this code has proof canonicality.

algomaxj: Ah, right, we have canonicality exactly because of the potential DoS vector.

algomaxj: Maybe it would be OK to just cap the implied depth of the tree at 64, since we can't represent more than 2**64 things anyway?

zeldovich: That could work.

In that case, the DoS vector is potentially 64 hashes per item in `elems`.  Which means the adversary would need to get the verifier to not just consider a large proof `hints`, but also try to verify a large number of elements.  So you'd still want to bound at least one of them, but this does seem potentially better than requiring the caller to bound both..

crypto/merklearray/partial.go
```diff
@@ -0,0 +1,107 @@
...
+// the proof.
+//
+// If doHash is false, fill in zero hashes, which suffices for constructing
+// a proof.
+func (pl partialLayer) up(s *siblings, l uint64, doHash bool) (partialLayer, error) {
+	positions := make([]uint64, 0, len(pl))
+	for pos := range pl {
+		positions = append(positions, pos)
+	}
+	sort.Slice(positions, func(i, j int) bool { return positions[i] < positions[j] })
```
algomaxj, line 64: If you want to avoid the sorting, you could change the `partialLayer` type to be a `[]IndexedHashable` where

```
type IndexedHashable struct {
    Index uint64
    Hash crypto.Hashable
}
```

Then you could kill the sorting, because as you append to the next `partialLayer`, things would stay sorted. Then you'd only have to sort the leaves layer up in `Verify` (or have callers to `Verify` pass a sorted slice of `IndexedHashable` -- you can verify that things are correctly sorted in the for loop below). To check for your sibling you'd just check the `Index` one after you in the partial layer.

I like the simplicity of maps but if we're optimizing for performance, this might be the way to go?

zeldovich: Good point.  I switched to slices, which noticeably sped up both `Prove` and `Verify`.

tsachiherman:
fyi : you'll need to add the license header to all the go files to make it pass the travis build.

zeldovich:
I'm pretty happy with this Merkle array API, since I have a working compact certificate implementation on top of it now.

@algomaxj or @tsachiherman , can you approve it so that I can merge it?

tsachiherman:
> I'm pretty happy with this Merkle array API, since I have a working compact certificate implementation on top of it now.
> 
> @algomaxj or @tsachiherman , can you approve it so that I can merge it?

I'll sync up with you in person.

zeldovich:
@tsachiherman , what does the assignment mean?

As far as I can tell, github tells me I'm not in the list of people who can merge PRs in this repo anymore, so I can't merge it myself.

tsachiherman:
It's just helping me to organize the PRs so that I'll know to associate the
PR with a person rather than looking at the PR creator.

PR #1275, opened 2020-07-21
Add blocks header to the header cache

tsachiherman:
## Summary

The typical flow is that we add a new block, and then try to evaluate the subsequent block.
When doing so, the previous block was already flushed to disk by the `block.syncer`, and so we're reloading the block from the `blockqueue` just so we can store it's header in the `ledger.headerCache`.

From functional perspective, there is nothing wrong here, but it's not very efficient. This PR adds the block header to the blocks header cache on Ledger.AddValidatedBlock, which ensure that subsequent calls to BlockHdr would not hit the disk.

## Test Plan

Added a unit test to verify that it's working correctly.

### DIFF
--- a/ledger/blockqueue.go
+++ b/ledger/blockqueue.go
@@ -39,9 +39,15 @@ type blockEntry struct {
 type blockQueue struct {
 	l *Ledger
 
+	// lastCommitted is the last committed round that was flushed to the block database
 	lastCommitted basics.Round
-	q             []blockEntry
+	// q is the list of blocks that have yet to be flushed to the block database, and are still pending to be flushed.
+	q []blockEntry
+	// lastCommittedEntry is the last entry that was flushed to the block database. It's a MRU level 1 cache, allowing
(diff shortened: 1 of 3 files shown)


tsachiherman:
> Why not just add a call to `l.headerCache.Put(vb.blk.Round(), vb.blk.BlockHeader)` to `Ledger.AddValidatedBlock()`?

Ahh.. I thought about that one. That would change the behavioral semantics, as it would cache the value *before* it's being persisted to disk ( it might still be a valid thing to do ).

zeldovich:
The current semantics already allow returning this block from `Ledger.BlockHdr()`, because it can be returned by the blockqueue before the blockqueue flushes it.

It's also the case that, if a block is passed to `AddValidatedBlock()`, then consensus must have already been reached on it.  Thus, even if we crash before persisting it to disk, we will not see a different block after restarting.

tsachiherman:
Sounds good. I will update the PR accordingly.

PR #1277, opened 2020-07-21
Add additional type information to teal programs in spec.

winder:
Add this additional type information to the account model to keep it consistent with indexer.


winder:
@tsachiherman you're right, this doesn't effect algod directly, it would be used by other tools which consume the algod spec. For example if/when we create our own README generator for the APIs they can attach additional information or link to other supporting documentation.

PR #1282, opened 2020-07-22
TEAL: fix txn and txna assemble and disassemble routines

algorandskiy:
## Summary

txn with two args is assembler-level alias for txna,
and this requires some additional checks

1. Fix copy-paste error in txna disassemble routine
2. Fix assembling 'txn Accounts' without second argument

## Test Plan

Added new unit tests for `txn/txna` and `gtxn/gtxna`

### DIFF
--- a/data/transactions/logic/assembler.go
+++ b/data/transactions/logic/assembler.go
@@ -707,6 +707,10 @@ func assembleTxn(ops *OpStream, spec *OpSpec, args []string) error {
 	if !ok {
 		return fmt.Errorf("txn unknown arg %s", args[0])
 	}
+	_, ok = txnaFieldSpecByField[fs.field]
+	if ok {
+		return fmt.Errorf("found txna field %s in txn op", args[0])
+	}
 	if fs.version > ops.Version {
 		return fmt.Errorf("txn %s available in version %d. Missed #pragma version?", args[0], fs.version)
 	}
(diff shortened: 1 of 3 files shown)


data/transactions/logic/assembler.go
```diff
@@ -760,6 +764,10 @@ func assembleGtxn(ops *OpStream, spec *OpSpec, args []string) error {
 	if !ok {
 		return fmt.Errorf("gtxn unknown arg %s", args[1])
 	}
+	_, ok = txnaFieldSpecByField[fs.field]
+	if ok {
+		return fmt.Errorf("gtxn unknown arg %s", args[1])
```
tsachiherman, line 769: two requests:
1. could you make the error string here differ than the one above ?
2. are you sure you need to print the arg[1] here in the error ? I'm asking because you didn't use it as a qualifier to determine that there was an error. ( i.e. you used fs.field, which you want to print out )

brianolson: I think here and in assembleTxn we know a little bit more, it matches a list, it is a known thing but in the wrong place? maybe we can note that in the error message?

algorandskiy: My intention was in preserving the same error as in original TEAL v1. If you think it is not important I'll change to a better one.

tsachiherman: I think that it's important to preserve the TEAL v1 language support, but it's ok to add more ( and better ! ) error messages for invalid TEAL programs. ( i.e. we've creating a better compiler )

algorandskiy: Fixed!
1. Use `found (g)txna field in (g)txn op` message
2. Update tests

> are you sure you need to print the arg[1] here in the error 

`arg[1]` here is the same as `fs.field.String()`, plus, we want to report back to the user exact string from his input

PR #1286, opened 2020-07-22
Add test for when the user does not provide a correct application json

bricerisingalgorand:
## Summary

This adds a test for goal clerk dryrun failures. Right now it just tests the failure case of not providing an application file, but it includes a method that can test any error message scenario we can expect

## Test Plan

I ran the test and ensured that the test is checking the expected values

### DIFF
--- a/test/e2e-go/cli/goal/expect/goalDryrunRestTest.exp
+++ b/test/e2e-go/cli/goal/expect/goalDryrunRestTest.exp
@@ -21,7 +21,7 @@ proc TestGoalDryrun { DRREQ_FILE TEST_PRIMARY_NODE_DIR } {
     }
 }
 
-proc TestGoalDryrunFailure { DRREQ_FILE TEST_PRIMARY_NODE_DIR EXPECTED_STATUS_CODE EXPECTED_MESSAGE} {
+proc TestGoalDryrunExitCode { DRREQ_FILE TEST_PRIMARY_NODE_DIR EXPECTED_STATUS_CODE EXPECTED_MESSAGE} {
     set MESSAGE_MATCHED 0
     spawn goal clerk dryrun-remote -d $TEST_PRIMARY_NODE_DIR -D $DRREQ_FILE -v
     expect {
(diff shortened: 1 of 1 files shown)


test/e2e-go/cli/goal/expect/goalDryrunRestTest.exp
```diff
@@ -88,6 +107,7 @@ if { [catch {
     TestGoalDryrun $DRREQ_FILE_1 $TEST_PRIMARY_NODE_DIR
     TestGoalDryrun $DRREQ_FILE_2 $TEST_PRIMARY_NODE_DIR
     TestGoalDryrun $DRREQ_FILE_3 $TEST_PRIMARY_NODE_DIR
+    TestGoalDryrunFailure "" $TEST_PRIMARY_NODE_DIR 1 "Cannot read file : open : no such file or directory"
```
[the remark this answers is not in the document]
bricerisingalgorand: Thanks! I added another test for this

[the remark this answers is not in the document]
bricerisingalgorand: https://github.com/algorand/go-algorand/pull/1286/commits/9c2bc527418e4edfa3a656e4db67a59c68ab98f0

test/e2e-go/cli/goal/expect/goalDryrunRestTest.exp
```diff
@@ -21,6 +21,25 @@ proc TestGoalDryrun { DRREQ_FILE TEST_PRIMARY_NODE_DIR } {
...
 
+proc TestGoalDryrunFailure { DRREQ_FILE TEST_PRIMARY_NODE_DIR EXPECTED_STATUS_CODE EXPECTED_MESSAGE} {
+    set MESSAGE_MATCHED 0
+    spawn goal clerk dryrun-remote -d $TEST_PRIMARY_NODE_DIR -D $DRREQ_FILE -v
+    expect {
+        timeout { ::AlgorandGoal::Abort "goal clerk dryrun-remote timeout" }
+        $EXPECTED_MESSAGE {puts "message matched"; set MESSAGE_MATCHED 1; close}
+        eof
+    }
+    catch wait result
```
algorandskiy, line 32: does it really work outside if `expect` block?

bricerisingalgorand: I believe so, the result variable does end up containing the exit code that we needed. Apparently this is how you wait for expect to hit eof

https://stackoverflow.com/questions/3299502/how-to-return-spawned-process-exit-code-in-expect-script

algorandskiy: I did not find it in the doc :(
To ensure, could you please check it by adding
```tcl
TestGoalDryrunFailure $DRREQ_FILE_3 $TEST_PRIMARY_NODE_DIR 0 "PASS"
```

Maybe `TestGoalDryrunFailure` might be renamed to `TestGoalDryrunExitCode`

PR #1289, opened 2020-07-22
Stateful teal test app read expect test

egieseke:
## Summary

Added an expect test for the `goal app read` command.  

## Test Plan

Ran `make integration`.


egieseke:
> What is a difference between this `app read` test and existing `test/scripts/e2e_subs/e2e-app-bootloader.sh` ?

Very similar, except that the Wallet is specified.

egieseke:
@algorandskiy I would like to commit this PR as is, and then create a new one to support adding the group transaction.  I created a new ticket to add the group transaction functionality here: https://github.com/algorand/go-algorand-internal/issues/668

algorandskiy:
I do not see a reason to merge it as is since it duplicates existing `e2e-app-bootloader.sh` test.
The problem is this PR is based on #1291 that is merged so that `statefulTealAppReadTest.exp` is already in master.

egieseke:
This duplicates #1291.

PR #1293, opened 2020-07-23
base64 encode binary keys/values in json `TealKeyValue`, `EvalDeltaKeyValue` model

algomaxj:
## Summary
- Byte keys and values should be base64 encoded when represented as json
- `EvalDeltaKeyValue` values were being base64 encoded, but their keys weren't

### DIFF
--- a/daemon/algod/api/server/v2/account.go
+++ b/daemon/algod/api/server/v2/account.go
@@ -17,6 +17,7 @@
 package v2
 
 import (
+	"encoding/base64"
 	"errors"
 
 	"github.com/algorand/go-algorand/crypto"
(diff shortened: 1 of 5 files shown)


daemon/algod/api/server/v2/account.go
```diff
@@ -138,10 +139,22 @@ func convertGeneratedTKV(akvs *generated.TealKeyValueStore) basics.TealKeyValue
 
 	tkv := make(basics.TealKeyValue)
 	for _, kv := range *akvs {
-		tkv[kv.Key] = basics.TealValue{
+		// Decode base-64 encoded map key
+		decodedKey, err := base64.StdEncoding.DecodeString(kv.Key)
+		if err != nil {
+			decodedKey = nil
```
tsachiherman, line 145: if `decodedKey = nil` , do you really want to `tkv[string(decodedKey)] = ... ` ?

algomaxj: Good point, failing silently is bad. I made it propagate up the error.

Issue #1212, opened 2020-06-29
goal clerk compile accepts programs that jumps after last instruction


fabrice102:
### Subject of the issue

If a program jumps to a label at the end of the program after which there is no instruction, `algod` will reject the program systematically. However, `goal clerk compile` does compile the program without issue.

See also https://github.com/algorand/pyteal/issues/5

### Your environment
* Software version: `algod -v`

```
8589934599
2.0.7.dev [HEAD] (commit #7b6d4d47)
go-algorand is licensed with AGPLv3.0
source code available at https://github.com/algorand/go-algorand
```

* Node status if applicable: `goal node status`

```
Last committed block: 7680192
Time since last block: 3.7s
Sync Time: 0.0s
Last consensus protocol: https://github.com/algorandfoundation/specs/tree/e5f565421d720c6f75cdd186f7098495caf9101f
Next consensus protocol: https://github.com/algorandfoundation/specs/tree/e5f565421d720c6f75cdd186f7098495caf9101f
Round for next consensus protocol: 7680193
Next consensus protocol supported: true
Last Catchpoint: 7680000#UINZMCTITNOSEMF2I2CTBRHHBK7CNCPJYWPSWJ7WRNZ5ABGXTQLA
Genesis ID: testnet-v1.0
Genesis hash: SGO1GKSzyE7IEPItTxCByw9x8FmnrCDexi9/cOUJOiI=
```

* Operating System details. macOS

### Steps to reproduce

Create a file `basic.teal`:

```
int 1
bnz l
l:
```

And run:
```bash
$ goal clerk compile basic.teal
basic.teal: ZV47DPGNBFFHWBNSI3WDA4RH5LCWIVTTZKJAP45DCCXR5J5ZOFJH7RCICU

$ goal clerk send --from-program basic.teal --to GCBRBIIMS7IVTHC3Y7ML6WAGGPK4L3LVSHDWXPTMD5YO2E7CFWNN5X2KOE --amount 10
Couldn't broadcast tx with algod: HTTP 400 Bad Request:   8 bnz target beyond end of program

$ goal clerk compile -D basic.teal.tok
// version 1
intcblock 1
intc_0
bnz label1
```

### Expected behaviour

`goal clerk compile` and `algod` should both either accept or reject the program.

### Actual behaviour

`goal clerk compile` accepts while `algod` rejects.

Decompilation also uses a non-existing label `label1`.

algorandskiy:
[...]

PR #1295, opened 2020-07-23
TEAL: assembler error on branching behind last instruction

algorandskiy:
## Summary

* TEAL v1 does not allow such program, so make assembler compatible
* TEAL v2 permits the mentioned behavior

## Test Plan

Added new test to validate v1 and v2 behavior

## Notes

1. Fixed disassembly listing for such programs - no unknown labels anymore
2. Switched some asm tests to `t.Parallel()`

### DIFF
--- a/data/transactions/logic/assembler.go
+++ b/data/transactions/logic/assembler.go
@@ -1072,6 +1072,16 @@ func (ops *OpStream) assemble(fin io.Reader) error {
 		err := fmt.Errorf("unknown opcode %v", opstring)
 		return lineErr(ops.sourceLine, err)
 	}
+
+	// backward compatibility: do not allow jumps behind last instruction in TEAL v1
+	if ops.Version <= 1 {
+		for label, dest := range ops.labels {
+			if dest == ops.Out.Len() {
+				return fmt.Errorf(":%d label %v is too far away", ops.sourceLine, label)
+			}
+		}
+	}
+
 	// TODO: warn if expected resulting stack is not len==1 ?
 	return ops.resolveLabels()
 }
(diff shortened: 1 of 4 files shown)


algorandskiy:
Fix for #1212

data/transactions/logic/assembler.go
```diff
@@ -1298,6 +1308,17 @@ func (dis *disassembleState) putLabel(label string, target int) {
 	dis.pendingLabels[target] = label
 }
 
+func (dis *disassembleState) outputLabelIfNeeded() error {
```
tsachiherman, line 1311: nit: coding style - 
```go
func (dis *disassembleState) outputLabelIfNeeded() (err error) {
 	if label, hasLabel := dis.pendingLabels[dis.pc]; hasLabel {
 		_, err := fmt.Fprintf(dis.out, "%s:\n", label)
 	}
 	return
 }
```

data/transactions/logic/assembler.go
```diff
@@ -1072,6 +1072,16 @@ func (ops *OpStream) assemble(fin io.Reader) error {
 		err := fmt.Errorf("unknown opcode %v", opstring)
 		return lineErr(ops.sourceLine, err)
 	}
+
+	// backward compatibility: do not allow jumps behind last instruction in TEAL v1
+	if ops.Version <= AssemblerDefaultVersion {
```
brianolson, line 1077: Shouldn't this simply be `<= 1`

brianolson: `AssemblerDefaultVersion` will change, but we want to pin this at <=1

algorandskiy: done

data/transactions/logic/assembler.go
algorandskiy: done

algorandskiy:
Fixed `TestBackwardCompatAssemble` a bit to extend the test beyond v2

PR #1296, opened 2020-07-24
Optimize accountsUpdateBalances implementation

tsachiherman:
## Summary
Existing code was calling `accountsUpdateBalances` for each round, applying the deltas to the trie and committing the changes to disk. However, on some of our testings, we have 1-2 changes per block, and we're pushing 2000-4000 blocks very quickly. This leads to a scenario where `commitRound` is asked to flush 500+ rounds to disk in a single batch. From
logical perspective, there is nothing wrong here.

Unfortunately, flushing 500 rounds * 2 changes =~ 1000 Merkle trie pages. ( i.e. the root page + one extra page update)
When flushing 1000 pages on slow machines such as travis, it can take a long time, and we want to avoid that.
( i.e. if a single flush takes 50ms on a NAS drive, it would take 50 seconds to complete.. )
In order to resolve that, I have moved the round iteration loop into the `accountsUpdateBalances`, and added periodic flush that depends on the number of pending changes. The Merkle trie implementation is sufficiently smart to omit unneeded changes, so that if some pages are being added and later on removed, it would just ignore these pages completly.

Note that for a normal running node, this is not an issue, since the incoming block update rate would not be that high.

## Solution Approach
I was contemplating whether a good fix would be to update the unit test to reflect a realistic usage of the `Ledger` interface or improve the `acctupdate` implementation. I went with the `acctupdate` path as this would marginally improve the overall performance.

## Test Plan
Use existing test

### DIFF
--- a/ledger/acctupdates.go
+++ b/ledger/acctupdates.go
@@ -49,6 +49,9 @@ const (
 	// trieRebuildAccountChunkSize defines the number of accounts that would get read at a single chunk
 	// before added to the trie during trie construction
 	trieRebuildAccountChunkSize = 512
+	// trieAccumulatedChangesFlush defines the number of pending changes that would be applied to the merkle trie before
+	// we attempt to commit them to disk while writing a batch of rounds balances to disk.
+	trieAccumulatedChangesFlush = 256
 )
 
 // trieCachedNodesCount defines how many balances trie nodes we would like to keep around in memory.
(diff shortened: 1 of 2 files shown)


algonautshant:
in the summary:
> However, on some of out testings

However, on some of **our** testings

ledger/acctupdates.go
```diff
@@ -1036,35 +1039,54 @@ func (au *accountUpdates) deleteStoredCatchpoints(ctx context.Context, dbQueries
...
+				}
+
 			}
-			if !deleted {
-				au.log.Warnf("failed to delete hash '%v' from merkle trie", deleteHash)
+			if !delta.new.IsZero() {
+				addHash := accountHashBuilder(addr, delta.new, protocol.Encode(&delta.new))
+				added, err = au.balancesTrie.Add(addHash)
+				if err != nil {
+					return err
```
algorandskiy, line 1069: is it OK to lose pending changes if at some point error happened? Maybe
```
defer func() {
	if accumulatedChanges > 0 {
		err = au.balancesTrie.Commit()
	}
}()
```
?

ledger/totals_test.go
```diff
@@ -69,7 +68,7 @@ func TestAlgoCountMarshalMsg(t *testing.T) {
 	inBuffer := make([]byte, 0, 128)
 	outBuffer, err := ac.MarshalMsg(inBuffer)
 	require.NoError(t, err)
-	fmt.Printf("out %d in %d\n", len(outBuffer), len(inBuffer))
+	//fmt.Printf("out %d in %d\n", len(outBuffer), len(inBuffer))
```
algorandskiy, line 71: forgotten debug output?

ledger/acctupdates.go
tsachiherman: yes, it's ok to loose changes if error occurs - 
in case of an error, the entire transaction would get rolled back, which includes all the pending operations.
The call to "au.balancesTrie.Commit()" would trickle down to a series of calls to `merkleCommitter.StorePage`, which uses the current transaction.

The reason we don't want to hold this off for too long is because it would create higher memory pressure for pending items. ( which in turn, make the single "Commit" call take a long time )

The number I've picked ( 256 ) is likely to be on the low end, and we probably could go higher.

ledger/totals_test.go
tsachiherman: We don't really need this debug output, unless there is a problem. I don't think it's very valuable at this time, and if we'll find that the subsequent entry fail, we could improve it by making it into a require.Truef(...) which would include the above information.

algorandskiy: I meant to delete it.
In logic tests Brian established a pattern like
```
if err != nil {
// print trace
}
require.NoError(t, err)
```

tsachiherman: updated.

ledger/acctupdates.go
```diff
@@ -49,6 +49,9 @@ const (
 	// trieRebuildAccountChunkSize defines the number of accounts that would get read at a single chunk
 	// before added to the trie during trie construction
 	trieRebuildAccountChunkSize = 512
+	// trieAccumulatedChangesFlush defines the number of pending changes that would be applied to the merkle trie before
+	// we attempt to commit them to disk while writing a batch of rounds balances to disk.
+	trieAccumulatedChangesFlush = 256
```
algonautshant, line 54: Does this mean that balancesTrie will no longer by in sync with the database at all times?
Does this comment need updating: 
```
// The Trie tracking the current account balances. Always matches the balances that were
// written to the database.
balancesTrie *merkletrie.Trie
```

tsachiherman:
No, it will be in sync since all the changes are committed in a single
database tramsaction.

PR #1299, opened 2020-07-24
Optimize accounts totals updates

tsachiherman:
## Summary
Existing code was calling `accountsNewRound` for each updated round, updating the account totals by applying the deltas. This works correctly, but not very efficiently.

If we have a series of updates that were batched together ( let's say 500 rounds ), we will be calling this function 500 times. On each such iteration, we'll load the current totals from disk, update the totals based on the reward changes and the new accounts data, and store it back to disk. In total we would have 500 read and 500 write operations.

With this change in place, we would load it at the beginning, and perform all the operations in memory. Than, we would save it to disk. As before, it all being executed within the same transaction, however, after the change we would perform total of 2 database operations compared to 1000 operations before.

This is a followup on the previous PR https://github.com/algorand/go-algorand/pull/1296, which optimized the accountsUpdateBalances.

## Solution Approach
I was contemplating whether a good fix would be to update the unit test to reflect a realistic usage of the `Ledger` interface or improve the `acctupdate` implementation. I went with the `acctupdate` path as this would marginally improve the overall performance.

## Test Plan
Existing tests were updated.


ledger/accountdb.go
```diff
@@ -618,39 +618,22 @@ func accountsPutTotals(tx *sql.Tx, totals AccountTotals, catchpointStaging bool)
 	return err
 }
 
-func accountsNewRound(tx *sql.Tx, updates map[basics.Address]accountDelta, creatables map[basics.CreatableIndex]modifiedCreatable, rewardsLevel uint64, proto config.ConsensusParams) (err error) {
-	var ot basics.OverflowTracker
-	totals, err := accountsTotals(tx, false)
-	if err != nil {
-		return
-	}
+func accountsNewRound(tx *sql.Tx, updates map[basics.Address]accountDelta, creatables map[basics.CreatableIndex]modifiedCreatable) (err error) {
```
algorandskiy, line 621: it would be helpful to have a func comment :)

ledger/acctupdates.go
```diff
@@ -1229,11 +1229,16 @@ func (au *accountUpdates) commitRound(offset uint64, dbRound basics.Round, lookb
 			treeTargetRound = dbRound + basics.Round(offset)
 		}
 		for i := uint64(0); i < offset; i++ {
-			err = accountsNewRound(tx, deltas[i], creatableDeltas[i], roundTotals[i+1].RewardsLevel, protos[i+1])
+			err = accountsNewRound(tx, deltas[i], creatableDeltas[i])
 			if err != nil {
 				return err
 			}
 		}
+		err = totalsNewRounds(tx, deltas[:offset], roundTotals[1:offset+1], protos[1:offset+1])
```
algorandskiy, line 1237: Not sure I understand the change. Before `totals.applyRewards` was called `offset` times due to the loop above.
Now it is called the same amount of times due to the loop inside `totalsNewRounds`. The only difference seems like in a number of `accountsTotals` calls. So can't we call it right before the `for i := uint64(0); i < offset; i++` loop above and leave `accountsNewRound` as-is except `accountsTotals` call (not a suggestion but a question).

ledger/accountdb.go
tsachiherman: will do.

ledger/acctupdates.go
tsachiherman: Your observation is correct. Both before and after the change, `totals.applyRewards` is called `offset` times. `totals.applyRewards` is in-memory function that perform few additions on the totals data structure, and is pretty fast. On the other hand, the `accountsTotals` and `accountsPutTotals` are sql calls which aren't that fast ( relatively ), and we want to optimize these.

tsachiherman: i.e. taking the `totals.applyRewards` out of the loop isn't the "gain" here. The gain is by calling `accountsTotals` and `accountsPutTotals` only once.

algomaxj, line 1237: This might not be possible, but could `roundTotals` or `protos` ever have fewer than 1 entry? For example, if `offset` is 0, the above loop will never be called. So this refactor does change the semantics of this block of code slightly.

tsachiherman: I think that having offset of 0 is not a possible scenario here. It's protected by the enqueueing function `committedUpTo` by doing:
```
...
	defer func() {
		au.accountsMu.RUnlock()
		if dc.offset != 0 {
			au.committedOffset <- dc
		}
	}()
...
```

The `roundTotals` as well as the `protos` are always having one more element compared to the `deltas`. The preamble retains that property:
```
	copy(deltas, au.deltas[:offset])
	copy(creatableDeltas, au.creatableDeltas[:offset])
	copy(roundTotals, au.roundTotals[:offset+1])
	copy(protos, au.protos[:offset+1])
```

The `offset -= uint64(au.dbRound - dbRound)` would never bring the `offset` to zero as long as the blocks numbers are growing.

tsachiherman:
Correct. We do already support thousands of writes per blocks. However, we
dont support that on travis.. where the Atomic operation became ridiculesly
slow when we use certain non-realistic operations, such as generating 2000
blocks and adding them to the ledger, all at once.
As I mentioned in the summary, the gains here are expected to be marginal
for the real life scenarios.

PR #1304, opened 2020-07-26
Fix a bug where goal inspect fails to process a rekeyed transaction

### LINKED ISSUE #1300: goal clerk inspect doesn't work with "rekey" transaction
### Subject of the issue
When trying to inspect a transaction with `sgnr` field assigned.  `goal inspect` fails to decode it.

algorotem:
## Summary

`goal clerk inspect` failed to work on rekeyed transactions. This PR fixes it. 

Fixes #1300


algorandskiy:
I think we need a test for this, how about extending `test/e2e-go/cli/goal/expect/goalTxValidityTest.exp` ?

PR #1305, opened 2020-07-27
Make /tmp into a tmpfs on travis linux/amd64

tsachiherman:
## Solution

Configure the travis linux/amd64 hosts to use a 768MB of in-memory tmp directory.
This change is intended to create a more deterministic running environment, where the network mapped disk would not negatively affect the unit tests.

### DIFF
--- a/scripts/travis/configure_dev.sh
+++ b/scripts/travis/configure_dev.sh
@@ -18,6 +18,8 @@ if [[ "${OS}" == "linux" ]]; then
         sudo apt-get update -y
         sudo apt-get -y install sqlite3
     elif [[ "${ARCH}" == "amd64" ]]; then
+        echo "/etc/fstab : "
+        sudo cat /etc/fstab
         sudo mv /tmp /old_tmp
         sudo mkdir -p /tmp
         sudo chmod 777 /tmp
--- a/scripts/travis/configure_dev.sh
+++ b/scripts/travis/configure_dev.sh
@@ -26,8 +26,9 @@ if [[ "${OS}" == "linux" ]]; then
         # none /var/ramfs tmpfs defaults,size=768m,noatime 0 2
 
         sudo umount -l /var/ramfs
+        sudo umount -l /dev/shm
         sudo sed '3d' /etc/fstab > fstab
-        sudo echo "none /var/ramfs tmpfs defaults,noatime,nosuid,nodev,size=256m,noatime,mode=0755 0 0" >> fstab
+        #sudo echo "none /var/ramfs tmpfs defaults,noatime,nosuid,nodev,size=256m,noatime,mode=0755 0 0" >> fstab
         sudo echo "tmpfs /tmp tmpfs rw,noatime,size=768m,noatime,mode=1777 0 0" >> fstab
         sudo cp fstab /etc/fstab
         sudo rm fstab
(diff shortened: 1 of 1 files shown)


scripts/travis/configure_dev.sh
```diff
@@ -17,6 +17,10 @@ if [[ "${OS}" == "linux" ]]; then
         set -e
         sudo apt-get update -y
         sudo apt-get -y install sqlite3
+    elif [[ "${ARCH}" == "amd64" ]]; then
+        sudo mkdir -p /mnt/ramdisk
+        sudo mount -t tmpfs -o rw,size=512M tmpfs /mnt/ramdisk
+        ln -s /mnt/ramdisk /tmp
```
onetechnical, line 23: If /tmp pre-exists, won't this create /tmp/ramdisk? I'm also concerned about obliterating the pre-existing files in /tmp, but that might be okay depending on what state of initialization we're in.

tsachiherman: As for your concerns - 
1. Yes, /tmp already exists - but it's part of the "/" directory. When doing `df -h`, I received:
```
Filesystem      Size  Used Avail Use% Mounted on
udev            3.9G     0  3.9G   0% /dev
tmpfs           797M  908K  796M   1% /run
/dev/sda1        68G   17G   52G  24% /
tmpfs           3.9G     0  3.9G   0% /dev/shm
tmpfs           5.0M     0  5.0M   0% /run/lock
tmpfs           3.9G     0  3.9G   0% /sys/fs/cgroup
/dev/loop0      109M  109M     0 100% /snap/google-cloud-sdk/138
/dev/loop1       97M   97M     0 100% /snap/core/9436
none            768M     0  768M   0% /var/ramfs
/dev/loop2       55M   55M     0 100% /snap/core18/1754
/dev/sda15      105M  3.6M  101M   4% /boot/efi
/dev/loop3       55M   55M     0 100% /snap/core18/1880
/dev/loop4       97M   97M     0 100% /snap/core/9665
/dev/loop5      119M  119M     0 100% /snap/google-cloud-sdk/142
tmpfs           797M     0  797M   0% /run/user/2000
```
2. The `/run` is read-only and I wasn't sure if I should attempt to change that.
3. I think that if we had any dependencies on the existing tmp directory, it would have caused build errors. Having the build passing seems to me like we're on the right path.

Would you have implemented the Ramdisk differently ? if so, please let me know how. I'm not particularly picky about the implementation of it.

onetechnical: The ramdisk setup is fine, I'm just not sure it'll be used unless we change our code to use /tmp/ramdisk instead of /tmp.

scripts/travis/configure_dev.sh
```diff
@@ -17,6 +17,12 @@ if [[ "${OS}" == "linux" ]]; then
         set -e
         sudo apt-get update -y
         sudo apt-get -y install sqlite3
+    elif [[ "${ARCH}" == "amd64" ]]; then
+        sudo mv /tmp /old_tmp
+        sudo mkdir -p /tmp
+        sudo chmod 777 /tmp
+        cp -r /old_tmp/ /tmp
```
onetechnical, line 24: Shouldn't this be after the mount command, not before?

tsachiherman:
I'm closing this for now. Will reopen as needed.

PR #1308, opened 2020-07-27
Move transaction apply methods into their own package

algomaxj:
In preparation for refactoring parts of `Applications`, it made sense to move the `apply` methods into their own package. With this change, the new `apply` package can refer to the `logic` package without creating a circular reference (since `logic` depends on `transactions`).

I also moved the `transactions.Balances` interface into the new `apply` package.


ledger/apply/apply.go
```diff
@@ -0,0 +1,57 @@
...
+	Move(src, dst basics.Address, amount basics.MicroAlgos, srcRewards *basics.MicroAlgos, dstRewards *basics.MicroAlgos) error
+
+	// Balances correspond to a Round, which mean that they also correspond
+	// to a ConsensusParams.  This returns those parameters.
+	ConsensusParams() config.ConsensusParams
+}
+
+// StateEvaluator is an interface that provides some Stateful TEAL
+// functionality that may be passed through to Apply from ledger, avoiding a
+// circular dependency between the logic and transactions packages
```
algorandskiy, line 52: seems like this comment about the dependency is not actual anymore. Maybe we even do not need the interface at all

algomaxj: Yeah -- one of the main reasons I'm doing this change is to get rid of the interface entirely! That's what I have on my WIP refactor branch.

algomaxj: Updated the comment.

PR #1311, opened 2020-07-28
Add rekey multisig support for goal 

algorotem:
If an account was rekeyed to a multisig account, there is no, currently, a way to sign and submit a transaction from this account using `goal`. This PR adds this ability. 

How to use it? 
Say that account `DIHP2DL3GOE62DS3VQGNYH4QD2JT3ZPKTQ3MUKVDJ7AFSGT7ZJHSZGXUGQ` was rekeyed to the multisig account `GTGUFZ7MEBVB2UMYNF2KEMTM4FOT73J2EFGQXVMZ6KS4ZUWTGA5JU3P6LY`, which is composed of 

```
SNCK2NULGJVILRSKF7BRML7R3ARJVZVM7NUGIGZAWGWOBGFZJ7JYHW2VQU
TIWJONWXCNZSIUAVVRLVHDENTDGQQMUGD4GFCAQGQWPKPDCH5GJJMNSUB4 
U5JNPKNJVVGBJDWH67OYI3UQE2QQMP4J5DHV3KQYPCHCORVRH46OPVLMCA
``` 

with threshold 1. 

to sign a transaction from the account using one of the keys in kmd, one should use the following commands. We assume the multisig account is not on the same machine, so the user needs to provide `goal` with the complete pre-image. 

First, generate the unsigned transaction with the relevant multisig params and the signer address. 
```
$ goal -d $ALGORAND_DIR clerk send -f DIHP2DL3GOE62DS3VQGNYH4QD2JT3ZPKTQ3MUKVDJ7AFSGT7ZJHSZGXUGQ -t DYAOIKTWYIV5MSQGBJTIA5IEQKYNY5QEPQDUVNUSV5BTMJO5DHQG4LMTNY -a $AMT --msig-params="1 SNCK2NULGJVILRSKF7BRML7R3ARJVZVM7NUGIGZAWGWOBGFZJ7JYHW2VQU TIWJONWXCNZSIUAVVRLVHDENTDGQQMUGD4GFCAQGQWPKPDCH5GJJMNSUB4 U5JNPKNJVVGBJDWH67OYI3UQE2QQMP4J5DHV3KQYPCHCORVRH46OPVLMCA" -o $OUT_FILE_NAME
```

Now simply sign it, as usual, using a key in kmd 
```
/goal -d ~/.algorand/pnet/Primary clerk multisig sign -t $OUT_FILE_NAME -a TIWJONWXCNZSIUAVVRLVHDENTDGQQMUGD4GFCAQGQWPKPDCH5GJJMNSUB4
```

Then it can be sent as usual using `goal clerk rawsend`
```
./goal -d ~/.algorand/pnet/Primary clerk rawsend -f $OUT_FILE_NAME
```


daemon/kmd/wallet/driver/sqlite_errors.go
```diff
@@ -43,5 +43,5 @@ var errWalletNotFound = fmt.Errorf("wallet not found")
 var errSQLiteWrongType = fmt.Errorf("sqlite wallet driver returned wrong wallet type")
 var errNameTooLong = fmt.Errorf("wallet name too long, must be <= %d bytes", sqliteMaxWalletNameLen)
 var errIDTooLong = fmt.Errorf("wallet id too long, must be <= %d bytes", sqliteMaxWalletIDLen)
-var errMsigWrongAddr = fmt.Errorf("given multisig preimage hashes to wrong address")
+var errMsigWrongAddr = fmt.Errorf("given multisig preimage hashes to neither Sender or AuthAddr")
```
tsachiherman, line 46: nit: syntax : neither -> nor

daemon/kmd/wallet/driver/ledger.go
```diff
@@ -355,7 +355,7 @@ func (lw *LedgerWallet) SignProgram(data []byte, src crypto.Digest, pw []byte) (
 }
 
 // MultisigSignTransaction implements the Wallet interface.
-func (lw *LedgerWallet) MultisigSignTransaction(tx transactions.Transaction, pk crypto.PublicKey, partial crypto.MultisigSig, pw []byte) (crypto.MultisigSig, error) {
+func (lw *LedgerWallet) MultisigSignTransaction(tx transactions.Transaction, pk crypto.PublicKey, partial crypto.MultisigSig, pw []byte, signer crypto.Digest) (crypto.MultisigSig, error) {
```
tsachiherman, line 358: I'm abit confused. It seems that you've done a lot in order to pass on the signer to this function, but when you reached this function it doesn't look like you're doing anything. What am I missing ?

algorotem: Yes, I basically had to add it because of the interface even though it doesn't do much. @algomaxj

ryanRfox:
The workflow passes as implemented. I feel it can be merged. However, `goal clerk inspect $OUT_FILE_NAME` returns the following error:
```
Cannot decode transactions from 1311.utxn: non-idempotent transformation to inspectSignedTxn (DeepEqual)
```

I am able to read the file with `cat $OUT_FILE_NAME | msgpacktool -d -32` so it's not a blocker. A subsequent feature request should be raised if `goal clerk inspect` cannot be modified to support this transaction structure within this PR.

algorotem:
> The workflow passes as implemented. I feel it can be merged. However, `goal clerk inspect $OUT_FILE_NAME` returns the following error:
> 
> ```
> Cannot decode transactions from 1311.utxn: non-idempotent transformation to inspectSignedTxn (DeepEqual)
> ```
> 
> I am able to read the file with `cat $OUT_FILE_NAME | msgpacktool -d -32` so it's not a blocker. A subsequent feature request should be raised if `goal clerk inspect` cannot be modified to support this transaction structure within this PR.

It got fixed in another PR. Hasn't been merged yet

test/e2e-go/kmd/e2e_kmd_wallet_multisig_test.go
```diff
@@ -224,6 +224,141 @@ func TestMultisigSign(t *testing.T) {
...
+			Threshold: 2,
+			Version:   1,
+			Subsigs:   []crypto.MultisigSubsig{{Key: pk1}, {Key: pk2}, {Key: pk3}},
+		},
+		WalletPassword: f.WalletPassword,
+	}
+
+	resp2 := kmdapi.APIV1POSTMultisigTransactionSignResponse{}
+	err = f.Client.DoV1Request(req2, &resp2)
+	require.Error(t, err)
```
tsachiherman, line 358: it would be nice to test here for a specific error

test/e2e-go/kmd/e2e_kmd_wallet_multisig_test.go
```diff
@@ -224,6 +224,141 @@ func TestMultisigSign(t *testing.T) {
...
+	resp, err = f.Client.GenerateKey([]byte(walletHandleToken))
+	require.NoError(t, err)
+	pk2 := addrToPK(t, resp.Address)
+	pk3 := crypto.PublicKey{1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1} // some public key we haven't imported
+
+	sender, err := f.Client.GenerateKey([]byte(walletHandleToken))
+	require.NoError(t, err)
+	pkSender := addrToPK(t, sender.Address)
+
+	// Create a 2-of-3 multisig account from the three public keys
```
tsachiherman, line 324: I think it would be an interesting to see if this scales up - can we create and verify that 149/253 keys are valid ?
( the simple use case of 2/3 is great. but it won't tell us if we have a bug related to a higher numbers )

cmd/goal/clerk.go
```diff
@@ -85,6 +87,7 @@ func init() {
 	sendCmd.Flags().StringVarP(&progByteFile, "from-program-bytes", "P", "", "Program binary to use as account logic")
 	sendCmd.Flags().StringSliceVar(&argB64Strings, "argb64", nil, "base64 encoded args to pass to transaction logic")
 	sendCmd.Flags().StringVarP(&logicSigFile, "logic-sig", "L", "", "LogicSig to apply to transaction")
+	sendCmd.Flags().StringVar(&msigParams, "msig-params", "", "Multisig pre image parameters - [threshold] [Address 1] [Address 2] ...\nUsed to add the necessary fields in case the account was rekeyed to a multisig account")
```
algomaxj, line 90: preimage

cmd/goal/clerk.go
```diff
@@ -389,6 +397,44 @@ var sendCmd = &cobra.Command{
 			}
 		}
 
+		// Handle the case where the user wants to send to an account that was rekeyed to a multisig account
+		if msigParams != "" {
+			// Decode params
+			params := strings.Split(msigParams, " ")
```
algomaxj, line 403: do you have to use quotation marks when using this flag? if so, maybe the split character should be a comma instead

cmd/goal/clerk.go
```diff
@@ -389,6 +397,44 @@ var sendCmd = &cobra.Command{
...
+			if len(params) < 3 {
+				reportErrorf(msigParseError, "Not enough arguments to create the multisig address.\nPlease make sure to specify the threshold and at least 2 addresses\n")
+			}
+
+			threshold, err := strconv.ParseUint(params[0], 10, 8)
+			if err != nil {
```
algomaxj, line 409: `if err != nil || threshold < 1 || threshold > 255`

test/e2e-go/kmd/e2e_kmd_wallet_multisig_test.go
algorotem: We don't have Error objects. I think that in order to do it properly it would a much larger change and outside the scope of this PR.

test/e2e-go/kmd/e2e_kmd_wallet_multisig_test.go
algorotem: I agree, but that's a bit outside the scope of this PR IMO.

cmd/goal/clerk.go
algorotem: yes, unfortunately, there is no way around it. The spaces are there to keep it consistent with the other commands (rather than commas)

PR #1318, opened 2020-07-29
Update stateful TEAL docs.

derbear:


derbear:
Note: this should wait for #1315 to merge.

PR #1319, opened 2020-07-29
Allow substring ops to return an empty string in TEAL.

derbear:
## Test Plan

A unit test was extended.


derbear:
Note: this PR should wait for #1318.

PR #1323, opened 2020-07-30
Disallow old TEAL versions for ApplicationCall transactions, disallow groups containing {Apps, RekeyTo} and v0/v1 TEAL

algomaxj:
## Summary

To prevent newly-introduced transaction fields from breaking assumptions made by older versions of TEAL, the TEAL environment prohibits the execution of TEAL on transaction groups where any transaction sets a field not present at the time the old TEAL version was introduced and instead causes those programs to fail.  For example, executing a TEAL version 1 program on a transaction with `RekeyTo` set to a nonzero address will cause the program to fail, regardless of the other contents of the program itself.

This is achieved by adding a `MinTealVersion` to the TEAL `EvalParams`. The `MinTealVersion` is computed for the entire transaction group. If the caller fails to set a `MinTealVersion`, it is computed on the fly inside of `eval.go` -- this is useful for testing.

This change also rejects v0 or v1 `ApprovalProgram`s and `ClearStateProgram`s for `Applications`.

## Test Plan

I added both unit tests and e2e tests for this change (I think e2e tests are important here to protect against the possibility that the the `EvalParams` are set up too unrealistically in the unit tests)


data/transactions/verify/txn.go
```diff
@@ -41,8 +41,9 @@ var logicErrTotal = metrics.MakeCounter(metrics.MetricName{Name: "algod_ledger_l
 // on a signed transaction.
 type Context struct {
 	Params
-	Group      []transactions.SignedTxn
-	GroupIndex int
+	Group          []transactions.SignedTxn
+	GroupIndex     int
+	MinTealVersion uint64
```
derbear, line 46: I think this should go into `Params`, not `Context`. `Params` is used for caching verified transactions, whereas `Context` is not. See here.

https://github.com/algorand/go-algorand/blob/master/data/pools/transactionPool.go#L423

derbear:
Also, I believe it will be possible to downgrade an `ApprovalProgram` or `ClearProgram` version with `UpdateApplication`, which will render the application unusable (though clearing will still work). Not sure whether we want to handle that case.

data/transactions/verify/txn.go
algomaxj: Yeah, I think the group hash committing to the other transactions should take care of it, but it's trivial to move this to `Params`, so I just did that.

data/transactions/logic/opcodes.go
```diff
@@ -28,6 +28,11 @@ const LogicVersion = 2
 // be maliciously or accidentally rekeyed.
 const rekeyingEnabledVersion = 2
 
+// appsEnabledVersion is the version of TEAL where ApplicationCall
+// functionality was enabled. We use this to disallow v0 and v1 TEAL programs
+// from being used with applications.
+const appsEnabledVersion = 2
```
algorandskiy, line 34: maybe here should be a comment like "do not edit" or a test that ensures `appsEnabledVersion == 2` to prevent accidental constants change.

data/transactions/logic/eval.go
```diff
@@ -91,6 +91,29 @@ func stackValueFromTealValue(tv *basics.TealValue) (sv stackValue, err error) {
...
 }
 
+// ComputeMinTealVersion calculates the minimum safe TEAL version that may be
+// used by a transaction in this group. It is important to prevent
+// newly-introduced transaction fields from breaking assumptions made by older
+// versions of TEAL. If one of the transactions in a group will execute a TEAL
+// program whose version predates a given field, that field must not be set
+// anywhere in the transaction group, or the group will be rejected.
+func ComputeMinTealVersion(group []transactions.SignedTxn) uint64 {
+	var minVersion uint64
```
algorandskiy, line 101: init to 1 ? I like version 1 better then 0

data/transactions/logic/eval_test.go
```diff
@@ -3716,18 +3840,14 @@ func TestAllowedOpcodesV2(t *testing.T) {
...
-						strings.Contains(err.Error(), "pc did not advance"),
-				)
 				_, err = Eval(program, ep)
 				require.Error(t, err, source)
 				require.Contains(t, err.Error(), "illegal opcode")
-				_, _, err = EvalStateful(program, ep)
-				require.Error(t, err, source)
-				require.Contains(t, err.Error(), "illegal opcode")
+
+				// It is not necessary to test CheckStateful and EvalStateful
```
algorandskiy, line 3847: I think it is better to leave `EvalStateful/CheckStateful` tests and call it with `ep. MinTealVersion = 1`. In this case another code path is triggered and actual opcodes version check happen.

data/transactions/verify/txn.go
```diff
@@ -54,23 +54,26 @@ type Context struct {
...
+	CurrSpecAddrs  transactions.SpecialAddresses
+	CurrProto      protocol.ConsensusVersion
+	MinTealVersion uint64
 }
 
 // PrepareContexts prepares verification contexts for a transaction
 // group.
 func PrepareContexts(group []transactions.SignedTxn, contextHdr bookkeeping.BlockHeader) []Context {
 	ctxs := make([]Context, len(group))
+	minTealVersion := logic.ComputeMinTealVersion(group)
```
algorandskiy, line 66: this probably should be done lazely only if logic/app call txn in the group.
Or even done in LogicSigSanityCheck?

data/transactions/logic/eval_test.go
algomaxj: Done, actually at some point I made a change to this PR that made changing this test not necessary.

data/transactions/logic/eval.go
algomaxj: I don't think we can do that, because then we'll always reject v0 programs.

data/transactions/logic/opcodes.go
algomaxj: Added a "do not edit" comment

data/transactions/verify/txn.go
algomaxj: I would prefer to leave this code very simple unless it slows things down too much. Transaction groups have at most 16 transactions, so I think this should be fast enough, but if it causes problems we can optimize.

data/transactions/logic/eval.go
derbear: ^+1

test/e2e-go/cli/goal/expect/goalDryrunRestTest.exp
```diff
@@ -78,7 +78,7 @@ if { [catch {
     puts "Primary Account Balance: $PRIMARY_ACCOUNT_BALANCE"
 
     set TEAL_PROG_FILE "$TEST_ROOT_DIR/trivial.teal"
-    exec echo int 1 > $TEAL_PROG_FILE
+    exec echo -e "#pragma version 2\nint 1" > $TEAL_PROG_FILE
```
algorandskiy, line 81: travis' macos `echo` does not have `-e` option and fails.
`printf "#pragma version 2\nint 1"` seems working

test/scripts/e2e_subs/e2e-app-bootloader.sh
```diff
@@ -32,7 +32,8 @@ sed -i"" -e "s/TMPL_APPROV_HASH/${TARGET_HASH}/g" ${TEMPDIR}/bootloader.teal
 sed -i"" -e "s/TMPL_CLEARSTATE_HASH/${TARGET_HASH}/g" ${TEMPDIR}/bootloader.teal
 
 # Create an app using filled-in bootloader template
-APPID=$(${gcmd} app create --creator ${ACCOUNT} --approval-prog ${TEMPDIR}/bootloader.teal --global-byteslices 1 --global-ints 0 --local-byteslices 0 --local-ints 0 --clear-prog <(echo 'int 1') | grep Created | awk '{ print $6 }')
+echo -e '#pragma version 2\nint 1' > "${TEMPDIR}/int1.teal"
```
algorandskiy, line 35: same here and below

PR #1324, opened 2020-07-31
Add e2e upgrade test for application over REST & Gossip

tsachiherman:
## Summary

This PR includes a new e2e test to cover the acceptable of applications creation transaction before and after an upgrade.
The test cover the REST endpoint and ensures that the node rejects application transaction before the upgrade takes place, and accept the transaction once the upgrade took place.

## Test Plan

This is an e2e test.

### DIFF
--- a/test/e2e-go/upgrades/application_support_test.go
+++ b/test/e2e-go/upgrades/application_support_test.go
@@ -35,7 +35,7 @@ import (
 
 // consensusTestUnupgradedProtocol is a version of ConsensusCurrentVersion
 // that allows the control of the upgrade from consensusTestUnupgradedProtocol to
-// consensusTestUnupgradedProtocol
+// test-fast-upgrade-future
 const consensusTestUnupgradedProtocol = protocol.ConsensusVersion("test-unupgraded-protocol")
 
 func makeApplicationUpgradeConsensus(t *testing.T) (appConsensus config.ConsensusProtocols) {
--- a/test/e2e-go/upgrades/application_support_test.go
+++ b/test/e2e-go/upgrades/application_support_test.go
@@ -285,3 +285,266 @@ int 1
 	require.Equal(t, creator, app.Params.Creator)
 	return
 }
+
+// TestApplicationsUpgrade tests that we can safely upgrade from a version that doesn't support applications
+// to a version that supports applications. It verify that prior to supporting applications, the node would not accept
+// any application transaction and after the upgrade is complete, it would support that.
+func TestApplicationsUpgradeOverGossip(t *testing.T) {
+	// set the small lambda to 500 for the duration of this test.
+	roundTimeMs := 500
+	lambda := os.Getenv("ALGOSMALLLAMBDAMSEC")
+	os.Setenv("ALGOSMALLLAMBDAMSEC", fmt.Sprintf("%d", roundTimeMs))
+	defer func() {
+		if lambda == "" {
+			os.Unsetenv("ALGOSMALLLAMBDAMSEC")
+		} else {
+			os.Setenv("ALGOSMALLLAMBDAMSEC", lambda)
+		}
+	}()
+
+	consensus := makeApplicationUpgradeConsensus(t)
+
+	var fixture fixtures.RestClientFixture
+	fixture.SetConsensus(consensus)
+	fixture.SetupNoStart(t, filepath.Join("nettemplates", "TwoNodes100SecondTestUnupgradedProtocol.json"))
+
+	// for the primary node, we want to have a different consensus which always enables applications.
+	primaryNodeUnupgradedProtocol := consensus[consensusTestFastUpgrade(protocol.ConsensusFuture)]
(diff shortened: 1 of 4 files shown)


test/testdata/nettemplates/TwoNodes100SecondTestUnupgradedProtocol.json
```diff
@@ -5,12 +5,12 @@
     "Wallets": [
       {
         "Name": "Wallet1",
-        "Stake": 0,
-        "Online": true
+        "Stake": 10,
```
algonautshant, line 8: This change does not have any side-effects to other tests that use this?

test/e2e-go/upgrades/application_support_test.go
```diff
@@ -0,0 +1,550 @@
...
+	require.Equal(t, schema, state.Schema)
+	require.Equal(t, 1, len(state.KeyValue))
+	value, ok = state.KeyValue["counter"]
+	require.True(t, ok)
+	require.Equal(t, uint64(1), value.Uint)
+
+	// call the app
+	tx, err = client.MakeUnsignedAppOptInTx(uint64(appIdx), nil, nil, nil)
+	require.NoError(t, err)
+	tx, err = client.FillUnsignedTxTemplate(user, 0, 0, fee, tx)
```
algonautshant, line 492: why the round is set to 0?

test/e2e-go/upgrades/application_support_test.go
```diff
@@ -0,0 +1,550 @@
...
+	require.NoError(t, err)
+	signedTxn, err := client.SignTransactionWithWallet(wh, nil, tx)
+	require.NoError(t, err)
+	round, err = client.CurrentRound()
+	require.NoError(t, err)
+
+	_, err = client.BroadcastTransaction(signedTxn)
+	require.NoError(t, err)
+
+	// this transaction is expect to reach the first node ( primary ), but to be rejected by the second node when transmitted over gossip.
```
algonautshant, line 406: Where are we checking that the transaction is rejected by the second node?

test/e2e-go/upgrades/application_support_test.go
```diff
@@ -0,0 +1,550 @@
...
+	"github.com/algorand/go-algorand/data/basics"
+	"github.com/algorand/go-algorand/data/transactions"
+	"github.com/algorand/go-algorand/data/transactions/logic"
+	"github.com/algorand/go-algorand/protocol"
+	"github.com/algorand/go-algorand/test/framework/fixtures"
+)
+
+// consensusTestUnupgradedProtocol is a version of ConsensusCurrentVersion
+// that allows the control of the upgrade from consensusTestUnupgradedProtocol to
+// consensusTestUnupgradedProtocol
```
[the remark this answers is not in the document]
tsachiherman: fixed.

test/e2e-go/upgrades/application_support_test.go
```diff
@@ -0,0 +1,550 @@
...
+	require.NoError(t, err)
+	require.Equal(t, uint64(0), pendingTx.TotalTxns)
+
+	// check creator's balance record for the app entry and the state changes
+	ad, err = client.AccountData(creator)
+	require.NoError(t, err)
+	require.Equal(t, 1, len(ad.AppParams))
+	var appIdx basics.AppIndex
+	var params basics.AppParams
+	for i, p := range ad.AppParams {
```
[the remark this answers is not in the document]
tsachiherman: no, but since it's a map, you can't directly take the first key/value from it without a for loop.

test/testdata/nettemplates/TwoNodes100SecondTestUnupgradedProtocol.json
tsachiherman: No. I checked all the tests that were using this json, and all seems to be happy with this change.

test/e2e-go/upgrades/application_support_test.go
tsachiherman: Setting it to zero allow the underlying implementation to fill it up for us.

test/e2e-go/upgrades/application_support_test.go
tsachiherman: we check the transaction pool of the second node : 
```
	// check that the secondary node doesn't have that transaction in it's transaction pool.
	pendingTx, err = secondary.GetPendingTransactions(1)
	require.NoError(t, err)
	require.Equal(t, uint64(0), pendingTx.TotalTxns)
```

algonautshant: Even if it is accepted, it won't be in the transaction pool forever right?

tsachiherman: correct. it would remain there until `lastRound` passes. Keep in mind that once a transaction is included in a transaction pool, it would remain there until it's no longer valid ( i.e. passed `lastRound` ) or included in a block.
We know that it wasn't included in a block since we tested both nodes, and we know that blocks are always uniform.

PR #1326, opened 2020-07-31
Filter timeout to 2s for period 0, 4s for period > 0

nicholasguoalgorand:
## Summary

Change filter timeout to a function instead of a constant. Filter Timeout is now 2s for Period 0 and 4s for Period > 0

Moved to https://github.com/algorand/go-algorand/pull/1401

## Test Plan

Unit tests for current version, new version, as well as protocol upgrade


agreement/service.go
```diff
@@ -19,6 +19,7 @@ package agreement
 //go:generate dbgen -i agree.sql -p agreement -n agree -o agreeInstall.go -h ../scripts/LICENSE_HEADER
 import (
 	"context"
+	"github.com/algorand/go-algorand/protocol"
```
derbear, line 22: Please move this import to the block of imports below.

agreement/service_test.go
```diff
@@ -974,6 +982,31 @@ func TestAgreementSynchronous5_50(t *testing.T) {
 	simulateAgreement(t, 5, 50, disabled)
 }
 
+func TestAgreementSynchronousFuture5(t *testing.T) {
```
derbear, line 985: We should also add a `TestAgreementSynchronousFuture1` which will run for short tests.

agreement/agreementtest/simulate.go
```diff
@@ -19,6 +19,7 @@ package agreementtest
 
 import (
 	"fmt"
+	"github.com/algorand/go-algorand/protocol"
```
derbear, line 22: Could you move this import to where the other `github.com/algorand/go-algorand` imports are?

agreement/common_test.go
```diff
@@ -214,6 +216,28 @@ func makeTestLedger(state map[basics.Address]basics.BalanceRecord) Ledger {
...
+	return l
+}
+
+func makeTestLedgerWithConsensusVersion(state map[basics.Address]basics.BalanceRecord, consensusVersion func(basics.Round) (protocol.ConsensusVersion, error)) Ledger {
+	l := new(testLedger)
+	l.entries = make(map[basics.Round]bookkeeping.Block)
+	l.certs = make(map[basics.Round]Certificate)
+	l.nextRound = 1
+
+	// deep copy of state
```
derbear, line 232: I don't think you introduced this comment, but we should toss it, since it's now misleading.

agreement/service_test.go
```diff
@@ -974,10 +984,47 @@ func TestAgreementSynchronous5_50(t *testing.T) {
 	simulateAgreement(t, 5, 50, disabled)
 }
 
+func TestAgreementSynchronousFuture1(t *testing.T) {
+	if testing.Short() {
+		t.Skip("Skipping agreement integration test")
+	}
```
derbear, line 990: Let's take this block out and always run the one-node test (they're relatively fast).

agreement/agreementtest/simulate.go
```diff
@@ -68,7 +69,7 @@ func (i *instant) TimeoutAt(d time.Duration) <-chan time.Time {
 		return ta
 	}
 
-	if d == agreement.FilterTimeout() && !i.HasPending("pseudonode") {
+	if d == agreement.FilterTimeout(1, protocol.ConsensusCurrentVersion) && !i.HasPending("pseudonode") {
```
derbear, line 72: I'm not completely sure, but this might need to be 0 and not one to work.

agreement/agreementtest/simulate.go
nicholasguoalgorand: Is there a script that can check for these? I feel like this is happening because my editor auto imports things

agreement/common_test.go
nicholasguoalgorand: should this be tossed in the original? aka line 212?

derbear: Yeah--let's remove it there as well.

agreement/agreementtest/simulate.go
derbear: Hmm, that's a good question. If `make sanity` doesn't do anything, then probably not currently...

nicholasguoalgorand:
moving to https://github.com/algorand/go-algorand/pull/1401

PR #1328, opened 2020-07-31
Do not save loop variable addresses in conversion routines

algorandskiy:
## Summary

REST V2 models use pointers to denote optional fields, and this a bit error prone when converting
maps or arrays of native data types to appropriate arrays of model

## Test Plan

Added tests

### DIFF
--- a/daemon/algod/api/server/v2/account.go
+++ b/daemon/algod/api/server/v2/account.go
@@ -183,33 +183,58 @@ func AccountToAccountData(a *generated.Account) (basics.AccountData, error) {
 	var assetParams map[basics.AssetIndex]basics.AssetParams
 	if a.CreatedAssets != nil && len(*a.CreatedAssets) > 0 {
 		assetParams = make(map[basics.AssetIndex]basics.AssetParams, len(*a.CreatedAssets))
+		var err error
 		for _, ca := range *a.CreatedAssets {
 			var metadataHash [32]byte
-			copy(metadataHash[:], *ca.Params.MetadataHash)
-			manager, err := basics.UnmarshalChecksumAddress(*ca.Params.Manager)
-			if err != nil {
-				return basics.AccountData{}, err
+			if ca.Params.MetadataHash != nil {
+				copy(metadataHash[:], *ca.Params.MetadataHash)
 			}
-			reserve, err := basics.UnmarshalChecksumAddress(*ca.Params.Reserve)
-			if err != nil {
-				return basics.AccountData{}, err
+			var manager, reserve, freeze, clawback basics.Address
+			if ca.Params.Manager != nil {
+				if manager, err = basics.UnmarshalChecksumAddress(*ca.Params.Manager); err != nil {
+					return basics.AccountData{}, err
+				}
 			}
-			freeze, err := basics.UnmarshalChecksumAddress(*ca.Params.Freeze)
-			if err != nil {
-				return basics.AccountData{}, err
+			if ca.Params.Reserve != nil {
+				if reserve, err = basics.UnmarshalChecksumAddress(*ca.Params.Reserve); err != nil {
+					return basics.AccountData{}, err
+				}
 			}
-			clawback, err := basics.UnmarshalChecksumAddress(*ca.Params.Clawback)
-			if err != nil {
-				return basics.AccountData{}, err
+			if ca.Params.Freeze != nil {
+				if freeze, err = basics.UnmarshalChecksumAddress(*ca.Params.Freeze); err != nil {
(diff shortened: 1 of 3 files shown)


daemon/algod/api/server/v2/account_test.go
```diff
@@ -62,38 +90,87 @@ func TestAccount(t *testing.T) {
...
-	require.Equal(t, 1, len(*conv.CreatedApps))
-	app := (*conv.CreatedApps)[0]
-	require.Equal(t, uint64(appIdx), app.Id)
-	require.Equal(t, params.ApprovalProgram, app.Params.ApprovalProgram)
-	require.Equal(t, params.GlobalStateSchema.NumUint, app.Params.GlobalStateSchema.NumUint)
-	require.Equal(t, params.GlobalStateSchema.NumByteSlice, app.Params.GlobalStateSchema.NumByteSlice)
-	require.NotNil(t, conv.AppsLocalState)
-	require.Equal(t, 1, len(*conv.AppsLocalState))
+	require.Equal(t, 2, len(*conv.CreatedApps))
+	for _, app := range *conv.CreatedApps {
```
algorandskiy, line 96: the loop is because order in resultant `CreatedApps` is not fixed due to random order of maps iteration

daemon/algod/api/server/v2/account.go
```diff
@@ -360,11 +385,12 @@ func AppParamsToApplication(creator string, appIdx basics.AppIndex, appParams *b
...
 // AssetParamsToAsset converts basics.AssetParams to generated.Asset
 func AssetParamsToAsset(creator string, idx basics.AssetIndex, params *basics.AssetParams) generated.Asset {
+	frozen := params.DefaultFrozen
 	assetParams := generated.AssetParams{
 		Creator:       creator,
 		Total:         params.Total,
 		Decimals:      uint64(params.Decimals),
-		DefaultFrozen: &params.DefaultFrozen,
+		DefaultFrozen: &frozen,
 		MetadataHash:  byteOrNil(params.MetadataHash[:]),
```
tsachiherman, line 394: The MetadataHash field also need to be handled. As is, it would not create a copy, but rather use the original pointer. ( since it's a true array and not a slice )

daemon/algod/api/server/v2/account.go
```diff
@@ -183,33 +183,58 @@ func AccountToAccountData(a *generated.Account) (basics.AccountData, error) {
...
+
+			var defaultFrozen bool
+			if ca.Params.DefaultFrozen != nil {
+				defaultFrozen = *ca.Params.DefaultFrozen
+			}
+			var url string
+			if ca.Params.Url != nil {
+				url = *ca.Params.Url
+			}
+			var unitName string
```
tsachiherman, line 222: huh.. is that really an issue ? I thought that strings are immutable.

daemon/algod/api/server/v2/account.go
```diff
@@ -183,33 +183,58 @@ func AccountToAccountData(a *generated.Account) (basics.AccountData, error) {
...
+				copy(metadataHash[:], *ca.Params.MetadataHash)
 			}
-			reserve, err := basics.UnmarshalChecksumAddress(*ca.Params.Reserve)
-			if err != nil {
-				return basics.AccountData{}, err
+			var manager, reserve, freeze, clawback basics.Address
+			if ca.Params.Manager != nil {
+				if manager, err = basics.UnmarshalChecksumAddress(*ca.Params.Manager); err != nil {
+					return basics.AccountData{}, err
+				}
```
tsachiherman, line 196: I'm surprised the addresses need to be copied. I would think that the `basics.UnmarshalChecksumAddress` would allocate storage when unmarshelling these.

algorandskiy: the problem is `ca.Params.Manager` can be `nil`

daemon/algod/api/server/v2/account.go
algorandskiy: possible `nil` deferefence is the issue

daemon/algod/api/server/v2/account.go
tsachiherman: I see. So you just fixed unrelated issue.. ok.

daemon/algod/api/server/v2/account.go
algorandskiy: yep, done

daemon/algod/api/server/v2/account.go
algorandskiy: yeah, this `AccountToAccountData` func is only used by `dryrun` and `tealdbg`

PR #1329, opened 2020-08-01
Update upgrade path to v24

tsachiherman:
## Summary

Update the configuration for upgrade path.
To be merged after the following PRs:
#1318 
#1323 

## Test Plan

Automated tests were updated


cmd/tealdbg/debugger_test.go
```diff
@@ -90,6 +90,7 @@ func (d *testDbgAdapter) eventLoop() {
 
 func TestDebuggerSimple(t *testing.T) {
 	proto := config.Consensus[protocol.ConsensusV23]
+	require.False(t, proto.Application)
```
algorandskiy, line 93: it does not matter here

tsachiherman: Would it be correct to use "ConsensusCurrentVersion" in here?

tsachiherman:
> We have bunch of `protocol.ConsensusFuture` usages in
> 
> ```
> data/basics/teal_test.go
> daemon/algod/api/server/v2/test/handlers_test.go
> ledger/apply/application_test.go
> test/e2e-go/features/transactions/accountv2_test.go
> test/e2e-go/upgrades/application_support_test.go
> ```
> 
> I think we need to switch to V24 there just to make things clear

I agree that we should, but it could be done after the branching, right ?

cmd/tealdbg/debugger_test.go
tsachiherman: or maybe I should ask : why version 23 was picked here ?

algorandskiy: Anything with LogicSig enabled, so `protocol.ConsensusV23` is just fine. I do not like `ConsensusCurrentVersion` since we might want disabling LogicSig one day.

algorandskiy: Because I was lazy to find a version where LogicSig was enabled, and `protocol.ConsensusV23` is known to be working.

tsachiherman: LogicSigVersion > 1 ? That was v18. LogicSigVersion > 2 ? that's V24..

tsachiherman: ok, I updated it to use the v18 and LogicSigVersion > 0 as a predicate.

config/consensus.go
```diff
@@ -691,63 +691,72 @@ func initConsensusProtocols() {
...
-	vFuture.MaxAppsCreated = 10
+	v24.MaxAppsCreated = 10
 
 	// Maximum number of apps a single account can opt into
-	vFuture.MaxAppsOptedIn = 10
+	v24.MaxAppsOptedIn = 10
+	Consensus[protocol.ConsensusV24] = v24
+
+	// v23 can be upgraded to v24, with an update delay of 7 days ( see calculation above )
+	v23.ApprovedUpgrades[protocol.ConsensusV24] = 140000
```
algorandskiy, line 753: Why do we need to set exact value here? It should be taken from `v20.DefaultUpgradeWaitRounds = 140000` when `ApprovedUpgrades[protocol.ConsensusV24] = 0` and it is within the margins set in
```
	v22.MinUpgradeWaitRounds = 10000
	v22.MaxUpgradeWaitRounds = 150000
```

algorandskiy, line 753: Seems like we need to check for default first and then err
```
		upgradeDelay := uint64(vote.UpgradeDelay)
		if upgradeDelay > params.MaxUpgradeWaitRounds || upgradeDelay < params.MinUpgradeWaitRounds {
			err = fmt.Errorf("applyUpgradeVote: proposed upgrade wait rounds %d out of permissible range [%d, %d]", upgradeDelay, params.MinUpgradeWaitRounds, params.MaxUpgradeWaitRounds)
			return
		}

		if upgradeDelay == 0 {
			upgradeDelay = params.DefaultUpgradeWaitRounds
		}
```

tsachiherman: I don't know how that would affect catchup. which is why I'd prefer to work around it for now. ( I don't want to run a full catchup to test that.. )

PR #1338, opened 2020-08-03
Add created/opted in apps to `goal account list`

algomaxj:
## Summary

It's not very pretty, but it matches what we do currently for assets. Here is example output:

```
[online]	VMQAA7D43Y6SZ7SJNMA6EPFEQ3DGD575OEKV4FP5EVZWSCRHKVLXZXQFZE	VMQAA7D43Y6SZ7SJNMA6EPFEQ3DGD575OEKV4FP5EVZWSCRHKVLXZXQFZE	4999999999995000 microAlgos	[created asset IDs: 4 (1234 foo), 5 (1234 bar)]	[created app IDs: 1, 2]	[opted in app IDs: 1]
```

## Testing

I manually tested this change using a private network.

### DIFF
--- a/cmd/goal/accountsList.go
+++ b/cmd/goal/accountsList.go
@@ -231,6 +231,21 @@ func (accountList *AccountsList) outputAccount(addr string, acctInfo v1.Account,
 		}
 		fmt.Printf("]")
 	}
+	if len(acctInfo.AppParams) > 0 {
+		fmt.Printf("\t[created apps:")
+		for aid := range acctInfo.AppParams {
+			fmt.Printf(" %d", aid)
+		}
+		fmt.Printf("]")
+	}
+	if len(acctInfo.AppLocalStates) > 0 {
+		fmt.Printf("\t[opted in apps:")
+		for aid := range acctInfo.AppLocalStates {
+			fmt.Printf(" %d", aid)
+		}
+		fmt.Printf("]")
+	}
+
 	if accountList.isDefault(addr) {
 		fmt.Printf("\t*Default")
 	}
--- a/cmd/goal/accountsList.go
+++ b/cmd/goal/accountsList.go
@@ -23,6 +23,7 @@ import (
 	"os"
 	"os/user"
 	"path/filepath"
+	"strings"
 
 	"github.com/algorand/go-algorand/daemon/algod/api/spec/v1"
 	"github.com/algorand/go-algorand/data/basics"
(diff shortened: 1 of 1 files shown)


algorotem:
I'm not sure I know how to parse it intuitively. 
what is the parenthesis around the assets? and is it the app index? If so, I'd separate it by commas.

cmd/goal/accountsList.go
```diff
@@ -231,6 +231,21 @@ func (accountList *AccountsList) outputAccount(addr string, acctInfo v1.Account,
 		}
 		fmt.Printf("]")
 	}
+	if len(acctInfo.AppParams) > 0 {
+		fmt.Printf("\t[created apps:")
+		for aid := range acctInfo.AppParams {
+			fmt.Printf(" %d", aid)
```
algorotem, line 237: Oh ok, I see. 
Yes, I'd sperate it with commas.

cmd/goal/accountsList.go
```diff
@@ -231,6 +231,21 @@ func (accountList *AccountsList) outputAccount(addr string, acctInfo v1.Account,
 		}
 		fmt.Printf("]")
 	}
+	if len(acctInfo.AppParams) > 0 {
+		fmt.Printf("\t[created apps:")
```
algorotem, line 235: Created apps IDs

cmd/goal/accountsList.go
```diff
@@ -231,6 +231,21 @@ func (accountList *AccountsList) outputAccount(addr string, acctInfo v1.Account,
 		}
 		fmt.Printf("]")
 	}
```
algorotem, line 233: I know it's not part of the PR but maybe worth fixing it now. Just add the word ID -> `Created assets IDs`.

cmd/goal/accountsList.go
algomaxj: I'm copying exactly what we do for assets. Do we really want commas for apps but spaces for assets?

cmd/goal/accountsList.go
algomaxj: done

cmd/goal/accountsList.go
algorotem: I think we want commas for both. 
Maybe a good opportunity to fix it.

algomaxj: alright, did commas and ID for both

cmd/goal/accountsList.go
algomaxj: done

cmd/goal/accountsList.go
```diff
@@ -225,12 +226,30 @@ func (accountList *AccountsList) outputAccount(addr string, acctInfo v1.Account,
...
+		var out []string
 		for curid, params := range acctInfo.AssetParams {
-			fmt.Printf(" %d (%d %s)", curid, params.Total, params.UnitName)
+			out = append(out, fmt.Sprintf("%d (%d %s)", curid, params.Total, params.UnitName))
 		}
-		fmt.Printf("]")
+		fmt.Printf("%s]", strings.Join(out, ", "))
 	}
+	if len(acctInfo.AppParams) > 0 {
+		fmt.Printf("\t[created app IDs: ")
```
tsachiherman, line 237: nit: unify the 
```
fmt.Printf("\t[created app IDs: ")
...
fmt.Printf("%s]", strings.Join(out, ", "))
```

into 
```
fmt.Printf("\t[created app IDs: %s]", strings.Join(out, ", "))
```

algomaxj: opted-in assets are not printed here -- they are printed elsewhere along with frozen status, creator, etc.

algomaxj: done!

PR #1348, opened 2020-08-04
Fix merketrie prioritization cache

tsachiherman:
## Summary

The `pagesPrioritizationMap` was misaligned with the `pagesPrioritizationList`, which prevented the `evict()` from correctly releasing no-longer needed nodes.

## Test Plan

Unit tests have been added to confirm fix

### DIFF
--- a/crypto/merkletrie/cache.go
+++ b/crypto/merkletrie/cache.go
@@ -72,7 +72,7 @@ type merkleTrieCache struct {
 
 	// a list of the pages priorities. The item in the front has higher priority and would not get evicted as quickly as the item on the back
 	pagesPrioritizationList *list.List
-	// the list element of each of the priorities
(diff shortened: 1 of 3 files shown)


crypto/merkletrie/cache.go
```diff
@@ -145,8 +145,9 @@ func (mtc *merkleTrieCache) getNode(nid storedNodeIdentifier) (pnode *node, err
 	return
 }
 
-// prioritizeNode make sure to move the priorities of the pages according to
-// the accessed node identifier
+// prioritizePage make sure to adjust the priority of the given node id.
```
algorandskiy, line 148: `prioritizePage` in the comment and `prioritizeNode` in the code

tsachiherman: i will fix comment in a subsequent PR

PR #1350, opened 2020-08-05
Add a fuzzer to the merkletrie cache testing

tsachiherman:
## Summary

This PR adds another test to the merkletrie cache implementation to ensure correctness on random trie operations.

## Test Plan

This is a unit test.

### DIFF
--- a/crypto/merkletrie/cache_test.go
+++ b/crypto/merkletrie/cache_test.go
@@ -155,6 +155,37 @@ func (spmc *smallPageMemoryCommitter) LoadPage(page uint64) (content []byte, err
 	return spmc.InMemoryCommitter.LoadPage(page)
 }
 
+func cacheEvictionFuzzer(t *testing.T, hashes []crypto.Digest, pageSize int64, evictSize int) {
+	var memoryCommitter smallPageMemoryCommitter
+	memoryCommitter.pageSize = pageSize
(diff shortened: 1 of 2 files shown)


crypto/merkletrie/cache_test.go
```diff
@@ -117,3 +123,80 @@ func TestCacheEviction3(t *testing.T) {
...
+// TestCacheEvictionFuzzer generates bursts of random Add/Delete operations on the trie, and
+// testing the correctness of the cache internal buffers priodically.
+func TestCacheEvictionFuzzer(t *testing.T) {
+	// create 2000 hashes.
+	leafsCount := 2000
+	hashes := make([]crypto.Digest, leafsCount)
+	for i := 0; i < len(hashes); i++ {
+		hashes[i] = crypto.Hash([]byte{byte(i % 256), byte((i / 256) % 256), byte(i / 65536)})
+	}
+	for _, pageSize := range []int64{2, 3, 8, 12, 17} {
```
algorandskiy, line 167: make page size and evict size random in a range, it would make more sense

tsachiherman: I was hoping to have some controlled dimensionality. I don't mind having some pure randomized version of these as well.

algorandskiy: yeah, generate some additional values and add to iterations

PR #1352, opened 2020-08-05
Fix Makefile to work with ubuntu 20.04

winder:
Installing python doesn't install the "python" command anymore. Switch over to "python3".

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -133,7 +133,7 @@ KMD_API_SWAGGER_INJECT := daemon/kmd/lib/kmdapi/bundledSpecInject.go
 
 $(KMD_API_SWAGGER_SPEC): $(KMD_API_FILES) crypto/libs/$(OS_TYPE)/$(ARCH)/lib/libsodium.a
 	cd daemon/kmd/lib/kmdapi && \
-		python genSwaggerWrappers.py $(KMD_API_SWAGGER_WRAPPER)
+		python3 genSwaggerWrappers.py $(KMD_API_SWAGGER_WRAPPER)
 	cd daemon/kmd && \
 		PATH=$(GOPATH1)/bin:$$PATH \
 		go generate ./...


algorandskiy:
Maybe `command -v python3` to ensure python3 is there?

winder:
Hopefully this is the correct test.

I created the following dockerfile and put it in the root of the go-algorand repo:
```
FROM ubuntu:18.04    
ENV GOLANG_VERSION 1.13
ENV DEBIAN_FRONTEND noninteractive    

RUN apt-get update && apt-get -y install wget git
RUN ln -fs /usr/share/zoneinfo/America/New_York /etc/localtime
RUN mkdir -p /tmp/go-algorand
WORKDIR "/tmp/go-algorand"

RUN wget --quiet https://dl.google.com/go/go${GOLANG_VERSION}.linux-amd64.tar.gz && tar -xvf go${GOLANG_VERSION}.linux-amd64.tar.gz && mv go /usr/local    

ENV GOROOT=/usr/local/go \    
    GOPATH=$HOME/go \    
    GOPROXY=https://gocenter.io,https://goproxy.io,direct    
ENV PATH=$GOPATH/bin:$GOROOT/bin:$PATH

COPY . /tmp/go-algorand
RUN scripts/configure_dev.sh
RUN make install

ENTRYPOINT ["/bin/bash"]    
```

I ran `make clean` in my go-algorand to ensure swagger.json is removed, so that this code path needs to be hit.

The docker file builds fine (`docker build --tag test:1.0 .`), running the file (`docker run -it test:1.0`) I was able to verify that the swagger file is generated.

tsachiherman:
@algojohnlee I think that this one seems a good change, unless you think it would break anything.

winder:
@algojohnlee this is follow-up from fixing the rest of the repo, at least for the scripts which are actually used. My system doesn't have python, somehow I missed this usage earlier.

https://github.com/algorand/go-algorand/pull/1199

PR #1359, opened 2020-08-06
core crypto primitives for compact certificates

zeldovich:
This PR implements the core cryptographic primitives for compact certificates.

To make this PR easier to review, it's a PR against a temporary branch that already has the merklearray changes (the same code that was reviewed in PR #1270).

The compact certificate construction that this is implementing is described in https://github.com/algorand/compactcert

This PR does not have any users of compact certificates yet, but I have an initial implementation of compact certificates working on top of it, so I'm pretty happy with its interface, and will open more PRs building on this code.


crypto/compactcert/builder.go
```diff
@@ -0,0 +1,241 @@
...
+}
+
+// coinIndex returns the position pos in the sigs array such that the sum
+// of all signature weights before pos is less than or equal to coinWeight,
+// but the sum of all signature weights up to and including pos exceeds
+// coinWeight.
+//
+// coinIndex works by doing a binary search on the sigs array.  The caller
+// should invoke it with lo=0 and hi=len(b.sigs).  The caller should make
+// sure that sigs[*].l is initialized before using coinIndex().
```
tsachiherman, line 143: typo `sigs[*].l`->`sigs[*].L`

crypto/compactcert/builder.go
```diff
@@ -0,0 +1,241 @@
...
+func (b *Builder) coinIndex(coinWeight uint64, lo uint64, hi uint64) (uint64, error) {
```
tsachiherman, line 144: I realize that this is a binary search like function; if we ever want to further optimize it, we could unwind it.

crypto/compactcert/builder.go
zeldovich: Oops, yeah, thanks

crypto/compactcert/builder.go
zeldovich: I tried rewriting it in a non-recursive style, and it actually seems cleaner (it's easier to see which part of the binary search bounds are changing on each goto).  Pushed that.

tsachiherman: yes, thanks for the change.

crypto/compactcert/builder.go
```diff
@@ -0,0 +1,244 @@
...
+
+// coinIndex returns the position pos in the sigs array such that the sum
+// of all signature weights before pos is less than or equal to coinWeight,
+// but the sum of all signature weights up to and including pos exceeds
+// coinWeight.
+//
+// coinIndex works by doing a binary search on the sigs array.  The caller
+// should invoke it with lo=0 and hi=len(b.sigs).  The caller should make
+// sure that sigs[*].L is initialized before using coinIndex().
+func (b *Builder) coinIndex(coinWeight uint64, lo uint64, hi uint64) (uint64, error) {
```
tsachiherman, line 144: could you create a trivial unit test for this function ? I tried to go over this, and I think the `hi = mid` might need to be `lo = mid`, or maybe the predicate isn't what I think of it.
( no rush about this one.. it would nice to have it before the package is to be merged )

crypto/compactcert/common.go
```diff
@@ -0,0 +1,88 @@
...
+	h := crypto.Hash(hashinput)
+
+	i := &big.Int{}
+	i.SetBytes(h[:])
+
+	w := &big.Int{}
+	w.SetUint64(signedWeight)
+
+	res := &big.Int{}
+	res.Mod(i, w)
```
algomaxj, line 43: I'm still working on understanding this PR, but in general is taking `hash value % signedWeight` sufficiently uniformly random? I see your comment mentions 'nearly uniform' but its unclear to me what this distribution actually looks like.

tsachiherman: If I understand it correctly, it's retaining the randomness of the hashed value. i.e. it has the same entropy of hash(rand()), which might be less than just rand().

zeldovich: My understanding is that, yes, this is sufficiently uniform given the scale of values we are considering (signedWeight being tiny compared to the size of the hash).

crypto/compactcert/verifier.go
```diff
@@ -0,0 +1,96 @@
...
+// and participants that were used to construct the Verifier.
+func (v *Verifier) Verify(c *Cert) error {
+	if c.SignedWeight < v.ProvenWeight {
+		return fmt.Errorf("cert signed weight %d < proven weight %d", c.SignedWeight, v.ProvenWeight)
+	}
+
+	// Verify all of the reveals
+	sigs := make(map[uint64]crypto.Hashable)
+	parts := make(map[uint64]crypto.Hashable)
+	for i, r := range c.Reveals {
```
algomaxj, line 54: Suppose `Reveals` contains entries such that `r.Pos` is some fixed value over and over, `r.Part` is always the correct entry for that position, and `r.SigSlot` has `L`s increasing by the weight of `r.Part.Weight` each time. In other words, imagine we fill the cert with the same guy until it exceeds `v.ProvenWeight`.

Then we can build a `SigCommit` such that the `SigCommit`/`SigProofs` check succeeds (since `SigCommit` comes from the certificate), and we'll only verify the one position we chose in `v.partcom`. Then, it looks like the `hashCoin` checks should succeed too.

Where does this break?

zeldovich: I think you're right: if the cert has multiple reveals for the same `r.Pos`, the code only checks the last one!  That is indeed pretty horrible.  I believe it should suffice to check that `r.Pos` is not already in `sigs`, in this first loop.  Does that sound right to you?

algomaxj: I think that's right -- or maybe more intuitively (for me) that it's not already in `parts`, since `partcom` is constructed by the verifier. But I think functionality that is equivalent.

algomaxj:
There's something about the security parameters that I find confusing here.

If I claim `signedWeight` is `1<<64 - 1`, and `provenWeight` is `2 * 1000**5` (taken from `TestNumReveals`), then [`numReveals` works out to 10 when `secKQ =128`.](https://play.golang.org/p/CkpEa6RxA1R)

Say I have 10 different accounts, each with a small amount of money. I fill the first 1/10th of the 2**64 slots with duplicate signatures from the first account, the second 1/10th with signatures from the second account, etc. Now I just need enough malleability to retry the proving process until I get coins selected once from each 1/10th.

If I'm not mistaken, the probability of choosing once from each 1/10th then works out to `1 / 10! = 1 / 3,628,800` which seems really not very unlikely? One way to rerandomize would just be to decrease `signedWeight` by one microalgo until you succeed.

algomaxj:
(edited the above to remove an incorrect calculation)

algomaxj:
I think the probability would be something like `10! / 10**10 ~= 0.00036288`, since you can pick any 1/10th to start, then 9 of the 10, then 8 of the 10 and so on.

zeldovich:
I wonder if we should enforce that `signedWeight` cannot claim to be higher than the `CompactCertVotersTotal` (not in this PR, but it's just the sum of weights of all participants in the participants merkle tree)..  I don't see any downside to that check (it should not preclude any legitimate sets of signatures), and might avoid some degenerate cases.

But to your proposed attack..  I think the adversary would need to pick some specific `L` value in each signature.  Since the weight of each signature is some small `epsilon` (enforced by the accounts having a small amount of money in the participants commitment), the pseudorandomly-chosen coin would need to fall in the interval `[L, L+epsilon)`.  For a small `epsilon`, I think the adversary would need a lot of tries to get this right.

justicz:
Where is it enforced that the attacker would need to pick a particular single `L` value, and not just fill up a huge interval as in the first attack?

justicz:
More concretely... say I have 10 accounts A, B, C... each with 1 microalgo.

I fill the first `(2**64-1) / 10` slots in my `SigCommit` tree with `A`'s signature and `L` values 0, 1, 2, ... `((2**64-1) / 10) - 1`

I fill the second `(2**64-1) / 10` slots in my `SigCommit` tree with `B`'s signature and `L` values `((2**64-1) / 10)`, `((2**64-1) / 10) + 1`, `((2**64-1) / 10) + 2`, `2 * ((2**64-1) / 10) - 1`

And so on.

Now when I pick my 10 random coins, if I get `10! / 10**10` lucky, haven't I selected once from each of the 10 keys, so I won't detect a collision?

justicz:
If this attack works, is it because we should be enforcing that the structure of the signature tree matches with the structure of the balance tree? Should there only be one valid `Pos` for a given account?

zeldovich:
The first loop in `Verifier.Verify()` ensures that each revealed signature is at the same position (`r.Pos`) as the corresponding participant public key in the participants array.  So there's only one useful `L` value that the adversary can use for each account whose signing key they control, for a given Merkle tree.

justicz:
Ah, OK, I see now. I had misunderstood/forgotten how `Pos` was being used by the underlying `merklearray` library. Thanks for clarifying!

crypto/compactcert/builder.go
zeldovich: Added a test that confirms the implementation is correct.

tsachiherman:
You'll need to rebase it with master and change it from a draft to a PR before it can be reviewed, approved and merged.

zeldovich:
Rebased against the merged merklearray PR

crypto/compactcert/builder.go
```diff
@@ -0,0 +1,244 @@
...
+	b.sigs[pos].part = p
+	b.sigs[pos].Sig.OneTimeSignature = sig
+	b.signedWeight += p.Weight
+	b.cert = nil
+	return nil
+}
+
+// Ready returns whether the certificate is ready to be built.
+func (b *Builder) Ready() bool {
+	return b.signedWeight >= b.Params.ProvenWeight
```
reyzin, line 113: >= should probably be >. When =, you are guaranteed to fail (unless k+q=0, which should never be the case).

crypto/compactcert/builder.go
```diff
@@ -0,0 +1,244 @@
...
+}
+
+// Build returns a compact certificate, if the builder has accumulated
+// enough signatures to construct it.
+func (b *Builder) Build() (*Cert, error) {
+	if b.cert != nil {
+		return b.cert, nil
+	}
+
+	if b.signedWeight < b.Params.ProvenWeight {
```
reyzin, line 171: < should probably be <= (see https://github.com/algorand/go-algorand/pull/1359/files#r477474352)

crypto/compactcert/builder.go
reyzin, line 144: Since coinIndex is iterative and not recursive, there is no need to have lo and hi as arguments. You can initialize them inside coinIndex, making the caller's life a bit easier.

crypto/compactcert/builder.go
```diff
@@ -0,0 +1,244 @@
...
+	"github.com/algorand/go-algorand/crypto/merklearray"
+	"github.com/algorand/go-algorand/data/basics"
+)
+
+//msgp:ignore sigslot
+type sigslot struct {
+	// part is the participant signing this message.  The participant
+	// information is tracked here for convenience, but it does not
+	// appear in the commitment to the sigs array.
+	part Participant
```
reyzin, line 32: I think having part in sigslot is confusing. The only thing you use for it is part.Weight (lines 75, 156, and 177), which is either 0 or participants[relevant index].Weight). Why not just have sigslot.Weight, initlized to 0, and replace line 104 with b.sigs[pos].Weight = p.Weight?

crypto/compactcert/builder.go
```diff
@@ -0,0 +1,244 @@
...
+}
+
+// coinIndex returns the position pos in the sigs array such that the sum
+// of all signature weights before pos is less than or equal to coinWeight,
+// but the sum of all signature weights up to and including pos exceeds
+// coinWeight.
+//
+// coinIndex works by doing a binary search on the sigs array.  The caller
+// should invoke it with lo=0 and hi=len(b.sigs).  The caller should make
+// sure that sigs[*].L is initialized before using coinIndex().
```
reyzin, line 143: This comment is a bit confused. Not only should it be initialized, it should also be current (an Add will ruin it). Low-key solution: update the comment. Belt-and-suspenders solution in case someone changes the code later in a weird way: have a boolean in Builder that tracks whether .Ls are valid. It's initialized on false. Becomes true after lines 176-178 (though maybe those lines should be their own function). Becomes false in Add (just like b.cert = nil line 107).

crypto/compactcert/builder.go
```diff
@@ -0,0 +1,244 @@
...
+		for _, r := range c.Reveals {
+			if r.Pos == pos {
+				alreadyRevealed = true
+			}
+		}
+
+		if alreadyRevealed {
+			continue
+		}
+
```
reyzin, line 221: 210-220 will require quadratic running time. No big deal at current sizes, but for some parameter settings we may end up with bigger certificates. May want to futureproof by first generating all the pos into an array of size nr, then sorting in n log n time, the deduplicating.

crypto/compactcert/builder.go
```diff
@@ -0,0 +1,244 @@
...
+	return b.signedWeight >= b.Params.ProvenWeight
+}
+
+// SignedWeight returns the total weight of signatures added so far.
+func (b *Builder) SignedWeight() uint64 {
+	return b.signedWeight
+}
+
+//msgp:ignore sigCommit
+type sigCommit []sigslot
```
reyzin, line 122: I find this name confusing -- sigCommit seems like a signature commitment

crypto/compactcert/builder.go
```diff
@@ -0,0 +1,244 @@
...
+	nr, err := b.numReveals(b.signedWeight)
+	if err != nil {
+		return nil, err
+	}
+
+	var proofPositions []uint64
+
+	for j := uint64(0); j < nr; j++ {
+		coin := hashCoin(j, c.SigCommit, c.SignedWeight)
+		pos, err := b.coinIndex(coin, 0, uint64(len(b.sigs)))
```
reyzin, line 201: BTW, thank you for realizing that simple binary search works here. My write-up had something more complicated for no good reason. Fixed in the write-up.

crypto/compactcert/builder.go
zeldovich: No reason.  Storing just the weight indeed seems cleaner; changed to that.

crypto/compactcert/builder.go
zeldovich: Indeed, I was expecting the `=` case to fail further down, but it seemed safe to allow either, and allowing `=` here seemed cleaner in terms of what must be enforced at this point.  It sounds like you'd prefer it to be `>` here?  I'd be OK with that; I don't have a strong preference.

crypto/compactcert/builder.go
zeldovich: What about `sigsArray` or `sigCommitArray`?

crypto/compactcert/builder.go
reyzin: Yeah, a slight preference. The formula for numReveals has the ratio of provenWeight to signedWeight raised to the power of numReveals, and that ratio better be <1 or else raising to the power doesn't help...

crypto/compactcert/builder.go
zeldovich: Indeed, that's a nice cleanup I should have done after converting it to be iterative.  Thanks!

crypto/compactcert/builder.go
zeldovich: I decided to store the reveals as a `map` in the `Cert` struct, instead of a list (slice).  This both addresses this point, and would have avoided the bug that Max spotted (where the previous list-based representation allowed an encoding to claim multiple reveals at the same position).

crypto/compactcert/builder.go
zeldovich: Indeed, seems easier to just add the runtime check (should be trivially cheap) rather than write a more elaborate comment.  Thanks!

crypto/compactcert/builder.go
zeldovich: Sure, fixed the two places you pointed out in `builder.go` as well as the corresponding check in `verifier.go`.

crypto/compactcert/builder.go
reyzin: sigsToCommit?

zeldovich: That seems clear; thanks.

crypto/compactcert/verifier.go
```diff
@@ -0,0 +1,96 @@
...
+	for j := uint64(0); j < nr; j++ {
+		coin := hashCoin(j, c.SigCommit, c.SignedWeight)
+		matchingReveal := false
+		for _, r := range c.Reveals {
+			if r.SigSlot.L <= coin && coin < r.SigSlot.L+r.Part.Weight {
+				matchingReveal = true
+				break
+			}
+		}
+
```
reyzin, line 89: Sort by L value and binary search to avoid quadratic running time? May be an overkill for the current sizes, but futureproofing?

zeldovich: I would worry about the (not huge, but also non-zero) added complexity introducing some subtle bug or corner case.  Partly because there isn't a built-in binary search tree in Go, so we'd need to implement it ourselves, and we have a slightly weird case of sorting these intervals (which potentially could be overlapping in arbitrary ways).

I suspect that if we have enough reveals to make this a serious consideration, these are no longer "compact" certificates. :-)

reyzin: We don't need a search tree. We just to need sort by L (equivalently, by pos -- maybe that can be done automatically, because pos is the key in the map?). And then perform the same binary search as when forming the cert. Now, it COULD be that the certificate is messed up and sorting by pos is not equivalent to sorting by L, or that intervals overlap, but that shouldn't happen in an honestly generated cert, and if we find yet another reason to reject a dishonestly generated one -- all the better. The binary search in builder code will not find the correct interval if it's not there. But I agree, this is not crucial.

crypto/compactcert/approxbigfloat.go
```diff
@@ -0,0 +1,90 @@
...
+
+	for (a.base & (1<<63)) == 0 {
+		a.base = a.base << 1
+		a.exp = a.exp - 1
+	}
+}
+
+// ge returns whether a>=b.
+func (a *approxBigFloat) ge(b *approxBigFloat) bool {
+	if a.exp > b.exp {
```
reyzin, line 48: I think this will fail on a=0, b=1, because a will have 0 exponent and b will have a negative exponent.

crypto/compactcert/approxbigfloat.go
```diff
@@ -0,0 +1,90 @@
...
+
+	a.base = hi
+	a.exp = a.exp + b.exp + 64
+
+	if (a.base & (1<<63)) == 0 {
+		a.base = (a.base << 1) | (lo >> 63)
+		a.exp = a.exp - 1
+	}
+}
+
```
reyzin, line 86: I think it won't multiply 0 or by 0 correctly.

crypto/compactcert/approxbigfloat.go
```diff
@@ -0,0 +1,90 @@
...
+	"fmt"
+	"math/bits"
+)
+
+// An approxBigFloat represents the number base*2^exp.  A canonical
+// representation is one where the highest bit of base is set, or
+// where base=0 and exp=0.  Every operation enforces canonicality
+// of results.
+type approxBigFloat struct {
+	base uint64
```
reyzin, line 29: I think the correct term for this is mantissa. Base in the case is 2, because 2^exp.

crypto/compactcert/approxbigfloat.go
zeldovich: Good catch!  I meant to disallow zero values, since we don't need to support them.  I added a runtime check to enforce that.

crypto/compactcert/approxbigfloat.go
zeldovich: Same as above: added a runtime check to avoid zero values.

crypto/compactcert/approxbigfloat.go
zeldovich: Fixed, thanks!

crypto/compactcert/common.go
```diff
@@ -0,0 +1,109 @@
...
+// 2^-k >= 2^q * (provenWeight / signedWeight) ^ numReveals
+//
+// which is equivalent to the following, which avoids any floating-point math:
+//
+// signedWeight ^ numReveals >= 2^(k+q) * provenWeight ^ numReveals
+//
+// We use bigFloat for this computation, which is deterministic but adds
+// an error factor (rounding-down) of at most 1-2^-31 at each multiplication step.
+// The second-order (multiplied) error factors are negligible, but the additive
+// error factors add up to at most a multiplicative factor of 1-numReveals*2^-31.
```
reyzin, line 62: I think it's not quite as good as this comment, but still good enough. I wrote up a careful analysis here https://github.com/algorand/compactcert/blob/master/compactcert.pdf section 8.1 (p.6). In particular, the analysis (method 2) shows a cheap way to provably avoid underestimating: subtract 2 from mantissa of signedWeight after its canonicalized.

zeldovich: Thanks for the careful analysis of this!

The error bounds on the security parameter seem small enough that it doesn't seem like method 1 vs method 2 matters much.  Do you think we should implement the `mantissa-2` trick?

reyzin: On the one hand, it's clear that what we are doing is fine. On the other hand, the argument for why it's fine is a bit subtle, and subtle security arguments are best avoided if possible. Just as a matter of simplicity. I am not a huge fan of my mantissa -2 solution, because it also involves a subtle argument. What about the following  simpler solution: add a boolean "roundUp" to setu64 and mul and set it to "false" for lhs / signedweight, and to "true" for rhs / provenweight. round Up will add 1 to the mantissa. The argument for why this is ok is much simpler: we round down on the side we are trying to prove is bigger, and round up on the side we are trying to prove is smaller, so we can't possible prove a false statement.

zeldovich: Good idea; explicit rounding seems much clearer.

Pushed an implementation of explicit rounding, with compile-time enforcement that we never mix rounding modes (through separate types for rounded-up and rounded-down values).

reyzin: Looks good! I like the type-enforced limitations. One suggestion: do you want to enforce it on ge? 
` func (a *bigFloatDn) ge(b *bigFloatUp)`

zeldovich: That seems like a nice idea.  Pushed it.

tsachiherman (approved):

PR #1361, opened 2020-08-06
implement tracking of top-N online accounts in the ledger

zeldovich:
This PR extends the ledger's account tracking to be able to efficiently get the top-N online accounts, for any round between the account base round and the latest round.

This will be used in a subsequent PR to periodically commit to a Merkle array of the top-N online accounts in the block header.

The most annoying complication in this PR has to do with our rewards scheme.  When you take rewards into account, then it's possible for there to be two accounts, A and B, where the balance of A is larger than B at some round R, and then the balance of A becomes smaller than the balance of B at a later round R'>R, even though no transactions executed on either account A or B.

This PR deals with the above complication by using a notion of a "normalized" balance, which is a hypothetical number of microalgos your account should have had at round 0 to accrue your current balance by the round in which your account was last modified.  Comparisons of these normalized balances are stable over time, and thus give us a deterministic way to compute the top-N accounts, so that all nodes agree on what they are.


zeldovich:
(Force-pushed to rebase against master, which had some conflicting changes.)

ledger/accountdb.go
```diff
@@ -63,6 +63,12 @@ var accountsSchema = []string{
 	`CREATE TABLE IF NOT EXISTS accountbase (
 		address blob primary key,
 		data blob)`,
+	`CREATE TABLE IF NOT EXISTS onlineaccountbase (
```
algomaxj, line 66: I think you may need to use the database migration system that was recently added: https://github.com/algorand/go-algorand/blob/master/ledger/acctupdates.go#L789

ledger/accountdb.go
```diff
@@ -266,9 +273,77 @@ func accountsInit(tx *sql.Tx, initAccounts map[basics.Address]basics.AccountData
...
+			return err
+		}
+
+		if data.Status != basics.Online {
+			continue
+		}
+
+		onlineData := accountDataToOnline(&data)
+		normBalance := onlineData.normBalance(proto)
+		_, err = tx.Exec("INSERT INTO onlineaccountbase (address, normalizedbalance, data) VALUES (?, ?, ?)",
```
tsachiherman, line 325: Consider using a prepared statement for this one.

ledger/accountdb.go
```diff
@@ -646,6 +777,25 @@ func accountsNewRound(tx *sql.Tx, updates map[basics.Address]accountDelta, creat
 			return
 		}
 
+		// Delete online accounts if they are no longer online.
+		// This takes advantage of the fact that a zero account
+		// value has Status=Offline, so this covers both deletion
+		// of accounts as well as switching to non-online state.
+		if data.old.Status == basics.Online && data.new.Status != basics.Online {
```
tsachiherman, line 784: this is a correct predicate, as it can safely skip the test for data.old.IsZero() because 0 == basics.Offline

ledger/accountdb_test.go
```diff
@@ -369,6 +372,53 @@ func checkAccounts(t *testing.T, tx *sql.Tx, rnd basics.Round, accts map[basics.
...
+		if data.Status == basics.Online {
+			onlineAccounts[addr] = accountDataToOnline(&data)
+		}
+	}
+
+	proto := config.Consensus[protocol.ConsensusCurrentVersion]
+	for i := 0; i < len(onlineAccounts); i++ {
+		dbtop, err := accountsOnlineTop(tx, 0, uint64(i))
+		require.NoError(t, err)
+		require.Equal(t, len(dbtop), i)
```
tsachiherman, line 387: nit: swap arguments of require.Equal : 
```
require.Equal(t, i, len(dbtop))
```
in case of an error, it would printout that the second argument is the "expected" and the third is the "actual"

ledger/accountdb.go
```diff
@@ -554,6 +629,50 @@ func (qs *accountsDbQueries) close() {
 	}
 }
 
+// accountsOnlineTop returns the top n online accounts starting at position offset
+// (that is, the top offset'th account through the top offset+n-1'th account).
+//
+// The accounts are sorted by their normalized balance and address.  The normalized
+// address has to do with the reward parts of online account balances.  See the
```
tsachiherman, line 636: "normalized address" -> "normalized balance"

ledger/accountdb.go
```diff
@@ -554,6 +629,50 @@ func (qs *accountsDbQueries) close() {
...
+// normalization procedure in onlineAccount.normBalance().
+//
+// Note that this does not check if the accounts have a vote key valid for any
+// particular round (past, present, or future).
+func accountsOnlineTop(tx *sql.Tx, offset, n uint64) (map[basics.Address]*onlineAccount, error) {
+	rows, err := tx.Query("SELECT address, data FROM onlineaccountbase ORDER BY normalizedbalance DESC, address DESC LIMIT ? OFFSET ?", n, offset)
```
tsachiherman, line 642: the `accountsOnlineTop` function is only called in a loop, always starting from offset 0.
In that case, why do we need the secondary sort by address ?
i.e. if we need to ensure stable order across calls, it would suffice to use _rowid_ which would allow you to simplify the index

ledger/accountdb.go
tsachiherman: given that this is a caching table ( i.e. it doesn't introduces a new data to the database but rather a computed value ), it doesn't need to affect the catchpoint files in any way. however, when restoring a catchpoint file, we need to handle this. You could follow the existing patten used for tables such as `assetcreators`.

zeldovich:
I was initially worried that the full account state might be pretty large, compared to just the state needed for my purpose (online balance and the voting key), which is why I materialized a separate table to cache top online accounts.

But, I do like the simplicity of just adding a column to `accountbase`.  And we can always optimize it later, if the database queries start becoming slow, if need be.

So, I'll get rid of the separate table, and revise this PR accordingly.  Thank you for the suggestion!

ledger/accountdb.go
zeldovich: Because the sort order must be exactly the same across all computers running algod

zeldovich:
I pushed a revised version.  It ends up being a little shorter, but mostly because `msgp` no longer generates encoding logic for the `onlineAccount` struct.  But it is simpler in that the database only has one table that stores account state.

ledger/accountdb.go
```diff
@@ -269,6 +278,76 @@ func accountsInit(tx *sql.Tx, initAccounts map[basics.Address]basics.AccountData
 	return nil
 }
 
+// accountsAddNormalizedBalance adds the normalizedonlinebalance column
+// to the accountbase table.
+func accountsAddNormalizedBalance(tx *sql.Tx, proto config.ConsensusParams) error {
+	var exists bool
+	err := tx.QueryRow("SELECT 1 FROM pragma_table_info('accountbase') WHERE name='normalizedonlinebalance'").Scan(&exists)
+	if err == nil {
```
tsachiherman, line 286: With the database versioning, you dont really need to verify that the column present/missing, but it wouldn't be incorrect either.

ledger/accountdb.go
```diff
@@ -87,10 +87,19 @@ var creatablesMigration = []string{
...
+	`CREATE INDEX IF NOT EXISTS onlineaccountbals
+		ON accountbase ( normalizedonlinebalance, address, data )
+		WHERE normalizedonlinebalance>0`,
+}
+
 var accountsResetExprs = []string{
 	`DROP TABLE IF EXISTS acctrounds`,
 	`DROP TABLE IF EXISTS accounttotals`,
 	`DROP TABLE IF EXISTS accountbase`,
+	`DROP TABLE IF EXISTS onlineaccountbase`,
```
tsachiherman, line 102: this line can be dropped, right ?

ledger/accountdb.go
tsachiherman: ahh, because you scanning only a subset of the online accounts. Make sense.

tsachiherman:
> I pushed a revised version. It ends up being a little shorter, but mostly because `msgp` no longer generates encoding logic for the `onlineAccount` struct. But it is simpler in that the database only has one table that stores account state.

thanks. I think it looks much better, and more imporrant : will perform much better on a high tx rate.

ledger/accountdb.go
zeldovich: It turns out to be convenient because in the accountdb test code, I don't want to be invoking acctupdates code, but I still need some way to add the normalizedonlinebalance column.  Without this check, I would need some other way to determine if this column has been added or not.

ledger/accountdb.go
zeldovich: Oops, yes, good catch.

tsachiherman:
In onlineTop, why don’t you cache the most recent modification to any account, i.e.
```
modifiedAccounts := make(map[basics.Address]accountDelta)
```
and use these in order to update the candidates ? or maybe I'm missing something ?

i.e. either way you want to get the "latest" state for an address.

zeldovich:
That seems like a reasonable optimization.  Implemented and pushed.

tsachiherman:
The changes that you've made so far looks good, but we also need to have the corresponding catchpoint tables changes. The `catchpointbalances` is using the same schema as `accountbase`, so that when flipping the switch, we can perform a quick rename from `catchpointbalances`-> `accountbase`.

As is, there are few missing changes:
1. in `applyCatchpointStagingBalances` you'll want to recreate the index `onlineaccountbals`. The index will get deleted by `DROP TABLE IF EXISTS accountbase_old`, so you'll need to recreate it against the new table.
2. the `catchpointbalances` schema need to be extended in `resetCatchpointStagingBalances`
3. adding items to `catchpointbalances` is done by `writeCatchpointStagingBalances` where you'll want to re-implement your changes to `accountsNewRound`

The above does looks like lots of changes, but it really does boil down to handful of lines. Beside these, I think that the code looks good.

tsachiherman:
btw - regarding your idea to materialize some of the accounts onto a separate table :
I think that your intuition is correct. We should do that ( eventually ), but as a MRU to accelerate the access to frequently used accounts. This would also go hand by hand with some more advanced data packing ( I don't have a concrete plan for that yet ).

zeldovich:
@tsachiherman , thanks for the pointers to where these normalized online balances need to be added for catchpoints.

I pushed the changes to support that; it'd be great if you can take a look when you have time, to see if I missed something.

One difference from what you were proposing to do lies in how indexes are handled.  When a table is renamed, SQLite actually moves the indexes along with the table.  So, there's no need to explicitly drop indices.  However, we do need to create an index for the `catchpointbalances` table.  The annoyance is that SQLite has no way to rename an index.  Thus, there's this hack to create a unique name for the index (using the current timestamp).

tsachiherman:
> @tsachiherman , thanks for the pointers to where these normalized online balances need to be added for catchpoints.
> 
> I pushed the changes to support that; it'd be great if you can take a look when you have time, to see if I missed something.
> 
> One difference from what you were proposing to do lies in how indexes are handled. When a table is renamed, SQLite actually moves the indexes along with the table. So, there's no need to explicitly drop indices. However, we do need to create an index for the `catchpointbalances` table. The annoyance is that SQLite has no way to rename an index. Thus, there's this hack to create a unique name for the index (using the current timestamp).

Thanks; I won't be able to take a look of these changes this week, but will do so right after the release.

ledger/accountdb.go
```diff
@@ -136,60 +148,95 @@ func writeCatchpointStagingCreatable(ctx context.Context, tx *sql.Tx, addr basic
 	return nil
 }
 
-func writeCatchpointStagingBalances(ctx context.Context, tx *sql.Tx, bals []encodedBalanceRecord) error {
-	insertStmt, err := tx.PrepareContext(ctx, "INSERT INTO catchpointbalances(address, data) VALUES(?, ?)")
+func writeCatchpointStagingBalances(ctx context.Context, tx *sql.Tx, bals []encodedBalanceRecord, proto config.ConsensusParams) error {
```
tsachiherman, line 151: I'm pretty confident that this isn't what you want to do:
Using the genesis config.ConsensusParams here means that accounts that would be using fast catchup would calculate the normalized online balances based on the genesis params and not by the most recent params.

I would suggest you'll:
1. add the calculated normalized online balances to the catchpoint file
2. "restore" it in the same way, without recalculating anything.
3. add the calculated normalized online balances to the hashing of the account, so that the merkletrie would include that.
4. update the version of the catchpoint file, so that it would "break" compatibility with older files.

I have yet had the need to perform #4, so I don't know how hard it would be. ( shouldn't be too hard ).

I will be available to help you with this one if you'd like next week.

zeldovich: The `proto` is used only for the value of `proto.RewardUnit`, which cannot change in a given network, so it's OK to pass the proto for the genesis block or the proto for a recent block.

tsachiherman: In that case, we should always use the genesis round proto in the call to 
`err = accountsNewRound(tx, deltas[i], creatableDeltas[i], protos[i+1])` as well.

zeldovich: That seems like a reasonable idea; thanks!  Pushed.

tsachiherman (approved):
looks good to me.

PR #1362, opened 2020-08-06
do not use deadlock.Mutex in logging

zeldovich:
Helps avoid recursive locking when the deadlock package tries to report
recursive locking and the error message goes through the logging layer.

### DIFF
--- a/daemon/algod/deadlock_test.go
+++ b/daemon/algod/deadlock_test.go
@@ -17,52 +17,41 @@
 package algod
 
 import (
-	"bytes"
 	"fmt"
-	"os"
-	"runtime"
+	"testing"
+	"time"
 
 	"github.com/algorand/go-deadlock"
 
+	"github.com/algorand/go-algorand/crypto"
 	"github.com/algorand/go-algorand/logging"
 )
 
-type dumpLogger struct {
-	logging.Logger
-	*bytes.Buffer
-}
+func TestDeadlockLogging(t *testing.T) {
+	logFn := fmt.Sprintf("/tmp/test.%s.%d.log", t.Name(), crypto.RandUint64())
+	archiveFn := fmt.Sprintf("%s.archive", logFn)
 
-func (logger *dumpLogger) dump() {
-	logger.Error(logger.String())
-}
+	l := logging.Base()
+	logWriter := logging.MakeCyclicFileWriter(logFn, archiveFn, 65536, time.Hour)
+	l.SetOutput(logWriter)
 
-var logger = dumpLogger{Logger: logging.Base(), Buffer: bytes.NewBuffer(make([]byte, 0))}
+	setupDeadlockLogger()
 
-var deadlockPanic func()
-
-func setupDeadlockLogger() {
+	deadlockCh := make(chan struct{})
 	deadlockPanic = func() {
-		logger.Panic("potential deadlock detected")
+		close(deadlockCh)
 	}
 
-	deadlock.Opts.LogBuf = logger
-	deadlock.Opts.OnPotentialDeadlock = func() {
-		// Capture all goroutine stacks
-		var buf []byte
-		bufferSize := 256 * 1024
-		for {
-			buf = make([]byte, bufferSize)
-			if runtime.Stack(buf, true) < bufferSize {
-				break
-			}
-			bufferSize *= 2
+	var mu deadlock.RWMutex
+	defer func() {
+		r := recover()
+		if r != nil {
+			fmt.Printf("Recovered: %v\n", r)
 		}
+	}()
 
-		// Run this code in a separate goroutine because it might grab locks.
-		go func() {
-			logger.dump()
-			fmt.Fprintln(os.Stderr, string(buf))
-			deadlockPanic()
-		}()
-	}
+	mu.RLock()
+	mu.RLock()
+
+	_ = <- deadlockCh
 }


tsachiherman:
It's not 100% clear to me where the issue was triggered : 
if the issue was triggered after the logging mutex was taken, than this won't help, since the sync.Mutex retain the same double-taking lock properties. ( i.e. no reference counting for the same goroutine )
If the issue is that we cannot log a recursive locking error while the logging is already taking the lock ( for other reason ), 
then I would try to decouple that :
in /Users/tsachi/go/src/github.com/algorand/go-algorand/daemon/algod/deadlockLogger.go
we should have the deadlock error messages being sent into the logger by a separate non-blocking channel.

zeldovich:
Here's a slightly contrived example change to illustrate the bug:

```
--- a/node/node.go
+++ b/node/node.go
@@ -318,6 +318,11 @@ func (node *AlgorandFullNode) Config() config.Local {
 
 // Start the node: connect to peers and run the agreement service while obtaining a lock. Doesn't wait for initial sync.
 func (node *AlgorandFullNode) Start() {
+       var foo deadlock.RWMutex
+
+       foo.RLock()
+       foo.RLock()
+
        node.mu.Lock()
        defer node.mu.Unlock()
 
```

With this PR, when I run `algod`, I quickly get a nice `panic` message.

Without this PR, when I run `algod`, it locks up.  After `killall -ABRT algod`, the stack trace shows the locked-up thread is deadlocked in the logging code:

```
goroutine 1 [semacquire]:
sync.runtime_SemacquireMutex(0xc0000c0044, 0x40e300, 0x1)
	runtime/sema.go:71 +0x49
sync.(*Mutex).lockSlow(0xc0000c0040)
	sync/mutex.go:138 +0xfe
sync.(*Mutex).Lock(...)
	sync/mutex.go:81
github.com/algorand/go-deadlock.(*lockOrder).storeLockOrder(0xc0000c0040, 0x4, 0x1c3c0e0, 0xc0003542e0, 0x1)
	github.com/algorand/go-deadlock@v0.2.1/deadlock.go:384 +0x34b
github.com/algorand/go-deadlock.(*lockOrder).preLock(0xc0000c0040, 0x4, 0x1c3c0e0, 0xc0003542e0, 0x1, 0x0)
	github.com/algorand/go-deadlock@v0.2.1/deadlock.go:409 +0xd3
github.com/algorand/go-deadlock.preLock(...)
	github.com/algorand/go-deadlock@v0.2.1/deadlock.go:173
github.com/algorand/go-deadlock.lock(0xc0006be8d0, 0x1c3c0e0, 0xc0003542e0, 0xc0002f6300)
	github.com/algorand/go-deadlock@v0.2.1/deadlock.go:199 +0x94
github.com/algorand/go-deadlock.(*Mutex).Lock(0xc0003542e0)
	github.com/algorand/go-deadlock@v0.2.1/deadlock.go:80 +0xc1
github.com/algorand/go-algorand/logging.(*logBuffer).append(0xc0003542c0, 0xc0000f9200, 0x831)
	github.com/algorand/go-algorand@/logging/logBuffer.go:47 +0x45
github.com/algorand/go-algorand/logging.writerTee.Write(0x1c3be60, 0xc0000cd8c0, 0xc0003542c0, 0xc0000f8900, 0x831, 0x900, 0x433cfc, 0x0, 0xffffffffffffffff)
	github.com/algorand/go-algorand@/logging/logBuffer.go:106 +0x70
github.com/sirupsen/logrus.(*Entry).write(0xc0009185a0)
	github.com/sirupsen/logrus@v1.0.5/entry.go:134 +0x1b8
github.com/sirupsen/logrus.Entry.log(0xc00012a120, 0xc0009164b0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x7fa800000002, ...)
	github.com/sirupsen/logrus@v1.0.5/entry.go:104 +0x238
github.com/sirupsen/logrus.(*Entry).Error(0xc000918550, 0xc00088bec8, 0x1, 0x1)
	github.com/sirupsen/logrus@v1.0.5/entry.go:169 +0x111
github.com/sirupsen/logrus.(*Entry).Errorln(0xc000918550, 0xc00088bf20, 0x2, 0x2)
	github.com/sirupsen/logrus@v1.0.5/entry.go:264 +0xc8
github.com/algorand/go-algorand/logging.logger.Error(0xc000248000, 0xc000138010, 0xc0006be5e0, 0x1, 0x1)
	github.com/algorand/go-algorand@/logging/log.go:223 +0xea
github.com/algorand/go-algorand/daemon/algod.(*dumpLogger).dump(...)
	github.com/algorand/go-algorand@/daemon/algod/deadlockLogger.go:36
github.com/algorand/go-algorand/daemon/algod.setupDeadlockLogger.func1()
	github.com/algorand/go-algorand@/daemon/algod/deadlockLogger.go:44 +0x103
github.com/algorand/go-deadlock.(*lockOrder).printRecursiveLocking(0xc0000c0040, 0x1, 0xc, 0xc00088c140, 0x6, 0x32, 0xc0009231e0, 0x5, 0x32, 0x1c3c100, ...)
	github.com/algorand/go-deadlock@v0.2.1/deadlock.go:314 +0x2b8
github.com/algorand/go-deadlock.(*lockOrder).checkRecursiveLocking(0xc0000c0040, 0x4, 0x1c3c100, 0xc00013e0a0, 0x1)
	github.com/algorand/go-deadlock@v0.2.1/deadlock.go:348 +0x248
github.com/algorand/go-deadlock.(*lockOrder).preLock(0xc0000c0040, 0x4, 0x1c3c100, 0xc00013e0a0, 0x1, 0x1)
	github.com/algorand/go-deadlock@v0.2.1/deadlock.go:405 +0x67
github.com/algorand/go-deadlock.preLock(...)
	github.com/algorand/go-deadlock@v0.2.1/deadlock.go:173
github.com/algorand/go-deadlock.lock(0xc0006be2d0, 0x1c3c100, 0xc00013e0a0, 0xc00013e001)
	github.com/algorand/go-deadlock@v0.2.1/deadlock.go:199 +0x94
github.com/algorand/go-deadlock.(*RWMutex).RLock(0xc00013e0a0)
	github.com/algorand/go-deadlock@v0.2.1/deadlock.go:152 +0xc1
github.com/algorand/go-algorand/node.(*AlgorandFullNode).Start(0xc00011ac00)
	github.com/algorand/go-algorand@/node/node.go:324 +0x68
github.com/algorand/go-algorand/daemon/algod.(*Server).Start(0xc00088dc10)
	github.com/algorand/go-algorand@/daemon/algod/server.go:153 +0x136
main.main()
	github.com/algorand/go-algorand@/cmd/algod/main.go:362 +0x134f
```

tsachiherman:
Yes, that was exactly what I meant in my second if clause above.
The `lockorder` class ( which is a global struct/lock inside the deadlock library ) is sending a log message while it's internal lock is taken. As a result, any other lock cannot take the `lockorder` lock - including the one used by the logging component.
To work around that, we need to break that locking dependency:
have the deadlock library callback go to a function that would never take a lock, and return right away. The function can spin a separate go-routine which would perform the logging ( which would happen only in case of a DEADLOCK alerts, so it's not something that we expect to see frequently ).

tsachiherman:
btw - the above example can be a good starting point for a unit test to confirm the correctness of the deadlock "linkage" with the logger.

tsachiherman:
Just to clarify, I meant something like this : https://github.com/algorand/go-algorand/compare/master...tsachiherman:tsachi/loggingdeadlock?expand=1

zeldovich:
Thanks for the suggestion.  I've revised this PR to take your suggested approach.

tsachiherman (approved):
I'm not a big fan of the global variable ( since there tend to lead to race issues), but given that it's being used only in one test, I think it's good enough.

PR #1363, opened 2020-08-06
Prevent warning message for block assembly when catching up

tsachiherman:
## Summary

During catchup, both the agreement and the transaction pool are being informed of new blocks. The transaction pool might be processing the blocks faster than the agreement messaging works, and as a result, the agreement might request the transaction pool to assemble a block for an older round.

Currently, the transaction pool would respond with an error, which would propagate to the agreement protocol, having the latter print a warning message. This is a benign message, as it does not indicate of a concrete issue but rather a normal system operation.

To resolve that, this PR add a special error returned by the AssembleBlock to indicate that the requested block round is too old. The agreement, in turn, would check this error code and would verify that "claim" against the ledger's round. If this is legit, the proposal generation for that round would be skipped. Otherwise, an error message would be printed out, indicating of the issue.

### DIFF
--- a/agreement/pseudonode.go
+++ b/agreement/pseudonode.go
@@ -266,7 +266,19 @@ func (n asyncPseudonode) makeProposals(round basics.Round, period period, accoun
 	deadline := time.Now().Add(AssemblyTime)
 	ve, err := n.factory.AssembleBlock(round, deadline)
 	if err != nil {
-		n.log.Errorf("pseudonode.makeProposals: could not generate a proposal for round %d: %v", round, err)
+		if err == ErrAssembleBlockRoundStale {
+			// check to see if we're truely stale.
+			ledgerNextRound := n.ledger.NextRound()
+			if ledgerNextRound == round {
+				// we've asked for the right round.. and the ledger doesn't think it's stale.
+				n.log.Errorf("pseudonode.makeProposals: could not generate a proposal for round %d, ledger and proposal generation are synced: %v", round, err)
+			} else if ledgerNextRound < round {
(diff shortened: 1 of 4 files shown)


agreement/pseudonode.go
```diff
@@ -266,7 +266,19 @@ func (n asyncPseudonode) makeProposals(round basics.Round, period period, accoun
 	deadline := time.Now().Add(AssemblyTime)
 	ve, err := n.factory.AssembleBlock(round, deadline)
 	if err != nil {
-		n.log.Errorf("pseudonode.makeProposals: could not generate a proposal for round %d: %v", round, err)
+		if err == ErrAssembleBlockRoundStale {
+			// check to see if we're truely stale.
+			ledgerNextRound := n.ledger.NextRound()
+			if ledgerNextRound == round {
```
algomaxj, line 272: I would get rid this check and the `else` block, and keep the changes to `agreement` as small as possible (i.e. `if err != ErrAssembleBlockRoundStale`, do the log)

This more complicated checks can be moved into `pool.AssembleBlock`, where you can compare against something like `pool.ledger.Latest()`

tsachiherman: If not under the `agreement`, then the `pool` isn't the best either. What about the AlgorandFullNode.AssembleBlock ? that's there the three component are joined.

algomaxj: That sounds like a good option to me -- as a general rule I just want to keep the agreement changes as small as possible so that we don't have to think about them (even though this one is probably benign).

PR #1364, opened 2020-08-06
e2e test for rekeying before and after upgrade

algorandskiy:
## Summary

e2e-go test for rekeying feature before and after upgrade

## Test Plan

This is a test

### DIFF
--- a/test/e2e-go/upgrades/rekey_support_test.go
+++ b/test/e2e-go/upgrades/rekey_support_test.go
@@ -1,0 +1,125 @@
+// Copyright (C) 2019-2020 Algorand, Inc.
+// This file is part of go-algorand
+//
+// go-algorand is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as
+// published by the Free Software Foundation, either version 3 of the
+// License, or (at your option) any later version.
+//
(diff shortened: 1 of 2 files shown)


test/e2e-go/upgrades/rekey_support_test.go
```diff
@@ -0,0 +1,121 @@
...
+		curStatus, err = client.Status()
+		require.NoError(t, err)
+
+		require.Less(t, int64(time.Now().Sub(startLoopTime)), int64(3*time.Minute))
+		time.Sleep(time.Duration(roundTimeMs) * time.Millisecond)
+		round = curStatus.LastRound
+	}
+
+	// now, that we have upgraded to the new protocol which supports rekey, try again.
+	_, err = client.BroadcastTransaction(rekey)
```
tsachiherman, line 116: I'm surprised you did not see any failures : The ordering of transaction within a block are not guaranteed.
You should wait until the first transaction is included in a block before sending the subsequent one.

algorandskiy: I guess they go into the same block since there no other traffic. I'll put them into a group

algorandskiy: ok, waiting is easier. Fixed.

PR #1367, opened 2020-08-07
Ensure correct genesis file is provided to node packages

bricerisingalgorand:
## Summary

For some reason it was trying to parse the parent branch of the current branch and that wasn't working. I'm not sure that would ever work to provide the correct genesis file for betanet, so I updated the script to return betanet for any commit in the rel/beta branch.

## Test Plan

I built a node package for rel/beta and master to confirm they both have their respective genesis files

## Urgency/ Impact
Without this calculation fix, branches can end up with the wrong genesis file, leading to a bunch of errors. Mostly internal issue but necessary to fix.

### DIFF
--- a/scripts/compute_branch_network.sh
+++ b/scripts/compute_branch_network.sh
@@ -14,9 +14,8 @@ if [ "${BRANCH}" = "rel/stable" ]; then
     exit 0
 fi
 
-#get parent of current branch
-#credit to https://stackoverflow.com/questions/3161204/find-the-parent-branch-of-a-git-branch
-BRANCHPARENT="$(git show-branch | grep '\*' | grep -v '${BRANCH}' | head -n1 | sed 's/.*\[\(.*\)\].*/\1/' | sed 's/[\^~].*//')"
+# Get branch that owns head commit
+BRANCHPARENT="$(git branch --contains HEAD | grep -v 'HEAD' | head -n1 | grep -Eo '\w+(\/\w*)*')"
 
 if [ "${BRANCHPARENT}" = "rel/stable" ]; then
     echo "testnet"
--- a/scripts/compute_branch_network.sh
+++ b/scripts/compute_branch_network.sh
@@ -15,7 +15,7 @@ if [ "${BRANCH}" = "rel/stable" ]; then
 fi
 
 # Get branch that owns head commit
-BRANCHPARENT="$(git branch --contains HEAD | grep -v 'HEAD' | head -n1 | grep -Eo '\w+(\/\w*)*')"
+BRANCHPARENT="$(git branch --contains HEAD | grep -v 'HEAD' | head -n1 | grep -Eo '\S+(\/\S*)*' | tail -1)"
 
 if [ "${BRANCHPARENT}" = "rel/stable" ]; then
     echo "testnet"


tsachiherman:
maybe I'm missing something, but it seems like we're repeating having the same issues over and over.
Eric already spend the time fixing it in https://github.com/algorand/go-algorand/commit/8840c1ff8b4bab1b6b462ec22f1dd9f904aa3b93
If we're going to fix it *again*, then I strongly suggest we'll find a way to automate a testing for that.

bricerisingalgorand:
This is kind of a difficult thing to write an automated test for since it requires knowledge of the git commit tree. The only way I can think to partially test this would be to clone another go-algorand repo outside the project, checkout each branch, copy over relevant scripts, commit changes and execute these scripts to see if the outputs are accurate. Even that test would be a slow flaky test that doesn't guarantee builds from individuals' machines. Do you have any ideas for alternative approaches?

tsachiherman:
> This is kind of a difficult thing to write an automated test for since it requires knowledge of the git commit tree. The only way I can think to partially test this would be to clone another go-algorand repo outside the project, checkout each branch, copy over relevant scripts, commit changes and execute these scripts to see if the outputs are accurate. Even that test would be a slow flaky test that doesn't guarantee builds from individuals' machines. Do you have any ideas for alternative approaches?

The test for that doesn't have to be shaped directly over this change. But given that we've already seen the same issue coming more than once, we should clearly find a way to detect it ourselves before it's getting publicly released, right ?

scripts/compute_branch_network.sh
```diff
@@ -14,9 +14,8 @@ if [ "${BRANCH}" = "rel/stable" ]; then
     exit 0
 fi
 
-#get parent of current branch
-#credit to https://stackoverflow.com/questions/3161204/find-the-parent-branch-of-a-git-branch
-BRANCHPARENT="$(git show-branch | grep '\*' | grep -v '${BRANCH}' | head -n1 | sed 's/.*\[\(.*\)\].*/\1/' | sed 's/[\^~].*//')"
+# Get branch that owns head commit
+BRANCHPARENT="$(git branch --contains HEAD | grep -v 'HEAD' | head -n1 | grep -Eo '\w+(\/\w*)*')"
```
egieseke, line 18: Does this work correctly if there are hyphens in the branch name?

bricerisingalgorand: Nice catch! I just updated it to work for branches with non letter characters

egieseke (approved):
Looks good.

btoll (approved):
I haven't tested this but I figure it can't hurt since the existing `BRANCHPARENT` often doesn't report the correct value.

onetechnical (changes requested):
I've been reviewing some test cases on this, and I'm not sure it quite does we want yet. Vet behavior changes versus:

- master
- feature branch (from master)
- rel/beta
- rel/beta PR (e.g. onetechnical/relbeta2.1.6)
- rel/stable
- rel/stable PR (e.g. onetechnical/relstable2.1.6)

What I've seen is that the answers are strange.

Brice and I kicked around the following as a potential:

`git branch --sort=committerdate --contains HEAD | grep -v 'HEAD' | grep -Eo '(master|rel/beta|rel/stable)' | head -n1`

However, this does have behavioral changes, so may not be quite right. I think this PR (surprisingly) needs to cook a little more.

onetechnical:
While we are mucking with BRANCHPARENT, we should look at this potential problem:

```
% git show-branch | grep '\*'
+++++++++++    ++++++ [release/2.1.2-beta~15] Update package*.yaml to reflect new syntax (#1365)
 % BRANCHPARENT="$(git show-branch | grep '\*' | grep -v '${BRANCH}' | head -n1 | sed 's/.*\[\(.*\)\].*/\1/' | sed 's/[\^~].*//' || ${BRANCH})"
% echo $BRANCHPARENT
release/2.1.2-beta
```

This will accidentally match because it is trapping the asterisk character.

btoll:
We're moving in the direction of explicitly passing all needed values to the build instead of having them be computed at runtime.  The focus of this ticket is on the most troublesome one.  If we prioritize moving forward with the former method, this ticket becomes moot.

Perhaps we shouldn't put more work into this ticket in favor of hardening the package build pipeline.  See this ticket for a breakdown of how that can happen: https://github.com/algorand/go-algorand-internal/issues/855

btoll:
I'm closing this in favor of https://github.com/algorand/go-algorand-internal/issues/855.

#855 will fix this particular issue as well as some others, and that will make the build more deterministic.

PR #1369, opened 2020-08-07
e2e test: create+delete create+clear in 1 tx

algonautshant:
create + delete in one tx should pass
create + clear in one tx should fail


test/scripts/e2e_subs/e2e-app-other.sh
```diff
@@ -0,0 +1,40 @@
+#!/bin/bash
+
+date '+keyreg-teal-test start %Y%m%d_%H%M%S'
```
algorandskiy, line 3: not a rekey test

test/scripts/e2e_subs/e2e-app-other.sh
```diff
@@ -0,0 +1,40 @@
...
+
+WALLET=$1
+
+gcmd="goal -w ${WALLET}"
+
+ACCOUNT=$(${gcmd} account list|awk '{ print $3 }')
+
+# approval program
+printf '#pragma version 2\nint 1' > "${TEMPDIR}/simple.teal"
+PROGRAM=($(${gcmd} clerk compile "${TEMPDIR}/simple.teal"))
```
algorandskiy, line 18: isn't too many parenthesis?

test/scripts/e2e_subs/e2e-app-other.sh
algonautshant: Fixed. Also fixed other tests with the same error.

test/scripts/e2e_subs/e2e-app-other.sh
algonautshant: This statement is not needed anyway. Thanks for pointing it out.

PR #1370, opened 2020-08-07
Add Rekeying to pingpong

algorandskiy:
## Summary

Add Rekeying support to pingpoing utility

In each iteration send two txns:
 - from -> to with RekeyTo=to
 - to -> from with RekeyTo=from and AuthAddr=to
This preserves accounts' private keys for subsequent runs

## Test Plan

```
pingpong run -d mynet1 --randomnote=true --numaccounts=200 --run=20 --groupsize=2 --rekey=true
pingpong run -d mynet1 --randomnote=true --numaccounts=200 --run=20 --groupsize=2
```

Report the same 2300 txns per 20 sec of run time.

S1/S2/S3 are still in pending.

### DIFF
--- a/shared/pingpong/pingpong.go
+++ b/shared/pingpong/pingpong.go
@@ -25,7 +25,7 @@ import (
 	"time"
 
 	"github.com/algorand/go-algorand/crypto"
-	"github.com/algorand/go-algorand/daemon/algod/api/spec/v1"
+	v1 "github.com/algorand/go-algorand/daemon/algod/api/spec/v1"
 	"github.com/algorand/go-algorand/data/basics"
 	"github.com/algorand/go-algorand/data/transactions"
 	"github.com/algorand/go-algorand/libgoal"
@@ -234,13 +234,17 @@ func RunPingPong(ctx context.Context, ac libgoal.Client, accounts map[string]uin
 
 		var totalSent, totalSucceeded uint64
 		for !time.Now().After(stopTime) {
+			minimumAmount := cfg.MinAccountFunds + (cfg.MaxAmt+cfg.MaxFee)*2
+			fromList := listSufficientAccounts(accounts, minimumAmount, cfg.SrcAccount)
+			// in group tests txns are sent back and forth, so both parties need funds
+			if cfg.GroupSize == 1 {
+				minimumAmount = 0
+			}
+			toList := listSufficientAccounts(accounts, minimumAmount, cfg.SrcAccount)
 
-			fromList := listSufficientAccounts(accounts, cfg.MinAccountFunds+(cfg.MaxAmt+cfg.MaxFee)*2, cfg.SrcAccount)
-			toList := listSufficientAccounts(accounts, 0, cfg.SrcAccount)
-
-			sent, succeded, err := sendFromTo(fromList, toList, accounts, assetParam, appParam, ac, cfg)
+			sent, succeeded, err := sendFromTo(fromList, toList, accounts, assetParam, appParam, ac, cfg)
 			totalSent += sent
-			totalSucceeded += succeded
+			totalSucceeded += succeeded
 			if err != nil {
 				_, _ = fmt.Fprintf(os.Stderr, "error sending transactions: %v\n", err)
 			}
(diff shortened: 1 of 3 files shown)


shared/pingpong/pingpong.go
```diff
@@ -410,19 +463,16 @@ func sendFromTo(fromList, toList []string, accounts map[string]uint64, assetPara
 			sendErr = client.BroadcastTransactionGroup(stxGroup)
 		}
 
-		if sendErr != nil && !cfg.Quiet {
-			_, _ = fmt.Fprintf(os.Stderr, "error sending transaction: %v\n", sendErr)
-		} else {
```
algorandskiy, line 415: removed `if sendErr != nil && !cfg.Quiet {` branch since `!cfg.Quiet` does not make much sense because of outputting the same error below under `if sendErr != nil {`

algorandskiy:
Thanks Eric, I'm preparing PR to `go-algorand-ci` to enable this in `cicd/algonet/runPerformanceTests.py`

algorandskiy:
Local S1 run:
1. Baseline 
```
algonet play pingpong-start -a "--groupsize 2 --randomnote true --tps 100 --rest 0  --refresh 1 --numaccounts 500"

TPS: 373.35 BS: 999520.05 TPB: 1592.48 RT: 4.27 elapsed: 0:03:10.640638
```

2. Rekey (a bit less than regular payment groups)
```
algonet play pingpong-start -a "--groupsize 2 --rekey true --randomnote true --tps 100 --rest 0  --refresh 1 --numaccounts 500"

TPS: 360.16 BS: 984070.93 TPB: 1539.41 RT: 4.27 elapsed: 0:00:00.000007
```

tsachiherman:
Are the degraded results are due to the nature of the test, or the actual rekeying itself ?
followup in case it's the former - is there anything we can do to eliminate that factor ?

algorandskiy:
There are two possible causes:
1. There are some non-zero `InvalidCount` metrics in both tests that. Pingpong slows down significantly in case of txn submission error.
2. The results themselves are different from the official tests because I run pingpong only for 60 sec, then get aggregated stats from ES like the official runner, and then take maximum TPS seen. So here we need to compare baseline vs test to see the performance impact, and do not compare official results with these ones.

algorandskiy:
There are three kinds of error in logs for the baseline test:
```json
"BroadcastSignedTxGroup","level":"info","line":451,"msg":"rejected by local pool: TransactionPool.Remember: transaction already in ledger: M25TBH6GBZZZGKYQQUUPBULXQYVWMEODBGRSAC2PWVOKJVPXS72Q

"BroadcastSignedTxGroup","level":"info","line":451,"msg":"rejected by local pool: TransactionPool.Remember: txn dead: round 45 outside of 39--44

"BroadcastSignedTxGroup","level":"info","line":451,"msg":"rejected by local pool: TransactionPool.Test: transaction pool have reached capacity
```

Freq are 149, 345, 19. Not sure why `dead` happen, validity is set for 5 rounds. Same for `already` - random note field is enabled.

algorandskiy:
On master `groupsize=2` also produces lots of rejected txns so that degraded performance. Max TPS seen:
```
TPS: 382.26 BS: 999269.09 TPB: 1636.96 RT: 4.28 elapsed: 0:03:45.732888
```

I guess we are fine in terms of **Rekeying** impact to performance (minimal) but I still investigating the reason txns rejection in group tests.

algorandskiy:
### S1/S2 results

| Test          | TPS | Change | Dup txn | Dead txn |
| --------   | ---- | -------- | -------- | --------- |
| Base S1   | 787 | -             | 0            | 689 |
| Rekey S1 | 722 | 92%       | 186         | 603 |
| Base S2  | 799 | -             | 210        | 376 |
| Rekey S2 | 670 | 83%      | 337        | 591 |

baseline cmd
```
"--groupsize 2 --randomlease true  --ra true --tps 100 --rest 0  --refresh 1 --numaccounts 500"
```

test cmd
```
"--rekey true --groupsize 2 --randomlease true  --ra true --tps 100 --rest 0  --refresh 1 --numaccounts 500"
```

Sounds like duplicate txn contribute to pingpong slowdown and TSP reduce.

PR #1372, opened 2020-08-10
Eliminate nil map entries on Merkle trie cache

tsachiherman:
## Summary

The Merkle trie was using the `pageToNIDsPtr` map to associate in-memory cached pages of nodes with the actual nodes.
During trie rearrangements, nodes are being deleted, which could lead to deletion of entries from this map. The map was
correctly deleting the internal list object, but left it as "nil" instead of completely eliminating it from the map. Over time, this 
could result is a memory leak.

The solution is quite straight-forward. Instead of just placing `nil` value on the map when the associated page gets empty, delete the page completely.

## Test Plan

Unit tests added to cover the relevant changes.

### DIFF
--- a/crypto/merkletrie/cache.go
+++ b/crypto/merkletrie/cache.go
@@ -67,8 +67,8 @@ type merkleTrieCache struct {
 
 	// pendingCreatedNID contains a list of the node ids that has been created since the last commit and need to be stored.
 	pendingCreatedNID map[storedNodeIdentifier]bool
-	// pendingDeletionPage contains a map of pages to delete, each pointing to a map of all the nodes ids that need to be removed.
-	pendingDeletionPages map[uint64]map[storedNodeIdentifier]*node
+	// pendingDeletionPage contains a map of pages to delete once committed.
+	pendingDeletionPages map[uint64]bool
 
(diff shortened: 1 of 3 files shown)


tsachiherman:
> I think the deletion logic is fine.
> The tests are questionable: I'd expect to see checks for `pageToNIDsPtr` not containing nil elements, or a test that somehow triggers deletion of all the nodes and then checks for empty caches.

The 3 new tests were supposed to cover these exact use cases by testing 
require.NotZerof(t, len(pageContent), "Memory page %d has zero nodes", page)

algorandskiy:
True, it checks all entries in the cache has some values but does not check the cache size/emptiness. Like shouldn't the cache be empty after removing all hashes from the tree?

tsachiherman:
> True, it checks all entries in the cache has some values but does not check the cache size/emptiness. Like shouldn't the cache be empty after removing all hashes from the tree?

Not always. The cache stores the recent used pages. If we reached to a stage where the tree became empty, we still want the newest page in cache, so ghat when a new item is added, we wont need to reliad the page from the cache.

I agree that we could add some special cases (i.e. empty trie), but these wont be common on production, right ?

PR #1373, opened 2020-08-10
Update the suggested fee mechanism to meet the current transaction pool fee structure 

algorotem:
## Summary

Our current `SuggetestedFee` mechanism keeps the MA of the prices. This mechanism is no longer relevant because the minimum fee to enter the queue is based solely on the number of pending whole blocks in the system. This PR reflects it. 

## Test Plan
Remember many transactions and check that the fee matches the expected one as a function of the number of whole blocks.


algorotem:
> I think that you might be missing taking the mutex lock on the transaction pool. looks good otherwise.

I thought about it and here is my reasoning for not doing it and let me know if I'm wrong -
`CurrentFeePerByte` is a read-only operations method. 
The external vars it uses `feeThresholdMultiplier` and `numPendingWholeBlocks` are being updated by another two functions (`Remember` and `Add`) and those lock the entire TransactionPool before they run. 
Now, adding a lock to protect these two var or to lock the entire TP for every read would hurt the performance and it's too high of a price to pay for the no harmful stale result that you may get from a race condition. 

Thoughts?

tsachiherman:
> CurrentFeePerByte

The problem with a race condition is that it would randomly fail our testing. I agree that getting a stale result here isn't that bad of an outcome. How's about having an atomic value in the transaction pool that would hold the latest calculated `CurrentFeePerByte` ?

algorotem:
> > CurrentFeePerByte
> 
> The problem with a race condition is that it would randomly fail our testing. I agree that getting a stale result here isn't that bad of an outcome. How's about having an atomic value in the transaction pool that would hold the latest calculated `CurrentFeePerByte` ?

That's a good idea. I'll do it.

PR #1377, opened 2020-08-11
REST API test script for /v2/applications

winder:
## Summary

e2e script test to make sure the `/v2/applications` endpoint is working.

### DIFF
--- a/test/scripts/e2e_subs/rest.sh
+++ b/test/scripts/e2e_subs/rest.sh
@@ -43,7 +43,7 @@ function test_applications_endpoint {
   APPID=$(${gcmd} app create --creator "${ACCOUNT}" --approval-prog "${TEMPDIR}/simple.teal" --clear-prog "${TEMPDIR}/simple.teal" --global-byteslices 0 --global-ints 2 --local-byteslices 0 --local-ints 0 | grep Created | awk '{ print $6 }')
 
   # Good request, non-existant app id
-  call_and_verify "Should not find app." "/v2/applications/$(("$APPID" + 1))" 'application does not exist'
+  call_and_verify "Should not find app." "/v2/applications/987654321" 'application does not exist'
   # Good request
   call_and_verify "Should contain app data." "/v2/applications/$APPID" '"global-state-schema":{"num-byte-slice":0,"num-uint":2}'
   # Good request, pretty response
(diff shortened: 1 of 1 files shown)


test/scripts/e2e_subs/rest.sh
```diff
@@ -0,0 +1,68 @@
...
+WALLET=$1
+gcmd="goal -w ${WALLET}"
+ACCOUNT=$(${gcmd} account list|awk '{ print $3 }')
+
+# REST Parameters
+PUB_TOKEN=$(cat "$ALGORAND_DATA"/algod.token)
+ADMIN_TOKEN=$(cat "$ALGORAND_DATA"/algod.admin.token)
+NET=$(cat "$ALGORAND_DATA"/algod.net)
+
+function call_admin {
```
btoll, line 19: Looks like this is a duplicate of `call`?

test/scripts/e2e_subs/rest.sh
```diff
@@ -0,0 +1,68 @@
...
+# $1 - test description.
+# $2 - query
+# $3 - substring that should be in the response
+function call_and_verify {
+  local RES
+  RES=$(call "$2")
+  if [[ "$RES" == *"$3"* ]]; then
+    return 0
+  fi
+  echo "Failed test: $1"
```
btoll, line 36: Could you further disambiguate the failed test when they may have the same name, ala "Should contain app data."?

winder: The script uses `set -x`, so all the output should be printed, so we should be alright even without this line. When the test passes it gets hidden by the framework

btoll: ok

brianolson:
LGTM

test/scripts/e2e_subs/rest.sh
winder: Not yet, I wanted to make it handy for when the scope of these tests expand.

algorotem: tiny - since it's the same code as the regular one, `admin/pub` can simply be an argument or just have a wrapper to one function.

winder:
Sure, I'll get that change in my followup PR that adds the test for /v2/assets

PR #1382, opened 2020-08-11
Avoid using  encodedAccountsRange and replace with a single query

tsachiherman:
## Summary

The encodedAccountsRange function was becoming too slow to be used on production systems.
It's performance degraded with size of data. ( i.e. first iteration is 1ms, after 5000 iterations you could get to 60ms ).
Instead, I replaced it with a single, long living, query which iterates over all the accounts.
In order to avoid a separate go-routine, I had to make some additional handling to the call stack.

Note that this function was used both in the rebuilding o the Merkle trie as well as during the generation of the catchpoint file. Both shown substantial performance gains.

## Test Plan

### DIFF
--- a/ledger/accountdb.go
+++ b/ledger/accountdb.go
@@ -938,3 +938,66 @@ func (mc *merkleCommitter) LoadPage(page uint64) (content []byte, err error) {
 func (mc *merkleCommitter) GetNodesCountPerPage() (pageSize int64) {
 	return merkleCommitterNodesPerPage
 }
+
+type encodedAccountsIterator struct {
+	rows            *sql.Rows
+	orederByAddress bool
+}
+
+// encodedAccountsRange returns an array containing the account data, in the same way it appear in the database
+// starting at entry startAccountIndex, and up to accountCount accounts long.
+func (iterator *encodedAccountsIterator) Iterate(ctx context.Context, tx *sql.Tx, accountCount int) (bals []encodedBalanceRecord, err error) {
+	if iterator.rows == nil {
+		if iterator.orederByAddress {
+			iterator.rows, err = tx.QueryContext(ctx, "SELECT address, data FROM accountbase ORDER BY address")
+		} else {
+			iterator.rows, err = tx.QueryContext(ctx, "SELECT address, data FROM accountbase")
+		}
+
+		if err != nil {
+			return
+		}
+	}
+
+	// gather up to accountCount encoded accounts.
+	bals = make([]encodedBalanceRecord, 0, accountCount)
+	var addr basics.Address
+	for iterator.rows.Next() {
+		var addrbuf []byte
+		var buf []byte
+		err = iterator.rows.Scan(&addrbuf, &buf)
+		if err != nil {
+			iterator.Close()
+			return
+		}
+
+		if len(addrbuf) != len(addr) {
+			err = fmt.Errorf("Account DB address length mismatch: %d != %d", len(addrbuf), len(addr))
+			return
+		}
+
+		copy(addr[:], addrbuf)
+
+		bals = append(bals, encodedBalanceRecord{Address: addr, AccountData: buf})
+		if len(bals) == accountCount {
+			// we're done with this iteration.
+			return
+		}
+	}
+
+	err = iterator.rows.Err()
+	if err != nil {
+		iterator.Close()
+		return
+	}
+	// we just finished reading the table.
+	iterator.Close()
+	return
+}
+
+func (iterator *encodedAccountsIterator) Close() {
+	if iterator.rows != nil {
+		iterator.rows.Close()
+		iterator.rows = nil
+	}
+}
--- a/ledger/accountdb.go
+++ b/ledger/accountdb.go
@@ -898,16 +898,17 @@ func (mc *merkleCommitter) GetNodesCountPerPage() (pageSize int64) {
 	return merkleCommitterNodesPerPage
 }
 
-type encodedAccountsIterator struct {
-	rows            *sql.Rows
-	orederByAddress bool
+// encodedAccountsBatchIter allows us to iterate over the accounts data stored in the accountbase table.
+type encodedAccountsBatchIter struct {
+	rows           *sql.Rows
+	orderByAddress bool
 }
 
-// encodedAccountsRange returns an array containing the account data, in the same way it appear in the database
-// starting at entry startAccountIndex, and up to accountCount accounts long.
-func (iterator *encodedAccountsIterator) Iterate(ctx context.Context, tx *sql.Tx, accountCount int) (bals []encodedBalanceRecord, err error) {
+// Next returns an array containing the account data, in the same way it appear in the database
+// returning accountCount accounts data at a time.
+func (iterator *encodedAccountsBatchIter) Next(ctx context.Context, tx *sql.Tx, accountCount int) (bals []encodedBalanceRecord, err error) {
 	if iterator.rows == nil {
-		if iterator.orederByAddress {
+		if iterator.orderByAddress {
 			iterator.rows, err = tx.QueryContext(ctx, "SELECT address, data FROM accountbase ORDER BY address")
 		} else {
 			iterator.rows, err = tx.QueryContext(ctx, "SELECT address, data FROM accountbase")
@@ -954,7 +955,8 @@ func (iterator *encodedAccountsIterator) Iterate(ctx context.Context, tx *sql.Tx
 	return
 }
 
-func (iterator *encodedAccountsIterator) Close() {
+// Close shuts down the encodedAccountsBatchIter, releasing database resources.
+func (iterator *encodedAccountsBatchIter) Close() {
 	if iterator.rows != nil {
 		iterator.rows.Close()
 		iterator.rows = nil
--- a/ledger/accountdb.go
+++ b/ledger/accountdb.go
@@ -751,47 +751,6 @@ func updateAccountsRound(tx *sql.Tx, rnd basics.Round, hashRound basics.Round) (
 	return
 }
 
-// encodedAccountsRange returns an array containing the account data, in the same way it appear in the database
-// starting at entry startAccountIndex, and up to accountCount accounts long.
-func encodedAccountsRange(ctx context.Context, tx *sql.Tx, startAccountIndex, accountCount int) (bals []encodedBalanceRecord, err error) {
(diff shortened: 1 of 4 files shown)


ledger/accountdb.go
```diff
@@ -938,3 +897,66 @@ func (mc *merkleCommitter) LoadPage(page uint64) (content []byte, err error) {
...
 }
+
+type encodedAccountsIterator struct {
+	rows            *sql.Rows
+	orederByAddress bool
+}
+
+// encodedAccountsRange returns an array containing the account data, in the same way it appear in the database
+// starting at entry startAccountIndex, and up to accountCount accounts long.
+func (iterator *encodedAccountsIterator) Iterate(ctx context.Context, tx *sql.Tx, accountCount int) (bals []encodedBalanceRecord, err error) {
```
algorandskiy, line 908: Please update the comment, there is no `encodedAccountsRange` anymore.

Strictly speaking, `Iterator` usually has `Next/Advance` method, or accepts a callback. Here we return batch of all available elements. Suggestion: call it `encodedAccountsBatchIter` and rename `Iterate` to `Next` or `NextBatch`

algorandskiy, line 908: Please update the comment, there is no `encodedAccountsRange` anymore.

Strictly speaking, `Iterator` usually has `Next/Advance` method, or accepts a callback. Here we return batch of all available elements. Suggestion: call it `encodedAccountsBatchIter` and rename `Iterate` to `Next` or `NextBatch`

ledger/acctupdates.go
```diff
@@ -866,23 +877,41 @@ func (au *accountUpdates) accountsInitialize(ctx context.Context, tx *sql.Tx) (b
...
-			if err != nil {
-				return 0, fmt.Errorf("accountsInitialize was unable to commit changes to trie: %v", err)
+			if pendingAccounts >= trieRebuildCommitFrequency {
+				// this trie Evict will commit using the current transaction.
+				// if anything goes wrong, it will still get rolled back.
+				_, err = trie.Evict(true)
+				if err != nil {
+					return 0, fmt.Errorf("accountsInitialize was unable to commit changes to trie: %v", err)
+				}
+				pendingAccounts = 0
```
algorandskiy, line 887: Q: how is it related to the perf fix of `encodedAccountsRange`?

ledger/catchpointwriter.go
```diff
@@ -222,7 +226,7 @@ func (cw *catchpointWriter) WriteStep(ctx context.Context) (more bool, err error
 }
 
 func (cw *catchpointWriter) readDatabaseStep(ctx context.Context, tx *sql.Tx) (err error) {
```
algorandskiy, line 228: Q: must it be called with DB lock? if so, please add a comment

ledger/catchpointwriter.go
```diff
@@ -118,7 +122,7 @@ func (cw *catchpointWriter) Abort() error {
 	return err
 }
 
-func (cw *catchpointWriter) WriteStep(ctx context.Context) (more bool, err error) {
```
algorandskiy, line 121: Q: must it be called with DB lock? if so, please add a comment

ledger/acctupdates.go
```diff
@@ -1511,37 +1538,54 @@ func (au *accountUpdates) generateCatchpoint(committedRound basics.Round, label
...
-		if more && err == nil {
-			// we just wrote some data, but there is more to be written.
-			// go to sleep for while.
-			select {
-			case <-time.After(100 * time.Millisecond):
-			case <-au.ctx.Done():
-				retryCatchpointCreation = true
+
+	var catchpointWriter *catchpointWriter
+	err = au.dbs.rdb.Atomic(func(ctx context.Context, tx *sql.Tx) (err error) {
```
algorandskiy, line 1543: before we had more granular locks and possibly better performance, did not we?
Plus, I'm a bit concerned about `select` and channels under db locks, that might be a source of locking problems.

ledger/accountdb.go
tsachiherman: good idea. will do.

ledger/acctupdates.go
tsachiherman: separating the Evict frequency from the chunk size decouple the database reading chunk from the trie commit/evict calls. Calling Evict too frequently causes us to drop pages that are needed further on. We don't want to skip the Evict, since the host system might not have enough memory to process the entire trie. That separation was giving us about 40% performance increase.

ledger/catchpointwriter.go
tsachiherman: are you referring to `accountUpdates.accountsMu` ? if so, then the answer is "no".
The writer is called in exactly two cases:
1. during the startup ( loadFromDisk ), where the writing go-routine has yet to be created. So obviously, in this case there won't be any other resource competing over a lock.
2. during commitRound. Since the commitRound itself is the only database writer, there is no chance that another go-routine would write to the database.

ledger/acctupdates.go
tsachiherman: I think that we're still having the same locks as before. Using WAL file allows us to have multiple readers, so having this reader should not be an issue. I would have generally preferred to avoid long-living database connection, but given that it's in-proc database, I think that it might be ok.

ledger/acctupdates.go
```diff
@@ -50,7 +50,9 @@ const (
 	pendingDeltasFlushThreshold = 128
 	// trieRebuildAccountChunkSize defines the number of accounts that would get read at a single chunk
 	// before added to the trie during trie construction
-	trieRebuildAccountChunkSize = 512
+	trieRebuildAccountChunkSize = 16384
```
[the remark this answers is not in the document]
tsachiherman: The initial number was aligned with the one in `BalancesPerCatchpointFileChunk`, which seems to be reasonable at the time. However, when rebuilding the trie from scratch, there is no reason to load such a small blocks at a time. If we have enough memory, we should load larger set of account on each iteration. Note that I also added the `trieRebuildCommitFrequency` which decouples the above from the actual Eviction. Together, these would utilize more memory during the building of the trie, and would shorten the time it takes to build it.

ledger/acctupdates.go
```diff
@@ -1511,37 +1538,54 @@ func (au *accountUpdates) generateCatchpoint(committedRound basics.Round, label
...
-			return
-		}
+		return
+	})
+
+	if err != nil {
+		au.log.Warnf("accountUpdates: generateCatchpoint: %v", err)
+		return
+	}
+	if catchpointWriter == nil {
```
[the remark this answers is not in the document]
tsachiherman: It shouldn't. The `db.Atomic` is not supposed to return with nil unless the callback function was called.
However, since there could "thoererically" be a bug in the `Atomic` function, I figured it would be safer to add this no-op check.

ledger/catchpointwriter.go
```diff
@@ -133,19 +137,19 @@ func (cw *catchpointWriter) WriteStep(ctx context.Context) (more bool, err error
 	}
 
 	// have we timed-out / canceled by that point ?
-	if more, err = hasContextDeadlineExceeded(ctx); more == true || err != nil {
+	if more, err = hasContextDeadlineExceeded(stepCtx); more == true || err != nil {
```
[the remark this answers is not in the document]
tsachiherman: The variable is called `more` to indicate that more calls to this function are needed in order to complete the writing process to the file, which was "interrupted" by the context deadline.

ledger/catchpointwriter.go
algorandskiy: > are you referring to accountUpdates.accountsMu

Not sure, was talking about `au.dbs.rdb.Atomic`

tsachiherman: The `Atomic` afaik, doesn't really take any locks on it's own, unless you're referring to it's variant `AtomicCommitWriteLock`. This is needed only for write locks, since we want to ensure we synchronize the writing to the database to be atomic. However, since the writing is executed on the same goroutine as the catchpoint writer, there is no need to take the lock. ( we can take the read lock, but there are no gains in doing that either )

ledger/acctupdates.go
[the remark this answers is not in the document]
algonautshant: > there is no reason to load such a small blocks at a time. 
The way I am understanding is that the database operation in not impacted by trieRebuildAccountChunkSize. 
The database is accessed by Next/Scan calls, and it will follow it's on buffering mechanism. 
It will be all the same if trieRebuildAccountChunkSize is set to 1. What am I missing here?

ledger/catchpointwriter.go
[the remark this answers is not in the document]
algonautshant: Will it be fair to interpret this is maybe-more? 
Since hasContextDeadlineExceeded does not know that more calls to this function are needed. 
It might very well be that no more calls are needed, but it coincided with the timeout.

ledger/acctupdates.go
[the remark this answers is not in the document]
tsachiherman: We have millions of accounts in the database. I don't know if we have enough memory to load all of them in, so I was processing it in chunks. Do you believe that increasing this from 16K accounts data to 64K accounts data, for instance, would improve the performance ? It's true that we could have used "1" here, but then we're exercising (slightly) larger execution path.
In the future, I'm thinking about batch operation for the Merkle trie, where we would benefit from this implementation.

ledger/catchpointwriter.go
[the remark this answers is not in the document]
tsachiherman: If you'll look at the locations that call `hasContextDeadlineExceeded` you'll see that we basically always need more "time" in these cases. But we want to exit prematurely if the caller context exceed the timeout.

ledger/accountdb.go
```diff
@@ -938,3 +897,68 @@ func (mc *merkleCommitter) LoadPage(page uint64) (content []byte, err error) {
...
+
+		if len(addrbuf) != len(addr) {
+			err = fmt.Errorf("Account DB address length mismatch: %d != %d", len(addrbuf), len(addr))
+			return
+		}
+
+		copy(addr[:], addrbuf)
+
+		bals = append(bals, encodedBalanceRecord{Address: addr, AccountData: buf})
+		if len(bals) == accountCount {
```
algonautshant, line 942: This code looks correct. 
Do we have a test for checking the edge cases? Particularly, if the number of addresses is equal to accountCount i.e. trieRebuildAccountChunkSize? 
Such a test is more useful if in the future changes are made to this code.

ledger/acctupdates.go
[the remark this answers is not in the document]
algonautshant: Thank you for the explanation. I am convinced.

PR #1383, opened 2020-08-12
Added TestListCreatables to test ListAssets and ListApplications

algonautshant:
* tests with all elements in cache, all synced to database, and combination of both
* tests the max results, max app index and max asset index

### DIFF
--- a/ledger/acctupdates_test.go
+++ b/ledger/acctupdates_test.go
@@ -862,7 +862,7 @@ func listAndCompareComb(t *testing.T, au *accountUpdates, expected map[basics.Cr
 	for a, b := range expected {
 		if b.ctype == basics.AssetCreatable {
 			if maxAss2 < a {
-				maxAss1 = maxAss1
+				maxAss1 = maxAss2
 				maxAss2 = a
 			} else if maxAss1 < a {
 				maxAss1 = a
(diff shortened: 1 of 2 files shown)


ledger/acctupdates_test.go
```diff
@@ -845,3 +845,161 @@ func TestAcctUpdatesDeleteStoredCatchpoints(t *testing.T) {
...
+// It repeats with different combinations of the limit parameters
+func listAndCompareComb(t *testing.T, au *accountUpdates, expected map[basics.CreatableIndex]modifiedCreatable) {
+
+	// test configuration parameters
+
+	// pick the second largest index for the app and asset
+	// This is to make sure exactly one element is left out
+	// as a result of max index
+	maxAss1 := basics.CreatableIndex(0)
+	maxAss2 := basics.CreatableIndex(0)
```
tsachiherman, line 859: I don't see where you're using `maxAss2` and `maxApp2`. 
Also, I would have implemented this as a `map[CreatableType]basics.CreatableIndex`, so that if we extend the types of creatables, it would work as well.

algonautshant: maxAss2 is used to keep track of the last two largest elements. 
It will be used more in my next update. 

Will postpone the implementation of the map for now. It makes sense in the test, but addition of new types will impact many other interfaces.

PR #1384, opened 2020-08-12
[WIP] Refactor applications

derbear:
## Summary

Refactor applications to reduce cloning, which should improve efficiency relative to various application parameters.

WIP

## Test Plan

WIP

Co-authored-by: Max Justicz <max@algorand.com>


data/basics/teal.go
```diff
@@ -28,11 +28,11 @@ import (
...
 const (
-	// SetBytesAction indicates that a TEAL byte slice should be stored at a key
-	SetBytesAction DeltaAction = 1
-
 	// SetUintAction indicates that a Uint should be stored at a key
-	SetUintAction DeltaAction = 2
+	SetUintAction DeltaAction = 1
+
+	// SetBytesAction indicates that a TEAL byte slice should be stored at a key
+	SetBytesAction DeltaAction = 2
```
derbear, line 35: @justicz Why are these switched here?

justicz: Unintentionally -- these changed on master at some point and I must have accidentally copied this from an older version.

derbear:
@justicz Must `AccountData.IsZero` be updated to correctly handle the maps introduced by applications in `AccountData`?

justicz:
@derbear I think this has come up before -- I think the existing check is already redundant (though updating it shouldn't hurt).

- The current check isn't even complete as-is, since it's not checking `AssetParams`
- As far as I can see, the only place we are zeroing out accounts is during `CloseOut` payments where we set `rec.AccountData = basics.AccountData{}`. I don't think there is a place where we can have an account that is all empty save for a length zero, non-nil map entry.

PR #1388, opened 2020-08-12
Add tests for /v2/assets endpoint

winder:
## Summary

Add e2e tests for the `/v2/assets/{id}` endpoint

### DIFF
--- a/test/scripts/e2e_subs/rest.sh
+++ b/test/scripts/e2e_subs/rest.sh
@@ -116,5 +116,3 @@ function test_assets_endpoint {
 # Run the tests.
 test_applications_endpoint
 test_assets_endpoint
-
-exit $EXIT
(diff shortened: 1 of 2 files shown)


tsachiherman:
I'm not sure where the `curl_out.txt` file is being created at. If it's being created at a unique place dedicated for this unit test, than it's fine. If not, it could ( theoretically ) interfere with other unit tests using the same file name concurrently.

Note that it's a minor issue, so I wouldn't hesitate merging the PR as is and have that addressed at some point in the future.

test/scripts/e2e_subs/rest.sh
```diff
@@ -16,23 +16,48 @@ PUB_TOKEN=$(cat "$ALGORAND_DATA"/algod.token)
...
 function call_and_verify {
-  local RES
-  RES=$(call "$2")
-  if [[ "$RES" != *"$3"* ]]; then
-    echo "Failed test - $2: $1"
-    exit 1
+  local CODE
+
+  set +e
+  CODE=$(call "$2" curl_out.txt)
```
brianolson, line 49: this should probably be `"${TEMPDIR}/curl_out.txt"`

brianolson:
@tsachiherman "${TEMPDIR}" is unique per sub-test so it could live under that

winder:
@brianolson thanks, thats perfect.

PR #1390, opened 2020-08-12
Extend ledger request timeout

tsachiherman:
## Summary

The recent increase in ledger size require the server side handler to allow longer timeouts.
This became a bigger issue that I was original hoping to deal with : The timeout configuration for the issue is the "WriteTimeout", which is defined once you start the HTTP server.

That, naturally, doesn't give us the required resolution needed in order to associate a specific request handler with a given timeout. In order to achieve that, I extended the request tracker so that it would keep track of all incoming connections, even across HTTP requests ( i.e. http requests might use http 1.1 and re-use the connection ).

From that point it was quite straight-forward : expose the entry point from the GossipNode interface and update the timeout on the handler.

## Test Plan

Tested manually.

### DIFF
--- a/network/requestTracker.go
+++ b/network/requestTracker.go
@@ -46,10 +46,11 @@ type TrackerRequest struct {
 	otherTelemetryGUID string
 	otherInstanceName  string
 	otherPublicAddr    string
+	connection         net.Conn
 }
 
 // makeTrackerRequest creates a new TrackerRequest.
-func makeTrackerRequest(remoteAddr, remoteHost, remotePort string, createTime time.Time) *TrackerRequest {
+func makeTrackerRequest(remoteAddr, remoteHost, remotePort string, createTime time.Time, conn net.Conn) *TrackerRequest {
 	if remoteHost == "" {
 		remoteHost, remotePort, _ = net.SplitHostPort(remoteAddr)
 	}
@@ -59,6 +60,7 @@ func makeTrackerRequest(remoteAddr, remoteHost, remotePort string, createTime ti
 		remoteAddr: remoteAddr,
 		remoteHost: remoteHost,
 		remotePort: remotePort,
+		connection: conn,
 	}
 }
 
@@ -196,7 +198,7 @@ func (rt *RequestTracker) Accept() (conn net.Conn, err error) {
 			return
 		}
 
-		trackerRequest := makeTrackerRequest(conn.RemoteAddr().String(), "", "", time.Now())
+		trackerRequest := makeTrackerRequest(conn.RemoteAddr().String(), "", "", time.Now(), conn)
 		rateLimitingWindowStartTime := trackerRequest.created.Add(-time.Duration(rt.config.ConnectionsRateLimitingWindowSeconds) * time.Second)
 
 		rt.hostRequestsMu.Lock()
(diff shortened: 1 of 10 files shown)


rpcs/ledgerService.go
```diff
@@ -165,6 +177,11 @@ func (ls *LedgerService) ServeHTTP(response http.ResponseWriter, request *http.R
 		response.Write([]byte(fmt.Sprintf("specified round number could not be parsed using base 36 : %v", err)))
 		return
 	}
+	if conn := ls.net.GetHTTPRequestConnection(request); conn != nil {
+		conn.SetWriteDeadline(time.Now().Add(maxCatchpointFileWritingDuration))
```
tsachiherman, line 181: Ideally, we would figure out what the file size is, and have a custom timeout per file. For now, I'm leaving it with a higher max value as I don't want to go into the rabbit hole of retrieving the file size.

network/requestTracker.go
```diff
@@ -59,13 +61,15 @@ func makeTrackerRequest(remoteAddr, remoteHost, remotePort string, createTime ti
...
 	}
 }
 
 // hostIncomingRequests holds all the requests that are originating from a single host.
 type hostIncomingRequests struct {
-	remoteHost string
-	requests   []*TrackerRequest // this is an ordered list, according to the requestsHistory.created
+	remoteHost             string
+	requests               []*TrackerRequest            // this is an ordered list, according to the requestsHistory.created
+	additionalHostRequests map[*TrackerRequest]struct{} // additional requests that aren't included in the "requests", and always assumed to be "alive".
```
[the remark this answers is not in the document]
tsachiherman: The purpose of the map is to track the *count* of entries for that tracker, while providing easy ( and efficient ) way to delete them.

network/requestTracker.go
```diff
@@ -102,7 +145,7 @@ func (ard *hostIncomingRequests) add(trackerRequest *TrackerRequest) {
 // countConnections counts the number of connection that we have that occured after the provided specified time
 func (ard *hostIncomingRequests) countConnections(rateLimitingWindowStartTime time.Time) (count uint) {
 	i := ard.findTimestampIndex(rateLimitingWindowStartTime)
-	return uint(len(ard.requests) - i)
+	return uint(len(ard.requests) - i + len(ard.additionalHostRequests))
```
algonautshant, line 148: We are not checking if the connections in additionalHostRequests are after rateLimitingWindowStartTime.
What is the catch? Maybe the comment should change?

brianolson:
I'm rusty looking at network code and this is extra twisty because of the request tracker, but I didn't see anything wrong with it.

network/requestTracker.go
tsachiherman: All the connections in ard.additionalHostRequests are such that have already reached the http handler.
As such, these will get deleted only when the connection.Close() is called.
The tracking of the last-30-seconds window doesn't apply for these since these might be a long-living connections.

The purpose of the 30-seconds window was to track incoming connections that never made it to the http handler. Once they did, we don't really care about their "window".

network/requestTracker.go
```diff
@@ -196,7 +277,7 @@ func (rt *RequestTracker) Accept() (conn net.Conn, err error) {
 			return
 		}
 
-		trackerRequest := makeTrackerRequest(conn.RemoteAddr().String(), "", "", time.Now())
+		trackerRequest := makeTrackerRequest(conn.RemoteAddr().String(), "", "", time.Now(), conn)
```
algonautshant, line 280: conn is being passed here, but isn't this nil at this point?
It gets initialized further down:

> 		conn = &requestTrackedConnection{Conn: conn, tracker: rt}

algonautshant: Never mind. I missed the initialization above.

PR #1391, opened 2020-08-12
Add a transaction type to store compact certificates

zeldovich:
This introduces a new transaction type, which stores a compact
certificate.  The block header keeps track of the last compact
certificate seen, and applying the compact certificate transaction bumps
this counter in the header, ensuring that all compact certs are seen
in-order.

Storing compact certs on the blockchain helps address a number of
issues regarding storage and dissemination of compact certificates.

This change also introduces a special address used to issue the
compact cert transaction.  This address is only allowed to issue
the compact cert transacton type, but does not pay any fee for
doing so.


zeldovich:
Rebased against master which now has the compactcert crypto merged in, and made this a non-draft PR.

(Accidentally closed and re-opened this PR as an unexpected side-effect of manipulating branches in github..)

data/pools/transactionPool.go
```diff
@@ -293,6 +293,16 @@ func (pool *TransactionPool) computeFeePerByte() uint64 {
 // checkSufficientFee take a set of signed transactions and verifies that each transaction has
 // sufficient fee to get into the transaction pool
 func (pool *TransactionPool) checkSufficientFee(txgroup []transactions.SignedTxn) error {
+	// Special case: the compact cert transaction, if issued from the
```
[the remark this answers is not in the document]
zeldovich: Economy of mechanism:

- These certificates are likely to be generated by nodes that aren't block proposers, so they need to be broadcast and queued by nodes that eventually propose a block.

- There might be multiple certs that appear in a single block, and they might take a non-trivial amount of block space; the tx pool needs to take them into account.

- The certs are much larger than our current block header, so they should not literally appear in the block header.  Having them as a transaction gives the right properties: the header has a hash commitment to the payset (which includes the cert transaction), and the header doesn't become huge.

- It's convenient to be able to look them up by sender address, like other transactions.

It's entirely possible to implement another part of the block that looks much like the current transaction set, but instead is used for "system messages".  It would mostly duplicate the transaction machinery: its own message broadcast protocol, its own message pool, its own hash commitment in the header, its own lookup API, etc.

algorandskiy (commented):
Nickolai, do I understand it right and anyone can send a compact cert txn, and it is accepted iff the cert itself matches what others calculate?
It would be also nice to see some tests.

zeldovich:
> Nickolai, do I understand it right and anyone can send a compact cert txn, and it is accepted iff the cert itself matches what others calculate?

That's correct.

> It would be also nice to see some tests.

Fair enough, will add some tests to confirm that it's not possible to issue unexpected transaction types from this special address.

data/transactions/transaction.go
```diff
@@ -378,14 +408,22 @@ func (tx Transaction) WellFormed(spec SpecialAddresses, proto config.ConsensusPa
...
+
 	for t, nonZero := range nonZeroFields {
 		if nonZero && t != tx.Type {
 			return fmt.Errorf("transaction of type %v has non-zero fields for type %v", tx.Type, t)
 		}
 	}
 
 	if tx.Fee.LessThan(basics.MicroAlgos{Raw: proto.MinTxnFee}) {
-		return makeMinFeeErrorf("transaction had fee %v, which is less than the minimum %v", tx.Fee, proto.MinTxnFee)
+		if tx.Type == protocol.CompactCertTx && tx.Sender == CompactCertSender && tx.Fee.IsZero() {
```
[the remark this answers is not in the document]
zeldovich: Good point, cleaned that up.

data/transactions/compactcert.go
```diff
@@ -0,0 +1,66 @@
...
+// ToBeHashed implements the crypto.Hashable interface
+func (a specialAddr) ToBeHashed() (protocol.HashID, []byte) {
+	return protocol.SpecialAddr, []byte(a)
+}
+
+// CompactCertSender is the computed address for sending out compact certs.
+var CompactCertSender basics.Address
+
+func init() {
+	CompactCertSender = basics.Address(crypto.HashObj(specialAddr("CompactCertSender")))
```
tsachiherman, line 65: It feels like it should be in the genesis; for a non specified addresses, we can use the special one you're suggested, but could allow the enduser to configure a network with his own secret address.

ledger/compactcert.go
```diff
@@ -0,0 +1,155 @@
...
+		SecKQ:        proto.CompactCertSecKQ,
+	}
+	return
+}
+
+// validateCompactCert checks that a compact cert is valid.
+func validateCompactCert(certHdr bookkeeping.BlockHeader, cert compactcert.Cert, votersHdr bookkeeping.BlockHeader, lastCertRnd basics.Round, atRound basics.Round) error {
+	proto := config.Consensus[certHdr.CurrentProtocol]
+
+	if certHdr.Round%basics.Round(proto.CompactCertRounds) != 0 {
```
tsachiherman, line 127: Could you please add a test for `basics.Round(proto.CompactCertRounds)` to ensure it's not zero ?

tsachiherman (commented):
Just two comments. overall looks good.

data/transactions/compactcert.go
zeldovich: Do you have a use case in mind for why someone would want to change this address?

ledger/compactcert.go
zeldovich: Good idea.  I think it was called only from places that already checked for this, but I like having this function be more defensive.  Pushed.

data/transactions/compactcert.go
tsachiherman: thinking about it - what's the value of having this address anyway ? This address is only used as the sender in a `CompactCertTx` transactions. Unless we're going to support other senders in the `CompactCertTx`, can't we just fill in an empty Sender ? ( since it's not providing any additional value, and only consume more space in the block )

zeldovich: Indeed, I actually had a zero value for this initially.  But then I ran into a number of complications where existing code required the sender to be non-zero.

tsachiherman (approved):
looks good

PR #1392, opened 2020-08-13
Improve client-side catchpoint catchup performance

tsachiherman:
## Summary

Existing code was naive: it would have created a new Merkle Trie object on each catchpoint file chunk.
This approach is correct, but inefficient - the internal trie cache need to reload all the relevant pages on each iteration.

Instead of doing so, I switched to persisted Merkle Trie object, and added the changes to it. It has the benefits of a persistent cache, and perform significantly faster.

## Test Plan

Tested manually against betanet and mainnet, and seems to be considerably faster.

### DIFF
--- a/ledger/catchupaccessor.go
+++ b/ledger/catchupaccessor.go
@@ -216,6 +216,10 @@ type CatchpointCatchupAccessorProgress struct {
 	ProcessedBytes    uint64
 	TotalChunks       uint64
 	SeenHeader        bool
+
+	// Having the cachedTrie here would help to accelarate the catchup process since the trie maintain an internal cache of nodes.
+	// While rebuilding the trie, we don't want to force and reload (some) of these nodes into the cache for each catchpoint file chunk.
+	cachedTrie *merkletrie.Trie
 }
 
 // ProgressStagingBalances deserialize the given bytes as a temporary staging balances
@@ -289,13 +293,19 @@ func (c *CatchpointCatchupAccessorImpl) processStagingBalances(ctx context.Conte
 	wdb := c.ledger.trackerDB().wdb
 	err = wdb.Atomic(func(ctx context.Context, tx *sql.Tx) (err error) {
 		// create the merkle trie for the balances
-		mc, err0 := makeMerkleCommitter(tx, true)
-		if err0 != nil {
-			return err0
-		}
-		trie, err := merkletrie.MakeTrie(mc, trieCachedNodesCount)
+		var mc *merkleCommitter
+		mc, err = makeMerkleCommitter(tx, true)
 		if err != nil {
 			return err
+		}
+
+		if progress.cachedTrie == nil {
+			progress.cachedTrie, err = merkletrie.MakeTrie(mc, trieCachedNodesCount)
+			if err != nil {
+				return err
+			}
+		} else {
+			progress.cachedTrie.SetCommitter(mc)
 		}
 
 		err = writeCatchpointStagingBalances(ctx, tx, balances.Balances)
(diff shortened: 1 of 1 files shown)


ledger/catchupaccessor.go
```diff
@@ -289,13 +293,19 @@ func (c *CatchpointCatchupAccessorImpl) processStagingBalances(ctx context.Conte
...
+		mc, err = makeMerkleCommitter(tx, true)
 		if err != nil {
-			return err
+			return
+		}
+
+		if progress.cachedTrie == nil {
+			progress.cachedTrie, err = merkletrie.MakeTrie(mc, trieCachedNodesCount)
+			if err != nil {
+				return err
```
algorandskiy, line 305: simpy `return` like in other places

ledger/catchupaccessor.go
```diff
@@ -289,13 +293,19 @@ func (c *CatchpointCatchupAccessorImpl) processStagingBalances(ctx context.Conte
...
-		trie, err := merkletrie.MakeTrie(mc, trieCachedNodesCount)
+		var mc *merkleCommitter
+		mc, err = makeMerkleCommitter(tx, true)
 		if err != nil {
-			return err
+			return
+		}
+
+		if progress.cachedTrie == nil {
+			progress.cachedTrie, err = merkletrie.MakeTrie(mc, trieCachedNodesCount)
```
algorandskiy, line 303: do we hold a lock while assigning?

ledger/catchupaccessor.go
```diff
@@ -324,30 +334,39 @@ func (c *CatchpointCatchupAccessorImpl) processStagingBalances(ctx context.Conte
...
 					if err != nil {
-						return err
+						return
 					}
 				}
 			}
 
 			hash := accountHashBuilder(balance.Address, accountData, balance.AccountData)
-			added, err := trie.Add(hash)
+			added, err2 := progress.cachedTrie.Add(hash)
```
algorandskiy, line 343: why `err2`? `err` does not look used in parallel here

ledger/catchupaccessor.go
```diff
@@ -324,30 +334,39 @@ func (c *CatchpointCatchupAccessorImpl) processStagingBalances(ctx context.Conte
...
+			if err2 != nil {
+				return err2
 			}
 		}
-		err = trie.Commit()
-		if err != nil {
-			return
+
+		// periodically, perform commit & evict to flush it to the disk and rebalance the cache memory utilization.
+		if (progress.ProcessedAccounts/trieRebuildCommitFrequency) < ((progress.ProcessedAccounts+uint64(len(balances.Balances)))/trieRebuildCommitFrequency) ||
```
algorandskiy, line 353: maybe move this complex cond into `progress.EvictNeeded(balances, trieRebuildCommitFrequency)` ?

ledger/catchupaccessor.go
tsachiherman: no, why should we ? do you see any potential contention ?
as far as I can tell, only this function is accessing this field.

ledger/catchupaccessor.go
tsachiherman: because of the `added`; I didn't want to hide the original err. I will change this by declaring the "added" as bool

ledger/catchupaccessor.go
tsachiherman: I don't really think it's a complicated expression.. but I don't mind moving it to a function either.

ledger/catchupaccessor.go
algorandskiy: if two threads runs this function then the second thread can reinitialize the cache.

tsachiherman: which two threads ? there can be only a single catchpoint "restore" at a time.

PR #1393, opened 2020-08-13
Add vacuumming to the database upgrade process

tsachiherman:
## Summary

This change adds another step to the database upgrade process by triggering the vacuuming on completion of the upgrade process.

## Test Plan

Tested and confirmed to be working manually by reviewing the log files.

### DIFF
--- a/ledger/accountdb.go
+++ b/ledger/accountdb.go
@@ -100,7 +100,7 @@ var accountsResetExprs = []string{
 // accountDBVersion is the database version that this binary would know how to support and how to upgrade to.
 // details about the content of each of the versions can be found in the upgrade functions upgradeDatabaseSchemaXXXX
 // and their descriptions.
-var accountDBVersion = int32(2)
+var accountDBVersion = int32(3)
 
 type accountDelta struct {
 	old basics.AccountData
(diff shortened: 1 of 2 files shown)


tsachiherman:
> The change is only showing the version upgrade.
> Where is the code that applies the vacuuming? Is it in another change?

Yes, it's in https://github.com/algorand/go-algorand/pull/1375
John conducted some performance tests around that to ensure the time it takes isn't prohibitive.

algonautshant:
Can you please comment on how this change effects OptimizeAccountsDatabaseOnStartup?

The way I understand is that irrespective of OptimizeAccountsDatabaseOnStartup, the upgrade to v3 will take place, and au.vacuumOnStartup will be set to true.

tsachiherman:
> Can you please comment on how this change effects OptimizeAccountsDatabaseOnStartup?
> 
> The way I understand is that irrespective of OptimizeAccountsDatabaseOnStartup, the upgrade to v3 will take place, and au.vacuumOnStartup will be set to true.

I think that your understanding is correct. However, it also leaves the ability for the end user to force a vacuum on every startup if he/she wishes to.

PR #1394, opened 2020-08-13
Avoid redundent info message for replacing block 0

tsachiherman:
## Summary

The current code does fix the issue of having incorrect TxnRoot on block 0, however, it also write to the log file the message `initBlocksDB replaced block 0` on every startup.

To address that, I have compared the old and new headers. If they are identical, the update is not needed. Note that the bug was in the TxnRoot field, but the test cover the entire block header.

## Test Plan

Tested manually to ensure correctness.

### DIFF
--- a/ledger/blockdb.go
+++ b/ledger/blockdb.go
@@ -166,6 +166,11 @@ func blockReplaceIfExists(tx *sql.Tx, log logging.Logger, blk bookkeeping.Block,
 	newBlk := protocol.Encode(&blk)
 	newCert := protocol.Encode(&cert)
 
+	// if the header hasn't been modified, just return.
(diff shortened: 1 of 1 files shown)


ledger/blockdb.go
```diff
@@ -166,6 +166,11 @@ func blockReplaceIfExists(tx *sql.Tx, log logging.Logger, blk bookkeeping.Block,
 	newBlk := protocol.Encode(&blk)
 	newCert := protocol.Encode(&cert)
 
+	// if the header hasn't been modified, just return.
+	if bytes.Compare(oldHdr[:], newHdr[:]) == 0 {
```
algorandskiy, line 170: nit: `bytes.Equal` ?

tsachiherman: sure. updated.

PR #1395, opened 2020-08-13
Increase connect timeout for e2e_client_runner

egieseke:
## Summary

Increase connect timeout for e2e_client_runner test script from 5 to 20 seconds.

## Test Plan

Running Travis tests.

### DIFF
--- a/test/scripts/e2e_client_runner.py
+++ b/test/scripts/e2e_client_runner.py
@@ -174,7 +174,7 @@ class RunSet:
         if self.algod and self.kmd:
             return
         # should run from inside self.lock
-        xrun(['goal', 'kmd', 'start', '-t', '200'], env=self.env, timeout=5)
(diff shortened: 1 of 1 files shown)


test/scripts/e2e_client_runner.py
```diff
@@ -66,6 +66,7 @@ def read_script_for_timeout(fname):
     return 200
 
 def _script_thread_inner(runset, scriptname):
+    time.sleep(20)
```
brianolson, line 69: I'd really like to avoid this sleep that slows down everything. Why wouldn't the system be ready yet? Haven't we already done things at this point to show that the network is up and ready? Is there a simple operation we can run in fast polling to check for things being ready?

egieseke: It seems that kmd is failing to start, so waiting 20 seconds will not help.

PR #1396, opened 2020-08-14
Allow algorand-indexer to be installed alongside other packages

btoll:
## Summary

The `preinst` maintenance script was not allowing for the indexer to be install alongside an `algorand` or `algorand-beta` package.

## Test Plan

Build the deb package doing `mule -f package.yaml package` in the root of the project directory.  This will put the package under the `./tmp/` directory, and then run the following to install `sudo dpkg -i tmp/...*.deb`.

### DIFF
--- a/installer/debian/preinst
+++ b/installer/debian/preinst
@@ -1,23 +1,27 @@
-#!/bin/bash
+#!/usr/bin/env bash
+
+set -e
 
 PKG_NAME=@PKG_NAME@
-VERSION="$2"
+INDEXER=algorand-indexer
 
-if test "$1" = install &&
-    dpkg-query --list 'algorand*' > /dev/null
+if [ "$1" != install ] || [ "$PKG_NAME" == $INDEXER ]
 then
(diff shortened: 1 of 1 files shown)


installer/debian/preinst
```diff
@@ -1,23 +1,27 @@
...
+
+set -e
 
 PKG_NAME=@PKG_NAME@
-VERSION="$2"
+INDEXER=algorand-indexer
 
-if test "$1" = install &&
-    dpkg-query --list 'algorand*' > /dev/null
+if [ "$1" != install ] || [ "$PKG_NAME" == $INDEXER ]
```
brianolson, line 8: PKG_NAME will never be algorand-indexer here though, right?

btoll: It could be, if a user did `sudo apt-get install algorand-indexer`.

brianolson: Yes, but then they'd get the preinst built into the indexer package (and there isn't one over there)

btoll: Oh, i see what you're saying, `algorand-indexer` isn't packaged with `preinst` so it will never be that value.  I had switched the order of operations and hadn't thought of that.  Thanks for catching that :)

PR #1397, opened 2020-08-14
Convert LedgerReader.BalanceRecord into LedgerReader.Lookup.

derbear:
LedgerReader.BalanceRecord is redundant because the address that is
passed in does not need to be returned to the agreement code.

## Test Plan

Existing tests were converted.


derbear:
May be of interest to @justicz

agreement/selector.go
```diff
@@ -64,7 +64,7 @@ func membership(l LedgerReader, addr basics.Address, r basics.Round, p period, s
 	balanceRound := balanceRound(r, cparams)
 	seedRound := seedRound(r, cparams)
 
-	record, err := l.BalanceRecord(balanceRound, addr)
+	record, err := l.AccountData(balanceRound, addr)
```
tsachiherman, line 67: this line got me confused for few seconds, as the `AccountData` is both the name of the data structure as well as the function name. I think that `GetAccountData` is a better choice for a name. ( note that the same applies for the previous `BalanceRecord` function name )

justicz:
LGTM, I'm glad we're not using the extra `BalanceRecord` field very many places!

data/ledger.go
```diff
@@ -186,17 +186,9 @@ func (l *Ledger) NextRound() basics.Round {
...
-		return basics.BalanceRecord{}, err
-	}
-
-	return basics.BalanceRecord{
-		Addr:        addr,
-		AccountData: data,
-	}, nil
+// AccountData implements agreement.Ledger.AccountData. It applies pending rewards to returned amounts.
+func (l *Ledger) AccountData(r basics.Round, addr basics.Address) (basics.AccountData, error) {
+	return l.Lookup(r, addr)
```
zeldovich, line 191: Since there's nothing interesting going on in this wrapper function now (even the order of arguments is exactly the same), perhaps we should delete it altogether and call this Lookup() in agreement's LedgerReader?

data/committee/committee.go
```diff
@@ -36,9 +36,18 @@ type Selector interface {
 	CommitteeSize(config.ConsensusParams) uint64
 }
 
+// BalanceRecord pairs an account's address with its associated data.
+//
+// This struct is used to decouple LedgerReader.AccountData from basics.BalanceRecord.
+//msgp:ignore BalanceRecord
```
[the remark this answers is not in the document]
derbear: I don't think we ever serialize anything of type `committee.Membership`.

agreement/selector.go
derbear: That makes sense. I've changed it to `Lookup` which is consistent to what `ledger.Ledger` does.

data/ledger.go
derbear: Sounds good--changed the interface to depend on `Lookup`.

PR #1401, opened 2020-08-14
Filter timeout to 2s for period 0, 4s for period > 0

nicholasguoalgorand:
## Summary

Change filter timeout to a function instead of a constant. Filter Timeout is now 2s for Period 0 and 4s for Period > 0

Moved https://github.com/algorand/go-algorand/pull/1326 to here.

## Test Plan

Unit tests for current version, new version, as well as protocol upgrade


agreement/types.go
```diff
@@ -22,6 +22,7 @@ import (
 	"github.com/algorand/go-algorand/config"
 	"github.com/algorand/go-algorand/data/basics"
 	"github.com/algorand/go-algorand/logging"
+	"github.com/algorand/go-algorand/protocol"
 )
 
 var filterTimeout = 2 * config.Protocol.SmallLambda
```
derbear, line 28: Could you delete this constant?

PR #1407, opened 2020-08-17
Disable deadlock detection on master

tsachiherman:
## Summary

Disable the deadlock by default on master; enable only when running travis tests or setting via ALGORAND_DEADLOCK

### DIFF
--- a/scripts/travis/build_test.sh
+++ b/scripts/travis/build_test.sh
@@ -9,6 +9,8 @@
 # Examples: scripts/travis/build_test.sh
 set -e
 
+ALGORAND_DEADLOCK=enable
+export ALGORAND_DEADLOCK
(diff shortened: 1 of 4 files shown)


scripts/travis/build_test.sh
```diff
@@ -9,6 +9,8 @@
 # Examples: scripts/travis/build_test.sh
 set -e
 
+ALGORAND_DEADLOCK=enable
+export ALGORAND_DEADLOCK
```
onetechnical, line 13: Nit: just `export ALGORAND_DEADLOCK=enable`

tsachiherman: I think that shellcheck doesn't like when you do that..

onetechnical: Is that true for variable assignments or just when there's command output trapping error codes? Either way it's fine.

PR #1408, opened 2020-08-17
Add CatchpointTracking to config file

tsachiherman:
## Summary

Add a CatchpointTracking option to the config file, so that non-archival nodes would not track catchpoints by defaults.


config/config.go
```diff
@@ -328,6 +328,13 @@ type Local struct {
 	// OptimizeAccountsDatabaseOnStartup controls whether the accounts database would be optimized
 	// on algod startup.
 	OptimizeAccountsDatabaseOnStartup bool `version[10]:"false"`
+
+	// CatchpointTracking determines if catchpoints are going to be tracked. The value is interpreted as follows:
+	// A of -1 means "don't track catchpoints".
```
algonautshant, line 333: type: A of -> A value of

config/config.go
```diff
@@ -328,6 +328,13 @@ type Local struct {
...
+	// A value of 1 means "track catchpoints as long as CatchpointInterval is also set to a positive non-zero value". If CatchpointInterval <= 0, no catchpoint tracking would be performed.
```
algonautshant, line 334: Should there be a comment on CatchpointInterval that it's effect may be overruled by CatchpointTracking?

config/config.go
tsachiherman: correct. I will fix it in the next pr ( just so it won't delay the merge too much )

config/config.go
tsachiherman: I can't see a good reason not to do so. Will do on next PR.

PR #1409, opened 2020-08-17
Slow down catchup while catchpoint file is being written

tsachiherman:
## Summary

Testing the algod against mainnet shown that memory utilization during catchup could be substantial.
This PR slows down the catchup when catchpoint file is being written. This would lower the memory pressure
on the accounts updater, as no additional blocks would get added.

Note that it will make a difference only on archival nodes. Non-archival nodes would behave as they do today.

### DIFF
--- a/ledger/acctupdates.go
+++ b/ledger/acctupdates.go
@@ -275,6 +275,22 @@ func (au *accountUpdates) close() {
 	<-au.commitSyncerClosed
 }
 
+// IsWritingCatchpointFile returns true when a catchpoint file is being generated. The function is used by the catchup service so
+// that the memory pressure could be decreased until the catchpoint file writing is complete.
+func (au *accountUpdates) IsWritingCatchpointFile() bool {
+	au.accountsMu.RLock()
+	defer au.accountsMu.RUnlock()
+	// if we're still writing the previous balances, we can't move forward yet.
+	select {
+	case <-au.catchpointWriting:
+		// the channel catchpointWriting is currently closed, meaning that we're currently not writing any
+		// catchpoint file.
+		return false
+	default:
+		return true
+	}
+}
+
 // Lookup returns the accound data for a given address at a given round. The withRewards indicates whether the
 // rewards should be added to the AccountData before returning. Note that the function doesn't update the account with the rewards,
 // even while it could return the AccoutData which represent the "rewarded" account data.
@@ -1579,6 +1595,10 @@ func (au *accountUpdates) generateCatchpoint(committedRound basics.Round, label 
 				db.ResetTransactionWarnDeadline(ctx, tx, time.Now().Add(1*time.Second))
 				select {
 				case <-time.After(100 * time.Millisecond):
+					chunkExecutionDuration *= 2
+					if chunkExecutionDuration > longChunkExecutionDuration {
+						chunkExecutionDuration = longChunkExecutionDuration
+					}
 				case <-au.ctx.Done():
 					retryCatchpointCreation = true
 					err2 := catchpointWriter.Abort()
--- a/ledger/acctupdates.go
+++ b/ledger/acctupdates.go
@@ -275,11 +275,11 @@ func (au *accountUpdates) close() {
 	<-au.commitSyncerClosed
 }
 
-// IsWritingCatchpointFile returns true when a catchpoint file is being generated. The function is used by the catchup service so
-// that the memory pressure could be decreased until the catchpoint file writing is complete.
+// IsWritingCatchpointFile returns true when a catchpoint file is being generated. The function is used by the catchup service
+// to avoid memory pressure until the catchpoint file writing is complete.
 func (au *accountUpdates) IsWritingCatchpointFile() bool {
-	au.accountsMu.RLock()
-	defer au.accountsMu.RUnlock()
+	au.accountsMu.Lock()
+	defer au.accountsMu.Unlock()
 	// if we're still writing the previous balances, we can't move forward yet.
 	select {
 	case <-au.catchpointWriting:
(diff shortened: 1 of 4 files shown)


ledger/ledger.go
```diff
@@ -584,6 +584,14 @@ func (l *Ledger) trackerEvalVerified(blk bookkeeping.Block, accUpdatesLedger led
 	return eval(context.Background(), accUpdatesLedger, blk, false, nil, nil)
 }
 
+// IsWritingCatchpointFile returns true when a catchpoint file is being generated. The function is used by the catchup service so
+// that the memory pressure could be decreased until the catchpoint file writing is complete.
```
yossigi, line 588: "memory pressure could be decreased" --> "to avoid memory pressure"

ledger/acctupdates.go
```diff
@@ -275,6 +275,22 @@ func (au *accountUpdates) close() {
...
 }
 
+// IsWritingCatchpointFile returns true when a catchpoint file is being generated. The function is used by the catchup service so
+// that the memory pressure could be decreased until the catchpoint file writing is complete.
+func (au *accountUpdates) IsWritingCatchpointFile() bool {
+	au.accountsMu.RLock()
+	defer au.accountsMu.RUnlock()
+	// if we're still writing the previous balances, we can't move forward yet.
+	select {
+	case <-au.catchpointWriting:
```
derbear, line 285: Out of curiosity, why is `au.catchpointWriting` a channel and not a boolean variable protected by `accountsMu`?

ledger/acctupdates.go
```diff
@@ -275,6 +275,22 @@ func (au *accountUpdates) close() {
 	<-au.commitSyncerClosed
 }
 
+// IsWritingCatchpointFile returns true when a catchpoint file is being generated. The function is used by the catchup service so
+// that the memory pressure could be decreased until the catchpoint file writing is complete.
+func (au *accountUpdates) IsWritingCatchpointFile() bool {
+	au.accountsMu.RLock()
```
yossigi, line 281: why do you need to take a read lock here? (the channel is already thread safe)

ledger/ledger.go
```diff
@@ -584,6 +584,14 @@ func (l *Ledger) trackerEvalVerified(blk bookkeeping.Block, accUpdatesLedger led
...
+func (l *Ledger) IsWritingCatchpointFile() bool {
+	l.trackerMu.RLock()
```
yossigi, line 590: why are the locks necessary?

ledger/acctupdates.go
```diff
@@ -1579,6 +1595,10 @@ func (au *accountUpdates) generateCatchpoint(committedRound basics.Round, label
 				db.ResetTransactionWarnDeadline(ctx, tx, time.Now().Add(1*time.Second))
 				select {
 				case <-time.After(100 * time.Millisecond):
+					chunkExecutionDuration *= 2
```
algonautshant, line 1598: Can you please comment why this duration is increasing here?

ledger/acctupdates.go
tsachiherman: It would probably be a good idea to use a different synchronization primitive at this point, but I didn't want to make such a big change for this fix.

ledger/ledger.go
tsachiherman: you're right. It was my bad. It should have been a `l.trackerMu.Lock()`.
explain:
`catchpointWriting` is being updated in two locations:
1. `initializeFromDisk`
2. `committedUpTo`

Since the channel itself is changing, it has to be protected. The former isn't an issue, but the latter is.
Given that the `committedUpTo` is updating the `catchpointWriting` while having the `l.trackerMu.RLock`, we want to have the `IsWritingCatchpointFile` taking the *Write* lock.

derbear: It seems a little counterintuitive to me to have a getter called `IsWritingCatchpointFile` take a write lock...

derbear: (especially when `committedUpTo` is the actual writer of the protected state)

yossigi: This does feel strange. I think it's an artifact of committedUpTo writing but taking only a read lock

derbear: I personally feel it would be easier in the long run to read the code if `au.catchpointWriting` were replaced by something from `sync/atomic`.

yossigi:
I agree, using atomics would make it easier

ledger/ledger.go
tsachiherman: I think that the reason it appear weird is because it's a different "synchronization set". It should either be a it's own mutex + bool, or an atomic variable. I lean toward the latter. Either way, I think that we should avoid doing that for this release, and make it toward the next one.

derbear:
What happens if right as `committedUpTo` returns, `dc.offset=0`?

https://github.com/algorand/go-algorand/blob/master/ledger/acctupdates.go#L414

tsachiherman:
> What happens if right as `committedUpTo` returns, `dc.offset=0`?
> 
> https://github.com/algorand/go-algorand/blob/master/ledger/acctupdates.go#L414

I'm not sure what the question is..
committedUpTo is guaranteed to be called with an increasing round numbers.
So, the offset would always be either non-zero ( since we had at least one round since the previous one )
or zero, in case we decided not to flush the current round ( and accumulate the pending changes )

derbear:
Ah, I see. Why do we still send a deferred commit even if we don't have a catchpoint round?

tsachiherman:
> ~Ah, I see. Why do we still send a deferred commit even if we don't have a catchpoint round?~
> 
> To clarify: is it intentional that we still send on `au.committedOffset` on this line? https://github.com/algorand/go-algorand/blob/master/ledger/acctupdates.go#L490

I don't believe this is the case. In this case, the `dc` data structure was not initialize and therefore dc.offset = 0

derbear:
Right, that makes sense to me.

PR #1415, opened 2020-08-18
Update CatchpointInterval config comment

tsachiherman:
## Summary

Update CatchpointInterval config comment, per review request.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -290,7 +290,7 @@ type Local struct {
 	// NetworkProtocolVersion overrides network protocol version ( if present )
 	NetworkProtocolVersion string `version[6]:""`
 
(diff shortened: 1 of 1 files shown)


config/config.go
```diff
@@ -290,7 +290,8 @@ type Local struct {
 	// NetworkProtocolVersion overrides network protocol version ( if present )
 	NetworkProtocolVersion string `version[6]:""`
 
-	// CatchpointInterval set the interval at which catchpoint are being generated.
+	// CatchpointInterval set the interval at which catchpoint are being generated. Setting this to 0 disable the catchpoint from being generated.
```
onetechnical, line 293:
```suggestion
	// CatchpointInterval sets the interval at which catchpoints are being generated. Setting this to 0 disables the catchpoint from being generated.
```

PR #1433, opened 2020-08-18
Dynamically install deb package in legacy pipeline

btoll:
## Summary

The legacy pipeline isn't distinguishing between stable and beta release packages. It needs to test algorand for the former and algorand-beta for the latter.

## Test Plan

https://ci.algorand.network/job/test-packages/

### DIFF
--- a/scripts/release/test/deb/test_algorand.sh
+++ b/scripts/release/test/deb/test_algorand.sh
@@ -5,6 +5,11 @@ set -ex
 export GOPATH=${HOME}/go
 export PATH=${GOPATH}/bin:/usr/local/go/bin:${PATH}
 
+PKG_NAME=algorand
+if [ "${CHANNEL}" = beta ]; then
+    PKG_NAME=algorand-beta
+fi
+
 apt-get update
 apt-get install -y gnupg2 curl software-properties-common python3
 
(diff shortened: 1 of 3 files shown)


scripts/release/test/deb/test_algorand.sh
```diff
@@ -5,6 +5,11 @@ set -ex
 export GOPATH=${HOME}/go
 export PATH=${GOPATH}/bin:/usr/local/go/bin:${PATH}
 
+PKG_NAME=algorand
+if [ "${CHANNEL}" = beta ]; then
```
egieseke, line 9: should this use double brackets?

scripts/release/test/rpm/run_centos.sh
```diff
@@ -9,6 +9,11 @@ echo
 date "+build_release begin TEST stage %Y%m%d_%H%M%S"
 echo
 
+if [ "$CHANNEL" = beta ]; then
+    echo "There is currently no support for RPM beta packages. Exiting RPM test stage..."
+    exit 0
```
egieseke, line 14: should exit code be 1?

btoll: No, because it should just exit out of the script and not fail the build.

scripts/release/test/deb/test_algorand.sh
btoll: It's fine, I've tested it.

It uses bash builtin '[`, which should act the same as the `/usr/bin/[` binary.

egieseke: OK.

PR #1435, opened 2020-08-19
add an API for using MsgOfInterest messages

zeldovich:
This adds an API that callers can use to cause the wsNetwork node to send MsgOfInterest messages to its peers so that peers can send additional message types not included in `defaultSendMessageTags`.

This is intended to be used by the compact cert code, so that relays use MsgOfInterest messages to request compact cert sigs, whereas regular non-relay nodes don't see compact cert sig messages.

### DIFF
--- a/network/msgOfInterest.go
+++ b/network/msgOfInterest.go
@@ -28,6 +28,7 @@ var errInvalidMessageOfInterest = errors.New("unmarshalMessageOfInterest: messag
 var errInvalidMessageOfInterestLength = errors.New("unmarshalMessageOfInterest: message length is too long")
 
 const maxMessageOfInterestTags = 1024
+const topicsEncodingSeparator = ","
 
 func unmarshallMessageOfInterest(data []byte) (map[protocol.Tag]bool, error) {
 	// decode the message, and ensure it's a valid message.
@@ -44,7 +45,7 @@ func unmarshallMessageOfInterest(data []byte) (map[protocol.Tag]bool, error) {
 	}
 	// convert the tags into a tags map.
 	msgTagsMap := make(map[protocol.Tag]bool, len(tags))
-	for _, tag := range strings.Split(string(tags), ",") {
+	for _, tag := range strings.Split(string(tags), topicsEncodingSeparator) {
 		msgTagsMap[protocol.Tag(tag)] = true
 	}
 	return msgTagsMap, nil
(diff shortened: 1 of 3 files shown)


tsachiherman:
looks great otherwise

zeldovich:
I added a lock to make sure that calls to `RegisterMessageInterest()` have no effect if invoked after `Start()`.  This choice was mostly to avoid extra locking in the peer accept path, which would have otherwise needed to grab some lock to look at `messagesOfInterestEnc`.

tsachiherman:
> I added a lock to make sure that calls to `RegisterMessageInterest()` have no effect if invoked after `Start()`. This choice was mostly to avoid extra locking in the peer accept path, which would have otherwise needed to grab some lock to look at `messagesOfInterestEnc`.

That makes perfect sense. I think that you might want to change 
```
	wn.messagesOfInterestEncoded = true
 	if wn.messagesOfInterest != nil {
          ...
```
to 
```
	wn.messagesOfInterestEncoded = (wn.messagesOfInterest != nil)
 	if wn.messagesOfInterest != nil {
          ...
```

so that it would work correctly even across Stop/Start

network/msgOfInterest.go
```diff
@@ -63,3 +63,19 @@ func MarshallMessageOfInterest(messageTags []protocol.Tag) []byte {
...
 	return topics.MarshallTopics()
 }
+
+// MarshallMessageOfInterestMap generates a message of interest message body
+// for the message tags that map to "true" in the map argument.
+func MarshallMessageOfInterestMap(tagmap map[protocol.Tag]bool) []byte {
+	tags := ""
+	for tag, flag := range tagmap {
+		if flag {
+			tags += "," + string(tag)
```
brianolson, line 73: Add a comment to protocol/tags.go that `,` cannot be in a tag?

brianolson: Or use a null byte or \x1e for record separator?

tsachiherman: I think that defining the const separator in `msgOfInterest.go` would do.
Adding a comment in `tags.go` is also a great idea.

zeldovich: It's too late to change the encoding (it's already in master and on mainnet); this PR is just adding a second function to produce this encoding from a map rather than a slice (which is just above the added function).  But documenting it is a good idea; will do that.

PR #1436, opened 2020-08-19
Add a test for goal app --app-account flag.

algonautshant:
Added a test for goal app --app-account flag
* The test creates four additional accounts to opt-in to the app, then passes the four accounts as --app-account flag to goal app call.
* The teal program writes the four passed account addresses into the local state of the account app. 
* The test verifies the correct addresses are written to the local state. 
* It also tests for the error when one flag is missing

Updated the README file for expect tests to mention that the test file suffix should be Test.exp


test/scripts/e2e_subs/tealprogs/appAccountParams.teal
```diff
@@ -0,0 +1,43 @@
...
+bnz skipped
+
+// Write the number of accounts passed as parameters
+int 0 // write to app caller local space
+byte "Num Accounts" // key
+txn NumAccounts // get number of accounts passed (value)
+app_local_put
+
+// Write the passed accounts to the local space
+int 0
```
algorandskiy, line 17: the fact accounts opted-in into the app is not used. when int 0 passed then the sender's local state is used. Plus, `txn Account 0` is `txn.Sender`, see the snippet:
```golang
	case Accounts:
		if arrayFieldIdx == 0 {
			// special case: sender
			sv.Bytes = txn.Sender[:]
```

I think a right way in code below should be
```
int X
byte "Account"
txn Accounts X
app_local_put
```
where X is from 0 to 4.

and then check local state not only of `PRIMARY_ACCOUNT_ADDRESS` but `ACCOUNT_1_ADDRESS` and so on.

algonautshant: Thanks. 
Updated: only one additional account opts-in, and it's local state is also written.

PR #1438, opened 2020-08-19
Auth Middleware Tests

winder:
## Summary

Adds some tests to the auth middleware. The unit test for the pprof url auth isn't great, so there is also an e2e test for that case.

### DIFF
--- a/daemon/algod/api/server/lib/middlewares/auth.go
+++ b/daemon/algod/api/server/lib/middlewares/auth.go
@@ -25,7 +25,10 @@ import (
 	"github.com/labstack/echo/v4"
 )
 
+const URLAuthPrefix = "/urlAuth/:" + TokenPathParam
 const urlAuthFormatter = "/urlAuth/%s"
+const InvalidTokenMessage = "Invalid API Token"
+const TokenPathParam = "token"
 
 // AuthMiddleware provides some data to the handler.
 type AuthMiddleware struct {
@@ -71,7 +74,7 @@ func (auth *AuthMiddleware) handler(next echo.HandlerFunc) echo.HandlerFunc {
 		}
 
 		// Handle debug routes with /urlAuth/:token prefix.
-		if ctx.Param("token") != "" {
+		if ctx.Param(TokenPathParam) != "" {
 			// For debug routes, we place the apiToken in the path itself
 			providedToken = []byte(ctx.Param("token"))
 
@@ -93,6 +96,6 @@ func (auth *AuthMiddleware) handler(next echo.HandlerFunc) echo.HandlerFunc {
 			}
 		}
 
-		return echo.NewHTTPError(http.StatusUnauthorized, "Invalid API Token")
+		return echo.NewHTTPError(http.StatusUnauthorized, InvalidTokenMessage)
 	}
 }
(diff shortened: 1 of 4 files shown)


test/scripts/e2e_subs/rest.sh
```diff
@@ -58,11 +86,63 @@ function test_applications_endpoint {
...
+    "freeze": "'
+  # Some invalid path parameters
+  call_and_verify "Asset parameter parsing error 1." "/v2/assets/-2" 400 "Invalid format for parameter asset-id"
+  call_and_verify "Asset parameter parsing error 2." "/v2/assets/not-a-number" 400 "Invalid format for parameter asset-id"
+
+  # Good request, but invalid query parameters
+  call_and_verify "Asset invalid parameter" "/v2/assets/$ASSET_ID?this-should-fail=200" 400 'parameter detected: this-should-fail'
+}
+
+function pprof_test {
```
winder, line 118: Pretty much everything except for this function is added with #1388

test/scripts/e2e_subs/rest.sh
```diff
@@ -112,7 +115,34 @@ function test_assets_endpoint {
   call_and_verify "Asset invalid parameter" "/v2/assets/$ASSET_ID?this-should-fail=200" 400 'parameter detected: this-should-fail'
 }
 
+function pprof_test {
```
winder, line 118: The pprof auth hits a special case which needs the router configured for a proper test. Covering that with this e2e test is simpler than configuring the router in a unit test.

daemon/algod/api/server/lib/middlewares/auth.go
```diff
@@ -25,7 +25,13 @@ import (
 	"github.com/labstack/echo/v4"
 )
 
+// URLAuthPrefix is the echo formatted url/path param which can be used for supplying an API token.
+const URLAuthPrefix = "/urlAuth/:" + TokenPathParam
 const urlAuthFormatter = "/urlAuth/%s"
+// InvalidTokenMessage is the message set when an invalid / missing token is found.
+const InvalidTokenMessage = "Invalid API Token"
+// TokenPathParam is the name of the path parameter used by URLAuthPrefix
+const TokenPathParam = "token"
```
brianolson, line 34: It took me a moment to realize this works because the Go compiler is smart; I was expecting to need this declared before it's used a few lines above.

daemon/algod/api/server/lib/middlewares/auth_test.go
```diff
@@ -0,0 +1,186 @@
...
+			"GET",
+			invalidTokenError,
+			"",
+		},
+	}
+
+	authFn := MakeAuth(testAPIHeader, tokens)
+	handler := authFn(success)
+
+	for _, test := range tests {
```
brianolson, line 163: use the testing sub-test method T.Run ?
https://golang.org/pkg/testing/#T.Run

daemon/algod/api/server/lib/middlewares/auth.go
winder: It's been a while since forward declaration spent much time interrupting my train of thought. I'll go ahead and move this up anyway

daemon/algod/api/server/lib/middlewares/auth_test.go
winder: Neat. Updated.

PR #1439, opened 2020-08-19
No need to return an error when requesting the same catchpoint multiple times

winder:
I think this error can be suppressed if we are asking to start a new catchpoint, but the catchpoint being requested is already in progress.


winder:
@tsachiherman adding in 201/200 was a lot more involved, let me know what you think about this second pass.

tsachiherman (approved):
nit: I would have moved the error message formatting into the `Error` function, and have the catchpoint label(s) stored inside the error structure.

thanks for the changes !

PR #1446, opened 2020-08-25
Fix accounting bug in page loading

tsachiherman:
## Summary

This PR contains two changes:
1. accounting bug fix in `merkleTrieCache::loadPage`, which could lead to incorrect `cachedNodeCount` values being used. In turn, it would cause the cache to evict pages ( or not to evict pages ) regardless of the eviction target. This could lead to excessive memory consumption.
2. modify the behavior of `merkleTrieCache::evict` so that it would re-prioritize the eviction of the root node page last. Currently, the hash calculation reloads multiple pages, which could lead to a situation where a call to `Evict(true)` would cause us to evict the root element. While this won't cause any issue on it's own, reloading the root element on the next Add/Delete operation is not desired. This change would place the root element page as the last place to consider evicting.
( note that a change to the hash recalculation page management is on it's way, and this one is a temporary patch )

## Bonus

As a side node, part of the fix in (1) was the elimination of following in the trivial case:
```
for nodeID, pnode := range decodedNodes {
	mtc.pageToNIDsPtr[page][nodeID] = pnode
}
```

This seems to consume notable amount of memory. Given that it's not usually needed, removing it on the trivial use case improves memory utilization.

## Test Plan

Unit tests were added to verify that change, and change was tested and against betanet and testnet.

### DIFF
--- a/crypto/merkletrie/cache.go
+++ b/crypto/merkletrie/cache.go
@@ -177,12 +177,15 @@ func (mtc *merkleTrieCache) loadPage(page uint64) (err error) {
 		return
 	}
 	if mtc.pageToNIDsPtr[page] == nil {
-		mtc.pageToNIDsPtr[page] = make(map[storedNodeIdentifier]*node, mtc.nodesPerPage)
+		mtc.pageToNIDsPtr[page] = decodedNodes
+		mtc.cachedNodeCount += len(mtc.pageToNIDsPtr[page])
+	} else {
+		mtc.cachedNodeCount -= len(mtc.pageToNIDsPtr[page])
+		for nodeID, pnode := range decodedNodes {
+			mtc.pageToNIDsPtr[page][nodeID] = pnode
+		}
+		mtc.cachedNodeCount += len(mtc.pageToNIDsPtr[page])
 	}
-	for nodeID, pnode := range decodedNodes {
-		mtc.pageToNIDsPtr[page][nodeID] = pnode
-	}
-	mtc.cachedNodeCount += len(mtc.pageToNIDsPtr[page])
 
 	// if we've just loaded a deferred page, no need to reload it during the commit.
 	if mtc.deferedPageLoad != page {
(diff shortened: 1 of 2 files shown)


crypto/merkletrie/cache.go
```diff
@@ -177,12 +177,15 @@ func (mtc *merkleTrieCache) loadPage(page uint64) (err error) {
 		return
 	}
 	if mtc.pageToNIDsPtr[page] == nil {
-		mtc.pageToNIDsPtr[page] = make(map[storedNodeIdentifier]*node, mtc.nodesPerPage)
-	}
-	for nodeID, pnode := range decodedNodes {
-		mtc.pageToNIDsPtr[page][nodeID] = pnode
+		mtc.pageToNIDsPtr[page] = decodedNodes
+		mtc.cachedNodeCount += len(mtc.pageToNIDsPtr[page])
```
algorandskiy, line 181: nit: `len(decodedNodes)`, one lookup less

tsachiherman: will do in a subsequent PR ( just to avoid retesting this one on travis ).

crypto/merkletrie/cache.go
```diff
@@ -438,10 +441,18 @@ func (mtc *merkleTrieCache) encodePage(nodeIDs map[storedNodeIdentifier]*node) [
...
+// evict releases the least used pages from cache until the number of elements in cache are less than cachedNodeCountTarget.
+// the root element page is being moved to the front so that it would get flushed last.
 func (mtc *merkleTrieCache) evict() (removedNodes int) {
 	removedNodes = mtc.cachedNodeCount
-
+	// check the root element ( if there is such ), and give it a higher priority, since we want
+	// to release the page with the root element last.
+	if mtc.mt.root != storedNodeIdentifierNull {
+		rootPage := uint64(mtc.mt.root) / uint64(mtc.nodesPerPage)
+		if element, has := mtc.pagesPrioritizationMap[rootPage]; has && element != nil {
```
algonautshant, line 452: I am not convinced why this change is needed.
If the reason for this is because we frequently use the root node page, it will already be taken care of by the least-recently-used eviction policy implemented by prioritizeNode, and keep this page in the cache. 

If this page is evicted frequently, could it be that there is a bug somewhere else causing it to happen?

tsachiherman: The use case here is a call to `trie.Evict(true)`, which would internally call `cache.commit()` followed by `cache.evict()`. 
When we're entering commit(), the root node page would be on the top. 
While executing `commit()`, we're trying to calculate the hashes of the various updated nodes. In many cases, we won't have the needed child node in memory, and we'll need to load it from disk. When we do this, that page is pushed to the front of the list. The update of the node hash value doesn't update its page position on the list.

Later on, when we get to call `evict()`, we find that the root page need to be evicted.
I have some bigger planned changes for this, which would hopefully eliminate the source of the problem - but I'd like to have a simple solution first, and address the "better" fix later on.

PR #1447, opened 2020-08-25
commit to Merkle tree of online participants in block header

zeldovich:
This PR commits to a Merkle tree of online participants in the block header, for use in validating compact certificates.

The main tricky aspect of this PR has to do with the fact that it takes a while to compute this Merkle tree, so the computation needs to happen over the course of several blocks, before the ledger can expect the Merkle root to be available.


data/basics/userBalance.go
```diff
@@ -381,6 +381,14 @@ func (u AccountData) Money(proto config.ConsensusParams, rewardsLevel uint64) (m
 	return e.MicroAlgos, e.RewardedMicroAlgos
 }
 
+// PendingRewards computes the amount of rewards (in microalgos) that
+// have yet to be added to the account balance.
+func PendingRewards(ot *OverflowTracker, microAlgos MicroAlgos, rewardsBase uint64, proto config.ConsensusParams, rewardsLevel uint64) MicroAlgos {
```
tsachiherman, line 386: nit: maybe move the `proto config.ConsensusParams` to be one of the first two parameters ? it's really weird that it's placed in the middle of the uint64 parameters.
(the function is perfectly fine otherwise )

data/bookkeeping/block.go
```diff
@@ -121,6 +121,21 @@ type (
 		// transactions have ever been committed (since TxnCounter
 		// started being supported).
 		TxnCounter uint64 `codec:"tc"`
+
+		// CompactCertVoters is the root of a Merkle tree containing the
+		// online accounts that will help sign a compact certificate.  The
+		// Merkle root, and the compact certificate, happen on blocks that
+		// are a multiple of ConsensusParams.CompactCertRounds.
```
tsachiherman, line 128: could you please add a note that on rounds that aren't multiple of ConsensusParams.CompactCertRounds, the `CompactCertVoters` would contains a zeroed out digest ?

ledger/eval.go
```diff
@@ -762,6 +763,18 @@ func (eval *BlockEvaluator) endOfBlock() error {
 		} else {
 			eval.block.TxnCounter = 0
 		}
+
+		if eval.proto.CompactCertRounds > 0 && eval.block.Round()%basics.Round(eval.proto.CompactCertRounds) == 0 {
```
tsachiherman, line 767: consider changing the CompactCertVoters(loopback) so that it have a similar signature to something like:
```
err := CompactCertVoters(basics.Round, *CompactCertVoters, *CompactCertVotersTotal)
```

so that you could collapse the logic you’ve added to finalValidation and endOfBlock into a single location.

ledger/voters.go
```diff
@@ -0,0 +1,337 @@
...
+	"sync"
+
+	"github.com/algorand/go-algorand/config"
+	"github.com/algorand/go-algorand/crypto"
+	"github.com/algorand/go-algorand/crypto/compactcert"
+	"github.com/algorand/go-algorand/crypto/merklearray"
+	"github.com/algorand/go-algorand/data/basics"
+	"github.com/algorand/go-algorand/data/bookkeeping"
+
+	"github.com/algorand/go-deadlock"
```
tsachiherman, line 30: in other files we usually place the go-deadlock before the go-algorand. I don't know why.

config/consensus.go
```diff
@@ -292,6 +292,34 @@ type ConsensusParams struct {
...
+	// publishing a Merkle commitment to the online accounts.  Namely,
+	// if block number N contains a Merkle commitment to the online
+	// accounts (which, incidentally, means N%CompactCertRounds=0),
+	// then the balances reflected in that commitment must come from
+	// block N-CompactCertVotersLookback.  This gives each node some
+	// time (CompactCertVotersLookback blocks worth of time) to
+	// construct this Merkle tree, so as to avoid placing the
+	// construction of this Merkle tree (and obtaining the requisite
+	// accounts and balances) in the critical path.
+	CompactCertVotersLookback uint64
```
tsachiherman, line 322: Do you have any suggestion for good numbers for `CompactCertVotersLookback` and `CompactCertRounds` ?

ledger/voters.go
```diff
@@ -0,0 +1,337 @@
...
+	if err != nil {
+		return err
+	}
+
+	participants := make(participantsArray, len(top))
+	addrToPos := make(map[basics.Address]uint64)
+	var totalWeight basics.MicroAlgos
+
+	for i, acct := range top {
+		var ot basics.OverflowTracker
```
tsachiherman, line 215: nit: move the `basics.OverflowTracker` outside the for loop.

data/bookkeeping/block.go
zeldovich: Good idea, thanks!

data/basics/userBalance.go
zeldovich: Yeah, that would be clearer indeed; thanks!

ledger/voters.go
zeldovich: Huh, yeah, I'll make that consistent.

config/consensus.go
zeldovich: `CompactCertVotersLookback` is relatively easy: something like 16 should be fine.

`CompactCertRounds` depends a lot on where we imagine using these compact certs, because it translates into how often you have to check compact certs (per unit time) to keep up with the state of the chain.  I've been thinking of this being somewhere around 128-1024.  Probably not smaller.  Perhaps a bit longer depending on the use case (e.g., I could imagine this being 8K or 16K, so a compact cert once a day..)

ledger/voters.go
zeldovich: You're right that we could move it out (since, on each iteration of the loop, either `ot` remains non-overflowed, or the loop returns an error).  But I think that would make the code harder to read, because `ot` is used independently for each iteration of the loop, and it's not used after the loop.

ledger/eval.go
zeldovich: The reason the `Ledger.CompactCertVoters()` function returns the entire `VotersForRound` is that I need other parts from that `VotersForRound` in the rest of the code.  Here's the complete compactcert implementation, if it's helpful, and the other place that I need to use `Ledger.CompactCertVoters()`: https://github.com/zeldovich/go-algorand/blob/compactcert/compactcert/builder.go#L48

But I agree it'd be good to factor out this common logic.  Pushed that change.

zeldovich:
Ah, good point about resetting the votersTracker on `accountUpdates.loadFromDisk()`.  I pushed a fix for that, just re-allocating a fresh `votersTracker` as you suggested.

I think this is more-or-less good to go, then.  I'll rebase against master and remove the draft flag when #1359 and #1361 get approved and merged.

ledger/voters.go
```diff
@@ -0,0 +1,337 @@
...
+func (vt *votersTracker) getVoters(r basics.Round) (*VotersForRound, error) {
+	tr, ok := vt.round[r]
+	if !ok {
+		// Not tracked: compact certs not enabled.
+		return nil, nil
+	}
+
+	// Wait for the Merkle tree to be constructed.
+	tr.mu.Lock()
+	for tr.Tree == nil {
```
tsachiherman, line 314: this could lead to a hang in case we had error during the `VotersForRound::loadTree`. maybe store the error generated during the `VotersForRound::loadTree` inside the `VotersForRound`, and make sure to broadcast when changing the error as well ?

zeldovich: That seems like a good plan.  I think originally I had no failure paths in the background `loadTree()` function, but indeed now there's possible errors that can cause this to hang forever.  Pushed a revised version.

algorandskiy (commented):
Looks good to me

data/basics/userBalance.go
```diff
@@ -381,6 +381,14 @@ func (u AccountData) Money(proto config.ConsensusParams, rewardsLevel uint64) (m
 	return e.MicroAlgos, e.RewardedMicroAlgos
 }
 
+// PendingRewards computes the amount of rewards (in microalgos) that
+// have yet to be added to the account balance.
+func PendingRewards(ot *OverflowTracker, proto config.ConsensusParams, microAlgos MicroAlgos, rewardsBase uint64, rewardsLevel uint64) MicroAlgos {
+	rewardsUnits := microAlgos.RewardUnits(proto)
+	rewardsDelta := ot.Sub(rewardsLevel, rewardsBase)
+	return MicroAlgos{Raw: ot.Mul(rewardsUnits, rewardsDelta)}
+}
```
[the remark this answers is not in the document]
zeldovich: Indeed it's basically the same code.  The reason I didn't merge them is that I liked the detailed panic message that describes the internal computation steps that overflowed (in the off chance that we do trigger overflow), and the function below does one more overflow-checked addition that could overflow and cause that message to be printed (so it's not just a matter of moving the panic-on-error into this new `PendingRewards()`)..

This code is rather deeply embedded in our protocol now, so it seemed not so bad to have a copy of this code, since it's not so likely to change.

algonautshant (approved):
Looks good. 
I have couple of questions to better understand the code.

ledger/voters.go
```diff
@@ -0,0 +1,351 @@
...
+	"github.com/algorand/go-algorand/crypto/compactcert"
+	"github.com/algorand/go-algorand/crypto/merklearray"
+	"github.com/algorand/go-algorand/data/basics"
+	"github.com/algorand/go-algorand/data/bookkeeping"
+)
+
+// The votersTracker maintains the Merkle tree for the most recent
+// commitments to online accounts for compact certificates.
+//
+// We maintain multiple Merkle trees: we might commit to a new Merkle tree in
```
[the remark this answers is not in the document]
zeldovich: What the comment refers to is that there might be multiple elements in the `round` map, and each one has a Merkle tree.

ledger/voters.go
```diff
@@ -0,0 +1,351 @@
...
+	//
+	// The first is to maintain the tree for an upcoming block -- that is,
+	// if X+Loookback<Latest.  The block evaluator can ask for the root of
+	// the tree to propose and validate a block.
+	//
+	// The second is to construct compact certificates.  Compact certificates
+	// are formed for blocks that are a multiple of CompactCertRounds, using
+	// the Merkle commitment to online accounts from the previous such block.
+	// Thus, we maintain X in the round map until we form a compact certificate
+	// for round X+CompactCertVotersLookback+CompactCertRounds.
```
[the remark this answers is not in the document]
zeldovich: It is indeed the case that `CompactCertVotersLookback` is on the order of 16, and `CompactCertRounds` is probably something like 128 or 1024.

However, I believe the comment is correct: the two intervals have to be added together.  We need the compact cert to certify the next block that is a multiple of `CompactCertRounds`, because that block will contain the commitment to the next group of voters (which will enable verifying the next compact cert, etc).  And the `CompactCertVotersLookback` addition is needed to give the ledger time to compute the Merkle tree, in case there are many voters.

PR #1452, opened 2020-08-26
export RekeyTo txn field in the algod REST API

zeldovich:
Looks like the RekeyTo field didn't get propagated into the REST API when it was added to transactions.

### DIFF
--- a/daemon/algod/api/server/v1/handlers/handlers.go
+++ b/daemon/algod/api/server/v1/handlers/handlers.go
@@ -114,6 +114,10 @@ func txEncode(tx transactions.Transaction, ad transactions.ApplyData) (v1.Transa
 		res.Lease = tx.Lease[:]
 	}
 
+	if !tx.RekeyTo.IsZero() {
+		res.RekeyTo = tx.RekeyTo.String()
+	}
+
 	return res, nil
 }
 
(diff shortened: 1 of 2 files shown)


algorandskiy:
It is available in API v2:
1. in account data - https://github.com/algorand/go-algorand/blob/master/daemon/algod/api/algod.oas2.json#L1151
2. in pending txn response as raw txn object - https://github.com/algorand/go-algorand/blob/master/daemon/algod/api/algod.oas2.json#L2064

zeldovich:
The first is really about the account state, not the transaction.

The latter means you have to decode the raw transaction with msgpack, as far as I can tell?  Which is OK, and perhaps this is what we're expecting the callers of the API to do (I don't know).  But the v1 API explicitly broke out each transaction field into a separate value in the JSON response, so the app didn't have to worry about msgpack to examine transactions.

I should say, I don't actually need this field for anything I'm doing; just thought our users might find it convenient.  But if our users are using the v2 API and decoding msgpack works for them, then this PR isn't needed.

pbennett:
Is decoding the msgpack even cleanly in the v2 api?  I want to say from prior hacking I had to hack things a fair bit to decode the msgpack via the v2 api only (I think a lot of it is private).
I think if users have to crack the raw block to get base fields then the API is missing fields.

algorotem:
`v2` has it in both forms (`msgpack` and `json`. It allows you to choose at the API layer). `v1` is deprecated by now so we don't add new features to it (and therefore the rekey isn't there). I'd have merged this one but there is more work to do here (updating the swagger file, the developer website, docs, etc), and given the deprecation and `v2`, I don't think it worth the time.

PR #1456, opened 2020-08-26
Add tealdbg to installation packages

algorandskiy:
## Summary

Add tealdbg to installation packages

## Test Plan

No tests

### DIFF
--- a/installer/rpm/algorand.spec
+++ b/installer/rpm/algorand.spec
@@ -27,7 +27,7 @@ This package provides an implementation of the Algorand protocol.
 mkdir -p %{buildroot}/usr/bin
 # NOTE: keep in sync with scripts/build_deb.sh bin_files
 # NOTE: keep in sync with %files section below
-for f in algocfg algod algoh algokey carpenter catchupsrv ddconfig.sh diagcfg goal kmd msgpacktool node_exporter; do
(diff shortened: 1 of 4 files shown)


algonautshant:
Is make install also picking this up?

algorandskiy:
I agree but this is a separate issue.

algorandskiy:
@algonautshant I think so, it implicitly calls `scripts/build_package.sh`

Priegle:
> I'm not sure if I like this change; and it's not that I think that we shouldn't distribute the tealdbg.
> 
> I think that we should redesign the packaging so that we will have a separate package for the core components ( i.e. algoh, algod, etc. ), and another package (let's call it algorand-dev ) for developers ( i.e. carpenter, catchupsrv, msgpacktool, tealdbg and maybe additional tools that are currently only available in "tools" ).
> 
> We might not be able to deliver the development tools for all the platforms, but we still want to have the core installation to work.
> 
> The rationale here is that our installation package is pretty big already, and most of the users would never use these tools - so why do we force them having them ?

@tsachiherman I think this makes sense, but obviously more work.  I'd like to get the debugger out there as quickly as possible as its a pretty critical component of writing stateful smart contracts.  What about a plan that gets it out now and then later we pull out the tools?

PR #1457, opened 2020-08-26
Bugfix: Reorder permission check for setting contract administrator.

derbear:
This fixes a security vulnerability where the contract-administrator
status of an address was checked after it was set.  Any address could
cause itself to be a contract administrator, regardless of whether it
were a contract administrator already.

This bugfix checks whether an application call's sender is a contract
administrator before modifying a target user's contract-administrator
status.

Thanks to @jasonpaulos for helping to find this bug.

## Test Plan

Added test case to e2e test.


jasonpaulos (dismissed):
The program looks good to me

PR #1462, opened 2020-08-31
Bugfix: fix error propagation at two sites.

derbear:
Correctly return errors in agreement/proposal.go and
ledger/apply/payment.go.

## Test Plan

Used existing tests.


derbear:
> Nice catches Derek

Thanks, these were leftover bugs found when @justicz ran `ineffassign` a while ago.

PR #1464, opened 2020-09-01
Adding tests for uncovered statements in data/transactions/logic

algonautshant:
Adding tests for: 
TxnFieldToTealValue
MinTealVersion Parameter check for Eval and Check
OpcodesByVersion branch where the opcode version order is changed

### DIFF
--- a/data/transactions/logic/opcodes_test.go
+++ b/data/transactions/logic/opcodes_test.go
@@ -84,12 +84,6 @@ func TestOpcodesByVersion(t *testing.T) {
 		OpSpecs2[idx] = cp
 	}
 
-	// To test the case where a newer version opcode is before an older version
-	// Change the order of opcode 0x01 so that version 2 comes before version 1
-	tmp := OpSpecs[1]
-	OpSpecs[1] = OpSpecs[4]
-	OpSpecs[4] = tmp
-
 	opSpecs := make([][]OpSpec, 2)
 	for v := uint64(1); v <= LogicVersion; v++ {
 		t.Run(fmt.Sprintf("v=%d", v), func(t *testing.T) {
--- a/data/transactions/logic/opcodes_test.go
+++ b/data/transactions/logic/opcodes_test.go
@@ -84,6 +84,12 @@ func TestOpcodesByVersion(t *testing.T) {
 		OpSpecs2[idx] = cp
 	}
 
+	// To test the case where a newer version opcode is before an older version
+	// Change the order of opcode 0x01 so that version 2 comes before version 1
+	tmp := OpSpecs[1]
+	OpSpecs[1] = OpSpecs[4]
+	OpSpecs[4] = tmp
+
 	opSpecs := make([][]OpSpec, 2)
 	for v := uint64(1); v <= LogicVersion; v++ {
 		t.Run(fmt.Sprintf("v=%d", v), func(t *testing.T) {


data/transactions/logic/opcodes_test.go
```diff
@@ -35,6 +35,12 @@ func TestOpSpecs(t *testing.T) {
 func TestOpcodesByVersion(t *testing.T) {
 	t.Parallel()
 
+	// To test the case where a newer version opcode is before an older version
```
brianolson, line 38: make this not Parallel() since it is modifying a global resource. I think the comment on OpcodesByVersion() may be out of date and my reading of that function right now, I don't see where it modifies the OpSpecs global? We could add to this test something that makes a copy of the global OpSpecs, runs OpcodesByVersion(), and checks that the global OpSpecs matches the original copy?
Then, at the end of this test, undo the swap and put OpSpecs back so that no other code sees any difference.

algorandskiy: >  I think the comment on OpcodesByVersion() may be out of date and my reading of that function right now

This is correct, I changed it but forgot to update the comment.

data/transactions/logic/eval_test.go
```diff
@@ -110,6 +110,41 @@ func TestEmptyProgram(t *testing.T) {
...
+	txn := transactions.Transaction{}
+	value := uint64(9999)
+	txn.FirstValid = basics.Round(value)
+	groupIndex := 0
+	field := FirstValid
+
+	tealValue, err := TxnFieldToTealValue(&txn, groupIndex, field)
+	require.Equal(t, basics.TealUintType, tealValue.Type)
+	require.Equal(t, value, tealValue.Uint)
+	fmt.Printf("%+v\n", tealValue)
```
algorandskiy, line 144: debug print

data/transactions/logic/opcodes_test.go
```diff
@@ -35,6 +35,12 @@ func TestOpSpecs(t *testing.T) {
...
+	// Change the order of opcode 0x01 so that version 2 comes before version 1
+	tmp := OpSpecs[1]
+	OpSpecs[1] = OpSpecs[4]
+	OpSpecs[4] = tmp
+
```
algorandskiy, line 43: I think this should be a separate test case with the shared implementation with the current one

data/transactions/logic/eval_test.go
algonautshant: fixed.

data/transactions/logic/opcodes_test.go
algonautshant: Done.

data/transactions/logic/opcodes_test.go
algonautshant: done.

data/transactions/logic/eval_test.go
```diff
@@ -110,6 +110,40 @@ func TestEmptyProgram(t *testing.T) {
...
+	txn := transactions.Transaction{}
+	value := uint64(9999)
+	txn.FirstValid = basics.Round(value)
+	groupIndex := 0
+	field := FirstValid
+
+	tealValue, err := TxnFieldToTealValue(&txn, groupIndex, field)
+	require.Equal(t, basics.TealUintType, tealValue.Type)
+	require.Equal(t, value, tealValue.Uint)
+	require.NoError(t, err)
```
tsachiherman, line 144: nit: move the NoError to be the first. If there is an error, the previous two tests are meaningless.

data/transactions/logic/eval_test.go
```diff
@@ -110,6 +110,40 @@ func TestEmptyProgram(t *testing.T) {
...
+	pass, err := Eval(program, params)
+	require.Error(t, err)
+	require.Contains(t, err.Error(), fmt.Sprintf("program version must be >= %d", appsEnabledVersion))
+	require.False(t, pass)
+}
+
+func TestTxnFieldToTealValue(t *testing.T) {
+
+	txn := transactions.Transaction{}
+	value := uint64(9999)
```
tsachiherman, line 136: nit: the test is not wrong, but it's a classic for testing several key values : 0, 1, 2, 2^32-1, 2^64-1

data/transactions/logic/opcodes_test.go
```diff
@@ -32,8 +32,57 @@ func TestOpSpecs(t *testing.T) {
...
+func TestOpcodesByVersionReordered(t *testing.T) {
+
+	// Make a copy to restore to the original
+	OpSpecsOrig := make([]OpSpec, len(OpSpecs))
+	for idx, opspec := range OpSpecs {
+		cp := opspec
+		OpSpecsOrig[idx] = cp
+	}
+	defer (func() {
```
tsachiherman, line 66: why the round brackets around the deferred function definition ?

algonautshant: fixed.

data/transactions/logic/eval_test.go
algonautshant: added the test for the extra values.

data/transactions/logic/eval_test.go
algonautshant: got it!

data/transactions/logic/eval_test.go
```diff
@@ -110,6 +110,42 @@ func TestEmptyProgram(t *testing.T) {
...
+	require.Contains(t, err.Error(), fmt.Sprintf("program version must be >= %d", appsEnabledVersion))
+	require.False(t, pass)
+}
+
+func TestTxnFieldToTealValue(t *testing.T) {
+
+	txn := transactions.Transaction{}
+	groupIndex := 0
+	field := FirstValid
+	values := [6]uint64{0, 1, 2, 2 ^ 32 - 1, 2 ^ 64 - 1}
```
tsachiherman, line 138: I didn't mean explicitly 2 ^ 32 - 1.. I meant 
```
values := [6]uint64{0, 1, 2, 0xffffffff, uint64(0xffffffffffffffff)}
```

algonautshant: fixed.

algonautshant:
No more changes here. This PR was merged with:
https://github.com/algorand/go-algorand/pull/1466

algonautshant:
No more changes here. This PR was merged with:
https://github.com/algorand/go-algorand/pull/1466

PR #1466, opened 2020-09-02
Shant/debugger tests

algonautshant:
Adding debugger coverage test.

### DIFF
--- a/data/transactions/logic/debugger_test.go
+++ b/data/transactions/logic/debugger_test.go
@@ -18,8 +18,10 @@ package logic
 
 import (
 	"os"
+	"encoding/base64"
 	"testing"
 
+	"github.com/algorand/go-algorand/data/basics"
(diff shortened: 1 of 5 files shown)


data/transactions/logic/debugger_test.go
```diff
@@ -125,3 +125,28 @@ func TestDebuggerHook(t *testing.T) {
 	require.Greater(t, testDbg.update, 1)
 	require.Equal(t, 1, len(testDbg.state.Stack))
 }
+
+func TestLineToPC(t *testing.T) {
```
tsachiherman, line 129: Add missing test case for LineToPC(`negative value`) and a high value integer, i.e. 0x7fffffff

algonautshant: Added.

algorandskiy (dismissed):

tsachiherman (approved):

PR #1467, opened 2020-09-02
Add versions endpoint to spec, fix versions model objects.

winder:
## Summary

Adds and fixes the versions endpoint and models in the spec. Only used for documentation.


daemon/algod/api/algod.oas2.json
```diff
@@ -96,6 +96,26 @@
         }
       }
     },
+    "/versions": {
+      "get": {
+        "description": "Retrieves the current version",
```
algorotem, line 101: Is that the correct description? 
I'd expect it to return an array of currently supported *API* versions.

The build is returned by the status

winder: I copied this from the v1 swagger file so it is hopefully correct. Double checking, it looks like versions returns both:

```
curl -s "localhost:8080/versions" -H "Authorization: Bearer 1422fe0dfc8e30ad88863b66849db35ba5b82729d34945aa98c05b36becf2a0c" |json_pp
{
   "build" : {
      "branch" : "will/genesis-endpoint",
      "build_number" : 83115,
      "channel" : "dev",
      "commit_hash" : "35821726+",
      "major" : 2,
      "minor" : 1
   },
   "genesis_hash_b64" : "sC3P7e2SdbqKJK0tbiCdK9tdSpbe6XeCGKdoNzmlj0E=",
   "genesis_id" : "devnet-v1.0",
   "versions" : [
      "v1",
      "v2"
   ]
}
```

algorotem: Sounds good to me. 
Since it's a new one - can we call the api version `api_versions` and not just `versions`?

algorotem: We should also probably change the description to something more specific along the lines of  "Retrieves the current API supported versions, binary build versions, and genesis info."

winder: The endpoint hasn't been changed, It was just missing from the documentation. So unfortunately we can't change the contents here.

algorotem: but we can make the description better. Let's do that and merge it in.

daemon/algod/api/algod.oas2.json
```diff
@@ -98,7 +98,7 @@
     },
     "/versions": {
       "get": {
-        "description": "Retrieves the current version",
+        "description": "Retrieves the API version, binary build versions, and genesis information.",
```
algorotem, line 101: just to be accurate - `Suported API versions`

algorotem (dismissed):
Approved. Let's just fix the comment.

algorotem (approved):

PR #1468, opened 2020-09-02
REST API - new endpoint /genesis.json

winder:
## Summary

Adds a new `/genesis.json` endpoint. It is occasionally useful to fetch the genesis.json file from a running instance. For example when configuring `indexer`, working with private networks or working in a sandbox.

## Testing

New e2e test to make sure the endpoint is returning genesis information.


cmd/algod/main.go
```diff
@@ -289,7 +289,7 @@ func main() {
 		}
 	}
 
-	err = s.Initialize(cfg, phonebookAddresses)
+	err = s.Initialize(cfg, phonebookAddresses, string(genesisText[:]))
```
brianolson, line 292: probably don't need `[:]`
why not reference into `github.com/algorand/go-algorand/daemon/algod/api/server/lib` here and skip bouncing off daemon/algod/server.go ?

winder: Thanks, I thought I had tried that. I didn't want to add another import to main.go which is why I opted to pass it into the initialize function.

algorotem (approved):

daemon/algod/api/algod.oas3.yml
```diff
@@ -1359,6 +1359,32 @@
   },
   "openapi": "3.0.1",
   "paths": {
+    "/genesis": {
+      "get": {
+        "description": "Returns the entire genesis file in json.",
+        "operationId": "SwaggerJSON",
```
algorotem, line 1365: is that a leftover from the Swagger one?

winder: Ack, fixed.

algorotem (approved):

PR #1476, opened 2020-09-03
Update rpm packaging

btoll:
## Summary

This updates the RPM packaging to allow for the following scenarios:

- Package a custom version from host filesystem:

    + `VERSION=2.1.8.976 PACKAGES_DIR=rpm_dir mule -f package-deploy.yaml package-deploy-rpm`

- Package a custom version and expose it to the host filesystem instead of pushing to S3:

    + `VERSION=2.1.8.976 NO_DEPLOY=true PACKAGES_DIR=rpm_dir mule -f package-deploy.yaml package-deploy-rpm`

- Package latest version and push to s3:algorand-releases (pulls latest version from staging):

    + `VERSION=2.1.6 mule -f package-deploy.yaml package-deploy-rpm`

- Package specified version and push to s3:algorand-releases (pulls latest version from staging):

    + `VERSION=2.1.1 mule -f package-deploy.yaml package-deploy-rpm`

In addition, it removes any calls in the containers to `docker` and `mule`.  This greatly simplifies things.

## Test Plan

`mule -f package-deploy.yaml package-deploy-rpm` with differing env vars prepended on cli.


tsachiherman:
I can't really see from this change if it would work or not, so I'll have to take your word on this one.
I think that we should have some automated testing for this, so we can detect future regressions.

tsachiherman:
I can't really see from this change if it would work or not, so I'll have to take your word on this one.
I think that we should have some automated testing for this, so we can detect future regressions.

btoll:
> I can't really see from this change if it would work or not, so I'll have to take your word on this one.
> I think that we should have some automated testing for this, so we can detect future regressions.

I changed the name of the ticket from "Fix" to "Update", as this is much more in line with what the scope is.  There was an issue that was causing an error, but that is addressed in this ticket here: https://github.com/algorand/go-algorand/pull/1469.

The current method to update the rpm packaging works, but it's a bit separate from the rest of the build process.  This ticket and others are going to bring that process more in line with the current build process.

I agree that there should be more tests, and I'm currently working on some.  However, I think that by misrepresenting the scope of this work I caused some confusion.

onetechnical:
There's been a lot of changes to RPM packaging over the past week, so I think we should re-vet this.

egieseke (approved):
Looks good.  However, I think that we should get in the habit of including a header describing the purpose of the script or yaml file, and some overview of how it works.

btoll:
> Looks good. However, I think that we should get in the habit of including a header describing the purpose of the script or yaml file, and some overview of how it works.

Yes, I agree.  My plan is to include a README for each mule task that explains what env vars are important to each script.  Part of this can then be copied to the shell scripts themselves.

bricerisingalgorand (approved):

PR #1480, opened 2020-09-03
Add support for algorelay deleting unused SRV and DNS entries

tsachiherman:
## Summary

This PR adds the deletion of a non-functional SRV/DNS relay entry in the service records.

This didn't used to be an issue, since relays were always only switching from non-functional to functional state. However, since relays can now become non-functional - we need an automated tool to help us removing these entries.

### DIFF
--- a/cmd/algorelay/relayCmd.go
+++ b/cmd/algorelay/relayCmd.go
@@ -228,6 +228,7 @@ var checkCmd = &cobra.Command{
 				if enabledRelaysDNSAlias[alias] {
 					continue
 				}
+
 				if relaysDNSAlias[alias] {
 					fmt.Printf("WARN : disabled relay %s has a _algobootstrap entry\n", bootstrap)
 				} else {
@@ -291,11 +292,6 @@ var updateCmd = &cobra.Command{
 				continue
 			}
 
-			if !relay.CheckSuccess {
-				fmt.Printf("[%d] OK: Skipping NotSuccessful %s\n", relay.ID, relay.Address)
-				// Don't output results if skipped
-				continue
-			}
 			const checkOnly = false
 			name, port, err := ensureRelayStatus(checkOnly, relay, nameDomainArg, srvDomainArg, defaultPortArg, context)
 			if err != nil {
@@ -307,7 +303,11 @@ var updateCmd = &cobra.Command{
 				})
 				anyUpdateError = true
 			} else {
-				fmt.Printf("[%d] OK: %s -> %s:%d\n", relay.ID, relay.Address, name, port)
+				if relay.CheckSuccess {
+					fmt.Printf("[%d] OK: %s -> %s:%d\n", relay.ID, relay.Address, name, port)
+				} else {
+					fmt.Printf("[%d] OK: %s removed ( if it was there )\n", relay.ID, relay.Address)
+				}
 				results = append(results, checkResult{
 					ID:      relay.ID,
 					Success: true,
@@ -380,6 +380,7 @@ func ensureRelayStatus(checkOnly bool, relay eb.Relay, nameDomain string, srvDom
 	}
 
 	targetDomainAlias := relay.DNSAlias + "." + nameDomain
+
 	if topmost != targetDomainAlias {
 		if checkOnly {
 			err = fmt.Errorf("topmost DNS name is not the assigned DNS Alias (wanted: %s, found %s)",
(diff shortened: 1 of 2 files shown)


cmd/algorelay/relayCmd.go
```diff
@@ -307,7 +303,11 @@ var updateCmd = &cobra.Command{
 				})
 				anyUpdateError = true
 			} else {
-				fmt.Printf("[%d] OK: %s -> %s:%d\n", relay.ID, relay.Address, name, port)
+				if relay.CheckSuccess {
```
algonautshant, line 306: If I understand correctly, this output seems to be redundant. 
Similar information is already output in ensureRelayStatus.

cmd/algorelay/relayCmd.go
```diff
@@ -415,50 +432,85 @@ func ensureRelayStatus(checkOnly bool, relay eb.Relay, nameDomain string, srvDom
...
+			// Add SRV entry to map to our DNSAlias
+			err = addSRVRecord(ctx.bootstrap.networkName, topmost, port, ctx.bootstrap.shortName, ctx.srvZoneID)
 			if err != nil {
 				return
 			}
-			fmt.Printf("[%d] Added metrics SRV Record: %s:%d\n", relay.ID, targetDomainAlias, metricsPort)
+			fmt.Printf("[%d] Added boostrap SRV Record: %s:%d\n", relay.ID, targetDomainAlias, port)
+		}
+	} else {
+		if matchCount > 0 {
```
algonautshant, line 465: We are not checking the err in this branch. 
Can you please add a comment if it need not be checked?

cmd/algorelay/relayCmd.go
tsachiherman: The `ensureRelayStatus` doesn't always output text, which is why this is needed.

cmd/algorelay/relayCmd.go
tsachiherman: The err here is not a good indicator to whether we should delete entries. The err is nil only in the case where the number of SRV entries is 1. This isn't what we're looking to get. ( and honestly, it's quite meaningless in this context. )

cmd/algorelay/relayCmd.go
```diff
@@ -380,26 +380,43 @@ func ensureRelayStatus(checkOnly bool, relay eb.Relay, nameDomain string, srvDom
...
+		} else {
+			// remove entry.
+			err = deleteDNSRecord(targetDomainAlias, topmost, ctx.nameZoneID)
+			if err != nil {
+				return
+			}
+			fmt.Printf("[%d] Removed DNS Record: %s -> %s\n", relay.ID, targetDomainAlias, topmost)
+		}
+	} else if !relay.CheckSuccess {
+		// remove entry.
```
algonautshant, line 411: In the other branch (topmost != targetDomainAlias), it is deleted only if checkOnly is false. 
Why checkOnly is not considered here?

cmd/algorelay/relayCmd.go
```diff
@@ -415,50 +432,85 @@ func ensureRelayStatus(checkOnly bool, relay eb.Relay, nameDomain string, srvDom
...
+			err = addSRVRecord(ctx.bootstrap.networkName, topmost, port, ctx.bootstrap.shortName, ctx.srvZoneID)
 			if err != nil {
 				return
 			}
-			fmt.Printf("[%d] Added metrics SRV Record: %s:%d\n", relay.ID, targetDomainAlias, metricsPort)
+			fmt.Printf("[%d] Added boostrap SRV Record: %s:%d\n", relay.ID, targetDomainAlias, port)
+		}
+	} else {
+		if matchCount > 0 {
+			err = clearSRVRecord(ctx.bootstrap.networkName, topmost, ctx.bootstrap.shortName, ctx.srvZoneID)
```
algonautshant, line 466: Similarly, checkOnly is not considered for clearSRVRecord (in all 3 cases below). 
Is this what was intended?

cmd/algorelay/relayCmd.go
tsachiherman: if you'll look carefully, you'll see that `ensureRelayStatus` is called with `checkOnly=true` only when `relay.CheckSuccess=true`. ( see `checkCmd` and `updateCmd`

algonautshant (approved):
Looks good.

PR #1482, opened 2020-09-03
Allow max and assetIndex to be optional

winder:
## Summary

The documentation says that these fields are not required, but there is a parsing error if they aren't provided.

## Test Plan

I ran the command locally with and without the fix, this is the v1 endpoint so I didn't want to spend too much time on testing.

### DIFF
--- a/daemon/algod/api/server/v1/handlers/handlers.go
+++ b/daemon/algod/api/server/v1/handlers/handlers.go
@@ -1321,20 +1321,28 @@ func Assets(ctx lib.ReqContext, context echo.Context) {
 
 	const maxAssetsToList = 100
 
+	var err error
+	var max int64 = maxAssetsToList
+	var assetIdx int64 = 0
+
 	// Parse max assets to fetch from db
-	max, err := strconv.ParseInt(r.FormValue("max"), 10, 64)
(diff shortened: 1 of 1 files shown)


algorotem:
I believe that it would be easier to return a better error message and correct the documentation. The limit is there for a reason - this can be potentially a very long list.

winder:
There is still a limit to return 100 assets, I'm just changing the no argument form such that `/v1/assets` is equivelent to `/v1/assets?max=100&assetIndex=0`. I could also set the default value to something smaller, how about 10?

Alternatively I could change the doc string, and make another PR to the docs repo manually editing the v1 REST documentation.

algorandskiy (approved):
Thank you for fixing it

rotemh:
> There is still a limit to return 100 assets, I'm just changing the no argument form such that `/v1/assets` is equivelent to `/v1/assets?max=100&assetIndex=0`. I could also set the default value to something smaller, how about 10?
> 
> 
> 
> Alternatively I could change the doc string, and make another PR to the docs repo manually editing the v1 REST documentation.

100 is a good number. I'll approve and merge it

rotemh (approved):

PR #1485, opened 2020-09-04
Add a benchmark to test restoring from catchpoint performance

tsachiherman:
## Summary

Before attempting to improve the catchpoint catchup performance, we need a way to systematically test how well we perform. Running this test, produces the following output:
```
goos: darwin
Add a benchmark to test restoring from catchpoint performance.
goarch: amd64
BenchmarkRestoringFromCatchpointFile/Restore-102400-8         	  102400	     64479 ns/op
--- BENCH: BenchmarkRestoringFromCatchpointFile/Restore-102400-8
    catchupaccessor_test.go:99: Last 64K	4.949638275s
BenchmarkRestoringFromCatchpointFile/Restore-204800-8         	  204800	     86044 ns/op
--- BENCH: BenchmarkRestoringFromCatchpointFile/Restore-204800-8
    catchupaccessor_test.go:99: Last 64K	7.556581895s
BenchmarkRestoringFromCatchpointFile/Restore-409600-8         	  409600	    105683 ns/op
--- BENCH: BenchmarkRestoringFromCatchpointFile/Restore-409600-8
    testingLogger.go:38: time="2020-09-04T12:24:42.944604 -0400" level=warning msg="dbatomic: tx surpassed expected deadline by 11.728647ms" callee="github.com/algorand/go-algorand/ledger.(*CatchpointCatchupAccessorImpl).processStagingBalances.func1" caller="/Users/tsachi/go/src/github.com/algorand/go-algorand/ledger/catchupaccessor.go:294" file=dbutil.go function="github.com/algorand/go-algorand/util/db.(*Accessor).atomic" line=355 readonly=false
    catchupaccessor_test.go:99: Last 64K	9.041332973s
PASS
ok  	github.com/algorand/go-algorand/ledger	68.104s
```

by itself, it's not good or bad. But it's a measurable starting point for performance tuning.

### DIFF
--- a/ledger/catchupaccessor_test.go
+++ b/ledger/catchupaccessor_test.go
@@ -1,0 +1,111 @@
+// Copyright (C) 2019-2020 Algorand, Inc.
+// This file is part of go-algorand
+//
+// go-algorand is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as
+// published by the Free Software Foundation, either version 3 of the
+// License, or (at your option) any later version.
+//
+// go-algorand is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with go-algorand.  If not, see <https://www.gnu.org/licenses/>.
+
+package ledger
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/algorand/go-algorand/config"
+	"github.com/algorand/go-algorand/crypto"
+	"github.com/algorand/go-algorand/data/basics"
+	"github.com/algorand/go-algorand/logging"
+	"github.com/algorand/go-algorand/protocol"
+)
+
+func benchmarkRestoringFromCatchpointFileHelper(b *testing.B) {
+	genesisInitState, _ := testGenerateInitState(b, protocol.ConsensusCurrentVersion)
+	const inMem = false
+	log := logging.TestingLog(b)
+	cfg := config.GetDefaultLocal()
+	cfg.Archival = false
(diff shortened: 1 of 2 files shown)


ledger/catchupaccessor_test.go
```diff
@@ -0,0 +1,111 @@
...
+
+	"github.com/algorand/go-algorand/config"
+	"github.com/algorand/go-algorand/crypto"
+	"github.com/algorand/go-algorand/data/basics"
+	"github.com/algorand/go-algorand/logging"
+	"github.com/algorand/go-algorand/protocol"
+)
+
+func benchmarkRestoringFromCatchpointFileHelper(b *testing.B) {
+	genesisInitState, _ := testGenerateInitState(b, protocol.ConsensusCurrentVersion)
```
algonautshant, line 37: Is there any reason to run this if b.N is less than 64*1024?
It is not called with such a value, but maybe just return here in case the value is less than 64*1024.

ledger/catchupaccessor_test.go
```diff
@@ -0,0 +1,111 @@
...
+			crypto.RandBytes(randomAccount.Address[:])
+			balances.Balances[i] = randomAccount
+		}
+		err = catchpointAccessor.ProgressStagingBalances(context.Background(), "balances.XX.msgpack", protocol.Encode(&balances), &progress)
+		require.NoError(b, err)
+		accounts += chunkSize
+	}
+	if !last64KStart.IsZero() {
+		last64KDuration := time.Now().Sub(last64KStart)
+		b.Logf("Last 64K\t%v\n", last64KDuration)
```
algonautshant, line 99: Can you also print the b.N size with the duration?

algonautshant (approved):
LGTM

ledger/catchupaccessor_test.go
tsachiherman: The b.N is printed two lines above it.. i.e.
```
BenchmarkRestoringFromCatchpointFile/Restore-102400-8         	  102400	     64479 ns/op
--- BENCH: BenchmarkRestoringFromCatchpointFile/Restore-102400-8
    catchupaccessor_test.go:99: Last 64K	4.949638275s
```
I.e. b.N = `102400`

algonautshant: OK. It's good :-)

PR #1486, opened 2020-09-04
Add unit tests for equal functions in teal.go

algorandskiy:
## Summary

Added tests for `EvalDelta.Equal` and `StateDelta.Equal`

## Test Plan

`EvalDelta.Equal` and `StateDelta.Equal` are 100% covered.

### DIFF
--- a/data/basics/teal_test.go
+++ b/data/basics/teal_test.go
@@ -111,3 +111,115 @@ func TestSatisfiesSchema(t *testing.T) {
 	err = tkv.SatisfiesSchema(schema)
 	a.NoError(err)
 }
+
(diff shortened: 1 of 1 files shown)


tsachiherman (changes requested):
Looks good - two requests, which are basically the same one : 
since both `StateDelta` and `EvalDelta.GlobalDelta`/`EvalDelta.LocalDelta` are maps, they could be a nil. We need to ensure that we test the use case where we compare a nil and an empty non-nil map.

algorandskiy:
Done

tsachiherman (approved):

PR #1493, opened 2020-09-09
Removing close from eof pattern match

algonautshant:
From expect man page:
Both expect and interact will detect when the current process exits
and implicitly do a close.

When eof pattern is detected, there will already be an implicit close.
Closing again will throw an error like: spawn_id: spawn id exp13 not open


tsachiherman (approved):
This pattern is being used in other places as well.. so we probably would need to revisit the usage of that pattern. 
Note that regardless of this, this test need to check the exit code from the goal command, which it doesn't do.

PR #1497, opened 2020-09-10
Add txn.Accounts and txn.ApplicationArgs support to tealdbg

algorandskiy:
## Summary

Tealdbg now can show content of txn.Accounts and txn.ApplicationArgs
Addresses now as shown as standard base32 strings rather than hex-encoded blobs

## Test Plan

### DIFF
--- a/cmd/tealdbg/cdtState.go
+++ b/cmd/tealdbg/cdtState.go
@@ -66,6 +66,29 @@ type cdtStateUpdate struct {
 	appState
 }
 
+type typeHint int
+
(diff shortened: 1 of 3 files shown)


tsachiherman (changes requested):
Change is good, but we need another unit test for TxnFieldToTealValue which exercises arrayFieldIdx != 0.

algorandskiy:
Updated the test, added arrays previews.

tsachiherman (approved):

PR #1499, opened 2020-09-10
Improve RestoringFromCatchpointFile benchmark precision

tsachiherman:
## Summary

This PR moves the preparation of the encoded accounts to be executed before the measuring starts.
This result in more accurate measurements of the actual time it takes to perform the writing.
In addition, it also ensures that the generated addresses are provided in increasing order ( which aligns with the catchpoint file accounts ordering )

output:
```
goos: darwin
goarch: amd64
BenchmarkRestoringFromCatchpointFile/Restore-102400-8         	  102400	     71381 ns/op	     82270 ns/last_64k_account
BenchmarkRestoringFromCatchpointFile/Restore-204800-8         	  204800	     79992 ns/op	     90898 ns/last_64k_account
BenchmarkRestoringFromCatchpointFile/Restore-409600-8         	  409600	     85741 ns/op	    106074 ns/last_64k_account
PASS
ok  	github.com/algorand/go-algorand/ledger	59.774s
```
## Test Plan

Not needed

### DIFF
--- a/ledger/catchupaccessor_test.go
+++ b/ledger/catchupaccessor_test.go
@@ -18,6 +18,7 @@ package ledger
 
 import (
 	"context"
+	"encoding/binary"
 	"fmt"
 	"os"
 	"strings"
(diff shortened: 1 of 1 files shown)


algonautshant (approved):
Looks good.
However, this benchmark is giving up the ability to run the benchmark with larger datasets which do not fit into the memory. 
For that, perhaps keeping the older benchmark is a good idea.

tsachiherman:
> Looks good.
> However, this benchmark is giving up the ability to run the benchmark with larger datasets which do not fit into the memory.
> For that, perhaps keeping the older benchmark is a good idea.

That's true. I don't think that at this point in time it's of much value ( but it might become one day )

PR #1500, opened 2020-09-11
Fix asset flags.

winder:
Fix inconsistencies in `goal asset` command.


cmd/goal/asset.go
```diff
@@ -110,7 +110,7 @@ func init() {
 	addTxnFlags(freezeAssetCmd)
 
 	infoAssetCmd.Flags().Uint64Var(&assetID, "assetid", 0, "ID of the asset to look up")
-	infoAssetCmd.Flags().StringVar(&assetUnitName, "asset", "", "Unit name of the asset to look up")
+	infoAssetCmd.Flags().StringVar(&assetUnitName, "unit", "", "Unit name of the asset to look up")
```
winder, line 113: I'm not sure about changing the flag name. unit is more accurate, but for I think for most people unit is also a fine name to use.

algorandskiy:
I think a new `unit` flag needs to be added and `asset` not removed but marked as deprecated.

cmd/goal/asset.go
algonautshant: How about following the naming convention stablished in:
```
Usage:
  goal asset create [flags]
...
      --unitname string             Name for the unit of asset
...
```

then, asset/unit would be **unitname**

algorotem: Does it actually look it up by asset (Tether USDt) or unit name (USDt)?

winder: It looks up by unit

algorotem: So seems like we should change it or follow Pavel's suggestion, no?

algonautshant (dismissed):
I think it will be better to have the unit flag consistent within goal asset. 
Now, it is --unit here, and --unitname in asset create. 

If --unit is preferred, then changing --unitname to --unit may need it to be deprecated... I understand the complexity. 

This is just a suggestion. 

Otherwise, looks good.

algorotem (approved):

PR #1501, opened 2020-09-11
Change database synchronous mode during fast catchup

tsachiherman:
## Summary

Reduce the database synchronization for the duration of the fast catchup. Reducing the synchronization to Normal, would make it possible to loose a transaction in case of a power failure. This is not crucial for the fast catchup, since we restart loading the accounts into the database when the algod process restarts.

Note that the database performance gains here are substantial - local tests shown 2x-2.5x performance gains.

## Perf gains

Before:
```
BenchmarkRestoringFromCatchpointFile/Restore-102400-8         	  102400	     64230 ns/op	     72418 ns/last_64k_account
BenchmarkRestoringFromCatchpointFile/Restore-204800-8         	  204800	     71893 ns/op	     89333 ns/last_64k_account
BenchmarkRestoringFromCatchpointFile/Restore-409600-8         	  409600	     82651 ns/op	    106174 ns/last_64k_account
```

After:
```
BenchmarkRestoringFromCatchpointFile/Restore-102400-8         	  102400	     26852 ns/op	     32910 ns/last_64k_account
BenchmarkRestoringFromCatchpointFile/Restore-204800-8         	  204800	     33752 ns/op	     44747 ns/last_64k_account
BenchmarkRestoringFromCatchpointFile/Restore-409600-8         	  409600	     40365 ns/op	     60342 ns/last_64k_account
```

## Test Plan

Added a test unit for the new database functionality


util/db/dbutil.go
```diff
@@ -453,3 +442,49 @@ type idemFn func(ctx context.Context, tx *sql.Tx) error
...
+// SetSynchronousMode updates the syncronous mode of the connection
+func (db *Accessor) SetSynchronousMode(ctx context.Context, mode SynchronousMode, fullfsync bool) (err error) {
+	if mode < SynchronousModeOff || mode > SynchronousModeExtra {
+		return fmt.Errorf("invalid value(%d) was provided to mode", mode)
+	}
+	_, err = db.Handle.ExecContext(ctx, fmt.Sprintf("PRAGMA synchronous=%d", mode))
+	if err != nil {
+		return err
+	}
+	if fullfsync {
```
brianolson, line 479: make this statement just be `if mode >= SynchronousModeFull` ? I think we can drop the argument and simplify the callers.

brianolson (commented):
I think I found one way it can be cleaned up, otherwise looks fine.

util/db/dbutil.go
tsachiherman: yes, I was thinking about that. I thought that we might want to have the database tier still provide that flexibility in case we'll want it in the future.
( I do realize that currently it's an extra parameter )

brianolson (approved):

PR #1502, opened 2020-09-14
Test for process return in expect tests

algonautshant:
Added ::AlgorandGoal::CheckProcessReturnedCode to test process return code
Added a test to test CheckProcessReturnedCode

### DIFF
--- a/test/e2e-go/cli/goal/expect/goalExpectCommon.exp
+++ b/test/e2e-go/cli/goal/expect/goalExpectCommon.exp
@@ -57,21 +57,38 @@ proc ::AlgorandGoal::Abort { ERROR } {
 }
 
 # Utility method to test the process returned value
-# Returns 0 when no error code is detected.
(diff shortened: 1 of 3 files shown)


test/e2e-go/cli/goal/expect/goalExpectCommon.exp
```diff
@@ -56,6 +56,24 @@ proc ::AlgorandGoal::Abort { ERROR } {
     exit 1
 }
 
+# Utility method to test the process returned value
```
tsachiherman, line 59: nit: please add a comment:
method returns 1 in case of a normal or abnormal process error or zero otherwise.

tsachiherman:
is looks like this detector works correctly, as it detected the 
```
process killed: SIGHUP hangup
```
however, before merging this in, you'll need to fix the source of the issue.

tsachiherman (approved):
looks great.  thanks for all the changes..

PR #1503, opened 2020-09-14
REST docs update: /v1/account/{addr}/transactions

winder:
## Summary

Update documentation as requested

### DIFF
--- a/daemon/algod/api/server/v1/handlers/handlers.go
+++ b/daemon/algod/api/server/v1/handlers/handlers.go
@@ -1615,9 +1615,7 @@ func Transactions(ctx lib.ReqContext, context echo.Context) {
 	// swagger:operation GET /v1/account/{address}/transactions Transactions
 	// ---
 	//     Summary: Get a list of confirmed transactions.
-	//     Description: Returns the list of confirmed transactions between within a date range. This call is available only when the indexer is running.
-	//     Produces:
-	//     - application/json
+	//     Description: Returns the list of confirmed transactions between within a date range. When indexer is disabled this call requires firstRound and lastRound and returns an error if firstRound is not available to the node. Transactions are returned in round, then block order.
 	//     Schemes:
 	//     - http
 	//     Parameters:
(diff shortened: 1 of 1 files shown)


daemon/algod/api/server/v1/handlers/handlers.go
```diff
@@ -1615,9 +1615,7 @@ func Transactions(ctx lib.ReqContext, context echo.Context) {
 	// swagger:operation GET /v1/account/{address}/transactions Transactions
 	// ---
 	//     Summary: Get a list of confirmed transactions.
-	//     Description: Returns the list of confirmed transactions between within a date range. This call is available only when the indexer is running.
-	//     Produces:
-	//     - application/json
+	//     Description: Returns the list of confirmed transactions between within a date range. When indexer is disabled this call requires firstRound and lastRound and returns an error if firstRound is not available to the node. Transactions are returned in round, then block order.
```
tsachiherman, line 1618: it took me few seconds to figure what you meant by " round, then block order."
( as block order is usually identical to round order )

I think you means in-block or on-block order ( I'm not sure which one is the better way to write this )

cmd/goal/asset.go
```diff
@@ -110,7 +110,7 @@ func init() {
 	addTxnFlags(freezeAssetCmd)
 
 	infoAssetCmd.Flags().Uint64Var(&assetID, "assetid", 0, "ID of the asset to look up")
-	infoAssetCmd.Flags().StringVar(&assetUnitName, "asset", "", "Unit name of the asset to look up")
+	infoAssetCmd.Flags().StringVar(&assetUnitName, "unit", "", "Unit name of the asset to look up")
```
tsachiherman, line 113: unfortunately, it's more than just a doc change, as suggested by the PR title, and it will break the goal backward compatibility.
( not to say it's not a good change. I think it's ok, but we need to make sure we communicate this one clearly )

winder: oops, I only meant to include the doc change. I'll get this part removed, it is part of a different PR.

tsachiherman (approved):
looks good.

PR #1504, opened 2020-09-14
Return 404 if v1/transaction/{txid} doesn't find the txn.

winder:
## Summary

Return 404 instead of 500 if the transaction isn't found.


brianolson (approved):

tsachiherman (approved):
I think that GetRoundByTxID should not have returned an sql error, but rather an internal error.

algorotem (approved):

PR #1507, opened 2020-09-14
Top-level compact certificate support

zeldovich:
This PR integrates all of the earlier pieces, in order to actually generate compact certificates, complete with an end-to-end test that the resulting compact certs can be validated by a client.


tsachiherman:
maybe now it would be a good time to rebase this against master ?

zeldovich:
Thanks for merging in the txn PR!

I've merged master back into this PR's branch, so the diffs should be ready for review.

tsachiherman:
@zeldovich , I tried to rebase it and it looks like it would require some additional work:
running
```
go test -v . --run TestArchivalCreatables --race
```
still fails.
It looks like the vote tracker would need to be initialized before the call to initializeCaches, since it's calling newBlockImpl, that calls into the voter tracker.
```
	writingCatchpointRound, _, err = au.accountsq.readCatchpointStateUint64(context.Background(), catchpointStateWritingCatchpoint)
	if err != nil {
		return err
	}

	writingCatchpointDigest, err = au.initializeCaches(lastBalancesRound, lastestBlockRound, basics.Round(writingCatchpointRound))
	if err != nil {
		return err
	}

	if writingCatchpointRound != 0 && au.catchpointInterval != 0 {
		au.generateCatchpoint(basics.Round(writingCatchpointRound), au.lastCatchpointLabel, writingCatchpointDigest, time.Duration(0))
	}

	au.voters = &votersTracker{}
	err = au.voters.loadFromDisk(l, au)
	if err != nil {
		return err
	}
```

zeldovich:
Thanks for merging in the recent changes!  Indeed, it looks like the right thing to do is to avoid calling into the voters tracker when `au.voters` is nil.  Initializing the voters tracker requires `acctupdates` to be fully initialized, in order to compute top-N accounts at various recent rounds.  Pushed this small fix.

tsachiherman:
This PR is still failing. One of the fail cases was because the `votersTracker` wasn't shutdown when `accountsupdate` was. 
I implemented the shutdown code here:
https://github.com/zeldovich/go-algorand/compare/compactcert...tsachiherman:tsachi/closevotes
( if you could rebase you branch, it should be containing *just* the diff )

compactcert/builder.go
```diff
@@ -0,0 +1,332 @@
...
+	builder, ok := ccw.builders[sfa.Round]
+	if !ok {
+		latest := ccw.ledger.Latest()
+		latestHdr, err := ccw.ledger.BlockHdr(latest)
+		if err != nil {
+			return network.Disconnect, err
+		}
+
+		if sfa.Round <= latestHdr.CompactCertLastRound {
+			// Already have a complete compact cert in ledger.
```
algonautshant, line 148: Why are we sure we have this?

algonautshant (commented):
I have a few questions. Otherwise, looks good.

However, more test needs to be added (e.g. the case where overlap is true in signer() is not tested).

compactcert/builder.go
```diff
@@ -0,0 +1,332 @@
...
+	for rnd, sigs := range roundSigs {
+		_, ok := ccw.builders[rnd]
+		if ok {
+			ccw.log.Warnf("initBuilders: round %d already present", rnd)
+			continue
+		}
+
+		builder, err := ccw.builderForRound(rnd)
+		if err != nil {
+			ccw.log.Warnf("initBuilders: builderForRound(%d): %v", rnd, err)
```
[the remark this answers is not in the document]
zeldovich: This code is only getting called when we have a `roundSigs` for that round (i.e., some pending signature for a compact cert for that round), so if we can't find the voters tracker, that's a bit suspicious (and we have to throw away this pending signature).  I think you are right that it's possible for some race condition to cause this warning even though the situation is benign (e.g., we committed a new compact cert already, and as a result, there's no voters tracker, but we crashed just before we deleted the pending sig from `roundSigs`).  But the race is unlikely, so it seems still worth logging a warning if this happens.

compactcert/builder.go
zeldovich: `latestHdr.CompactCertLastRound` gets updated when that block contains a new compact cert transaction.  This means that blocks leading up to `latestHdr` must already store compact cert transactions for all blocks up to and including `latestHdr.CompactCertLastRound`.

node/node.go
```diff
@@ -271,6 +274,14 @@ func MakeFull(log logging.Logger, rootDir string, cfg config.Local, phonebookAdd
 		}
 	}
 
+	compactCertPathname := filepath.Join(genesisDir, config.CompactCertFilename)
+	compactCertAccess, err := db.MakeAccessor(compactCertPathname, false, false)
+	if err != nil {
+		log.Errorf("Cannot load compact cert data: %v", err)
```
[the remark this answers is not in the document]
zeldovich: Yes -- we do exactly the same thing for the agreement state database (see `crashPathname` and `crashAccess` above).

compactcert/signer.go
```diff
@@ -0,0 +1,174 @@
...
+		sigkeys, nextrnd, err = getSignedLast(tx)
+		return
+	})
+
+	if err != nil {
+		nextrnd := ccw.ledger.Latest() + 1
+		ccw.log.Warnf("ccw.signer(): using nextrnd=%d, cannot obtain last signed: %v", nextrnd, err)
+	} else {
+		// Check if we have any keys in common with the last signed set.
+		// If so, try to sign the next round after that.  Otherwise, if
```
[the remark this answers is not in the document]
zeldovich: I expect it should be `<= ccw.ledger.Latest()`

config/consensus.go
```diff
@@ -811,6 +811,13 @@ func initConsensusProtocols() {
 	// FilterTimeout for period 0 should take a new optimized, configured value, need to revisit this later
 	vFuture.AgreementFilterTimeoutPeriod0 = 4 * time.Second
 
+	// Enable compact certificates.
+	vFuture.CompactCertRounds = 128
+	vFuture.CompactCertTopVoters = 1024 * 1024
+	vFuture.CompactCertVotersLookback = 16
+	vFuture.CompactCertWeightThreshold = 30
+	vFuture.CompactCertSecKQ = 128
+
```
nicholasguoalgorand, line 820: Do we need to initialize these values in previous consensus versions?

test/e2e-go/features/compactcert/compactcert_test.go
```diff
@@ -0,0 +1,128 @@
...
+	r.NoError(err)
+
+	var lastCertBlock v1.Block
+	libgoal := fixture.LibGoalClient
+	for rnd := uint64(1); rnd <= consensusParams.CompactCertRounds*4; rnd++ {
+		fixture.WaitForRound(rnd, 30*time.Second)
+		blk, err := libgoal.Block(rnd)
+		r.NoError(err)
+
+		// fmt.Printf("Round %d, block %v\n", rnd, blk)
```
nicholasguoalgorand, line 66: remove this?

daemon/algod/api/spec/v1/model.go
```diff
@@ -524,6 +524,11 @@ type Transaction struct {
 	// required: true
 	ApplicationCall *ApplicationCallTransactionType `json:"app,omitempty"`
 
+	// CompactCert
+	//
+	// required: true
```
nicholasguoalgorand, line 529: why is required true?

nicholasguoalgorand (commented):
Looks good, had some pretty minor questions

config/consensus.go
zeldovich: Zero values are the correct initialization.

test/e2e-go/features/compactcert/compactcert_test.go
zeldovich: I changed it to `t.Logf()` instead, which captures what I actually wanted: it's a useful debug message if the test is failing, but doesn't need to be printed otherwise.

daemon/algod/api/spec/v1/model.go
zeldovich: I actually have no idea.  I don't know what's going on with the `required:` annotations there for swagger (I think).  If you can tell me what it should be instead, I'm happy to change it.  I just copied it from other entries that seemed semantically identical (like the `ApplicationCall` above it).

compactcert/signer.go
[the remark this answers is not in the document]
algonautshant: I am not able to follow this code.

Here is what I understand: please correct me.
nextrnd is determining where to start signing.
If getSignedLast returns an error, or if there is no overlap between last signed keys and the worker account keys, then nextrnd = ccw.ledger.Latest() + 1, otherwise, it will  start earlier. 

So, if there is an error (getSignedLast), then we start signing from a later round? Why is this?

brianolson (commented):
a couple notes, probably more Monday

algorandskiy (commented):
I'd like to see couple unit tests for `builder.go` and `signer.go` in compactcert package.
e2e test is good but unit test is unit test.

compactcert/builder.go
```diff
@@ -0,0 +1,332 @@
...
+		return
+	})
+	if err != nil {
+		ccw.log.Warnf("broadcastSigs: getPendingSigs: %v", err)
+		return
+	}
+
+	for rnd, sigs := range roundSigs {
+		for _, sig := range sigs {
+			// Randomize which sigs get broadcast over time.
```
derbear, line 258: How long does it take in expectation to get the whole cert in the common case?

compactcert/signer.go
```diff
@@ -0,0 +1,174 @@
...
+		// there is no overlap in keys, start with latest round.
```
derbear, line 57: What's the significance of having an overlap?

compactcert/signer.go
```diff
@@ -0,0 +1,174 @@
...
+	keys := ccw.accts.Keys()
+	if len(keys) == 0 {
+		// No keys, nothing to do.
+		return
+	}
+
+	// Compact cert gets signed by the next round after the block,
+	// because by the time agreement is reached on the block,
+	// ephemeral keys for that round could be deleted.
+	sigKeyRound := hdr.Round + 1
```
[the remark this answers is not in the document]
derbear: (Though this might be rare in practice with the key dilution)

compactcert/builder.go
```diff
@@ -0,0 +1,332 @@
...
+			addr64 := binary.LittleEndian.Uint64(sig.signer[:])
+			if addr64%(proto.CompactCertRounds/2) != uint64(brnd)%(proto.CompactCertRounds/2) {
```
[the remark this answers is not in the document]
brianolson: I think I get it, and @derbear 's comment. CompactCertRounds = 128
For round N..N+63 send 1/64 th of the 'pending sigs from this node', for N+64..N+127 send 1/64 th of the 'pending sigs' (from any node).

compactcert/signer.go
```diff
@@ -0,0 +1,174 @@
...
+				ccw.log.Warnf("ccw.signer(): BlockHdr(%d): %v", nextrnd, err)
+				switch err.(type) {
+				case ledger.ErrNoEntry:
+					nextrnd = ccw.ledger.Latest() + 1
+
+				default:
+					time.Sleep(1 * time.Second)
+				}
+			} else {
+				ccw.signBlock(hdr)
```
brianolson, line 91: I think this loop could cause a node in catchup to generate signature messages for blocks long in the past where the signature is long irrelevant. Maybe in the setup to `signer()` set a floor of nextrnd at (ledger.Latest() - CompactCertRounds) ?

compactcert/builder.go
```diff
@@ -0,0 +1,332 @@
...
+		})
+	})
+	if err != nil {
+		return network.Ignore, err
+	}
+
+	return network.Broadcast, nil
+}
+
+func (ccw *Worker) builder() {
```
algorandskiy, line 188: How do we handle catchup? Sounds like signatures are not needed during catchup on a blockchain that already has compact signatures

daemon/algod/api/server/v1/handlers/handlers.go
```diff
@@ -92,6 +92,8 @@ func txEncode(tx transactions.Transaction, ad transactions.ApplyData) (v1.Transa
 		res = assetFreezeTxEncode(tx, ad)
 	case protocol.ApplicationCallTx:
 		res = applicationCallTxEncode(tx, ad)
+	case protocol.CompactCertTx:
```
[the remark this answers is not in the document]
zeldovich: Mostly because, as far as I can tell, the e2e testing code uses the v1 API, so it was easiest to extend it that way.

Somewhere (I forget where, unfortunately) Rotem commented that the v2 API doesn't need to be explicitly extended to support this because it just sends back the entire msgpack-encoded transaction, so the v2 API client can get whatever information it wants directly.

compactcert/db.go
```diff
@@ -0,0 +1,159 @@
...
+		_, err := tx.Exec(tableCreate)
+		if err != nil {
+			return fmt.Errorf("could not create table %v", err)
+		}
+	}
+
+	return nil
+}
+
+func getSignedLast(tx *sql.Tx) ([]crypto.OneTimeSignatureVerifier, basics.Round, error) {
```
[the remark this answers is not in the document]
zeldovich: I agree we could store the msgpack serialization of this thing instead.

What's the `metastate` table you are referring to?  I'm probably missing something here.

compactcert/signer.go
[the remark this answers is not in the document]
zeldovich: Sorry for the confusion here!  I tried to add some description to the schema (in `db.go`) which might help make this less confusing.

Your understanding is correct.  The role of the stored `signed_last` information is to make sure that, when this node starts running, it will resume signing blocks where it last stopped.  In particular, if this node was stopped, and in the meantime many blocks appeared, on restart this node will sign those new blocks.

If there's an error, we don't know where we last left off signing, so we won't sign blocks in this missing window.  That doesn't seem too bad, since compact certificates can be formed even if not every node signs.  Since we don't know where we last left off, the only real other option is to block this node's startup altogether, which seems less desirable.

compactcert/signer.go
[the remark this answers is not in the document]
zeldovich: I think this will be logged by `OneTimeSignatureSecrets.Sign()` in `crypto/onetimesig.go`.

compactcert/builder.go
```diff
@@ -0,0 +1,332 @@
...
+		case <-ccw.ctx.Done():
+			return
+
+		case <-ccw.ledger.Wait(nextrnd):
+			// Continue on
+		}
+
+		hdr, err := ccw.ledger.BlockHdr(nextrnd)
+		if err != nil {
+			ccw.log.Warnf("ccw.builder: BlockHdr(%d): %v", nextrnd, err)
```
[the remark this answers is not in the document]
zeldovich: Good catch; thanks!

compactcert/builder.go
zeldovich: I would expect it to take slightly more than `CompactCertRounds/2` (i.e., 64) rounds.  The "slightly more" is to allow one of the relay nodes to broadcast the compact cert transaction, and then to allow that transaction to make it into a block, which probably takes another 1-2 rounds.

One degenerate case might be that we have fewer than 64 participation keys, in which case we might get 100% of the signatures earlier than 64 rounds, so then we'd be able to form the certificate at that point.

If some nodes are offline (and thus we have fewer than 100% sigs even after 64 rounds), the cert will be formed slightly later (somewhere between 64 and 128 rounds), depending on how much less than 100% we get.

compactcert/signer.go
zeldovich: Overlap means we know where we last left off signing blocks before this node stopped, and we have (some of) the same participation keys that the node had when it last stopped, so we can resume from there.

If the set of participation keys is completely different, or there weren't any participation keys before or now, don't bother trying to resume where we last left off -- just start signing from the current block.

compactcert/builder.go
```diff
@@ -0,0 +1,332 @@
...
+
+	err = ccw.db.Atomic(func(ctx context.Context, tx *sql.Tx) error {
+		return addPendingSig(tx, sfa.Round, pendingSig{
+			signer:       sfa.Signer,
+			sig:          sfa.Sig,
+			fromThisNode: sender == nil,
+		})
+	})
+	if err != nil {
+		return network.Ignore, err
```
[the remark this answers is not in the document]
zeldovich: I didn't have any concrete error in mind that would cause this execution path, but if something goes wrong, it seemed better to not propagate this message.  With the schema change that you proposed, this might indeed return an error due to a uniqueness constraint violation, although I would expect earlier code in this function (the check for `builder.Present(pos)`) to have the same effect.

compactcert/builder.go
zeldovich: There's two possible directions in terms of how catchup and compact certs could interact.

One direction is: compact certs could be used to significantly speed up catchup.  This is not done in this PR, but hopefully should be not too hard to implement.  The block validation logic can be borrowed from the compactcert e2e test, and used during catchup to quickly validate a chain of blocks at multiples of CompactCertRounds.  Then use the prev pointer (`BlockHdr.Branch`) to validate previous block hashes.

The other direction is: during catchup, the compact cert code will try to sign blocks for compact certs, and build compact certs.  If compact certs have already been formed for those blocks (most likely), this will be harmless.  But the code tries to be careful and make sure that, even if a node was not running at some point, restarting it will cause it to produce all the same compact cert signatures, in case the network was for some reason waiting for our signature.

compactcert/builder.go
```diff
@@ -0,0 +1,332 @@
...
+	return res, nil
+}
+
+func (ccw *Worker) initBuilders() {
+	ccw.mu.Lock()
+	defer ccw.mu.Unlock()
+
+	var roundSigs map[basics.Round][]pendingSig
+	err := ccw.db.Atomic(func(ctx context.Context, tx *sql.Tx) (err error) {
+		roundSigs, err = getPendingSigs(tx)
```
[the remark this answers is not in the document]
zeldovich: Relay nodes build compact certificates using signatures from all participating accounts, regardless of which node that signature came from.  (In the likely deployment scenario, a relay node does not have any participation keys of its own, and thus all signatures come from other nodes.)

zeldovich:
Thank you for all of the review feedback (and for your patience while I was distracted with other things over the last month or two).

A big source of confusion was the signer code, and its associated tracking of `signed_last`.  I re-worked that whole logic (see commit 57ffe9b2a that was recently pushed) to hopefully avoid much of the confusion, avoid tracking `signed_last` altogether, avoid some of the potential long looping, etc.

node/node.go
```diff
@@ -819,6 +819,53 @@ func (node *AlgorandFullNode) oldKeyDeletionThread() {
...
+		// we already computed, since we can then delete ephemeral keys that
+		// were already used to compute a signature (stored in the compact
+		// cert db).
+		ccSigs, err := node.compactCert.LatestSigsFromThisNode()
+		if err != nil {
+			node.log.Warnf("Cannot look up latest compact cert sigs: %v", err)
+			continue
+		}
+
+		nextRoundFunc := func(part account.Participation) basics.Round {
```
tsachiherman, line 846: passing the nextRoundFunc to the DeleteOldKeys is technically correct, but makes it much harder to follow.
Could you have that implementation as part of the DeleteOldKeys or in  `accountManager.go` ?

zeldovich: Yeah, it's not the cleanest code..  The reason I tried doing it this way (with a callback func) is to avoid having the lower layers depend on knowing about compact certificates.  In particular, currently the compactcert package imports the data package, and we would need to import the compactcert package from data if we wanted to invoke `node.compactCert.LatestSigsFromThisNode()` there.

Just for comparison, I pushed another version of this code: `node.go` calls `LatestSigsFromThisNode()`, but then the callback is inlined inside of `accountManager.go`.  If this looks cleaner to you, I'm happy with it too.

data/basics/units.go
```diff
@@ -130,3 +130,8 @@ func (round Round) SubSaturate(x Round) Round {
 
 	return round - x
 }
+
+// RoundUpToMultipleOf rounds up round to the next multiple of n.
+func (round Round) RoundUpToMultipleOf(n Round) Round {
```
tsachiherman, line 135: I'm not sure if `n` needs to be a `basics.Round` here. But I can't see any concrete harm done here.

algorandskiy (changes requested):
Nickolai thank you for answering questions. The code looks good but I there are no tests in `compactcert` package. This creates maintainability problem and needs to be addressed.

tsachiherman (approved):

Issue #914, opened 2020-03-17
Agreement service votes validation conflicts with ledger balance history


tsachiherman:
## Issue
The issue that we've observed was logged as follows:
```
malformed proposal for (5487228, 0): unauthenticatedVote.verify: could not get membership parameters: Service.initializeVote (r=5487228): Failed to obtain balance record for address IVKDCE6MS44YVGMQQFVXCDABW2HKULKIXMLDS2AEOIA6P2OGMVHVJ64MZI in round 5486908: round 5486908 before dbRound 5486910
```

The above has few variation ranging between proposal and vote verification.

## Analysis
The verification code is relying on the `membership` function, which is turn is using the `balanceRound` to calculate the round at which we obtain the balance for the account in question. The round number is calculated as : 
balanceRound = vote/proposal round - 2 * SeedRefreshInterval * SeedLookback =  vote/proposal round - 2 * 80 * 2 = vote/proposal round - 320

On the other hand, the ledger account manager is holding back the balances for the past MaxBalLookback round ( = 320 ).

If the agreement service was the only party that is moving the ledger forward, this would not be an issue. However, since the catchup is moving the ledger as well, it does get to scenarios where such as : 
1. ledger round is 1000 ( earliest balance round is 680 ).
2. the agreement service received a message to be verified for round 1000. That went onto a background queue.
3. the catchup service moved the ledger to round 1001 ( earliest balance round is 681 ).
4. the verifier is trying to verify the message and needs the balance for round 1000-320=680. however, this balance is no longer available.

derbear:
This should be safe--if the catchup service has already moved on to the next round, then it's fine to try to send votes from old rounds.

Perhaps it would make sense to drop the logging level to `info` in such a case? Is this something worth doing?

tsachiherman:
Yes, this sounds correct. If we could confirm that this is the case, then we should lower the logging level as well as log a different message. ( i.e. message that explain why this membership isn't really required )

derbear:
Okay. Something to keep in mind is that this is only fine in this case--if the failure is due to something else, then something has definitely gone wrong.

I think we'll need to have the `LedgerReader` methods return an error of a specific type and then switch on it at the top.

derbear:
Or, something easier might be to check the current round and suppress the warning if we've moved past it.

tsachiherman:
Yes; also, we need to make sure not to disconnect from the source of the message in that case.

PR #1508, opened 2020-09-15
Vote validation error refactor

nicholasguoalgorand:
## Summary

When an extremely old vote is received, the verifier will not be able to fetch the membership and will mark it as ProposalMalformed and cause a WARN log. This change downgrades that to a DEBUG and marks it as ProposalFiltered by using the cancelled tag.

Resolves https://github.com/algorand/go-algorand/issues/914

## Test Plan

Wrote additional unit test that simulates the described scenario to test this change

### DIFF
--- a/agreement/asyncVoteVerifier.go
+++ b/agreement/asyncVoteVerifier.go
@@ -19,10 +19,8 @@ package agreement
 import (
 	"context"
 	"errors"
-	"fmt"
 	"sync"
 
-	"github.com/algorand/go-algorand/data/basics"
 	"github.com/algorand/go-algorand/util/execpool"
 )
 
@@ -61,14 +59,13 @@ type AsyncVoteVerifier struct {
 	ctxCancel       context.CancelFunc
 }
 
-// RoundOffsetError is an error for when agreement far behind ledger
-type RoundOffsetError struct {
-	Round   basics.Round
-	DbRound basics.Round
+// LedgerDroppedRoundError is an error for when agreement far behind ledger
+type LedgerDroppedRoundError struct {
+	Err error
 }
 
-func (e *RoundOffsetError) Error() string {
-	return fmt.Sprintf("round %d before dbRound %d", e.Round, e.DbRound)
+func (e *LedgerDroppedRoundError) Error() string {
+	return e.Err.Error()
 }
 
 // MakeAsyncVoteVerifier creates an AsyncVoteVerifier with workers as the number of CPUs
@@ -118,7 +115,7 @@ func (avv *AsyncVoteVerifier) executeVoteVerification(task interface{}) interfac
 		v, err := req.uv.verify(req.l)
 		req.message.Vote = v
 
-		var e *RoundOffsetError
+		var e *LedgerDroppedRoundError
 		cancelled := errors.As(err, &e)
 
 		return &asyncVerifyVoteResponse{v: v, index: req.index, message: req.message, err: err, cancelled: cancelled, req: &req}
@@ -136,7 +133,7 @@ func (avv *AsyncVoteVerifier) executeEqVoteVerification(task interface{}) interf
 		// request was not cancelled, so we verify it here and return the result on the channel
 		ev, err := req.uev.verify(req.l)
 
-		var e *RoundOffsetError
+		var e *LedgerDroppedRoundError
 		cancelled := errors.As(err, &e)
 
 		return &asyncVerifyVoteResponse{ev: ev, index: req.index, message: req.message, err: err, cancelled: cancelled, req: &req}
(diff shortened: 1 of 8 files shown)


agreement/asyncVoteVerifier.go
```diff
@@ -104,7 +117,14 @@ func (avv *AsyncVoteVerifier) executeVoteVerification(task interface{}) interfac
 		// request was not cancelled, so we verify it here and return the result on the channel
 		v, err := req.uv.verify(req.l)
 		req.message.Vote = v
-		return &asyncVerifyVoteResponse{v: v, index: req.index, message: req.message, err: err, req: &req}
+
+		cancelled := false
+		var e *RoundOffsetError
+		if ok := errors.As(err, &e); ok {
+			cancelled = true
+		}
```
tsachiherman, line 125: I think that you could write this as
```
var e *RoundOffsetError
canceled := errors.Is(err, &e)
```

agreement/asyncVoteVerifier.go
```diff
@@ -58,6 +61,16 @@ type AsyncVoteVerifier struct {
 	ctxCancel       context.CancelFunc
 }
 
+// RoundOffsetError is an error for when agreement far behind ledger
+type RoundOffsetError struct {
```
tsachiherman, line 65: Can we have this error be defined in the ledger rather than in the agreement ?
We might need to have both, and "bridge" them in the node to avoid additional dependencies.
(alternative - move it to a shared location such as data)

nicholasguoalgorand: So when I do this there is a circular import dependency. I'm not sure what the best solution would be here.

agreement/voteAggregator_test.go
```diff
@@ -839,3 +839,38 @@ func TestVoteAggregatorFiltersVoteNextRound(t *testing.T) {
 	require.NoError(t, err)
 	require.NoErrorf(t, res, "Votes from next round not correctly filtered")
 }
+
+func TestVoteAggregatorOldVote(t *testing.T) {
+	ledger := makeTestLedgerMaxBlocks(readOnlyGenesis100, 320)
```
tsachiherman, line 844: take a look on 
```
func balanceRound(r basics.Round, cparams config.ConsensusParams) basics.Round {
	return r.SubSaturate(basics.Round(2 * cparams.SeedRefreshInterval * cparams.SeedLookback))
}
```
this is how the 320 number is generated, and we should ensure that this is what we test against.
i.e. replace the 320 with the above calculation. you can use the current consensus protocol params to get the numerical values.

agreement/asyncVoteVerifier.go
```diff
@@ -58,6 +59,15 @@ type AsyncVoteVerifier struct {
...
 
+// LedgerDroppedRoundError is an error for when agreement far behind ledger
+type LedgerDroppedRoundError struct {
+	Err error
+}
+
+func (e *LedgerDroppedRoundError) Error() string {
+	return e.Err.Error()
+}
+
```
tsachiherman, line 70: This would be  better placed in abstract.go, where the interfaces are. We should also add some more detailed documentation for the reasons this error is used ( notify the agreement that the requested round cannot, and will not be available as it's too far back ), and why ( the agreement need to make the distinction between a malicious vote and a vote that it cannot verify ).

ledger/acctupdates.go
```diff
@@ -213,6 +213,16 @@ type deferedCommit struct {
 	lookback basics.Round
 }
 
+// RoundOffsetError is an error for when requested round is behind earliest stored db entry
+type RoundOffsetError struct {
+	Round   basics.Round
+	DbRound basics.Round
```
tsachiherman, line 219: rename Round -> requestedRound and DbRound -> dbRound.
we don't need to export the fields here, only the error struct.

agreement/asyncVoteVerifier.go
tsachiherman: Also, since you've created such a nice internal error, why don't you export it via Unwrap ? it will avoid breaking the chain.

node/impls.go
```diff
@@ -111,3 +114,14 @@ func (l agreementLedger) EnsureDigest(cert agreement.Certificate, verifier *agre
 	// 4. no other senders to this channel exists
 	l.UnmatchedPendingCertificates <- catchup.PendingUnmatchedCertificate{Cert: cert, VoteVerifier: verifier}
 }
+
+func (l agreementLedger) Lookup(rnd basics.Round, addr basics.Address) (basics.AccountData, error) {
```
tsachiherman, line 118: Add some documentation here explaining why you're doing that.

tsachiherman:
I think that this one looks good. You'll need to convert this one from Draft to a PR by hitting the "Ready for review".

tsachiherman (approved):
looks good.

PR #1510, opened 2020-09-15
Add metrics for the various ledger components

brianolson:
## Summary

Measure count of various database options and the sum of milliseconds spent on those operations. This profiling should let us know what's slow and if anything needs to be optimized or if db ops are fast enough.

## Test Plan

Run this on a node in catchup, a node doing private network tests, a node caught up with mainnet/testnet and processing current traffic, and nodes in an S1 stress test.


ledger/catchupaccessor.go
```diff
@@ -174,6 +174,7 @@ func (c *CatchpointCatchupAccessorImpl) SetLabel(ctx context.Context, label stri
 // ResetStagingBalances resets the current staging balances, preparing for a new set of balances to be added
 func (c *CatchpointCatchupAccessorImpl) ResetStagingBalances(ctx context.Context, newCatchup bool) (err error) {
 	wdb := c.ledger.trackerDB().wdb
+	// TODO: db metrics
```
tsachiherman, line 177: The `CatchpointCatchupAccessorImpl` is used only during fast catchup, so we won't get any metrics on normal operations.

tsachiherman (commented):
I like the blockqueue metrics, but think that the accountupdate ones would need to have benchmarks per exported function first, and going down to a concrete metric later.

brianolson:
Now after running a mainnet catchpoint and a few rounds, shows metrics like this
```
ledger_accountsinit_count{} 2
ledger_accountsinit_ms{} 26
ledger_commitround_count{} 99
ledger_commitround_ms{} 3491
ledger_vacuum_count{} 2
ledger_vacuum_ms{} 19417
ledger_init_count{} 2
ledger_init_ms{} 0
ledger_sync_blockput_count{} 374
ledger_sync_blockput_ms{} 4189
ledger_sync_blockforget_count{} 374
ledger_sync_blockforget_ms{} 3603
ledger_getblock_count{} 1355
ledger_getblock_ms{} 2
ledger_getblockhdr_count{} 4
ledger_getblockhdr_ms{} 0
ledger_resetstagingbalances_count{} 1
ledger_resetstagingbalances_ms{} 9
ledger_processstagingcontent_count{} 1
ledger_processstagingcontent_ms{} 6
ledger_processstagingbalances_count{} 11114
ledger_processstagingbalances_ms{} 605303
ledger_verifycatchpoint_count{} 1
ledger_verifycatchpoint_ms{} 1
ledger_storebalancesround_count{} 1
ledger_storebalancesround_ms{} 82
ledger_storefirstblock_count{} 1
ledger_storefirstblock_ms{} 10
ledger_catchpointstoreblock_count{} 1000
ledger_catchpointstoreblock_ms{} 12011
ledger_catchpoint_finishblocks_count{} 1
ledger_catchpoint_finishblocks_ms{} 23
ledger_catchpoint_ensureblock1_count{} 1
ledger_catchpoint_ensureblock1_ms{} 0
ledger_catchpoint_finish_bals_count{} 1
ledger_catchpoint_finish_bals_ms{} 16
ledger_initblocksdb_count{} 1
ledger_initblocksdb_ms{} 23
ledger_verifygenhash_count{} 2
ledger_verifygenhash_ms{} 0
algod_network_incoming_connections{} 0
algod_network_outgoing_connections{} 4
algod_network_rx_buffer_micros_total{} 2979291
algod_network_rx_handle_micros_total{} 174771
algod_network_broadcasts_total{} 2323
algod_network_broadcast_queue_micros_total{} 14210
algod_network_broadcast_send_micros_total{} 11128
algod_network_peers{} 4
algod_network_incoming_peers{} 0
algod_network_outgoing_peers{} 4
algod_network_sent_bytes_total{} 236436
algod_network_received_bytes_total{} 116356440
algod_network_message_received_total{} 279289
algod_network_message_sent_total{} 6954
algod_network_message_sent_queue_micros_total{} 256538
algod_transaction_messages_dropped_backlog{} 516
algod_agreement_handled{} 4788
algod_tx_pool_count{} 0
algod_ledger_transactions_total{} 0
algod_ledger_reward_claims_total{} 100
algod_ledger_transactions_total{} 14966
algod_ledger_reward_claims_total{} 786
algod_ledger_round{} 9080786
```

tsachiherman:
just be reminded that `ledger_commitround_ms{} 3491` is a counter, so you would need to subtract the previous value from it before interpreting it's value. ( or maybe a gauge is more suitable for this one ? )

brianolson:
Right, `ledger_commitround_ms{} 3491` is for all time, so differential readings would be needed if you wanted to see performance during different phases of a test (Grafana has builtin tools for this). In this case I took the metrics just after algod had spent most of its time since startup doing a catchpoint fast-catchup. So you can see `ledger_processstagingbalances_ms{} 605303` 605 seconds of actual IO time happened doing that part of it across `ledger_processstagingbalances_count{} 11114` calls. (This was on my MBP)

ledger/acctupdates.go
```diff
@@ -1895,7 +1910,10 @@ func (au *accountUpdates) vacuumDatabase(ctx context.Context) (err error) {
 		}
 	}()
 
+	start := time.Now()
+	ledger_vacuum_count.Inc(nil)
 	vacuumStats, err := au.dbs.wdb.Vacuum(ctx)
+	counterMicros(ledger_vacuum_micros, start)
```
tsachiherman, line 1916: In this particular case, I think that we should use the existing `vacuumElapsedTime` and add it to the `ledger_vacuum_micros` counter. I'd rather avoid having several "variations" for the same operation.

util/metrics/counter.go
```diff
@@ -36,6 +36,18 @@ func MakeCounter(metric MetricName) *Counter {
 	return c
 }
 
+func NewCounter(name, desc string) *Counter {
```
tsachiherman, line 39: It would be cleaner if you were to pack the name&desc in a `MetricName` and pass it to `MakeCounter`

ledger/blockqueue.go
```diff
@@ -306,11 +326,34 @@ func (bq *blockQueue) getBlockCert(r basics.Round) (blk bookkeeping.Block, cert
...
 	})
+	counterMicros(ledger_getblockcert_micros, start)
 	err = updateErrNoEntry(err, lastCommitted, latest)
 	return
 }
+
+// TODO: promote this to a utility function in util/metrics ?
+func counterMicros(counter *metrics.Counter, start time.Time) {
+	counter.AddUint64(uint64(time.Now().Sub(start).Microseconds()), nil)
+}
```
tsachiherman, line 344: Yes, add a method to counter that would do so. maybe `AddDurationSinceMicroseconds(..)` ?

brianolson:
@tsachiherman there's a difference in conventions I'd like you to weigh in on.
metrics are typically `ledger_accountsonlinetop_count`
and go variables would be `ledgerAccountsonlinetopCount`
I think it would be convenient for them to be same so that grepping the code turns up both the string and the variable.
go lint is complaining that I made the go variables look like metrics with underscores. I don't _think_ there's anything downstream in grafana that cares about the metrics variable names, so I'm inclined to break that convention and make new metrics names that look like go variables. Thoughts or a vote one way or the other?

tsachiherman:
> @tsachiherman there's a difference in conventions I'd like you to weigh in on.
> metrics are typically `ledger_accountsonlinetop_count`
> and go variables would be `ledgerAccountsonlinetopCount`
> I think it would be convenient for them to be same so that grepping the code turns up both the string and the variable.
> go lint is complaining that I made the go variables look like metrics with underscores. I don't _think_ there's anything downstream in grafana that cares about the metrics variable names, so I'm inclined to break that convention and make new metrics names that look like go variables. Thoughts or a vote one way or the other?

I think that I would rather seeing using go-naming conventions for go code, and have the metrics have their existing naming convention style. 
Note that entries like this 
```
var ledger_init_count = metrics.NewCounter("ledger_init_count", "calls to init block queue")
```
would probably be best if modified slightly - 
The counter `ledger_init_count` representing the block queue initialization, so it should be "ledger_blockqueue_init_count".
and so forth.

I would extend that to the corresponding counters as well. `ledger_init_count` is already within the ledger package, and so it doesn't add much to call it `ledger_XXX`; instead, calling it `blockQueueInitCount` makes much more sense.

tsachiherman:
How about using "μs" instead of "ms" to ensure nobody mix up millisecond with microseconds ?

ledger/ledger.go
```diff
@@ -643,3 +651,8 @@ type txlease struct {
 	sender basics.Address
 	lease  [32]byte
 }
+
+var ledgerInitblocksdbCount = metrics.NewCounter("ledgerInitblocksdbCount", "calls")
+var ledgerInitblocksdbMicros = metrics.NewCounter("ledgerInitblocksdbMicros", "µs spent")
+var ledgerVerifygenhashCount = metrics.NewCounter("ledgerVerifygenhashCount", "calls")
+var ledgerVerifygenhashMicros = metrics.NewCounter("ledgerVerifygenhashMicros", "µs spent")
```
tsachiherman, line 658: align counters naming conventions with the rest ?

tsachiherman (commented):
looks good; just one comment above.

tsachiherman (approved):

PR #1512, opened 2020-09-15
Improve network template validation

tsachiherman:
## Summary

The network template validator was failing accounting for small fractions. This is a meaningless situation that could safely ignored here. When the actual account would be allocation, the allocator would ensure the reminder algos would be assigned to the last account.

## Test Plan

Unit test updated for this use case.


algorandskiy (approved):
Looks good as a quick fix. Maybe we want to switch to `decimal` in `WalletData` to complete eliminate the problem.

tsachiherman:
> decimal

I'm ok if you'd like to pursue that; `decimal` is not a public struct, so I don't know how easy it would be to use it directly.

PR #1513, opened 2020-09-15
Improve goal node start error reporting in case of algod errored

tsachiherman:
## Summary

Improve the error reporting in case we're unable to start algod : 
if there is an error code, it would be returned as part of the error.
The newly reported error aligns with the go 1.13 errors objects and allow the client to inspect its content using Unwrap().

## Test Plan

Added an expect test to verify it's working.


test/e2e-go/cli/goal/expect/goalNodeTest.exp
```diff
@@ -36,6 +36,24 @@ if { [catch {
...
+    lassign [exec find $TEST_PRIMARY_NODE_DIR -name "ledger.tracker.sqlite"] PRIMARY_TRACKER_DATABASE_FILE
+    exec find $TEST_PRIMARY_NODE_DIR -name "ledger.tracker.sqlite*" -delete
+    exec -- echo "1234" > $PRIMARY_TRACKER_DATABASE_FILE
+
+    # try to start the primary node, and observe the expected failuire
+    set timeout 15
+    spawn goal node start -d $TEST_PRIMARY_NODE_DIR
+    expect {
+        timeout { close; ::AlgorandGoal::Abort "starting node exceeded timeout" }
+        "^Algorand node failed to start: node exited with an error code, check node\.log for more details : exit status 1" {puts "Expected failuire : node failed to start"; close}
```
algonautshant, line 49: should add the eof check line: 
`	eof { ::AlgorandGoal::Abort "eof received before the expected output " }`
Otherwise, the eof will match when the `"^Algorand node failed to start...` is not matched, and proceed without reporting an error.

algonautshant, line 49: `\.` is not necessary. `.` will be sufficient.

algonautshant (commented):
Looks good. 
The test must be updated otherwise the error will not be caught.

algonautshant (approved):

test/e2e-go/cli/goal/expect/goalNodeTest.exp
tsachiherman: added.

tsachiherman: I've converted this to be a regex.

algonautshant (approved):

PR #1515, opened 2020-09-16
Remove devtools binaries from algorand package (rpm)

btoll:
## Summary

Remove devtools binaries from algorand rpm package

## Test Plan

Jenkins job.


tsachiherman (approved):
looks like the correct set of binaries for the algorand package to me.

onetechnical (approved):

PR #1518, opened 2020-09-17
Cache the most recent Seed/Circulation calls to reduce pressure on ledger

tsachiherman:
## Summary

This PR adds a cache for the data.Ledger for the `Circulation` and `Seed` calls.
The goal of this PR is to reduce the contingency pressure on the downstream accounts update lock.

The high-level plan is to eventually eliminate this code, and make sure that the agreement service is sufficiently smart to perform a similar logic.

## Test Plan

Unit test added.

### DIFF
--- a/data/ledger.go
+++ b/data/ledger.go
@@ -45,22 +45,34 @@ type Ledger struct {
 
 	log logging.Logger
 
+	// a two-item moving window cache for the total number of online circulating coins
 	lastRoundCirculation atomic.Value
-	lastRoundSeed        atomic.Value
+	// a two-item moving window cache for the round seed
+	lastRoundSeed atomic.Value
 }
 
+// roundCirculation is the cache for the circulating coins
 type roundCirculation struct {
-	prevRound       basics.Round
+	// prevRound is the round number for the previously fetched online circulating coins
+	prevRound basics.Round
+	// prevOnlineMoney is the amount of coins in circulation for the round prevRound
 	prevOnlineMoney basics.MicroAlgos
-	curRound        basics.Round
-	curOnlineMoney  basics.MicroAlgos
+	// curRound is the round number for the last fetched online circulating coins
+	curRound basics.Round
+	// curOnlineMoney is the amount of coins in circulation for the round curRound
+	curOnlineMoney basics.MicroAlgos
 }
 
+// roundSeed is the cache for the seed
 type roundSeed struct {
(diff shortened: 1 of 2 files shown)


data/ledger.go
```diff
@@ -43,6 +44,23 @@ type Ledger struct {
...
+
+type roundCirculation struct {
+	prevRound       basics.Round
+	prevOnlineMoney basics.MicroAlgos
+	curRound        basics.Round
+	curOnlineMoney  basics.MicroAlgos
+}
+
+type roundSeed struct {
+	prevRound basics.Round
```
nicholasguoalgorand, line 60: prevRound indicates the previously stored round, not the most previous round right? Might be helpful to include a comment clarifying this / rename the field

nicholasguoalgorand:
Do we know if this makes a performance improvement or are we just guessing that it does? Are there any specific figures that would be helpful to include in the pr?

data/ledger.go
tsachiherman: I'm not sure what `most previous round` would be. It's a two element moving window cache.
The implementation here is the trivial one, allowing to reduce the pressure from the accounts update.
My intent is eventually to wrap this into the agreement itself, and delete this code completely.
( when the agreement is making ledger advancements, it should query these values and store these locally.
we should go "outside" the agreement "box" only when needed )

tsachiherman:
> Do we know if this makes a performance improvement or are we just guessing that it does? Are there any specific figures that would be helpful to include in the pr?

I have measured that it reduces the timing from 800-1000ns to 350-450ns. However, I'm less interested in the absolute gains here, and more concern about lowering the pressure on the ledger code itself. See my previous nodes about moving further.

data/ledger.go
nicholasguoalgorand: right so `prevRound` could be confused with `curRound - 1`. But this is not the case here, so I suggested a comment possibly clarifying.

tsachiherman: I've added some documentation which I hope would be clear enough.

tsachiherman: I'm not a big fan of a/b; how do you feel about an array of two elements ?

brianolson: 2 element array sounds great

tsachiherman: ok, done.

brianolson (approved):

Issue #1505, opened 2020-09-14
FilterTimeout changes followup


tsachiherman:
1. rename FilterTimeoutSmallLambdas and FilterTimeoutPeriod0SmallLambdas in consensus.go to AgreementFilterTimeout and AgreementFilterTimeoutPeriod0
2. AgreementFilterTimeout/AgreementFilterTimeoutPeriod0 are already defined in units of uint64. That's great. however, to take advantage of these, we need to change the scale. I'd suggest to change the scale to nanoseconds to align with time.Duration. Given that we're defining the filter timeout here and not the small lambdas anymore, we can just assign concrete values here. i.e. 4000000000
3. for now, we should avoid the following:
```
       // FilterTimeout 2s instead of 4s for period 0
	vFuture.FilterTimeoutPeriod0SmallLambdas = 1
```
instead, you can hard code it here to have the 4000000000 value above, so that we will be reminded that we need to update this value.
4. In func FilterTimeout, you can now just use the value from the protocol params without any multiplication. Add a comment explaining what this means. i.e. that the return value is 2 times the small lambda, and the protocol changes should account for that.
5. Find where the SmallLambda from config.Global, is being used, and update all the locations that are using it as a "filter timeout" to use the filter timeout from the protocol params. If not sure, you can always use the current protocol params to derive the value.

PR #1520, opened 2020-09-17
Filter Timeout Change Refactor

nicholasguoalgorand:
## Summary

https://github.com/algorand/go-algorand/issues/1505
Change new filtertimeout variables to AgreementFilterTimeout and AgreementFilterTimeoutPeriod0 with type time.Duration

## Test Plan

Reran unit tests


agreement/types.go
```diff
@@ -31,13 +31,10 @@ var recoveryExtraTimeout = config.Protocol.SmallLambda
...
 func FilterTimeout(p period, v protocol.ConsensusVersion) time.Duration {
-	var smallLambdas uint64
 	if p == 0 {
-		smallLambdas = config.Consensus[v].FilterTimeoutPeriod0SmallLambdas
-	} else {
-		smallLambdas = config.Consensus[v].FilterTimeoutSmallLambdas
+		return config.Consensus[v].AgreementFilterTimeoutPeriod0
 	}
-	return config.Protocol.SmallLambda * time.Duration(smallLambdas)
+	return config.Consensus[v].AgreementFilterTimeout
```
tsachiherman, line 37: Add a comment here stating that the timeout is expected to be 2 * SmallLambda, and it was moved into the consensus protocol parameter.

config/consensus.go
```diff
@@ -137,8 +137,8 @@ type ConsensusParams struct {
 	DownCommitteeSize      uint64
 	DownCommitteeThreshold uint64
 
-	FilterTimeoutSmallLambdas             uint64
-	FilterTimeoutPeriod0SmallLambdas      uint64
+	AgreementFilterTimeout        time.Duration
+	AgreementFilterTimeoutPeriod0 time.Duration
```
tsachiherman, line 141: Add explanations around these variables - what do they mean, how they should be configured, etc.

config/consensus.go
```diff
@@ -792,7 +792,7 @@ func initConsensusProtocols() {
 	vFuture.ApprovedUpgrades = map[protocol.ConsensusVersion]uint64{}
 
 	// FilterTimeout 2s instead of 4s for period 0
-	vFuture.FilterTimeoutPeriod0SmallLambdas = 1
+	vFuture.AgreementFilterTimeoutPeriod0 = 2 * time.Second
```
tsachiherman, line 795: for now, set it to `4 * time.Seconds` and update the comment that we should revisit this value.

tsachiherman (commented):
Few notes below; In addition, take a look on `node_test.go` : 
the following line should be updated:

var expectedAgreementTime = 2*config.Protocol.BigLambda + 3*config.Protocol.SmallLambda + 2*time.Second

The  3*config.Protocol.SmallLambda should be converted to `config.Protocol.SmallLambda + cparams.agreementfiltertimeout`

nicholasguoalgorand:
> Few notes below; In addition, take a look on
> 
> 1. `node_test.go` :
>    the following line should be updated:
> 
> var expectedAgreementTime = 2_config.Protocol.BigLambda + 3_config.Protocol.SmallLambda + 2*time.Second
> 
> The 3*config.Protocol.SmallLambda should be converted to `config.Protocol.SmallLambda + cparams.agreementfiltertimeout`
> 
> 1. `basic_test.go`:
>    this line `time.Sleep(2 * config.Protocol.SmallLambda)` need to be updated to be `cparams.agreementfiltertimeout`
> 
> once you're done with this change, you'll have some more work refactoring the `ALGOSMALLLAMBDAMSEC` into the `agreementfiltertimeout` that you just created.

What consensus version should `cparams` be, current or future?

Next for the `ALGOSMALLLAMBDAMSEC` refactor, is it okay to just create env variables for `ALGOFILTERTIMEOUT`?

tsachiherman:
> > Few notes below; In addition, take a look on
> > 
> > 1. `node_test.go` :
> >    the following line should be updated:
> > 
> > var expectedAgreementTime = 2_config.Protocol.BigLambda + 3_config.Protocol.SmallLambda + 2*time.Second
> > The 3*config.Protocol.SmallLambda should be converted to `config.Protocol.SmallLambda + cparams.agreementfiltertimeout`
> > 
> > 1. `basic_test.go`:
> >    this line `time.Sleep(2 * config.Protocol.SmallLambda)` need to be updated to be `cparams.agreementfiltertimeout`
> > 
> > once you're done with this change, you'll have some more work refactoring the `ALGOSMALLLAMBDAMSEC` into the `agreementfiltertimeout` that you just created.
> 
> What consensus version should `cparams` be, current or future?
> 
> Next for the `ALGOSMALLLAMBDAMSEC` refactor, is it okay to just create env variables for `ALGOFILTERTIMEOUT`?

Definitely use params with current version. This test is not intended to test the variations of the filtertimeout, so using the current protocol version is the correct thing to do.

As for the ALGOSMALLLAMBDAMSEC refactor, no - it should be a configurable consensus where you programmatically set the FilterTimeout. The fact that we had that ALGOSMALLLAMBDAMSEC variable was a temporary placeholder. ( that small refactor should be done outside the scope of this PR )

test/e2e-go/features/transactions/accountv2_test.go
```diff
@@ -39,6 +40,8 @@ func TestAccountInformationV2(t *testing.T) {
 	proto, ok := config.Consensus[protocol.ConsensusFuture]
 	a.True(ok)
 	os.Setenv("ALGOSMALLLAMBDAMSEC", "200")
+	proto.AgreementFilterTimeoutPeriod0 = 400 * time.Millisecond
+	proto.AgreementFilterTimeout = 400 * time.Millisecond
```
tsachiherman, line 44: if you've set the proto.AgreementFilterTimeoutPeriod0/proto.AgreementFilterTimeout, than you should be able to remove the `os.Setenv("ALGOSMALLLAMBDAMSEC", "200")` and ```defer func() {
 		os.Unsetenv("ALGOSMALLLAMBDAMSEC")
	}()```, right ?

test/e2e-go/upgrades/send_receive_upgrade_test.go
```diff
@@ -117,6 +117,10 @@ func generateFastUpgradeConsensus() (fastUpgradeProtocols config.ConsensusProtoc
 		}
 
 		fastUpgradeProtocols[consensusTestFastUpgrade(proto)] = fastParams
+
+		// support the ALGOSMALLLAMBDAMSEC = 500 env variable
+		fastParams.AgreementFilterTimeout = time.Second
```
tsachiherman, line 122: as before - if you've set this, you should be able to get rid of the `os.Setenv("ALGOSMALLLAMBDAMSEC", "500")`.
make sure that the various tests still takes have the same execution time.

tsachiherman (approved):
@nicholasguoalgorand , the changes here are good, but its not complete yet.
I'm going to merge this in just to unblock the performance testing, but the rest of the `ALGOSMALLLAMBDAMSEC` still need to be updated.

nicholasguoalgorand:
I wasn't sure if the env variables could be removed since small lambda is still used in other variables ie deadlineTimeout, etc

tsachiherman:
True, but these tests are only adjusting the small lambda to run faster,
and it would suffice to update the filtertimeout in these cases.
( we don't have any tests that arbitrarily check various small lambda
values )

nicholasguoalgorand:
Cool, then I can make another pr that removes those? We might be able to completely deprecate the env variable?

tsachiherman:
Yes, that was my intent.

PR #1525, opened 2020-09-18
Fix gpg agent forwarding

btoll:
## Summary

Revert to using `egrep` instead of `grep -E`.

## Test Plan

./forward_gpg_agent


btoll:
Close as duplicate

PR #1527, opened 2020-09-18
Decouple database round and acctupdate round

tsachiherman:
## Summary

We used to have a rwmutex for synchronizing the access to the account updates data structures and the underlying database. That approach allowed as to treat the database as a contiguous part of the in-memory data structures.

Unfortunately, doing so would also create certain undesired latencies - specifically, while writing to the database, we're unable to make read operations of the internal data structures.

Breaking this dependency require some double-checking of the round number as part of the database queries. If the round number of the database is newer than the in-memory data, we would need to wait for the memory round number to be updated and re-evaluate the query.

This change opens up the possibility for few other account data caching changes which will be implemented in subsequent PRs.

## Test Plan

1. Use existing tests which already cover the account update functions.
2. Deploy a test network to measure performance impact. ( results: no notable gains / losses )


ledger/accountdb.go
```diff
@@ -422,17 +422,17 @@ func accountsDbInit(r db.Queryable, w db.Queryable) (*accountsDbQueries, error)
 	var err error
 	qs := &accountsDbQueries{}
 
-	qs.listCreatablesStmt, err = r.Prepare("SELECT asset, creator FROM assetcreators WHERE asset <= ? AND ctype = ? ORDER BY asset desc LIMIT ?")
+	qs.listCreatablesStmt, err = r.Prepare("SELECT rnd, asset, creator FROM acctrounds LEFT JOIN assetcreators ON assetcreators.asset <= ? AND assetcreators.ctype = ? WHERE acctrounds.id='acctbase' ORDER BY assetcreators.asset desc LIMIT ?")
```
zeldovich, line 425: Why not just `SELECT .. FROM acctrounds, assetcreators ON ...`?  This would be a cross-join in SQLite, so that if there are no matching rows in the `assetcreators` table, this would return no rows as before.  I think that would require fewer special cases for dealing with the resulting row set, compared to what you're currently handling in this PR.

zeldovich:
This seems plausible, but I'm wondering for what workload will this really help?  SQLite has reasonably coarse-grained locking (which is the reason we have separate sqlite database files for the blocks DB vs. the accounts DB).  So if we miss in the in-memory caches, we will still wait, just waiting on the sqlite lock instead of our Golang lock.  What I worry about is that we'll inevitably miss on something when evaluating a block, so we will need to wait for the lock to be released.

Relaxing the locking discipline here certainly adds a non-trivial amount of complexity, so it'd be good to be sure this complexity is worth it.

ledger/accountdb.go
tsachiherman: I'm not sure if that's what we want here. We *want* to receive the database round even when there are not rows, since a mismatching round number between the database and the internal data structure means we need to re-evaluate.

zeldovich: Ah, I see, yeah, you're right.  Good point!

tsachiherman:
> This seems plausible, but I'm wondering for what workload will this really help? SQLite has reasonably coarse-grained locking (which is the reason we have separate sqlite database files for the blocks DB vs. the accounts DB). So if we miss in the in-memory caches, we will still wait, just waiting on the sqlite lock instead of our Golang lock. What I worry about is that we'll inevitably miss on something when evaluating a block, so we will need to wait for the lock to be released.
> 
> Relaxing the locking discipline here certainly adds a non-trivial amount of complexity, so it'd be good to be sure this complexity is worth it.

The type of workload that I'm looking at is such that we're writing to disk for more than half of the round time. That might seems like a lot today, but not so much if we want to increase the thoughput by a factor of 10.
When we do that, consider the following:

```
Writing |---DB writes ----|      | ---- memory writes ---- |
Read      |-1-|        |-2-| |-3-| 
```
Read 1,2 can happen from either memory/db without any issue.
Read 3 will fail the round check, and will wait for the memory write to complete.

The timing of the memory writes are in the order of milliseconds. The timing for the disk writing are in the order of seconds.

As for where this would help us - there are few places that could use better account data caching which are currently not performing to their max:
1. The agreement service, trying to evaluate votes. Getting the account data for x-320 rounds is something that we can create a "special" cache for, but we don't want this cache to be locked with the database writing.
2. Proposal verification ( i.e. before cert vote ). We want to verify that against the latest without any writing slowing us down.

We currently can't really see the benefit, as the disk throughput is sufficiently high. however, with increased number of transactions on a large populated database, I believe we'll hit the limit pretty easily.

zeldovich:
In your example scenario, read 1 can't happen from db because sqlite doesn't allow read transactions concurrently with a write transaction on the same table.

tsachiherman:
> In your example scenario, read 1 can't happen from db because sqlite doesn't allow read transactions concurrently with a write transaction on the same table.

My understanding was that it's locking it only for the during of the *commit* and not for the duration of the transaction.
But let me write a unit test to confirm the hypothesis.

zeldovich:
Now that I understand sqlite's concurrency guarantees (thanks for #1532)..  (I deleted my earlier comment where I was confused about `AtomicCommitWriteLock()`.)

Is the cost of flushing to disk (i.e. `accountsNewRound()` called by `accountUpdates.commitRound()`) coming largely from the execution time of `accountsNewRound()`, or is it the commit statement issued by the enclosing `AtomicCommitWriteLock()`?

If it's the former, then perhaps we can call `accountsNewRound()` without holding the `accountsMu` write-lock, and only grab the write-lock before trying to commit the overall transaction?

If the cost is dominated by the time it takes to run the commit, do we actually still get useful concurrency?  Your comment in this PR says that sqlite does locking during the commit, so we might not be able to run a `COMMIT` concurrently with a `SELECT`.  So, if the `COMMIT` takes a long time, sqlite will anyway prevent concurrent `SELECT` statements, and we won't get the benefit from issuing them concurrently.

tsachiherman:
> Now that I understand sqlite's concurrency guarantees (thanks for #1532).. (I deleted my earlier comment where I was confused about `AtomicCommitWriteLock()`.)
> 
> Is the cost of flushing to disk (i.e. `accountsNewRound()` called by `accountUpdates.commitRound()`) coming largely from the execution time of `accountsNewRound()`, or is it the commit statement issued by the enclosing `AtomicCommitWriteLock()`?
> 
> If it's the former, then perhaps we can call `accountsNewRound()` without holding the `accountsMu` write-lock, and only grab the write-lock before trying to commit the overall transaction?
> 
> If the cost is dominated by the time it takes to run the commit, do we actually still get useful concurrency? Your comment in this PR says that sqlite does locking during the commit, so we might not be able to run a `COMMIT` concurrently with a `SELECT`. So, if the `COMMIT` takes a long time, sqlite will anyway prevent concurrent `SELECT` statements, and we won't get the benefit from issuing them concurrently.

The previous implementation, which was using AtomicCommitWriteLock was doing just that : it was taking the write lock just before calling the sql transaction commit. Since all transactions ( whether it's read or write ) are atomic, it allowed us to execute select statement for reading while performing inserts on a separate write transaction.

Some preliminary tests shown that the commit time is very "non-stable". In some cases, it just write the changes to the WAL file, whereas at other times, it would also merge the WAL file back to the main database file. ( suspect: WAL checkpointing )
From time consumption perspective, writing the account data takes about the same time as the changes to the Merkle trie flushes. Also, the time it takes to perform the commit, is roughly equivalent to the time it takes to insert/delete all the entries on the write transaction.

The main issue that I'm trying to address here ( as I'm sure you've already figured out ) is that we have a "window" of time, at which we cannot *evaluate* the in-memory data structures. 

Oh, and as for my former comment regarding the commit-level locking - I'm pretty sure that there is a database level locking, however the duration of that lock are just enough to write a single page which would serve as the pointer to either the "new" entries or the "old" entries. I'm pretty certain that the commit back ( i.e. WAL checkpointing )is part of that.

ledger/acctupdates.go
```diff
@@ -1374,82 +1410,156 @@ func (au *accountUpdates) newBlockImpl(blk bookkeeping.Block, delta StateDelta)
...
+				au.log.Errorf("accountUpdates.lookupImpl: database round %d is behind in-memory round %d", dbRound, currentDbRound)
+			}
+			au.accountsMu.RLock()
+			needUnlock = true
+			for currentDbRound >= au.dbRound && currentDeltaLen == len(au.deltas) {
+				au.accountsReadCond.Wait()
+			}
+		} else {
+			// in non-sync mode, we don't wait since we already assume that we're syncronized.
+			au.log.Errorf("accountUpdates.lookupImpl: database round %d mismatching in-memory round %d", dbRound, currentDbRound)
```
brianolson, line 1491: if this is an error should we also return err?

ledger/acctupdates.go
```diff
@@ -1374,82 +1410,156 @@ func (au *accountUpdates) newBlockImpl(blk bookkeeping.Block, delta StateDelta)
...
-	if offset == uint64(len(au.deltas)) {
-		macct, ok := au.accounts[addr]
-		if ok {
-			return macct.data, nil
+	var offset uint64
+	var rewardsProto config.ConsensusParams
+	var rewardsLevel uint64
+	if withRewards {
+		defer func() {
+			data = data.WithUpdatedRewards(rewardsProto, rewardsLevel)
```
brianolson: only do this if err == nil?

ledger/acctupdates.go
```diff
@@ -1374,82 +1410,156 @@ func (au *accountUpdates) newBlockImpl(blk bookkeeping.Block, delta StateDelta)
...
+			if dbRound < currentDbRound {
+				au.log.Errorf("accountUpdates.getCreatorForRoundImpl: database round %d is behind in-memory round %d", dbRound, currentDbRound)
+			}
+			au.accountsMu.RLock()
+			unlock = true
+			for currentDbRound >= au.dbRound && currentDeltaLen == len(au.deltas) {
+				au.accountsReadCond.Wait()
+			}
+		} else {
+			au.log.Errorf("accountUpdates.getCreatorForRoundImpl: database round %d mismatching in-memory round %d", dbRound, currentDbRound)
```
brianolson, line 1563: log err == return err?

brianolson (commented):
I tried to do a close read and didn't find anything wrong by inspection. A couple small possible improvements.

ledger/acctupdates.go
tsachiherman: yes, it won't hurt.

ledger/acctupdates.go
tsachiherman: added

algonautshant (commented):
Looks great. 
I have some clarification questions and couple of typos.

ledger/acctupdates.go
```diff
@@ -1374,82 +1410,156 @@ func (au *accountUpdates) newBlockImpl(blk bookkeeping.Block, delta StateDelta)
...
+		rewardsProto = au.protos[offset]
+		rewardsLevel = au.roundTotals[offset].RewardsLevel
+
+		// Check if this is the most recent round, in which case, we can
+		// use a cache of the most recent account state.
+		if offset == uint64(len(au.deltas)) {
+			macct, ok := au.accounts[addr]
 			if ok {
-				return d.new, nil
+				return macct.data, nil
```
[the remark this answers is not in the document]
tsachiherman: I believe that it will. The deferred function would get called and replace the return value in place.

ledger/acctupdates.go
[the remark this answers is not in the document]
tsachiherman: this function replaces the output account data with one with rewards.

ledger/acctupdates.go
```diff
@@ -350,181 +352,216 @@ func (au *accountUpdates) ListApplications(maxAppIdx basics.AppIndex, maxResults
...
+				res = append(res, loc)
+			}
 			return res, nil
 		}
-
-		// Creatable was deleted
-		if _, ok := deletedCreatables[loc.Index]; ok {
-			continue
+		if dbRound < currentDbRound {
+			au.log.Errorf("listCreatables: database round %d is behind in-memory round %d", dbRound, currentDbRound)
```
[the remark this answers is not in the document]
tsachiherman: There is no "legit" codepath that would do that, other than someone swapping the database while the algod is running.
This is not a use case where we can truely recover from; I'm going to change that into returning an error here - but I don't expect it to be happen on any normal path.

ledger/acctupdates.go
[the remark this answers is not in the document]
algonautshant: Now I get it. It updates the data with the rewards.

ledger/acctupdates.go
```diff
@@ -350,181 +352,216 @@ func (au *accountUpdates) ListApplications(maxAppIdx basics.AppIndex, maxResults
...
+		}
+		au.accountsMu.RLock()
+		for currentDbRound >= au.dbRound && currentDeltaLen == len(au.deltas) {
```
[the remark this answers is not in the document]
tsachiherman: it's not. The `currentDbRound` is the value of `au.dbRound` on the last time we took the lock and evaluated the memory content.

What I was looking for is to have a change in the `au.dbRound`. Since `au.dbRound` can move only forward, the normal condition is that `currentDbRound < au.dbRound`. however, since we're blocking in this for-loop, I need to take the complement of that expression.
`!(currentDbRound < au.dbRound)` -> `currentDbRound >=  au.dbRound`

ledger/accountdb.go
```diff
@@ -505,39 +512,49 @@ func (qs *accountsDbQueries) listCreatables(maxIdx basics.CreatableIndex, maxRes
...
+			copy(addr[:], buf)
+		}
 		return nil
 	})
 	return
 }
 
-func (qs *accountsDbQueries) lookup(addr basics.Address) (data basics.AccountData, err error) {
+// lookup looks up for a the account data given it's address. It returns the current database round and the matching
+// account data, if such was found. If matching account data was found for the given address, the empty account data would
```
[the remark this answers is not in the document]
tsachiherman: fixed.

ledger/acctupdates.go
```diff
@@ -1374,82 +1410,156 @@ func (au *accountUpdates) newBlockImpl(blk bookkeeping.Block, delta StateDelta)
...
+	var offset uint64
+	for {
+		currentDbRound := au.dbRound
+		currentDeltaLen := len(au.deltas)
+		offset, err = au.roundOffset(rnd)
+		if err != nil {
+			return basics.Address{}, false, err
+		}
+
+		// If this is the most recent round, au.creatables has will have the latest
```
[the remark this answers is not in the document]
tsachiherman: fixed.

algonautshant (dismissed):
Thanks for the explanation and the changes.

ledger/acctupdates.go
```diff
@@ -1374,82 +1434,164 @@ func (au *accountUpdates) newBlockImpl(blk bookkeeping.Block, delta StateDelta)
...
-	// state and we can skip scanning backwards over creatableDeltas
-	if offset == uint64(len(au.deltas)) {
-		// Check if we already have the asset/creator in cache
-		creatableDelta, ok := au.creatables[cidx]
-		if ok {
-			if creatableDelta.created && creatableDelta.ctype == ctype {
-				return creatableDelta.creator, true, nil
-			}
-			return basics.Address{}, false, nil
+	defer func() {
```
algorandskiy, line 1531: nit: moving this `defer` inside the `if` cond and restoring the lock right after `au.accountsq.lookupCreator` will allow to get rid of `unlock` variable.

ledger/acctupdates.go
```diff
@@ -350,181 +375,217 @@ func (au *accountUpdates) ListApplications(maxAppIdx basics.AppIndex, maxResults
...
-				Type:    delta.ctype,
-				Index:   cidx,
-				Creator: delta.creator,
-			})
-		} else {
-			// Mark deleted creatables for exclusion from the results set
-			deletedCreatables[cidx] = true
+		// Check in-memory created creatables, which will always be newer than anything
+		// in the database
+		var res []basics.CreatableLocator
```
algorandskiy, line 415: nit: `res = append(res, loc[:max(maxResults, len(loc)]...)`  + cond. I mean we can get rid loop here.

algorandskiy (approved):
LGTM but too complex.

ledger/acctupdates.go
tsachiherman: I was considering that, but given that the loop could take multiple iterations, we don't want to defer the function call multiple times, right ?

ledger/acctupdates.go
tsachiherman: that's a good point, and I'll do that next in a separate PR.

PR #1528, opened 2020-09-20
Fix heuristicFormatStr logic

tsachiherman:
## Summary

The goalAppAccountAddressTest was randomally failing. The culprit was the heuristicFormatStr which was testing to see if the string is unicode printable before printing it out. To address this issue, I made two modifications by adding steps 1 and 2 below :
1. A byte array that corresponds to a checksumed address, would be printed as address.
2. A json formatteble string, would be printed as is.
3. A byte array that corresponds to a non-checksumed address, would be printed as address
4. The string would be printed as is.

Note that the checking if the byte array was printable was misleading : it was checking if it's a valid unicode string, instead of checking if it's a json string. I think that we might want to add another option for forcing a specific bytes output formatter, but that's already a separate issue.

## Test Plan

Existing (failing) test provides coverage for this use case.

### DIFF
--- a/cmd/goal/application.go
+++ b/cmd/goal/application.go
@@ -1048,17 +1048,7 @@ func jsonPrintable(str string) bool {
 }
 
 func heuristicFormatStr(str string) string {
-	// if it's a base32 address string which has a valid checksum, then make the assumption it's an address
-	if len(str) == 58 {
-		// try to decode the string and perform a checksum validation on it.
-		_, err := basics.UnmarshalChecksumAddress(str)
-		// if err is nil, then the checksum is valid
-		if err == nil {
-			return str
-		}
-	}
-
-	// otherwise, see if we can print it as a json output
+	// See if we can print it as a json output
 	if jsonPrintable(str) {
 		return str
 	}
--- a/cmd/goal/application.go
+++ b/cmd/goal/application.go
@@ -1048,11 +1048,13 @@ func jsonPrintable(str string) bool {
 }
 
 func heuristicFormatStr(str string) string {
-	// if it's a 32 byte string which has a valid checksum, then make the assumption it's an address
-	if len(str) == 32 {
-		addr, err := basics.UnmarshalChecksumAddress(str)
+	// if it's a base32 address string which has a valid checksum, then make the assumption it's an address
+	if len(str) == 58 {
+		// try to decode the string and perform a checksum validation on it.
+		_, err := basics.UnmarshalChecksumAddress(str)
+		// if err is nil, then the checksum is valid
 		if err == nil {
-			return addr.String()
+			return str
 		}
 	}
 
--- a/cmd/goal/application.go
+++ b/cmd/goal/application.go
@@ -25,6 +25,7 @@ import (
 	"strconv"
 	"strings"
 	"unicode"
+	"unicode/utf8"
 
 	"github.com/spf13/cobra"
 
@@ -919,7 +920,7 @@ var deleteAppCmd = &cobra.Command{
 	},
 }
 
-func printable(str string) bool {
+func unicodePrintable(str string) bool {
 	for _, r := range str {
 		if !unicode.IsPrint(r) {
 			return false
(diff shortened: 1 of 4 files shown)


cmd/goal/application.go
```diff
@@ -928,17 +929,146 @@ func printable(str string) bool {
...
 
+func jsonPrintable(str string) bool {
+	// htmlSafeSet holds the value true if the ASCII character with the given
+	// array position can be safely represented inside a JSON string, embedded
+	// inside of HTML <script> tags, without any additional escaping.
+	//
+	// All values are true except for the ASCII control characters (0-31), the
+	// double quote ("), the backslash character ("\"), HTML opening and closing
+	// tags ("<" and ">"), and the ampersand ("&").
+	var htmlSafeSet = [utf8.RuneSelf]bool{
```
brianolson, line 940: would it be better to generate this programatically? Or flip the logic and have a small set of unsafe runes? `unsafe := unsafeSet[c]` would only need the few exception chars set true.

cmd/goal/application.go
```diff
@@ -928,17 +929,146 @@ func printable(str string) bool {
...
+			return false
+		}
+	}
+	return true
+}
+
 func heuristicFormatStr(str string) string {
-	if printable(str) {
+	// if it's a 32 byte string which has a valid checksum, then make the assumption it's an address
+	if len(str) == 32 {
```
brianolson, line 1052: Do you mean 58? `len('XFVTOHVXBJRMGMDNUGC7KA3IFK7QCJ2F7FMZDZOEC4TIUKEXIE6FID6S3U')==58`

cmd/goal/application.go
```diff
@@ -928,17 +929,146 @@ func printable(str string) bool {
...
+		addr, err := basics.UnmarshalChecksumAddress(str)
+		if err == nil {
+			return addr.String()
```
brianolson, line 1055: addr.String() _should_ be == `str`, but I guess this would make it more canonical if it isn't quite?

cmd/goal/application.go
tsachiherman: no; if you'll look where the data is coming from, you'll see that it's the byte slice stored in the TEALValue, which contains an address. The use case which you refer to is the one below, where this is printable.

cmd/goal/application.go
tsachiherman: no, it's not. look in 
```
// String returns a string representation of Address
func (addr Address) String() string {
	addrWithChecksum := make([]byte, crypto.DigestSize+checksumLength)
	copy(addrWithChecksum[:crypto.DigestSize], addr[:])
	// calling addr.GetChecksum() here takes 20ns more than just rolling it out, so we'll just repeat that code.
	shortAddressHash := crypto.Hash(addr[:])
	copy(addrWithChecksum[crypto.DigestSize:], shortAddressHash[len(shortAddressHash)-checksumLength:])
	return base32Encoder.EncodeToString(addrWithChecksum)
}
```

cmd/goal/application.go
tsachiherman: I wasn't that creative. I just copied it from the Golang json encoding library. That's how the json library is determining if it's going to encode that as is, or whether it would apply some additional formatting.
unfortunately, the go library keep this array as an internal array, so I had to duplicate it.

cmd/goal/application.go
brianolson: This len(str)==32 clause and the one on line 1065 do opposite things, and I think only one of them can be right.
This one parses from base32 text to a [32]byte Address, the other one encodes to base32. Lines 1065-1069 assume that a [32]byte or []byte has been typecast to string and that if we typecast it back and copy out the bytes we get an Address.
Here I don't expect 32 chars of base32 encoded data will come to anything meaningful, but 58 chars is the base32 encoded length of a [32]byte Address with checksum.
If this code works, I expect the Unmarshal() hits the `err != nil` case, falls through, and eventually just returns exactly the text that was passed in (which may happen to have been a base32 Address-with-checksum).

tsachiherman: I believe you're correct. Let me revise it.

tsachiherman: updated. thanks for the feedback!

cmd/goal/application.go
```diff
@@ -928,17 +929,148 @@ func printable(str string) bool {
...
+		if htmlSafeSet[r] == false {
+			return false
+		}
+	}
+	return true
+}
+
 func heuristicFormatStr(str string) string {
-	if printable(str) {
+	// if it's a base32 address string which has a valid checksum, then make the assumption it's an address
```
brianolson, line 1051: I believe this is now correct as intended, but I also think it doesn't add anything. Everything in the base32 set is in the jsonPrintable set, so falling through to the next clause would be as good. If you still want to explicitly call out that addresses should pass through, I won't block it.

tsachiherman: The trigger for this change was the `goalAppAccountAddressTest` which was failing. It was failing because *some* of the valid addresses were a valid unicode printable strings, but invalid json printable strings.
as a result, the output of `goal ...` used to contain "encoded" entries for some of the account addresses.

I'm going to remove the top section completely, and leave only the json printing test, as it seems to suffice.

brianolson (approved):

PR #1532, opened 2020-09-21
Add concurrent read and write test from database table

tsachiherman:
## Summary

Test to see if sqlite blocks when we attempt to read while writing to a database table.

## Test Plan

This is a test.


zeldovich (approved):
Indeed, looks like you are correct about the concurrency offered by sqlite; I found that https://sqlite.org/wal.html has a good explanation of what's going on.

Issue #1530, opened 2020-09-21
TEAL - arguments counts and traps


jeapostrophe:
Application contracts can check the number of arguments they have been provided via `NumAppArgs`, but normal arguments do not have a similar field or opcode to look at.

For application contracts, there's no discussion in the docs about what happens if you attempt to read an argument after this limit. I expect it should trap. For non-application contracts, I'd expect the same behavior, but again, this is not discussed.

Finally, application contracts in the same group can read each other's arguments via `gtxna` but non-application contracts cannot read each other's arguments (nor can application contracts read the arguments of non-application contracts in the same group.)

jeapostrophe:
The code causes a trap on reading past the last arg in sig mode: https://github.com/algorand/go-algorand/blob/master/data/transactions/logic/eval.go#L1089

And in app mode: https://github.com/algorand/go-algorand/blob/master/data/transactions/logic/eval.go#L1329

jeapostrophe:
This line could have an indirection

https://github.com/algorand/go-algorand/blob/master/data/transactions/logic/eval.go#L1089

like this line

https://github.com/algorand/go-algorand/blob/master/data/transactions/logic/eval.go#L1485

And then you'd be able to read the arguments of other transactions.

jeapostrophe:
Here's an implementation: https://github.com/algorand/go-algorand/pull/1533

jeapostrophe:
The pull-request does not address the docs problem.

jannotti:
Agreed that all the array txn opcodes should discuss their behavior when reading part the end. I'm going to close this issue because it is mostly talking about the args issue which has been addressed (to the extent that it will be addressed, anyway)

PR #1533, opened 2020-09-21
TEAL - Add Args, NumArgs

jeapostrophe:
https://github.com/algorand/go-algorand/issues/1530

### DIFF
--- a/data/transactions/logic/eval.go
+++ b/data/transactions/logic/eval.go
@@ -1249,9 +1249,9 @@ func (cx *evalContext) assetParamsEnumToValue(params *basics.AssetParams, field 
 }
 
 // TxnFieldToTealValue is a thin wrapper for txnFieldToStack for external use
-func TxnFieldToTealValue(txn *transactions.Transaction, groupIndex int, field TxnField, arrayFieldIdx uint64) (basics.TealValue, error) {
+func TxnFieldToTealValue(stxn *transactions.SignedTxn, groupIndex int, field TxnField, arrayFieldIdx uint64) (basics.TealValue, error) {
 	cx := evalContext{EvalParams: EvalParams{GroupIndex: groupIndex}}
-	sv, err := cx.txnFieldToStack(txn, field, arrayFieldIdx, groupIndex)
+	sv, err := cx.txnFieldToStack(stxn, field, arrayFieldIdx, groupIndex)
 	return sv.toTealValue(), err
 }
 
@@ -1271,7 +1271,8 @@ func (cx *evalContext) getTxID(txn *transactions.Transaction, groupIndex int) tr
 	return txid
 }
 
-func (cx *evalContext) txnFieldToStack(txn *transactions.Transaction, field TxnField, arrayFieldIdx uint64, groupIndex int) (sv stackValue, err error) {
+func (cx *evalContext) txnFieldToStack(stxn *transactions.SignedTxn, field TxnField, arrayFieldIdx uint64, groupIndex int) (sv stackValue, err error) {
+	txn := &stxn.Txn
 	err = nil
 	switch field {
 	case Sender:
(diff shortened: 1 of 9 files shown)


jeapostrophe:
I don't know if this requires a version change now or if those get bundled in the future during a release process.

jeapostrophe:
For the failing test, I was under the assumption that if I looked at `Lsig` and it wasn't there, there would be a trap and it would fail and that would be fine. It seems like that's not the case. I've never written a Go program before, but presumably there's a way in the 1330s of `eval.go` to check if `Lsig` isn't present and return 0. I'll look into this shortly, but if it easy for someone else, that'd be great.

jeapostrophe:
I updated the tests to pass

jeapostrophe:
I don't think the travis error is related to this change.

data/transactions/logic/eval.go
```diff
@@ -1325,6 +1326,14 @@ func (cx *evalContext) txnFieldToStack(txn *transactions.Transaction, field TxnF
 		sv.Uint = uint64(txn.ApplicationID)
 	case OnCompletion:
 		sv.Uint = uint64(txn.OnCompletion)
+	case Args:
```
brianolson, line 1329: We don't need `txn Args N` because we already have the `arg N` ops

tsachiherman:
@brianolson - these changes would require a consensus protocol upgrade for sure, right ?
( what about teal version ? would we it also require bumping this one up ? )

brianolson:
@tsachiherman Yes, this would require both a protocol upgrade and TEAL version upgrade. So if we decide to do this, I expect it to get batched along with a bunch of other changes and maybe released several months from now? (There were about six months between TEAL 1 and TEAL 2.)

tsachiherman:
Thanks for the confirmation @algobolson; @jeapostrophe - at this time, I'm unaware of any planned changes to the TEAL implementation. We will definitely consider to include these changes ( or the intent behind them ), at some later release.

jeapostrophe:
@tsachiherman Thanks. I'll monitor the pull-request to see if merges cleanly. If you get around to the point of deciding, please pull me back in and I can update it.

data/transactions/logic/eval.go
algorandskiy: +1 to Brian, `arg N` is already in place. `NumArgs` looks legit.

jeapostrophe: Hi @algorandskiy You are right that `arg N` is the same `txna Args N`, but there is no way to do `gtxna Args N` without this patch. Meaning that txn 0 cannot look at txn 1's arguments.

If you'd like to see an example of this checkout --- https://github.com/reach-sh/reach-lang/blob/master/examples/overview/build/index.main.mjs#L166 --- and if you'd like an explanation, you can watch my video --- https://www.youtube.com/watch?v=6QXISWGoul4 --- where I explain how I compile Reach to Algorand

jeapostrophe:
Sorry for the delay posting my comments... they got trapped in Github's pending code review system.

data/transactions/logic/eval.go
algorandskiy: Yeah, now I see why, thanks for the explanation. I'm a bit worried about the design and consistency - currently `txna`/`gtxna` access transaction fields not signature fields. Sounds like we either want `txn` accessing all signature fields or to have another opcode for signature fields accessing. Anyway, thank you for the proposal! Now it needs to want to be batched with other for TEAL v3.

jeapostrophe: I don't have a strong opinion on the name. I can see a case for (A) `txn/gtxn/txna/gtxna` are for _all transaction_ information and (B) making sure that users know that a transaction's signature information is in a sub-category from the transaction's general information. I, perhaps false, feel pretty confident in my understanding of Algorand and I don't find it intuitive to think about them differently, so that's why I prefer (A)

jannotti:
Closing based on discussion about dangers of exposing lsig args to other transactions. In theory, there's nothing wrong with it, if the semantics are well understood - lsig args are not signed. But in practice people are likely to misuse more than they would gain any advantage. The notes field is signed, and could be a better place to share info across txns in a group.

PR #1536, opened 2020-09-22
Deprecate algolambdamsec environment variable

nicholasguoalgorand:
## Summary

Since we are now using AgreementFilterTimeout as a consensParam, we can remove the algolambdamsec environment variable from the codebase and eliminate that tech debt.

## Test Plan

Run unit tests again to make sure functionality still the same

### DIFF
--- a/config/consensus.go
+++ b/config/consensus.go
@@ -21,7 +21,6 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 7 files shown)


tsachiherman:
it looks good to me; let's see if it passes the automated tests.

tsachiherman (approved):
Look good.

PR #1539, opened 2020-09-22
Correctly account for bloom filter binary marshaling size

tsachiherman:
## Summary

The http read size limiter for the tx sync was not configured correctly, causing failures when bloom filter requests reached or were near capacity.

### Quick summary

The allowed body buffer read size should have been 8 bytes longer.

## Test Plan

Use the performance system, where the issue was found as a test bed.


brianolson (approved):
for bonus points we could have a unit test that bloom.go Filter.MarshalBinary() returns a blob whose len() agrees with BinaryMarshalLength(); but I believe this is correct.

tsachiherman:
> for bonus points we could have a unit test that bloom.go Filter.MarshalBinary() returns a blob whose len() agrees with BinaryMarshalLength(); but I believe this is correct.

I'll get the bonus on a separate PR. ;-)

PR #1540, opened 2020-09-22
Windows version

mxmauro:
This commit makes modifications to source code to be able to generate native Windows binaries with the help of MSYS2 build environment.

Some of the original shell scripts for building the apps were sightly modified also to work under Windows environments.

The instructions, also located in `/scripts/windows/instructions.md` to build the executable files are:

1. Download and install `MSYS2` package from [here](https://www.msys2.org/)

2. Run `MSYS2 MingW 64-bit` application to open the MSYS2 terminal.

3. Update MSYS2 package and dependency manager by running the following commands:

	```
	pacman -Syu --disable-download-timeout
	```

	NOTE: It is very likely MSYS2 will ask to close the window and repeat the command for furter updates. Check `MSYS2` web page for additional support.

4. Install GIT on MSYS2 by executing the following command:

	```
	pacman -S --disable-download-timeout --noconfirm git
	```

5. Clone repository with `git clone https://github.com/algorand/go-algorand`

6. Switch to source code directory with `cd go-algorand`.

7. Run `./scripts/windows/install_deps.sh` to install required dependencies.

8. Run `make`.

In `/home/{you-user}/go/bin` you will find all the apps. There is no installer package right know but you can copy `genesis.json` and create a `config.json` under `data` subdirectory and run `goal node start -d data` as you would do under other OSes.

image


crypto/memcpy_chk_windows.c
```diff
@@ -0,0 +1,35 @@
...
+extern void _cgo_panic(void *, int);
+
+void * __memcpy_chk (void *dstpp, const void *srcpp, size_t len, size_t dstlen)
+{
+	if (dstlen < len) {
+		struct { const char *p; } a;
+
+		a.p = "panic from __memcpy_chk";
+		crosscall2(_cgo_panic, &a, sizeof a);
+		*(int*)1 = 1;
```
tsachiherman, line 32: can you release the 
```
*(int*)1 = 1;
```
with a `RaiseException` call, and avoid the memcpy ?

if we were to use a static code analysis, it could flag this line, but using the RaiseException it would work just fine.

libgoal/lockedFileWindows.go
```diff
@@ -0,0 +1,95 @@
...
+	winLockfileExclusiveLock   = 0x00000002
+	winLockfileSharedLock      = 0x00000000
+)
+
+const ErrorLockViolation syscall.Errno = 0x21 // 33
+
+// makeLocker create a windows file locker.
+// note that the desired way is to use the OFD locker, which locks on the file descriptor level.
+// falling back to the non-OFD lock would allow obtaining two locks by the same process. If this becomes
+// and issue, we might want to use flock, which wouldn't work across NFS.
```
tsachiherman, line 48: I'm pretty sure that this comment is not relevant to this implementation.

util/util_windows.go
```diff
@@ -0,0 +1,28 @@
+// +build darwin freebsd netbsd openbsd
```
tsachiherman, line 1: didn't you meant to remove this line ?

tsachiherman (commented):
If we'll put aside that build environment changes, most of the go-code changes looks very well written and I don't have any reason to block these.
Unfortunately, I don't have the setup to test the build environment, so I can't approve this change. Maybe with some additional changes to the travis automated build, I could see how this compile successfully.

mxmauro:
Hi @tsachiherman thanks for the feedback. I'll check it better and upload fixes. BTW: Do you know the reason of the `golint` error Travis is throwing?

util/util_windows.go
mxmauro: Fixed. Funny bug hahaha

mxmauro:
Hi @tsachiherman finally I could get rid of the issues.

Something weird I found, `util_windows.go` has a wrong license header but `check_license.sh` script started to modify all .go files because of this.

I don't how to setup Travis CI. But I asked for assistance to @rfustino to also check on Windows. Should be straight-forward. Just Windows 10 + MSYS2 and running a script to install dependencies.

Of course I can assist in whatever you need.

util/util_windows.go
```diff
@@ -0,0 +1,82 @@
...
+		stime = filetimeToDuration(&kernelTime).Nanoseconds()
+	} else {
+		utime = 0
+		stime = 0
+	}
+	return
+}
+
+func filetimeToDuration(ft *syscall.Filetime) time.Duration {
+	n := int64(ft.HighDateTime)<<32 + int64(ft.LowDateTime) // in 100-nanosecond intervals
```
tsachiherman, line 68: Adding the `LowDataTime` is fine.. but you'd like to use uint64 here to avoid the negative sign.
The same holds true for the `HighDateTime`, but I don't know if we'll see that bit set any time soon.

cmd/kmd/mlock_windows.go
```diff
@@ -0,0 +1,24 @@
...
+// You should have received a copy of the GNU Affero General Public License
+// along with go-algorand.  If not, see <https://www.gnu.org/licenses/>.
+
+package main
+
+import (
+	"github.com/algorand/go-algorand/logging"
+)
+
+func tryMlockall(_ logging.Logger) {
```
tsachiherman, line 23: could you please add an info message here, something like 
log.Infof("running on windows -- mlockall not available")
and add a comment that we should look into solutions such as 
https://github.com/elastic/elasticsearch/pull/9186

logging/usage_windows.go
```diff
@@ -0,0 +1,65 @@
...
+
+import (
+	"context"
+	"sync"
+	"syscall"
+	"time"
+)
+
+// UsageLogThread utility logging method
+func UsageLogThread(ctx context.Context, log Logger, period time.Duration, wg *sync.WaitGroup) {
```
tsachiherman, line 27: The `UsageLogThread` implementation should be unified for all the platforms. Only the underlying time acquisition portion should be platform dependent.

util/util_windows.go
mxmauro: In the new code changed the retrieval of process times to `int64` like the nanoseconds used in Linux

util/util_windows.go
mxmauro: Done

logging/usage_windows.go
mxmauro: Reimplemented code and process times are retrieved from a function in `utils` also used in network test file.

scripts/configure_dev.sh
```diff
@@ -61,6 +100,21 @@ elif [ "${OS}" = "darwin" ]; then
...
     install_or_upgrade python3
+elif [ "${OS}" = "windows" ]; then
+    pacman -S --disable-download-timeout --noconfirm git automake autoconf m4 libtool mingw-w64-x86_64-python3 make mingw-w64-x86_64-gcc mingw-w64-x86_64-go mingw-w64-x86_64-boost mingw-w64-x86_64-python unzip procps
+    if [ $? -ne 0 ]
+    then
+        echo "Error installing pacman dependencies"
+        exit 1
+    fi
+
+    export GOPATH=$HOME/go
```
algonautshant, line 111: Why configure_dev needs to export GOPATH here (only for windows)?

scripts/configure_dev.sh
```diff
@@ -41,6 +41,45 @@ function install_or_upgrade {
     fi
 }
 
+function install_windoows_shellcheck() {
```
algonautshant, line 44: This function is not needed here, since a file is added for the same purpose: scripts/windows/install_shellcheck.sh

algonautshant (commented):
Looks very nice. 
I have some comments/questions.

Makefile
```diff
@@ -1,9 +1,16 @@
-export GOPATH		:= $(shell go env GOPATH)
-GOPATH1		:= $(firstword $(subst :, ,$(GOPATH)))
+UNAME := $(shell uname)
+ifneq (, $(findstring MINGW,$(UNAME)))
+	#Gopath is not saved across sessions, probably existing Windows env vars, override them
+	export GOPATH := ${HOME}/go
```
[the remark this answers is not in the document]
mxmauro: Hi @algonautshant , The whole build process, including the installation of Go, runs inside MSYS2 ecosystem. `HOME` defaults  to `/home/{your-windows-username}`.

Although you can access the whole disk with `/c/...`, `/home` and the rest of the common Unix/Linux directories are located inside MSYS2 installation path.

Makefile
```diff
@@ -1,9 +1,16 @@
-export GOPATH		:= $(shell go env GOPATH)
-GOPATH1		:= $(firstword $(subst :, ,$(GOPATH)))
+UNAME := $(shell uname)
+ifneq (, $(findstring MINGW,$(UNAME)))
```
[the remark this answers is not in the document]
mxmauro: Yes, may be. Bash and Linux is not my strong suite. :)

Makefile
```diff
@@ -38,6 +45,10 @@ endif
 endif
 endif
 
+ifneq (, $(findstring MINGW,$(UNAME)))
+EXTLDFLAGS := -static-libstdc++ -static-libgcc
```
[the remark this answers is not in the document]
mxmauro: Do you prefer to add to this line or remove the ones at the beginning like the rest of the file?

network/wsNetwork_windows.go
```diff
@@ -0,0 +1,23 @@
...
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with go-algorand.  If not, see <https://www.gnu.org/licenses/>.
+
+// +build windows
+
+package network
+
+func (wn *WebsocketNetwork) rlimitIncomingConnections() error {
+	return nil
```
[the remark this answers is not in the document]
mxmauro: There is no default limit on Windows. If I return an error, the test that uses it will fail to run. Are you sure to change this?

scripts/configure_dev.sh
mxmauro: I completely forgot about an old attempt to move all the actions of the `/scripts/windows` directory inside `configure_dev.sh` to normalize set up of development environment. I'll fix it.

About exporting `GOPATH`, because the whole process runs under some sort of virtual environment, I assume Go is not installed and I need it to run `configure_dev-deps.sh`, install shellcheck (there is no package available under MSYS2, and copy `libwinpthread-1.dll` (cannot compile statically because of `karalabe/hid` dependency using libusb)

Makefile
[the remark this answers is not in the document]
algonautshant: I like the indentation. I prefer adding indentation here as well. Thanks.

network/wsNetwork_windows.go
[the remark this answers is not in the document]
algonautshant: In that case, I guess let's keep it as it is. Thanks.

Makefile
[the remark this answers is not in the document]
algonautshant: Thank you.

algonautshant (dismissed):

mxmauro:
Hi @algonautshant , I moved the installation of dependencies to the `configure_dev.sh` script and also merged almost the latest changes from master.

Despite of this, I'm a bit lost on the failed Travis build about external arm64.

algonautshant:
@mxmauro your job is good. It was one of the tests timed-out. A rerun passed.

mxmauro:
Thanks @algonautshant. Next step would be to create Travis support and an installation package using Wix (although need to see how to integrate Wix Toolset and MSYS2)

scripts/configure_dev.sh
```diff
@@ -61,11 +89,29 @@ elif [ "${OS}" = "darwin" ]; then
...
+    if [ $? -ne 0 ]
+    then
+        echo "Error installing pacman dependencies"
+        exit 1
+    fi
+
+    # Golang probably is not installed under MSYS2 so add the environment variable temporarily
+    export GOPATH=$HOME/go
+
+    install_windoows_shellcheck
```
tsachiherman, line 103: typo : `windoows` -> `windows`

scripts/configure_dev.sh
```diff
@@ -61,11 +89,29 @@ elif [ "${OS}" = "darwin" ]; then
...
+    if [ $? -ne 0 ]
+    then
+        exit 1
+    fi
+
+    # This is required because http://github.com/karalabe/hid library compiles with non-static libraries
+    cp /mingw64/bin/libwinpthread-1.dll $GOPATH/bin/
```
tsachiherman, line 110: I think it would be cleaner to add the `/mingw64/bin` directory to the path on compiling machines.
For packaged deployment, we need to copy the above file anyway.

nodecontrol/NodeController.go
```diff
@@ -129,7 +131,7 @@ func killPID(pid int) error {
 		}
 		select {
 		case <-waitLong:
-			return syscall.Kill(pid, syscall.SIGKILL)
+			return util.KillProcess(pid, syscall.SIGTERM)
```
tsachiherman, line 134: Unless there is a good reason to change the SIGKILL to SIGTERM, I'd like to stick with the existing logic and keep it with SIGTERM. SIGTERM can be caught and ignored, whereas SIGKILL cannot.

scripts/configure_dev.sh
mxmauro: Actually is already there when you open the MSYS2 MingW64 terminal:

This is my current configuration:

```
Blaster64@BlasterNotebook MINGW64 ~
$ echo $PATH
/mingw64/bin:/usr/local/bin:/usr/bin:/bin:/c/Windows/System32:/c/Windows:/c/Windows/System32/Wbem:/c/Windows/System32/WindowsPowerShell/v1.0/:/usr/bin/site_perl:/usr/bin/vendor_perl:/usr/bin/core_perl
```

nodecontrol/NodeController.go
mxmauro: I saw it tonight and was about to fix it.

scripts/configure_dev.sh
mxmauro: I hate my "mechanic" keyboard

util/process_windows.go
```diff
@@ -0,0 +1,67 @@
...
+
+import (
+	"os"
+	"unsafe"
+
+	"golang.org/x/sys/windows"
+)
+
+// KillProcess kills a running OS process
+func KillProcess(pid int, _ os.Signal) error {
```
tsachiherman, line 29: This implementation is a great fit for the os.Signal SIGKILL. It doesn't emulate the SIGTERM correctly on windows.
Implementing it correctly require some extra work ( which can be deferred to a follow-up PR ):
1. Add a hidden console window to the algod and configure a SetConsoleCtrlHandler
2. In KillProcess, provide handling for SIGINT:
    Find the window that is associated with the process, and use GenerateConsoleCtrlEvent
3. In KillProcess, provide handling for SIGTERM:
    Find the window that is associated with the process, and send WM_QUIT using SendMessage.

scripts/configure_dev.sh
tsachiherman: if it's already there, why do you need to copy it ? windows DLL loader would already search on the PATH, and would find it. Is there any usage of this DLL as a library source file ? i.e. building the stub out of an existing DLL ?

Makefile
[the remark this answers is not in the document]
bricerisingalgorand: SInce this is `ifneq`, this will overwrite the `EXTLDFLAGS` set for linux in the previous if block won't it?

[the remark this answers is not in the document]
tsachiherman: this is a tricky expression -
on linux, the following expression `$(findstring MINGW,$(UNAME))` would evaluate to an empty string, right ?
if is, the expression `ifeq (, <empty string>)` is clearly true, and the `ifneq (, <empty string>)` is false.

On the other hand, on windows, it's the opposite. A comment above could clarify things up.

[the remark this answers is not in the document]
mxmauro: It is a similar approach to the used to test flags: https://www.gnu.org/software/make/manual/make.html#Testing-Flags

scripts/configure_dev.sh
mxmauro: Done

scripts/configure_dev.sh
mxmauro: Because you usually don't run the app inside MSYS2. The goal is to create executable files that can be distributed and used anywhere.

Despite it is a simple wrapper to convert pthread api into native Windows ones, I'll continue trying to remove that dependency and see if I can compile the whole code statically.

About source code, here: https://github.com/mirror/mingw-w64/tree/master/mingw-w64-libraries/winpthreads

util/process_windows.go
mxmauro: Go already sets up a SetConsoleCtrlHandler but the problem is how to send the signal to the target process.

Workarounds includes injecting a remote thread or attaching the target console (I don't like this approach)

Whatever the method, it will have some complexity.

tsachiherman (approved):
looks great, thank you for your contribution.

mxmauro:
You are welcome.

Next steps:

1. Investigate about Travis CI
2. If suits for you, use github.com/kardianos/service or build some sort of helper app to be able to launch algod as a service.
3. Installer package.

PR #1542, opened 2020-09-23
Add detailed error logging in case of failed blocks download

tsachiherman:
## Summary

Prior to this PR, a failed block download during `processStageBlocksDownload` would not get logged. Instead, it would retry to download the block following the configured `cs.config.CatchupBlockDownloadRetryAttempts` limits.

This PR adds an info logging in case of such a failure. The intent of this PR is to provide the node runner with the information about the download error so that she would be able to take concrete actions around that specific failure.

## Test Plan

Tested manually.

### DIFF
--- a/catchup/catchpointService.go
+++ b/catchup/catchpointService.go
@@ -466,7 +466,7 @@ func (cs *CatchpointCatchupService) processStageBlocksDownload() (err error) {
 				}
 				if attemptsCount <= uint64(cs.config.CatchupBlockDownloadRetryAttempts) {
 					// try again.
-					cs.log.Infof("Failed to download block %d. %v", topBlock.Round()-basics.Round(blocksFetched), err)
+					cs.log.Infof("Failed to download block %d on attempt %d out of %d. %v", topBlock.Round()-basics.Round(blocksFetched), attemptsCount, cs.config.CatchupBlockDownloadRetryAttempts, err)
 					continue
 				}
 				return cs.abort(fmt.Errorf("processStageBlocksDownload failed after multiple blocks download attempts"))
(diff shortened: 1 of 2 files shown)


catchup/catchpointService.go
```diff
@@ -466,6 +466,7 @@ func (cs *CatchpointCatchupService) processStageBlocksDownload() (err error) {
 				}
 				if attemptsCount <= uint64(cs.config.CatchupBlockDownloadRetryAttempts) {
 					// try again.
+					cs.log.Infof("Failed to download block %d. %v", topBlock.Round()-basics.Round(blocksFetched), err)
```
pbennett, line 469: Perhaps show attempt x of y in message    
ie:
Failed to download block %d (attempt %d of %d). %v

catchup/httpFetcher.go
```diff
@@ -100,8 +100,13 @@ func (hf *HTTPFetcher) GetBlockBytes(ctx context.Context, r basics.Round) (data
 		return nil, errNoBlockForRound
 	default:
 		hf.log.Warn("http block fetcher response status code : ", response.StatusCode)
-		response.Body.Close()
```
algonautshant, line 103: why response Body is no longer closed in this case?

algonautshant: because it is closed in rpcs.ResponseBytes :-)

algonautshant (approved):
Looks good.

catchup/catchpointService.go
tsachiherman: I don't know if that's a super useful piece of information here, but I don't mind adding it either.

algonautshant (approved):

PR #1546, opened 2020-09-23
Extend expect timeout when checking account balance

egieseke:
## Summary

Extend timeout for the Expect common function `WaitForAccountBalance` to wait for the account balance to reflect the expected amount.  Changed from 20 to 40 seconds.  

## Test Plan

Ran Integration tests.

### DIFF
--- a/test/e2e-go/cli/goal/expect/goalExpectCommon.exp
+++ b/test/e2e-go/cli/goal/expect/goalExpectCommon.exp
@@ -453,7 +453,7 @@ proc ::AlgorandGoal::WaitForAccountBalance { WALLET_NAME ACCOUNT_ADDRESS EXPECTE
                 puts "Account balance OK: $ACCOUNT_BALANCE"; break
             } else {
                puts "Account balance: '$ACCOUNT_BALANCE' does not match expected balance: '$EXPECTED_BALANCE'"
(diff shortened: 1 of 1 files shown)


test/e2e-go/cli/goal/expect/goalExpectCommon.exp
```diff
@@ -453,7 +453,7 @@ proc ::AlgorandGoal::WaitForAccountBalance { WALLET_NAME ACCOUNT_ADDRESS EXPECTE
                 puts "Account balance OK: $ACCOUNT_BALANCE"; break
             } else {
                puts "Account balance: '$ACCOUNT_BALANCE' does not match expected balance: '$EXPECTED_BALANCE'"
-               if { $i >= 10 } then { ::AlgorandGoal::Abort "Account balance $ACCOUNT_BALANCE does not match expected amount: $EXPECTED_BALANCE"; break;}
+               if { $i >= 20 } then { ::AlgorandGoal::Abort "Account balance $ACCOUNT_BALANCE does not match expected amount: $EXPECTED_BALANCE"; break;}
```
brianolson, line 456: Maybe make these two messages different? The non-error case might end with ", still waiting..." ? The timeout case might end with, ", waited too long, FAIL" ?

egieseke: Good thinking, I updated the messaging.

brianolson (approved):

Issue #1543, opened 2020-09-23
Resolve warning message in logs: peer write error websocket: close sent


egieseke:
The following Warning message is visible in the performance test logs, "peer write error websocket: close sent",

`Message:peer write error websocket: close sent Data.log:{"details":{"Address":"3.133.123.122","HostName":"e1e2d17b-af57-483b-b076-7684a85a603e:N18","Incoming":true,"InstanceName":"HrhxJnO4t2Ph623h","Reason":"ReadError"},"file":"telemetry.go","function":"github.com/algorand/go-algorand/logging.(*telemetryState).logTelemetry","instanceName":"Rv0W1a4Hw84L2i0I","level":"info","line":213,"msg":"/Network/DisconnectPeer","name":":4560","session":"","time":"2020-09-23T15:12:08.101307Z"} {"file":"wsPeer.go","function":"github.com/algorand/go-algorand/network.(*wsPeer).writeLoopSend","level":"warning","line":569,"msg":"peer write error websocket: close sent","name":":4560","time":"2020-09-23T15:12:08.101338Z"} Host:08cbd266-a955-4cc0-b0cc-7188fb83d697:R3 @timestamp:September 23rd 2020, 11:12:08.101 Data.file:wsPeer.go`

`http://telemetry.cicd-performance.algodev.network:5601/app/kibana#/discover?_g=(refreshInterval:(pause:!t,value:0),time:(from:now-4h,mode:quick,to:now))&_a=(columns:!(_source),index:'*',interval:auto,query:(language:lucene,query:error),sort:!('@timestamp',desc))`

it is a warning, but includes "error" in the message

Tsachi Herman  11:16 AM
let’s open an issue for that one, and place it in a lower priority?
I’m pretty much convinced it’s a non-issue, but we should track all the issues.
just please add a note that we’ve had it for a long while and it’s not known to cause any concrete issues.

PR #1549, opened 2020-09-23
remove expected ReadError events

tsachiherman:
## Summary

Avoid ReadError event & log entry in case the other side gracefully closed the websocket.
This is not an error, and should not be marked as such. Instead, we'll be using a new reason called 'DisconnectRequest'.

## Test Plan

Unit test added

## Linked Issues

This PR would provide a partial fix for #1543. Additional changes would be required.


brianolson (approved):

tsachiherman:
thanks @brianolson !

PR #1560, opened 2020-09-29
more precise CompactCertWeightThreshold

zeldovich:
Two related changes to how CompactCertWeightThreshold works:

1. For precision, treat it as a fraction of (1/2^32)ths, instead of a
   fraction of (1/100)ths.  This also makes the integer math easier to
   implement for other verifiers.

2. The decaying threshold of acceptable cert weights now goes from full
   (all signatures required) to CompactCertWeightThreshold, instead of
   going down to 0.  The 0 was meaningless anyway, because if a compact
   cert has less than CompactCertWeightThreshold of signatures, it would
   be rejected elsewhere.

### DIFF
--- a/config/consensus.go
+++ b/config/consensus.go
@@ -326,16 +326,18 @@ type ConsensusParams struct {
 	// accounts and balances) in the critical path.
 	CompactCertVotersLookback uint64
 
-	// CompactCertWeightThreshold is the percentage of top voters weight
+	// CompactCertWeightThreshold specifies the fraction of top voters weight
 	// that must sign the message (block header) for security.  The compact
 	// certificate ensures this threshold holds; however, forming a valid
 	// compact certificate requires a somewhat higher number of signatures,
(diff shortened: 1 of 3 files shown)


tsachiherman:
Keep in mind that once #1507 is merged in, it will be deployed to devnet. So, if you want this to be applied before the other is deployed into mainnet, let me know so that I can merge the two on the same day.

zeldovich:
Ah, thanks for the reminder.  Is devnet running vFuture?

tsachiherman:
> Ah, thanks for the reminder. Is devnet running vFuture?

no. It's running current protocol version.

zeldovich:
Cool -- in that case, should be fine for the top-level PR to get merged without this change, and this can come later, since it will only affect networks running with vFuture (so, hopefully, none).

tsachiherman (approved):
looks good.

PR #1567, opened 2020-09-29
Remove Travis token, which is not needed for open source repos

onetechnical:
## Summary

A Travis token was included for a build status image, but this token is not needed for open source repositories.

## Test Plan

Just ensure build image continues to work for README.md and is removed from buildhost/README.md.

### DIFF
--- a/scripts/buildhost/README.md
+++ b/scripts/buildhost/README.md
@@ -1,5 +1,3 @@
-[![Build Status](https://travis-ci.com/algorand/go-algorand.svg?branch=master)](https://travis-ci.com/algorand/go-algorand)
-
 buildhost
 ====================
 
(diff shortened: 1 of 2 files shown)


scripts/buildhost/README.md
```diff
@@ -1,4 +1,4 @@
-[![Build Status](https://travis-ci.com/algorand/go-algorand.svg?token=25XP72ADqbCQJ3TJVC9S&branch=master)](https://travis-ci.com/algorand/go-algorand)
+[![Build Status](https://travis-ci.com/algorand/go-algorand.svg?branch=master)](https://travis-ci.com/algorand/go-algorand)
```
tsachiherman, line 1: this line is not needed. The build status is irrelevant on this file.
( we don't compile the buildhost directory, so there is no "build status" here )

tsachiherman (changes requested):
it makes sense to fix this on the root `READ.me`, but not on the `./scripts/buildhost/README.md`

tsachiherman (approved):

scripts/buildhost/README.md
onetechnical: That is a very good point. Fixed that and a formatting bug.

tsachiherman:
Last failure was caused due to a bug in one of the unit tests. I just fixed it, so you might want to rebase your branch if this build keep failing.

PR #1570, opened 2020-09-29
Add Console Logger Option

Subumbo:
## Summary

## Test Plan


config/config.go
```diff
@@ -153,6 +153,10 @@ type Local struct {
 	// Valid units are 's' seconds, 'm' minutes, 'h' hours
 	LogArchiveMaxAge string `version[4]:""`
 
+	// Log to console out only.
+	// Used when running node within docker container
+	LogToConsole bool `version[13]:"false"`
```
tsachiherman, line 158: Could you implement this feature in a slightly different way ?
Instead of having the `LogToConsole`, add `LogFileName`. If the LogFileName is "-", then set the output to StdOut. Otherwise, write it to file ( i.e. as is ).

This would address the issue of having a hard-coded named log file, allowing redirection to stdout, just like the goal commands. When doing so, you'll also need to take care of the carpenter so that it would pick the correct file. ( i.e. use the config package to load the config.json file, etc. )

Subumbo: Thank you for your feedback Great idea, will do so tomorrow evening CEST.

test/testdata/configs/config-v13.json
```diff
@@ -0,0 +1,80 @@
...
+    "GossipFanout": 4,
+    "IncomingConnectionsLimit": 10000,
+    "IncomingMessageFilterBucketCount": 5,
+    "IncomingMessageFilterBucketSize": 512,
+    "IsIndexerActive": false,
+    "LedgerSynchronousMode": 2,
+    "LogArchiveMaxAge": "",
+    "LogArchiveName": "node.archive.log",
+    "LogSizeLimit": 1073741824,
+    "LogToConsole": true,
```
tsachiherman, line 51: same here.

installer/config.json.example
```diff
@@ -48,6 +48,7 @@
     "LogArchiveMaxAge": "",
     "LogArchiveName": "node.archive.log",
     "LogSizeLimit": 1073741824,
+    "LogToConsole": false,
```
tsachiherman, line 51: don't forget to update this to `"LogFileName" : "node.log",`

tsachiherman (changes requested):
don't forget to change the LogToConsole -> LogFileName
looks good otherwise.

tsachiherman:
btw - you'll need to fill in the Summary/Test Plan above, and convert this to a non-draft PR.

daemon/algod/server.go
```diff
@@ -78,7 +79,14 @@ func (s *Server) Initialize(cfg config.Local, phonebookAddresses []string, genes
 			maxLogAge = 0
 		}
 	}
-	logWriter := logging.MakeCyclicFileWriter(liveLog, archive, cfg.LogSizeLimit, maxLogAge)
+
+	var logWriter io.Writer
+	if cfg.LogFileName == "-" {
```
Subumbo, line 84: @tsachiherman where is a good place to test this logic? server_test.go, could be a place but I did not find any tests on logging in there. Perhaps higher up the test pyramid?

tsachiherman: I agree that server_test.go isn't the best place. It's more of a UI feature, as can be tested using expect testing.
We already have a test for algod in `./go-algorand/test/e2e-go/cli/algod/expect`. You can take `algodTelemetryLocationTest.exp` as a starting point and try to create another test to verify your change.
Using expect the first time might not be very trivial, but it's getting straight forward once you get used to it.

( as an alternative, I'm also ok if you'll have a go test for that by capturing the output vs. verifying the output goes to node.log file )

tsachiherman:
It has been more than 3 month since you last made change to this PR. If you're not interested pursuing this one anymore, that's fine - I'll just close this one. Otherwise, please update this PR.

tsachiherman:
I'm going to close this PR due to inactivity. If you'd like to revise your work here, please open a separate PR. thanks!

PR #1574, opened 2020-09-30
Fix random failure in goalNodeTest.exp

tsachiherman:
## Summary

Expect test was closing the process connection instead of letting it end on its own. In turn, it lead to incorrect usage of CheckProcessReturnedCode, which resulted in random failure. Solution is to properly wait for process to terminate and examine the return code without calling close.

## Test Plan

Use travis to confirm a good fix.


algonautshant (approved):
Looks good.

PR #1579, opened 2020-10-01
bugfix: fix random failure in goalAppAccountAddressTest.exp

tsachiherman:
## Summary

The test was closing the process after receiving the expected error message. Closing the current process was causing an incorrect error code retrieval. Instead of closing it, we should just leave it alone and wait for the expected eof event.

## Test Plan

This is a test; let travis execute that and confirm it's working.


algonautshant (approved):
LGTM

PR #1580, opened 2020-10-01
Added more deb and rpm package tests

btoll:
## Summary

Added the following tests:

- verify_control_files.sh
- verify_genesis_file.sh
- verify_package_binaries.sh

The tests are separated into two locations:

- ./test/tests/pre for tests that should run before any packages are installed (these test control files, for example)
- ./test/tests/post for tests that should run after packages are installed (basically, everything else)

This prevents getting into potentially messy logic that needs to happen if packages needs to be (re-)installed/removed before any tests are run.

Workflow:

mule -> make -> test.sh (sets up env vars, installs packages using `dpkg` or `rpm`) -> run all tests in `pre/` -> run all tests in `post/`

## Test Plan

`mule -f package-test.yaml package-test`


scripts/release/mule/package/deb/package.sh
```diff
@@ -60,7 +60,12 @@ if [[ ! "$ALGORAND_PACKAGE_NAME" =~ devtools ]]; then
         cp "installer/${data}" "${PKG_ROOT}/var/lib/algorand"
     done
 
-    cp "./installer/genesis/${DEFAULTNETWORK}/genesis.json" "${PKG_ROOT}/var/lib/algorand/genesis.json"
+    genesis_dirs=("devnet" "testnet" "mainnet" "betanet")
+    for dir in "${genesis_dirs[@]}"; do
+        mkdir -p "${PKG_ROOT}/var/lib/algorand/genesis/${dir}"
+        cp "./installer/genesis/${dir}/genesis.json" "${PKG_ROOT}/var/lib/algorand/genesis/${dir}/genesis.json"
+    done
+    cp "./installer/genesis/${DEFAULT_RELEASE_NETWORK}/genesis.json" "${PKG_ROOT}/var/lib/algorand/genesis.json"
```
btoll, line 68: Bug fix.

scripts/release/mule/package/deb/package.sh
```diff
@@ -90,7 +95,7 @@ Dpkg::Options {
 EOF
 
 mkdir -p "${PKG_ROOT}/DEBIAN"
-if [[ "$PKG_NAME" =~ devtools ]]; then
+if [[ "$ALGORAND_PACKAGE_NAME" =~ devtools ]]; then
```
btoll, line 98: Bug fix.

scripts/release/mule/package/deb/package.sh
```diff
@@ -103,6 +108,7 @@ for ctl_file in $(ls "${CTL_FILES_DIR}"); do
     < "${CTL_FILES_DIR}/${ctl_file}" \
       sed -e "s,@ARCH@,${ARCH}," \
           -e "s,@VER@,${VER}," \
+          -e "s,@PKG_NAME@,$ALGORAND_PACKAGE_NAME," \
```
btoll, line 111: Bug fix.

scripts/release/mule/package/rpm/package.sh
```diff
@@ -39,10 +39,10 @@ else
 fi
 trap 'rm -rf $TEMPDIR' 0
 < "./installer/rpm/$INSTALLER_DIR/$INSTALLER_DIR.spec" \
-    sed -e "s,@ALGORAND_PACKAGE_NAME@,$REQUIRED_ALGORAND_PACKAGE," \
+    sed -e "s,@PKG_NAME@,$ALGORAND_PACKAGE_NAME," \
         -e "s,@VER@,$FULLVERSION," \
         -e "s,@ARCH@,$ARCH," \
-        -e "s,@REQUIRED_ALGORAND_PKG@,$ALGORAND_PACKAGE_NAME," \
+        -e "s,@REQUIRED_ALGORAND_PKG@,$REQUIRED_ALGORAND_PACKAGE," \
```
btoll, line 45: Bug fixes.

scripts/release/mule/test/test.sh
```diff
@@ -10,35 +11,52 @@ export ARCH_TYPE
...
-export SHA
+pushd "$PKG_DIR"
 
 if ! $USE_CACHE
 then
-    mule -f package-test.yaml "package-test-setup-$PKG_TYPE"
+    SRC_DIR="s3://algorand-staging/releases/$CHANNEL/$VERSION"
+
+    # deb
+    aws s3 cp "$SRC_DIR/algorand_${CHANNEL}_${OS_TYPE}-${ARCH_TYPE}_${VERSION}.deb" .
```
egieseke, line 29: should these copy commands be conditional based on the OS_TYPE and ARCH_TYPE?

scripts/release/mule/test/tests/post/verify_genesis_file.sh
```diff
@@ -0,0 +1,27 @@
...
+
+if [ ! -f /var/lib/algorand/genesis.json ]
+then
+    echo "[$0] The genesis file is not present."
+    exit 1
+fi
+
+if [ "$NETWORK" != "$(jq -r '.network' /var/lib/algorand/genesis.json)" ]
+then
+    echo "[$0] The network \`$NETWORK\` set in \`genesis.json\` is incorrect."
```
egieseke, line 22: This message is confusing.  Please update this message to include the expected and actual network values.  Consider showing the full path of the genesis.json file.

scripts/release/mule/test/tests/post/verify_genesis_file.sh
```diff
@@ -0,0 +1,27 @@
...
+    exit 1
+fi
+
+echo "[$0] The network \`$NETWORK\` set in \`genesis.json\` is correct."
```
egieseke, line 26: Consider rephrasing to "The network set in genesis.json is correctly set to $NETWORK".  Show full path to genesis.json file.

scripts/release/mule/test/test.sh
```diff
@@ -10,35 +11,52 @@ export ARCH_TYPE
...
 else
-    ./scripts/release/mule/test/util/test_package.sh
+    yum install yum-cron -y
+    rpm -i algorand-"$VERSION"-1."$ARCH_BIT".rpm
+    rpm -i algorand-devtools-"$VERSION"-1."$ARCH_BIT".rpm
 fi
 
+popd
+
+for test in $(ls ./scripts/release/mule/test/tests/post/*.sh)
```
egieseke, line 58: Check for "\*test.sh" rather than "*.sh" to help prevent inadvertently breaking when an unrelated utility script is added to the directory.

scripts/release/mule/test/tests/post/verify_package_binaries.sh
```diff
@@ -0,0 +1,79 @@
...
+then
+    dpkg -L algorand > "$RPMTMP/algorand.install"
+    dpkg -L algorand-devtools > "$RPMTMP/algorand-devtools.install"
+else
+    rpm -ql algorand > "$RPMTMP/algorand.install"
+    rpm -ql algorand-devtools > "$RPMTMP/algorand-devtools.install"
+fi
+
+ALGORAND_BINS=(
+    /usr/bin/algocfg
```
egieseke, line 23: Can the bin dir be factored out?

scripts/release/mule/test/test.sh
```diff
@@ -10,35 +11,52 @@ export ARCH_TYPE
...
+pushd "$PKG_DIR"
+
+if [ "$PKG_TYPE" = deb ]
 then
-    ./scripts/release/mule/test/tests/run_tests -b "$BRANCH" -c "$CHANNEL" -h "$SHA" -r "$VERSION"
+    dpkg -i algorand_*"$VERSION"*.deb
+    dpkg -i algorand-devtools*"$VERSION"*.deb
 else
-    ./scripts/release/mule/test/util/test_package.sh
+    yum install yum-cron -y
```
onetechnical, line 51: Is this required because we're installing from package rather than repository?

scripts/release/mule/test/tests/post/verify_package_binaries.sh
```diff
@@ -0,0 +1,79 @@
...
+if [ "$PKG_TYPE" = deb ]
+then
+    dpkg -L algorand > "$RPMTMP/algorand.install"
+    dpkg -L algorand-devtools > "$RPMTMP/algorand-devtools.install"
+else
+    rpm -ql algorand > "$RPMTMP/algorand.install"
+    rpm -ql algorand-devtools > "$RPMTMP/algorand-devtools.install"
+fi
+
+ALGORAND_BINS=(
```
onetechnical, line 22: This is okay, but I'd prefer if we could dynamically fetch this list from the script doing the packaging so we don't have to update in multiple places. Maybe over-optimizing, though.

onetechnical (approved):
I think this is okay, had a couple questions, though.

scripts/release/mule/test/test.sh
btoll: I'm not sure I understand the question.  The naming convention for the packages are pretty firmly set, and if there are different os types and arch types this should just work.

scripts/release/mule/test/test.sh
btoll: Yes.

scripts/release/mule/test/test.sh
btoll: The point of the `post/` and `pre/` directories are to run tests.  There shouldn't be any scripts added to these dirs that aren't tests.  It was designed this way.

scripts/release/mule/test/tests/post/verify_package_binaries.sh
btoll: Yes, I called this out during the review, I agree it is a brittle area and should be hardened.  My preference is to commit as-is and come up with a design later.

scripts/release/mule/test/tests/post/verify_package_binaries.sh
btoll: Yes, but why?  It's better if we match the full path as stated in the control files, in my opinion.

scripts/release/mule/test/tests/post/verify_genesis_file.sh
btoll: I'll update to show the full path.

scripts/release/mule/test/tests/post/verify_genesis_file.sh
btoll: I'll update to show the full path.

scripts/release/mule/test/test.sh
btoll: I just added a comment to clarify.

onetechnical (approved):

scripts/release/mule/test/test.sh
```diff
@@ -10,35 +11,55 @@ export ARCH_TYPE
...
+pushd "$PKG_DIR"
+
+if [ "$PKG_TYPE" = deb ]
 then
-    ./scripts/release/mule/test/tests/run_tests -b "$BRANCH" -c "$CHANNEL" -h "$SHA" -r "$VERSION"
+    dpkg -i algorand_*"$VERSION"*.deb
+    dpkg -i algorand-devtools*"$VERSION"*.deb
 else
-    ./scripts/release/mule/test/util/test_package.sh
+    # We need to install this since it's not being installed by a package manger.
```
onetechnical, line 51:
```suggestion
    # We need to install this since it's not being installed by a package manager.
```

onetechnical (approved):

scripts/release/mule/test/tests/post/verify_package_binaries.sh
egieseke: Seems redundant to specify /usr/bin/ for each binary rather than just the binary name.

scripts/release/mule/test/test.sh
egieseke: Sometimes test files share utility scripts to factor out common code, so good to have a way to distinguish.

btoll: Then the shared scripts should go somewhere else.  These dirs are designed to only hold tests.

scripts/release/mule/test/tests/post/verify_package_binaries.sh
btoll: What if there's a false positive for `/etc/algocfg`?  The binaries should be in a specific location prescribed by the control files, which is what we should match.

scripts/release/mule/test/test.sh
btoll: Already fixed.

scripts/release/mule/test/tests/post/verify_package_binaries.sh
egieseke: Agree, just saying that the /usr/bin could be appended in the for loop rather than repeated for each binary.

btoll: Can we just move this along?  I'd really like to get unblocked.

scripts/release/mule/test/test.sh
egieseke: Are deb and rpm packaging appropriate for all combinations of OS and ARCH?

egieseke (approved):

PR #1581, opened 2020-10-01
Coverage tests for ledger/acctupdates

algonautshant:
Coverage tests for IsWritingCatchpointFile and GetCatchpointStream

### DIFF
--- a/ledger/acctupdates_test.go
+++ b/ledger/acctupdates_test.go
@@ -21,7 +21,9 @@ import (
 	"context"
 	"database/sql"
 	"fmt"
+	"io/ioutil"
 	"os"
+	"path/filepath"
 	"runtime"
 	"sync"
 	"testing"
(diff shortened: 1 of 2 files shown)


ledger/acctupdates_test.go
```diff
@@ -1030,3 +1030,92 @@ func TestListCreatables(t *testing.T) {
 		expectedDbImage, numElementsPerSegement)
 	listAndCompareComb(t, au, expectedDbImage)
 }
+
+func TestIsWritingCatchpointFile(t *testing.T) {
```
tsachiherman, line 1034: I think that this test is not very meaningful, but it's ok to have it for now.
( i.e. it's not wrong.. just overly dependent in the internal implementation )

ledger/acctupdates_test.go
```diff
@@ -1030,3 +1030,92 @@ func TestListCreatables(t *testing.T) {
...
+	au.initialize(conf, ".", proto, accts[0])
+	defer au.close()
+
+	err := au.loadFromDisk(ml)
+	require.NoError(t, err)
+
+	filesToCreate := 4
+
+	// Prepare the directory as GetCatchpointStream expects
+	err = os.Mkdir("catchpoints", 0777)
```
tsachiherman, line 1068: Create this directory under the temp directory, and make sure to defer the deletion of that directory right after using RemoveAll. See `TestBasicCatchpointWriter` as an example.

ledger/acctupdates_test.go
```diff
@@ -1030,3 +1030,92 @@ func TestListCreatables(t *testing.T) {
...
+	require.NoError(t, err)
+
+	// Create the catchpoint files with dummy data
+	for i := 0; i < filesToCreate; i++ {
+		f, err := os.Create(fmt.Sprintf("catchpoints/%d.catchpoint", i))
```
tsachiherman, line 1073: I think that `ioutil.WriteFile` could help you do this section much simpler, but there is nothing incorrect here.

ledger/acctupdates_test.go
```diff
@@ -1030,3 +1030,92 @@ func TestListCreatables(t *testing.T) {
...
+	for i := 0; i < filesToCreate; i++ {
+		err := au.accountsq.storeCatchpoint(context.Background(), basics.Round(i), fmt.Sprintf("catchpoints/%d.catchpoint", i), "", 0)
+		require.NoError(t, err)
+	}
+
+	dataRead := make([]byte, 3)
+	var n int
+
+	// File on disk, and database has the record
+	reader, err := au.GetCatchpointStream(1)
```
tsachiherman, line 1094: pass in basics.Round(1) instead of just 1.

tsachiherman (changes requested):
Few small requests, overall structure is good.

ledger/acctupdates_test.go
```diff
@@ -1030,3 +1032,89 @@ func TestListCreatables(t *testing.T) {
...
+	}()
+	catchpointsDirectory := filepath.Join(temporaryDirectroy, "catchpoints")
+	err = os.Mkdir(catchpointsDirectory, 0777)
+	require.NoError(t, err)
+
+	au.dbDirectory = temporaryDirectroy
+
+	// Create the catchpoint files with dummy data
+	for i := 0; i < filesToCreate; i++ {
+		fileName := filepath.Join("catchpoints", fmt.Sprintf("%d.catchpoint", i))
```
tsachiherman, line 1082: join with `catchpointsDirectory` here to avoid the subsequent `filepath.Join(temporaryDirectroy, fileName)`

algonautshant: I will need fileName without temoraryDirectory on line 1088.
The reason why I need this arrangement is to match the filename in GetCatchpointStream:
```
	// if the database doesn't know about that round, see if we have that file anyway:
	fileName := filepath.Join("catchpoints", catchpointRoundToPath(round))
	catchpointPath := filepath.Join(au.dbDirectory, fileName)

```

tsachiherman (approved):

PR #1583, opened 2020-10-02
Add goal account info command

### LINKED ISSUE #1487: Goal command `goal account balance` only display the balance of ALGO
Please find the code below:
https://github.com/algorand/go-algorand/blob/8fd86a225714e10e014828e2f4037d6e521194ed/cmd/goal/account.go#L523-L538

It is supposed to display the balance of both ALGO and assets, while here it only outputs the balance of ALGO.
One of our developers found this problem when he was trying to use this in his product.

jasonpaulos:
## Summary

The description of `goal account balance` says it displays balances for algos and assets, but actually only shows algos. I have corrected the description and added a new command that shows information about an account's assets and applications, `goal account info`.

I have also removed the asset information that `goal account list` currently reports. Instead of having asset balances always listed, I have added the flag `--info` to `goal account list` to generate this same report as `goal account info` for each account in a wallet.

Examples:

```
$ goal account info -a 5K6OH2HKK3HORMIAQT7JYCUUYJTWS5KQ4AL7ITPI2SOCBKA7LMEEJYT5LA -d data
Created Assets:
        ID 2678221, A-Coin, supply 1000 AC
        ID 2678222, <unnamed>, supply 1000 BC
        ID 2678223, C-Coin, supply 1000 units
        ID 2678224, D-Coin, supply 10.00 DC
Held Assets:
        ID 2678221, A-Coin, balance 1000 AC
        ID 2678222, <unnamed>, balance 1000 BC
        ID 2678223, C-Coin, balance 1000 units
        ID 2678224, D-Coin, balance 10.00 DC (frozen)
Created Apps:
        ID 2685370, global state used 0/1 uints, 1/1 byte slices
        ID 2685371, global state used 0/0 uints, 0/0 byte slices
Opted In Apps:
        ID 2685371, local state used 0/1 uints, 1/2 byte slices
```

```
$ goal account list -d data
[offline]       Unnamed-0       5K6OH2HKK3HORMIAQT7JYCUUYJTWS5KQ4AL7ITPI2SOCBKA7LMEEJYT5LA      99993000 microAlgos     [created asset IDs: 2678223 (1000 ), 2678224 (1000 DC), 2678221 (1000 AC), 2678222 (1000 BC)]      [created app IDs: 2685370, 2685371]     [opted in app IDs: 2685371]     *Default
```

```
$ goal account list --info -d data
[offline]       Unnamed-0       5K6OH2HKK3HORMIAQT7JYCUUYJTWS5KQ4AL7ITPI2SOCBKA7LMEEJYT5LA      99993000 microAlgos     [created asset IDs: 2678221 (1000 AC), 2678222 (1000 BC), 2678223 (1000 ), 2678224 (1000 DC)]      [created app IDs: 2685370, 2685371]     [opted in app IDs: 2685371]     *Default
Created Assets:
        ID 2678221, A-Coin, supply 1000 AC
        ID 2678222, <unnamed>, supply 1000 BC
        ID 2678223, C-Coin, supply 1000 units
        ID 2678224, D-Coin, supply 10.00 DC
Held Assets:
        ID 2678221, A-Coin, balance 1000 AC
        ID 2678222, <unnamed>, balance 1000 BC
        ID 2678223, C-Coin, balance 1000 units
        ID 2678224, D-Coin, balance 10.00 DC (frozen)
Created Apps:
        ID 2685370, global state used 0/1 uints, 1/1 byte slices
        ID 2685371, global state used 0/0 uints, 0/0 byte slices
Opted In Apps:
        ID 2685371, local state used 0/1 uints, 1/2 byte slices
```

Closes #1487.

## Test Plan

I have added a goal expect test to verify the command output is correct.

### DIFF
--- a/cmd/goal/account.go
+++ b/cmd/goal/account.go
@@ -31,6 +31,7 @@ import (
 	"github.com/algorand/go-algorand/config"
 	"github.com/algorand/go-algorand/crypto"
 	"github.com/algorand/go-algorand/crypto/passphrase"
+	"github.com/algorand/go-algorand/daemon/algod/api/spec/v1"
 	algodAcct "github.com/algorand/go-algorand/data/account"
 	"github.com/algorand/go-algorand/data/basics"
 	"github.com/algorand/go-algorand/data/transactions"
@@ -61,6 +62,7 @@ var (
 	importDefault      bool
 	mnemonic           string
 	dumpOutFile        string
+	listAccountInfo    bool
 )
 
 func init() {
@@ -68,6 +70,7 @@ func init() {
 	accountCmd.AddCommand(deleteCmd)
 	accountCmd.AddCommand(listCmd)
 	accountCmd.AddCommand(renameCmd)
+	accountCmd.AddCommand(infoCmd)
 	accountCmd.AddCommand(balanceCmd)
 	accountCmd.AddCommand(rewardsCmd)
 	accountCmd.AddCommand(changeOnlineCmd)
@@ -115,6 +118,13 @@ func init() {
 	// Lookup info for multisig account flag
 	infoMultisigCmd.Flags().StringVarP(&accountAddress, "address", "a", "", "Address of multisig account to look up")
 	infoMultisigCmd.MarkFlagRequired("address")
+
+	// Account list flags
+	listCmd.Flags().BoolVar(&listAccountInfo, "info", false, "Include additional information about each account's assets and applications")
+
+	// Info flags
+	infoCmd.Flags().StringVarP(&accountAddress, "address", "a", "", "Account address to look up (required)")
+	infoCmd.MarkFlagRequired("address")
 
 	// Balance flags
 	balanceCmd.Flags().StringVarP(&accountAddress, "address", "a", "", "Account address to retrieve balance (required)")
@@ -440,7 +450,7 @@ var infoMultisigCmd = &cobra.Command{
 var listCmd = &cobra.Command{
 	Use:   "list",
 	Short: "Show the list of Algorand accounts on this machine",
-	Long:  `Show the list of Algorand accounts on this machine. Indicates whether the account is [offline] or [online], and if the account is the default account for goal. Also displays balances and asset information.`,
+	Long:  `Show the list of Algorand accounts on this machine. Indicates whether the account is [offline] or [online], and if the account is the default account for goal. Also displays account information with --info.`,
 	Args:  validateNoPosArgsFn,
 	Run: func(cmd *cobra.Command, args []string) {
 		dataDir := ensureSingleDataDir()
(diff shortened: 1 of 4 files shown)


EvanJRichard (approved):
Looking good to me, :ship: pending tests passing!

tsachiherman:
Could you please confirm that with @rotemh ? I'm a bit habitant about this one since it was kicked off by a third-party request and did no come as internal design. Specifically, it breaks the existing goal interface, and seems to be missing the applications.

When I originally read that ticket, I was thinking of changing the `Long` string so it would align with the current implementation.

algorotem:
@tsachiherman are we worried about someone who uses goal's output as an input to another program? If this is the case, In my view, it's ok to change it because goal was never defined (nor designed) to be a dev tool in that sense.

With that, I do think that we need to keep the "feel" consistent across commands and this one should be closer to `account list`. 

Per the Applications, this is the `balance` command and feels "incorrect" to me to have the applications their as well. Maybe the better solution is to deprecate this one and create a new `goal account info` that returns all balances and applications?

tsachiherman:
Yes, your plan sounds much better:
1. remove asset information from `goal account list`
2. keep `goal account balance` as is, updating the `Long` to indicate that it won't be providing the asset information
3. adding `goal account info` which would be able to provide assets and app information for any specified account.

one thing to note is that it would be nice if we can use the output of #1 as the input to #3. Alternatively, add a flag to #1 so that it would add the account information.

jasonpaulos:
@tsachiherman I've updated this PR with the above changes. Let me know if it looks good to you.

tsachiherman (changes requested):
@jasonpaulos, great work implementing these, I have few more (smaller) requests related to the code changes - 
Calling `reportErrorf` in `printAccountInfo` is correct, but would result in weird output, as only some of the account information would be presented.
( i.e. calling `reportErrorf` would terminate the process )
I have few suggestions that could improve this use case:
1. Accumulate the output string, and output it with a single `reportInfof` ( which also better align with separating the formatting and outputting )
2. Promote the error case to the top, so you could identify the error case first
3. The error output of "Error processing command: %s" is a pretty generic; I think that we can do better in this case. ( i.e. "Unable to retrieve asset information for asset %d referred by account %s : %v" )

jasonpaulos:
@tsachiherman nice catch, I forgot to remove that `reportErrorf`. I made changes to buffer the response and print any errors first. However, I chose not to use `reportInfof` for the time being because `unicodePrintable` thinks tabs (`\t`) are non-printable characters.

Let me know if you think this is acceptable for now, I should switch to using spaces for the indent, or I should modify `unicodePrintable` to allow tabs.

tsachiherman:
> @tsachiherman nice catch, I forgot to remove that `reportErrorf`. I made changes to buffer the response and print any errors first. However, I chose not to use `reportInfof` for the time being because `unicodePrintable` thinks tabs (`\t`) are non-printable characters.
> 
> Let me know if you think this is acceptable for now, I should switch to using spaces for the indent, or I should modify `unicodePrintable` to allow tabs.

I think that it's better, but we definitely need to have the errors sent to the stderr, plus exiting with a non-zero error code.
( i.e. that's why I suggested using the reportXXX functions )

tsachiherman (approved):
looks good; it's more common in go to return an error object compared to returning a success flag, but the code looks like it would work just fine.

PR #1584, opened 2020-10-02
bugfix: correct accounting of transactions in checkPendingQueueSize

tsachiherman:
## Summary

The checkPendingQueueSize function is called per new transaction group that is being added to the transaction pool. Unfortunately, it tests whether a single entry can be added to the transaction pool, whereas it should have checked wether all the transaction included in the incoming transaction group would be able to fit into the transaction pool.

## Test Plan

Unit test added.


algonautshant (approved):
Looks good.

`func (pool *TransactionPool) Test` is not tested. 
Can you use it in the test to get that also covered?

tsachiherman:
> Looks good.
> 
> `func (pool *TransactionPool) Test` is not tested.
> Can you use it in the test to get that also covered?

Will do in a subsequant pr.

PR #1585, opened 2020-10-02
avoid printing control characters from goal

### LINKED ISSUE #1569: [Bugcrowd] A malicious transaction file can control all output of `goal clerk inspect` (different vector)
[PDF of Bug description](https://app.zenhub.com/files/191266671/30b609c6-5b75-4a65-8939-5dc6baa070fb/download)

tsachiherman:
## Summary

Goal occasionally print out messages (including error messages) that are based on its inputs. If any of these inputs contains control characters, these control characters could get outputted to the stdout/stderr. This PR ensures that we don't output control characters to the standard out/err when called using the reportInfo/reportWarn/reportError variations.

( it's ok to emit control characters that we intend to; but we don't want to output these unintentionally )

## Test Plan

Added a unit test. Resolves #1569


cmd/goal/commands.go
```diff
@@ -492,36 +492,50 @@ func ensurePassword() []byte {
...
-	fmt.Printf(format+"\n", args...)
-	// log.Infof(format, args...)
+	reportInfoln(fmt.Sprintf(format, args...))
 }
 
 func reportWarnln(args ...interface{}) {
 	fmt.Print("Warning: ")
-	fmt.Println(args...)
-	// log.Warnln(args...)
+
```
nicholasguoalgorand, line 510: Why not just invoke `reportInfoln` here?

tsachiherman: I didn't want to mixup the info level handlers and the warning level handlers.

nicholasguoalgorand (approved):
made one small comment, otherwise looks good to me

PR #1600, opened 2020-10-06
Adding support for KMD wallet password supplied as environment variable

gislik:
## Summary

Repeated usage of `goal` requires a password to be entered each time the wallet needs to be unlocked which introduces a friction for the user. This pull request allows the user to provide the wallet password in the environment variable `$ALGORAND_KMD_PASSWORD`.

## Security

Exposing passwords using environment variables is generally [safe](https://security.stackexchange.com/questions/14000/environment-variable-accessibility-in-linux/14009#14009) since the environment of a process is only available to the user (euid) running the process.

## Test Plan

Unit tests were written to account for three different use cases:

1. User provides correct password in the environment variable (should succeed)
2. User provides incorrect password in the environment variable (should fail)
3. User does not provide password in the environment variable (should exit with status code 1)

~~~~sh
$ go test ./cmd/goal -run 'TestEnsurePasswordForWallet.*'
ok      github.com/algorand/go-algorand/cmd/goal
~~~~

No integration tests are required at this stage but if this pull request is merged it can simplify other unit tests so that they don't have to depend on `expect`.

### DIFF
--- a/cmd/diagcfg/messages.go
+++ b/cmd/diagcfg/messages.go
@@ -21,11 +21,11 @@ const (
 	loggingNotEnabled    = "Remote logging is currently disabled"
 	loggingEnabled       = "Remote logging is enabled.  Node = %s, Guid = %s\n"
 
-	metricNoConfig                          = "Unable to load configuration file : %s\n"
-	metricConfigReadingFailed               = "Failed to read configuration file : %s\n"
-	metricReportingStatus                   = "Metric reporting is %s\n"
-	metricSaveConfigFailed                  = "Metric configuration file could not be saved : %s\n"
-	metricDataDirectoryEmpty                = "no data directory was specified. Please use either -d or set environment variable ALGORAND_DATA"
+	metricNoConfig            = "Unable to load configuration file : %s\n"
+	metricConfigReadingFailed = "Failed to read configuration file : %s\n"
+	metricReportingStatus     = "Metric reporting is %s\n"
+	metricSaveConfigFailed    = "Metric configuration file could not be saved : %s\n"
+	metricDataDirectoryEmpty  = "no data directory was specified. Please use either -d or set environment variable ALGORAND_DATA"
 
 	telemetryConfigReadError = "Could not read telemetry config: %s\n"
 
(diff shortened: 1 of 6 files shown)


tsachiherman (changes requested):
@gislik, I’m going to block this PR since using environment variables, like in your change, doesn’t scale up.

I have several concerns around using environment variable for this purpose:
1. User might run multiple network instance and/or wallets on a single machine. Each would need its own password.
2. Storing the password in the environment might be a security concern, since other processes might have access to that information.
goal already supports password “caching”. So you could send the password once, and it will be available on all subsequent calls. ( in the code it’s called Wallet Handle )
3. Generally speaking, I’d like to avoid going onto a path where the end user opts into an implementation where the password is available to anyone other than the desired process.

For these reasons, we avoided implementing the password passed in as a command line parameter. You can use the terminal redirection, such as 
expect -c ‘spawn goal account -d . list -w wallet2; expect {Please enter the password for wallet*} {send “1234\r”; exp_continue} eof {};’
( i.e. use the password 1234 with wallet2 )

Passing the password is also available on the SDKs as well, allowing you to tunnel end-user credentials directly to KMD.

If you don’t care about passwords, you can always create a password-free wallet.

gislik:
Roger that @tsachiherman.

Thank you for a very detailed reply and I understand your reasoning.

When using `goal` the caching did not seem to kick in for me. Passwords copied from a password manager on a mac are flushed from the clipboard with time. The clipboard also gets filled with all sorts of data like addresses so managing passwords between invocations of `goal` become painful. 

Is it possible that the cache is time-bound, and if so is there a way to configure the duration? That would alleviate the pain.

tsachiherman:
@gislik , the kmd process is the one that caches the underlaying password using a wallet handle mechanism. The kmd by default remains in memory for only 60 seconds before going away.
You can either change the time the kmd would remain in memory by altering the kmd.config file, or start the kmd via goal providing a higher timeout value.

I found two references that supports that in the documentation:
1. https://developer.algorand.org/docs/reference/node/config/
2. https://developer.algorand.org/docs/features/accounts/create/

gislik:
Thank you for the reference. This will help.

tsachiherman:
No, but thanks!

PR #1601, opened 2020-10-06
Fix bug in TestGetBlockHTTP and TestGetBlockWS

tsachiherman:
## Summary

Fix uncovered bug in buildTestHTTPPeerSource implementation.
The following seems to create an issue for go 1.15.X, whereas it passes on go 1.14.X:
```
func buildTestHTTPPeerSource(rootURLs ...string) *httpTestPeerSource {
	for url := range rootURLs {
		...
	}
}
buildTestHTTPPeerSource()
```

In this case, the affect is no-op : since the function parameter list is always empty, the looping never executed. The call to testHTTPPeer(url) is bogus, but never noted since it's not being used.

## Test Plan

This is a unit test. Unit test was executed and working correctly.

### DIFF
--- a/catchup/fetcher_test.go
+++ b/catchup/fetcher_test.go
@@ -567,7 +567,7 @@ func (p *testHTTPPeer) GetHTTPPeer() network.HTTPPeer {
 
 func buildTestHTTPPeerSource(rootURLs ...string) *httpTestPeerSource {
 	peers := []network.Peer{}
-	for url := range rootURLs {
+	for _, url := range rootURLs {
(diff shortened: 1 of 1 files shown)


brianolson:
I think I can do one better, see #1602

jannotti (approved):
This clearly the right fix if we intend to use this function, though Brian's deletion of the entire thing seems right if we're only going to call the function with no args.  I'll bow out of the decision about which is right until I know more.  I don't know what the original test goal was.

PR #1602, opened 2020-10-06
cleanup buggy dead code

brianolson:
Remove dead test code that had buggy statements in it (that were unexecuted).

### DIFF
--- a/catchup/fetcher_test.go
+++ b/catchup/fetcher_test.go
@@ -565,14 +565,6 @@ func (p *testHTTPPeer) GetHTTPPeer() network.HTTPPeer {
 	return p
 }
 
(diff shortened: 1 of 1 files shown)


jannotti (approved):
I guess this is right unless we can see what peers it was originally trying to accumulate and test against.  This seems like a pretty empty test that looks like it wasn't *intended* to be so empty.

tsachiherman (approved):
I generally don't like the removal of the functionality. however, I have no proof that this functionality even work as intended ( since it was clearly buggy, and currently unused ).

If it was a product code, I would spend some more time here and would ensure we have an empty constructor; but given that this is a test code, and this struct is used only in this file, I'll take your solution.

PR #1603, opened 2020-10-06
make sanity

brianolson:
This is just the formatting changes resulting from running `make sanity`
After this it should be safe for someone to run `make sanity` and not be burdened with irrelevant changes to their changes.


tsachiherman (approved):
looks ok.

PR #1604, opened 2020-10-07
Simplify mule signing task

btoll:
## Summary

Simplify package signing by only having one task that signs all build artifacts instead of delineating them by type.

Note that there is some churn in `scripts/release/mule/package/deb/package.sh` because of the removal of the (unnecessary) brackets around the env vars, but it makes the script more readable (in my opinion).

## Test Plan

`mule -f package-sign.yaml package-sign`

### DIFF
--- a/scripts/release/mule/package/deb/package.sh
+++ b/scripts/release/mule/package/deb/package.sh
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
-# shellcheck disable=2038,2045,2064
+# shellcheck disable=2038,2045,2064,2129
 
 set -ex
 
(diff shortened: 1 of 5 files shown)


scripts/release/mule/package/deb/package.sh
```diff
@@ -125,11 +126,46 @@ Copyright: Algorand developers <dev@algorand.com>
...
+cat << EOF >> "$STATUSFILE"
+
+dpkg-l:
+EOF
+
+dpkg -l >> "$STATUSFILE"
+
+popd
+
+############################################################
```
btoll, line 168: Here we need to capture the details of the build env for the `build_status` file.

bricerisingalgorand (approved):

scripts/release/mule/sign/sign.sh
```diff
@@ -16,75 +15,47 @@ BRANCH=${BRANCH:-$(git rev-parse --abbrev-ref HEAD)}
...
+done
 
-    for item in *"$VERSION"*."$1"
-    do
-        gpg -u "$SIGNING_KEY_ADDR" --detach-sign "$item"
-    done
-}
+for file in *.rpm
+do
+    gpg -u rpm@algorand.com --detach-sign "$file"
```
onetechnical, line 45: I find it weird that we parameterize `$SIGNING_KEY_ADDR` but then hardcode `rpm@algorand.com`. Maybe do both, and we can do "test" signing, or sign off on nightlies with a different key?

onetechnical (changes requested):

scripts/release/mule/sign/sign.sh
btoll: The reason is because the only the rpm packages use this user to sign and everything else uses the dev@ user.  We could stick that in a variable, but it's only used once...

Yes, we'll need to sign nightlies with a different key, but I'm not sure how that will affect these scripts yet.

onetechnical: We don't allow overriding `$SIGNING_KEY_ADDR` right now, so I'll grudgingly approve it, but this is probably better variabled out, because I can almost guarantee we'll want to generalize it for nightly / bespoke builds.

onetechnical (approved):

scripts/release/mule/sign/sign.sh
btoll: Yes, I agree, I'd just rather not do that before we actually tackle that.

PR #1605, opened 2020-10-07
Correct Telemetry uploading message reported in node.log

algonautshant:
## Summary

Added GetTelemetryRemoteEnabled to check if telemetry is enabled to log remotely (i.e. logging.config Enable param).

Before this change, GetTelemetryEnabled was checked to report "Telemetry Enabled".
This may be true, even if telemetry is not sending remotely. 
GetTelemetryEnabled returns true if logging.config params Enable or SendToLog is true, or if config.json param TelemetryToLog is true.

## Test Plan

This may be tested with an e2e expect test. Not yet added.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -269,7 +269,7 @@ type Local struct {
 	EnableProfiler bool `version[0]:"false"`
 
 	// TelemetryToLog records messages to node.log that are normally sent to remote event monitoring
-	TelemetryToLog bool `version[5]:"true"`
+	TelemetryToLog bool `"true"`
 
 	// DNSSecurityFlags instructs algod validating DNS responses.
 	// Possible fla values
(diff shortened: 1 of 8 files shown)


config/config.go
```diff
@@ -269,7 +270,7 @@ type Local struct {
 	EnableProfiler bool `version[0]:"false"`
 
 	// TelemetryToLog records messages to node.log that are normally sent to remote event monitoring
-	TelemetryToLog bool `version[5]:"true"`
```
algorandskiy, line 272: we should not modify versions... they denote config version the param first appeared and its default value

algonautshant: reverted.

logging/log.go
```diff
@@ -400,6 +404,12 @@ func (l logger) GetTelemetryURI() string {
 	return telemetryConfig.URI
 }
 
+// GetTelemetryRemoteEnabled returns true if remote logging is enabled
+// This is decided by Enable parameter in logging.config
+func (l logger) GetTelemetryRemoteEnabled() bool {
```
tsachiherman, line 409: It's not about remote logging, it's purely about sending events ( and maybe a subset of the log messages ) to the telemetry server. A better name could be "GetTelemetryUploadingEnabled"

tsachiherman:
Please also update the PR title to reflect its content.

tsachiherman (approved):

Issue #1432, opened 2020-08-18
Bug: node reports `AssembleBlock: assembled block round does not match: XXX != YYY"`


tsachiherman:
Research "pseudonode.makeProposals: could not generate a proposal for round 7330065: AssembleBlock: assembled block round does not match: 7330064 != 7330065"

PR #1611, opened 2020-10-08
Refactor AssembleBlock error

nicholasguoalgorand:
## Summary

https://github.com/algorand/go-algorand/issues/1432

AssembleBlock would produce an error during catchup that does not interfere with the node. Bumping the error down to an warn message.

## Test Plan

Added unit test for the change


tsachiherman (changes requested):
Two requests:

1. The new case would need to have the following, right ?
```
stats.StopReason = telemetryspec.AssembleBlockTimeout
```

2. Please modify the unit test to verify that the warning message is being printed out.

tsachiherman (changes requested):
The `callbackLogger` that you've used is fine.. but the test isn't sufficiently deterministic :
1. You want to ensure that you get the warning message exactly once. ( and it's the correct one )
2. You want to ensure that you get no error messages. 

Note that for tests, there is a wrapper for the logging called logging.TestingLog; it tends to be quite noisy, so I would recommend adjusting the log level on that one to Info.
( this doesn't have to be part of this PR; just an fyi )

node/assemble_test.go
```diff
@@ -161,3 +162,74 @@ func BenchmarkAssembleBlock(b *testing.B) {
...
+
+func TestAssembleBlockTransactionPoolBehind(t *testing.T) {
+	const numUsers = 100
+	expectingLog := false
+	baseLog := logging.TestingLog(t)
+	baseLog.SetLevel(logging.Info)
+	log := &callbackLogger{
+		Logger: baseLog,
+		WarnfCallback: func(s string, args ...interface{}) {
+			require.True(t, expectingLog)
```
tsachiherman, line 183: I think it would be a good idea to reset the expectingLog after that statement, right ?
i.e. to ensure we don't get 5 calls with that content.

tsachiherman (approved):
overall looks good. You get get extra brownie point for resetting the `expectingLog` after the callback, plus ensuring that this callback took place.
( I would just increment a counter, and test the counter after the AssembleBlock/AddBlock calls; but that's just one way of doing so )

tsachiherman (approved):
looks good.

PR #1612, opened 2020-10-09
Repack merkle trie pages before flush

tsachiherman:
## Summary

This PR makes the following changes in the merkle trie implementation:
1. The `NodesCountPerPage` was moved from the `Committer` interface into a merkle trie configuration structure.
    That allows us to put all the configuration in a single place.
2. Calling `Commit` on the trie would return a `CommitStats`, describing what happened under the hood. This is useful for debugging and performance tuning purposes.
3. Added child reallocation for nodes with high child count. This is particularly valuable for nodes that are on the top of the trie, since these tend to reload large number of page during the hash calculation.
4. New nodes packing. Between any two "Commit" calls, nodes are being added to the trie. These node aren't always filling their corresponding pages completely. As a result, we were storing higher number of pages than we need to. This PR packs all the new nodes, in order, into a new set page pages. That allows us to reduce the number of new pages that need to be flushed, hence, reducing the number of `StorePage` calls.
5. Few minor node optimization were made reclaiming discarded node memory when creating a new one.
6. Tweaking to the `CatchpointCatchupAccessor` allows dynamic Commit rate, which improves UI.
7. Preallocate the hash accumulation buffer, and have it allocated only once.

## Performance Gains

The following were measured using the `BenchmarkRestoringFromCatchpointFile` benchmark:

Accounts Number | Before Change (Current) | After Change 
-- | -- | -- 
102,400 | 31,799ns | 29,356ns 
204,800 | 42,833ns | 34,812ns
409,600 | 56,923ns | 42,854ns
819,200 | 81,288ns | 57,019ns

Note that the growth is composed of both the account data writing as well as the merkle trie writing.
The performance gains when we have 7M accounts are roughly 100% **.

** projected performance gains. We don't have a benchmark for that.

## Test Plan

Unit tests were updated.
Tested against mainnet.


crypto/merkletrie/cache.go
```diff
@@ -168,27 +187,37 @@ func (mtc *merkleTrieCache) getNode(nid storedNodeIdentifier) (pnode *node, err
...
 	return
 }
 
 // prioritizeNode make sure to adjust the priority of the given node id.
 // nodes are prioritized based on the page the belong to.
-// a new page would be placed on front, and an older page would get moved
+// a new page would be placed on front, and an existing page would get moved
 // to the front.
-func (mtc *merkleTrieCache) prioritizeNode(nid storedNodeIdentifier) {
+func (mtc *merkleTrieCache) prioritizeNode(nid storedNodeIdentifier, frontSide bool) {
```
algorandskiy, line 199: How about splitting this function into two `prioritizeNodeFront` and `prioritizeNodeBack` since they do not share any code now except element finding.
The reason for proposal is to get rid of boolean param from the signature and increase readability.

crypto/merkletrie/cache_test.go
```diff
@@ -391,7 +399,7 @@ func TestCachePageReloading(t *testing.T) {
 
 	earlyCachedNodeCount := mt1.cache.cachedNodeCount
 	// reloading existing cached page multiple time should not cause increase cached node count.
-	page := uint64(mt1.nextNodeID) / uint64(memoryCommitter.GetNodesCountPerPage())
+	page := uint64(mt1.nextNodeID-1) / uint64(defaultTestMemoryConfig.NodesCountPerPage)
```
algorandskiy, line 402: why '-1' ?

crypto/merkletrie/committer.go
```diff
@@ -41,7 +40,9 @@ func (mc *InMemoryCommitter) StorePage(page uint64, content []byte) error {
 	if content == nil {
 		delete(mc.memStore, page)
 	} else {
-		mc.memStore[page] = content
+		storedContent := make([]byte, len(content))
```
algorandskiy, line 43: why copying? Seems like all `StorePage`'s clients obtain a fresh array from encode/serialize

crypto/merkletrie/node.go
```diff
@@ -365,6 +372,37 @@ func deserializeNode(buf []byte) (n *node, s int) {
 		i++
 	}
 	n.children = make([]childEntry, i, i)
-	copy(n.children[:], childEntries[:i])
+	copy(n.children, childEntries[:i])
 	return
 }
+
+func (n *node) getUniqueChildPageCount(nodesPerPage int64) uint64 {
+	uniquePages := make(map[int64]struct{}, len(n.children))
```
algorandskiy, line 380: I wonder if map here works here faster than array of bool.

ledger/catchupaccessor.go
```diff
@@ -388,10 +389,27 @@ func (progress *CatchpointCatchupAccessorProgress) EvictAsNeeded(balancesCount u
...
+	}
 	// periodically, perform commit & evict to flush it to the disk and rebalance the cache memory utilization.
-	if (progress.ProcessedAccounts/trieRebuildCommitFrequency) < ((progress.ProcessedAccounts+balancesCount)/trieRebuildCommitFrequency) ||
+	if (progress.ProcessedAccounts/progress.evictFrequency) < ((progress.ProcessedAccounts+balancesCount)/progress.evictFrequency) ||
 		(progress.ProcessedAccounts+balancesCount) == progress.TotalAccounts {
+		evictStart := time.Now()
 		_, err = progress.cachedTrie.Evict(true)
+		if err == nil {
+			evictDelta := time.Now().Sub(evictStart)
+			if evictDelta > 2000*time.Millisecond {
```
algorandskiy, line 402: make 2000 and 1024 named constants?

crypto/merkletrie/cache.go
```diff
@@ -368,60 +516,135 @@ func (mtc *merkleTrieCache) commit() error {
...
+		// if we aren't going to reallocate any nodes, no need to allocate (maybe)
+		// new pages for these.
+		goto skipContentDeletion
+	}
+
+	if _, has := mtc.pageToNIDsPtr[nextPage]; !has {
+		pageMap := make(map[storedNodeIdentifier]*node, mtc.nodesPerPage)
+		mtc.reallocatedPages[nextPage] = pageMap
+		mtc.pageToNIDsPtr[nextPage] = pageMap
+		mtc.pagesPrioritizationMap[nextPage] = mtc.pagesPrioritizationList.PushFront(nextPage)
```
algorandskiy, line 583: this can be a local func parameterized with nextPage or lastPage

crypto/merkletrie/cache.go
tsachiherman: sounds reasonable.

crypto/merkletrie/cache.go
tsachiherman: true, but I don't like the idea of making another function here. I think that a better implementation would be to set up a local variable with `nextPage`/`lastPage` and perform the above if that was set.

crypto/merkletrie/cache_test.go
tsachiherman: because the page for nextNodeID might node exists if nextNodeID happen to be a multiple of `NodesCountPerPage`. On the other hand, after we commit, the last added node ( i.e. nextNodeID-1 ) should be stored, and therefore - it's page is there as well.

ledger/catchupaccessor.go
tsachiherman: sounds good.

algorandskiy (approved):

PR #1617, opened 2020-10-12
Added `package-upload` mule task

btoll:
## Summary

Upload build assets to `algorand-staging` at any point in the build pipeline process.

## Test Plan

`mule -f package-upload.yaml package-upload`

### DIFF
--- a/package-upload.yaml
+++ b/package-upload.yaml
@@ -1,0 +1,11 @@
+tasks:
+  - task: s3.BucketCopy
+    name: upload
+    src: $HOME/projects/go-algorand/tmp/node_pkgs/linux/amd64
+    dest: s3://algorand-staging/releases/$CHANNEL/$VERSION
+
+jobs:
+  package-upload:
+    tasks:
+      - s3.BucketCopy.upload
+


egieseke (approved):
Looks good.

package-upload.yaml
```diff
@@ -0,0 +1,11 @@
+tasks:
+  - task: s3.BucketCopy
+    name: upload
+    src: $HOME/projects/go-algorand/tmp/node_pkgs/linux/amd64
```
bricerisingalgorand, line 4: is this meant to only upload our linux-amd64 builds?

btoll: Currently, yes.  Since the eks pipeline uses stashing, this shouldn't be needed for the cluster, but it's a nice to have when doing a local build.

I could parameterize the OS and ARCH_TYPE, but that seems a bit burdensome to do so, at least now.

bricerisingalgorand: I've been working on this make target to do the uploads, it's not quite working yet though for reasons I don't understand. I think having this yaml file works too though

```
archive:
	CHANNEL=$(CHANNEL) \
	FULLBUILDNUMBER=$(FULLBUILDNUMBER) \
	aws s3 cp tmp/node_pkgs s3://algorand-internal/channel/${CHANNEL}/${FULLBUILDNUMBER} --recursive --exclude "*" --include "*${CHANNEL}*${FULLBUILDNUMBER}*"
```

bricerisingalgorand (approved):

package-upload.yaml
btoll: Yes, just a simple make target would work well, too.  We should have a ticket to investigate getting that working :)

PR #1618, opened 2020-10-12
optional network propagation tracing

brianolson:
Add network propagation tracing to `algod`
Compiled out by default, it can be enabled by building with:
`make GOTAGSCUSTOM=msgtrace`
And setting NetworkMessageTraceServer in config.json
`algocfg set -p NetworkMessageTraceServer -v server.addr:6225`
Traces when a block proposal message is handled.


config/config.go
```diff
@@ -349,6 +349,9 @@ type Local struct {
 	// and is expected to happen only on either startup ( after enabling the catchpoint interval, or on certain database upgrades ) or during fast catchup. The values specified here
 	// and their meanings are identical to the ones in LedgerSynchronousMode.
 	AccountsRebuildSynchronousMode int `version[12]:"1"`
+
+	// TraceServer is a host:port to report graph propagation trace info to.
+	TraceServer string `version[12]:""`
```
tsachiherman, line 354: TraceServer -> NetworkMessageTraceServer

node/node.go
```diff
@@ -268,6 +272,17 @@ func MakeFull(log logging.Logger, rootDir string, cfg config.Local, phonebookAdd
...
+		node.tracer = &graphtrace.NopClientSingleton
+	} else {
+		node.tracer, err = graphtrace.NewTcpClient(cfg.TraceServer)
+		if err != nil {
+			log.Errorf("unable to create trace client: %v", err)
+			return nil, err
+		}
+		gossip.Trace = node.tracer
+	}
+
```
tsachiherman, line 285: It doesn't seems that you really need it here. instead, just pass the cfg.TraceServer to 
gossip.InitializeNetworkMessageTracer(..).
The InitializeNetworkMessageTracer would be a function pointer which (by default) would be an empty function, and with the build tag would contain the above initialization.

Same goes for the usage of the tracer.. which would allow you to compile it with and without via a build tag.

node/node.go
```diff
@@ -268,6 +271,9 @@ func MakeFull(log logging.Logger, rootDir string, cfg config.Local, phonebookAdd
 		}
 	}
 
+	node.tracer = messagetracer.NewTracer().Init(cfg)
```
tsachiherman, line 274: it would be nice if you could pass the logger to the NewTracer so that we could stop using the logging.Base().

agreement/gossip/network.go
```diff
@@ -87,6 +91,13 @@ func (i *networkImpl) processVoteMessage(raw network.IncomingMessage) network.Ou
 }
 
 func (i *networkImpl) processProposalMessage(raw network.IncomingMessage) network.OutgoingMessage {
+	if Trace != nil {
+		hasher := fnv.New64a()
+		hasher.Write(raw.Data)
+		var msgBytes []byte = []byte("prop        ")
+		hash := hasher.Sum(msgBytes[0:4])
+		Trace.Trace(hash)
```
tsachiherman, line 99: Could you move this implementation into the `graphtraceMessageTracer.Trace` ? It would be much cleaner that way.

tsachiherman (changes requested):
two requests above, looks good otherwise.

tsachiherman:
you've modified the config.Local ( which is ok ), but now you'll need to `make generate` which would update the `config/local_defaults.go` and `installer/config.json.example`, you'll need to update the testing config json ( from the auto-gen `installer/config.json.example` ) and set the version on the `TraceServer` to 13, because we've already released version 12.

brianolson:
Finally passes all the checks, I think this is ready.

agreement/gossip/network.go
```diff
@@ -87,6 +92,9 @@ func (i *networkImpl) processVoteMessage(raw network.IncomingMessage) network.Ou
 }
 
 func (i *networkImpl) processProposalMessage(raw network.IncomingMessage) network.OutgoingMessage {
+	if Trace != nil {
+		Trace.HashTrace("prop", raw.Data)
```
tsachiherman, line 96: could you add a string constant for "prop" in "github.com/algorand/go-algorand/network/messagetracer" ?
( just so we'll have a single location with all the constants we use for the hash trace )

agreement/gossip/network.go
```diff
@@ -38,6 +39,10 @@ var (
 var messagesHandled = metrics.MakeCounter(metrics.AgreementMessagesHandled)
 var messagesDropped = metrics.MakeCounter(metrics.AgreementMessagesDropped)
 
+// Trace is the client for sending trace records for network propagation statistics.
+// It is public to allow configuration to be pushed in from node/node.go
+var Trace messagetracer.MessageTracer
```
tsachiherman, line 44: could you move this to the `networkImpl` and make it non-public ?
The node is also the one that calls the WrapNetwork, so it would be trivial to pass it in right after creation.

tsachiherman (changes requested):
Two small requests, but looks great otherwise.

tsachiherman (dismissed):

tsachiherman (approved):
Please update the PR title to be more descriptive, and when merging in, also ensure that you are squashing and provide a release notes description for this PR.

PR #1619, opened 2020-10-13
Change telemetryConfig from global variable to a member of the logger 

algonautshant:
log.go has a global TelemetryConfig variable for the general Base logger: 
`	telemetryConfig TelemetryConfig`
However, this is a struct, and the functions pass it around by value.
Many query functions answer by reading the values from the global variable. e.g. `s.log.GetTelemetrySession()`
When modifications are made, it happens to the local copy of the struct, which leads to inconsistencies. e.g. 
```
func makeTelemetryState(cfg TelemetryConfig, hookFactory hookFactory) (telemetryState, error) {
...
	if cfg.Enable {
		if cfg.SessionGUID == "" {
			cfg.SessionGUID = uuid.NewV4().String()
...
```

note: cfg.SessionGUID is an actual bug which triggered this PR.

This change converts the by value passing of the telemetryConfig to by pointer. 
A reason to pass around by value would be to maintain multiple versions of the telemetryConfig. I did not discover any such use case. 

The next step after this would be the elimination the global variable.

### DIFF
--- a/cmd/diagcfg/telemetry.go
+++ b/cmd/diagcfg/telemetry.go
@@ -53,7 +53,7 @@ func maybeUpdateDataDirFromEnv() {
 	}
 }
 
-func readTelemetryConfigOrExit() logging.TelemetryConfig {
+func readTelemetryConfigOrExit() *logging.TelemetryConfig {
 	maybeUpdateDataDirFromEnv()
 	cfg, err := logging.ReadTelemetryConfigOrDefault(&dataDir, "")
 	if err != nil {
@@ -63,7 +63,7 @@ func readTelemetryConfigOrExit() logging.TelemetryConfig {
 	return cfg
 }
 
-func saveTelemetryConfig(cfg logging.TelemetryConfig) {
+func saveTelemetryConfig(cfg *logging.TelemetryConfig) {
 	globalPath, err := config.GetConfigFilePath(logging.TelemetryConfigFilename)
 	if err != nil {
 		fmt.Fprintf(os.Stderr, "%s\n", err)
--- a/cmd/diagcfg/telemetry.go
+++ b/cmd/diagcfg/telemetry.go
@@ -53,7 +53,7 @@ func maybeUpdateDataDirFromEnv() {
 	}
 }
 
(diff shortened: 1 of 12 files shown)


cmd/diagcfg/telemetry.go
```diff
@@ -53,7 +53,7 @@ func maybeUpdateDataDirFromEnv() {
 	}
 }
 
-func readTelemetryConfigOrExit() logging.TelemetryConfig {
+func readTelemetryConfigOrExit() *logging.TelemetryConfig {
```
tsachiherman, line 56: this one doesn't need to be a pointer.

cmd/diagcfg/telemetry.go
```diff
@@ -63,7 +63,7 @@ func readTelemetryConfigOrExit() logging.TelemetryConfig {
 	return cfg
 }
 
-func saveTelemetryConfig(cfg logging.TelemetryConfig) {
+func saveTelemetryConfig(cfg *logging.TelemetryConfig) {
```
tsachiherman, line 66: I think that this isn't needed either.

tsachiherman (changes requested):
Can you go over the changes and have the structure pointer only where really needed ?
i.e. it looks like you've replaced `TelemetryConfig` with `*TelemetryConfig` which isn't really needed.

logging/log.go
```diff
@@ -52,7 +52,7 @@ type Level uint32
 // Create a general Base logger
 var (
 	baseLogger      Logger
-	telemetryConfig TelemetryConfig
+	telemetryConfig *TelemetryConfig
```
brianolson, line 55: Maybe this one should actually be owned here? Is this the one canonical value of TelemetryConfig, and we should pass references to this value? Is there one part of the code that owns the value if not here?

algonautshant: It is not clear when this global is accessed. And when this goes out of sync with a local variable, then it can be a bug. 
I will to do away with this global variable.

tsachiherman (approved):
looks good to me. I don't like the changes on the `logging.TelemetryOverride`, but we should defer this to a separate PR.

PR #1623, opened 2020-10-14
Add syncMu to AlgorandFullNode OnNewBlock

nicholasguoalgorand:
## Summary

Adding a new mutex to AlgorandFullNode so that OnNewBlock wouldn't be blocked by oldKeyDeletionThread during catchup.

### DIFF
--- a/node/node.go
+++ b/node/node.go
@@ -115,6 +115,9 @@ type AlgorandFullNode struct {
 
 	log logging.Logger
 
+	// use syncMu for locking lastRoundTimestamp and hasSyncedSinceStartup
+	// syncMu added so OnNewBlock wouldn't be blocked by oldKeyDeletionThread during catchup
+	syncMu                deadlock.Mutex
 	lastRoundTimestamp    time.Time
 	hasSyncedSinceStartup bool
 
@@ -578,7 +581,9 @@ func (node *AlgorandFullNode) GetPendingTransaction(txID transactions.Txid) (res
 // Status returns a StatusReport structure reporting our status as Active and with our ledger's LastRound
 func (node *AlgorandFullNode) Status() (s StatusReport, err error) {
 	node.mu.Lock()
+	node.syncMu.Lock()
 	defer node.mu.Unlock()
+	defer node.syncMu.Unlock()
 
 	s.LastRoundTimestamp = node.lastRoundTimestamp
 	s.HasSyncedSinceStartup = node.hasSyncedSinceStartup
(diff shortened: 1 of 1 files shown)


node/node.go
```diff
@@ -578,7 +581,9 @@ func (node *AlgorandFullNode) GetPendingTransaction(txID transactions.Txid) (res
 // Status returns a StatusReport structure reporting our status as Active and with our ledger's LastRound
 func (node *AlgorandFullNode) Status() (s StatusReport, err error) {
 	node.mu.Lock()
```
tsachiherman, line 583: It would be sufficient to take the new lock during the setup of these two output variables, i.e.
```
node.syncMu.Lock()
s.LastRoundTimestamp = node.lastRoundTimestamp
s.HasSyncedSinceStartup = node.hasSyncedSinceStartup
node.mu.Unlock()

node.mu.Lock()
defer node.syncMu.Unlock()
...
```

node/node.go
```diff
@@ -115,6 +115,9 @@ type AlgorandFullNode struct {
 
 	log logging.Logger
 
+	// use syncMu for locking lastRoundTimestamp and hasSyncedSinceStartup
+	// syncMu added so OnNewBlock wouldn't be blocked by oldKeyDeletionThread during catchup
+	syncMu                deadlock.Mutex
 	lastRoundTimestamp    time.Time
 	hasSyncedSinceStartup bool
```
tsachiherman, line 122: When writing a variable description in go, try to always start with the variable name; go-vet would complain about that for public functions.

node/node.go
```diff
@@ -115,6 +115,9 @@ type AlgorandFullNode struct {
 
 	log logging.Logger
 
+	// use syncMu for locking lastRoundTimestamp and hasSyncedSinceStartup
+	// syncMu added so OnNewBlock wouldn't be blocked by oldKeyDeletionThread during catchup
+	syncMu                deadlock.Mutex
```
tsachiherman, line 120: could you change the variable name to something more concrete. ( since all mutexes are meant for syncing.. )
maybe lastRoundTimeMu ?

tsachiherman (changes requested):
Few small requests, but otherwise looks good.

node/node.go
brianolson: the two variables together might be 'sync status', so this could be the 'sync status mutex'

tsachiherman (approved):
looks good. thanks!

PR #1625, opened 2020-10-15
Brice/perf10x

bricerisingalgorand:
## Summary

Implement performance test scenario that generates a network similar to real world conditions. Most of the data comes from speedtest.com to generate an appropriate network profile for different groups of ec2 instances based on the network performances of different continents.

## Test Plan

I tested this by running the scripts and generating a performance test network based of it


scripts/configure_dev.sh
```diff
@@ -69,6 +69,13 @@ function install_windows_shellcheck() {
     return 0
 }
 
+function install_python_packages() {
+    pip3 install \
+        argparse \
+        simplejson \
+        awscli
```
tsachiherman, line 76: Are you using awscli ? I haven't see usage of it.

tsachiherman (changes requested):
I'm hesitant about adding these pip3 packages. The number of people that would use these is going to be pretty low, and forcing everyone to install these seems like using a bigger hammer than needed.

What about adding a makefile at the destination directory to preinstall the prerequisites when invoked ? ( note that most of the time it won't be invoked, since we already having the "output" of these )

as a sidenote, the fail reason for travis here was the pip3 invocation..

tsachiherman (changes requested):
Could you change the perf10x recipe name to "mainnet-model" ?
We might end up creating multiple network recipes for this project, so naming it after the project name might not be the most obvious choice.

tsachiherman (changes requested):
The `test/testdata/deployednettemplates/recipes/generate-recipe/` directory seems to be abit misplaced : it's not a recipe but rather a recipe-generation engine.
Could you move it to `test/testdata/deployednettemplates/generate-recipe` ?

tsachiherman:
Few requests above, but looks great otherwise.

test/testdata/deployednettemplates/recipes/generate-recipe/generate_network.py
```diff
@@ -0,0 +1,164 @@
+#!/usr/bin/env python3
+
+import argparse
+import simplejson
```
brianolson, line 4: I'd rather stick to standard library and I don't see simplejson doing anything special over standard library json

test/testdata/deployednettemplates/recipes/generate-recipe/generate_network.py
```diff
@@ -0,0 +1,164 @@
...
+
+    template_path = os.path.abspath(os.path.dirname(args.template))
+    script_path = os.path.dirname(__file__)
+    topology = build_topology(template_dict)
+
+    gen_dir = f"{template_path}/generated"
+    if not os.path.isdir(gen_dir):
+        os.mkdir(gen_dir)
+
+    subprocess.run(['cp', f"{script_path}/recipe.json", f"{template_path}/recipe.json"])
```
brianolson, line 22: `import shutil` probably has `cp` in a way better than subprocess

scripts/configure_dev.sh
```diff
@@ -69,6 +69,13 @@ function install_windows_shellcheck() {
     return 0
 }
 
+function install_python_packages() {
+    pip3 install \
+        argparse \
```
brianolson, line 74: argparse is standard library and shouldn't need pip install. swapping in standard library json then removes this entire need for pip doing anything to Python

bricerisingalgorand: Thanks! I removed this and switched to the json lib

test/testdata/deployednettemplates/recipes/generate-recipe/generate_network.py
bricerisingalgorand: Thanks, i switched to shutil

bricerisingalgorand:
Thanks for the feedback! I switched to standard lib python packages to remove the need for the pip3 installs in configure_dev, and made the suggested folder structure changes

tsachiherman (approved):
looks good to me. thanks for the changes.

PR #1626, opened 2020-10-15
extend account lookup to include validity range

tsachiherman:
## Summary

Extend the existing implementation of account updates to provide already-available information for round range validity for account data queries. The long term goal here is to take advantage of that in the agreement implementation so that the number of queries that would reach the ledger would get reduced.

The change in a nutshell:
The function `accountUpdates. Lookup` got broken into
`LookupWithRewards(rnd basics.Round, addr basics.Address) (data basics.AccountData, err error)`
and
`LookupWithoutRewards(rnd basics.Round, addr basics.Address) (basics.AccountData, basics.Round, error)`

The latter includes a validity round as the second return value.

## Test Plan

Unit tests were extended to cover the changes.

### DIFF
--- a/ledger/acctupdates.go
+++ b/ledger/acctupdates.go
@@ -366,11 +366,16 @@ func (au *accountUpdates) IsWritingCatchpointFile() bool {
 	}
 }
 
-// Lookup returns the account data for a given address at a given round. The withRewards indicates whether the
-// rewards should be added to the AccountData before returning. Note that the function doesn't update the account with the rewards,
-// even while it could return the AccoutData which represent the "rewarded" account data.
-func (au *accountUpdates) Lookup(rnd basics.Round, addr basics.Address, withRewards bool) (data basics.AccountData, err error) {
-	return au.lookupImpl(rnd, addr, withRewards, true /* take lock*/)
+// LookupWithRewards returns the account data for a given address at a given round.
+// Note that the function doesn't update the account with the rewards,
+// even while it does return the AccoutData which represent the "rewarded" account data.
+func (au *accountUpdates) LookupWithRewards(rnd basics.Round, addr basics.Address) (data basics.AccountData, err error) {
+	return au.lookupWithRewardsImpl(rnd, addr)
+}
+
+// LookupWithoutRewards returns the account data for a given address at a given round.
+func (au *accountUpdates) LookupWithoutRewards(rnd basics.Round, addr basics.Address) (data basics.AccountData, validThrough basics.Round, err error) {
+	return au.lookupWithoutRewardsImpl(rnd, addr, true /* take lock*/)
 }
 
 // ListAssets lists the assets by their asset index, limiting to the first maxResults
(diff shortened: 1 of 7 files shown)


ledger/acctupdates.go
```diff
@@ -1478,14 +1481,14 @@ func (au *accountUpdates) lookupImpl(rnd basics.Round, addr basics.Address, with
...
+		// check if we've had this address modified in the past rounds. ( i.e. if it's in the deltas )
+		macct, indeltas := au.accounts[addr]
+		if indeltas {
+			// Check if this is the most recent round, in which case, we can
+			// use a cache of the most recent account state.
+			if offset == uint64(len(au.deltas)) {
 				return macct.data, nil
 			}
-		} else {
 			// Check if the account has been updated recently.  Traverse the deltas
```
algonautshant, line 1492: The updated comment in lookupWithoutRewardsImpl is very nice. Will be good to copy it here as well. 
			// the account appears in the deltas, but we don't know if it appears in the
			// delta range of [0..offset], so we'll need to check :
			// Traverse the deltas backwards to ensure that later updates take
			// priority if present.

ledger/acctupdates.go
```diff
@@ -1478,14 +1481,14 @@ func (au *accountUpdates) lookupImpl(rnd basics.Round, addr basics.Address, with
...
 			// backwards to ensure that later updates take priority if present.
 			for offset > 0 {
```
algonautshant, line 1494: if ok { ...
					// the returned validThrough here is not optimal, but it still correct. We could get a more accurate value by scanning
					// the deltas forward, but this would be time consuming loop, which might not pay off.

algonautshant (approved):
Looks very good. 
I noticed that LookupWithRewards and LookupWithoutRewards have many common segments. It is great that both functions are simpler now. Also, the new comments are very helpful. Will be nice to copy over the comments to the same segments to both functions.

ledger/acctupdates.go
tsachiherman: The comment is not relevant here, since `lookupWithRewardsImpl` doesn't return validity range. ( i.e. it can't due to the rewards )

algonautshant (approved):

ledger/acctupdates.go
```diff
@@ -1497,6 +1502,82 @@ func (au *accountUpdates) lookupImpl(rnd basics.Round, addr basics.Address, with
...
+		var dbRound basics.Round
+		// No updates of this account in the in-memory deltas; use on-disk DB.
+		// The check in roundOffset() made sure the round is exactly the one
+		// present in the on-disk DB.  As an optimization, we avoid creating
+		// a separate transaction here, and directly use a prepared SQL query
+		// against the database.
+		data, dbRound, err = au.accountsq.lookup(addr)
+		if dbRound == currentDbRound {
+			return data, err
+		}
```
algorandskiy, line 1517: this code is run without a lock, what would happen if another thread adds an update to in-memory? Do we return outdated info from the DB?

ledger/acctupdates.go
```diff
@@ -1497,6 +1502,82 @@ func (au *accountUpdates) lookupImpl(rnd basics.Round, addr basics.Address, with
...
+			// priority if present.
+			for offset > 0 {
+				offset--
+				d, ok := au.deltas[offset][addr]
+				if ok {
+					// the returned validThrough here is not optimal, but it still correct. We could get a more accurate value by scanning
+					// the deltas forward, but this would be time consuming loop, which might not pay off.
+					return d.new, rnd, nil
+				}
+			}
```
algorandskiy, line 1572: nit: maybe remove all the comment and put "see lookupWithRewardsImpl above"? Like there is only 3 lines of code and 2x more comments.

algorandskiy (commented):
Looks good, one question - see below

ledger/acctupdates.go
tsachiherman: yes, another thread could make a concurrent memory update. however, the query is always for historical data, so as long as we can guarantee that the account data is the correct one for the given round, it doesn't really matter.

ledger/acctupdates.go
tsachiherman: It's true that we have lots of comments, but I think I'd rather that vs. the alternative. Functions tend to change over time, and I don't know how long the "lookup" comment would hold true.

ledger/acctupdates.go
tsachiherman: I.e. the case where we return an error since the round is not in the [dbround..dbround+len(delta)], is not an issue since it applied regardless of whether we have a lock or not.

algorandskiy (approved):

PR #1627, opened 2020-10-15
Output error when 'goal node stop' is run on invalid directory

nicholasguoalgorand:
## Summary

Output error when 'goal node stop' is run on invalid directory, gives user feedback instead of saying node successfully stopped.

## Test Plan

Ran the commands in terminal


tsachiherman (changes requested):
Few quests:
1. add expect test to prove the correctness of your change.
2. The existing pattern of `alreadyStopped bool, err error` is common. If err is set, the alreadyStopped is expected to be ignored. Unfortunately, this isn't what we're after. We need to return here few possible errors that would need to be handled differently on the client side. My suggestion:

a. create an error struct for alreadyStopped.
b. create an error struct for invalidDataDirectory
c. change StopAlgod so that it would return only a single error, and add error testing & formatting upstream.

nodecontrol/algodControl.go
```diff
@@ -133,7 +149,11 @@ func (nc NodeController) algodRunning() (isRunning bool) {
 }
 
 // StopAlgod reads the net file and kills the algod process
-func (nc *NodeController) StopAlgod() (alreadyStopped bool, err error) {
+func (nc *NodeController) StopAlgod() (err error) {
+	// Check for valid data director
+	if !util.IsDir(nc.algodDataDir) {
```
algonautshant, line 154: This comment is for something not part of this PR, but I would like to raise it here:
Why do we say "kill node"? Shouldn't it be "stop node"? 
```
 $ goal node stop -d /tmp/a.txt 
Cannot kill node: /tmp/a.txt is not a valid directory
```
Instead:
`Cannot stop node: /tmp/a.txt is not a valid directory.`

algonautshant (commented):
Looks good. 
I have one comment and one (unrelated) comment.

nodecontrol/algodControl.go
```diff
@@ -40,6 +40,22 @@ const StdErrFilename = "algod-err.log"
 // StdOutFilename is the name of the file in <datadir> where stdout will be captured if not redirected to host
 const StdOutFilename = "algod-out.log"
 
+type NodeAlreadyStoppedError struct {
```
algonautshant, line 43: Travis failure: Comment is missing here.
`nodecontrol/algodControl.go:43:6: exported type NodeAlreadyStoppedError should have comment or be unexported
`

nodecontrol/algodControl.go
```diff
@@ -40,6 +40,22 @@ const StdErrFilename = "algod-err.log"
...
+	algodDataDir string
+}
+
+func (e *NodeAlreadyStoppedError) Error() string {
+	return fmt.Sprintf("Node in directory %s has already stopped", e.algodDataDir)
+}
+
+type InvalidDataDirError struct {
```
algonautshant, line 51: Travis failure: Comment is missing here.

`nodecontrol/algodControl.go:51:6: exported type InvalidDataDirError should have comment or be unexported`

nodecontrol/algodControl.go
```diff
@@ -40,6 +40,22 @@ const StdErrFilename = "algod-err.log"
 // StdOutFilename is the name of the file in <datadir> where stdout will be captured if not redirected to host
 const StdOutFilename = "algod-out.log"
 
+type NodeAlreadyStoppedError struct {
+	algodDataDir string
+}
+
+func (e *NodeAlreadyStoppedError) Error() string {
+	return fmt.Sprintf("Node in directory %s has already stopped", e.algodDataDir)
```
tsachiherman, line 48: nit: place the %s in a single-quotes, so that the user would be able to distinguish between the messages and the argument.

nodecontrol/algodControl.go
```diff
@@ -40,6 +40,22 @@ const StdErrFilename = "algod-err.log"
...
+}
+
+type InvalidDataDirError struct {
+	algodDataDir string
+}
+
+func (e *InvalidDataDirError) Error() string {
+	return fmt.Sprintf("%s is not a valid directory", e.algodDataDir)
```
tsachiherman, line 56: This is not the "correct" error for this use case. using util.IsDir would return true if the directory exists.
It doesn't tell you if a directory of that name could be created.
In that case, you want a "MissingDataDirError" which would format into
"the provided data directory '%s' does not exists"
You could get more information by calling os.Stat and getting the precise underlying error.

nodecontrol/algodControl.go
```diff
@@ -133,7 +149,11 @@ func (nc NodeController) algodRunning() (isRunning bool) {
 }
 
 // StopAlgod reads the net file and kills the algod process
-func (nc *NodeController) StopAlgod() (alreadyStopped bool, err error) {
+func (nc *NodeController) StopAlgod() (err error) {
+	// Check for valid data director
```
tsachiherman, line 153: nit: typo : director -> directory

nodecontrol/algodControl.go
tsachiherman: yes, I agree that the user-facing error message should refrain of using the "kill" terminology, since it's the signal name and not the operation.

nodecontrol/algodControl.go
```diff
@@ -143,8 +163,7 @@ func (nc *NodeController) StopAlgod() (alreadyStopped bool, err error) {
 			return
 		}
 	} else {
-		err = nil
-		alreadyStopped = true
+		return &NodeAlreadyStoppedError{algodDataDir: nc.algodDataDir}
```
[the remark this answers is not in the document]
tsachiherman: I think it's ok, but would drop the "configuration" and add single quotes around the directory name.

tsachiherman (changes requested):
requested few changes. looks pretty good.

nodecontrol/algodContol_test.go
```diff
@@ -0,0 +1,17 @@
...
+//
+// go-algorand is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with go-algorand.  If not, see <https://www.gnu.org/licenses/>.
+
+package nodecontrol
```
tsachiherman, line 17: wait! I'm confused - is that an empty file ?!

nodecontrol/algodControl.go
tsachiherman: @algonautshant - could you open another PR for the kill vs stop ?
it's in goal,
```
errorKill                         = "Cannot kill node: %s"
```
and would affect some of the changes made here ( but maybe more )

nodecontrol/algodContol_test.go
nicholasguoalgorand: strange, git and renaming files work strange

test/e2e-go/cli/algod/expect/algodTelemetryLocationTest.exp
```diff
@@ -19,6 +19,7 @@ if { [catch {
     #allows script to be run outside of go context
     exec mkdir -p $TEST_PRIMARY_NODE_DIR
 
+    exec goal node start -d $TEST_PRIMARY_NODE_DIR
```
tsachiherman, line 22: I'm looking on this one and I'm wondering why we need to make the call to `goal node stop` ?
Can you try to run it without it to see if the test passes ?

tsachiherman (approved):

PR #1628, opened 2020-10-16
Added S3_SOURCE env var to specify source of packages

btoll:
## Summary

Adds an `S3_SOURCE` var that allows the packages to be downloaded from different locations for testing and signing.

To do so, `S3_SOURCE` must be set.

The impetus for this grew out of one of our conversations where we stated that it would be good to get the packages from `algorand-internal` and only push signed assets to staging.

## Test Plan

`S3_SOURCE=the-staging-area mule -f package-sign.yaml package-sign`


onetechnical (dismissed):
Looks fine to me.

egieseke (dismissed):
Looks good.

bricerisingalgorand (dismissed):

onetechnical (approved):
I think this is good, but please test if you can.

scripts/release/mule/test/test.sh
```diff
@@ -21,17 +21,17 @@ PKG_DIR="./tmp/node_pkgs/$OS_TYPE/$ARCH_TYPE"
 
 pushd "$PKG_DIR"
 
-if ! $USE_CACHE
+if [ -n "$S3_SOURCE" ]
```
bricerisingalgorand, line 24: Is this going to always download from s3 now, instead of being able to use already downloaded binaries? This might not mesh well with the folder structure the package pipeline uses, since the prefix isn't the same between os/arch's

egieseke (approved):
Looks good.

scripts/release/mule/test/tests/post/verify_genesis_file.sh
```diff
@@ -22,7 +22,7 @@ EXPECTED_NETWORK=$(jq -r '.network' $GEN_FILE)
 
 if [ "$NETWORK" != "$EXPECTED_NETWORK" ]
 then
-    echo "[$0] The network value \`$NETWORK\` in \`$GEN_FILE\` is incorrect, it does not match $EXPECTED_NETWORK."
+    echo "[$0] The network value \`$NETWORK\` in \`$GEN_FILE\` is incorrect, it does not match \`$EXPECTED_NETWORK\`."
```
tsachiherman, line 25: I reviewed this file, and got confused.. the `EXPECTED_NETWORK` variable is mis-named.
It should be `ACTUAL_NETWORK`, right ?
the network that we expect to find is the one in `NETWORK`. ( which is fine.. )

note that this is a nice-to-have only. The functionality of this code is correct.

tsachiherman (commented):
I think that besides my comment above, the changes here seems to be valid, but I have no idea how to test these.

scripts/release/mule/test/test.sh
btoll: No, it will only download from s3 if `S3_SOURCE` is specified.

PR #1634, opened 2020-10-20
tealdbg: listen for upcoming spinoff connections

algorandskiy:
## Summary

Use:
$ tealdbg debug -q
$ curl -X http://localhost:9293/spinoff --data-binary @dr.msgp

The /spinoff handler returns text/plain and
- HTTP 200 and ws://abc/sessionid in case of success
- HTTP 400 and error text in case of error
An URL in response body (ws://abc/) can be used for interactive debugging
with Chrome DevTools or for scripted debugging with CDT proto 1.1

A race condition is also fixed: dbg state access in the evaluator and in debugging core go-routine.
A possible deadlock on start is fixed: attempt to resume execution before initialization completed.

## Test Plan

Added e2e test

### DIFF
--- a/cmd/tealdbg/cdtdbg.go
+++ b/cmd/tealdbg/cdtdbg.go
@@ -33,6 +33,7 @@ import (
 type CdtFrontend struct {
 	mu         deadlock.Mutex
 	sessions   map[string]cdtSession
+	latestSids []string
 	router     *mux.Router
 	apiAddress string
 	verbose    bool
@@ -84,6 +85,7 @@ func (a *CdtFrontend) SessionStarted(sid string, debugger Control, ch chan Notif
 	s.states = debugger.GetStates(nil)
 
 	a.sessions[sid] = *s
+	a.latestSids = append(a.latestSids, sid)
 }
 
 // SessionEnded removes the session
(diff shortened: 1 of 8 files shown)


tsachiherman:
it's still data racing..

algorandskiy:
It is another one :) Fixed now.

cmd/tealdbg/cdtdbg.go
```diff
@@ -84,6 +85,7 @@ func (a *CdtFrontend) SessionStarted(sid string, debugger Control, ch chan Notif
 	s.states = debugger.GetStates(nil)
 
 	a.sessions[sid] = *s
+	a.latestSid = sid
```
tsachiherman, line 88: maybe I'm missing something here: it looks as if a call to `SessionEnded` could leave you with a `a.latestSid` that doesn't have a matching session object, causing `URL()` to return an empty string ?

cmd/tealdbg/debugger_test.go
```diff
@@ -142,19 +143,17 @@ func TestSession(t *testing.T) {
...
 
-	sig := make(chan struct{})
 	ackCount := 0
+	done := make(chan struct{})
 	ackFunc := func() {
-		sig <- struct{}{}
 		<-s.acknowledged
 		ackCount++
-		<-sig
+		done <- struct{}{}
```
tsachiherman, line 151: if would be more reliable to use defer statement at the beginning of the function, as it handles panic use cases as well.

tsachiherman (commented):
Overall looks great, a question below:

cmd/tealdbg/debugger_test.go
algorandskiy: I do not think panic needs to be handled here, it the test panics let it print the dump

cmd/tealdbg/cdtdbg.go
algorandskiy: Your are right, although this implementation works for a current use scenario, it breaks in others. I'll fix it.

algorandskiy: changed to ordered array of session ids, return the last.

tsachiherman (approved):
looks great. thanks for the changes.

PR #1639, opened 2020-10-21
Netgoal general templates

brianolson:
Allow `netgoal build` to more generally fill in template values.
`netgoal build ... -v SomeConfigField=foo` will replace `{{SomeConfigField}}` in template files with `foo`
I have used this to construct a test cluster that configures an experimental config.json field I hacked into a development `algod`

### DIFF
--- a/netdeploy/remote/deployedNetwork.go
+++ b/netdeploy/remote/deployedNetwork.go
@@ -45,6 +45,16 @@ var ErrDeployedNetworkInsufficientHosts = fmt.Errorf("target network requires mo
 // ErrDeployedNetworkNameCantIncludeWildcard is returned by Validate if network name contains '*'
 var ErrDeployedNetworkNameCantIncludeWildcard = fmt.Errorf("network name cannont include wild-cards")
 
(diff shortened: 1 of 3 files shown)


netdeploy/remote/deployedNetwork.go
```diff
@@ -104,12 +100,21 @@ func replaceTokens(original string, buildConfig BuildConfig) (expanded string, e
...
+	openIndex := strings.Index(expanded, "{{")
+	closeIndex := strings.Index(expanded, "}}")
+	if openIndex >= 0 || closeIndex >= 0 {
+		if openIndex < 0 {
+			openIndex = 0
+		}
+		if closeIndex < 0 {
+			closeIndex = len(expanded) - 2
+		}
+		return "", fmt.Errorf("config file contains unrecognized token: %s", expanded[openIndex:closeIndex+2])
```
tsachiherman, line 117: could you create a error type `type ErrDeployedNetworkTokenError struct` for that ? The consumer would be able to use `errors.As` to determine if it's an error of the given type.

tsachiherman (changes requested):
looks good, one request regarding exposed error types.

tsachiherman (approved):

Issue #1633, opened 2020-10-19
Please create TEAL Version 3 for collecting TEAL changes for future releases


jeapostrophe:
@jannotti and I were just talking about it would be useful for `go-algorand` to have a "TEAL version 3" in the consensus parameters and in the TEAL evaluator that is currently turned off and intended to be turned off until some later date, but that can be optionally enabled in development nodes (not TestNet or BetaNet) for testing Algorand dApps. A pattern like this would ensure that existing TEAL contract behavior doesn't change, but would facilitate developing more advanced TEAL programs, the Reach compiler, and the future Algorand layer-2 contracts.

There are many simple things that could be added and tested quickly if there were a structure like this in place:
- [ ] stateless args visible to other txns (#1533 / #1640 )
- [ ] swap (#1534 / #1640)
- [ ] bit shifts (#1062 / #1640 )
- [ ] if-then-else opcode for a conditional-move, but stack machine style (#1640)
- [ ] assert opcode (#1640)
- [ ] guarantees about integer constants, and maybe more 1 byte options.
- [ ] immediate varuint integers
- [ ] a `stringsplice` opcode like `substring`/`substring3` for injecting new bytes into a region (~ #1640)
- [ ] efficient msgpack-based byte string access, like LLVM's getelementptr
- [ ] accurate cost accounting and backward jumps (#1211)
- [ ] experimenting with different program limits (#1005)

I am very happy to implement most of these things, but I want to have a process in place for them to get in and testable beyond my own Algorand fork. I can wait a long time for it to be "live", but I don't want to wait a long time for it to be "eventually live".

Thanks!

jeapostrophe:
I work around most of these issues today, but the Reach compiler has to produce much much larger TEAL programs than it otherwise would and it stresses the limits of TEAL quite a bit. For example, consider `stringsplice`

Here's the expected behavior:
```teal
bytes "0123456789"
bytes "aaaa"
stringsplice 3
bytes "012aaaa789"
=
```

If you don't have this op-code then you need to do something like:
```teal
...original...
// stack: [ bytes ]
dup
// stack: [ bytes, bytes ]
substring 0 3
// stack: [ before, bytes ]
...new value...
concat
// stack: [ before', bytes ]
swap
// stack: [ bytes, before' ]
substring 3 len
// stack: [ after, before' ]
swap
// stack: [ before', after ]
concat
// stack: [ bytes' ]
```

[...]

PR #1640, opened 2020-10-22
Add some TEAL3 opcodes

jeapostrophe:
I don't expect you to take this as-is, without addressing #1633, and adding a version 3, but I am working on a program and these instructions are very useful for me.

For example, I have an example program that verifies the rules of a simple game in a TEAL contract. My original compiler as of Monday used ~6000 bytes for the program, then I fixed a dumb thing and got it to 3262 bytes, but then I did a lot of aggressive optimizations over the last few days and got it down to 946 bytes, but I expect the argument size to be 77 bytes, so `946 + 77 = 1023`, so I needed to do something else. These new instructions, especially `assert` and `ite` drop this program down to 864 bytes, which is good for me.

(You can get `assert` via `bz fail; .....; fail: int 0; return <EOF>`, but `bz` is three bytes, so that's a lot when most of a TEAL program is checking or failing.)

(You can get `cond; true; false; ite` via `cond; bz l_false; true; bz l_join; false; l_join:` which is 6 bytes vs 1, so it's a pretty decent win. `ite` is only useful if `true` and `false` are simple, so it is like a conditional move or the `?:` operator.)

(Swap is a big win in size, because `store 0; store 1; load 0; load 1` is 8 bytes plus pinning two slots.)

### DIFF
--- a/cmd/tealdbg/cdtState.go
+++ b/cmd/tealdbg/cdtState.go
@@ -145,7 +145,7 @@ func (s *cdtState) getObjectDescriptor(objID string, preview bool) (desc []cdt.R
 				return
 			}
 			if len(s.txnGroup) > 0 {
-				return makeTxnImpl(&s.txnGroup[idx].Txn, idx, preview), nil
+				return makeTxnImpl(&s.txnGroup[idx], idx, preview), nil
 			}
 		} else if parentObjID, ok := decodeArrayLength(objID); ok {
 			switch parentObjID {
@@ -357,7 +357,7 @@ func prepareGlobals(globals []basics.TealValue) []fieldDesc {
 	return result
 }
 
-func prepareTxn(txn *transactions.Transaction, groupIndex int) []fieldDesc {
+func prepareTxn(stxn *transactions.SignedTxn, groupIndex int) []fieldDesc {
 	result := make([]fieldDesc, 0, len(logic.TxnFieldNames))
 	for field, name := range logic.TxnFieldNames {
 		if field == int(logic.FirstValidTime) ||
@@ -367,7 +367,7 @@ func prepareTxn(txn *transactions.Transaction, groupIndex int) []fieldDesc {
 		}
 		var value string
 		var valType string = "string"
-		tv, err := logic.TxnFieldToTealValue(txn, groupIndex, logic.TxnField(field), 0)
+		tv, err := logic.TxnFieldToTealValue(stxn, groupIndex, logic.TxnField(field), 0)
 		if err != nil {
 			value = err.Error()
 			valType = "undefined"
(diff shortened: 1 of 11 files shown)


jeapostrophe:
p.s. I will look into how to add opcode tests to the suite and put in tests for these... it wasn't obvious to me so I haven't yet figured it out.

pbennett:
Adding more powerful opcodes to TEAL would be fantastic.  
I'd recommend you add BenchmarkXXXX functions which exercise these opcodes, so their computational impact can be assessed.  In particular, a benchmark with a 'real' program created from Reach, and two sub-benchmarks, one using the TEAL2 opcodes, and another using the TEAL3 opcodes would be an excellent starting point.  
Comparing the two would trivial, and effective in showing the value and usefulness of a few more expressive opcodes (or not! real data matters...)

Here's a nice example of sub-benchmarks if you're not familiar:
https://medium.com/justforfunc/analyzing-the-performance-of-go-functions-with-benchmarks-60b8162e61c6

jeapostrophe:
Thanks Patrick, just a quick FYI, these instructions are not about making the TEAL program faster, they are all about conserving byte size, because of the strict 1000 byte limit on the program and its arguments. Your comment makes me think you're thinking of speed. Without these instructions, I don't know how to compile the Reach program to be small enough because it is just too big, so I can't benchmark it at all in TEAL2, ya know?

pbennett:
I think speed has to always be a consideration.  1000+TPS (and with work ongoing to increase that), means that those transactions need to be contracts as well and perform comparably.  Just requiring fewer bytes for the same operations could impact the performance in numerous, and not always obvious ways.  Performance could be significantly improved.
However, if performance is significantly impacted, that could be a problem unto itself.  If it means certain code isn't possible at all because of size restrictions though, I think that's a reasonable argument to make in return.
If performance is unchanged, yet size reduced, or even better, if performance is increased - both make it IMO a no-brainer.

jeapostrophe:
I just added four more instructions that are simplified versions of `substring/strsplice` for bytes. This cuts another 100-ish bytes of my current program and finally gets it + arguments below the limit. (I originally thought the arguments were 77 bytes, but actually they were 209 for one step, so I needed to make some more cuts.)

It would probably be nice(r) to have a `bitset/get` that worked on bytes, but when I was implementing it, I felt that code was too complicated to work on both ints and bytes, so this was simpler and serves the purpose well.

brianolson:
Your assert could be a couple bytes shorter using the `err` op. `...; bz fail; ...; fail: err`

jannotti:
I'm going to close this as everything (except splice) is in teal3. I think teal4 will make splice less important (as functions will be possible)

PR #1644, opened 2020-10-22
Refactor parseIntcblock

nicholasguoalgorand:
## Summary

Modify the error messages for parseIntcblock to distinguish fro, parseBytecblock

## Test Plan

None

### DIFF
--- a/data/transactions/logic/assembler.go
+++ b/data/transactions/logic/assembler.go
@@ -1322,7 +1322,7 @@ func disDefault(dis *disassembleState, spec *OpSpec) {
 	_, dis.err = fmt.Fprintf(dis.out, "%s\n", spec.Name)
 }
 
-var errShortIntcblock  = errors.New("intcblock ran past end of program")
+var errShortIntcblock = errors.New("intcblock ran past end of program")
 var errTooManyIntc = errors.New("intcblock with too many items")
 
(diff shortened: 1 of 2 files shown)


tsachiherman:
You'll need to add a unit test that confirm the correctness of the modified functions, but beside an extra space after `errShortIntcblock  =`, the change looks good.

data/transactions/logic/assembler.go
```diff
@@ -1322,6 +1322,7 @@ func disDefault(dis *disassembleState, spec *OpSpec) {
 	_, dis.err = fmt.Fprintf(dis.out, "%s\n", spec.Name)
 }
 
+var errShortIntcblock  = errors.New("intcblock ran past end of program")
```
brianolson, line 1325: run `go fmt` or `make sanity`

tsachiherman (changes requested):
Two small issues, looks good otherwise:
- Remove node.go and network.go from this PR. You haven't changes these, and it shouldn't be in this PR.
- The unit test for parseIntcblock is good, but we also need one for checkIntConstBlock. You can populate the content of the evalContext object you pass to checkIntConstBlock for test purposes.

tsachiherman (approved):

PR #1646, opened 2020-10-23
Ensure syncStartNS is managed correctly across catchpoint file writing

### LINKED ISSUE #1529: catchupTime resets incorrectly while still catching up
### Subject of the issue
When performing a classic sync (non-catchpoint), the `catchupTime` property returned for `status` resets to 0 whenever it detects a catchpoint file being written.  This breaks tools that depend on the non-zero -> zero transition to detect the node being caught up.

This catchpoint-related change, in `catchup/service.go:pipelinedFetch()` broke the implementation:
```
	// if we're writing a catchpoint file, stop catching up to reduce the memory pressure. Once we finish writing the file we
	// could resume with the catchup.
	if s.ledger.IsWritingCatchpointFile() {
		s.log.Info("Catchup is stopping due to catchpoint file being written")
		return
	}
```

`catchup/service.go:sync()` resets `s.syncStartNS` when starting to sync.  Returning early from `pipelinedFetch()` causes `sync()` to return; so resuming the catchup means `s.syncStartNS` is reset to `Now()` after every interruption, when logically we don't want to reset it when being interrupted in this instance.  (`catchupTime` is computed based on `s.syncStartNS`, so this reset opens the possibility for the caller to see a `catchupTime` value of 0 when the node hasn't caught up).

### Your environment
* Software version: v2.1.3
* Node status if applicable:
* Operating System details. (irrelevant)

### Steps to reproduce
Start a new non-relay node with indexer enabled (this disables catchpoint catchup).  Loop calling `v2/status` until catchupTime transitions from non-zero to zero; compare lastRound to expected value (polling at 30-second interval, we see this trigger before syncing to round 50,000)

### Expected behaviour
catchupTime should not be reset when intentionally interrupted due to catchpoint file write.  This also messes up the `initialSyncComplete` logic - I forget where that's used.

### Actual behaviour
See above

tsachiherman:
## Summary

The `catchup.Service.sync` function was incorrectly setting/resetting the `syncStartNS` in the case of the catchup was paused due to a catchpoint file being written.

The PR changes the function logic as follows :
- When the function starts, it sets the `syncStartNS` only if it's zero. That allows as to "resume" previously suspended catchup.
- When the function ends, it tests to see if the catchup termination was due to catchpoint file being written. If so, it doesn't reset the `syncStartNS`.

Also, the PR separate the handling of an agreement-derived `sync()` from a non-agreement derived `sync()`. That simplify the logic of the `sync()` function, and defer the agreement-logic to be implemented in `syncCert()`.

Resolves #1529

### DIFF
--- a/catchup/service.go
+++ b/catchup/service.go
@@ -73,10 +73,11 @@ type Service struct {
 	auth            BlockAuthenticator
 	parallelBlocks  uint64
 	deadlineTimeout time.Duration
-	// catchpointWriting defines whether we've ran into a state where the ledger is currently busy writing the
-	// catchpoint file. If so, we want to pospone all the catchup process until the catchpoint file writing is complete,
+
+	// suspendForCatchpointWriting defines whether we've ran into a state where the ledger is currently busy writing the
+	// catchpoint file. If so, we want to suspend the catchup process until the catchpoint file writing is complete,
 	// and resume from there without stopping the catchup timer.
-	catchpointWriting bool
+	suspendForCatchpointWriting bool
 
 	// The channel gets closed when the initial sync is complete. This allows for other services to avoid
 	// the overhead of starting prematurely (before this node is caught-up and can validate messages for example).
@@ -394,7 +395,7 @@ func (s *Service) pipelinedFetch(seedLookback uint64) {
 			// could resume with the catchup.
 			if s.ledger.IsWritingCatchpointFile() {
 				s.log.Info("Catchup is stopping due to catchpoint file being written")
-				s.catchpointWriting = true
+				s.suspendForCatchpointWriting = true
 				return
 			}
 			completedRounds[round] = true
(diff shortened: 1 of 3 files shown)


algonautshant (commented):
node status reported Sync time  depends on the state of the catchup service. 
Will it be possible/helpful to report the state of catchup service in node status report?
I still think context is missing in the node status for the reported sync time.

catchup/service.go
```diff
@@ -73,6 +73,10 @@ type Service struct {
 	auth            BlockAuthenticator
 	parallelBlocks  uint64
 	deadlineTimeout time.Duration
+	// catchpointWriting defines whether we've ran into a state where the ledger is currently busy writing the
+	// catchpoint file. If so, we want to pospone all the catchup process until the catchpoint file writing is complete,
+	// and resume from there without stopping the catchup timer.
+	catchpointWriting bool
```
algonautshant, line 79: I want to propose couple of different names for catchpointWriting:
syncInterruptedWaitingCatchpointWriting
syncWaitingForCatchpointWriting

catchpointWriting is lacking context, and I found it difficult to understand the logic without this context.

algonautshant (approved):
Looks good.
I only have one comment on the catchpointWriting var name.

algonautshant (approved):

PR #1656, opened 2020-10-27
Set TRAVIS_BRANCH env var in jenkins mule agents

bricerisingalgorand:
Set TRAVIS_BRANCH env var in jenkins mule agents so that compute branch will be set correction. This needs to be done because the full git refs aren't pulled by the jobs. Reusing the TRAVIS_BRANCH env var so that we do not have to change any scripts, but we may want to consider revising later on.


egieseke (approved):
Looks good.

onetechnical (approved):

PR #1660, opened 2020-10-28
Avoid printing control characters for displaying assets

nicholasguoalgorand:
## Summary

Asset names can have control characters, which will be displayed to the terminal when run commands such as `goal asset info`. This allows malicious asset names to delete terminal input and / or print arbitrary input.

## Test Plan

Expect test

### DIFF
--- a/cmd/goal/asset.go
+++ b/cmd/goal/asset.go
@@ -603,7 +603,7 @@ var infoAssetCmd = &cobra.Command{
 
 		fmt.Printf("Asset ID:         %d\n", assetID)
 		fmt.Printf("Creator:          %s\n", params.Creator)
-		fmt.Printf("Asset name:       %s\n", params.AssetName)
+		reportInfof("Asset name:       %s\n", params.AssetName)
 		fmt.Printf("Unit name:        %s\n", params.UnitName)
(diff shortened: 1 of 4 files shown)


cmd/goal/asset.go
```diff
@@ -603,7 +603,7 @@ var infoAssetCmd = &cobra.Command{
 
 		fmt.Printf("Asset ID:         %d\n", assetID)
 		fmt.Printf("Creator:          %s\n", params.Creator)
-		fmt.Printf("Asset name:       %s\n", params.AssetName)
+		reportInfof("Asset name:       %s\n", params.AssetName)
```
tsachiherman, line 606: I think that `UnitName` would be as susceptible as `AssetName`, wouldn't it ?

tsachiherman (commented):
Could you also review `printAccountInfo` under the same context ?

cmd/goal/asset.go
nicholasguoalgorand: unit name is limited to 8 characters, would this make any difference?

tsachiherman: if these contain control characters, it might.

tsachiherman (approved):
Looks good to me, I'll wait for feedback from @jasonpaulos before merging this in.

jasonpaulos (approved):
Looks like this fixes the issue for `goal account info/list` too.

PR #1662, opened 2020-10-28
Default to branch name when branch parent fails

bricerisingalgorand:
Since jenkins doesn't pull the full git tree with refs, this will have compute_branch_network use the current branch instead of the branch parent when that rev parse fails


egieseke (approved):

onetechnical (approved):
Approve, but also see #1367

PR #1665, opened 2020-10-28
Fix computed network for packaging

onetechnical:
## Summary

The RPM packaging script wasn't pulling the right network due to hardcoding. Modified this and the deb script to use the passed branch to calculate the correct network. Also modified the compute branch network script to return betanet if passed rel/beta.

## Test Plan

Ensure packaging pipeline selects correct networks for packages.

### DIFF
--- a/scripts/compute_branch_network.sh
+++ b/scripts/compute_branch_network.sh
@@ -12,6 +12,9 @@ fi
 if [ "${BRANCH}" = "rel/stable" ]; then
     echo "testnet"
     exit 0
+elif [ "${BRANCH}" = "rel/beta" ]; then
+    echo "betanet"
+    exit 0
 fi
 
 #get parent of current branch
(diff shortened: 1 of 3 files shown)


btoll (approved):

scripts/compute_branch_network.sh
```diff
@@ -12,6 +12,9 @@ fi
 if [ "${BRANCH}" = "rel/stable" ]; then
     echo "testnet"
     exit 0
+elif [ "${BRANCH}" = "rel/beta" ]; then
```
bricerisingalgorand, line 15: oh, this makes our branch parent stuff unnecessary doesn't it?

bricerisingalgorand (approved):

egieseke (approved):
Looks OK.

scripts/compute_branch_network.sh
onetechnical: Not quite. This catches if rel/beta is passed as an argument, or if `./scripts/compute_branch.sh` actually returns rel/beta. If you do a `checkout -f`, then BRANCH is HEAD, and BRANCHPARENT will be the relevant parent. Hopefully.

PR #1667, opened 2020-10-29
Harden mule signing task

btoll:
## Summary

Improved the signing shell script in the following ways:
1. Fixed gpg warning by setting correct permissions in the docker container.
1. Instead of assuming the build status file is present, check for it first.

## Test Plan

`mule -f package-signing package-signing`

### DIFF
--- a/scripts/release/mule/sign/sign.sh
+++ b/scripts/release/mule/sign/sign.sh
@@ -17,6 +17,14 @@ PKG_DIR="./tmp/node_pkgs/$OS_TYPE/$ARCH_TYPE"
 SIGNING_KEY_ADDR=dev@algorand.com
 STATUSFILE="build_status_${CHANNEL}_${VERSION}"
 
+# It seems that copying/mounting the gpg dir from another machine can result in insecure
+# access privileges, so set the correct permissions to avoid the following warning:
+#
+#   gpg: WARNING: unsafe permissions on homedir '/root/.gnupg'
+#
+find /root/.gnupg -type d -exec chmod 700 {} \;
+find /root/.gnupg -type f -exec chmod 600 {} \;
+
 cd "$PKG_DIR"
 
 if [ -n "$S3_SOURCE" ]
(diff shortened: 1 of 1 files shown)


scripts/release/mule/sign/sign.sh
```diff
@@ -56,8 +64,12 @@ shasum -a 512 *.tar.gz *.deb *.rpm >> "$HASHFILE"
...
 gpg -u "$SIGNING_KEY_ADDR" --clearsign "$HASHFILE"
 
-gpg -u "$SIGNING_KEY_ADDR" --clearsign "$STATUSFILE"
-gzip -c "$STATUSFILE.asc" > "$STATUSFILE.asc.gz"
+# The status file *should* be present, but let's check to be safe.
+if [ -f "$STATUSFILE" ]
+then
+    gpg -u "$SIGNING_KEY_ADDR" --clearsign "$STATUSFILE"
+    gzip -c "$STATUSFILE.asc" > "$STATUSFILE.asc.gz"
+fi
```
egieseke, line 72: Is it considered an error if the status file does not exist?  Should this print an error message and return an error code if it is missing?

onetechnical: I think there's a separate bug where the status file is getting put in the wrong place. The file is not required and not published, so I think this is a workaround; alternatively we could fix the placement of the status file and nip this in the bud.

scripts/release/mule/sign/sign.sh
```diff
@@ -47,7 +55,7 @@ do
     gpg -u rpm@algorand.com --detach-sign "$file"
 done
 
-HASHFILE=hashes_${CHANNEL}_${OS_TYPE}_${ARCH_TYPE}_${VERSION}
+HASHFILE="hashes_${CHANNEL}_${OS_TYPE}_${ARCH_TYPE}_${VERSION}"
```
bricerisingalgorand, line 58: Will this sign all of the packages for different architectures in tmp/node_dir, or just the ones matching the host computers cpu architecture?

scripts/release/mule/sign/sign.sh
btoll: Yes, optimally the build status file is put back in its original location.

scripts/release/mule/sign/sign.sh
btoll: The latter.

scripts/release/mule/sign/sign.sh
btoll: I'm going to remove that conditional.

onetechnical (approved):
This looks good to me, but will require fixing build status location. #1668

egieseke (approved):
Looks good.

scripts/release/mule/sign/sign.sh
bricerisingalgorand: Cool, for this one we can run this for each os/arch we'll need. Later on maybe we can update this like the package scripts to iterate on all available files

bricerisingalgorand (approved):

PR #1669, opened 2020-10-29
Update mainnet-model for scale and distribution

tsachiherman:
## Summary

Updates the mainnet-model for scale ( increasing the number of staked nodes from 25 to 50 ) and
redistribute the relays, staked nodes and non-staked nodes according to mainnet distribution.

### DIFF
--- a/daemon/algod/api/client/restClient.go
+++ b/daemon/algod/api/client/restClient.go
@@ -57,25 +57,6 @@ const (
 	APIVersionV2 APIVersion = "v2"
 )
 
-// unauthorizedRequestError is generated when we receive 401 error from the server. This error includes the inner error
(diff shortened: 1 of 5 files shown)


brianolson:
Do we want to _duplicate_ mainnet or _model_ it? A model could be smaller but proportional.

bricerisingalgorand (approved):

tsachiherman:
> Do we want to _duplicate_ mainnet or _model_ it? A model could be smaller but proportional.

It is much smaller. We need certain size to ensure the number of uniquely generated proposal is roughly the same.
i.e. we have 20 proposals, so having 25 nodes might not be enough distribution given that some continents have much more stake than others.

tsachizehub (approved):
re-approving.

PR #1674, opened 2020-10-30
Support multiple OS and ARCH for signing

btoll:
## Summary

This supports multiple OS and ARCH by downloading assets recursively from s3.  It then will determine multiple OS by inspecting the direct child directories of `tmp/node_pkgs` to determine how to proceed.

Assets are separated by ARCH as usual, i.e. tmp/node_pkgs/OS/ARCH, but when uploaded to staging they all go in the same object (flat directory structure).

## Test Plan

`mule -f package-sign.yaml package-sign`
`mule -f package-upload.yaml package-upload`

### DIFF
--- a/scripts/release/mule/sign/sign.sh
+++ b/scripts/release/mule/sign/sign.sh
@@ -1,5 +1,7 @@
 #!/usr/bin/env bash
-# shellcheck disable=2035,2129
+# shellcheck disable=2035,2129,2162
+
+# TODO: This needs to be reworked a bit to support Darwin.
 
 set -exo pipefail
 
@@ -7,12 +9,11 @@ echo
 date "+build_release begin SIGN stage %Y%m%d_%H%M%S"
 echo
 
-ARCH_TYPE=$(./scripts/archtype.sh)
 OS_TYPE=$(./scripts/ostype.sh)
 VERSION=${VERSION:-$(./scripts/compute_build_number.sh -f)}
 BRANCH=${BRANCH:-$(./scripts/compute_branch.sh)}
 CHANNEL=${CHANNEL:-$(./scripts/compute_branch_channel.sh "$BRANCH")}
-PKG_DIR="./tmp/node_pkgs/$OS_TYPE/$ARCH_TYPE"
+PKG_DIR="./tmp/node_pkgs"
 SIGNING_KEY_ADDR=dev@algorand.com
 STATUSFILE="build_status_${CHANNEL}_${VERSION}"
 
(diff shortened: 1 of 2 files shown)


scripts/release/mule/sign/sign.sh
```diff
@@ -29,36 +30,53 @@ cd "$PKG_DIR"
...
+                do
+                    gpg -u "$SIGNING_KEY_ADDR" --detach-sign "$file"
+                done
+
+                for file in *.rpm
+                do
+                    gpg -u rpm@algorand.com --detach-sign "$file"
+                done
+
+                HASHFILE="hashes_${CHANNEL}_${OS_TYPE}_${ARCH_TYPE}_${VERSION}"
```
bricerisingalgorand, line 64: should ARCH_TYPE and OS_TYPE be ${os}_${arch}?

scripts/release/mule/sign/sign.sh
```diff
@@ -29,36 +30,53 @@ cd "$PKG_DIR"
...
-for file in *.rpm
-do
-    gpg -u rpm@algorand.com --detach-sign "$file"
+# https://unix.stackexchange.com/a/46259
+# Grab the directories directly underneath (max-depth 1) ./tmp/node_pkgs/ into a space-delimited string.
+# This will help us target `linux`, `darwin` and (possibly) `windows` build assets.
+# Note the surrounding parens turns the string created by `find` into an array.
+OS_TYPES=($(find . -mindepth 1 -maxdepth 1 -type d -printf '%f\n'))
+for os in "${OS_TYPES[@]}"; do
+    if [ "$os" = linux ]
```
bricerisingalgorand, line 44: Do we need to also sign the darwin builds?

btoll: Yes, that's my understanding.

scripts/release/mule/sign/sign.sh
btoll: Oh, yeah, I believe so.  Good call.

scripts/release/mule/sign/sign.sh
onetechnical: Eventually, but we aren't building them yet. Because of the ARCHS loop below, it'd probably need to be a separate block anyway.

scripts/release/mule/sign/sign.sh
onetechnical: ^^^ I think this is the case

onetechnical (changes requested):

scripts/release/mule/sign/sign.sh
btoll: Fixed.

bricerisingalgorand (approved):

egieseke (approved):
Looks OK.

onetechnical (approved):

bricerisingalgorand (approved):

egieseke (approved):
Looks good.

PR #1666, opened 2020-10-29
Change rpm package file to match rename

btoll:
## Summary

Change rpm package file to match rename.

## Test Plan

`mule -f package-sign package-sign`

PR #1675, opened 2020-10-30
Revert "Change rpm package file to match rename (#1666)"

onetechnical:
This reverts commit 450fd8c7984528070fc412aa322af304b54c2a65.

## Summary

While we need to change the arch type for other reasons, we do not need to rename the RPM. A separate commit is reverting the name change, so this is reverting the related changes.

## Test Plan

Standard release testing.


bricerisingalgorand (approved):

egieseke (approved):
Looks good.

PR #1678, opened 2020-10-30
Update goal error message for unauthorized access

tsachiherman:
## Summary

Update error message to be more informative when the libgoal is failing to connect to a node with 401 error.
New error would look like this
```
Cannot contact Algorand node: Unauthorized request to `http://127.0.0.1:5160/v2/status` when using token `d1187151e420f660573403f2068c0948659e4eb9f4c9ecc17d38e72665179d1b` : {"message":"Invalid API Token"}
```

instead of

```
Cannot contact Algorand node: HTTP 401 : {"message":"Invalid API Token"}
.
```

## Test Plan

Update existing test and add a new one.

### DIFF
--- a/daemon/algod/api/client/restClient.go
+++ b/daemon/algod/api/client/restClient.go
@@ -117,7 +117,7 @@ func filterASCII(unfilteredString string) (filteredString string) {
 // extractError checks if the response signifies an error (for now, StatusCode != 200 or StatusCode != 201).
 // If so, it returns the error.
 // Otherwise, it returns nil.
-func (client RestClient) extractError(resp *http.Response) error {
+func extractError(resp *http.Response) error {
 	if resp.StatusCode == 200 || resp.StatusCode == 201 {
 		return nil
 	}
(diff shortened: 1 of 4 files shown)


daemon/algod/api/client/restClient.go
```diff
@@ -64,6 +64,19 @@ var rawRequestPaths = map[string]bool{
...
+// as well as the likely parameters that caused the issue.
+type unauthorizedRequestError struct {
+	errorString string
+	apiToken    string
+	url         string
+}
+
+// Error format an error string for the unauthorizedRequestError error.
+func (e unauthorizedRequestError) Error() string {
+	return fmt.Sprintf("Unauthorized request to `%s` when using token `%s` : %s", e.url, e.apiToken, e.errorString)
```
algonautshant, line 77: Why not embed filterASCII in this function?

algonautshant (approved):
Looks good. 
Just wondering if unauthorizedRequestError Error function can call filterASCII just in case it gets unfiltered string.

daemon/algod/api/client/restClient.go
tsachiherman: I wanted to use the filtered string for the other errors as well. We shouldnt have errors like that with unfiltered content since its being written to the console.

PR #1693, opened 2020-11-10
Dynamically adjust catchpoint file serving timeouts based on file size

tsachiherman:
## Summary

The catchpoint file serving used to have a hard-coded timeout for all catchpoint file requests. This is sub-optimal, since large catchpoint files would be served over longer period of time.

This PR dynamically adjust the timeout based on the file size, taking advantage of the sizing information we already store in the database, and passing it upstream.

## Test Plan

Unit tests updated.

### DIFF
--- a/rpcs/ledgerService.go
+++ b/rpcs/ledgerService.go
@@ -47,14 +47,11 @@ const (
 	// e.g. .Handle(LedgerServiceLedgerPath, &ls)
 	LedgerServiceLedgerPath = "/v{version:[0-9.]+}/{genesisID}/ledger/{round:[0-9a-z]+}"
 
-	// maxCatchpointFileSize is a rough estimate for the worst-case scenario we're going to have of all the accounts data per a single catchpoint file chunk.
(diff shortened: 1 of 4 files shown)


rpcs/ledgerService.go
```diff
@@ -199,6 +191,20 @@ func (ls *LedgerService) ServeHTTP(response http.ResponseWriter, request *http.R
 		}
 	}
 	defer cs.Close()
+	if conn := ls.net.GetHTTPRequestConnection(request); conn != nil {
+		maxCatchpointFileWritingDuration := 2 * time.Minute
```
algonautshant, line 195: Is this giving an additional 2 minutes?

algonautshant (approved):
Looks good.

rpcs/ledgerService.go
tsachiherman: Yes (exactly as before). The intent here is that regardless of the file size, we would always have some "connection-establishing" timing, and add from there.

algorandskiy (approved):

PR #1694, opened 2020-11-11
Add block hash and transaction ids in /v2/block

mxmauro:
`/v1/block` provides the block hash and transaction id but `/v2/blocks` does not.

Although it is possible to calculate them, it involves decoding and re-encoding portions of the block and doing some math that can be slow on certain languages.

In addition, the node knows all protocol versions and when to add genesis hash and ids by checking the `hgi` parameter.

### DIFF
--- a/daemon/algod/api/server/v2/handlers.go
+++ b/daemon/algod/api/server/v2/handlers.go
@@ -167,11 +167,28 @@ func (v2 *Handlers) GetBlock(ctx echo.Context, round uint64, params generated.Ge
 		return internalError(ctx, err, errFailedLookingUpLedger, v2.Log)
 	}
 
+	// Calculate transaction IDs
+	txIds := make([]string, 0)
+	for _, txib := range block.Payset {
+		var tx transactions.SignedTxn
+
+		tx, _, err = block.DecodeSignedTxn(txib)
+		if err != nil {
+			return internalError(ctx, err, err.Error(), v2.Log)
+		}
+
+		txIds = append(txIds, tx.Txn.ID().String())
+	}
+
 	// Encoding wasn't working well without embedding "real" objects.
 	response := struct {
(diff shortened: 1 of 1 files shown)


algorotem:
Hi Mauro,

Thanks for the PR! 

Unfortunately, I cannot accept it. The block's `hash` and transactions' `txid` were not included intentionally. During the design of `v2`, we have decided to provide the data "as is" and therefore txids and hashes are not part of it.
Those fields, including many additional others, can be retrieved via Indexer.

mxmauro:
Hi @algorotem no problem. My only concern is the TXID. As you know it is based on the protocol version and `hgi` parameter.

Based on the live networks, also the indexer has to do some checks depending if the block's protocol version is 16. I believe the addition of `RequireGenesisHash=true` that is part of txid calculation won't change in the future, right?

Kind regards.

pbennett:
I'm curious - so are you saying that even for the last 1000 blocks with a pruning node, the ONLY way to get certain data via an API and SDK call is to also run an indexer, which then also requires a full node?  
I should be able to run a pruning node and through the API get *all* data easily with the API which I'd see on any block explorer.  To force a full node AND an indexer (adding additional 50% to already significant storage requiremets) just to be able to fully extract a block is a big step backwards in the v2 api.

algorotem:
@mxmauro - Not in the foreseeable future. Changes like that require a protocol change and are very well thought. If any proposal like that would happen, you'll know about it well ahead of time and be able to prepare. But again, I truly don't see why it would be changed. 

@pbennett - As always, thanks for your comments. Nothing to be concern about, you don't need a full node nor an indexer to get this information for the last 1000 blocks. These are easily accessible through the API and SDKs will take care of computing the TXIDs and Blockhashes for you.

mxmauro:
Excellent @algorotem , thanks for the feedback.

PR #1697, opened 2020-11-13
Ordered account insertion into merkle trie (I)

tsachiherman:
## Summary

Inserting ordered hashes into the merkle trie improve its insertion speed dramatically. The goal of this change is to modify the insertion that doing so that these would always be in-order. 

We currently have two such use cases : 
- [x] Rebuilding a local merkle trie from the accounts themselves
- [ ]  Node catching up using the fast catchup. This would be handled on the server side by ordering the accounts on the catchpoint file.

This PR addresses the first part, using an ordered accounts hashes to be inserted into the trie.
The new `orderedAccountsIter` is intended to replace the `encodedAccountsBatchIter`, but that would only be possible on the subsequent PR.

## Performance Plan

I added a benchmark to test the gains:

| Number of accounts   |     Before the change      |  With the change | Performance gains |
|:----------:|:-------------:|:------:|:------:|
| 6,000,000 |  293 seconds | 80 seconds | 266% |

The 6M accounts were chosen since it's roughly the number of accounts we currently have on mainnet.

## Test Plan

Existing tests already cover the functional portion of this PR.


ledger/accountdb.go
```diff
@@ -1131,3 +1131,205 @@ func (iterator *encodedAccountsBatchIter) Close() {
...
+		}
+		// prepare the insert statement into the temporary table
+		iterator.insertStmt, err = iterator.tx.PrepareContext(ctx, "INSERT INTO accountsiteratorhashes(address, hash) VALUES(?, ?)")
+		if err != nil {
+			return
+		}
+		iterator.step = 3
+		return
+	}
+	if iterator.step == 3 {
```
nicholasguoalgorand, line 1205: add comment here briefly explaining step 3

ledger/accountdb.go
```diff
@@ -1131,3 +1131,205 @@ func (iterator *encodedAccountsBatchIter) Close() {
...
+		}
+
+		if err != nil {
+			return
+		}
+		iterator.step = 6
+		return
+	}
+
+	if iterator.step == 6 {
```
nicholasguoalgorand, line 1275: add comment here briefly explaining step 6

ledger/accountdb.go
```diff
@@ -1131,3 +1131,205 @@ func (iterator *encodedAccountsBatchIter) Close() {
...
 	}
 }
+
+// orderedAccountsIter allows us to iterate over the accounts addresses in the order of the account hashes.
+type orderedAccountsIter struct {
+	step         int
+	rows         *sql.Rows
+	tx           *sql.Tx
+	accountCount int
+	accountData  bool
```
nicholasguoalgorand, line 1141: can we rename "accountData" to something more descriptive like "includeAccountData" or whatever the appropriate field name would be?

nicholasguoalgorand (commented):
Added some minor comments. Question: how efficient is using sql's orderby / indexing as a sorting algorithm?

ledger/accountdb.go
tsachiherman: I changes the number of constants with description.

ledger/accountdb.go
tsachiherman: same

ledger/accountdb.go
tsachiherman: done.

tsachiherman:
> Added some minor comments. Question: how efficient is using sql's orderby / indexing as a sorting algorithm?

I'm not sure that we have a good alternative. We don't have enough memory to load it all in anyway.

ledger/accountdb.go
```diff
@@ -1131,3 +1131,232 @@ func (iterator *encodedAccountsBatchIter) Close() {
...
+			}
+
+			count++
+			if count == iterator.accountCount {
+				// we're done with this iteration.
+				acct = make([]accountAddressHashData, count, count)
+				return
+			}
+		}
+		acct = make([]accountAddressHashData, count, count)
```
algonautshant, line 1269: Why wasting time allocating an empty array here, which seems to me is never used?

nicholasguoalgorand (approved):
Updated changes look good

algonautshant (commented):
Looks good. 
I feel like this can be simplified. So I need to understand why all these cases are needed.

ledger/accountdb.go
```diff
@@ -1131,3 +1131,232 @@ func (iterator *encodedAccountsBatchIter) Close() {
...
+
+// orderedAccountsIterStep is used to by orderedAccountsIter to define the current step
+//msgp:ignore orderedAccountsIterStep
+type orderedAccountsIterStep int
+
+const (
+	// startup step
+	oaiStepStartup = orderedAccountsIterStep(0)
+	// delete old ordering table if we have any leftover from previous invocation
+	oaiStepDeleteOldOrderingTable = orderedAccountsIterStep(0)
```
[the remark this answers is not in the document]
tsachiherman: It's just so that we can given a meaningful name to this stage.

ledger/accountdb.go
```diff
@@ -1131,3 +1131,232 @@ func (iterator *encodedAccountsBatchIter) Close() {
...
+// Next returns an array containing the account address, hash and potentially data
+// the Next function works in multiple processing stages, where it first processs the current accounts and order them
+// followed by returning the ordered accounts. In the first phase, it would return empty accountAddressHashData array
+// whose size matches with the data that was processed, and set the ordered to false. On the second phase, the acct
+// would contain valid data ( and optionally the account data as well, if was asked in makeOrderedAccountsIter) and
+// would set the ordered to true. If err is sql.ErrNoRows it means that the iterator have completed it's work and no further
+// accounts exists. Otherwise, the caller is expected to keep calling "Next" to retrieve the next set of accounts
+// ( or let the Next function make some progress toward that goal )
+func (iterator *orderedAccountsIter) Next(ctx context.Context) (acct []accountAddressHashData, ordered bool, err error) {
+	if iterator.step == oaiStepDeleteOldOrderingTable {
```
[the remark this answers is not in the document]
tsachiherman: The breaking into stages is designed so that the entire process can be paused at any moment. To accomplish that, I broke that into granular sql statements.

ledger/accountdb.go
```diff
@@ -1131,3 +1131,232 @@ func (iterator *encodedAccountsBatchIter) Close() {
...
+			hash := accountHashBuilder(addr, accountData, buf)
+			_, err = iterator.insertStmt.ExecContext(ctx, addrbuf, hash)
+			if err != nil {
+				return
+			}
+
+			count++
+			if count == iterator.accountCount {
+				// we're done with this iteration.
+				acct = make([]accountAddressHashData, count, count)
```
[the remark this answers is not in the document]
tsachiherman: The consumer needs the sizing information in order to print logging information; I could have added another return value, but that would make it even harder to read..

ledger/accountdb.go
```diff
@@ -1131,3 +1131,232 @@ func (iterator *encodedAccountsBatchIter) Close() {
...
+		_, err = iterator.tx.ExecContext(ctx, "CREATE INDEX accountsiteratorhashesidx ON accountsiteratorhashes(hash)")
+		if err != nil {
+			return
+		}
+		iterator.step = oaiStepSelectFromOrderedTable
+		return
+	}
+	if iterator.step == oaiStepSelectFromOrderedTable {
+		// select the data from the ordered table
+		if iterator.fetchAccountData {
```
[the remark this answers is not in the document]
tsachiherman: You can't see it in this PR, but the next one is going to use this in the catchpoint writing, where we need to go to sleep if we've exceeded the timeout.

ledger/acctupdates.go
```diff
@@ -1119,56 +1119,65 @@ func (au *accountUpdates) accountsInitialize(ctx context.Context, tx *sql.Tx) (b
...
 
-			if pendingAccounts >= trieRebuildCommitFrequency {
-				// this trie Evict will commit using the current transaction.
-				// if anything goes wrong, it will still get rolled back.
-				_, err = trie.Evict(true)
-				if err != nil {
-					return 0, fmt.Errorf("accountsInitialize was unable to commit changes to trie: %v", err)
-				}
-				pendingAccounts = 0
+			if !wasOrdered && ordered {
```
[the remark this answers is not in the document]
tsachiherman: The reason to avoid passing just the count was to reduce number of return parameters. The number of parameters becoming too "messy". however, I think that we could change it into
```
func (iterator *orderedAccountsIter) Next(ctx context.Context) (acct []accountAddressHashData, processedRecords int, err error)
```
where:
acct would be populated only on the result.
processedRecords would be populated when we build the index.
err would retain the same logic as today.

is that sounds like a good solution?

ledger/accountdb.go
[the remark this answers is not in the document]
algonautshant: but here we are making an allocation right? Depending on the magnitude of 'count', this could be significant.

ledger/acctupdates.go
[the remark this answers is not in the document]
algonautshant: Yes. this is very good. 
The current implementation is also confusing, because, the reader expects it to be used somewhere, but cannot find where.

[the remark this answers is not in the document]
tsachiherman: ok, I applied the changes.

ledger/accountdb.go
```diff
@@ -1293,6 +1297,7 @@ func (iterator *orderedAccountsIter) Next(ctx context.Context) (acct []accountAd
 		}
 
 		if err != nil {
+			iterator.Close(ctx)
 			return
 		}
 		iterator.step = oaiStepIterateOverOrderedTable
```
algonautshant, line 1303: Don't we need to close the iterator on line 1325?
```
			if len(addrbuf) != len(addr) {
				err = fmt.Errorf("Account DB address length mismatch: %d != %d", len(addrbuf), len(addr))
				return
			}
```

algonautshant (approved):
Looks great.
Thanks for the changes. 
Just one comment about closing the iterator.

ledger/accountdb.go
tsachiherman: yes, we do. I'll add that!

ledger/accountdb.go
```diff
@@ -1131,3 +1131,236 @@ func (iterator *encodedAccountsBatchIter) Close() {
...
+
+// accountAddressHashData is used by Next to return a single account address, hash and account data.
+type accountAddressHashData struct {
+	address            basics.Address
+	digest             []byte
+	encodedAccountData []byte
+}
+
+// Next returns an array containing the account address, hash and potentially data
+// the Next function works in multiple processing stages, where it first processs the current accounts and order them
```
algorandskiy, line 1191: processs -> processes

algonautshant (approved):

algorandskiy (approved):
LGTM, not sure although why do we need such level of granularity/steps within the `Next()` function

ledger/accountdb.go
tsachiherman: will update on subsequent PR. thanks!

PR #1698, opened 2020-11-13
Ordered account insertion into merkle trie (II)

tsachiherman:
## Summary

Inserting ordered hashes into the merkle trie improve its insertion speed dramatically. The goal of this change is to modify the insertion that doing so that these would always be in-order. 

We currently have two such use cases : 
- [x] Rebuilding a local merkle trie from the accounts themselves
- [x]  Node catching up using the fast catchup. This would be handled on the server side by ordering the accounts on the catchpoint file.

This PR addresses the second part, using the `orderedAccountsIter` to insert the accounts data to the tarball file.

## Performance

I added a benchmark to test the gains:

| Number of accounts   |     Before the change      |  With the change | Performance loss |
|:----------:|:-------------:|:------:|:------:|
| 6,000,000 |  22 seconds | 59 seconds | x2.68 |

The 6M accounts were chosen since it's roughly the number of accounts we currently have on mainnet.
The performance loss here is expected, but it would get gained on the client side allowing faster catchups.

## Test Plan

Existing tests already cover the functional portion of this PR.


tsachiherman:
closing this PR as I'd like to research this one further on.

ledger/accountdb.go
```diff
@@ -1208,7 +1143,11 @@ func (iterator *orderedAccountsIter) Next(ctx context.Context) (acct []accountAd
 	}
 	if iterator.step == oaiStepCreateOrderingTable {
 		// create the temporary table
-		_, err = iterator.tx.ExecContext(ctx, "CREATE TABLE accountsiteratorhashes(address blob, hash blob)")
+		if iterator.fetchAccountData {
+			_, err = iterator.tx.ExecContext(ctx, "CREATE TABLE accountsiteratorhashes(accountbase_rawid INTEGER, hash blob)")
```
algonautshant, line 1147: Have you considered adding accountbase.data column to accountsiteratorhashes? 
It looks like the accountbase.data is being read once during the select in oaiStepInsertAccountData, and once during the join in oaiStepSelectFromOrderedTable. 

Copying them over to accountsiteratorhashes will avoid the join, as well as scanning all pages with data from accountbase, most of which will not be in the result anyway.

tsachiherman: Yes, I did. Unfortunately adding column means that we'll pay a higher price on deletion/insertion to the table.
I was trying to avoid doing that if possible.
regardless, I think that this PR is going to be shelved. If you could review https://github.com/algorand/go-algorand/pull/1712 PR instead, it would be great. ( it's a subset of this one )

PR #1700, opened 2020-11-14
Add app state read-write into pingpong

algorandskiy:
## Summary

Fix application program generation so that it writes local and global states. Also allowed more that `MaxAppsCreated` application creation across multiple accounts.
Three new cli options added:
1.  `numappoptin` sets how many apps every account opts in
2. `appproggk` num of global state writes
3. `appproglk` num of local state writes

Initialization might take longer since lots of opt ins happens. To offset this opt-in and app-create txn are sent as txn groups.

## Test Plan

Tested manually
```
pingpong run -d $NETWORKDIR/Node --numapp 100 --appprogops 300 --numappoptin 10 --appproggk 10 --appproglk 10 --tps 1000 --rest 0  --refresh 1 --numaccounts 500 --duration 180 --minaccount 34000000 > /dev/null &
```


shared/pingpong/pingpong.go
```diff
@@ -540,7 +547,16 @@ func constructTxn(from, to string, fee, amt, aidx uint64, client libgoal.Client,
 	}
 
 	if cfg.NumApp > 0 { // Construct app transaction
-		txn, err = client.MakeUnsignedAppNoOpTx(aidx, nil, nil, nil, nil)
+		// generate random assetID or appId if we send asset/app txns
+		accounts := cinfo.OptIns[aidx]
+		if len(accounts) > 0 {
+			limit := len(accounts) - 4
+			if limit > 0 {
+				start := rand.Intn(limit)
```
tsachiherman, line 555: for this test it would probably be good enough, but note that 4 consecutive accounts won't give us random access testing. maybe use rand.Perm and pick the first four ?

shared/pingpong/accounts.go
```diff
@@ -450,73 +563,136 @@ func prepareApps(accounts map[string]uint64, client libgoal.Client, cfg PpConfig
...
+			appParams[k] = v
+		}
+	}
+
+	// time to opt in to these apps
+	if cfg.NumAppOptIn > 0 {
+		optIns = make(map[uint64][]string)
+		for addr := range accounts {
+			var txgroup []transactions.Transaction
+			addrOptedIn := make(map[uint64]bool)
```
tsachiherman, line 646: consider using rand.Perm, and remove the entry from the indexing array once you're done with it.

tsachiherman (changes requested):
looks good, few small requests.

shared/pingpong/accounts.go
algorandskiy: done

shared/pingpong/pingpong.go
algorandskiy: done

egieseke (approved):
The teal testing has not been very reliable due to flooding of the transaction queue during the account setup.  Were you able to resolve that?

tsachiherman (dismissed):

shared/pingpong/accounts.go
```diff
@@ -357,47 +360,70 @@ func genBigNoOpAndBigHashes(numOps uint32, numHashes uint32, hashSize string) []
...
+	writeLocBlock := `
+		int 1           // [1, n]
+		+               // [1+n]
+		dup             // [1+n, 1+n]
+		dup             // [1+n, 1+n, 1+n]
+		store 0         // [1+n, 1+n]
+		txn NumAccounts // [N, 1+n, 1+n]
+		int 1           // [1, N, 1+n, 1+n]
+		-               // [N-1, 1+n, 1+n],  exclude sender
+		%               // [A, 1+n], A = 1+n mod N
```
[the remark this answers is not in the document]
algorandskiy: ah, you've meant a comment. Right, fixed, thanks!

tsachiherman (approved):

PR #1701, opened 2020-11-15
Static linking in Windows executables

mxmauro:
Most of Windows binaries depends on the `MINGW` shared `pthreads` library. This addition copies it to the output directory on Windows builds.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 UNAME := $(shell uname)
-ifneq (, $(findstring MINGW,$(UNAME)))
+ifneq (,$(findstring MINGW,$(UNAME)))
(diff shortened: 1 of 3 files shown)


tsachiherman:
@mxmauro , I'm a big opponent of copying DLL files, especially those which aren't yours.
Instead of adding this copy, why don't we compile the binary with static linking ?
This would not only solve this issue but also would simplify the creation of a distribution package.

mxmauro:
Hi @tsachiherman me too. Believe me I spent too much time with Karalabe's HID dependency (https://github.com/karalabe/hid).

But now I was able to do static linking. No idea what to say.

tsachiherman (approved):
looks good.

PR #1702, opened 2020-11-15
Use go tooling to get go version string in go.mod

btoll:
Looks like this approach may not work b/c go modules are disabled in travis.  All of the builds failed there.  Is this intentional or is it a default setting?
```
go: modules disabled inside GOPATH/src by GO111MODULE=auto; see 'go help modules'
```

## Summary

Use go tooling to get go version string in `go.mod` rather than depending on a regular expression.

## Test Plan

`./scripts/get_golang_version.sh`


tsachiherman (commented):
I generally like the idea, two points:
1. The script is called get_golang_version.sh, but it's doing more than just that.
    It also verify that the go.mod would match the hard-coded version. I think that we might
    want to either rename that to "check_golang_version" or maybe break the two
    functionalities apart.
2. We need to make sure that it would always work. For instance, if we have go 1.10 installed ( which is the default supported go version on ubuntu ), we want to be able to tell that the go version is too old rather than generating some error here.

scripts/check_golang_version.sh
```diff
@@ -0,0 +1,19 @@
...
+set -eo pipefail
+
+GOLANG_VERSION=$(./scripts/get_golang_version.sh)
+
+# Check to make sure that it matches what is specified in go.mod
+
+GOMOD_VERSION=$(go mod edit -json | jq -r ".Go")
+TRIMMED_GOLANG_VERSION=$(awk -F. '{ print $1 "." $2 }' <<< "${GOLANG_VERSION}")
+
+if ! [[ "${GOLANG_VERSION}" == "${GOMOD_VERSION}" || "${TRIMMED_GOLANG_VERSION}" == "${GOMOD_VERSION}" ]]
```
brianolson, line 12: Do we want == or >=
Is 1.14.8 allowed? IS 1.15.x allowed?

btoll: My understanding was that thought was to pin this to a precise version.

( Note that most of this file is existing code. )

brianolson: For automated systems maybe but for human developers we should certainly set a lower bound and allow newer things to be okay

scripts/configure_dev.sh
```diff
@@ -1,6 +1,8 @@
 #!/usr/bin/env bash
 set -e
 
+./scripts/check_golang_version.sh
```
brianolson, line 4: This is the human case where we should be flexible. I think it's normal for a new developer to run configure_dev.sh. The travis cases pinning to a specific version are just a check that travis is actually doing what we said to, right?

btoll: Yes, that's a good point.

There should be at least a minimum version check.  For instance, if I try to install on ubuntu:18.04 using the version of go that I've downloaded from `apt`, I'm going to be working with go1.10 and that causes other failures when running `make`.

What do you and @tsachiherman feel would be a good minimum version?

scripts/check_golang_version.sh
btoll: Ok, that's a good point.

scripts/configure_dev.sh
tsachiherman: The minimal value would be one that is capable of building our source code. So it has to be above 0.13. I dont know if we require 0.14 or not.

scripts/check_golang_version.sh
tsachiherman: The pinning should be for our build system.. not for other cases.
I think that it's legit if we fail to build with a newer version due to breaking changes.

scripts/configure_dev.sh
brianolson: Going forward >= 0.14 would be good. golang.org is currently only releasing security patches to 0.14 and 0.15. I think this signals that everyone should migrate off 0.13

scripts/check_golang_version.sh
onetechnical: I think we would prefer to pin to the major version specified in go.mod (1.14 currently), no? I don't think this should be hardcoded to 1.12.

onetechnical (changes requested):
Generally I like the approach, but don't like the 1.12 hardcoding.

scripts/configure_dev.sh
btoll: I will make it so :)

scripts/check_golang_version.sh
btoll: > I think we would prefer to pin to the major version specified in go.mod (1.14 currently), no? I don't think this should be hardcoded to 1.12.

The hardcoded version (1.12) is the first version of go that fully supports the go modules command that we use to get the pinned version from `go.mod`, so we need to check this first to disambiguate that failure from the comparison check and possible error later in the script.

scripts/get_golang_version.sh
```diff
@@ -1,16 +1,18 @@
...
+MIN=1.14
+GO_MOD_SUPPORT=1.12
 
-if ! [[ ${GOLANG_VERSION} == "${GOMOD_VERSION}" || ${TRIMMED_GOLANG_VERSION} == "${GOMOD_VERSION}" ]]
+if [ "$1" = all ]
 then
-    echo "go version mismatch, go mod version ${GOMOD_VERSION} does not match required version ${GOLANG_VERSION}"
-    exit 1;
+    echo $BUILD $MIN $GO_MOD_SUPPORT
+elif [ "$1" = dev ]
```
tsachiherman, line 12: it's a bit weird that you're comparing "$1" to dev rather than "dev".

btoll: I'm not really following.  What's weird about it?

onetechnical: I think just that it's not explicitly stating it's a string. It does look a little strange, but am guessing it works.

scripts/check_golang_version.sh
```diff
@@ -0,0 +1,41 @@
+#!/usr/bin/env bash
+# shellcheck disable=2072
+
```
onetechnical, line 3: A comment here explaining the purpose of this script would be good...indicate the default case, and why it would be run with 'dev'.

scripts/check_golang_version.sh
```diff
@@ -0,0 +1,41 @@
...
+    then
+        echo "[$0] WARNING: The version of go on the system (${SYSTEM_GOLANG_VERSION}) is below the recommended version (${MIN_VERSION}) and therefore may not build correctly."
+        echo "[$0] Please update to at least ${MIN_VERSION}"
+    fi
+else
+    # Check to make sure that it matches what is specified in go.mod
+    GOMOD_VERSION=$(go mod edit -json | jq -r ".Go")
+    TRIMMED_BUILD_VERSION=$(awk -F. '{ print $1 "." $2 }' <<< "${BUILD_VERSION}")
+
+    if ! [[ "${BUILD_VERSION}" == "${GOMOD_VERSION}" || "${TRIMMED_BUILD_VERSION}" == "${GOMOD_VERSION}" ]]
```
onetechnical, line 33: Pre-existing code, I know, but why not just always compare to $TRIMMED_BUILD_VERSION here?

scripts/release/build/build_algod_docker.sh
```diff
@@ -38,9 +38,9 @@ DOCKER_IMAGE="algorand/algod_${CHANNEL_VERSION}:${DOCKER_TAG}"
 RESULT_DIR="${HOME}/node_pkg/"
 DOCKERFILE="${HOME}/go/src/github.com/algorand/go-algorand/docker/build/algod.Dockerfile"
 START_ALGOD_FILE="start_algod_docker.sh"
-# Use go version specified by get_golang_version.sh
+# Use go build version specified by get_golang_version.sh
 pushd "${HOME}/go/src/github.com/algorand/go-algorand"
-if ! GOLANG_VERSION=$("./scripts/get_golang_version.sh")
+if ! GOLANG_VERSION=$("./scripts/check_golang_version.sh")
```
onetechnical, line 43: Semantically, I'm a little bit bothered by this. We're defining the version of golang we want with a script named 'check_golang_version.sh', not 'get_golang_version.sh'. I think splitting the check and get functionalities are fine, but it makes uses such as this cognitively difficult.

scripts/get_golang_version.sh
```diff
@@ -1,16 +1,18 @@
 #!/usr/bin/env bash
-# Required GOLANG Version for project specified here
-GOLANG_VERSION="1.14.7"
 
-# Check to make sure that it matches what is specified in go.mod
+# Required GOLANG versions for project specified here.
```
onetechnical, line 3: A little more doc here would be helpful to explain the reason for the 'all', 'dev', and default outputs.

onetechnical (changes requested):
Structurally, I'm a little bothered by how we're using the check_golang_version.sh calls; they don't quite feel semantically correct. However, I'd begrudgingly approve, if you add a little documentation around the changes, as we're splitting functionality into two scripts with more interfaces.

scripts/get_golang_version.sh
btoll: I'm in the habit of not using the double-quotes when there are no shell parameters to expand.

scripts/check_golang_version.sh
btoll: I'll add comments.

scripts/check_golang_version.sh
btoll: In case the pinned version in `go.mod` has a patch version.

scripts/release/build/build_algod_docker.sh
btoll: I agree, I've always been bothered by this haha
I'll look into making it less cognitively jarring.

scripts/get_golang_version.sh
btoll: Will add comments.

scripts/release/build/build_algod_docker.sh
btoll: I just pushed up a commit that changes the checks in all the scripts that use it.  The new idiom looks like this:
```
if ! ./scripts/check_golang_version.sh
then
    exit 1
fi
# Get the go build version.
GOLANG_VERSION=$(./scripts/get_golang_version.sh)
```
It's more expensive, but I think the benefit is that it reads a lot better.  The previous way had bugged me because the old version of `get_golang_version.sh` was pulling double-duty in a not-so-obvious way from a cursory read of the code.

For example, it wasn't just "getting" the golang version, it was also checking it.  And if the check failed the calling script would echo the return value captured in the `GOLANG_VERSION` variable before exiting.  That was just weird.

scripts/travis/build.sh
```diff
@@ -34,16 +34,17 @@ SCRIPTPATH="$( cd "$(dirname "$0")" ; pwd -P )"
...
-    echo "${GOLANG_VERSION}"
     exit 1
 fi
+# Get the go build version.
+GOLANG_VERSION=$(./scripts/get_golang_version.sh)
 
 curl -sL -o ~/gimme https://raw.githubusercontent.com/travis-ci/gimme/master/gimme
 chmod +x ~/gimme
-eval $(~/gimme "${GOLANG_VERSION}")
+eval "$(~/gimme "${GOLANG_VERSION}")"
```
btoll, line 47: This satisfies shellchecker.

scripts/check_golang_version.sh
onetechnical: I don't think that's allowed. That's how we had it initially to pin the version, but it didn't work, which is why we stick with the minor version.

scripts/release/build/build_algod_docker.sh
onetechnical: I agree - longer code in this case makes more sense, and it's not _that_ much more expensive, as it's a build env execution, not a run time execution.

scripts/check_golang_version.sh
btoll: I'm going to push a commit that compares the strings.  I had posed this idea when this code was initially merged, and I think it's a succinct and safe way to verify build version.

For example, as long as we compare either string to the beginning of the other, we shouldn't get any false positives.

```
[[ 1.14 =~ ^1.14 ]] # true
[[ 1.14.7 =~ ^1.14 ]] # true
[[ 1.14.7 =~ ^14.7 ]] # false
```

scripts/get_golang_version.sh
btoll: I added the double-quotes.

onetechnical:
Shifting back to in progress column, as it appears to be failing tests

onetechnical (approved):

scripts/travis/build.sh
bricerisingalgorand: Are you able to use double quotes again like that inside of `$()`?

bricerisingalgorand (approved):
This looks good to me as long as the quoting I commented on works as intended. I would've thought bash would interpret that as 3 separate strings so the variable eval would be unquoted (which is probably fine in the context, but I'm surprised shellcheck likes that)

scripts/travis/build.sh
btoll: Yes.  The outer quotes would be expanding anything returned by the command, while the inner quotes are still interpolating the variable.

PR #1704, opened 2020-11-16
Specify swagger version v0.25.0 to fix build.

egieseke:
## Summary

Fix a build problem by specifying version 0.25.0 of the go swagger module.  The latest released version is stricter and breaking the build.  

## Test Plan

Tested with scripts/configure_dev-deps.sh to ensure that the correct version is installed.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,7 @@ module github.com/algorand/go-algorand
 go 1.14
 
 require (
-        github.com/go-swagger/go-swagger/cmd/swagger v0.25.0
+	github.com/go-swagger/go-swagger/cmd/swagger v0.25.0
 	github.com/algorand/go-codec/codec v0.0.0-20190507210007-269d70b6135d
 	github.com/algorand/go-deadlock v0.2.1
 	github.com/algorand/msgp v1.1.45
--- a/go.mod
+++ b/go.mod
@@ -3,6 +3,7 @@ module github.com/algorand/go-algorand
 go 1.14
 
 require (
+        github.com/go-swagger/go-swagger/cmd/swagger v0.25.0
 	github.com/algorand/go-codec/codec v0.0.0-20190507210007-269d70b6135d
 	github.com/algorand/go-deadlock v0.2.1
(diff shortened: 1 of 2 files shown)


tsachiherman (commented):
It would be much more straight forward to add the swagger in the `go.mod` file, wouldn't it ?

egieseke:
> It would be much more straight forward to add the swagger in the `go.mod` file, wouldn't it ?

I tried that first, but that caused problems when running make 
go: github.com/go-swagger/go-swagger/cmd/swagger@v0.25.0: reading github.com/go-swagger/go-swagger/cmd/swagger/cmd/swagger/go.mod at revision cmd/swagger/v0.25.0: unknown revision cmd/swagger/v0.25.0
make: *** [clean] Error 1`

tsachiherman:
> > It would be much more straight forward to add the swagger in the `go.mod` file, wouldn't it ?
> 
> I tried that first, but that caused problems when running make
> go: github.com/go-swagger/go-swagger/cmd/swagger@v0.25.0: reading github.com/go-swagger/go-swagger/cmd/swagger/cmd/swagger/go.mod at revision cmd/swagger/v0.25.0: unknown revision cmd/swagger/v0.25.0
> make: *** [clean] Error 1`

yes, that's probably because the swagger is a sub-directory there. have you tried to add just `github.com/go-swagger/go-swagger v0.25.0` ?

egieseke:
> github.com/go-swagger/go-swagger v0.25.0

That does not seem to install the correct swagger version, since it results in the swagger errors.

tsachiherman (approved):
I think that we should really try to place it in the `go.mod` file, but I don't want to block the build system until that happens.

egieseke:
> I think that we should really try to place it in the `go.mod` file, but I don't want to block the build system until that happens.

I created this ticket to move the swagger dependency to the `go.mod` file.  https://github.com/algorand/go-algorand/issues/1707

algorotem (approved):

PR #1636, opened 2020-10-21
Fixes to algoh timeout e2e test

algorandskiy:
## Summary

TestAlgohWithExpect constantly fails on arm64, this PR addresses it.

* Make Abort to cleanup
* Ensure cleanup is done on errors

TODO: figure out why node startup takes longer than expected.

## Test Plan

This is a test enhancement PR

PR #1706, opened 2020-11-16
Fix intermittent  failure of algohTimeoutTest.exp 

egieseke:
## Summary

Fix error running algohTimeoutTest.exp by providing sufficient time for the network to start.

## Test Plan

Ran make integration and testing on Travis with PR.

### DIFF
--- a/test/e2e-go/cli/algoh/expect/algohTimeoutTest.exp
+++ b/test/e2e-go/cli/algoh/expect/algohTimeoutTest.exp
@@ -13,7 +13,15 @@ if { [catch {
     exec mkdir -p $TEST_ALGO_DIR
     
     set TIME_STAMP [clock seconds]
-    set timeout 60
+
+    set archType [exec $TEST_DATA_DIR/../../scripts/archtype.sh]
+    puts "arch type: $archType"
+    if {[regexp "arm64" $archType]} {
+       set timeout 60
+    } else {
+       set timeout 5
+    }
+    puts "set timeout to $timeout"
 
     set TEST_ROOT_DIR $TEST_ALGO_DIR/root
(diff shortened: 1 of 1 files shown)


egieseke:
I am fairly confident that the cause of the intermittent test failure was due to a short 5-second timeout for the network startup.  I reset that to 60 seconds.  The test should be more reliable now.

algorandskiy:
I have a similar PR #1636 and @tsachiherman did not let it in until we understand why it takes a long for nodes to start. I guess I have a similar issue now after running exclusive app pingpongs - network startup takes significantly longer after even a single 180 sec test run.

egieseke:
> I have a similar PR #1636 and @tsachiherman did not let it in until we understand why it takes a long for nodes to start. I guess I have a similar issue now after running exclusive app pingpongs - network startup takes significantly longer after even a single 180 sec test run.

This test is adversely affecting the productivity of the dev team by causing builds to fail and delaying PRs.  We should commit either your fix or mine and create a separate ticket to look into the problem of slow startup.

tsachiherman:
Do you see this test failing on all platforms, or only a specific one ?
If it's a platform specific, than isolating the slow activity and extending it's timeout for that particular platform makes sense.
In general, i'm opposing to increasing test timeout since I'd like to see consistant product improvement rather than consistant tech debt that eventually get deleted by Ian.

tsachiherman:
Just to be clear, if you'd like to tackle this properly in the coming sprint, than yes, by all means - "unblock" the master and take the work by creating a ticket.

tsachiherman (approved):
I think that we'll need to figure out why this platform is so slow and fix it. But it's ok to do this another day!

PR #1709, opened 2020-11-17
Derive the swagger version from the go.mod file

tsachiherman:
## Summary

We have the swagger version hard coded in the `configure_dev-deps.sh` file, whereas all the other go modules have their version stored in the `go.mod` file.

This change takes the module version from the `go.mod` file for the swagger as well.

This is a followup on https://github.com/algorand/go-algorand/pull/1704

## Test Plan

Since tip-of-tree on the swagger repo would break our build, a successful build ensures that the solution works as expected.


go.mod
```diff
@@ -17,6 +17,7 @@ require (
 	github.com/gen2brain/beeep v0.0.0-20180718162406-4e430518395f
 	github.com/getkin/kin-openapi v0.22.0
 	github.com/go-chi/chi v4.1.2+incompatible // indirect
+	github.com/go-swagger/go-swagger v0.25.0
```
egieseke, line 20: Do we want this dependency in `go.mod` since it is used by our build process but not depended on by our software?

logging/telemetryCommon.go
```diff
@@ -59,15 +59,24 @@ type TelemetryConfig struct {
...
-	ReportHistoryLevel logrus.Level
-	FilePath           string // Path to file on disk, if any
-	ChainID            string `json:"-"`
-	SessionGUID        string `json:"-"`
+	MinLogLevel        logrus.Level `json:"-"` // these are the logrus.Level, but we can't use it directly since on logrus version 1.4.2 they added
+	ReportHistoryLevel logrus.Level `json:"-"` // text marshalers which breaks our backward compatibility.
+	FilePath           string       // Path to file on disk, if any
+	ChainID            string       `json:"-"`
+	SessionGUID        string       `json:"-"`
 	UserName           string
```
egieseke, line 67: Are these telemetry config changes appropriate for this PR?

egieseke (commented):
Are the telemetry changes intentional?

go.mod
tsachiherman: It's currently required by our makefile.. so I would lean toward "yes".
But that's not the only way to achieve that : 
we have have another go.mod/go.sum in a separate directory, and fetch the go-swagger from there.

I think that it makes it simpler to know that all the dependencies are in one location, though.

logging/telemetryCommon.go
tsachiherman: The downside of having a single go.mod for both the go-swagger and our code is that we need to synchronize versions of shared components. These changes are backward compatible, so I could do these without applying this PR, and have the PR applied afterward. ( but I don't think it would add much more meaningful granularity.. )

If you think that these changes need to be splitted off, please let me know and I'll create a separate PR for these.

btoll:
@tsachiherman The changes make sense, but I still don't understand why getting/installing swagger in `configure_dev-deps.sh` in the previous iteration worked and it didn't adding it to `go.mod`.

Can you explain this a bit further please?

tsachiherman:
> @tsachiherman The changes make sense, but I still don't understand why getting/installing swagger in `configure_dev-deps.sh` in the previous iteration worked and it didn't adding it to `go.mod`.
> 
> Can you explain this a bit further please?

Yes, of course. The `scripts/configure_dev-deps.sh` was doing a `cd && GO111MODULE=on go get "$1@${VERSION}" 2>&1` call. Note the `cd`, which would change the current directory to the user's home directory inside the subshell.
This means that any go.mod would be looked at the user's home directory.

I think that we need to address this as well, but that's already a separate issue.

btoll:
> > @tsachiherman The changes make sense, but I still don't understand why getting/installing swagger in `configure_dev-deps.sh` in the previous iteration worked and it didn't adding it to `go.mod`.
> > Can you explain this a bit further please?
> 
> Yes, of course. The `scripts/configure_dev-deps.sh` was doing a `cd && GO111MODULE=on go get "$1@${VERSION}" 2>&1` call. Note the `cd`, which would change the current directory to the user's home directory inside the subshell.
> This means that any `go.mod` would be looked at the user's home directory.
> 
> I think that we need to address this as well, but that's already a separate issue.

Ok, I see.  Thanks for jumping on the ticket.

btoll (approved):

PR #1710, opened 2020-11-17
Consistently report line numbers in TEAL error messages.

jannotti:
## Summary

Previously, we rarely reported line numbers when reporting errors
during assembly.  When we did, we reported them in a strange
way (filename, followed by two colons, then line number, then message
with no colon between line number and the message).

Now line numbers are always reported, and they use a more standard
format:
filename.teal: ##: error message

This makes it much easier to integrate `goal clerk compile` output
into other tools.  I've created an emacs TEAL mode that highlights
errors using this output, and a vscode module would probably be
similarly easy.

## Test Plan

All existing assembler unit tests pass (two were changed because they depended on the line number following the colon).  Most were also augmented to check that the proper line number was reported.

### DIFF
--- a/data/transactions/logic/assembler.go
+++ b/data/transactions/logic/assembler.go
@@ -1183,10 +1183,6 @@ func (ops *OpStream) Bytes() (program []byte, err error) {
 	return
 }
 
-func (ops *OpStream) addLine(err error) error {
-	return &lineError{Line: ops.sourceLine, Err: err}
-}
-
 func (ops *OpStream) error(problem interface{}) error {
 	switch p := problem.(type) {
 	case string:
--- a/data/transactions/logic/assembler.go
+++ b/data/transactions/logic/assembler.go
@@ -352,7 +352,7 @@ func assembleInt(ops *OpStream, spec *OpSpec, args []string) error {
 	}
 	val, err := strconv.ParseUint(args[0], 0, 64)
 	if err != nil {
-		return ops.addLine(err)
+		return ops.error(err)
 	}
 	return ops.Uint(val)
 }
@@ -361,14 +361,14 @@ func assembleInt(ops *OpStream, spec *OpSpec, args []string) error {
 func assembleIntC(ops *OpStream, spec *OpSpec, args []string) error {
 	constIndex, err := strconv.ParseUint(args[0], 0, 64)
 	if err != nil {
-		return ops.addLine(err)
+		return ops.error(err)
 	}
 	return ops.Intc(uint(constIndex))
 }
 func assembleByteC(ops *OpStream, spec *OpSpec, args []string) error {
 	constIndex, err := strconv.ParseUint(args[0], 0, 64)
 	if err != nil {
-		return ops.addLine(err)
+		return ops.error(err)
 	}
 	return ops.Bytec(uint(constIndex))
 }
@@ -528,7 +528,7 @@ func assembleByte(ops *OpStream, spec *OpSpec, args []string) error {
 	}
 	val, _, err = parseBinaryArgs(args)
 	if err != nil {
-		return ops.addLine(err)
+		return ops.error(err)
 	}
 	return ops.ByteLiteral(val)
 }
@@ -542,7 +542,7 @@ func assembleIntCBlock(ops *OpStream, spec *OpSpec, args []string) error {
 	for i, xs := range args {
 		cu, err := strconv.ParseUint(xs, 0, 64)
 		if err != nil {
-			return ops.addLine(err)
+			return ops.error(err)
 		}
 		l = binary.PutUvarint(scratch[:], cu)
 		ops.Out.Write(scratch[:l])
@@ -559,7 +559,7 @@ func assembleByteCBlock(ops *OpStream, spec *OpSpec, args []string) error {
 	for len(rest) > 0 {
 		val, consumed, err := parseBinaryArgs(rest)
 		if err != nil {
-			return ops.addLine(err)
+			return ops.error(err)
 		}
 		bvals = append(bvals, val)
 		rest = rest[consumed:]
@@ -585,7 +585,7 @@ func assembleAddr(ops *OpStream, spec *OpSpec, args []string) error {
 	}
 	addr, err := basics.UnmarshalChecksumAddress(args[0])
 	if err != nil {
-		return ops.addLine(err)
+		return ops.error(err)
 	}
 	return ops.ByteLiteral(addr[:])
 }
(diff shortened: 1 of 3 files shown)


tsachiherman (commented):
You might want to review https://blog.golang.org/go1.13-errors
having the ability to break apart the errors string from the line numbers might be valuable for a debugger.

data/transactions/logic/assembler_test.go
```diff
@@ -745,11 +658,13 @@ func TestAssembleRejectNegJump(t *testing.T) {
 	t.Parallel()
 	text := `wat:
 int 1
-bnz wat`
+bnz wat
+int 2`
```
algorandskiy, line 662: why `int 2` here and below?

data/transactions/logic/assembler_test.go
```diff
@@ -267,141 +267,54 @@ pop
...
+	// test for the correct line number in the error is more
+	// meaningful.
+	source := "int 1\n" + line + "\nint 1\n"
+	_, err := AssembleStringWithVersion(source, ver)
+	if expected == "" {
+		require.NoError(t, err)
+	} else {
+		require.Error(t, err)
+		if strings.HasSuffix(expected, "...") {
+			require.Contains(t, "^"+err.Error(), "^2: "+expected[:len(expected)-3])
```
algorandskiy, line 281: Is `^` really needed?

data/transactions/logic/assembler.go
```diff
@@ -885,17 +888,23 @@ func init() {
...
-func (lew *lineErrorWrapper) Error() string {
-	return fmt.Sprintf(":%d %s", lew.Line, lew.Err.Error())
+func addLine(line int, err error) error {
+	return &lineError{Line: line, Err: err}
+}
+func newLineError(line int, msg string) error {
+	return addLine(line, errors.New(msg))
+}
+func fmtLineError(line int, format string, args ...interface{}) error {
+	return addLine(line, fmt.Errorf(format, args...))
```
algorandskiy, line 903: two functions look exceeding, maybe leave `fmtLineError` only?

data/transactions/logic/assembler.go
```diff
@@ -868,7 +871,7 @@ func asmDefault(ops *OpStream, spec *OpSpec, args []string) error {
 	}
 	err = ops.Out.WriteByte(spec.Opcode)
 	if err != nil {
-		return lineErr(ops.sourceLine, err)
+		return ops.addLine(err)
```
algorandskiy, line 874: I think `lineErr` is a better name than `addLine`. The latter assumes adding error as line (?) to somewhere rather than returning a new error object.

algorandskiy (commented):
I like the idea of unifying error messages. Some comments below.

data/transactions/logic/assembler_test.go
jannotti: I was making the test harder to pass.  In my initial implementation, a bug that occurred was that this particular error reported the last line *always*.  So, to make things harder, I added another line to ensure I was reporting the error on the bnz line, not just the last line.

data/transactions/logic/assembler_test.go
jannotti: Again, I'm making the test a bit harder to pass and a more "real" test.  It's conceivable that an error could report line 12 instead of line 2, and without the extra delimiter, Contains() would still report success because the message started with "2".

data/transactions/logic/assembler.go
jannotti: I'm ok with that.  I was unsure if calling Errorf() on a static string without any args to substitute in was considered bad style, so I wanted to give the option to avoid it.

data/transactions/logic/assembler.go
jannotti: The point is that it is just adds a line to an existing error.  I want to make clear that this is not producing the error, it's only adding the line information to the existing error.  I don't feel very strongly about it.  I considered withLine() and annotate() as possibilities as well.

jannotti:
> You might want to review https://blog.golang.org/go1.13-errors
> having the ability to break apart the errors string from the line numbers might be valuable for a debugger.

I could add a trivial Unwrap() to lineError.  It would return the inner error.  There are no callers yet - I'm amenable either way.

jannotti:
> > You might want to review https://blog.golang.org/go1.13-errors
> > having the ability to break apart the errors string from the line numbers might be valuable for a debugger.
> 
> I could add a trivial Unwrap() to lineError.  It would return the inner error.  There are no callers yet - I'm amenable either way.

Yeah, I'll definitely do that. I didn't know about errors.Is and errors.As. I would want them usable without the caller going back to add the Unwrap later. Thanks.

data/transactions/logic/assembler.go
```diff
@@ -349,7 +352,7 @@ func assembleInt(ops *OpStream, spec *OpSpec, args []string) error {
 	}
 	val, err := strconv.ParseUint(args[0], 0, 64)
 	if err != nil {
-		return err
+		return ops.addLine(err)
```
algorandskiy, line 355: I still think `ops.wrap(err)` or similar would make more sense than `ops.addLine(err)`

algorandskiy (commented):
The change looks good to me except such naming.
@brianolson Could you also take a look please?

data/transactions/logic/assembler.go
```diff
@@ -1177,6 +1186,18 @@ func (ops *OpStream) Bytes() (program []byte, err error) {
 	return
 }
 
+func (ops *OpStream) addLine(err error) error {
```
brianolson, line 1189: Another option for readability within this module is that ops.error(x interface{}) could type-switch on x and handle either string or error. (Kinda like log.Print())

brianolson (commented):
Generally I like it and I think this will make editor integration easier. A couple tweaks and ideas on code readability.

data/transactions/logic/assembler.go
```diff
@@ -1187,15 +1184,22 @@ func (ops *OpStream) Bytes() (program []byte, err error) {
 }
 
 func (ops *OpStream) addLine(err error) error {
```
algorandskiy, line 1186: seems like this function is not used anymore, please remove.

algorandskiy (changes requested):
The last fix needed remove `addLine` it is good to be merged in!

data/transactions/logic/assembler.go
jannotti: Thanks, removing.

algorandskiy (approved):

PR #1713, opened 2020-11-18
Add transaction evaluation distribution metrics

tsachiherman:
## Summary

Add the following fields to the `AssembleBlockStats`:
* ProcessingTime
        // 10 buckets: 0-0.1ms, 0.1-02ms .. 0.9ms-1ms
	// 9 buckets: 1ms-2ms .. 9ms-10ms
	// 9 buckets: 10ms-20ms .. 90ms-100ms
	// 9 buckets: 100ms-200ms .. 900ms-1s
	// 1 bucket: 1s+
* BlockGenerationDuration - The time it took to execute the `pendingBlockEvaluator.GenerateBlock()`
* TransactionsLoopStartTime - how long did it take between the time we started the deadline and the time when the block was ready ?

## Test Plan

Ran and tested on performance testing environment. Works as expected. Sample output:
```json
"metrics": {
        "StartCount": 10611,
        "IncludedCount": 4649,
        "InvalidCount": 0,
        "MinFee": 1000,
        "MaxFee": 10000,
        "AverageFee": 9965,
        "MinLength": 215,
        "MaxLength": 229,
        "MinPriority": 0,
        "MaxPriority": 0,
        "CommittedCount": 0,
        "StopReason": "block-full",
        "TotalLength": 999787,
        "EarlyCommittedCount": 4617,
        "Nanoseconds": 307555590,
        "ProcessingTime": [
          4625,          11,          8,          1,          0,          1,          0,          0,          0,          0,
          0,          1,          1,          0,          0,          0,          0,          0,          2,
          0,          0,          0,          0,          0,          0,          0,          0,          0,
          0,          0,          0,          0,          0,          0,          0,          0,          0,
          0
        ],
        "BlockGenerationDuration": 80980171,
        "TransactionsLoopStartTime": 0
      }
```


data/pools/transactionPool.go
```diff
@@ -698,15 +718,21 @@ func (pool *TransactionPool) recomputeBlockEvaluator(committedTxIds map[transact
 	}
 
 	pool.assemblyMu.Lock()
+	if (pool.assemblyDeadline != time.Time{}) {
```
algorandskiy, line 721: mb `pool.assemblyDeadline.IsZero()` ?

data/pools/transactionPool.go
```diff
@@ -698,15 +718,21 @@ func (pool *TransactionPool) recomputeBlockEvaluator(committedTxIds map[transact
...
+		asmStats.TransactionsLoopStartTime = int64(firstTxnGrpTime.Sub(pool.assemblyDeadline.Add(-config.ProposalAssemblyTime)))
```
algorandskiy, line 722: why `Add(-config.ProposalAssemblyTime)` and not `Sub` ?

algorandskiy (commented):
LGTM, couple questions below

data/pools/transactionPool.go
```diff
@@ -582,16 +593,23 @@ func (pool *TransactionPool) addToPendingBlockEvaluatorOnce(txgroup []transactio
 					stats.StopReason = telemetryspec.AssembleBlockFull
 				} else {
 					stats.StopReason = telemetryspec.AssembleBlockTimeout
+					// if the block is not full, it means that the above transaction made it to the block, so we want to add it here.
+					stats.ProcessingTime.AddTransaction(transactionGroupDuration)
```
brianolson, line 597: the stats nil check below makes me notice that there isn't one here, but apparently it works out that this code is only reached through paths that set stats

data/pools/transactionPool.go
tsachiherman: Because `Add` returns `time.Time` whereas `Sub` returns `time.Duration`

brianolson: Time.Add(Duration) -> Time
Time.Sub(Time) -> Duration

brianolson: It took me a bit to figure out that this is (deadline - allowedTime) to get the implicit start time, and then (first txn grp time - start time) shows how long the setup time to get there is. Dunno if there's a short comment that would clarify this intent and reasoning.

data/pools/transactionPool.go
tsachiherman: sounds good.

logging/telemetryspec/metric.go
```diff
@@ -129,3 +136,49 @@ type RoundTimingMetrics struct {
 func (m RoundTimingMetrics) Identifier() Metric {
 	return roundTimingMetricsIdentifier
 }
+
+type transactionProcessingTimeDistibution struct {
+	// 10 buckets: 0-100Kns, 100Kns-200Kns .. 900Kns-1ms
```
brianolson, line 141: I slightly wish this was uniform log scale buckets. To keep things int it could be power of two buckets? <= 1ns, <= 2ns, <= 4ns, etc. It would make AddTransaction simpler. Not at all a big deal.

data/pools/transactionPool.go
tsachiherman: I think that stat != nil is always the case when we recompute.

logging/telemetryspec/metric.go
tsachiherman: I baffled with that as well. I think that this would become more valuable as we try to optimize non-trivial transactions. 
Having a "pure" logarithmic scale would place too many "entries" in a single bucket and would hurt the resolution we'd like to have.
To be concrete : for the block assembly, we really want *all* the transactions to be on the lower bucket. Any time this is not the case, we need to try and figure out why.

data/pools/transactionPool.go
tsachiherman: yes, confirmed. I'll remove the stat != nil check.

brianolson (approved):

data/pools/transactionPool.go
algonautshant, line 597: Do we care not to set ProcessingTime if ErrNoSpace?
Can't just set this once on line 590?

data/pools/transactionPool.go
```diff
@@ -582,16 +593,23 @@ func (pool *TransactionPool) addToPendingBlockEvaluatorOnce(txgroup []transactio
...
 					pool.assemblyResults.err = fmt.Errorf("could not generate block for %d: %v", pool.assemblyResults.roundStartedEvaluating, gerr)
 				} else {
 					pool.assemblyResults.blk = lvb
 				}
+				stats.BlockGenerationDuration = uint64(time.Now().Sub(blockGenerationStarts))
+				pool.assemblyResults.stats = *stats
 				pool.assemblyCond.Broadcast()
+			} else {
+				// add the transaction time only if we didn't ended up finishing the block.
+				stats.ProcessingTime.AddTransaction(transactionGroupDuration)
```
algonautshant, line 612: Why stats in not set in this branch?
`pool.assemblyResults.stats = *stats`

algonautshant (approved):
Looks good. 
Have couple of questions.

data/pools/transactionPool.go
tsachiherman: no; because we want to have the "numbers" to match up.

data/pools/transactionPool.go
tsachiherman: because there is no need to do that yet.. it will get updated when the block get finalized.

PR #1715, opened 2020-11-20
Add applications to min balance requirements calculation

algorandskiy:
## Summary

Improve apps support in pingpong:
1. Add apps min balance requirements into required balance calculation, similar to assets
2. Use SuggestedFee for accounts funding to compete with pending transactions in the pool

## Test Plan

Checked against local network.

### DIFF
--- a/shared/pingpong/pingpong.go
+++ b/shared/pingpong/pingpong.go
@@ -127,8 +127,13 @@ func prepareNewAccounts(client libgoal.Client, cfg PpConfig, wallet []byte, acco
 }
 
 // determine the min balance per participant account
-func computeAccountMinBalance(cfg PpConfig) (requiredBalance uint64) {
-	const minActiveAccountBalance uint64 = 100000 // min balance for any active account
+func computeAccountMinBalance(client libgoal.Client, cfg PpConfig) (requiredBalance uint64, err error) {
+	proto, err := getProto(client)
+	if err != nil {
+		return
+	}
+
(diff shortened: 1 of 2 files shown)


shared/pingpong/pingpong.go
```diff
@@ -141,6 +146,12 @@ func computeAccountMinBalance(cfg PpConfig) (requiredBalance uint64) {
 	}
 	if cfg.MaxFee != 0 {
 		fee = cfg.MaxFee
+	} else {
+		// follow the same logic as constructTxn
+		fee, err = client.SuggestedFee()
```
tsachiherman, line 151: This is correct - functionally. However, it will reduce the pingpong throughput..

algorandskiy: this is called only for funding accounts that is done in between test runs. I guess one call per `--duration N` is not a big deal.

tsachiherman: ahh. I didn't get that it's only one call per batch. In this case, I think that it's fine.

tsachiherman (approved):

shared/pingpong/pingpong.go
cce: Was just running the `ineffassign` linter and saw that this change makes it so the `if cfg.MinFee > fee { fee = cfg.MinFee }` section above is ineffectual/useless, since this else expression will overwrite whatever fee might have been set above. Is that OK? Should I remove the `if cfg.MinFee > fee` code?

PR #1716, opened 2020-11-20
Windows scripts fix

mxmauro:
Depending on how MSYS2 is installed, GO language and repository being build, some times Go modules ends outside of MSYS2.

These patches ensure all the code is contained within MSYS2 environment.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -178,7 +178,7 @@ build: buildsrc gen
 buildsrc: crypto/libs/$(OS_TYPE)/$(ARCH)/lib/libsodium.a node_exporter NONGO_BIN deps $(ALGOD_API_SWAGGER_INJECT) $(KMD_API_SWAGGER_INJECT)
 	mkdir -p tmp/go-cache && \
 	touch tmp/go-cache/file.txt && \
-	GOCACHE=$(SRCPATH)/tmp/go-cache && go install $(GOTRIMPATH) $(GOTAGS) $(GOBUILDMODE) -ldflags="$(GOLDFLAGS)" ./...
+	export GOCACHE=$(SRCPATH)/tmp/go-cache && go install $(GOTRIMPATH) $(GOTAGS) $(GOBUILDMODE) -ldflags="$(GOLDFLAGS)" ./...
 
 SOURCES_RACE := github.com/algorand/go-algorand/cmd/kmd
 
--- a/Makefile
+++ b/Makefile
@@ -54,7 +54,7 @@ export GOBUILDMODE := -buildmode=exe
 endif
 
 GOTAGS      := --tags "$(GOTAGSLIST)"
-GOTRIMPATH	:= $(shell GOPATH=$(GOPATH) && go help build | grep -q .-trimpath && echo -trimpath)
+GOTRIMPATH	:= $(shell export GOPATH=$(GOPATH) && go help build | grep -q .-trimpath && echo -trimpath)
 
 GOLDFLAGS_BASE  := -X github.com/algorand/go-algorand/config.BuildNumber=$(BUILDNUMBER) \
 		 -X github.com/algorand/go-algorand/config.CommitHash=$(COMMITHASH) \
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -178,7 +178,7 @@ build: buildsrc gen
 buildsrc: crypto/libs/$(OS_TYPE)/$(ARCH)/lib/libsodium.a node_exporter NONGO_BIN deps $(ALGOD_API_SWAGGER_INJECT) $(KMD_API_SWAGGER_INJECT)
 	mkdir -p tmp/go-cache && \
 	touch tmp/go-cache/file.txt && \
-	GOCACHE=$(SRCPATH)/tmp/go-cache go install $(GOTRIMPATH) $(GOTAGS) $(GOBUILDMODE) -ldflags="$(GOLDFLAGS)" ./...
+	GOCACHE=$(SRCPATH)/tmp/go-cache && go install $(GOTRIMPATH) $(GOTAGS) $(GOBUILDMODE) -ldflags="$(GOLDFLAGS)" ./...
```
tsachiherman, line 181: I think that you'll need to make it into a "export GOCACHE=..." if you want to add the &&

mxmauro: Didn't have any issue but I'll do because it works better with `&&`

scripts/configure_dev-deps.sh
```diff
@@ -31,6 +31,11 @@ function install_go_module {
     fi
 }
 
+UNAME=$(uname)
+if [[ "${UNAME}" == *"MINGW"* ]]; then
+	export GOPATH=$HOME/go
+fi
+
```
tsachiherman, line 38: I don't feel that this is the right thing to do. It seems to be windows specific, and it's would be better if we could align all the platforms. Given that this file being invoked via configure_dev.sh, which being invoked manually, I would suggest adding this to the caller. Also, if we want to avoid taking this to the caller, we could add a make larget for this one, since the makefile already export the GOPATH correctly.

scripts/build_package.sh
```diff
@@ -34,7 +34,7 @@ fi
 
 UNAME=$(uname)
 if [[ "${UNAME}" == *"MINGW"* ]]; then
-	GOPATH1=$HOME/go
+	export GOPATH=$HOME/go
```
tsachiherman, line 37: I think that there are three issues with this if statement:
1. we shouldn't compare here against MINGW; instead, we should use the `scripts/ostype.sh` to determine OS and use that.
2. the fact the we always update the GOPATH here is likely to be wrong. We definitely want to export it, but the caller should have set it's value already.
3. the fact that we have a difference between the windows and linux here suggest that something isn't aligned.. we should align the behavior and expectations.

mxmauro: 1. May be better although Windows is the only different than the rest. Wll have to patch several places.
2. Yes but the problem is `MSYS2` package only sets `GOROOT`. I think `GOPATH` is not added because Go is installed globally.
3. Windows is a totally different world, actually the usage of `MSYS2` is to avoid converting all the scripts to batch or powershell. But `MSYS2` also has its limitations, that's why these sort of patches.

scripts/configure_dev-deps.sh
mxmauro: I have the doubt if this script is always meant to be executed by the parent script or if it can be executed standalone. Please confirm me it is always the former, so I'll do the change

tsachiherman: At this time, the script is always meant to be invoked by `script/configure_dev.sh`; there is no use case that I can see where the script need to be invoked on its own. ( i.e. if any of the dependencies changes, the user would run the `script/configure_dev.sh` which would update all the required dependencies ).

The `script/configure_dev.sh` doesn't really care about go, so we could definitely add a check there to ensure that the GOPATH was previously set by the user.
I think that it would get solved better by adding a "configure" target to make.
i.e.
configure:
    ./scripts/configure_dev.sh

mxmauro: Ok let me see it on the weekend and see necessary changes to documentation.

tsachiherman: thanks! I really appreciate your contribution !

mxmauro: You are welcome 👍

hernandp:
Hi guys, regarding the problem that GOPATH will point to %USERPROFILE% as base directory instead of the MSYS2 filesystem, we found that asking the user to execute:

```
echo 'export GOPATH=$HOME/go' >> $HOME/.bashrc
```
and restarting the environment is enough for the subsequent `make` stage to succeed.

FYI, This is already in documentation: 
https://developer.algorand.org/tutorials/compile-and-run-the-algorand-node-natively-windows/#1-environment-setup

Maybe this PR can be closed by now.
Thanks.

tsachiherman:
@hernandp, 

Thanks for looking into that. I'll close the PR.

PR #1718, opened 2020-11-23
add graph propagation trace to mainnet and scenario1

brianolson:
## Summary

Allow for setting graph propagation trace server when testing mainnet-model and scenario1

## Test Plan

Has been used in test clusters.

### DIFF
--- a/test/testdata/deployednettemplates/recipes/mainnet-model/network-tpl.json
+++ b/test/testdata/deployednettemplates/recipes/mainnet-model/network-tpl.json
@@ -6,17 +6,17 @@
   "instances": {
     "relays": {
       "config": "./configs/relay.json",
-      "type": "c5.xlarge",
+      "type": "c5d.xlarge",
       "count": 25
     },
     "participatingNodes": {
       "config": "./configs/node.json",
-      "type": "c5.xlarge",
+      "type": "c5d.xlarge",
       "count": 50
     },
     "nonParticipatingNodes": {
       "config": "./configs/nonPartNode.json",
-      "type": "c5.xlarge",
+      "type": "c5d.xlarge",
       "count": 10
     }
   },
@@ -28,7 +28,7 @@
         "participatingNodes": 0,
         "nonParticipatingNodes": 0
       },
-      "region": "us-west-1"
+      "region": "us-east-2"
     },
     {
       "name": "us-n",
@@ -37,7 +37,7 @@
         "participatingNodes": 14,
         "nonParticipatingNodes": 35
       },
-      "region": "us-west-1"
+      "region": "us-east-2"
     },
     {
       "name": "eu-r",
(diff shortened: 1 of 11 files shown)


test/testdata/deployednettemplates/recipes/mainnet-model/network-tpl.json
```diff
@@ -6,17 +6,17 @@
   "instances": {
     "relays": {
       "config": "./configs/relay.json",
-      "type": "c5.xlarge",
+      "type": "c5d.xlarge",
```
tsachiherman, line 9: Have you seen any measurable gains when switching the EBS -> NVMe ?
For the mainnet model, I do think that we shouldn't use EBS, but NVMe might be an overspec for non-relays. ( i.e. we should use a non-nvme SSD or spinning rust )

test/testdata/deployednettemplates/recipes/mainnet-model/network-tpl.json
```diff
@@ -28,7 +28,7 @@
         "participatingNodes": 0,
         "nonParticipatingNodes": 0
       },
-      "region": "us-west-1"
+      "region": "us-east-2"
```
tsachiherman, line 31: We had 2 reasons for using us-west-2 here:
1. It was easier to spot the generated instances, since us-west-1 wasn't used other the by this recipe.
2. There is a instance limit on each zone, and we might have reached our limit on us-east-2. ( especially when we run multiple networks concurrently ).

I can see how having the telemetry server nearby would be a good thing. I'm fine with relocating this once you verify that the latter would not break our runs.

bricerisingalgorand: Our telemetry server automation sets up a self contained telemetry server right? Maybe updating that to support other regions might be better so that we can continue not worrying about instance limits in the future

tsachiherman: No, I think that the performance testing environment is using a pre-deployed telemetry server on us-east-1. I think that it's ok to auto-deploy a new telemetry server instance dynamically, but this means that we also need to preserve the collected data once we're done with this run.

test/testdata/deployednettemplates/recipes/mainnet-model/network-tpl.json
brianolson: Yes. This was crucial to getting the initial 1000 TPS benchmark. We should always be using local SSD for the tracker db.

test/testdata/deployednettemplates/recipes/mainnet-model/network-tpl.json
brianolson: I made the deploy_ephemeral_telemetry.sh script take an argument for what region to run in. I think it's important to run the telemetry server in the same region as the cluster. In my experimentation in the last couple weeks it was easiest to get things running in us-east-2 but I'm not super attached to that region.

PR #1719, opened 2020-11-23
TEAL: fix return type check in assembler

algorandskiy:
## Release notes

This PR addresses inconsistency in the spec where operands pushing order is different from return values order.
There is no changes in TEAL evaluation so that no protocol upgrade is required.

Corresponding conning to the spec is https://github.com/algorandfoundation/specs/commit/21adc34b722ff365076b7e83a797190e7f99d6a2

Please note, the consensus protocol version is pointing to the not up-to-date spec https://github.com/algorandfoundation/specs/releases/tag/August2020-TEALv2 that includes only Rekey but not applications.

## Summary

* app and asset getters return a pair
* order in opcode specification was "top-left"
* type check expects "bottom-left" order

## Test Plan

New test added

### DIFF
--- a/data/transactions/logic/README.md
+++ b/data/transactions/logic/README.md
@@ -352,6 +352,12 @@ For version 1, subsequent bytes after the varuint are program opcode bytes. Futu
 
 It is important to prevent newly-introduced transaction fields from breaking assumptions made by older versions of TEAL. If one of the transactions in a group will execute a TEAL program whose version predates a given field, that field must not be set anywhere in the transaction group, or the group will be rejected. For example, executing a TEAL version 1 program on a transaction with RekeyTo set to a nonzero address will cause the program to fail, regardless of the other contents of the program itself.
 
+This requirement is enforced as follows:
+
+* For every transaction, compute the earliest TEAL version that supports all the fields and and values in this transaction. For example, a transaction with a nonzero RekeyTo field will have version (at least) 2.
+
+* Compute the largest version number across all the transactions in a group (of size 1 or more), call it `maxVerNo`. If any transaction in this group has a TEAL program with a version smaller than `maxVerNo`, then that TEAL program will fail.
+
 ## Varuint
 
 A '[proto-buf style variable length unsigned int](https://developers.google.com/protocol-buffers/docs/encoding#varint)' is encoded with 7 data bits per byte and the high bit is 1 if there is a following byte and 0 for the last byte. The lowest order 7 bits are in the first byte, followed by successively higher groups of 7 bits.
--- a/data/transactions/logic/assembler.go
+++ b/data/transactions/logic/assembler.go
@@ -53,6 +53,7 @@ type OpStream struct {
 	Out     bytes.Buffer
 	Version uint64
 	Trace   io.Writer
+	Stderr  io.Writer
 	vubytes [9]byte
 
 	intc        []uint64
@@ -1003,7 +1004,7 @@ func (ops *OpStream) checkArgs(spec OpSpec) error {
 		if !typecheck(argType, stype) {
 			msg := fmt.Sprintf("%s arg %d wanted type %s got %s", spec.Name, i, argType.String(), stype.String())
 			if len(ops.labelReferences) > 0 {
-				fmt.Fprintf(os.Stderr, "warning: %d: %s; but branches have happened and assembler does not precisely track types in this case\n", ops.sourceLine, msg)
+				fmt.Fprintf(ops.Stderr, "warning: %d: %s; but branches have happened and assembler does not precisely track types in this case\n", ops.sourceLine, msg)
 			} else {
 				return lineErr(ops.sourceLine, errors.New(msg))
 			}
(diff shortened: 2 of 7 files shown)


data/transactions/logic/assembler.go
```diff
@@ -1003,7 +1004,7 @@ func (ops *OpStream) checkArgs(spec OpSpec) error {
 		if !typecheck(argType, stype) {
 			msg := fmt.Sprintf("%s arg %d wanted type %s got %s", spec.Name, i, argType.String(), stype.String())
 			if len(ops.labelReferences) > 0 {
-				fmt.Fprintf(os.Stderr, "warning: %d: %s; but branches have happened and assembler does not precisely track types in this case\n", ops.sourceLine, msg)
+				fmt.Fprintf(ops.Stderr, "warning: %d: %s; but branches have happened and assembler does not precisely track types in this case\n", ops.sourceLine, msg)
```
tsachiherman, line 1007: that's a nice one ;-)

data/transactions/logic/assembler.go
```diff
@@ -1226,7 +1227,7 @@ func AssembleStringWithVersionEx(text string, version uint64) ([]byte, map[int]i
 	}
 
 	sr = strings.NewReader(text)
-	ops := OpStream{Version: version}
+	ops := OpStream{Version: version, Stderr: os.Stderr}
```
tsachiherman, line 1230: I'm not sure if this one is ok. I don't think that the assembler should write output directly to the stdout/stderr. Instead, it should receive a place to write the error to.
i.e. only the "goal" command line utility should be the one that actually writes to the console. All the rest of the utility functions should return errors / write to provided interfaces, etc.

having said that, I'm ok with deferring this change to a future PR.

algorandskiy: I agree, I only established the basis needed for tests and it opens options for further improvements.

tsachiherman (dismissed):
looks good to me, would be happy to see additional approval before merging this in.

data/transactions/logic/assembler.go
brianolson: #1710 did a lot to work towards doing this better, probably room for another round after both of these are in

data/transactions/logic/opcodes.go
```diff
@@ -154,16 +154,16 @@ var OpSpecs = []OpSpec{
 	{0x60, "balance", opBalance, asmDefault, disDefault, oneInt, oneInt, 2, runModeApplication, opSizeDefault},
 	{0x61, "app_opted_in", opAppCheckOptedIn, asmDefault, disDefault, twoInts, oneInt, 2, runModeApplication, opSizeDefault},
 	{0x62, "app_local_get", opAppGetLocalState, asmDefault, disDefault, oneInt.plus(oneBytes), oneAny, 2, runModeApplication, opSizeDefault},
-	{0x63, "app_local_get_ex", opAppGetLocalStateEx, asmDefault, disDefault, twoInts.plus(oneBytes), oneInt.plus(oneAny), 2, runModeApplication, opSizeDefault},
+	{0x63, "app_local_get_ex", opAppGetLocalStateEx, asmDefault, disDefault, twoInts.plus(oneBytes), oneAny.plus(oneInt), 2, runModeApplication, opSizeDefault},
```
brianolson, line 157: The problem with this change is that this changes the *spec*
The implementation in `opAppGetLocalStateEx()` isn't the spec.
`(cd data/transactions/logic && rm -f TEAL_opcodes.md; make)`
Then you see it changes the opcode spec.

brianolson (changes requested):
This change can't go forward until we carefully consider whether the published spec (though backwards of what was intended) is what we need to follow or if the deployed Go code is actually authoritative.

data/transactions/logic/opcodes.go
algorandskiy: I guess the spec can be simply extended to specify pushing order. Currently it is not clear in the spec doc with direction the stack grows. I updated the spec and the spec generator.

data/transactions/logic/README.md
```diff
@@ -352,6 +352,12 @@ For version 1, subsequent bytes after the varuint are program opcode bytes. Futu
 
 It is important to prevent newly-introduced transaction fields from breaking assumptions made by older versions of TEAL. If one of the transactions in a group will execute a TEAL program whose version predates a given field, that field must not be set anywhere in the transaction group, or the group will be rejected. For example, executing a TEAL version 1 program on a transaction with RekeyTo set to a nonzero address will cause the program to fail, regardless of the other contents of the program itself.
 
+This requirement is enforced as follows:
+
+* For every transaction, compute the earliest TEAL version that supports all the fields and and values in this transaction. For example, a transaction with a nonzero RekeyTo field will have version (at least) 2.
+
+* Compute the largest version number across all the transactions in a group (of size 1 or more), call it `maxVerNo`. If any transaction in this group has a TEAL program with a version smaller than `maxVerNo`, then that TEAL program will fail.
+
```
algorandskiy, line 360: This part present in [the spec](https://github.com/algorandfoundation/specs/blob/master/dev/TEAL.md#encoding-and-versioning) but is missing in our repo

tsachiherman (approved):

PR #1720, opened 2020-11-23
Remove extra newline from version output.

winder:
##  Summary

There's an extra newline at the end of `goal -v` which looks weird in one of my scripts. In general I think these sort of helper functions to leave the formatting to the caller, which it is in this case where it adds a 2nd newline.

### DIFF
--- a/config/version.go
+++ b/config/version.go
@@ -113,7 +113,7 @@ func GetCurrentVersion() Version {
 // FormatVersionAndLicense prints current version and license information
 func FormatVersionAndLicense() string {
 	version := GetCurrentVersion()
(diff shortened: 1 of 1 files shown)


tsachiherman (commented):
Looks good to me, but we will need the @btoll and @onetechnical here as well, since they use this functionality as part of their build system.

btoll:
> Looks good to me, but we will need the @btoll and @onetechnical here as well, since they use this functionality as part of their build system.

This won't affect anything in the package build pipeline, but I can't speak to anything else that could perhaps be relying on the output.

onetechnical:
I think we do some regex checking for versions in tests but I don't think this should affect anything.

onetechnical (approved):

PR #1721, opened 2020-11-23
Implement opt-in in dryrun

algorandskiy:
## Summary

Current dryrun implementation manages `Balances` interface by itself so it also needs to process app opt-in transactions.

## Test Plan

New e2e test added


tsachiherman (approved):
looks good.

PR #1723, opened 2020-11-24
Fix rpm upload

btoll:
## Summary

This fixes the `archive` target that wasn't properly uploading stable rpm packages because "stable" isn't in the filename.  Instead, it is sufficient to just search for the version number.

## Test Plan

Tested in the eks cluster by running the pipeline here: https://jenkins.algodev.network/job/go-algorand.muleci/

You can see the expected build artifacts here: https://s3.console.aws.amazon.com/s3/buckets/algorand-internal?region=us-east-1&prefix=channel/dev/2.2.91422/linux/amd64/&showversions=false


onetechnical (approved):

egieseke (approved):
Looks good.

PR #1724, opened 2020-11-25
Fix opt-ins and timeouts in pingpong assets testing

algorandskiy:
## Summary

Current asset allocation and distribution implementation leads to invalid asset transactions:
 * opt-in (when `proto.MaxAssetsPerAccount` is exceeded)
 * transfer (when transferring to/from non-opted in accounts)

The fix preserves unconditional assets creation by every account but also tracks opt ins and stops when `proto.MaxAssetsPerAccount` is reached. `OptIns map[Creatable][]Address` is used to select random sender/receiver for random asset ID.

## Test Plan

Run e2e `pingpongText` with current settings and `--numasset=100 --numaccounts 10` to exceed limits.

### DIFF
--- a/shared/pingpong/accounts.go
+++ b/shared/pingpong/accounts.go
@@ -130,12 +130,19 @@ func throttleTransactionRate(startTime time.Time, cfg PpConfig, totalSent uint64
 // Step 1) Create X assets for each of the participant accounts
 // Step 2) For each participant account, opt-in to assets of all other participant accounts
 // Step 3) Evenly distribute the assets across all participant accounts
-func prepareAssets(accounts map[string]uint64, client libgoal.Client, cfg PpConfig) (resultAssetMaps map[uint64]v1.AssetParams, err error) {
+func prepareAssets(accounts map[string]uint64, client libgoal.Client, cfg PpConfig) (resultAssetMaps map[uint64]v1.AssetParams, optIns map[uint64][]string, err error) {
+	proto, err := getProto(client)
+	if err != nil {
+		return
+	}
 
 	var startTime = time.Now()
 	var totalSent uint64 = 0
 	resultAssetMaps = make(map[uint64]v1.AssetParams)
 
+	// optIns contains own and explicitly opted-in assets
+	optIns = make(map[uint64][]string)
+	assetsInAcct := make(map[string]int)
 	// 1) Create X assets for each of the participant accounts
 	for addr := range accounts {
 		addrAccount, addrErr := client.AccountInformation(addr)
@@ -146,6 +153,7 @@ func prepareAssets(accounts map[string]uint64, client libgoal.Client, cfg PpConf
 		}
 
 		toCreate := int(cfg.NumAsset) - len(addrAccount.AssetParams)
+		assetsInAcct[addr] = toCreate
 
 		fmt.Printf("Creating %v create asset transaction for account %v \n", toCreate, addr)
 		fmt.Printf("cfg.NumAsset %v, addrAccount.AssetParams %v\n", cfg.NumAsset, addrAccount.AssetParams)
@@ -186,21 +194,36 @@ func prepareAssets(accounts map[string]uint64, client libgoal.Client, cfg PpConf
 			totalSent++
 			throttleTransactionRate(startTime, cfg, totalSent)
 		}
-		account, accountErr := client.AccountInformation(addr)
-		if accountErr != nil {
-			fmt.Printf("Cannot lookup source account %v\n", addr)
-			err = accountErr
-			return
-		}
-
-		assetParams := account.AssetParams
-		fmt.Printf("Configured  %d assets %+v\n", len(assetParams), assetParams)
-
 	}
 
-	time.Sleep(time.Second * 15)
+	// wait until all the assets created
+	for addr := range accounts {
+		var account v1.Account
+		for {
+			account, err = client.AccountInformation(addr)
+			if err != nil {
+				fmt.Printf("Warning, cannot lookup source account")
+				return
+			}
+			if len(account.AssetParams) >= assetsInAcct[addr] {
+				break
+			}
+			time.Sleep(time.Second)
+		}
+		assetParams := account.AssetParams
+		if !cfg.Quiet {
+			fmt.Printf("Configured  %d assets %+v\n", len(assetParams), assetParams)
+		}
+		// add own asset to opt-ins since asset creators are auto-opted in
+		for k := range account.AssetParams {
+			optIns[k] = append(optIns[k], addr)
+		}
+	}
 
-	// 2) For each participant account, opt-in to assets of all other participant accounts
+	// 2) For each participant account, opt-in up to proto.MaxAssetsPerAccount assets of all other participant accounts
+
+	// optInsByAddr tracks only explicitly opted-in assets
+	optInsByAddr := make(map[string]map[uint64]bool)
 	for addr := range accounts {
 		if !cfg.Quiet {
 			fmt.Printf("Opting in assets from account %v\n", addr)
@@ -228,11 +251,14 @@ func prepareAssets(accounts map[string]uint64, client libgoal.Client, cfg PpConf
 					if !cfg.Quiet {
 						fmt.Printf("Opting in assets to account %v \n", addr2)
 					}
-					_, addrErr2 := client.AccountInformation(addr2)
+					acct, addrErr2 := client.AccountInformation(addr2)
 					if addrErr2 != nil {
 						fmt.Printf("Cannot lookup optin account\n")
 						err = addrErr2
 						return
+					}
+					if len(acct.Assets) >= proto.MaxAssetsPerAccount {
+						break
 					}
 
 					// opt-in asset k for addr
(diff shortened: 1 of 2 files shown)


shared/pingpong/accounts.go
```diff
@@ -228,12 +251,15 @@ func prepareAssets(accounts map[string]uint64, client libgoal.Client, cfg PpConf
...
 						fmt.Printf("Opting in assets to account %v \n", addr2)
 					}
-					_, addrErr2 := client.AccountInformation(addr2)
+					acct, addrErr2 := client.AccountInformation(addr2)
 					if addrErr2 != nil {
 						fmt.Printf("Cannot lookup optin account\n")
 						err = addrErr2
 						return
 					}
+					if len(acct.Assets) >= proto.MaxAssetsPerAccount {
```
algonautshant, line 260: Make the receiving address loop the outer loop, so that once the limit is reached, it can continue to the next address, instead of coming here every time while the limit is reached. 
In other words, for each "addr", for each "k" in addr, addr2 that reached the limit, will visit this line and break.

shared/pingpong/accounts.go
```diff
@@ -288,7 +335,13 @@ func prepareAssets(accounts map[string]uint64, client libgoal.Client, cfg PpConf
 
 			assetAmt := assetParams[k].Total / uint64(len(accounts))
 			for addr2 := range accounts {
-				if addr != addr2 {
+				optedIn := false
```
algonautshant, line 338: Why not `continue` here if `addr == addr2`?

algonautshant (dismissed):
Looks good. 
I have a few suggestions and questions.

shared/pingpong/accounts.go
```diff
@@ -186,21 +194,36 @@ func prepareAssets(accounts map[string]uint64, client libgoal.Client, cfg PpConf
...
+	// wait until all the assets created
+	for addr := range accounts {
+		var account v1.Account
+		for {
+			account, err = client.AccountInformation(addr)
+			if err != nil {
+				fmt.Printf("Warning, cannot lookup source account")
+				return
+			}
+			if len(account.AssetParams) >= assetsInAcct[addr] {
```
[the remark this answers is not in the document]
algorandskiy: The idea is we know that we should create `assetsInAcct[addr]` assets and we wait until `len(account.AssetParams)` has all of them - it can have less if creation transactions are pending.
But. you are right, `assetsInAcct[addr]` needs to be initialized to `int(cfg.NumAsset) + len(addrAccount.AssetParams)` sinse some assets might pre-exist.

shared/pingpong/accounts.go
algorandskiy: because it was written this way. I'll fix, it makes sence.

shared/pingpong/accounts.go
algorandskiy: completely removed 3 nested looks and do only two - on assets and on accounts

algorandskiy: well, it actually does not matter, same complexity

shared/pingpong/accounts.go
```diff
@@ -186,133 +194,156 @@ func prepareAssets(accounts map[string]uint64, client libgoal.Client, cfg PpConf
...
+	// wait until all the assets created
+	allAssets := make(map[uint64]string, cfg.NumAsset)
+	for addr := range accounts {
+		var account v1.Account
+		start := time.Now()
+		for {
+			account, err = client.AccountInformation(addr)
+			if err != nil {
+				fmt.Printf("Warning, cannot lookup source account")
+				return
```
egieseke, line 208: Maybe it is better to record the error and continue rather than return, to make it more robust.

shared/pingpong/accounts.go
```diff
@@ -186,133 +194,156 @@ func prepareAssets(accounts map[string]uint64, client libgoal.Client, cfg PpConf
...
+			}
+			if len(account.AssetParams) >= numCreatedAssetsByAddr[addr] {
+				break
+			}
+			if time.Now().After(start.Add(3 * time.Minute)) {
+				err = fmt.Errorf("asset creation took too long")
+				fmt.Printf("Error: %s\n", err.Error())
+				return
```
egieseke, line 216: same here.

egieseke (commented):
Not sure it is a good idea to return on error.  It may cause the ping pong test to fail.

shared/pingpong/accounts.go
```diff
@@ -186,133 +194,156 @@ func prepareAssets(accounts map[string]uint64, client libgoal.Client, cfg PpConf
...
+		for {
+			account, err = client.AccountInformation(addr)
+			if err != nil {
+				fmt.Printf("Warning, cannot lookup source account")
+				return
+			}
+			if len(account.AssetParams) >= numCreatedAssetsByAddr[addr] {
+				break
+			}
+			if time.Now().After(start.Add(3 * time.Minute)) {
```
tsachiherman, line 213: create a `deadline = start.Add(3 * time.Minute)` outside the for loop and test against it.

shared/pingpong/accounts.go
```diff
@@ -186,133 +194,156 @@ func prepareAssets(accounts map[string]uint64, client libgoal.Client, cfg PpConf
...
+				err = fmt.Errorf("asset creation took too long")
+				fmt.Printf("Error: %s\n", err.Error())
+				return
+			}
+			time.Sleep(time.Second)
```
tsachiherman, line 218: Sleeping here is not what you really want to do. instead, you want to track the current round number and call `WaitForRound`. Before the round would change, there is no reason to keep polling the asset information.

shared/pingpong/accounts.go
```diff
@@ -186,133 +194,156 @@ func prepareAssets(accounts map[string]uint64, client libgoal.Client, cfg PpConf
...
-		account, accountErr := client.AccountInformation(addr)
-		if accountErr != nil {
-			fmt.Printf("Cannot lookup source account %v\n", addr)
-			err = accountErr
-			return
-		}
+	}
 
+	// wait until all the assets created
+	allAssets := make(map[uint64]string, cfg.NumAsset)
```
tsachiherman, line 200: That's an interesting way of doing it. I'm not sure it's the way I would do that - 
I think it would be more robust to send a transaction and verify that my transaction was included in a block. This way, there is no need to poke into the account. ( i.e. the account could have modified by a another transaction which wasn't sent in this context ).

shared/pingpong/accounts.go
```diff
@@ -186,133 +194,156 @@ func prepareAssets(accounts map[string]uint64, client libgoal.Client, cfg PpConf
...
-			// append the asset to the result assets
-			resultAssetMaps[k] = assetParams[k]
+
+			totalSent++
+			throttleTransactionRate(startTime, cfg, totalSent)
 		}
+		// append the asset to the result assets
+		resultAssetMaps[k] = assetParams[k]
 	}
 	time.Sleep(time.Second * 10)
```
tsachiherman, line 348: I'm not sure what's the purpose of the 10 second sleep. If we want to make sure that all our transactions were included in a block(s), we should test that explicitly.

shared/pingpong/pingpong.go
```diff
@@ -384,6 +384,11 @@ func sendFromTo(
 	amt := cfg.MaxAmt
 	fee := cfg.MaxFee
 
+	assetsByCreator := make(map[string][]*v1.AssetParams)
```
tsachiherman, line 387: nit: size the map according to the size of the cinfo.AssetParams

tsachiherman (commented):
The fact that you need to query the account using `client.AccountInformation` is a failure point. I think that sending transaction and waiting for it to be included in a block is the more robust approach. If the transaction is not included in a block after a while, we have a serious issue that we need to address - but that goes outside of the scope of the pingpong.

shared/pingpong/accounts.go
algorandskiy: there are too many transactions to store IDS and lookup in blocks. Plus, we need to know asset id that does not appear in txn we are looking for (it is a tx counter but it is implementation details).

shared/pingpong/accounts.go
algorandskiy: done

shared/pingpong/accounts.go
algorandskiy: done

shared/pingpong/accounts.go
algorandskiy: done

shared/pingpong/accounts.go
algorandskiy: done

shared/pingpong/accounts.go
algorandskiy: yeah, I think that was the intention.

algorandskiy: done

shared/pingpong/pingpong.go
algorandskiy: the problem is exact size is not known, each address creates multiple assets. So let it grow.

algorandskiy:
Fixes:
* Continue on account info lookup error
* Use WaitForRound to reduce number of iterations
* Add a waiter to the end of asset creation

egieseke:
Note that one error condition when running ping-pong script is due to the transaction pool overflowing and rejecting new transactions.   This is a problem when setting up for asset tests, which requires many create and option asset transactions.  Also, you don't want to take too long preparing for the tests, so that you can measure the actual performance of asset exchanges and also the performance of the smart contracts.

algorandskiy:
With this fix we'll be sending 500 (accounts) * 1000 (max asset) opt ins instead of 500 * 500 * 100 transactions for the old version, that is 50 times less, and significant reduce of the pressure on txnpool.

tsachiherman (approved):
looks good to me. thanks for the changes!

PR #1728, opened 2020-11-30
Fix DNSBootstrapArray returning empty strings

tsachiherman:
## Summary

The `DNSBootstrapArray` function is expected to return a list of the DNS bootstrap entries. Internally, it converts the config file DNSBootstrapID using a list of DNS bootstrap entries. However, in the case of having the `DNSBootstrapID` set to an empty string, the function would also return an array containing an empty string.

The above behavior, in turn, trigger a downstream failure when searching for the telemetry SRV records.

## Test Plan

Update a unit test to cover this use case.


brianolson (approved):

tsachiherman:
Thanks @brianolson

algorandskiy (approved):

PR #1732, opened 2020-11-30
Fix a race in arm test in algoh startup and algod.* files creation

algorandskiy:
## Summary

Spawned algoh returns too fact, and despite the fix for `goal node start` startup code the test still fails:
```
algoh startup successful
spawn goal node wait -d /home/ubuntu/go/src/github.com/algorand/go-algorand/tmp/out/e2e/92019-1606765255748/algohTimeoutTest/algod/root/Node -w 61
Cannot contact Algorand node: open /home/ubuntu/go/src/github.com/algorand/go-algorand/tmp/out/e2e/92019-1606765255748/algohTimeoutTest/algod/root/Node/algod.token: no such file or directory
```
Added a file waiter for `algoh` startup.

## Test Plan

This is test fix

### DIFF
--- a/test/e2e-go/cli/algoh/expect/algohTimeoutTest.exp
+++ b/test/e2e-go/cli/algoh/expect/algohTimeoutTest.exp
@@ -55,6 +55,7 @@ if { [catch {
     expect {
         "^Logging to: *" {puts "algoh startup successful"}
         timeout {::Algoh::Abort "algoh failed to start";}
+        eof { ::Algoh::CheckEOF "algoh failed to start" }
     }
 
     # allow algoh time to put files in Node dir
--- a/test/e2e-go/cli/algoh/expect/algohTimeoutTest.exp
+++ b/test/e2e-go/cli/algoh/expect/algohTimeoutTest.exp
@@ -57,6 +57,9 @@ if { [catch {
         timeout {::Algoh::Abort "algoh failed to start";}
     }
 
+    # allow algoh time to put files in Node dir
+    ::Algoh::WaitForFile $TEST_NODE_DIR/algod.pid
+
     # wait until Node approves blocks to the network
     set timeout 60
     spawn goal node wait -d $TEST_NODE_DIR -w 61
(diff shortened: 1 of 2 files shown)


test/e2e-go/cli/algoh/expect/algohTimeoutTest.exp
```diff
@@ -57,6 +57,9 @@ if { [catch {
         timeout {::Algoh::Abort "algoh failed to start";}
     }
 
+    # allow algoh time to put files in Node dir
+    ::Algoh::WaitForFile $TEST_NODE_DIR/algod.pid
+
```
tsachiherman, line 62: It's ok to wait for the file, but maybe the reason for this is because the above spawn should have waited until the EOF have reached ?

tsachiherman (commented):
I think that you might want to revise what is done in the expect portion; we need to wait for EOF before going forward.

test/e2e-go/cli/algoh/expect/algohTimeoutTest.exp
algorandskiy: I always thought `spawn` waits.

algorandskiy:
Seems like the last failure left
```
sleep time 9
spawn node status
node status check complete, current round is 0
Current Round: '0' is less than wait for round: '1'
Aborting with Error:  Current Round 0 did not reach 1
...
--- FAIL: TestGoalWithExpect/catchpointCatchupTest.exp (128.12s)
```

test/e2e-go/cli/algoh/expect/algohTimeoutTest.exp
tsachiherman: no, spawn definitely doesn't wait. When you use expect, it *could* be waiting if you include an eof section.
also, there is no error code testing here, which is needed. ( i.e. error code testing implies waiting. )

tsachiherman: I'm ok in merging this, but not sure if the  ::Algoh::WaitForFile is still needed.

algorandskiy: It is needed, spawn returns earlier than files gets created, and we do have tools to control spawn in a way we did with `goal node start`

tsachiherman (approved):
Looks good.

PR #1733, opened 2020-11-30
Improve error logging of elastic search telemetry module

tsachiherman:
## Summary

This PR addresses several key locations which lacked error message printouts.

## Test Plan

These changes were tested manually. Automating a test for these seems to be quite an intensive task, as it requires to emulate an elastic search server..

### DIFF
--- a/logging/telemetryhook.go
+++ b/logging/telemetryhook.go
@@ -202,21 +202,24 @@ func (hook *dummyHook) waitForEventAndReady() bool {
 	return true
 }
 
+// the elasticClientLogger is used to bridge the elastic library error reporting
+// into our own logging system.
 type elasticClientLogger struct {
-	elastic.Logger
-	level logrus.Level
+	logger Logger       // points to the underlying logger which would perform the logging
+	level  logrus.Level // indicate what logging level we want to use for the logging
 }
 
+// Printf tunnel the log string into the log file.
 func (el elasticClientLogger) Printf(format string, v ...interface{}) {
 	switch el.level {
 	case logrus.DebugLevel:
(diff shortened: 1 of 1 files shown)


logging/telemetryhook.go
```diff
@@ -87,7 +86,10 @@ func createAsyncHookLevels(wrappedHook logrus.Hook, channelDepth uint, maxQueueD
 					}
 					hook.Unlock()
 					if entry != nil {
-						hook.wrappedHook.Fire(entry)
+						err := hook.wrappedHook.Fire(entry)
```
algonautshant, line 89: Do we also need to capture the error returned from Fire here (in func logTelemetry)?
logging/telemetry.go 228:	t.hook.Fire(entry)

algonautshant (approved):
Looks good. 
Just one comment.

logging/telemetryhook.go
```diff
@@ -199,6 +202,24 @@ func (hook *dummyHook) waitForEventAndReady() bool {
 	return true
 }
 
+type elasticClientLogger struct {
+	elastic.Logger
```
brianolson, line 206: to be a elastic logger just needs Printf, which always points to Base(), so I think we don't need this part of the struct.

logging/telemetryhook.go
tsachiherman: The Fire function on the `logging/telemetry.go 228` goes to `asyncTelemetryHook.Fire` which aways returns nil, so adding a logging there is correct, but won't really be helpful.

brianolson (approved):

Issue #1190, opened 2020-06-21
Node doesn't sync.


Myridium:
### Subject of the issue
The node doesn't sync. Similar issue to #571, probably.

### Your environment
* Software version: 
```
8590010404
2.0.75812.dev, commit #eeedd36d6
```
* Node status if applicable:
```
Last committed block: 0
Time since last block: 0.0s
Sync Time: 0.0s
Last consensus protocol: https://github.com/algorandfoundation/specs/tree/5615adc36bad610c7f165fa2967f4ecfa75125f0
Next consensus protocol: https://github.com/algorandfoundation/specs/tree/5615adc36bad610c7f165fa2967f4ecfa75125f0
Round for next consensus protocol: 1
Next consensus protocol supported: true
Last Catchpoint: 
Genesis ID: mainnet-v1.0
Genesis hash: wGHE2Pwdvd7S12BL5FaOP20EGYesN73ktiC1qzkkit8=
```

* Operating System details. 
Arch Linux x64

* Last part of `node.log`:
```
{"Context":"Agreement","Hash":"AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA","ObjectPeriod":0,"ObjectRound":0,"ObjectStep":0,"Period":0,"Round":1,"Sender":"","Step":9,"Type":"VoteAttest","Weight":0,"WeightTotal":0,"file":"actions.go","function":"github.com/algorand/go-algorand/agreement.pseudonodeAction.do","level":"info","line":372,"msg":"attested to {{} 0 AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAY5HFKQ AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA} at (1, 0, 9)","time":"2020-06-21T22:40:54.165248+10:00"}
{"file":"wsNetwork.go","function":"github.com/algorand/go-algorand/network.(*WebsocketNetwork).meshThread","level":"info","line":1321,"msg":"got no DNS addrs for network mainnet","name":"","time":"2020-06-21T22:41:14.763916+10:00"}
[...]

PR #1736, opened 2020-12-02
Use OS and fallback DNS servers in DNSSEC resolver

algorandskiy:
## Summary

This addresses #1190 by unifying DNSSEC resolver with a regular resolver: try system, try fallback if available, try default hardcoded, error.

Changes:
1. General refactoring: decouple DNSSEC to components and move tests around, remove some dead code, better naming and encapsulation.
2. Introduce ResolveController to ReadFromSRV in order to abstract resolution attempts from concrete resolver
3. New ResolverIf interface is compatible with net.Resolver and is implemented by network.Resolver and dnssec.Resolver

## Test Plan

Existing test suite passed. Added new tests, including testing real network with system, fallback and default configurations.


network/wsNetwork.go
```diff
@@ -581,9 +580,9 @@ func (wn *WebsocketNetwork) GetPeers(options ...PeerOption) []Peer {
 }
 
 func (wn *WebsocketNetwork) setup() {
-	var preferredResolver *dnssec.Resolver
+	var preferredResolver dnssec.ResolverIf
 	if wn.config.DNSSecurityRelayAddrEnforced() {
-		preferredResolver = &dnssec.DefaultResolver
+		preferredResolver = dnssec.MakeDefaultDnssecResolver(wn.config.FallbackDNSResolverAddress)
```
algorandskiy, line 585: `MakeDefaultDnssecResolver` has stuff from `/etc/resolv.conf` and `FallbackDNSResolverAddress` that might fail to handle DNSSEC requests. Maybe it makes sense to add track failed server and do not query them again to speedup things

tools/network/bootstrap.go
```diff
@@ -37,35 +35,28 @@ func ReadFromSRV(service string, protocol string, name string, fallbackDNSResolv
...
+		log.Infof("ReadFromBootstrap: DNS LookupSRV failed when using system resolver: %v", sysLookupErr)
+
+		var fallbackLookupErr error
+		if fallbackDNSResolverAddress != "" {
+			fallbackResolver := controller.FallbackResolver()
+			_, records, fallbackLookupErr = fallbackResolver.LookupSRV(context.Background(), service, protocol, name)
 		}
-	} else {
-		var sysLookupErr error
-		_, records, sysLookupErr = net.LookupSRV(service, protocol, name)
```
algorandskiy, line 50: `net.LookupSRV` is the same as `net.DefaultResolver. LookupSRV(ctx, ...)`

tools/network/bootstrap.go
```diff
@@ -37,35 +35,28 @@ func ReadFromSRV(service string, protocol string, name string, fallbackDNSResolv
...
-			var resolver Resolver
-			// try to resolve the address. If it's an dotted-numbers format, it would return that right away.
-			// if it's a named address, we would attempt to parse it and might fail.
-			// ( failing isn't that bad; the resolver would internally try to use 8.8.8.8 instead )
-			if DNSIPAddr, err2 := net.ResolveIPAddr("ip", fallbackDNSResolverAddress); err2 == nil {
-				resolver.DNSAddress = *DNSIPAddr
-			} else {
-				log.Infof("ReadFromBootstrap: Failed to resolve fallback DNS resolver address '%s': %v; falling back to default fallback resolver address", fallbackDNSResolverAddress, err2)
-			}
+		if fallbackLookupErr != nil || fallbackDNSResolverAddress == "" {
```
tsachiherman, line 50: this doesn't align with the previous : If the ` fallbackLookupErr != nil`, it's already guarantee that `fallbackDNSResolverAddress != ""`. Also, please add the `fallbackDNSResolverAddress` to the info message.

tools/network/bootstrap.go
```diff
@@ -37,35 +35,28 @@ func ReadFromSRV(service string, protocol string, name string, fallbackDNSResolv
...
+			log.Infof("ReadFromBootstrap: DNS LookupSRV failed when using fallback resolver: %v", fallbackLookupErr)
 
-			_, records, err = resolver.LookupSRV(context.Background(), service, protocol, name)
-			if err != nil {
-				err = fmt.Errorf("ReadFromBootstrap: DNS LookupSRV failed when using system resolver(%v) as well as via %s due to %v", sysLookupErr, resolver.EffectiveResolverDNS(), err)
+			fallbackResolver := controller.DefaultResolver()
+			var defaultLookupErr error
+			_, records, defaultLookupErr = fallbackResolver.LookupSRV(context.Background(), service, protocol, name)
+			if defaultLookupErr != nil {
+				err = fmt.Errorf("ReadFromBootstrap: DNS LookupSRV failed when using system resolver(%v), fallback resolver(%v), as well as using default DNS due to %v", sysLookupErr, fallbackLookupErr, defaultLookupErr)
```
tsachiherman, line 57: `default DNS` => `default resolver`

tools/network/bootstrap.go
algorandskiy: Old implementation internally used `resolver.DNSAddress` if set (fallback) or its default resolver otherwise.
New implementation explicitly checks for fallback address and uses it if set, or continues with default servers otherwise. So a new implementation will not fail if fallback is set to a dead server.

tools/network/resolveController.go
```diff
@@ -0,0 +1,88 @@
...
+
+// ResolverIf is re-import from dnssec.ResolverIf
+type ResolverIf interface {
+	dnssec.ResolverIf
+}
+
+// compile-time check
+var _ dnssec.ResolverIf = &net.Resolver{}
+var _ dnssec.ResolverIf = &Resolver{}
+var _ dnssec.ResolverIf = &dnssec.Resolver{}
```
tsachiherman, line 35: nit: add these compile-time checks to the `XXX_test.go` file so that they won't get instantiated on the "production" binary.

tools/network/resolveController.go
```diff
@@ -0,0 +1,88 @@
...
+}
+
+// NewResolveController creates a new ResolveController
+func NewResolveController(secure bool, fallbackDNSResolverAddress string) ResolveController {
+	return ResolveController{secure, fallbackDNSResolverAddress}
+}
+
+// SystemResolver returns a resolver that uses OS-defined DNS servers
+func (c *ResolveController) SystemResolver() ResolverIf {
+	log := logging.Base()
```
tsachiherman, line 50: add the Logger to the ResolveController and have it passed during constructor.

tools/network/resolveController.go
```diff
@@ -0,0 +1,88 @@
...
+	return net.DefaultResolver
+}
+
+// FallbackResolver returns a resolver that uses fallback DNS address
+func (c *ResolveController) FallbackResolver() ResolverIf {
+	log := logging.Base()
+	var dnsIPAddr *net.IPAddr
+	var err error
+	if dnsIPAddr, err = net.ResolveIPAddr("ip", c.fallback); err != nil {
+		log.Debugf("resolving fallback %s failed with %s", c.fallback, err.Error())
```
tsachiherman, line 69: add single-quotes around printing the `c.fallback` string, for better readability of space characters

tools/network/dnssec/config.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+	return systemConfig(f)
+}
+
+func systemConfig(configFile io.Reader) (servers []ResolverAddress, timeout time.Duration, err error) {
+	if configFile == nil {
+		err = fmt.Errorf("empty config reader")
+		return
+	}
+	cc, err := dns.ClientConfigFromReader(configFile)
```
tsachiherman, line 77: it's interesting that this function has hard-coded "53" as the returned Port number.

tools/network/dnssec/config.go
```diff
@@ -0,0 +1,89 @@
...
+	if err != nil {
+		return
+	}
+	for _, addr := range cc.Servers {
+		servers = append(servers, MakeResolverAddress(addr, cc.Port))
+	}
+	timeout = DefaultTimeout
+	if cc.Timeout != 0 && len(servers) > 0 {
```
tsachiherman, line 85: Could you add an upper bound for the Timeout here ? The `dns` library doesn't provide any upper bound for that.

tools/network/dnssec/resolver.go
```diff
@@ -23,53 +44,63 @@ import (
...
+	dc := MakeDNSClient(servers, timeout)
+	tc := &QueryWrapper{dc}
+	return &Resolver{client: dc, trustChain: makeTrustChain(tc), maxHops: DefaultMaxHops}
+}
 
-	if len(rs.servers) == 0 {
-		rs.servers = append(rs.servers, defaultDnssecAwareNSServers...)
+// MakeDefaultDnssecResolver returns a resolver with all possible DNS servers:
+// system, fallback, default
+func MakeDefaultDnssecResolver(fallbackAddress string) ResolverIf {
```
tsachiherman, line 80: same note regarding the logger as before.

tools/network/dnssec/trustedzone.go
```diff
@@ -0,0 +1,208 @@
...
+}
+
+// verifyDS checks DS RRSIG using this zone ZSK
+// cacheOutdated indicates no ZSK in this zone to verify the provided RRSIG
+func (tz *trustedZone) verifyDS(rrSet []dns.RR, rrSig []dns.RRSIG, t time.Time) (cacheOutdated bool, err error) {
+	// parent zone's DNSKEY RRSET is cached but DS is just came from the network
+	// and might be signed by newer key
+	kt := make([]uint16, 0, len(rrSig))
+	for _, sig := range rrSig {
+		kt = append(kt, sig.KeyTag)
```
tsachiherman, line 62: nit: `kt[i] = sig.KeyTag` would work faster.

tools/network/dnssec/trustedzone.go
```diff
@@ -0,0 +1,208 @@
...
+	if !tz.checkKeys(kt) {
+		cacheOutdated = true
+		return
+	}
+
+	// at least one signature must be valid
+	verifiedSig := verifyRRSig(rrSet, rrSig, t, tz.zsk)
+	if len(verifiedSig) == 0 {
+		requestedZone := (rrSig)[0].Hdr.Name
+		err = fmt.Errorf("DS signature verification failed for %s", requestedZone)
```
tsachiherman, line 74: format the requestedZone within single quotes.

tsachiherman (commented):
Few small requests, but looks great otherwise.

tools/network/resolveController.go
algorandskiy: they will be stripped out anyway since not used

tools/network/dnssec/config.go
algorandskiy: done

algorandskiy (commented):
Fixed review notes.

tools/network/dnssec/trustedzone.go
algorandskiy: done

tools/network/dnssec/trustedzone.go
algorandskiy: done

tools/network/dnssec/resolver.go
algorandskiy: done

tools/network/resolveController.go
algorandskiy: done

tools/network/resolveController.go
algorandskiy: done

tools/network/bootstrap.go
algorandskiy: done

tsachiherman (approved):
Looks good, thanks for the changes.

vishalgupta96:
Hi Team, 

@algorandskiy , @tsachiherman ... think my algorand node stopped syncing at Block#11665201 because of these recent changes. I was running algorand:2.3.0 earlier and we upgraded it to 2.4.1 also but it did not help.

Below is the config file i am using and the error message i can see in logs:
Error:

{"file":"bootstrap.go","function":"github.com/algorand/go-algorand/tools/network.ReadFromSRV","level":"info","line":43,"msg":"ReadFromBootstrap: DNS LookupSR
V failed when using system resolver: no answer for (_algobootstrap._tcp.mainnet.algorand.network., 33) from DNS servers [127.0.0.11:53]","time":"2021-02-13T0
4:07:41.698141Z"}
{"file":"wsNetwork.go","function":"github.com/algorand/go-algorand/network.(*WebsocketNetwork).meshThread","level":"info","line":1341,"msg":"got no DNS addrs
 for network mainnet","name":"127.0.0.1:4160","time":"2021-02-13T04:07:47.700072Z"}


Config.json:
{
    "Archival": true,
    "isIndexerActive": true,
    "GossipFanout": 4,
    "BaseLoggerDebugLevel": 1,
    "IncomingConnectionsLimit": -1,
    "NetAddress": "127.0.0.1:4160",
    "EndpointAddress": "0.0.0.0:8887",
    "LogSizeLimit": 1073741824,
    "EnableMetricReporting": false,
    "EnableOutgoingNetworkMessageFiltering": true,
    "EnableIncomingMessageFilter": false,
    "CatchupParallelBlocks": 50,
    "DNSBootstrapID": "mainnet.algorand.network",
    "DNSSecurityFlags": 0
}

Please help asap... its impacting our release, need some resolution within a day or two.

tsachiherman:
Could you try to drop the DNSXXX entries from the config file, add a
Version 12 entry and try again ?

vishalgupta96:
@tsachiherman : Hi , sorry i could not get what you are suggesting here?

Are you suggesting to remove below entries from config file?
"DNSBootstrapID": "mainnet.algorand.network",
"DNSSecurityFlags": 0

Actually we have tried removing them but we are still facing the same issue. 
Which "Version 12" entry you are asking us to add? Could you please specify some details.

Or pls provide the config file i should use?

tsachiherman:
If you'll look at any of the config.json.example files, you'll see that it always starts with a version field:
```
{
    "Version": 15,
    "AccountsRebuildSynchronousMode": 1,
    "AnnounceParticipationKey": true,
    ...
```

The version field is used for supporting configuration file upgrades. In your example, the version field was missing.
Could you try to see if having this field resolves the issue ?
( I'm suspecting that the two changes you've made to `DNSBootstrapID` and `DNSSecurityFlags` would no longer be required )

vishalgupta96:
Hi @tsachiherman ... I can't see a config.json.example file. But as you suggested, I tried adding "Version" : 12 , and I am getting below error now in logs:

{"Context":"Agreement","Hash":"","ObjectPeriod":0,"ObjectRound":0,"ObjectStep":0,"Period":0,"Round":11665202,"Sender":"","Step":15,"Type":"StepTimeout","Weight":0,"WeightTotal":0,"file":"trace.go","function":"github.com/algorand/go-algorand/agreement.(*tracer).logFastTimeout","level":"info","line":204,"msg":"timeout fired on (11665202, 0, 15) with value 3h12m40.602871493s (napping: false)","time":"2021-02-15T13:52:38.070762Z"}

image

tsachiherman:
That's an info message. It's not an indication of any issue. try to see if the node is catching up.

algorandskiy:
It looks like the following happened:
1. Before this change DNSSEC was used public DNSSEC-aware resolvers.
2. After this change the code tries system-configured resolver as DNSSEC first, then fallback from config, and then the public resolvers are before. So it should word as before.

Couple questions:
1. Do you have 8.8.8.8:53 available from your network?
2. What are other log entries from `ReadFromSRV` func?

vishalgupta96:
> That's an info message. It's not an indication of any issue. try to see if the node is catching up.

@tsachiherman : No it is not syncing.. it is stuck at 11665201.

@algorandskiy : I am able to ping 8.8.8.8 from my VM and able to get the reply... but when i tried to ping 8.8.8.8 53 .. i am unable to connect it seems. Could you give some more background about these and do we need to have 53 port opened?

algorandskiy:
There is a list of public DNSSEC-aware servers that are used when both system and fallback (from the config file) fail.
```
var DefaultDnssecAwareNSServers = []ResolverAddress{"1.1.1.1:53", "208.67.222.222:53", "8.8.8.8:53", "77.88.8.8:53", "8.26.56.26:53", "180.76.76.76:53"}
```

algorandskiy:
@vishalgupta96 Could you please create an issue and attach
1. part of the log from the restart until stuck
2. a config file
3. `aglod -v` output

PR #1737, opened 2020-12-02
Fix goalAsset expect test

tsachiherman:
## Summary

The expect test had few bugs; all of the following bugs were addressed : 
* Missing wait for round 1 before sending transaction. This is required since nodes at startup might take a short while before network start making progress. We need network to make progress so that the timeout on the subsequent command would have some known value.
* The asset create should have called with expect instead of exec so that we can monitor it's output.
* The asset info calls should have been tested for negative return value.
* The network should have been stopped at the end of the test.
* In case of an error, the test was outputting "goalFormattingTest" message instead of suggesting the correct test name ( i.e. goalAssetTest )

## Test Plan

This is a test.


algorandskiy (approved):

algorandskiy:
Seems like there is one more
```
One or more than one KMD instances remains running:
ubuntu   28414     1  0 06:11 ?        00:00:00 /home/ubuntu/go/bin-race/kmd -d /home/ubuntu/go/src/github.com/algorand/go-algorand/tmp/out/e2e/9225-1606885944829/multisigCreationDeletionTest/algod/root/Primary/kmd-v0.5 -t 60
```

PR #1739, opened 2020-12-02
Small transaction pool rebuild & assembly optimizations

tsachiherman:
## Summary

Optimize new round block evaluator rebuild performance, and improve block assembly deadline accuracy.

## Test Plan

Existing tests provides sufficient coverage for the functionality. The changes were tested on the performance system and demonstrated minor gains ( as expected ) and notable improved block assembly accuracy.


data/pools/transactionPool.go
```diff
@@ -132,18 +132,28 @@ type poolAsmResults struct {
 	// roundStartedEvaluating is the round which we were attempted to evaluate last. It's a good measure for
 	// which round we started evaluating, but not a measure to whether the evaluation is complete.
 	roundStartedEvaluating basics.Round
+	// assemblyComplete is *not* protected via the pool.assemblyMu lock and should be accessed only from the OnNewBlock goroutine.
+	// it's equivilent to the "ok" variable, and used for avoiding taking the lock.
+	assemblyComplete bool
```
algonautshant, line 137: Maybe call this assemblyCompletedOrAbandoned to reflect that fact that the block might not be really assembled when this is set to true.

data/pools/transactionPool.go
```diff
@@ -584,11 +607,21 @@ func (pool *TransactionPool) addToPendingBlockEvaluatorOnce(txgroup []transactio
...
+		if !pool.assemblyResults.assemblyComplete {
+			pool.assemblyMu.Lock()
+			defer pool.assemblyMu.Unlock()
+			if pool.assemblyRound > pool.pendingBlockEvaluator.Round() {
+				// the block we're assembling now isn't the one the the AssembleBlock is waiting for. While it would be really cool
+				// to finish generating the block, it would also be pointless to spend time on it.
+				// we're going to set the ok and assemblyComplete to "true" so we can complete this loop asap
+				pool.assemblyResults.ok = true
+				pool.assemblyResults.assemblyComplete = true
+				stats.StopReason = telemetryspec.AssembleBlockAbandon
```
algonautshant, line 619: is this used somewhere?

tsachiherman, line 619: This one rarely happen on normal cases, but it does occur when doing catchup.

data/pools/transactionPool.go
```diff
@@ -558,6 +568,19 @@ func (pool *TransactionPool) OnNewBlock(block bookkeeping.Block, delta ledger.St
...
+// or whether we've ran out of time. It takes into consideration the assemblyDeadline that was set by the AssembleBlock function as well as the
+// projected time it's going to take to call the GenerateBlock function before the block assembly would be ready.
+// The function expects that the pool.assemblyMu lock would be taken before being called.
+func (pool *TransactionPool) isAssemblyTimedOut() bool {
+	if pool.assemblyDeadline.IsZero() {
+		// we have no deadline, so no reason to timeout.
+		return false
+	}
+	generateBlockDuration := generateBlockBaseDuration + time.Duration(pool.pendingBlockEvaluator.TxnCounter())*generateBlockTransactionDuration
+	return time.Now().After(pool.assemblyDeadline.Add(-generateBlockDuration))
```
algonautshant, line 581: Why is this '-' generatBlockDuration?

tsachiherman: I think we already covered this in another PR.. the `time.Add` returns a `time.Time`; the `time.Sub` returns a `time.Duration`. If we want to subtract a duration from a `time.Time` and get a `time.Time`, the correct function is the `time.Add` and passing a negative `time.Duration`.

data/pools/transactionPool.go
tsachiherman: I don't believe so; however, just for completeness, I figure that I should update the statistics so that they won't be left uninitialized while the `ok` variable was set.

data/pools/transactionPool.go
tsachiherman: I'm fine with changing this to assemblyCompletedOrAbandoned, although I don't think it adds much information. ( I'll wait for the other reviewers to comment here before making this change )

data/pools/transactionPool.go
algonautshant: My concern is something else. 
The way I understand this is that generateBlockDuration is based on generateBlockTransactionDuration, which I assume, is an upper limit.
And using the upper limit for subtracting from the allowed time, might be subtracting too much, and triggering a timeout.

tsachiherman: These numbers were tuned by trial and error; the block assembly time in the case of a timeout, even with these changes is always more than 250ms, suggesting that these values are well within reason.
Note that if you'll do the math, you'll see that the `generateBlockDuration` never ends up being more than 13ms on a block of a 4650 transactions ( which is our current limit ).

This change is important as we're trying to push for higher transaction count, and need to have a better accuracy when stopping to build the block.

algonautshant (approved):
Thank you for the detailed explanations.
Looks great!

data/pools/transactionPool.go
tsachiherman: done.

algonautshant (approved):

PR #1740, opened 2020-12-03
Fix releases-page task

btoll:
## Summary

This is the last major piece of the package build pipeline that needed to be completed.  It fixes the releases-page `mule` task as well as setting the `acl` for every synced object to "public-read".

Removed outdated docs.

## Test Plan

```
NETWORK=betanet VERSION=2.3.0 mule -f package-deploy.yaml releases-page
```
( However, don't do this or it will create a new key in s3 that is public-facing. )


onetechnical (approved):
Didn't run it but looks fine

egieseke (approved):
Looks good.

PR #1742, opened 2020-12-03
Fix random limit order test

tsachiherman:
## Summary

The asset transfer goal command should have a timeout of an order of few rounds, rather than 5 seconds. This is needed since we wait for the transaction to be included in a block, and there is no guarantee that the transaction would make it to the next block.

## Test Plan

This is a test.


algorandskiy (commented):
Why permissions change?

tsachiherman:
> Why permissions change?

making all the exp files to have the same permissions.

algorandskiy (approved):

PR #1743, opened 2020-12-03
benchmark account db random reads and BlockEvaluator

brianolson:
## Summary

Add Benchmarks for account db random reads from sqlite3 ram and disk.
Add Benchmark for BlockEvaluator building a big block.

## Test Plan

Benchmarks are tests!


ledger/eval_test.go
```diff
@@ -318,3 +318,53 @@ func TestPrepareAppEvaluators(t *testing.T) {
...
+		}
+		st := txn.Sign(keys[sender])
+		err = eval.Transaction(st, transactions.ApplyData{})
+		require.NoError(b, err)
+	}
+
+	validatedBlock, err := eval.GenerateBlock()
+	require.NoError(b, err)
+
+	l.AddValidatedBlock(*validatedBlock, agreement.Certificate{})
```
tsachiherman, line 369: while this benchmark is good on it's own, I think that what we're after is to test the 
```
func (l *Ledger) Validate(ctx context.Context, blk bookkeeping.Block, txcache VerifiedTxnCache, executionPool execpool.BacklogPool) (*ValidatedBlock, error)
```
function which is being used from the agreement. Ideally, if we were to conduct any improvement, we could leave this function signature the same and modify the internals.

ledger/accountdb_test.go
```diff
@@ -730,6 +737,41 @@ func BenchmarkReadingAllBalances(b *testing.B) {
...
+
+	qs, err := accountsDbInit(dbs.rdb.Handle, dbs.wdb.Handle)
+	require.NoError(b, err)
+
+	b.ResetTimer()
+	// read all the balances in the database, shuffled by map-iterator hash order
+	for addr, _ := range accounts {
+		_, _, err = qs.lookup(addr)
+		require.NoError(b, err)
+	}
```
tsachiherman, line 764: since this is a benchmark, you should wrap this portion in a for loop : 
```
for i:=0;i<b.N;i++ {
...
}
```
so that if it takes less than 1 second, go would try to execute it twice.

tsachiherman: nit: add something in the lines of 
```
b.ReportMetric(float64(len(accounts)), "accounts/iteration")
```
so that the benchmark feedback would be clear what is being tested.

ledger/accountdb_test.go
```diff
@@ -712,8 +707,20 @@ func BenchmarkReadingAllBalances(b *testing.B) {
...
+
+func benchmarkReadingAllBalances(b *testing.B, inMemory bool) {
+	proto := config.Consensus[protocol.ConsensusCurrentVersion]
+	//b.N = 50000
+	dbs, _ := dbOpenTest(b, inMemory)
+	setDbLogging(b, dbs)
+	defer dbs.close()
+
+	benchmarkInitBalances(b, dbs, proto)
+	tx, err := dbs.rdb.Handle.Begin()
```
tsachiherman, line 723: the second part of this function doesn't seems to be needed for the benchmark, and looks like a test

brianolson: The part after b.ResetTimer() is the real benchmark, before that is just setup.
The hash might be meaningless. It was like that when I got here I just rearranged it.

ledger/accountdb_test.go
brianolson: `b.N` goes into the number of accounts set up by `benchmarkInitBalances()`, so the loop to read back all the accounts is `b.N` long.

ledger/eval_test.go
brianolson: I expanded the test to use the built block to also run Validate

ledger/eval_test.go
```diff
@@ -318,3 +323,225 @@ func TestPrepareAppEvaluators(t *testing.T) {
...
+	avbDone := time.Now()
+	avbTime := avbDone.Sub(blockBuildDone)
+	b.ReportMetric(float64(avbTime)/float64(numTxns), "ns/AddValidatedBlock_tx")
+
+	// test speed of block validation
+	// This should be the same as the eval line in ledger.go AddBlock()
+	// This is pulled out to isolate eval() time from db ops of AddValidatedBlock()
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
+		_, err = eval(context.Background(), l2, validatedBlock.blk, false, nil, nil)
```
tsachiherman, line 538: it's "almost" right; the fourth parameter should be "true" to have it perform validation.
I think it would be more reliable to just use
```
_, err = l2.Validate(context.Background(), validatedBlock.blk, nil, nil)
```
```

Comment (tsachiherman) [reply ledger/eval_test.go:538]: on unrelated topic, the `(l *Ledger) Validate` looks like it should be moved to `ledger.go`. but that's not really related to your change.

Comment (brianolson) [reply ledger/eval_test.go:538]: Do we care about comparing how much time crypto validation takes in this test? In my mind that's the part that's parallelizeable and probably not the bottleneck (I'm betting the serial txn accounting will be the blocker). I guess what I'll do in the next hour is make it both ways and see how different the time is.

Comment (tsachiherman) [reply ledger/eval_test.go:538]: One number that I'd like to get from this experiment is "how long does it takes to perform a validation in the worst case scenario for a 10MB block". We learned from telemetry that it's about 3 seconds. I want to hear from you how much time does it truly takes. This would help us to determine if this is the bottleneck or not.
A subsequent benchmark would be to see if parallelizing some of that (optimistically) could really reduce time validation time. My guess is "yes", but I don't want to say so before having a concrete test.

Comment (brianolson) [reply ledger/eval_test.go:538]: for a 50,000 txn block, 3,766,519,337 ns aka 3.7 seconds (while using sqlite3 in RAM)

**Comment by brianolson** [comment]
Usage for this is like
```
BLOCK_EVAL_PPROF=1 time go test -run=zzz -bench=BenchmarkBlockEvaluatorRAMCrypto -benchtime 60s
go tool pprof -http :8082 $(ls -t BenchmarkBlockEvaluatorRAM*_cpuprof |head -1)
```

File: ledger/eval_test.go (hunk 4 above)
Comment (tsachiherman) [reply ledger/eval_test.go]: could you summarize the results in the PR summary ? It would be great to have it documented.
I think that the missing validation case is typical for when the txpool is full and we can't accept more txns.

**Comment by brianolson** [comment]
```
go test -run zzz -bench BenchmarkBlock -cpu 1,4,8 -json > /tmp/bench$(date +%Y%m%d_%H%M%S).json
jq -r .Output $(ls -t /tmp/bench*json|head -1)|grep eval_val

BenchmarkBlockEvaluatorRAMCrypto        	       1	3455589469 ns/op	     69113 ns/eval_validate_tx
BenchmarkBlockEvaluatorRAMCrypto-4      	       1	3390001942 ns/op	     67801 ns/eval_validate_tx
BenchmarkBlockEvaluatorRAMCrypto-8      	       1	3382257414 ns/op	     67647 ns/eval_validate_tx
BenchmarkBlockEvaluatorRAMNoCrypto      	       1	1068593703 ns/op	     21373 ns/eval_validate_tx
BenchmarkBlockEvaluatorRAMNoCrypto-4    	       1	1038342142 ns/op	     20768 ns/eval_validate_tx
BenchmarkBlockEvaluatorRAMNoCrypto-8    	       1	1071763836 ns/op	     21437 ns/eval_validate_tx
BenchmarkBlockEvaluatorDiskCrypto       	       1	3436579347 ns/op	     68732 ns/eval_validate_tx
BenchmarkBlockEvaluatorDiskCrypto-4     	       1	3424601686 ns/op	     68493 ns/eval_validate_tx
BenchmarkBlockEvaluatorDiskCrypto-8     	       1	3400353415 ns/op	     68008 ns/eval_validate_tx
BenchmarkBlockEvaluatorDiskNoCrypto     	       1	1303251352 ns/op	     26066 ns/eval_validate_tx
BenchmarkBlockEvaluatorDiskNoCrypto-4   	       1	1304357969 ns/op	     26088 ns/eval_validate_tx
BenchmarkBlockEvaluatorDiskNoCrypto-8   	       1	1305665732 ns/op	     26114 ns/eval_validate_tx```

tsachiherman:
you'll need to lint the PR.

ledger/accountdb_test.go
```diff
@@ -662,19 +662,14 @@ func randomCreatable(uniqueAssetIds map[basics.CreatableIndex]bool) (
...
 }
 
-func BenchmarkReadingAllBalances(b *testing.B) {
-	proto := config.Consensus[protocol.ConsensusCurrentVersion]
-	//b.N = 50000
-	dbs, _ := dbOpenTest(b, true)
-	setDbLogging(b, dbs)
-	defer dbs.close()
-
+func benchmarkInitBalances(b *testing.B, dbs dbPair, proto config.ConsensusParams) (updates map[basics.Address]basics.AccountData) {
```
tsachiherman, line 665: I think it would make it much more flexible if you were to change the function signature to receive a `testing.TB` instead of `testing.B` and add the accounts number as the argument.

ledger/accountdb_test.go
```diff
@@ -662,19 +662,14 @@ func randomCreatable(uniqueAssetIds map[basics.CreatableIndex]bool) (
...
 	tx, err := dbs.wdb.Handle.Begin()
 	require.NoError(b, err)
 
 	secrets := crypto.GenerateOneTimeSignatureSecrets(15, 500)
 	pubVrfKey, _ := crypto.VrfKeygenFromSeed([32]byte{0, 1, 2, 3})
-	updates := map[basics.Address]basics.AccountData{}
+	//updates := map[basics.Address]basics.AccountData{}
```
tsachiherman, line 671: Remove commented out line.

ledger/accountdb_test.go
```diff
@@ -712,8 +707,27 @@ func BenchmarkReadingAllBalances(b *testing.B) {
...
+func cleanupTestDb(dbs dbPair, dbName string, inMemory bool) {
+	dbs.close()
+	if !inMemory {
+		os.Remove(dbName)
+	}
+}
+
+func benchmarkReadingAllBalances(b *testing.B, inMemory bool) {
+	proto := config.Consensus[protocol.ConsensusCurrentVersion]
+	//b.N = 50000
```
tsachiherman, line 724: remove commented out line.

ledger/accountdb_test.go
```diff
@@ -730,6 +744,41 @@ func BenchmarkReadingAllBalances(b *testing.B) {
...
+	setDbLogging(b, dbs)
+	defer cleanupTestDb(dbs, fn, inMemory)
+
+	accounts := benchmarkInitBalances(b, dbs, proto)
+
+	qs, err := accountsDbInit(dbs.rdb.Handle, dbs.wdb.Handle)
+	require.NoError(b, err)
+
+	b.ResetTimer()
+	// read all the balances in the database, shuffled by map-iterator hash order
```
tsachiherman, line 767: I'm afraid that it won't be sufficiently random; both the insertion and query are using the same map, so the items would have the same order..

ledger/eval_test.go
```diff
@@ -318,3 +323,238 @@ func TestPrepareAppEvaluators(t *testing.T) {
...
+	l.AddValidatedBlock(*validatedBlock, agreement.Certificate{})
+
+	avbDone := time.Now()
+	avbTime := avbDone.Sub(blockBuildDone)
+	b.ReportMetric(float64(avbTime)/float64(b.N), "ns/AddValidatedBlock_tx")
+
+	// test speed of block validation
+	// This should be the same as the eval line in ledger.go AddBlock()
+	// This is pulled out to isolate eval() time from db ops of AddValidatedBlock()
+	_, err = eval(context.Background(), l2, validatedBlock.blk, false, nil, nil)
```
tsachiherman, line 442: you've passed false to "validate" here.. I'm pretty sure that you're not going to make any validation here.

tsachiherman (changes requested):
few small requests, overall looks good.

ledger/eval_test.go
brianolson: turned out to be a unused copy of a prior version of the benchmark function. deleted.

tsachiherman (approved):
Looks good

PR #1744, opened 2020-12-03
Add e2e test with some interesting state for Indexer tests.

winder:
## Summary

Add a new e2e test script that creates and closes things. The main purpose here is to create some interesting state for Indexer tests.

### DIFF
--- a/test/scripts/e2e_subs/create_destroy_optin_optout.sh
+++ b/test/scripts/e2e_subs/create_destroy_optin_optout.sh
@@ -1,0 +1,88 @@
+#!/bin/bash
+# This script creates some state required for Indexer testing
+# TIMEOUT=400
+
+date '+create_destroy_optin_optout start %Y%m%d_%H%M%S'
+
+set -e
+set -x
+set -o pipefail
+export SHELLOPTS
+
+WALLET=$1
+
+# Directory of this bash program
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+
+gcmd="goal -w ${WALLET}"
+
+ACCOUNT=$(${gcmd} account list | awk '{ print $3 }')
+ACCOUNTB=$(${gcmd} account new | awk '{ print $6 }')
+${gcmd} clerk send -a 1000000 -f $ACCOUNT -t $ACCOUNTB
+
+#############################################
+# Account - create / close / create / close #
+#############################################
+ACCOUNT_NEW=$(${gcmd} account new | grep "Created" | awk '{ sub("\r", "", $NF); print $NF }')
+${gcmd} clerk send -a 100000 -f $ACCOUNT -t $ACCOUNT_NEW
+${gcmd} clerk send -a 0 -f $ACCOUNT_NEW -t $ACCOUNT -c $ACCOUNT
+${gcmd} clerk send -a 100000 -f $ACCOUNT -t $ACCOUNT_NEW
+${gcmd} clerk send -a 0 -f $ACCOUNT_NEW -t $ACCOUNT -c $ACCOUNT
+
+#####################################
+# Account - create / close / create #
+#####################################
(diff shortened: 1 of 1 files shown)


test/scripts/e2e_subs/create_destroy_optin_optout.sh
```diff
@@ -0,0 +1,88 @@
...
+gcmd="goal -w ${WALLET}"
+
+ACCOUNT=$(${gcmd} account list | awk '{ print $3 }')
+ACCOUNTB=$(${gcmd} account new | awk '{ print $6 }')
+${gcmd} clerk send -a 1000000 -f $ACCOUNT -t $ACCOUNTB
+
+#############################################
+# Account - create / close / create / close #
+#############################################
+ACCOUNT_NEW=$(${gcmd} account new | grep "Created" | awk '{ sub("\r", "", $NF); print $NF }')
```
tsachiherman, line 26: is there any particular reason you're going this differently as the above ACCOUNTB ?
( I'm not particular, but would rather be consistent )

tsachiherman (approved):
looks good to me.

test/scripts/e2e_subs/create_destroy_optin_optout.sh
brianolson: I also suspect ACCOUNT_NEW would look better as ACCOUNTC

test/scripts/e2e_subs/create_destroy_optin_optout.sh
```diff
@@ -0,0 +1,88 @@
...
+${gcmd} clerk send -a 100000 -f $ACCOUNT -t $ACCOUNT_NEW
+${gcmd} clerk send -a 0 -f $ACCOUNT_NEW -t $ACCOUNT -c $ACCOUNT
+${gcmd} clerk send -a 100000 -f $ACCOUNT -t $ACCOUNT_NEW
+${gcmd} clerk send -a 0 -f $ACCOUNT_NEW -t $ACCOUNT -c $ACCOUNT
+
+#####################################
+# Account - create / close / create #
+#####################################
+ACCOUNT_NEW=$(${gcmd} account new | grep "Created" | awk '{ sub("\r", "", $NF); print $NF }')
```
brianolson, line 35: And I think this would be clearer with the same `account new` incantation as ACCOUNTB, but probably titled here ACCOUNTD just to keep clear that it's a different account than the previous ACCOUNT_NEW

winder:
@brianolson @tsachiherman fixed the inconsistent variable name

test/scripts/e2e_subs/create_destroy_optin_optout.sh
```diff
@@ -23,19 +23,19 @@ ${gcmd} clerk send -a 1000000 -f $ACCOUNT -t $ACCOUNTB
 #############################################
 # Account - create / close / create / close #
 #############################################
-ACCOUNT_NEW=$(${gcmd} account new | grep "Created" | awk '{ sub("\r", "", $NF); print $NF }')
-${gcmd} clerk send -a 100000 -f $ACCOUNT -t $ACCOUNT_NEW
-${gcmd} clerk send -a 0 -f $ACCOUNT_NEW -t $ACCOUNT -c $ACCOUNT
-${gcmd} clerk send -a 100000 -f $ACCOUNT -t $ACCOUNT_NEW
-${gcmd} clerk send -a 0 -f $ACCOUNT_NEW -t $ACCOUNT -c $ACCOUNT
+ACCOUNTC=$(${gcmd} account new | grep "Created" | awk '{ sub("\r", "", $NF); print $NF }')
```
brianolson, line 26: It's still weird that this isn't the same form as the `account new` on line 20
Why this form of `grep|awk` vs the above `awk '{ print $6 }'` ?

winder: I was fooling around with this trying to get "the last word", I guess I never landed on a consistent formula

tsachiherman (approved):

PR #1746, opened 2020-12-03
Add travis_retry to build and test commands in case of random failures.

winder:
## Summary

The build always seems to fail with intermittent errors. See if adding `travis_retry` helps.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -52,7 +52,7 @@ jobs:
           packages:
             - awscli
       script:
-        - scripts/travis/external_build.sh ./scripts/travis/build_test.sh
+        - travis_retry scripts/travis/external_build.sh ./scripts/travis/build_test.sh
     - # same stage, parallel job
       name: External ARM64 Integration Test
       os: linux
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,18 +29,18 @@ jobs:
     - stage: build_commit
       os: linux
       script:
-        - scripts/travis/build_test.sh
+        - travis_retry scripts/travis/build_test.sh
 
     - stage: build_pr
       os: linux
       name: Ubuntu AMD64 Build
       script:
-        - scripts/travis/build_test.sh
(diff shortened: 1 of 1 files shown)


tsachiherman:
not sure what I think about this one:
- we already have a retrying mechanism during the configure_dev step so that internet-based failures would get retried anyway.
- build errors should be deterministic, as there aren't "unexpected variables" there.
- tests should be reliable. If not, we should fix these rather then implementing a patch.

But I'll leave it to @onetechnical to make the call on this one.

winder:
@tsachiherman I don't disagree. But I also can't recall the last time tests on this repo succeeded for me without at least one stage needing a restart. The reality is that the Travis build check is annoying and fails so frequently that I don't bother to look at errors unless it fails a few times in a row.

Happy to hear what others think. Looks like this isn't even going to work because one of the external builds has failed this time. Do you think those can be retried with this mechanism as well?

tsachiherman:
> @tsachiherman I don't disagree. But I also can't recall the last time tests on this repo succeeded for me without at least one stage needing a restart. The reality is that the Travis build check is annoying and fails so frequently that I don't bother to look at errors unless it fails a few times in a row.
> 
> Happy to hear what others think. Looks like this isn't even going to work because one of the external builds has failed this time. Do you think those can be retried with this mechanism as well?

I think that there is no reason why it wouldn't work with the external builders, but there could be a timing issue here that need to be resolved.

Also, please don't forget that most of our builds takes quite some time. And there is also a hard time limit on travis builds that cannot be overwritten.

winder:
@tsachiherman the timeout issue might be a blocker for this change. I think the default number of retries with this command is 3. So if there was a legitimate and consistent error that occurs 45 minutes into one of our integration test builds, the 3rd retry would fail with a timeout at the end instead of the actual error.

onetechnical (approved):
I agree that it feels less than deterministic, but reality is that we do retries. I do worry philosophically, but practically think it's worth trying.

PR #1747, opened 2020-12-04
Get system DNS on Windows

algorandskiy:
## Summary

Use GetNetworkParams to get list system DNS servers on Windows.
The main issue with the implementation it relies on the fact `GetNetworkParams` allocates list of `IP_ADDR_STRING` for `DnsServerList` field directly after the input struct. In the same time it is a part of interface - `FIXED_INFO` has a fixed size across versions (no version or size field) but the API accepts size of allocated memory region for `FIXED_INFO` and overlay.

## Test Plan

Use existing tests

### DIFF
--- a/tools/network/dnssec/config_windows.go
+++ b/tools/network/dnssec/config_windows.go
@@ -1,0 +1,126 @@
+// Copyright (C) 2019-2020 Algorand, Inc.
+// This file is part of go-algorand
+//
+// go-algorand is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as
+// published by the Free Software Foundation, either version 3 of the
+// License, or (at your option) any later version.
+//
+// go-algorand is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with go-algorand.  If not, see <https://www.gnu.org/licenses/>.
+
+// +build windows
+
+package dnssec
+
+import (
+	"fmt"
+	"runtime/debug"
+	"time"
+	"unsafe"
+
+	"golang.org/x/sys/windows"
+)
+
+var (
+	dll               = windows.NewLazyDLL("iphlpapi.dll")
+	networkParamsProc = dll.NewProc("GetNetworkParams")
+)
+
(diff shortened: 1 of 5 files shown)


tools/network/dnssec/config_windows.go
```diff
@@ -0,0 +1,120 @@
...
+const ipAddrStringSizeof = 48
+
+type fixedInfoWithOverlay struct {
+	fixedInfo
+	overlay [ipAddrStringSizeof * 32]uint8 // space for max 32 IP_ADDR_STRING entries in the overlay
+}
+
+func SystemConfig() (servers []ResolverAddress, timeout time.Duration, err error) {
+	var fi fixedInfoWithOverlay
+	var ulSize uint32
```
tsachiherman, line 91: You missed here setting the ulSize to the size of the fi data structure.

tools/network/dnssec/config_windows.go
```diff
@@ -0,0 +1,120 @@
...
+		ip := make([]byte, ip_size)
+		for i := 0; i < len(p.IpAddress) && p.IpAddress[i] != 0; i++ {
+			ip[i] = p.IpAddress[i]
+		}
+		servers = append(servers, MakeResolverAddress(string(ip), "53"))
+
+		if p.Next == 0 {
+			break
+		}
+		p = (*ipAddrString)(unsafe.Pointer(p.Next))
```
pzbitskiy, line 116: The bigger problem is this conversion that triggers `go vet` to complain, and a chance to get `fi` collected by GC.
Maybe I need to give `cgo` another change and return an struct with max of 24 arrays 16 bytes each.

tools/network/dnssec/config_windows.go
pzbitskiy: Yep, thanks

algorandskiy: done

tools/network/dnssec/config_windows.go
algorandskiy: Strange, it does not complain in command line. Fixed GC.

tsachiherman (approved):
The usage of the `GetNetworkParams` isn't really done correctly - it could break if the user has configured more than 32 DNS servers. I think that it's a big improvement for now, but would be happy to make sure writing this in a more "windows-compliant" way ( as described in the [getnetworkparams](https://docs.microsoft.com/en-us/windows/win32/api/iphlpapi/nf-iphlpapi-getnetworkparams) docs )

algorandskiy:
I agree, another option is `cgo`. If I remember right Windows gives only 2 placeholders for DNS servers in GUI.

PR #1751, opened 2020-12-07
Parallelize proposal transactions signature verification

tsachiherman:
## Summary

Our existing code was verifying the transactions in parallel with the insertion to the evaluator, but that won't fully utilize the host computational abilities. To push it to the end, we need to break the transaction verification into subsets and parallelize the verification.

This is the first PR, which would handle the verification parallelization. The subsequent PR would ensure that we don't revalidate the same transaction coming from two different proposals, regardless if the transaction appear in the transaction pool or not.

## Test Plan

Added unit test to verify changed code correctness.

## Performance

Testing the performance gains between running `verify.TxnPool` vs using the new `verify.PaysetGroups`:
| Transaction Count | Before ( serialized ) | After ( parallelized ) |
|:--:|:--:|:--:|
|1 (amortized) |65,200 ns| 14,600 ns|
|5,000|326 ms|73 ms|
|10,000|656 ms|146 ms|
|50,000|3.260 s|730 ms|

Tests were conducted on a MacBook Pro 2.7 GHz Quad-Core Intel Core i7

### DIFF
--- a/data/transactions/verify/txn.go
+++ b/data/transactions/verify/txn.go
@@ -336,6 +336,7 @@ func PaysetGroups(ctx context.Context, payset [][]transactions.SignedTxn, blk bo
 		RewardsPool: blk.BlockHeader.RewardsPool,
 	}
 	var txnGroups [][]transactions.SignedTxn
+	// prepare up to 16 concurrent worksets.
 	worksets := make(chan struct{}, 16)
 	worksDoneCh := make(chan interface{}, 16)
 	zeroAddress := basics.Address{}
@@ -343,10 +344,13 @@ func PaysetGroups(ctx context.Context, payset [][]transactions.SignedTxn, blk bo
 	tasksCtx, cancelTasksCtx := context.WithCancel(ctx)
 	defer cancelTasksCtx()
 	for processing >= 0 {
-		if txnGroupIdx < len(payset) && (txnGroups == nil) {
-			txnCounter := 0
-			for i := txnGroupIdx + 1; i < len(payset); i++ {
+		if txnGroupIdx < len(payset) && (len(txnGroups) == 0) {
+			txnCounter := 0 // how many transaction we already included in the current workset.
+			for i := txnGroupIdx; i < len(payset); i++ {
 				if txnCounter+len(payset[i]) > txnPerGroupThreshold {
+					if i == txnGroupIdx {
+						i++
+					}
 					txnGroups = payset[txnGroupIdx:i]
 					txnGroupIdx = i
 					break
--- a/data/transactions/verify/txn.go
+++ b/data/transactions/verify/txn.go
@@ -138,19 +138,25 @@ func Txn(s *transactions.SignedTxn, ctx Context) error {
 	if !ok {
 		return protocol.Error(ctx.CurrProto)
 	}
-	if err := s.Txn.WellFormed(ctx.CurrSpecAddrs, proto); err != nil {
-		return err
-	}
 
-	zeroAddress := basics.Address{}
-	if s.Txn.Src() == zeroAddress {
-		return errors.New("empty address")
-	}
 	if !proto.SupportRekeying && (s.AuthAddr != basics.Address{}) {
 		return errors.New("nonempty AuthAddr but rekeying not supported")
 	}
 
-	return stxnVerifyCore(s, &ctx)
+	return txnCoreCheck(s, &ctx, proto)
+}
+
+// txnCoreCheck perform the WellFormed test as well as the signature verification test
+func txnCoreCheck(s *transactions.SignedTxn, ctx *Context, proto config.ConsensusParams) error {
+	if err := s.Txn.WellFormed(ctx.CurrSpecAddrs, proto); err != nil {
+		return err
+	}
+
+	if s.Txn.Src().IsZero() {
+		return errors.New("empty address")
+	}
+
+	return stxnVerifyCore(s, ctx)
 }
 
 type asyncVerifyContext struct {
@@ -337,7 +343,6 @@ func LogicSig(txn *transactions.SignedTxn, ctx *Context) error {
 //
 // This version of verify is performing the verification over the provided execution pool.
 func PaysetGroups(ctx context.Context, payset [][]transactions.SignedTxn, blk bookkeeping.Block, verificationPool execpool.BacklogPool) (err error) {
-	txnGroupIdx := 0
 	proto, ok := config.Consensus[blk.BlockHeader.CurrentProtocol]
 	if !ok {
 		return protocol.Error(blk.BlockHeader.CurrentProtocol)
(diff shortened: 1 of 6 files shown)


data/transactions/verify/txn.go
```diff
@@ -314,3 +314,113 @@ func LogicSig(txn *transactions.SignedTxn, ctx *Context) error {
...
+		return nil
+	}
+	spec := transactions.SpecialAddresses{
+		FeeSink:     blk.BlockHeader.FeeSink,
+		RewardsPool: blk.BlockHeader.RewardsPool,
+	}
+	var txnGroups [][]transactions.SignedTxn
+	// prepare up to 16 concurrent worksets.
+	worksets := make(chan struct{}, 16)
+	worksDoneCh := make(chan interface{}, 16)
```
algorandskiy, line 341: why 16?

data/transactions/verify/txn.go
```diff
@@ -314,3 +314,113 @@ func LogicSig(txn *transactions.SignedTxn, ctx *Context) error {
...
+								return err
+							}
+							if signTxn.Txn.Src() == zeroAddress {
+								return errors.New("empty address")
+							}
+
+							if err := stxnVerifyCore(&signTxn, &ctxs[k]); err != nil {
+								return err
+							}
+						}
```
algorandskiy, line 399: Looks the same as `verify.Txn` if `if !proto.SupportRekeying` moves into the loop.

data/transactions/verify/txn.go
```diff
@@ -314,3 +314,113 @@ func LogicSig(txn *transactions.SignedTxn, ctx *Context) error {
...
+					txnGroupIdx = i
+					break
+				}
+				if i == len(payset)-1 {
+					txnGroups = payset[txnGroupIdx:]
+					txnGroupIdx = len(payset)
+					break
+				}
+				txnCounter += len(payset[i])
+			}
```
algorandskiy, line 364: possibly needs to be a method

ledger/eval.go
```diff
@@ -995,40 +981,36 @@ func eval(ctx context.Context, l ledgerForEvaluator, blk bookkeeping.Block, vali
...
 				return StateDelta{}, err
-			default:
 			}
-			return StateDelta{}, ctx.Err()
 		default:
 		}
 
-		if validate {
-			txvalidator.txgroups <- txgroup
-		}
```
algorandskiy, line 1031: another parallelism idea: run NumCPU-1 goroutines that equieue tasks received via `txvalidator.txgroups` channel. It possibly will have much less code but the same perf/parallelism.

algorandskiy (commented):
The code (especially `verify. PaysetGroups`) is quite hard to follow, please use existing `verify.Txn` function in the pool's task.

algorandskiy:
I'd love to see detailed explanation in the commit message like "split entire list of txn groups to X blocks and post them to processing in parallel".

ledger/eval.go
```diff
@@ -1043,10 +1025,15 @@ func eval(ctx context.Context, l ledgerForEvaluator, blk bookkeeping.Block, vali
 
 	// If validating, do final block checks that depend on our new state
 	if validate {
-		close(txvalidator.txgroups)
-		err, gotErr := <-txvalidator.done
-		if gotErr && err != nil {
-			return StateDelta{}, err
+		// wait for the validation to complete.
+		select {
+		case err, open := <-txvalidator.done:
```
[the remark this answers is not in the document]
tsachiherman: no, if you'll look at the code, you'll see that it closes the channel when it's done and there is no validation error.
You could ask then, why do we need `if err != nil {`, since we already know it would be a non-nil; but I don't really think it adding that extra `if` statement would hurt much here.

ledger/eval.go
tsachiherman: The execution pool is already using the NumCPU; that's why we use it..

data/transactions/verify/txn.go
tsachiherman: yes, but placing it outside the loop is more efficient as it's being executed once per transaction group rather then once per transaction. And thinking about that, I should take it one loop upward since it's *normally* not executed.

data/transactions/verify/txn.go
tsachiherman: I'll add const to define these with some explanation.

data/transactions/verify/txn.go
```diff
@@ -314,3 +314,113 @@ func LogicSig(txn *transactions.SignedTxn, ctx *Context) error {
...
+	}
+	var txnGroups [][]transactions.SignedTxn
+	// prepare up to 16 concurrent worksets.
+	worksets := make(chan struct{}, 16)
+	worksDoneCh := make(chan interface{}, 16)
+	zeroAddress := basics.Address{}
+	processing := 0
+	tasksCtx, cancelTasksCtx := context.WithCancel(ctx)
+	defer cancelTasksCtx()
+	for processing >= 0 {
```
brianolson, line 360: It's clever that producer and consumer are both in this loop, but I think it would have been clearer split out. Probably spawn a go routine for the producer and then run the consumer loop in this function below that.

data/transactions/verify/txn.go
tsachiherman: I think it would make it quite confusing to have it as a separate function, given that it would need to be something like:
```
func getNextWorkset(payset [][]transactions.SignedTxn, txnGroupIdx *int) (txnGroups [][]transactions.SignedTxn) {
                        txnCounter := 0 // how many transaction we already included in the current workset.
			// scan starting from the current position until we filled up the workset.
			for i := *txnGroupIdx; i < len(payset); i++ {
				if txnCounter+len(payset[i]) > txnPerWorksetThreshold {
					if i == *txnGroupIdx {
						i++
					}
					txnGroups = payset[txnGroupIdx:i]
					*txnGroupIdx = i
					return
				}
				if i == len(payset)-1 {
					txnGroups = payset[*txnGroupIdx:]
					*txnGroupIdx = len(payset)
					return
				}
				txnCounter += len(payset[i])
			}
                        return nil
} 
```

ledger/eval.go
```diff
@@ -923,58 +924,43 @@ func (eval *BlockEvaluator) GenerateBlock() (*ValidatedBlock, error) {
...
+	defer close(validator.done)
+	verifyParams := verify.Params{
+		CurrSpecAddrs: transactions.SpecialAddresses{
+			FeeSink:     validator.block.BlockHeader.FeeSink,
+			RewardsPool: validator.block.BlockHeader.RewardsPool,
+		},
+		CurrProto: validator.block.BlockHeader.CurrentProtocol,
+	}
+	var unverifiedTxnGroups [][]transactions.SignedTxn
+	unverifiedTxnGroups = make([][]transactions.SignedTxn, len(validator.txgroups))
```
brianolson, line 944: this was probably supposed to be the 3 arg form make(type, len, cap) with len=0

ledger/eval.go
```diff
@@ -1043,10 +1025,15 @@ func eval(ctx context.Context, l ledgerForEvaluator, blk bookkeeping.Block, vali
 
 	// If validating, do final block checks that depend on our new state
 	if validate {
-		close(txvalidator.txgroups)
-		err, gotErr := <-txvalidator.done
-		if gotErr && err != nil {
-			return StateDelta{}, err
+		// wait for the validation to complete.
+		select {
```
brianolson, line 1029: don't need a one option select{}, just do the receive statement?

data/transactions/verify/txn.go
tsachiherman: I don't think that making the producer a separate go-routine makes sense here. The time we're spending on the producer is by order of several magnitudes smaller than the consumer. i.e. the loop is waiting in the select statement almost fo the entire time.

ledger/eval.go
tsachiherman: good catch.

ledger/eval.go
tsachiherman: hmm.. no, it's a bug:
it should also test for `ctx.Done()` since the context might be canceled. Will be fixed shortly.

data/transactions/verify/txn.go
```diff
@@ -314,3 +329,116 @@ func LogicSig(txn *transactions.SignedTxn, ctx *Context) error {
...
+		case <-tasksCtx.Done():
+			return tasksCtx.Err()
+		case worksets <- struct{}{}:
+			if len(txnGroups) > 0 {
+				err := verificationPool.EnqueueBacklog(ctx, func(arg interface{}) interface{} {
+					// check if we've canceled the request while this was in the queue.
+					if tasksCtx.Err() != nil {
+						return tasksCtx.Err()
+					}
+					txnGroups := arg.([][]transactions.SignedTxn)
```
algorandskiy, line 393: I still think `verify.Txn` should be called here inside loops. most likely the benchmark will not show any noticeable difference but the code would be clearer

data/transactions/verify/txn.go
algorandskiy: or
```
type workset struct {
  payset [][]transactions.SignedTxn
  idx int
}

func (w *workset) next() (txnGroups [][]transactions.SignedTxn) {
...
}

func (w *workset) completed() bool {
  return w.idx >= len(w.payset)
}
```

data/transactions/verify/txn.go
algorandskiy: I suggest to measure. Benefits: better code, no duplicate code.

data/transactions/verify/txn.go
tsachiherman: that looks much better; I'll add that.

data/transactions/verify/txn.go
```diff
@@ -314,3 +335,124 @@ func LogicSig(txn *transactions.SignedTxn, ctx *Context) error {
...
+			return tasksCtx.Err()
+		case worksets <- struct{}{}:
+			if len(nextWorkset) > 0 {
+				err := verificationPool.EnqueueBacklog(ctx, func(arg interface{}) interface{} {
+					// check if we've canceled the request while this was in the queue.
+					if tasksCtx.Err() != nil {
+						return tasksCtx.Err()
+					}
+					txnGroups := arg.([][]transactions.SignedTxn)
+					if !proto.SupportRekeying {
```
algorandskiy, line 379: Current
```
BenchmarkPaysetGroups-8   	   78248	     16482 ns/op
BenchmarkPaysetGroups-8   	   75268	     16444 ns/op

BenchmarkTxnPool-8   	   18794	     63580 ns/op
BenchmarkTxnPool-8   	   18135	     64093 ns/op
```

With vanila `Txn(...)` call and the only one two loops instead of 4:
```
					txnGroups := arg.([][]transactions.SignedTxn)
					for _, signTxnsGrp := range txnGroups {
						ctxs := PrepareContexts(signTxnsGrp, blk.BlockHeader)
						for k, signTxn := range signTxnsGrp {
							if err := Txn(&signTxn, ctxs[k]); err != nil {
								return err
							}
						}
					}
```
```
BenchmarkPaysetGroups-8   	   80364	     15178 ns/op
BenchmarkPaysetGroups-8   	   82190	     14846 ns/op

BenchmarkTxnPool-8   	   18781	     66758 ns/op
BenchmarkTxnPool-8   	   18808	     63886 ns/op
```

algorandskiy: I guess just calling `Txn` is desirable since it is not slower and gives twice less code and no code duplication.

brianolson (dismissed):
Makes my benchmarks in #1743 faster. Seems good.

algorandskiy (approved):

PR #1754, opened 2020-12-07
Fix bug revealed by TestFullCatchpointWriter

tsachiherman:
## Summary

When the `CatchpointCatchupAccessorImpl.processStagingBalances` attempts to write the staged changes to disk, it's doing that in parallel : writing the accounts data, the assets and the staging hashes.

On test systems which use in-memory database, we have a shared-lock that prevents more than a single writer. This creates writing conflict which doesn't exists in disk writer.

To work around that limitation, when using an in-memory database, we'll be waiting for each of these writers to complete their work before moving to the next one.

## Test Plan

The TestFullCatchpointWriter test covers this issue.


algorandskiy (approved):
Nice

PR #1755, opened 2020-12-08
Fix download location when signing build assets

btoll:
## Summary

Also, created `running_in_docker.sh` script to determine when a process
is running inside a docker container. This will allow the process to run
correctly when run as a mule task or as a shell script.

## Test Plan

```
NETWORK=mainnet S3_SOURCE=algorand-internal/channel VERSION=2.3.0 ./scripts/release/mule/sign/sign.sh
NETWORK=mainnet S3_SOURCE=algorand-internal/channel VERSION=2.3.0 mule -f package-sign.yaml package-sign
```


scripts/release/mule/sign/sign.sh
```diff
@@ -22,14 +22,16 @@ OS_TYPE=$(./scripts/release/mule/common/ostype.sh)
...
-find /root/.gnupg -type f -exec chmod 600 {} \;
-
+if ./scripts/release/mule/common/running_in_docker.sh
+then
+    # It seems that copying/mounting the gpg dir from another machine can result in insecure
+    # access privileges, so set the correct permissions to avoid the following warning:
+    #
+    #   gpg: WARNING: unsafe permissions on homedir '/root/.gnupg'
+    #
+    find /root/.gnupg -type d -exec chmod 700 {} \;
```
bricerisingalgorand, line 32: Maybe we should use an env var to set this path, then we could just always package that in our docker images. This may not always work since not every container runs as root. As we develop our automation we'll want to move away from root containers since there are some vulnerabilities with them

onetechnical (dismissed):

scripts/release/mule/sign/sign.sh
btoll: I'll change this by adding another env var to the package-sign.yaml file.  It will be optional and will default to /root if not set.

I think optional is better than forcing that to be defined, since in the overwhelming number of use cases it will be /root.  You bring up a good point though, and once we move away from root containers this can be revisited (and hopefully this will since have been rewritten in python).

scripts/release/mule/sign/sign.sh
```diff
@@ -21,15 +21,23 @@ SIGNING_KEY_ADDR=dev@algorand.com
...
-#
-#   gpg: WARNING: unsafe permissions on homedir '/root/.gnupg'
+# Note that we don't want to use $GNUPGHOME here because that is a documented env var for the gnupg
+# project and if it's set in the environment mule will automatically pick it up, which could have
+# unintended consequences and be hard to debug.
 #
-find /root/.gnupg -type d -exec chmod 700 {} \;
-find /root/.gnupg -type f -exec chmod 600 {} \;
+# By naming it something other than $GNUPGHOME, it's essentially acting as an opt-in.
+GPG_DIR=${GPG_DIR:-/root/.gnupg}
```
onetechnical, line 29: Good idea, especially if we eventually stop running containers as root, as we should. It does mean we never run this when running scripts _locally_, but that's probably fine.

onetechnical (approved):

egieseke (approved):
Looks good.

PR #1756, opened 2020-12-08
Propagate block proposals before transaction validation

nicholasguoalgorand:
## Summary

Propagate block proposals before we validate it (still after we check it matches proposal votes). This change reduces the validation latency when propagating block proposals across the network.

## Test Plan

Mainnet model performance tests went from timing out > 17s to taking 10-11 seconds per round.
Reran unit tests.
Tested for backwards compatibility by running node with nodes on master branch and this branch


agreement/player.go
```diff
@@ -568,8 +572,12 @@ func (p *player) handleMessageEvent(r routerHandle, e messageEvent) (actions []a
 		}
 		up := e.Input.UnauthenticatedProposal
 
-		a := relayAction(e, protocol.ProposalPayloadTag, compoundMessage{Proposal: up, Vote: uv})
-		actions = append(actions, a)
+		// relay as the proposer
+		if e.TaskIndex == 1 {
```
tsachiherman, line 576: I'm sure there is a more elegant way to determine if this proposal was locally generated or not. One way to do so is to use the e.Input.MessageHandle and see if it's zero; messages that came from the "outside" would have this as a non-zero value. @derbear might have other suggestions.

tsachiherman (approved):

PR #1757, opened 2020-12-08
Create a unified transaction verification cache

tsachiherman:
## Summary

The existing transaction cache was always around the transaction entries that we had in our transaction pool. That has been working well for scenarios where the transaction pool is not congested. However, when we get to situations where the transaction pool is congested, it creates issues in the following two scenarios:
1. A transaction is being received while the transaction pool is full. After we verify it's signature, we find that we won't be able to insert it into the transaction pool, and drop it. A subsequent block that we attempt to verify could include this transaction, and we will need to re-validate it's signature again.
2. A node receive a proposal for verification. After verifying the proposal, the node receive a second proposal with a lower hash value ( and with a similar set of transactions ). At that point, the node would attempt to re-verify all the repeated transactions ( assuming they aren't present in the transaction pool ).

To address both issues, I've extracted the verified transaction cache out of the transaction pool into a separate object that is being held by the ledger. This object is always being used when verifying a transaction, and any verified transaction is being "set" in that object.

## Test Plan

Unit tests were added and updated.

## Performance Testing

The changes were tested using scenario1 and scenario2 networks; no regression was noted.


data/transactions/verify/txn_test.go
```diff
@@ -61,11 +70,12 @@ func generateTestObjects(numTxs, numAccs int, blockRound basics.Round) ([]transa
 		addresses[i] = addr
 	}
 	var iss, exp int
+	u := uint64(0)
 	for i := 0; i < numTxs; i++ {
 		s := rand.Intn(numAccs)
 		r := rand.Intn(numAccs)
 		a := rand.Intn(1000)
-		f := config.Consensus[protocol.ConsensusCurrentVersion].MinTxnFee + uint64(rand.Intn(10))
+		f := config.Consensus[protocol.ConsensusCurrentVersion].MinTxnFee + uint64(rand.Intn(10)) + u
```
tsachiherman, line 78: this was done to ensure that we don't end up with identical transactions when generating large number of txns.

ledger/eval.go
```diff
@@ -966,9 +956,9 @@ func (validator *evalTxValidator) run() {
 // used by Ledger.Validate() Ledger.AddBlock() Ledger.trackerEvalVerified()(accountUpdates.loadFromDisk())
 //
 // Validate: eval(ctx, blk, true, txcache, executionPool)
-// AddBlock: eval(context.Background(), blk, false, nil, nil)
-// tracker:  eval(context.Background(), blk, false, nil, nil)
-func eval(ctx context.Context, l ledgerForEvaluator, blk bookkeeping.Block, validate bool, txcache VerifiedTxnCache, executionPool execpool.BacklogPool) (StateDelta, error) {
+// AddBlock: eval(context.Background(), blk, false, txcache, nil)
```
algonautshant, line 959: "l ledgerForEvaluator" is missing in all three comment statements.

data/transactions/verify/txn.go
```diff
@@ -362,19 +317,21 @@ func PaysetGroups(ctx context.Context, payset [][]transactions.SignedTxn, blk bo
...
 					txnGroups := arg.([][]transactions.SignedTxn)
-					for _, signTxnsGrp := range txnGroups {
-						ctxs := PrepareContexts(signTxnsGrp, blk.BlockHeader)
-						for k, signTxn := range signTxnsGrp {
-							if err := Txn(&signTxn, ctxs[k]); err != nil {
-								return err
-							}
+					groupCtxs := make([]*GroupContext, len(txnGroups))
+					for i, signTxnsGrp := range txnGroups {
+						groupCtxs[i], grpErr = TxnGroup(signTxnsGrp, blkHeader, nil)
```
algorandskiy, line 328: why is the cache param nil here? `AddPayset` is used only here, so... maybe let `TxnGroup` add a group?
I do not think few additional locks make a difference there

data/transactions/verify/verifiedTxnCache.go
```diff
@@ -0,0 +1,292 @@
...
+	// in the cache, the new entry overrides the old one.
+	Add(txgroup []transactions.SignedTxn, groupCtx *GroupContext)
+	// AddPayset works in a similar way to Add, but is intended for adding an array of transaction groups, along with their corresponding contexts.
+	AddPayset(txgroup [][]transactions.SignedTxn, groupCtxs []*GroupContext) error
+	// GetUnverifiedTranscationGroups compares the provided payset against the currently cached transactions and figure which transaction groups aren't fully cached.
+	GetUnverifiedTranscationGroups(payset [][]transactions.SignedTxn, CurrSpecAddrs transactions.SpecialAddresses, CurrProto protocol.ConsensusVersion) [][]transactions.SignedTxn
+	// UpdatePinned replaces the pinned entries with the one provided in the pinnedTxns map. This is typically expected to be a subset of the
+	// already-pinned transactions. If a transaction is not currently pinned, and it's can't be found in the cache, a errMissingPinnedEntry error would be generated.
+	UpdatePinned(pinnedTxns map[transactions.Txid]transactions.SignedTxn) error
+	// Pin pins the given transaction group.
```
algorandskiy, line 66: It sounds like "a baker bakes"

data/transactions/verify/verifiedTxnCache.go
```diff
@@ -0,0 +1,292 @@
...
+	Pin(txgroup []transactions.SignedTxn) error
+}
+
+// VerifiedTransactionCacheImpl provides an implementation of the VerifiedTransactionCache interface
+type verifiedTransactionCacheImpl struct {
```
algorandskiy, line 71: I used to Impl suffixes after C++ but in Go `verifiedTransactionCache` looks as a proper naming

algorandskiy (commented):
Som initial minor remarks. I need to take another look later.

data/transactions/verify/txn.go
tsachiherman: many transaction groups will be of size 1. I think that we shouldn't take the lock if we don't have to..
after all, taking the lock takes 3000-5000 ns; multiply this by 10000 and you'll end up with some notable delay.

data/transactions/verify/verifiedTxnCache.go
```diff
@@ -0,0 +1,292 @@
...
+	}
+	return
+}
+
+// add is the internal implementation of Add/AddPayset which adds a transaction group to the buffer.
+func (v *verifiedTransactionCache) add(txgroup []transactions.SignedTxn, groupCtx *GroupContext) {
+	if len(v.buckets[v.base])+len(txgroup) > entriesPerBucket {
+		// move to the next bucket while deleting the content of the next bucket.
+		v.base = (v.base + 1) % len(v.buckets)
+		v.buckets[v.base] = make(map[transactions.Txid]*GroupContext, entriesPerBucket)
```
algorandskiy, line 248: probably it is better pre-allocate to max(entriesPerBucket, len(txgroup))

data/transactions/verify/verifiedTxnCache.go
```diff
@@ -0,0 +1,292 @@
...
+	}
+	v.pinned = pinned
+	return err
+}
+
+// Pin sets a given transaction group as pinned items, after they have already been verified.
+func (v *verifiedTransactionCache) Pin(txgroup []transactions.SignedTxn) (err error) {
+	v.bucketsLock.Lock()
+	defer v.bucketsLock.Unlock()
+	transcationMissing := false
```
algorandskiy, line 205: transcationMissing -> transactionMissing

data/transactions/verify/verifiedTxnCache.go
```diff
@@ -0,0 +1,292 @@
...
+				// move it to the pinned items :
+				v.pinned[txID] = ctx
+				delete(v.buckets[bucketIdx], txID)
+				found = true
+				baseBucket = bucketIdx
+				break
+			}
+		}
+		if !found {
+			transcationMissing = true
```
algorandskiy, line 234: break? since we going to error anyway

data/transactions/verify/verifiedTxnCache.go
```diff
@@ -0,0 +1,292 @@
...
+			// it's already pinned; keep going.
+			continue
+		}
+
+		// entry isn't in pinned; maybe we have it in one of the buckets ?
+		found := false
+		// we use the (base + W) % W trick here so we can go backward and wrap around the zero.
+		for offsetBucketIdx := baseBucket + len(v.buckets); offsetBucketIdx > baseBucket; offsetBucketIdx-- {
+			bucketIdx := offsetBucketIdx % len(v.buckets)
+			if ctx, has := v.buckets[bucketIdx][txID]; has {
```
algorandskiy, line 224: nit: we might stop earlier if we track how many buckets are in use. Maybe not a big deal, will only help on non-full cache.

data/transactions/verify/verifiedTxnCache.go
```diff
@@ -0,0 +1,292 @@
...
+// UpdatePinned replaces the pinned entries with the one provided in the pinnedTxns map. This is typically expected to be a subset of the
+// already-pinned transactions. If a transaction is not currently pinned, and it's can't be found in the cache, a errMissingPinnedEntry error would be generated.
+func (v *verifiedTransactionCache) UpdatePinned(pinnedTxns map[transactions.Txid]transactions.SignedTxn) (err error) {
+	v.bucketsLock.Lock()
+	defer v.bucketsLock.Unlock()
+	pinned := make(map[transactions.Txid]*GroupContext, len(pinnedTxns))
+	for txID := range pinnedTxns {
+		if groupEntry, has := v.pinned[txID]; has {
+			pinned[txID] = groupEntry
+			continue
```
algorandskiy, line 178: what's the point of replacing an entry with itself?

data/transactions/verify/verifiedTxnCache.go
```diff
@@ -0,0 +1,292 @@
...
+}
+
+// verifiedTransactionCache provides an implementation of the VerifiedTransactionCache interface
+type verifiedTransactionCache struct {
+	// bucketsLock is the lock for syncornizing the access to the cache
+	bucketsLock deadlock.Mutex
+	// buckets is the circular cache buckets buffer
+	buckets []map[transactions.Txid]*GroupContext
+	// pinned is the pinned transactions entries map.
+	pinned map[transactions.Txid]*GroupContext
```
algorandskiy, line 77: seems like `pinned` is used as a kind of a cache within the cache for faster lookup in `GetUnverifiedTranscationGroups`. Could please put a comment here and maybe in the interface about pinning?

algorandskiy (commented):
as gerrit says "+1, Looks good to me, but someone else must approve"

data/pools/transactionPool.go
```diff
@@ -467,73 +453,14 @@ func (pool *TransactionPool) Lookup(txid transactions.Txid) (tx transactions.Sig
...
 	return pool.statusCache.check(txid)
 }
 
-// Verified returns whether a given SignedTxn is already in the
-// pool, and, since only verified transactions should be added
-// to the pool, whether that transaction is verified (i.e., Verify
-// returned success).  This is used as an optimization to avoid
-// re-checking signatures on transactions that we have already
-// verified.
-func (pool *TransactionPool) Verified(txn transactions.SignedTxn, params verify.Params) bool {
```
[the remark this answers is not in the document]
tsachiherman: it used to be called before I implemented the UnverifiedTxnGroups, which does the same, but for an array.

data/transactions/verify/verifiedTxnCache.go
tsachiherman: not quite. It's not about faster lookup; it's about non-perishable by incoming transactions. I'll add some comments to explain that.

data/transactions/verify/verifiedTxnCache.go
tsachiherman: don't get confused between the current `v.pinned` and the newly built `pinned`.

data/transactions/verify/verifiedTxnCache.go
tsachiherman: I think that after the first cycle, all the bucket will be in use ( although they might contain "old" entries ).
My intent here was to try and avoid deleting the old maps entries.

data/transactions/verify/verifiedTxnCache.go
tsachiherman: ahh yes.. Failing to pin a transaction within group ( or part of it ) isn't a good thing, but it shouldn't prevent us from pinning the rest of the entries. ( i.e. in the worst case scenario, we will need to verify the signature again for that particular transaction ).
The caller should log this, but there is nothing that really can be done at that point. ( and it's not really harmful either )

data/transactions/verify/verifiedTxnCache.go
tsachiherman: the number of transaction in a group is 16. the entriesPerBucket is in the order of several thousands. 
when allocating a new bucket, we want to have large buckets, and have each bucket contain all the transactions of a single txn group.

data/transactions/verify/verifiedTxnCache.go
```diff
@@ -0,0 +1,295 @@
...
+			if !entryGroup.Equal(groupCtx) {
+				break
+			}
+
+			if entryGroup.signedGroupTxns[txnIdx].Sig != txn.Sig || (!entryGroup.signedGroupTxns[txnIdx].Msig.Equal(txn.Msig)) || (!entryGroup.signedGroupTxns[txnIdx].Lsig.Equal(&txn.Lsig)) || (entryGroup.signedGroupTxns[txnIdx].AuthAddr != txn.AuthAddr) {
+				break
+			}
+			verifiedTxn++
+		}
+		if verifiedTxn != len(signedTxnGroup) || verifiedTxn == 0 {
```
algonautshant, line 165: why also append when verifiedTxn == 0?

data/transactions/verify/verifiedTxnCache.go
```diff
@@ -0,0 +1,295 @@
...
+				break
+			}
+		}
+		if !found {
+			err = errMissingPinnedEntry
+		}
+
+	}
+	v.pinned = pinned
+	return err
```
algonautshant, line 201: What does it mean if an err is returned? The old pinned cache is discarded, and a new one is partial.
The call is returning an error, but it is not meant to discard the results of the call.

data/transactions/verify/verifiedTxnCache.go
```diff
@@ -0,0 +1,295 @@
...
+		for offsetBucketIdx := v.base + len(v.buckets); offsetBucketIdx > v.base; offsetBucketIdx-- {
+			bucketIdx := offsetBucketIdx % len(v.buckets)
+			if groupEntry, has := v.buckets[bucketIdx][txID]; has {
+				pinned[txID] = groupEntry
+				found = true
+				break
+			}
+		}
+		if !found {
+			err = errMissingPinnedEntry
```
algonautshant, line 196: Why is this an error? It is not an error when a txn is cycled out of the circular buffer, it is what we intend right?

data/transactions/verify/verifiedTxnCache.go
```diff
@@ -0,0 +1,295 @@
...
+	return err
+}
+
+// Pin sets a given transaction group as pinned items, after they have already been verified.
+func (v *verifiedTransactionCache) Pin(txgroup []transactions.SignedTxn) (err error) {
+	v.bucketsLock.Lock()
+	defer v.bucketsLock.Unlock()
+	transactionMissing := false
+	if len(v.pinned)+len(txgroup) > maxPinnedEntries {
+		// reaching this number likely means that we have an issue not removing entries from the pinned map.
```
algonautshant, line 210: What if the same txns are in txgroup and in v.pinned? Can this happen?

algonautshant (approved):
Looks great. 
I have some clarification questions.

data/transactions/verify/txn.go
```diff
@@ -52,126 +52,89 @@ const txnPerWorksetThreshold = 32
...
-	cx := asyncVerifyContext{s: s, outCh: outCh, ctx: &ctx}
-	verificationPool.EnqueueBacklog(context.Background(), stxnAsyncVerify, &cx, nil)
-	if err, hasErr := <-outCh; hasErr {
-		return err
-	}
-	return nil
+// Equal compares two group contexts to see if they would represent the same verification context for a given transaction.
+func (g *GroupContext) Equal(other *GroupContext) bool {
+	return g.specAddrs == other.specAddrs &&
+		g.consensusVersion == other.consensusVersion &&
```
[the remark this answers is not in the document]
tsachiherman: Exactly. For any given node, this always true.

data/transactions/verify/verifiedTxnCache.go
```diff
@@ -0,0 +1,295 @@
...
+	"errors"
+
+	"github.com/algorand/go-deadlock"
+
+	"github.com/algorand/go-algorand/data/transactions"
+	"github.com/algorand/go-algorand/data/transactions/logic"
+	"github.com/algorand/go-algorand/protocol"
+)
+
+const entriesPerBucket = 8179 // the default bucket size; a prime number could promote a lower hash collisions in case the hash function isn't perfect.
```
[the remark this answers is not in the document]
tsachiherman: I believe it will consume more memory, but it will be bound regardless.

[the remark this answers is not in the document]
tsachiherman: Keep in mind that this number only adjust the entries per bucket. The number of buckets is derived from yhe config value.

data/transactions/verify/verifiedTxnCache.go
```diff
@@ -0,0 +1,295 @@
...
+	for txID := range pinnedTxns {
+		if groupEntry, has := v.pinned[txID]; has {
+			pinned[txID] = groupEntry
+			continue
+		}
+
+		// entry isn't in pinned; maybe we have it in one of the buckets ?
+		found := false
+		// we use the (base + W) % W trick here so we can go backward and wrap around the zero.
+		for offsetBucketIdx := v.base + len(v.buckets); offsetBucketIdx > v.base; offsetBucketIdx-- {
```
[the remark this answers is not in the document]
tsachiherman: It depends on the usage. Proposal validation would cause full buckets, transaction gossiping that goes into the txpool would first go into the buckets and then moved into the pinned map.

data/transactions/verify/verifiedTxnCache.go
tsachiherman: It's not a valid txn, and therefore would return in the list. I think it would never happen, but it's good to be on the safe side.

data/transactions/verify/verifiedTxnCache.go
tsachiherman: Yes, the intent is to cycle out old messages. However, if we fail to pin a txn, it means that our cyclic buffer is too small. The error here is just to inform the caller that the cyclic buffer isn't sized optimally. It shouldnt cause any issues other the re-checking the signature later on.

data/transactions/verify/verifiedTxnCache.go
```diff
@@ -0,0 +1,295 @@
...
+		for txnIdx := 0; txnIdx < len(signedTxnGroup); txnIdx++ {
+			txn := &signedTxnGroup[txnIdx]
+			id := txn.Txn.ID()
+			// check pinned first
+			entryGroup := v.pinned[id]
+			// if not found in the pinned map, try to find in the verified buckets:
+			if entryGroup == nil {
+				// try to look in the previously verified buckets.
+				// we use the (base + W) % W trick here so we can go backward and wrap around the zero.
+				for offsetBucketIdx := baseBucket + len(v.buckets); offsetBucketIdx > baseBucket; offsetBucketIdx-- {
```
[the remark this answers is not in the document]
tsachiherman: We do.. that's why we have the baseBucket local variable.

data/transactions/verify/verifiedTxnCache.go
```diff
@@ -0,0 +1,295 @@
...
+					bucketIdx := offsetBucketIdx % len(v.buckets)
+					if params, has := v.buckets[bucketIdx][id]; has {
+						entryGroup = params
+						baseBucket = bucketIdx
+						break
```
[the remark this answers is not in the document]
tsachiherman: We already have a dup checking in the txhandler, and we're filtering out the unverified txn before inserting the txgroup to the cache.

data/transactions/verify/verifiedTxnCache.go
```diff
@@ -0,0 +1,295 @@
...
+					}
+				}
+			}
+
+			if entryGroup == nil {
+				break
+			}
+
+			if !entryGroup.Equal(groupCtx) {
```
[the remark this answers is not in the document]
tsachiherman: No, this verifies that the parameters under which we originally tested the validity of this transaction are still the same as of the current block.

data/transactions/verify/verifiedTxnCache.go
tsachiherman: There is no much we can do, and it won't cause any real harm. But it would be an indication that increasing the cyclic buffer size could be a good idea.

data/transactions/verify/verifiedTxnCache.go
tsachiherman: This is a failsafe check. I dont expect we would ever reach this.. if we get anywhere near 500000 pinned txn, we have bigger issues to take care of.. remember that every new round flushes the pinned map to the size of the txpool, which is much much smaller.

data/transactions/verify/verifiedTxnCache.go
```diff
@@ -0,0 +1,295 @@
...
+	}
+	if transactionMissing {
+		err = errMissingPinnedEntry
+	}
+	return
+}
+
+// add is the internal implementation of Add/AddPayset which adds a transaction group to the buffer.
+func (v *verifiedTransactionCache) add(txgroup []transactions.SignedTxn, groupCtx *GroupContext) {
+	if len(v.buckets[v.base])+len(txgroup) > entriesPerBucket {
```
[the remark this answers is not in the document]
tsachiherman: The group is small <17, and the bucket is large 7-8k. Leaving few empty entries won't change much, and would help ensuring all the group txn are in the same bucket.

data/transactions/verify/verifiedTxnCache.go
```diff
@@ -0,0 +1,295 @@
...
+		// move to the next bucket while deleting the content of the next bucket.
+		v.base = (v.base + 1) % len(v.buckets)
+		v.buckets[v.base] = make(map[transactions.Txid]*GroupContext, entriesPerBucket)
+	}
+	currentBucket := v.buckets[v.base]
+	for _, txn := range txgroup {
+		currentBucket[txn.ID()] = groupCtx
```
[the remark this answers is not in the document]
tsachiherman: See my answer above..

algorandskiy (approved):

PR #1761, opened 2020-12-09
Code style fixes DNSSEC client

algorandskiy:
## Summary

* Fix comments for config
* Use struct embedding instead of aggregation
  to get rid of useless explicit forwarding

## Test Plan

Use existing tests


tsachiherman (approved):
looks good.

Issue #1738, opened 2020-12-02
Scalability issues in the implementation of TEAL data types


aybehrouz:
## Summary

The code which implements Teal data types needs to be refactored to prevent storing unnecessary nil values and to make adding new data types easier.

## Scope/Requirements

In the source code we have a struct named `TealValue` which seems to be created for representing a teal data element. Currently a teal data element can be a byte slice or uint64. `TealValue` struct has this declaration:
```go
// TealValue contains type information and a value, representing a value in a
// TEAL program
type TealValue struct {
	_struct struct{} `codec:",omitempty,omitemptyarray"`

	Type  TealType `codec:"tt"`
	Bytes string   `codec:"tb"`
	Uint  uint64   `codec:"ui"`
}
```

Similarly, somewhere else we have a `StackValue` struct with similar declaration: (I don't know why this struct was created. couldn't we just define a stack of `TealValue` instead of creating a new `StackValue` type?)
```go
// stackValue is the type for the operand stack.
// Each stackValue is either a valid []byte value or a uint64 value.
// If (.Bytes != nil) the stackValue is a []byte value, otherwise uint64 value.
type stackValue struct {
	Uint  uint64
	Bytes []byte
}
```
All these structs contain a field for every possible data type inside them and only one of these fields will be non-nil. In my opinion this is not the right way for representing a teal data element and this representation is not scalable. in the future we may need more Teal data types like `[]int` or `float`. With the current code, adding them will not be so easy and needs a lot of changes in different parts of the code. **Also with more data types every struct will have more nil fields which wastes memory and when the struct is serialized to the ledger will waste the storage space too.**

For a situation like this In OO programing we usually define a  `TealValue` (or `TealData`) super-class and a specific teal data type, for example `TealInt`, would be a subclass of it. In go language this can be implemented using interfaces, something like this:
```go
//TealValue is our super class
type TealValue interface {
	SayHiLikeATeal()
}

//TealInt is a subclass of TealValue representing an integer data element
type TealInt struct {
	value int
}

func (ti TealInt) Value() int {
	return ti.value
}

func (ti TealInt) SayHiLikeATeal() {
	fmt.Print("Hi! I'm a TealInt!")
}

//TealBytes is a subclass of TealValue representing a byte slice data element
type TealBytes struct {
	value []byte
}

func (tb TealBytes) SayHiLikeATeal() {
	fmt.Print("Hi! I'm a TealBytes!")
}

func (tb TealBytes) Value() []byte {
[...]

PR #1763, opened 2020-12-10
Improve TEAL's memory architecture

aybehrouz:
## Summary

In the current version of Algorand, the state of a stateful contract is represented by key-value pairs. As the following example from official developer docs shows, these pairs are usually used for storing script variables. variable names are the keys for variable values and key-value pairs are stored in the state storage:
```
int 0
byte "MyLocalKey"
int 50
app_local_put
```
This approach has several short comings:
 
- A map's implementation always keeps a copy of its keys. When we use a map for storing variables, besides the value of variables, their names are stored in the state storage too.
- In this method a variable name is essentially a part of the byte-code. In the above example `"MyLocalKey"` acts like a constant and hence is a part of the script's byte-code. When we store the byte code of a smart contract we are storing a copy of its variable names too. 

Programmers are taught to use long readable names in their code. This is how we code. We don't care about length of our variables names. Algorand is keeping **two copies** of these names in its blockchain wasting a lot of storage.

My proposed solution for this issue is simple. We could use an array-style data structure instead of a map. Variables could be stored and loaded using indices and the mapping between names of variables and their addresses (indices) could be done in the compiler/assembler using symbol tables. There will be no need for storing variable names.
 
This memory model is simply like an array which can hold any TEAL datatype. Every contract or account will keep an array-style memory segment in its state which has a fixed size and smart contract's schema defines its size: 
```
Static Memory:	                   Heap:
|0|<empty>
|1|    -------------------> teal.UInt
|2|<empty>
|3|    -------------------> teal.ConstByteArray
|4|<empty>
```
State manipulation opcodes could be changed to include a direct address field for manipulating a memory location. For example, `app_global_get` would be changed to something like this:
```
app_global_get
    Opcode: 0x64 {uint8 position in global static memory space to load from}
    Pops: None
    Pushes: any
    LogicSigVersion >= 2
    Mode: Application
```
With these changes, actually the state storage space looks like the scratch space so state manipulation opcodes will be similar to `load` and `store` opcodes which manipulate the scratch space.

This will also make TEAL scripts much more readable because variables could be handled by compiler and the code will look more like a conventional program. For example, a TEAL script could look like this after this change:
```
define var myVariable
int 50
app_local_put myVariable
```

I've developed a package which implements an array-style memory, as described above. This memory can save and restore a state snapshot, so when a smart contract fails, its state can be restored. Also, I've provided a cleaner implementation for TEAL data types which is a little bit more efficient than the current implementation. I've addressed this issue in https://github.com/algorand/go-algorand/issues/1738 before. This package fixes [that issue](https://github.com/algorand/go-algorand/issues/1738).

Of course, there is still a lot of work to be done. I hope this pull request could be a starting point for the community to discuss this issue and start working on it.

## Test Plan

For testing, I've created a unit test which tests several scenarios.


brianolson:
Thank you for your interest but given our need to continue to support existing TEAL apps this sort of change is unlikely. Array-style and other schemes were proposed during design but map/dict style won as most general and good-enough performance. Future expansions of TEAL are still likely to make it more capable or more able to efficiently be a solution for future applications.

tsachiherman:
As @brianolson mentioned above, once we have a feature released, we have to support it forever, even if just for the sake of validating old blocks. Improving the node's memory utilization is always an important task, but I'm not really sure that the TEAL keys is where I would start optimizing.
If you'd like to try and optimize the memory utilization, why don't you let the node to perform catchup and try to pprof its memory utilization ? I'm pretty sure there are much bigger fish there.
Alternatively, you could use the pingpong utility to generate lots of TEAL transactions and pprof that process as well. I wouldn't bet that the TEAL code would be the one on the top.

aybehrouz:
>  once we have a feature released, we have to support it forever

Indeed! That's why you should be really careful with your design decisions when u are designing blockchain protocols. Smart contracts can not be updated and if a protocol upgrade accidentally changes some smart contracts behavior its possible that a lot of money gets trapped inside contract accounts (this actually happened for Ethereum) or other things like that...

Anyways, **by using version tags for TEAL scripts, newer versions of TEAL need not to be backward compatible. Also, its possible to disallow installation of new smart contracts that use an older version of TEAL.** for example, if we optimize the memory model in TEAL3 we can disallow installation of TEAL2 contracts so the storage problems will be limited to existing TEAL2 contracts.

Im concerned about this particular issue because this issue arises from a bad protocol design. As you said, protocol upgrades regarding smart contracts, is not easy. As we're talking some new smart contracts are being added to Algorand's blockchain and they will waste more storage by storing their variable names as keys. these keys will stay there forever. No one can remove them later. That's why we should be quick and solving this issue is a high priority.
_(there might be some subtle methods for optimizing existing byte-codes. I do not consider them here... also note that those methods are always risky)_

> If you'd like to try and optimize the memory utilization, why don't you let the node to perform catchup and try to pprof its memory utilization ? I'm pretty sure there are much bigger fish there.

I don't think this is a good idea because currently Algorand is not under a realistic work load. No one is really using Algorand's blockchain. There are not so many smart contracts like DEXs, lending pools, etc so the information you'll get from pprof will not be so useful and in the future when the load pattern of Algorand changes we may see different behavior .

At the end, I really hope that I'm wrong about this one... Consider this scenario:
I have a SC that people can deposit Algo, USDT or USDC to it. I need to keep the balance of every user so I make these keys:
```
"user_balance_Algo"
"user_balance_USDT" 
"user_balance_USDC" 
```
my DAPP becomes successful and 150 million users deposit some money to it. How many copies of strings: `"user_balance_Algo ","user_balance_USDT"` and `"user_balance_USDC"` will be stored on the Algorand's state storage?

tsachiherman:
@aybehrouz, your arguments here are correct. If you have a DAPP that has 150M user accounts, and you store data in their local data, it would have lot's of duplicate information ( which is inarguably undesirable ).
However, as the creator of this successful application, you have some motivation to optimize your partake here and try to make your application as compact as possible ( i.e. to save money for both yourself and others ).
In order to do so, you could use the following keys:
"user_balance_Algo" -> "a"
"user_balance_USDT" -> "t"
"user_balance_USDC"  -> "c"

I know that it doesn't seems to be as attractive as using arrays, but it could reduce the problem from "a change that has to happen" into "a change that could happen, and it would be great if we've had it earlier". I haven't tried to analyze what amount of changes this would require, but I would imagine it would be quite an extensive effort.

aybehrouz:
@tsachiherman 
When I opened this PR I also provided the implementation of a memory package which could've been a starting point for changing the current memory style of TEAL. I didn't try to integrate it to the main code because after creating the PR I didn't get any positive feedback from the community or the foundation, so I supposed, even after a lot of work if I integrate it to the main code, still it may not be accepted.

On the other hand, I think Algorand needs to replace its current persistence layer with a zero knowledge database (ZK-EDB). I know that already some prototypes like Aardvark or Vault exist for this. maybe TEAL memory style and persistent layer could be changed together and some other improvements like what I mentioned in this closed [issue](https://github.com/algorand/go-algorand/issues/1773), could be done at that time too.

Anyway, still I think this package could be useful. It has many useful functionalities and is tested. later u may find a use for it somewhere, specially if u decide to change TEAL's memory to an array-style memory.  I suggest that you merge it to one of the development branches so in case someone wants to use it, there will be a copy of it.

Everything I did was just because I wanted to help Algorand...

tsachiherman:
> @tsachiherman
> When I opened this PR I also provided the implementation of a memory package which could've been a starting point for changing the current memory style of TEAL. I didn't try to integrate it to the main code because after creating the PR I didn't get any positive feedback from the community or the foundation, so I supposed, even after a lot of work if I integrate it to the main code, still it may not be accepted.
> 
> On the other hand, I think Algorand needs to replace its current persistence layer with a zero knowledge database (ZK-EDB). I know that already some prototypes like Aardvark or Vault exist for this. maybe TEAL memory style and persistent layer could be changed together and some other improvements like what I mentioned in this closed [issue](https://github.com/algorand/go-algorand/issues/1773), could be done at that time too.
> 
> Anyway, still I think this package could be useful. It has many useful functionalities and is tested. later u may find a use for it somewhere, specially if u decide to change TEAL's memory to an array-style memory. I suggest that you merge it to one of the development branches so in case someone wants to use it, there will be a copy of it.
> 
> By the way, I was just trying to help...

Unless I misunderstand you, the current database already has the ZK-EDB properties; we call it catchpoint tracking, but it's essentially the same thing : a way to (continuously) hash an entire database and have a commitment for it's content. At this point, for performance reasons, we use it only primarily on relays and archival nodes - although you can enable it on a non-archival node as well.

I think that your package as is looks good, but I don't know if anyone would pick it up for integrating with the codebase. I'll have a copy of it, if you're ok with that, for future reference.

tsachiherman:
@aybehrouz , this PR contains a new package that isn't linked to anything. Unless you're going to keep working on this one, I'd rather closing this PR ( since nobody is reviewing it ). Once you have a ready to deploy solution, please don't hesitate to share that.

aybehrouz:
I can understand...

However, I still think, current methods of Algorand for serialization of data are not efficient enough. If they remain unchanged, they'll become problematic in the future.

The issues exist in both protocol (usage of maps for representing state of smart contracts) and implementation (using msgpack in map mode https://github.com/algorand/go-algorand/issues/1773, having lots of redundant information in structs https://github.com/algorand/go-algorand/issues/1738)

I propose three solution at different levels:

1. **Doing a quick patch as in https://github.com/algorand/go-algorand/pull/1802:** this way we could mitigate the problem of smart contracts storage in the protocol level, but the implementation would remain inefficient.
2. **Using this PR:** it needs more work but it could mitigate both protocol and implementation problems. However, this solution  is not completely optimal.
3. **Using a VM as my [proposal](https://github.com/aybehrouz/AVM):** It solves all serialization issues and it will make improvements in many other areas.

Unfortunately, code maintainers are not giving me enough feedback. I don't know if they think any of my proposals look good, should I keep working or not? which one of these proposals should I work on? .....

@tsachiherman 
Also, please let me know If u want me to close my other PR.

tsachiherman:
@aybehrouz, I know that it's hard to get answers to some of these design questions.

I'm afraid that these aren't easy challenges. I agree that *one day*, we would definitely need to address these, but at this time, there is no sufficient business motivator for doing that. In order to implement any of these, it would require substantial effort of Algorand developers ( to complete the implementation, test, benchmark, etc. ).

These type of resources aren't available for tasks that aren't on the Algorand roadmap or mandatory ( i.e. critical security bug, for instance ). Another note is that so far Algorand has maintained its backward compatibility. I realize that at some point in the future we might need to break that, but it's going to require an approval from both the product team as well as from the community. As an engineer, I agree that it would be "better" to make the change earlier. However, since it's not a limiting factor ( today ), it would be a hard sell.

Note that I really appreciate the fact that you're challenging the development team here at Algorand. I think that new ideas and fresh views of the system are important - these keep us better "connected" to the community requests & needs.

I hope that this helps.

PR #1764, opened 2020-12-11
Fix catchup concurrently from a single host

tsachiherman:
## Summary

When trying to perform a catchup from a single host, the following error comes up:
```
 dial tcp 127.0.0.1:8088: can't assign requested address
```

The cause of this is that GO runtime runs of out connection. When doing so, it creates new one. And the new connection that was created has a system-wide timeout. The solution for that is to increase the number of MaxIdleConnsPerHost to the expected number of concurrent requests. 

Secondly, once having that solved, the catchup started working fast. In fact, it was working so fast that the accounts update could not keep up, causing the memory utilization to grow up and the "dbatomic" to report slow flushing performance.
To resolve that, we need to hold off newly fetched block until the queue isn't too large. Once we have that, it would be safe to add these to the ledger.

## Test Plan

Tested manually, as automated testing is highly complicated for these use cases and won't add too much value.

### DIFF
--- a/catchup/service.go
+++ b/catchup/service.go
@@ -204,11 +204,11 @@ func (s *Service) fetchAndWrite(fetcher Fetcher, r basics.Round, prevFetchComple
 			if !hasLookback {
 				select {
 				case <-s.ctx.Done():
-					s.log.Debugf("fetchAndWrite(%v): Aborted while waiting for lookback block to ledger after failing once", r)
+					s.log.Infof("fetchAndWrite(%d): Aborted while waiting for lookback block to ledger after failing once : %v", r, err)
 					return false
 				case hasLookback = <-lookbackComplete:
 					if !hasLookback {
-						s.log.Debugf("fetchAndWrite(%v): lookback block doesn't exist, won't try to retrieve block again", r)
+						s.log.Infof("fetchAndWrite(%d): lookback block doesn't exist, won't try to retrieve block again : %v", r, err)
 						return false
 					}
 				}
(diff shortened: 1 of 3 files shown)


catchup/service.go
```diff
@@ -262,11 +262,26 @@ func (s *Service) fetchAndWrite(fetcher Fetcher, r basics.Round, prevFetchComple
...
-				err := s.ledger.AddBlock(*block, *cert)
+				// make sure the ledger wrote enough of the account data to disk, since we don't want the ledger to hold a large amount of data in memory.
+				proto, err := s.ledger.ConsensusParams(r.SubSaturate(1))
+				if err != nil {
+					s.log.Errorf("fetchAndWrite(%d): Unable to determine consensus params for round %d: %v", r, r-1, err)
+					return false
+				}
+				ledgerBacklogRound := r.SubSaturate(basics.Round(proto.MaxBalLookback))
+				select {
+				case <-s.ledger.Wait(ledgerBacklogRound):
```
algonautshant, line 273: What is the reason for choosing ledgerBacklogRound here? 
I don't mind, but I'm curious.

algonautshant (approved):
Looks good.

catchup/service.go
tsachiherman: The account updates stores the latest proto.MaxBalLookback rounds in memory. when we get to a higher number, it means that the "distance" between the blocks database and the account database grows. ( i.e. the account database would be lagging behind ).
That "lagging behind" get translated into few problems:
1. the memory consumption grows ( until the data is flushed ).
2. when the account updates try to flush the changes, it would try to flush "too many" changes. This would lead to dbatomic warning messages.
3. If you'll kill the node at any point, it would need to try and recover on startup. To do so, it would need to rescan all the blocks that weren't included in the account database. If you have 1-2 blocks, that's fine. But if you have 1M blocks, it would cause the node to get "hanged" at startup until the issue is resolved.

For these reasons, I think that it would make sense to hold off the writing until we're done flushing the old rounds.. than keep going.

pzbitskiy (approved):

PR #1768, opened 2020-12-15
Set CadaverSizeTarget to 0 for Scenario testing

nicholasguoalgorand:
## Summary

Set CadaverSizeTarget to 0 so that we don't suffer huge performance losses in tests

## Test Plan

### DIFF
--- a/test/testdata/deployednettemplates/recipes/scenario1/node.json
+++ b/test/testdata/deployednettemplates/recipes/scenario1/node.json
@@ -5,7 +5,7 @@
     "TelemetryURI": "{{TelemetryURI}}",
     "EnableMetrics": false,
     "MetricsURI": "{{MetricsURI}}",
-    "ConfigJSONOverride": "{  \"TxPoolExponentialIncreaseFactor\": 1, \"DNSBootstrapID\": \"<network>.algodev.network\", \"DeadlockDetection\": -1, \"PeerPingPeriodSeconds\": 30, \"BaseLoggerDebugLevel\": 4, \"CadaverSizeTarget\": 0  }",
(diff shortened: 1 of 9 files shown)


test/testdata/deployednettemplates/recipes/scenario1/node.json
```diff
@@ -5,18 +5,18 @@
...
-        "APIToken": "{{APIToken}}",
-        "EnableBlockStats": true,
-        "EnableTelemetry": true,
-        "TelemetryURI": "{{TelemetryURI}}",
-        "EnableMetrics": true,
-        "MetricsURI": "{{MetricsURI}}",
-        "ConfigJSONOverride": "{  \"TxPoolExponentialIncreaseFactor\": 1, \"DNSBootstrapID\": \"<network>.algodev.network\", \"DeadlockDetection\": -1, \"PeerPingPeriodSeconds\": 30, \"EnableAgreementReporting\": true, \"EnableAgreementTimeMetrics\": true, \"EnableAssembleStats\": true, \"EnableProcessBlockStats\": true, \"BaseLoggerDebugLevel\": 4, \"EnableProfiler\": true }",
-        "FractionApply": 0.2
-    }
+    "ConfigJSONOverride": "{  \"TxPoolExponentialIncreaseFactor\": 1, \"DNSBootstrapID\": \"<network>.algodev.network\", \"DeadlockDetection\": -1, \"PeerPingPeriodSeconds\": 30, \"BaseLoggerDebugLevel\": 4, \"CadaverSizeTarget\": 0  }",
```
tsachiherman, line 8: is there any specific reason you removed the `EnableProfiler` flag ?

tsachiherman (changes requested):
only one comment regarding the EnableProfiler on scenario 1; also, could you confirm that the mainnet-model didn't require that because it was already using 0 CadaverSizeTarget ?

egieseke (approved):
Looks OK.

test/testdata/deployednettemplates/recipes/scenario1/node.json
nicholasguoalgorand: I removed it because it wasn't in the net.json file, i'll add it back

tsachiherman (approved):

PR #1770, opened 2020-12-16
Check minimal go version 1.14 prior the build

### LINKED ISSUE #1690: Add a minimal go version check to the build process
**Is your feature request related to a problem? Please describe.**

Trying to compile the algorand repository with the default go version provided on Ubuntu 18.04 doesn't work.
The default version provided by Ubuntu 18.04 is too old.

**Describe the solution you'd like**

Add a test to the build system to check the go compiler version. If the compiler version isn't supported,
the build should fail with an explicit error message

**Additional context**

Issue requested by https://forum.algorand.org/t/cant-complete-catchup-operation-on-mainnet-timeout-after-45-minutes-4m-accounts/2188/6

**Urgency/Priority**
High priority because it's a quick win (in between other tasks). Otherwise, if we need to, we can easily deprioritize.

algorandskiy:
## Summary

Ensure go is 1.14+ prior the build

## Test Plan

Tested manually

Closes #1690

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -175,10 +175,13 @@ build: buildsrc gen
 # get around a bug in go build where it will fail
 # to cache binaries from time to time on empty NFS
 # dirs
-buildsrc: crypto/libs/$(OS_TYPE)/$(ARCH)/lib/libsodium.a node_exporter NONGO_BIN deps $(ALGOD_API_SWAGGER_INJECT) $(KMD_API_SWAGGER_INJECT)
+buildsrc: check-go-version crypto/libs/$(OS_TYPE)/$(ARCH)/lib/libsodium.a node_exporter NONGO_BIN deps $(ALGOD_API_SWAGGER_INJECT) $(KMD_API_SWAGGER_INJECT)
 	mkdir -p tmp/go-cache && \
 	touch tmp/go-cache/file.txt && \
 	GOCACHE=$(SRCPATH)/tmp/go-cache go install $(GOTRIMPATH) $(GOTAGS) $(GOBUILDMODE) -ldflags="$(GOLDFLAGS)" ./...
+
+check-go-version:
+	./scripts/check_golang_version.sh build
 
 SOURCES_RACE := github.com/algorand/go-algorand/cmd/kmd
 
@@ -298,7 +301,7 @@ dump: $(addprefix gen/,$(addsuffix /genesis.dump, $(NETWORKS)))
 install: build
 	scripts/dev_install.sh -p $(GOPATH1)/bin
 
-.PHONY: default fmt vet lint check_shell sanity cover prof deps build test fulltest shorttest clean cleango deploy node_exporter install %gen gen NONGO_BIN
+.PHONY: default fmt vet lint check_shell sanity cover prof deps build test fulltest shorttest clean cleango deploy node_exporter install %gen gen NONGO_BIN check-go-version
 
 ###### TARGETS FOR CICD PROCESS ######
 include ./scripts/release/mule/Makefile.mule
--- a/scripts/check_golang_version.sh
+++ b/scripts/check_golang_version.sh
@@ -12,14 +12,6 @@
 
 set -eo pipefail
 
-# Exit early if running in travis. The reason for this is that travis starts a build with an older version of go that will
-# fail the first check (GO_MOD_SUPPORT), and this isn't necessary for travis builds.
-#
-# https://docs.travis-ci.com/user/environment-variables/#default-environment-variables
-if [[ -n $CI ]] && [[ -n $TRAVIS ]]; then
-    exit 0
-fi
-
 read -ra GOLANG_VERSIONS <<< "$(./scripts/get_golang_version.sh all)"
 BUILD_VERSION="${GOLANG_VERSIONS[0]}"
 MIN_VERSION="${GOLANG_VERSIONS[1]}"
--- a/scripts/check_golang_version.sh
+++ b/scripts/check_golang_version.sh
@@ -16,8 +16,7 @@ set -eo pipefail
 # fail the first check (GO_MOD_SUPPORT), and this isn't necessary for travis builds.
 #
 # https://docs.travis-ci.com/user/environment-variables/#default-environment-variables
-if $CI && $TRAVIS
-then
+if [[ -n $CI ]] && [[ -n $TRAVIS ]]; then
     exit 0
 fi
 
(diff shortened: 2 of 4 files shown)


tsachiherman (changes requested):
you'll need to add the `check-go-version` target as a dependency of `buildsrc` and `gen` as well, since when the user would invoke `make -j4`, you want the `check-go-version` to complete before attempting any of the other two.


I'm not sure why you need to change the `GOPATH` in check_go_version.sh; it was already exported from the makefile, so it should be ready to go.

scripts/check_go_version.sh
```diff
@@ -0,0 +1,27 @@
...
+
+# check_go_version.sh - Check current go version against minimal required
+#
+# Syntax:   check_go_version.sh
+
+GOMINVERSION_MAJOR=1
+GOMINVERSION_MINOR=14
+
+UNAME=$(uname)
+if [[ "${UNAME}" == *"MINGW"* ]]; then
```
brianolson, line 11: Why are we concerned with GOPATH at all? It won't affect what `go` is found on $PATH

algorandskiy: fixed

algorandskiy:
moved `check-go-version` to `buildsrc`

btoll:
@algorandskiy I'm just getting back now and am in a meeting so I haven't really been able to look at this, but I recently did https://github.com/algorand/go-algorand/commit/547b4ee832fe71f2db6b22f5df144bed9b29f2b6 which is similar to what your PR is doing.

Makefile
```diff
@@ -175,11 +175,14 @@ build: buildsrc gen
 # get around a bug in go build where it will fail
 # to cache binaries from time to time on empty NFS
 # dirs
-buildsrc: crypto/libs/$(OS_TYPE)/$(ARCH)/lib/libsodium.a node_exporter NONGO_BIN deps $(ALGOD_API_SWAGGER_INJECT) $(KMD_API_SWAGGER_INJECT)
+buildsrc: check-go-version crypto/libs/$(OS_TYPE)/$(ARCH)/lib/libsodium.a node_exporter NONGO_BIN deps $(ALGOD_API_SWAGGER_INJECT) $(KMD_API_SWAGGER_INJECT)
```
tsachiherman, line 178: The default build target is `default`, which has the `build` target.
In order to make the `build`, it will try and follow the targets for `buildsrc` and `gen` in parallel. ( if -j is provided ).
But we want to make sure that before `gen` we have the ran the `check-go-version`, so you'll need to add that as a dependency for `gen` as well.

Plus, since it's not a real file, you'll need to add that to the `.PHONY` list.

algorandskiy: `gen` does not really matter and indirectly depends on `buildsrc` if I understand it right.
```
gen/%/genesis.json: gen/%.json gen/generate.go buildsrc
gen: $(addsuffix gen, $(NETWORKS)) mainnetgen
```

algorandskiy:
> I'm just getting back now and am in a meeting so I haven't really been able to look at this, but I recently did 547b4ee which is similar to what your PR is doing.

I missed it. But this script kind of non-working:
```
if $CI && $TRAVIS
then
    exit 0
fi
```

Check env and hardcode `MIN=1.15`:
```
$ echo $TRAVIS

$ echo $CI

```

Run
```
./scripts/check_golang_version.sh dev
+ exit 0
```

algorandskiy:
Added to .PHONY, reused Bill's check_golang_version.sh script

scripts/check_golang_version.sh
```diff
@@ -16,7 +16,7 @@ set -eo pipefail
 # fail the first check (GO_MOD_SUPPORT), and this isn't necessary for travis builds.
 #
 # https://docs.travis-ci.com/user/environment-variables/#default-environment-variables
-if $CI && $TRAVIS
+if [[ ! -z $CI ]] && [[ ! -z $TRAVIS ]]
```
btoll, line 19: Since travis sets these env vars as booleans, the check works without the brackets.  What you changed *should* still work, we'll find out soon enough :)

algorandskiy: The problem is these are not set when invoked from Makefile locally and the condition always met

algorandskiy:
> What do you think about something like this?
Done

btoll (dismissed):

algorandskiy:
This a nice one
```
+ ARCH=arm64
+ ./scripts/check_golang_version.sh
./scripts/check_golang_version.sh: line 29: go: command not found
```

tsachiherman:
nice; maybe `export PATH` in the top of the makefile, just like the windows version ?

algorandskiy:
Added `CI=true` and `TRAVIS=true` to external build env.

The condition in `check_golang_version.sh` is true when CI/TRAVIS is unset.
```
  if $CI && $TRAVIS ; then echo 1; fi
```

scripts/check_golang_version.sh
```diff
@@ -16,8 +16,9 @@ set -eo pipefail
 # fail the first check (GO_MOD_SUPPORT), and this isn't necessary for travis builds.
 #
 # https://docs.travis-ci.com/user/environment-variables/#default-environment-variables
-if $CI && $TRAVIS
-then
+#
+# TRAVIS_BRANCH is set for external ARM builds, and CI/TRAVIS are not set
+if [[ -n $CI ]] && [[ -n $TRAVIS ]] || [[ -n $TRAVIS_BRANCH ]]; then
```
tsachiherman, line 21: generally speaking, you should avoid using travis-specific variables outside the "/scripts/travis" directory.

algorandskiy: It looks like a comment for #1702. Open to suggestions? call `./scripts/travis/istravis_build.sh` here?

tsachiherman: no, just export the path from the makefile. The reason it doesn't find the go compiler is because the makefile doesn't export the PATH to this script.

algorandskiy: It fails not from Makefile but from `travis/build.sh` when go is not installed:
```
if ! ./scripts/check_golang_version.sh
then
    exit 1
fi
# Get the go build version.
GOLANG_VERSION=$(./scripts/get_golang_version.sh)

curl -sL -o ~/gimme https://raw.githubusercontent.com/travis-ci/gimme/master/gimme
```

algorandskiy: just because the old condition `if $CI && $TRAVIS` returns true if CI/TRAVIS unset, as it happens in external build.

tsachiherman: so I guess that's where you need to fix it, right ? i.e. avoid checking the Golang version if no go is installed.

algorandskiy: I'm not sure what is better. Under CI you have either older go and `[ -n $CI ]] && [[ -n $TRAVIS ]]` skip that or no go at all. So the question why do check go version in `travis/build.sh` @btoll ?

tsachiherman: yes, agree with @algorandskiy ; sounds to me like this check isn't very meaningful in travis specific executions since it's getting skipped anyway.

btoll (approved):
Looks good @algorandskiy, thanks for fixing up.

PR #1771, opened 2020-12-17
Disable cadaver for mainnet model

nicholasguoalgorand:
## Summary

Disable cadaver logging for mainnet model for higher performance

## Test Plan


tsachiherman:
you'll need to rebase against master, and leave only the mainnet-model files. looks good otherwise.

tsachiherman (approved):
Looks good

PR #1772, opened 2020-12-18
Compact accounts deltas before persisting to disk

tsachiherman:
## Summary

When we run catchup, we often attempt to persist multiple rounds in a single sqlite transaction. When doing so, we were iterating throughout the rounds we were attempting to update and flushing all the deltas for each individual round.

While this work correctly, it's also not the most efficient, since we are updating the same account multiple times. To address that issue, this PR compact the deltas into a single map, which corresponds to the entire change set across the updated rounds range.

## Test Plan

Unit test and benchmark added. Ran a test against mainnet and it seems to be working correctly with roughly 30% performance increase.

### DIFF
--- a/ledger/acctupdates.go
+++ b/ledger/acctupdates.go
@@ -1361,7 +1361,7 @@ func (au *accountUpdates) deleteStoredCatchpoints(ctx context.Context, dbQueries
 	return nil
 }
 
-// accountsUpdateBalances applies the given deltas array to the merkle trie
+// accountsUpdateBalances applies the given deltas map to the merkle trie
 func (au *accountUpdates) accountsUpdateBalances(accountsDeltas map[basics.Address]accountDelta) (err error) {
 	if au.catchpointInterval == 0 {
 		return nil
@@ -1830,6 +1830,8 @@ func (au *accountUpdates) commitRound(offset uint64, dbRound basics.Round, lookb
 		}
 	}
 
+	// compact all the deltas - when we're trying to persist multiple rounds, we might have the same account
+	// being updated multiple times. When that happen, we can safely omit the intermediate updates.
 	compactDeltas, compactCreatableDeltas := compactDeltas(deltas, creatableDeltas)
 
 	var catchpointLabel string
@@ -1859,12 +1861,12 @@ func (au *accountUpdates) commitRound(offset uint64, dbRound basics.Round, lookb
 			}
 			treeTargetRound = dbRound + basics.Round(offset)
 		}
-		//for i := uint64(0); i < offset; i++ {
+
 		err = accountsNewRound(tx, compactDeltas, compactCreatableDeltas, genesisProto)
 		if err != nil {
 			return err
 		}
-		//}
+
 		err = totalsNewRounds(tx, deltas[:offset], roundTotals[1:offset+1], protos[1:offset+1])
 		if err != nil {
 			return err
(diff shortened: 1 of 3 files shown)


ledger/acctupdates.go
```diff
@@ -1971,6 +1974,61 @@ func (au *accountUpdates) commitRound(offset uint64, dbRound basics.Round, lookb
...
+		for _, roundCreatable := range creatableDeltas {
+			for creatableIdx, creatable := range roundCreatable {
+				if prev, has := outCreatableDeltas[creatableIdx]; has {
+					outCreatableDeltas[creatableIdx] = modifiedCreatable{
+						ctype:   creatable.ctype,
+						created: creatable.created,
+						creator: creatable.creator,
+						// The following ndeltas is not currently used
+						// TODO : it should be "converted" to counting the number of updates for a particular creatable entry and replace the creatableFlushcount in commitRound.
+						ndeltas: creatable.ndeltas + prev.ndeltas,
```
algorandskiy, line 2020: they are set in `newBlockImpl` and checked in `commitRound`. Shouldn't `ndeltas` increment per seen entry instead of summing?

tsachiherman: as with the comment, the content of this field is not being used *at all*. I will provide proper implementation on a subsequent PR, where these would be used instead the existing flushCount. ( but we're not there yet - that's why I added the comment ).

tsachiherman: and yes, it will become a counter of how many updates we had across the rounds.

algonautshant (approved):
Looks good.

ledger/acctupdates.go
```diff
@@ -1971,6 +1974,61 @@ func (au *accountUpdates) commitRound(offset uint64, dbRound basics.Round, lookb
 
 }
 
+// compactDeltas takes an arary of account map deltas ( one array entry per round ), and corresponding creatables array, and compact the arrays into a single
+// map that contains all the account deltas changes. While doing that, the function eliminate any intermediate account changes.
+// The function returns a map that could be the same as the input map, and therefore both the outputs of this function as well as the inputs to this function
+// should be kept read-only ( unless we don't care about the correctness of the data anymore ).
```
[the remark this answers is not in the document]
tsachiherman: right now, the input is already a read-only. The reason it's needed is because in the case of `if len(accountDeltas) == 1 && len(creatableDeltas) == 1`, we're copying the input maps into the output..
so, a change to either one of them would change the other.

This is not a real issue, since we don't modify either of these.

ledger/acctupdates.go
```diff
@@ -1971,6 +1974,61 @@ func (au *accountUpdates) commitRound(offset uint64, dbRound basics.Round, lookb
...
+func compactDeltas(accountDeltas []map[basics.Address]accountDelta, creatableDeltas []map[basics.CreatableIndex]modifiedCreatable) (outAccountDeltas map[basics.Address]accountDelta, outCreatableDeltas map[basics.CreatableIndex]modifiedCreatable) {
+	// this is a common case where we have only a single round that we try to flush. In this case, we're just going to return
+	// the first entry.
+	if len(accountDeltas) == 1 && len(creatableDeltas) == 1 {
```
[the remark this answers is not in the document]
tsachiherman: the length of both arrays is always identical ( look at the caller ), and should always be 1 or greater.
For functional correctness, what you've suggested won't be correct since we're indexing the first entry inside this if clause, which would cause an access violation. I think that the `len=0` is already handled correctly in the subsequent two if statements.

PR #1776, opened 2020-12-20
systemd: run algod as a user service

btoll:
## Summary

Add a user service for `algod` that is part of the tarball downloaded by `updater`.  In addition, add functionality to `update.sh` to invoke `systemctl` with the `--user` flag to stop the user service.

Installing the service as usual has not changed:
`sudo ./systemd-setup.sh $USER $GROUP`

Installing as a user service:
`./systemd-setup-user.sh $USER`

Docs: https://github.com/algorand/docs/pull/279

## Test Plan

Added to `systemd-setup.sh` to install `systemd` scripts in correct location depending on how the script is invoked.

Tested the old path by installing system-wide as root or as a user in the sudo group.  Tested the new path by invoking the shell script as outlined above and running all `systemctl` commands with the `--user` flag.

Tested on both Linux and MacOS.  For the latter, ensured that the node was started/stopped using `goal`.

---
Test `./test_update.sh`:
```
./test_update.sh
```

Test `./test_systemd-setup.sh`:
```
# Test a systemd user service.
./test_systemd-setup.sh -u $USER -g --user

# Test a systemd system service.
sudo ./test_systemd-setup.sh -u $USER -g $USER
```


installer/algorand@.service.template-user
```diff
@@ -0,0 +1,30 @@
+## This file should be installed as:
+##
+##   /home/$USER/.config/systemd/user/algorand@.service
+##
+## and then running "systemctl --user daemon-reload".
+##
+## To enable and start algod for a particular directory, run:
+##
+##   systemctl --user enable algorand@$(systemd-escape /home/algo/algorand/testnet/data/relay)
+##   systemctl --user start  algorand@$(systemd-escape /home/algo/algorand/testnet/data/relay)
```
btoll, line 10: The instructions are different from the other template.

installer/algorand@.service.template-user
```diff
@@ -0,0 +1,30 @@
...
+## This assumes a directory layout where the algod binary lives at
+## /home/algo/algorand/testnet/algod, and the data lives under
+## /home/algo/algorand/testnet/data/relay.
+##
+## We do not use systemd user services because they are disabled in CentOS.
+
+[Unit]
+Description=Algorand daemon under %I
+After=network.target
+AssertPathExists=%I
```
btoll, line 21: The service unit for a user cannot have user and group keys.

cmd/updater/systemd-setup.sh
```diff
@@ -1,26 +1,48 @@
...
 
 USER="$1"
-GROUP="$2"
 
-SCRIPTPATH="$( cd "$(dirname "$0")" ; pwd -P )"
-SEDARGS="-e s,@@USER@@,${USER}, -e s,@@GROUP@@,${GROUP},"
+if ! id -u "${USER}"> /dev/null; then
+    echo "$0 [ERROR] User \`$USER\` does not exist on system"
+    exit 1
+fi
```
btoll, line 39: Should we check for group as well?

cmd/updater/update.sh
```diff
@@ -330,14 +330,27 @@ function validate_update() {
     return 0
 }
 
+function run_systemd_action() {
+    local action=$1
+    local data_dir=$2
+
+    # Get the owner of a running process.
+    owner=$(awk '{ print $1 }' <(ps aux | grep "[a]lgod"))
```
tsachiherman, line 338: the grep here is a bit weak predicate.. if could match other processes that you didn't mean.. maybe do some extra filtering in the awk ?

cmd/updater/update.sh
```diff
@@ -330,14 +330,27 @@ function validate_update() {
...
+    if [[ "$owner" = "root" ]] || grep sudo <(groups "$owner"); then
+        sudo -n systemctl "${action}" "algorand@$(systemd-escape "${data_dir}")"
```
tsachiherman, line 340: I think that before you're going to do that, you need to verify that you're running as root. If not, this operation is going to fail anyway.
( and it need to be checked at the very beginning of the update script ).

tsachiherman, line 340: I think that before you're going to do that, you need to verify that you're running as root. If not, this operation is going to fail anyway.
( and it need to be checked at the very beginning of the update script ).

tsachiherman:
does that work under mac ? if not, we need to test that it's a mac and abort with an error code.

cmd/updater/update.sh
btoll: I can see if it matches the data directory as well.

cmd/updater/update.sh
btoll: Ok, will look at this more.

installer/algorand@.service.template-user
```diff
@@ -0,0 +1,30 @@
...
+## To enable and start algod for a particular directory, run:
+##
+##   systemctl --user enable algorand@$(systemd-escape /home/algo/algorand/testnet/data/relay)
+##   systemctl --user start  algorand@$(systemd-escape /home/algo/algorand/testnet/data/relay)
+##
+## This assumes a directory layout where the algod binary lives at
+## /home/algo/algorand/testnet/algod, and the data lives under
+## /home/algo/algorand/testnet/data/relay.
+##
+## We do not use systemd user services because they are disabled in CentOS.
```
brianolson, line 16: This line doesn't make sense to me, since this file is establishing a systemd user service

btoll: Haha, yeah, I'll change that.

btoll:
> does that work under mac ? if not, we need to test that it's a mac and abort with an error code.

I'll abort with an error code if it's mac.  I'm not sure how it ever would have worked currently once it would hit the `systemctl` command...

I created this ticket if we want to support `launchd` in the future: https://github.com/algorand/go-algorand/issues/1779

tsachiherman (commented):
it looks ok to me, but it would require quite some testing. I don't see any testing here, and I wonder if we have any testing facilities for this script.
( I'm well aware that testing this script isn't easy.. but as it's getting more sophisticated, we need to develop better ways to validate the correctness of this solution )

cmd/updater/systemd-setup.sh
```diff
@@ -1,26 +1,48 @@
...
+# shellcheck disable=2086,2164
 #
 # Run this script to set up the systemd service for algod.
-# The arguments are the username and groupname that algod
-# should run as.
+# The arguments are the username and groupname that algod should run as.
+# If wanting to run as a user service, substitute "--user" for the groupname.
+
+setup_user() {
+    mkdir -p "/home/$1/.config/systemd/user"
```
bricerisingalgorand, line 9: This may be a better way to get the user home dir in case anyone does any nonstandard setups
```
eval echo ~$1
```

cmd/updater/systemd-setup.sh
bricerisingalgorand: I think that makes sense 👍

cmd/updater/systemd-setup.sh
btoll: https://unix.stackexchange.com/questions/332579/is-it-always-safe-to-use-eval-echo

cmd/updater/update.sh
```diff
@@ -330,20 +335,63 @@ function validate_update() {
...
+                systemctl --user "${action}" "algorand@$(systemd-escape "${data_dir}")" &> /dev/null
+            fi
+        elif [ "$user_service" -ne 4 ]; then
+            if [ "$action" = "start" ]; then
+                systemctl --user "${action}" "algorand@$(systemd-escape "${data_dir}")" &> /dev/null
+            fi
+        fi
+    fi
+}
+
```
egieseke, line 380: Consider adding some echo statements to indicate what path is chosen for the given data dir.

cmd/updater/update.sh
```diff
@@ -330,20 +335,63 @@ function validate_update() {
...
+        if [ "$user_service" -eq 0 ]; then
+            process_owner=$(awk '{ print $1 }' <(ps aux | grep "[a]lgod -d ${data_dir}"))
+            if [ "$action" = "stop" ] && [ "$(whoami)" = "$process_owner" ]; then
+                systemctl --user "${action}" "algorand@$(systemd-escape "${data_dir}")" &> /dev/null
+            fi
+        elif [ "$user_service" -ne 4 ]; then
+            if [ "$action" = "start" ]; then
+                systemctl --user "${action}" "algorand@$(systemd-escape "${data_dir}")" &> /dev/null
+            fi
+        fi
```
egieseke, line 377: Should you include an else clause for $user_service == 4 and print an error.

cmd/updater/update.sh
```diff
@@ -330,20 +335,63 @@ function validate_update() {
...
 function shutdown_node() {
     echo Stopping node...
     if [ "$(pgrep -x algod)" != "" ] || [ "$(pgrep -x kmd)" != "" ] ; then
-        if [ -f ${BINDIR}/goal ]; then
-            for DD in ${DATADIRS[@]}; do
-                if [ -f ${DD}/algod.pid ] || [ -f ${DD}/**/kmd.pid ] ; then
+        if [ -f "${BINDIR}/goal" ]; then
+            for DD in "${DATADIRS[@]}"; do
+                if [ -f "${DD}/algod.pid" ] || [ -f "${DD}"/**/kmd.pid ] ; then
                     echo Stopping node and waiting...
```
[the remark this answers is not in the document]
btoll: Will add.

cmd/updater/update.sh
btoll: Good catch, it should return a non-zero exit code so the startup code will start with `goal`.

cmd/updater/update.sh
btoll: Will look at adding some helpful info.

cmd/updater/update.sh
btoll: Thinking about it, I may re-work the caller a bit.  It seems non-intuitive to return an error here since a service doesn't have to be installed.

cmd/updater/systemd-setup.sh
bricerisingalgorand: I mean, we could make sure the string is an actual user first if we're worried about us maliciously using our scripts

bricerisingalgorand: If there's a better way that's fine too, we just need to make sure we're actually using the user's home directory

bricerisingalgorand: Since this script is specified for bash, we don't need the eval apparently. That part's only necessary on sh, so this should work
```
echo ~"$1"
```

bricerisingalgorand: Actually never mind, the eval is still necessary I think. It worked on my mac, but on ubuntu I had to use the eval

```
root@22112cf00d84:/# useradd -d /tmp/hodor hodor
root@22112cf00d84:/# echo ~"hodor"
~hodor
root@22112cf00d84:/# eval echo ~"hodor"
/tmp/hodor
```

btoll: I fixed this.  It's getting it from `/etc/passwd` which is the most reliable way.

cmd/updater/update.sh
```diff
@@ -330,20 +340,75 @@ function validate_update() {
...
+}
+
+function has_user_service() {
+    local dd="$1"
+    local path
+
+    path=$(awk -F= '{ print $2 }' <(systemctl --user show -p FragmentPath "algorand@$(systemd-escape "$dd")"))
+    if [ "$path" != "" ]; then
+        return 0
+#        state=$(systemctl --user show -p ActiveState --value "algorand@$(systemd-escape "$dd")")
```
tsachiherman, line 361: This need to be dropped, right ?

btoll: Yes, in fact I should combine those two functions because they're virtually identical.

cmd/updater/update.sh
```diff
@@ -93,6 +100,9 @@ while [ "$1" != "" ]; do
             shift
             TOOLS_OUTPUT_DIR=$1
             ;;
+        -z)
+            TESTING=true
```
tsachiherman, line 104: This is a released file, so i'm not sure if we want test mode, but it would make sense to have a dryrun mode.

btoll: I'll make the change.

btoll: Fixed

cmd/updater/update.sh
btoll: I removed the commented-out code and combined the two functions into one.

cmd/updater/update.sh
btoll: I'm going to resolve this conversation since this particular code has been refactored and the comments are outdated.

installer/algorand@.service.template-user
```diff
@@ -0,0 +1,28 @@
...
+## /home/algo/algorand/testnet/algod, and the data lives under
+## /home/algo/algorand/testnet/data/relay.
+
+[Unit]
+Description=Algorand daemon under %I
+After=network.target
+AssertPathExists=%I
+
+[Service]
+ExecStart=/bin/bash -c "exec /%I/../algod -d %I"
```
brianolson, line 22: It's awfully fragile that the bin dir must be relative to the data dir. Is that documented? Can we fix that and substitute in the bin dir when installing this in systemd-setup ? (e.g. use `sed` instead of `cp`). Then we could drop the `bash` layer to and just have `ExecStart=/path/to/algod -d %I`

cmd/updater/update.sh
```diff
@@ -330,20 +340,67 @@ function validate_update() {
     return 0
 }
 
+function check_service() {
```
brianolson, line 343: there's nothing in common between the "user" and "system" branches of this function. It might be clearer as two different functions check_service_user() and check_service_system()

cmd/updater/systemd-setup.sh
```diff
@@ -1,26 +1,60 @@
...
 #
 # Run this script to set up the systemd service for algod.
-# The arguments are the username and groupname that algod
-# should run as.
+# The arguments are the username and groupname that algod should run as.
+# If wanting to run as a user service, substitute "--user" for the groupname.
+
+set -e
+
+setup_user() {
```
brianolson, line 10: There's nothing in common between setup_user() and setup_root(), would this script be clearer as two scripts? systemd-setup.sh and systemd-user-setup.sh ?

installer/algorand@.service.template-user
btoll: I agree that it's fragile.  I'll look into this as you suggested.

I do have some new docs here: https://github.com/algorand/docs/pull/279

cmd/updater/update.sh
btoll: I initially had them as two separate functions, but then the only bits there were different between them were `--user` and it seemed better to combine them.

cmd/updater/systemd-setup.sh
btoll: I'll break these into two scripts.

btoll: Done and retested.

installer/algorand@.service.template-user
btoll: No, it's not documented.

egieseke (approved):

bricerisingalgorand (approved):
I think this looks good assuming it's been tested very thoroughly

bricerisingalgorand:
I ran Ben's test script and performed these tests using the systemd setup script locally and they all passed. I believe this PR accomplishes the goal of supporting user systemd services

```
vagrant@vagrant:~/node$ ./systemd-setup-user.sh `whoami`
[INFO] BINDIR is unset, setting to cwd.
[SUCCESS] systemd user service has been installed.
vagrant@vagrant:~/node$ systemctl start --user algorand@$(systemd-escape `pwd`/data)
vagrant@vagrant:~/node$ systemctl status --user algorand@$(systemd-escape `pwd`/data)
● algorand@-home-vagrant-node-data.service - Algorand daemon under /home/vagrant/node/data
   Loaded: loaded (/home/vagrant/.config/systemd/user/algorand@.service; disabled; vendor preset: enabled)
   Active: active (running) since Thu 2021-02-04 04:55:27 UTC; 5s ago
 Main PID: 3602 (algod)
   CGroup: /user.slice/user-1000.slice/user@1000.service/algorand.slice/algorand@-home-vagrant-node-data.service
           └─3602 /home/vagrant/node/algod -d /home/vagrant/node/data
Feb 04 04:55:27 vagrant systemd[1299]: Started Algorand daemon under /home/vagrant/node/data.
Feb 04 04:55:27 vagrant algod[3602]: Telemetry configured from '/home/vagrant/.algorand/logging.config'
Feb 04 04:55:27 vagrant algod[3602]: Logging to:  /home/vagrant/node/data/node.log
Feb 04 04:55:27 vagrant algod[3602]: Deadlock detection is set to: disabled (Default state is 'disable')
Feb 04 04:55:27 vagrant algod[3602]: Initializing the Algorand node... Success!
Feb 04 04:55:27 vagrant algod[3602]: Node running and accepting RPC requests over HTTP on port 127.0.0.1:8080. Press Ctrl-C to exit
Feb 04 04:55:27 vagrant algod[3602]: ⇨ http server started on 127.0.0.1:8080
```

```
vagrant@vagrant:~/node$ ./goal node stop -d data
The node was successfully stopped.
vagrant@vagrant:~/node$ systemctl status --user algorand@$(systemd-escape `pwd`/data)
● algorand@-home-vagrant-node-data.service - Algorand daemon under /home/vagrant/node/data
   Loaded: loaded (/home/vagrant/.config/systemd/user/algorand@.service; disabled; vendor preset: enabled)
   Active: activating (auto-restart) since Thu 2021-02-04 04:57:38 UTC; 2s ago
  Process: 3602 ExecStart=/home/vagrant/node/algod -d /home/vagrant/node/data (code=exited, status=0/SUCCESS)
 Main PID: 3602 (code=exited, status=0/SUCCESS)
vagrant@vagrant:~/node$ ./goal node start -d data
Algorand node was already started!
vagrant@vagrant:~/node$ systemctl status --user algorand@$(systemd-escape `pwd`/data)
● algorand@-home-vagrant-node-data.service - Algorand daemon under /home/vagrant/node/data
   Loaded: loaded (/home/vagrant/.config/systemd/user/algorand@.service; disabled; vendor preset: enabled)
   Active: active (running) since Thu 2021-02-04 04:57:44 UTC; 39s ago
 Main PID: 3624 (algod)
   CGroup: /user.slice/user-1000.slice/user@1000.service/algorand.slice/algorand@-home-vagrant-node-data.service
           └─3624 /home/vagrant/node/algod -d /home/vagrant/node/data
Feb 04 04:57:44 vagrant systemd[1299]: Started Algorand daemon under /home/vagrant/node/data.
Feb 04 04:57:44 vagrant algod[3624]: Telemetry configured from '/home/vagrant/.algorand/logging.config'
Feb 04 04:57:44 vagrant algod[3624]: Logging to:  /home/vagrant/node/data/node.log
Feb 04 04:57:44 vagrant algod[3624]: Deadlock detection is set to: disabled (Default state is 'disable')
Feb 04 04:57:44 vagrant algod[3624]: Initializing the Algorand node... Success!
Feb 04 04:57:44 vagrant algod[3624]: Node running and accepting RPC requests over HTTP on port 127.0.0.1:8080. Press Ctrl-C to exit
Feb 04 04:57:44 vagrant algod[3624]: ⇨ http server started on 127.0.0.1:8080
```

PR #1777, opened 2020-12-21
Convert the catchpointWriting from a channel to an atomic variable

tsachiherman:
## Summary

Convert the catchpointWriting from a channel to an atomic variable. In the past, we used to have a value in having this as a channel, however, this is no longer the case. I have converted this to an atomic variable to improve code simplicity and readability.

## Test Plan

Unit tests were updated.

### DIFF
--- a/ledger/acctupdates_test.go
+++ b/ledger/acctupdates_test.go
@@ -1046,11 +1046,11 @@ func TestIsWritingCatchpointFile(t *testing.T) {
 
 	au := &accountUpdates{}
 
-	au.catchpointWriting = make(chan struct{}, 1)
+	au.catchpointWriting = -1
 	ans := au.IsWritingCatchpointFile()
 	require.True(t, ans)
 
-	close(au.catchpointWriting)
+	au.catchpointWriting = 0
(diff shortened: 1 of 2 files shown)


algorandskiy (approved):

ledger/acctupdates_test.go
```diff
@@ -1046,11 +1046,11 @@ func TestIsWritingCatchpointFile(t *testing.T) {
 
 	au := &accountUpdates{}
 
-	au.catchpointWriting = make(chan struct{}, 1)
+	au.catchpointWriting = -1
```
algonautshant, line 1049: Why not use the same statement used in the production code for setting the value?

algonautshant (approved):

ledger/acctupdates_test.go
tsachiherman: you mean the `atomic.StoreInt32` ? There is no concurrency here, so I'm not sure why that would make a difference. Keep in mind that it's literally just store a value, nothing else; I think that in the context of this test function isn't not really important. If it was used in a scenario where we try to modify it while there is some go-routine running, than yes - it would be a good idea.

PR #1778, opened 2020-12-21
Add make and g++ to configure_dev.sh

algorandskiy:
## Summary

When building on vanila Ubuntu 20.04 make and g++ are missing

## Test Plan

This is a build script, tested on a Ubuntu 20.04

### DIFF
--- a/scripts/configure_dev.sh
+++ b/scripts/configure_dev.sh
@@ -75,7 +75,7 @@ if [ "${OS}" = "linux" ]; then
     fi
 
     sudo apt-get update
-    sudo apt-get install -y libboost-all-dev expect jq autoconf shellcheck sqlite3 python3-venv
+    sudo apt-get install -y libboost-all-dev expect jq autoconf shellcheck sqlite3 python3-venv make g++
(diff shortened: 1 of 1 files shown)


scripts/configure_dev.sh
```diff
@@ -75,7 +75,7 @@ if [ "${OS}" = "linux" ]; then
     fi
 
     sudo apt-get update
-    sudo apt-get install -y libboost-all-dev expect jq autoconf shellcheck sqlite3 python3-venv
+    sudo apt-get install -y libboost-all-dev expect jq autoconf shellcheck sqlite3 python3-venv make g++
```
tsachiherman, line 78: are these missing on vanilla Ubuntu 18.04 ? If not, we need to branch.

onetechnical: Wouldn't `build-essential` be preferable here?

onetechnical (commented):
Suggest a different package for building.

algorandskiy:
both make and g++ are installed as dependencies of something on ubuntu 18.04. So closing

PR #1782, opened 2020-12-22
Pass transaction groups correctly to logicsig evaluator

tsachiherman:
## Summary

For the `goal clerk sign`, it would perform the following:
1. Group the transactions from the input file according to their Group field.
2. When using a logic signature, create a verification GroupContext and verify each transaction according to it's location within the group.
3. Retain the original ordering of the groups, based on the first transaction from each of the transaction groups.

For the `goal clerk dryrun`, it would perform the following:
1. Add GroupIndex and TxnGroup to the eval params.

### DIFF
--- a/cmd/goal/clerk.go
+++ b/cmd/goal/clerk.go
@@ -726,7 +726,7 @@ var signCmd = &cobra.Command{
 		dec := protocol.NewDecoderBytes(data)
 		// read the entire file and prepare in-memory copy of each signed transaction, with grouping.
 		txnGroups := make(map[crypto.Digest][]*transactions.SignedTxn)
-		groupsOrder := make([]crypto.Digest, 0)
+		var groupsOrder []crypto.Digest
 		txnIndex := make(map[*transactions.SignedTxn]int)
 		count := 0
 		for {
--- a/cmd/goal/clerk.go
+++ b/cmd/goal/clerk.go
@@ -724,46 +724,77 @@ var signCmd = &cobra.Command{
 
 		var outData []byte
 		dec := protocol.NewDecoderBytes(data)
+		// read the entire file and prepare in-memory copy of each signed transaction, with grouping.
+		txnGroups := make(map[crypto.Digest][]*transactions.SignedTxn)
+		groupsOrder := make([]crypto.Digest, 0)
+		txnIndex := make(map[*transactions.SignedTxn]int)
 		count := 0
 		for {
-			// transaction file comes in as a SignedTxn with no signature
-			var uncheckedTxn transactions.SignedTxn
-			err = dec.Decode(&uncheckedTxn)
+			uncheckedTxn := new(transactions.SignedTxn)
+			err = dec.Decode(uncheckedTxn)
 			if err == io.EOF {
 				break
 			}
 			if err != nil {
 				reportErrorf(txDecodeError, txFilename, err)
 			}
-
-			var signedTxn transactions.SignedTxn
-			if lsig.Logic != nil {
-				proto, _ := getProto(protoVersion)
-				uncheckedTxn.Lsig = lsig
-				blockHeader := bookkeeping.BlockHeader{
-					UpgradeState: bookkeeping.UpgradeState{
-						CurrentProtocol: proto,
-					},
-				}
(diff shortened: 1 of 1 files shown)


cmd/goal/clerk.go
```diff
@@ -724,46 +724,77 @@ var signCmd = &cobra.Command{
 
 		var outData []byte
 		dec := protocol.NewDecoderBytes(data)
+		// read the entire file and prepare in-memory copy of each signed transaction, with grouping.
+		txnGroups := make(map[crypto.Digest][]*transactions.SignedTxn)
+		groupsOrder := make([]crypto.Digest, 0)
```
brianolson, line 729: don't need to `make()`, `append()` and `range` will do the right thing with a nil array; can just `var groupsOrder []crypto.Digest`; OR preallocate so the append()s go faster `make([]T, 0, 5)`

brianolson (approved):
I feel like the use case of using `goal clerk sign` to sign several txns at once is a little weird, but this code looks fine.

cmd/goal/clerk.go
```diff
@@ -724,46 +724,77 @@ var signCmd = &cobra.Command{
 
 		var outData []byte
 		dec := protocol.NewDecoderBytes(data)
+		// read the entire file and prepare in-memory copy of each signed transaction, with grouping.
+		txnGroups := make(map[crypto.Digest][]*transactions.SignedTxn)
```
algorandskiy, line 728: nit: It is OK to use obj copies instead of pointers in CLI code (imo code without pointers looks cleaner, and this is a CLI so mem efficiency has low impact)

algorandskiy (commented):
Looks good
1. please consider switching back from pointers to values
2. ensure to preserve PR message in the commit when merge

cmd/goal/clerk.go
```diff
@@ -724,46 +724,77 @@ var signCmd = &cobra.Command{
...
 			if err != nil {
 				reportErrorf(txDecodeError, txFilename, err)
 			}
+			group := uncheckedTxn.Txn.Group
+			if group.IsZero() {
+				// create a dummy group.
+				randGroupBytes := crypto.Digest{}
+				crypto.RandBytes(randGroupBytes[:])
+				group = randGroupBytes
+			}
```
jannotti, line 747: I don't think there's anything wrong here, but I would like to understand how this works.  I expected that the rule for group id was that if it was set, it must be the hash of the entire group, but this code makes me think there's nothing in the blockchain that checks that (since you're allocating a random group id).  Is that right?  The "rule" that a group id must be the hash of the whole group is just a best practice that should be done by users as they decide what transactions to sign. (If they check the hash, they know they're not signing something surprising.) But the blockchain does not care.

cmd/goal/clerk.go
tsachiherman: good point; however, in this case where I can't tell ( easily ) ahead of time, I'll just use
```
var groupsOrder []crypto.Digest
```

cmd/goal/clerk.go
tsachiherman: I think that you missed the point of me using pointers... Pointers are unique, whereas `transactions.SignedTxn` might not be. If the input has the same transaction multiple times, and we load it all into a map and then "dump" it, we would skip duplicate transactions. Another reason is that it allows me to "tag" each transaction with it's index within the input file, which could be problematic if we were using `transactions.SignedTxn`.

cmd/goal/clerk.go
tsachiherman: The blockchain does care. look in `func (eval *BlockEvaluator) TestTransactionGroup(txgroup []transactions.SignedTxn) error`.

The group here is used *only* for the purpose of grouping the transactions from the input file before being processed. It does not *modify* the Group value within the transaction itself.
The allocation of the random group here is used only for the purpose of preserving the original transaction ordering of the input file.

cmd/goal/clerk.go
algorandskiy: ah, I see it is really only needed for  proper index displaying `txnIndex[txnGroups[group][i]]` in the error message.

tsachizehub (approved):

PR #1785, opened 2020-12-22
Better kmd permission error message. Closes #1784

### LINKED ISSUE #1784: kmd permission error message could be more actionable
## Summary

When the data dir is too permissive we give an error which says what the current permissions are. It should also tell you what they should be.

## Scope/Requirements
change error message in `nodecontrol/kmdControl.go`

winder:
Better kmd permission error

### DIFF
--- a/nodecontrol/kmdControl.go
+++ b/nodecontrol/kmdControl.go
@@ -192,7 +192,7 @@ func (kc *KMDController) StartKMD(args KMDStartArgs) (alreadyRunning bool, err e
 			return false, errors.New("bad kmd data dir")
 		}
 		if (dataDirStat.Mode() & 0077) != 0 {
-			logging.Base().Errorf("%s: kmd data dir exists but is too permissive (%o), change to (%o)", kc.kmdDataDir, dataDirStat.Mode()&0777, dataDirStat.Mode()^0077)
+			logging.Base().Errorf("%s: kmd data dir exists but is too permissive (%o), change to (%o)", kc.kmdDataDir, dataDirStat.Mode()&0777, DefaultKMDDataDirPerms)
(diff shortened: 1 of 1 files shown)


nodecontrol/kmdControl.go
```diff
@@ -192,7 +192,7 @@ func (kc *KMDController) StartKMD(args KMDStartArgs) (alreadyRunning bool, err e
 			return false, errors.New("bad kmd data dir")
 		}
 		if (dataDirStat.Mode() & 0077) != 0 {
-			logging.Base().Errorf("%s: kmd data dir exists but is too permissive (%o)", kc.kmdDataDir, dataDirStat.Mode()&0777)
+			logging.Base().Errorf("%s: kmd data dir exists but is too permissive (%o), change to (%o)", kc.kmdDataDir, dataDirStat.Mode()&0777, dataDirStat.Mode()&0077)
```
tsachiherman, line 195: I believe that you meant to have the following on the second one, right ? 
`dataDirStat.Mode()^0077`

tsachiherman (changes requested):
I think the predicate would need a small tweak..

nodecontrol/kmdControl.go
vdsprakash: "_dataDirStat.Mode()&0077_"  can be replaced with "_dataDirStat.Mode()&0644_" to show the right permissions. Just a suggestion.

winder: Thank you for the suggestions, this should be fixed now.

winder:
@tsachiherman ack. Looks like there is a default permission constant that can be used for the suggested value.

tsachiherman (approved):
looks good to me.

PR #1792, opened 2020-12-27
bugfix: incorrect block download retry accounting during fast catchup

tsachiherman:
## Summary

Incorrect block download retry accounting during fast catchup. The existing code was counting successful block retrieval as an retry attempt, which would "exhaust" the `CatchupBlockDownloadRetryAttempts` before the catchup is complete, causing the fast catchup to fail in case the `CatchupBlockDownloadRetryAttempts` was set too low.

The workaround of this is to set the `CatchupBlockDownloadRetryAttempts` to a value that is higher by 1000 than then number of desired attempts count.

## Test Plan

Tested by running fast catchup and observing the failed attempts logs.

### DIFF
--- a/catchup/catchpointService.go
+++ b/catchup/catchpointService.go
@@ -453,7 +453,7 @@ func (cs *CatchpointCatchupService) processStageBlocksDownload() (err error) {
 	blocksFetched := uint64(1) // we already got the first block in the previous step.
 	var blk *bookkeeping.Block
 	var client FetcherClient
-	for attemptsCount := uint64(1); blocksFetched <= lookback; {
+	for retryCount := uint64(1); blocksFetched <= lookback; {
 		if err := cs.ctx.Err(); err != nil {
 			return cs.stopOrAbort()
 		}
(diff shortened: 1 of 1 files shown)


catchup/catchpointService.go
```diff
@@ -453,7 +453,7 @@ func (cs *CatchpointCatchupService) processStageBlocksDownload() (err error) {
 	blocksFetched := uint64(1) // we already got the first block in the previous step.
 	var blk *bookkeeping.Block
 	var client FetcherClient
-	for attemptsCount := uint64(1); blocksFetched <= lookback; attemptsCount++ {
+	for attemptsCount := uint64(1); blocksFetched <= lookback; {
```
algonautshant, line 456: Since this is only counting the failed attempts, can you call it failedAttemptsCount or retryCount instead of attemptsCount?

algonautshant (approved):
Looks good.
Maybe change the variable name?

catchup/catchpointService.go
tsachiherman: Calling it "retryCount" makes sense.

algonautshant (approved):
Thanks for the update.

PR #1794, opened 2020-12-28
bugfix: Network peers list memory leak

tsachiherman:
## Summary

Removing the wsPeer object pointer from the peers list would allow it to be recycled.

The original code was creating a sub-slice and not removing the entry from the underlying array.
Since the peer was left on the underlying array, it would not get GC'ed.

## Test Plan

Existing tests confirms correct behavior.

### DIFF
--- a/network/peersheap.go
+++ b/network/peersheap.go
@@ -33,6 +33,7 @@ func (ph peersHeap) Pop() interface{} {
 	wn := ph.wn
 	end := len(wn.peers) - 1
 	p := wn.peers[end]
+	wn.peers[end] = nil // remove the entry from the peers list, so that the GC can recycle it's memory as needed.
 	wn.peers = wn.peers[:end]
 	return p
 }


brianolson (approved):
It feels like we found a bug in the GC. I get how it happens. I'm not sure exactly what the Go runtime GC contract spec is, but there's an argument that a pointer off the end of the length of an array isn't an *active* pointer and shouldn't prevent GC.

tsachiherman:
> It feels like we found a bug in the GC. I get how it happens. I'm not sure exactly what the Go runtime GC contract spec is, but there's an argument that a pointer off the end of the length of an array isn't an _active_ pointer and shouldn't prevent GC.

Arrays are immutable objects in Go. Just like strings. The array size can never be resized, only its element can get changed.  This does simplify the GC and move much of the length/cap to be handled by the slice - but it comes with the cost of having the enduser do some maintenance in case some GC'd items are stored. see https://github.com/golang/go/wiki/SliceTricks for the comment regarding the memory leak.

zeldovich (approved):
Nice find!

PR #1795, opened 2020-12-29
Add explicit error for stale network message writes

tsachiherman:
## Summary

This PR adds an explicit write error for disconnecting due to stale messages, which could now be distinguished from other write errors.

## Test Plan

Tested against telemetry server.


brianolson (approved):
I like it. Elegant little improvement.

PR #1799, opened 2020-12-30
Improve agreement cadaver & crash database error handling

tsachiherman:
## Summary

This PR contains few changes to improve error handling and file permissions -
1. cadaver file now being created with 0600 instead of 0666 and being modified to 0666 after creation. This allow us to make this modification symmetrically if the file is opened for appending, which ensures that when we attempt to rename the file to ".archive" the file always has the 0666 permission.
2. add handling for failure to close the cadaver file.
3. modify the reset() function so that it would log the error, in case it encountered any.
4. modify the restore() function to review the returned error code from agreeInstallDatabase and take advantage of the result.

### DIFF
--- a/agreement/cadaver.go
+++ b/agreement/cadaver.go
@@ -68,9 +68,14 @@ func (c *cadaver) filename() string {
 }
 
 func (c *cadaver) init() (err error) {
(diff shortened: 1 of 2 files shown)


nicholasguoalgorand (approved):
Looks good

PR #1800, opened 2020-12-30
Improve node error code checking

tsachiherman:
## Summary

This PR handles the following missing error code checking:
1. Add a test to verify the error in the creation of the genesis directory.
2. Add logging in case `node.loadParticipationKeys` returns an error
3. Add logging in case `os.Rename` has failed.
4. Remove `fmt.Println(msg)`, as the node reporting output is the log file, and not the console.

### DIFF
--- a/node/node.go
+++ b/node/node.go
@@ -182,11 +182,9 @@ func MakeFull(log logging.Logger, rootDir string, cfg config.Local, phonebookAdd
 
 	// create initial ledger, if it doesn't exist
 	err = os.Mkdir(genesisDir, 0700)
-	if err != nil {
-		if !os.IsExist(err) {
-			log.Errorf("Unable to create genesis directroy: %v", err)
-			return nil, err
-		}
+	if err != nil && !os.IsExist(err) {
+		log.Errorf("Unable to create genesis directroy: %v", err)
+		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


node/node.go
```diff
@@ -181,7 +181,13 @@ func MakeFull(log logging.Logger, rootDir string, cfg config.Local, phonebookAdd
 	ledgerPathnamePrefix := filepath.Join(genesisDir, config.LedgerFilenamePrefix)
 
 	// create initial ledger, if it doesn't exist
-	os.Mkdir(genesisDir, 0700)
+	err = os.Mkdir(genesisDir, 0700)
+	if err != nil {
```
algorandskiy, line 185: nit: `if err != nil && !os.IsExist(err)` instead of nested ifs

algorandskiy (commented):
LGTM

algorandskiy:
> Add a test to verify the error in the creation of the genesis directory.

No test

tsachiherman:
> > Add a test to verify the error in the creation of the genesis directory.
> 
> No test

I meant in-code test ( i.e. check ), not a unit test. I think that adding a unit test for this is a good idea.

algorandskiy (approved):

PR #1801, opened 2020-12-30
Fix return code testing & fix data directroy cloning

tsachiherman:
## Summary

This PR addresses several different issues:
1. The clone functionality wasn't working ( and according to the code, it looks as it wasn't working for over a year.. ); The changes here would make it work again. The might be additional caveats around the current functionality, but that could be left for another PR.
2. The cleanUpZombieKMD function now reports back an error, if there is a meaningful error to report ( i.e. other than "file not found" ).
3. The algod/kmd start function checks the return code of the Close() function and return the result to the caller. This doesn't really expected to fail, but in case it does, it will be reported upstream.

### DIFF
--- a/nodecontrol/algodControl.go
+++ b/nodecontrol/algodControl.go
@@ -191,7 +191,12 @@ func (nc *NodeController) StartAlgod(args AlgodStartArgs) (alreadyRunning bool, 
 		files := nc.setAlgodCmdLogFiles(algodCmd)
 		// Descriptors will get dup'd after exec, so OK to close when we return
 		for _, file := range files {
-			defer file.Close()
+			defer func(file *os.File) {
+				localError := file.Close()
+				if localError != nil {
+					err = localError
+				}
+			}(file)
 		}
 	}
 
@@ -283,8 +288,19 @@ func (nc NodeController) GetAlgodPath() string {
 
 // Clone creates a new DataDir based on the controller's DataDir; if copyLedger is true, we'll clone the ledger.sqlite file
 func (nc NodeController) Clone(targetDir string, copyLedger bool) (err error) {
-	os.RemoveAll(targetDir)
-	err = os.Mkdir(targetDir, 0700)
+	err = os.RemoveAll(targetDir)
+	if err != nil {
+		if !os.IsNotExist(err) {
+			return fmt.Errorf("unable to delete directory '%s' : %v", targetDir, err)
+		}
+	}
+	sourceFolderStat, statError := os.Stat(nc.algodDataDir)
+	if statError != nil {
+		err = statError
+		return
+	}
+
+	err = os.Mkdir(targetDir, sourceFolderStat.Mode())
 	if err != nil && !os.IsExist(err) {
 		return
 	}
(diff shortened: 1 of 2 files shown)


nodecontrol/algodControl.go
```diff
@@ -283,8 +288,18 @@ func (nc NodeController) GetAlgodPath() string {
...
 // Clone creates a new DataDir based on the controller's DataDir; if copyLedger is true, we'll clone the ledger.sqlite file
 func (nc NodeController) Clone(targetDir string, copyLedger bool) (err error) {
-	os.RemoveAll(targetDir)
-	err = os.Mkdir(targetDir, 0700)
+	err = os.RemoveAll(targetDir)
+	if err != nil && !os.IsNotExist(err) {
+		return fmt.Errorf("unable to delete directory '%s' : %v", targetDir, err)
+	}
+
+	sourceFolderStat, statError := os.Stat(nc.algodDataDir)
```
algonautshant, line 296: Why not use the err here instead of introducing  a new variable statError?

nodecontrol/algodControl.go
```diff
@@ -283,8 +288,18 @@ func (nc NodeController) GetAlgodPath() string {
...
+	if statError != nil {
+		err = statError
+		return
+	}
+
+	err = os.Mkdir(targetDir, sourceFolderStat.Mode())
```
algonautshant, line 302: Here is a good idea to capture the error into a new variable, and copy it to err only if the error is not IsExist.
Otherwise, os.IsExist error may be returned.

nodecontrol/algodControl.go
```diff
@@ -316,19 +331,25 @@ func (nc NodeController) Clone(targetDir string, copyLedger bool) (err error) {
 		}
 
 		genesisFolder := filepath.Join(nc.algodDataDir, genesis.ID())
+		genesisFolderStat, statError := os.Stat(genesisFolder)
```
algonautshant, line 334: Why not use the err here instead of introducing a new variable statError?

nodecontrol/algodControl.go
```diff
@@ -316,19 +331,25 @@ func (nc NodeController) Clone(targetDir string, copyLedger bool) (err error) {
...
+		if statError != nil {
+			err = statError
+			return
+		}
+
 		targetGenesisFolder := filepath.Join(targetDir, genesis.ID())
-		err = os.Mkdir(targetGenesisFolder, 0770)
-		if err != nil {
+		err = os.Mkdir(targetGenesisFolder, genesisFolderStat.Mode())
+		if err != nil && !os.IsExist(err) {
```
algonautshant, line 342: Same comment as above (use a different variable for the returned error).

algonautshant (approved):
Looks good. 
A few small comments.

nodecontrol/algodControl.go
tsachiherman: I find that the go language syntax of 
```
a, b := func()
```
is misleading. It's not clear if b was defined before, or is defined within this line.
I'll find another way to do that.

nodecontrol/algodControl.go
tsachiherman: good idea.

nodecontrol/algodControl.go
algonautshant: Oops. I missed that. It is fine the way it is.

algonautshant (approved):
Thanks for the updates.

Issue #1773, opened 2020-12-18
Inefficient serialization method 


aybehrouz:
## Summary
msgpack codec should be configured in `toarray` mode instead of `map` mode.
## Scope/Requirements
In a conversation that I had with @jannotti I noticed that the codec serializer is configured in its default mode which means it encodes structs as maps. Therefore every field of any struct will be serialized and stored with a tag name. For example, a `TealValue` struct:
```
type TealValue struct {
	_struct struct{} `codec:",omitempty,omitemptyarray"`

	Type  TealType `codec:"tt"`
	Bytes string   `codec:"tb"`
	Uint  uint64   `codec:"ui"`
}
```
is encoded in this way:
```
temp := basics.TealValue{
		Type:  basics.TealUintType,
		Bytes: "",
		Uint:  123456789,
	}
------------------------
serialization:
 ��tt�ui�I��  length:13
```
As u can see strings "tt" and "ui" which are tag names are encoded and they are a part of the serialization. 6 bytes are used for storing tag names while the uint itself is only 5 bytes! By using `codec:toarray` structs can be serialized without saving those tag names on the disk...

## Urgency/Relative Priority
...

jannotti:
The trouble with that approach is that the serialized forms become extremely fragile. If the struct is rearranged, or a field is added anywhere except the end, all existing serialized forms would break.

We do use short names because of this, but we're not quite willing to throw out that extra safety for the 2x or so or might save.

ian-algorand:
While we appreciate your contribution, we likely won't pick up this change in the foreseeable future. We might end up wrapping some of this into some bigger changes we have planned but it likely won't get done here. Closing this but as always, let us know if you have any additional thoughts or comments.

PR #1802, opened 2020-12-30
Add teal converter

aybehrouz:
## Summary
This PR is a follow up to my previous PR https://github.com/algorand/go-algorand/pull/1763. As I mentioned in that PR the current approach for storing state of smart contracts has several problems. On the other hand, changing the memory architecture of TEAL is not easy and it will take some time. While this development is in progress addition of new smart contracts can not be stopped. These new contracts will use current inefficient approach for storing their state and the data they save can not be changed and optimized later.

Therefore, I recommend changing TEAL to a newer version **as soon as possible**. The new version should use an efficient array-style memory model. Fortunately, **applying this change does not require any change in the code-base. All scripts written in new TEAL can be safely converted to the current version of TEAL.** Because The current version of TEAL uses a more general memory model the conversion can be done safely and algorithmically. 

For example, in the new version of TEAL we may want to change `app_global_get` opcode to something like this:
```
app_global_get
    Opcode: 0x64 {uint8 position in global static memory space to load from}
    Pops: None
    Pushes: any
    LogicSigVersion >= 3
    Mode: Application
```
This instruction can safely be converted to two TEALv2 instructions:
```
app_global_get 12 -----> byte "12"
                         app_global_get
```

After introduction of the new TEAL the installation of TEALv2 contracts should be disallowed. Later, when the support for new memory model is added to the code-base the stored state of contracts that are written in the new TEAL can be safely converted and optimized to new array-style memory model.

### DIFF
--- a/tools/teal/converter/converter.go
+++ b/tools/teal/converter/converter.go
@@ -1,0 +1,17 @@
+// Copyright (C) 2019-2020 Algorand, Inc.
+// This file is part of go-algorand
+//
+// go-algorand is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as
+// published by the Free Software Foundation, either version 3 of the
+// License, or (at your option) any later version.
+//
+// go-algorand is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with go-algorand.  If not, see <https://www.gnu.org/licenses/>.
+
+package converter


aybehrouz:
The converter is almost finished. After this, the support for new instructions needs to be added in the assembler.

algorandskiy:
> After introduction of the new TEAL the installation of TEALv2 contracts should be disallowed

This is not likely to happen.

What are the benefits of having `app_local_get const`? It disallows dynamic key names that is a limiting factor.

If you have concerns about space efficiency of key-value storage in the blockchain or about current apps limits then please consider targeting these areas (better binary representation, apps evaluation performance).

aybehrouz:
>What are the benefits of having app_local_get const?

I've explained this in my previous PR https://github.com/algorand/go-algorand/pull/1763. I'll try to summarize it here again. 

- With a key-value state storage efficient storage of program variables is not possible. In a conventional compiler, program variables are handled using symbol tables and symbol tables can be stored much more efficiently in an array-style memory. (you won't need to store the variable names)
- In the current TEAL, variable names become constants in the code and hence they will make the byte-code of the program longer. 

> It disallows dynamic key names that is a limiting factor.

A TEAL program natively has access to a dictionary with accountAddr or AppID as its keys. I don't think there is any urgent needs for having a native map data structure. Even if that's the case, I think maps should be added as a data type and its not wise to have an unconventional key-value memory style for Algorand virtual machine.

>If you have concerns about space efficiency of key-value storage in the blockchain or about current apps limits then please consider targeting these areas (better binary representation, apps evaluation performance).

I've addressed these issues too. I recommend taking a look at issues https://github.com/algorand/go-algorand/issues/1738 and https://github.com/algorand/go-algorand/issues/1773. Also in my other PR https://github.com/algorand/go-algorand/issues/1738 I've implemented a memory segment which is my proposal as a replacement for the current key-value storage. It has a custom serialization. If u store a small program's state using it and then compare its binary representation size with the size of state of that program in the current implementation of Algorand I guess u would see at least 3x-4x improvement.

PR #1804, opened 2020-12-31
bugfix: clear roundDigest when fast catchup is switching off

tsachiherman:
## Summary

The `roundDigest` variable in the `accountUpdates` structure was expected to be cleared out during `initializeFromDisk`, but it wasn't. As a result, a node that has caught up using the fast catchup would produce incorrect catchpoint labels.

## Test Plan

Add a unit test to verify the correctness of the fix. The unit test is far more comprehensive than required, and would detect other potential failure cases that would end up having a different catchpoint label.

### DIFF
--- a/ledger/acctupdates_test.go
+++ b/ledger/acctupdates_test.go
@@ -26,6 +26,7 @@ import (
 	"os"
 	"path/filepath"
 	"runtime"
+	"strings"
 	"sync"
 	"testing"
 	"time"
@@ -43,6 +44,7 @@ import (
 type mockLedgerForTracker struct {
 	dbs      dbPair
 	blocks   []blockEntry
+	deltas   []StateDelta
 	log      logging.Logger
 	filename string
 	inMemory bool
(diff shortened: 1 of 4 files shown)


ledger/acctupdates_test.go
```diff
@@ -57,6 +59,44 @@ func makeMockLedgerForTracker(t testing.TB, inMemory bool) *mockLedgerForTracker
 	return &mockLedgerForTracker{dbs: dbs, log: dblogger, filename: fileName, inMemory: inMemory}
 }
 
+// Fork creates another database which has the same content as the current one. Works only for non-memory databases.
+func (ml *mockLedgerForTracker) Fork(t testing.TB) *mockLedgerForTracker {
```
algorandskiy, line 63: nit: make non-exported

ledger/acctupdates_test.go
```diff
@@ -71,11 +111,20 @@ func (ml *mockLedgerForTracker) Latest() basics.Round {
...
+
 func (ml *mockLedgerForTracker) trackerEvalVerified(blk bookkeeping.Block, accUpdatesLedger ledgerForEvaluator) (StateDelta, error) {
-	delta := StateDelta{
-		hdr: &bookkeeping.BlockHeader{},
+	if len(ml.deltas) > int(blk.Round()) {
+		return ml.deltas[uint64(blk.Round())], nil
+	} else {
+		return StateDelta{
+			hdr: &bookkeeping.BlockHeader{},
+		}, nil
```
algorandskiy, line 126: this looks not obvious at all, could you please comment/explain/make a separate method?

ledger/acctupdates_test.go
```diff
@@ -1343,3 +1392,118 @@ func TestCompactDeltas(t *testing.T) {
...
+	accts[0][testPoolAddr] = pooldata
+
+	sinkdata := basics.AccountData{}
+	sinkdata.MicroAlgos.Raw = 1000 * 1000 * 1000 * 1000
+	sinkdata.Status = basics.NotParticipating
+	accts[0][testSinkAddr] = sinkdata
+
+	au := &accountUpdates{}
+	cfg := config.GetDefaultLocal()
+	cfg.CatchpointInterval = 500
```
algorandskiy, line 1422: looks like setting `CatchpointInterval=20` would speedup the test with the same level of reproducibility

ledger/acctupdates_test.go
```diff
@@ -1343,3 +1392,118 @@ func TestCompactDeltas(t *testing.T) {
...
+		if uint64(i)%cfg.CatchpointInterval == 0 {
+			au.waitAccountsWriting()
+			catchpointLabels[i] = au.GetLastCatchpointLabel()
+			ledgerHistory[i] = ml.Fork(t)
+			defer ledgerHistory[i].close()
+		}
+	}
+
+	// test in revese what happens when we try to repeat the exact same blocks.
+	for startingRound := basics.Round((testCatchpointLabelsCount - 1) * cfg.CatchpointInterval); startingRound > basics.Round(cfg.CatchpointInterval); startingRound -= basics.Round(cfg.CatchpointInterval) {
```
algorandskiy, line 1485: nit: startingRound -> sRound, then maybe `for` cond will be short enough to be seen without scrolling

ledger/acctupdates_test.go
```diff
@@ -1343,3 +1392,118 @@ func TestCompactDeltas(t *testing.T) {
...
+		ml.addMockBlock(blockEntry{block: blk}, delta)
+		accts = append(accts, totals)
+		rewardsLevels = append(rewardsLevels, rewardLevel)
+		roundDeltas[i] = delta
+
+		// if this is a catchpoint round, save the label.
+		if uint64(i)%cfg.CatchpointInterval == 0 {
+			au.waitAccountsWriting()
+			catchpointLabels[i] = au.GetLastCatchpointLabel()
+			ledgerHistory[i] = ml.Fork(t)
```
algorandskiy, line 1479: do we need to care about ledger files deletion?

tsachiherman: Files should be deleted by the close method. I'll double check that tomorrow.

ledger/acctupdates_test.go
tsachiherman: It needs to be greater than the lookback, which defaults to 320.
And yes, we can use a custom protocol here yo do that.

tsachiherman: done

ledger/acctupdates_test.go
tsachiherman: I moved the `startingRound` to it's own line before the for loop, so that it would now fit nicely in view.

algorandskiy (approved):

PR #1805, opened 2020-12-31
Add the ability to disable catchup in config file

tsachiherman:
## Summary

Add the ability to disable catchup in config file.

## Test Plan

Tested manually.

### DIFF
--- a/catchup/service.go
+++ b/catchup/service.go
@@ -465,6 +465,10 @@ func (s *Service) periodicSync() {
 				sleepDuration = s.deadlineTimeout
 				continue
 			}
+			// if the catchup is disabled in the config file, just skip it.
+			if s.parallelBlocks == 0 {
+				continue
+			}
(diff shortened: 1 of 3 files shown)


algonautshant (approved):
Looks good.

catchup/service.go
```diff
@@ -465,6 +465,10 @@ func (s *Service) periodicSync() {
 				sleepDuration = s.deadlineTimeout
 				continue
 			}
+			// if the catchup is disabled in the config file, just skip it.
```
algorandskiy, line 468: disabling it here still makes one catchup step via `s.sync()` above. Maybe exit earlier or not even run `periodicSync`?

tsachiherman: You're correct regarding the `s.sync()` above, but not running `periodicSync` completely is a problem since the agreement service depends on it ( i.e. it needs the `s.unmatchedPendingCertificates` case )

algonautshant (approved):

PR #1807, opened 2021-01-01
Improve network package error handling ( WIP )

tsachiherman:
## Summary

Make several small changes to the network package - 
Avoid using `crypto.NewHash`/`hasher.Write` since it has a returned error that need to be checked. Instead, pass a byte-array to crypto.Hash which would do the same thing.

## Testing

Unit test was added to confirm that the changed in the hashing result in the exact same hash as before.


network/messageFilter.go
```diff
@@ -48,12 +48,7 @@ func makeMessageFilter(bucketsCount, maxBucketSize int) *messageFilter {
 // CheckMessage checks if the given tag/msg already in the collection, and return true if it was there before the call.
 // Prepends our own random secret to the message to make it hard to abuse hash collisions.
 func (f *messageFilter) CheckIncomingMessage(tag protocol.Tag, msg []byte, add bool, promote bool) bool {
-	hasher := crypto.NewHash()
-	hasher.Write(f.nonce[:])
-	hasher.Write([]byte(tag))
-	hasher.Write(msg)
-	var digest crypto.Digest
-	hasher.Sum(digest[:0])
+	digest := crypto.Hash(append(append(f.nonce[:], []byte(tag)...), msg...))
```
brianolson, line 51: I think I deliberately laid it out the other way so as to not do memcpy() operations in append(). I'd be curious to see a benchmark of each coding style.

brianolson (commented):
Lots of good things around err checks and crypto.RandBytes() but I'm not sure about the hasher usage change

network/messageFilter.go
tsachiherman: I'll be splitting this PR into two: the one for the hashing and all the rest.

PR #1808, opened 2021-01-02
Update go files to have license range of 2021

tsachiherman:
## Summary

Update the license range from 2019-2020 to 2019-2021.

## Test Plan

No tests updates here, as this is only an in-code comment. The testing for this is done via the license check shell script.

### DIFF
--- a/README.md
+++ b/README.md
@@ -189,6 +189,6 @@ A number of packages provide utilities for the various components:
 
 Please see the [COPYING_FAQ](COPYING_FAQ) for details about how to apply our license.
 
-Copyright (C) 2019-2020, Algorand Inc.
+Copyright (C) 2019-2021, Algorand Inc.
 
 [developer site url]: https://developer.algorand.org/
--- a/agreement/abstractions.go
+++ b/agreement/abstractions.go
@@ -1,4 +1,4 @@
-// Copyright (C) 2019-2020 Algorand, Inc.
+// Copyright (C) 2019-2021 Algorand, Inc.
 // This file is part of go-algorand
 //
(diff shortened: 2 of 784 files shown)


pzbitskiy (changes requested):
I disagree. Let's update the checker script and allow 2019-202x

tsachiherman:
> I disagree. Let's update the checker script and allow 2019-202x

Both Adam and myself have looked into that in the past.
The goal here is to get the broadest legal protection while reducing the potential licensing ambiguity to the minimum.

I know that it looks like a tidious thing, but if you'll evaluate the actual cost, you'll find that it won't realy make a difference. 

Btw - I think that it's reasonable to convert a off-by-one year to be a warning rather then a build failuire.

brianolson:
Do we allow 20xx-2021 ? It seems wrong to me to claim 2019 copyright on a file that was created during 2020.

onetechnical (approved):
Looks simple enough. I wonder if we could automate this.

tsachiherman:
> Do we allow 20xx-2021 ? It seems wrong to me to claim 2019 copyright on a file that was created during 2020.

It would be reasonable to have it claim "2019" if one could say he was working on this "feature" for a very long time, right ?
( i.e. it doesn't have to be the "publishing date" )

It wouldn't be wrong to have some of the files start at 2021-2021, for instance. however, I think that there is real added value ( and please correct me if I'm wrong here ) in making it more flexible that we absolutely need to get legal coverage that we want.

The fact that the build "fail" might sounds like a bad thing, but it's more of a reminder to us, not a really a product failure. ( keep in mind that it would only fail on travis builds.. not locally )

PR #1811, opened 2021-01-05
Add BenchmarkWritingRandomBalancesDisk

tsachiherman:
## Summary

Add a benchmark to test the performance difference between writing to the accountbase table using the address as the primary key, versus using the rowid.

Benchmark results:

| Writing Method  | Database accounts number | Rate |
| ------------- | ------------- | ------------- |
| Using address  | 5,000,000 | 151,635 ns/acct_update |
| Using rowid  | 5,000,000 | 131,031 ns/acct_update |
| Using address  | 2,000,000 |  136,688 ns/acct_update |
| Using rowid  | 2,000,000 |  116,865 ns/acct_update |
| Using address  | 500,000 | 60,462 ns/acct_update |
| Using rowid  | 500,000 |  56,764 ns/acct_update |

Running parameters-
1. Insertions were made using prepared statements.
2. Random account data size on insertion was 200 bytes.

## Remarks
1. While the update speed itself seems to be variable with the amount of data, it seems that the delta between the two is (roughly) fixed - it's always around 5,000ns-25,000ns or ~15%.
2. Using rowid seems to be a clear winner here, in all cases.

## Test Plan

Benchmark is a test.

### DIFF
--- a/ledger/accountdb_test.go
+++ b/ledger/accountdb_test.go
@@ -36,6 +36,7 @@ import (
 	"github.com/algorand/go-algorand/data/basics"
 	"github.com/algorand/go-algorand/logging"
 	"github.com/algorand/go-algorand/protocol"
+	"github.com/algorand/go-algorand/util/db"
 )
 
 func randomAddress() basics.Address {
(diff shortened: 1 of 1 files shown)


brianolson:
I'm tempted to expand this branch as a test on account update based on other databases to compete with sqlite3

tsachiherman:
Based on your comment, I guess it's "the benchmark is ok to be merged in", right ?

tsachiherman:
Based on your comment, I guess it's "the benchmark is ok to be merged in", right ?

algorandskiy (approved):
The benchmark looks ok.
Proposal: shouldn't we have say a weekly job that runs all the benchmark and collects data into a dashboard?

tsachiherman:
I think that this proposal would make a good DevOp project.

PR #1816, opened 2021-01-08
Eval prefetch

brianolson:
## Summary

Improve speed of BlockEvaluator eval() by pre-fetching account data, decreasing latency at various points during block accounting.

## Test Plan

passes existing tests and improves benchmark:
```
branch eval_prefetch
BenchmarkBlockEvaluatorRAMCrypto-8   	      12	1372329400 ns/op	     27455 ns/eval_validate_tx

master
BenchmarkBlockEvaluatorRAMCrypto-8   	      12	1779370174 ns/op	     35603 ns/eval_validate_tx
```


tsachiherman (commented):
Given that this improves the performance, I think that it would mean we need to keep focusing on this one. There are few aspects of this implementation that are quite premature and would not get us the full potential gains -
1. Instead of calling `state.lookup`, prepare a map ( no duplicates ) of the entries you'd need to load.
2. Use the execution pool to process batches of accounts. If each account fetching takes 30,000ns, grouping these into a 1ms buckets would be a good choice. The transaction signature verification currently using that patten.
3. I'm not convinced that we would benefit for having the prefetching in a go-routine, as it's only forcing us to take additional locks. Maybe we can block the eval until we prefetch all the accounts ? This would also imply the during the eval there is no need to lock before getting the value ( since it's guaranteed to be preloaded )

brianolson:
I tried using a local map (no locking) of seen addresses so that state.lookup() would only be called once per address, it was slower. Seems to be more efficient to just use the map behind the lock to see if an addr has been looked up yet.

2 sounds like a two stage pipeline. A first stage for fetching accounts and a second stage for doing the eval(). These pipeline stages could run in parallel but with less locking.

3. I bet (fetch all) (eval) would be slower than interleaved fetch-eval-with-locks, but it's worth an experiment. Should be easy to try...

brianolson:
3. tested this morning, no locking prefetch-then-eval is faster than master but not faster than interleaved prefetch+eval

ledger/eval.go
```diff
@@ -963,6 +991,7 @@ func eval(ctx context.Context, l ledgerForEvaluator, blk bookkeeping.Block, vali
 	if err != nil {
 		return StateDelta{}, err
 	}
+	go prefetchThread(ctx, eval.state.lookupParent, blk.Payset)
```
tsachiherman, line 994: The `ctx` here might not be the best context to use. I think it would be better to copy the 
```
validationCtx, validationCancel := context.WithCancel(ctx)
defer validationCancel()
```
from few lines below and use it here. That way, whenever this function exits OR the calling context is canceled, the prefetchThread would abort.

ledger/eval.go
```diff
@@ -1035,6 +1064,38 @@ func eval(ctx context.Context, l ledgerForEvaluator, blk bookkeeping.Block, vali
...
+		maybelookup(stxn.Txn.AssetReceiver)
+		maybelookup(stxn.Txn.AssetCloseTo)
+		maybelookup(stxn.Txn.FreezeAccount)
+		for _, xa := range stxn.Txn.Accounts {
+			maybelookup(xa)
+		}
+		if bail == 0 {
+			bail = 10
+			select {
+			case <-ctx.Done():
```
tsachiherman, line 1089: have you observed that calling this `<-ctx.Done()` is making it slow ? It's kind of unexpected that we're doing that only every 10 transactions. ( where the timing isn't really clear )

tsachiherman:
Keep in mind that the sqlite supports multiple concurrent readers. The existing code doesn't take advantage of that.

ledger/eval.go
brianolson: No, just a guess. I probably pulled this pattern from code where just calling time() is enough of a slowdown that it shouldn't be done too often.

brianolson:
idea: parallel sqlite3 fetch
problem: roundCowBase.lookup() calls Ledger.LookupWithoutRewards() which has a lock in it, and that calls accountUpdates.LookupWithoutRewards() which also takes a lock.
Updating all this to make parallel reads work could be a separate ticket.

tsachiherman:
> idea: parallel sqlite3 fetch
> problem: roundCowBase.lookup() calls Ledger.LookupWithoutRewards() which has a lock in it, and that calls accountUpdates.LookupWithoutRewards() which also takes a lock.
> Updating all this to make parallel reads work could be a separate ticket.

It's a read lock, so it won't be crucial. However, I agree that if you'd like to improve it, you should prepare a "list" of addresses you'd like to fetch and get them all with a single lock invocation.

I would start by making sure you have few go-routines that makes the calls in parallel. Implementing `LookupMultiple` could be next, if you find that the locking performance is an issue.

brianolson:
I missed that they were read-locks, yes they should allow multiple read threads. But, Looks like there isn't actually parallelism available at the sqlite3 layer. Adding more read threads just added to the time in CPU profile spent waiting on sqlite3, and the whole system was overall slower. A `LookupMultiple()` API may yet be more efficient.

brianolson:
I also tried a pipelined version that released txgroups and their prefetched account records over a chan, no locks, but this was also not faster than the simpler method implemented in this branch.

tsachiherman (commented):
I think that I have a better solution to the deadlock issue -
instead of 
```
defer validationCancel()
```
use
```
var wg sync.WaitGroup
wg.Add(1)
defer func() {
     validationCancel()
     wg.Wait()
}()

...
...

go prefetchThread(..., wg)
```

and inside prefetchThread : 
```
func prefetchThread(..., wg) {
    defer wg.Done()
```

you could use a channel to achieve the same, but I think that WaitGroup is more elegant.

ledger/archival_test.go
```diff
@@ -678,11 +678,6 @@ func makeSignedTxnInBlock(tx transactions.Transaction) transactions.SignedTxnInB
 
 func TestArchivalFromNonArchival(t *testing.T) {
 	// Start in non-archival mode, add 2K blocks, restart in archival mode ensure only genesis block is there
-	deadlockDisable := deadlock.Opts.Disable
-	deadlock.Opts.Disable = true
-	defer func() {
-		deadlock.Opts.Disable = deadlockDisable
-	}()
```
algorandskiy, line 685: I guess it was disabled here to speed up the test

ledger/eval.go
```diff
@@ -1075,23 +1103,29 @@ func (validator *evalTxValidator) run() {
...
 	if err != nil {
 		return ledgercore.StateDelta{}, err
 	}
 
+	validationCtx, validationCancel := context.WithCancel(ctx)
+	defer validationCancel()
+
+	// If validationCtx or underlying ctx are Done, end prefetch
+	if usePrefetch {
+		go prefetchThread(validationCtx, eval.state.lookupParent, blk.Payset)
```
algorandskiy, line 1120: why `eval.state.lookupParent` and not `eval.state` ?
`lookupParent` is a ledger the goes directly into accoutdb without any intermediate caching. Or does this relies on `au.baseAccounts` cache?
If instead `eval.state` is used then all these accounts are 100% cached within `roundCowState.mods`.

algorandskiy (commented):
Please clarify `prefetchThread` comment below.

ledger/archival_test.go
tsachiherman: no - it definitely wasn't placed here for speeding tests.
It was placed here since it would prevent random deadlocks on this test on travis.

 If you recall, we had quite many of these.. and this addressed the issue. Changing a global variable can lead to a data race, like in this case. The textbook solution I've used is to ensure each test run to completion ( i.e. shutting down all it's child threads ) before exiting.

ledger/eval.go
algobolson: `eval.state.lookupParent` is a roundCowBase with caching
`eval.state` is just a roundCowState.
I figure I may as well go straight to the layer with the caching.

tsachiherman (approved):

algorandskiy (approved):

PR #1819, opened 2021-01-11
Fix deprication warning on sortition compilation

algorandskiy:
## Summary

Boost complains about non c++11 compiler, so make it happy.

## Test Plan

Use existing test suite


tsachiherman (approved):
looks good.

PR #1822, opened 2021-01-12
Move ledger.dbPair to db.Pair

tsachiherman:
## Summary

Attempting to restructure the ledger package into sub-packages shows that we would have many cycles.
In this PR I'm going to address the dbPair, which is shared among many of the files and would become an issue.
The dbPair structure is a very light-weight struct that doesn't have much to do with the ledger itself, and therefore a good candidate to be extracted outside.

## Test Plan


algorandskiy:
Please rebase after merging #1818 to simplify review

tsachiherman:
> Please rebase after merging #1818 to simplify review

Yes, that's why it's still a draft ;-)

algorandskiy (approved):
LGTM

PR #1823, opened 2021-01-12
Remove ApplyData from proposal when propagating

nicholasguoalgorand:
## Summary

Do not send ApplyData when sending transactions since they are to be recomputed anyway. This will save bandwidth as well as set up for future changes that uses txId's in blocks instead of full transactions

## Test Plan

Reran unit tests


agreement/actions.go
```diff
@@ -151,7 +151,7 @@ func (a networkAction) do(ctx context.Context, s *Service) {
 	case protocol.ProposalPayloadTag:
 		msg := a.CompoundMessage
 		payload := transmittedPayload{
-			unauthenticatedProposal: msg.Proposal,
+			unauthenticatedProposal: msg.Proposal.Compressed(),
```
tsachiherman, line 154: I think that you would want to have the digest in the vote include the "Compressed" proposal.

tsachiherman:
am I missing anything in https://github.com/algorand/go-algorand/compare/master...tsachiherman:tsachi/dropad?expand=1 ?

nicholasguoalgorand:
> am I missing anything in https://github.com/algorand/go-algorand/compare/master...tsachiherman:tsachi/dropad?expand=1 ?

Tried this, tests kept failing. Turns out this won't work since nodes that generate the block will skip calling the AddBlock and will jump straight to AddValidatedBlock

agreement/proposal.go
```diff
@@ -19,6 +19,7 @@ package agreement
 import (
 	"context"
 	"fmt"
+	"github.com/algorand/go-algorand/data/transactions"
```
tsachiherman, line 22: The agreement service should not be familiar with the concept of "transactions". Let upstream components to provide you with two versions of the block.

ledger/eval.go
```diff
@@ -768,19 +774,19 @@ func (eval *BlockEvaluator) transaction(txn transactions.SignedTxn, evalParams *
...
-			if !ad.Equal(applyData) {
-				return fmt.Errorf("transaction %v: applyData mismatch: %v != %v", txid, ad, applyData)
-			}
-		} else {
-			if !ad.Equal(transactions.ApplyData{}) {
-				return fmt.Errorf("transaction %v: applyData not supported", txid)
-			}
-		}
-	}
+	//// Validate applyData if we are validating an existing block.
```
tsachiherman, line 777: With your change, you've dropped the usage of the input parameter ad. If this is the case, remove it from the caller and clean up the code.

agreement/proposal.go
```diff
@@ -70,11 +69,17 @@ func (p unauthenticatedProposal) value() proposalValue {
 	return proposalValue{
 		OriginalPeriod:   p.OriginalPeriod,
 		OriginalProposer: p.OriginalProposer,
-		BlockDigest:      p.Digest(),
-		EncodingDigest:   crypto.HashObj(p),
+		BlockDigest:      p.Compressed().Digest(),
+		EncodingDigest:   crypto.HashObj(p.Compressed()),
```
tsachiherman, line 73: Avoid calling p.Compress() twice; it's not that cheap to iterate over all the transactions...

agreement/proposal.go
```diff
@@ -238,14 +243,18 @@ func proposalForBlock(address basics.Address, vrf *crypto.VRFSecrets, ve Validat
...
 	value := proposalValue{
 		OriginalPeriod:   period,
 		OriginalProposer: address,
-		BlockDigest:      proposal.Block.Digest(),
-		EncodingDigest:   crypto.HashObj(proposal),
+		BlockDigest:      compressedProposal.Block.Digest(),
+		EncodingDigest:   crypto.HashObj(compressedProposal),
 	}
-	return proposal, value, nil
+	return uncompressedProposal, value, nil
```
algorandskiy, line 257: this looks confusing - why do we include compressed hash but return non-compressed. Please comment out the code

data/bookkeeping/block.go
```diff
@@ -254,6 +254,20 @@ func (block *Block) Seed() committee.Seed {
 	return block.BlockHeader.Seed
 }
 
+// Compressed returns the Block without ApplyData
+func (block *Block) Compressed() Block {
+	c := *block
+	c.Payset = nil
+	for _, stb := range block.Payset {
```
algorandskiy, line 261: do we really need to copy payset?

ledger/eval.go
```diff
@@ -768,20 +763,6 @@ func (eval *BlockEvaluator) transaction(txn transactions.SignedTxn, evalParams *
 		return fmt.Errorf("transaction %v: %v", txid, err)
 	}
 
-	// Validate applyData if we are validating an existing block.
```
algorandskiy, line 771: if we do not check entire apply data, then we need to check maybe a hash of it? Can't find it.

data/bookkeeping/block.go
tsachiherman: We want to create two "versions" - one for sending on the network, and one that would be used locally if this proposal is accepted by the network.
The copy is "stripping" the Payset from the ApplyData.

ledger/eval.go
tsachiherman: yes, I believe you'll find it in finalValidation : we compute the commit hash for the block.
The flat commit hash is computed as the hash of the entire payset ( with apply data ).
so, by verifying the root hash, we can avoid comparing all the inner transactions.

algorandskiy: There is a code path to avoid this validation: `TransactionPool.addToPendingBlockEvaluatorOnce` calls `BlockEvaluator.TransactionGroup` directly so `finalValidation` is not called.

nicholasguoalgorand: `finalValidation` is called by `eval` which is called by `Validate`, am I missing something?

agreement/proposal.go
nicholasguoalgorand: thanks for suggestion, tried to make this cleaner + added comment

ledger/eval.go
nicholasguoalgorand: (all in `ledger/eval.go`)

tsachiherman: The codepath through `TransactionPool.addToPendingBlockEvaluatorOnce` ( which truly doesn't have the `finalValidation` ), is always providing dummy apply data anyway, so I don't think it's meaningful anyway.
( i.e. `addToPendingBlockEvaluatorOnce` converts a `[]transactions.SignedTxn` into `[]transactions.SignedTxnWithAD` with a simple copy.. )

algorandskiy: Then I'm good

ledger/eval.go
```diff
@@ -601,39 +601,34 @@ func (eval *BlockEvaluator) testTransaction(txn transactions.SignedTxn, cow *rou
...
 		}
 
 		// Initialize group without ApplyData lazily
 		if groupNoAD == nil {
 			groupNoAD = make([]transactions.SignedTxn, len(txgroup))
 			for j := range txgroup {
-				groupNoAD[j] = txgroup[j].SignedTxn
+				groupNoAD[j] = txgroup[j]
 			}
 			minTealVersion = logic.ComputeMinTealVersion(groupNoAD)
```
tsachiherman, line 633: I think that with your change, this entire section can now be dropped and we can use the txgroup directly in the preparation of the EvalParams below.

data/bookkeeping/block.go
tsachiherman: But you're correct that we should make sure we're doing that exactly one time.

ledger/eval.go
nicholasguoalgorand: must have missed this, thanks for pointing it out, fixing now

ledger/eval.go
```diff
@@ -624,19 +623,15 @@ func (eval *BlockEvaluator) prepareEvalParams(txgroup []transactions.SignedTxn)
...
 
-		// Initialize group without ApplyData lazily
-		if groupNoAD == nil {
-			groupNoAD = make([]transactions.SignedTxn, len(txgroup))
-			for j := range txgroup {
-				groupNoAD[j] = txgroup[j]
-			}
-			minTealVersion = logic.ComputeMinTealVersion(groupNoAD)
+		// Initialize minTealVersion lazily
+		if minTealVersion == 0 {
```
tsachiherman, line 627: I think that you meant `if i == 0`, right ?

nicholasguoalgorand: No because I'm trying to follow the original logic of computing this value lazily, ie only if we encounter a txn of type `protocol.ApplicationCallTx`. However, checking if the value is 0 might not be the right thing to do here, perhaps introduce another variable. Thoughts?

tsachiherman: yes, you're correct about this one, but comparing minTealVersion to 0 isn't the right solution either since a zero is a legit return code from logic.ComputeMinTealVersion.
given that the entire purpose of this loop ( now ) is to calculate the minTealVersion, I would suggest dropping the if statement completely and break right after calling ` logic.ComputeMinTealVersion`. Does that sounds like a good solution?

tsachiherman: I would suggest creating another loop before this one, to find calculate the value of the minTealVersion and break once it's calculated.

gmalouf:
@tsachiherman / @nicholasguoalgorand feel free to re-open when/if becomes relevant.

PR #1825, opened 2021-01-12
Restructure ledger package: ledger/ledgercore

tsachiherman:
## Summary

The ledger package become too big; this PR is doing the first step by creating a `ledger/ledgercore` and moving the following structures there:
- AccountDelta ( formally, `ledger.accountDelta` )
- ModifiedCreatable ( formally, `ledger.modifiedCreatable` )
- Txlease ( formally, `ledger.txlease` )
- StateDelta ( formally, `ledger.StateDelta` )
- AccountTotals ( formally, `ledger.AccountTotals` )
- TransactionInLedgerError ( formally, `ledger. TransactionInLedgerError` )
- LeaseInLedgerError ( formally, `ledger. LeaseInLedgerError` )
- ErrNoEntry ( formally, `ledger.ErrNoEntry ` )

Next step is to add the `ledgerForEvaluator` to the `ledgercore`, however, this depends on `VotersForRound`, which is still being modified by Nickolai's compactcert [PR](https://github.com/algorand/go-algorand/pull/1507).

## Test Plan

Update existing unit tests.


cmd/goal/node.go
```diff
@@ -132,15 +132,15 @@ var catchupCmd = &cobra.Command{
 func catchpointCmdArgument(cmd *cobra.Command, args []string) error {
 	catchpointsCount := 0
 	for _, arg := range args {
-		_, _, err := ledger.ParseCatchpointLabel(arg)
+		_, _, err := common.ParseCatchpointLabel(arg)
```
algorandskiy, line 135: I feel like `common` not a good name, it is too generic and it is hard to say what is it when first seeing line like this. Maybe `ledgerTypes` or `ledgerCommon`, or `lgdata`, `lgtypes`, `lgcom`, etc.
Especially we already have `daemon/algod/api/server/common` that is fine inside server but it leaked out to goal and libgoal and... not really common there.

ledger/common/catchpointlabel.go
```diff
@@ -14,7 +14,7 @@
 // You should have received a copy of the GNU Affero General Public License
```
algorandskiy, line 14: this should probably be catchpoint subpackage or even go to `catchup/catchpoint`

algorandskiy (commented):
Good work. I only think `common` name is too generic and does not make sense outside of ledger package. Please consider another name, I proposed some options below.

ledger/common/catchpointlabel.go
tsachiherman: I'm not a big fan of that dependency graph direction. what do you think about moving it to `data/basics` ?

cmd/goal/node.go
tsachiherman: I think that you have a good point; but this need to be further hashed.
Most of the structures/interfaces in /ledger/common are internal to `/ledger/...`; so I think that it makes sense to keep it as `/ledger/common`. however, for the very few interfaces/struct that need to be exported, we might want to place these into a separate, public, package; maybe `/ledger/ledgercore` ?

tsachiherman: note - the Golang internal packages aren't a good fit here since `/ledger/internal` isn't accessible by `/ledger/trackers`

algorandskiy (approved):

PR #1827, opened 2021-01-13
Remove check_golang_version check when deploying in travis

btoll:
## Summary

The call to `check_golang_version.sh` is causing the travis run to fail when deploying:
https://travis-ci.com/github/algorand/go-algorand/jobs/470973034#L142

I believe that removing this check is safe.  We're installing the version of go that may be needed immediately after this check using `gimme`, so any subsequent operations will be sure to succeed because it will be installing the build version (1.14.7).

Also, the `check_golang_version` script is intended as a safeguard for dev and building and I don't believe is needed in this deploy script.

## Test Plan

Hope and pray.

### DIFF
--- a/scripts/travis/deploy_packages.sh
+++ b/scripts/travis/deploy_packages.sh
@@ -12,10 +12,6 @@ set -e
 
 SCRIPTPATH="$( cd "$(dirname "$0")" ; pwd -P )"
 
-if ! "${SCRIPTPATH}/../check_golang_version.sh"
-then
-    exit 1
-fi
 # Get the go build version.
 GOLANG_VERSION=$("${SCRIPTPATH}/../get_golang_version.sh")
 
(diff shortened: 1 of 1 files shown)


scripts/travis/deploy_packages.sh
```diff
@@ -12,10 +12,6 @@ set -e
 
 SCRIPTPATH="$( cd "$(dirname "$0")" ; pwd -P )"
 
-if ! "${SCRIPTPATH}/../check_golang_version.sh"
```
onetechnical, line 15: Instead of deleting this, should we move it after the gimme, to verify that it installed successfully?

onetechnical (changes requested):
I would actually modify travis.yaml temporarily in this branch to run the deploy stages to verify in all environments. You could filter out the integration tests to verify, but it will still take a long time. I think this would _probably_ work but I also want to confirm that test_release.sh works as well.

We may as well disable the External ARM Deploy, as we don't use that for nightlies and I don't think it's completed in months; this would save an hour plus in travis testing validation.

scripts/travis/deploy_packages.sh
btoll: I'm not sure the script even needs to do the check.

btoll: Re-added it after the check.

bricerisingalgorand (approved):

onetechnical (approved):
I'd have liked to have had run through the travis Deploy jobs before approving, but this seems simple enough and it's blocking nightly builds.

PR #1828, opened 2021-01-14
New recipe for Devnet network.

egieseke:
## Summary

Reinitialized Devnet using algonet with a new Devnet recipe file.

## Test Plan

Used algonet and the new devnet recipe to create a new network.  Tested with a local node to ensure that it was possible to connect to the running network.


tsachiherman (commented):
What's about creating it with substentially longer part keys ?
I.e. mayby 30m rounds instead of 3m rounds

egieseke:
> What's about creating it with substentially longer part keys ?
> I.e. mayby 30m rounds instead of 3m rounds

test/testdata/deployednettemplates/recipes/devnet/gen_topology.py
```diff
@@ -0,0 +1,31 @@
...
+        node_name = node_type + str(i+1) + "-" + network
+        region = regions[i%region_count]
+        if (first ):
+            first = False
+        else:
+            f.write(",")
+        f.write ("\n    {\n      \"Name\": \"" + node_name + "\",\n      \"Template\": \"" + region + region_size + "\"\n    }"  )
+
+f.write("\n  ]\n}\n")
+f.close()
```
btoll, line 30: Use a context manager ("with ...") for this block.  It's safer to let python manage it for us.

test/testdata/deployednettemplates/recipes/devnet/gen_topology.py
```diff
@@ -0,0 +1,31 @@
...
+    "AWS-EU-NORTH-1",
+    "AWS_AP_SOUTHEAST-2"
+]
+network="DevNet"
+f = open("topology.json", "w")
+f.write("{ \"Hosts\":\n  [")
+
+region_count = len(regions)
+first = True
+for  x in node_types:
```
btoll, line 16: Spacing

btoll: And should this be `node_types.items()`?

test/testdata/deployednettemplates/recipes/devnet/gen_topology.py
```diff
@@ -0,0 +1,31 @@
...
+    node_count = node_types[x]
+    region_size = node_size[x]
+    for i in range(node_count):
+        node_name = node_type + str(i+1) + "-" + network
+        region = regions[i%region_count]
+        if (first ):
+            first = False
+        else:
+            f.write(",")
+        f.write ("\n    {\n      \"Name\": \"" + node_name + "\",\n      \"Template\": \"" + region + region_size + "\"\n    }"  )
```
btoll, line 27: I would look at the `json` stdlib module so you're not having to write json by hand.

test/testdata/deployednettemplates/recipes/devnet/gen_topology.py
```diff
@@ -0,0 +1,31 @@
...
+region_count = len(regions)
+first = True
+for  x in node_types:
+    node_type = x
+    node_count = node_types[x]
+    region_size = node_size[x]
+    for i in range(node_count):
+        node_name = node_type + str(i+1) + "-" + network
+        region = regions[i%region_count]
+        if (first ):
```
btoll, line 23:
```
if first:
```

btoll (changes requested):

tsachiherman:
> > What's about creating it with substentially longer part keys ?
> > I.e. mayby 30m rounds instead of 3m rounds
> 
> From an operational perspective, it is probably better to follow the guidance that we provide users of less than 3 million rounds and come up with a process to automatically monitor and then renew the participation keys so that we are familiar with the process. Otherwise, we will be caught off guard again in 30 million rounds.

I agree that it would better to come up with such a process. Specifying 3M rounds give you ~6 month. Using 30M gives you 5 years. Unless you have a concrete plan to complete this monitoring project within 6 month ( which is what we've had in our backlog for over a year ), it would be smarter choice to push it considerably further.

egieseke:
> > > What's about creating it with substentially longer part keys ?
> > > I.e. mayby 30m rounds instead of 3m rounds
> > 
> > 
> > From an operational perspective, it is probably better to follow the guidance that we provide users of less than 3 million rounds and come up with a process to automatically monitor and then renew the participation keys so that we are familiar with the process. Otherwise, we will be caught off guard again in 30 million rounds.
> 
> I agree that it would better to come up with such a process. Specifying 3M rounds give you ~6 month. Using 30M gives you 5 years. Unless you have a concrete plan to complete this monitoring project within 6 month ( which is what we've had in our backlog for over a year ), it would be smarter choice to push it considerably further.

Metrika agreed to add a monitor for devnet participation keys.

test/testdata/deployednettemplates/recipes/devnet/gen_topology.py
egieseke: Fixed spacing, added .keys() for clarity.

test/testdata/deployednettemplates/recipes/devnet/gen_topology.py
egieseke: removed

test/testdata/deployednettemplates/recipes/devnet/gen_topology.py
egieseke: Updated to use json.dumps().

test/testdata/deployednettemplates/recipes/devnet/gen_topology.py
egieseke: Updated to use with block.

btoll:
@egieseke Thanks for the changes.  This looks good to me, if @tsachiherman is satisfied I'm ready to approve.

installer/genesis/devnet/genesis.json
```diff
@@ -17,276 +17,249 @@
...
-      "addr": "F5FALDSUFYO5LQU4OQY2HPZTXSUPRHUAERZ7IDG5QCBC76AOHEPK2VUJ34",
-      "comment": "pp2",
-      "state": {
-        "algo": 50000000000000
-      }
     }
   ],
   "fees": "A7NMWS3NT3IUDMLVO26ULGXGIIOUQ3ND2TXSER6EBGRZNOBOUIQXHIBGDE",
-  "id": "v1.0",
+  "id": "v1",
```
onetechnical, line 261: Why did we change this from v1.0? I know it's arbitrary, but isn't this inconsistent with our other network genesis files?

installer/genesis/devnet/genesis.json
```diff
@@ -17,276 +17,249 @@
...
   "network": "devnet",
-  "proto": "https://github.com/algorand/spec/tree/a26ed78ed8f834e2b9ccb6eb7d3ee9f629a6e622",
-  "rwd": "7777777777777777777777777777777777777777777777777774MSJUVU",
-  "timestamp": 1560210489
+  "proto": "https://github.com/algorandfoundation/specs/tree/3a83c4c743f8b17adfd73944b4319c25722a6782",
```
onetechnical, line 263: Also curious about starting this from most recent protocol, rather than first protocol. Though I suppose this doesn't matter because the mainnet network also did not start with the first.

installer/genesis/devnet/genesis.json
onetechnical: Actually, won't this also make all the indexes different for other queries? Was this intentional to split the networks?

test/testdata/deployednettemplates/recipes/devnet/genesis.json
```diff
@@ -0,0 +1,113 @@
+{
+	"NetworkName": "",
+	"VersionModifier": "",
+	"ConsensusProtocol": "",
+	"FirstPartKeyRound": 0,
+	"LastPartKeyRound": 3000000,
```
onetechnical, line 6: I kind of agree with Tsachi; I think having a higher count (e.g. 10,000,000) would be nice here. We can always rotate keys early for practice, and it gives us a little more time. Not required for this, though.

onetechnical (commented):
Curious about the id change in genesis.json mostly; want to discuss before approving.

installer/genesis/devnet/genesis.json
egieseke: Tsachi asked to confirm that it is the latest, which it is.

test/testdata/deployednettemplates/recipes/devnet/genesis.json
egieseke: OK, adjusting to 10,000,000.

installer/genesis/devnet/genesis.json
tsachiherman: yes, it will make the gossip port ( 4160 ) queries different. I think that it won't affect the indexer ( if there is such ).

installer/genesis/devnet/genesis.json
tsachiherman: it is ConsensusV24, which is currently the current default one:
```
const ConsensusCurrentVersion = ConsensusV24
```

installer/genesis/devnet/genesis.json
egieseke: Updated id to "v1.0".

test/testdata/deployednettemplates/recipes/devnet/genesis.json
egieseke: Restarted network with the id of v1.0 and participation key expiration set to 10,000,000 rounds.

onetechnical (approved):

bricerisingalgorand:
https://github.com/algorand/go-algorand/blob/master/gen/devnet.json

I think this needs to be updated too right?
Otherwise looks great to me! We should make sure local builds also end up with the correct genesis file in this branch.

bricerisingalgorand:
It looks like that thing is part of a make target that makes a random devnet like genesis file every time it's ran. I'm down to merge unless anyone else is worried about it

bricerisingalgorand (approved):

PR #1830, opened 2021-01-14
Increase pending proposals queue

tsachiherman:
## Summary

The number of proposals generated derives from the consensus params. At this time, we have (up to) 20 proposals.
In a rare network event, when all the relays connected to a given proposer are lagging behind, the winning proposal could be dropped out, causing a 21 second round.

This was verified by lowering the proposalBufferSize dramatically, confirming that the only dropped agreement messages were proposals, and that a 21 second rounds were introduced.

## Test Plan

Tested by lowering the proposalBufferSize to 2, and running a deployed network. Once every few *hours*, we would have a 21-second round.

### DIFF
--- a/agreement/gossip/network.go
+++ b/agreement/gossip/network.go
@@ -32,7 +32,7 @@ import (
 
 var (
 	voteBufferSize     = 10000
-	proposalBufferSize = 5
+	proposalBufferSize = 2
 	bundleBufferSize   = 7
 )
 
--- a/agreement/gossip/network.go
+++ b/agreement/gossip/network.go
@@ -32,12 +32,22 @@ import (
 
 var (
 	voteBufferSize     = 10000
-	proposalBufferSize = 14
+	proposalBufferSize = 5
 	bundleBufferSize   = 7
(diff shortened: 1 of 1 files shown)


onetechnical:
Thanks for testing this! I'm psyched you could recreate it!

egieseke (approved):
Looks good.  Are there logs for the dequeued messages?

tsachiherman:
> Looks good. Are there logs for the dequeued messages?

No. For performance & security reasons, there aren't any logs for an individual dropped message. ( since we don't want an attacker to force us writing to disk.. )

onetechnical (approved):

pbennett:
You should probably at least track metrics for it so it could be tracked w/ prometheus or other telemetry options.

tsachiherman:
> You should probably at least track metrics for it so it could be tracked w/ prometheus or other telemetry options.

Yes, that's how I confirmed that this change was contributed to this issue. What I wasn't able to confirm was that there aren't other causes for this issue.

PR #1838, opened 2021-01-16
Switch from map of modified accounts to an array

algorandskiy:
## Summary

* map[Address]AccountData replaced by AccountDeltas type
* AccountDeltas maintains a plain array for accessing elements
  and a map for quick lookup when needed
* map[Address]accountDeltaCount replaced by compactAccountDeltas type
  similarly to AccountDeltas

## Test Plan

Added new tests for AccountDeltas and compactAccountDeltas, updated existing tests


ledger/ledgercore/statedelta.go
```diff
@@ -69,3 +69,80 @@ type StateDelta struct {
...
+// AccountsDelta stores orderd accounts and allow fast lookup by address
+type AccountsDelta struct {
+	accts      []basics.BalanceRecord
+	acctsCache map[basics.Address]int
+}
+
+// MakeStateDelta creates a new instance of StateDelta
+func MakeStateDelta(hdr *bookkeeping.BlockHeader, prevTimestamp int64) StateDelta {
+	return StateDelta{
+		Accts:         AccountsDelta{},
```
tsachiherman, line 82: allocation the `ad.acctsCache` here would save you from checking it on every `addAccount`.

tsachiherman: maybe I'm being too greedy here, but initializing the above slice/acctsCache with some size could avoid reallocations down the road.

ledger/txtail_test.go
```diff
@@ -58,7 +58,7 @@ func TestTxTailCheckdup(t *testing.T) {
 		txleases := make(map[ledgercore.Txlease]basics.Round)
 		txleases[ledgercore.Txlease{Sender: basics.Address(crypto.Hash([]byte{byte(rnd % 256), byte(rnd / 256), byte(2)})), Lease: crypto.Hash([]byte{byte(rnd % 256), byte(rnd / 256), byte(3)})}] = rnd + leasevalidity
 
-		tail.newBlock(blk, ledgercore.StateDelta{Accts: make(map[basics.Address]basics.AccountData), Hdr: &blk.BlockHeader, Txids: txids, Txleases: txleases})
+		tail.newBlock(blk, ledgercore.StateDelta{Accts: ledgercore.AccountsDelta{}, Hdr: &blk.BlockHeader, Txids: txids, Txleases: txleases})
```
tsachiherman, line 61: I'd follow your own pattern:  use MakeStateDelta(), and the follow up by updating the rest of the fields.

ledger/ledgercore/statedelta.go
```diff
@@ -69,3 +69,80 @@ type StateDelta struct {
...
+}
+
+// Len returns number of stored accounts
+func (ad *AccountsDelta) Len() int {
+	return len(ad.accts)
+}
+
+// GetByIdx returns address and AccountData
+// It does NOT check boundaries.
+func (ad *AccountsDelta) GetByIdx(i int) (basics.Address, basics.AccountData) {
```
tsachiherman, line 123: nit: it looks that the two locations that do need this function would benefit more from having a direct access to the underlying slice.

tsachiherman (commented):
I think that it looks very promising.

ledger/ledgercore/statedelta.go
pzbitskiy: This is done on purpose to allow using non-initialized `AccountsDelta` in acctupdates. I think I'll run benchmarks to double check both options when complete.

ledger/ledgercore/statedelta.go
pzbitskiy: you mean returning pointers or `GetAddrIdx`/`GetDataIdx` methods?

tsachiherman: Return the inner slice instead of tunnelling each iteration through a getter.

ledger/ledgercore/statedelta.go
tsachiherman: Try to see if this nil checking can be avoided by design.

ledger/txtail_test.go
algorandskiy: done

ledger/acctupdates.go
```diff
@@ -139,7 +139,8 @@ type accountUpdates struct {
 	dbRound basics.Round
 
 	// deltas stores updates for every round after dbRound.
-	deltas []map[basics.Address]basics.AccountData
+	// deltas []map[basics.Address]basics.AccountData
```
tsachiherman, line 142: nit: remove this commented out line.

ledger/acctupdates.go
```diff
@@ -1350,14 +1352,15 @@ func (au *accountUpdates) deleteStoredCatchpoints(ctx context.Context, dbQueries
 }
 
 // accountsUpdateBalances applies the given deltas map to the merkle trie
-func (au *accountUpdates) accountsUpdateBalances(accountsDeltas map[basics.Address]accountDeltaCount) (err error) {
+func (au *accountUpdates) accountsUpdateBalances(accountsDeltas accountDeltasWithCount) (err error) {
```
tsachiherman, line 1355: nit: update the comment:
```
// accountsUpdateBalances applies the given account changes to the merkle trie
```

ledger/cow.go
```diff
@@ -68,15 +68,8 @@ func makeRoundCowState(b roundCowParent, hdr bookkeeping.BlockHeader, prevTimest
...
-		mods: ledgercore.StateDelta{
-			Accts:         make(map[basics.Address]basics.AccountData),
-			Txids:         make(map[transactions.Txid]basics.Round),
-			Txleases:      make(map[ledgercore.Txlease]basics.Round),
-			Creatables:    make(map[basics.CreatableIndex]ledgercore.ModifiedCreatable),
-			Hdr:           &hdr,
-			PrevTimestamp: prevTimestamp,
-		},
-		sdeltas: make(map[basics.Address]map[storagePtr]*storageDelta),
+		mods:         ledgercore.MakeStateDelta(&hdr, prevTimestamp),
```
tsachiherman, line 71: In startEvaluator, we have the paysetHint; we should use it to preinitialize the state delta allocation.
The number of transactions and the deltas are typically growing in a linear fashion, so preallocating twice as much in the array initializer would be a good idea.

tsachiherman (changes requested):
I think that you could avoid maintaining the `accountDeltasWithCount.addresses` array. The information there is used only in `getByIdx`, where you already have that in `deltas.old.addr`.

ledger/acctupdates.go
```diff
@@ -2000,22 +2005,23 @@ func (au *accountUpdates) commitRound(offset uint64, dbRound basics.Round, lookb
...
+func compactDeltas(accountDeltas []ledgercore.AccountDeltas, creatableDeltas []map[basics.CreatableIndex]ledgercore.ModifiedCreatable, baseAccounts lruAccounts) (outAccountDeltas accountDeltasWithCount, unavailableBaseAccounts []basics.Address, outCreatableDeltas map[basics.CreatableIndex]ledgercore.ModifiedCreatable) {
 	if len(accountDeltas) > 0 {
-		// the sizes of the maps here aren't super accurate, but would hopefully be a rough estimate for a resonable starting point.
-		outAccountDeltas = make(map[basics.Address]accountDeltaCount, 1+len(accountDeltas[0])*len(accountDeltas))
-		unavailableBaseAccounts = make([]basics.Address, 0, 1+len(accountDeltas[0])*len(accountDeltas))
+		// the sizes of the maps here aren't super accurate, but would hopefully be a rough estimate for a reasonable starting point.
+		size := accountDeltas[0].Len()*len(accountDeltas) + 1
+		unavailableBaseAccounts = make([]basics.Address, 0, size)
```
tsachiherman, line 2012: this would be a good place to preallocate the arrays in outAccountDeltas.

ledger/cow.go
```diff
@@ -207,22 +200,13 @@ func (cb *roundCowState) child() *roundCowState {
...
-		mods: ledgercore.StateDelta{
-			Accts:         make(map[basics.Address]basics.AccountData),
-			Txids:         make(map[transactions.Txid]basics.Round),
-			Txleases:      make(map[ledgercore.Txlease]basics.Round),
-			Creatables:    make(map[basics.CreatableIndex]ledgercore.ModifiedCreatable),
-			Hdr:           cb.mods.Hdr,
-			PrevTimestamp: cb.mods.PrevTimestamp,
-		},
-		sdeltas: make(map[basics.Address]map[storagePtr]*storageDelta),
+		mods:         ledgercore.MakeStateDelta(cb.mods.Hdr, cb.mods.PrevTimestamp, 1),
```
tsachiherman, line 203: I think we can do better in this case, but we can defer that for later.

tsachiherman:
Agree regarding the accountDeltasWithCount  -> compactAccountDeltas change;
I will look into the performance issue you've noted tomorrow.
Looks great.

algorandskiy:
Renamed. I do not have performance issues, just telling some optimization I thought make sense were not really measurable due to run-to-run fluctuations.

ledger/accountdb.go
```diff
@@ -192,6 +203,156 @@ func prepareNormalizedBalances(bals []encodedBalanceRecord, proto config.Consens
...
+			}
+		}
+	}
+	return
+}
+
+// accountsLoadOld updates the entries on the deltas.old map that matches the provided addresses.
+// The round number of the persistedAccountData is not updated by this function, and the caller is responsible
+// for populating this field.
+func (a *compactAccountDeltas) accountsLoadOld(tx *sql.Tx) (err error) {
```
tsachiherman, line 250: how's about updating the `compactAccountDeltas.misses` so that it would be set to nil once you exit from this function ? ( i.e. since that array is no longer needed )

ledger/accountdb.go
```diff
@@ -192,6 +203,156 @@ func prepareNormalizedBalances(bals []encodedBalanceRecord, proto config.Consens
...
+		switch err {
+		case nil:
+			if len(acctDataBuf) > 0 {
+				persistedAcctData := &persistedAccountData{addr: addr, rowid: rowid.Int64}
+				err = protocol.Decode(acctDataBuf, &persistedAcctData.accountData)
+				if err != nil {
+					return err
+				}
+				a.updateOld(idx, *persistedAcctData)
+				// a.upsertOld(*persistedAcctData)
```
tsachiherman, line 273: omit commented out lines.

ledger/accountdb.go
algorandskiy: done

algorandskiy:
Updated per review.

tsachiherman (approved):
looks good.

tsachiherman (approved):

algorandskiy:
Rebased

tsachiherman (approved):

PR #1839, opened 2021-01-18
Nitpicky, codebase refactoring and simplication

alexprut:
## Summary

Pretty trivial and nitpicky.

* Remove redundant code

## Test Plan

* Build
* Lint

### DIFF
--- a/gen/generate.go
+++ b/gen/generate.go
@@ -85,7 +85,7 @@ func GenerateGenesisFiles(genesisData GenesisData, consensus config.ConsensusPro
 	// Backwards compatibility with older genesis files: if the consensus
 	// protocol version is not specified, default to V0.
 	proto := genesisData.ConsensusProtocol
-	if proto == ("") {
+	if proto == "" {
 		proto = protocol.ConsensusCurrentVersion
 	}
 
--- a/gen/generate.go
+++ b/gen/generate.go
@@ -85,7 +85,7 @@ func GenerateGenesisFiles(genesisData GenesisData, consensus config.ConsensusPro
 	// Backwards compatibility with older genesis files: if the consensus
 	// protocol version is not specified, default to V0.
 	proto := genesisData.ConsensusProtocol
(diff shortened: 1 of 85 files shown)


gen/generate.go
```diff
@@ -85,7 +85,7 @@ func GenerateGenesisFiles(genesisData GenesisData, consensus config.ConsensusPro
 	// Backwards compatibility with older genesis files: if the consensus
 	// protocol version is not specified, default to V0.
 	proto := genesisData.ConsensusProtocol
-	if proto == protocol.ConsensusVersion("") {
+	if proto == ("") {
```
tsachiherman, line 88: why `("")` and not `""` ?

tsachiherman (commented):
The changes in this PR looks like they were created by an automated tool.
Could you share which tool you've used ?

I generally find it useless to apply these kind of changes unless we can also detect these kind of issues as part of the check-in process.

gen/generate.go
alexprut: Good catch, fixed

alexprut:
@tsachiherman most of the changes were highlighted / signaled by the GoLand IDE, while reading / learning the codebase, the changes were made manually.

> I generally find it useless to apply these kind of changes unless we can also detect these kind of issues as part of the check-in process.

I tend to agree, please feel free to close the PR if you find it useless. I have not look at the entire codebase yet, where can I take a look at the linter and configure it?

tsachiherman:
I don't have GoLand to test with; could you try and figure out how the GoLand is generating these ? ( I know that Visual Studio Code is using gofmt to achieve that ).

Once we integrate that linter into the travis build script, we can include the changes you've provided here.

tsachiherman:
I'm closing it since as is, we won't be able to maintain these changes. Thanks for the contribution. 
If you decide that you'd like to add an automated linter and make these changes, I'd be happy to review these changes over.

PR #1840, opened 2021-01-18
implement a transaction load generator

tsachiherman:
## Summary

In order to implement the "max attained throughput" on the mainnet metrics dashboard we need to generate at least a single block that would contain as many transactions as we can. The utility implemented here could assist us doing so.

## Test Plan

Tested manually.
Config file should be provided based on the example config file.
A mnemonic can be generated using 
```
algokey generate | grep "Private key mnemonic:"
```

### DIFF
--- a/cmd/loadgenerator/config.go
+++ b/cmd/loadgenerator/config.go
@@ -1,0 +1,55 @@
+// Copyright (C) 2019-2021 Algorand, Inc.
+// This file is part of go-algorand
+//
+// go-algorand is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as
+// published by the Free Software Foundation, either version 3 of the
+// License, or (at your option) any later version.
+//
+// go-algorand is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with go-algorand.  If not, see <https://www.gnu.org/licenses/>.
+
+package main
+
+import (
+	"encoding/json"
+	"net/url"
+	"os"
+)
+
+const configFileName = "loadgenerator.config"
+
+type config struct {
+	AccountMnemonic string
+	ClientURL       *url.URL `json:"-"`
+	APIToken        string
+	RoundModulator  uint64
+	RoundOffset     uint64
+	Fee             uint64
+}
+
+type fileConfig struct {
+	config
+	ClientURL string `json:"ClientURL"`
+}
+
+func loadConfig() (cfg config, err error) {
+	var fd *os.File
+	fd, err = os.Open(configFileName)
+	if err != nil {
+		return config{}, err
+	}
+	jsonDecoder := json.NewDecoder(fd)
+	var fileCfg fileConfig
+	err = jsonDecoder.Decode(&fileCfg)
+	if err == nil {
+		cfg = fileCfg.config
+		cfg.ClientURL, err = url.Parse(fileCfg.ClientURL)
+	}
+	return
+}
--- a/cmd/loadgenerator/config.go
+++ b/cmd/loadgenerator/config.go
@@ -25,12 +25,18 @@ import (
 const configFileName = "loadgenerator.config"
 
 type config struct {
+	// AccountMnemonic is the mnemonic of the account from which we would like to spend Algos.
 	AccountMnemonic string
-	ClientURL       *url.URL `json:"-"`
-	APIToken        string
-	RoundModulator  uint64
-	RoundOffset     uint64
-	Fee             uint64
+	// ClientURL is the url ( such as http://127.0.0.1:8080 ) that would be used to communicate with a node REST endpoint
+	ClientURL *url.URL `json:"-"`
+	// APIToken is the API token used to communicate with the node.
+	APIToken string
+	// RoundModulator is the modulator used to determine of the current round is the round at which transactions need to be sent.
+	RoundModulator uint64
+	// RoundOffset is the offset used to determine of the current round is the round at which transactions need to be sent.
+	RoundOffset uint64
+	// Fee is the amount of algos that would be specified in the transaction fee field.
+	Fee uint64
 }
 
 type fileConfig struct {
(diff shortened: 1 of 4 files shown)


cmd/loadgenerator/main.go
```diff
@@ -0,0 +1,222 @@
...
+		fmt.Fprintf(os.Stderr, "unable to load config : %v\n", err)
+		os.Exit(1)
+	}
+	fmt.Printf("Configuration file loaded successfully.\n")
+
+	seed := loadMnemonic(cfg.AccountMnemonic)
+	privateKey := crypto.GenerateSignatureSecrets(seed)
+	publicKey := basics.Address(privateKey.SignatureVerifier)
+
+	fmt.Printf("Spending account public key : %v\n", publicKey.String())
```
algonautshant, line 63: What is the purpose of printing this out?

cmd/loadgenerator/main.go
```diff
@@ -0,0 +1,222 @@
...
+			continue
+		}
+		if isSpendRound(cfg, nodeStatus.LastRound) {
+			// time to send transactions.
+			return
+		}
+		fmt.Printf("Current round %d, waiting for spending round %d\n", nodeStatus.LastRound, nextSpendRound(cfg, nodeStatus.LastRound))
+	waitForNextBlock:
+		// wait for the next round.
+		nodeStatus, err = restClient.WaitForBlock(basics.Round(nodeStatus.LastRound))
```
algonautshant, line 142: What am I missing here? It this waiting for the current round? Isn't this trivial?
Did you mean waiting for nextRpendRound?

cmd/loadgenerator/main.go
```diff
@@ -0,0 +1,222 @@
...
+			fmt.Fprintf(os.Stderr, "unable to check status : %v", err)
+			time.Sleep(1 * time.Second)
+			continue
+		}
+		if isSpendRound(cfg, nodeStatus.LastRound) {
+			// time to send transactions.
+			return
+		}
+		fmt.Printf("Current round %d, waiting for spending round %d\n", nodeStatus.LastRound, nextSpendRound(cfg, nodeStatus.LastRound))
+	waitForNextBlock:
```
algonautshant, line 140: Why the repetition inside the labeled block?
Why not just wait for nextPsendRound and continue here?

cmd/loadgenerator/main.go
```diff
@@ -0,0 +1,222 @@
...
+	for {
+		waitForSpendingRound(restClient, cfg)
+		queueFull := generateTransactions(restClient, cfg, privateKey, publicKey)
+		if queueFull {
+			waitForNonSpendingRound(restClient, cfg)
+		}
+	}
+}
+
+func waitForNonSpendingRound(restClient client.RestClient, cfg config) {
```
algonautshant, line 97: No need for this function. Just use the  waitForSpendingRound with a parameter, and check 
if isPendingRound == spendingOrNonSpending.

algonautshant (commented):
Looks good. Some comments and questions. 
One more function to review.

cmd/loadgenerator/main.go
tsachiherman: When starting off this utility, the enduser needs to know which account he/she need to fund with money.. so, printing this out might be a good idea.

cmd/loadgenerator/main.go
```diff
@@ -0,0 +1,222 @@
...
+				Amount:   basics.MicroAlgos{Raw: 0},
+			},
+			Type: protocol.PaymentTx,
+		}
+		crypto.RandBytes(tx.Note[:])
+		txns[i] = tx.Sign(privateKey)
+	}
+
+	// create multiple go-routines to send all these requests.
+	const nroutines = 8
```
algonautshant, line 196: Maybe this should be parametrized either in the config or as a constant at the top?

cmd/loadgenerator/main.go
tsachiherman: nope; this one was abit tricky for me too. If you'll look on the server side implementation of WaitForBlock you'll see that it's waiting for the round+1 to get committed. 
while this isn't trivial, changing this behavior at the source would cause more failures than we want to have; for sure not something to be done in the context of this solution.

algonautshant (approved):

cmd/loadgenerator/main.go
tsachiherman: it would repeat the "Status" call:
1. The first time. ( so that we would have a solid knowledge of what is the current round )
2. Every time we had a communication error.

Otherwise, it would perform the WaitForBlock, blocking for exactly one round at a time.

cmd/loadgenerator/config.go
```diff
@@ -0,0 +1,55 @@
...
+)
+
+const configFileName = "loadgenerator.config"
+
+type config struct {
+	AccountMnemonic string
+	ClientURL       *url.URL `json:"-"`
+	APIToken        string
+	RoundModulator  uint64
+	RoundOffset     uint64
```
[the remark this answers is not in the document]
tsachiherman: yes, I can do that.

cmd/loadgenerator/main.go
tsachiherman: I'll make it equal to twice the number of cpus.

cmd/loadgenerator/main.go
algonautshant: Then why not wait for next spending round - 1?

cmd/loadgenerator/main.go
```diff
@@ -0,0 +1,222 @@
...
+		nodeStatus, err = restClient.Status()
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "unable to check status : %v", err)
+			time.Sleep(1 * time.Second)
+			continue
+		}
+		if !isSpendRound(cfg, nodeStatus.LastRound) {
+			return
+		}
+	waitForNextBlock:
```
nicholasguoalgorand, line 110: can we replace this goto block with another for loop, or are using goto's generally okay?

cmd/loadgenerator/main.go
nicholasguoalgorand, line 140: ditto on the goto

cmd/loadgenerator/main.go
tsachiherman: I'll refactor the function.

nicholasguoalgorand (commented):
Looks good, had one comment

cmd/loadgenerator/main.go
tsachiherman: Waiting for "next spending round - 1" can take a long time to reach. The server is limiting the request to 1 minute anyway. In general, I don't like to leave the server holding a connection for a long time, so keeping it for a single round seems fair.

algonautshant (approved):
Looks good.
Thanks for the changes.

PR #1841, opened 2021-01-18
Fix insufficient fee error message formatting

tsachiherman:
## Summary

This PR modifies the error message generated when the amount of fee is below the `MinTxnFee` from
```
transaction had fee {100}, which is less than the minimum 1000
```
to
```
transaction had fee 100, which is less than the minimum 1000
```


algonautshant (approved):
Looks good.

PR #1842, opened 2021-01-19
Improvements/bugfixes for tealdbg

jasonpaulos:
## Summary

Some transactions are not able to be fully expanded and viewed in tealdbg when using the Chrome DevTools frontend. This specifically affects transactions with empty `Accounts` or `ApplicationArgs` fields.

This bug happened because empty fields were being represented as nil slices in tealdbg, which translates to a `null` JSON value instead of an empty array, causing the Chrome frontend to fail to display any part of the expanded object. I've fixed this and made the following additional fixes in this PR:
* Made the object overflow field for Chrome more useful.
* Fixed an issue where the web frontend fails because the stack is empty.
* Fixed an issue where the web frontend fails to decode JSON with application local state because of integer map keys.

<details>
<summary>Current Chrome behavior (note the expanded arrow at gtxn[0] but nothing is shown)</summary>

</details>

<details>
<summary>Fixed Chrome behavior</summary>

</details>

## Test Plan

No tests added because these are frontend changes.

### DIFF
--- a/cmd/tealdbg/cdtState.go
+++ b/cmd/tealdbg/cdtState.go
@@ -429,6 +429,7 @@ func prepareArray(array []basics.TealValue) []fieldDesc {
 }
 
 func makePreview(fields []fieldDesc) (prop []cdt.RuntimePropertyPreview) {
+	prop = make([]cdt.RuntimePropertyPreview, 0, len(fields))
 	for _, field := range fields {
 		v := cdt.RuntimePropertyPreview{
 			Name:  field.Name,
@@ -441,6 +442,7 @@ func makePreview(fields []fieldDesc) (prop []cdt.RuntimePropertyPreview) {
 }
 
 func makeIntPreview(n int) (prop []cdt.RuntimePropertyPreview) {
+	prop = make([]cdt.RuntimePropertyPreview, 0, n)
 	for i := 0; i < n; i++ {
 		v := cdt.RuntimePropertyPreview{
 			Name:  strconv.Itoa(i),
@@ -480,7 +482,8 @@ func makeArrayPreview(array []basics.TealValue) cdt.RuntimeObjectPreview {
 	fields := prepareArray(array)
 
 	length := len(fields)
-	if length > maxArrayPreviewLength {
+	overflow := length > maxArrayPreviewLength
+	if overflow {
 		length = maxArrayPreviewLength
 	}
 	prop := makePreview(fields[:length])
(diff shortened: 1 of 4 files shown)


algorandskiy (dismissed):

cmd/tealdbg/cdtState.go
```diff
@@ -441,6 +442,7 @@ func makePreview(fields []fieldDesc) (prop []cdt.RuntimePropertyPreview) {
 }
 
 func makeIntPreview(n int) (prop []cdt.RuntimePropertyPreview) {
+	prop = make([]cdt.RuntimePropertyPreview, 0, n)
```
tsachiherman, line 445: I like that you preallocated the array.. unfortunately, if n is negative, the function would now panic.

cmd/tealdbg/cdtState.go
```diff
@@ -694,13 +695,15 @@ func makeLocalScope(s *cdtState, preview bool) (desc []cdt.RuntimePropertyDescri
 
 func makeGlobals(s *cdtState, preview bool) (desc []cdt.RuntimePropertyDescriptor) {
 	fields := prepareGlobals(s.globals)
+	desc = make([]cdt.RuntimePropertyDescriptor, 0, len(fields))
 	for _, field := range fields {
 		desc = append(desc, makePrimitive(field))
 	}
```
tsachiherman, line 701: you can get extra bonus points by writing it as 
```
desc = make([]cdt.RuntimePropertyDescriptor, len(fields), len(fields))
for i, field := range fields {
  		desc[i] = makePrimitive(field)
  	}
```

cmd/tealdbg/cdtState.go
```diff
@@ -694,13 +695,15 @@ func makeLocalScope(s *cdtState, preview bool) (desc []cdt.RuntimePropertyDescri
...
 	return
 }
 
 func makeTxn(s *cdtState, preview bool) (desc []cdt.RuntimePropertyDescriptor) {
+	desc = make([]cdt.RuntimePropertyDescriptor, 0)
```
tsachiherman, line 706: could you move the `desc` after the `if` statement ?

tsachiherman (changes requested):
one small issue in `makeIntPreview`; beside that there are few more bonus points that you can get all around.

tsachiherman (changes requested):
one small issue in `makeIntPreview`; beside that there are few more bonus points that you can get all around.

tsachiherman (approved):

tsachiherman (approved):

PR #1844, opened 2021-01-19
diagcfg : Read from global logging config if data dir not provided

bricerisingalgorand:
## Summary

When a datadir is not provided, this will have diagcfg read from the default logging config in the global config path. It will then generate a default config if that file does not exist

## Test Plan

I ran this locally and was able to produce correctly configured default logging configs

### DIFF
--- a/logging/telemetry.go
+++ b/logging/telemetry.go
@@ -97,7 +97,7 @@ func ReadTelemetryConfigOrDefault(dataDir *string, genesisID string) (cfg Teleme
 		configPath := filepath.Join(*dataDir, TelemetryConfigFilename)
 		cfg, err = LoadTelemetryConfig(configPath)
 	}
-	if err != nil && os.IsNotExist(err) || !dataDirProvided {
+	if (err != nil && os.IsNotExist(err)) || !dataDirProvided {
 		var configPath string
 		configPath, err = config.GetConfigFilePath(TelemetryConfigFilename)
 		if err != nil {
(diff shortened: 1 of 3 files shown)


logging/telemetry.go
```diff
@@ -92,11 +92,12 @@ func makeTelemetryState(cfg TelemetryConfig, hookFactory hookFactory) (*telemetr
...
 func ReadTelemetryConfigOrDefault(dataDir *string, genesisID string) (cfg TelemetryConfig, err error) {
 	err = nil
-	if dataDir != nil && *dataDir != "" {
+	dataDirProvided := dataDir != nil && *dataDir != ""
+	if dataDirProvided {
 		configPath := filepath.Join(*dataDir, TelemetryConfigFilename)
 		cfg, err = LoadTelemetryConfig(configPath)
 	}
-	if err != nil && os.IsNotExist(err) {
+	if err != nil && os.IsNotExist(err) || !dataDirProvided {
```
tsachiherman, line 100: The change is reasonable, however, I'd like to ask for few more small changes:
* change the function parameter from `*string` to `string`. Change these on the caller as well. ( there is no concrete current use case where we attempt to pass in nil )
* in the `if err != nil && os.IsNotExist(err) || !dataDirProvided` statement, make sure to use round brackets if there is more than a single logical expression. i.e. `if (a && b) || c
* Implement a unit test for this function. Unit test should fail prior to your change and pass with your change.

tsachiherman (changes requested):
Change looks good, few small requests.

logging/telemetry_test.go
```diff
@@ -18,6 +18,8 @@ package logging
 
 import (
 	"fmt"
+	"github.com/algorand/go-algorand/config"
```
tsachiherman, line 21: nit: coding style - the import list should be grouped as follows:
```
import (
 	"fmt"
 	"os"
 	"testing"
 	"time"

	"github.com/sirupsen/logrus"

	"github.com/algorand/go-deadlock"

	"github.com/algorand/go-algorand/logging/telemetryspec"
)
```

tsachiherman (approved):

logging/telemetry_test.go
bricerisingalgorand: Thanks! I just fixed that

tsachiherman (approved):
looks good.

PR #1845, opened 2021-01-19
Fix regression to loadTelemetryConfig introduced in 2.3.0

tsachiherman:
## Summary

As part of a refactoring effort in https://github.com/algorand/go-algorand/pull/1709, a small bug was introduced that changed the intended behavior of `loadTelemetryConfig`. This PR restores the previous behavior and adds a unit test for that.

## Test Plan

Unit test added.

### DIFF
--- a/logging/telemetryConfig.go
+++ b/logging/telemetryConfig.go
@@ -133,8 +133,9 @@ func loadTelemetryConfig(path string) (TelemetryConfig, error) {
 		return createTelemetryConfig(), err
 	}
 	defer f.Close()
-	cfg := createTelemetryConfig()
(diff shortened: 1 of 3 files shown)


onetechnical:
This sounds right to me. Were you able to verify the negative case in testing?

tsachiherman:
> This sounds right to me. Were you able to verify the negative case in testing?

You mean to verify that without the fix to `loadTelemetryConfig` the unit test consistently fail ? yes, of course. I also went back to 97673c3f ( the commit prior to  #1709 ), and confirmed that the unit test passes there.

algonautshant (approved):
Looks good.

PR #1848, opened 2021-01-20
testing: make e2e_client_runner more robust

tsachiherman:
## Summary

This PR improves the existing e2e client runner in two different ways:
1. It extends the time the kmd would remain in memory from 200 seconds to 3600 seconds.
    Given that some of our tests have a timeout of 300 seconds, it only make sense to align it with *at least* the duration of the longest test. Since we don't don't when each of the tests start running, we need to ensure it would be sufficiently large.
2. The existing wallet initialization bootstrapping code was polling for up to 5 seconds for the funds to be moved to the wallet. The changes here would ensure that:
    - The funds were moved successfully.
    - The operation is completed within a known number of rounds.
    - Instead of polling, it would wait for a new round. That sounds reduce the cpu load when running many e2e tests concurrently.
     

## Test Plan

See the above summary for details.

### DIFF
--- a/test/scripts/e2e_client_runner.py
+++ b/test/scripts/e2e_client_runner.py
@@ -174,7 +174,7 @@ class RunSet:
         if self.algod and self.kmd:
             return
         # should run from inside self.lock
-        xrun(['goal', 'kmd', 'start', '-t', '200'], env=self.env, timeout=5)
+        xrun(['goal', 'kmd', 'start', '-t', '3600'], env=self.env, timeout=5)
         algodata = self.env['ALGORAND_DATA']
         self.kmd = openkmd(algodata)
         self.algod = openalgod(algodata)
@@ -279,6 +279,7 @@ def goal_network_stop(netdir, normal_cleanup=False):
     logger.info('stop network in %s', netdir)
     try:
         xrun(['goal', 'network', 'stop', '-r', netdir], timeout=10)
+        xrun(['goal', 'kmd', 'stop'], env=self.env, timeout=5)
     except Exception as e:
         logger.error('error stopping network', exc_info=True)
         if normal_cleanup:
--- a/test/scripts/e2e_client_runner.py
+++ b/test/scripts/e2e_client_runner.py
@@ -79,15 +79,21 @@ def _script_thread_inner(runset, scriptname):
     # send one million Algos to the test wallet's account
     params = algod.suggested_params()
     round = params['lastRound']
-    txn = algosdk.transaction.PaymentTxn(sender=maxpubaddr, fee=params['minFee'], first=round, last=round+100, gh=params['genesishashb64'], receiver=addr, amt=1000000000000, flat_fee=True)
+    max_init_wait_rounds = 5
+    txn = algosdk.transaction.PaymentTxn(sender=maxpubaddr, fee=params['minFee'], first=round, last=round+max_init_wait_rounds, gh=params['genesishashb64'], receiver=addr, amt=1000000000000, flat_fee=True)
     stxn = kmd.sign_transaction(pubw, '', txn)
     txid = algod.send_transaction(stxn)
     ptxinfo = None
-    for i in range(50):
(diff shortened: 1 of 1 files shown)


test/scripts/e2e_client_runner.py
```diff
@@ -79,15 +79,21 @@ def _script_thread_inner(runset, scriptname):
...
     txid = algod.send_transaction(stxn)
     ptxinfo = None
-    for i in range(50):
+    for i in range(max_init_wait_rounds):
         txinfo = algod.pending_transaction_info(txid)
         if txinfo.get('round'):
             break
-        time.sleep(0.1)
+        status = algod.status_after_block(round_num=round)
+        round = status['lastRound']
```
algorandskiy, line 92: I think sleep is for good here, some delay between requests is better than nothing

test/scripts/e2e_client_runner.py
```diff
@@ -174,8 +180,8 @@ def _connect(self):
         if self.algod and self.kmd:
             return
         # should run from inside self.lock
-        xrun(['goal', 'kmd', 'start', '-t', '200'], env=self.env, timeout=5)
         algodata = self.env['ALGORAND_DATA']
+        xrun(['goal', 'kmd', 'start', '-t', '3600','-d', algodata], env=self.env, timeout=5)
```
algorandskiy, line 184: looks to high

test/scripts/e2e_client_runner.py
```diff
@@ -271,14 +277,16 @@ def wait(self, timeout):
...
+def goal_network_stop(netdir, env, normal_cleanup=False):
     global already_stopped, already_deleted
     if already_stopped or already_deleted:
         return
 
     logger.info('stop network in %s', netdir)
     try:
         xrun(['goal', 'network', 'stop', '-r', netdir], timeout=10)
+        algodata = env['ALGORAND_DATA']
+        xrun(['goal', 'kmd', 'stop', '-d', algodata], timeout=5)
```
algorandskiy, line 289: move into a separate try/catch block to prevent kmd keep running if network stopping falied.

test/scripts/e2e_client_runner.py
```diff
@@ -381,7 +389,7 @@ def main():
     retcode = 0
     xrun(['goal', 'network', 'create', '-r', netdir, '-n', 'tbd', '-t', os.path.join(gopath, 'src/github.com/algorand/go-algorand/test/testdata/nettemplates/TwoNodes50EachFuture.json')], timeout=90)
     xrun(['goal', 'network', 'start', '-r', netdir], timeout=90)
-    atexit.register(goal_network_stop, netdir)
+    atexit.register(goal_network_stop, env, netdir)
```
algorandskiy, line 392: wrong args order: netdir, env

test/scripts/e2e_client_runner.py
tsachiherman: The status_after_block is a blocking function, that would return when the target round+1 is reached.
I'm not sure why some delay would be better; could you expand on this one ?
(i.e. this change reduces the number of requests from ~50 to 4. sleeping won't help here )

test/scripts/e2e_client_runner.py
tsachiherman: yes, that's why I added the kmd shutdown.

test/scripts/e2e_client_runner.py
tsachiherman: done.

test/scripts/e2e_client_runner.py
tsachiherman: done.

test/scripts/e2e_client_runner.py
brianolson: the status_after_block api waiting for a block is better than sleep. good change.

algorandskiy: did not get it is blocking call

algorandskiy (approved):

PR #1849, opened 2021-01-20
non-utf8 app value keys for test

brianolson:
## Summary

Replace some app value keys "foo" with a non-utf8 string to test binary cleanness.

## Test Plan

This is a test. It should pass.


tsachiherman (approved):
This changes is good, but we also need a local storage setter as it affect different map.

PR #1850, opened 2021-01-20
Implicit support for Ubuntu and Arch Linux

HLFH:
## Summary

Algorand does not support by default Arch Linux, a GNU/Linux distribution often used in server environment for its KISS capabilities.
This PR brings the support for Arch Linux. How? When it is a GNU/Linux distribution, the script `distribtype.sh` specifies the distribution and allows the `configure_dev.sh` script to run flawlessly on Ubuntu and ArchLinux, and install the required dependencies.

## Test plan
It works. Which testing command lines do you advise?

### DIFF
--- a/scripts/configure_dev.sh
+++ b/scripts/configure_dev.sh
@@ -69,8 +69,10 @@ function install_windows_shellcheck() {
 }
 
 if [ "${OS}" = "linux" ]; then
-    DISTRIB=$("$SCRIPTPATH"/distribtype.sh)
-    if ! which sudo > /dev/null; then
+    . /etc/os-release
+    DISTRIB=$ID
+
+    if ! which sudo > /dev/null; then    
         if [ "${DISTRIB}" = "ubuntu" ]; then
             apt-get update
             apt-get -y install sudo
--- a/scripts/configure_dev.sh
+++ b/scripts/configure_dev.sh
@@ -69,13 +69,24 @@ function install_windows_shellcheck() {
 }
 
 if [ "${OS}" = "linux" ]; then
+    DISTRIB=$("$SCRIPTPATH"/distribtype.sh)
     if ! which sudo > /dev/null; then
-        apt-get update
-        apt-get -y install sudo
+        if [ "${DISTRIB}" = "ubuntu" ]; then
+            apt-get update
+            apt-get -y install sudo
+        elif [ "${DISTRIB}" = "arch" ]; then
+            pacman -Syy 
+            pacman -S sudo
+        fi
     fi
-
-    sudo apt-get update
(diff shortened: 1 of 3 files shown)


scripts/configure_dev.sh
```diff
@@ -69,13 +69,24 @@ function install_windows_shellcheck() {
 }
 
 if [ "${OS}" = "linux" ]; then
+    DISTRIB=$("$SCRIPTPATH"/distribtype.sh)
```
tsachiherman, line 72: if this is the *only* place where you need to call `os-release`, than it would make sense to have it called explicitly here instead of creating a separate file.

scripts/configure_dev.sh
```diff
@@ -69,13 +69,24 @@ function install_windows_shellcheck() {
...
     if ! which sudo > /dev/null; then
-        apt-get update
-        apt-get -y install sudo
+        if [ "${DISTRIB}" = "ubuntu" ]; then
```
tsachiherman, line 74: with this change, any unspecified distribution would stop working. Just to ensure backward compatibility, could you default to ubuntu ? i.e.
```
if [ "${DISTRIB}" = "arch" ]; then
   ...
else
   # ubuntu ( or compatible) distribution
   ...
fi
```

tsachiherman:
make sure to sign on the CLA; the travis tests should verify that there are no compilation issues on the various supported platforms.

scripts/configure_dev.sh
tsachiherman: nice to have: Adding a warning for a non-ubuntu distributions in this case might also be a good idea.

tsachiherman:
@HLFH, I'd like to move this issue forward so that we can either merge it or close it.

If you can complete the requested changes, I will try to get this approved & merged. Otherwise, I'd like to close this issue.

scripts/configure_dev.sh
HLFH: Ok, I am removing this file.

scripts/configure_dev.sh
HLFH: I have defaulted to Ubuntu in the last commit.

scripts/configure_dev.sh
```diff
@@ -69,13 +69,26 @@ function install_windows_shellcheck() {
...
+            pacman -Syy
+            pacman -S sudo
+        else
+            apt-get update
+            apt-get -y install sudo
+        fi
+    fi
+    
+    if [ "${DISTRIB}" = "arch" ]; then
+        pacman -Syy
```
onetechnical, line 86: Shouldn't these two commands be sudo'ed?

tsachiherman:
I'm closing this PR since there is no activity on this ticket, as well as that a separate ticket (https://github.com/algorand/go-algorand/pull/2073) was already addressing the issue.

PR #1853, opened 2021-01-21
add a non-utf8 Note

brianolson:
## Summary

To better test the binary-clean-ness of the tools at every stage, add a non-utf8 Note field

## Test Plan

This is a test. It should pass. Downstream tests (Indxer) will also benefit.


tsachiherman (changes requested):
Could you please add a comment explaining what is the purpose of this ?
( i.e. so that future code reviewer going over this test would understand why it's important to have this )

tsachiherman (approved):

PR #1854, opened 2021-01-22
Adjust timeouts in assets-app.sh/sectok-app.sh tests

algorandskiy:
## Summary

* goal waits for txn submitted at round x till x+2
  if submitting at the beginning of the round it is full 3 * 4.5 seconds per txn
* assets-app.sh has 23 txns that is 23 * 3 * 4.5 = 311 seconds
* sectok-app.sh has 27 txns that is 27 * 3 * 4.5 = 365 seconds

## Test Plan

This is fix for tests

### DIFF
--- a/test/scripts/e2e_subs/assets-app.sh
+++ b/test/scripts/e2e_subs/assets-app.sh
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
-# TIMEOUT=300
+# TIMEOUT=320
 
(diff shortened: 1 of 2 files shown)


tsachiherman (approved):
given that these increments are quite small, I tend to approve. However, I think that there are two issues that needs to be addresses:
1. see if the goal is really waiting the correct number of rounds.
2. consider breaking these into smaller tests that can be executed concurrently.

PR #1855, opened 2021-01-22
tests for the compactcert package

zeldovich:
This PR adds unit tests to the compactcert package, as suggested by Pavel.

As part of introducing these unit tests, this PR refactors the worker threads to be more predictable when the test harness is quickly advancing rounds in the simulated ledger.

### DIFF
--- a/compactcert/abstractions.go
+++ b/compactcert/abstractions.go
@@ -1,0 +1,58 @@
+// Copyright (C) 2019-2021 Algorand, Inc.
+// This file is part of go-algorand
+//
+// go-algorand is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as
+// published by the Free Software Foundation, either version 3 of the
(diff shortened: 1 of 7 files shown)


tsachiherman:
The positive testing over this package looks pretty good. I wouldn't mind having some better coverage on the expected fail cases ( which aren't covered by the unit tests ) : 
```
go test -coverprofile=coverage.out
go tool cover -html=coverage.out
```

algorandskiy (commented):
+1. This is a good change, thank you Nickolai.
I only a bit concern about `initBuilders` function, in tests it seems to produce empty `roundSigs` so the init loop is not really exercised.

zeldovich:
This `go tool cover` is pretty neat; thanks for the pointer!  I had not seen it before.  I pushed a few additional tests to cover some of the larger chunks of code that weren't being tested before.

algorandskiy (approved):

PR #1856, opened 2021-01-22
Change block header to allow for multiple compact certs in the future

zeldovich:
This is a somewhat superficial change to how compact cert state appears in
the block header.  Instead of having the three fields (CompactCertVoters,
CompactCertVotersTotal, and CompactCertNextRound) live directly in the
block header, they now live in a map indexed by the type of compact cert.
There's just one type for now, but this change will enable us, in the
future, to cleanly support multiple types of compact certs.  (There might
be a type of compact cert for zero-knowledge proofs, or another type of
compact cert for PQ security.)


go.mod
```diff
@@ -6,7 +6,7 @@ require (
 	github.com/algorand/go-codec/codec v0.0.0-20190507210007-269d70b6135d
 	github.com/algorand/go-deadlock v0.2.1
 	github.com/algorand/graphtrace v0.0.0-20201117160756-e524ed1a6f64
-	github.com/algorand/msgp v1.1.45
+	github.com/algorand/msgp v1.1.46
```
tsachiherman, line 9: I'm not sure why we have this here, since it's already identical to the master.

zeldovich: I added it here before you introduced a separate PR for it.  Will rebase against master, which should cause this diff to go away.

tsachiherman (commented):
I'm probably missing something, but where does the blockheader.CompactCert field is initialized ? how would it work with older blocks that doesn't have that map ?
i.e. it's ok to have the map empty.. but the usage of nil maps here seems questionable :
```
if sfa.Round < latestHdr.CompactCert[protocol.CompactCertBasic].CompactCertNextRound {
...
}
```

zeldovich:
The `CompactCert` field is initialized in `ledger/eval.go` in `BlockEvaluator.endOfBlock()`, if compact certs are enabled.

If compact certs are not enabled, this field is `nil`, which is equivalent to an empty map for the purposes of lookups (index expressions), so lookups will return the zero value.

algorandskiy (approved):

PR #1865, opened 2021-01-25
implement peer selector logic for catchup

tsachiherman:
## Summary

In order to support pseudo archival relays, we need to provide two-tier approach to the selection of peers in the catchup logic. This PR adds the `peerSelector` struct, which encapsulates this logic.

It also adds an important performance feedback loop - the result of the request is being fed back into the subsequent peer selection. A faster peer would be be favorable over a slower peer. A peer that failed providing a past block, would drop to the bottom of the list. 

## Test Plan

Unit test added, existing tests updated.

peerSelector.go test coverage 100% : 
```
go test .  -coverprofile=count.out; go tool cover -html=count.out
```


catchup/peerSelector.go
```diff
@@ -0,0 +1,287 @@
...
+// peerAddress returns the peer's underlying address. The network.Peer object cannot be compared
+// to itself, since the network package dynamically creating a new instance on every network.GetPeers() call.
+// The method retrun the peer address or an empty string if the peer is not one of HTTPPeer/UnicastPeer
+func peerAddress(peer network.Peer) string {
+	if httpPeer, ok := peer.(network.HTTPPeer); ok {
+		return httpPeer.GetAddress()
+	} else if unicastPeer, ok := peer.(network.UnicastPeer); ok {
+		return unicastPeer.GetAddress()
+	}
+	return ""
```
algorandskiy, line 206: what else can be in a peers list?

catchup/peerSelector.go
```diff
@@ -0,0 +1,287 @@
...
+}
+
+// refreshAvailablePeers reload the available peers from the network package, add new peers along with their
+// corresponding initial rank, and deletes peers that have been dropped by the network package.
+func (ps *peerSelector) refreshAvailablePeers() {
+	evalPeers := make(map[string]network.Peer)
```
algorandskiy, line 212: `existingPeers` looks a better name

catchup/peerSelector.go
```diff
@@ -0,0 +1,287 @@
...
+				delete(evalPeers, peerAddress)
+				continue
+			}
+			// it's an entry which we did not had before.
+			sortNeeded = ps.addToPool(peer, initClass.initialRank, initClass) || sortNeeded
+		}
+	}
+
+	// delete the "old" entries.
+	for poolIdx := len(ps.pools) - 1; poolIdx >= 0; poolIdx-- {
```
algorandskiy, line 238: please explain "old" - the peers that do not exist in network anymore

algorandskiy (commented):
Looks good, few minor questions/comments below.

catchup/catchpointService.go
```diff
@@ -320,20 +334,37 @@ func (cs *CatchpointCatchupService) processStageLastestBlockDownload() (err erro
...
-			fetcher := fetcherFactory.New()
-			blk, _, client, err = fetcher.FetchBlock(cs.ctx, blockRound)
+			peer, err = cs.blocksDownloadPeerSelector.GetNextPeer()
+			if err != nil {
+				err = fmt.Errorf("processStageLastestBlockDownload: unable to obtain a list of peers to retrieve the latest block from")
+				return cs.abort(err)
+			}
+			httpPeer, validPeer := peer.(network.HTTPPeer)
+			if !validPeer {
+				cs.log.Warnf("processStageLastestBlockDownload: non-HTTP peer was provided by the peer selector")
```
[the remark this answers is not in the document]
tsachiherman: The `GetNextPeer` could return a non-HTTPPeer if the peerClass was configured with PeersConnectedOut or PeersConnectedIn. While this is not the case for the catchpointService, I think that it would be better to verify that.
( we will need this later on when integrating with the classic catchup )

catchup/peerSelector.go
tsachiherman: There *shouldn't* be anything else. however, since the Peer is opaque interface, we need to verify. The robustness of this function allows us to avoid checking that in other places.

catchup/catchpointService.go
```diff
@@ -472,9 +509,27 @@ func (cs *CatchpointCatchupService) processStageBlocksDownload() (err error) {
...
+				if retryCount <= uint64(cs.config.CatchupBlockDownloadRetryAttempts) {
+					// try again.
+					retryCount++
+					continue
+				}
+				return cs.abort(fmt.Errorf("processStageBlocksDownload: recurring non-HTTP peer was provided by the peer selector"))
+			}
+			fetcher := makeHTTPFetcher(cs.log, httpPeer, cs.net, &cs.config)
+			blockDownloadStartTime := time.Now()
+			blk, _, err = fetcher.FetchBlock(cs.ctx, topBlock.Round()-basics.Round(blocksFetched))
```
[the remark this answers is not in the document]
tsachiherman: wasn't trivial.. done.

algorandskiy (dismissed):
IMO it looks better now, isn't it? :)

catchup/peerSelector.go
```diff
@@ -0,0 +1,287 @@
...
+	}
+}
+
+// findPeer look into the peer pool and find the given peer.
+// The method returns the pool and peer indices if a peer was found, or (-1, -1) otherwise.
+func (ps *peerSelector) findPeer(peer network.Peer) (poolIdx, peerIdx int) {
+	peerAddr := peerAddress(peer)
+	if peerAddr != "" {
+		for i, pool := range ps.pools {
+			for j, localPeerEntry := range pool.peers {
```
algonautshant, line 268: This is called every time RankPeer is called.
Why not keep a map from address to rank? Then, knowing the rank of the address, can do binary search on pools (since it is sorted), and locate the pool where the address/peer resides. 
refreshAvailablePeers, RankPeer, PeerDownloadDurationToRank, findPeer and GetNextPeer will all benefit from such a data structure, and is not costly to maintain it.

catchup/peerSelector.go
```diff
@@ -0,0 +1,287 @@
...
+}
+
+// calculate the duration rank by mapping the range of [minDownloadDuration..maxDownloadDuration] into the rank range of [minRank..maxRank]
+func downloadDurationToRank(downloadDuration, minDownloadDuration, maxDownloadDuration time.Duration, minRank, maxRank int) (rank int) {
+	if downloadDuration < minDownloadDuration {
+		downloadDuration = minDownloadDuration
+	} else if downloadDuration > maxDownloadDuration {
+		downloadDuration = maxDownloadDuration
+	}
+	rank = minRank + int((downloadDuration-minDownloadDuration).Nanoseconds()*int64(maxRank-minRank)/(maxDownloadDuration-minDownloadDuration).Nanoseconds())
```
algonautshant, line 285: Can you please add some explanation on this formula?

algonautshant (commented):
Looks very good. 
I have one concern: when the rank is updated from the download duration, it does not take into consideration the previous performance of the peer (or did I miss that?). 
Maybe it is a very good performing peer, but that instant had a bad download? Then its rank is gonna be downgraded despite the past performances.
Maybe average the rank with the previous value? Or keep a moving window of most recent performances for the rank?

catchup/peerSelector.go
```diff
@@ -0,0 +1,287 @@
...
+// makePeerSelector creates a peerSelector, given a networkGetPeers and peerClass array.
+func makePeerSelector(net networkGetPeers, initialPeersClasses []peerClass) *peerSelector {
+	selector := &peerSelector{
+		net:         net,
+		peerClasses: initialPeersClasses,
+	}
+	return selector
+}
+
+// GetNextPeer returns the next peer. It randomally selects a peer from the lowest
```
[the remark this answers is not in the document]
tsachiherman: I agree that the "lowest rank pool" is not clear enough.  I'll change that to 
```
// GetNextPeer returns the next peer. It randomally selects a peer from a pool that has
// the lowest rank value. Given that the peers are grouped by their ranks, allow us to
// prioritize peers based on their class and/or performance.
```

catchup/peerSelector.go
```diff
@@ -0,0 +1,287 @@
...
+}
+
+// peerSelector is a helper struct used to select the next peer to try and connect to
+// for various catchup purposes. Unlike the underlying network GetPeers(), it allows the
+// client to provide feedback regarding the peer's performance, and to have the subsequent
+// query(s) take advantage of that intel.
+type peerSelector struct {
+	mu          deadlock.Mutex
+	net         networkGetPeers
+	peerClasses []peerClass
```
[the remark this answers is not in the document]
tsachiherman: I don't expect any use of this struct to try and have the same PeerOption. It won't break anything, but it would not be useful.

catchup/peerSelector.go
```diff
@@ -0,0 +1,287 @@
...
+	defer ps.mu.Unlock()
+
+	poolIdx, peerIdx := ps.findPeer(peer)
+	if poolIdx < 0 || peerIdx < 0 {
+		return false
+	}
+
+	// we need to remove the peer from the pool so we can place it in a different location.
+	pool := ps.pools[poolIdx]
+	class := pool.peers[peerIdx].class
```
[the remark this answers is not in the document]
tsachiherman: sure.

catchup/peerSelector.go
tsachiherman: I think that we can definitely should consider this in the future, if we see that this operation takes notable amount of time.

At this time, we don't have enough relays to really justify that. The good part is that if we do want to optimize it, the entire optimization can be done within this file.

catchup/peerSelector.go
tsachiherman: will do.

catchup/peerSelector.go
```diff
@@ -0,0 +1,287 @@
...
+// and define tnet network.PeerOption that would be used to retrieve that type of
+// peer
+type peerClass struct {
+	initialRank int
+	peerClass   network.PeerOption
+}
+
+// the networkGetPeers is a subset of the network.GossipNode used to ensure that we can create an instance of the peerSelector
+// for testing purposes, providing just the above function.
+type networkGetPeers interface {
```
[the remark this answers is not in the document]
tsachiherman: interfaces in go are usually named "XXX-er", so I think that `peersRetriever` is a good alternative. Do you agree ?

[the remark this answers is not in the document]
algonautshant: Yes. peersRetriever is good.

algonautshant (approved):

PR #1867, opened 2021-01-27
Added Repeat configuration flag to support single invocation.

egieseke:
## Summary

Add configuration flag to loadgenerator utility to specify if it it should repeat after the first round.

## Test Plan

Tested manually.

### DIFF
--- a/cmd/loadgenerator/main.go
+++ b/cmd/loadgenerator/main.go
@@ -17,6 +17,7 @@
 package main
 
 import (
+	"flag"
 	"fmt"
 	"os"
 	"runtime"
@@ -35,6 +36,12 @@ import (
 )
 
 const transactionBlockSize = 800
+
+var runOnce = flag.Bool("once", false, "Terminate after first spend loop")
+
+func init() {
+	flag.Parse()
+}
 
 var nroutines = runtime.NumCPU() * 2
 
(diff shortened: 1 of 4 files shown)


cmd/loadgenerator/main.go
```diff
@@ -177,12 +181,12 @@ func generateTransactions(restClient client.RestClient, cfg config, privateKey *
...
-				_, err = restClient.SendRawTransaction(txns[x])
-				if err != nil {
-					if strings.Contains(err.Error(), "txn dead") || strings.Contains(err.Error(), "below threshold") {
+				_, err2 := restClient.SendRawTransaction(txns[x])
+				if err2 != nil {
+					if strings.Contains(err2.Error(), "txn dead") || strings.Contains(err2.Error(), "below threshold") {
 						break
 					}
-					fmt.Fprintf(os.Stderr, "unable to send transaction : %v\n", err)
+					fmt.Fprintf(os.Stderr, "unable to send transaction : %v\n", err2)
```
tsachiherman, line 189: the function doesn't return the `err` variable, so I can't see why you would need to allocate another `err2` variable.. anything that I'm missing ?

tsachiherman (changes requested):
I think that adding a "-once" command line is more suitable than adding a configuration option.

cmd/loadgenerator/main.go
egieseke: I was seeing a memory segmentation fault when this is handling errors.

tsachiherman: huh.. I think that you're correct in that it should be a unique variable per go-routine.
Let me confirm that tomorrow morning with race enabled.

cmd/loadgenerator/main.go
```diff
@@ -36,6 +37,12 @@ import (
 
 const transactionBlockSize = 800
 
+var runOnce = flag.Bool("once", false, "Terminate after first spend loop")
+
+func init() {
+	flag.Parse()
```
tsachiherman, line 43: I would have put this in the first line of `main`, but your implementation is not wrong.

tsachiherman:
I will approve this after testing it with the -race flag. I'd hope to see that the issue you found is the only issue.

tsachiherman (approved):

PR #1868, opened 2021-01-27
Each channel should get its keys by its unique bucket/prefix

btoll:
## Summary

The API call that the releases page script is using will only return a max of 1000 keys, and as the number of objects increases on s3 this adds to the fragility of the script as the returned objects probably won't contain all of the ones that we're interested in.

This fix will target each channel's prefix, which is much more robust and predictable.

This is a blocker for the package deployment.

## Test Plan

Generated a releases page index.html and tested the links by hand.

### DIFF
--- a/scripts/release/mule/deploy/releases_page/generate_releases_page.py
+++ b/scripts/release/mule/deploy/releases_page/generate_releases_page.py
@@ -6,13 +6,13 @@ import sys
 import boto3
 
 staging_bucket = "algorand-dev-deb-repo"
-staging_prefix = "http://algorand-dev-deb-repo.s3-website-us-east-1.amazonaws.com/"
+#staging_prefix = "http://algorand-dev-deb-repo.s3-website-us-east-1.amazonaws.com/"
+staging_prefix = "https://algorand-dev-deb-repo.s3.amazonaws.com/"
 key_url = "https://releases.algorand.com/key.pub"
 releases_bucket = "algorand-releases"
 releases_prefix = "https://releases.algorand.com/"
 html_tpl = "html.tpl"
 styles_url = "releases_page.css"
-tokens = ["stable", "beta", "indexer"]
 
 def get_stage_release_set(response):
     prefix = None
@@ -116,11 +116,11 @@ def getContent(url):
 
     return content
 
-def build_page(channels):
+def build_page(tables):
     html = getContent(html_tpl).replace("{styles}", getContent(styles_url))
 
-    for n in tokens:
-        html = html.replace("".join(["{", n, "}"]), "".join(channels[n]))
+    for n in tables:
+        html = html.replace("".join(["{", n, "}"]), "".join(tables[n]))
 
     sys.stdout.write(html)
 
@@ -133,12 +133,16 @@ def get_furl(release_files, fname, skey):
 
 def main():
     s3 = boto3.client("s3")
-    channels = {}
+    tables = {}
 
     for channel in ["stable", "beta", "indexer"]:
-        staging_response = s3.list_objects_v2(Bucket=staging_bucket, Prefix="releases/" + channel + "/", MaxKeys=100)
+        staging_response = s3.list_objects_v2(Bucket=staging_bucket, Prefix="".join(("releases/", channel, "/")), MaxKeys=100)
         release_sets = get_stage_release_set(staging_response)
-        releases_response = s3.list_objects_v2(Bucket=releases_bucket)
+        if channel == "indexer":
+            prefix = "indexer/"
+        else:
+            prefix = "".join(("channel/", channel, "/"))
+        releases_response = s3.list_objects_v2(Bucket=releases_bucket, Prefix=prefix, MaxKeys=100)
         release_files = objects_by_fname(releases_response["Contents"])
 
         table = []
@@ -151,7 +155,7 @@ def main():
             for fname, info in files.items():
                 if "file" not in info:
                     continue
-                furl = get_furl(release_files, fname, info['file'])
+                furl = get_furl(release_files, fname, info["file"])
                 ftext = '<div class="fname"><a href="{}">{}</a></div>'.format(furl, fname)
                 sig = info.get(".sig")
                 stext = ""
@@ -183,9 +187,9 @@ def main():
             if len(files.items()) > 1:
                 table.append('<tbody><tr class="spacer"><td></td></tr></tbody>')
 
-        channels[channel] = table
+        tables[channel] = table
 
-    build_page(channels)
+    build_page(tables)
 
 if __name__ == "__main__":
     main()


scripts/release/mule/deploy/releases_page/generate_releases_page.py
```diff
@@ -6,13 +6,13 @@
 import boto3
 
 staging_bucket = "algorand-dev-deb-repo"
-staging_prefix = "http://algorand-dev-deb-repo.s3-website-us-east-1.amazonaws.com/"
+#staging_prefix = "http://algorand-dev-deb-repo.s3-website-us-east-1.amazonaws.com/"
```
btoll, line 9: I wanted to keep the old url around for a bit.

brianolson:
There are better ways to fix this, that I have been arguing for since I got here 2 years ago.
If we want to get the newest stuff from an S3 bucket, the correct way is to write records with a reverse time string in the path. (e.g. https://github.com/algorand/go-algorand/blob/master/scripts/release/prod/reverse_hex_timestamp )

tsachiherman:
I don't know what the limit on the AWS query side, but to me, it looks like you're the one who limits it:
```
MaxKeys=100
```

btoll:
> There are better ways to fix this, that I have been arguing for since I got here 2 years ago.
> If we want to get the newest stuff from an S3 bucket, the correct way is to write records with a reverse time string in the path. (e.g. https://github.com/algorand/go-algorand/blob/master/scripts/release/prod/reverse_hex_timestamp )

We are doing this as part of the deployment process: https://github.com/algorand/go-algorand/blob/master/scripts/release/mule/deploy/releases_page/generate_releases_page.sh#L21

The problem is that we're not doing that everywhere.  If we're looking up objects in `algorand-releases` and it's not bubbling the newest stuff to the top, then we start running into issues such as what this is trying to address.

My opinion is that we need to be more disciplined about how we structure our s3 buckets and their prefixes.  Currently, it's difficult to do as you suggest because we're referencing things all over the place.

btoll:
> I don't know what the limit on the AWS query side, but to me, it looks like you're the one who limits it:
> 
> ```
> MaxKeys=100
> ```

https://docs.aws.amazon.com/cli/latest/reference/s3api/list-objects-v2.html

tsachiherman:
> > I don't know what the limit on the AWS query side, but to me, it looks like you're the one who limits it:
> > ```
> > MaxKeys=100
> > ```
> 
> https://docs.aws.amazon.com/cli/latest/reference/s3api/list-objects-v2.html

right; this is the proper location for the documentation.. and the documentation claims that the default is 1000. but you're limiting this to 100, which sounds like making the issue more substantial.

Also, you might want to review the documentation again. I vaguely remember something about the prefix filtering being done on the client side rather than the server side.

egieseke (approved):

btoll:
> > > I don't know what the limit on the AWS query side, but to me, it looks like you're the one who limits it:
> > > ```
> > > MaxKeys=100
> > > ```
> > 
> > 
> > https://docs.aws.amazon.com/cli/latest/reference/s3api/list-objects-v2.html
> 
> right; this is the proper location for the documentation.. and the documentation claims that the default is 1000. but you're limiting this to 100, which sounds like making the issue more substantial.
> 
> Also, you might want to review the documentation again. I vaguely remember something about the prefix filtering being done on the client side rather than the server side.

Setting it to the max 1000 does not return all of the objects as there are many more than that currently in s3.  That was the first thing that I did.

tsachiherman:
> > > > I don't know what the limit on the AWS query side, but to me, it looks like you're the one who limits it:
> > > > ```
> > > > MaxKeys=100
> > > > ```
> > > 
> > > 
> > > https://docs.aws.amazon.com/cli/latest/reference/s3api/list-objects-v2.html
> > 
> > 
> > right; this is the proper location for the documentation.. and the documentation claims that the default is 1000. but you're limiting this to 100, which sounds like making the issue more substantial.
> > Also, you might want to review the documentation again. I vaguely remember something about the prefix filtering being done on the client side rather than the server side.
> 
> Setting it to the max 1000 does not return all of the objects as there are many more than that currently in s3. That was the first thing that I did.

So, if setting the max to 1000 gave you only the first 1000 results, why did you reduce it to 100 ?

btoll:
> > > > > I don't know what the limit on the AWS query side, but to me, it looks like you're the one who limits it:
> > > > > ```
> > > > > MaxKeys=100
> > > > > ```
> > > > 
> > > > 
> > > > https://docs.aws.amazon.com/cli/latest/reference/s3api/list-objects-v2.html
> > > 
> > > 
> > > right; this is the proper location for the documentation.. and the documentation claims that the default is 1000. but you're limiting this to 100, which sounds like making the issue more substantial.
> > > Also, you might want to review the documentation again. I vaguely remember something about the prefix filtering being done on the client side rather than the server side.
> > 
> > 
> > Setting it to the max 1000 does not return all of the objects as there are many more than that currently in s3. That was the first thing that I did.
> 
> So, if setting the max to 1000 gave you only the first 1000 results, why did you reduce it to 100 ?

I'm not sure the last time I touched that.  We use that to control the number of releases that appear on the releases page.

tsachiherman:
> > > > > > I don't know what the limit on the AWS query side, but to me, it looks like you're the one who limits it:
> > > > > > ```
> > > > > > MaxKeys=100
> > > > > > ```
> > > > > 
> > > > > 
> > > > > https://docs.aws.amazon.com/cli/latest/reference/s3api/list-objects-v2.html
> > > > 
> > > > 
> > > > right; this is the proper location for the documentation.. and the documentation claims that the default is 1000. but you're limiting this to 100, which sounds like making the issue more substantial.
> > > > Also, you might want to review the documentation again. I vaguely remember something about the prefix filtering being done on the client side rather than the server side.
> > > 
> > > 
> > > Setting it to the max 1000 does not return all of the objects as there are many more than that currently in s3. That was the first thing that I did.
> > 
> > 
> > So, if setting the max to 1000 gave you only the first 1000 results, why did you reduce it to 100 ?
> 
> I'm not sure the last time I touched that. We use that to control the number of releases that appear on the releases page.

The thing I'm confused about here is what was the trigger here. I realize the the latest release wasn't appearing on the generated page. However, I would think that changing the 100 -> 1000 would solve the issue.
( what is the number of objects in that bucket anyway ? ) 

I can't imagine that the latest release bumped the number of entries in that bucket from <100 to over 1000.

btoll:
> > > > > > > I don't know what the limit on the AWS query side, but to me, it looks like you're the one who limits it:
> > > > > > > ```
> > > > > > > MaxKeys=100
> > > > > > > ```
> > > > > > 
> > > > > > 
> > > > > > https://docs.aws.amazon.com/cli/latest/reference/s3api/list-objects-v2.html
> > > > > 
> > > > > 
> > > > > right; this is the proper location for the documentation.. and the documentation claims that the default is 1000. but you're limiting this to 100, which sounds like making the issue more substantial.
> > > > > Also, you might want to review the documentation again. I vaguely remember something about the prefix filtering being done on the client side rather than the server side.
> > > > 
> > > > 
> > > > Setting it to the max 1000 does not return all of the objects as there are many more than that currently in s3. That was the first thing that I did.
> > > 
> > > 
> > > So, if setting the max to 1000 gave you only the first 1000 results, why did you reduce it to 100 ?
> > 
> > 
> > I'm not sure the last time I touched that. We use that to control the number of releases that appear on the releases page.
> 
> The thing I'm confused about here is what was the trigger here. I realize the the latest release wasn't appearing on the generated page. However, I would think that changing the 100 -> 1000 would solve the issue.
> ( what is the number of objects in that bucket anyway ? )
> 
> I can't imagine that the latest release bumped the number of entries in that bucket from <100 to over 1000.

Sorry, I should have provided more context.

It's querying the `algorand-releases` bucket to get its objects and then it's going through each of the staged assets that were just synced to `algorand-dev-deb-repo` to see if they're present in the returned object.  Unfortunately, the number of objects is so great that the lookups fail.

The answer may not be what I have done here but to refactor the script.  Since this was blocking the release, I fixed it quickly so it targets the prefix of each channel.

tsachiherman:
Yes, I understand that.. but all I'm questioning is the decision to convert
```
releases_response = s3.list_objects_v2(Bucket=releases_bucket)
```
into
```
releases_response = s3.list_objects_v2(Bucket=releases_bucket, Prefix=prefix, MaxKeys=100)
```

I think that adding the prefix is fine.. but the MaxKeys should be either dropped to set to 1000. Otherwise, you'll run into the same issue again very soon.

brianolson:
S3 sorting and prefix filtering is done server side, which is why the reverse-timestamp pattern is powerful in S3.
`The problem is that we're not doing that everywhere`
And I have consistently argued that we _should_ be doing it everywhere, and repeatedly raised the warning that our scripts could break, and apparently now they have.

algorandskiy:
I agree with Brian about naming.

>The API call that the releases page script is using will only return a max of 1000 keys

FYI, `list_objects_v2` supports pages of `MaxKeys` of size by using `ContinuationToken` arg and `IsTruncated`/`NextContinuationToken` fields from the response.

btoll:
> I agree with Brian about naming.
> 
> > The API call that the releases page script is using will only return a max of 1000 keys
> 
> FYI, `list_objects_v2` supports pages of `MaxKeys` entries in size by using `ContinuationToken` arg and `IsTruncated`/`NextContinuationToken` fields from the response.

I like the approach as well.  I'm happy to refactor this to use that approach.  At the time of fixing this, I was trying to be pragmatic and be as surgical as possible.

Also, I wonder if looking up the objects in `algorand-releases` is complicating it.  Why not just construct the page from the contents of the `algorand-dev-deb-repo` bucket?

brianolson:
We should have release artifacts in algorand-releases because that gets served up as https://releases.algorand.com and that looks good as a secure official download site.
The schema of the algorand-releases bucket is limited by go-algorand/cmd/updater which does S3 API queries of it. I have previously written a fix for this (2019-12) but it was rejected as not necessary.

btoll:
> We should have release artifacts in algorand-releases because that gets served up as https://releases.algorand.com and that looks good as a secure official download site.
> The schema of the algorand-releases bucket is limited by go-algorand/cmd/updater which does S3 API queries of it. I have previously written a fix for this (2019-12) but it was rejected as not necessary.

I kind of remember that.  It sounds good to me.

bricerisingalgorand (approved):

brianolson:
follow up: better long term fix in #1871
This fixes the `updater` dependency; then we can start writing build products to the better S3 schema.

tsachiherman:
just to be clear - if the code changes here fix the problem that you should merge these in. My suggestion above was just a note regarding future potential failures.

btoll:
> just to be clear - if the code changes here fix the problem that you should merge these in. My suggestion above was just a note regarding future potential failures.

I think I should still throttle the number of returns using `MaxKeys`, shouldn't I?  Otherwise, it will return older releases that we may not be interested in posting to the releases page.

brianolson:
If we want to show 'newest N' on the releases page that should be configured explicitly, e.g. to show the newest 5 releases. The S3 fetch should do batches of whatever size is right for that API (100-1000) then we filter that for what we want.

onetechnical:
Closing this; we are using a reverse hex timestamp, so I'm not sure this is meaningful here.

PR #1871, opened 2021-01-28
Updater schema2021

brianolson:
## Summary

Allow improving the schema of release S3 buckets to make queries for "newest things first" easier.
`updater` would query the new schema first, probably return something very quickly, or if during a transitional phase nothing is there yet it would fall back to the old schema.

## Test Plan

This needs more testing, probably including practical testing in the dev-release buckets.
PR made early for design review, more testing is a TODO


util/s3/s3Helper.go
```diff
@@ -262,17 +264,56 @@ func (helper *Helper) GetPackageFilesVersion(channel string, pkgFiles string, sp
...
+	prefix := fmt.Sprintf("latest/%s", channel)
+	input := s3.ListObjectsInput{
+		Bucket:  &helper.bucket,
+		Prefix:  &prefix,
+		MaxKeys: aws.Int64(50),
+	}
+	if specificVersion != 0 {
+		input.MaxKeys = aws.Int64(1000)
+	}
+	result, err := svc.ListObjects(&input)
```
algorandskiy, line 281: two suggestions:
1) switch to `ListObjectsV2` since the old one most likely will be deprecated one day
2) use double loop to bypass possible 1000 keys limit, something like this:
```
var result s3.ListObjectsOutput{IsTruncated: True}
for result.IsTruncated {
    result, err = svc.ListObjectsV2(&input)
    for _, item := range result.Contents {
        ...
    }
    if result.IsTruncated {
        input.Marker = result. NextMarker
    }
}

algorandskiy (commented):
+1, one suggestion below

gmalouf:
Closing as has been put in ice box / will need to be revisited down the line.

PR #1872, opened 2021-01-28
Fix tealdbg Accounts array

jasonpaulos:
## Summary

Currently the `Accounts` array is not displayed properly in tealdbg. The debugger does not account for the special case that index 0 of the `Accounts` array is always the sender. This means that if the ForeignAccounts txn field contains `n` accounts, the `Accounts` array in tealdbg contains the sender's account followed by `n-1` of the ForeignAccounts. The last account never gets displayed.

This PR fixes this problem by making tealdbg aware that the `Accounts` array has 1 more member (the sender's account) than `len(txn.Accounts)`. As a consequence of this, the sender's account will always appear at index 0 of the `Accounts` array in the debugger, even when the transaction's ForeignAccounts array is empty. This aligns with the behavior of the TEAL op `txna Accounts 0` always returning the sender's address, even in stateless TEAL.

## Test Plan

Tested visually on the following txn group:

```
txns.msgp[0]
{
  ...
}
txns.msgp[1]
{
  "sig": "eGj0ng3AH2AtKDCKJ8kGKxaLIcx2lNuY5/icVVGk5eklrqPK8wby4lgytE7gRg5VbZoZlh+gI0BIUfm/kzP8Bg==",
  "txn": {
    "apaa": [
      "c3dhcA==",
      "Zmk="
    ],
    "apat": [
      "ECMT7R5ZU2U4RWFV3D3GOE3ORSJ4WTZNUTPAUJH5KT7YYLGGMOVVUA2KAA"
    ],
    "apid": 18167,
    "fee": 1000,
    "fv": 115378,
    "gen": "sandnet-v1",
    "gh": "fNM6YsUZ/negqJ9mnBt348s38pbuoe/Xyw9iT6VUbzg=",
    "grp": "H6B8jJ3AACvyz7uEnjxeccFJL6sR5GyYQNi8nF4WZYo=",
    "lv": 116378,
    "snd": "XZ7SYOHYQAEES7XIEJ554GET66IEDEA7ABMGGH3ELQZNTSTK36BMCAAY7A",
    "type": "appl"
  }
}
txns.msgp[2]
{
  ...
}
txns.msgp[3]
{
  ...
}
```

<details>
<summary>Behavior before this PR:</summary>
<p>Note that the Accounts array should contain "ECMT7R5ZU2U4RWFV3D3GOE3ORSJ4WTZNUTPAUJH5KT7YYLGGMOVVUA2KAA", but it only contains the sender's address.</p>

</details>

<details>
<summary>Behavior with this PR:</summary>
<p>Now the Accounts array correctly shows the sender's address at index 0 and the first ForeignAccount argument at index 1.</p>

</details>

### DIFF
--- a/cmd/tealdbg/cdtState.go
+++ b/cmd/tealdbg/cdtState.go
@@ -721,7 +721,7 @@ func makeTxnImpl(txn *transactions.Transaction, groupIndex int, preview bool) (d
 		var length int
 		switch logic.TxnField(fieldIdx) {
 		case logic.Accounts:
-			length = len(txn.Accounts)
+			length = len(txn.Accounts) + 1
 		case logic.ApplicationArgs:
 			length = len(txn.ApplicationArgs)
 		}
@@ -765,7 +765,7 @@ func makeTxnArrayField(s *cdtState, groupIndex int, fieldIdx int) (desc []cdt.Ru
 		var length int
 		switch logic.TxnField(fieldIdx) {
 		case logic.Accounts:
-			length = len(txn.Accounts)
+			length = len(txn.Accounts) + 1
 		case logic.ApplicationArgs:
 			length = len(txn.ApplicationArgs)
 		}


tsachiherman:
what is the behavior if any of the `ForeignAccounts` is identical to the sender account ?

jasonpaulos:
If the sender's address is one of the ForeignAccounts, then they'll appear more than once in that array. For example:

Txn:
```
{
  "sig": "uHW6ExRSkzZrifHKMPKngyBHXnNUcuuCISMYSiNaJJXGtft1imecyYtG27b0liUxXeTdH05xuLSFOIe6NKqHDA==",
  "txn": {
    "apat": [
      "6REYQZ6WLEPNXXDDPHRXRWGTI7PLS5RMI5TLUMNL6Q4ISW3WBQMLGQFHJM"
    ],
    "apid": 18213,
    "fee": 1000,
    "fv": 116972,
    "gen": "sandnet-v1",
    "gh": "fNM6YsUZ/negqJ9mnBt348s38pbuoe/Xyw9iT6VUbzg=",
    "lv": 117972,
    "snd": "6REYQZ6WLEPNXXDDPHRXRWGTI7PLS5RMI5TLUMNL6Q4ISW3WBQMLGQFHJM",
    "type": "appl"
  }
}
```

tealdbg:
sender in ForeignAccounts

The sender's address is at index 0 and 1. I believe this makes sense because both of the ops `txna Accounts 0` and `txna Accounts 1` will return the sender's address in this case.

tsachiherman:
Looks fine with me, but I'd like to see @algorandskiy feedback as well. Do we have any unit test infrastructure that you could modify to include this use case ?

algorandskiy (approved):

PR #1875, opened 2021-01-28
Handle #pragma errors more like all others

### LINKED ISSUE #1874: goal compile crash on bad pragma
### Subject of the issue
When compiling a TEAL program with an error related to #pragma handling, goal crashes instead of reporting the error.

### Steps to reproduce
goal clerk compile example.teal
while example.teal contains "#pragma version 3"

### Expected behaviour
An error for the unsupported version should be reported.

### Actual behaviour
goal crashes

jannotti:
This moves #pragma handling into the main loop of assembly, allowing
simpler error reporting.  Multiple errors that include pragma errors
and "normal" errors can both be reported.  Since this fixes #1874, I
want to get it into mainline, rather than waiting for the teal3
branch.

## Summary

## Test Plan

### DIFF
--- a/data/transactions/logic/assembler.go
+++ b/data/transactions/logic/assembler.go
@@ -942,10 +942,6 @@ type lineError struct {
 	Err  error
 }
 
-func fmtLineError(line int, format string, args ...interface{}) error {
-	return &lineError{Line: line, Err: fmt.Errorf(format, args...)}
-}
-
 func (le *lineError) Error() string {
 	return fmt.Sprintf("%d: %s", le.Line, le.Err.Error())
 }
@@ -1086,14 +1082,18 @@ func (ops *OpStream) assemble(fin io.Reader) error {
 			continue
 		}
 		if strings.HasPrefix(line, "#pragma") {
-			// all pragmas must be be already processed in advance
 			ops.trace("%d: #pragma line\n", ops.sourceLine)
+			ops.pragma(line)
 			continue
 		}
 		fields := fieldsFromLine(line)
 		if len(fields) == 0 {
 			ops.trace("%d: no fields\n", ops.sourceLine)
 			continue
+		}
+		// we're going to process opcodes, so fix the Version
+		if ops.Version == assemblerNoVersion {
+			ops.Version = AssemblerDefaultVersion
 		}
 		opstring := fields[0]
 		spec, ok := opsByName[ops.Version][opstring]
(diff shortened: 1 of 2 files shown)


data/transactions/logic/assembler.go
```diff
@@ -1143,6 +1143,49 @@ func (ops *OpStream) assemble(fin io.Reader) error {
...
+			return ops.error("no version value")
+		}
+		value := fields[2]
+		var ver uint64
+		if ops.pending.Len() > 0 {
+			return ops.error("#pragma version is only allowed before instructions")
+		}
+		ver, err := strconv.ParseUint(value, 0, 64)
+		if err != nil {
+			return ops.error(err)
```
brianolson, line 1167: should this have a nicer comment? `("bad version number %#v: %v", value, err)` or so?

data/transactions/logic/assembler_test.go
```diff
@@ -1325,114 +1326,63 @@ func TestStringLiteralParsing(t *testing.T) {
...
 
-	text = `#pragma version a`
-	sr = strings.NewReader(text)
-	ps = PragmaStream{}
-	err = ps.Process(sr)
-	require.Error(t, err)
-	require.Contains(t, err.Error(), "1: strconv.ParseUint")
-	require.Equal(t, uint64(0), ps.Version)
+	testProg(t, `#pragma version a`, assemblerNoVersion,
+		expect{1, "strconv..."})
```
brianolson, line 1344: per previous comment, "bad version number" instead of the raw strconv

brianolson (commented):
Generally I like it. One error message could be nicer.

brianolson:
I think @algorandskiy wrote `#pragma version` initially, so it's probably worth a 5 minute skim to make sure we didn't miss something non-obvious about its prior implementation.

data/transactions/logic/assembler_test.go
```diff
@@ -1492,10 +1432,7 @@ len
 	require.NoError(t, err)
 	require.Equal(t, ops2.Program, ops.Program)
 
-	_, err = AssembleString("#pragma unk")
-	require.Error(t, err)
-	require.Contains(t, err.Error(), "1: unsupported pragma directive: unk")
-
+	testProg(t, "#pragma unk", assemblerNoVersion, expect{1, "unsupported..."})
```
algorandskiy, line 1435: *unsupported pragma*, `unsupported...` does not distinguish *unsupported version* and *unsupported pragma*

data/transactions/logic/assembler_test.go
```diff
@@ -1446,16 +1396,11 @@ int 1
...
-
-	_, err = AssembleStringWithVersion(text, 2)
-	require.Error(t, err)
-	require.Contains(t, err.Error(), "version mismatch")
+	testProg(t, text, 0, expect{1, "version mismatch..."})
+	testProg(t, text, 2, expect{1, "version mismatch..."})
+	testProg(t, text, assemblerNoVersion)
 
 	ops, err = AssembleStringWithVersion(text, assemblerNoVersion)
-	require.NoError(t, err)
```
algorandskiy, line 1458: no error check makes sense, otherwise ops is nil and sigfault instead of nice test failure message

algorandskiy (commented):
Looks good, couple minor notes, see below.

Plus, two more tests would be useful:
1. AssembleStringWithVersion(xx, 100) should fail
2. `echo "#pragma version 3" | goal clerk compile -` should not fail with SIGSEGV. Looks like it came from errors refactoring PR, so better to have it covered.

data/transactions/logic/assembler.go
jannotti: Agreed, I think I'll leave out the strconv error entirely, in fact.  We have a lot of these in existing errors.  If make similar changes as I go through teal3 updates.

data/transactions/logic/assembler_test.go
jannotti: Agreed, not sure how that got removed.  Thanks.

jannotti:
Thanks for all the feedback.  All changes will be committed soon.  And yes, I definitely caused the SEGV when I did the multi-error handling..  When I saw that I decided to get this into mainline instead of off on a feature branch.

algorandskiy:
I guess we need to add `echo "#pragma version 100" | goal clerk compile` into some existing e2e test script and merge this.

jannotti:
I did add that test to unit tests (obviously at a lower-level, but at a
place that would have caught the bug this PR fixes), but e2e would be good
as well.  I'll try to figure that out, I'm not as familiar with it as the
unit tests.

algorandskiy:
I think the easiest is a new line in `test/scripts/e2e_subs/e2e-teal.sh`

jannotti:
> I think the easiest is a new line in `test/scripts/e2e_subs/e2e-teal.sh`

Thanks, added.

algorandskiy (approved):

PR #1876, opened 2021-01-28
Optimize LimitedReaderSlurper memory utilization

tsachiherman:
## Summary

The memory profiler shown that the `LimitedReaderSlurper` is one of the bigger memory consumer on a relay.
This behavior aligns well with the original intent - preallocate memory so that we won't need to reallocate it on every message. The unintended consequence of that was that a relay that has many incoming active connections might be using more memory than it really needs to.

To address this solution, I have used series of reading buffers. These buffers would be dynamically allocated on per-need basis. In order to retain original performance characteristics, I have set the size of the base buffer to be larger than the average message. Doing so would allow us to avoid allocation on *most* of the messages, and allocate buffers only when truly needed.

Another advantage is that the allocated memory is of a fixed, and small size. Allocating smaller buffers improves the performance when compared with larger buffers.

## Test Plan

Unit tests added, and memory benchmark was added.


network/limited_reader_slurper.go
```diff
@@ -24,52 +24,117 @@ import (
...
+			}
+
+			// make another buffer
+			s.allocateNextBuffer()
+		}
 
-	for s.size <= s.Limit {
-		more, err := reader.Read(s.buf[s.size:])
+		readBuffer = s.buffers[s.lastBuffer]
+		n, err := reader.Read((readBuffer[:cap(readBuffer)])[len(readBuffer):])
```
algorandskiy, line 90: mb `chunk := readBuffer[:cap(readBuffer)]; reader.Read(chunk[len(readBuffer ):]`  for readability?

network/wsPeer.go
```diff
@@ -39,6 +39,7 @@ import (
 )
 
 const maxMessageLength = 4 * 1024 * 1024 // Currently the biggest message is VB vote bundles. TODO: per message type size limit?
+const avarageMessageLength = 2 * 1024    // Most of the messages are smaller than this size, which makes it into a good base allocation.
```
algorandskiy, line 42: avarageMessageLength -> averageMessageLength

network/limited_reader_slurper.go
```diff
@@ -24,52 +24,117 @@ import (
...
+
+// MakeLimitedReaderSlurper creates a LimitedReaderSlurper instance with the provided base and max memory allocations.
+func MakeLimitedReaderSlurper(baseAllocation, maxAllocation uint64) *LimitedReaderSlurper {
+	if baseAllocation > maxAllocation {
+		baseAllocation = maxAllocation
+	}
+	lrs := &LimitedReaderSlurper{
+		remainedUnallocatedSpace: maxAllocation - baseAllocation,
+		lastBuffer:               0,
+		buffers:                  make([][]byte, 1+(maxAllocation-baseAllocation+allocationStep)/allocationStep),
```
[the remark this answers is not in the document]
tsachiherman: hmm. I'll update the formula and add a test for that.

network/limited_reader_slurper.go
```diff
@@ -24,52 +24,117 @@ import (
...
+				switch {
+				case n > 0:
+					// yes, there was at least one extra byte - return ErrIncomingMsgTooLarge
+					return ErrIncomingMsgTooLarge
+				case err == io.EOF:
+					// no, no more data. just return nil
+					return nil
+				case err == nil:
+					// if we received err == nil and n == 0, we should retry calling the Read function.
+					continue
```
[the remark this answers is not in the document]
tsachiherman: we can't do that. We need to wait here until we have enough data or the reader wouldn't provide more data for the current message. ( it won't lead to an infinite loop due to the nature of the web socket )
a value of zero returned bytes and non-error is a "normal" case for the io.Reader.

network/limited_reader_slurper.go
tsachiherman: sure.

network/wsPeer.go
tsachiherman: done.

algorandskiy (approved):

PR #1879, opened 2021-01-30
add a Merkle commitment to transactions in block header

zeldovich:
This will enable short proofs of transactions being present in a given
block.

The proof includes both the transaction ID (which is likely the most
useful part, knowing that some transaction ID was committed), as well as
the hash of the entire SignedTxn and ApplyData (in case some application
wants to know what LogicSig.Args were used, or what the ApplyData was
for rewards calculation).

The Merkle tree uses the same merklearray code that's already used for
compact certificates.


zeldovich:
I imagine that, to make this more useful for applications, it would be good for algod to expose an API for generating Merkle proofs for particular transactions in a block.  This is pretty straightforward, except for figuring out where to stick this in the new v2 REST API.  Is there some larger plan for where the v2 API is documented and how to add new things to the API, or should I just pick some sensible-looking path for this new API functionality and add it?

tsachiherman:
Could you please describe how the validation of that would take place ?
i.e. for a node that have the block where the transaction resides, a block number plus a transaction index would be enough.
For a node that does not have the block, the proof won't be enough to validate.

what am I missing ?

tsachiherman:
> I imagine that, to make this more useful for applications, it would be good for algod to expose an API for generating Merkle proofs for particular transactions in a block. This is pretty straightforward, except for figuring out where to stick this in the new v2 REST API. Is there some larger plan for where the v2 API is documented and how to add new things to the API, or should I just pick some sensible-looking path for this new API functionality and add it?

Talk to @rotemh before making changes; we had plans to add some type of RPC for functionalities that aren't RESTful by nature.

tsachiherman:
Could you please add a benchmark that test the worst case scenario for calculating the merkle tree root of, let's say, 40,000 transactions ?

tsachiherman:
Do you think it would be possible to populate the TxnRoot with this new hash ? It sounds to me like you're hashing(roughly) the same data.
The non-flat PaysetCommitFlat can be removed from the source code; we switched to `PaysetCommitFlat = true` before launch ( which was v17 ), so we can safely drop this.

zeldovich:
> Could you please describe how the validation of that would take place ?
> i.e. for a node that have the block where the transaction resides, a block number plus a transaction index would be enough.
> For a node that does not have the block, the proof won't be enough to validate.
> 
> what am I missing ?

The intent is for this feature to be used in combination with compact certificates.  The compact cert (plus previous-block hashes) will enable validating block headers, and this feature will enable validating that some transaction ID appeared in that block.  (For validation, the caller will need the Merkle proof, as well as the transaction's position within this block.)

zeldovich:
> Do you think it would be possible to populate the TxnRoot with this new hash ? It sounds to me like you're hashing(roughly) the same data.
> The non-flat PaysetCommitFlat can be removed from the source code; we switched to `PaysetCommitFlat = true` before launch ( which was v17 ), so we can safely drop this.

In principle, yes.  The reason I added it as a separate field is because I was a bit worried that other users might be relying on our `TxnRoot` field.  If you expect that's not the case, then yes, we should be able to replace `TxnRoot` with `TxnMerkleRoot`, since it should ensure the consistency of the same `Payset` information.  That would avoid the cost of encoding and hashing the `Payset` array.

What's your sense as to whether anyone is using `TxnRoot`?

zeldovich:
> Could you please add a benchmark that test the worst case scenario for calculating the merkle tree root of, let's say, 40,000 transactions ?

I added `BenchmarkTxnRoots`; as a first cut, it takes 17 msec to commit to a max-sized block of small transactions using FlatCommit, and about 170 msec to commit using the Merkle tree.  I'll poke around to see if there's something we can speed up in the Merkle tree commit path, but even as-is, this seems plausible.

zeldovich:
> > I imagine that, to make this more useful for applications, it would be good for algod to expose an API for generating Merkle proofs for particular transactions in a block. This is pretty straightforward, except for figuring out where to stick this in the new v2 REST API. Is there some larger plan for where the v2 API is documented and how to add new things to the API, or should I just pick some sensible-looking path for this new API functionality and add it?
> 
> Talk to @rotemh before making changes; we had plans to add some type of RPC for functionalities that aren't RESTful by nature.

@rotemh , let me know where you think we should add an API to generate a Merkle proof for a transaction in a committed block; thanks!

tsachiherman:
regardless of the other PR, note that a binary merkle tree computation is task that is generally highly parallelizable. I'd like to revisit the benchmarks after completing the rest of the optimizations.

tsachiherman:
> > Do you think it would be possible to populate the TxnRoot with this new hash ? It sounds to me like you're hashing(roughly) the same data.
> > The non-flat PaysetCommitFlat can be removed from the source code; we switched to `PaysetCommitFlat = true` before launch ( which was v17 ), so we can safely drop this.
> 
> In principle, yes. The reason I added it as a separate field is because I was a bit worried that other users might be relying on our `TxnRoot` field. If you expect that's not the case, then yes, we should be able to replace `TxnRoot` with `TxnMerkleRoot`, since it should ensure the consistency of the same `Payset` information. That would avoid the cost of encoding and hashing the `Payset` array.
> 
> What's your sense as to whether anyone is using `TxnRoot`?

As far as I know, the algod is the only consumer of the `TxnRoot` *logic*. Other parties might cache/store this value, but none that I'm aware of are attempting to calculate it. I haven't looked into the details, but I'd like to verify with you that the merkle tree root calculation is accounting for **every single bit in the payset**. If this isn't the case, then we should probably have both.

Another thing that came to mind is to create the merke tree of the transaction groups rather then the transactions themselves. The rationale here is that we're committing to the transactions along with the arguments and apply data - but the TEAL in one transaction can access a different transaction in the same transaction group.

zeldovich:
> As far as I know, the algod is the only consumer of the `TxnRoot` _logic_. Other parties might cache/store this value, but none that I'm aware of are attempting to calculate it. I haven't looked into the details, but I'd like to verify with you that the merkle tree root calculation is accounting for **every single bit in the payset**. If this isn't the case, then we should probably have both.

Yes, the Merkle tree commits to every bit in the payset.

It sounds like we should replace the current TxnRoot with the new TxnMerkleRoot, then.

FWIW, with some of the other changes in the pipeline (PRs #1880 and #1882), computing the Merkle root is now about 2x faster than computing the flat commitment.

> Another thing that came to mind is to create the merke tree of the transaction groups rather then the transactions themselves. The rationale here is that we're committing to the transactions along with the arguments and apply data - but the TEAL in one transaction can access a different transaction in the same transaction group.

I could see an argument in either direction, and I would prefer to just commit to individual transactions.  If someone wants to verify an entire transaction group, they can ask for a Merkle proof for every transaction in that group, then.  The merklearray code happens to be particularly efficient for proofs of adjacent elements.

Priegle:
> I imagine that, to make this more useful for applications, it would be good for algod to expose an API for generating Merkle proofs for particular transactions in a block. This is pretty straightforward, except for figuring out where to stick this in the new v2 REST API. Is there some larger plan for where the v2 API is documented and how to add new things to the API, or should I just pick some sensible-looking path for this new API functionality and add it?

@rotemh

algorotem:
> > > I imagine that, to make this more useful for applications, it would be good for algod to expose an API for generating Merkle proofs for particular transactions in a block. This is pretty straightforward, except for figuring out where to stick this in the new v2 REST API. Is there some larger plan for where the v2 API is documented and how to add new things to the API, or should I just pick some sensible-looking path for this new API functionality and add it?
> > 
> > 
> > Talk to @rotemh before making changes; we had plans to add some type of RPC for functionalities that aren't RESTful by nature.
> 
> @rotemh , let me know where you think we should add an API to generate a Merkle proof for a transaction in a committed block; thanks!

I'd put it under `GET /v2/blocks/{round}/transactions/{txid}/proof`

test/e2e-go/upgrades/application_support_test.go
```diff
@@ -286,7 +286,7 @@ func TestApplicationsUpgradeOverGossip(t *testing.T) {
 	fixture.SetupNoStart(t, filepath.Join("nettemplates", "TwoNodes100SecondTestUnupgradedProtocol.json"))
 
 	// for the primary node, we want to have a different consensus which always enables applications.
-	primaryNodeUnupgradedProtocol := consensus[consensusTestFastUpgrade(protocol.ConsensusFuture)]
+	primaryNodeUnupgradedProtocol := consensus[consensusTestFastUpgrade(protocol.ConsensusCurrentVersion)]
```
tsachiherman, line 289: unrelated to your change, but I don't think that this is correct anymore.
We also seen to remove application support from `primaryNodeUnupgradedProtocol` for the test to be correct.

data/bookkeeping/txn_merkle.go
```diff
@@ -0,0 +1,98 @@
...
+
+	var buf [2 * crypto.DigestSize]byte
+	copy(buf[:crypto.DigestSize], txid[:])
+	copy(buf[crypto.DigestSize:], stib[:])
+
+	return protocol.TxnMerkleLeaf, buf[:]
+}
+
+// Hash implements an optimized version of crypto.HashObj(tme).
+func (tme *txnMerkleElem) Hash() crypto.Digest {
```
tsachiherman, line 88: you're hashing both the `stib` as well as the `txid`, but eventually they both commit to the same underlying transaction. is there any benefit that I'm missing in creating a
hash(hash(txn)+hash(stxinblock)+nonce) vs just hash(stxinblock) ?

test/e2e-go/upgrades/application_support_test.go
zeldovich: Yeah, that whole test isn't terribly meaningful anymore, as far as I can tell.

I just did enough to make it not fail erroneously, but it may well be the case that it's worth removing altogether.

data/bookkeeping/txn_merkle.go
zeldovich: The benefit of having the txid separately is that it simplifies verification of a transaction being committed, without requiring the verifier to decode the SignedTxnInBlock and determine if it's the same transaction as a particular txid it's looking for.

tsachiherman: yes, that's what I thought.. but that's not what you're doing - 
i.e. if you want to accomplish that, you need to have the actual merkle tree element contain the tx.ID() rather than a hash of the `txid` and `stib`. ( otherwise, you would have to have the hash(txib) in order to find it out ).

test/e2e-go/upgrades/application_support_test.go
tsachiherman: let me fix that in a separate ticket ? ( I'll add you to review that ).

data/bookkeeping/txn_merkle.go
zeldovich: Part of the proof is the hash of `stib`; the verifier doesn't have to do anything other than confirm that, this supplied `stib` value happens to be in the Merkle tree along with the txid it's expecting.

tsachiherman: right, I see that - but what it the added value here ? i.e. it sounds like we're forcing the user to have both of the values in all cases if he wants verify that it's in the Merkle tree. constructing the hash to be a "string" of the first hash followed by the second hash ( at leaf) would make much better "construct", as it would allow us to verify that the entry is there by comparing the bytes.

zeldovich: That is exactly what's going on here.  `txid` and `stib` are `crypto.Digest` hashes of the transaction ID and the `SignedTxnInBlock` respectively.

tsachiherman: so, what I'm probably missing is the last line `return crypto.Hash(s)` - why do you hash the entire constructed array instead of just returning it ?

zeldovich: Oh, I see what you are asking about.  This `Hash()` function returns the hash of the Merkle tree leaf, rather than the Merkle leaf itself.  (It is returned by `txnMerkleArray.GetHash()` above, and the `merklearray` code expects it to return the hash of the leaf.  I recently made this optimization/change in one of the earlier PRs leading up to this one.)

tsachiherman: given that the verifier is going to have just the hashes of the entire merkle tree ( and not the underlying data elements ), it would be a big saving if the verifier could prove that an element belongs to the tree just by looking on the hash value. this also consist with merkle proofs.

tsachiherman: another point ( unrelated to the previous one ):
do you think it would make sense to construct the array as
<hash(txid) + nonce + hash(stib)> instead of <nonce + hash(txid) +  hash(stib)> ?
I'm suggesting that so that we can perform log(n) lookups into the leafs instead of examining all of them.

zeldovich: I'm probably missing what you're thinking about here; sorry about that.

The verifier just needs the hashes leading up to the root.  The prover can indeed prove that an element is in the tree by supplying the hash of the leaf (i.e., the value returned by `txnMerkleElem.Hash()`).  If the verifier knows the entire `SignedTxn` and its associated `ApplyData`, it can compute both the `stib` hash and the `txid` on its own.  If the verifier knows the `txid` it's looking for, but doesn't know (or doesn't want to compute) the entire `SignedTxnInBlock`, it can ask for the proof to also include the `stib` hash, so that it can concatenate its known `txid` with this supplied `stib` hash to verify that its `txid` lives in the tree.

We could construct two Merkle trees: one for just the transaction IDs, and one for just the `SignedTxnInBlock`s.  But it seems slightly more efficient to have a single Merkle tree where the leaves are the concatenation of the transaction ID and the `SignedTxnInBlock` hash.

tsachiherman: That's the thing - I don't think that we *need* to have two trees:
if the leaf hash would consist abit longer byte array ( let's say 66 bytes ), we can achieve both at the same time - 
the `prover` would provide a merkle proof that contains all the elements (sibling) from the root and up to the txn leaf.
The `verifier` see that the txn leaf "long hash" contain either the txid or the hash(stib), and can verify the path up to the root.
And when the verifier is asked to provide a proof for either txid or hash(stib), it can do so without rebuilding the SignedTxn but rather just iterate on the leaves and find the correct leaf in log(n) steps.

zeldovich: I believe the design you describe is basically equivalent to the code in this PR.  The only difference, I think, is that you are describing this design in terms of having a Merkle tree of 2N nodes, and in this PR, I have this design implemented as using a `merklearray` Merkle tree of N nodes, where each node is effectively another 2-node subtree.

tsachiherman: I haven't thought about it in the terms of `N nodes merkle tree` vs `2N nodes merkle tree`, but you are correct in that my thinking would achieve that. The trigger for me was what seems to be a meaningless hashing - a hashing of previously hashed value, that only take away functionality.
I don't think that having your implementation as is would be a bad thing; In fact, from performance perspective, it's already slightly faster (7678093ns vs 6956778ns ). 
By skipping the hashing on the leaf node, I expect it to be ~10% faster, which would become meaningful if we were to increase the block size substentially.

zeldovich: I don't believe there is any meaningless hashing going on here: this hash reduces `2*crypto.DigestSize` bytes to just `crypto.DigestSize` bytes, just like the hashes of internal Merkle tree nodes.

tsachiherman: no intend to insult, the meaningless is ( from my perspective ) is just because reducing the size isn't really our goal.

If I were to reduce the number of bytes, I would have taken the first 16 bytes from the first one and add the first 16 bytes from the second one. Given that the inputs are already "secure hashes" on their own, it wouldn't reduce the overall security of the solution, and provide look-ability into the hashes value themselves.

Another point is that not all the hashes in the merkle tree has to be of the same length. I know that it might seems abit "unorthodox" to do so, but if we can get better value of it, than it's probably worth considering.

zeldovich: No insult taken -- I meant "meaningless" only in a technical sense.  :)

In some sense, I think reducing the size is pretty much the only goal of the Merkle tree.  Otherwise, the previous plan of hashing the entire payset is just as secure.

Indeed, having the same-size hashes in the Merkle tree at all layers simplifies the `merklearray` code, which is a nice benefit.

zeldovich:
I just pushed a new REST API for generating Merkle proofs, along with an e2e test that verifies a transaction's Merkle proof.  So, I think this should be more-or-less complete, assuming the Travis tests don't trip up on anything.

data/bookkeeping/txn_merkle.go
zeldovich: Maybe as a way of understanding what you are hoping to improve: what operation do you think we could make more efficient?  I'm thinking there's three things going on with this Merkle tree: (1) online nodes generating the tree root when we build the block, (2) REST API nodes generating a proof for a specific transaction in a block, and (3) some other code that verifies a proof against a known txn root hash from a block header.

tsachiherman: I'm happy you've broken that like that; I think that all the above would benefit from that.
The online nodes generating the tree root, would not be required to make additional hashing call for the leaf elements.
The REST API nodes generating a proof for multiple transactions ids, would not need to calculate any hash(s). Instead, they would be able to perform a binary search on the leafs and find the leaf that matches the desired transaction. From there, the proof could be obtained by looking at the siblings.
And last, the verification of the leaf node becomes a simple byte array compare. This become meaningful when "exporting" the proof to another platform that isn't go-oriented where you don't want to have another "magic" number that would be "corresponding" to the transaction id on the Algorand platform. I can also see us supporting this on TEAL, but I'm not really sure in what form.

daemon/algod/api/server/v2/handlers.go
```diff
@@ -182,6 +182,64 @@ func (v2 *Handlers) GetBlock(ctx echo.Context, round uint64, params generated.Ge
...
+
+	ledger := v2.Node.Ledger()
+	block, _, err := ledger.BlockCert(basics.Round(round))
+	if err != nil {
+		return internalError(ctx, err, errFailedLookingUpLedger, v2.Log)
+	}
+
+	proto := config.Consensus[block.CurrentProtocol]
+	if proto.PaysetCommit != config.PaysetCommitMerkle {
+		return internalError(ctx, err, "protocol does not support Merkle proofs", v2.Log)
```
tsachiherman, line 202: I think this should be 412 rather than 5XX since there is no issue with the server; the network is currently at "fault" here and re-trying will not be helpful. One notation is that it's a "deterministic" error : trying making the request again would not help.

@rotemh - what do you think should the error code be in this case ?

daemon/algod/api/server/v2/handlers.go
```diff
@@ -182,6 +182,64 @@ func (v2 *Handlers) GetBlock(ctx echo.Context, round uint64, params generated.Ge
...
+	}
+
+	tree, err := block.TxnMerkleTree()
+	if err != nil {
+		return internalError(ctx, err, "building Merkle tree", v2.Log)
+	}
```
tsachiherman, line 208: maybe move this section into the `if txns[idx].Txn.ID() == txID {` clause since it's being used only if the transaction is found ?

daemon/algod/api/server/v2/handlers.go
zeldovich: I changed it to 404.  Logically, I think you could make the argument that 404 makes some sense: this object does not exist, and will not exist, if the protocol doesn't support proofs at that block number.  412 seems to be relatively specific to `If-Modified-Since` and `If-None-Match`, which is not related to the error condition we're considering here.  But if @algorotem has a convincing argument for something else, I'm happy to change it.

tsachiherman:
Does it makes sense to build a proof for multiple transactions at a time ?

The code changes looks good to me. I'll be waiting for @algorotem approval for the REST API changes.

data/bookkeeping/txn_merkle.go
zeldovich: For (1), I agree that the cost to generate the root commitment would be slightly lower if we were to concatenate the `txid` and `stib` in the leaf, rather than hash them together.

For (2), I believe the number of hashes needed to generate a proof is exactly the same (zero) once the tree is built.  After `merklearray` builds a tree, it retains all intermediate hashes, and generating a proof for any particular element just requires choosing the hashes along the path to the root.  This would be true both for generating a proof for one transaction or for multiple transactions.

For (3), I believe no byte comparison of leaves is needed in either case.  The procedure for verifying a proof, in both this PR's design and in your proposed alternative design, is to take the information we expect to be in the leaf node, and hash it (along with appropriate siblings) until we get a candidate root hash.  Then, we just compare the root hash we computed to the root hash we expect to see, from the block header.

One downside of the alternative design we're discussing here is that the proof becomes slightly longer: the first sibling is 2*64 bytes, instead of just 64 bytes long.

I think I'd prefer keeping the code as-is in this PR, since it leads to a simpler `merklearray` (supporting longer leaves would require changing that code), and I don't think there's a clear advantage to larger leaves.

zeldovich:
> Does it makes sense to build a proof for multiple transactions at a time ?

I was thinking about that..  For transaction groups, which I was expecting to be one case where the user might want such a multi-transaction proof, I don't think it's necessary to do the proof at the Merkle tree level.  A proof for any transaction in a tx group means that the rest of the transactions in the group must also be there -- no additional Merkle proofs is required.

It's possible that a user might want to prove multiple independent transactions in the same block.  The `merklearray` code has logic to make such proofs more efficient than just two independent proofs.  The main reason I didn't add support for it is because I couldn't figure out an elegant REST API path for listing multiple transaction IDs. :-)  But also because it seems like a relatively unlikely case: if the user wants to prove multiple independent transactions, in the general case there's no reason to expect they would appear in the same block anyway.

data/bookkeeping/txn_merkle.go
```diff
@@ -0,0 +1,98 @@
...
+// header), or to generate proofs of membership for transactions that are
+// in this block.
+func (block Block) TxnMerkleTree() (*merklearray.Tree, error) {
+	return merklearray.Build(&txnMerkleArray{block: block})
+}
+
+// txnMerkleArray is a representation of the transactions in this block,
+// along with their ApplyData, as an array for the merklearray package.
+type txnMerkleArray struct {
+	block Block
```
algorotem, line 40: Are we using `Block` and not `transactions.Payset` because we need the `DecodeSignedTxn`?

data/bookkeeping/txn_merkle.go
```diff
@@ -0,0 +1,98 @@
...
+
+// Length implements the merklearray.Array interface.
+func (tma *txnMerkleArray) Length() uint64 {
+	return uint64(len(tma.block.Payset))
+}
+
+// Get implements the merklearray.Array interface.
+func (tma *txnMerkleArray) GetHash(pos uint64) (crypto.Digest, error) {
+	if pos >= uint64(len(tma.block.Payset)) {
+		return crypto.Digest{}, fmt.Errorf("txnMerkleArray.Get(%d) out of bounds %d", pos, len(tma.block.Payset))
```
algorotem, line 51: Super nit but consider changing the error to be more specific about the bounds. Maybe something along the lines of `txnMerkleArray.Get(%d) failed because you asked for transaction index %d but there are only %d transactions in the block` or `txnMerkleArray.Get(%d) out of bounds. Payset length - %d`

daemon/algod/api/server/v2/handlers.go
```diff
@@ -182,6 +182,64 @@ func (v2 *Handlers) GetBlock(ctx echo.Context, round uint64, params generated.Ge
...
+
+			proof, err := tree.Prove([]uint64{uint64(idx)})
+			if err != nil {
+				return internalError(ctx, err, "generating proof", v2.Log)
+			}
+
+			var proofconcat []byte
+			for _, proofelem := range proof {
+				proofconcat = append(proofconcat, proofelem[:]...)
+			}
```
algorotem, line 225: Did we benchmark these? Might be a fairly expensive group of operations for an API call.

data/bookkeeping/txn_merkle.go
zeldovich: Yes.

data/bookkeeping/txn_merkle.go
zeldovich: I'm not quite sure what you're suggesting.  I believe the error message currently includes both the index requested (the first `%d`) as well as the expected bound (the second `%d`).  Are you thinking the wording should be more verbose?

daemon/algod/api/server/v2/handlers.go
zeldovich: There are some benchmarks for various components of what's going on here (see various benchmarks in `crypto/merklearray`).  I think a rough upper bound would be `BenchmarkMerkleCommit/Item1000/Count10000` which takes 17 msec (although that's for a 10 MByte block containing 10K transactions of 1 KByte each).  Doesn't seem super expensive but certainly people operating publicly-accessible REST APIs (maybe Purestake?) should pay attention.

data/bookkeeping/txn_merkle.go
algorotem: Yes, I think it can be tricky for someone to guess what is the second number if it's out of context. But again, very nit.

daemon/algod/api/server/v2/handlers.go
algorotem: Fair enough, 17ms is not a disaster for such a call.

tsachiherman (approved):

PR #1880, opened 2021-01-31
faster computation of transaction ID and length

zeldovich:
This change optimizes ID() and GetEncodedLength() for Transaction, SignedTxn, and SignedTxnInBlock.  There are two sources of overhead that this change avoids: interface conversions (to msgp.Marshaler) and memory allocations (for temporary slices to hold the encoding).

To avoid interface conversions, the code invokes the msgp-generated MarshalMsg() method directly.

To avoid memory allocation overhead, the code uses a sync.Pool to track existing dirty slices that can be used for encoding.  This seems to be more efficient than letting the Go GC manage slices, probably because the encoder is OK with dirty (non-zero) slices.

This significantly improves performance according to BenchmarkEncoding (in data/transactions).  Before this change:

    BenchmarkEncoding/EncodeLen-32          276892        4363 ns/op
    BenchmarkEncoding/ID-32                 169942        7131 ns/op

After this change:

    BenchmarkEncoding/EncodeLen-32         1963832         593 ns/op
    BenchmarkEncoding/ID-32                1109022        1042 ns/op

The broader use of this change is in significantly speeding up the Merkle tree computation for PR #1879.

### DIFF
--- a/ledger/eval.go
+++ b/ledger/eval.go
@@ -733,7 +733,7 @@ func (eval *BlockEvaluator) transactionGroup(txgroup []transactions.SignedTxnWit
 		txibs = append(txibs, txib)
 
 		if eval.validate {
-			groupTxBytes += len(protocol.Encode(&txib))
+			groupTxBytes += txib.GetEncodedLength()
 			if eval.blockTxBytes+groupTxBytes > eval.proto.MaxTxnBytesPerBlock {
 				return ErrNoSpace
 			}
--- a/data/pools/transactionPool.go
+++ b/data/pools/transactionPool.go
@@ -763,7 +763,7 @@ func (pool *TransactionPool) AssembleBlock(round basics.Round, deadline time.Tim
 
 				for i, txib := range payset {
 					fee := txib.Txn.Fee.Raw
-					encodedLen := len(protocol.Encode(&txib))
+					encodedLen := txib.GetEncodedLength()
 
 					stats.IncludedCount++
 					totalFees += fee
--- a/data/transactions/signedtxn.go
+++ b/data/transactions/signedtxn.go
@@ -70,7 +70,16 @@ func (s SignedTxnInBlock) ID() {
 
 // GetEncodedLength returns the length in bytes of the encoded transaction
 func (s SignedTxn) GetEncodedLength() int {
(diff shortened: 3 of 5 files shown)


pbennett:
Just in case it wasn't known, running benchmark tests with -benchmem flag will track memory allocations as well which would be a great comparison to include here (in the already great performance improvment).

ledger/eval.go
```diff
@@ -733,7 +733,7 @@ func (eval *BlockEvaluator) transactionGroup(txgroup []transactions.SignedTxnWit
 		txibs = append(txibs, txib)
 
 		if eval.validate {
-			groupTxBytes += len(protocol.Encode(&txib))
+			groupTxBytes += txib.GetEncodedLength()
```
tsachiherman, line 736: extra brownie points ( overall solution ): we've currently triple-encoding the payset -
1. we encode it to calculate the block size.
2. we encode it to calculate the commit hash ( either via flat, or as Merkle tree ).
3. we encode it as a whole for the purpose of preparing the proposal ( I know that this isn't always the case, but this case happens to be on the critical path ).

Caching the encoded data of the *first* block could help us repeating the process.

tsachiherman (changes requested):
I think that these changes are generally good, but require so more work:
1. complete the merge of the msgp, and revision it.
2. add a PR to upgrade the msgp
3. Apply this change, and add error handling for these -
   * The error case for these should not be "panic". We need to use proper error handling and align it the the optimal business decision at each case. ( I'm pretty confident that panic isn't the correct action )

tsachiherman:
Could you please rebase this PR on top of master ?

tsachiherman (approved):
I think that the idea here is good, but the real gains were attained "by-accident" : 
What your change really did was to pass a previously allocated buffer back to the pool so that it can be "pulled" over later on. This lead avoiding reallocating-and-copying smaller buffers ( which is caused by append ).

If you want to make this even better, pass a `hint` to the `GetEncodingBuf` and if the retrieved buffer capacity is below the `hint`, just allocate a new one.

Having said that, I don't see any reason to block this one. It gets us performances gains as is, and it's not interfering with anything else.

PR #1881, opened 2021-01-31
use the error-free msgp encoding

zeldovich:
This PR uses msgp version 1.1.47, which eliminates the possibility of MarshalMsg returning an error.


tsachiherman (approved):
looks good.

PR #1882, opened 2021-02-01
speed up merklearray

zeldovich:
This PR speeds up merklearray, both with some small absolute performance wins (avoiding crypto.HashObj overheads), as well as parallelism in hashing the leaves and the internal nodes.

For tiny trees (10 elements where each element is something like 10-100 bytes), the parallelism adds a slight overhead (about 30-40 microseconds on my test machine, on top of a total runtime of 25-80 microseconds).  But for anything larger (either more elements or more complex-to-hash elements), this ends up being an improvement.

This might have some benefit for the current use of merklearray (namely, compact certificates), but should be even more useful for the upcoming use of merklearray for hashing all transactions in a block.

### DIFF
--- a/crypto/compactcert/builder.go
+++ b/crypto/compactcert/builder.go
@@ -129,12 +129,12 @@ func (sc sigsToCommit) Length() uint64 {
 	return uint64(len(sc))
 }
 
-func (sc sigsToCommit) Get(pos uint64) (crypto.Hashable, error) {
+func (sc sigsToCommit) GetHash(pos uint64) (crypto.Digest, error) {
 	if pos >= uint64(len(sc)) {
-		return nil, fmt.Errorf("pos %d past end %d", pos, len(sc))
+		return crypto.Digest{}, fmt.Errorf("pos %d past end %d", pos, len(sc))
 	}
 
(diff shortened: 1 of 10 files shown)


tsachiherman (approved):
The parallelism logic you've applied here is logically correct, but it wouldn't be as fast as it could be since you're forcing it to calculations of a single row.

I was thinking of grouping the bottom tree row into power-of-two groups and processing these in a recursive manner, by throttling the number of base row elements per go-routine.

zeldovich:
Yeah, I'm not sure exactly what's the best plan here.  I tried quite a number of schemes, and it's been tricky to find a balance that works well.  One constraint is that, since the work we're doing can be pretty cheap (like hashing an interior leaf), the overhead for distributing work across threads has to be minimal.  At the same time, the work could end up being costly (like large transactions at the leaves), in which case pre-assigning large batches of work limits our parallelism (though it achieves low overhead).  Simple code is also important since this is security-critical.

My sense is, this implementation captures most of the benefit, and gets pretty good performance across the board (both small and large leaves, and both small and large trees), so it's probably worth merging for now, and we can revisit if we think there are significant performance opportunities remaining.

PR #1886, opened 2021-02-02
Add asset close to amount field to transaction 

algorotem:
## Summary

This PR adds `AssetClosingAmount` to the `ApplyData`, conditioned by a new `ConsensusParam.EnableAssetCloseAmount`. A corresponding `AssetClosingAmount` was added to the REST API to align with the variable already presents on the indexer v2.

## Test Plan

Unit test added for `AssetTransfer`.


daemon/algod/api/spec/v1/model.go
```diff
@@ -675,6 +675,11 @@ type AssetTransferTransactionType struct {
 	//
 	// required: false
 	CloseTo string `json:"closeto"`
+
+	// CloseToAmount is amount of the remaining funds that were transferred to the close to address (if closing).
+	//
+	// required: false
+	CloseToAmount uint64 `json:"close-to-amount"`
```
tsachiherman, line 682: nit: maybe `AssetClosingAmount` ? we already have a field called `AssetID`, so the prefix "Asset" isn't foreign.

ledger/apply/asset.go
```diff
@@ -330,6 +330,10 @@ func AssetTransfer(ct transactions.AssetTransferTxnFields, header transactions.H
 		// and putIn will short circuit (so bypassFreeze doesn't matter)
 		_, bypassFreeze := dst.AssetParams[ct.XferAsset]
 
+		// Add the close amount to ApplyData.
+		closeAmount := sndHolding.Amount
+		ad.AssetClosingAmount = closeAmount
+
```
tsachiherman, line 336: This part has to be enclosed in a protocol upgrade parameter condition. specifically, you'll need to PR https://github.com/algorandfoundation/specs/blob/3a83c4c743f8b17adfd73944b4319c25722a6782/dev/ledger.md and have it merged. Once it's merged, you'll have a commit hash that you can use for the consensus protocol string.

tsachiherman (dismissed):
Once we have a consensus param we should be able to merge this in. If you don't have that yet, you can add a new parameter and enable this on "future" protocol.

daemon/algod/api/spec/v1/model.go
algorotem: it's `v1` I just matched it to the rest. That being said, I did change `close-amount-to` to `closeamountto` to match the pattern there.

ledger/apply/asset.go
algorotem: https://github.com/algorandfoundation/specs/pull/34

daemon/algod/api/server/v2/handlers.go
```diff
@@ -442,17 +442,18 @@ func (v2 *Handlers) PendingTransactionInformation(ctx echo.Context, txid string,
...
-		ClosingAmount    *uint64                        `codec:"closing-amount,omitempty"`
-		ConfirmedRound   *uint64                        `codec:"confirmed-round,omitempty"`
-		GlobalStateDelta *generated.StateDelta          `codec:"global-state-delta,omitempty"`
-		LocalStateDelta  *[]generated.AccountStateDelta `codec:"local-state-delta,omitempty"`
-		PoolError        string                         `codec:"pool-error"`
-		ReceiverRewards  *uint64                        `codec:"receiver-rewards,omitempty"`
-		SenderRewards    *uint64                        `codec:"sender-rewards,omitempty"`
-		Txn              transactions.SignedTxn         `codec:"txn"`
+		AssetIndex         *uint64                        `codec:"asset-index,omitempty"`
+		AssetClosingAmount *uint64                        `codec:"asset-closing-amount, omitempty"`
```
zeldovich, line 446: Does our code do the right thing with a blank space in the codec tag?

algorotem: Good catch, thanks!

daemon/algod/api/server/v1/handlers/handlers.go
```diff
@@ -315,6 +315,10 @@ func assetTransferTxEncode(tx transactions.Transaction, ad transactions.ApplyDat
 		xfer.CloseTo = tx.AssetTransferTxnFields.AssetCloseTo.String()
 	}
 
+	if ad.AssetClosingAmount != 0 {
```
algorandskiy, line 318: why do we care about v1?

ledger/apply/payment_test.go
```diff
@@ -101,15 +103,23 @@ func (balances mockBalances) PutWithCreatable(basics.Address, basics.AccountData
 	return nil
 }
 
-func (balances mockBalances) Get(basics.Address, bool) (basics.AccountData, error) {
+func (balances mockBalances) Get(addr basics.Address, withPendingRewards bool) (basics.AccountData, error) {
+	if ad, ok := balances.b[addr]; ok {
+		return ad, nil
+	}
 	return basics.AccountData{}, nil
```
algorandskiy, line 110: branch is not needed, if `addr` not in the map, it returns empty `AccountData`

ledger/apply/payment_test.go
```diff
@@ -101,15 +103,23 @@ func (balances mockBalances) PutWithCreatable(basics.Address, basics.AccountData
...
 	return basics.Address{}, true, nil
 }
 
-func (balances mockBalances) Put(basics.Address, basics.AccountData) error {
+func (balances mockBalances) Put(addr basics.Address, ad basics.AccountData) error {
+	if _, ok := balances.b[addr]; ok {
+		balances.b[addr] = ad
+		return nil
+	}
+	balances.b[addr] = basics.AccountData{}
```
algorandskiy, line 122: why not `balances.b[addr] = ad` unconditionally?

algorandskiy (commented):
Looks ok, couple minor notes

daemon/algod/api/server/v1/handlers/handlers.go
algorotem: Some partners still use it

algorandskiy (approved):

algorandskiy (approved):

tsachiherman (approved):

ledger/apply/mockBalances.go
```diff
@@ -0,0 +1,84 @@
...
+package apply
+
+import (
+	"github.com/algorand/go-algorand/config"
+	"github.com/algorand/go-algorand/data/basics"
+	"github.com/algorand/go-algorand/data/transactions/logic"
+	"github.com/algorand/go-algorand/protocol"
+)
+
+type mockBalances struct {
```
winder, line 26: Could you add some comments describing how to use this tool?

Also, you moved it out of the `test` package, do we have some sort of testutils spot to put things like this instead of putting test code into release packages?

winder (commented):
I realize that this has already been merged, but I think there are a couple more changes needed.

Overall, the changes LGTM.

ledger/apply/mockBalances.go
algorotem: Not sure about the "tool" side of it. It mocks the "real" balances which are well documented. 

I didn't. It was in `payment_test.go` and used by the other test files in the `apply` directory.  - I mainly put it in its own file and made it a bit more generic that it would be easier to use and clear that it doesn't belong to a specific test suite.

daemon/algod/api/server/v2/handlers.go
```diff
@@ -467,6 +468,7 @@ func (v2 *Handlers) PendingTransactionInformation(ctx echo.Context, txid string,
 		response.ConfirmedRound = &r
 
 		response.ClosingAmount = &txn.ApplyData.ClosingAmount.Raw
+		response.AssetClosingAmount = &txn.ApplyData.AssetClosingAmount
```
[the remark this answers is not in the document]
algorotem: Yes, `Raw` is a member of `MicroAlgo`. Assets amount are `uint64` so we don't need it there.

Interesting, @tsachiherman and I talked about it but from a quick check, it seemed like it would remove it. Yes, if this is the case we will push a quick patch.

ledger/apply/mockBalances.go
tsachiherman: I think that I agree with @winder here. it's perfectly fine to move it to it's own file, but unless we're using it from outside the package, we should keep it as a _test.go file.

PR #1887, opened 2021-02-02
assert in teal v3

jannotti:
This a first pass on the simplest new opcode, `assert`.  Would like to get feedback on whether I'm approaching backward compatibility properly.


tsachiherman:
try to rebase with master; your PR includes multiple files that are unrelated and already merged in.

jannotti:
> try to rebase with master; your PR includes multiple files that are unrelated and already merged in.

I agree with the idea, but I don't think it's going to be possible over time on a long-lived feature branch. Unless I misunderstand, I shouldn't be rebasing commits that have become "public" by being pushed to feature/teal3, as that rewrites git history in a way that can affect others.

I probably could have done it for the first commit, the merge commits seem inevitable beyond that.

tsachiherman:
> > try to rebase with master; your PR includes multiple files that are unrelated and already merged in.
> 
> I agree with the idea, but I don't think it's going to be possible over time on a long-lived feature branch. Unless I misunderstand, I shouldn't be rebasing commits that have become "public" by being pushed to feature/teal3, as that rewrites git history in a way that can affect others.
> 
> I probably could have done it for the first commit, the merge commits seem inevitable beyond that.

You're correct saying that it might not be the best solution for a long-lived branch. I'm going to defer that to @algojohnlee to suggest the best git management strategies. The thing is that at this time, it's pretty easy to see what's going on on that branch, since it's not that different from master. However, as both are going to change over time, it would become much harder to review the changes.

data/transactions/logic/doc.go
```diff
@@ -111,6 +111,7 @@ var opDocList = []stringString{
 	{"app_global_del", "delete key A from a global state of the current application"},
 	{"asset_holding_get", "read from account specified by Txn.Accounts[A] and asset B holding field X (imm arg) => {0 or 1 (top), value}"},
 	{"asset_params_get", "read from asset Txn.ForeignAssets[A] params field X (imm arg) => {0 or 1 (top), value}"},
+	{"assert", "immediately fail unless value X is not zero"},
```
brianolson, line 114: 'fail unless X is a non-zero number'
it will also fail if X is a byte string

brianolson (commented):
I'm not sure what's going on with the branching, but the assert opcode looks okay.

onetechnical:
> > > try to rebase with master; your PR includes multiple files that are unrelated and already merged in.
> > 
> > 
> > I agree with the idea, but I don't think it's going to be possible over time on a long-lived feature branch. Unless I misunderstand, I shouldn't be rebasing commits that have become "public" by being pushed to feature/teal3, as that rewrites git history in a way that can affect others.
> > I probably could have done it for the first commit, the merge commits seem inevitable beyond that.
> 
> You're correct saying that it might not be the best solution for a long-lived branch. I'm going to defer that to @algojohnlee to suggest the best git management strategies. The thing is that at this time, it's pretty easy to see what's going on on that branch, since it's not that different from master. However, as both are going to change over time, it would become much harder to review the changes.

This is true, but jannotti is going to periodically re-merge master into feature/teal3, until we're ready to merge all to master and release.

jannotti:
Perhaps my best bet is to try to separate those unavoidable merge commits and my normal changes. I'll make a PR for merging mainline to the branch which I hope I can get almost automatic "review" for. Maybe I'll only do that one a week. Then I'll work exclusively on teal3 oriented changes and make PRs that only include those changes. Those PRs will be more worthwhile to look at. I'd LOVE to avoid bothering to ask for reviews of the periodic merge commits, but we were unable to set up the feature branch so that I would have authority to merge without review.

For this PR it would be helpful to look past the first few files, which do indeed represent merges, and let me know if the approach to adding teal functionality send correct.

data/transactions/logic/assembler_test.go
```diff
@@ -1067,21 +1104,21 @@ func TestDisassembleTxna(t *testing.T) {
 func TestDisassembleGtxna(t *testing.T) {
 	t.Parallel()
 	// check gtxn and gtxna are properly disassembled
-	gtxnSample := "// version 2\ngtxn 0 Sender\n"
+	gtxnSample := fmt.Sprintf("// version %d\ngtxn 0 Sender\n", AssemblerMaxVersion)
```
tsachiherman, line 1107: The changes that you've made to TestDisassembleGtxna aren't wrong. however, once we start having more version of TEAL, we also need to expand these testings. In particular, I think that we have bunch of  TestDisassembleXXX that formats a string to update it's version, assemble, disassemble and compare the disassembled string vs. the constructed one.
I think that this "pattern" can be refactored to as a data-driven test cases, and we should also check it across multiple assembler versions (i.e. 2, 3, etc. ) and not just the latest one.

data/transactions/logic/assembler_test.go
```diff
@@ -869,21 +901,26 @@ func TestAssembleDisassembleCycle(t *testing.T) {
...
-		2: bigTestAssembleNonsenseProgram,
-		1: bigTestAssembleNonsenseProgram[:strings.Index(bigTestAssembleNonsenseProgram, "dup2")],
+		1: v1Nonsense,
+		2: v1Nonsense + v2Nonsense,
+		3: v1Nonsense + v2Nonsense + v3Nonsense,
 	}
 
+	// This confirms that each program compiles to the same bytes
+	// (except the leading version indicator), when compiled under
+	// oriringal and max versions. That doesn't *have* to be true,
```
tsachiherman, line 911: typo: oriringal -> original

data/transactions/logic/assembler_test.go
jannotti: Good idea. Disassembly is functionality I haven't seen in actual use. Are there strong requirements about what it looks like, and being able to map back and forth to disassembled form? Or is it just to display while the debugger walks through your code?

data/transactions/logic/assembler_test.go
```diff
@@ -228,14 +235,17 @@ func TestAssemble(t *testing.T) {
...
-			!strings.HasPrefix(spec.Name, "arg") {
+			!strings.HasPrefix(spec.Name, "arg") &&
+			spec.Version <= 2 {
 			t.Errorf("test should contain op %v", spec.Name)
 		}
 	}
-	ops, err := AssembleStringWithVersion(bigTestAssembleNonsenseProgram, AssemblerMaxVersion)
+	// First, we test v2, not AssemblerMaxVersion. A higher version is
+	// allowed to differ (and must, in the first byte).
+	ops, err := AssembleStringWithVersion(v1Nonsense+v2Nonsense, 2)
```
tsachiherman, line 248: I think that this type of test would be better implemented with a TEAL fuzzer, creating random TEAL programs, assembling and disassembling them. ( the rationale here is that the nonsense programs you've picked here covers only small number of opcodes. we need to check all the opcodes being assembled/disassembled )

data/transactions/logic/assembler.go
```diff
@@ -1117,7 +1117,13 @@ func (ops *OpStream) assemble(fin io.Reader) error {
 			ops.createLabel(opstring[:len(opstring)-1])
 			continue
 		}
-		ops.errorf("unknown opcode: %v", opstring)
+		// unknown opcode, let's report a good error if version problem
+		spec, ok = opsByName[AssemblerMaxVersion][opstring]
+		if ok {
+			ops.errorf("%v opcode was introduced in TEAL v%d", opstring, spec.Version)
```
tsachiherman, line 1123: nit: consider replacing `%v` with `%s`

data/transactions/logic/assembler_test.go
brianolson: I think disassembly should be able to go into assembly and produce the same binary.

data/transactions/logic/opcodes.go
```diff
@@ -164,6 +164,8 @@ var OpSpecs = []OpSpec{
 
 	{0x70, "asset_holding_get", opAssetHoldingGet, assembleAssetHolding, disAssetHolding, twoInts, oneAny.plus(oneInt), 2, runModeApplication, opSize{1, 2, nil}},
 	{0x71, "asset_params_get", opAssetParamsGet, assembleAssetParams, disAssetParams, oneInt, oneAny.plus(oneInt), 2, runModeApplication, opSize{1, 2, nil}},
+
+	{0x72, "assert", opAssert, asmDefault, disDefault, oneInt, nil, 3, modeAny, opSizeDefault},
```
tsachiherman, line 168: I find this opcode to be redundant, as it's a synonymous to the byte code of `bnz [offset=100000]`, and I don't see massive usage of this command either. I would think that making it into a semantic sugar, or a macro would make more sense since the "saving" here are 2 bytes for a command that rarely used. Also keep in mind that since teal is a stack-language, the common pattern is to accumulate bunch of condition on the stack, and to make multiple && at the end. ( which are 1 byte long instruction, naturally )

If you do believe that this is an important command, I won't be blocking this as I don't see any real harm here other than "wasting" the instructions space.

tsachiherman: to me, it looks like the `assert` is a better fit for a functional programming rather then stack-based languages.

brianolson: The `bnz Infinity` option is an important point. It's possible that `assert` could be implemented as a psuedo-op in the assembler but we don't need an actual opcode for it.

data/transactions/logic/assembler_test.go
jannotti: Then I should change the version comment to #pragma version?

data/transactions/logic/assembler_test.go
jannotti: Agreed that fuzzer would be a nice addition.  But this test is better than you think - it actually does require that every opcode that is available actually appears. That was @brianolson , I assume?  Anyway, point taken that adding fuzzing would be nice, but I like this test as at least a quick sanity that every opcode works in at least one simple use case.

data/transactions/logic/opcodes.go
jannotti: While `assert` is definitely not the most important opcode in the world (I only did it first because it was so simple, and want to check that I'm doing all the right things for addition) it has two important properties that I think make it a clear win.

1) It allows a TEAL programmer to express predicates *locally*.  They can avoid the && chain pattern, which is is error-prone.  You must skip the && on your first check, but nowhere else.  It runs all of your code when you could have exited early. You can't reliably take a predicate check clause and insert it somewhere else.  Once you've messed up the chain, even when the assembler catches your typing error (which it can only do if you've avoided jumps in your entire program), it is hard to figure out what the assembler is actually complaining about.

2) `assert` is just as compact as the && pattern (one byte per condition checked), so programmers have no reason to avoid it.  TEAL users do run into our resource limits.  If we given them a 3-byte alternative (jnz Infinity pseudo-op) they will avoid it, and continue to use the error-prone approach.

tsachiherman: I agree with both of your observations.

data/transactions/logic/assembler_test.go
brianolson: Yes! That's something we missed when `#pragma version` was added.

jannotti:
I'd like to move this one into feature/teal3 with the feedback I've got so far, then do something similar for a one or two new fields.

jannotti:
I'm going to close this, merge mainline, then prepare a new one to apply cleanly.

PR #1891, opened 2021-02-03
Fix omitempty for AssetCloseAmount

algorotem:
## Summary

As a fix to #1886, now API v1 omits the `closetoamount` in case of the 0 value.


winder (approved):
LGTM

tsachiherman (approved):

PR #1894, opened 2021-02-05
Cleanup github templates.

winder:
## Summary

Attempt to make the GitHub integration a little friendlier.

1. There is a new links feature. Rather than using a comment in the questions template, give a link to the community forum.
2. There is a new security feature. Rather than using a comment in each template, give a link to the vulnerability disclosure page.
3. Use headers consistently instead of bold in some.
4. Comment out all descriptions, this way you don't have to delete anything when creating a new issue.

You can see most of these features in action on the react project:
https://github.com/facebook/react/issues/new/choose


brianolson:
Generally I like it. I wish there was a way to preview the result.

How prominent is the forum link? Should we have that _and_ links in the issue templates to redirect people away from issues to the forums? If someone thinks "I'll make an issue" and doesn't look for the forum link, it might be good if the issue template was a second chance for them to notice the forum link.

winder:
@brianolson take a look at the description link, I would say that the forum is equally prominent to the existing choices and the security-report link is more prominent.

onetechnical (approved):

PR #1895, opened 2021-02-05
goal: avoid creating invalid consensus.json file

tsachiherman:
## Summary

Using `goal network create` would create a network directories containing a `consensus.json` file. That file would contain the `null` string.

The above file should not have been created. In case no consensus protocol changes are needed for a given network, the expected behavior is of no `consensus.json` file being generated.

## Test Plan

Add a unit test.

### DIFF
--- a/config/consensus.go
+++ b/config/consensus.go
@@ -433,9 +433,19 @@ func checkSetAllocBounds(p ConsensusParams) {
 }
 
 // SaveConfigurableConsensus saves the configurable protocols file to the provided data directory.
+// if the params contains zero protocols, the existing consensus.json file will be removed if exists.
 func SaveConfigurableConsensus(dataDirectory string, params ConsensusProtocols) error {
 	consensusProtocolPath := filepath.Join(dataDirectory, ConfigurableConsensusProtocolsFilename)
 
+	if len(params) == 0 {
+		// we have no consensus params to write. In this case, just delete the existing file
+		// ( if any )
+		err := os.Remove(consensusProtocolPath)
+		if os.IsNotExist(err) {
+			return nil
+		}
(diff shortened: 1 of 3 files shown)


config/consensus.go
```diff
@@ -433,9 +433,19 @@ func checkSetAllocBounds(p ConsensusParams) {
...
 
 // SaveConfigurableConsensus saves the configurable protocols file to the provided data directory.
+// if the params contains zero protocols, the existing consensus.json file will be removed if exists.
 func SaveConfigurableConsensus(dataDirectory string, params ConsensusProtocols) error {
 	consensusProtocolPath := filepath.Join(dataDirectory, ConfigurableConsensusProtocolsFilename)
 
+	if len(params) == 0 {
+		// we have no consensus params to write. In this case, just delete the existing file
+		// ( if any )
+		err := os.Remove(consensusProtocolPath)
```
algonautshant, line 443: This is imposing additional requirements to the directory permissions. 
Before this change, if the user had no directory write permissions and the file existed, still could write the file. But now, will error.
Instead, reset the file to empty file irrespective of len(params), then write the file if params are there.
If the params are empty, then you get an empty file. 
If there are no write permissions to the directory, then if the file exists, will not fail. If the file does not exist, then it will fail anyway.

tsachiherman: if there is an error as a result of this operation, then the error would be propagated upstream and would be prompted to the end user via "goal". The user could, then, correct the permission issues.

The user being unable to write to the data directory is a serious issue that would prevent other functions of working correctly, regardless of this issue - so we'd better surface that error.

algonautshant (approved):
Looks good.

PR #1896, opened 2021-02-05
agreement: FilterTimeout is using binary ConsensusCurrentVersion instead of current protocol version

tsachiherman:
## Summary

The agreement mainLoop function was bootstrapping the Deadline with the parameters based on the `protocol.ConsensusCurrentVersion`. This might be incorrect when the binary supports protocols that haven't (yet) been adopted by the network.

## Test Plan

Added unit test.


nicholasguoalgorand (approved):
Looks good to me

PR #1898, opened 2021-02-06
network: refactor broadcastThread

tsachiherman:
## Summary

The broadcastThread implementation was sub-optimal, and this PR addresses the issues we had there -
1. The broadcastThread is now using a single broadcasting go-routine rather than 4.
2. The broadcastThread/innerBroadcast used to drop queued messages just because there are no current peers. Instead, it will hold off until there is an available peer before queuing up the messages to the peers.
3. The peers array is being updated only if there is enough "wait time" between consecutive messages. In case of a high message burst, the peers would not get updated to avoid taking the peers lock.
4. The node's broadcast call was replaced with non-blocking to align with previous behavior.
5. The single thread implementation would ensure that queued messages would also be sent to the underlying peers at the order in which they were enqueued. This, in turn would ensure message id monotonicity.

## Test Plan

Unit tests updated, performance tests shown no regression.

### DIFF
--- a/network/wsNetwork.go
+++ b/network/wsNetwork.go
@@ -1174,13 +1174,42 @@ func (wn *WebsocketNetwork) broadcastThread() {
 		}
 	}
 
-	// waitForPeers waits until there is at least a single peer connected.
-	// it return true if there is at least one peer, or false if the network context has expired.
-	waitForPeers := func() bool {
+	// waitForPeers waits until there is at least a single peer connected or pending request expires.
+	// in any of the above two cases, it returns true.
+	// otherwise, false is returned ( the network context has expired )
+	waitForPeers := func(request *broadcastRequest) bool {
+		// waitSleepTime defines how long we'd like to sleep between consecutive tests that the peers list have been updated.
+		const waitSleepTime = 5 * time.Millisecond
+		// requestDeadline is the request deadline. If we surpass that deadline, the function returns true.
+		var requestDeadline time.Time
+		// sleepDuration is the current iteration sleep time.
+		var sleepDuration time.Duration
+		// initialize the requestDeadline if we have a request.
+		if request != nil {
+			requestDeadline = request.enqueueTime.Add(maxMessageQueueDuration)
+		} else {
+			sleepDuration = waitSleepTime
+		}
+
 		// wait until the we have at least a single peer connected.
 		for len(peers) == 0 {
+			// adjust the sleep time in case we have a request
+			if request != nil {
+				// we want to clamp the sleep time so that we won't sleep beyond the expiration of the request.
+				now := time.Now()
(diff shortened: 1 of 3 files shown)


network/wsNetwork.go
```diff
@@ -1165,36 +1163,129 @@ func (wn *WebsocketNetwork) sendFilterMessage(msg IncomingMessage) {
...
-	var peers []*wsPeer
+
 	slowWritingPeerCheckTicker := time.NewTicker(wn.slowWritingPeerMonitorInterval)
 	defer slowWritingPeerCheckTicker.Stop()
+	peers, lastPeersChangeCounter := wn.peerSnapshot([]*wsPeer{})
+	// updatePeers update the peers list if their peer change counter has changed.
+	updatePeers := func() {
+		if curPeersChangeCounter := atomic.LoadInt32(&wn.peersChangeCounter); curPeersChangeCounter != lastPeersChangeCounter {
+			peers, lastPeersChangeCounter = wn.peerSnapshot(peers[:0])
+			// clear out the unused portion of the peers array to allow the GC to cleanup unused peers.
```
brianolson, line 1174: Would it make sense to do the GC cleanup inside peerSnapshot() since we're already dealing with the len() and cap() of the passed in value there? Pass in not `peers[:0]` but plain `peers` and inside peerSnapshot() before writing new peer values, nil-out any excess slots if the new peers list is shorter.

brianolson (commented):
Generally I like it. One notion of code rearrangement but just aesthetic and could go either way.

network/wsNetwork.go
tsachiherman: I think that it's a great idea. I will implement that and update the review.

tsachiherman:
@nicholasguoalgorand, please take a look on this PR, as it's related to the project you're working on.

network/wsNetwork.go
```diff
@@ -2027,6 +2122,7 @@ func (wn *WebsocketNetwork) removePeer(peer *wsPeer, reason disconnectReason) {
 		if peer.throttledOutgoingConnection {
 			atomic.AddInt32(&wn.throttledOutgoingConnections, int32(1))
 		}
+		atomic.AddInt32(&wn.peersChangeCounter, 1)
```
algorandskiy, line 2125: I understand the idea of `peersChangeCounter` but not sure I like adding even more state to not the simplest code.
I'd suggest encapsulation but peers array is already heavily used around.

algorandskiy (approved):

network/wsNetwork.go
tsachiherman: Yes, it's not the simplest code, and I agree that we should see if it can be encapsulated somehow without negative performance impact.

PR #1899, opened 2021-02-08
testing: put some unicode in an AsssetName to help test Indexer

brianolson:
## Summary

In development Indexer got stuck on mainnet data similar to this.

## Test Plan

This is algod test code that should continue to pass.

### DIFF
--- a/test/scripts/e2e_subs/asset-misc.sh
+++ b/test/scripts/e2e_subs/asset-misc.sh
@@ -16,7 +16,7 @@ ACCOUNTB=$(${gcmd} account new|awk '{ print $6 }')
 ACCOUNTC=$(${gcmd} account new|awk '{ print $6 }')
 ACCOUNTD=$(${gcmd} account new|awk '{ print $6 }')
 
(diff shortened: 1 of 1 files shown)


test/scripts/e2e_subs/asset-misc.sh
```diff
@@ -16,7 +16,9 @@ ACCOUNTB=$(${gcmd} account new|awk '{ print $6 }')
 ACCOUNTC=$(${gcmd} account new|awk '{ print $6 }')
 ACCOUNTD=$(${gcmd} account new|awk '{ print $6 }')
 
-${gcmd} asset create --creator ${ACCOUNT} --name asset-misc --unitname amisc --total 1000000000000
+ASSET_NAME='Birlot : décollage'
```
tsachiherman, line 19: The change is good, however, I would suggest pushing it to the limits : 
use the following asset name
```
Birlot : décollage vs. ࠶🦪1
```
which is 32 bytes when encoded using utf-8

PR #1900, opened 2021-02-09
bugfix: prevent misleading agreement error message

tsachiherman:
## Summary

The agreement was printing out the following error message:
```
 (1) unable to retrieve consensus version for round 8764517, defaulting to binary consensus version
```

The culprit was a recent change to the [agreement initialization code](https://github.com/algorand/go-algorand/pull/1896) that would try to use the consensus version of the next round in order to find out some of the agreement runtime parameters.

This PR improves the data.Ledger.ConsensusVersion logic, allowing it to "predict" future protocol versions, based on known and previously agreed upon block headers.

## Test Plan

Unit test added.

### DIFF
--- a/data/basics/units.go
+++ b/data/basics/units.go
@@ -121,7 +121,7 @@ func OneTimeIDForRound(round Round, keyDilution uint64) crypto.OneTimeSignatureI
 	}
 }
 
-// SubSaturate subtracts two rounds with saturation arithmetic that does not
+// SubSaturate subtracts x rounds with saturation arithmetic that does not
 // wrap around past zero, and instead returns 0 on underflow.
 func (round Round) SubSaturate(x Round) Round {
 	if round < x {
--- a/data/ledger.go
+++ b/data/ledger.go
@@ -308,15 +308,48 @@ func (l *Ledger) ConsensusParams(r basics.Round) (config.ConsensusParams, error)
 }
 
 // ConsensusVersion gives the consensus version agreed on in a given round,
-// returning an error if we don't have that round or we have an
-// I/O error.
+// returning an error if the consensus version could not be figured using
+// either the block header for the given round, or the latest block header.
 // Implements agreement.Ledger.ConsensusVersion
 func (l *Ledger) ConsensusVersion(r basics.Round) (protocol.ConsensusVersion, error) {
 	blockhdr, err := l.BlockHdr(r)
-	if err != nil {
+	if err == nil {
+		return blockhdr.UpgradeState.CurrentProtocol, nil
+	}
+	// try to see if we can figure out what the version would be.
+	latestRound := l.Latest()
+	// if the request round was for an older round, then just say the we don't know.
+	if r < latestRound {
 		return "", err
 	}
-	return blockhdr.UpgradeState.CurrentProtocol, nil
+	// the request was for a future round. See if we have any known plans for the next round.
+	latestBlockhdr, err := l.BlockHdr(latestRound)
+	// if we have the lastest block header, look inside and try to figure out if we can deduce the
+	// protocol version for the given round.
+	if err == nil {
+		// check to see if we have a protocol upgrade.
+		if latestBlockhdr.NextProtocolSwitchOn == 0 {
+			// no protocol upgrade taking place, which implies the next round has the same version as the current one.
+			if r == (latestBlockhdr.Round + 1) {
+				return latestBlockhdr.CurrentProtocol, nil
+			}
+			// otherwise, we can't really tell.
+			return "", ledgercore.ErrNoEntry{Round: r, Latest: latestRound, Committed: latestRound}
+		}
+		// in this case, we do have a protocol upgrade taking place.
+		if r < latestBlockhdr.NextProtocolSwitchOn {
+			// if we're in the voting duration or uprade waiting period, then the protocol version is the current version.
+			return latestBlockhdr.CurrentProtocol, nil
+		}
+		// if the requested round aligns with the protocol version switch version and we've passed the voting period, then we know that on the switching round
+		// we will be using the next protocol.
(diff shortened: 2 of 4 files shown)


data/basics/units.go
```diff
@@ -121,7 +121,7 @@ func OneTimeIDForRound(round Round, keyDilution uint64) crypto.OneTimeSignatureI
 	}
 }
 
-// SubSaturate subtracts two rounds with saturation arithmetic that does not
+// SubSaturate subtracts x rounds with saturation arithmetic that does not
```
tsachiherman, line 124: this is a documentation bug..

data/ledger.go
```diff
@@ -308,15 +308,48 @@ func (l *Ledger) ConsensusParams(r basics.Round) (config.ConsensusParams, error)
...
-	return blockhdr.UpgradeState.CurrentProtocol, nil
+	// the request was for a future round. See if we have any known plans for the next round.
+	latestBlockhdr, err := l.BlockHdr(latestRound)
+	// if we have the lastest block header, look inside and try to figure out if we can deduce the
+	// protocol version for the given round.
+	if err == nil {
+		// check to see if we have a protocol upgrade.
+		if latestBlockhdr.NextProtocolSwitchOn == 0 {
+			// no protocol upgrade taking place, which implies the next round has the same version as the current one.
+			if r == (latestBlockhdr.Round + 1) {
```
algonautshant, line 333: Is there a minimum rounds in the future that a protocol upgrade can be scheduled?
If so, then here that offset can be used instead of the + 1.

data/ledger.go
```diff
@@ -308,15 +308,48 @@ func (l *Ledger) ConsensusParams(r basics.Round) (config.ConsensusParams, error)
...
+			return "", ledgercore.ErrNoEntry{Round: r, Latest: latestRound, Committed: latestRound}
+		}
+		// in this case, we do have a protocol upgrade taking place.
+		if r < latestBlockhdr.NextProtocolSwitchOn {
+			// if we're in the voting duration or uprade waiting period, then the protocol version is the current version.
+			return latestBlockhdr.CurrentProtocol, nil
+		}
+		// if the requested round aligns with the protocol version switch version and we've passed the voting period, then we know that on the switching round
+		// we will be using the next protocol.
+		if r == latestBlockhdr.NextProtocolSwitchOn && latestBlockhdr.Round >= latestBlockhdr.NextProtocolVoteBefore {
```
algonautshant, line 346: Why latestBlockhdr.Round >= latestBlockhdr.NextProtocolVoteBefore?
Does NextProtocolVoteBefore mean that at that round, there is still voting, which might be voted down?
Shouldn't it be '>' instead of '>='?

data/ledger.go
```diff
@@ -308,15 +308,48 @@ func (l *Ledger) ConsensusParams(r basics.Round) (config.ConsensusParams, error)
...
 // Implements agreement.Ledger.ConsensusVersion
 func (l *Ledger) ConsensusVersion(r basics.Round) (protocol.ConsensusVersion, error) {
 	blockhdr, err := l.BlockHdr(r)
-	if err != nil {
+	if err == nil {
+		return blockhdr.UpgradeState.CurrentProtocol, nil
+	}
+	// try to see if we can figure out what the version would be.
+	latestRound := l.Latest()
+	// if the request round was for an older round, then just say the we don't know.
```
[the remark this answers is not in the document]
tsachiherman: exactly.

data/ledger.go
tsachiherman: The `NextProtocolVoteBefore` if I understand correct is the threshold. At that round we already know if the new protocol was adopted or not. ( i.e. since the counting start at 0 )

data/ledger.go
tsachiherman: yes, I believe that you're correct. We can use the voting period of the current protocol to guide us how many rounds would guarantee to have the same version. I'll make the change.

tsachiherman: updated.

data/ledger.go
algonautshant: I agree. One more point. Should we also take into account MinUpgradeWaitRounds?

data/ledger.go
```diff
@@ -308,15 +308,53 @@ func (l *Ledger) ConsensusParams(r basics.Round) (config.ConsensusParams, error)
...
+	if err == nil {
+		// check to see if we have a protocol upgrade.
+		if latestBlockhdr.NextProtocolSwitchOn == 0 {
+			// no protocol upgrade taking place, we have *at least* UpgradeVoteRounds before the protocol version would get changed.
+			// it's safe to ignore the error case here since we know that we couldn't reached to this "known" round
+			// without having the binary supporting this protocol version.
+			currentConsensusParams, _ := config.Consensus[latestBlockhdr.CurrentProtocol]
+			// we're using <= here since there is no current upgrade on this round, and if there will be one on the subsequent round
+			// it would still be correct until (latestBlockhdr.Round + currentConsensusParams.UpgradeVoteRounds)
+			if r <= latestBlockhdr.Round+basics.Round(currentConsensusParams.UpgradeVoteRounds) {
```
algonautshant, line 338: Should MinUpgradeWaitRounds also be taken into account here? i.e.
currentConsensusParams.UpgradeVoteRounds + MinUpgradeWaitRounds?
It is probably not important, I can't think of why one would check too much into the future.

tsachiherman: I don't think that it really matter, and if we go this path it's becoming really complicated.
You see, in order to figure that out, we need to know which protocol we're going to upgrade *to* ( and there could be several ! ). Then, depending on the upgrade path, we find what the delay is. If the delay is zero, we're going to default to `DefaultUpgradeWaitRounds`, otherwise, the upgrade is bounded to [min..max].
I think that if we will need it in the future, we will always be able to implement it. Right now, we just need the `current round + 1`

algonautshant (approved):
Looks great.
Thanks for the updates and the explanations.

PR #1903, opened 2021-02-11
Remove the network v1 fetcher via websocket connections.

algonautshant:
## Summary

Remove the network v1 fetcher service via websocket connections.

## Test Plan

This is only removing unused code. The existing tests should be sufficient.

### DIFF
--- a/test/e2e-go/features/catchup/basicCatchup_test.go
+++ b/test/e2e-go/features/catchup/basicCatchup_test.go
@@ -25,6 +25,7 @@ import (
 	"github.com/stretchr/testify/require"
 
 	"github.com/algorand/go-algorand/config"
+	"github.com/algorand/go-algorand/network"
 	"github.com/algorand/go-algorand/protocol"
 	"github.com/algorand/go-algorand/test/framework/fixtures"
 )
@@ -76,13 +77,22 @@ func TestBasicCatchup(t *testing.T) {
 // The current versions are the original v1 and the upgraded to v2.1
 func TestCatchupOverGossip(t *testing.T) {
 	t.Parallel()
+
+	versions := network.SupportedProtocolVersions
+	require.LessOrEqual(t, len(versions), 3)
+
 	// ledger node upgraded version, fetcher node upgraded version
-	runCatchupOverGossip(t, false, false)
+	for i := 0; i < len(versions); i++ {
+		runCatchupOverGossip(t, "", "")
+		runCatchupOverGossip(t, versions[i], "")
+		runCatchupOverGossip(t, "", versions[i])
+		runCatchupOverGossip(t, versions[i], versions[i])
+	}
 }
 
 func runCatchupOverGossip(t *testing.T,
-	ledgerNodeDowngrade,
-	fetcherNodeDowngrade bool) {
+	ledgerNodeDowngradeTo,
+	fetcherNodeDowngradeTo string) {
 
 	if testing.Short() {
 		t.Skip()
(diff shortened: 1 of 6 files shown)


tsachiherman:
take a look on `/go-algorand/test/e2e-go/features/catchup/basicCatchup_test.go`; it needs to be updated as well.

test/e2e-go/features/catchup/basicCatchup_test.go
```diff
@@ -78,12 +78,6 @@ func TestCatchupOverGossip(t *testing.T) {
 	t.Parallel()
 	// ledger node upgraded version, fetcher node upgraded version
 	runCatchupOverGossip(t, false, false)
-	// ledger node older version, fetcher node upgraded version
-	runCatchupOverGossip(t, true, false)
-	// ledger node upgraded older version, fetcher node older version
-	runCatchupOverGossip(t, false, true)
-	// ledger node older version, fetcher node older version
-	runCatchupOverGossip(t, true, true)
```
tsachiherman, line 86: I think that I have a better proposal for this test - 
improve it so that it would know how to read the list of SupportedProtocolVersions and dynamically use these.

The motivation here is that I expect to have another network version soon, and this test seems to be a good test case for that.

tsachiherman: If the list contains just a single protocol then we can just execute runCatchupOverGossip with (false, false)

test/e2e-go/features/catchup/basicCatchup_test.go
```diff
@@ -76,19 +77,22 @@ func TestBasicCatchup(t *testing.T) {
...
+
 	// ledger node upgraded version, fetcher node upgraded version
-	runCatchupOverGossip(t, false, false)
-	// ledger node older version, fetcher node upgraded version
-	runCatchupOverGossip(t, true, false)
-	// ledger node upgraded older version, fetcher node older version
-	runCatchupOverGossip(t, false, true)
-	// ledger node older version, fetcher node older version
-	runCatchupOverGossip(t, true, true)
+	for i := 0; i < len(versions); i++ {
```
algonautshant, line 85: This looks redundant to go over the more recent version (i.e. no downgrade). 
Index 0 is iterator over here to test the code path which loads the version from the configuration file instead of just relying on the default. Moreover, the configuration might have the version set to other than the most recent supported version.

test/e2e-go/features/catchup/basicCatchup_test.go
```diff
@@ -76,19 +77,22 @@ func TestBasicCatchup(t *testing.T) {
...
+		runCatchupOverGossip(t, "", "")
```
tsachiherman, line 86: This line doesnt need to be within the loop

test/e2e-go/features/catchup/basicCatchup_test.go
```diff
@@ -76,19 +77,22 @@ func TestBasicCatchup(t *testing.T) {
...
+		runCatchupOverGossip(t, versions[i], "")
```
tsachiherman, line 87: You can skip a loop iteration if the default protocol is identical to the loop protocol

tsachiherman (changes requested):
The test needs few more touches, but looks great otherwise.

test/e2e-go/features/catchup/basicCatchup_test.go
algonautshant: The default protocol is empty, which will be equivalent to setting it to SupportedProtocolVersions[0] or equivalent to not downgrading.
Skipping loop iteration 0 will serve this purpose.

test/e2e-go/features/catchup/basicCatchup_test.go
algonautshant: Will put this outside the loop, and replace it with:
`runCatchupOverGossip(t, supportedVersions[0], supportedV[0])`
The default value from the config file will be empty. So, the value will be picked up from supportedVersions index 0.

tsachiherman (approved):
looks good, thanks for the changes.

PR #1906, opened 2021-02-11
Numerous of the simpler teal3 additions. opcodes and fields.

jannotti:
## Summary 

Several new opcodes:
assert
min_balance
getbit
setbit
getbyte
setbyte
swap
select
dig
stxn
stxna
pushbytes
pushint

Several new txn fields available:
ForeignAssets
NumForeignAssets
ForeignApps
NumForeignApps
GlobalStateInts
GlobalStateByteslices
LocalStateInts
LocalStateByteslices
LogicArgs
NumLogicArgs

One new global field: CreatorAddress

## Test Plan

Still work to be done here, especially on how to manage the consensus upgrade.


data/transactions/logic/eval.go
```diff
@@ -1678,6 +1845,67 @@ func opSubstring3(cx *evalContext) {
 	cx.stack = cx.stack[:prev]
 }
 
+func opGetBit(cx *evalContext) {
```
tsachiherman, line 1848: if we're already adding get/set bit, can we add that functionality to support byte arrays as well ?

data/transactions/logic/opcodes.go
```diff
@@ -164,6 +166,28 @@ var OpSpecs = []OpSpec{
...
+
+	// There must be at least one thing on the stack for dig, but
+	// it would be nice if we did better checking than that.
+	{0x80, "dig", opDig, asmDefault, disDefault, oneAny, twoAny, 3, modeAny, opSize{1, 2, nil}},
+
+	// Like gtxn, but gets txn index from stack, rather than immediate arg
+	{0x81, "stxn", opStxn, assembleStxn, disTxn, oneInt, oneAny, 3, modeAny, opSize{1, 2, nil}},
+	{0x82, "stxna", opStxna, assembleStxna, disTxna, oneInt, oneAny, 3, modeAny, opSize{1, 3, nil}},
+
+	// Immediate bytes and ints. Smaller code size for single use of constant.
```
brianolson, line 188: It would be good for the assembler to help optimize this for TEAL 3.
`int` assembler text could become intc_N, intc, or pushint (`inti` ?) for most-used, less-used, used-once
same for `byte`

jannotti: I agree it should. But I'm going to have to delay that to the teal4 (which is the new name for the second half teal3 changes, so is coming by end of Q1).

data/transactions/logic/TEAL_opcodes.md
```diff
@@ -727,3 +741,110 @@ params: account index, asset id. Return: did_exist flag (1 if exist and 0 otherw
...
+- Opcode: 0x79
+- Pops: *... stack*, {any A}, {any B}, {uint64 C}
+- Pushes: any
+- selects one of two values to retain: A, B, C -> A ? B : C
+- LogicSigVersion >= 3
+
+## dig
+
+- Opcode: 0x80 {uint8 depth}
+- Pops: *... stack*, any
```
brianolson, line 813: I wonder if there's a better way to say this?
```
*... stack ...*, any, * ... stack ...*
*... stack ...*, any, * ... stack ...*, any
```

data/transactions/logic/TEAL_opcodes.md
```diff
@@ -727,3 +741,110 @@ params: account index, asset id. Return: did_exist flag (1 if exist and 0 otherw
...
+
+- Opcode: 0x82 {uint8 transaction field index}{uint8 transaction field array index}
+- Pops: *... stack*, uint64
+- Pushes: any
+- push value from an array field from transaction A in the current group
+- LogicSigVersion >= 3
+
+## pushbytes
+
+- Opcode: 0x83
```
brianolson, line 838: note the immediate data `{varuint length}{[length]byte data}`

data/transactions/logic/TEAL_opcodes.md
```diff
@@ -727,3 +741,110 @@ params: account index, asset id. Return: did_exist flag (1 if exist and 0 otherw
...
+- Pops: _None_
+- Pushes: []byte
+- push the following program bytes to the stack
+- LogicSigVersion >= 3
+
+## pushint
+
+- Opcode: 0x84
```
brianolson, line 846: note the immediate data after the opcode `{varuint data}`

data/transactions/logic/assembler.go
```diff
@@ -157,7 +157,6 @@ func (ops *OpStream) Intc(constIndex uint) {
 	} else {
 		ops.trace("intc %d %d", constIndex, ops.intc[constIndex])
 	}
-	ops.tpush(StackUint64)
```
brianolson, line 160: Does this happen somewhere else now?

data/transactions/logic/eval.go
```diff
@@ -1678,6 +1845,124 @@ func opSubstring3(cx *evalContext) {
...
+			return
+		}
+		byteVal := target.Bytes[byteIdx]
+
+		bitIdx := idx % 8
+		// We saying that bit 9 (the 10th bit), for example,
+		// is the 2nd bit in the second byte, and that "2nd
+		// bit" here means almost-highest-order bit, because
+		// we're thinking of the bits in the byte itself as
+		// being big endian. So this looks "reversed"
```
brianolson, line 1876: Okay yeah it's weird that uint64 is little-bit-endian but byte-array is the other way around. What if we treated uint64 as big endian [8]byte ? Or either way just consistent.

data/transactions/logic/assembler.go
jannotti: Yes, a whole bunch of opcodes were simplified by moving the type stack handling into checkArgs.  It used to only check that the stack was the right shape at opcode entry, but not it also performs the update for the shape of the stack when the opcode finishes. Since we now proceed after errors in order report more, it's quite nice that it does type stack management at the opcode start. Even if the opcode ends up erroring out - the type stack will still be updated so we won't get cascading errors as we continue assembling.  Only opcodes that return different types (field access mainly) need to do explicit management now.

data/transactions/logic/eval.go
jannotti: The trouble is that I just don't think you can get around the fact that people expect "bit 0" to mean "lower order bit" of an integer, and for "byte 0" (and therefore "bit 0") to mean the "left-most" byte (or bit) in a string.

I convinced myself it was reasonable. The direction depends on the type being indexed. I'm amenable to counter-arguments though.  Have a look at the unit test to be convinced though.  I think when you see the operations in action, they do what you expect.

data/transactions/logic/TEAL_opcodes.md
jannotti: Thanks, got it

data/transactions/logic/eval.go
jannotti: Here's what I find convincing
```
	// bit 10 is the 3rd bit (from the high end) in the second byte
	testAccepts(t, "byte 0xfff0; int 10; getbit; int 1; ==", 3)
	testAccepts(t, "byte 0xfff0; int 12; getbit; int 0; ==", 3)
```

You really want the 10 to see the "f" part of the third byte, and 12 to see the 0 part.

brianolson: ok but
```
int 0xfff0; int 12; getbit; int 0; ==
```
should do the same thing?

jannotti: I would say no.  I think that everyone has decided that integer bit indexing is based on math, rather than layout.  In C, whether you're on a big-endian or little-endian machine, 0xfff0 & (1 << 2),  is going to take a 0 bit from that low-order nybble.  I think that's what people think of as bit indexing.  That expression is "give me bit 2", and it yields the same thing on x86 and sparc.

data/transactions/logic/assembler.go
```diff
@@ -862,14 +827,61 @@ func assembleGtxna(ops *OpStream, spec *OpSpec, args []string) error {
...
+	fs, ok := txnFieldSpecByName[args[0]]
+	if !ok {
+		return ops.errorf("stxn unknown field: %v", args[0])
+	}
+	_, ok = txnaFieldSpecByField[fs.field]
+	if ok {
+		return ops.errorf("found gtxna field %v in gtxn op", args[0])
+	}
+	if fs.version > ops.Version {
+		return ops.errorf("gtxn %s available in version %d. Missed #pragma version?", args[0], fs.version)
```
brianolson, line 846: a couple missed gtxn -> stxn

data/transactions/logic/doc.go
```diff
@@ -161,11 +179,12 @@ var opDocExtraList = []stringString{
...
 	{"b", "See `bnz` for details on how branches work. `b` always jumps to the offset."},
 	{"intcblock", "`intcblock` loads following program bytes into an array of integer constants in the evaluator. These integer constants can be referred to by `intc` and `intc_*` which will push the value onto the stack. Subsequent calls to `intcblock` reset and replace the integer constants available to the script."},
-	{"bytecblock", "`bytecblock` loads the following program bytes into an array of byte string constants in the evaluator. These constants can be referred to by `bytec` and `bytec_*` which will push the value onto the stack. Subsequent calls to `bytecblock` reset and replace the bytes constants available to the script."},
+	{"bytecblock", "`bytecblock` loads the following program bytes into an array of byte-array constants in the evaluator. These constants can be referred to by `bytec` and `bytec_*` which will push the value onto the stack. Subsequent calls to `bytecblock` reset and replace the bytes constants available to the script."},
 	{"*", "Overflow is an error condition which halts execution and fails the transaction. Full precision is available from `mulw`."},
```
brianolson, line 183: would love to see some extended example doc of setbit in the opDocExtra

data/transactions/logic/eval_test.go
```diff
@@ -200,26 +208,14 @@ func TestWrongProtoVersion(t *testing.T) {
...
-			require.NoError(t, err)
-			require.True(t, cost < 1000)
-			var txn transactions.SignedTxn
-			txn.Lsig.Logic = ops.Program
-			pass, err := Eval(ops.Program, defaultEvalParams(nil, &txn))
-			require.True(t, pass)
-			require.NoError(t, err)
-		})
-	}
+	testAccepts(t, "int  2; int 3; + ;int  5;==", 1)
```
brianolson, line 213: nice. this improved framework doing more testing is great.

brianolson (commented):
almost there

data/transactions/logic/TEAL_opcodes.md
```diff
@@ -727,3 +741,110 @@ params: account index, asset id. Return: did_exist flag (1 if exist and 0 otherw
...
+- pop a target A (integer or byte-array), and index B. Pushes the Bth bit of A.
+- LogicSigVersion >= 3
+
+## setbit
+
+- Opcode: 0x75
+- Pops: *... stack*, {any A}, {uint64 B}, {uint64 C}
+- Pushes: uint64
+- pop a target A, index B, and bit C. Sets the Bth bit of A to C, and push the result
+- LogicSigVersion >= 3
```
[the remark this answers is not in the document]
jannotti: Now in unit test
```
	testAccepts(t, "byte 0x0000; int 3; int 1; setbit; byte 0x1000; ==", 3)
	testAccepts(t, "byte 0x0000; int 15; int 1; setbit; byte 0x0001; ==", 3)
	testAccepts(t, "int 0x0000; int 3; int 1; setbit; int 0x0008; ==", 3)
	testAccepts(t, "int 0x0000; int 12; int 1; setbit; int 0x1000; ==", 3)
```
I'll try to figure out where to put this in docs.  Do our website docs get built automatically later, by the way?  Or do I need to talk to Jason W?

data/transactions/logic/doc.go
jannotti: How about:
```
	{"getbit", "bit indexing begins with low-order bits in integers. Bit 4 is the 1 in int 0x0010. Indexing begins in the first bytes of a byte-string (as seen in getbyte and substring). Bits 0 through 11 are 1 in byte 0xfff000000000"},
	{"setbit", "see explanation of bit ordering in getbit"},
```

brianolson: I like putting the docs in setbit because starting with 0x0000 it's really clear to see what happens.

jannotti: How about this:
```
{"getbit", "see explanation of bit ordering in setbit"},
{"setbit", "bit indexing begins with low-order bits in integers. Setting bit 4 to 1 on the integer 0 yields 16 (`int 0x0010`, or 2^4). Indexing begins in the first bytes of a byte-string (as seen in getbyte and substring). Setting bits 0 through 11 to 1 in a 4 byte-array of 0s yields `byte 0xfff00000`"},
```

brianolson: Yup. This is unambiguous; thanks.

data/transactions/logic/assembler.go
```diff
@@ -1143,6 +1173,49 @@ func (ops *OpStream) assemble(fin io.Reader) error {
 	return nil
 }
 
+func (ops *OpStream) pragma(line string) error {
```
algorandskiy, line 1176: Is is the same as in master? maybe back merge master into teal3 to reduce the diff?

data/transactions/logic/assembler.go
```diff
@@ -909,43 +921,40 @@ func assembleAssetParams(ops *OpStream, spec *OpSpec, args []string) error {
...
 	ops.checkArgs(*spec)
-	if len(spec.Returns) > 0 {
-		ops.tpusha(spec.Returns)
-		ops.trace(" pushes(%s", spec.Returns[0].String())
-		if len(spec.Returns) > 1 {
-			for _, rt := range spec.Returns[1:] {
-				ops.trace(", %s", rt.String())
-			}
-		}
-		ops.trace(")")
```
algorandskiy, line 922: why did this go? meaning both type stack and trace messages

data/transactions/logic/opcodes.go
```diff
@@ -164,6 +168,28 @@ var OpSpecs = []OpSpec{
...
+	{0x74, "getbit", opGetBit, asmDefault, disDefault, anyInt, oneInt, 3, modeAny, opSizeDefault},
+	{0x75, "setbit", opSetBit, asmDefault, disDefault, anyIntInt, oneInt, 3, modeAny, opSizeDefault},
+	{0x76, "getbyte", opGetByte, asmDefault, disDefault, byteInt, oneInt, 3, modeAny, opSizeDefault},
+	{0x77, "setbyte", opSetByte, asmDefault, disDefault, byteIntInt, oneBytes, 3, modeAny, opSizeDefault},
+	{0x78, "swap", opSwap, asmDefault, disDefault, twoAny, twoAny, 3, modeAny, opSizeDefault},
+	{0x79, "select", opSelect, asmDefault, disDefault, twoAny.plus(oneInt), oneAny, 3, modeAny, opSizeDefault},
+
+	// There must be at least one thing on the stack for dig, but
+	// it would be nice if we did better checking than that.
+	{0x80, "dig", opDig, asmDefault, disDefault, oneAny, twoAny, 3, modeAny, opSize{1, 2, nil}},
```
algorandskiy, line 184: move to 0x4b, right after `dup2` op to preserve logical grouping

data/transactions/logic/opcodes.go
```diff
@@ -164,6 +168,28 @@ var OpSpecs = []OpSpec{
...
 	{0x70, "asset_holding_get", opAssetHoldingGet, assembleAssetHolding, disAssetHolding, twoInts, oneAny.plus(oneInt), 2, runModeApplication, opSize{1, 2, nil}},
 	{0x71, "asset_params_get", opAssetParamsGet, assembleAssetParams, disAssetParams, oneInt, oneAny.plus(oneInt), 2, runModeApplication, opSize{1, 2, nil}},
+
+	{0x72, "assert", opAssert, asmDefault, disDefault, oneInt, nil, 3, modeAny, opSizeDefault},
+	{0x73, "min_balance", opMinBalance, asmDefault, disDefault, oneInt, oneInt, 3, runModeApplication, opSizeDefault},
+
+	{0x74, "getbit", opGetBit, asmDefault, disDefault, anyInt, oneInt, 3, modeAny, opSizeDefault},
+	{0x75, "setbit", opSetBit, asmDefault, disDefault, anyIntInt, oneInt, 3, modeAny, opSizeDefault},
+	{0x76, "getbyte", opGetByte, asmDefault, disDefault, byteInt, oneInt, 3, modeAny, opSizeDefault},
+	{0x77, "setbyte", opSetByte, asmDefault, disDefault, byteIntInt, oneBytes, 3, modeAny, opSizeDefault},
```
algorandskiy, line 178: move to 0x53 to preserve grouping by op type

data/transactions/logic/opcodes.go
```diff
@@ -164,6 +168,28 @@ var OpSpecs = []OpSpec{
...
+	{0x78, "swap", opSwap, asmDefault, disDefault, twoAny, twoAny, 3, modeAny, opSizeDefault},
+	{0x79, "select", opSelect, asmDefault, disDefault, twoAny.plus(oneInt), oneAny, 3, modeAny, opSizeDefault},
```
algorandskiy, line 180: move to 0x4d to preserve grouping by op type, right after `dig` = 0x4b

data/transactions/logic/opcodes.go
```diff
@@ -164,6 +168,28 @@ var OpSpecs = []OpSpec{
...
+
+	// There must be at least one thing on the stack for dig, but
+	// it would be nice if we did better checking than that.
+	{0x80, "dig", opDig, asmDefault, disDefault, oneAny, twoAny, 3, modeAny, opSize{1, 2, nil}},
+
+	// Like gtxn, but gets txn index from stack, rather than immediate arg
+	{0x81, "stxn", opStxn, assembleStxn, disTxn, oneInt, oneAny, 3, modeAny, opSize{1, 2, nil}},
+	{0x82, "stxna", opStxna, assembleStxna, disTxna, oneInt, oneAny, 3, modeAny, opSize{1, 3, nil}},
```
algorandskiy, line 188: move to 0x38 after `gtxna`

algorandskiy (commented):
Looks good.
Please 1) consider re-arranging opcodes by group as suggested.
2) merge back `master` into `teal3` (e2e fixes, pragma, etc) to reduce the diff and simplify review.

data/transactions/logic/TEAL_opcodes.md
```diff
@@ -727,3 +741,114 @@ params: account index, asset id. Return: did_exist flag (1 if exist and 0 otherw
...
+
+## dig
+
+- Opcode: 0x80 {uint8 depth}
+- Pops: *... stack*, any
+- Pushes: *... stack*, any, any
+- duplicate value N from the top of stack
+- LogicSigVersion >= 3
+
+## stxn
```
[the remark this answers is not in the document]
jannotti: That's true.  I meant it as "stack txn" or "slot txn", but we do use stxn for signed all over the place.  What do you think of the "ge" in getxn as meaning? group extended?

I considered adding an i, for indirect: gtxni/gtxnia but the trouble there is that "i" can also mean "immediate" which describes the normal gtxn exactly.  So "i" doesn't help much.

With substring, we have substring3, which I guess means that it takes 3 args on stack, unlike the normal one which only takes 1.  So I guess that would mean gtxn1/gtxna1 for this opcode.  But that's ugly, and gtxn2 isn't much better (and less principled).

What about atxn/atxna for "alternate"? It also it nice because we normally call the top-of-stack "A" when describing an opcode that takes one thing from the stack, so an alternate way to think of "atxn" is that it's txn that reads the index from "A".

data/transactions/logic/assembler.go
jannotti: It is.  I had to pull it in so that I could get access to the `make integration` fixes so I could right v24 tests.

data/transactions/logic/assembler.go
jannotti: It's done in checkArgs() now, so most opcodes don't have to do any extra work.  (Just opcodes that have alternate return types, like field accesses)

data/transactions/logic/opcodes.go
jannotti: Excellent. I didn't notice we had left gaps.

data/transactions/logic/opcodes.go
```diff
@@ -33,14 +33,52 @@ const rekeyingEnabledVersion = 2
...
-	cost      int
-	size      int
-	checkFunc opCheckFunc
+// opDetails records details such as non-standard costs, immediate
+// arguments, or dynamic layout controlled by a check function.
+type opDetails struct {
+	Cost       int
+	Size       int
+	checkFunc  opCheckFunc
+	Immediates []immediate
```
brianolson, line 42: consolodate with doc.go `opcodeImmediateNoteList`, either move those into opDetails.Immediates or move the new opDetails.Immediates data into opcodeImmediateNoteList

data/transactions/logic/assembler_test.go
```diff
@@ -1029,7 +1113,7 @@ func TestAssembleAsset(t *testing.T) {
 func TestDisassembleSingleOp(t *testing.T) {
 	t.Parallel()
 	// test ensures no double arg_0 entries in disassembly listing
-	sample := "// version 2\narg_0\n"
+	sample := fmt.Sprintf("// version %d\narg_0\n", AssemblerMaxVersion)
```
winder, line 1116: Should these tests run against v2 and v3 instead of just the latest?

data/transactions/logic/eval_test.go
```diff
@@ -3972,3 +3807,161 @@ func TestRekeyFailsOnOldVersion(t *testing.T) {
...
+	testRejects(t, "int 1;assert;int 0", 3)
+	testPanics(t, "int 0;assert;int 1", 3)
+	testPanics(t, obfuscate("assert;int 1"), 3)
+	testPanics(t, obfuscate(`byte "john";assert;int 1`), 3)
+}
+
+func TestBits(t *testing.T) {
+	t.Parallel()
+	testAccepts(t, "int 1; int 0; getbit; int 1; ==", 3)
+	testAccepts(t, "int 1; int 1; getbit; int 0; ==", 3)
```
algorandskiy, line 3880: how about a loop over a random number with say 24 out of 64 bits set, and checking `getbit` returns a correct value for each of 64 possible positions?

(same for bytes operand)

data/transactions/logic/eval_test.go
```diff
@@ -3972,3 +3807,161 @@ func TestRekeyFailsOnOldVersion(t *testing.T) {
...
+}
+
+func TestSelect(t *testing.T) {
+	t.Parallel()
+
+	testAccepts(t, "int 1; byte 0x1231; int 0; select", 3) // selects the 1
+	testRejects(t, "int 0; byte 0x1232; int 0; select", 3) // selects the 0
+
+	testAccepts(t, "int 0; int 1; int 1; select", 3)      // selects the 1
+	testPanics(t, "int 1; byte 0x1233; int 1; select", 3) // selects the bytes
```
algorandskiy, line 3933: add stack underflow test when only 3 ops out of 2 provided

algorandskiy (commented):
Looks good, a few notes and suggestions

data/transactions/logic/assembler_test.go
jannotti: Yes, thanks.  Adding in cleanup PR.

data/transactions/logic/opcodes.go
```diff
@@ -33,16 +33,54 @@ const rekeyingEnabledVersion = 2
...
+var opBranch = opDetails{1, 3, checkBranch, []immediate{{"target", immLabel}}}
 
-// OpSpec defines one byte opcode
+func costly(cost int) opDetails {
+	return opDetails{cost, 1, nil, nil}
+}
+
+func immediates(name string, rest ...string) opDetails {
+	num := 1 + len(rest)
+	immediates := make([]immediate, num, len(rest)+1)
```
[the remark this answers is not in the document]
jannotti: Thanks, fixing in cleanup PR.

data/transactions/logic/assembler.go
```diff
@@ -330,16 +353,16 @@ func (ops *OpStream) AssetParams(val uint64) {
...
 	}
-	ops.pending.WriteByte(opsByName[ops.Version]["asset_params_get"].Opcode)
+	ops.pending.WriteByte(OpsByName[ops.Version]["asset_params_get"].Opcode)
 	ops.pending.WriteByte(uint8(val))
 	ops.tpush(AssetParamsFieldTypes[val])
 	ops.tpush(StackUint64)
 }
 
 func assembleInt(ops *OpStream, spec *OpSpec, args []string) error {
+	ops.checkArgs(*spec)
```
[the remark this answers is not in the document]
jannotti: Excellent point.  I moved it into the the assembly loop and now we catch more mistakes (because some assembly routines were not calling it before.). It will be in a cleanup PR.

ledger/applications.go
```diff
@@ -74,6 +75,16 @@ func (al *logicLedger) Balance(addr basics.Address) (res basics.MicroAlgos, err
 	return record.MicroAlgos, nil
 }
 
+func (al *logicLedger) MinBalance(addr basics.Address, proto *config.ConsensusParams) (res basics.MicroAlgos, err error) {
+	// Fetch record with pending rewards applied
+	record, err := al.cow.Get(addr, true)
```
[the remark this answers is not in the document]
jannotti: Thanks will go into cleanup

tsachiherman:
do we still need this PR ?

PR #1909, opened 2021-02-13
WIP: 1000+ assets

algorandskiy:
## Summary

Implement asset holding saving into a separate table

## Test Plan

New tests added, most of new code covered.


ledger/accountdb.go
```diff
@@ -64,6 +66,9 @@ var accountsSchema = []string{
 	`CREATE TABLE IF NOT EXISTS accountbase (
 		address blob primary key,
 		data blob)`,
+	`CREATE TABLE IF NOT EXISTS accountext (
+		id integer primary key,
+		data blob)`,
```
tsachiherman, line 71: I think that we'll want this to be a database schema upgrade.

ledger/accountdb.go
```diff
@@ -114,17 +120,109 @@ var accountsResetExprs = []string{
...
+	// Holding "frozen" flag
+	// same number of elements as in AmountsAssetIndicesOffsets
+	Frozens []bool `codec:"f,allocbound=1048576"`
+}
+
+// ExtendedAssetHolding TODO
+type ExtendedAssetHolding struct {
+	_struct struct{}             `codec:",omitempty,omitemptyarray"`
+	Count   uint32               `codec:"c"`
+	Groups  []AssetsHoldingGroup `codec:"gs,allocbound=100"`
```
tsachiherman, line 177: I think that we'll want to allow higher number of allocbound here.

ledger/accountdb.go
```diff
@@ -245,42 +343,69 @@ func makeCompactAccountDeltas(accountDeltas []ledgercore.AccountDeltas, baseAcco
...
 	selectStmt, err := tx.Prepare("SELECT rowid, data FROM accountbase WHERE address=?")
 	if err != nil {
 		return
 	}
 	defer selectStmt.Close()
+	loadStmt, err := tx.Prepare("SELECT data FROM accountext WHERE id=?")
+	if err != nil {
+		return
+	}
+	defer loadStmt.Close()
```
tsachiherman, line 361: I think that lazy initialization here would be better since on the "normal case", I don't expect to see too many accounts that have many holdings.

jannotti:
Can we add rest APIs, along the lines of `/v2/accounts/{account_id}/holdings/{asset_id}` and `/v2/accounts/{account_id}/applications/{app_id}` (or perhaps v3?)

Currently, the /v2/accounts API returns all holdings, but that might be a very big object, so callers might prefer to ask for single sub-objects. (We might also want an option on `/v2/accounts/{account_id}&brief=true` that returns just the IDs of assets and apps, to keep things reasonably sized. Callers would use the above to get details where needed.

tsachiherman:
These are all good ideas, but they go abit outside of the scope of this PR.

I would like to see us completing this change first, and only then
extending out interfaces as you've suggested.

algorandskiy:
Replaced by #1932 and #1933 ,...

PR #1910, opened 2021-02-15
Remove GOPATH dependencies in integration tests

jannotti:
This allows running integration tests without a repo under GOPATH.

## Summary

Tests had been using GOPATH, but mostly just to find the top of the repo to
refer to tests.  This breaks if doing module oriented stuff outside of
GOPATH. These changes make it so that e2e.sh uses the full name of the
e2e test files, and e2e_client_runner executes the e2e scripts with
cwd set to top of repo.  That allowed removing the few places those
scripts used GOPATH, in favor of relative filenames.

## Test Plan

It *is* the tests.  Was able to run integration tests in my local development environment,
which was impossible without these changes, as I don't work under GOPATH.  That is the way
forward with Go, so we should get ready.  I will be checking if CI broke in any way by watching
it!


brianolson:
Looks good so far. Is this still 'draft'? I'd approve it.

brianolson (approved):

algorandskiy (approved):

PR #1914, opened 2021-02-16
Deprecate support of reverse catchup protocol ( part 2 ) 

algonautshant:
## Summary

This PR is the second iteration of network v1 cleanup
- Moving RequestBlock from WsFetcherService to wsFetcherClient since RequestBlock is the only used function of WsFetcherService
- removing wsFetcherService which is no longer used. No messages are sent with the tags:
   - UniEnsBlockResTag  Tag = "US" 
   - UniCatchupResTag   Tag = "UT"

## Test Plan

The existing tests are sufficient here because no no functionality is being added.


catchup/wsFetcher.go
```diff
@@ -168,3 +163,45 @@ func (w *wsFetcherClient) Close() error {
 	w.pendingCtxs = make(map[context.Context]context.CancelFunc)
 	return nil
 }
+
+// RequestBlock send a request for block <round> and wait until it receives a response or a context expires.
+func (w *wsFetcherClient) RequestBlock(ctx context.Context, round basics.Round) (rpcs.WsGetBlockOut, error) {
```
tsachiherman, line 168: `RequestBlock` -> `requestBlock`

network/topics.go
```diff
@@ -25,8 +25,13 @@ import (
...
 const (
-	requestHashKey = "RequestHash"
-	ErrorKey       = "Error" // used for passing an error message
+	requestHashKey     = "RequestHash"
+	ErrorKey           = "Error"           // used for passing an error message
+	RoundKey           = "roundKey"        // Block round-number topic-key in the request
+	RequestDataTypeKey = "requestDataType" // Data-type topic-key in the request (e.g. block, cert, block+cert)
+	BlockDataKey       = "blockData"       // Block-data topic-key in the response
+	CertDataKey        = "certData"        // Cert-data topic-key in the response
+	BlockAndCertValue  = "blockAndCert"    // block+cert request data (as the value of requestDataTypeKey)
```
tsachiherman, line 34: move these to blockService.go, to keep it as application-level constants.

catchup/wsFetcher.go
tsachiherman: Please review the refactored output of this function and the usage. It looks like you've eliminated few error cases, so make sure to take advantage of that.

protocol/tags.go
```diff
@@ -37,23 +37,7 @@ const (
 	TxnTag             Tag = "TX"
 	UniCatchupReqTag   Tag = "UC"
 	UniEnsBlockReqTag  Tag = "UE"
-	UniEnsBlockResTag  Tag = "US"
-	UniCatchupResTag   Tag = "UT"
+	UniEnsBlockResTag  Tag = "US" // was used for wsfetcherservice
```
tsachiherman, line 40: I would comment out these tags to ensure we don't use these.

tsachiherman: Also, you would need to eliminate the use of UniCatchupReqTag/UniEnsBlockReqTag, but it's ok to defer that.

tsachiherman (approved):

PR #1916, opened 2021-02-17
Deprecate support of reverse catchup protocol ( part 3 )

algonautshant:
## Summary

In this PR, network v1 cleanup 3rd iteration:
- Simplified requestBlock return
- Stopped the use of UniCatchupReqTag in the requests (to eventually eliminate it in the next version)
- Storing the tag in the fetcher structs and passing it via NewOverGossip is not needed, because the fetcher will not use only a single tag value, and have it hard coded in the request function.
   - Removed the tag parameter from NewOverGossip interface 
   - Removed the tag field from the structs WsFetcher and wsFetcherClient

## Test Plan

The existing tests are sufficient here because no no functionality is being added.

### DIFF
--- a/catchup/fetcher.go
+++ b/catchup/fetcher.go
@@ -54,7 +54,7 @@ type FetcherFactory interface {
 	// Create a new fetcher
 	New() Fetcher
 	// Create a new fetcher that also fetches from backup peers over gossip network utilising given message tag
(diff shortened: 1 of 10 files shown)


tsachiherman (approved):
looks good.

PR #1920, opened 2021-02-20
Trim StateDelta fields before storing

nicholasguoalgorand:
## Summary

Copy the txLeases and creatableDeltas maps before storing in txTail and acctupdates, respectively. This is so that if the two maps are fairly empty, we do not want to hold as much memory as allocated for them as nodes have limited memory.
## Test Plan

Ran unit tests, integration tests for correctness
Ran memory profiler on performance test with only payment txns to show that memory held for these two maps are very low


tsachiherman (changes requested):
While the code you've added is factually correct, I think that we could do better.
The reallocation of these two maps is strongly related to the potential memory saving - which cannot be evaluated in the current implementation. What I would suggest is the following:
In StateDelta, add a int variable that would store the initial hint. ( i.e. initialTransactionsCount ).
Add a method to StateDelta that would optimize the above two maps. The method would be called by eval.finalValidation.
When implementing the method, use the initialTransactionsCount and length of the maps to figure out how much memory is being saved. Perform the reallocation only if we surpass a certain threshold.

The intent here is to make sure that we reallocate only in cases where there is a notable amount of memory savings, and that we won't see any regression on the "other" cases.

ledger/ledgercore/statedelta.go
```diff
@@ -149,3 +152,31 @@ func (ad *AccountDeltas) upsert(br basics.BalanceRecord) {
 	}
 	ad.acctsCache[addr] = last
 }
+
+func (sd *StateDelta) Compress() {
```
tsachiherman, line 156: nit: don't use the word "Compress" since we don't really compress anything here. Maybe OptimizeAllocatedMemory ?

ledger/ledgercore/statedelta.go
```diff
@@ -149,3 +152,31 @@ func (ad *AccountDeltas) upsert(br basics.BalanceRecord) {
...
+	sd.Accts.accts = sd.Accts.accts[:len(sd.Accts.accts)]
```
tsachiherman, line 157: this line is a no-op. The allocation for an array is the cap of the array, so just re-slicing it won't do much.
If you really want to reallocate it, you should create another array, and copy the data.

ledger/ledgercore/statedelta.go
```diff
@@ -149,3 +152,31 @@ func (ad *AccountDeltas) upsert(br basics.BalanceRecord) {
...
+
+	if len(sd.Accts.acctsCache) < sd.initialTransactionsCount / 2 {
```
tsachiherman, line 159: The idea is correct, but you're not getting full-value for your efforts. What you want is to make this condition evaluate the memory cost gains of the reallocation before you do it.
in the case of the map, you want to do something like this:
```
initial_cost = sd.initialTransactionsCount * sizeofSuchAMapPerElement
current_cost = len(sd.Accts.acctsCache) * sizeofSuchAMapPerElement
if  initial_cost > current_cost && initial_cost-current_cost > minGainsTreshold {
   .. re-allocate..
}
```

create memory benchmark to measure the sizeofSuchAMapPerElement, and figure out a reasonable value for minGainsTreshold.

Note that when performing this trick on an array, you need to use the array's cap to figure out its underlying array size. The len would only give you the slice length.

ledger/ledgercore/statedelta.go
```diff
@@ -68,6 +68,9 @@ type StateDelta struct {
 
 	// previous block timestamp
 	PrevTimestamp int64
+
+	// initial hint for allocating data structures for StateDelta
+	initialTransactionsCount int
```
tsachiherman, line 73: don't forget to initialize this one.

tsachiherman (changes requested):
In your changes to the "OptimizeAllocatedMemory", you used the patten of 
```
if len(sd.Accts.accts) < sd.initialTransactionsCount-500 {
  …
```

While it could be a good number, there is no way the code reader could understand what the memory footprint of that would be.
The `OptimizeAllocatedMemory` is designed to optimize the allocated memory, and as such, need to perform the calculations
In terms of bytes ( kb/mb ) rather than entries count.

Three examples for that to illustrate why this is an issue:
- A `basics.BalanceRecord` is very likely to consume more space than a map entry. ( maybe I'm wrong, but you'll need to provide a benchmark to precisely measure that )
- Some maps are kept for 320 rounds, whereas some other are kept for 1000 round. The implies that we're likely want to optimize the memory pressure for the latter by order of 3.
- For the reader of the code, it's not clear exactly how much memory is going to be saved here, since the calculations aren't being done on a KB basis.

General speaking, I think that if the saving is less than 50KB, there is no need to try and optimize the allocation.

Last, the code isn't documented at all. We need to have detailed documentation explaining why we're doing that.

tsachiherman (changes requested):
The changes that you've made are good; They align the behavior with what we'd like to have.
From technical perspective, though, there are few style-related changes needed :
1. For each of the conditionals, create a constant for the element size;
	i.e. accountArrayEntrySize, accountMapCacheEntrySize, etc, with values 232, 64, .. and add a comment that points these to the corresponding benchmark.
2. The 320/1000 need to be replaces with the corresponding consensus params MaxBalLookback and MaxTxnLife. You can pass into the `OptimizeAllocatedMemory` function the consensus params for the round.
3. The 50000000 number is a good number, but it should be a constant. declare it at the top of the file. Maybe deltaStateTargetOptimizationThreshold ?

tsachiherman (approved):

tsachiherman:
I'm going to punt on this one until after branching off the next release.

PR #1923, opened 2021-02-23
Universal fetcher for http and ws fetchers

algonautshant:
## Summary
UniversalFetcher is a fetcher that returns a block given a peer. 
It uses the current websocket fetcher and the http fetcher implementations.  

## Test Plan
Tests added for http and ws cases.


catchup/universalFetcher.go
```diff
@@ -0,0 +1,83 @@
...
+// MakeUniversalFetcher returns a fetcher for http and ws peers.
+func MakeUniversalFetcher(config config.Local, net network.GossipNode, log logging.Logger) UniversalFetcher {
+	return UniversalFetcher{
+		config: config,
+		net:    net,
+		log:    log}
+}
+
+// FetchBlock returns a block from the peer. The peer can be either an http or ws peer.
+func (uf *UniversalFetcher) FetchBlock(ctx context.Context, round basics.Round, peer network.Peer) (blk *bookkeeping.Block,
```
tsachiherman, line 47: Instead of using the makeHTTPFetcher/MakeWsFetcher inside the UniversalFetcher, copy their implementation inside the UniversalFetcher so that you could deprecate them.

catchup/service.go
```diff
@@ -167,15 +174,17 @@ func (s *Service) innerFetch(fetcher Fetcher, r basics.Round) (blk *bookkeeping.
 			cf()
 		}
 	}()
-	return fetcher.FetchBlock(ctx, r)
+	peer, err := s.blocksDownloadPeerSelector.GetNextPeer()
```
tsachiherman, line 177: this line need to be protected with a mutex.

catchup/universalFetcher.go
```diff
@@ -0,0 +1,132 @@
...
+	newBlockFetcher() blockFetcher
+}
+
+type universalBlockFetcherFactory struct {
+	log    logging.Logger
+	net    network.GossipNode
+	config config.Local
+}
+
+func makeUniversalBlockFetcherFactory(log logging.Logger, net network.GossipNode, config config.Local) blockFetcherFactory {
```
tsachiherman, line 60: do we really need this ? it seems like a redundant factory.

algonautshant: I introduced it to use the mock fetcher in the block fetcher service tests.

catchup/service.go
algonautshant: There is a mutex protection inside GetNextPeer. Is that enough?

tsachiherman: Yes, I forgot about this one. I guess it's making your implementation even simpler.

catchup/universalFetcher.go
tsachiherman: Given that it's being used only for testing suggest that the testing are the ones that need to be implemented to changed to adapt to the new code, right ?

catchup/universalFetcher_test.go
```diff
@@ -0,0 +1,131 @@
...
+	cfg := config.GetDefaultLocal()
+	cfg.EnableCatchupFromArchiveServers = true
+
+	ledger, next, b, err := buildTestLedger(t)
+	if err != nil {
+		t.Fatal(err)
+		return
+	}
+
+	version := "2.1"
```
tsachiherman, line 45: this version is no longer meaningful, right ?

catchup/universalFetcher_test.go
```diff
@@ -0,0 +1,131 @@
...
+	require.Equal(t, int64(duration), int64(0))
+}
+
+// TestUGetBlockUnsupported tests the handling of an unsupported peer
+func TestUGetBlockUnsupported(t *testing.T) {
+	fetcher := universalBlockFetcher{}
+	peer := ""
+	block, cert, duration, err := fetcher.fetchBlock(context.Background(), 1, peer)
+	require.Error(t, err)
+	require.Contains(t, err.Error(), "FetchBlock: UniversalFetcher only supports HTTPPeer or UnicastPeer")
```
tsachiherman, line 127: "or" -> "and"

test/e2e-go/cli/goal/expect/goalExpectCommon.exp
```diff
@@ -46,7 +46,7 @@ package require Tcl 8.0
 proc ::AlgorandGoal::Abort { ERROR } {
     puts "Aborting with Error: $ERROR"
 
-    if { "$::GLOBAL_TEST_ALGO_DIR" != "" && "$::GLOBAL_TEST_ROOT_DIR" != "" } {
+    if { [info exists ::GLOBAL_TEST_ALGO_DIR] && [info exists ::GLOBAL_TEST_ROOT_DIR] } {
```
tsachiherman, line 49: this doesn't really related to this PR , right ?

algonautshant: no. But when the test was failing, the abort script was generating more errors. I'll make this into a separate PR.

catchup/universalFetcher.go
```diff
@@ -0,0 +1,104 @@
...
+		return nil, nil, time.Duration(0), err
+	}
+	return block, cert, downloadDuration, err
+}
+
+func processBlockBytes(fetchedBuf []byte, r basics.Round, debugStr string) (blk *bookkeeping.Block, cert *agreement.Certificate, err error) {
+	var decodedEntry rpcs.EncodedBlockCert
+	if uint64(r) == 0 {
+		r = 0
+	}
```
tsachiherman, line 87: not needed, right ?

algonautshant: Yes. no need.

catchup/service.go
```diff
@@ -345,6 +342,30 @@ func (s *Service) pipelinedFetch(seedLookback uint64) {
...
+			})
+	} else {
+		peerSelector = makePeerSelector(
+			s.net,
+			[]peerClass{
+				{initialRank: peerRankInitialFirstPriority, peerClass: network.PeersPhonebookArchivers},
+				{initialRank: peerRankInitialSecondPriority, peerClass: network.PeersConnectedOut},
+				{initialRank: peerRankInitialThirdPriority, peerClass: network.PeersPhonebookRelays},
+			})
+	}
```
tsachiherman, line 363: nit: refactor out these `makePeerSelector` call into external method `createPeerSelector`, that would get an argument `pipelineFetch bool`. Internally, the method would determine the best set of peer classes, and would call `makePeerSelector`, returning the peerSelector object.
i.e.
```
func (s *Service) makePeerSelector(createPeerSelector bool) *peerSelector {
   var peerClasses []peerClass
   if createPeerSelector {
      if s.cfg.NetAddress != "" { // Relay node
          peerClasses = ...
      } else {
      }
   } else {
      ...
   }
   return makePeerSelector(s.net, peerClasses)
}
```

algonautshant: done

catchup/service.go
```diff
@@ -557,20 +554,16 @@ func (s *Service) syncCert(cert *PendingUnmatchedCertificate) {
...
-		if fetcher.OutOfPeers(cert.Round) {
-			fetcher.Close()
-			// refresh peers and try again
-			logging.Base().Warn("fetchRound found no outgoing peers")
-			s.net.RequestConnectOutgoing(true, s.ctx.Done())
-			fetcher = s.latestRoundFetcherFactory.NewOverGossip()
+		peer, getPeerErr := peerSelector.GetNextPeer()
+		if getPeerErr != nil {
+			s.log.Debugf("fetchRound: was unable to obtain a peer to retrieve the block from")
+			break
```
tsachiherman, line 562: the `fetchRound` method need to ensure that it doesn't exit before the round is fetched. Therefore, in this case, you want to call `s.net.RequestConnectOutgoing(true, s.ctx.Done())` and `continue`. The `peerSelector.GetNextPeer` would internally try to pick new relays.

tsachiherman (changes requested):
one more request regarding the `fetchRound` method. looks great otherwise.

tsachiherman (approved):

PR #1924, opened 2021-02-23
Retrieve assets when needed in tealdbg

jasonpaulos:
## Summary

Currently tealdbg is able to look up the creator of an application and retrieve their balance record to provide accurate information about the app during a debugging session. The same is not true for assets, meaning if the asset creator's balance record is not provided in another way, tealdbg is unable to provide asset parameter information to the TEAL program.

This PR fixes this by letting tealdbg look up balance records for asset creators just like it already does for applications.

## Test Plan

Given the following transaction on testnet that references a foreign asset:
```javascript
{
  "txn": {
    "apap": "AiACAAEmAg1hc3NldF90b3RhbF8wDWFzc2V0X3RvdGFsXzEicQA1ADUBKDQAZyk0AWcjQw==",
    "apas": [
      10458941
    ],
    "apgs": {
      "nui": 2
    },
    "apsu": "AiABASI=",
    "fee": 1000,
    "fv": 12527718,
    "gen": "testnet-v1.0",
    "gh": "SGO1GKSzyE7IEPItTxCByw9x8FmnrCDexi9/cOUJOiI=",
    "lv": 12528718,
    "snd": "XX4WKKGGY4X5TI4JSG7Z2ALXI7OJYSJYVF2F7D7H6Q32Y7ON5VOF3B5SXE",
    "type": "appl"
  }
}
```

And running tealdbg like so:
```
tealdbg debug -t txn.json -i "MY_TESTNET_INDEXER" -r 12527737 -a 14181575
```

<details>
<summary>Before this PR, asset params would not be loaded if the creator is not part of the transaction group:</summary>

</details>

<details>
<summary>With this PR, the asset parameter information is available as expected:</summary>

</details>

### DIFF
--- a/cmd/tealdbg/localLedger.go
+++ b/cmd/tealdbg/localLedger.go
@@ -20,6 +20,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
+	"math"
 	"math/rand"
 	"net/http"
 
@@ -57,6 +58,16 @@ type ApplicationIndexerResponse struct {
 	CurrentRound uint64 `json:"current-round"`
 }
 
+// AssetIndexerResponse represents the Asset Response object from querying indexer
+type AssetIndexerResponse struct {
+
+	// Application index and its parameters
+	Asset generated.Asset `json:"asset,omitempty"`
+
+	// Round at which the results were computed.
+	CurrentRound uint64 `json:"current-round"`
+}
+
 type localLedger struct {
 	balances        map[basics.Address]basics.AccountData
 	txnGroup        []transactions.SignedTxn
@@ -83,6 +94,8 @@ func makeBalancesAdapter(
 	apps := []basics.AppIndex{appIdx}
 	apps = append(apps, txn.Txn.ForeignApps...)
 
+	assets := txn.Txn.ForeignAssets
+
 	// populate balances from the indexer if not already
 	if indexerURL != "" {
 		for _, acc := range accounts {
@@ -97,6 +110,16 @@ func makeBalancesAdapter(
 		}
 		for _, app := range apps {
 			creator, err := getAppCreatorFromIndexer(indexerURL, indexerToken, app)
+			if err != nil {
+				return nil, AppState{}, err
+			}
+			balances[creator], err = getBalanceFromIndexer(indexerURL, indexerToken, creator, round)
+			if err != nil {
+				return nil, AppState{}, err
+			}
+		}
+		for _, asset := range assets {
+			creator, err := getAssetCreatorFromIndexer(indexerURL, indexerToken, asset)
 			if err != nil {
 				return nil, AppState{}, err
 			}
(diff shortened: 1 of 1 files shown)


jasonpaulos:
By the way I would love to see more comprehensive tests for tealdbg to catch things like this, but that seems like a much larger issue. Currently I'm just fixing problems as I see them.

cmd/tealdbg/localLedger.go
```diff
@@ -138,8 +161,35 @@ func makeBalancesAdapter(
...
+					AssetParams: map[basics.AssetIndex]basics.AssetParams{
+						aid: makeAssetParams(addr),
+					},
+					Assets: map[basics.AssetIndex]basics.AssetHolding{
+						aid: makeAssetHolding(),
+					},
+				}
+				balances[addr] = ad
+			}
+		}
```
algorandskiy, line 192: why does it needed at all if these assets are already in `balances`?

algorandskiy (commented):
> By the way I would love to see more comprehensive tests for tealdbg to catch things like this,

what does prevent you from adding such tests? for example, a test for this new functionality?

cmd/tealdbg/localLedger.go
jasonpaulos: If indexer is not being used then all the referenced assets in the transaction may not exist. If the painless option is enabled, this code generates a balance record for each asset that doesn't exist.

algorandskiy: ah, sorry, overlooked `!ok`

jasonpaulos:
> what does prevent you from adding such tests? for example, a test for this new functionality?

I meant that it's pretty difficult to test some behavior of tealdbg with the current testing infrastructure, since a lot of the features rely on a frontend. I think it would be really great if we could have more in depth tests to make sure our JSON responses match what CDT wants, or even run a headless version of Chrome to verify fields in the frontend evaluate to what we expect.

That being said, I just thought of a way to unit test this change without doing any of the above. I'm going to revert this to a draft until I implement that unit test.

algorandskiy:
Yes, take a look into local_test, there are couple examples that runs `Setup` that exercises `makeBalancesAdapter` with your changes.

cmd/tealdbg/localLedger.go
```diff
@@ -211,6 +261,34 @@ func getAppCreatorFromIndexer(indexerURL string, indexerToken string, app basics
 	return creator, nil
 }
 
+func getAssetCreatorFromIndexer(indexerURL string, indexerToken string, asset basics.AssetIndex) (basics.Address, error) {
+	queryString := fmt.Sprintf("%s/v2/assets/%d", indexerURL, asset)
+	client := &http.Client{}
+	request, err := http.NewRequest("GET", queryString, nil)
+	request.Header.Set("X-Indexer-API-Token", indexerToken)
```
tsachiherman, line 268: it would be nicer if you could move the ""X-Indexer-API-Token" to a const at the top of the file.

cmd/tealdbg/localLedger.go
```diff
@@ -211,6 +261,34 @@ func getAppCreatorFromIndexer(indexerURL string, indexerToken string, app basics
...
+	resp, err := client.Do(request)
+	if err != nil {
+		return basics.Address{}, fmt.Errorf("asset request error: %s", err)
```
tsachiherman, line 271: nit: when formatting errors, use either `%v` or `%w` for the error object.

cmd/tealdbg/localLedger.go
```diff
@@ -211,6 +261,34 @@ func getAppCreatorFromIndexer(indexerURL string, indexerToken string, app basics
...
+	}
+	defer resp.Body.Close()
+	if resp.StatusCode != 200 {
```
tsachiherman, line 274: nit: is there any constant in the http package that equal to the magic number 200 ? Maybe StatusOK ?

jasonpaulos:
I'm closing this for now, but I'll reopen once I make the changed needed for this to be ready.

PR #1926, opened 2021-02-24
Use total stake instead of online money

aybehrouz:
### Summary

Currently `AccountData` struct has a `VotingStake` method, but there is no way for calculating the sum of voting stake of all accounts. We can only calculate the total money of online accounts. As a result, it's not possible to easily implement different agreement protocols by changing the behavior of `VotingStake` method.

This PR will add `TotalStake` method to `LedgerReader` interface which returns the sum of all accounts voting stake.

### Test Plan

I think current tests should be enough.
To be honest, I'm not sure if I've made all the necessary changes to the code.

### DIFF
--- a/agreement/abstractions.go
+++ b/agreement/abstractions.go
@@ -146,6 +146,15 @@ type LedgerReader interface {
 	// protocol may lose liveness.
 	Circulation(basics.Round) (basics.MicroAlgos, error)
 
+	// TotalStake returns the total amount of accounts stake at the conclusion of a
+	// given round.
+	//
+	// This method returns an error if the given Round has not yet been
+	// confirmed. It may also return an error if the given Round is
+	// unavailable by the storage device. In that case, the agreement
+	// protocol may lose liveness.
+	TotalStake(basics.Round) (basics.MicroAlgos, error)
+
 	// LookupDigest returns the Digest of the entry that was agreed on in a
 	// given round.
 	//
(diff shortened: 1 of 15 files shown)


tsachiherman:
Could you explain what is the difference between the total online account money and the total accumulated voting stake ?
I.e. how could an account voting stake balance be different from the account online balance ?

tsachiherman:
I grabbed some of your typos fixes in this PR and moved them to https://github.com/algorand/go-algorand/pull/1927; I also added you to the THANKS.md file.

aybehrouz:
They must be equal in the current code. 

This is actually a refactoring PR. If we use the total accumulated voting stake instead of the online money that would make it possible to implement different agreement protocols by changing `VotingStake()` method. For instance, if we want to prevent accounts with low balance from participating in the consensus then we can change this method to something like this: (without considering the required logic for handling protocol versions)
```go
func (u AccountData) VotingStake() MicroAlgos {
	if u.Status != Online || u.MicroAlgos.LessThan(Threshold) {
		return MicroAlgos{Raw: 0}
	}
	return u.MicroAlgos
}
```
In this case, we need to calculate the total stake instead of total online money. Unfortunately in the current code the sum of voting stake is not used in the agreement protocol.

I was experimenting with a different agreement protocol and I needed to use a different `VotingStake()` method.

aybehrouz:
> I grabbed some of your typos fixes in this PR and moved them to #1927; I also added you to the THANKS.md file.

I think you should add "jetbrains" to that file instead of my name! ;D

cause my IDE corrected those typos. xD

PR #1929, opened 2021-02-25
Preload account data in parallel on the evaluator

tsachiherman:
## Summary

Modify the `eval()` to load the account data before the evaluation of each transaction group. 

## Test Plan

Existing test provide sufficient coverage of the change.

### DIFF
--- a/ledger/eval.go
+++ b/ledger/eval.go
@@ -22,8 +22,6 @@ import (
 	"fmt"
 	"sync"
 
-	"github.com/algorand/go-deadlock"
-
 	"github.com/algorand/go-algorand/config"
 	"github.com/algorand/go-algorand/crypto"
 	"github.com/algorand/go-algorand/crypto/compactcert"
@@ -73,9 +71,6 @@ type roundCowBase struct {
 	// are beyond the scope of this cache.
 	// The account data store here is always the account data without the rewards.
 	accounts map[basics.Address]basics.AccountData
-
-	// accountsMu is the accounts read-write mutex, used to syncronize the access ot the accounts map.
-	accountsMu deadlock.RWMutex
 }
 
 func (x *roundCowBase) getCreator(cidx basics.CreatableIndex, ctype basics.CreatableType) (basics.Address, bool, error) {
@@ -86,18 +81,13 @@ func (x *roundCowBase) getCreator(cidx basics.CreatableIndex, ctype basics.Creat
 // first, and if it cannot find it there, it would defer to the underlaying implementation.
 // note that errors in accounts data retrivals are not cached as these typically cause the transaction evaluation to fail.
 func (x *roundCowBase) lookup(addr basics.Address) (basics.AccountData, error) {
-	x.accountsMu.RLock()
 	if accountData, found := x.accounts[addr]; found {
-		x.accountsMu.RUnlock()
 		return accountData, nil
 	}
-	x.accountsMu.RUnlock()
 
 	accountData, _, err := x.l.LookupWithoutRewards(x.rnd, addr)
 	if err == nil {
-		x.accountsMu.Lock()
 		x.accounts[addr] = accountData
-		x.accountsMu.Unlock()
 	}
 	return accountData, err
 }
@@ -1154,7 +1144,7 @@ func eval(ctx context.Context, l ledgerForEvaluator, blk bookkeeping.Block, vali
 		//wg.Add(1)
 		//go prefetchThread(validationCtx, eval.state.lookupParent, blk.Payset, &wg)
 	}
-	paysetgroupsCh := loadAccounts(l, blk.Round()-1, eval.state, paysetgroups)
+	paysetgroupsCh := loadAccounts(l, blk.Round()-1, paysetgroups)
 
 	var txvalidator evalTxValidator
 	if validate {
@@ -1173,13 +1163,18 @@ func eval(ctx context.Context, l ledgerForEvaluator, blk bookkeeping.Block, vali
 
 	}
 
+	base := eval.state.lookupParent.(*roundCowBase)
+
 	for {
 		select {
 		case txgroup, ok := <-paysetgroupsCh:
 			if !ok {
 				break
 			}
-			err = eval.TransactionGroup(txgroup)
+			for _, br := range txgroup.balances {
+				base.accounts[br.Addr] = br.AccountData
+			}
+			err = eval.TransactionGroup(txgroup.group)
 			if err != nil {
 				return ledgercore.StateDelta{}, err
 			}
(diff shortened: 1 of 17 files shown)


ledger/eval.go
```diff
@@ -1218,31 +1223,147 @@ func eval(ctx context.Context, l ledgerForEvaluator, blk bookkeeping.Block, vali
...
+							AccountData: acctData,
+						}
+						if err == nil {
+							for i, wg := range task.groups {
+								wg.balances[task.groupIndices[i]] = br
+								wg.done <- nil
+							}
+						} else {
+							for i, wg := range task.groups {
+								wg.balances[task.groupIndices[i]] = br
```
brianolson, line 1332: drop this line because we're going to throw this away and error out?

brianolson (commented):
Looks good, a few tweaks, created a sub-PR with comments I wish I'd had
https://github.com/tsachiherman/go-algorand/pull/43

ledger/eval.go
```diff
@@ -1171,22 +1163,35 @@ func eval(ctx context.Context, l ledgerForEvaluator, blk bookkeeping.Block, vali
...
 			}
-		default:
-		}
-
-		err = eval.TransactionGroup(txgroup)
-		if err != nil {
-			return ledgercore.StateDelta{}, err
+			continue
 		}
+		break
```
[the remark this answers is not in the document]
tsachiherman: I needed this because of the lines 1171-1172 : 
in case we're running out of transaction groups in the channel, we need to exit the for loop. Using "break" won't exit the for loop since we're in a select statement.

ledger/eval.go
```diff
@@ -1218,31 +1223,147 @@ func eval(ctx context.Context, l ledgerForEvaluator, blk bookkeeping.Block, vali
...
+					addr:         addr,
+					groups:       make([]*groupTask, 1, 4),
+					groupIndices: make([]int, 1, 4),
+				}
+				task.groups[0] = wg
+				task.groupIndices[0] = wg.balancesCount
+				accountsChannels[addr] = task
+				addressesCh <- task
+			} else {
+				task := accountsChannels[addr]
```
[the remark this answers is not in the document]
tsachiherman: good idea.

ledger/eval.go
```diff
@@ -1218,31 +1223,147 @@ func eval(ctx context.Context, l ledgerForEvaluator, blk bookkeeping.Block, vali
...
+			wg.balancesCount++
+			totalBalances++
 		}
-	}
+		if len(groups) > 0 {
+			task := &addrTask{
+				addr: feeSinkAddr,
+			}
+			addressesCh <- task
+			accountsChannels[feeSinkAddr] = task
```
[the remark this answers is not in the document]
tsachiherman: Yes, I was thinking about that. The rewards pool is being accessed in the startEvaluator, right ? so, by the time we execute this function, that account should already be in the baseState, right ?

Also, the `startEvaluator` needs this regardless if there are any transactions.

brianolson (approved):
LGTM, can wait for @algorandskiy to triple check if you like

tsachiherman:
I think that I'd like to wait, since he is working on adjacent changes.

ledger/eval.go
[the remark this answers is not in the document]
algorandskiy: I agree with Brian - add `break outer` - see suggested change.

ledger/eval.go
```diff
@@ -1171,22 +1163,35 @@ func eval(ctx context.Context, l ledgerForEvaluator, blk bookkeeping.Block, vali
...
 	}
 
-	for _, txgroup := range paysetgroups {
+	base := eval.state.lookupParent.(*roundCowBase)
+
+	for {
 		select {
+		case txgroup, ok := <-paysetgroupsCh:
+			if !ok {
+				break
```
algorandskiy, line 1172:
```suggestion
outer:
	for {
		select {
		case txgroup, ok := <-paysetgroupsCh:
			if !ok {
				break outer
```

algorandskiy (commented):
What is a performance gain?

ledger/eval.go
tsachiherman: It's the first time I'm seeing a use for the `break label`, but I don't see why not to use it.

tsachiherman: The performance gains are about 20% faster account loading, which is notable when you have full blocks.
When looking on the overall `eval()` gains, it's roughly 10% of the total time.

The key here is that we want to perform all the account and sql operation done in parallel, and (potentially) ahead of transaction processing. We might want to include the asset loading here as well, but this could wait.

algorandskiy: yep, in couple places we have goto instead of `break label` or `continue label` and there goto is fine. Here I think the code will be much simpler with this construct.

ledger/eval.go
```diff
@@ -1218,39 +1221,188 @@ func eval(ctx context.Context, l ledgerForEvaluator, blk bookkeeping.Block, vali
...
+		totalBalances := 0
+		initAccount := func(addr basics.Address, wg *groupTask) {
+			if addr.IsZero() {
+				return
+			}
+			if task, have := accountsChannels[addr]; !have {
+				task := &addrTask{
+					address:      addr,
+					groups:       make([]*groupTask, 1, 4),
+					groupIndices: make([]int, 1, 4),
```
algorandskiy, line 1272: why do you set `cap=4`?

ledger/eval.go
```diff
@@ -1218,39 +1221,188 @@ func eval(ctx context.Context, l ledgerForEvaluator, blk bookkeeping.Block, vali
...
+			accountsChannels[feeSinkAddr] = task
+		}
+
+		// iterate over the transaction groups and add all their account addresses to the list
+		groupsReady := make([]*groupTask, len(groups))
+		for i, group := range groups {
+			groupWg := &groupTask{}
+			groupsReady[i] = groupWg
+			for _, stxn := range group {
+				initAccount(stxn.Txn.Sender, groupWg)
```
algorandskiy, line 1300: can you please check if moving
```
			if addr.IsZero() {
				return
			}
```
from the func would give some perf gain?

ledger/eval.go
```diff
@@ -1218,39 +1221,188 @@ func eval(ctx context.Context, l ledgerForEvaluator, blk bookkeeping.Block, vali
...
+				address: feeSinkAddr,
+			}
+			addressesCh <- task
+			accountsChannels[feeSinkAddr] = task
+		}
+
+		// iterate over the transaction groups and add all their account addresses to the list
+		groupsReady := make([]*groupTask, len(groups))
+		for i, group := range groups {
+			groupWg := &groupTask{}
```
algorandskiy, line 1297: what does `groupWg` mean? **wg** might mean **work group** but then the abbreviation is similar to **wait group** from `sync` package.

ledger/eval.go
```diff
@@ -1218,39 +1221,188 @@ func eval(ctx context.Context, l ledgerForEvaluator, blk bookkeeping.Block, vali
...
-		maybelookup(stxn.Txn.CloseRemainderTo)
-		maybelookup(stxn.Txn.AssetSender)
-		maybelookup(stxn.Txn.AssetReceiver)
-		maybelookup(stxn.Txn.AssetCloseTo)
-		maybelookup(stxn.Txn.FreezeAccount)
-		for _, xa := range stxn.Txn.Accounts {
-			maybelookup(xa)
+		defer close(outChan)
+
+		accountsChannels := make(map[basics.Address]*addrTask)
```
algorandskiy, line 1260: `accountsChannels` -> `accountTasks` ?

algorandskiy (commented):
Few more minor comments/questions. How about adding a unit test ensuring output order from loadAccount's channel?

ledger/eval.go
tsachiherman: I changes it to be just "task"

ledger/eval.go
tsachiherman: np

ledger/eval.go
tsachiherman: I think it wouldn't be a good idea. Removing it means that we would create an entry to that in `accountTasks` which would need to update all the `groupTask` when it's done.

ledger/eval.go
tsachiherman: I'm not a big fan of this part either. Let me see if I preallocate and sub-allocate that.

tsachiherman: Preallocating wasn't as trivial as I initially thought; I'll defer that to a separate task.
The capacity = 4 is a good fit for most of the trivial use cases. using just cap=1 is usually not a smart move, since there is almost always more than a single account involved.

tsachiherman:
I'm going to punt on this one until after branching off the release. These changes aren't needed for the coming release, and it would be better to incorporate these changes along with the changes @pzbitskiy is working on.

ledger/eval.go
algorandskiy: ok, this is just a guess, not exact number, good enough

algorandskiy (approved):
LGTM

PR #1932, opened 2021-03-02
1k+ assets (part 1)

algorandskiy:
## Summary

This PR introduces `PersistedAccountData` data type that is `AccountData` + `ExtendedAssetHoldings`. It will be a blob stored into `accountbase` table instead of plain `AccountData`
Two abbreviations introduced: **pad** is for `PersistedAccountData` and **dbad** is for `DbAccountData` = `PersistedAccountData` + **rowId**.


ledger/ledgercore/persistedacctdata.go
```diff
@@ -0,0 +1,114 @@
...
+	// assetId3 = AmountsAssetIndicesOffsets[2] + assetIdx2
+	AssetOffsets []basics.AssetIndex `codec:"ao,allocbound=MaxHoldingGroupSize"`
+
+	// Holding amount
+	// same number of elements as in AmountsAssetIndicesOffsets
+	Amounts []uint64 `codec:"a,allocbound=MaxHoldingGroupSize"`
+
+	// Holding "frozen" flag
+	// same number of elements as in AmountsAssetIndicesOffsets
+	Frozens []bool `codec:"f,allocbound=MaxHoldingGroupSize"`
```
tsachiherman, line 69: Would this be the most efficient way to store array of bits ? What if we would use byte array for this one, and encode the bits in a little/big endian fashion ?

ledger/ledgercore/persistedacctdata.go
```diff
@@ -0,0 +1,114 @@
...
+}
+
+// ExtendedAssetHolding is AccountData's extension for storing asset holdings
+type ExtendedAssetHolding struct {
+	_struct struct{}             `codec:",omitempty,omitemptyarray"`
```
tsachiherman, line 74: nit: add empty line after `_struct` line

ledger/ledgercore/persistedacctdata.go
```diff
@@ -0,0 +1,114 @@
...
+	Count   uint32               `codec:"c"`
+	Groups  []AssetsHoldingGroup `codec:"gs,allocbound=4096"` // 1M asset holdings
```
tsachiherman, line 76: nit: define const for the 4096

ledger/ledgercore/persistedacctdata.go
```diff
@@ -0,0 +1,114 @@
...
+
+	//msgp:ignore loaded
+	loaded bool
+}
+
+// PersistedAccountData represents actual data stored in DB
+type PersistedAccountData struct {
+	_struct struct{} `codec:",omitempty,omitemptyarray"`
```
tsachiherman, line 84: nit: empty line after `_struct`

ledger/ledgercore/persistedacctdata.go
algorandskiy: done

ledger/ledgercore/persistedacctdata.go
algorandskiy: done

ledger/ledgercore/persistedacctdata.go
algorandskiy: done

ledger/ledgercore/persistedacctdata.go
algorandskiy: Yes, makes sense, let's add it in subsequent PR

tsachiherman (approved):

PR #1933, opened 2021-03-02
1k+ assets (part 2)

algorandskiy:
## Summary

Introduce `ApplicationDbgLedger` interface to use in tealdbg and dryrun. This is to properly isolate internal `legderForCowBase` and allow its extension.

### DIFF
--- a/cmd/tealdbg/localLedger.go
+++ b/cmd/tealdbg/localLedger.go
@@ -23,12 +23,10 @@ import (
 	"math/rand"
 	"net/http"
 
-	"github.com/algorand/go-algorand/config"
 	"github.com/algorand/go-algorand/crypto"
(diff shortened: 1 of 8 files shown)


tsachiherman:
it doesn't look like it was made "on top of the branch". Could you take another look ?

algorandskiy:
Because part 1 was squashed during the merge and the common ancestor is the base branch again instead of part1 commits. This broke the entire idea of separate consequent PRs and all need to be re-created from different branches :(

algorandskiy:
Updated. Now need to `rebase -i upstream/feature/1000-assets` and remove squashed commits manually after each squashed merge.

tsachiherman (approved):
looks good.

PR #1935, opened 2021-03-02
1k+ assets (part 4)

algorandskiy:
## Summary

`accountdb` changes to support `ExtendedAssetHolding`: a new accountext table, migrations, tests.

### DIFF
--- a/ledger/accountdb.go
+++ b/ledger/accountdb.go
@@ -21,6 +21,7 @@ import (
 	"context"
 	"database/sql"
 	"fmt"
+	"sort"
 	"time"
 
 	"github.com/mattn/go-sqlite3"
@@ -35,6 +36,7 @@ import (
 // accountsDbQueries is used to cache a prepared SQL statement to look up
 // the state of a single account.
 type accountsDbQueries struct {
+	// initialized in accountsDBInit
 	listCreatablesStmt          *sql.Stmt
 	lookupStmt                  *sql.Stmt
 	lookupCreatorStmt           *sql.Stmt
@@ -46,6 +48,9 @@ type accountsDbQueries struct {
 	insertCatchpointStateUint64 *sql.Stmt
 	selectCatchpointStateString *sql.Stmt
 	insertCatchpointStateString *sql.Stmt
+
+	// these are lazily initialized
+	loadAssetHoldingGroupStmt *sql.Stmt
 }
 
 var accountsSchema = []string{
@@ -104,6 +109,7 @@ var createOnlineAccountIndex = []string{
 var accountsResetExprs = []string{
 	`DROP TABLE IF EXISTS acctrounds`,
 	`DROP TABLE IF EXISTS accounttotals`,
+	`DROP TABLE IF EXISTS accountext`,
 	`DROP TABLE IF EXISTS accountbase`,
 	`DROP TABLE IF EXISTS assetcreators`,
 	`DROP TABLE IF EXISTS storedcatchpoints`,
@@ -114,7 +120,7 @@ var accountsResetExprs = []string{
 // accountDBVersion is the database version that this binary would know how to support and how to upgrade to.
 // details about the content of each of the versions can be found in the upgrade functions upgradeDatabaseSchemaXXXX
 // and their descriptions.
-var accountDBVersion = int32(4)
+var accountDBVersion = int32(5)
 
 // dbAccountData is used for representing a single account stored on the disk. In addition to the
 // basics.AccountData, it also stores complete referencing information used to maintain the base accounts
@@ -256,6 +262,11 @@ func (a *compactAccountDeltas) accountsLoadOld(tx *sql.Tx) (err error) {
 		return
 	}
 	defer selectStmt.Close()
+	loadStmt, err := tx.Prepare("SELECT data FROM accountext WHERE id=?")
+	if err != nil {
+		return
+	}
+	defer loadStmt.Close()
 	defer func() {
 		a.misses = nil
 	}()
@@ -263,23 +274,51 @@ func (a *compactAccountDeltas) accountsLoadOld(tx *sql.Tx) (err error) {
 	var acctDataBuf []byte
 	for _, idx := range a.misses {
 		addr := a.addresses[idx]
+		dbad := dbAccountData{addr: addr}
 		err = selectStmt.QueryRow(addr[:]).Scan(&rowid, &acctDataBuf)
 		switch err {
 		case nil:
+			dbad.rowid = rowid.Int64
 			if len(acctDataBuf) > 0 {
-				persistedAcctData := &dbAccountData{addr: addr, rowid: rowid.Int64}
-				err = protocol.Decode(acctDataBuf, &persistedAcctData.pad)
+				err = protocol.Decode(acctDataBuf, &dbad.pad)
 				if err != nil {
 					return err
 				}
-				a.updateOld(idx, *persistedAcctData)
+
+				// fetch holdings that are in new
+				// these are either new or modified, and needed to be written back later.
+				// to simplify upcoming reconciliation load them now
+				_, delta := a.getByIdx(idx)
+				if len(delta.new.Assets) > 0 && len(dbad.pad.Assets) == 0 {
+					dbad.pad.Assets = make(map[basics.AssetIndex]basics.AssetHolding, len(delta.new.Assets))
+				}
+				for aidx := range delta.new.Assets {
+					if _, ok := dbad.pad.AccountData.Assets[aidx]; !ok {
+						gi, ai := dbad.pad.ExtendedAssetHolding.FindAsset(aidx, 0)
+						if gi != -1 {
+							g := &dbad.pad.ExtendedAssetHolding.Groups[gi]
+							groupData, err := loadAssetHoldingGroupData(loadStmt, g.AssetGroupKey)
+							if err != nil {
+								return err
+							}
+							g.Load(groupData)
+							_, ai = dbad.pad.ExtendedAssetHolding.FindAsset(aidx, gi)
+							if ai != -1 {
+								// found!
+								dbad.pad.AccountData.Assets[aidx] = groupData.GetHolding(ai)
+							} else {
+								// no such asset => newly created
+							}
+						}
+					}
+				}
 			} else {
 				// to retain backward compatability, we will treat this condition as if we don't have the account.
-				a.updateOld(idx, dbAccountData{addr: addr, rowid: rowid.Int64})
 			}
+			a.updateOld(idx, dbad)
 		case sql.ErrNoRows:
 			// we don't have that account, just return an empty record.
-			a.updateOld(idx, dbAccountData{addr: addr})
+			a.updateOld(idx, dbad)
 			err = nil
 		default:
 			// unexpected error - let the caller know that we couldn't complete the operation.
@@ -552,10 +591,6 @@ func accountsInit(tx *sql.Tx, initAccounts map[basics.Address]basics.AccountData
 		for addr, data := range initAccounts {
 			_, err = tx.Exec("INSERT INTO accountbase (address, data) VALUES (?, ?)",
 				addr[:], protocol.Encode(&data))
-			if err != nil {
-				return err
-			}
-
 			totals.AddAccount(proto, data, &ot)
 		}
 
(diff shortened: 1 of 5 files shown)


ledger/accountdb.go
```diff
@@ -1026,74 +1080,395 @@ func accountsPutTotals(tx *sql.Tx, totals ledgercore.AccountTotals, catchpointSt
...
+	}
+	eah.SetLoaded()
+	return holdings, eah, nil
+}
 
-	var insertCreatableIdxStmt, deleteCreatableIdxStmt, deleteByRowIDStmt, insertStmt, updateStmt *sql.Stmt
+func loadHoldingGroup(stmt *sql.Stmt, g ledgercore.AssetsHoldingGroup, holdings map[basics.AssetIndex]basics.AssetHolding) (map[basics.AssetIndex]basics.AssetHolding, ledgercore.AssetsHoldingGroup, error) {
+	if holdings == nil {
+		holdings = make(map[basics.AssetIndex]basics.AssetHolding, g.Count)
+	}
```
algorandskiy, line 1102: in part-7 this moved to a caller as part of optimization - sometimes holdings are not needed

ledger/accountdb.go
```diff
@@ -1026,74 +1080,395 @@ func accountsPutTotals(tx *sql.Tx, totals ledgercore.AccountTotals, catchpointSt
...
+			pad.ExtendedAssetHolding = ledgercore.ExtendedAssetHolding{}
+		} else {
+			// Reconcile groups data:
+			// identify groups, load, update, then delete and insert, dump to the disk
+
+			pad.ExtendedAssetHolding = oldPad.ExtendedAssetHolding
+			if len(updated) > 0 {
+				sort.SliceStable(updated, func(i, j int) bool { return updated[i] < updated[j] })
+				gi, ai := 0, 0
+				for _, aidx := range updated {
```
algorandskiy, line 1275: this entire block re-written after introducing `delta.holding` that track deletion in next PR with tests. Please skip with branch for now

ledger/accountdb.go
```diff
@@ -256,30 +262,63 @@ func (a *compactAccountDeltas) accountsLoadOld(tx *sql.Tx) (err error) {
...
+
+				// fetch holdings that are in new
+				// these are either new or modified, and needed to be written back later.
+				// to simplify upcoming reconciliation load them now
+				_, delta := a.getByIdx(idx)
+				if len(delta.new.Assets) > 0 && len(dbad.pad.Assets) == 0 {
+					dbad.pad.Assets = make(map[basics.AssetIndex]basics.AssetHolding, len(delta.new.Assets))
+				}
+				for aidx := range delta.new.Assets {
+					if _, ok := dbad.pad.AccountData.Assets[aidx]; !ok {
```
tsachiherman, line 296: nit: maybe early continue instead of nesting ?
i.e. 
```
if _, ok := dbad.pad.AccountData.Assets[aidx];  ok {
    continue
}
```

ledger/accountdb.go
```diff
@@ -552,10 +591,6 @@ func accountsInit(tx *sql.Tx, initAccounts map[basics.Address]basics.AccountData
 		for addr, data := range initAccounts {
 			_, err = tx.Exec("INSERT INTO accountbase (address, data) VALUES (?, ?)",
 				addr[:], protocol.Encode(&data))
-			if err != nil {
-				return err
-			}
```
tsachiherman, line 557: I think that we do want to have this. Otherwise, the `totals.AddAccount` might get overflowed, and the function would return `fmt.Errorf("overflow computing totals")`, while hiding the true error.

ledger/accountdb.go
```diff
@@ -631,6 +666,18 @@ func accountsAddNormalizedBalance(tx *sql.Tx, proto config.ConsensusParams) erro
...
+func createAccountExtTable(tx *sql.Tx) error {
+	tableCreate := `CREATE TABLE IF NOT EXISTS accountext (
+		id integer primary key,
+		data blob)`
+	_, err := tx.Exec(tableCreate)
+	if err != nil {
+		return err
+	}
+
+	return nil
```
tsachiherman, line 678: nit: simplify - 
```
return err
```

ledger/accountdb.go
```diff
@@ -1026,74 +1080,395 @@ func accountsPutTotals(tx *sql.Tx, totals ledgercore.AccountTotals, catchpointSt
 	return err
 }
 
-// accountsNewRound updates the accountbase and assetcreators tables by applying the provided deltas to the accounts / creatables.
-// The function returns a dbAccountData for the modified accounts which can be stored in the base cache.
-func accountsNewRound(tx *sql.Tx, updates compactAccountDeltas, creatables map[basics.CreatableIndex]ledgercore.ModifiedCreatable, proto config.ConsensusParams, lastUpdateRound basics.Round) (updatedAccounts []dbAccountData, err error) {
+func loadHoldings(stmt *sql.Stmt, eah ledgercore.ExtendedAssetHolding) (map[basics.AssetIndex]basics.AssetHolding, ledgercore.ExtendedAssetHolding, error) {
```
tsachiherman, line 1083: you're calling this method from two places, and in both you're ignoring the returned `ledgercore.ExtendedAssetHolding`. If this is an input/output parameter, maybe you should pass it via a pointer ? I think that the fact you're doing it by copy might be an issue here since you're changing an internal array element, which would change the source, but then - you're not updating the source correctly.

ledger/accountdb.go
```diff
@@ -1026,74 +1080,395 @@ func accountsPutTotals(tx *sql.Tx, totals ledgercore.AccountTotals, catchpointSt
...
+	if len(eah.Groups) == 0 {
+		return nil, ledgercore.ExtendedAssetHolding{}, nil
+	}
+	var err error
+	holdings := make(map[basics.AssetIndex]basics.AssetHolding, eah.Count)
+	for gi, g := range eah.Groups {
+		holdings, eah.Groups[gi], err = loadHoldingGroup(stmt, g, holdings)
```
tsachiherman, line 1090: The usage of `g` here is really misleading. While it's technically correct, it would be much better to pass a pointer to the group data ( i.e. `&each.Groups[gi]` ) to loadHoldingGroup and let it perform the update.

ledger/accountdb.go
```diff
@@ -1026,74 +1080,395 @@ func accountsPutTotals(tx *sql.Tx, totals ledgercore.AccountTotals, catchpointSt
...
+	err = stmt.QueryRow(key).Scan(&buf)
 	if err != nil {
+		return ledgercore.AssetsHoldingGroupData{}, err
+	}
+
+	err = protocol.Decode(buf, &group)
+	return
+}
+
+func accountsNewCreate(qabi *sql.Stmt, qaei *sql.Stmt, addr basics.Address, ad basics.AccountData, proto config.ConsensusParams, updatedAccounts []dbAccountData, updateIdx int) ([]dbAccountData, error) {
```
tsachiherman, line 1127: rename `proto` -> `genesisProto`.

ledger/accountdb.go
```diff
@@ -1026,74 +1080,395 @@ func accountsPutTotals(tx *sql.Tx, totals ledgercore.AccountTotals, catchpointSt
...
+	assetsThreshold := config.Consensus[protocol.ConsensusV18].MaxAssetsPerAccount
```
tsachiherman, line 1128: This need to be based on the current round protocol version. This makes things harder, since it means that we'll need to adjust the committedUpTo() so that it would break updates on protocol boundaries.

ledger/accountdb.go
```diff
@@ -1026,74 +1080,395 @@ func accountsPutTotals(tx *sql.Tx, totals ledgercore.AccountTotals, catchpointSt
...
+		pad.AccountData = ad
+		pad.AccountData.Assets = nil
+	}
+
+	var result sql.Result
+	var err error
+	for i := 0; i < len(pad.ExtendedAssetHolding.Groups); i++ {
+		result, err = qaei.Exec(pad.ExtendedAssetHolding.Groups[i].Encode())
+		if err != nil {
+			break
```
tsachiherman, line 1143: breaking here is abit more risky, since it depends on the flow-control of the method. I would suggest returning here with the error and without updatedAccounts.
( in case of an update error, the entire transaction is being rolled back, so doing the updates won't really help )

ledger/accountdb.go
```diff
@@ -1026,74 +1080,395 @@ func accountsPutTotals(tx *sql.Tx, totals ledgercore.AccountTotals, catchpointSt
...
+			updatedAccounts[updateIdx].rowid, err = result.LastInsertId()
+			updatedAccounts[updateIdx].pad = pad
+		}
+	}
+	return updatedAccounts, err
+}
+
+func accountsNewDelete(qabd *sql.Stmt, qaed *sql.Stmt, addr basics.Address, dbad dbAccountData, updatedAccounts []dbAccountData, updateIdx int) ([]dbAccountData, error) {
+	result, err := qabd.Exec(dbad.rowid)
+	if err == nil {
```
tsachiherman, line 1164: nit: early exit:
```
if err != nil {
   return updatedAccounts, err
}
...
```

ledger/accountdb.go
```diff
@@ -1026,74 +1080,395 @@ func accountsPutTotals(tx *sql.Tx, totals ledgercore.AccountTotals, catchpointSt
...
+				}
+			}
+		}
+	}
+	return updatedAccounts, err
+}
+
+// accountsNewUpdate reconciles old and new AccountData in delta.
+// Precondition: all modified assets must be loaded into old's groupData and cached into Assets
+func accountsNewUpdate(qabu, qabq, qaeu, qaei, qaed *sql.Stmt, addr basics.Address, delta accountDelta, proto config.ConsensusParams, updatedAccounts []dbAccountData, updateIdx int) ([]dbAccountData, error) {
```
tsachiherman, line 1187: I'm skipping reviewing this method for now. Will review it in subsequent PRs.

ledger/accountdb.go
```diff
@@ -1026,74 +1080,395 @@ func accountsPutTotals(tx *sql.Tx, totals ledgercore.AccountTotals, catchpointSt
...
+	if q.insertStmt, err = db.Prepare("INSERT INTO accountbase (address, normalizedonlinebalance, data) VALUES (?, ?, ?)"); err != nil {
+		q.close()
 		return
 	}
-	defer deleteByRowIDStmt.Close()
 
-	insertStmt, err = tx.Prepare("INSERT INTO accountbase (address, normalizedonlinebalance, data) VALUES (?, ?, ?)")
-	if err != nil {
+	if q.updateStmt, err = db.Prepare("UPDATE accountbase SET normalizedonlinebalance = ?, data = ? WHERE rowid = ?"); err != nil {
+		q.close()
```
tsachiherman, line 1372: instead of repeating the q.close() for each entry, you could do
```
defer func() {
    if err != nil {
        q.close()
    }
}()
```

tsachiherman (commented):
looks very impressive, few requested changes.

ledger/accountdb.go
algorandskiy: done

ledger/acctupdates.go
```diff
@@ -1329,6 +1335,25 @@ func (au *accountUpdates) upgradeDatabaseSchema3(ctx context.Context, tx *sql.Tx
...
+
+	err = createAccountExtTable(tx)
+	if err != nil {
+		return 0, err
+	}
+
+	// update version
+	_, err = db.SetUserVersion(ctx, tx, upgraded)
+	if err != nil {
+		return 0, fmt.Errorf("accountsInitialize unable to update database schema version from %d to %d: %v", err, current, upgraded)
```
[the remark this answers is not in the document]
algorandskiy: fixed

ledger/accountdb.go
tsachiherman: I'm still worried about the usage of this method.  It's creating a *copy* and making changes that would propagate to the origin. It's only valid when used in the following "pattern":
```go
_, extholdings = loadHoldings(stmt, extholdings)
```
if you were to use the following, it would be broken:
```go
_, newholding = loadHoldings(stmt, extholdings)
```
since the `extholdings` would included the loaded groups, but not the loaded flag.

algorandskiy: yes you right. I fixed it by writing groups into a temp array.

PR #1936, opened 2021-03-02
1k+ assets (part 5)

algorandskiy:
## Summary

1. Extend `apply.Balances` interface with `GetEx` methods. 
2. Use `Allocate`/`Deallocate` for asset holdings adding/removing.
3. Implement ledger.LookupFull that loads all the assets.

### DIFF
--- a/ledger/accountdb.go
+++ b/ledger/accountdb.go
@@ -49,8 +49,7 @@ type accountsDbQueries struct {
 	selectCatchpointStateString *sql.Stmt
 	insertCatchpointStateString *sql.Stmt
 
-	// these are lazily initialized
-	loadAssetHoldingGroupStmt *sql.Stmt
+	loadAccountGroupDataStmt *sql.Stmt
 }
 
 var accountsSchema = []string{
@@ -158,9 +157,10 @@ type compactAccountDeltas struct {
 }
 
 type accountDelta struct {
-	old     dbAccountData
-	new     basics.AccountData
-	ndeltas int
+	old      dbAccountData
+	new      ledgercore.PersistedAccountData
+	holdings map[basics.AssetIndex]ledgercore.HoldingAction
+	ndeltas  int
 }
 
 // catchpointState is used to store catchpoint related variables into the catchpointstate table.
@@ -226,22 +226,26 @@ func makeCompactAccountDeltas(accountDeltas []ledgercore.AccountDeltas, baseAcco
 	for _, roundDelta := range accountDeltas {
 		for i := 0; i < roundDelta.Len(); i++ {
 			addr, acctDelta := roundDelta.GetByIdx(i)
+			hmap := roundDelta.GetHoldingDeltas(addr)
 			if prev, idx := outAccountDeltas.get(addr); idx != -1 {
 				outAccountDeltas.update(idx, accountDelta{ // update instead of upsert economizes one map lookup
-					old:     prev.old,
-					new:     acctDelta,
-					ndeltas: prev.ndeltas + 1,
+					old:      prev.old,
+					new:      acctDelta,
+					holdings: hmap,
+					ndeltas:  prev.ndeltas + 1,
 				})
 			} else {
 				// it's a new entry.
 				newEntry := accountDelta{
-					new:     acctDelta,
-					ndeltas: 1,
+					new:      acctDelta,
+					holdings: hmap,
+					ndeltas:  1,
 				}
 				if baseAccountData, has := baseAccounts.read(addr); has {
 					newEntry.old = baseAccountData
 					outAccountDeltas.insert(addr, newEntry) // insert instead of upsert economizes one map lookup
 				} else {
+					// missing old entries will be populated in accountsLoadOld
 					outAccountDeltas.insertMissing(addr, newEntry)
 				}
 			}
@@ -285,27 +289,53 @@ func (a *compactAccountDeltas) accountsLoadOld(tx *sql.Tx) (err error) {
 					return err
 				}
 
-				// fetch holdings that are in new
-				// these are either new or modified, and needed to be written back later.
-				// to simplify upcoming reconciliation load them now
-				_, delta := a.getByIdx(idx)
-				if len(delta.new.Assets) > 0 && len(dbad.pad.Assets) == 0 {
-					dbad.pad.Assets = make(map[basics.AssetIndex]basics.AssetHolding, len(delta.new.Assets))
-				}
-				for aidx := range delta.new.Assets {
-					if _, ok := dbad.pad.AccountData.Assets[aidx]; !ok {
+				// accountsLoadOld is called from commitRound as db.Atomic => safe to load data without additional sync
+				if dbad.pad.ExtendedAssetHolding.Count > 0 {
+					// fetch holdings that are in new
+					// these are either new or modified, and needed to be written back later.
+					// to simplify upcoming reconciliation load them now
+					_, delta := a.getByIdx(idx)
+					if len(delta.new.Assets) > 0 && len(dbad.pad.AccountData.Assets) == 0 {
+						dbad.pad.AccountData.Assets = make(map[basics.AssetIndex]basics.AssetHolding, len(delta.new.Assets))
+					}
+
+					// load created and deleted asset holding groups
+					for aidx, action := range delta.holdings {
+						gi := -1
+						if action == ledgercore.ActionCreate {
+							// use FindGroup to find a possible matching group for insertion
+							gi = dbad.pad.ExtendedAssetHolding.FindGroup(aidx, 0)
+						} else if action == ledgercore.ActionDelete {
+							gi, _ = dbad.pad.ExtendedAssetHolding.FindAsset(aidx, 0)
+						}
+						if gi != -1 {
+							g := &dbad.pad.ExtendedAssetHolding.Groups[gi]
+							if !g.Loaded() {
+								groupData, err := loadHoldingGroupData(loadStmt, g.AssetGroupKey)
+								if err != nil {
+									return err
+								}
+								g.Load(groupData)
+							}
+						}
+					}
+
+					// load updated assets
+					for aidx := range delta.new.Assets {
 						gi, ai := dbad.pad.ExtendedAssetHolding.FindAsset(aidx, 0)
 						if gi != -1 {
 							g := &dbad.pad.ExtendedAssetHolding.Groups[gi]
-							groupData, err := loadAssetHoldingGroupData(loadStmt, g.AssetGroupKey)
-							if err != nil {
-								return err
+							if !g.Loaded() {
+								groupData, err := loadHoldingGroupData(loadStmt, g.AssetGroupKey)
+								if err != nil {
+									return err
+								}
+								g.Load(groupData)
 							}
-							g.Load(groupData)
 							_, ai = dbad.pad.ExtendedAssetHolding.FindAsset(aidx, gi)
 							if ai != -1 {
 								// found!
-								dbad.pad.AccountData.Assets[aidx] = groupData.GetHolding(ai)
+								dbad.pad.AccountData.Assets[aidx] = g.GetHolding(ai)
 							} else {
 								// no such asset => newly created
 							}
@@ -771,6 +801,9 @@ func accountsRound(tx *sql.Tx) (rnd basics.Round, hashrnd basics.Round, err erro
 	return
 }
 
+const lookupAcctBaseQuery string = "SELECT accountbase.rowid, rnd, data FROM acctrounds LEFT JOIN accountbase ON address=? WHERE id='acctbase'"
+const loadAcctExtQuery string = "SELECT data FROM accountext WHERE id=?"
+
 func accountsDbInit(r db.Queryable, w db.Queryable) (*accountsDbQueries, error) {
 	var err error
 	qs := &accountsDbQueries{}
@@ -780,7 +813,7 @@ func accountsDbInit(r db.Queryable, w db.Queryable) (*accountsDbQueries, error) 
 		return nil, err
 	}
 
-	qs.lookupStmt, err = r.Prepare("SELECT accountbase.rowid, rnd, data FROM acctrounds LEFT JOIN accountbase ON address=? WHERE id='acctbase'")
+	qs.lookupStmt, err = r.Prepare(lookupAcctBaseQuery)
 	if err != nil {
 		return nil, err
 	}
@@ -830,7 +863,7 @@ func accountsDbInit(r db.Queryable, w db.Queryable) (*accountsDbQueries, error) 
 		return nil, err
 	}
 
-	qs.loadAssetHoldingGroupStmt, err = r.Prepare("SELECT data from accountext WHERE id=?")
+	qs.loadAccountGroupDataStmt, err = r.Prepare(loadAcctExtQuery)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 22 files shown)


ledger/cow.go
```diff
@@ -126,15 +128,48 @@ func (cb *roundCowState) getCreator(cidx basics.CreatableIndex, ctype basics.Cre
...
+	// data from cb.mods.Accts is newer than from lookupParent.lookupHolding, so add the asset if any
+	if holding, ok := d.AccountData.Assets[basics.AssetIndex(cidx)]; ok {
+		ad.Assets[basics.AssetIndex(cidx)] = holding
+	}
+	return ledgercore.PersistedAccountData{AccountData: ad}, nil
+
+	// if err != nil {
+	// 	// save ExtendedHolding portion if possible for later reference
+	// 	cb.mods.Accts.Register(addr, d)
+	// }
```
algorandskiy, line 170: removed in part-6

ledger/appcow.go
```diff
@@ -200,53 +200,6 @@ func errAlreadyStorage(addr basics.Address, aidx basics.AppIndex, global bool) e
 	return fmt.Errorf("%v has already opted in to app %d", addr, aidx)
 }
 
-// Allocate creates kv storage for a given {addr, aidx, global}. It is called on app creation (global) or opting in (local)
-func (cb *roundCowState) Allocate(addr basics.Address, aidx basics.AppIndex, global bool, space basics.StateSchema) error {
-	// Check that account is not already opted in
-	allocated, err := cb.allocated(addr, aidx, global)
-	if err != nil {
-		return err
-	}
```
algorandskiy, line 209: moved to **creatablecow.go** since where two methods for both apps and assets

ledger/cow.go
```diff
@@ -126,15 +128,48 @@ func (cb *roundCowState) getCreator(cidx basics.CreatableIndex, ctype basics.Cre
...
 	if ok {
-		return d, nil
+		return ledgercore.PersistedAccountData{AccountData: d}, nil
 	}
 
 	return cb.lookupParent.lookup(addr)
 }
 
+// lookupWithHolding is gets account data but also fetches asset holding or app local data for a specified creatable
+func (cb *roundCowState) lookupHolding(addr basics.Address, cidx basics.CreatableIndex, ctype basics.CreatableType) (data ledgercore.PersistedAccountData, err error) {
```
algorandskiy, line 141: use `cidx basics.CreatableIndex` for possible generalization in future

ledger/acctupdates.go
```diff
@@ -1612,6 +1631,86 @@ func (au *accountUpdates) lookupWithRewards(rnd basics.Round, addr basics.Addres
 	}
 }
 
+// lookupFullWithRewards returns the full account data for a given address at a given round.
+// The rewards are added to the AccountData before returning. Note that the function doesn't update the account with the rewards,
+// even while it does return the AccoutData which represent the "rewarded" account data.
+func (au *accountUpdates) lookupFullWithRewards(rnd basics.Round, addr basics.Address) (pad ledgercore.PersistedAccountData, err error) {
```
tsachiherman, line 1637: this would not work correctly - you need to make all the reads from the database within the same `RLock`. Taking the lock twice means that in-between there could be a database write.

ledger/apply/apply.go
```diff
@@ -31,6 +31,9 @@ type Balances interface {
 	// A non-nil error means the lookup is impossible (e.g., if the database doesn't have necessary state anymore)
 	Get(addr basics.Address, withPendingRewards bool) (basics.AccountData, error)
 
+	// GetEx is like Get(addr, false), but also loads specific creatable
+	GetEx(addr basics.Address, cidx basics.CreatableIndex, ctype basics.CreatableType) (basics.AccountData, error)
```
tsachiherman, line 35: I think that a single `Get` method would be preferable, that would have the following parameters:
Get(addr basics.Address, withPendingRewards bool, cidx basics.CreatableIndex, ctype basics.CreatableType)
where we ignore cidx of -1, and adding support for ctype of "AssetParams" or something like that.
(i.e. so that this Get call would be good for asset holding/asset params and applications )

ledger/apply/asset.go
```diff
@@ -102,7 +103,14 @@ func AssetConfig(cc transactions.AssetConfigTxnFields, header transactions.Heade
...
 
-		return balances.PutWithCreatable(header.Sender, record, created, nil)
+		// Save asset params and holding, and register new creatable
+		err = balances.PutWithCreatable(header.Sender, record, created, nil)
+		if err != nil {
+			return err
+		}
+
+		// register new holding
+		return balances.Allocate(header.Sender, basics.CreatableIndex(newidx), basics.AssetCreatable, false, basics.StateSchema{})
```
tsachiherman, line 113: maybe: pass the `created` to `Allocate`, instead of passing all the arguments ?

ledger/apply/asset.go
```diff
@@ -140,6 +148,8 @@ func AssetConfig(cc transactions.AssetConfigTxnFields, header transactions.Heade
 
 		delete(record.Assets, cc.ConfigAsset)
 		delete(record.AssetParams, cc.ConfigAsset)
+
+		balances.Deallocate(creator, basics.CreatableIndex(cc.ConfigAsset), basics.AssetCreatable, false)
```
tsachiherman, line 152: I'm not a big fan of the usage of the `basics.CreatableLocator` above, since it promotes the `missed field initialization` type-bug. I'd rather seeing the object being constructed as a result of `makeCreatableLocator(..)`

ledger/cow.go
```diff
@@ -126,15 +128,43 @@ func (cb *roundCowState) getCreator(cidx basics.CreatableIndex, ctype basics.Cre
...
+		} else {
+			_, exist = ad.AppLocalStates[basics.AppIndex(cidx)]
+		}
+
+		if exist {
+			return ledgercore.PersistedAccountData{AccountData: ad}, nil
+		}
+	}
+
+	d, err := cb.lookupParent.lookupHolding(addr, cidx, ctype)
```
tsachiherman, line 156: is this always needed ? i.e. in case the `ad` already contain the holding, there is no need to lookup, right ?

ledger/ledgercore/statedelta.go
```diff
@@ -70,12 +70,24 @@ type StateDelta struct {
...
+)
+
 // AccountDeltas stores ordered accounts and allows fast lookup by address
 type AccountDeltas struct {
 	// actual data
 	accts []basics.BalanceRecord
 	// cache for addr to deltas index resolution
 	acctsCache map[basics.Address]int
+	// holdings keeps track of created and deleted holdings per address
+	holdings map[basics.Address]map[basics.AssetIndex]HoldingAction
```
tsachiherman, line 90: I can't see here were you're using this `holding` structure, but it looks here as we have some first-level duplication here : the data inside this map should go into the `accts`.

tsachiherman (changes requested):
it looks like series of steps in the right direction, although it would require few more tweaks. ( I'm mostly concerned about the database locks, but there are few other points as well )

ledger/acctupdates.go
algorandskiy: thanks, fill fix here and below.

ledger/apply/apply.go
algorandskiy: I want to emphasize the fact `Ex` method is more expensive.

ledger/apply/asset.go
algorandskiy: I'll check if it suitable for apps as well

ledger/apply/asset.go
algorandskiy: could you elaborate please? not sure I follow the point of using `makeCreatableLocator` here

ledger/cow.go
algorandskiy: yes, if ad in `mods` and the asset exists, it is returned above

ledger/ledgercore/statedelta.go
algorandskiy: this map tracks created and **deleted** holdings. See `SetHoldingDelta` below.

ledger/ledgercore/statedelta.go
```diff
@@ -70,12 +70,24 @@ type StateDelta struct {
 	PrevTimestamp int64
 }
 
+// HoldingAction is an enum of actions on holdings
+type HoldingAction uint64
```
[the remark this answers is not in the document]
algorandskiy: will do

algorandskiy:
Rebased, requires #2069 to be merged first.

algorandskiy:
Rebased again.

ledger/accountdb.go
```diff
@@ -900,31 +933,86 @@ func (qs *accountsDbQueries) lookupCreator(cidx basics.CreatableIndex, ctype bas
...
+		// Return the zero value of data
+		return data, fmt.Errorf("unable to query account data for address %v : %w", addr, err)
+	}
+
+	return
+}
+
+func lookupExt(rdb db.Accessor, addr basics.Address, extension func(*sql.Stmt, *ledgercore.PersistedAccountData) error) (data dbAccountData, err error) {
+	rdb.Atomic(func(ctx context.Context, tx *sql.Tx) error {
+		lookupStmt, err := tx.Prepare(lookupAcctBaseQuery)
```
tsachiherman, line 969: don't forget to `Close` the prepared statement once you no longer need it.

ledger/acctupdates.go
```diff
@@ -1794,7 +1814,10 @@ func (au *accountUpdates) lookupWithRewards(rnd basics.Round, addr basics.Addres
 			// Check if this is the most recent round, in which case, we can
 			// use a cache of the most recent account state.
 			if offset == uint64(len(au.deltas)) {
-				return macct.data, nil
+				if full && macct.data.ExtendedAssetHolding.Count != 0 {
+					macct.data.Assets, macct.data.ExtendedAssetHolding, err = au.accountsq.loadHoldings(macct.data.ExtendedAssetHolding)
```
tsachiherman, line 1818: in order to make sure that the return is valid, you need to join it with the round number and compare it with `currentDbRound`. Otherwise, you might have just created a mixed-up balance record of old account data with new extended holding.

ledger/acctupdates.go
```diff
@@ -1911,7 +1992,16 @@ func (au *accountUpdates) lookupWithoutRewards(rnd basics.Round, addr basics.Add
 		// present in the on-disk DB.  As an optimization, we avoid creating
 		// a separate transaction here, and directly use a prepared SQL query
 		// against the database.
-		dbad, err = au.accountsq.lookup(addr)
+
+		if extension != nil {
+			dbad, err = lookupExt(au.dbs.Rdb, addr, extension)
```
tsachiherman, line 1997: it's abit hard to follow, but I'm not sure if the extension method here is called from the *same* transaction or not. It's vital to call it from the same read transaction - otherwise, the `dbad.round` comparison below would not be sufficient.

ledger/apply/asset.go
tsachiherman: sure;
How can you tell that you did not forget to initialize one of the fields in the `deleted` object ?
Creating a wrapper function around that would ensure that you won't be forgetting to initialize all the field members, even when you add a new field to the structure.

ledger/ledgercore/statedelta.go
tsachiherman: yes, I realize that.. I'm wondering why wound't we add this map to the `PersistedBalanceRecord` instead ?

tsachiherman (changes requested):
Few open questions above; one particular issue is the missing atomicity verification on database reads.

ledger/acctupdates.go
algorandskiy: not sure I understand. `au.accountsq.lookup` uses prepared queries from `au.dbs.Rdb`, same way as `lookupExt` does. `lookupExt` prepares and runs two queries on the same transaction derived from `au.dbs.Rdb.Atomic`. It looks like as a complete equivalent of the old code.

ledger/accountdb.go
```diff
@@ -900,31 +933,86 @@ func (qs *accountsDbQueries) lookupCreator(cidx basics.CreatableIndex, ctype bas
...
+	if err == sql.ErrNoRows {
+		// Return the zero value of data
+		return data, fmt.Errorf("unable to query account data for address %v : %w", addr, err)
+	}
+
+	return
+}
+
+func lookupExt(rdb db.Accessor, addr basics.Address, extension func(*sql.Stmt, *ledgercore.PersistedAccountData) error) (data dbAccountData, err error) {
+	rdb.Atomic(func(ctx context.Context, tx *sql.Tx) error {
```
tsachiherman, line 968: You'll need to add `err = rdb.Atomic` in order to catch `Commit` errors as well.

tsachiherman: Also, I think that you'd like to use `db.Retry` here.

ledger/acctupdates.go
tsachiherman: ah - I missed that. Thanks.

ledger/apply/asset.go
algorandskiy: Ok, I'll submit a separate PR to master with this fix since it is not related to this PR.

ledger/ledgercore/statedelta.go
algorandskiy: It is possible but `PersistedBalanceRecord` is also used in eval for accounts prefetching and in that context it does not need holdings delta at all. I'll revisit your suggestion as move forward with params implementation.

ledger/acctupdates.go
```diff
@@ -1856,9 +1888,18 @@ func (au *accountUpdates) lookupWithRewards(rnd basics.Round, addr basics.Addres
...
+				goto retryAcquireLock
+			}
 		} else {
 			dbad, err = au.accountsq.lookup(addr)
 		}
+
+		if err != nil {
+			return ledgercore.PersistedAccountData{}, err
+		}
+
```
algorandskiy, line 1902: not sure we miss this error check in master, added

ledger/acctupdates.go
algorandskiy: done

ledger/acctupdates.go
```diff
@@ -1825,26 +1894,88 @@ func (au *accountUpdates) lookupWithRewards(rnd basics.Round, addr basics.Addres
 		// present in the on-disk DB.  As an optimization, we avoid creating
 		// a separate transaction here, and directly use a prepared SQL query
 		// against the database.
-		dbad, err = au.accountsq.lookup(addr)
+		if full {
+			dbad, err = lookupFull(au.dbs.Rdb, addr)
+			var mismatchRoundErr *MismatchingDatabaseRoundError
+			if errors.As(err, &mismatchRoundErr) {
+				goto retryAcquireLock
+			}
```
algorandskiy, line 1902: MismatchingDatabaseRoundError might happen when round advances between accountbase and accountext reading, not sure to to test it.

ledger/acctupdates.go
```diff
@@ -1814,7 +1874,16 @@ func (au *accountUpdates) lookupWithRewards(rnd basics.Round, addr basics.Addres
...
 			au.baseAccounts.writePending(macct)
-			return macct.pad.AccountData, nil
+			if full && macct.pad.ExtendedAssetHolding.Count != 0 {
+				macct.pad.Assets, macct.pad.ExtendedAssetHolding, err = au.accountsq.loadHoldings(macct.pad.ExtendedAssetHolding, macct.round)
+				var mismatchRoundErr *MismatchingDatabaseRoundError
+				if errors.As(err, &mismatchRoundErr) {
+					// go to waiting
+					// note, au.accountsMu is still held
+					goto retryLocked
+				}
```
algorandskiy, line 1884: MismatchingDatabaseRoundError might happen when `au.baseAccounts` has older data than DB if round advances between `au.baseAccounts` and accountext reading. Retrying should sufficient - on next iteration we either do not hit this branch or data becomes synchronized.

ledger/acctupdates.go
tsachiherman: yes, I agree it might be a hard thing to test.

ledger/acctupdates.go
tsachiherman: I believe that if you just want to check if the error is a specific error, a type assertion would suffice. But using `errors.As` would not be incorrect.

tsachiherman (approved):

PR #1937, opened 2021-03-02
1k+ assets (part 6)

algorandskiy:
## Summary

Use `PersistedAccountData` in `StateDelta` and everywhere internally inside `ledger` package. Use holding delta information for asset holding group data reconciliation.

## Test Plan

Added tests verifying account updates.


algorandskiy:
Merged with #1936

PR #1940, opened 2021-03-04
Reconstruct balances database from blocks

tsachiherman:
## Summary

During the accounts update initialization, we call the initializeCaches method to preload the most recent 320 rounds.
These 320 rounds are the difference between where the accounts database round and the latest blocks database round.

If (from any reason), the difference is more than 320 rounds, then all the rounds in that range are being loaded into memory. This, in turn, creates an issue when we have to recover from a large gap, as it would run out of memory.

To address this issue, this PR reconstruct the accounts database in "chunks" of 1000 rounds. This method prevent the node from going into a situation where it needs to handle large quantity of rounds sitting in memory.

## Test Plan

Unit test added. Verified manually that recovered account database end up generating the same catchpoints as on published mainnet.

### DIFF
--- a/ledger/acctupdates.go
+++ b/ledger/acctupdates.go
@@ -94,7 +94,7 @@ const initializeCachesRoundFlushInterval = 1000
 const initializingAccountCachesMessageTimeout = 3 * time.Second
 
 // accountsUpdatePerRoundHighWatermark is the warning watermark for updating accounts data that takes
-// longer then expected. We set it up here for one second per round, so that if we're bulk updating
+// longer than expected. We set it up here for one second per round, so that if we're bulk updating
 // four rounds, we would allow up to 4 seconds. This becomes important when supporting balances recovery
 // where we end up batching up to 1000 rounds in a single update.
 const accountsUpdatePerRoundHighWatermark = 1 * time.Second
@@ -993,8 +993,9 @@ func (au *accountUpdates) initializeCaches(lastBalancesRound, lastestBlockRound,
 		// flush to disk if any of the following applies:
 		// 1. if we have loaded up more than initializeCachesRoundFlushInterval rounds since the last time we flushed the data to disk
 		// 2. if we completed the loading and we loaded up more than 320 rounds.
-		if blk.Round()-lastFlushedRound > initializeCachesRoundFlushInterval ||
-			(lastestBlockRound == blk.Round() && lastBalancesRound+basics.Round(blk.ConsensusProtocol().MaxBalLookback) < lastestBlockRound) {
+		flushIntervalExceed := blk.Round()-lastFlushedRound > initializeCachesRoundFlushInterval
+		loadCompleted := (lastestBlockRound == blk.Round() && lastBalancesRound+basics.Round(blk.ConsensusProtocol().MaxBalLookback) < lastestBlockRound)
+		if flushIntervalExceed || loadCompleted {
 			// adjust the last flush time, so that we would not hold off the flushing due to "working too fast"
 			au.lastFlushTime = time.Now().Add(-balancesFlushInterval)
 
--- a/ledger/acctupdates.go
+++ b/ledger/acctupdates.go
@@ -967,6 +967,16 @@ func (au *accountUpdates) initializeCaches(lastBalancesRound, lastestBlockRound,
 	const accountsCacheLoadingMessageInterval = 5 * time.Second
 	lastProgressMessage := time.Now().Add(-accountsCacheLoadingMessageInterval / 2)
 
+	// rollbackSynchronousMode ensures that we switch to "fast writing mode" when we start flushing out rounds to disk, and that
+	// we exit this mode when we're done.
+	rollbackSynchronousMode := false
+	defer func() {
+		if rollbackSynchronousMode {
+			// restore default synchronous mode
+			au.dbs.Wdb.SetSynchronousMode(context.Background(), au.synchronousMode, au.synchronousMode >= db.SynchronousModeFull)
+		}
+	}()
+
 	for blk := range blocksStream {
 		delta, err = au.ledger.trackerEvalVerified(blk, &accLedgerEval)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


ledger/acctupdates.go
```diff
@@ -896,28 +926,140 @@ func (au *accountUpdates) initializeCaches(lastBalancesRound, lastestBlockRound,
...
+			au.accountsMu.Unlock()
+
+			// flush the account data
+			au.committedUpTo(blk.Round())
+
+			// wait for the writing to complete.
+			au.waitAccountsWriting()
+
+			// The au.dbRound after writing should be ~320 behind the block round.
+			roundsBehind := blk.Round() - au.dbRound
```
algorandskiy, line 1025: Is it true for the case when accounts behind blocks for 1M rounds? If not how would not the condition below about the process?

ledger/acctupdates.go
```diff
@@ -896,28 +926,140 @@ func (au *accountUpdates) initializeCaches(lastBalancesRound, lastestBlockRound,
...
+	// we exit this mode when we're done.
+	rollbackSynchronousMode := false
+	defer func() {
+		if rollbackSynchronousMode {
+			// restore default synchronous mode
+			au.dbs.Wdb.SetSynchronousMode(context.Background(), au.synchronousMode, au.synchronousMode >= db.SynchronousModeFull)
 		}
+	}()
 
+	for blk := range blocksStream {
```
algorandskiy, line 980: I probably miss something but how is reading from the `blocksStream` chan different from have a loop `for roundNumber := lastBalancesRound + 1; roundNumber <= lastestBlockRound; roundNumber++` ?

ledger/acctupdates.go
tsachiherman: on "normal" use case, after writing the accounts for a given block, the au.dbRound would be at most 320 rounds behind that block. The goal of this test is to see that there hasn't been any errors generated during the `commitRound` execution. If there were any, there is no point in keep going.
As the node below suggest, a catchpoint can tweak that, so we need to take that into account as well. This "issue" comes because we're not calling `committedUpTo` for every round, but rather in 1000 rounds "jumps".

ledger/acctupdates_test.go
```diff
@@ -1545,3 +1545,95 @@ func TestReproducibleCatchpointLabels(t *testing.T) {
...
+	ml2 := makeMockLedgerForTracker(t, true, int(initialRounds), protocolVersion)
+	ml2.log.SetLevel(logging.Warn)
+	defer ml2.Close()
+
+	// and "fix" it to contain the blocks and deltas from before.
+	ml2.blocks = ml.blocks
+	ml2.deltas = ml.deltas
+
+	au = &accountUpdates{}
+	au.initialize(config.GetDefaultLocal(), ".", proto, accts[0])
```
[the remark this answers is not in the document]
tsachiherman: The au would start a new tracker database ( i.e. the one defined in `ml2` ), and a pseudo blocks database. The blocks database comes "pre-loaded" with many blocks.. 
The `lastestBlockRound` is taken from the `l.Latest()` and not from the tracker database.

ledger/acctupdates.go
```diff
@@ -896,28 +926,141 @@ func (au *accountUpdates) initializeCaches(lastBalancesRound, lastestBlockRound,
...
+	writeAccountCacheMessageCompleted := make(chan struct{})
+	defer func() {
+		close(skipAccountCacheMessage)
+		select {
+		case <-writeAccountCacheMessageCompleted:
+			au.log.Infof("initializeCaches completed initializing account data caches")
+		default:
+		}
+	}()
+	go func() {
```
algorandskiy, line 939: please add a comment like "this goroutine logs a message once if the parent function have not completed in `initializingAccountCachesMessageTimeout` seconds

ledger/acctupdates.go
```diff
@@ -896,28 +926,141 @@ func (au *accountUpdates) initializeCaches(lastBalancesRound, lastestBlockRound,
...
+			lastFlushedRound = blk.Round()
+		}
+
+		// if enough time have passed since the last time we wrote a message to the log file then give the user an update about the progess.
+		if time.Now().Sub(lastProgressMessage) > accountsCacheLoadingMessageInterval {
+			// drop the initial message if we're got to this point since a message saying "still initializing" that comes after "is initializing" doesn't seems to be right.
+			select {
+			case skipAccountCacheMessage <- struct{}{}:
+				// if we got to this point, we should be able to close the writeAccountCacheMessageCompleted channel to have the "completed initializing" message written.
+				close(writeAccountCacheMessageCompleted)
```
algorandskiy, line 1050: why do we close `writeAccountCacheMessageCompleted` here and in goroutine above? it will trigger "completed" message on exit but in fact we can fail just on another block applying iteration (since we are still in a loop at the moment of closing the channel).

algorandskiy (commented):
Thank you for explanation, two more questions on log file updating

ledger/acctupdates.go
tsachiherman: ahh. nice corner case.
I've added
```
if err == nil {
```
around the logging of the `completed` message, so that in case of an error, it wouldn't write that.

algorandskiy (approved):
Functionally looks OK. I do not like additional complexity added by logging. Maybe implement "initCacheLogger" and move all this additional logging code there and leaving only
```
defer logger.complete()
go logger.progress()
//...
if time.Now().Sub(lastProgressMessage) > accountsCacheLoadingMessageInterval { 
logger.triggerProgress()
}
```
or something similar?

tsachiherman:
> Functionally looks OK. I do not like additional complexity added by logging. Maybe implement "initCacheLogger" and move all this additional logging code there and leaving only this:
> 
> ```go
> defer logger.complete()
> go logger.progress()
> //...
> if time.Now().Sub(lastProgressMessage) > accountsCacheLoadingMessageInterval { 
> logger.triggerProgress()
> }
> ```
> 
> or something similar? The main intent is to have less code in the function to make clearer for undestanding.

Sounds like a good idea. I will create a ticket for that so we could implement it separately.

PR #1941, opened 2021-03-04
Fixes for KMD on Windows

mxmauro:
This PR aims to fix different Windows issues that appeared while attempted to use wallets.

1. Windows uses a totally different security model for directories than *nix like OSes.

On this release, the check will be bypassed but the installer will be changed to add the proper access control list.

2. Windows 10 build 16257 added the ANSI compatible terminal which is not enabled by default. Sending escape codes will simply add garbage to the output.

This behavior was only seen when an account is created and the mnemonic displayed. The colorization was removed in Windows builds.

3. `Signal(0)` handling. *nix OSes can use `signal(0)` to check if a process is valid and if the caller can send signals to it.

Code was modified to enforce a similar behavior on Windows.


mxmauro:
Closing because undesired history mixing.

PR #1942, opened 2021-03-04
Fixes for KMD on Windows

mxmauro:
This PR aims to fix different Windows issues that appeared while attempted to use wallets.

1. Windows uses a totally different security model for directories than *nix like OSes.
On this release, the check will be bypassed but the installer will be changed to add the proper access control list.

2. Windows 10 build 16257 added the ANSI compatible terminal which is not enabled by default. Sending escape codes will simply add garbage to the output.
This behavior was only seen when an account is created and the mnemonic displayed. The colorization was removed in Windows builds.

3. `Signal(0)` handling. *nix OSes can use signal(0) to check if a process is valid and if the caller can send signals to it.
Code was modified to enforce a similar behavior on Windows.

### DIFF
--- a/cmd/goal/wallet.go
+++ b/cmd/goal/wallet.go
@@ -21,6 +21,7 @@ import (
 	"bytes"
 	"fmt"
 	"os"
+	"runtime"
 	"strings"
 
 	"github.com/spf13/cobra"
@@ -166,7 +167,11 @@ var newWalletCmd = &cobra.Command{
 
 				// Display the mnemonic to the user
 				reportInfoln(infoPrintedBackupPhrase)
-				reportInfof(infoBackupPhrase, mnemonic)
+				if runtime.GOOS == "windows" {
+					reportInfof(infoBackupPhraseWindows, mnemonic)
+				} else {
+					reportInfof(infoBackupPhrase, mnemonic)
+				}
 			}
 		}
 
--- a/cmd/goal/wallet.go
+++ b/cmd/goal/wallet.go
@@ -21,7 +21,6 @@ import (
 	"bytes"
 	"fmt"
 	"os"
-	"runtime"
 	"strings"
 
 	"github.com/spf13/cobra"
@@ -167,11 +166,7 @@ var newWalletCmd = &cobra.Command{
 
 				// Display the mnemonic to the user
 				reportInfoln(infoPrintedBackupPhrase)
-				if runtime.GOOS == "windows" {
-					reportInfof(infoBackupPhraseWindows, mnemonic)
-				} else {
-					reportInfof(infoBackupPhrase, mnemonic)
-				}
+				reportInfof(infoBackupPhrase, mnemonic)
 			}
 		}
 
--- a/nodecontrol/kmdControl.go
+++ b/nodecontrol/kmdControl.go
@@ -200,8 +200,7 @@ func (kc *KMDController) StartKMD(args KMDStartArgs) (alreadyRunning bool, err e
 			logging.Base().Errorf("%s: kmd data dir exists but is not a directory", kc.kmdDataDir)
 			return false, errors.New("bad kmd data dir")
 		}
-		if !isDirectorySafe(dataDirStat) {
-			logging.Base().Errorf("%s: kmd data dir exists but is too permissive (%o), change to (%o)", kc.kmdDataDir, dataDirStat.Mode()&0777, DefaultKMDDataDirPerms)
+		if !kc.isDirectorySafe(dataDirStat) {
 			return false, errors.New("kmd data dir not secure")
 		}
 	} else {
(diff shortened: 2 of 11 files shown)


nodecontrol/kmdControl_non_windows.go
```diff
@@ -0,0 +1,29 @@
+// +build !windows
```
algorandskiy, line 1: call the file `_unix.go` or `_common.go` as other non-windows files in the repo

algorandskiy (commented):
a couple of naming remarks

nodecontrol/kmdControl_non_windows.go
mxmauro: Done. Also moved the `+build` option below the copyright notice to match the location of other files

cmd/goal/wallet.go
```diff
@@ -166,7 +167,11 @@ var newWalletCmd = &cobra.Command{
 
 				// Display the mnemonic to the user
 				reportInfoln(infoPrintedBackupPhrase)
-				reportInfof(infoBackupPhrase, mnemonic)
+				if runtime.GOOS == "windows" {
+					reportInfof(infoBackupPhraseWindows, mnemonic)
+				} else {
+					reportInfof(infoBackupPhrase, mnemonic)
+				}
```
tsachiherman, line 174: instead of having the difference here, define two constants for `\x1B[32m` and `\033[0m` and use these.
The constants should be defined separately for windows and *nix.

nodecontrol/kmdControl.go
```diff
@@ -200,7 +200,7 @@ func (kc *KMDController) StartKMD(args KMDStartArgs) (alreadyRunning bool, err e
 			logging.Base().Errorf("%s: kmd data dir exists but is not a directory", kc.kmdDataDir)
 			return false, errors.New("bad kmd data dir")
 		}
-		if (dataDirStat.Mode() & 0077) != 0 {
+		if !isDirectorySafe(dataDirStat) {
```
tsachiherman, line 203: add the logging line to the non-windows implementation. ( so that all the special permissions testing and logging would remain there )

nodecontrol/kmdControl_windows.go
```diff
@@ -0,0 +1,25 @@
...
+
+package nodecontrol
+
+import (
+	"os"
+)
+
+func isDirectorySafe(_ os.FileInfo) bool {
+	return true
+}
```
tsachiherman, line 25: add end of line

util/process_windows.go
```diff
@@ -19,23 +19,88 @@
...
+	}
 
-		err = p.Kill()
+	return process, nil
+}
+
+// KillProcess kills a running OS process
+func KillProcess(pid int, signal os.Signal) error {
+	p, err := FindProcess(pid)
+	if err == nil {
```
tsachiherman, line 74: coding style:
avoid deep branching and exit early.
i.e. use 
```
if err != nil {
    return err
}
...
```

util/process_windows.go
```diff
@@ -65,3 +130,12 @@ func getChildrenProcesses(parentPid int) []*os.Process {
...
 }
+
+func waitUntilProcessEnds(pid int) {
+	h, err := syscall.OpenProcess(syscall.SYNCHRONIZE, false, uint32(pid))
+	if err == nil {
+		_, _ = syscall.WaitForSingleObject(h, processTerminateWaitInMs)
+
+		_ = syscall.CloseHandle(h)
+	}
+}
```
tsachiherman, line 141: add end of line

util/process_windows.go
```diff
@@ -19,23 +19,88 @@
...
+			return nil, nil
 		}
+	}
 
-		err = p.Kill()
+	return process, nil
+}
+
+// KillProcess kills a running OS process
+func KillProcess(pid int, signal os.Signal) error {
```
tsachiherman, line 72: Your `KillProcess` implementation could be a bit better:
* When killing processes, consider that there could be a deep nested tree. Your implementation currently only handle the first layer of child processes.
* When killing processes, you'd better try to terminate them all at once, without blocking.
* Last, after you killed all the child processes, you should reiterate on the child processes to ensure no new children were spawn during the function execution.

tsachiherman (changes requested):
Overall looks promising. I've added few requests for various improvements.

nodecontrol/kmdControl.go
mxmauro: Hi @tsachiherman , I don't understand what do you mean with this. The `kmd data dir exists but is too permissive` is still there.

tsachiherman: what I meant was to move the line
```
logging.Base().Errorf("%s: kmd data dir exists but is too permissive (%o), change to (%o)", kc.kmdDataDir, dataDirStat.Mode()&0777, DefaultKMDDataDirPerms)
```
from kmdControl.go into kmdControl_common.go, so that all directory-mode bit masks would be handled in there.

mxmauro: Got it. Tomorrow morning I'll do all the changes

tsachiherman (approved):

PR #1944, opened 2021-03-05
comments. add lock to test to prevent rare race

brianolson:
## Summary

Comments to help understand the code better. Rename 'compare' to 'sameParams' to be more descriptive.
Add a lock to a test to prevent rare race crash.

## Test Plan

Ran tests many times, race crash hasn't happened since adding lock.

### DIFF
--- a/txnsync/emulatorTimer_test.go
+++ b/txnsync/emulatorTimer_test.go
@@ -57,6 +57,8 @@ func (g *guidedClock) TimeoutAt(delta time.Duration) <-chan time.Time {
 		close(c)
 		return c
 	}
+	g.mu.Lock()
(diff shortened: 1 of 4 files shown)


tsachiherman (approved):
looks great.

txnsync/emulatorTimer_test.go
```diff
@@ -92,6 +94,7 @@ func (g *guidedClock) Advance(adv time.Duration) {
 		ch       chan time.Time
 	}
 	expiredClocks := []entryStruct{}
+	g.mu.Lock()
```
tsachiherman, line 97: I wonder why I didn't ran into these when running it with "-race". but it's perfectly fine regardless.

PR #1945, opened 2021-03-05
Improve emulator logging & send request on offset change.

tsachiherman:
## Summary

This PR handles two separate issues:
1. On the non-relay, it sends a message every time the offset for the node changes. ( i.e. ~4 times a second).
    This helps to reduce the amount of redundant transactions are sent to the non-relay.
2. Added a destination to the logged message so that we can see on the emulation the name of the destination node.

## Test Plan

Unit tests were updated.


brianolson (approved):
LGTM. This gives me ideas that I'll probably put in a follow on PR.

PR #1947, opened 2021-03-05
Improve txnsync emulator performance and memory profile

tsachiherman:
## Summary

Avoid using the `deadlock.Mutex` in the emulator, and use a channel instead.
The `deadlock.Mutex` was consuming considerable amount of memory, due to it's internally allocated timer.

edit: following on rare data races found on travis, I avoided using the main state's clock in the async methods. The clock is being updated once every round, so we can't really control that unless we add a synchronization primitive or find a better way.

## Test Plan

This is a test.

### DIFF
--- a/txnsync/emulatorNode_test.go
+++ b/txnsync/emulatorNode_test.go
@@ -22,8 +22,6 @@ import (
 	"sync/atomic"
 	"time"
 
-	"github.com/algorand/go-deadlock"
-
 	"github.com/algorand/go-algorand/crypto"
 	"github.com/algorand/go-algorand/data/basics"
 	"github.com/algorand/go-algorand/data/transactions"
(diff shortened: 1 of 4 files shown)


brianolson:
How's the performance if we just don't use the deadlock library? That would seem simpler and more foolproof than reinventing sync.Mutex

tsachiherman:
The deadlock library does have it's pros, but we need to be careful when
picking the syncronization primitive for the application.

If you can spot another location that has high frequency of
locking/unlocking, then i would think that it could be a good option.

brianolson:
I still think it's better to use standard library sync.Mutex instead of reinventing mutex behavior with a channel. I think we need to add some kind of pragma-comment to the always-use-deadlock checker to exempt this usage.

tsachiherman:
I like the idea of adding a pragma like to ignore the dead.mutex
enforcement. I think that it's an overkill for a unit test, though.

brianolson (approved):

PR #1948, opened 2021-03-05
create a msgStat object so that we don't have to sscanf(sprintf())

brianolson:
## Summary

Create a msgStat{} object for logging to unify rx and tx logging and to make debug logging better without the need for sscanf(sprintf())

## Test Plan

Minor logging tweaks. Existing tests still pass.

### DIFF
--- a/txnsync/mainloop.go
+++ b/txnsync/mainloop.go
@@ -302,3 +302,55 @@ func (s *syncState) updatePeersRequestParams(peers []*Peer) {
 		}
 	}
 }
+
+func (s *syncState) logMsgStats(mode msgMode, mstat msgStats) {
+	var modeString string
+	var tofrom string
+	switch mode {
+	case modeIncoming:
+		modeString = "Incoming"
+		tofrom = "from"
+	case modeOutgoing:
+		modeString = "Outgoing"
+		tofrom = "to"
+	}
+
+	s.log.Infof(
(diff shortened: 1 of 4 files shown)


tsachiherman:
I like the idea, I will have some code styling comments later on.

tsachiherman (changes requested):
I think that the idea is good, but we need to make it more structured:
* Create a logger object for the txsync, which will be in txnsync/logger.go
* The logger object would implement all the logging.Logger functions by streaming them to the base logger passed in the MakeTranscationSyncService
* The logger object would implement an interface "txnsync.Logger" which would have the OutgoingMessage and IncomingMessage function.
* The rest should align with your existing implementation.

txnsync/mainloop.go
```diff
@@ -302,3 +302,55 @@ func (s *syncState) updatePeersRequestParams(peers []*Peer) {
...
+		mr.printMsgStats(mode, &mstat)
+	}
+}
+
+type msgMode int
+
+const (
+	modeZero     msgMode = 0
+	modeIncoming msgMode = iota
+	modeOutgoing msgMode = iota
```
tsachiherman, line 341: the common pattern is to define the first one as "iota", and leave the rest blank. i.e. 
```
modeZero     msgMode = iota
modeIncoming
modeOutgoing
```

tsachiherman (approved):
looks good.

brianolson:
integration tests found a race in usage of syncState.clock
I may investigate that more today.

PR #1955, opened 2021-03-09
Fix parameter check in abort procedure

algonautshant:
## Summary

When a test fails and the abort routine is called, some variables used there may not be defined. This causes the script to report error about that. This is not related to the test failure and is a test environment error that should not be there. 

This change first checks if the variable is defined. 

## Test Plan

This is on an exception path that does not get exercised. I tested it by forcing a failure to verify this fixes the problem.

### DIFF
--- a/test/e2e-go/cli/goal/expect/goalExpectCommon.exp
+++ b/test/e2e-go/cli/goal/expect/goalExpectCommon.exp
@@ -46,7 +46,7 @@ package require Tcl 8.0
 proc ::AlgorandGoal::Abort { ERROR } {
     puts "Aborting with Error: $ERROR"
 
-    if { "$::GLOBAL_TEST_ALGO_DIR" != "" && "$::GLOBAL_TEST_ROOT_DIR" != "" } {
+    if { [info exists ::GLOBAL_TEST_ALGO_DIR] && [info exists ::GLOBAL_TEST_ROOT_DIR] } {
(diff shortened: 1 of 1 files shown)


tsachiherman (changes requested):
I don't think this is a good fix.. It will prevent the expect error, but would leave the network running, which would miss the point of having this method to start with. Instead, I would suggest the following:

```
proc ::AlgorandGoal::Abort { ERROR } {
     puts "Aborting with Error: $ERROR"

    if { [info exists ::GLOBAL_TEST_ROOT_DIR] } {
        # terminate child algod processes, if there are active child processes the test will hang on a test failure
        puts "GLOBAL_TEST_ROOT_DIR $::GLOBAL_TEST_ROOT_DIR"
        puts "GLOBAL_NETWORK_NAME $::GLOBAL_NETWORK_NAME"
        ::AlgorandGoal::StopNetwork $::GLOBAL_NETWORK_NAME $::GLOBAL_TEST_ROOT_DIR
    }

    if { [info exists ::GLOBAL_TEST_ALGO_DIR] } {
        puts "GLOBAL_TEST_ALGO_DIR $::GLOBAL_TEST_ALGO_DIR"
        ::AlgorandGoal::StopNode $::GLOBAL_TEST_ROOT_DIR
    }
    exit 1
}

```

tsachiherman (approved):

PR #1956, opened 2021-03-09
locally allow sync.Mutex

brianolson:
## Summary

use comments to locally allow sync.Mutex instead of rewriting it to the much slower deadlock.Mutex

```go
// algofix allow sync
// algofix require deadlock
```

## Test Plan

Added unit test of deadlock replacement, normal and with new comment directives

### DIFF
--- a/cmd/algofix/deadlock.go
+++ b/cmd/algofix/deadlock.go
@@ -18,6 +18,7 @@ package main
 
 import (
 	"go/ast"
+	"strings"
 )
 
 func init() {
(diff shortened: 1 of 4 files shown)


tsachiherman:
I think that the implementation is not sufficiently strict, allowing unintended declaration as a code-placement.
( i.e. it opens up to copy&paste errors resulting from changing of struct definitions within the file )

My suggestion is that we would use tags for that : 
i.e. mu sync.Mutex `algofix:allow sync.Mutex`
which has a "default" scope of a single variable.

tsachiherman (approved):
looks good.

PR #1957, opened 2021-03-10
Fix for wrong channel when packaging.

egieseke:
## Summary

If the channel is set, don't reset.

## Test Plan

Tested with Jenkins pipeline


onetechnical (approved):
This looks fine to me.

algobarb (approved):
Seems like a simple change

bricerisingalgorand (approved):

PR #1958, opened 2021-03-10
add to Peer interface DataGet and DataPut

brianolson:
for misc service/app data associated with a peer

## Summary

For use by future expansion on protocols, add misc data to a Peer, which goes away when the peer closes.

## Test Plan

Code is pretty trivial, but I should add at least one unit test...

### DIFF
--- a/network/ping.go
+++ b/network/ping.go
@@ -28,7 +28,7 @@ func pingHandler(message IncomingMessage) OutgoingMessage {
 	if len(message.Data) > 8 {
 		return OutgoingMessage{}
 	}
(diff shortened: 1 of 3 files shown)


tsachiherman (commented):
I think you'd better off just coping the getPeerData/setPeerData from the txnsync branch, and add unit testing for that. your implementation also adds that to the wsPeerCore ( which is used by the catchup as HTTPPeer ) instead of the wsPeer ( i.e. UnicastPeer ), which is where we need it.

brianolson:
already implemented on feature/txnsync branch

brianolson:
already implemented on feature/txnsync branch

PR #1959, opened 2021-03-11
Txn cache cleanup

brianolson:
## Summary

Rename 'transactionLru' to 'transactionCache' because it is a FIFO policy cache, not LRU. Add unit test.

## Test Plan

Adds unit test.


tsachiherman (approved):
looks good to me.

PR #1962, opened 2021-03-11
allow struct field to be sync.Mutex if tagged

brianolson:
## Summary

Allow struct field to be sync.Mutex if tagged `algofix:"allow sync.Mutex"`
deadlock.Mutex can be much slower and some code (in progress on another branch) needs to be specified to use standard library sync.Mutex

## Test Plan

Includes new unit test for algofix of code with sync.Mutex, both replacing with deadlock.Mutex and tagged to remain as sync.Mutex

### DIFF
--- a/cmd/algofix/deadlock.go
+++ b/cmd/algofix/deadlock.go
@@ -18,6 +18,7 @@ package main
 
 import (
 	"go/ast"
+	"strings"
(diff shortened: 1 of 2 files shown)


tsachiherman (approved):
looks good.

PR #1963, opened 2021-03-12
Add DisableLocalhostConnectionRateLimit configuration option

tsachiherman:
## Summary

The current connection rate limit doesn't differentiate between connections that are coming from localhost vs connections that are coming from a remote machine. This means that the connection rate limit is unable to distinguish between multiple hosts running locally. This becomes a problem, since when running a local host network, one node can saturate the connection limits, and prevent other nodes from connecting to a pseudo relay.

To work around that, a flag was added `DisableLocalhostConnectionRateLimit` that would allow us to skip the connection limit for incoming connection whose IP address is the localhost. To retain backward compatibility, the flag is still there, and users can toggle it off.

## Test Plan

Update existing unit test; test manually against large local network.


network/requestTracker.go
```diff
@@ -501,3 +507,13 @@ func (rt *RequestTracker) getForwardedConnectionAddress(header http.Header) (ip
 	}
 	return
 }
+
+// isLocalhost returns true if the given host is a localhost address.
+func isLocalhost(host string) bool {
+	for _, v := range []string{"localhost", "127.0.0.1", "[::1]", "::1"} {
```
brianolson, line 513: Go is also currently representing localhost as "[::]"
(this is what I see in a algod.net, `[::]:4160`)

brianolson (commented):
I like it. I'm pretty sure there are times I would have used this in some dev testing.
I think we need at least one more string for 'localhost' `[::]`

network/requestTracker.go
```diff
@@ -285,8 +285,11 @@ func (rt *RequestTracker) Accept() (conn net.Conn, err error) {
 		rt.hostRequests.pruneRequests(rateLimitingWindowStartTime)
 		originConnections := rt.hostRequests.countOriginConnections(trackerRequest.remoteHost, rateLimitingWindowStartTime)
 
+		remoteHostIsNonLocal := (!rt.config.DisableLocalhostConnectionRateLimit) || (!isLocalhost(trackerRequest.remoteHost))
```
algonautshant, line 288: I prefer naming this rateLimitedRemoteHost
This can be local host but reads remote Host Is Non Local.

algonautshant (approved):
Looks good.
It is a good idea to expand the bandwidth for local connections.

brianolson (dismissed):

tsachizehub (approved):

PR #1965, opened 2021-03-12
Improve ConsensusVersion errors

tsachiherman:
## Summary

Improve ConsensusVersion error output to include a correct Committed round.
In particular - the `ledgercore.ErrNoEntry` has a field `Committed` which wasn't initialized correctly.

## Test Plan

Unit test updated. Providing a test for this change is really hard due to the fact that the underlaying Ledger object is on a different package than the data.Ledger, which makes it hard to mock this. The dependency between the two is also using struct embedding, which prevents mocking.


ledger/README.md
```diff
@@ -26,7 +26,8 @@ The ledger exposes the following functions for managing the blocks:
 
 - `Latest()` returns the last block added to the ledger.
 
-- `LatestCommitted()` returns the last block written to durable storage.
+- `LatestCommitted()` returns the last block written to durable storage
+  as well as the latest block added to the ledger.
```
algonautshant, line 30: Maybe, to be more precise:
returns **the round of** the last block written to durable storage as well as **the round of the** latest block added to the ledger

ledger/ledger.go
```diff
@@ -504,8 +504,10 @@ func (l *Ledger) Latest() basics.Round {
 
 // LatestCommitted returns the last block round number written to
 // persistent storage.  This block, and all previous blocks, are
-// guaranteed to be available after a crash.
-func (l *Ledger) LatestCommitted() basics.Round {
+// guaranteed to be available after a crash. In addition, it returns
+// the latest block added to the ledger ( which will be flushed
```
algonautshant, line 508: the latest block **round number** added to the ledger

algonautshant (dismissed):
Looks good.
Couple of comments on the comments.

tsachizehub (approved):

PR #1966, opened 2021-03-13
Add networking infrastructure for cancelling sends

nicholasguoalgorand:
## Summary

When we send a large message (ie a proposal payload), sometimes we realize after starting sending the message that we don't need to send the message (ie the proposal isn't the best one, the payload is malformed). Thus we wish to have support to be able to cancel sending the message after it is enqueued / starting to be sent.

This pr adds the functionality of broadcasting an array of messages, and the ability to cancel sending after any subset of messages in the array is sent.

## Test Plan

Wrote a few unit tests, also reran other unit / integration tests


network/wsPeer.go
```diff
@@ -20,6 +20,7 @@ import (
 	"context"
 	"encoding/binary"
 	"fmt"
+	"github.com/algorand/go-algorand/logging"
```
tsachiherman, line 23: move to the correct section.

network/wsPeer.go
```diff
@@ -48,15 +49,10 @@ const averageMessageLength = 2 * 1024    // Most of the messages are smaller tha
 const msgsInReadBufferPerPeer = 10
 
 var networkSentBytesTotal = metrics.MakeCounter(metrics.NetworkSentBytesTotal)
-var networkSentBytesByTag = metrics.NewTagCounter("algod_network_sent_bytes_{TAG}", "Number of bytes that were sent over the network per message tag")
```
tsachiherman, line 51: why does your change requires the removal of the metrics ?

network/wsPeer.go
```diff
@@ -534,11 +545,40 @@ func (wp *wsPeer) handleFilterMessage(msg IncomingMessage) {
...
+
+	// wait for another peer to finish sending before sending
+	if msgs.pacer != nil {
+		<-msgs.pacer
+
+		defer func (pacer chan int) {
+			pacer <- 1
+		}(msgs.pacer)
+	}
+
```
tsachiherman, line 564: I'm not a big fan of the pacer concept. It's not scalable, and slow peers would temporarily cause faster peers to get "hang". The better way to approach that is with a callback method.

network/wsPeer.go
nicholasguoalgorand: @yossigi, hey was there a reason these were removed? If not, I will add them back in the right locations.

network/wsPeer.go
nicholasguoalgorand: @yossigi maybe it makes sense to leave the pacer changes out then?

brianolson (approved):
All looks good to me.

network/wsNetwork_test.go
```diff
@@ -289,6 +289,84 @@ func TestWebsocketNetworkUnicast(t *testing.T) {
...
+	data := make([][]byte, 100)
+	for i := range data {
+		tags[i] = protocol.TxnTag
+		data[i] = []byte(string(i))
+	}
+
+	ctx, cancel := context.WithCancel(context.Background())
+
+	netA.BroadcastArray(ctx, tags, data, false, nil)
+	cancel()
```
tsachiherman, line 361: this isn't a reliable way to test this feature. it will randomly fail.
you need to find a way to "hang" the receiver at the 50th message, and then call cancel. Then, when resuming the receiver, it should not end up receiving all the messages.

network/wsNetwork_test.go
```diff
@@ -1184,7 +1266,7 @@ func TestDelayedMessageDrop(t *testing.T) {
 
 	currentTime := time.Now()
 	for i := 0; i < 10; i++ {
-		err := netA.broadcastWithTimestamp(protocol.TxnTag, []byte("foo"), currentTime.Add(time.Hour*time.Duration(i-5)))
+		err := netA.broadcastWithTimestamp(protocol.TxnTag, []byte(string(i)), currentTime.Add(time.Hour*time.Duration(i-5)))
```
tsachiherman, line 1269: hmm. Any particular reason this is needed ?

network/wsNetwork_test.go
```diff
@@ -1302,9 +1384,9 @@ func TestForceMessageRelaying(t *testing.T) {
 
 	// send 5 messages from both netB and netC to netA
 	for i := 0; i < 5; i++ {
-		err := netB.Relay(context.Background(), protocol.TxnTag, []byte{1, 2, 3}, true, nil)
+		err := netB.Relay(context.Background(), protocol.TxnTag, []byte{byte(i), 2, 3}, true, nil)
```
tsachiherman, line 1387: is that because of the filtering you've added ( but excluded from this PR ) ? if so, than this change shouldn't be included here either.

network/wsPeer.go
```diff
@@ -534,11 +551,30 @@ func (wp *wsPeer) handleFilterMessage(msg IncomingMessage) {
 	}
 	var digest crypto.Digest
 	copy(digest[:], msg.Data)
-	//wp.net.log.Debugf("add filter %v", digest)
+	wp.net.log.Debugf("add filter %v", digest)
```
tsachiherman, line 554: please restore this comment. It affect the performance of the method, and generally not really helpful

network/wsPeer.go
```diff
@@ -534,11 +551,30 @@ func (wp *wsPeer) handleFilterMessage(msg IncomingMessage) {
...
 	wp.outgoingMsgFilter.CheckDigest(digest, true, true)
 }
 
-func (wp *wsPeer) writeLoopSend(msg sendMessage) disconnectReason {
+var emptyHash = crypto.Digest{}
```
tsachiherman, line 558: you're not using this.. remove it ?

network/wsPeer.go
```diff
@@ -534,11 +551,30 @@ func (wp *wsPeer) handleFilterMessage(msg IncomingMessage) {
...
+
+func (wp *wsPeer) writeLoopSend(msgs sendMessages) disconnectReason {
+	for i, msg := range msgs.msgs {
+		select {
+		case <-msg.ctx.Done():
+			logging.Base().Infof("cancelled large send, msg %v out of %v", i, len(msgs.msgs))
```
tsachiherman, line 564: Avoid logging messages in the network library. Enable these while you're working on a concrete test. 
( it's ok to leave these commented out in the code for that purpose )

network/wsPeer.go
```diff
@@ -622,24 +658,53 @@ func (wp *wsPeer) writeLoopCleanup(reason disconnectReason) {
...
+	filteredCount := 0
+	filtered := make([]bool, len(data), len(data))
+	for i := range data {
+		if wp.outgoingMsgFilter != nil && len(data[i]) > messageFilterSize && wp.outgoingMsgFilter.CheckDigest(digest[i], false, false) {
+			//wp.net.log.Debugf("msg drop as outbound dup %s(%d) %v", string(data[:2]), len(data)-2, digest)
+			// peer has notified us it doesn't need this message
+			outgoingNetworkMessageFilteredOutTotal.Inc(nil)
+			outgoingNetworkMessageFilteredOutBytesTotal.AddUint64(uint64(len(data)), nil)
+
+			filtered[i] = true
```
tsachiherman, line 680: we can do much better than this : instead of storing a boolean value, store the index of the entries that were not dropped. This would simplify the subsequent loop since you'll know exactly what to look for.

tsachiherman (changes requested):
few more requested changes.

network/wsPeer.go
```diff
@@ -316,12 +327,14 @@ func (wp *wsPeer) Respond(ctx context.Context, reqMsg IncomingMessage, responseT
 	return nil
 }
 
+const maxBroadcastArraySize = 30000
```
tsachiherman, line 330: this variable is not used.. can you remove it ?

tsachiherman (approved):

PR #1969, opened 2021-03-14
Check compile and dryrun programs against consensus limits

jannotti:
This adds extra checks on `goal clerk compile` and `goal clerk dryrun` to error out if your TEAL is too big or costly.


jannotti:
I have PR #1970 that addresses `compile` and `dryrun`. Not sure if you also want to do the various commands that also submit transactions.

tsachiherman (changes requested):
The change itself it good - two requests : 
1. Move error strings to `messages.go`. If we'll need at any point to translate the command line utility, it would make it much easier if all the messages would be in a single location.
2. Add goal expect test to generate this errors and verify the expected output is generated.

jasonpaulos:
I'm wondering if it makes sense to add a flag like `--force` or `--ignore-size` to `goal clerk compile` which would maintain the current behavior of always compiling.

jannotti:
There is a way to change the protocol level, -P, which I think is a global
option.  So if the use case is someone trying to compile something longer
for a new protocol version that would accept it, they could also use the -P
option to use the higher limits.

tsachiherman (dismissed):

cmd/goal/messages.go
```diff
@@ -156,6 +156,9 @@ const (
 	tealsignTooManyArg    = "--set-lsig-arg-idx too large, maximum of %d arguments"
 	tealsignInfoWroteSig  = "Wrote signature for %s to LSig.Args[%d]"
 
+	tealLogicSigSize = "%s: logicsig program size too large: %d > %d"
```
brianolson, line 159: I dislike the messages.go pattern; the strings are only used once and make more sense in context. The other two added reportErrorf() calls in clerk.go have inline strings.

jannotti: I somewhat agree, but you and @tsachiherman should arm wrestle and tell me who wins.  I don't feel too strongly about it.

tsachiherman: I know that it's a bit of a pain to maintain it *right now*, but the separation between code and text is meant to help us with few things ( even if we don't get the gains right now ) : 
- avoiding duplicate strings
- correcting strings ( typo, or just something more systematic such as the usage of the word "Algo" )
- Support for localization
- Support for non-ASCII output characters on different platforms.

brianolson: I used to believe the i18n benefit, and still kinda do, but also saw a system with effectively `func i18n(string) string` so that at a call site you could `fmt.Printf(i18n("Hello World, %s"), foo)` and it might substitute in "Hola Mundo, %s" based on a translation table lookup.
Anyway, I don't feel strong enough about it to block this PR. Might be a topic for `#dev` slack discussion to get a sense of people's feelings on style.

jannotti:
Yeah, something like that used to be the way in c/c++, except I think they
used _(s) instead of i18n(s). That was really nice because it lets you see
the string in context. You're much more likely to improve a message that
way.

tsachizehub (approved):

PR #1970, opened 2021-03-14
Add comments to disassembly of constant load lines to show constant

jannotti:
Now, a line like "int 76" which previously might have looked like
"intc_1", will look like "intc_1 // 76"

With bytes, the comment is an address or a quoted string if that seems
appropriate.


data/transactions/logic/assembler.go
```diff
@@ -1612,6 +1645,57 @@ func disBytecblock(dis *disassembleState, spec *OpSpec) {
...
+		suffix = "_2"
+		b = 2
+	case 0x2b:
+		suffix = "_3"
+		b = 3
+	case 0x27:
+		b = uint(dis.program[dis.pc+1])
+		suffix = fmt.Sprintf(" %d", b)
+	}
+	_, dis.err = fmt.Fprintf(dis.out, "bytec%s // %s\n", suffix, guessByteFormat(dis.byteConstants[b]))
```
brianolson, line 1696: I like this structure better, with one printf, can we do this for disIntc too?

brianolson (commented):
I like it. Cool feature. One idea for code beautification in disIntc()

data/transactions/logic/assembler.go
jannotti: Yup, done now, with a refactor to get better dryrun output too.

tsachiherman:
looks good to me, but would require a conflict resolving first.

brianolson (dismissed):

data/transactions/logic/assembler.go
```diff
@@ -1485,218 +1483,282 @@ func checkByteConstBlock(cx *evalContext) int {
...
-		_, dis.err = fmt.Fprintf(dis.out, " 0x%s", hex.EncodeToString(bv))
-		if dis.err != nil {
-			return
+		dis.bytec = append(dis.bytec, bv)
+		out += fmt.Sprintf(" 0x%s", hex.EncodeToString(bv))
+	}
+	return out, nil
+}
+
+func allPrintableASCII(bytes []byte) bool {
```
algorandskiy, line 1548: `unicode.IsPrint`?

data/transactions/logic/assembler.go
```diff
@@ -1485,218 +1483,282 @@ func checkByteConstBlock(cx *evalContext) int {
...
 		}
 	}
-	_, dis.err = dis.out.Write([]byte("\n"))
+	return true
 }
+func guessByteFormat(bytes []byte) string {
+	var short basics.Address
 
-func disPushInt(dis *disassembleState, spec *OpSpec) {
+	if len(bytes) == len(short) {
```
algorandskiy, line 1559: this is not 100% accurate, sometimes it can be a random hash that is also 32 bytes long. I had a similar issue in tealdbg with types guessing :(

data/transactions/logic/assembler.go
jannotti: Excellent, yes, I'll convert to that.

data/transactions/logic/assembler.go
jannotti: What did you do there?  I figured we want to be as helpful as possible, and there's no strictly correct thing to do, so I would try to present the most likely correct thing.  Nothing goes wrong if it's wrong, but it's nice to get it right.

One thing I realize is that I will certainly reverse the order, so that a 32 byte printable string is printed as a string, if it's entirely printable.  That's vanishingly unlikely for a key.

algorandskiy: This was for stack/scratch values: I check if printable, and either leave as a string or convert to 0xAABBCC... as assembler/disassembler. I removed address type guessing in a preference of correctness, and wanted to implement tooltip on hover with address string, but switched to something else.

Working solution for untyped addresses would be running the code and assigning `Address` type if a value compared/assigned/retrieved from known addresses (like `txn.Sender` field). But seems like it does not worth efforts :)

tsachizehub (approved):

PR #1972, opened 2021-03-14
Fix random e2e test failure in TestAccountInformationV2

tsachiherman:
## Summary

Fix random failure in `TestAccountInformationV2` e2e test.

## Test Plan

This is a test.

### DIFF
--- a/test/e2e-go/features/transactions/accountv2_test.go
+++ b/test/e2e-go/features/transactions/accountv2_test.go
@@ -162,9 +162,11 @@ int 1
 	a.NoError(err)
 	round, err = client.CurrentRound()
 	a.NoError(err)
-	_, err = client.BroadcastTransaction(signedTxn)
+	txid, err := client.BroadcastTransaction(signedTxn)
 	a.NoError(err)
-	client.WaitForRound(round + 2)
+	// ensure transaction is accepted into a block within 5 rounds.
+	confirmed := fixture.WaitForAllTxnsToConfirm(round+5, map[string]string{txid: signedTxn.Txn.Sender.String()})
+	a.True(confirmed)
 
 	// check creator's balance record for the app entry and the state changes
 	ad, err = client.AccountData(creator)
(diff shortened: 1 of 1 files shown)


test/e2e-go/features/transactions/accountv2_test.go
```diff
@@ -207,14 +209,17 @@ int 1
...
 	a.NoError(err)
-	_, err = client.BroadcastTransaction(signedTxn)
+	txid, err = client.BroadcastTransaction(signedTxn)
 	a.NoError(err)
 	_, err = client.WaitForRound(round + 3)
 	a.NoError(err)
 	// Ensure the txn committed
 	resp, err := client.GetPendingTransactions(2)
 	a.NoError(err)
 	a.Equal(uint64(0), resp.TotalTxns)
```
tsachiherman, line 219: Checking the pending transactions is good, but it's not sufficiently deterministic. If the node were to reset, the transaction pool would get clear out, and the transaction might or might not be there. Checking the client.TransactionInformation is much more robust.

test/e2e-go/features/transactions/accountv2_test.go
```diff
@@ -257,7 +262,7 @@ int 1
 	a.Equal(uint64(1), value.Uint)
 
 	// 2 global state update in total, 1 local state updates
-	checkEvalDelta(t, &client, round, round+5, 2, 1)
+	checkEvalDelta(t, &client, round+2, round+5, 2, 1)
```
tsachiherman, line 265: The above transaction was sent on round+3, so we won't find it before that point. On the flip side, we might find the previous transaction, and run into the wrong conclusions.

test/e2e-go/features/transactions/accountv2_test.go
algorandskiy: Just remove GetPendingTransactions then

algorandskiy (approved):

test/e2e-go/features/transactions/accountv2_test.go
tsachiherman: It wouldn't be adding it, but since it's already there, let's keep it for now so that this test would also provide a coverage for that method ?

shiqizng (approved):

PR #1974, opened 2021-03-15
Fixing bug: deleting old catchpoint directories when deleting files

algoidan:
## Summary
When removing catchpoint files, Algod now also removes the directory if it is empty.
When algod boots up it now cleans empty catchpoint directories if there are any.    

## Test Plan
Uint tests that where added:
-No empty directories are left after removing catchpoint files.
-algod cleans empty catchpoint directories.


tsachiherman:
Please sign with the CLA above; you'll need this in order to pass travis CLA check.

ledger/acctupdates.go
```diff
@@ -60,6 +62,8 @@ const (
 	// trieAccumulatedChangesFlush defines the number of pending changes that would be applied to the merkle trie before
 	// we attempt to commit them to disk while writing a batch of rounds balances to disk.
 	trieAccumulatedChangesFlush = 256
+	// directory name in which all the catchpoints files are stored
+	catchpointDirName = "catchpoints"
```
tsachiherman, line 66: It's great that you added this. I wasn't aware that we have multiple instances of that string in the codebase. However, please also make sure you're replacing the string with that constant uniformly - there are additional places in the _test files that use it. look in acctupdates_test.go and catchpointwriter_test.go

ledger/acctupdates.go
```diff
@@ -1473,6 +1483,72 @@ func (au *accountUpdates) upgradeDatabaseSchema3(ctx context.Context, tx *sql.Tx
...
+
+	// update version
+	_, err = db.SetUserVersion(ctx, tx, 5)
+	if err != nil {
+		return 0, fmt.Errorf("accountsInitialize unable to update database schema version from 4 to 5: %v", err)
+	}
+	return 5, nil
+}
+
+func  removeEmptyDirsOnSchemaUpgrade() (err error) {
```
tsachiherman, line 1504: The algorithm you've used here is non-recursive and would only handle the first-tier directories. We need a recursive solution here.

ledger/acctupdates.go
```diff
@@ -1473,6 +1483,72 @@ func (au *accountUpdates) upgradeDatabaseSchema3(ctx context.Context, tx *sql.Tx
...
+	f, err := os.Open(catchpointDirName)
```
tsachiherman, line 1505: The starting point isn't `catchpointDirName`, since that directory is relative to the CWD. instead, you should use the `au.dbDirectory` and use it in the same way it's being used above.

ledger/acctupdates.go
```diff
@@ -2328,6 +2399,37 @@ func catchpointRoundToPath(rnd basics.Round) string {
...
+		err = nil
+	} else {
+		// we can't delete the file, abort -
+		return fmt.Errorf("unable to delete old catchpoint file '%s' : %v", absCatchpointFileName, err)
+	}
+	catchpointRootDir := filepath.Join(au.dbDirectory, catchpointDirName)
+	currentCatchpointDir := filepath.Dir(absCatchpointFileName)
+
+	for currentCatchpointDir != catchpointRootDir {
+		filesInCatchupDir, err := os.ReadDir(currentCatchpointDir)
```
tsachiherman, line 2417: `ReadDir` is quite a slow method, since it reads all the entries. look https://stackoverflow.com/questions/30697324/how-to-check-if-directory-on-path-is-empty for alternative solution.

ledger/acctupdates.go
```diff
@@ -2328,6 +2399,37 @@ func catchpointRoundToPath(rnd basics.Round) string {
...
+		// it's ok if the file doesn't exist.
+		err = nil
+	} else {
+		// we can't delete the file, abort -
+		return fmt.Errorf("unable to delete old catchpoint file '%s' : %v", absCatchpointFileName, err)
+	}
+	catchpointRootDir := filepath.Join(au.dbDirectory, catchpointDirName)
+	currentCatchpointDir := filepath.Dir(absCatchpointFileName)
+
+	for currentCatchpointDir != catchpointRootDir {
```
tsachiherman, line 2416: I can't say that this iteration won't work.. but it's making me a bit uncomfortable to reconstruct the directory paths and then loop until we get to our root. I think that a more robust solution here would be to use `filepath.SplitList` and *construct* the known path from the file itself. That path would not include the root, so you don't need to worry about your stop condition.

ledger/acctupdates_test.go
```diff
@@ -881,30 +883,157 @@ func TestAcctUpdatesDeleteStoredCatchpoints(t *testing.T) {
...
 	require.NoError(t, err)
 
-	dummyCatchpointFilesToCreate := 42
+	const dummyCatchpointFilesToCreate = 43
+	const catchpointDir string = "./catchpoints"
 
+	dummyCatchpointFiles := make([]string, dummyCatchpointFilesToCreate)
 	for i := 0; i < dummyCatchpointFilesToCreate; i++ {
-		f, err := os.Create(fmt.Sprintf("./dummy_catchpoint_file-%d", i))
+		file := fmt.Sprintf("./%v/%v/%v/dummy_catchpoint_file-%d", catchpointDir, i/10, i/2, i)
```
tsachiherman, line 891: when formatting decimal numbers, use `%d` instead of `%v`.

ledger/acctupdates_test.go
```diff
@@ -881,30 +883,157 @@ func TestAcctUpdatesDeleteStoredCatchpoints(t *testing.T) {
...
+	defer ml.Close()
+
+	accts := []map[basics.Address]basics.AccountData{randomAccounts(20, true)}
+	au := &accountUpdates{}
+	conf := config.GetDefaultLocal()
+	conf.CatchpointInterval = 1
+	au.initialize(conf, ".", proto, accts[0])
+	defer au.close()
+
+	err = au.loadFromDisk(ml)
```
tsachiherman, line 963: There is no test here to ensure the said directories were truly removed.

ledger/acctupdates.go
```diff
@@ -1473,6 +1483,72 @@ func (au *accountUpdates) upgradeDatabaseSchema3(ctx context.Context, tx *sql.Tx
...
+		files, err := ioutil.ReadDir(path)
+		if err != nil {
+			return err
+		}
+		if len(files) == 0 {
+			emptyDir = append(emptyDir, path)
+		}
+		return nil
+	})
+	return emptyDir,err
```
tsachiherman, line 1549: add a space after the `,`

ledger/acctupdates.go
```diff
@@ -1473,6 +1483,72 @@ func (au *accountUpdates) upgradeDatabaseSchema3(ctx context.Context, tx *sql.Tx
...
+func GetEmptyDirs() (emptyDir []string, err error) {
+	emptyDir = make([]string, 0)
+	err = filepath.WalkDir(catchpointDirName, func(path string, d fs.DirEntry, errIn error) error {
+		if errIn != nil {
+			return errIn
+		}
+		if !d.IsDir() {
+			return nil
+		}
+		files, err := ioutil.ReadDir(path)
```
tsachiherman, line 1540: See my previous remarks regarding ReadDir

ledger/acctupdates.go
```diff
@@ -1473,6 +1483,72 @@ func (au *accountUpdates) upgradeDatabaseSchema3(ctx context.Context, tx *sql.Tx
...
+		}
+		for _, emptyDirPath := range emptyDir {
+			os.Remove(emptyDirPath)
+		}
+	}
+	return nil
+}
+
+func GetEmptyDirs() (emptyDir []string, err error) {
+	emptyDir = make([]string, 0)
```
tsachiherman, line 1532: it's ok to allocate zero length slices, but not required.

ledger/acctupdates.go
```diff
@@ -1473,6 +1483,72 @@ func (au *accountUpdates) upgradeDatabaseSchema3(ctx context.Context, tx *sql.Tx
...
+}
+
+func  removeEmptyDirsOnSchemaUpgrade() (err error) {
+	f, err := os.Open(catchpointDirName)
+	isNotExists := os.IsNotExist(err)
+	f.Close()
+	if isNotExists {
+		return nil
+	}
+	for {
```
tsachiherman, line 1511: Why do we need this for-loop ?

ledger/acctupdates_test.go
```diff
@@ -881,30 +883,157 @@ func TestAcctUpdatesDeleteStoredCatchpoints(t *testing.T) {
...
+	au.generateCatchpoint(basics.Round(3000010), "0#ABC2", crypto.Digest{}, time.Second)
+	au.generateCatchpoint(basics.Round(3000015), "0#ABC3", crypto.Digest{}, time.Second)
+	au.generateCatchpoint(basics.Round(3000020), "0#ABC4", crypto.Digest{}, time.Second)
+
+	const catchpointDir string = "./catchpoints"
+	defer func() {
+		os.RemoveAll(catchpointDir)
+	}()
+
+	numberOfCatchpointFiles, err := getNumberOfCatchpointFilesInDir(catchpointDir)
```
tsachiherman, line 1029: generated files in test should be created in the os's TEMP directory.

tsachiherman (changes requested):
overall looks very promising. Few small changes were requested.

ledger/acctupdates.go
algoidan: Done

ledger/acctupdates.go
algoidan: Done

ledger/acctupdates.go
algoidan: Done

ledger/acctupdates_test.go
algoidan: oops...
Done!

ledger/acctupdates.go
algoidan: Done

ledger/acctupdates.go
algoidan: Done

ledger/acctupdates.go
algoidan: It seems like we don't want to load the full list of catchpoint files into memory, So we execute the loop in chunks.

tsachiherman:
The package "io/fs" is a new package added in 1.16. I don't know when we're going to switch to 1.16.
I would suggest using `filepath.Walk` method from the `path/filepath` package.

ledger/acctupdates.go
```diff
@@ -2328,6 +2412,51 @@ func catchpointRoundToPath(rnd basics.Round) string {
...
+			continue
+		}
+
+		isEmpty, err:= IsDirectoryEmpty(absSubdir)
+		if err != nil {
+			return fmt.Errorf("unable to read old catchpoint directory '%s' : %v", subDirectoriesToScan[i], err)
+		}
+		if isEmpty{
+			err = os.Remove(absSubdir)
+			if err != nil {
```
tsachiherman, line 2450: In this case, we can ignore `os.IsNotExist(err)`.

ledger/acctupdates.go
```diff
@@ -1473,6 +1482,86 @@ func (au *accountUpdates) upgradeDatabaseSchema3(ctx context.Context, tx *sql.Tx
...
+	var emptyDir []string
+	err := filepath.Walk(PathToScan,func(path string, f os.FileInfo, errIn error) error {
+		if errIn != nil  {
+			return errIn
+		}
+		if !f.IsDir() {
+			return nil
+		}
+		isEmpty, err := IsDirectoryEmpty(path)
+		if err != nil {
```
tsachiherman, line 1554: in case you get a file-doesn't-exist here, you can just ignore it and keep going. ( it just mean that it was deleted in the background by some other process. )

tsachiherman:
@algorandskiy - please be aware of these changes. I'm planning to merge these along with your asset changes so that we can make a single database upgrade.

algorandskiy:
yeah, `feature/1000-assets` branch a bit behind, so the merge will be not an easiest one.

tsachiherman:
@algoidan , could you please update the PR description ( Summary & Test Plan ) to reflect your work here ?
( it will be needed before the PR can be merged in )

gmalouf:
@tsachiherman / @algoidan - is this going in as part of the ledger refactor or can it be independent?

tsachiherman:
> @tsachiherman / @algoidan - is this going in as part of the ledger refactor or can it be independent?

It would require a pseudo schema upgrade so that we can do the "fix" only once. It's a pretty simple change, but was blocked for a long time.

tsachiherman:
I'm closing this one since it was already merged into the feature/unlimited-assets branch.

PR #1976, opened 2021-03-16
TEAL 3 changes for tealdbg

jasonpaulos:
## Summary

Most of the new TEAL 3 features already work with the debugger, but a few changes were needed to make the new array fields display properly on the Chrome frontend.

Also, I noticed the disassembly for `pushbytes` was missing a trailing newline, which made the disassembled program for debugging look strange.

## Test Plan

Added a unit test for `pushbytes` and `pushint` disassembly.


algorandskiy (approved):

jannotti (approved):
This looks right to me.

Issue #1809, opened 2021-01-02
[Investigation] Allow tealdbg to listen to any address


fabrice102:
**Is your feature request related to a problem? Please describe.**

When using `tealdbg` from inside sandbox (https://github.com/algorand/sandbox), `tealdbg` needs to allow listening from any address and not just localhost. (This is true on macOS. It is possible that Docker on Ubuntu does not require that.)

Currently, the only solution I know how to use `tealdbg` inside sandbox is the following:
1. Add `- 9392:9392` inside the ports section of `docker-compose.yml`: https://github.com/algorand/sandbox/blob/0c15c8f5e0256770ef434333fee4ca55bb794ce0/docker-compose.yml#L21
2. Enter sandbox: `./sandbox enter algod`
3. Install `socat`: `apt install socat`
4. Create TEAL script, e.g., `echo "int 1" > a.teal`
5. Run `tealdbg` in the background with new port (e.g., 1234): `tealdbg debug --remote-debugging-port 1234 a.teal &`
6. Redirect port 1234 to 9392 because 1234 is only bound locally (see netstat below): `socat tcp-listen:9392,reuseaddr,fork tcp:localhost:1234`
7. Open Chrome Developer tools and visit the address displayed by `tealdbg` where 1234 is replaced by 9392.

For information: netstat
```
root@b4a8696ccc6d:~/testnetwork/Node# netstat  -ltnp
Active Internet connections (only servers)
Proto Recv-Q Send-Q Local Address           Foreign Address         State       PID/Program name
tcp        0      0 0.0.0.0:9392            0.0.0.0:*               LISTEN      2794/socat
tcp        0      0 127.0.0.1:8080          0.0.0.0:*               LISTEN      14/algod
tcp        0      0 127.0.0.1:1234          0.0.0.0:*               LISTEN      2739/tealdbg
tcp        0      0 127.0.0.1:44739         0.0.0.0:*               LISTEN      14/algod
tcp        0      0 127.0.0.11:42925        0.0.0.0:*               LISTEN      -
tcp6       0      0 :::4001                 :::*                    LISTEN      23/algod
tcp6       0      0 :::4002                 :::*                    LISTEN      32/kmd
```

**Describe the solution you'd like**

Add a flag to `tealdbg` to allow to listen to any address rather than just `127.0.0.1`.
[...]

PR #1977, opened 2021-03-16
tealdbg: listen on specified address

algorandskiy:
## Summary

Fix for #1809

## Test Plan

Updated tealdbg e2e test

### DIFF
--- a/cmd/tealdbg/main.go
+++ b/cmd/tealdbg/main.go
@@ -135,6 +135,7 @@ var roundNumber uint64
 var timestamp int64
 var runMode runModeValue = runModeValue{makeCobraStringValue("auto", []string{"signature", "application"})}
 var port int
+var iface string
 var noFirstRun bool
 var noBrowserCheck bool
 var noSourceMap bool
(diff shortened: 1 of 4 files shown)


tsachiherman (commented):
How's about replacing it with an endpoint, so it would be more consistent with the rest of the platform ? I know that it will break backward compatibility though..

algorandskiy:
`--remote-debugging-port` is here to have chrome-like option so that one can use VSCode's CDT plugin to drive debugging and have VSCode as UI.

tsachiherman (approved):

fabrice102:
I have checked on macOS. This worked for me by doing the following:
1. open the port 9392 in sandbox in `Dockerfile.yml`
2. inside sandbox (`./sandbox enter algod`), run:
    ```bash
    tealdbg debug --remote-debugging-port 9392 --listen "" a.teal
    ```
3. visit `devtools://devtools/bundled/js_app.html?experiments=true&v8only=false&ws=127.0.0.1:9392/7d35cd8846288e164178efed7123a00efd0af5d906fed89b1c42fcc7593e3a7d` instead of the URL displayed by `tealdbg` (i.e., adding 127.0.0.1 after `ws=`)

Note: I've tried to use the IP address 172.18.0.3 of the container instead of `""` for listen (IP got by `docker inspect -f '{{.Name}} - {{range .NetworkSettings.Networks}}{{.IPAddress}}{{end}}' algorand-sandbox-algod`, that matches `ifconfig` inside the sandbox docker container). This works when visiting `devtools://devtools/bundled/js_app.html?experiments=true&v8only=false&ws=127.0.0.1:9392/7d35cd8846288e164178efed7123a00efd0af5d906fed89b1c42fcc7593e3a7d` but not when visiting `devtools://devtools/bundled/js_app.html?experiments=true&v8only=false&ws=172.18.0.3:9392/7d35cd8846288e164178efed7123a00efd0af5d906fed89b1c42fcc7593e3a7d`, things seem to work but the source code does not appear.
So I don't know.

TLDR: Improve significantly the current state of the art. People need to manually fix the URL displayed by `tealdbg`, but it's quite painless. And fixing that in the code seems complex and not necessary.

PR #1979, opened 2021-03-17
skip printing the logs when program is in terminated state.

shiqizng:
script prints logs for all test cases that has non-zero return code. When the process got killed it  has a return code 15, so this caused verbose logging.

### DIFF
--- a/test/scripts/e2e_client_runner.py
+++ b/test/scripts/e2e_client_runner.py
@@ -113,16 +113,12 @@ def _script_thread_inner(runset, scriptname):
     runset.running(scriptname, p)
     timeout = read_script_for_timeout(scriptname)
     try:
-        # Popen return code: https://docs.python.org/3/library/subprocess.html#subprocess.Popen.returncode
-        # 0 means success. A negative value -N indicates that the child was terminated by signal N (POSIX only).
         retcode = p.wait(timeout)
     except subprocess.TimeoutExpired as te:
         sys.stderr.write('{}\n'.format(te))
         retcode = -1
     dt = time.time() - start
 
-    if retcode != 0:
-        logger.debug("script file: ", scriptname, "retcode: ", retcode)
 
     if runset.terminated:
         logger.info('Program terminated before %s finishes.', scriptname)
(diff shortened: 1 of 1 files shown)


test/scripts/e2e_client_runner.py
```diff
@@ -113,11 +113,22 @@ def _script_thread_inner(runset, scriptname):
     runset.running(scriptname, p)
     timeout = read_script_for_timeout(scriptname)
     try:
+        # Popen return code: https://docs.python.org/3/library/subprocess.html#subprocess.Popen.returncode
```
brianolson, line 116: excessive comment; retcode=0 meaning success is standard unix for decades. We can drop these two new lines of comment.

test/scripts/e2e_client_runner.py
```diff
@@ -113,11 +113,22 @@ def _script_thread_inner(runset, scriptname):
...
+        # 0 means success. A negative value -N indicates that the child was terminated by signal N (POSIX only).
         retcode = p.wait(timeout)
     except subprocess.TimeoutExpired as te:
         sys.stderr.write('{}\n'.format(te))
         retcode = -1
     dt = time.time() - start
+
+    if retcode != 0:
+        logger.debug("script file: ", scriptname, "retcode: ", retcode)
+
```
brianolson, line 126: don't need this, there's retcode != 0 logging just below

brianolson (changes requested):

test/scripts/e2e_client_runner.py
shiqizng: code updated.

test/scripts/e2e_client_runner.py
shiqizng: removed these lines.

brianolson (approved):
Thanks. I'm glad this turned out to be simple with most of the parts built.

algorandskiy (approved):
@shiqizng for the next PRs, please fork `algorand/go-algorand` repo and submit PRs from the fork.

PR #1980, opened 2021-03-17
Fixes for pending transaction endpoint

jasonpaulos:
## Summary

I discovered that #1004 accidentally changed the `total-transactions` field in the `getPendingTransactions` handler function. That value is supposed to be the total number of pending transactions, but currently it is only the number of transactions returned by the current API call. I've fixed this and improved the names of local variables to make this easier to see.

While testing this, I also discovered a bug with the `max` parameter. When set to 0, `getPendingTransactions` should return all pending transactions. The current behavior is correct when there is no `max` query param in the API request, but when explicitly setting `max` to 0 (i.e.`/v2/transactions/pending?max=0`) the endpoint always returns 0 transactions.

Since the documentation states `If max=0, returns all pending txns`, I believe this is undesirable so I've also fixed this.

## Test Plan

Unit test for `getPendingTransactions`  have been improved.

### DIFF
--- a/daemon/algod/api/server/v2/handlers.go
+++ b/daemon/algod/api/server/v2/handlers.go
@@ -22,6 +22,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"math"
 	"net/http"
 	"time"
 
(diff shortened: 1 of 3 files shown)


daemon/algod/api/server/v2/handlers.go
```diff
@@ -584,10 +584,10 @@ func (v2 *Handlers) getPendingTransactions(ctx echo.Context, max *uint64, format
 	}
 
 	// Convert transactions to msgp / json strings
-	txnArray := make([]transactions.SignedTxn, 0)
-	for _, txn := range txns {
+	topTxns := make([]transactions.SignedTxn, 0)
+	for _, txn := range txnPool {
 		// break out if we've reached the max number of transactions
-		if max != nil && uint64(len(txnArray)) >= *max {
+		if max != nil && *max != 0 && uint64(len(topTxns)) >= *max {
```
tsachiherman, line 590: Nit : evaluate the max variable outside the loop, and initialize another variable to the desired limit.
If no limit, use maxint.

This would simplify the in-loop predicate.

tsachiherman (commented):
Looks good, one nit

winder (approved):
LGTM

PR #1982, opened 2021-03-17
Report algod panic in algoh

nicholasguoalgorand:
## Summary

When algoh spawns an algod instance and the algod panics, algoh simply exits without logging the panic. This change logs the algod panic before the algoh exits.

## Test Plan

Started an algoh with algod, hardcoded a panic when it hits round 10, sees that `algod-err.log` contains the panic message

### DIFF
--- a/cmd/algoh/main.go
+++ b/cmd/algoh/main.go
@@ -145,7 +145,6 @@ func main() {
 		}
 		err = cmd.Wait()
 		if err != nil {
-			captureErrorLogs(algohConfig, errorOutput, output, absolutePath, true)
 			reportErrorf("error waiting for algod: %v", err)
 		}
 		close(done)
@@ -363,7 +362,7 @@ func captureErrorLogs(algohConfig algoh.HostConfig, errorOutput stdCollector, ou
 
 func reportErrorf(format string, args ...interface{}) {
 	fmt.Fprintf(os.Stderr, format, args...)
-	logging.Base().Fatalf(format, args...)
+	logging.Base().Warnf(format, args...)
 }
 
 func sendLogs() {
--- a/cmd/algoh/main.go
+++ b/cmd/algoh/main.go
@@ -145,6 +145,7 @@ func main() {
 		}
 		err = cmd.Wait()
 		if err != nil {
+			captureErrorLogs(algohConfig, errorOutput, output, absolutePath, true)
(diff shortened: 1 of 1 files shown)


cmd/algoh/main.go
```diff
@@ -145,6 +145,7 @@ func main() {
 		}
 		err = cmd.Wait()
 		if err != nil {
+			captureErrorLogs(algohConfig, errorOutput, output, absolutePath, true)
```
tsachiherman, line 148: we want to capture the logs in all the exit cases, so let's do it correctly:
after the `err = cmd.Wait()`, add the following
```
defer func() {
                // capture logs if algod terminated prior to blockWatcher starting
		if !blockWatcherInitialized {
			captureErrorLogs(algohConfig, errorOutput, output, absolutePath, true)
		}
}
```
and remove the same section from below.
Also, in `reportErrorf`, make sure to call `Warnf` instead of `Fatalf`. We don't want to exit the process right away, ( the closing of the `done` channel would do that anyway, and we don't need the `algoh` stack trace.
There is a race condition around the `blockWatcherInitialized`, but I'm going to leave it for now.

cmd/algoh/main.go
```diff
@@ -362,7 +362,7 @@ func captureErrorLogs(algohConfig algoh.HostConfig, errorOutput stdCollector, ou
 
 func reportErrorf(format string, args ...interface{}) {
 	fmt.Fprintf(os.Stderr, format, args...)
-	logging.Base().Fatalf(format, args...)
+	logging.Base().Warnf(format, args...)
```
egieseke, line 365: Would Errorf be better than Warnf?

egieseke (approved):

cmd/algoh/main.go
tsachiherman: hmm. The thing is this - 
Errorf ends up writing a stack trace dump into the log file. This stack trace is completely useless, since it's the `algoh` stack trace. It also doesn't indicate an error in the `algoh`, since the `algoh` was working flawlessly.

nicholasguoalgorand: also the algoh hasn't errored, the algod has right?

egieseke:
As soon as this is merged, I will run another scenario2 test.

PR #1990, opened 2021-03-19
copy GetPeerData()/SetPeerData() from txnsync branch

brianolson:
## Summary

Add GetPeerData()/SetPeerData() to network interface for data to be associated with a peer by other parts of the system so that data gets garbage collected when peer goes away.

## Test Plan

Trivial feature. No testing yet.

### DIFF
--- a/network/wsNetwork.go
+++ b/network/wsNetwork.go
@@ -201,6 +201,10 @@ type GossipNode interface {
 
 	// SubstituteGenesisID substitutes the "{genesisID}" with their network-specific genesisID.
 	SubstituteGenesisID(rawURL string) string
+
+	GetPeerData(peer Peer, key string) interface{}
+
+	SetPeerData(peer Peer, key string, value interface{})
 }
 
 // IncomingMessage represents a message arriving from some peer in our p2p network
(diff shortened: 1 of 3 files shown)


network/wsNetwork.go
```diff
@@ -201,6 +201,10 @@ type GossipNode interface {
 
 	// SubstituteGenesisID substitutes the "{genesisID}" with their network-specific genesisID.
 	SubstituteGenesisID(rawURL string) string
+
+	GetPeerData(peer Peer, key string) interface{}
```
tsachiherman, line 205: Please add comments on the public interface.

network/wsPeer.go
```diff
@@ -209,6 +209,11 @@ type wsPeer struct {
 	// throttledOutgoingConnection determines if this outgoing connection will be throttled bassed on it's
 	// performance or not. Throttled connections are more likely to be short-lived connections.
 	throttledOutgoingConnection bool
+
+	// clientDataStore is a generic key/value store used to store client-side data entries associated with a particular peer.
+	clientDataStore map[string]interface{}
+
+	clientDataStoreMu deadlock.Mutex
```
tsachiherman, line 216: please add the trivial comment `clientDataStoreMu` synchronizes access to the `clientDataStore` map

tsachiherman (changes requested):
please add a trivial unit test. The goal of the unit test is to verify "normal" operation as well as to ensure we have full coverage. ( i.e. pass nil as the peer object to confirm that the method doesn't crash ).
when this PR approved and merge, create a follow-up to merge the changes from here onto the feature/txnsync branch.

tsachiherman (approved):
looks great!

PR #1991, opened 2021-03-19
Avoid compact cert validation on non-validate/non-generate path

tsachiherman:
## Summary

This PR addresses two issues with the compact cert implementation:
1. During evaluation, the compact certificate transaction were evaluated regardless if the evaluator was used with validate, generate nor neither specified. From performance perspective, it's inefficient. From practical perspective, it currently trigger a separate issue since it fails fetching the block 0 header during the accounts update caches initializations. With this change, the accounts update caches initialization could remain as is.
2. In case the compact certificate is not the last transaction in a block, no compact certificate would be generated. This is due to a bug in the initialization of the cow object.

## Test Plan

Modify existing unit tests and e2e tests to reproduce and confirm fix.

### DIFF
--- a/ledger/eval.go
+++ b/ledger/eval.go
@@ -804,8 +804,11 @@ func (eval *BlockEvaluator) transaction(txn transactions.SignedTxn, evalParams *
 		}
 	}
 
-	// in case of a CompactCertTx transaction, we want to validate it only in validate or generate mode. This will deviate the cow's CompactCertNext depending of
+	// in case of a CompactCertTx transaction, we want to "apply" it only in validate or generate mode. This will deviate the cow's CompactCertNext depending of
 	// whether we're in validate/generate mode or not, however - given that this variable in only being used in these modes, it would be safe.
+	// The reason for making this into an exception is that during initialization time, the accounts update is "converting" the recent 320 blocks into deltas to
+	// be stored in memory. These deltas don't care about the compact certificate, and so we can improve the node load time. Additionally, it save us from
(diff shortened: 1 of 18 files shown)


ledger/eval.go
```diff
@@ -801,6 +804,17 @@ func (eval *BlockEvaluator) transaction(txn transactions.SignedTxn, evalParams *
 		}
 	}
 
+	// in case of a CompactCertTx transaction, we want to "apply" it only in validate or generate mode. This will deviate the cow's CompactCertNext depending of
+	// whether we're in validate/generate mode or not, however - given that this variable in only being used in these modes, it would be safe.
+	// The reason for making this into an exception is that during initialization time, the accounts update is "converting" the recent 320 blocks into deltas to
+	// be stored in memory. These deltas don't care about the compact certificate, and so we can improve the node load time. Additionally, it save us from
+	// performing the validation during catchup, which is another performance boost.
+	if (eval.validate || eval.generate) && txn.Txn.Type == protocol.CompactCertTx {
```
zeldovich, line 812: It seems like a natural place for this code would be in the `case protocol.CompactCertTx` branch inside of `applyTransaction()`; why not put it there?  Is it because `applyTransaction()` is not a method on `BlockEvaluator`?  It doesn't seem to be used by anything except `BlockEvaluator.transaction()`, so should be reasonable to make it a method so that we can access `eval.validate` and `eval.generate` there..

tsachiherman: that's a good point. I'll make that change.

zeldovich (approved):

PR #1995, opened 2021-03-21
make deadlockLogger thread-safe

tsachiherman:
## Summary

The deadlock logger we've used was not thread-safe. That was fine until the point where a deadlock would be detected, at which there was a chance of creating a subsequent failure while reporting the first one.

This PR re-implement the logger in a safe-way : it inherently prevent concurrent calls to `onPotentialDeadlock` as well as ensures that all the writes using the `Writer` interface are synchronized.

## Test Plan

Existing unit test was updated. New unit test was added.


algonautshant (dismissed):
Looks good.

algonautshant (approved):

PR #1997, opened 2021-03-22
Improve txsync and txpool data exchange

tsachiherman:
## Summary

This PR improves the transaction pool and transaction sync data exchange, preparing it for higher throughput.
In essence, it's making the following modifications:
1. Add a processing go-routine between the transaction sync and the transaction pool, which takes the "load" of the transaction verification times as well as the load from the transaction pool locking time.
2. Add pendingLastestLocal to the transaction pool, and take advantage of that in the transaction sync.
3. In the transaction handler, add support for processing lists of transaction groups in chunks - allowing us to take advantage of less locking on the transaction pool as well as using the batch transaction verification ( when that would become available ).

## Test Plan

Use the performance testing to ensure no regression occur.
Use the emulator to ensure no functional changes.


data/txHandler.go
```diff
@@ -263,12 +263,6 @@ func (handler *TxHandler) processIncomingTxn(rawmsg network.IncomingMessage) net
 // Note that this also checks the consistency of the transaction's group hash,
 // which is required for safe transaction signature caching behavior.
 func (handler *TxHandler) checkAlreadyCommitted(tx *txBacklogMsg) (processingDone bool) {
-	txids := make([]transactions.Txid, len(tx.unverifiedTxGroup))
-	for i := range tx.unverifiedTxGroup {
-		txids[i] = tx.unverifiedTxGroup[i].ID()
-	}
-	logging.Base().Debugf("got a tx group with IDs %v", txids)
-
```
tsachiherman, line 271: this code section was just printing transaction ids.. since it's not needed, I removed it.

txnsync/emulatorTimer_test.go
```diff
@@ -112,6 +112,7 @@ func (g *guidedClock) Advance(adv time.Duration) {
 	// fire expired clocks
 	for _, entry := range expiredClocks {
 		entry.ch <- g.zero.Add(g.adv)
+		close(entry.ch)
```
tsachiherman, line 115: closing the channel is important, since it's handling the case where we have two "users" for the same timeout value -> in which case we need the channel to be closed.

txnsync/emulator_test.go
```diff
@@ -541,7 +541,7 @@ func TestEmulatedTwoNodesFourRelays(t *testing.T) {
 						{
 							uploadSpeed:   1000000,
 							downloadSpeed: 1000000,
-							target:        1,
+							target:        2,
```
tsachiherman, line 544: this was a bug in the test case.

data/txHandler.go
```diff
@@ -319,7 +313,75 @@ func (handler *TxHandler) processDecoded(unverifiedTxGroup []transactions.Signed
 		logging.Base().Warnf("unable to pin transaction: %v", err)
 	}
 
-	return network.OutgoingMessage{}, false
+	return false
+}
+
+func (handler *TxHandler) filterAlreadyCommitted(unverifiedTxGroups []transactions.SignedTxGroup) (filteredGroups []transactions.SignedTxGroup) {
```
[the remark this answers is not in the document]
tsachiherman: done.

brianolson (commented):
I think I found a race due to introduced asynchrony, otherwise looks fine

node/txnSyncConn.go
```diff
@@ -199,19 +201,15 @@ func (tsnc *transcationSyncNodeConnector) Handle(raw network.IncomingMessage) ne
...
 	for _, txGroup := range txGroups {
-		if err := tsnc.txHandler.Handle(txGroup.Transactions); err != nil {
-			// we had some failuire, disconnect from peer.
-			tsnc.node.net.Disconnect(networkPeer)
-			break
-		}
+		txCount += len(txGroup.Transactions)
 	}
-	return
+	tsnc.txHandler.HandleTransactionGroups(networkPeer, txGroups)
```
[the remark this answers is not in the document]
tsachiherman: I think that you've got the gist here. Because we want to make it a background operation, we can no longer know what the size of the transaction pool exactly. So, we make a "best efforts attempt".
The returned number here doesn't have to be precise. It only helps us to determine if we need to adjust the beta, which get modified only on really large "chunks", so the txCount might not be as critical anyway.

brianolson (approved):

PR #1998, opened 2021-03-22
Fix/improve kmd speed in test

algoidan:
## Summary
Speed up tests on ARM64 machines by configuring kmd with allow_unsafe_scrypt = true.


test/scripts/e2e.sh
```diff
@@ -45,6 +45,19 @@ TEST_RUN_ID=$(${SCRIPT_PATH}/testrunid.py)
 export TEMPDIR=${SRCROOT}/tmp/out/e2e/${TEST_RUN_ID}
 echo "Test output can be found in ${TEMPDIR}"
 
+
+# some ARM64 testing machines have memory issues which cause some tests to fail .
```
brianolson, line 49:
```suggestion
# ARM64 has an unoptimized scrypt() which can cause tests to timeout.
# Run kmd with scrypt() configured to run less secure and fast to go through the motions for test.
```

brianolson (approved):
I think the comment could be tweaked based on my understanding of the underlying issue, but otherwise looks good.

brianolson (approved):

PR #2001, opened 2021-03-22
Xor filter bloom alternative

brianolson:
## Summary

xor filter is more efficient than traditional bloom filter (better false positive rate per serialized bit), and faster (see benchmarks)

## Test Plan

Many new tests and benchmarks were written to show that new algorithm performs better for our kind of data.


brianolson:
bloom filter target 0.004 false positive rate
xor produces smaller serialized filters

```
 filter for 1000 * [32]byte, bloom 1446 bytes, xor8 1274 bytes
 false positives bloom 386/100000, xor 401/100000
 filter for 5000 * [32]byte, bloom 7192 bytes, xor8 6194 bytes
 false positives bloom 373/100000, xor 399/100000
 filter for 10000 * [32]byte, bloom 14375 bytes, xor8 12344 bytes
 false positives bloom 397/100000, xor 360/100000
 filter for 50000 * [32]byte, bloom 71836 bytes, xor8 61546 bytes
 false positives bloom 379/100000, xor 363/100000
 filter for 100000 * [32]byte, bloom 143662 bytes, xor8 123046 bytes
 false positives bloom 392/100000, xor 438/100000
```

bloom filter target 0.01 false positive rate
bloom filter produces smaller serialized filters; xor still gets 0.004 fp rate

```
 filter for 1000 * [32]byte, bloom 1207 bytes, xor8 1274 bytes
 false positives bloom 1071/100000, xor 401/100000
 filter for 5000 * [32]byte, bloom 6000 bytes, xor8 6194 bytes
 false positives bloom 1092/100000, xor 399/100000
 filter for 10000 * [32]byte, bloom 11991 bytes, xor8 12344 bytes
 false positives bloom 957/100000, xor 360/100000
 filter for 50000 * [32]byte, bloom 59916 bytes, xor8 61546 bytes
 false positives bloom 980/100000, xor 363/100000
 filter for 100000 * [32]byte, bloom 119822 bytes, xor8 123046 bytes
 false positives bloom 999/100000, xor 438/100000
```

brianolson:
xor filter uses a lot more scratch space while building the filter
```
=== RUN   TestMemXor
 build mem[5000]: 215 mallocs, 0 frees, 596856 bytes allocated
 load  mem[5000]: 3 mallocs, 0 frees, 6592 bytes allocated
--- PASS: TestMemXor (0.00s)
=== RUN   TestMemBloom
 build mem[5000]: 7 mallocs, 0 frees, 16704 bytes allocated
 load  mem[5000]: 6 mallocs, 0 frees, 8480 bytes allocated
--- PASS: TestMemBloom (0.00s)
```

brianolson:
Creating a persistent Builder object I was able to get memory allocations down to about half. You can see the savings in the second half of this test:

```
=== RUN   TestMemXor
    xor_test.go:195: build mem[5000]: 194 mallocs, 0 frees, 604112 bytes allocated
    xor_test.go:196: load  mem[5000]: 3 mallocs, 0 frees, 6600 bytes allocated
    xor_test.go:195: build mem[5000]: 195 mallocs, 0 frees, 268976 bytes allocated
    xor_test.go:196: load  mem[5000]: 3 mallocs, 0 frees, 6600 bytes allocated
--- PASS: TestMemXor (0.01s)
```

The xorfilter benchmarks well for speed:

```
BenchmarkCreateLargeBloomFilter-16    	      20	  54640981 ns/op
BenchmarkCreateLargeXorFilter-16      	      36	  28692618 ns/op

BenchmarkBloomFilterTest-16           	11311885	       106.0 ns/op
BenchmarkXorFilterTest-16             	95593315	        13.39 ns/op
```

brianolson:
map with hinted capacity for many fewer malloc()
```
=== RUN   TestMemXor
    xor_test.go:196: build mem[5000]: 26 mallocs, 0 frees, 490784 bytes allocated
    xor_test.go:197: load  mem[5000]: 6 mallocs, 0 frees, 104976 bytes allocated
    xor_test.go:196: build mem[5000]: 16 mallocs, 0 frees, 153504 bytes allocated
    xor_test.go:197: load  mem[5000]: 6 mallocs, 0 frees, 104976 bytes allocated
--- PASS: TestMemXor (0.00s)
```
xor build with hinted map also faster
```
BenchmarkCreateLargeBloomFilter-16    	      21	  53671552 ns/op
BenchmarkCreateLargeXorFilter-16      	      62	  18011036 ns/op
```

util/bloom/bloom_test.go
```diff
@@ -169,18 +170,21 @@ func TestMarshalJSON(t *testing.T) {
 	}
 }
 
-func BenchmarkCreateLargeFilter(b *testing.B) {
+const largeFilterElements = 150000
+
+// BenchmarkCreateLargeBloomFilter should have the same structure as xor_test.go BenchmarkCreateLargeXorFilter
+func BenchmarkCreateLargeBloomFilter(b *testing.B) {
```
tsachiherman, line 176: you "broke" the barrier here between the bloom filter, and the xor filter by including the Marshaling as part of the "PutXX" benchmarks. I think that for your new testing it is correct, however, why don't you retain the old one as a separate benchmark ?

util/bloom/xor.go
```diff
@@ -0,0 +1,155 @@
...
+)
+
+// XorBuilder is a local alias for xorfilter.Builder
+type XorBuilder = xorfilter.Builder
+
+// XorFilter is a faster more efficient alternative to a Bloom filter
+// An XorFilter object can be used as is or with optional adittional setup.
+type XorFilter struct {
+	xor     *xorfilter.Xor8
+	holding map[uint64]bool
```
tsachiherman, line 34: having this map isn't the optimal data structure here. I would think that an array of arrays would be a better choice. 
But don't feel obligated to change this unless you profile it and see that it becomes an issue - in particular, you want to make sure that the number of memory allocations isn't linear with the number of elements.

tsachiherman (commented):
looks like a very promising PR. Try to make final cleanups, comments, etc.

util/bloom/xor.go
brianolson: It changes the test but maintains comparability between xor and bloom. The xor filter does the majority of its work in MarshalBinary() so I had to put that in its benchmark, but then it seemed only fair to compare to to the bloom filter going through the same steps. Do we keep archived history of performance on benchmarks? In this case the benchmark isn't being used to see if changing code makes it more optimized, but to compare two different implementations.

util/bloom/bloom_test.go
brianolson: It changes the test but maintains comparability between xor and bloom. The xor filter does the majority of its work in MarshalBinary() so I had to put that in its benchmark, but then it seemed only fair to compare to to the bloom filter going through the same steps. Do we keep archived history of performance on benchmarks? In this case the benchmark isn't being used to see if changing code makes it more optimized, but to compare two different implementations.

tsachiherman: exactly. I agree that your changes are good and needed in order to compare "apples" to "apples".
I don't have any particular sentiment for the existing benchmark, just figured there is no good reason to "drop it" unless it's in our way. Just call the old one "BenchmarkSetOnLargeBloomFilter".

brianolson: I tried adding back an unaltered benchmark but it seems out of place. It doesn't add anything other than being a copy of the new Benchmark but without MarshalBinary()

brianolson: Since we don't care about the historic performance of the benchmark from before this PR to after it I think it's okay to change the benchmark. I don't think we gain anything by having a duplicate of the benchmark with one line different.

util/bloom/xor.go
brianolson: The xor-filter build step wants a de-duplicated set of uint64. There might be a more efficient `set<uint64>` type, but this was easy and still benchmarks faster to build the filter than the old bloom filter.

brianolson:
This is failing one integration test, but it passes locally. Probably safe to merge to feature branch?

util/bloom/bloom.go
```diff
@@ -118,22 +118,46 @@ func BinaryMarshalLength(numElements int, falsePositiveRate float64) int64 {
...
+		if f.numHashes > maxHashes {
+			return nil, errors.New("too many hashes")
+		}
+		copy(f.prefix[:], data[4:8])
+		f.data = data[8:]
+		f.preimageStagingBuffer = make([]byte, len(f.prefix), len(f.prefix)+32)
+		f.hashStagingBuffer = make([]uint32, f.numHashes+3)
+		copy(f.preimageStagingBuffer, f.prefix[:])
+		return f, nil
+	*/
```
tsachiherman, line 160: remove commented out lines.

txnsync/bloomFilter.go
```diff
@@ -45,29 +46,40 @@ type transactionsRange struct {
...
-	outFilter.filter, err = bloom.UnmarshalBinary(enc.BloomFilter)
 	if err != nil {
 		return bloomFilter{}, err
 	}
 	return outFilter, nil
 }
 
 func (bf *bloomFilter) encode() (out encodedBloomFilter) {
-	out.BloomFilterType = byte(multiHashBloomFilter)
+	out.BloomFilterType = byte(invalidBloomFilter)
```
tsachiherman, line 72: I think that you meant `xorBloomFilter`, right ?

txnsync/bloomFilter.go
```diff
@@ -137,8 +149,10 @@ func makeBloomFilter(encodingParams requestParams, txnGroups []transactions.Sign
 			}
 		}
 
-		sizeBits, numHashes := bloom.Optimal(len(filtedTransactionsIDs), bloomFilterFalsePositiveRate)
-		result.filter = bloom.New(sizeBits, numHashes, shuffler)
+		//sizeBits, numHashes := bloom.Optimal(len(filtedTransactionsIDs), bloomFilterFalsePositiveRate)
+		//result.filter = bloom.New(sizeBits, numHashes, shuffler)
```
tsachiherman, line 153: remove commented out lines.

txnsync/bloomFilter.go
```diff
@@ -137,8 +149,10 @@ func makeBloomFilter(encodingParams requestParams, txnGroups []transactions.Sign
...
+		result.filter = bloom.NewXor(len(filtedTransactionsIDs), &s.xorBuilder)
```
tsachiherman, line 154: it's ok that you've used the new filter here, but you also need to add this on lines 123-124 ( or unify these somehow )

tsachiherman (dismissed):
few small requests. Looks good otherwise.

brianolson:
It is again to the state where a few integration tests fail and I don't know if I should take them seriously.

brianolson:
latest build only missed e2e TestCompactCerts on mac

tsachiherman:
> latest build only missed e2e TestCompactCerts on mac

I think that the latest change that you've make ( i.e. the one with the script time ), was needed to work around a bug in the txsync code. I think that the code changes here are good, but I'm waiting to complete the bugfix first and then retest it with these changes.

util/bloom/xor.go
```diff
@@ -0,0 +1,165 @@
...
+type XorBuilder = xorfilter.Builder
+
+// XorFilter is a faster more efficient alternative to a Bloom filter
+// An XorFilter object can be used as is or with optional adittional setup.
+type XorFilter struct {
+	//xor     *xorfilter.Xor8
+	xor     *xorfilter.Xor32
+	holding map[uint64]bool
+
+	b *XorBuilder
```
tsachiherman, line 37: nit: change `b` -> `builder`

util/bloom/xor.go
```diff
@@ -0,0 +1,165 @@
...
+// MarshalBinary implements encoding.BinaryMarshaller interface
+func (xf *XorFilter) MarshalBinary() ([]byte, error) {
+	if len(xf.holding) != 0 {
+		keys := make([]uint64, len(xf.holding))
+		pos := 0
+		for k := range xf.holding {
+			keys[pos] = k
+			pos++
+		}
+		var err error
```
tsachiherman, line 80: I think that if you're already doing a dedup, you'd want to `keys = keys[:pos]` so that you won't be passing empty duplicated entries to Populate, right ?

util/bloom/xor.go
```diff
@@ -0,0 +1,165 @@
...
+	//xor     *xorfilter.Xor8
+	xor     *xorfilter.Xor32
+	holding map[uint64]bool
+
+	b *XorBuilder
+}
+
+// NewXor returns an XorFilter with an internal map created with a size hint and an optional *XorBuilder (may be nil)
+// The Builder is not thread safe and should only be used by one thread at a time.
+func NewXor(hint int, builder *XorBuilder) *XorFilter {
```
tsachiherman, line 42: I see no issues with you using the `builder` here; however, I think that it should be mandatory to use a builder and not optional.

tsachiherman: also, you can add a hint flag here to indicate the expected number of elements.

util/bloom/xor.go
tsachiherman: yes, and that's fine - what I'm saying is that we should leave the de-dup to `MarshalBinary` and use the "lightest" data structure here.

util/bloom/xor.go
```diff
@@ -0,0 +1,165 @@
...
+	"github.com/algorand/xorfilter"
+)
+
+// XorBuilder is a local alias for xorfilter.Builder
+type XorBuilder = xorfilter.Builder
+
+// XorFilter is a faster more efficient alternative to a Bloom filter
+// An XorFilter object can be used as is or with optional adittional setup.
+type XorFilter struct {
+	//xor     *xorfilter.Xor8
```
tsachiherman, line 33: remove commented out code.

util/bloom/xor.go
```diff
@@ -0,0 +1,165 @@
...
+	k := binary.BigEndian.Uint64(x)
+	xf.holding[k] = true
+}
+
+// Test checks whether x is present in the filter.
+// May return (rare) erroneous true values, but false is precise.
+func (xf *XorFilter) Test(x []byte) bool {
+	k := binary.BigEndian.Uint64(x)
+	if xf.holding != nil {
+		return xf.holding[k]
```
tsachiherman, line 63: I think that this shouldn't be here. i.e. when we test against a bloom filter, we should make the test against the "real" xor filter ( which you have below ) rather then the `holding`, which are *temporary* array before the information get marshaled.

util/bloom/xor.go
```diff
@@ -0,0 +1,165 @@
...
+	}
+	if xf.xor != nil {
+		return xf.xor.Contains(k)
+	}
+	return false
+}
+
+// MarshalBinary implements encoding.BinaryMarshaller interface
+func (xf *XorFilter) MarshalBinary() ([]byte, error) {
+	if len(xf.holding) != 0 {
```
tsachiherman, line 73: clear the `xf.holding` after generating the inner xor filter.

util/bloom/xor.go
```diff
@@ -0,0 +1,165 @@
...
+	pos := 0
+	pos += binary.PutUvarint(out[pos:], xf.xor.Seed)
+	pos += binary.PutUvarint(out[pos:], uint64(xf.xor.BlockLength))
+	pos += binary.PutUvarint(out[pos:], uint64(len(xf.xor.Fingerprints)))
+	for _, v := range xf.xor.Fingerprints {
+		binary.LittleEndian.PutUint32(out[pos:], v)
+		pos += 4
+	}
+	//copy(out[pos:], xf.xor.Fingerprints) // for Xor8
+	//pos += len(xf.xor.Fingerprints)
```
tsachiherman, line 104: remove commented out lines.

util/bloom/xor.go
```diff
@@ -0,0 +1,165 @@
...
+	}
+	xor.BlockLength = uint32(blockLength)
+	pos += dp
+	lenFingerprints, dp := binary.Uvarint(data[pos:])
+	if dp < 0 {
+		return ErrBadBinary
+	}
+	pos += dp
+	if lenFingerprints > 0 {
+		//xor.Fingerprints = make([]byte, lenFingerprints)
```
tsachiherman, line 134: remove commented out lines.

util/bloom/xor.go
brianolson: pretty sure keys was allocated to the precise length, no extra positions.

util/bloom/xor.go
brianolson: In planned usage it won't be used. A filter will get built on one side, marshaled, unmarshaled, then used. The unmarshaled version won't have the holding map that was used for building. But if we have it there's no sense in throwing it away and not using it. (Real usage will throw away the whole XorFilter object and .holding and it will all get GC-ed.)

util/bloom/xor.go
brianolson: won't we just throw away the whole XorFilter soon on the sending side?

go.mod
```diff
@@ -9,6 +9,7 @@ require (
 	github.com/algorand/msgp v1.1.47
 	github.com/algorand/oapi-codegen v1.3.5-algorand5
 	github.com/algorand/websocket v1.4.2
+	github.com/algorand/xorfilter v0.0.0-20210423194602-fadd755a812a
```
tsachiherman, line 12: we need to create a release for the xor filter and use that release version here.

tsachiherman (changes requested):
I think that now after I reviewed both the xorfilter repo as well as this change, I have a better understanding of what you did here.

There are few design decision here that could align us for better performance.

The xor filter works great until there are duplicates. The duplicate retry rate, as defined in the xorfilter repo is 100. This is excessively high and non beneficial for us ( since it would make the runtime unpredictable ).

The txsync is already written so that there won't be any duplicate transactions. ( it's being done by the transaction pool ). However, the xorfilter does have a potential fail case when it picks a "bad" random seed.

To address these:
1. Remove the duplicate testing from the `util/bloom/xor.go`, and return the error back to the caller.
2. In `util/bloom/xor.go`, fall back from the xor/xor32 filter to the regular bloom filter.
3. Make sure you can pass the retry could to Populate32
4. In `XorFilter.Set`, using BigEndian.Uint64 isn't good enough. Instead, use the xor of all the four uint64 in that array. ( in fact, I don't mind making this into [32]byte, since this is our only usage of it anyway ).

brianolson:
There's a bit of a mismatch in that the xorfilter only considers 64 bits of any input. We guarantee no duplicate 256 bit TXID, but there's a small nonzero chance of duplication within a 64 bit prefix of that, hence my duplicate checking in the xor filter building. I haven't looked into the xorfilter algorithm to see if it would make sense to expand its input to fully consider a 256 bit input but that might make sense? (The algorithm scales readily on output to save 8..32 bits of filter data per element, but I haven't looked at the input side.)

tsachiherman:
> There's a bit of a mismatch in that the xorfilter only considers 64 bits of any input. We guarantee no duplicate 256 bit TXID, but there's a small nonzero chance of duplication within a 64 bit prefix of that, hence my duplicate checking in the xor filter building. I haven't looked into the xorfilter algorithm to see if it would make sense to expand its input to fully consider a 256 bit input but that might make sense? (The algorithm scales readily on output to save 8..32 bits of filter data per element, but I haven't looked at the input side.)

I think that ideally, the answer is yes; however, this optimization is using the uint64 over and over, and it would be pretty expensive to make that change.
Maybe try to look for a quick hash for [32]byte -> uint64 ?

brianolson:
If our hash is doing it's job, the '256 bit number' should be uniformly distributed and indistinguishable from random, so any subset of its bits should be uniformly distributed, so a 64 bit slice of its 256 bits should have just as good collision properties as a 64 bit hash of the 256 bits. Right?

tsachiherman:
> If our hash is doing it's job, the '256 bit number' should be uniformly distributed and indistinguishable from random, so any subset of its bits should be uniformly distributed, so a 64 bit slice of its 256 bits should have just as good collision properties as a 64 bit hash of the 256 bits. Right?

Yes. It's just doesn't feel right to trim it blindly and ignore the rest of the bits.

brianolson:
Even with the weird map[] to de-duplicate the [32]byte to uint64 truncation, with a little more RAM overhead the xorfilter is a bunch faster than the traditional bloom filter. I'm a little worried about getting lost in trying to have an ideal xorfilter vs comparing it vs the other algorithm. The tests here are also written to check false-positive-collision against the Algorand level [32]byte inputs, so I think we know it's working well at that level.

tsachiherman:
> Even with the weird map[] to de-duplicate the [32]byte to uint64 truncation, with a little more RAM overhead the xorfilter is a bunch faster than the traditional bloom filter. I'm a little worried about getting lost in trying to have an ideal xorfilter vs comparing it vs the other algorithm. The tests here are also written to check false-positive-collision against the Algorand level [32]byte inputs, so I think we know it's working well at that level.

The reason I think we should remove it is since it's indistinguishable ( from xor filter perspective ) with "bad" random seed. So, since we're going to handle that case anyway, and since we don't really have any duplicate use case by design, it would be better removed.

Separately, I'm trying to eliminate as many memory allocations as possible. This allocation of the map ( and it's internals ) is a high-frequency allocated ( in particular on relays ).

I don't think that merging this code as is would be a bad thing; but I think that we will need to add the above changes before calling it "done".

brianolson:
I still think duplicate testing has to happen because the xorfilter _will_ break if given duplicates, so avoiding that seems good.

100 iterations is too many. How many actually happen? I tweaked the benchmark on 10,000 uint64 keys:
`iterations min=1/avg=1.11/max=6 (N=314342)`
So I'b be perfectly happy with an iteration limit of 10, then fallback to the standard bloom filter.

I also still think the 8 byte slice of the well distributed 32 byte key is a fine way to fit into xorfilter's uint64 world. They're not doing fancy hashes inside xorfilter, so let's not overthink this. SHA512_256 is already way beyond what xorfilter is doing.

tsachiherman (approved):

PR #2003, opened 2021-03-23
Add comments to releases page script.

onetechnical:
## Summary

The releases page script could use a little explanation, so added comments for consideration.

## Test Plan

This is a comments / line formatting fix.


algobarb (approved):
Always good to add useful comments.

egieseke (approved):

PR #2006, opened 2021-03-24
Improve memory allocations in cloneAssetHoldings and cloneAssetParams

algorandskiy:
## Summary

before mem 1.83GB, allocs 1773556
after  mem 1.22GB, allocs 962952

## Test Plan

Added a benchmark

### DIFF
--- a/network/limited_reader_slurper.go
+++ b/network/limited_reader_slurper.go
@@ -77,7 +77,7 @@ func (s *LimitedReaderSlurper) Read(reader io.Reader) error {
 					// if we received err == nil and n == 0, we should retry calling the Read function.
 					continue
 				default:
-					// if we recieved a non-io.EOF error, return it.
+					// if we received a non-io.EOF error, return it.
 					return err
 				}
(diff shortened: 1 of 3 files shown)


tsachiherman (commented):
This change looks fine; I'd like to hold it off until we can evaluate its proven metrics on a loaded system test ( if it will go into master, it would be harder to roll back this change ).

network/limited_reader_slurper.go
```diff
@@ -77,7 +77,7 @@ func (s *LimitedReaderSlurper) Read(reader io.Reader) error {
 					// if we received err == nil and n == 0, we should retry calling the Read function.
 					continue
 				default:
-					// if we recieved a non-io.EOF error, return it.
+					// if we received a non-io.EOF error, return it.
```
tsachiherman, line 80: I found ~10 more instances of this typo. we can fix those in subsequent PRs.

tsachiherman (approved):

PR #2009, opened 2021-03-24
Attempt to pull from a custom cache of images

onetechnical:
## Summary

Docker Hub now limits the number of pulls against its repository for unauthenticated users. While we can login to Travis in job using secret variables, these are not available in forked repositories. As a result, PRs coming from forks that hit the docker pull limit will start failing. This is most notable in integration tests.

To work around this, we cache the images requested to our custom public repository that does not require authentication. This is not a good solution, but will help work around pulls in Travis for forked repositories for now. Bandwidth usage should be monitored, both from activity as well as for potential for abuse.

## Test Plan

See if integration tests can pull images.

### DIFF
--- a/scripts/release/test/util/test_package.sh
+++ b/scripts/release/test/util/test_package.sh
@@ -35,7 +35,7 @@ build_images () {
     # We'll use this simple tokenized Dockerfile.
     # https://serverfault.com/a/72511
     TOKENIZED=$(echo -e "\
-FROM {{OS}}\n\n\
+FROM public.ecr.aws/i3h3n7g0/{{OS}}\n\n\
 WORKDIR /root\n\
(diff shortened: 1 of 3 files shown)


algorandskiy:
weird, although 3.4.7 was released few hours ago.
```
Building wheel for cryptography (PEP 517): finished with status 'error'
8495
  error: can't find Rust compiler
```

They switched to Rust 1.41 quite long ago in 3.4.5
> 3.4.5 - 2021-02-13
Various improvements to type hints.
Lower the minimum supported Rust version (MSRV) to >=1.41.0. This change improves compatibility with system-provided Rust on several Linux distributions.

onetechnical:
Using the approach in #2012 instead.

PR #2012, opened 2021-03-25
Cache docker images in Travis and re-load them

onetechnical:
## Summary

To get around Docker Hub pull limits, we can try to explicitly cache docker images between builds. This seems like a safer approach than #2009.

## Test Plan

Verify docker images are saved by running this, then re-running and seeing if it's cached.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -165,6 +165,7 @@ jobs:
 cache:
   directories:
     - crypto/lib
+    - docker_cache
 
 before_install:
 - |-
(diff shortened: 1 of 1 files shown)


algobarb (approved):

egieseke (approved):

.travis.yml
```diff
@@ -198,6 +200,7 @@ before_cache:
      $msys2 pacman --sync --clean --noconfirm
      ;;
    esac
+   docker save -o $HOME/docker_cache/images.tar $(docker images -a -q)
```
bricerisingalgorand, line 203: oh that's fancy, i never knew you could tar several images together

bricerisingalgorand (approved):

PR #2015, opened 2021-03-25
Optimistic RoundCowState updates

tsachiherman:
## Summary

This PR includes two memory-allocation optimizations in the RoundCowState model:
1. Transactions insertion to the RoundCowState rarely fails.
    This, in turn means that creating a child RoundCowState, storing the deltas there and later on pushing them to the parent would not be the most efficient way to achieve this. Instead, we push the changes *directly* to the parent, and rolling these changes back in case the transaction verification fails. That wouldn't work correctly for *any* change, but would be fine for Txid and Txleases.
2.  Dynamically allocate Creatables.
     Most of the transactions we have do not create/destroy assets. Instead, they are mostly asset transfer or payment transactions. In these cases, there is no reason to allocate the Creatables map for each child RoundCowState.

## Test Plan

Use the existing test to verify correctness.


algorandskiy (commented):
I'm not sure I like `Txids`/`Txleases` part since it makes boundaries between `BlockEvaluator` and `StateDelta` even more fuzzy, and makes the code even more complex.
What is memory gain? I mean we already allocate enough space in parent cow to hold entire evaluation set, so there should not be any re-allocs when copying from children. Every children doubles memory, but they are sequential and no more than 16 txn in each child

tsachiherman:
Yes, I feel the same way. It reduces the amount of memory child cowroundstate object allocates.

From the bench that I was doing, it *looks* like it was ~10% less memory. In an ideal GC system, it should have made no difference whatsoever. However, GO tend to leave some objects around, longer than we would really wish it did.

tsachiherman:
btw - the memory allocation in the children isn't so much about their size, but more about their count.

tsachiherman:
I'm going go shelf this one for now. If https://github.com/algorand/go-algorand/pull/2016 would be sufficient to address most of the concerns, then I'm fine with it.

PR #2026, opened 2021-03-29
clear unused test code

algonautshant:
## Summary
- Remove testing tools which are no longer used
- Add timeout to TestAbruptWrites to avoid occasional failures

## Test Plan
No new code or functionality added. Existing test should be sufficient.

### DIFF
--- a/catchup/service_test.go
+++ b/catchup/service_test.go
@@ -262,6 +262,9 @@ func TestServiceFetchBlocksOneBlock(t *testing.T) {
 	require.Equal(t, *block, localBlock)
 }
 
+// TestAbruptWrites emulates the fact that the agreement can also generate new rounds
+// When caught up, and the agreement service is taking the lead, the sync() stops and
+// yields to the agreement. Agreement is emulated by the go func() loop in the test
 func TestAbruptWrites(t *testing.T) {
 	numberOfBlocks := 100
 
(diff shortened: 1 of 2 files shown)


catchup/service_test.go
```diff
@@ -317,6 +317,13 @@ func TestAbruptWrites(t *testing.T) {
 	s.testStart()
 
 	s.sync()
+
+	for try:=0; try < 100; try++ {
+		if remote.LastRound() == local.LastRound() {
+			break
+		}
+		time.Sleep(10*time.Millisecond)
+	}
```
tsachiherman, line 326: Under what condition does `s.sync()` fail to complete the synchronization before returning ?

algonautshant: Ledger `local` is getting blocks added to it through `sync()` and through the for loop in `go func()` above in the test.

When a block is written to the ledger via `mockedLedger AddBlock` above (service_test.go line 311), it closes the wait chan for that round, which triggers the context cancel function in service and `fetchAndWrite` gets "context canceled" err in `innerFetch`. 

Eventually, `fetchAndWrite` reaches the `catchupRetryLimit` for the round and returns. When `fetchAndWrite` returns `false`, `pipelineCallback` returns 0, which eventually leads to round 0 returned from `completed` chan in `pipelinedFetch` loop, and returns. This ends the `sync()`. 

When `sync()` returns, it happens that the for loop above is still adding blocks to the local ledger. This wait is for that to finish. 

To properly fix this, I need to understand the expected behavior of the test. Why two different routines are writing to the ledger simultaneously, and which eventually ends up the go func in the test writing all the blocks.

tsachiherman: It emulates that fact that the agreement can also generate new rounds.

tsachiherman: I think that in this case, you'd want to have the `wg.Wait()` right after the `s.sync`

tsachiherman (approved):
looks good.

PR #2027, opened 2021-03-29
ledger: split committed round range by consensus protocol version

tsachiherman:
## Summary

Split the update range to contain a single consensus version at a time. This change is a pre-requisite for the 1000 assets upcoming change.

## Test Plan

- [x] Unit tests added.
- [x] Tested reconstructing tracker database from scratch.
- [x] Tested catching up a mainnet node from scratch.


ledger/acctupdates.go
```diff
@@ -712,6 +711,19 @@ func (au *accountUpdates) committedUpTo(committedRound basics.Round) (retRound b
 
 	offset = uint64(newBase - au.dbRound)
 
+	// check if this update chunk spans across multiple consensus versions. If so, break it so that each update would tackle only a single
+	// consensus version.
+	if au.versions[1] != au.versions[offset] {
```
algorandskiy, line 716: why `version[1]` and not `[0]` ?

ledger/acctupdates_test.go
```diff
@@ -1637,3 +1637,256 @@ func TestCachesInitialization(t *testing.T) {
...
+		ml.addMockBlock(blockEntry{block: blk}, delta)
+		au.newBlock(blk, delta)
+		accts = append(accts, totals)
+		rewardsLevels = append(rewardsLevels, rewardLevel)
+	}
+
+	newVersionBlocksCount := uint64(47)
+	newVersion := protocol.ConsensusV21
+	// add 47 more rounds that contains blocks using a newer consensus version, and stuff it with MaxBalLookback
+	for i := basics.Round(initialRounds + extraRounds); i < basics.Round(initialRounds+initialProtoParams.MaxBalLookback+extraRounds+newVersionBlocksCount); i++ {
```
algorandskiy, line 1715: move `basics.Round(initialRounds+initialProtoParams.MaxBalLookback+extraRounds+newVersionBlocksCount` into a separate var and re-use below in `committedUpTo` to make it obvious what happens.

ledger/acctupdates_test.go
```diff
@@ -1637,3 +1637,256 @@ func TestCachesInitialization(t *testing.T) {
...
+		delta.Accts.MergeAccounts(updates)
+		ml.addMockBlock(blockEntry{block: blk}, delta)
+		au.newBlock(blk, delta)
+		accts = append(accts, totals)
+		rewardsLevels = append(rewardsLevels, rewardLevel)
+	}
+	// now, commit and verify that the committedUpTo method broken the range correctly.
+	au.committedUpTo(basics.Round(initialRounds + extraRounds + initialProtoParams.MaxBalLookback))
+	au.waitAccountsWriting()
+	require.Equal(t, basics.Round(initialRounds+extraRounds)-1, au.dbRound)
```
algorandskiy, line 1857: Who is this loop different from the last one from the prev test? there is an additional `newVersionBlocksCount` value but it looks not really needed since we add more than `initialRounds+extraRounds)-1` and check that only `initialRounds+extraRounds)-1` is committed at single `committedUpTo` call.

algorandskiy (commented):
Looks good, couple questions.

ledger/acctupdates.go
tsachiherman: The versions array is shifted by one ( just like it's `protos` predecessor), and so the indexing there start with one ( and for the same reason - we address au.versions[offset] and not au.versions[offset-1] )

ledger/acctupdates_test.go
tsachiherman: This test ensures that we know not only to break the rounds at the right place, but also to continue from the right place on the subsequent write.

Now that the code is properly written, it would be impossible for you to create a case where it's not happening, but during development it did, and this was the test I used to verify it against.

tsachiherman: btw - I agree that there is no need for writing `extraRounds` here. One round would be enough. But for verification purposes, having multiple is usually better than having just one.

ledger/acctupdates.go
algorandskiy: and what is in `au.protos[0]`?

algorandskiy (approved):

ledger/acctupdates.go
tsachiherman: keep in mind that offset 0, which aligns with `dbRound` is already stored on disk. When we persist rounds, we are going to start off dbRound+1, which perfectly align with the above.

PR #2028, opened 2021-03-29
Filter out automated testing useless verbosed output

tsachiherman:
## Summary

Filter out automated testing useless verbosed output.

## Test Plan

Add a unit test with "sample" cases and their corresponding expected output.


algonautshant (approved):
Very nice!

PR #2031, opened 2021-03-30
Fix testing bug : avoid datarace on a node crash during e2e test

tsachiherman:
## Summary

Synchronize testing.T access to avoid generating a datarace in case of a node crash feedback.
When the node controller notice that the node has exit, it reports it back to the test fixture.
However, the test fixture cannot report that to the underlying test since doing so would create a data race : the testing.TB is not meant to support concurrency.

To address that, I've created an abstraction above the testing.TB called TestingTB, which would retain the same functionality, but would use a mutex to synchronize the access.

## Test Plan

Tested manually to confirm correctness.

### DIFF
--- a/test/e2e-go/features/catchup/basicCatchup_test.go
+++ b/test/e2e-go/features/catchup/basicCatchup_test.go
@@ -35,7 +35,7 @@ func TestBasicCatchup(t *testing.T) {
 		t.Skip()
 	}
 	t.Parallel()
-	a := require.New(t)
+	a := require.New(fixtures.SynchronizedTest(t))
 
 	// Overview of this test:
 	// Start a two-node network (primary has 0%, secondary has 100%)
@@ -99,7 +99,7 @@ func runCatchupOverGossip(t *testing.T,
 	if testing.Short() {
 		t.Skip()
 	}
-	a := require.New(t)
+	a := require.New(fixtures.SynchronizedTest(t))
 	// Overview of this test:
 	// Start a two-node network (Primary with 0% stake, Secondary with 100% stake)
 	// Kill the primary for a few blocks. (Note that primary only has incoming connections)
@@ -198,7 +198,7 @@ func TestStoppedCatchupOnUnsupported(t *testing.T) {
 		t.Skip()
 	}
 	t.Parallel()
-	a := require.New(t)
+	a := require.New(fixtures.SynchronizedTest(t))
 
 	consensus := make(config.ConsensusProtocols)
 	// The following two protocols: testUnupgradedProtocol and testUnupgradedToProtocol
--- a/test/e2e-go/features/catchup/catchpointCatchup_test.go
+++ b/test/e2e-go/features/catchup/catchpointCatchup_test.go
@@ -82,7 +82,7 @@ func TestBasicCatchpointCatchup(t *testing.T) {
 	if testing.Short() {
 		t.Skip()
 	}
-	a := require.New(t)
+	a := require.New(fixtures.SynchronizedTest(t))
 	log := logging.TestingLog(t)
 
 	// Overview of this test:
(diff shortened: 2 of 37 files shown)


test/framework/fixtures/kmdFixture.go
```diff
@@ -94,72 +94,72 @@ func (f *KMDFixture) ShutdownImpl(preserveData bool) {
...
+	require.NoError(f.t, err)
 	err = os.Mkdir(f.dataDir, 0750)
-	require.NoError(t, err)
+	require.NoError(f.t, err)
 
 	// Set up the kmd data dir within the main datadir
 	f.kmdDir = filepath.Join(f.dataDir, nodecontrol.DefaultKMDDataDir)
 	err = os.Mkdir(f.kmdDir, nodecontrol.DefaultKMDDataDirPerms)
-	require.NoError(t, err)
+	require.NoError(f.t, err)
```
algonautshant, line 122: Any reason why the require.New pattern is not used in this file?

test/e2e-go/features/catchup/catchpointCatchup_test.go
```diff
@@ -82,7 +82,7 @@ func TestBasicCatchpointCatchup(t *testing.T) {
 	if testing.Short() {
 		t.Skip()
 	}
-	a := require.New(t)
+	a := require.New(fixtures.SynchronizedTest(t))
```
algonautshant, line 85: Does this need to be updated as well?
```
func (ec *nodeExitErrorCollector) Print() {
...
		require.NoError(ec.t, err, ec.messages[i])

```

test/e2e-go/features/catchup/basicCatchup_test.go
```diff
@@ -35,7 +35,7 @@ func TestBasicCatchup(t *testing.T) {
 		t.Skip()
 	}
 	t.Parallel()
-	a := require.New(t)
+	a := require.New(fixtures.SynchronizedTest(t))
```
algonautshant, line 38: Does this need to be updated?
```
func TestCatchupOverGossip(t *testing.T) {
	t.Parallel()
	supportedVersions := network.SupportedProtocolVersions
	require.LessOrEqual(t, len(supportedVersions), 3)

```

algonautshant (approved):
Looks very nice. 
A few comments.

test/framework/fixtures/fixture.go
```diff
@@ -50,3 +63,116 @@ type Fixture interface {
...
+	defer st.Unlock()
+	st.t.Skipf(format, args...)
+}
+func (st *synchTest) Skipped() bool {
+	st.Lock()
+	defer st.Unlock()
+	return st.t.Skipped()
+}
+
+/*func (st *synchTest) TempDir() string {
```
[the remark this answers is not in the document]
tsachiherman: it's only on go 1.15, which we don't use. I'll drop the comments, since they aren't needed.

test/framework/fixtures/kmdFixture.go
tsachiherman: The reason for the `require.New` pattern was to avoid repeating passing the testing.T, since it would require using the `SynchronizedTest` to ensure its synchronized. 
For this fixture, we have that guarantee already, since when we store the KMDFixture.t in Initialize() we always use `SynchronizedTest`.

test/e2e-go/features/catchup/catchpointCatchup_test.go
tsachiherman: Yes, but the change won't be on that line.

test/e2e-go/features/catchup/basicCatchup_test.go
tsachiherman: yes, I'll update that.

algonautshant (approved):

PR #2032, opened 2021-03-30
node controller: delete pid file on process killing

tsachiherman:
## Summary

Existing code wait 30 seconds between sending SIGTERM and SIGKILL. When sending the SIGKILL, the process is deleted, but the pid file remains on disk.

Leaving the pid file on disk could cause subsequent failures - and we could easily avoid them by clearing this file if we SIGKILL'ed the process.

## Test Plan

Tested manually.

### DIFF
--- a/nodecontrol/NodeController.go
+++ b/nodecontrol/NodeController.go
@@ -111,26 +111,26 @@ func (nc NodeController) stopProcesses() (kmdAlreadyStopped bool, err error) {
 	return
 }
 
-func killPID(pid int) error {
+func killPID(pid int) (killed bool, err error) {
 	process, err := util.FindProcess(pid)
 	if process == nil || err != nil {
-		return err
+		return false, err
 	}
 
 	err = util.KillProcess(pid, syscall.SIGTERM)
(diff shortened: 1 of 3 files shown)


nodecontrol/NodeController.go
```diff
@@ -111,26 +111,26 @@ func (nc NodeController) stopProcesses() (kmdAlreadyStopped bool, err error) {
 	return
 }
 
-func killPID(pid int) error {
+func killPID(pid int) (killed bool, err error) {
```
algonautshant, line 114: Instead of returning killed, and deleting the pid file, why not pass the pid file path to killPID, and delete the file?

algonautshant (approved):
Looks good.

nodecontrol/NodeController.go
tsachiherman: I think that the `killPID` should be focused on *just* killing the process ( without worrying how the pid was obtained ).
The caller, who have the knowledge of how the pid was obtained, can delete the file.
( either directly or indirectly ).

PR #2033, opened 2021-03-30
adding extra docs for pushbytes and pushint

barnjamin:
Mention that the args to ops pushbytes and pushint do not get added to the bytecblock or intcblock.

This replaces another PR with docs generated using Makefile since it was created with the edit feature.


brianolson (approved):

jannotti (approved):
I'm ok with this.  There could be an argument against it since *none* of the opcodes actually add to the constant blocks, only the pseudo-ops do that.  But I can understand the confusion, so let's clear it up.

PR #2038, opened 2021-04-01
Fix ClearState program applying when it errs

algorandskiy:
## Summary

* Apps logic needs to ignore result and error from logic evaluator but but fail on other errors
* The bug happened in app refactor PR
* Added unit and e2e test

## Test Plan

Added e2e test that ensures success with clear state programs that always fail or err.
Added unit test that check clear state error does not lead to `apply.ApplicationCall` error

### DIFF
--- a/ledger/appcow.go
+++ b/ledger/appcow.go
@@ -25,6 +25,7 @@ import (
 	"github.com/algorand/go-algorand/data/transactions"
 	"github.com/algorand/go-algorand/data/transactions/logic"
 	"github.com/algorand/go-algorand/ledger/apply"
+	"github.com/algorand/go-algorand/ledger/ledgercore"
 	"github.com/algorand/go-algorand/protocol"
 )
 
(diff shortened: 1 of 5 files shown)


tsachiherman (approved):
Looks good.

algorandskiy:
Added `ledgercore.LogicEvalError` to distinguish logic eval errors from others.
Relevant code from pre-refactor can found [here](https://github.com/algorand/go-algorand/blob/573a34c457e9ac5e4e4f63355f0b295ced001083/ledger/apply/application.go#L382)

tsachiherman:
> Added `ledgercore.LogicEvalError` to distinguish logic eval errors from others.
> Relevant code from pre-refactor can found [here](https://github.com/algorand/go-algorand/blob/573a34c457e9ac5e4e4f63355f0b295ced001083/ledger/apply/application.go#L382)

It will take me a bit longer to verify it's correctness.

tsachiherman (approved):
Yes, the fix looks to be correct, and it correctly retain previous behavior.
One note is that I was surprised to see the usage of the LogicEvalError; I'm usually think of interfaces as a pointer to structures, and was surprised to see that you're using it without dereferencing it. Your usage here is fine, but be aware that if you had to make any changes to the error object, you would be using a pointer receiver on the Error method, and then being forced to dereference the `LogicEvalError`.

algorandskiy:
We actually have similar type assertion for errors in `switch/case` without accessing the error itself.

PR #2042, opened 2021-04-05
Tealdbg don't assume program version

jasonpaulos:
## Summary

Currently the TEAL debugger always compiles programs with the latest TEAL version. This means that programs using prior versions are unable to be properly debugged. This PR fixes this issue.

## Test Plan

I've added a regression test case to `tealdbgTest.exp` that uses a version lower than the maximum version the protocol supports.

### DIFF
--- a/cmd/tealdbg/local.go
+++ b/cmd/tealdbg/local.go
@@ -336,6 +336,9 @@ func (r *LocalRunner) Setup(dp *DebugParams) (err error) {
 			if IsTextFile(data) {
 				source := string(data)
 				ops, err := logic.AssembleString(source)
+				if ops.Version > r.proto.LogicSigVersion {
+					return fmt.Errorf("Program version (%d) is beyond the maximum supported protocol version (%d)", ops.Version, r.proto.LogicSigVersion)
+				}
 				if err != nil {
 					errorLines := ""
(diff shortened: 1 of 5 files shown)


tsachiherman:
Do we need to add a check to ensure that the protocol version supports the requested TEAL version ?

jasonpaulos:
@tsachiherman good point, I'll add a check for this.

cmd/tealdbg/local.go
```diff
@@ -335,7 +335,10 @@ func (r *LocalRunner) Setup(dp *DebugParams) (err error) {
 			r.runs[i].program = data
 			if IsTextFile(data) {
 				source := string(data)
-				ops, err := logic.AssembleStringWithVersion(source, r.proto.LogicSigVersion)
+				ops, err := logic.AssembleString(source)
```
algorandskiy, line 338: Old code allowed to control assembly version by setting`--proto` cli option. New code relies only on `#pragma version` statement.

Since `--proto` is not used by assembler anymore, this needs to be rephrased:
```go
debugCmd.Flags().StringVarP(&proto, "proto", "p", "", "Consensus protocol version for TEAL")
```
to something like "protocol version for TEAL evaluation".

algorandskiy (approved):

PR #2049, opened 2021-04-07
bugfix: ensure loading of a merkle trie deferred page during commit

tsachiherman:
## Summary

When using the MerkleTrie, the trie avoid making loads from disk for pages that aren't needed. In particular, it won't load the latest page ( known as the deferred page ) until it needs to commit it.

The implementation had a bug in the `loadPage` method, where it would reset the loading page flag incorrectly.

## Test Plan

1. Unit test added.
2. Tested against mainnet and testnet.


algonautshant (approved):
Looks good.

PR #2050, opened 2021-04-08
Extend catchpointdump utility to verify merkle trie 

tsachiherman:
## Summary

This PR extends the `catchpointdump` utility and add the ability to scan a tracker database and retrieve the merkle trie statistics. The statistics themselves aren't very useful on their own, but the scan process verify the structure of the merkle trie, allowing us to verify the consistency of a given database.

## Test Plan

Tested manually against multiple databases.

### DIFF
--- a/crypto/merkletrie/committer_test.go
+++ b/crypto/merkletrie/committer_test.go
@@ -221,11 +221,9 @@ func TestIterativeCommits(t *testing.T) {
 	require.NoError(t, err)
 
 	for r := 0; r < 100; r++ {
-		/*if r%2 == 1 {*/
 		newMC := mc.Duplicate(true)
 		mt, _ = MakeTrie(newMC, memConfig)
 		mc = newMC
-		//}
 
 		for k := r * 5; k < r*7+len(hashes); k++ {
 			i := k % len(hashes)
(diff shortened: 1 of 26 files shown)


crypto/merkletrie/committer_test.go
```diff
@@ -187,3 +193,51 @@ func TestMultipleCommits(t *testing.T) {
...
+		require.True(t, added)
+		require.NoError(t, err)
+	}
+	_, err := mt.Commit()
+	require.NoError(t, err)
+
+	for r := 0; r < 100; r++ {
+		newMC := mc.Duplicate(true)
+		mt, _ = MakeTrie(newMC, memConfig)
+		mc = newMC
```
algonautshant, line 226: Why mc is duplicated to newMC, and then the original (mc) is discarded?

algonautshant (approved):
Looks good.

crypto/merkletrie/committer_test.go
tsachiherman: I wanted to ensure that there is no chance of memory slices being shared. It is intended to simulate persisting to disk and reloading from disk.

PR #2053, opened 2021-04-08
Use shared encoding buffer for the encoded transaction groups

tsachiherman:
## Summary

Use shared encoding buffer.

## Test Plan

Existing tests already provide coverage for the modified functionality.


nicholasguoalgorand:
The purpose of this change is to just use one buffer so msgp doesn't keep allocating new buffers correct?

tsachiherman:
it's basically using the sync.Pool buffer to recycle buffers. msgp likes to grow buffers - which becomes extremely inefficient.
By recycling buffers we're ensuring that we don't allocate new buffers if the ones in the pool are sufficiently large.
The sync.Pool has some special handling with the GC.

PR #2054, opened 2021-04-08
make the msgp less verbosed on the success case

tsachiherman:
## Summary

The message pack generator is very noisy. It tends to emit lot of messages that aren't being used.
This PR cache the output of the message pack so that it would only present these in case of an error.

## Test Plan

Tested on travis.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -107,7 +107,18 @@ generate: deps
 msgp: $(patsubst %,%/msgp_gen.go,$(MSGP_GENERATE))
 
 %/msgp_gen.go: deps ALWAYS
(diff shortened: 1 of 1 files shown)


algobarb (approved):
So in the success case, it outputs into ./$@.out and if it fails, it will printf the .out file to the console, then it removes the out file again at the very end whether it succeeds or not. So basically a try/catch use case.

Interesting way to go about it, it looks like a small change/non-breaking so I can approve it, but why do we rm the out file twice in the fail case? Why don't we just have the `rm -f ./$@.out` at the end instead?

PR #2060, opened 2021-04-12
Fine tune catchup peer selection logic

algonautshant:
## Summary

The peer selector has the following goals:
1. Block malicious hosts that return invalid blocks
2. Pick hosts that perform better than others
3. Adapt to changes to the performance of hosts
4. Avoid hosts that download fails

The current implementation of the peer selector meets all these goals, but it is not flexible to random failures.
- It will rank the peers by the first download time. When a good performing peer returns a bad value in the first download, it will have a low rank and may never be selected. 
- When the download fails from a good performing peer, that peer may never be considered again (until it runs out of all other peers)
- The top performing peer is selected all the time, until it's performance degrades. It will never consider other peers as long as the top peer's rank is better than others obtained in the beginning.

These situations make a big difference when using archival buckets. When switching from a peer in the same region to a peer in a different region, the performance difference is huge. If a local peer fails occasionally, the peer selector needs to be smart enough to evaluate it based on it's overall performance, and not just the most recent performance. 

This change enhances the peer selector so that it implements the above 4 points, and is resilient to occasional performance changes:
- It established the initial ranks with multiple download value points
- It penalizes a peer for frequently being selected, to also consider other peers
- A single download failure will be forgiven if the peer is otherwise performing well

These flexibility is implemented with:
- Historical ranks. A moving window of the last 100 ranks is stored, and the new ranks is calculates by taking the average of these values.
- Frequency penalty. When a peer is frequently selected, it will be penalized for that. When a peer is not selected, then the penalty will be reduced. This is done by storing a window of its selection frequency, and the penalty is higher (exponential) if the peer is continuously selected. The penalty will quickly get high enough to favor the next best ranking peer. This way, the selection will be diversified.  

## Test Plan
Tests added to confirm the new features are properly implemented.

### DIFF
--- a/catchup/peerSelector.go
+++ b/catchup/peerSelector.go
@@ -197,8 +197,9 @@ func (hs *historicStats) updateRequestPenalty(counter uint64) float64 {
 	return hs.computerPenalty()
 }
 
-func (hs *historicStats) resetRequestPenalty(steps int, initialRank int) (int) {
+func (hs *historicStats) resetRequestPenalty(steps int, initialRank int) int {
 	if len(hs.requestGaps) == 0 || len(hs.rankSamples) == 0 {
+		fmt.Println("_")
 		return initialRank
 	}
 	if steps == 0 {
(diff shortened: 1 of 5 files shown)


tsachiherman (approved):
looks great. Thanks!

PR #2039, opened 2021-04-02
Bandwidth watch

brianolson:
## Summary

Expand test/heapwatch/heapWatch.py to also gather /metrics and block headers.
Add a reporting script to measure differences across metrics and blocks.

## Test Plan

```sh
rm -rf ~/Algorand/tn3/
bash test/heapwatch/bwstart.sh ~/Algorand/tn3/
sleep 950
bash test/heapwatch/stop.sh ~/Algorand/tn3/

python3 test/heapwatch/metrics_delta.py --report ~/Algorand/tn3/report.csv --deltas ~/Algorand/tn3/txsd.csv ~/Algorand/tn3/heaps/Primary*metrics
```

(run again with test/heapwatch/* scripts copied onto branch `feature/txnsync`, compare `algod_network_sent_bytes_*` )

PR #2063, opened 2021-04-12
netgoal: fixup small amounts of float64 roundoff

brianolson:
## Summary

Genesis blocks (now usually for test networks) initial stake must add up to exactly the total money, but the network description is in float64 and there can be small roundoff to fixup.

## Test Plan

Added unit test that has 5 out of 10,000,000,000,000,000 micro-Algos roundoff in naive wallet generation.

### DIFF
--- a/gen/generate.go
+++ b/gen/generate.go
@@ -19,6 +19,7 @@ package gen
 import (
 	"fmt"
 	"io/ioutil"
+	"math"
 	"os"
 	"path/filepath"
 	"runtime"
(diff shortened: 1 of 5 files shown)


brianolson:
found on the way to building https://github.com/algorand/go-algorand/pull/2039

gen/generate_test.go
```diff
@@ -100,3 +103,22 @@ func TestLoadSingleRootKeyConcurrent(t *testing.T) {
 	}
 	wg.Wait()
 }
+
+func TestGenesisRoundoff(t *testing.T) {
+	defer func() {
+		verboseOutWriter = os.Stdout
```
tsachiherman, line 109: I like the idea of having the ability to redirect output during tests. However, this would cause race conditions.
Maybe you could pass the output file writer as an argument to the `GenerateGenesisFiles `/ `setupGenerateGenesisFiles` method instead ? i.e. instead of passing the `verbose` flag, pass in this *os.File target.

brianolson: ok, I like the idea of an io.Writer as the `verbose` flag instead of a bool. changing...

gen/generate.go
```diff
@@ -79,28 +113,43 @@ func GenerateGenesisFiles(genesisData GenesisData, consensus config.ConsensusPro
...
+				} else {
+					if allocation[i].Stake > consensusParams.MinBalance {
+						allocation[i].Stake--
+						sum--
+					}
+				}
+				i = (i + 1) % len(allocation)
+			}
+		} else {
+			panic(fmt.Sprintf("Amounts don't add up to TotalMoney - off by %v", int64(TotalMoney)-int64(sum)))
```
shiqizng, line 137: what kinds of values would lead to this panic? one of the inputs to this func() is genesis data which could be created from netgoal generate with specified parameters, should I generate an another genesis file if I reach this panic line?

gen/generate.go
```diff
@@ -79,28 +113,43 @@ func GenerateGenesisFiles(genesisData GenesisData, consensus config.ConsensusPro
...
-	proto := genesisData.ConsensusProtocol
-	if proto == protocol.ConsensusVersion("") {
-		proto = protocol.ConsensusCurrentVersion
+		fsum := float64(sum)
+		ftot := float64(TotalMoney)
+		if (math.Abs((fsum-ftot)/ftot) < 0.01) && (u64absDiff(sum, TotalMoney) < 10000) {
+			if verbose {
+				fmt.Fprintf(verboseOutWriter, "doing roundoff fixup expected total money %d actual sum %d\n", TotalMoney, sum)
+			}
+			// wallet stake is a float and roundoff might happen but we might be close enough to do fixup
```
shiqizng, line 122: do you know why wallet stake is a float while genesis stake is an int64? shouldn't wallet stake also be represented in int64 to avoid floating-point problem?

gen/generate_test.go
```diff
@@ -100,3 +103,18 @@ func TestLoadSingleRootKeyConcurrent(t *testing.T) {
 	}
 	wg.Wait()
 }
+
+func TestGenesisRoundoff(t *testing.T) {
+	verbosity := strings.Builder{}
+	genesisData := DefaultGenesis
+	genesisData.NetworkName = "wat"
+	genesisData.ConsensusProtocol = protocol.ConsensusCurrentVersion // TODO: also check ConsensusFuture ?
```
tsachiherman, line 111: I don't think that adding a different `ConsensusProtocol` would make any difference in the way we setup the accounts, wouldn't it ?

tsachiherman (dismissed):
looks good to me.

gen/generate.go
brianolson: It's a float64 percent; [0.0 .. 100.0]; for being human-convenent.

tsachiherman (approved):

PR #2066, opened 2021-04-13
Log error conditions of invalid incoming messages

tsachiherman:
## Summary

Log error conditions of invalid incoming messages.

## Test Plan

Changes are trivial.


nicholasguoalgorand (approved):
This looks good to me. Just checking, but is Info the right log level to use instead of Warn?

tsachiherman:
> This looks good to me. Just checking, but is Info the right log level to use instead of Warn?

For load-related messages, I think that we don't want to send warns. But if we find that we want to change the log level later on, I'm pretty much flexible around that.

PR #2067, opened 2021-04-13
goal: improve waitForCommit logic

tsachiherman:
## Summary

The existing implementation of `waitForCommit` had two deficiencies:
- It was waiting for two rounds at the time, instead of waiting for a single round before re-testing the pending transactions.
- It wasn't limiting it's execution time to the transaction's last valid ( on the client side ).
   - Note that there is a server side handling for that, which returns an error "kicked out of local node pool".
   - This would ensure we have a proper working client in case the server-side restarts.

## Test Plan

Existing unit tests provide good coverage. I've added an e2e test to verify the transaction timeout.

### DIFF
--- a/cmd/goal/clerk.go
+++ b/cmd/goal/clerk.go
@@ -151,17 +151,21 @@ var clerkCmd = &cobra.Command{
 	},
 }
 
-func waitForCommit(client libgoal.Client, txid string, lastWaitRound uint64) error {
+func waitForCommit(client libgoal.Client, txid string, transactionLastValidRound uint64) error {
 	// Get current round information
 	stat, err := client.Status()
 	if err != nil {
 		return fmt.Errorf(errorRequestFail, err)
 	}
 
+	// check if we've already committed to the block number equals to the transaction's last valid round.
+	// if this is the case, the transaction would not be included in the blockchain, and we can exit right
+	// here.
+	if stat.LastRound >= transactionLastValidRound {
+		return fmt.Errorf(errorTransactionExpired, txid)
+	}
+
 	for {
-		if stat.LastRound > lastWaitRound {
-			return fmt.Errorf(errorTransactionExpired, txid)
-		}
 		// Check if we know about the transaction yet
 		txn, err := client.PendingTransactionInformation(txid)
 		if err != nil {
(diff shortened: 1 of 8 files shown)


cmd/goal/clerk.go
```diff
@@ -151,13 +151,20 @@ var clerkCmd = &cobra.Command{
...
 	// Get current round information
 	stat, err := client.Status()
 	if err != nil {
 		return fmt.Errorf(errorRequestFail, err)
 	}
 
+	// check if we've already committed to the block number equals to the transaction's last valid round.
+	// if this is the case, the transaction would not be included in the blockchain, and we can exit right
+	// here.
+	if stat.LastRound >= transactionLastValidRound {
```
algonautshant, line 164: Why this check here? 
If the txn is accepted, but the node is past transactionLastValidRound, it will correctly return. 
But, wouldn't the error errorTransactionExpired be wrong?

algonautshant (dismissed):
Looks good.
I have one question.

cmd/goal/clerk.go
tsachiherman: yes, I think that you're correct. If the transaction was sent, we have to test it's acceptance using the `client.PendingTransactionInformation` before testing for expiration.

cmd/goal/clerk.go
```diff
@@ -151,7 +151,7 @@ var clerkCmd = &cobra.Command{
 	},
 }
 
-func waitForCommit(client libgoal.Client, txid string) error {
+func waitForCommit(client libgoal.Client, txid string, transactionLastValidRound uint64) error {
```
algorandskiy, line 154: maybe we want 0 as a default value?

algorandskiy (approved):

algonautshant (approved):

cmd/goal/clerk.go
tsachiherman: Using the value of 0 would be useful for cases where we don't know what the last valid of a transaction. I don't believe we have such a use case, do we ? ( maybe if we refactor the `rawsendCmd` ).

algorandskiy (approved):

PR #2069, opened 2021-04-13
Merge master into feature/1000k-assets

algorandskiy:
## Summary

Merge + fixes


ledger/appdbg.go
```diff
@@ -62,9 +62,10 @@ func MakeDebugBalances(l ApplicationDbgLedger, round basics.Round, proto protoco
 	w := ledgerForCowBaseWrapper{l}
 
 	base := &roundCowBase{
-		l:     &w,
-		rnd:   round - 1,
-		proto: config.Consensus[proto],
+		l:        &w,
+		rnd:      round - 1,
```
tsachiherman, line 66: not really a problem here, but in general we try to avoid `round-1` and using round.SubSaturate(1) instead.

algorandskiy: there are two places like this in new code, I'll fix in part-5 PR

PR #2072, opened 2021-04-14
go-algorand 2.5.4 re-merge

onetechnical:
## Summary

Re-merge of 2.5.4-stable to master. Do NOT squash merge this PR.

## Test Plan

Standard branch testing.


tsachiherman (approved):
I'm surprised that only now we're updating the `VersionMinor = 5` on master.. but that's fine.

onetechnical:
> I'm surprised that only now we're updating the `VersionMinor = 5` on master.. but that's fine.

This is just because we do the versioning in rel/beta. I'm a little torn on this; I agree in that I feel like we should be running the next version (e.g. `VersionMinor = 6`), but we'd also have to special-case when there's a VersionMajor upgrade too.

PR #2075, opened 2021-04-14
goal node restart message when node already started

algonautshant:
## Summary
When starting a node which is already started, two messages are show: already started and successfully started:

$ goal node start
Algorand node was already started!

Only the already started message should be shown.

Also, when node restart command is given, there is a check in the code if the node was already running. 
This check comes after the node was stopped, so should never return as the node is still running. But in case it does, a message is added to notify the user. 

## Test Plan
Added a test to make sure no other output is printed.


tsachiherman (changes requested):
Looks good; could you add expect test for that ?

tsachiherman (approved):

PR #2077, opened 2021-04-15
catchup: cross verify genesis.json hash

tsachiherman:
## Summary

When using the fast catchup, the block authenticity is verified against the catchpoint label. That method bypass the local genesis.json authentication since it's not required to establish secure catchup. From that point and on, the caught-up node would be able to proceed as usual. However, when the node is restarted, it compares the on-disk genesis.json to the first 320 rounds being loaded into memory. Failing with that comparison, the node would refuse to start up.

The solution here is to prevent the node from completing the fast catchup before testing that the genesis.json's hash matches the one on the first downloaded block. From there, it would be guaranteed to be identical due to the reverse hash check.

## Test Plan

Manually tested against "corrupted" genesis.json file. Requires full catchup test.

### DIFF
--- a/catchup/catchpointService.go
+++ b/catchup/catchpointService.go
@@ -339,6 +339,8 @@ func (cs *CatchpointCatchupService) processStageLastestBlockDownload() (err erro
 	if ledgerBlock, err := cs.ledger.Block(blockRound); err == nil {
 		blk = &ledgerBlock
 	}
+	var protoParams config.ConsensusParams
+	var ok bool
 
 	for {
 		attemptsCount++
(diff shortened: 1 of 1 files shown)


catchup/catchpointService.go
```diff
@@ -339,6 +339,8 @@ func (cs *CatchpointCatchupService) processStageLastestBlockDownload() (err erro
 	if ledgerBlock, err := cs.ledger.Block(blockRound); err == nil {
 		blk = &ledgerBlock
 	}
+	var protoParams config.ConsensusParams
+	var ok bool
```
algonautshant, line 344: nit: the variables can be declared inside the for loop for tighter scoping, since they are not needed outside.

algonautshant (approved):
Looks good.

catchup/catchpointService.go
tsachiherman: My intention here was to avoid having these variables get re-initialized on each iteration. ( it won't cause any issue if they would, though )

PR #2079, opened 2021-04-16
Fix proposal propagation

nicholasguoalgorand:
## Summary

When a relay receives a proposal for a future round, the proposal fails to be relayed. This change fixes that by sending it when the agreement advances to the corresponding round.

## Test Plan

Wrote unit test to simulate the scenario


agreement/player.go
```diff
@@ -391,9 +391,15 @@ func (p *player) enterRound(r routerHandle, source event, target round) []action
...
 		e := e.(payloadProcessedEvent)
-		msg := message{MessageHandle: 0, Tag: protocol.ProposalPayloadTag, UnauthenticatedProposal: e.UnauthenticatedPayload} // TODO do we want to keep around the original handle?
-		a := verifyPayloadAction(messageEvent{T: payloadPresent, Input: msg}, p.Round, e.Period, e.Pinned)
-		actions = append(actions, a)
+
+		up := e.UnauthenticatedPayload
+		uv := e.Vote.u()
+		msg := message{MessageHandle: 0, Tag: protocol.ProposalPayloadTag, UnauthenticatedProposal: up} // TODO do we want to keep around the original handle?
+
+		ra := relayAction(messageEvent{T: payloadPresent, Input: msg}, protocol.ProposalPayloadTag, compoundMessage{Proposal: up, Vote: uv})
```
tsachiherman, line 399: I'm pretty sure that on a relay action, you need to pass the message handle so it would prevent sending the proposal back to it's sender.

tsachiherman:
in this change, we're checking if we had any non-verified proposal that we've received in round X-1, when we enter round X. Then, we attempt to forward this proposal.

This is all nice, but it conflicting with the early sending of the proposal. If we have had sent it already, we wouldn't need to worry to send it once we start a new round ( clearly - this proposal isn't ours since we haven't generated our own proposal yet ).

tsachiherman:
another note:
I think it would be better to try and contain all the relaying logic inside `player.handleMessageEvent`. The reason for that is so that we can test the relaying logic by testing this method directly instead of doing a bigger-scope testing, which might have undesired complexities.

nicholasguoalgorand:
> in this change, we're checking if we had any non-verified proposal that we've received in round X-1, when we enter round X. Then, we attempt to forward this proposal.
> 
> This is all nice, but it conflicting with the early sending of the proposal. If we have had sent it already, we wouldn't need to worry to send it once we start a new round ( clearly - this proposal isn't ours since we haven't generated our own proposal yet ).

The reason why I structured the change like this is to mirror the original behavior before the early sending, ie relay the proposal right before we verify it. The other approach is likely fine, but might require a bit more thought.

agreement/player.go
```diff
@@ -17,6 +17,7 @@
 package agreement
 
 import (
+	"github.com/algorand/go-algorand/logging"
```
tsachiherman, line 20: nit: move this one down.

agreement/player_test.go
```diff
@@ -1802,6 +1802,118 @@ func TestPlayerRePropagatesFreshestBundle(t *testing.T) {
 	require.Truef(t, pM.getTrace().Contains(resynchEvent), "Player should try to repropagate freshest bundle")
 }
 
+func TestPlayerPropagatesProposalPayload(t *testing.T) {
```
algonautshant, line 1805: Can you please describe in comments what these tests are checking? 
Only one of the three fail without the fix.

agreement/state_machine_test.go
```diff
@@ -150,6 +150,15 @@ func (t ioTrace) ContainsFn(compareFn func(b event) bool) bool {
 	return false
 }
 
+func (t ioTrace) countAction() (count int) {
```
tsachiherman, line 153: I don't see where you use this method. did I miss anything ?

agreement/player_permutation_test.go
```diff
@@ -0,0 +1,748 @@
...
+	payload.Block = ve.Block()
+	payload.SeedProof = randomVRFProof()
+
+	return &proposal{unauthenticatedProposal: payload, ve: ve}
+}
+
+var payload = makeRandomProposalPayload(r)
+var pV = payload.value()
+
+var verifyError = makeSerErrStr("test error")
```
tsachiherman, line 48: `verifyError` -> `errTestVerifyFailed`

agreement/player_permutation_test.go
```diff
@@ -0,0 +1,748 @@
...
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/algorand/go-algorand/config"
+	"github.com/algorand/go-algorand/data/basics"
+	"github.com/algorand/go-algorand/protocol"
+)
+
+const r = round(209)
+const p = period(0)
```
tsachiherman, line 32: make r, p, payload and pV either non-global, or name them so it would be clear what they refer to, since they are accessible from other test files in the agreement package.

agreement/state_machine_test.go
nicholasguoalgorand: I use this to verify that the number of actions emitted by a player receiving a messageEvent lines up with what if should be, ie
```
			require.Equalf(t, trace.countAction(), 1, "Plyaer should not emit extra actions, player: %v, event: %v", playerN, eventN)
```

tsachiherman: sounds good, thanks.

tsachiherman (approved):

PR #2080, opened 2021-04-16
Add darwin builds to pipeline mule.yaml

bricerisingalgorand:
## Summary

Update Jenkins job config to support darwin builds

## Test Plan

I tested this locally and @algobarb tested it on an aws mac instance that is configured to do our darwin builds


algobarb (commented):
We are pivoting from using AWS to host the Darwin Builds.

PR #2083, opened 2021-04-19
reduce verbosity of logged messages

tsachiherman:
## Summary

This PR reduces the severity of two benign logged errors/warnings:
1. On the catchup service, the following warning was extremely common:
```
fetchAndWrite: block retrieval exceeded retry limit
```
This PR moves this to be a info messages rather than a warning message ( in most cases ), and extend it to include the round number.
2. On the AlgorandFullNode struct, we used to log a message like this:
```
AlgorandFullNode.loadParticipationKeys: cannot load db XXX.part: database is locked
```
This was caused by concurrent database access to the same part key database. This is a benign error : if we already updating the file, then we don't need to re-evaluate it's content.

## Test Plan

Logged level changes were tested manually.

### DIFF
--- a/catchup/service.go
+++ b/catchup/service.go
@@ -183,7 +183,18 @@ func (s *Service) fetchAndWrite(r basics.Round, prevFetchCompleteChan chan bool,
 
 		// Stop retrying after a while.
 		if i > catchupRetryLimit {
-			s.log.Errorf("fetchAndWrite: block retrieval exceeded retry limit")
+			if _, initialSync := s.IsSynchronizing(); initialSync {
+				// on the initial sync, it's completly expected that we won't be able to get all the "next" blocks. therefore info should suffice.
+				s.log.Infof("fetchAndWrite(%d): block retrieval exceeded retry limit", r)
+			} else {
+				// on any subsequent sync, we migth be looking for multiple rounds into the future, so it's completly reasonable that we would fail retrieving the future
+				// block. Generate a warning here only if we're failing to retrieve X+1 or below. all other retrievals should not generate a warning.
+				if r > s.ledger.NextRound() {
+					s.log.Infof("fetchAndWrite(%d): block retrieval exceeded retry limit", r)
+				} else {
+					s.log.Warnf("fetchAndWrite(%d): block retrieval exceeded retry limit", r)
+				}
+			}
 			return false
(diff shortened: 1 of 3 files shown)


catchup/service.go
```diff
@@ -183,7 +183,18 @@ func (s *Service) fetchAndWrite(r basics.Round, prevFetchCompleteChan chan bool,
 
 		// Stop retrying after a while.
 		if i > catchupRetryLimit {
-			s.log.Errorf("fetchAndWrite: block retrieval exceeded retry limit")
+			if _, initialSync := s.IsSynchronizing(); initialSync {
+				// on the initial sync, it's completly expected that we won't be able to get all the "next" blocks. therefore info should suffice.
+				s.log.Infof("fetchAndWrite(%d): block retrieval exceeded retry limit", r)
+			} else {
+				// on any subsequent sync, we migth be looking for multiple rounds into the future, so it's completly reasonable that we would fail retrieving the future
```
algonautshant, line 190: minor: the comment line is too long.

catchup/service.go
```diff
@@ -183,7 +183,18 @@ func (s *Service) fetchAndWrite(r basics.Round, prevFetchCompleteChan chan bool,
 
 		// Stop retrying after a while.
 		if i > catchupRetryLimit {
-			s.log.Errorf("fetchAndWrite: block retrieval exceeded retry limit")
+			if _, initialSync := s.IsSynchronizing(); initialSync {
```
algorandskiy, line 186: nit: format string here once.

node/node.go
```diff
@@ -734,6 +734,13 @@ func (node *AlgorandFullNode) loadParticipationKeys() error {
 		// Fetch a handle to this database
 		handle, err := node.getExistingPartHandle(filename)
 		if err != nil {
+			if err.Error() == "database is locked" {
```
algorandskiy, line 737: can we type asset here instead of message comparison?

algonautshant (dismissed):
Looks good.

node/node.go
tsachiherman: no, since the sqlite3 doesn't have a specific type for that error, but I have encapsulated it so it would be cleaner.

algonautshant (approved):

PR #2084, opened 2021-04-19
Print out intermediate decoding errors

algorandskiy:
## Summary

Better errors on txn and dryrun objects decoding.

## Test Plan

Use existing tests

### DIFF
--- a/cmd/tealdbg/local.go
+++ b/cmd/tealdbg/local.go
@@ -60,14 +60,14 @@ func txnGroupFromParams(dp *DebugParams) (txnGroup []transactions.SignedTxn, err
 
 	// 1. Attempt json - a single transaction
 	var txn transactions.SignedTxn
-	err = protocol.DecodeJSON(data, &txn)
+	err1 := protocol.DecodeJSON(data, &txn)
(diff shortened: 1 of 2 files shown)


cmd/tealdbg/local.go
```diff
@@ -128,6 +128,16 @@ func balanceRecordsFromParams(dp *DebugParams) (records []basics.BalanceRecord,
 		records = append(records, record)
 	}
 
+	// if conversion failed report all intermediate decoding errors
+	if err != nil {
+		if err1 != nil {
+			log.Printf("Decoding as JSON txn failed: %s", err1.Error())
```
tsachiherman, line 134: you're not decoding a txn here, you're decoding a balance record. 
i.e.
"Decoding blob as a JSON encoded single balance record failed : %s"

algorandskiy:
Ny bad, wrong function

cmd/tealdbg/local.go
algorandskiy: Fixed, thanks

tsachiherman (approved):
looks good to me.

PR #2086, opened 2021-04-21
ledger: avoid producing empty local delta

algorandskiy:
## Summary

Ensure local delta does not have entries with empty key-value stores

## Test Plan

New test added

### DIFF
--- a/config/consensus.go
+++ b/config/consensus.go
@@ -348,6 +348,9 @@ type ConsensusParams struct {
 
 	// update the initial rewards rate calculation to take the reward pool minimum balance into account
 	InitialRewardsRateCalculation bool
+
+	// NoEmptyLocalDeltas updates how ApplyDelta.EvalDelta.LocalDeltas are stored
+	NoEmptyLocalDeltas bool
 }
 
 // PaysetCommitType enumerates possible ways for the block header to commit to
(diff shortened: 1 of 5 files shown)


ledger/appcow.go
```diff
@@ -467,7 +467,12 @@ func (cb *roundCowState) BuildEvalDelta(aidx basics.AppIndex, txn *transactions.
 				if err != nil {
 					return basics.EvalDelta{}, err
 				}
-				evalDelta.LocalDeltas[addrOffset] = sdelta.kvCow.serialize()
+
+				d := sdelta.kvCow.serialize()
+				// ignore empty deltas
+				if len(d) != 0 {
```
tsachiherman, line 473: while it's probably correct, I think that you ignored the case where `len(d) == 0` and `len(evalDelta.LocalDeltas[addrOffset]) != 0`.
i.e. you would need to add a `delete(evalDelta.LocalDeltas, addrOffset)` here. I'm not sure how this would work in the previous release, and it could be a  non-issue here as well.

ledger/appcow_test.go
```diff
@@ -413,7 +413,7 @@ func TestCowBuildDelta(t *testing.T) {
 	a.Equal(
 		basics.EvalDelta{
 			GlobalDelta: basics.StateDelta{},
-			LocalDeltas: map[uint64]basics.StateDelta{0: {}},
+			LocalDeltas: map[uint64]basics.StateDelta{},
```
winder, line 416: Is there another test case needed for empty sender delta + non-empty other account deltas?

ledger/appcow.go
algorandskiy: 1. all txn have independent eval delta
2. there is only one kv per addr+appId combination
3. I'll add couple more tests to document this behavior

ledger/appcow_test.go
algorandskiy: Not really needed, but added this test to document this scenario

algorandskiy:
Added couple more tests per review

config/consensus.go
```diff
@@ -872,9 +875,25 @@ func initConsensusProtocols() {
...
+	v27.NoEmptyLocalDeltas = true
+
+	Consensus[protocol.ConsensusV27] = v27
+
+	// v26 or v25 can be upgraded to v27, with an update delay of 3 days
+	// 60279 = (3 * 24 * 60 * 60 / 4.3)
+	// for the sake of future manual calculations, we'll round that down
+	// a bit :
+	v26.ApprovedUpgrades[protocol.ConsensusV27] = 60000
+	v25.ApprovedUpgrades[protocol.ConsensusV27] = 60000
```
tsachiherman, line 892: for simplicity, let's limit this to v26 ? The v25 was on betanet only anyway.

ledger/appcow.go
```diff
@@ -467,7 +467,11 @@ func (cb *roundCowState) BuildEvalDelta(aidx basics.AppIndex, txn *transactions.
 				if err != nil {
 					return basics.EvalDelta{}, err
 				}
-				evalDelta.LocalDeltas[addrOffset] = sdelta.kvCow.serialize()
+
+				d := sdelta.kvCow.serialize()
+				if !cb.proto.NoEmptyLocalDeltas || len(d) != 0 {
+					evalDelta.LocalDeltas[addrOffset] = d
+				}
```
tsachiherman, line 474: this part need to be contingent upon the consensus parameter that you've added.

tsachiherman (changes requested):
only one comment. I don't think we need this line:
```go
v25.ApprovedUpgrades[protocol.ConsensusV27] = 60000
```

otherwise, I think that it looks good.

config/consensus.go
algorandskiy: Fixed

tsachiherman (approved):
looks good to me.

tsachiherman (approved):

PR #2087, opened 2021-04-21
Copy database files to nodes

shiqizng:
This PR enables database files generation in create_and_deploy_recipe.sh and upload them to S3, and copies these files to all nodes during algonet apply update  when available.

### DIFF
--- a/netdeploy/remote/nodecfg/nodeConfigurator.go
+++ b/netdeploy/remote/nodecfg/nodeConfigurator.go
@@ -33,12 +33,13 @@ import (
 )
 
 type nodeConfigurator struct {
-	config           remote.HostConfig
-	dnsName          string
-	genesisFile      string
-	genesisData      bookkeeping.Genesis
-	relayEndpoints   []srvEntry
-	metricsEndpoints []srvEntry
+	config                remote.HostConfig
+	dnsName               string
+	genesisFile           string
+	genesisData           bookkeeping.Genesis
+	bootstrappedDataFiles []string
+	relayEndpoints        []srvEntry
+	metricsEndpoints      []srvEntry
 }
 
 type srvEntry struct {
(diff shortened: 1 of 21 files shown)


scripts/create_and_deploy_recipe.sh
```diff
@@ -73,6 +74,9 @@ while [ "$1" != "" ]; do
             shift
             BUCKET="$1"
             ;;
+        --gendbfiles)
```
tsachiherman, line 77: I think that the gendbfiles functionality need to be implied.
i.e. if the recipe contains pre-generated block generation, then we should always use these.

If you want to add an option to override that, (i.e. disable that ), that's fine, of course.

netdeploy/remote/nodecfg/nodeConfigurator.go
```diff
@@ -134,6 +149,23 @@ func (nc *nodeConfigurator) prepareNodeDirs(configs []remote.NodeConfig, rootCon
 			return
 		}
 
+		// Copy the bootstrapped files into current ledger folder
+		if nc.bootstrappedBlockFile != "" {
+			fmt.Fprintf(os.Stdout, "... copying block database file to ledger folder ...\n")
+			_, err = util.CopyFile(nc.bootstrappedBlockFile, filepath.Join(nodeDest, genesisDir, fmt.Sprintf("%s.block.sqlite", config.LedgerFilenamePrefix)))
+			if err != nil {
+				return
```
tsachiherman, line 157: in case of an error, provide more detailed error - 
i.e. fmt.Errorf("failed to copy database file %s from %s to %s : %w", ... )

netdeploy/remote/nodecfg/nodeConfigurator.go
```diff
@@ -134,6 +149,23 @@ func (nc *nodeConfigurator) prepareNodeDirs(configs []remote.NodeConfig, rootCon
...
+			}
+		}
+
+		if nc.bootstrappedTrackerFile != "" {
```
tsachiherman, line 161: the `nc.bootstrappedBlockFile` and `nc.bootstrappedTrackerFile` should be deployed together, or not at all.
so, if you have only one of them, either fail right away, or issue a warning message and don't copy just one file.

shiqizng:
I need to close this PR and open an another one since this one is not updating with the latest commit.

PR #2088, opened 2021-04-21
catchup: add deep block validation support

tsachiherman:
## Summary

Add a config key `CatchupBlockValidateMode` that would control which types of validation the catchup would perform prior to accept a block.

The default validation remains the same as before. However, it allows the end-user to remove/add certain validation "toggles". This enable us to add a deep validation support, that perform a complete validation of a block.

## Test Plan

Tested manually, and against known-good and known-bad cases.


algorandskiy (dismissed):

config/config.go
```diff
@@ -388,6 +388,17 @@ type Local struct {
...
+
+	// CatchupBlockValidateMode defines the degree of validation used by the catchup service. This field is a bit-field, where the
+	// default enabled checks have their corresponding bit set as zero, and the default disabled checks have their corresponding bit clear.
+	// bit 0, default enable: verify the block certificate
+	// bit 1, default enable: verify payset committed hash in block header matches payset hash
+	// bit 2, default disable: verify the transaction signatures on the block are valid
+	// bit 3, default disable: verify that the hash of the recomputed payset matches the hash of the payset committed in the block header.
+	// not all permutation of the above bitset are currently functional. In particular, the two that are functional are:
+	// 0 : default behavior.
+	// 12 : perform all supported validation methods.
```
algonautshant, line 400: Isn't 12 the default position?
i.e. only CatchupVerifyCertificate and CatchupVerifyPaysetHash return true?

config/config.go
```diff
@@ -388,6 +388,17 @@ type Local struct {
...
 	EnableBlockServiceFallbackToArchiver bool `version[16]:"true"`
+
+	// CatchupBlockValidateMode defines the degree of validation used by the catchup service. This field is a bit-field, where the
+	// default enabled checks have their corresponding bit set as zero, and the default disabled checks have their corresponding bit clear.
+	// bit 0, default enable: verify the block certificate
+	// bit 1, default enable: verify payset committed hash in block header matches payset hash
+	// bit 2, default disable: verify the transaction signatures on the block are valid
+	// bit 3, default disable: verify that the hash of the recomputed payset matches the hash of the payset committed in the block header.
+	// not all permutation of the above bitset are currently functional. In particular, the two that are functional are:
+	// 0 : default behavior.
```
algonautshant, line 399: Isn't 0 for performing all supported validation methods?

config/config.go
```diff
@@ -388,6 +388,17 @@ type Local struct {
...
+	// 12 : perform all supported validation methods.
+	CatchupBlockValidateMode int `version[17]:"0"`
```
algonautshant, line 401: Shouldn't the default position be 12? 
If we wanted the behavior before this change?

config/config.go
```diff
@@ -388,6 +388,17 @@ type Local struct {
 	// BlockServiceCustomFallbackEndpoints is empty.
 	// The archiver is randomly selected, if none is available, will return StatusNotFound (404).
 	EnableBlockServiceFallbackToArchiver bool `version[16]:"true"`
+
+	// CatchupBlockValidateMode defines the degree of validation used by the catchup service. This field is a bit-field, where the
+	// default enabled checks have their corresponding bit set as zero, and the default disabled checks have their corresponding bit clear.
+	// bit 0, default enable: verify the block certificate
```
brianolson, line 394: I think it's confusing that some 1 bits turn on a feature and some 1 bits turn off a feature. If we go with the mspgack default-0 value, can we have the default be 'all checks on' and then any 1 bit turns off a check? Also I'm hoping that 'all checks on' is a good default and these things will mostly be turned off to make some dev and test operations faster.

config/config.go
algonautshant: I see the formulas are different for the upper bits. This is correct.

config/config.go
algonautshant: I see the formulas are different for the upper bits. This is correct.

config/config.go
algonautshant: I see the formulas are different for the upper bits. This is correct.

config/config.go
tsachiherman: it's not the msgp default here.. it's the json marshaler.
and no - all checks is *not* what we want by default. I think that the descriptions need to be further clarified.

config/config.go
```diff
@@ -388,6 +388,17 @@ type Local struct {
 	// BlockServiceCustomFallbackEndpoints is empty.
 	// The archiver is randomly selected, if none is available, will return StatusNotFound (404).
 	EnableBlockServiceFallbackToArchiver bool `version[16]:"true"`
+
+	// CatchupBlockValidateMode defines the degree of validation used by the catchup service. This field is a bit-field, where the
```
algonautshant, line 392: Suggested rewording of the comments:

CatchupBlockValidateMode is a development and testing configuration used by the catchup service. 
It can be used to omit certain validations to speed up the catchup process, or to apply extra validations which are redundant in normal operation.
This field is a bit-field with:
bit 0: (default 0) 0: verify the block certificate; 1: skip this validation
bit 1: (default 0) 0: verify payset committed hash in block header matches payset hash; 1: skip this validation
bit 2: (default 0) 0: don't verify the transaction signatures on the block are valid; 1: skip this redundant validation
bit 3: (default 0) 0: don't verify that the hash of the recomputed payset matches the hash of the payset committed in the block header; 1: skip this redundant validation
Note: not all permutations of the above bitset are currently functional. In particular, the ones that are functional are:
0 : default behavior.
3: speed up catchup by skipping necessary validations
12 : perform all validation methods (normal and additional)

PR #2089, opened 2021-04-22
This is a new recipe with 6 participating nodes and 30 relays in diff…

algonautshant:
## Summary
This is a new recipe with 6 participating nodes and 30 relays in different regions.

From the README.txt
This configuration creates a network of 6 participating nodes and 30
relays, such that the relays are distributed in 3 regions, and the
nodes are in a 4th region.

This scenario was designed to stress-test the situation where the
agreement proposal arrives for the next round.

See the captured round times [here](https://drive.google.com/file/d/1Xno-JwPG20LFBSRyan8RfdVY17nTKxA9/view?usp=sharing). Pingpong script starts around round 700 where the time increases from 4400 msec.

### DIFF
--- a/test/testdata/deployednettemplates/recipes/scenario4/README.txt
+++ b/test/testdata/deployednettemplates/recipes/scenario4/README.txt
@@ -3,4 +3,8 @@ relays, such that the relays are distributed in 3 regions, and the
 nodes are in a 4th region.
 
 This scenario was designed to stress-test the situation where the
-agreement votes arrive late from different rounds. 
+agreement proposal arrives for the next round.
+
+
+
+
--- a/test/testdata/deployednettemplates/recipes/scenario4/README.txt
+++ b/test/testdata/deployednettemplates/recipes/scenario4/README.txt
@@ -4,7 +4,3 @@ nodes are in a 4th region.
 
 This scenario was designed to stress-test the situation where the
 agreement proposal arrives for the next round.
-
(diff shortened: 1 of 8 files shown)


test/testdata/deployednettemplates/recipes/scenario4/README.txt
```diff
@@ -0,0 +1,6 @@
+This configuration creates a network of 6 participating nodes and 30
+relays, such that the relays are distributed in 3 regions, and the
+nodes are in a 4th region.
+
+This scenario was designed to stress-test the situation where the
+agreement votes arrive late from different rounds.
```
tsachiherman, line 6: `agreement votes arrive late from different rounds` -> `agreement proposal arrives for the next round.`

algorandskiy (dismissed):
should not we also generate and commit `topology.json` as for other scenarios?

algonautshant:
> should not we also generate and commit `topology.json` as for other scenarios?

Good idea. I added it.

tsachiherman:
I'm not against adding this one - but I don't think the name "scenario4" is the one I'd like to have.
The scenario 1-3 are defined as exponentially increased difficulty.. and scenario4 doesn't fit into that.

tsachiherman:
Given that this pr wasn't effective in reproing the issue, i'm going to close this pr. If needed, we can recreate it.

PR #2090, opened 2021-04-22
Prevent empty accounts stored in DB after opting out from an app

algorandskiy:
## Summary

Fix for empty accounts stored in DB after app clear state or app removal

## Test Plan

New tests added

### DIFF
--- a/ledger/appcow_test.go
+++ b/ledger/appcow_test.go
@@ -413,7 +413,7 @@ func TestCowBuildDelta(t *testing.T) {
 	a.Equal(
 		basics.EvalDelta{
 			GlobalDelta: basics.StateDelta{},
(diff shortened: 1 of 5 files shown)


ledger/appcow_test.go
```diff
@@ -418,6 +419,18 @@ func TestCowBuildDelta(t *testing.T) {
 		ed,
 	)
 
+	// check v27 behavior
+	cow.proto = config.Consensus[protocol.ConsensusCurrentVersion]
```
tsachiherman, line 423: if you want a specific behavior, you should call it by name; There is no reason to avoid using the protocol.V27

algorandskiy:
GH fails to update it after resolving conflicts, replaced by #2091

PR #2092, opened 2021-04-22
Fix initial balances in ledger tests

algorandskiy:
## Summary

Some tests compare data against old released behavior and increasing initial balance breaks hard-coded records

## Test Plan

This is a test

### DIFF
--- a/ledger/ledger_test.go
+++ b/ledger/ledger_test.go
@@ -66,7 +66,7 @@ func sign(secrets map[basics.Address]*crypto.SignatureSecrets, t transactions.Tr
 	}
 }
 
-func testGenerateInitState(tb testing.TB, proto protocol.ConsensusVersion) (genesisInitState InitState, initKeys map[basics.Address]*crypto.SignatureSecrets) {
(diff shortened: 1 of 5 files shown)


ledger/ledger_test.go
```diff
@@ -1579,6 +1581,12 @@ func TestListAssetsAndApplications(t *testing.T) {
 
 func TestLedgerMemoryLeak(t *testing.T) {
 	t.Skip() // for manual runs only
+	savedBaseAlgo := baseAlgoPerAccount
+	baseAlgoPerAccount = 10000000000
```
tsachiherman, line 1585: this might cause data races if we run two tests that access this variable concurrently.

tsachiherman: passing this parameter in is the "safe" approach.

algorandskiy: this test is for manual running only (see `t.Skip()`) so I think it is ok to use a global var for it. Passing as a param is safer, agree, but requires 22 modifications that are not really necessary in all these cases.

algorandskiy: done, moved to args.

tsachiherman (approved):

PR #2093, opened 2021-04-22
DB data migration for empty accounts

algorandskiy:
## Summary

Delete empty entries, update merkleTrie hashes.

## Test Plan

Rely on existing tests

### DIFF
--- a/ledger/accountdb.go
+++ b/ledger/accountdb.go
@@ -114,7 +114,7 @@ var accountsResetExprs = []string{
 // accountDBVersion is the database version that this binary would know how to support and how to upgrade to.
 // details about the content of each of the versions can be found in the upgrade functions upgradeDatabaseSchemaXXXX
 // and their descriptions.
-var accountDBVersion = int32(4)
+var accountDBVersion = int32(5)
 
 // persistedAccountData is used for representing a single account stored on the disk. In addition to the
 // basics.AccountData, it also stores complete referencing information used to maintain the base accounts
@@ -629,6 +629,45 @@ func accountsAddNormalizedBalance(tx *sql.Tx, proto config.ConsensusParams) erro
 	}
 
 	return rows.Err()
+}
+
+// removeEmptyAccountData removes empty AccountData msgp-encoded entries from accountbase table
+// and optionally returns list of addresses that were eliminated
+func removeEmptyAccountData(tx *sql.Tx, queryAddresses bool) (num int64, addresses []basics.Address, err error) {
+	if queryAddresses {
+		rows, err := tx.Query("SELECT address FROM accountbase where length(data) = 1 and data = x'80'") // empty AccountData is 0x80
+		if err != nil {
+			return 0, nil, err
+		}
+		defer rows.Close()
+
+		for rows.Next() {
+			var addrbuf []byte
+			err = rows.Scan(&addrbuf)
+			if err != nil {
+				return 0, nil, err
+			}
+			var addr basics.Address
+			if len(addrbuf) != len(addr) {
+				err = fmt.Errorf("Account DB address length mismatch: %d != %d", len(addrbuf), len(addr))
+				return 0, nil, err
+			}
+			copy(addr[:], addrbuf)
+			addresses = append(addresses, addr)
+		}
+	}
+
+	result, err := tx.Exec("DELETE from accountbase where length(data) = 1 and data = x'80'")
+	if err != nil {
+		return 0, nil, err
+	}
+	num, err = result.RowsAffected()
+	if err != nil {
+		// something wrong on getting rows count but data deleted, ignore the error
+		num = int64(len(addresses))
+		err = nil
+	}
+	return num, addresses, err
 }
 
 // accountDataToOnline returns the part of the AccountData that matters
--- a/ledger/acctupdates.go
+++ b/ledger/acctupdates.go
@@ -1232,6 +1232,12 @@ func (au *accountUpdates) accountsInitialize(ctx context.Context, tx *sql.Tx) (b
 					au.log.Warnf("accountsInitialize failed to upgrade accounts database (ledger.tracker.sqlite) from schema 3 : %v", err)
 					return 0, err
 				}
+			case 4:
+				dbVersion, err = au.upgradeDatabaseSchema4(ctx, tx)
+				if err != nil {
+					au.log.Warnf("accountsInitialize failed to upgrade accounts database (ledger.tracker.sqlite) from schema 4 : %v", err)
+					return 0, err
+				}
 			default:
 				return 0, fmt.Errorf("accountsInitialize unable to upgrade database from schema version %d", dbVersion)
 			}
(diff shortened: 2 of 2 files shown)


ledger/accountdb.go
```diff
@@ -631,6 +631,46 @@ func accountsAddNormalizedBalance(tx *sql.Tx, proto config.ConsensusParams) erro
...
+			err = rows.Scan(&addrbuf, &buf)
+			if err != nil {
+				return 0, nil, err
+			}
+			var addr basics.Address
+			if len(addrbuf) != len(addr) {
+				err = fmt.Errorf("Account DB address length mismatch: %d != %d", len(addrbuf), len(addr))
+				return 0, nil, err
+			}
+			copy(addr[:], addrbuf)
```
tsachiherman, line 656: Make sure to test buf content is 80h. Otherwise, the hash deletion might fail.

ledger/accountdb.go
```diff
@@ -631,6 +631,46 @@ func accountsAddNormalizedBalance(tx *sql.Tx, proto config.ConsensusParams) erro
...
+			addresses = append(addresses, addr)
+		}
+	}
+
+	result, err := tx.Exec("DELETE from accountbase where length(data) = 1")
```
tsachiherman, line 661: Limit this to deletion where the data is 80h

algorandskiy:
fixed both remarks by adding `and data = x'80'` into queries.

ledger/accountdb.go
```diff
@@ -631,6 +631,46 @@ func accountsAddNormalizedBalance(tx *sql.Tx, proto config.ConsensusParams) erro
 	return rows.Err()
 }
 
+// removeEmptyAccountData removes empty AccountData msgp-encoded entries from accountbase table
+// and optionally returns list of addresses that were eliminated
+func removeEmptyAccountData(tx *sql.Tx, queryAddresses bool) (num int64, addresses []basics.Address, err error) {
+	if queryAddresses {
+		rows, err := tx.Query("SELECT address, data FROM accountbase where length(data) = 1 and data = x'80'") // empty AccountData is 0x80
```
tsachiherman, line 638: you're not using the data argument. maybe eliminate it ?

ledger/acctupdates.go
```diff
@@ -1487,6 +1493,44 @@ func (au *accountUpdates) upgradeDatabaseSchema3(ctx context.Context, tx *sql.Tx
...
+	numDeleted, addresses, err := removeEmptyAccountData(tx, queryAddresses)
+	if err != nil {
+		return 0, err
+	}
+
+	if queryAddresses {
+		var totalHashesDeleted int
+		for _, addr := range addresses {
+			hash := accountHashBuilder(addr, basics.AccountData{}, []byte{0x80})
+			deleted, err := au.balancesTrie.Delete(hash)
```
tsachiherman, line 1510: Please do double-check me, but I believe that the `au.balancesTrie` here would always be nil. in order to address this, do the following:
```go
if queryAddresses {
    mc, err0 := MakeMerkleCommitter(tx, false)
    trie, err := merkletrie.MakeTrie(mc, TrieMemoryConfig)
    for _, addr := range addresses {
       ... 
       trie.Delete(hash)
    }
    au.Commit()
}
```

ledger/acctupdates.go
```diff
@@ -1487,6 +1493,44 @@ func (au *accountUpdates) upgradeDatabaseSchema3(ctx context.Context, tx *sql.Tx
 	return 4, nil
 }
 
+// upgradeDatabaseSchema4 does not change the schema but migrates data:
+// remove empty AccountData entries from accountbase table
+func (au *accountUpdates) upgradeDatabaseSchema4(ctx context.Context, tx *sql.Tx) (updatedDBVersion int32, err error) {
+
+	queryAddresses := au.balancesTrie != nil
```
tsachiherman, line 1500: check `au.catchpointInterval == 0` instead, since you can't use the `au.balancesTrie` at this point of initialization.

ledger/accountdb.go
algorandskiy: fixed

ledger/acctupdates.go
algorandskiy: Thank you, you are right, it is created right after migrations...

ledger/acctupdates.go
algorandskiy: done

tsachiherman (approved):
looks good. Tested against a data directory that contained an empty account data, the account data was removed, and the merkle trie root correctly re-aligned.

ledger/accountdb.go
```diff
@@ -631,6 +631,45 @@ func accountsAddNormalizedBalance(tx *sql.Tx, proto config.ConsensusParams) erro
...
+				return 0, nil, err
+			}
+			var addr basics.Address
+			if len(addrbuf) != len(addr) {
+				err = fmt.Errorf("Account DB address length mismatch: %d != %d", len(addrbuf), len(addr))
+				return 0, nil, err
+			}
+			copy(addr[:], addrbuf)
+			addresses = append(addresses, addr)
+		}
```
tolikzinovyev, line 657: I think we need to check for errors after reading rows.
https://golang.org/pkg/database/sql/#Rows.Next

ledger/acctupdates.go
```diff
@@ -1487,6 +1493,59 @@ func (au *accountUpdates) upgradeDatabaseSchema3(ctx context.Context, tx *sql.Tx
...
+			au.log.Errorf("upgradeDatabaseSchema4: failed to create merkle trie: %v", err)
+			goto done
+		}
+
+		var totalHashesDeleted int
+		for _, addr := range addresses {
+			hash := accountHashBuilder(addr, basics.AccountData{}, []byte{0x80})
+			deleted, err := trie.Delete(hash)
+			if err != nil {
+				au.log.Errorf("upgradeDatabaseSchema4: failed to delete hash '%s' from merkle trie for account %v: %v", hex.EncodeToString(hash), addr, err)
```
tolikzinovyev, line 1525: shouldn't we exit here?

ledger/acctupdates.go
```diff
@@ -1487,6 +1493,59 @@ func (au *accountUpdates) upgradeDatabaseSchema3(ctx context.Context, tx *sql.Tx
...
+			} else {
+				if !deleted {
+					au.log.Warnf("upgradeDatabaseSchema4: failed to delete hash '%s' from merkle trie for account %v", hex.EncodeToString(hash), addr)
+				} else {
+					totalHashesDeleted++
+				}
+			}
+		}
+		trie.Commit()
```
tolikzinovyev, line 1534: shouldn't we check the error?

ledger/acctupdates.go
```diff
@@ -1487,6 +1493,59 @@ func (au *accountUpdates) upgradeDatabaseSchema3(ctx context.Context, tx *sql.Tx
...
+		au.log.Infof("upgradeDatabaseSchema4: deleted %d hashes", totalHashesDeleted)
+	}
+
+done:
+	au.log.Infof("upgradeDatabaseSchema4: deleted %d rows", numDeleted)
+
+	// update version
+	_, err = db.SetUserVersion(ctx, tx, 5)
```
tolikzinovyev, line 1542: after jumping to `done` because of an error, `err` will be overwritten and we might exit without an error

tolikzinovyev (commented):
Came across this PR. I hope my comments are useful!

ledger/acctupdates.go
tsachiherman: In general, I would say "yes". however, for the deletion of keys from the merkle trie, we let it fail since we don't want the node to become non-functional due to a non-critical path code. ( i.e. it shouldn't happen. )

ledger/acctupdates.go
tsachiherman: yes, we should check for an error here.

ledger/acctupdates.go
tsachiherman: see my comment above regarding the merkle trie keys.

ledger/accountdb.go
tsachiherman: that is correct. After a `for rows.Next() {}` loop, we should test the `rows.Err()` to ensure the loop did not terminate prematurely due to an error.

PR #2099, opened 2021-04-24
Remove logfilters from Makefile

onetechnical:
These may be masking exit codes.

## Summary

Integration tests may be masking test failures after the integration of logfilter to the Makefile. Because it's not easy to add a pipefail to Makefiles, this pull request just removes them.

## Test Plan

See if test failures are recorded after submitting this change.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,6 +14,7 @@ SRCPATH     := $(shell pwd)
 ARCH        := $(shell ./scripts/archtype.sh)
 OS_TYPE     := $(shell ./scripts/ostype.sh)
 S3_RELEASE_BUCKET = $$S3_RELEASE_BUCKET
+.SHELLFLAGS := -euo pipefail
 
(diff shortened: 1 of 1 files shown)


tsachiherman:
just wondering if
```bash
(set -o pipefail; go test $(GOTAGS) -race $(UNIT_TEST_SOURCES) -timeout 3600s | logfilter)
```
wouldn't archive the same goal without all the verbosed output.

onetechnical:
> just wondering if
> 
> ```shell
> (set -eo pipefail; go test $(GOTAGS) -race $(UNIT_TEST_SOURCES) -timeout 3600s | logfilter)
> ```
> 
> wouldn't archive the same goal without all the verbosed output.

I wasn't sure if that would work in the Makefile to trap the error code...willing to try it, though!

onetechnical:
The approach in #2100 will work (explicit failure analysis in logfilter for error codes) rendering this unnecessary.

PR #2100, opened 2021-04-24
logfilter: exit with non-zero when test fails

tsachiherman:
## Summary

Exit from the logfilter process with non-zero output when "FAIL" was detected on input stream.
This allows us to pipe the input of the test without using `-o pipefail`.

## Test Plan

Unit test updated.


algorandskiy (approved):

onetechnical:
Excellent

PR #2102, opened 2021-04-26
Fix accessing int64 atomic variables on arm32

tsachiherman:
## Summary

Fix accessing int64 atomic variables on arm32, by aligning variables on 64-bit aligned address.

### DIFF
--- a/crypto/merklearray/worker.go
+++ b/crypto/merklearray/worker.go
@@ -25,6 +25,11 @@ import (
 // workerState describes a group of goroutines processing a sequential list
 // of maxidx elements starting from 0.
 type workerState struct {
+	// maxidx is the total number of elements to process, and nextidx
+	// is the next element that a worker should process.
+	maxidx  uint64
+	nextidx uint64
+
 	// nworkers is the number of workers that can be started.
 	// This field gets decremented once workers are launched,
 	// and represents the number of remaining workers that can
@@ -43,11 +48,6 @@ type workerState struct {
 	// wg tracks outstanding workers, to determine when all workers
 	// have finished their processing.
 	wg sync.WaitGroup
-
-	// maxidx is the total number of elements to process, and nextidx
-	// is the next element that a worker should process.
-	maxidx  uint64
-	nextidx uint64
 }
 
 func newWorkerState(max uint64) *workerState {


crypto/merklearray/worker.go
```diff
@@ -25,6 +25,11 @@ import (
 // workerState describes a group of goroutines processing a sequential list
 // of maxidx elements starting from 0.
 type workerState struct {
+	// maxidx is the total number of elements to process, and nextidx
```
algonautshant, line 28: Do these other cases also need a similar treatment? 

```
./ledger/acctupdates.go388:	return atomic.LoadInt32(&au.catchpointWriting) != 0
./ledger/bulletin.go42:	if atomic.CompareAndSwapUint32(&notifier.notified, 0, 1) {
./util/metrics/segment.go54:	concurrentCounter := atomic.AddUint32(&segment.concurrentCounter, uint32(1))
./util/metrics/segment.go84:	concurrentCounter := atomic.AddUint32(&segInstance.segment.concurrentCounter, ^uint32(0))
./network/wsNetwork.go1225:		if curPeersChangeCounter := atomic.LoadInt32(&wn.peersChangeCounter); curPeersChangeCounter != lastPeersChangeCounter {
./network/wsNetwork.go2053:	if atomic.AddInt32(&wn.throttledOutgoingConnections, int32(-1)) >= 0 {
./network/wsNetwork.go2228:		if atomic.CompareAndSwapInt32(&wn.ready, 0, 1) {
./network/wsNetwork_test.go158:	if atomic.LoadInt32(&mch.shouldWait) > 0 {
./network/wsNetwork_test.go752:	return atomic.LoadInt32(&peer.didInnerClose) != 0
./network/wsPeer.go366:	if atomic.LoadInt32(&wp.didInnerClose) == 0 {
./network/wsPeer.go610:		if atomic.LoadInt32(&wp.didInnerClose) == 0 {
./catchup/service.go129:	if atomic.CompareAndSwapUint32(&s.initialSyncNotified, 0, 1) {
./logging/telemetryOperation.go41:	if !atomic.CompareAndSwapInt32(&op.pending, 1, 0) {
```

algonautshant (approved):
Looks good. 
Perhaps it is another task separate from this, but I noticed some other cases where a similar treatment might be needed just in case.

crypto/merklearray/worker.go
tsachiherman: I'm pretty sure that the same was already applied to the rest of them. We generally make these change pretty quickly since the failure is very straight-forward.

PR #2105, opened 2021-04-26
go-algorand 2.5.6-beta

onetechnical:
GitHub Logo

**This release will require a protocol upgrade.**
This release contains a consensus protocol upgrade, which implements the following spec:
https://github.com/algorandfoundation/specs/tree/d050b3cade6d5c664df8bd729bf219f179812595

## Changes

1. General
    - Bug Fix - Fix accessing int64 atomic variables on arm32 which was causing a panic on startup on ARM32
1. Ledger
    - Bug Fixes
        - Ensure local delta does not have entries with empty key-value stores
        - Prevent storing empty accounts in DB after opting out from an app
        - DB data migration for empty accounts
        - compatibility mode for account index calculation in eval delta

## Additional Resources
* [Algorand Forum](https://forum.algorand.org)
* [Developer Documentation](https://developer.algorand.org)

## Test Plan

Standard release testing.

Verify consensus upgrade.


ledger/acctupdates.go
```diff
@@ -1324,6 +1330,63 @@ func (au *accountUpdates) upgradeDatabaseSchema3(ctx context.Context, tx *sql.Tx
...
+	numDeleted, addresses, err := removeEmptyAccountData(tx, queryAddresses)
+	if err != nil {
+		return 0, err
+	}
+
+	if queryAddresses && len(addresses) > 0 {
+		mc, err := makeMerkleCommitter(tx, false)
+		if err != nil {
+			// at this point record deleted and DB is pruned for account data
+			// if hash deletion fails just log it and do not about startup
```
egieseke, line 1347: about -> abort

ledger/appcow.go
```diff
@@ -457,17 +488,29 @@ func (cb *roundCowState) BuildEvalDelta(aidx basics.AppIndex, txn *transactions.
...
+					addrOffset = sdelta.accountIdx
+				} else {
+					addrOffset, err = txn.IndexByAddress(addr, txn.Sender)
+					if err != nil {
+						return basics.EvalDelta{}, err
+					}
+				}
+
+				d := sdelta.kvCow.serialize()
+				// noEmptyDeltas restricts prodicing empty local deltas in general
```
egieseke, line 508: prodicing -> producing

egieseke (approved):
Looks ok.

algobarb (commented):
Don't we need to update `config/version.go` before we merge ? I'm guessing this is another case in which we had to cherry-pick the commits?

tsachiherman:
> Don't we need to update `config/version.go` before we merge ? I'm guessing this is another case in which we had to cherry-pick the commits?

only for major/minor changes. not for patch/hotfix. that's where the buildnumber.dat comes into play.

tsachiherman (approved):

PR #2106, opened 2021-04-26
catchpoint: truncate output file

algorandskiy:
## Summary

Dump utility have not truncated output dump file so it might have some unexpected data.

## Test Plan

Manually tested


tsachiherman:
The change is good - but we have two more places in the `catchpointdump` utility that have the exact same issue. mind fixing these as well ?

algorandskiy:
Oh, I see, fixed.

tsachiherman (approved):
looks good.

PR #2109, opened 2021-04-27
More space efficient txnsync msg encoding

nicholasguoalgorand:
## Summary

Compress the TxnSync message before sending it and decompress it after receiving it. We can take advantage of the fact that some fields in transactions are duplicated as well as remove msgp tags by breaking arrays of txns into arrays of individual fields.

On a sample set of 20k txns on mainnet, this was able to encode the txnsync msg with ~37% less data.

## Test Plan

Wrote unit tests that encoded the txnsync message, then decoded it to make sure the data was still preserved. This was done on different transaction sets, some hard coded, some generated via reflection.


txnsync/txngroups.go
```diff
@@ -58,3 +227,522 @@ func decodeTransactionGroups(bytes []byte) (txnGroups []transactions.SignedTxGro
...
+			stub.Lsig = append(stub.Lsig, txn.Lsig)
+			stub.SignedTxns[i].Lsig = transactions.LogicSig{}
+		}
+		if !txn.AuthAddr.MsgIsZero() {
+			stub.BitmaskAuthAddr.SetBit(i)
+			stub.AuthAddr = append(stub.AuthAddr, txn.AuthAddr)
+			stub.SignedTxns[i].AuthAddr = basics.Address{}
+		}
+	}
+	stub.BitmaskAuthAddr = trimBitmask(stub.BitmaskAuthAddr)
```
tsachiherman, line 337: consider refactoring trimBitmask as a method of bit mask

protocol/txntype.go
```diff
@@ -47,3 +47,31 @@ const (
...
+	default:
+		return 7
+	}
+}
+
+// ByteToTxType converts a byte encoding to TxType
+func ByteToTxType(b byte) TxType {
+	txTypes := []TxType{PaymentTx, KeyRegistrationTx, AssetConfigTx, AssetTransferTx, AssetFreezeTx, ApplicationCallTx, CompactCertTx, UnknownTx}
+	return txTypes[b]
+}
```
tsachiherman, line 77: These would better be in txsync and not in protocol. 
consider defining an enum for these instead of using numeric values.

txnsync/txngroups.go
```diff
@@ -58,3 +227,522 @@ func decodeTransactionGroups(bytes []byte) (txnGroups []transactions.SignedTxGro
...
+	stub.BitmaskTxType = make(bitmask, bitmaskLen)
+	for i, txn := range stub.SignedTxns {
+		txTypeByte := protocol.TxTypeToByte(txn.Txn.Type)
+		if txTypeByte != 0 {
+			stub.BitmaskTxType.SetBit(i)
+			stub.TxType = append(stub.TxType, txTypeByte)
+		}
+		stub.SignedTxns[i].Txn.Type = ""
+	}
+	stub.BitmaskTxType = trimBitmask(stub.BitmaskTxType)
```
tsachiherman, line 355: That's interesting. So, you've picked a transaction type ( payment ), and optimized for that transaction type.
On the flip side, the rest of the transaction types are going to be stored quite inefficiently ( i.e. 8 bits/entry instead of 3 ).
Given that all transactions have a type, why don't we identify the most common transaction type, and optimize for that one ?

txnsync/txngroups.go
```diff
@@ -58,3 +227,522 @@ func decodeTransactionGroups(bytes []byte) (txnGroups []transactions.SignedTxGro
...
+	stub.BitmaskLastValid = make(bitmask, bitmaskLen)
+	stub.BitmaskNote = make(bitmask, bitmaskLen)
+	stub.BitmaskGenesisID = make(bitmask, bitmaskLen)
+	stub.BitmaskGenesisHash = make(bitmask, bitmaskLen)
+	stub.BitmaskLease = make(bitmask, bitmaskLen)
+	stub.BitmaskRekeyTo = make(bitmask, bitmaskLen)
+	for i, txn := range stub.SignedTxns {
+		if !txn.Txn.Sender.MsgIsZero() {
+			stub.BitmaskSender.SetBit(i)
+			stub.Sender = append(stub.Sender, txn.Txn.Sender)
```
tsachiherman, line 380: using a single byte array ([]byte) instead of array of arrays ([][]byte) would result in more efficient encoding. This is relatively easy to do for bytes array of fixed size ( such as `Address` = 32 bytes )

txnsync/txngroups.go
```diff
@@ -17,33 +17,202 @@
...
+
+// assumed to be in mode 0, sets bit at index to 1
+func (b bitmask) SetBit(index int) {
+	byteIndex := index/8 + 1
+	b[byteIndex] |= 1 << (index % 8)
+}
+
+func (b bitmask) EntryExists(index int, hint int) (bool, int) {
+	option := 0
+	if len(b) > 0 {
```
tsachiherman, line 170: if `len(b)` == 0, then we can exit right here. no need to further conditionals, right ?

txnsync/txngroups.go
```diff
@@ -17,33 +17,202 @@
...
+	}
+	switch option {
+	case 0: // if we have the bit 1 then we have an entry at the corresponding bit index.
+		byteIndex := index/8 + 1
+		return byteIndex < len(b) && (b[byteIndex]&(1<<(index%8)) != 0), 0
+	case 1: // if we have the bit 0 then we have an entry at the corresponding bit index.
+		byteIndex := index/8 + 1
+		return byteIndex >= len(b) || (b[byteIndex]&(1<<(index%8)) == 0), 0
+	case 2: // contains a list of bytes designating the transaction bit index
+		for hint < len(b) && index <= int(b[index]) {
```
tsachiherman, line 181: sounds like a binary search would be more suitable here. but for compactness sake, we should use differential list.

txnsync/txngroups.go
```diff
@@ -58,3 +227,522 @@ func decodeTransactionGroups(bytes []byte) (txnGroups []transactions.SignedTxGro
...
+	}
+
+	addGroupHashes(txnGroups)
+
+	return txnGroups, nil
+}
+
+func addGroupHashes(txnGroups []transactions.SignedTxGroup) {
+	for _, txns := range txnGroups {
+		if len(txns.Transactions) > 1 {
```
tsachiherman, line 294: this is abit more tricky : 
A transaction group of size 1 is allowed to have a transaction group field. so, if you were to add a bitmask specifically for that case, it would be great.

txnsync/txngroups.go
```diff
@@ -58,3 +227,522 @@ func decodeTransactionGroups(bytes []byte) (txnGroups []transactions.SignedTxGro
...
+	for i, txn := range stub.SignedTxns {
+		txTypeByte := protocol.TxTypeToByte(txn.Txn.Type)
+		if txTypeByte != 0 {
+			stub.BitmaskTxType.SetBit(i)
+			stub.TxType = append(stub.TxType, txTypeByte)
+		}
+		stub.SignedTxns[i].Txn.Type = ""
+	}
+	stub.BitmaskTxType = trimBitmask(stub.BitmaskTxType)
+	deconstructTxnHeader(stub)
```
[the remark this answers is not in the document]
nicholasguoalgorand: Do you mean I should call only one of these based on the type of txn?

txnsync/txngroups.go
nicholasguoalgorand: In practice calling this method over all the indices will iterate over the entire list just once.

Also, what is a differential list / why would that be helpful?

txnsync/txngroups.go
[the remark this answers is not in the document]
tsachiherman: Yes, I believe this is the case ( please correct me if I'm wrong here ).
The function `func (tx Transaction) WellFormed` verifies that.

txnsync/txngroups.go
tsachiherman: differential list goes as follows:
instead of storing a list of [50, 60, 70, 100], we'll store a list of [50, 10, 10, 30]. While this doesn't seems like a reduction is size, it would allow msgp to use less bytes for encoding these. ( since msgp spends less bytes on integer less than 127 ).
Another benefits is the increase the frequency of "small" numbers. This helps to improve the compression rate, if we were to apply a compression on top of that.

txnsync/txngroups_test.go
```diff
@@ -129,9 +128,11 @@ func TestTxnGroupEncodingSmall(t *testing.T) {
 	require.ElementsMatch(t, inTxnGroups, out)
 }
 
-func TestTxnGroupEncodingLarge(t *testing.T) {
-	dat, err := ioutil.ReadFile("/Users/nicholas/Downloads/txns.txt")
-	require.NoError(t, err)
+func txnGroupsData() ([]transactions.SignedTxGroup, error) {
+	dat, err := ioutil.ReadFile("txns.txt")
```
tsachiherman, line 132: I'm not opposing to create a test-data file that would be used for verification. But if you're creating that, why don't you place it in `./test/testdata/transactions`

txnsync/txngroups.go
tsachiherman: something about the commas above is skewed. do you mean 124,745,162 ?

brianolson: Yup. 124 million, miscounted/typeod places above. A bit more than 10x the number of Payment txns.

txnsync/txngroups.go
```diff
@@ -17,23 +17,417 @@
...
+	RekeyTo        []byte  `codec:"rekey,allocbound=maxAddressBytes"`
+	BitmaskRekeyTo bitmask `codec:"rekeybm"`
+}
+
+type encodedKeyregTxnFields struct {
+	_struct                 struct{}       `codec:",omitempty,omitemptyarray"`
+	VotePK                  []byte         `codec:"votekey,allocbound=maxAddressBytes"`
+	BitmaskVotePK           bitmask        `codec:"votekeybm"`
+	SelectionPK             []byte         `codec:"selkey,allocbound=maxAddressBytes"`
+	BitmaskSelectionPK      bitmask        `codec:"selkeybm"`
```
tsachiherman, line 115: The `SelectionPK`, `VotePK` and `KeyDilution` would always share the same bitmask.

txnsync/txngroups.go
```diff
@@ -17,23 +17,417 @@
...
+	BitmaskCloseRemainderTo bitmask `codec:"closebm"`
+}
+
+type encodedAssetConfigTxnFields struct {
+	_struct struct{} `codec:",omitempty,omitemptyarray"`
+
+	ConfigAsset        []basics.AssetIndex `codec:"caid,allocbound=maxEncodedTransactionGroup"`
+	BitmaskConfigAsset bitmask             `codec:"caidbm"`
+
+	AssetParams        []basics.AssetParams `codec:"apar,allocbound=maxEncodedTransactionGroup"`
```
tsachiherman, line 144: this one is going to be pretty inefficient - we should "flatten" this one onto the `encodedAssetConfigTxnFields` so that it won't include all the child keys over and over.

txnsync/txngroups.go
```diff
@@ -17,23 +17,417 @@
...
+//msgp:allocbound program config.MaxAppProgramLen
+type program []byte
+
+type encodedApplicationCallTxnFields struct {
+	_struct struct{} `codec:",omitempty,omitemptyarray"`
+
+	ApplicationID        []basics.AppIndex `codec:"apid,allocbound=maxEncodedTransactionGroup"`
+	BitmaskApplicationID bitmask           `codec:"apidbm"`
+
+	OnCompletion        []transactions.OnCompletion `codec:"apan,allocbound=maxEncodedTransactionGroup"`
```
tsachiherman, line 201: this field could really benefit from better packing. `transactions.OnCompletion` contains only few values, and we could easily fit 2 of them per a single byte.
Applications are going to be pretty common - so we'd better optimize the packer for that.

txnsync/txngroups.go
```diff
@@ -17,23 +17,417 @@
...
+}
+
+//msgp:allocbound applicationArgs transactions.EncodedMaxApplicationArgs
+type applicationArgs [][]byte
+
+//msgp:allocbound addresses transactions.EncodedMaxAccounts
+type addresses []basics.Address
+
+//msgp:allocbound appIndeces transactions.EncodedMaxForeignApps
+type appIndeces []basics.AppIndex
```
tsachiherman, line 187: it's `appIndices`

txnsync/txngroups.go
```diff
@@ -17,23 +17,417 @@
...
+	Accounts        []addresses `codec:"apat,allocbound=maxEncodedTransactionGroup"`
+	BitmaskAccounts bitmask     `codec:"apatbm"`
+
+	ForeignApps        []appIndeces `codec:"apfa,allocbound=maxEncodedTransactionGroup"`
+	BitmaskForeignApps bitmask      `codec:"apfabm"`
+
+	ForeignAssets        []assetIndeces `codec:"apas,allocbound=maxEncodedTransactionGroup"`
+	BitmaskForeignAssets bitmask        `codec:"apasbm"`
+
+	LocalStateSchema        []basics.StateSchema `codec:"apls,allocbound=maxEncodedTransactionGroup"`
```
tsachiherman, line 216: you'll need to break that into it's component to avoid the keys appear over and over in the encoded form.

txnsync/txngroups.go
```diff
@@ -17,23 +17,417 @@
...
+	case protocol.AssetTransferTx:
+		return assetTransferTx
+	case protocol.AssetFreezeTx:
+		return assetFreezeTx
+	case protocol.ApplicationCallTx:
+		return applicationCallTx
+	case protocol.CompactCertTx:
+		return compactCertTx
+	default:
+		return unknownTx
```
tsachiherman, line 271: There is no `unknownTx`. We shouldn't generate a value for that. instead, if we get it here, we should exit with an error.

txnsync/txngroups.go
```diff
@@ -60,6 +454,953 @@ func decodeTransactionGroups(bytes []byte) (txnGroups []transactions.SignedTxGro
...
+	stub.BitmaskTxType = make(bitmask, bitmaskLen)
+	stub.TxType = make([]byte, 0, int(stub.TotalTransactionsCount))
+	setupDeconstructTxnHeader(stub)
+	setupDeconstructKeyregTxnFields(stub)
+	setupDeconstructPaymentTxnFields(stub)
+	setupDeconstructAssetConfigTxnFields(stub)
+	setupDeconstructAssetTransferTxnFields(stub)
+	setupDeconstructAssetFreezeTxnFields(stub)
+	setupDeconstructApplicationCallTxnFields(stub)
+	setupDeconstructCompactCertTxnFields(stub)
```
tsachiherman, line 596: why not to switch on the type in order to make the right call ? It makes it cleaner, and reduce the extra conditional within each of these methods.

txnsync/txngroups.go
```diff
@@ -60,6 +454,953 @@ func decodeTransactionGroups(bytes []byte) (txnGroups []transactions.SignedTxGro
...
+	}
+	if !txn.Txn.FirstValid.MsgIsZero() {
+		stub.BitmaskFirstValid.SetBit(i)
+		stub.FirstValid = append(stub.FirstValid, txn.Txn.FirstValid)
+	}
+	if !txn.Txn.LastValid.MsgIsZero() {
+		stub.BitmaskLastValid.SetBit(i)
+		stub.LastValid = append(stub.LastValid, txn.Txn.LastValid)
+	}
+	if txn.Txn.Note != nil && len(txn.Txn.Note) > 0 {
```
tsachiherman, line 666: having `len(txn.Txn.Note) > 0` already makes the `txn.Txn.Note != nil` not needed since len(nil) = 0

txnsync/txngroups.go
```diff
@@ -60,6 +454,953 @@ func decodeTransactionGroups(bytes []byte) (txnGroups []transactions.SignedTxGro
...
+			stub.BitmaskVoteLast.SetBit(i)
+			stub.VoteLast = append(stub.VoteLast, txn.Txn.VoteLast)
+		}
+		if txn.Txn.VoteKeyDilution > 0 {
+			stub.BitmaskVoteKeyDilution.SetBit(i)
+			stub.VoteKeyDilution = append(stub.VoteKeyDilution, txn.Txn.VoteKeyDilution)
+		}
+		if txn.Txn.Nonparticipation {
+			stub.BitmaskNonparticipation.SetBit(i)
+			stub.Nonparticipation = append(stub.Nonparticipation, txn.Txn.Nonparticipation) // can probably get rid of this
```
tsachiherman, line 738: given that it's a boolean the above bitset should be enough.

tsachiherman:
Regarding the performance issue you've mentioned : the reason it's so slow is due to the following pattern:
```golang
stub.AssetAmount = append(stub.AssetAmount, txn.Txn.AssetAmount)
```

Since you haven't pre-allocated this array, it would allocate a new one and copy the old elements into the new one.
This is a n^2 efficiency, which is what you're seeing. preallocating this on the first time you're hitting this field would eliminate the issue.

txnsync/txngroups.go
nicholasguoalgorand: yes, I put a comment reminding myself to get rid of that

nicholasguoalgorand:
> Regarding the performance issue you've mentioned : the reason it's so slow is due to the following pattern:
> 
> ```go
> stub.AssetAmount = append(stub.AssetAmount, txn.Txn.AssetAmount)
> ```
> 
> Since you haven't pre-allocated this array, it would allocate a new one and copy the old elements into the new one.
> This is a n^2 efficiency, which is what you're seeing. preallocating this on the first time you're hitting this field would eliminate the issue.

I agree that these are good changes to make. However, these weren't causing the performance issues I observed. I have already dealt with them and most of the computation time is now in the msgp encoding part

txnsync/txngroups.go
```diff
@@ -17,23 +17,473 @@
...
+}
+
+func getSlice(b []byte, index int, size int) ([]byte, error) {
+	if index*size+size > len(b) {
+		return nil, errDataMissing
+	}
+	return b[index*size : index*size+size], nil
+}
+
+func getHalfByte(b []byte, index int) (byte, error) {
```
tsachiherman, line 464: nit: half byte -> nibble

tsachiherman:
The `setupDeconstructXXX` path is following the patten of preallocating all the arrays/bitmasks and then filling them as needed. That would create a heavy load on the GC, and not for a good reason..
If you could dynamically allocate these structures, it would be much better.

txnsync/txngroups.go
```diff
@@ -17,23 +17,524 @@
...
+	"github.com/algorand/go-algorand/logging"
 	"github.com/algorand/go-algorand/protocol"
 )
 
-const maxEncodedTransactionGroup = 10000
-const maxEncodedTransactionGroupEntries = 10000
+const maxEncodedTransactionGroup = 30000
+const maxEncodedTransactionGroupEntries = 30000
+const maxBitmaskSize = (maxEncodedTransactionGroupEntries+7)/8 + 1
+const signatureSize = 64
```
tsachiherman, line 34: Avoid hardcoding this value. use `len(crypto.Signature)` instead.

txnsync/txngroups.go
```diff
@@ -17,23 +17,524 @@
...
+const maxSignatureBytes = maxEncodedTransactionGroupEntries * signatureSize
+const addressSize = 32
```
tsachiherman, line 36: same as above. use `crypto.DigestSize`

txnsync/txngroups.go
```diff
@@ -17,23 +17,524 @@
...
+const maxAddressBytes = maxEncodedTransactionGroupEntries * addressSize
+
+var errDataMissing = fmt.Errorf("failed to decode: data missing")
```
tsachiherman, line 39: nit: avoid using `fmt.Errorf` unless you're formatting an error. For a fixed-string error, use `errors.New`

txnsync/txngroups.go
```diff
@@ -17,23 +17,524 @@
...
+	case protocol.AssetTransferTx:
+		return assetTransferTx
+	case protocol.AssetFreezeTx:
+		return assetFreezeTx
+	case protocol.ApplicationCallTx:
+		return applicationCallTx
+	case protocol.CompactCertTx:
+		return compactCertTx
+	default:
+		logging.Base().Errorf("invalid txtype") // TODO: (nguo) perform proper error handling here instead
```
tsachiherman, line 361: please do..

tsachiherman (changes requested):
Reviewed the code, some issues I've came across:

1. Leaving the SignedTxns and TxnGroups inside txGroupsEncodingStub makes it really to hard to figure out if it's doing the right thing.
I think that these shouldn't be part of the structure, and these should be passed on the stack instead.
2. The TxnGroups field itself is only being used by encodeTransactionGroupsOld and decodeTransactionGroupsOld, which aren't being used.
3. Bitmask type 3 is not being used or tested anywhere. Please add unit test to the bitmask explicitly which tests all the code paths there.
4. When allocating dynamically fields, you're using `if stub.BitmaskVersion == nil ..`, which isn't incorrect, but might be more error prone. It would be better to test `if len(stub.BitmaskVersion)==0 …` instead.
5. Use operator pointer methods when applicable. i.e. use `func (stub *txGroupsEncodingStub) deconstructCompactCertTxnFields(i int, txn transactions.SignedTxn) ` instead of `func deconstructCompactCertTxnFields(stub *txGroupsEncodingStub, i int, txn transactions.SignedTxn)`
6. None of the unit tests cover the case where GenesisID != "". We should definitely cover that.
7. Neither GenesisID nor GenesisHash should be encoded into encodedTxnHeaders. These are global network parameters, and can ( and should! ) be retrieved from the ledger.

These above would require another pass through, so I'll make another review after these are done. I think that the txngroups.go file is too large. I would suggest breaking this file into 3:
- encodedgroupstypes.go
- encodedgroupsmarshalers.go
- encodedgroupsunmarshalers.go

txnsync/bitmask.go
```diff
@@ -0,0 +1,161 @@
...
+			if i != 0 {
+				(*b)[i] = 255 - (*b)[i] // invert bits
+			}
+		}
+	}
+}
+
+func bytesNeededBitmask(elements int) int {
+	return (elements+7)/8 + 1
+}
```
tsachiherman, line 161: Add EOL

txnsync/mainloop.go
```diff
@@ -18,6 +18,7 @@ package txnsync
 
 import (
 	"context"
+	"github.com/algorand/go-algorand/crypto"
```
tsachiherman, line 21: move this import into the group below.

txnsync/service.go
```diff
@@ -18,6 +18,7 @@ package txnsync
 
 import (
 	"context"
+	"github.com/algorand/go-algorand/crypto"
```
tsachiherman, line 21: move import line into the next group.

txnsync/txngroups.go
```diff
@@ -17,24 +17,24 @@
 package txnsync
 
 import (
+	"fmt"
+	"github.com/algorand/go-algorand/crypto"
```
tsachiherman, line 21: move import line into the next group.

txnsync/txngroups.go
```diff
@@ -60,10 +99,46 @@ func decodeTransactionGroups(bytes []byte) (txnGroups []transactions.SignedTxGro
...
+}
+
 func releaseEncodedTransactionGroups(buffer []byte) {
 	if buffer == nil {
 		return
 	}
 
 	protocol.PutEncodingBuf(buffer[:0])
-}
+}
```
tsachiherman, line 144: add EOL

txnsync/txngroups.go
```diff
@@ -60,10 +99,46 @@ func decodeTransactionGroups(bytes []byte) (txnGroups []transactions.SignedTxGro
...
+	if len(bytes) == 0 {
+		return nil, nil
+	}
+	var stub txGroupsEncodingStub
+	_, err = stub.UnmarshalMsg(bytes)
+	if err != nil {
+		return nil, err
+	}
+
+	stx := make([]transactions.SignedTxn, stub.TotalTransactionsCount)
```
tsachiherman, line 112: add a test here to ensure `stub.TotalTransactionsCount` isn't too large. We don't want someone to send us 10G for this field, and we will crash as a result.

txnsync/txngroups.go
```diff
@@ -60,10 +99,46 @@ func decodeTransactionGroups(bytes []byte) (txnGroups []transactions.SignedTxGro
...
+
+	err = stub.reconstructSignedTransactions(stx, genesisID, genesisHash)
+	if err != nil {
+		return nil, err
+	}
+
+	txnGroups = make([]transactions.SignedTxGroup, stub.TransactionGroupCount)
```
tsachiherman, line 119: same as above; validate `stub.TransactionGroupCount`

tsachiherman (changes requested):
TestTxnGroupEncodingLarge is failing due to missing `txns1.txt` file. Did you intend to include it with this PR ?

txnsync/txngroups.go
```diff
@@ -44,11 +44,50 @@ func encodeTransactionGroups(inTxnGroups []transactions.SignedTxGroup) []byte {
...
+				index++
+			}
+		}
+	}
+	stub.finishDeconstructSignedTransactions()
+
+	return stub.MarshalMsg(protocol.GetEncodingBuf()[:0]), nil
+}
+
+func decodeTransactionGroupsOld(bytes []byte) (txnGroups []transactions.SignedTxGroup, err error) {
```
tsachiherman, line 86: decodeTransactionGroupsOld along with txGroupsEncodingStubOld seems to be redundant.

txnsync/encodedgroupsmarshalers.go
```diff
@@ -0,0 +1,826 @@
...
+import (
+	"fmt"
+
+	"github.com/algorand/go-algorand/crypto"
+	"github.com/algorand/go-algorand/data/basics"
+	"github.com/algorand/go-algorand/data/transactions"
+	"github.com/algorand/go-algorand/protocol"
+)
+
+func squeezeByteArray(b []byte) []byte {
```
tsachiherman, line 28: maybe `squeezeByteArray` -> `compactNibblesArray`, since each element of the array is a nibble ?

nicholasguoalgorand:
> TestTxnGroupEncodingLarge is failing due to missing `txns1.txt` file. Did you intend to include it with this PR ?

I don't know if I should include it or remove the benchmarks / large test. What do you think?

txnsync/txngroups.go
nicholasguoalgorand: I left this in the code to compare performance, should I go ahead and delete all of these?

txnsync/encodedgroupsmarshalers.go
```diff
@@ -0,0 +1,826 @@
...
+		var txGroup transactions.TxGroup
+		txGroup.TxGroupHashes = txGroupHashes[index : index+len(txns.Transactions)]
+		for i, tx := range txns.Transactions {
+			txGroup.TxGroupHashes[i] = crypto.HashObj(tx.Txn)
+		}
+		groupHash := crypto.HashObj(txGroup)
+		for i := range txns.Transactions {
+			if exists := b.EntryExists(index, txnCount); exists {
+				txns.Transactions[i].Txn.Group = groupHash
+			}
```
tsachiherman, line 51: The predicate here is too loose. The group hash need to be added if one of the following applies
1. this transaction group contains more than 1 transaction
2. the bitmask is set.

txnsync/encodedgroupsmarshalers.go
```diff
@@ -0,0 +1,826 @@
...
+	}
+	return b[0 : len(b)/2]
+}
+
+func addGroupHashes(txnGroups []transactions.SignedTxGroup, txnCount int, b bitmask) {
+	index := 0
+	txGroupHashes := make([]crypto.Digest, txnCount)
+	for _, txns := range txnGroups {
+		var txGroup transactions.TxGroup
+		txGroup.TxGroupHashes = txGroupHashes[index : index+len(txns.Transactions)]
```
tsachiherman, line 43: that's pretty inefficient. preallocate 16 entries, and sub-slice them here.
i.e. 
```golang
txGroupHashes := make([]crypto.Digest, 16)
...
txGroup.TxGroupHashes = txGroupHashes[: len(txns.Transactions)]
```

txnsync/encodedgroupsmarshalers.go
```diff
@@ -0,0 +1,826 @@
...
+func (stub *txGroupsEncodingStub) finishDeconstructLsigs() {
+	stub.BitmaskLogic.trimBitmask(int(stub.TotalTransactionsCount))
+	stub.BitmaskLogicArgs.trimBitmask(int(stub.TotalTransactionsCount))
+}
+
+func (stub *txGroupsEncodingStub) deconstructTransactions(i int, txn transactions.SignedTxn) error {
+	bitmaskLen := bytesNeededBitmask(int(stub.TotalTransactionsCount))
+	txTypeByte, err := TxTypeToByte(txn.Txn.Type)
+	if err != nil {
+		return fmt.Errorf("failed to deconstructTransactions: %v", err)
```
tsachiherman, line 184: `%v` -> `%w` so that we can detect the child error using Unwrap.

txnsync/encodedgroupsmarshalers.go
```diff
@@ -0,0 +1,826 @@
...
+	}
+	if txTypeByte != 0 {
```
tsachiherman, line 186: two things - 
1. avoid comparing to magic numbers.
2. 0 seems to be a payment. How come it didn't break any of the unit testing ?

txnsync/encodedgroupsmarshalers.go
```diff
@@ -0,0 +1,826 @@
...
+				}
+				index++
+			} else {
+				stub.BitmaskTxType.SetBit(i)
+				newTxTypes = append(newTxTypes, offset)
+			}
+		}
+		stub.TxType = newTxTypes
+		stub.TxTypeOffset = offset
+	}
```
tsachiherman, line 243: extract this section into its own method.

txnsync/encodedgroupsmarshalers.go
nicholasguoalgorand: So the way I did it was to first decompose it such that payment txns were omitted via the bitmask. Later I checked if a different txtype were more common, and if so, I recomputed bitmask / txtype field to fit the other txtype. So the result produced is still correct.

But now that you mention it, I can make this code more clean by not assuming everything is a payment txn at the beginning.

txnsync/txngroups.go
tsachiherman: I think that it's good to leave it as a reference; in that case:
1. document it, so we'll know that it's not being used.
2. move this method to the XXX_test.go file.

tsachiherman:
> > TestTxnGroupEncodingLarge is failing due to missing `txns1.txt` file. Did you intend to include it with this PR ?
> 
> I don't know if I should include it or remove the benchmarks / large test. What do you think?

Unless the file is exceptionally large, I would add it. If it's more than 500Kb, just gzip it and add a code for the benchmark to dynamically decompress it. I think that reference test cases are extremely important.

txnsync/encodedgroupstypes.go
```diff
@@ -0,0 +1,357 @@
...
+	BitmaskVersion   bitmask `codec:"msigvbm"`
+	Threshold        []byte  `codec:"msigthr,allocbound=maxEncodedTransactionGroup"`
+	BitmaskThreshold bitmask `codec:"msigthrbm"`
+	// splitting subsigs further make the code much more complicated / does not give gains
+	Subsigs        [][]crypto.MultisigSubsig `codec:"subsig,allocbound=maxEncodedTransactionGroup,allocbound=crypto.MaxMultisig"`
+	BitmaskSubsigs bitmask                   `codec:"subsigsbm"`
+}
+
+type encodedLsigs struct {
+	_struct          struct{}   `codec:",omitempty,omitemptyarray"`
```
tsachiherman, line 74: nit: add line separator after `_struct`

txnsync/encodedgroupstypes.go
```diff
@@ -0,0 +1,357 @@
...
+
+	Sender             []byte              `codec:"snd,allocbound=maxAddressBytes"`
+	BitmaskSender      bitmask             `codec:"sndbm"`
+	Fee                []basics.MicroAlgos `codec:"fee,allocbound=maxEncodedTransactionGroup"`
+	BitmaskFee         bitmask             `codec:"feebm"`
+	FirstValid         []basics.Round      `codec:"fv,allocbound=maxEncodedTransactionGroup"`
+	BitmaskFirstValid  bitmask             `codec:"fvbm"`
+	LastValid          []basics.Round      `codec:"lv,allocbound=maxEncodedTransactionGroup"`
+	BitmaskLastValid   bitmask             `codec:"lvbm"`
+	Note               [][]byte            `codec:"note,allocbound=maxEncodedTransactionGroup"` // TODO whats the correct allocbound?
```
tsachiherman, line 109: The allocbound definition should be 
 `codec:"note,allocbound=maxEncodedTransactionGroup,allocbound=config.MaxTxnNoteBytes"`

txnsync/encodedgroupstypes.go
```diff
@@ -0,0 +1,357 @@
...
+
+	Lease        []byte  `codec:"lx,allocbound=maxAddressBytes"`
+	BitmaskLease bitmask `codec:"lxbm"`
+
+	RekeyTo        []byte  `codec:"rekey,allocbound=maxAddressBytes"`
+	BitmaskRekeyTo bitmask `codec:"rekeybm"`
+}
+
+type encodedKeyregTxnFields struct {
+	_struct                 struct{}       `codec:",omitempty,omitemptyarray"`
```
tsachiherman, line 124: nit: add EOL after this line

txnsync/encodedgroupsunmarshalers.go
```diff
@@ -0,0 +1,795 @@
...
+		return b[index/2] / 16, nil
+	}
+	return b[index/2] % 16, nil
+}
+
+func (stub *txGroupsEncodingStub) reconstructSignedTransactions(signedTxns []transactions.SignedTxn, genesisID string, genesisHash crypto.Digest) error {
+	var index int
+	index = 0
+	for i := range signedTxns {
+		if exists := stub.BitmaskSig.EntryExists(i, int(stub.TotalTransactionsCount)); exists {
```
tsachiherman, line 50: looks as if you could have reduced this to 
```golang
if stub.BitmaskSig.EntryExists(i, int(stub.TotalTransactionsCount)) {
```
and make it (slightly) simpler.

tsachiherman: Another thing - I see that in all the calls to `EntryExists` you're casting the second parameter to int. It might be a good indicator that either this parameter should have been an int, or that the method parameter should have been different.

nicholasguoalgorand: The main problem is that I'd like `stub.TotalTransactionsCount` to be an int and msgp doesn't let me

nicholasguoalgorand: The method parameter doesn't make much sense to be a uint64

txnsync/encodedgroupsunmarshalers.go
```diff
@@ -0,0 +1,795 @@
...
+	for i := range signedTxns {
+		if exists := stub.BitmaskAuthAddr.EntryExists(i, int(stub.TotalTransactionsCount)); exists {
+			slice, err := getSlice(stub.AuthAddr, index, crypto.DigestSize)
+			if err != nil {
+				return err
+			}
+			copy(signedTxns[i].AuthAddr[:], slice)
+			index++
+		}
+	}
```
tsachiherman: I was reviewing the executing of this section - in some cases it would be extremely inefficient:
let's say you have a 500 signed transactions. And there is only a single transaction with AuthAddr.
In that case, you would perform this loop 500 times, calling EntryExists over and over.

The better implementation for this would be to use the bitmask to figure out what the indices are, and then go back and fill up the data accordingly. This shouldn't be too difficult to implement -
have the bitmap expose a Iterate method that makes a callback to an anonymous function.
i.e.
```golang
err := stub.BitmaskAuthAddr.Iterate(stub.TotalTransactionsCount, func (i int) error {
     slice, err := getSlice(stub.AuthAddr, index, crypto.DigestSize)
     if err != nil {
	return err
     }
     copy(signedTxns[i].AuthAddr[:], slice)
     index++
})
if err != nil {
    return err
}
```

txnsync/txngroups.go
nicholasguoalgorand: Okay, I moved the decode / encode to the test file but left the data structures since msgp wouldnt compile them and left a comment

txnsync/bitmask.go
```diff
@@ -0,0 +1,226 @@
...
+	switch option {
+	case 0:
+		for i := 0; i < entries; i++ {
+			byteIndex := i/8 + 1
+			if byteIndex < len(*b) && ((*b)[byteIndex]&(1<<(i%8)) != 0) {
+				if err := callback(i); err != nil {
+					return err
+				}
+			}
+		}
```
tsachiherman, line 178: hmm. I don't think you want to iterate on entries, as you'd have the same inefficiency for which you implemented this method to avoid. instead,
```golang
for i, v := range b[1:] {
   for j := 0; j < 8 && v > 0; j++, v>>=1 {
       if v & 1 != 0 {
            if err := callback(8*i+j); err != nil {
                 return err
            }
       }
   }
}
```

txnsync/encodedgroupsmarshalers.go
```diff
@@ -0,0 +1,818 @@
...
+func (stub *txGroupsEncodingStub) finishDeconstructTransactions() {
+	stub.finishDeconstructTxType()
+	stub.finishDeconstructTxnHeader()
+	stub.finishDeconstructKeyregTxnFields()
+	stub.finishDeconstructPaymentTxnFields()
+	stub.finishDeconstructAssetConfigTxnFields()
+	stub.finishDeconstructAssetTransferTxnFields()
+	stub.finishDeconstructAssetFreezeTxnFields()
+	stub.finishDeconstructApplicationCallTxnFields()
+	stub.finishDeconstructCompactCertTxnFields()
```
tsachiherman, line 220: shouldn't you use the `stub.TxType` here to determine which one of these would be called ?

txnsync/encodedgroupsmarshalers.go
```diff
@@ -0,0 +1,818 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to deconstructTransactions: %w", err)
+	}
+	if len(stub.BitmaskTxType) == 0 {
+		stub.BitmaskTxType = make(bitmask, bitmaskLen)
+		stub.TxType = make([]byte, 0, int(stub.TotalTransactionsCount))
+	}
+	stub.TxType = append(stub.TxType, txTypeByte)
+	stub.deconstructTxnHeader(i, txn)
+	switch txn.Txn.Type {
```
tsachiherman, line 192: use the `txTypeByte` here instead of the `txn.Txn.Type`, since the `txTypeByte` is known to be "validated".
I know that at this point, the `txn.Txn.Type` would also be valid here, but it makes it more clear since this variable is always in "valid" state ( or none ).

txnsync/encodedgroupsmarshalers.go
```diff
@@ -0,0 +1,818 @@
...
+	stub.finishDeconstructAssetConfigTxnFields()
+	stub.finishDeconstructAssetTransferTxnFields()
+	stub.finishDeconstructAssetFreezeTxnFields()
+	stub.finishDeconstructApplicationCallTxnFields()
+	stub.finishDeconstructCompactCertTxnFields()
+}
+
+func (stub *txGroupsEncodingStub) finishDeconstructTxType() {
+	offset := byte(0)
+	count := make(map[int]int)
```
tsachiherman, line 225: given that there are exactly 8 transaction types, you'd do *much* better allocating a `[8]int` instead of using a map here.

txnsync/encodedgroupsmarshalers.go
```diff
@@ -0,0 +1,818 @@
...
+			stub.BitmaskGenesisID = make(bitmask, bitmaskLen)
+		}
+		stub.BitmaskGenesisID.SetBit(i)
+	}
+	if !txn.Txn.GenesisHash.MsgIsZero() {
+		if len(stub.BitmaskGenesisHash) == 0 {
+			stub.BitmaskGenesisHash = make(bitmask, bitmaskLen)
+		}
+		stub.BitmaskGenesisHash.SetBit(i)
+	}
```
tsachiherman, line 302: The genesis hash is not optional. There is no need for `BitmaskGenesisHash` since it's always presents.

txnsync/encodedgroupsmarshalers.go
```diff
@@ -0,0 +1,818 @@
...
+	if !txn.Txn.Group.MsgIsZero() {
+		if len(stub.BitmaskGroup) == 0 {
+			stub.BitmaskGroup = make(bitmask, bitmaskLen)
+		}
+		stub.BitmaskGroup.SetBit(i)
```
tsachiherman, line 307: I think that we should set the bit only on transactions group of size 1. ( setting this bit on transaction groups > 1 is redundant, since we can extrapolate that information from the size of the transaction group ).

txnsync/encodedgroupstypes.go
```diff
@@ -0,0 +1,369 @@
...
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with go-algorand.  If not, see <https://www.gnu.org/licenses/>.
+
+package txnsync
+
+import (
+	"errors"
+	"github.com/algorand/go-algorand/data/transactions"
```
tsachiherman, line 21: move this line to the group below.

txnsync/encodedgroupsunmarshalers.go
```diff
@@ -0,0 +1,925 @@
...
+func (stub *txGroupsEncodingStub) reconstructSignedTransactions(signedTxns []transactions.SignedTxn, genesisID string, genesisHash crypto.Digest) (err error) {
+	var index int
+	index = 0
+	err = stub.BitmaskSig.Iterate(int(stub.TotalTransactionsCount), func(i int) error {
+		slice, err := getSlice(stub.Sig, index, len(crypto.Signature{}))
+		if err != nil {
+			return err
+		}
+		copy(signedTxns[i].Sig[:], slice)
+		index++
```
tsachiherman, line 55: The pattern that you've been using here is abit tricky ( but it is correct ).
I think that avoiding having the index and the slice would simplify the implementation and would be
more maintainable :

```golang
// inside the anonymous function
err := nextSlice(&stub.Sig, signedTxns[i].Sig[:], len(crypto.Signature{})
if err != nil {
    return err
}
// the nextSlice definition - copy the next slice and slide the src window.
func nextSlice(src *[]byte, dst []byte, size int) ([]byte, error) {
    if len(*src) < size {
         return errDataMissing
    }
    copy(dst[:], (*src)[: size])
    // slice the src window so next call would get the next entry.
    *src = (*src)[size:]
    return nil
}
```

txnsync/encodedgroupsunmarshalers.go
```diff
@@ -0,0 +1,925 @@
...
+	var index int
+	index = 0
+	err = stub.BitmaskVersion.Iterate(int(stub.TotalTransactionsCount), func(i int) error {
+		if index >= len(stub.Version) {
+			return errDataMissing
+		}
+		signedTxns[i].Msig.Version = stub.Version[index]
+		index++
+		return nil
+	})
```
tsachiherman, line 95: similar to the above, this clearly indicate that you're maintaining an index which is increased on each call to the anonymous function. I think that moving this index into the Iterate method, and passing this onto the anonymous function via a parameter would simplify the implementation.

tsachiherman: also, maybe pass the `len(stub.Version)` here as an argument to the `stub.BitmaskVersion.Iterate` so that the Iterate could perform the array length validation ?

txnsync/bitmask.go
```diff
@@ -0,0 +1,241 @@
...
+	"errors"
+)
+
+//msgp:allocbound bitmask maxBitmaskSize
+type bitmask []byte
+
+// assumed to be in mode 0, sets bit at index to 1
+func (b *bitmask) SetBit(index int) {
+	byteIndex := index/8 + 1
+	(*b)[byteIndex] ^= 1 << (index % 8)
```
tsachiherman, line 30: I think that it would be better to change this line to
```golang
(*b)[byteIndex] |= 1 << (index % 8)
```
otherwise, you've implemented toggleBit

txnsync/txngroups_test.go
```diff
@@ -47,23 +224,287 @@ func TestTxnGroupEncoding(t *testing.T) {
...
+			txn.Lsig.Logic = []byte("logic")
+			if i%3 != 0 {
+				txn.Sig = crypto.Signature{}
+			}
+			if i%3 != 1 {
+				txn.Msig = crypto.MultisigSig{}
+			}
+			if i%3 != 2 {
+				txn.Lsig = transactions.LogicSig{}
+			}
```
tsachiherman, line 438: coding style - use a switch statement, and initialize these to have some meaningful content.

txnsync/txngroups_test.go
```diff
@@ -47,23 +224,287 @@ func TestTxnGroupEncoding(t *testing.T) {
...
+	encodedGroupsBytes, err := encodeTransactionGroups(inTxnGroups)
 	require.NoError(t, err)
-	require.Equal(t, inTxnGroups, out)
+	out, err := decodeTransactionGroups(encodedGroupsBytes, genesisID, genesisHash)
+	require.NoError(t, err)
+	require.ElementsMatch(t, inTxnGroups, out)
+}
+
+func txnGroupsData() (txnGroups []transactions.SignedTxGroup, genesisID string, genesisHash crypto.Digest, err error) {
+	dat, err := ioutil.ReadFile("../test/testdata/mainnettransactions")
```
tsachiherman, line 267: Looks to me like this a file that contains all the blocks rather than the transactions.
While this is fine, please note that large portion of the data is going to be the certificate, which you couldn't care less about ( and consumes large portion of this binary file size ).

tsachiherman:
I was doing some performance profiling on this; it looks like the encoding is faster than before, and the decoding is slower than before. The following would double the decoding speed:
```golfing
func addGroupHashes(txnGroups []transactions.SignedTxGroup, txnCount int, b bitmask) {
	index := 0
	txGroupHashes := make([]crypto.Digest, 16)
	for _, txns := range txnGroups {
		if len(txns.Transactions) == 1 && !b.EntryExists(index, txnCount) {
			index++
			continue
		}

		var txGroup transactions.TxGroup
		txGroup.TxGroupHashes = txGroupHashes[:len(txns.Transactions)]
		for i, tx := range txns.Transactions {
			txGroup.TxGroupHashes[i] = crypto.HashObj(tx.Txn)
		}
		groupHash := crypto.HashObj(txGroup)
		for i := range txns.Transactions {
			txns.Transactions[i].Txn.Group = groupHash
			index++
		}
	}
}
```

since the `crypto.HashObj` is the slow element here. Additional optimization on that method would be to try and use the Iterate method.

txnsync/txngroups_test.go
nicholasguoalgorand: the point of this was such that if i%3 = 0, it would have a normal signature, 1 for multisig, and 2 for logicsig. I will refactor this to make it clearer.

tsachiherman:
Looking into the encoding performance, changing the various
```golang
func (stub *txGroupsEncodingStub) deconstructAssetParams(i int, txn transactions.SignedTxn)
```
to
```golang
func (stub *txGroupsEncodingStub) deconstructAssetParams(i int, txn *transactions.SignedTxn)
```

would almost double the encoding performance.

PR #2115, opened 2021-04-28
Add performance metrics for account updates

shiqizng:
This PR enables metrics logging configurations for account updates' commitRound(...).

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -228,11 +228,11 @@ type Local struct {
 	// Generate ProcessBlockMetrics telemetry event
 	EnableProcessBlockStats bool `version[0]:""`
 
-	// Generate ProcessBlockMetrics telemetry event
-	EnableAccountUpdatesStats bool `version[0]:""`
-
 	// Generate AccountUpdates telemetry event
-	AccountUpdatesStatsFrequency time.Duration `version[17]:"120"`
+	EnableAccountUpdatesStats bool `version[17]:"false"`
+
+	// Frequency for generating accountUpdates telemetry event
+	AccountUpdatesStatsFrequency time.Duration `version[17]:"5"`
 
 	// SuggestedFeeSlidingWindowSize is number of past blocks that will be considered in computing the suggested fee
 	SuggestedFeeSlidingWindowSize uint32 `version[3]:"50"`
--- a/config/config.go
+++ b/config/config.go
@@ -228,12 +228,6 @@ type Local struct {
 	// Generate ProcessBlockMetrics telemetry event
 	EnableProcessBlockStats bool `version[0]:""`
 
-	// Generate AccountUpdates telemetry event
-	EnableAccountUpdatesStats bool `version[17]:"false"`
-
-	// Frequency for generating accountUpdates telemetry event
-	AccountUpdatesStatsFrequency time.Duration `version[17]:"5"`
-
 	// SuggestedFeeSlidingWindowSize is number of past blocks that will be considered in computing the suggested fee
 	SuggestedFeeSlidingWindowSize uint32 `version[3]:"50"`
 
@@ -408,6 +402,12 @@ type Local struct {
 	// 12 : perform all validation methods (normal and additional). These extra tests helps to verify the integrity of the compiled executable against
 	//      previously used executabled, and would not provide any additional security guarantees.
 	CatchupBlockValidateMode int `version[17]:"0"`
+
+	// Generate AccountUpdates telemetry event
+	EnableAccountUpdatesStats bool `version[17]:"false"`
+
+	// Frequency for generating accountUpdates telemetry event
+	AccountUpdatesStatsFrequency time.Duration `version[17]:"5"`
 }
 
 // Filenames of config files within the configdir (e.g. ~/.algorand)
--- a/config/config.go
+++ b/config/config.go
@@ -406,8 +406,9 @@ type Local struct {
 	// Generate AccountUpdates telemetry event
 	EnableAccountUpdatesStats bool `version[17]:"false"`
 
-	// Frequency for generating accountUpdates telemetry event
-	AccountUpdatesStatsFrequency time.Duration `version[17]:"5"`
+	//
+	// Time interval in ns for generating accountUpdates telemetry event
+	AccountUpdatesStatsInterval time.Duration `version[17]:"5000000000"`
 }
 
 // Filenames of config files within the configdir (e.g. ~/.algorand)
--- a/config/config.go
+++ b/config/config.go
@@ -228,6 +228,12 @@ type Local struct {
 	// Generate ProcessBlockMetrics telemetry event
 	EnableProcessBlockStats bool `version[0]:""`
 
+	// Generate ProcessBlockMetrics telemetry event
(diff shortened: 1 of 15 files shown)


config/config.go
```diff
@@ -228,6 +228,12 @@ type Local struct {
 	// Generate ProcessBlockMetrics telemetry event
 	EnableProcessBlockStats bool `version[0]:""`
 
+	// Generate AccountUpdates telemetry event
+	EnableAccountUpdatesStats bool `version[17]:"false"`
+
+	// Frequency for generating accountUpdates telemetry event
+	AccountUpdatesStatsFrequency time.Duration `version[17]:"5"`
+
```
tsachiherman, line 236: could you please move these two to the end of the struct ?

config/config.go
```diff
@@ -228,6 +228,12 @@ type Local struct {
 	// Generate ProcessBlockMetrics telemetry event
 	EnableProcessBlockStats bool `version[0]:""`
 
+	// Generate AccountUpdates telemetry event
+	EnableAccountUpdatesStats bool `version[17]:"false"`
+
+	// Frequency for generating accountUpdates telemetry event
+	AccountUpdatesStatsFrequency time.Duration `version[17]:"5"`
```
tsachiherman: The value here is not in the correct units. The `time.Duration` is given in units of nanoseconds, so you'd need to specify a value of `5000000000` for 5 seconds. ( which are a good interval )
Also, given that we're using this as an interval, maybe we should change the name of this field to `AccountUpdatesStatsInterval`

ledger/acctupdates.go
```diff
@@ -249,6 +249,11 @@ type accountUpdates struct {
 
 	// the synchronous mode that would be used while the accounts database is being rebuilt.
 	accountsRebuildSynchronousMode db.SynchronousMode
+
+	// Metrics collection
+	logAccountUpdateMetrics bool
+	logAccountUpdateFreq    time.Duration
+	lastMetricsLogTime      time.Time
```
tsachiherman, line 256: please provide more detailed description for these fields.

ledger/acctupdates.go
```diff
@@ -2138,26 +2152,53 @@ func (au *accountUpdates) commitRound(offset uint64, dbRound basics.Round, lookb
...
+			err = compactDeltas.accountsLoadOld(tx)
+			if err != nil {
+				return err
+			}
+			stats.OldAccountPreloadDuration = time.Now().Sub(begin)
+		} else {
+			err = compactDeltas.accountsLoadOld(tx)
+			if err != nil {
+				return err
+			}
```
tsachiherman, line 2166: this isn't incorrect, but I think that duplicating the "actual work code" is not desired and would create more confusion.
what do you think about the following pattern: 
```go
if updateStats {
      stats.OldAccountPreloadDuration = time.Now().UnixNano()
}
err = compactDeltas.accountsLoadOld(tx)
if err != nil {
	return err
}
if updateStats {
      stats.OldAccountPreloadDuration = time.Now().UnixNano() - stats.OldAccountPreloadDuration
}
```

Note that in case of a back to back timing, you'd like to have only a single `time.Now()` call.

ledger/acctupdates.go
```diff
@@ -2034,6 +2044,10 @@ func (au *accountUpdates) commitSyncer(deferedCommits chan deferredCommit) {
 
 // commitRound write to the database a "chunk" of rounds, and update the dbRound accordingly.
 func (au *accountUpdates) commitRound(offset uint64, dbRound basics.Round, lookback basics.Round) {
+	var stats telemetryspec.AccountsUpdateMetrics
+
+	updateStats := au.logAccountUpdateMetrics && time.Now().Sub(au.lastMetricsLogTime) >= au.logAccountUpdateFreq
```
tsachiherman, line 2049: The duration of this method is known.. so you don't want to skew the metrics timing based on that.
i.e. the update time you place in `au.lastMetricsLogTime` should be the same as in this query:
```golang
var updateStats bool
if au.logAccountUpdateMetrics {
    now := time.Now()
    if now.Sub(au.lastMetricsLogTime) >= au.logAccountUpdateFreq {
       updateStats = true
       au.lastMetricsLogTime = now
    }
}
```

tsachiherman (commented):
once you'll finish making the changes on the config.go, you should run `make generate`. It will regenerate all the config files, where you'll need to add to this PR.

ledger/acctupdates.go
shiqizng: I thought having fewer `if updateStats` checks would keep the code more readable. But this pattern actually makes more sense since duplicating the actual work code would require duplicating any future changes, too.

ledger/acctupdates.go
```diff
@@ -2138,28 +2166,51 @@ func (au *accountUpdates) commitRound(offset uint64, dbRound basics.Round, lookb
...
+		if updateStats {
+			stats.MerkleTrieUpdateDuration = time.Duration(time.Now().UnixNano())
+		}
+
 		err = au.accountsUpdateBalances(compactDeltas)
 		if err != nil {
 			return err
 		}
 
+		if updateStats {
```
tsachiherman, line 2196: when having back-to-back timing, use the *same* value you received from `time.Now()`.
This would allow you to ensure there is no time-slippage between the different blocks, and that they always sum up.

tsachiherman:
You'll need to update `./test/testdata/configs/config-v17.json` with the updated content. You can take the content from the auto-gen `installer/config.json.example`

ledger/acctupdates.go
shiqizng: do you mean use time.Now().UnixNano() instead of time.Duration(time.Now().UnixNano())?

shiqizng: do you mean use time.Now().UnixNano() instead of time.Duration(time.Now().UnixNano())?

tsachiherman: no, I mean -
```golang
if updateStats {
     now := time.Duration(time.Now().UnixNano())
     stats.MerkleTrieUpdateDuration = now - stats.MerkleTrieUpdateDuration
     stats.AccountsWritingDuration = now
}
```
so that it would be the *same* `time.Now()`

ledger/acctupdates.go
```diff
@@ -2180,6 +2231,10 @@ func (au *accountUpdates) commitRound(offset uint64, dbRound basics.Round, lookb
 		return
 	}
 
+	if updateStats {
+		stats.DatabaseCommitDuration = time.Duration(time.Now().UnixNano()) - stats.DatabaseCommitDuration
```
tsachiherman, line 2235: my intention here was to measure the time of the database commit method. While you can't directly see it here ( i.e. it's hidden inside `Atomic` ), that is the method that can take a bulky amount of time.
In order to measure that, you'll need to subtract the "content" of the Atomic function call, i.e.
```golang
stats.DatabaseCommitDuration = time.Duration(time.Now().UnixNano()) - stats.DatabaseCommitDuration - stats.AccountsWritingDuration - stats.MerkleTrieUpdateDuration - stats.OldAccountPreloadDuration
```

tsachiherman (approved):
looks good to me.

tsachiherman (approved):

PR #2122, opened 2021-04-30
add support for a single node private network

tsachiherman:
## Summary

This PR adds the support for creating a single node private network. On such a node, the entire network functionality need to be disabled, since there is no "other" parties to communicate with.

## Test Plan

Expect test was added.


tools/network/telemetryURIUpdateService.go
```diff
@@ -60,7 +60,7 @@ func (t *telemetryURIUpdater) Start() {
 		updateTelemetryURI := func() {
 			endpointURL := t.lookupTelemetryURL()
 
-			if endpointURL != nil && endpointURL.String() != t.log.GetTelemetryURI() {
+			if endpointURL != nil && endpointURL.String() != t.log.GetTelemetryURI() && false == t.cfg.DisableNetwork {
```
winder, line 63: Was there a specific problem with telemetry? There might be cases where it would be useful for a single-node deployment and I don't see a technical reason why it would have to be disabled.

winder (approved):
If we can leave telemetry alone I think that would be better, other than that LGTM.

tools/network/telemetryURIUpdateService.go
tsachiherman: So, it's true that for a single-node network this isn't an issue. However, the flag is "DisableNetwork", and as such, no outgoing network activity is expected.. including no telemetry.
( note that I didn't want to disable the telemetry since we currently use it for logging to file ).

network/wsNetwork.go
```diff
@@ -723,6 +727,9 @@ func (wn *WebsocketNetwork) setup() {
 
 // Start makes network connections and threads
 func (wn *WebsocketNetwork) Start() {
+	if wn.config.DisableNetwork {
```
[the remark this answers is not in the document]
tsachiherman: sure. I'll make that change.

tools/network/telemetryURIUpdateService.go
winder: Maybe call the property "DisableNetworking" in that case, I read it as "DisableAlgorandNetwork"

catchup/service.go
```diff
@@ -498,7 +498,7 @@ func (s *Service) periodicSync() {
 			sleepDuration = time.Duration(crypto.RandUint63()) % s.deadlineTimeout
 			continue
 		case <-time.After(sleepDuration):
-			if sleepDuration < s.deadlineTimeout {
```
algonautshant, line 501: What is the purpose of periodicSync() if there is not network? 
Maybe return it before the loop?

algonautshant (dismissed):
Looks good.
Couple of comments.

catchup/service.go
```diff
@@ -498,7 +498,7 @@ func (s *Service) periodicSync() {
...
+			if sleepDuration < s.deadlineTimeout || s.cfg.DisableNetwork {
```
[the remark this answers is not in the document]
tsachiherman: sure.

catchup/service.go
tsachiherman: we still need it for the `unmatchedPendingCertificates`, which could come from the agreement protocol. I believe that it shouldn't happen, but it would be simpler just to let it pass through.

algonautshant (approved):

PR #2125, opened 2021-05-02
improve the support for overlapping participation keys

tsachiherman:
## Summary

improve the support for overlapping participation keys

## Test Plan

todo

### DIFF
--- a/agreement/agreementtest/keyManager.go
+++ b/agreement/agreementtest/keyManager.go
@@ -25,23 +25,14 @@ import (
 type SimpleKeyManager []account.Participation
 
 // Keys implements KeyManager.Keys.
-func (m SimpleKeyManager) Keys() []account.Participation {
+func (m SimpleKeyManager) Keys(rnd basics.Round) []account.Participation {
 	var km []account.Participation
 	for _, acc := range m {
-		km = append(km, acc)
-	}
-	return km
-}
-
-// HasLiveKeys returns true if we have any Participation
-// keys valid for the specified round range (inclusive)
-func (m SimpleKeyManager) HasLiveKeys(from, to basics.Round) bool {
-	for _, acc := range m {
-		if acc.OverlapsInterval(from, to) {
-			return true
+		if acc.FirstValid <= rnd && rnd <= acc.LastValid {
+			km = append(km, acc)
 		}
 	}
-	return false
+	return km
 }
 
 // DeleteOldKeys implements KeyManager.DeleteOldKeys.
--- a/agreement/pseudonode.go
+++ b/agreement/pseudonode.go
@@ -67,14 +67,16 @@ type pseudonode interface {
 
 // asyncPseudonode creates proposals and votes asynchronously.
 type asyncPseudonode struct {
-	factory   BlockFactory
-	validator BlockValidator
-	keys      KeyManager
-	ledger    Ledger
-	log       serviceLogger
-	quit      chan struct{}   // a quit signal for the verifier goroutines
-	closeWg   *sync.WaitGroup // frontend waitgroup to get notified when all the verifier goroutines are done.
-	monitor   *coserviceMonitor
+	factory                BlockFactory
+	validator              BlockValidator
+	keys                   KeyManager
+	ledger                 Ledger
+	log                    serviceLogger
+	quit                   chan struct{}   // a quit signal for the verifier goroutines
+	closeWg                *sync.WaitGroup // frontend waitgroup to get notified when all the verifier goroutines are done.
+	monitor                *coserviceMonitor
+	participationKeysRound basics.Round            // the round to which the participationKeys matches
+	participationKeys      []account.Participation // the list of the participation keys for round participationKeysRound
 
 	proposalsVerifier *pseudonodeVerifier // dynamically generated verifier goroutine that manages incoming proposals making request.
 	votesVerifier     *pseudonodeVerifier // dynamically generated verifier goroutine that manages incoming votes making request.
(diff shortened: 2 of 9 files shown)


brianolson (commented):
looks good, only two little quibbles

agreement/agreementtest/keyManager.go
```diff
@@ -24,24 +24,15 @@ import (
...
-	return km
-}
-
-// HasLiveKeys returns true if we have any Participation
-// keys valid for the specified round range (inclusive)
-func (m SimpleKeyManager) HasLiveKeys(from, to basics.Round) bool {
-	for _, acc := range m {
-		if acc.OverlapsInterval(from, to) {
-			return true
+		if acc.FirstValid <= votingRound && votingRound <= acc.LastValid {
```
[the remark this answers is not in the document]
tsachiherman: Just for the sake of consistency, I'll make these all use `OverlapsInterval()`. From logical perspective, there shouldn't be any difference.

agreement/pseudonode_test.go
```diff
@@ -374,3 +379,67 @@ func (n serializedPseudonode) MakeVotes(ctx context.Context, r round, p period,
...
+}
+
+func (k *KeyManagerProxy) VotingKeys(votingRound, balanceRound basics.Round) []account.Participation {
+	return k.target(votingRound, balanceRound)
+}
+
+func TestPseudonodeLoadingOfParticipationKeys(t *testing.T) {
+	t.Parallel()
+
+	logging.Base().SetLevel(logging.Warn)
```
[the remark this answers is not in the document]
tsachiherman: exactly. and it was triggering a datarace. I fixed it by creating a local Logger, which means that we avoid using the global variable.

agreement/pseudonode.go
```diff
@@ -244,21 +270,6 @@ func (n asyncPseudonode) makePseudonodeVerifier(voteVerifier *AsyncVoteVerifier)
 	return pv
 }
 
-// getParticipations retrieves the participation accounts for a given round.
-func (n asyncPseudonode) getParticipations(procName string, round basics.Round) []account.Participation {
```
winder, line 248: You could hide the caching inside this function instead of requiring the caller to call `loadParticipationKeys`

tsachiherman: I honestly don't see a huge difference here.. the function effectively was renamed from `getParticipations` -> `loadRoundParticipationKeys`, plus the addition of the caching logic.

agreement/pseudonode.go
```diff
@@ -193,33 +195,57 @@ func (n asyncPseudonode) MakeVotes(ctx context.Context, r round, p period, s ste
...
 func (n asyncPseudonode) makeProposalsTask(ctx context.Context, r round, p period) pseudonodeProposalsTask {
-	participation := n.getParticipations("asyncPseudonode.makeProposalsTask", r)
+	n.loadRoundParticipationKeys(r)
 
 	pt := pseudonodeProposalsTask{
 		pseudonodeBaseTask: pseudonodeBaseTask{
 			node:          &n,
 			context:       ctx,
-			participation: participation,
+			participation: n.participationKeys,
```
winder, line 229: It's sections like this that make me a little uneasy. `n.participationKeys` is always available, so someone unfamiliar with the code might mistakenly use stale data.

Previously you were required to call the function, now the developer must remember to update the cache.

tsachiherman: So, the only way to work around that (nicely) would be to create a caching object that would cache the data, and return a slice. While this is clearly doable, I think that for this use case, it's definitely crossing the line to the over engineering realm. ( i.e. it's an internal functionality within an object; the object is fairy simple; the usage is very localized and have exactly two instance )

winder: Maybe I'm not explaining myself properly, I was thinking it would simply be a helper function and allow the existing code to work with the improved algorithm. Effectively this:

```
func (n asyncPseudonode) getParticipations(procName string, round basics.Round) []account.Participation {
    n.loadRoundParticipationKeys(r)
    return n.participationKeys
}
```

tsachiherman: I honestly don't see how adding this function would help - If I were to read this, I would probably think of that as a `spaghetti code`. But Iets see the other reviewer's for their view on that ? @brianolson @nicholasguoalgorand

brianolson: I kinda like the accessor that ensures cache state. If implemented, add a comment to the variable that it should be read from the accessor.

nicholasguoalgorand: Why are we using the logic of setting `n.participationKeys` in `loadRoundParticipationKeys` when we call `loadRoundParticipationKeys` every time before accessing `n.participationKeys` anyway? Could we have `loadRoundParticipationKeys` just return the list of keys?

winder (approved):

agreement/pseudonode.go
```diff
@@ -193,42 +195,61 @@ func (n asyncPseudonode) MakeVotes(ctx context.Context, r round, p period, s ste
...
-			out:           make(chan externalEvent),
+			node:    &n,
+			context: ctx,
+			out:     make(chan externalEvent),
 		},
 		round:  r,
 		period: p,
 	}
-	if len(participation) == 0 {
+	if !pt.populateParticipationKeys(r) {
```
winder, line 233: It's a "functional programming" approach to avoid side effects, so this misses the point a little bit by providing a similar guarantee using a side effect. To your point earlier, this is a private function that isn't used widely, so I don't think this nit is important enough to worry about further.

PR #2126, opened 2021-05-03
Dynamic teal

jannotti:
<!--
Thanks for submitting a pull request! We appreciate the time and effort you spent to get this far.

If you haven't already, please make sure that you've reviewed the CONTRIBUTING guide:
https://github.com/algorand/go-algorand/blob/master/CONTRIBUTING.md#code-guidelines
## Summary

Change to a dynamic cost model, starting in in teal v4.  This means that programs will be aborted if they use more than their allotted cost, rather than rejected if the sum of their (static) instruction sequence is too long.

This means that we can (and this PR does) allow loops and a sort of "function" call in TEAL.  Loops are added by allowing negative back jumps.  We continue to check that jump targets are the beginnings of instructions as an added protection against jumping into constants or similar.

Functions are added with "callsub" and "retsub".  Callsub stores the proper return address on a hidden call stack, retsub pops and returns to such addresses.  This means these address are never directly visible nor used as TEAL values - so indirect jumps are not yet supported.

## Test Plan

Units tests that cover show backward jumps are allowed only in v4 and above.  Code samples that work with backjumps and the the call/retsub opcodes. (One implements factorial, showing recursion works.)


data/transactions/logic/eval.go
```diff
@@ -1254,6 +1255,22 @@ func opB(cx *evalContext) {
 	cx.nextpc = target
 }
 
+func opCallSub(cx *evalContext) {
+	cx.callstack = append(cx.callstack, cx.pc+3)
```
algorandskiy, line 1259: I suggest to use a hidden stack for a chain of ret addresses with call to store and ret to pop. No other access. This is because of a long history of security problems in x86 and ret addresses on a stack.

jannotti: Yes, that's what this is.  cx.callstack inaccessible from any other instruction, except retsub accesses and returns to.

data/transactions/logic/eval.go
```diff
@@ -218,22 +224,28 @@ func (ep EvalParams) log() logging.Logger {
...
+	program   []byte // txn.Lsig.Logic ?
+	pc        int
+	nextpc    int
+	err       error
+	intc      []uint64
+	bytec     [][]byte
+	version   uint64
+	scratch   [256]stackValue
+
+	cost int // cost incurred so far
```
algorandskiy, line 238: what is the point of having `cost` as a member of context? I guess the less shared vars the better (imo returning a pair (cost, err) from checkers is fine)

data/transactions/logic/eval.go
```diff
@@ -774,6 +771,41 @@ func opAddw(cx *evalContext) {
 	cx.stack[last].Uint = sum
 }
 
+func uint128(hi uint64, lo uint64) *big.Int {
+	whole := new(big.Int).SetUint64(hi)
```
algorandskiy, line 775: it is interesting why does not `big` package like stack-allocated `Int`s... Have you tried
```
whole := &big.Int{}
whole.SetUint64(hi)
...
```
?

data/transactions/logic/eval_test.go
```diff
@@ -536,34 +537,46 @@ int 1                   // ret 1
 `, 2)
 }
 
+func TestUint128(t *testing.T) {
+	x := uint128(0, 3)
+	require.Equal(t, x.String(), "3")
+	x = uint128(1, 3)
+	require.Equal(t, x.String(), "18446744073709551619")
+	x = uint128(1, 5)
+	require.Equal(t, x.String(), "18446744073709551621")
```
algorandskiy, line 546: add uint128(0, 0) and uint128(uint64(-1), uint64(-1)) tests?

data/transactions/logic/eval_test.go
```diff
@@ -536,34 +537,46 @@ int 1                   // ret 1
...
+	testAccepts(t, `int 0; int 7777; int 1; int 0; divw;
+                        int 7777; ==; assert;
+                        int 0; ==; assert;
+                        int 0; ==; assert;
+                        int 0; ==; assert; int 1`, 4)
+
+	// 10:0 / 0:0 ==> panic
+	testPanics(t, `int 10; int 0; int 0; int 0; divw;
+	               pop; pop; pop; pop; int 1`, 4)
+}
```
algorandskiy, line 575: add 0 / 1, and (-1, -1) / (-1, -1) tests ?

data/transactions/logic/eval_test.go
```diff
@@ -2781,18 +2803,18 @@ bytecblock 0x01234576 0xababcdcd 0xf000baad
 done:
 int 1`, v)
 			require.NoError(t, err)
-			//t.Log(hex.EncodeToString(ops.Program))
+			//t.Log(hex.EncodeToString(ops.Program))                  brhilo
```
algorandskiy, line 2806: what is it?)

algorandskiy (commented):
Few comments and suggestions.
Please also consider adding backward branching tests into backward compatibility tests - like  the program rejected before teal4 and accepted after

data/transactions/logic/eval_test.go
jannotti: done. I assume you mean a test for largest uint64, I used
x = uint128(^uint64(0), ^uint64(0)) // maximum uint128 = 2^64-1
require.Equal(t, x.String(), "340282366920938463463374607431768211455")

data/transactions/logic/eval_test.go
jannotti: done

data/transactions/logic/eval_test.go
jannotti: I'm labelling the branch (br) high byte (hi) and low byte (lo) in the string below this line.  I moved it to a comment on a line on its own, and explained.  I'm labeling it like this because this test explicitly clobbering those bytes to test what happens so it was useful for me "mark" them.

data/transactions/logic/eval.go
jannotti: I think that's fine too, but why prefer two lines / statements instead of a single expression?

I wanted to use syntax that gave a single expression that produced a big.Int from a u64, especially since I needed similar syntax two lines later to create the low word.

data/transactions/logic/eval.go
jannotti: I agree, I can probably get rid of it.  This is the "old" static kind of cost, it's not new code with this PR.  go just reformatted the line because I got rid of stepCount, above it.

jannotti: I take that back. It could have been removed, if we only cared about the old static check. It would be reasonable for checkStep to simply return the cost, as you suggest, rather than modify state.  The program checker would sum them, and if it's too high, reject the program.

But when doing dynamic checking, you want to stop as soon as you decode the instruction that would exceed the budget.  That means the total cost so far is truly part of the execution environment.  That final instruction should never even run.  So you would have to pass the running total into the step() method.  But that's exactly what the evalContext should represent.

That is, this code in `step()` needs to work as it does:
```
	cx.cost += deets.Cost
	if cx.cost > cx.budget() {
		cx.err = fmt.Errorf("%3d %s budget exceeded", cx.pc, spec.Name)
		return
	}
	spec.op(cx)
```

jannotti:
@algorandskiy You asked about testing old backjumps.  This test does that

data/transactions/logic/eval.go
```diff
@@ -218,22 +224,28 @@ func (ep EvalParams) log() logging.Logger {
...
+
+	// Set of PC values that branches we've seen so far might
+	// go. So, if checkStep() skips one, that branch is trying to
+	// jump into the middle of a multibyte instruction
+	branchTargets map[int]bool
+
+	// Set of PC values that we have begun a checkStep() with. So
+	// if a back jump is going to a value that isn't here, it's
+	// jumping into the middle of multibyte instruction.
+	instructionStarts map[int]bool
```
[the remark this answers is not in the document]
jannotti: Thanks! I forgot that I got pulled away from this work before I checked back branches for alignment.  Added.

test/scripts/e2e_subs/v26/teal-v3-only.sh
```diff
@@ -0,0 +1,101 @@
+#!/bin/bash
+
+date '+teal-v2-only start %Y%m%d_%H%M%S'
```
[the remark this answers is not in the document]
jannotti: yes, thanks

cmd/goal/clerk.go
```diff
@@ -1097,7 +1094,7 @@ var dryrunCmd = &cobra.Command{
 			}
 			pass, err := logic.Eval(txn.Lsig.Logic, ep)
 			// TODO: optionally include `inspect` output here?
-			fmt.Fprintf(os.Stdout, "tx[%d] cost=%d trace:\n%s\n", i, cost, sb.String())
+			fmt.Fprintf(os.Stdout, "tx[%d] trace:\n%s\n", i, sb.String())
```
algorandskiy, line 1097: why not printing the cost after eval?

data/transactions/logic/README_in.md
```diff
@@ -27,7 +27,7 @@ A program can either authorize some delegated action on a normal private key sig
...
 
-The TEAL bytecode plus the length of any Args must add up to less than 1000 bytes (consensus parameter LogicSigMaxSize). Each TEAL op has an associated cost estimate and the program cost estimate must total less than 20000 (consensus parameter LogicSigMaxCost). Most ops have an estimated cost of 1, but a few slow crypto ops are much higher.
+The TEAL bytecode plus the length of any Args must add up to less than 1000 bytes (consensus parameter LogicSigMaxSize). Each TEAL op has an associated cost and the program cost must total less than 20000 (consensus parameter LogicSigMaxCost). Most ops have a cost of 1, but a few slow crypto ops are much higher. Prior to v4, program costs was estimated as the static sum of all opcode costs in a program (ignoring conditionals that might skip some code).  Beginning with v4, a program's cost is tracked dynamically, while being evaluated. If the program exceeds its budget, it fails.
```
algorandskiy, line 30: double space at "some code). Beginning"

```suggestion
The TEAL bytecode plus the length of any Args must add up to less than 1000 bytes (consensus parameter LogicSigMaxSize). Each TEAL op has an associated cost and the program cost must total less than 20000 (consensus parameter LogicSigMaxCost). Most ops have a cost of 1, but a few slow crypto ops are much higher. Prior to v4, program costs was estimated as the static sum of all opcode costs in a program (ignoring conditionals that might skip some code). Beginning with v4, a program's cost is tracked dynamically, while being evaluated. If the program exceeds its budget, it fails.
```

data/transactions/logic/opcodes.go
```diff
@@ -226,6 +231,16 @@ var OpSpecs = []OpSpec{
...
+
+	// "Function oriented"
+	{0x88, "callsub", opCallSub, assembleBranch, disBranch, nil, nil, 4, modeAny, opBranch},
+	{0x89, "retsub", opRetSub, asmDefault, disDefault, nil, nil, 4, modeAny, opDefault},
+	// Leave a little room for indirect function calls, or similar
+
+	// More math
+	// shl, shr
+	// divw, modw convenience
+	// expmod
```
algorandskiy, line 243: useless comment for this PR

data/transactions/verify/txn.go
```diff
@@ -214,13 +214,10 @@ func LogicSigSanityCheck(txn *transactions.SignedTxn, groupIndex int, groupCtx *
...
 		MinTealVersion: &groupCtx.minTealVersion,
 	}
-	cost, err := logic.Check(lsig.Logic, ep)
+	err := logic.Check(lsig.Logic, ep)
 	if err != nil {
 		return err
 	}
-	if cost > int(groupCtx.consensusParams.LogicSigMaxCost) {
-		return fmt.Errorf("LogicSig.Logic too slow, %d > %d", cost, groupCtx.consensusParams.LogicSigMaxCost)
-	}
```
algorandskiy, line 223: this looks like can break backward compatibility. I think for TEAL v1-v3 we still need this check.

algorandskiy (commented):
Looks good except removing cost checks from txn verifier for pre-eval checks. Scenario: submit a long v2 program with jump to the end on a new and old node before protocol upgrade. It will fail on the old (sum up check) and pass on new.

cmd/goal/clerk.go
jannotti: Because it's hidden inside the evaluation function and not exposed.  If we find it important enough, I could rework the interfaces.

data/transactions/logic/opcodes.go
jannotti: true, but it'll go away with the math pr that follows.  I'd rather not remove, as it'll just cause a conflict when I rebase the math pr to this one.

data/transactions/verify/txn.go
algorandskiy: I see now, thanks for pointing out

algorandskiy:
LGTM

algorandskiy (dismissed):

data/transactions/logic/eval.go
```diff
@@ -427,26 +431,31 @@ func eval(program []byte, cx *evalContext) (pass bool, err error) {
 	return cx.stack[0].Uint != 0, nil
 }
 
-// CheckStateful should be faster than EvalStateful.
-// Returns 'cost' which is an estimate of relative execution time.
-func CheckStateful(program []byte, params EvalParams) (cost int, err error) {
+// CheckStateful should be faster than EvalStateful.  It can perform
+// static checks and reject programs that are invalid. Returns a cost
```
brianolson, line 435: it no longer returns a cost?

test/scripts/e2e_subs/v26/teal-v3-only.sh
```diff
@@ -0,0 +1,101 @@
...
+ACCOUNT=$(${gcmd} account list|awk '{ print $3 }')
+
+# prints:
+# Created new account with address UCTHHNBEAUWHDQWQI5DGQCTB7AR4CSVNU5YNPROAYQIT3Y3LKVDFAA5M6Q
+ACCOUNTB=$(${gcmd} account new|awk '{ print $6 }')
+
+cat >${TEMPDIR}/true.teal<<EOF
+#pragma version 3
+int 1
+return
```
brianolson, line 23: don't need `return`

test/scripts/e2e_subs/v26/teal-v3-only.sh
```diff
@@ -0,0 +1,101 @@
...
+set +o pipefail
+# v4 opcodes with v3 pragma fails
+printf "#pragma version 3\nbegin: int 1\nb begin" | ${gcmd} clerk compile - 2>&1 | grep "back jump support"
+set -o pipefail
+
+# Although we are in an earlier version, v4 can be compiled, it just can't be used.
+cat >${TEMPDIR}/true4.teal<<EOF
+#pragma version 4
+int 1
+return
```
brianolson, line 69: s/return//

brianolson (commented):
Lots of good stuff here. A few small tweaks.

Probably worth having a paragraph in README_in.md about the callstack, how it's invisible except to two instructions, and any depth limit.

jannotti:
> Lots of good stuff here. A few small tweaks.
> 
> Probably worth having a paragraph in README_in.md about the callstack, how it's invisible except to two instructions, and any depth limit.

I'll add something in TEAL cannot do section - I'll create a new line about lack of indirect jumps and this will naturally fit there. (Though I'll do it the next PR, as I want to avoid having this thing run through CI again, until/unless I get feedback that needs a code change.)

My current thinking is NOT to restrict the call depth, and to simply let that be limited by the runtime of the program.  That means a program of `loop: callsub loop` would get 20,000 deep in a logicsig, but that's just a 20k int array that immediately goes away.  Didn't seem to matter.  Thoughts?

test/scripts/e2e_subs/v26/teal-v3-only.sh
jannotti: Right this is a remnant from the teal-v2 test, where I wanted to be clear that I had v3 code (where return was introduced). I think it still makes *some* sense here because of that, but the one below should be changed to some v4 opcode.

jannotti: Nope, I'm wrong.  return was in v2.  So it's entirely extraneous. I guess I got in the habit of doing that in tests when I wasn't sure of the stack.

data/transactions/logic/eval.go
jannotti: True, I'll fix the comment.  For versions that do static checking, Check() errs directly, rather than returning a high cost for checking on the outside.

jasonpaulos (approved):
Looks good!

> My current thinking is NOT to restrict the call depth, and to simply let that be limited by the runtime of the program.  That means a program of `loop: callsub loop` would get 20,000 deep in a logicsig, but that's just a 20k int array that immediately goes away.  Didn't seem to matter.  Thoughts?

From a programmer/debugger standpoint this seems like a good approach, though I can't comment on whether that's a significant memory usage or not.

data/transactions/logic/assembler.go
```diff
@@ -1148,8 +1155,8 @@ func (ops *OpStream) resolveLabels() {
 		}
 		// all branch instructions (currently) are opcode byte and 2 offset bytes, and the destination is relative to the next pc as if the branch was a no-op
 		naturalPc := lr.position + 3
-		if dest < naturalPc {
-			ops.errorf("label %v is before reference but only forward jumps are allowed", lr.label)
+		if ops.Version < backBranchEnabledVersion && dest < naturalPc {
+			ops.errorf("label %v is a back reference, back jump support was introduced in TEAL v4", lr.label)
```
tsachiherman, line 1159: nit: given that `label` is a string, using `%s` is more appropriate.

data/transactions/logic/assembler_test.go
```diff
@@ -263,44 +293,33 @@ func TestAssemble(t *testing.T) {
...
-	expectedBytes, _ = hex.DecodeString("032008b7a60cf8acd19181cf959a12f8acd19181cf951af8acd19181cf15f8acd191810f01020026050212340c68656c6c6f20776f726c6421208dae2087fbba51304eb02b91f656948397a7946390e8cb70fc9ea4d95f92251d024242047465737400320032013202320328292929292a0431003101310231043105310731083109310a310b310c310d310e310f3111311231133114311533000033000133000233000433000533000733000833000933000a33000b33000c33000d33000e33000f3300113300123300133300143300152d2e0102222324252104082209240a220b230c240d250e230f23102311231223132314181b1c2b171615400003290349483403350222231d4a484848482a50512a63222352410003420000432105602105612105270463484821052b62482b642b65484821052b2106662b21056721072b682b692107210570004848210771004848361c0037001a0031183119311b311d311e311f3120210721051e312131223123312431253126312731283129312a312b312c312d312e312f4478222105531421055427042106552105082106564c4d4b02210538212106391c0081e80780046a6f686e")
-	if bytes.Compare(expectedBytes, ops.Program) != 0 {
-		// this print is for convenience if the program has been changed. the hex string can be copy pasted back in as a new expected result.
-		t.Log(hex.EncodeToString(ops.Program))
+
+	t.Parallel()
```
tsachiherman, line 297: I think that what you really meant was to place the t.Parallel() inside the for loop to allow the different versions to run concurrently to each other, right ?

data/transactions/logic/TEAL_opcodes.md
```diff
@@ -851,3 +861,19 @@ pushbytes args are not added to the bytecblock during assembly processes
 - LogicSigVersion >= 3
 
 pushint args are not added to the intcblock during assembly processes
+
+## callsub target
+
+- Opcode: 0x88
+- Pops: _None_
+- Pushes: _None_
+- branch unconditionally to TARGET, saving the next instruction on the call stack
```
tsachiherman, line 870: I think that we should document better that this stack is separated from the user-accessible stack.

tsachiherman (commented):
Few minor issues mentioned above, but overall looks great.
I would be happy to see more extensive testing around the new teal opcodes - maybe some teal code that is doing a dual/triple recursion ?

data/transactions/logic/assembler.go
jannotti: I'm happy to change, though I think I'd rather change to %#v and get quotes around the label.

data/transactions/logic/assembler_test.go
jannotti: I think that would be better, yes.  I think I added it here so that this test as a whole could be run concurrently (it wasn't previously marked).  But I think it is ok for the subtests as well.  Might as well.

jannotti: Actually, it would require some restructuring, since the `v` in the for loop is shared across go routines. Leaving it on the outside at least gets the test as a whole going concurrently with the rest.  I don't think there's a ton of value in the higher-level of concurrency of individual subtests. (throughout this file, we don't do subtests in parallel)

data/transactions/logic/TEAL_opcodes.md
jannotti: Added.

jannotti:
> maybe some teal code that is doing a dual/triple recursion ?

I'll do a mutually recursive odd() / even() implementation.

brianolson (approved):

jannotti:
Aw crap.  I committed some changes to address @tsachiherman comments.  So now approvals are dismissed.  Once I get through CI, I'll ping for them again.  Sorry.

algorandskiy (approved):

PR #2128, opened 2021-05-03
Add tests for ledger/compactcert.go

brianolson:
## Summary

Add tests.

## Test Plan

This is tests.

### DIFF
--- a/compactcert/builder.go
+++ b/compactcert/builder.go
@@ -321,7 +321,7 @@ func (ccw *Worker) tryBuilding() {
 
 	for rnd, b := range ccw.builders {
 		firstValid := ccw.ledger.Latest() + 1
-		acceptableWeight := ledger.AcceptableCompactCertWeight(b.votersHdr, firstValid)
+		acceptableWeight := ledger.AcceptableCompactCertWeight(b.votersHdr, firstValid, nil)
 		if b.SignedWeight() < acceptableWeight {
 			// Haven't signed enough to build the cert at this time..
 			continue
--- a/compactcert/builder.go
+++ b/compactcert/builder.go
@@ -27,6 +27,7 @@ import (
 	"github.com/algorand/go-algorand/data/basics"
 	"github.com/algorand/go-algorand/data/transactions"
 	"github.com/algorand/go-algorand/ledger"
+	"github.com/algorand/go-algorand/logging"
 	"github.com/algorand/go-algorand/network"
 	"github.com/algorand/go-algorand/protocol"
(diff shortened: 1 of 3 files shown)


brianolson:
86.4% coverage, most of what's not hit is overflow cases that might not actually be possible to hit

ledger/compactcert_test.go
```diff
@@ -0,0 +1,171 @@
...
+	firstValid = basics.Round(29000)
+	config.Consensus[votersHdr.CurrentProtocol] = proto
+	cc.CompactCertVotersTotal.Raw = 0x7fffffffffffffff
+	votersHdr.CompactCert[protocol.CompactCertBasic] = cc
+	proto.CompactCertWeightThreshold = 0x7fffffff
+	config.Consensus[votersHdr.CurrentProtocol] = proto
+	out = AcceptableCompactCertWeight(votersHdr, firstValid)
+	require.Equal(t, uint64(0x4cd35a85213a92a2), out)
+
+	// Covers everything except "overdlow that shouldn't happen" branches
```
zeldovich, line 143: Minor typo: `overdlow`

zeldovich (dismissed):
Looks good to me

zeldovich (approved):

tsachiherman (changes requested):
I think that the unit test is well written - but the usage of a global variable increase the likelihood for a data race. Instead of using a global variable, pass the logger object to the `AcceptableCompactCertWeight` method ( this is quite tedious, but it's only an issues for non-objects. Objects should "inherit" the logger hierarchically to avoid that very issue )

compactcert/builder.go
```diff
@@ -321,7 +321,7 @@ func (ccw *Worker) tryBuilding() {
 
 	for rnd, b := range ccw.builders {
 		firstValid := ccw.ledger.Latest() + 1
-		acceptableWeight := ledger.AcceptableCompactCertWeight(b.votersHdr, firstValid)
+		acceptableWeight := ledger.AcceptableCompactCertWeight(b.votersHdr, firstValid, nil)
```
tsachiherman, line 324: nit: pass in `logging.Base()` instead of `nil` in
```golang
acceptableWeight := ledger.AcceptableCompactCertWeight(b.votersHdr, firstValid, nil)
```
so that you could avoid the `logger == nil` check in 
```golang
if logger == nil {
 	logger = logging.Base()
 }
```

tsachiherman (approved):
looks good to me.

PR #2129, opened 2021-05-04
Add ParticipationKeysRefreshInterval to the node config

tsachiherman:
## Summary

This PR adds a ParticipationKeysRefreshInterval config option that would replace the hard-coded 60-second value used.

## Test Plan

No testing provided in this PR. Subsequent PR ( https://github.com/algorand/go-algorand/pull/2125 ) will be using these config option in it's e2e test.

### DIFF
--- a/agreement/fuzzer/fuzzer_test.go
+++ b/agreement/fuzzer/fuzzer_test.go
@@ -84,7 +84,7 @@ func MakeFuzzer(config FuzzerConfig) *Fuzzer {
 		crashAccessors:   make([]db.Accessor, config.NodesCount),
 		accounts:         make([]account.Participation, config.NodesCount),
 		balances:         make(map[basics.Address]basics.AccountData),
-		accountAccessors: make([]db.Accessor, config.NodesCount*2),
+		accountAccessors: make([]db.Accessor, config.NodesCount),
 		ledgers:          make([]*testLedger, config.NodesCount),
 		agreementParams:  make([]agreement.Parameters, config.NodesCount),
 		tickGranularity:  time.Millisecond * 300,
@@ -196,7 +196,7 @@ func (n *Fuzzer) initAccountsAndBalances(rootSeed []byte, onlineNodes []bool) er
 		if err != nil {
 			return err
 		}
-		n.accountAccessors[i*2+0] = rootAccess
+		n.accountAccessors[i] = rootAccess
 
 		seed = sha256.Sum256(seed[:])
 		root, err := account.ImportRoot(rootAccess, seed)
--- a/data/account/participation.go
+++ b/data/account/participation.go
@@ -106,7 +106,7 @@ func (part Participation) VotingSigner() crypto.OneTimeSigner {
 }
 
 // GenerateRegistrationTransaction returns a transaction object for registering a Participation with its parent.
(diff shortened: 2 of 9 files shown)


data/account/participation.go
```diff
@@ -203,19 +202,24 @@ func (part PersistedParticipation) Persist() error {
...
 
 		_, err = tx.Exec("INSERT INTO ParticipationAccount (parent, vrf, voting, firstValid, lastValid, keyDilution) VALUES (?, ?, ?, ?, ?, ?)",
 			part.Parent[:], rawVRF, rawVoting, part.FirstValid, part.LastValid, part.KeyDilution)
 		if err != nil {
-			return fmt.Errorf("Participation.persist: failed to insert account: %v", err)
+			return fmt.Errorf("failed to insert account: %w", err)
 		}
 		return nil
 	})
+
```
tsachiherman, line 218: The change in this method ensures that in all the error cases, we will have the "PersistedParticipation.Persist" text as part of the error ( i.e. including if the error was generated by the Commit method )

netdeploy/network.go
```diff
@@ -146,6 +147,7 @@ func (n Network) NodeDataDirs() []string {
 	for _, nodeDir := range n.nodeDirs {
 		directories = append(directories, n.getNodeFullPath(nodeDir))
 	}
+	sort.Strings(directories)
```
tsachiherman, line 150: this change ensures that the output of the method is consistent; it's not so by default since the `nodeDirs` is a map.

agreement/fuzzer/fuzzer_test.go
```diff
@@ -196,7 +196,7 @@ func (n *Fuzzer) initAccountsAndBalances(rootSeed []byte, onlineNodes []bool) er
 		if err != nil {
 			return err
 		}
-		n.accountAccessors[i*2+0] = rootAccess
+		n.accountAccessors[i] = rootAccess
```
tsachiherman, line 199: we used to store here both the rootkey + partkey, but then stopped using it for the part keys. As a result, we need only half of the size of the array.

data/account/participation.go
```diff
@@ -203,19 +202,24 @@ func (part PersistedParticipation) Persist() error {
 	voting := part.Voting.Snapshot()
 	rawVoting := protocol.Encode(&voting)
 
-	return part.Store.Atomic(func(ctx context.Context, tx *sql.Tx) error {
+	err := part.Store.Atomic(func(ctx context.Context, tx *sql.Tx) error {
 		err := partInstallDatabase(tx)
 		if err != nil {
-			return fmt.Errorf("Participation.persist: failed to install database: %v", err)
+			return fmt.Errorf("failed to install database: %w", err)
```
tolikzinovyev, line 208: this is nice

tolikzinovyev (approved):

data/account/participation.go
tsachiherman: I'm trying to figure if this was a sarcastic comment or not 😅

tolikzinovyev: No, it's not. :)  I've recently encountered a situation where `%w` would have been useful.

algonautshant (approved):
Looks great!

PR #2134, opened 2021-05-05
bugfix: netgoal errors when bootstrap files was missing

tsachiherman:
## Summary

netgoal was failing to build network when the bootstrap files was missing.

## Test Plan

test manually.

### DIFF
--- a/cmd/netgoal/network.go
+++ b/cmd/netgoal/network.go
@@ -142,7 +142,10 @@ func runBuildNetwork() (err error) {
 		net.GenesisData.VersionModifier = networkGenesisVersionModifier
 	}
 
-	bootstrappedFile := resolveFile(r.BootstrappedFile, templateBaseDir)
+	var bootstrappedFile string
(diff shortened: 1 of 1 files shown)


cmd/netgoal/network.go
```diff
@@ -201,7 +201,9 @@ func resolveFile(filename string, baseDir string) string {
 	if filepath.IsAbs(filename) {
 		return filename
 	}
-
```
algorandskiy, line 204: maybe put a condition at
```
if r.BootstrappedFile != "" {
    bootstrappedFile := resolveFile(r.BootstrappedFile, templateBaseDir)
    ...
}
``` ?

tsachiherman: sure.

algorandskiy (approved):

algonautshant (approved):
Yes. With this change, skip-dbfiles is not needed to get the old behavior.

PR #2135, opened 2021-05-06
logfilter: provide proper handling for panicing unit tests

tsachiherman:
## Summary

The logfilter utility was not handling unit test panics correctly - the panic result was omitted from the output. 

## Test Plan

Add a test case to verify the output is correct.


algonautshant (approved):
LGTM

PR #2136, opened 2021-05-06
Updating to fork from libsodium v1.0.18

braderhart:
## Summary

Updating to fork from libsodium v1.0.18. This will make it easier to track future releases from upstream and enables a potential conversion to a patch instead of keeping entire libsodium fork in main repo.

## Test Plan

Ran `make generate` and `make sanity`. Additional testing is requested as I had to make some modifications to run on Alpine Linux.


tsachiherman:
I agree with the idea behind this PR : upgrading to the latest would make it easier to maintain upstream compatibility.
However, if we were to adopt this change, we would need to prove that it's 100% compatible with the existing signature signing, verification and correctness. Past versions of libsodium were inconsistent ( between themselves ) in regards to the verification method (https://hdevalence.ca/blog/2020-10-04-its-25519am).
Without providing full set of tests the would prove that this change won't generate a regression, we cannot take this PR. Note that if you can point to an existing bug in the current implementation that affects `algod`, then we might want to port a fix for that.

braderhart:
@tsachiherman I'll see if I can figure out what validation tests I would need to write. Obviously no rush, but if there is anyone that can help with this it would be greatly appreciated.

tsachiherman:
I think that you picked a really hard challenge here - just the portion of "defining what a sufficient unit test for this change would be" is a pretty large task on its own.

braderhart:
From the article you referenced, it appears it would be important to have a set of validation criteria that could be tested against. If I reach out through email are you open to discuss more and determine if this would be a good issue to tackle now or if there may be other ways in which I can contribute for the time being?

tsachiherman:
Let me reach to few internal resources, and I'll give you an update next week.

tsachiherman:
We have an internal resource that is working on some refactoring of the libsodium. I don't know ( yet ) how this project would end up, so I'd like to put this ticket on hold for the time being.

braderhart:
Thanks for the update.

PR #2137, opened 2021-05-06
add support for more strict key registration transaction verification

tsachiherman:
## Summary

The existing key registration transaction verification allows quite much flexibility on what is written to the account data.
This PR adds few more constraints and would allow us to make certain assumptions regarding future key registration transactions.

Specifically, this adds the following checks:
1. checking that [VotePK/SelectionPK/VoteKeyDilution] are all set or all clear.
2. checking that the VoteFirst is less or equal to VoteLast
3. checking that in the case of going offline, both the VoteFirst and VoteLast are clear
4. checking that in the case of going online the VoteLast is non-zero and greater then the current network round.

## Test Plan

Unit tests added.


data/transactions/transaction_test.go
```diff
@@ -164,3 +164,258 @@ func TestWellFormedErrors(t *testing.T) {
...
+						}
+					}
+				}
+			}
+		}
+	}
+
+	curProto.SupportBecomeNonParticipatingTransactions = false
+	curProto.EnableKeyregCoherencyCheck = false
+	expectedErrors := []error{
```
nicholasguoalgorand, line 214: nit, make a for loop that populates expectedErrors instead of listing them

tsachiherman:
> One more test that might be useful:
> 
> * In case of going online, checking that VoteFirst is not after FirstValid (or is not after the current round - whichever is easier to implement)
> 
> What do you think?

I think that you're correct. The implementation itself would need to be a two-phase -
1. When checking the transaction construction itself ( before trying to match it to a specific round ), we need to ensure that the VoteFirst <= FirstValid+1.
2. When checking the transaction against a specific round, we need to ensure that VoteFirst <= <current voting round>+1.

The situation where someone can send a voting that would start in the future means that until that point in time, that account would be considered to be online, which would not be beneficial to the network.

config/consensus.go
```diff
@@ -351,6 +351,15 @@ type ConsensusParams struct {
...
 	// NoEmptyLocalDeltas updates how ApplyDelta.EvalDelta.LocalDeltas are stored
 	NoEmptyLocalDeltas bool
+
+	// EnableKeyregCoherencyCheck enable the following extra checks on key registration transactions:
+	// 1. checking that [VotePK/SelectionPK/VoteKeyDilution] are all set or all clear.
+	// 2. checking that the VoteFirst is less or equal to VoteLast.
+	// 3. checking that in the case of going offline, both the VoteFirst and VoteLast are clear.
+	// 4. checking that in the case of going online the VoteLast is non-zero and greater then the current network round.
+	// 5. checking that in the case of going online the VoteFirst is less or equal to the LastValid+1.
+	// 6. checking that in the case of going online the VoteFirst is less or equal to the <voting round>+1.
```
fabrice102, line 361: What is `<voting round>`?

tsachiherman: would probably be a good idea to clarify this further. in here, `<voting round>` is the round in which we attempt to apply the said transaction. ( i.e. the voting using the participation key itself can happen not before the round is accepted )

data/transactions/transaction.go
```diff
@@ -253,18 +261,48 @@ func (tx Transaction) WellFormed(spec SpecialAddresses, proto config.ConsensusPa
...
+			// The trio of [VotePK, SelectionPK, VoteKeyDilution] needs to be all zeros or all non-zero for the transaction to be valid.
+			if !((tx.KeyregTxnFields.VotePK == crypto.OneTimeSignatureVerifier{} && tx.KeyregTxnFields.SelectionPK == crypto.VRFVerifier{} && tx.KeyregTxnFields.VoteKeyDilution == 0) ||
+				(tx.KeyregTxnFields.VotePK != crypto.OneTimeSignatureVerifier{} && tx.KeyregTxnFields.SelectionPK != crypto.VRFVerifier{} && tx.KeyregTxnFields.VoteKeyDilution != 0)) {
+				return errKeyregTxnNonCoherentVotingKeys
+			}
+
+			// if it's a going offline transaction
+			if tx.KeyregTxnFields.VoteKeyDilution == 0 {
+				// check that we don't have any VoteFirst/VoteLast fields.
+				if tx.KeyregTxnFields.VoteFirst != 0 || tx.KeyregTxnFields.VoteLast != 0 {
```
brianolson, line 279: (VoteFirst == 0 && VoteLast == 0) is incompatible with the above (VoteFirst > VoteLast)

data/transactions/transaction.go
```diff
@@ -242,6 +242,14 @@ func (tx Transaction) MatchAddress(addr basics.Address, spec SpecialAddresses) b
 	return false
 }
 
+var errKeyregTxnFirstVotingRoundGreaterThanLastVotingRound = fmt.Errorf("transaction first voting round need to be less than its last voting round")
```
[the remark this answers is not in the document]
tsachiherman: yup.

data/transactions/transaction.go
tsachiherman: why ? for going offline transaction, `tx.KeyregTxnFields.VoteFirst > tx.KeyregTxnFields.VoteLast` is false for (0,0).
and we still want to check that both VoteFirst and VoteLast are explicitly zero.

algorandskiy (approved):

nicholasguoalgorand (approved):

PR #2139, opened 2021-05-07
Teal4 math

jannotti:
## Summary

Add a lot of math oriented opcodes that have been requested by teal users.

`sqrt`, `exp`, `expw`, `divmodw`

byteslice oriented math - a set of opcodes that will treat byteslices as arbitrary precision uints.

## Test Plan

Unit tests

Byte math costs were established with a bunch of new operation benchmarks.


data/transactions/logic/doc.go
```diff
@@ -111,6 +118,23 @@ var opDocByName = map[string]string{
...
+	"b<":  "A is less than B, where A and B are byteslices interpreted as big-endian unsigned integers => { 0 or 1}",
+	"b>":  "A is greater than B, where A and B are byteslices interpreted as big-endian unsigned integers => { 0 or 1}",
+	"b<=": "A is less than or equal to B, where A and B are byteslices interpreted as big-endian unsigned integers => { 0 or 1}",
+	"b>=": "A is greater than or equal to B, where A and B are byteslices interpreted as big-endian unsigned integers => { 0 or 1}",
+	"b==": "A is equals to B, where A and B are byteslices interpreted as big-endian unsigned integers => { 0 or 1}",
+	"b!=": "A is not equal to B, where A and B are byteslices interpreted as big-endian unsigned integers => { 0 or 1}",
+	"b%":  "A modulo B, where A and B are byteslices interpreted as big-endian unsigned integers. Panic if B is zero.",
+	"b|":  "A bitwise-or B, where A and B are byteslices interpreted as big-endian unsigned integers.",
+	"b&":  "A bitwise-and B, where A and B are byteslices interpreted as big-endian unsigned integers.",
+	"b^":  "A bitwise-xor B, where A and B are byteslices interpreted as big-endian unsigned integers.",
```
jasonpaulos, line 137: Why not have `b~` for bitwise invert as well?

jannotti: Saying all this outloud led to a better way.  bitwise operation no longer truncate to the shortest byteslice that can hold the value.  They return a byteslice as long as their longest input.  This is more likely to be what people want, as they would use these operations when treating the slices as arrays of bits.  It also makes the definition of b~ more straightforward.

data/transactions/logic/eval.go
```diff
@@ -911,39 +938,18 @@ func opLt(cx *evalContext) {
...
-	prev := last - 1
-	cond := cx.stack[prev].Uint > cx.stack[last].Uint
-	if cond {
-		cx.stack[prev].Uint = 1
-	} else {
-		cx.stack[prev].Uint = 0
-	}
-	cx.stack = cx.stack[:last]
+	opSwap(cx)
+	opLt(cx)
```
algorandskiy, line 942: I thought entire point of duplicating code in each opcode was in reducing number of additional operations.
Like `opLe` now is `opSwap + opLt + opNot` that is (5 + 6 + 3) = 14 instead of 6 in old version (counted mem read/write, swap is expensive since no direct mem to mem copy in a single cpu operation).
IMO this is questionable, maybe benchmark it?

data/transactions/logic/eval_test.go
```diff
@@ -598,243 +606,52 @@ func TestDivZero(t *testing.T) {
...
-	}
+	testPanics(t, "intc_1", 1)
 }
 
 func TestBytecTooFar(t *testing.T) {
 	t.Parallel()
-	for v := uint64(1); v <= AssemblerMaxVersion; v++ {
-		t.Run(fmt.Sprintf("v=%d", v), func(t *testing.T) {
-			ops, err := AssembleStringWithVersion(`bytec_1
-btoi`, v)
```
algorandskiy, line 820: is `btoi` not needed in the test?

data/transactions/logic/eval_test.go
```diff
@@ -4145,3 +3735,157 @@ main:
...
+}
+
+func TestLog2(t *testing.T) {
+	t.Parallel()
+	testAccepts(t, "int 0; log2; int 0; ==", 4)
+	testAccepts(t, "int 1; log2; int 1; ==", 4)
+	testAccepts(t, "int 2; log2; int 2; ==", 4)
+	testAccepts(t, "int 4; log2; int 3; ==", 4)
+	testAccepts(t, "int 5; log2; int 3; ==", 4)
+	testAccepts(t, "int 8; log2; int 4; ==", 4)
```
algorandskiy, line 3810: I though log2(8) = 3

data/transactions/logic/eval_test.go
```diff
@@ -4145,3 +3735,157 @@ main:
...
+}
+
+func TestBytesMath(t *testing.T) {
+	t.Parallel()
+	testAccepts(t, "byte 0x01; byte 0x01; b+; byte 0x02; ==", 4)
+	testAccepts(t, "byte 0x01FF; byte 0x01; b+; byte 0x0200; ==", 4)
+
+	effs := strings.Repeat("ff", 64)
+	// 64 byte long inputs are accepted, even if they produce longer outputs
```
algorandskiy, line 3819: check two 64-bytes long inputs?

algorandskiy (commented):
Nice work, couple comments

data/transactions/logic/eval.go
```diff
@@ -584,7 +587,31 @@ func (cx *evalContext) step() {
 		cx.err = fmt.Errorf("pc=%3d dynamic cost budget of %d exceeded, executing %s", cx.pc, cx.budget(), spec.Name)
 		return
 	}
+
+	preheight := len(cx.stack)
 	spec.op(cx)
+
+	if cx.err == nil {
+		postheight := len(cx.stack)
+		if spec.Name != "return" && postheight-preheight != len(spec.Returns)-len(spec.Args) {
```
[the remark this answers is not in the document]
jannotti: This error would only occur because of a bug in our implementation, not in teal code.  I don't know how user-friendly it needs to be.  The info *we* need is there:  There's a bug in the named opcode and it didn't manage the stack properly.

data/transactions/logic/eval.go
jannotti: I did this when I found writing the equivalent byteslice operators to be tedious, repetitive, and error-prone.   It's hard for me to imagine any appreciable performance effect here, with all this data in cache.  I'll write a little benchmark to convince myself.

data/transactions/logic/eval_test.go
jannotti: Yikes, yes, thank you!  Off by one everywhere.

data/transactions/logic/eval_test.go
jannotti: Good catch.  The code still panics without it, but it's a better test with it in.  We want the program to have be plausibly accepted, so we're really testing that reading the bytec_1 fails.  In fact, I think "int 1; return" is an even better way to finish, so that we have the best possible change of accepting if bytec_1 doesn't fail.

data/transactions/logic/eval_test.go
jannotti: That's what that test does.  `effs` if 64 repeated "ff"s.  So when it gets sprintf'd into the string, you get a 64 byte, all ones, byte array.

algorandskiy: I must be missed the last line in this test, thanks.

data/transactions/logic/eval_test.go
jannotti: The reason I made this error is that integer log2 is just "what's the highest bit set in this integer (minus 1)".

I changed this so it's "correct", and fails when the supplied integer is 0.  But I wonder if I should just rename it "highbit" and give it the old semantics.  That sounds clearer to me, and is probably what people would actually need this opcode for (so they can decide how far to bitshift a value, for example).  That's why this is an operation that's supported by the bits package.  With the log2 semantics, callers need to check if the value is 0 first, to avoid panics.

data/transactions/logic/eval.go
jannotti: Here are the result.  I put them in order of overhead introduced by the shorter, clearer code. Each one is on more function call than the last.

1. Benchmark <          16.9-17.8 ns/op
2. Benchmark >=         17.3-17.5 ns/op
3. Benchmark >          20.0-23.3 ns/op
4. Benchmark <=         21.8-22.0 ns/op


3 ns seems like an ok price to pay, for the largest overhead. 

The biggest jump is between 2 & 3, because it involves opSwap.  I suppose I could be convinced to repeat the code to eliminate that one.  Then they'd all be so close the difference is in the noise.

algorandskiy: Thank you for benchmarking it! I think I agree, 3ns/op is OK price for cleaner code.

data/transactions/logic/eval.go
```diff
@@ -584,7 +587,31 @@ func (cx *evalContext) step() {
 		cx.err = fmt.Errorf("pc=%3d dynamic cost budget of %d exceeded, executing %s", cx.pc, cx.budget(), spec.Name)
 		return
 	}
+
+	preheight := len(cx.stack)
 	spec.op(cx)
+
+	if cx.err == nil {
```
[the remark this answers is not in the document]
jannotti: I don't think so.  It only checks things that would have previously caused an error.  But now it does it in a central place, so that future opcodes won't have to do it themselves.  In particular, in the past only concat could create a 4k slice, so only concat checked its output.  Now we check here, in case we have introduced a new way to create a 4k slice (I don't think we have, since byteslice math is input limited, but this seems like a useful thing to centralize, especially as we might add other ways to hit 4k)

data/transactions/logic/eval.go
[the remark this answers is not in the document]
algorandskiy: I simply do not remember checks for popped vs pushed. But I guess you are right and these problem would be already discovered.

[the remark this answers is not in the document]
jannotti: Yes, that's true, those checks are new.  But they are checks on the results of individual opcode operations, so I don't think it is possible to construct a strange program that would trip them.  If they catch a problem, it would mean that one of our existing opcodes has a pretty important bug that we have missed for a long time.  I think that's unlikely.  But it does help mistakes as we write new opcodes.

data/transactions/logic/eval_test.go
jannotti: I decided to call the op "bitlen" and define it on uints and byte-arrays.  It returns the index of the highest bit, so it returns 0 on 0, rather than panic. (And 1 => 1, 2 => 2, 4=> 3, 8 => 4). It corresponds directly to bits.Len in Go.

jasonpaulos (approved):
This looks good to me, including the `teal.tmLanguage.json` file.

data/transactions/logic/eval.go
```diff
@@ -1085,6 +1072,355 @@ func opBitNot(cx *evalContext) {
...
+		cx.err = errors.New("byte math would have negative result")
+		return
+	}
+	cx.stack[prev].Bytes = result.Bytes()
+	cx.stack = cx.stack[:last]
+}
+
+func opBytesPlus(cx *evalContext) {
+	result := new(big.Int)
+	opBytesBinOp(cx, result, result.Add)
```
algorandskiy, line 1257: clever

data/transactions/logic/eval.go
```diff
@@ -1085,6 +1072,355 @@ func opBitNot(cx *evalContext) {
...
+	cx.stack[prev].Bytes = fresh
+	cx.stack = cx.stack[:last]
+	return fresh, other
+}
+
+func opBytesBitOr(cx *evalContext) {
+	a, b := opBytesBinaryLogicPrep(cx)
+	for i := range a {
+		a[i] = a[i] | b[i]
+	}
```
algorandskiy, line 1387: another nice one

algorandskiy (approved):
Looks good. Please consider adding one more test for all wide opcodes in a sequence to ensure format compatibility - like mulw, addw, divmodw, and something like (A * B / B == A) for mulw + divmodw. A seprate PR is OK.

algorandskiy:
@brianolson please take a look

data/transactions/logic/eval_test.go
```diff
@@ -3187,31 +2819,161 @@ func benchmarkBasicProgram(b *testing.B, source string) {
...
+			benchmarkOperation(b, bench[1], bench[2], bench[3]);
+		})
+	}
+}
+
+func BenchmarkHash(b *testing.B) {
+	hashes := []string{"sha256", "keccak256", "sha512_256"}
+	for _, hash := range hashes {
+		b.Run(hash+"-small", func(b *testing.B) { // hash 32 bytes
+			benchmarkOperation(b, "byte 0x1234567890", hash,
```
brianolson, line 2958: nit: for each of these its only hashing 5 bytes on the first execution, and then {32,128,512} for the next 1999. It could be perfectly consistent by using a 32 byte constant and running `dup; concat; dup; concat; {hash}` as the repeated op?

brianolson (commented):
one tiny testing quibble, but the 0.05% change probably doesn't matter.
Otherwise no objections. LGTM.

onetechnical:
Brian's comments to be addressed in a separate PR.

PR #2141, opened 2021-05-07
remerge v2.5.6-stable

onetechnical:
## Summary

Finishing re-merge of v2.5.6-stable release.

## Test Plan

Standard integration testing. Changes should commits only, no file changes.


algobarb:
Why do we do this if files don't change? Is it to keep the commit history or something?

onetechnical:
> Why do we do this if files don't change? Is it to keep the commit history or something?

Yes. It ensures that the branches do not diverge and share a lineage. If there are fixes downstream, they may also be captured here, though that is less common.

algobarb (approved):

egieseke (approved):

PR #2143, opened 2021-05-10
txsync: transaction sync profiler

tsachiherman:
## Summary

This PR adds a lightweight profiler to the transaction sync. The profiler measures the execution time of the various internal components in the transaction sync and provide feedback for their relative CPU consumption.

## Test Plan

Test output on Kibana

### DIFF
--- a/txnsync/profiler.go
+++ b/txnsync/profiler.go
@@ -66,8 +66,6 @@ type profiler struct {
 	profileSum time.Duration
 	// profileSpan is the max span of the array ( or - the window ) that we would like to maintain.
 	profileSpan time.Duration
-	// spanReached flag is used to indicate whether we've already reached the window or not.
-	spanReached bool
 	// lastProfileLog is the last time we've logged to the telemetry.
 	lastProfileLog time.Duration
 	// logInterval defines what is the frequency at which we send an event to the telemetry. Zero to disable.
@@ -137,14 +135,19 @@ func (p *profiler) prune() {
 
 		p.profile = p.profile[1:]
 		e.times = e.times[1:]
-		p.spanReached = true
 	}
 }
 
 func (p *profiler) maybeLogProfile() {
-	if !p.spanReached || p.logInterval == 0 {
+	// do we have the log profile enabled ?
+	if p.logInterval == 0 {
 		return
 	}
+	// do we have enough samples ? ( i.e. at least 50% sample time )
+	if p.profileSum < p.profileSpan/2 {
+		return
+	}
+	// have we send metrics recently ?
 	curTime := p.clock.Since()
 	if curTime-p.lastProfileLog <= p.logInterval {
 		return
--- a/txnsync/profiler.go
+++ b/txnsync/profiler.go
@@ -30,7 +30,7 @@ type profElements int
 const (
 	profElementIdle = iota
 	profElementTxChange
-	profElementNewRounnd
+	profElementNewRound
 	profElementPeerState
 	profElementIncomingMsg
 	profElementOutgoingMsg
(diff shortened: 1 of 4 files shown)


txnsync/profiler.go
```diff
@@ -0,0 +1,205 @@
...
+)
+
+// The profiler struct provides profiling information regarding the main loop performance
+// characteristics. Using it provides statistics information about the recent duty cycle utilization,
+// that could be used when trying to throlle the accuracy and performance of the transaction sync.
+type profiler struct {
+	// clock used as the source clock for measurements.
+	clock timers.WallClock
+	// elements contains the elements we want to measure.
+	elements []*element
```
tolikzinovyev, line 57: I'm guessing the main purpose of `elements` is to reduce the amount of boilerplate code. How about having multiple public variables here for each element and having `profile` contain pointers to one of the elements? Then `prune` can be as concise as it is now and the user of `profiler` can also access elements without any extra code. I think it will be just as concise.

My main problem with `elements` is that it makes readability worse. It took me some time to understand the code.

tolikzinovyev (commented):
It would be nice to have unit tests.

txnsync/profiler.go
```diff
@@ -0,0 +1,205 @@
...
+		e.total -= dt
+		if !e.detached {
+			p.profileSum -= dt
+		}
+
+		p.profile = p.profile[1:]
+		e.times = e.times[1:]
+		p.spanReached = true
+	}
+	p.logProfile()
```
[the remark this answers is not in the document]
tsachiherman: sounds reasonable to me.

txnsync/profiler.go
```diff
@@ -0,0 +1,205 @@
...
+	log logging.Logger
+
+	// profile contains all the elements indices, in order of arrival. It allows us to maintain a moving window.
+	profile []int
+	// profileSum is the total amount of time tracked by the profile array.
+	profileSum time.Duration
+	// profileSpan is the max span of the array ( or - the window ) that we would like to maintain.
+	profileSpan time.Duration
+	// spanReached flag is used to indicate whether we've already reached the window or not.
+	spanReached bool
```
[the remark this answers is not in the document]
tsachiherman: no, it's not. The first time logProfile() is called, it's still false. It turns to true only once we have a "full window" of at least `profileSpan` worth of sampling time.
let's say that profileSpan = 200ms.
and let's say that we have 5 samples:
10ms -> it will still be false
150ms -> it will still be false
30ms -> it will still be false
20ms -> it will still be true
10ms -> it will still be true

so, only after the 4th sample it would set it to true.

txnsync/profiler.go
tsachiherman: I think that having all these defined on the object itself is quite an ugly implementation. I was trying to create a simpler struct interface, and enumerate the different elements using a simple constants instead.
Having said that, I do agree that we could do better with the documentation here so that it would be more clear why we chose to do that in this way. I'll try to improve the docs, but please don't be shy, and if you'd like to suggest a specific description, please let me know.

txnsync/profiler.go
```diff
@@ -0,0 +1,213 @@
...
+		IncomingMsgOps:               uint64(len(p.elements[profElementIncomingMsg].times)),
+		OutgoingMsgOps:               uint64(len(p.elements[profElementOutgoingMsg].times)),
+		NextOffsetOps:                uint64(len(p.elements[profElementNextOffset].times)),
+		GetTxnGroupsOps:              uint64(len(p.elements[profElementGetTxnsGroups].times)),
+		AssembleMessageOps:           uint64(len(p.elements[profElementAssembleMessage].times)),
+		SendMessageOps:               uint64(len(p.elements[profElementSendMessage].times)),
+		MakeBloomFilterOps:           uint64(len(p.elements[profElementMakeBloomFilter].times)),
+		SelectPendingTransactionsOps: uint64(len(p.elements[profElementTxnsSelection].times)),
+
+		TotalDuration:                    time.Duration(p.profileSum),
```
tolikzinovyev, line 175: conversion is not needed

txnsync/profiler.go
```diff
@@ -0,0 +1,213 @@
...
+func (p *profiler) maybeLogProfile() {
+	// do we have the log profile enabled ?
+	if p.logInterval == 0 {
+		return
+	}
+	// do we have enough samples ? ( i.e. at least 50% sample time )
+	if p.profileSum < p.profileSpan/2 {
+		return
+	}
+	// have we send metrics recently ?
```
tolikzinovyev, line 150: nit: sent

txnsync/profiler.go
tolikzinovyev: I guess it's not important right now. When this code needs to be merged to master, it can be revisited.

tolikzinovyev (approved):

PR #2144, opened 2021-05-11
Fix Update Script

algobarb:
## Summary

A change in 2.5.4 broke the upgrade behavior. When a node is upgraded with the new scripts, the algod -v updates, but the node doesn't update properly when you check `curl http://$(cat data/node1/algod.net)/versions`.

## Test Plan

I tested on AWS ubuntu machines with the go-algorand source code and building the binaries with the installer/algorand@.service.template change. I ran one with the change and one without the change for 2.5.4.

### Steps taken
```
sudo snap install go --classic
sudo apt-get install build-essential
sudo apt install make
git clone https://github.com/algorand/go-algorand.git
cd ~/go-algorand && git checkout tags/v2.5.4-stable -b feature/fixupdate
cd ~/go-algorand && ./scripts/configure_dev.sh
 make install
cd /home/ubuntu/go/bin
export ALGORAND_DATA=~/.algorand
./goal node start -d ~/.algorand
./update.sh -i -c stable-254 -p . -d ~/.algorand -n
sudo ./systemd-setup.sh ubuntu ubuntu
sudo systemctl daemon-reload
export ALGORAND_DATA=~/.algorand
sudo systemctl start algorand@$(systemd-escape $ALGORAND_DATA)


./update.sh -i -c stable-254 -p . -d ~/.algorand -n
curl http://$(cat ~/.algorand/algod.net)/versions # 2.5.4
./algod -v # 2.5.4

./update.sh -u -c stable-tmp -p . -d ~/.algorand
./algod -v # becomes latest 2.5.6 
curl http://$(cat ~/.algorand/algod.net)/versions # remains 2.5.4 without the execstart change but updates to 2.5.6 with the new change

```
New Steps
```
sudo snap install go --classic
sudo apt-get install build-essential
sudo apt install make
git clone https://github.com/algorand/go-algorand.git
cd ~/go-algorand && git checkout tags/v2.5.4-stable -b feature/fixupdate

# Make changes and in cmd/updater/update.sh add the following
# cp /home/ubuntu//go-algorand/cmd/updater/update.sh "${UPDATESRCDIR}/bin/${FILENAME}"
# add it above the following line so that the update script doesn't get overwritten while testing.:
# exec "${UPDATESRCDIR}/bin/${FILENAME}" ${INSTALLOPT} -r -c ${CHANNEL} ${DATADIRSPEC} ${NOSTART} ${BINDIRSPEC} ${HOSTEDSPEC} ${GENESIS_NETWORK_DIR_SPEC} ${UNKNOWNARGS[@]}

cd ~/go-algorand && ./scripts/configure_dev.sh
 make install
cd /home/ubuntu/go/bin
export ALGORAND_DATA=~/.algorand
sudo ./systemd-setup.sh ubuntu ubuntu
sudo systemctl daemon-reload
export ALGORAND_DATA=~/.algorand
sudo systemctl start algorand@$(systemd-escape $ALGORAND_DATA)

curl http://$(cat ~/.algorand/algod.net)/versions # 2.5.4
./algod -v # 2.5.4

./update.sh -u -c stable-tmp -p . -d ~/.algorand
./algod -v # becomes latest 2.5.6 
curl http://$(cat ~/.algorand/algod.net)/versions # remains 2.5.4 without the execstart change but updates to 2.5.6 with the new change
```
Future Improvement: 
Create a regular test that could test this scenario and alert for issue.

### DIFF
--- a/cmd/updater/update.sh
+++ b/cmd/updater/update.sh
@@ -393,8 +393,9 @@ function shutdown_node() {
         if [ -f "${BINDIR}/goal" ]; then
             for DD in "${DATADIRS[@]}"; do
                 if [ -f "${DD}/algod.pid" ] || [ -f "${DD}"/**/kmd.pid ] ; then
-                    echo "Stopping node with data directory ${DD} and waiting..."
+                    # Before shutting down the node, we have to shut down the systemctl, otherwise it will spin up a node with the old binary
                     run_systemd_action stop "${DD}"
+                    sleep 5
                     "${BINDIR}/goal" node stop -d "${DD}"
                     sleep 5
                 else
@@ -626,7 +627,7 @@ function apply_fixups() {
 # and that it's a valid directory.
 # Unless it's an install
 if [ ! -d "${BINDIR}" ]; then
-    if [ "${UPDATETYPE}" = "install" ]; then
+    if [ "${UPDATETYPE}" == "install" ]; then
         mkdir -p "${BINDIR}"
     else
         fail_and_exit "Missing or invalid binaries path specified '${BINDIR}'"
@@ -645,7 +646,7 @@ fi
 if [ ${RESUME_INSTALL} -eq 0 ] && ! $DRYRUN; then
     validate_channel_specified
 
-    if [ "${UPDATETYPE}" = "migrate" ]; then
+    if [ "${UPDATETYPE}" == "migrate" ]; then
         download_update_for_current_version
     else
         check_and_download_update
@@ -680,6 +681,7 @@ else
 fi
 
 # Shutdown node before backing up so data is consistent and files aren't locked / in-use.
+
 shutdown_node
 
 if ! $DRYRUN; then
@@ -721,5 +723,4 @@ if [ "${NOSTART}" != "" ]; then
 else
     startup_nodes
 fi
-
 exit 0
--- a/cmd/updater/update.sh
+++ b/cmd/updater/update.sh
@@ -601,8 +601,6 @@ function apply_fixups() {
     done
 }
 
-
-
 #--------------------------------------------
 # Main Update Driver
 
--- a/cmd/updater/update.sh
+++ b/cmd/updater/update.sh
@@ -374,6 +374,11 @@ function run_systemd_action() {
                 echo "systemd system service: $action"
                 return 0
             fi
+        else
+            if sudo -n systemctl "$action" "algorand@$(systemd-escape "$data_dir")"; then
+                echo "systemd system service: $action"
+                return 0
+            fi
         fi
     fi
 
@@ -563,16 +568,16 @@ function startup_node() {
     fi
 
     CURDATADIR=$1
-    echo Starting node in ${CURDATADIR}...
+    echo Restarting node in ${CURDATADIR}...
 
     check_install_valid
     if [ $? -ne 0 ]; then
         fail_and_exit "Installation does not appear to be valid"
     fi
 
-    if ! run_systemd_action start "${CURDATADIR}"; then
+    if ! run_systemd_action restart "${CURDATADIR}"; then
         echo "No systemd services, starting node with goal."
-        ${BINDIR}/goal node start -d "${CURDATADIR}" ${HOSTEDFLAG}
+        ${BINDIR}/goal node restart -d "${CURDATADIR}" ${HOSTEDFLAG}
     fi
 }
 
@@ -665,10 +670,12 @@ if [ ${RESUME_INSTALL} -eq 0 ] && ! $DRYRUN; then
     # Note that the SCRIPTPATH we're passing in should be our binaries directory, which is what we expect to be
     # passed as the last argument (if any)
     echo "Starting the new update script to complete the installation..."
+    echo "copy the update script"
+    cp /home/ubuntu//go-algorand/cmd/updater/update.sh "${UPDATESRCDIR}/bin/${FILENAME}"
     exec "${UPDATESRCDIR}/bin/${FILENAME}" ${INSTALLOPT} -r -c ${CHANNEL} ${DATADIRSPEC} ${NOSTART} ${BINDIRSPEC} ${HOSTEDSPEC} ${GENESIS_NETWORK_DIR_SPEC} ${UNKNOWNARGS[@]}
 
     # If we're still here, exec failed.
-    fail_and_exit "Error executing the new update script - unable to continue"
+    # fail_and_exit "Error executing the new update script - unable to continue"
 else
     # We're running the script from our expanded update, which is located in the last script's ${TEMPDIR}/a/bin
     # We need to define our TEMPDIR and UPDATESRCDIR to match those values; we do so by making them relative
--- a/cmd/updater/update.sh
+++ b/cmd/updater/update.sh
@@ -552,7 +552,7 @@ function startup_node() {
     fi
 
     if ! run_systemd_action restart "${CURDATADIR}"; then
-        echo "No systemd services, starting node with goal."
+        echo "No systemd services, restarting node with goal."
         ${BINDIR}/goal node restart -d "${CURDATADIR}" ${HOSTEDFLAG}
     fi
 }
--- a/cmd/updater/update.sh
+++ b/cmd/updater/update.sh
@@ -392,31 +392,6 @@ function run_systemd_action() {
     return 1
 }
 
-function shutdown_node() {
-    echo Stopping node...
-    if [ "$(pgrep -x algod)" != "" ] || [ "$(pgrep -x kmd)" != "" ] ; then
-        if [ -f "${BINDIR}/goal" ]; then
-            for DD in "${DATADIRS[@]}"; do
-                if [ -f "${DD}/algod.pid" ] || [ -f "${DD}"/**/kmd.pid ] ; then
-                    # Before shutting down the node, we have to shut down the systemctl, otherwise it will spin up a node with the old binary
-                    run_systemd_action stop "${DD}"
-                    sleep 5
-                    "${BINDIR}/goal" node stop -d "${DD}"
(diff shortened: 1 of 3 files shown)


installer/algorand@.service.template
```diff
@@ -30,7 +30,7 @@ After=network.target
 AssertPathExists=%I
 
 [Service]
-ExecStart=@@BINDIR@@/algod -d %I
+ExecStart=/bin/bash -c "exec /%I/../../algod -d %I"
```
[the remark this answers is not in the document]
bricerisingalgorand: Hmm that's interesting, I'll see if I can do that too

bricerisingalgorand (dismissed):

egieseke (dismissed):

algobarb:
After troubleshooting it more with John this afternoon, we narrowed down that `systemctl start/stop/restart` issue is due to an existing `algod.lock` file. From there we realized that the update script was not stopping systemctl properly before running goal stop node. This means that when the node stops, systemctl tries to spin up another algod with the old binary even if the binaries have been updated.

Taking a look at the code, the `systemctl stop` command is called before the `goal node stop` command; however, we need to give it some extra time for the `systemctl stop` to execute. Therefore, I added an extra `sleep 5` on line 398.

algobarb:
Troubleshooting today revealed that we need to ensure that systemctl is shut down properly when we are not the root user by using sudo. 

We also need to restart systemctl at the end rather than just a simple start so that it uses the new binaries.

From initial testing, it looks like this works. I could look into cleaning up the stopping node commands.

cmd/updater/update.sh
```diff
@@ -374,6 +374,11 @@ function run_systemd_action() {
                 echo "systemd system service: $action"
                 return 0
             fi
+        else
+            if sudo -n systemctl "$action" "algorand@$(systemd-escape "$data_dir")"; then
+                echo "systemd system service: $action"
```
egieseke, line 379: Should this command be with sudo, `sudo systemd system service: $action`

cmd/updater/update.sh
```diff
@@ -664,10 +645,12 @@ if [ ${RESUME_INSTALL} -eq 0 ] && ! $DRYRUN; then
     # Note that the SCRIPTPATH we're passing in should be our binaries directory, which is what we expect to be
     # passed as the last argument (if any)
     echo "Starting the new update script to complete the installation..."
+    echo "copy the update script"
+    cp /home/ubuntu//go-algorand/cmd/updater/update.sh "${UPDATESRCDIR}/bin/${FILENAME}"
     exec "${UPDATESRCDIR}/bin/${FILENAME}" ${INSTALLOPT} -r -c ${CHANNEL} ${DATADIRSPEC} ${NOSTART} ${BINDIRSPEC} ${HOSTEDSPEC} ${GENESIS_NETWORK_DIR_SPEC} ${UNKNOWNARGS[@]}
 
     # If we're still here, exec failed.
-    fail_and_exit "Error executing the new update script - unable to continue"
+    # fail_and_exit "Error executing the new update script - unable to continue"
```
egieseke, line 653: Can this be removed, or should there be an error message at this point?

cmd/updater/update.sh
```diff
@@ -721,5 +701,4 @@ if [ "${NOSTART}" != "" ]; then
 else
     startup_nodes
 fi
-
-exit 0
+exit 0
```
egieseke, line 704: Add new line

cmd/updater/update.sh
```diff
@@ -664,10 +645,12 @@ if [ ${RESUME_INSTALL} -eq 0 ] && ! $DRYRUN; then
     # Note that the SCRIPTPATH we're passing in should be our binaries directory, which is what we expect to be
     # passed as the last argument (if any)
     echo "Starting the new update script to complete the installation..."
+    echo "copy the update script"
+    cp /home/ubuntu//go-algorand/cmd/updater/update.sh "${UPDATESRCDIR}/bin/${FILENAME}"
```
egieseke, line 649: Is ${FILENAME} required for the target of the cp command?

egieseke (commented):
Added some questions.

cmd/updater/update.sh
onetechnical: While not affecting functionality, it would be nice in the debug to indicate it was run with sudo.

cmd/updater/update.sh
onetechnical: I think this is debug code that should be removed. (I believe it was to local copy the update.sh script, but should not be done in final code.)

cmd/updater/update.sh
onetechnical: I think this should be restored.

onetechnical (changes requested):

cmd/updater/update.sh
algobarb: Yes sorry, will restore. Thank you for catching it

cmd/updater/update.sh
algobarb: Sorry I couldn't actually use my branch while debugging since I was testing with 2.5.4 code. I must've copy and pasted some wrong changes into the file as I was working on the "not to commit but use to test" and "to-commit" files.

algobarb: Will remove debugging code

cmd/updater/update.sh
algobarb: done

onetechnical (changes requested):
If we strip shutdown, we should also remove the RESTART_NODE var.

Note that I also see a separate issue regarding backup files. While not in the initial scope of this, we should fix it as well.

onetechnical (dismissed):
This looks fine to me, provided testing was fine. Any thoughts on if we should include https://github.com/algobarb/go-algorand/pull/2 ?

cmd/updater/update.sh
```diff
@@ -387,30 +392,6 @@ function run_systemd_action() {
...
-                    # Clean up zombie (algod|kmd).net files
-                    rm -f "${DD}/algod.net" "${DD}"/**/kmd.net
-                fi
-            done
-        fi
-    else
-        echo ... node not running
-    fi
-
-    RESTART_NODE=1
```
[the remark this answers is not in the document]
algobarb: Moved RESTART_NODE=1  to where shutdown_node was previously in the script.

algobarb:
> This looks fine to me, provided testing was fine. Any thoughts on if we should include [algobarb#2](https://github.com/algobarb/go-algorand/pull/2) ?
Testing took a few tries on Travis, but it passed. I can add in the Datadirs one as well.

cmd/updater/update.sh
```diff
@@ -367,17 +367,23 @@ function run_systemd_action() {
     local data_dir=$2
     local process_owner
 
+    # If the service is system-level, check if it's root or sudo
     if check_service system "$data_dir"; then
         process_owner=$(awk '{ print $1 }' <(ps aux | grep "[a]lgod -d ${data_dir}"))
-        if $IS_ROOT || grep sudo <(groups "$process_owner" &> /dev/null); then
+        if $IS_ROOT; then
             if systemctl "$action" "algorand@$(systemd-escape "$data_dir")"; then
                 echo "systemd system service: $action"
```
egieseke, line 375: Should the echo reflect the full contents of the command above?

cmd/updater/update.sh
```diff
@@ -367,17 +367,23 @@ function run_systemd_action() {
...
                 return 0
             fi
+        elif grep sudo <(groups "$process_owner" &> /dev/null); then
+            if sudo -n systemctl "$action" "algorand@$(systemd-escape "$data_dir")"; then
+                echo "sudo -n systemd system service: $action"
+                return 0
```
egieseke, line 381: Same here for echo.

onetechnical (approved):
Still looks good. We should test with/without the hosted flag too.

cmd/updater/update.sh
```diff
@@ -562,16 +544,16 @@ function startup_node() {
...
+    echo Restarting node in ${CURDATADIR}...
 
     check_install_valid
     if [ $? -ne 0 ]; then
         fail_and_exit "Installation does not appear to be valid"
     fi
 
-    if ! run_systemd_action start "${CURDATADIR}"; then
+    if ! run_systemd_action restart "${CURDATADIR}"; then
         echo "No systemd services, starting node with goal."
```
egieseke, line 555: Update echo to say "restarting".

egieseke (dismissed):

cmd/updater/update.sh
algobarb: ok

cmd/updater/update.sh
algobarb: I'm not sure if it should or not? It might make more sense to.

cmd/updater/update.sh
```diff
@@ -695,15 +677,15 @@ if ! $DRYRUN; then
...
+    for DD in ${DATADIRS[@]}; do
         if ! install_new_data "${DD}"; then
             fail_and_exit "Error installing data files into ${DD}"
         fi
     done
 
     copy_genesis_files
 
-    for DD in "${DATADIRS[@]}"; do
+    for DD in ${DATADIRS[@]}; do
```
bricerisingalgorand, line 688: Does this work without the quotes? I'm not positive bash will interpret the bash array correctly without them

algobarb: This was John's change. His PR notes are here: https://github.com/algobarb/go-algorand/pull/2
I'm running it again to make sure this works as expected.

bricerisingalgorand: looks like they don't matter actually in a simple example, so i think this should be ok. Removing them might be unnecessary risk unless you had a reason to remove them though.

onetechnical (approved):

PR #2147, opened 2021-05-12
Increase ASA URL size 

jdtzmn:
## Summary

Increase the ASA URL size from a limiting 32 bytes to 96 bytes to support NFTs with longer IPFS URLs. 

Closes algorand/go-algorand-internal#1179.

## Test Plan

Integration tests for before and after the change.

### DIFF
--- a/test/scripts/e2e_subs/asset-misc.sh
+++ b/test/scripts/e2e_subs/asset-misc.sh
@@ -18,7 +18,10 @@ ACCOUNTD=$(${gcmd} account new|awk '{ print $6 }')
 
 ASSET_NAME='Birlot : décollage vs. ࠶🦪'
 
-${gcmd} asset create --creator ${ACCOUNT} --name "${ASSET_NAME}" --unitname amisc --total 1000000000000
+# to ensure IPFS URLs longer than 32 characters are supported
+ASSET_URL="/ipfs/QmXoypizjW3WknFiJnKLwHCnL72vedxjQkDDP1mXWo6uco/wiki/"
+
+${gcmd} asset create --creator ${ACCOUNT} --name "${ASSET_NAME}" --unitname amisc --total 1000000000000 --asseturl "${ASSET_URL}"
 
 ASSET_ID=$(${gcmd} asset info --creator $ACCOUNT --unitname amisc|grep 'Asset ID'|awk '{ print $3 }')
 
--- a/test/scripts/e2e_subs/asset-misc.sh
+++ b/test/scripts/e2e_subs/asset-misc.sh
@@ -21,7 +21,7 @@ ASSET_NAME='Birlot : décollage vs. ࠶🦪'
 # to ensure IPFS URLs longer than 32 characters are supported
 ASSET_URL="/ipfs/QmXoypizjW3WknFiJnKLwHCnL72vedxjQkDDP1mXWo6uco/wiki/"
 
-${gcmd} asset create --creator ${ACCOUNT} --name "${ASSET_NAME}" --unitname amisc --total 1000000000000 --asseturl "${ASSET_URL}"
+${gcmd} asset create --creator "${ACCOUNT}" --name "${ASSET_NAME}" --unitname amisc --total 1000000000000 --asseturl "${ASSET_URL}"
 
 ASSET_ID=$(${gcmd} asset info --creator $ACCOUNT --unitname amisc|grep 'Asset ID'|awk '{ print $3 }')
 
--- a/test/scripts/e2e_subs/asset-misc.sh
+++ b/test/scripts/e2e_subs/asset-misc.sh
@@ -19,7 +19,7 @@ ACCOUNTD=$(${gcmd} account new|awk '{ print $6 }')
 ASSET_NAME='Birlot : décollage vs. ࠶🦪'
 
 # to ensure IPFS URLs longer than 32 characters are supported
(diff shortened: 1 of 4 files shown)


test/scripts/e2e_subs/asset-misc.sh
```diff
@@ -18,7 +18,10 @@ ACCOUNTD=$(${gcmd} account new|awk '{ print $6 }')
 
 ASSET_NAME='Birlot : décollage vs. ࠶🦪'
 
-${gcmd} asset create --creator ${ACCOUNT} --name "${ASSET_NAME}" --unitname amisc --total 1000000000000
+# to ensure IPFS URLs longer than 32 characters are supported
+ASSET_URL="/ipfs/QmXoypizjW3WknFiJnKLwHCnL72vedxjQkDDP1mXWo6uco/wiki/"
```
jannotti, line 22: Can you pad this out to 96 bytes?  We'd like it to fail if anyone ever gets the bad idea to shrink the limit, as that would break existing code (once this is out).

I realize that such an idea is so bad that it is unlikely to occur, but I still think that's what we should be testing.

test/scripts/e2e_subs/asset-misc.sh
```diff
@@ -18,7 +18,10 @@ ACCOUNTD=$(${gcmd} account new|awk '{ print $6 }')
...
+
+${gcmd} asset create --creator ${ACCOUNT} --name "${ASSET_NAME}" --unitname amisc --total 1000000000000 --asseturl "${ASSET_URL}"
```
jannotti, line 24: Consider installing "shellcheck" which would complain about the lack of quotes around ${ACCOUNT}.

Our codebase is definitely not totally compliant, but I try to be whenever adding code.

test/scripts/e2e_subs/v26/short-asset-url-only.sh
```diff
@@ -0,0 +1,28 @@
...
+
+gcmd="goal -w ${WALLET}"
+
+ACCOUNT=$(${gcmd} account list|awk '{ print $3 }')
+ASSET_NAME='Birlot : décollage vs. ࠶🦪'
+
+set +o pipefail
+# longer than 32-byte ASA URLs should fail
+${gcmd} asset create \
+    --creator ${ACCOUNT} \
```
jannotti, line 20: "${ACCOUNT}"

test/scripts/e2e_subs/v26/short-asset-url-only.sh
```diff
@@ -0,0 +1,28 @@
...
+    --name "${ASSET_NAME}" \
+    --unitname amisc \
+    --total 1000000000000 \
+    --asseturl "12345678901234567890123" \
+    | grep "is too long (max 32 bytes)"
+set -o pipefail
+
+date '+short-asset-url-only finish %Y%m%d_%H%M%S'
```
jannotti, line 28: Rainy day activity: Figure out the right set of random bash commands that would produce the name of the script without the .sh at the end, and substitute that in, here and at the start.  That way we don't have to update these every time we make a new one.

jannotti (changes requested):
Things look good, though you will want to rebase this once the PR it's based on is merged, so we don't get my commit log added twice.

tsachiherman:
I do like the idea of expanding the size of the URL, but think that we should also increase the min required balance for these assets. This is a general observation as the number of assets on the network grows - note that the min balance here is for asset "owners" and not for asset holders. Also - if we have that degree of flexibility, we could also allow much longer URLs. i.e. if someone is willing to lock enough Algos, I don't see any reason to prevent him from having an asset that has 4K long url.

test/scripts/e2e_subs/asset-misc.sh
jdtzmn: Will do!

test/scripts/e2e_subs/v26/short-asset-url-only.sh
jdtzmn: Yes, I noticed it as well but I didn't want to mess anything up so I copied it verbatim from the `asset-misc.sh` file. I'll update it.

jannotti:
> If someone is willing to lock enough Algos, I don't see any reason to prevent him from having an asset that has 4K long url.

That is generally my thought process as well, though more for apps than ASAs.  With ASAs, I felt we'd be better served by keeping things as static and backward compatible as possible (all ASAs have the same min balance requirement.) to have quick support for common NFT standards (refer to an ipfs url).

If we have to go any further (like give ASAs some extra space to arbitrary use) I'd certainly push for a min balance bump. Clearly in that case, callers would be new code (requesting such space) and could be expected to understand the min balance bump.

winder:
For what it's worth, a partner asked recently about maximum sizes for ASA strings so that they could enact stricter Indexer queries on their load balancers. We would need to make sure this sort of change is documented and communicated.

jannotti:
> We would need to make sure this sort of change is documented and communicated.

Is updating the foundation spec sufficient? (I mean, in one sense it is clearly not - we'd have to point out the update, but I'm asking if any other documentation needs to be changed?)

The foundation spec update is part of the PR I'm accumulating for this consensus upgrade: https://github.com/algorandfoundation/specs/pull/43

winder:
@jannotti making sure it's in the release notes and notifying @JasonWeathersby for the DevRel blog posts should be enough.

jannotti (approved):
Looks good.

jannotti (approved):

PR #2149, opened 2021-05-13
ledger: disable minimum balance check when not validating or generating

tolikzinovyev:
## Summary

This is needed for indexer that doesn't store `AccountData.TotalSchema` that allows computing account's minimum balance.

## Test Plan

The change is trivial.

### DIFF
--- a/ledger/eval.go
+++ b/ledger/eval.go
@@ -859,17 +859,22 @@ func (eval *BlockEvaluator) transaction(txn transactions.SignedTxn, evalParams *
 			continue
 		}
 
-		dataNew := data.WithUpdatedRewards(eval.proto, rewardlvl)
-		effectiveMinBalance := dataNew.MinBalance(&eval.proto)
-		if dataNew.MicroAlgos.Raw < effectiveMinBalance.Raw {
-			return fmt.Errorf("transaction %v: account %v balance %d below min %d (%d assets)",
-				txid, addr, dataNew.MicroAlgos.Raw, effectiveMinBalance.Raw, len(dataNew.Assets))
-		}
+		// Only do those checks if needed. It is useful to skip them if the user cannot provide
+		// account data that contains enough information to compute the correct minimum balance
+		// (the case with indexer).
+		if eval.validate || eval.generate {
+			dataNew := data.WithUpdatedRewards(eval.proto, rewardlvl)
+			effectiveMinBalance := dataNew.MinBalance(&eval.proto)
+			if dataNew.MicroAlgos.Raw < effectiveMinBalance.Raw {
+				return fmt.Errorf("transaction %v: account %v balance %d below min %d (%d assets)",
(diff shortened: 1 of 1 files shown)


ledger/eval.go
```diff
@@ -859,17 +859,22 @@ func (eval *BlockEvaluator) transaction(txn transactions.SignedTxn, evalParams *
...
-		dataNew := data.WithUpdatedRewards(eval.proto, rewardlvl)
-		effectiveMinBalance := dataNew.MinBalance(&eval.proto)
-		if dataNew.MicroAlgos.Raw < effectiveMinBalance.Raw {
-			return fmt.Errorf("transaction %v: account %v balance %d below min %d (%d assets)",
-				txid, addr, dataNew.MicroAlgos.Raw, effectiveMinBalance.Raw, len(dataNew.Assets))
-		}
+		// Only do those checks if needed. It is useful to skip them if the user cannot provide
+		// account data that contains enough information to compute the correct minimum balance
+		// (the case with indexer).
+		if eval.validate || eval.generate {
```
winder, line 865: Please avoid the indentation to match the rest of the code by changing this to

```
if !eval.validate && !eval.generate {
    continue
}
```

winder (commented):
Could you call `applyTransaction` directly instead of disabling the verification logic?

tolikzinovyev:
> Could you call `applyTransaction` directly instead of disabling the verification logic?

I think it's not so easy. There is logic outside of `applyTransaction()`.

ledger/eval.go
tsachiherman: I think that we should take one step back here. 

This entire for loop could be skipped in case `!eval.validate && !eval.generate` holds true. 
Please take a close look, but if this true, I think that we could extract this for-loop into a separate method. That would address the indentation issue, provide faster code and work for the indexer - all at the same time.

tolikzinovyev: You are right, thanks!

ledger/eval.go
```diff
@@ -837,42 +880,15 @@ func (eval *BlockEvaluator) transaction(txn transactions.SignedTxn, evalParams *
...
-			}
-		}
-	}
+  // Only do those checks if we are validating or generating. It is useful to skip them
+  // if we cannot provide account data that contains enough information to
+  // compute the correct minimum balance (the case with indexer which does not store it).
+  if eval.validate || eval.generate {
+    err := eval.checkMinBalance(cow)
+    if err != nil {
+      return fmt.Errorf("transaction: %v err: %w", txid, err)
```
tsachiherman, line 889: `"transaction: %v err: %w"` -> `"transaction %v: %w"`

just to retain the previous error strings, which might be used elsewhere.

tsachiherman (changes requested):
run `make fmt` to format your changes.

tsachiherman (approved):

PR #2150, opened 2021-05-13
txnsync: improve outgoing message error handling

tsachiherman:
## Summary

Add error handling for outgoing messages:
1. If we fail to send a message to a peer, disconnect the peer ( it's probably already disconnected. But we can't tell that for sure ).
2. If the txnsync internally unable to enqueue the response for the successfully sent message, disconnect from peer. ( we could probably do better here, but that's an issue for future PR ).
3. In the emulator, detect txnsync error and panic on it.

## Test Plan

1. Emulator
2. Performance testing
3. Reviewing log files under load.

### DIFF
--- a/txnsync/emulatorNode_test.go
+++ b/txnsync/emulatorNode_test.go
@@ -315,10 +315,13 @@ func (n *emulatedNode) step() {
 			dm := peer.deferredSentMessages[0]
 			peer.deferredSentMessages = peer.deferredSentMessages[1:]
 			peer.mu.Unlock()
-			dm.callback(true, dm.seq)
+			err := dm.callback(true, dm.seq)
 			n.unblock()
 			n.waitBlocked()
 			peer.mu.Lock()
+			if err != nil {
+				panic(err)
+			}
 		}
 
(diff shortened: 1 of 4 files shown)


txnsync/emulatorNode_test.go
```diff
@@ -315,10 +315,13 @@ func (n *emulatedNode) step() {
 			dm := peer.deferredSentMessages[0]
 			peer.deferredSentMessages = peer.deferredSentMessages[1:]
 			peer.mu.Unlock()
-			dm.callback(true, dm.seq)
+			err := dm.callback(true, dm.seq)
 			n.unblock()
 			n.waitBlocked()
 			peer.mu.Lock()
+			if err != nil {
```
tolikzinovyev, line 322: why not check `err` above?

tolikzinovyev (approved):

brianolson (approved):

txnsync/emulatorNode_test.go
tsachiherman: yes, that's a good point. I'll fix this in a subsequent PR.

PR #2152, opened 2021-05-13
Return an empty string instead of null when making a proof for transaction with no siblings.

winder:
## Summary

A nil array is serialized to null which looks strange in the API response. Initialize to an empty array.

## Test Plan

Manually checked a transaction with no siblings.

Before:
```
$ curl -H "Authorization: Bearer $token" "$net/v2/blocks/14089979/transactions/ZEFJNJD3LRTVW2R6PPJ3G5RW26MP3Q3AW5VGVSGFDHG5JU2QJRIA/proof?pretty"
{
  "idx": 0,
  "proof": null,
  "stibhash": "RdW1Eb+3FnedfvEio+XeNF54dJzXa/3JRYSuXaC7Lvk="
}
```

After:
```
$ curl -H "Authorization: Bearer $token" "$net/v2/blocks/14089979/transactions/ZEFJNJD3LRTVW2R6PPJ3G5RW26MP3Q3AW5VGVSGFDHG5JU2QJRIA/proof?pretty"
{
  "idx": 0,
  "proof": "",
  "stibhash": "RdW1Eb+3FnedfvEio+XeNF54dJzXa/3JRYSuXaC7Lvk="
}
```

### DIFF
--- a/daemon/algod/api/server/v2/utils.go
+++ b/daemon/algod/api/server/v2/utils.go
@@ -184,7 +184,7 @@ func getCodecHandle(formatPtr *string) (codec.Handle, string, error) {
 
 	switch format {
 	case "json":
-		return protocol.JSONStrictHandle, "application/json", nil
+		return protocol.JSONHandle, "application/json", nil
 	case "msgpack":
 		fallthrough
 	case "msgp":
--- a/daemon/algod/api/server/v2/utils.go
+++ b/daemon/algod/api/server/v2/utils.go
@@ -184,7 +184,7 @@ func getCodecHandle(formatPtr *string) (codec.Handle, string, error) {
 
 	switch format {
 	case "json":
-		return protocol.JSONHandle, "application/json", nil
+		return protocol.JSONStrictHandle, "application/json", nil
 	case "msgpack":
 		fallthrough
 	case "msgp":
--- a/test/scripts/e2e_subs/serial/rest-proof-endpoint.sh
+++ b/test/scripts/e2e_subs/serial/rest-proof-endpoint.sh
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
-# TIMEOUT=600
+# TIMEOUT=60
 
 my_dir="$(dirname "$0")"
 #"$my_dir/rest.sh" "$@"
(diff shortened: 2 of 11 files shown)


tsachiherman (changes requested):
Could you update the e2e test for this to confirm this particular use case is working as expected?

winder:
@tsachiherman I added a test. Unfortunately, the test uncovered an unrelated bug that could have downstream side effects in the SDKs.

daemon/algod/api/server/v2/utils.go
```diff
@@ -184,7 +184,7 @@ func getCodecHandle(formatPtr *string) (codec.Handle, string, error) {
 
 	switch format {
 	case "json":
-		return protocol.JSONHandle, "application/json", nil
+		return protocol.JSONStrictHandle, "application/json", nil
```
tsachiherman, line 187: is this really needed for this PR ? I'm worried that it might cause some regression elsewhere. i.e.
```golang
// JSONStrictHandle is the same as JSONHandle but with MapKeyAsString=true
// for correct maps[int]interface{} encoding
var JSONStrictHandle *codec.JsonHandle
```

winder: Without this change the test becomes more complicated because `NUM_TRANSACTIONS=$(cat "${TEMPDIR}/curl_out.txt" | jq '.block.txns | length')` doesn't work.

winder: I just thought of an alternative and have removed this change.

That said, we need to use JSONStrictHandle. The API should not respond with invalid JSON.

tsachiherman (dismissed):
looks good. I think that enabling StrictJSON is fine for the v2/blocks/round endpoint.
( but I don't know if that's also true for the rest of our endpoints )

test/scripts/e2e_subs/serial/rest-proof-endpoint.sh
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env bash
+# TIMEOUT=600
```
tsachiherman, line 2: do we really need a timeout of 10m for this one ?

winder: Thanks, forgot about this. A long timeout is no longer required.

tsachiherman (approved):

PR #2153, opened 2021-05-13
txsync: avoid reevaluating old transactions during transaction selection

tsachiherman:
## Summary

This PR avoids reevaluating old transactions during transaction selection by maintaining the `lastTransactionSelectionGroupCounter`. This variable would get reset when new bloom filter arrives ( since we want to re-evaluate past transactions against the new bloom filter ),  as well as when the requested bloom filter changes.
In order to optimize that further, the transaction encoded size was moved to the transaction pool, ensuring that it would be executed only one time.

## Test Plan

1. Verified using the emulator
2. Tested on performance system
3. Tested manually on s1 network /w pingpong


txnsync/peer.go
```diff
@@ -215,6 +211,11 @@ func (p *Peer) selectPendingTransactions(pendingTransactions []transactions.Sign
 			}
 		}
 
+		if p.recentSentTransactions.contained(txID) {
+			// we already sent that transaction. no need to send again.
+			continue
```
algonautshant, line 216: When can this happen?

algonautshant (approved):
LGTM

txnsync/peer.go
tsachiherman: The other node send us periodically a new bloom filter. When this happen, we fetch a refreshed list of transactions from the transaction pool, and filter these using this method. The refreshed list might contain some of the transactions that were previously sent to the peer. We can't use the bloom filter for that, since the bloom filter isn't being updated frequently enough.

txnsync/peer.go
```diff
@@ -228,26 +229,28 @@ func (p *Peer) selectPendingTransactions(pendingTransactions []transactions.Sign
...
 		selectedTxnIDs = append(selectedTxnIDs, txID)
 
-		// calculate the total size of the transaction group.
-		for txidx := range pendingTransactions[grpIdx].Transactions {
-			encodingBuf := protocol.GetEncodingBuf()
-			accumulatedSize += len(pendingTransactions[grpIdx].Transactions[txidx].MarshalMsg(encodingBuf))
-			protocol.PutEncodingBuf(encodingBuf)
-		}
+		// add the size of the transaction group
+		accumulatedSize += pendingTransactions[grpIdx].EncodedLength
```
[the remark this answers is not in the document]
tsachiherman: The rationale here is that we want to make sure we have at least a single transaction in the message, so we would always "make progress", and not fall into situations where no transactions are being exchanged due to low bandwidth calculations.

txnsync/peer.go
winder: Is this assuming that a transaction only needs to be evaluated once during the selection process?

What about a multiple spend case. If relay A processes txA which drains an account and relay B processes txB which also drains the account. Is this optimization still valid to skip?

tsachiherman: multiple spend ( or transaction validity in general ) is not something the transaction sync would try/need to address. The goal fo the transaction sync is to make sure the transaction gets to all intended parties( ideally only once). The validation is done on the transaction pool side.

txnsync/peerscheduler.go
```diff
@@ -97,6 +97,21 @@ func (p *peerScheduler) nextPeers() (outPeers []*Peer) {
 		bucket := heap.Remove(p, 0).(peerBucket)
 		outPeers = append(outPeers, bucket.peer)
 	}
+
+	// in many cases, we'll have only a single peer; however, in case we have multiple
+	// ( which is more likely when we're "running late" ), we want to make sure to remove
+	// duplicate ones.
```
winder, line 103: Are you doing this here to avoid making a map each time `Push` is called? It's a bit confusing to do it this way, and don't you also need to remove duplicate peers from the `p.peers`?

tsachiherman: That's a good question. The reason for this is that the `next` timestamp ( which is now the `p.peers[0].next` ) is going to be dynamically throttled to allow different "groups" of peers to be selected. Because of that, it's not possible to do that during the `Push`.

This PR is one of a series of PR required for implementing the throttler. So - it lays the foundation for the throttler without introducing the actual implementation.

tsachiherman: and no, there is no need to remove duplicate `p.peers` instance, since we want to have the ability to have multiple-schedules for a single peer.

txnsync/peer.go
```diff
@@ -228,26 +228,28 @@ func (p *Peer) selectPendingTransactions(pendingTransactions []transactions.Sign
...
 		if accumulatedSize > windowLengthBytes {
 			windowSizedReached = true
 		}
 	}
 
+	if grpIdx == len(pendingTransactions) {
+		grpIdx--
+	}
+	if grpIdx >= 0 {
+		p.lastTransactionSelectionGroupCounter = pendingTransactions[grpIdx].GroupCounter
```
winder, line 250: I think it would be clearer to handle the empty case directly, I was a bit confused by the `grpIdx--` and `grpIdx >= 0` logic.

Something like `grpIdx != 0 && grpIdx == len(pendingTransactions)`, then access `grpIdx - 1`.

Alternatively update `p.lastTransactionSelectionGroupCounter` on each iteration of the loop.

tsachiherman: I wanted to avoid `p.lastTransactionSelectionGroupCounter` on each loop iteration. I'll update the predicate as you've suggested.

tsachiherman: I think that this expression wasn't good enough. Let me run few more test on this one and I'll ask for another review once I'm done.

txnsync/peer.go
```diff
@@ -228,26 +228,35 @@ func (p *Peer) selectPendingTransactions(pendingTransactions []transactions.Sign
...
+		// add the size of the transaction group
+		accumulatedSize += pendingTransactions[grpIdx].EncodedLength
+
 		if accumulatedSize > windowLengthBytes {
 			windowSizedReached = true
 		}
 	}
 
+	// update the lastTransactionSelectionGroupCounter if needed -
+	// if we selectged any transaction to be sent, update the lastTransactionSelectionGroupCounter with the latest
```
winder, line 247:
```suggestion
	// if we selected any transactions to be sent, update the lastTransactionSelectionGroupCounter with the latest
```

txnsync/peer.go
```diff
@@ -228,26 +228,35 @@ func (p *Peer) selectPendingTransactions(pendingTransactions []transactions.Sign
...
+	// group counter. If the startIndex was *after* the last pending transaction, it means that we don't
+	// need to update the lastTransactionSelectionGroupCounter since it's already ahead of everything in the pending transactions.
+	if grpIdx >= 0 && startIndex < len(pendingTransactions) {
+		if grpIdx == len(pendingTransactions) {
+			if grpIdx > 0 {
+				p.lastTransactionSelectionGroupCounter = pendingTransactions[grpIdx-1].GroupCounter + 1
+			}
+		} else {
+			p.lastTransactionSelectionGroupCounter = pendingTransactions[grpIdx].GroupCounter
```
winder, line 256: No `+ 1` here because the counter indicates the next group to be processed?

txnsync/peer.go
winder: This feels like premature optimization, assigning `p.lastTransactionSelectionGroupCounter` is simpler.

txnsync/peer.go
tsachiherman: exactly.

winder (approved):
Just a couple of nits, otherwise look good.

txnsync/peer.go
tsachiherman: I agree that just assigning the `p.lastTransactionSelectionGroupCounter` is truly simpler. However, I found that this method was taking a considerable amount of the total txnsync time, so I'm trying to optimize it. Moving operations out of the loop is my first go-to solution here.

txnsync/peer.go
tsachiherman: ok, fixed.

PR #2155, opened 2021-05-14
txpool: Fix incorrect datatype checking in recomputeBlockEvaluator

tsachiherman:
## Summary

When the recomputeBlockEvaluator runs, it tests all the pending transactions against the ledger to detect duplicate transactions ( which happen all the time; these aren't an issue ).
The block evaluator reports the error by returning `*ledgercore.TransactionInLedgerError` and not `ledgercore.TransactionInLedgerError` as the code was testing against.

## Note
Yet another one character bug fix.

## Test Plan

The bug would only affect the reported block assembly stats, and therefore the issue was validated against the telemetry.

### DIFF
--- a/data/pools/transactionPool.go
+++ b/data/pools/transactionPool.go
@@ -695,7 +695,7 @@ func (pool *TransactionPool) recomputeBlockEvaluator(committedTxIds map[transact
 			}
 
 			switch err.(type) {
-			case ledgercore.TransactionInLedgerError:
+			case *ledgercore.TransactionInLedgerError:
 				asmStats.CommittedCount++
 				stats.RemovedInvalidCount++
 			case transactions.TxnDeadError:


data/pools/transactionPool.go
```diff
@@ -695,7 +695,7 @@ func (pool *TransactionPool) recomputeBlockEvaluator(committedTxIds map[transact
 			}
 
 			switch err.(type) {
-			case ledgercore.TransactionInLedgerError:
+			case *ledgercore.TransactionInLedgerError:
```
algonautshant, line 698: This looks confusing, and can be mistaken somewhere else. 
Why not return the error by value, and make the Txid inside it a pointer, if copying it is a concern?
type TransactionInLedgerError struct {                                                                           
        Txid *transactions.Txid                                                                                   
}

algonautshant (approved):
Looks correct. 
However, I prefer that the error be returned by value, and the pointer be inside the error struct.

data/pools/transactionPool.go
tsachiherman: I agree. We clearly need to figure this out. However, as you've noted, it could be a sensitive issue where fixing it incorrectly could lead to other issues. That's why I opted in for the simple fix.

PR #2157, opened 2021-05-17
"Rent" larger program size

shiqizng:
Allow paying for more program space


data/transactions/transaction.go
```diff
@@ -359,6 +359,10 @@ func (tx Transaction) WellFormed(spec SpecialAddresses, proto config.ConsensusPa
 			return fmt.Errorf("tx.GlobalStateSchema too large, max number of keys is %d", proto.MaxGlobalSchemaEntries)
 		}
 
+		if tx.ExtraProgramPages > proto.MaxExtraAppProgramPages {
```
jannotti, line 362: Very small complaint - seem like we ought to move this check up, so that tx.ExtraProgramPages is checked before using it on lines 346 and 350.  No difference in result, but it seem right.

jannotti (commented):
This is looking really good so far. I mentioned a small unit test that can happen for the min_balance opcode, I don't know how many unit tests we have a txns - I expect you'll most need to do most checks as integration tests.

data/transactions/transaction.go
```diff
@@ -343,12 +343,12 @@ func (tx Transaction) WellFormed(spec SpecialAddresses, proto config.ConsensusPa
...
-		if len(tx.ApprovalProgram) > proto.MaxAppProgramLen {
-			return fmt.Errorf("approval program too long. max len %d bytes", proto.MaxAppProgramLen)
+		if len(tx.ApprovalProgram) > ((1 + tx.ExtraProgramPages) * proto.MaxAppProgramLen) {
+			return fmt.Errorf("approval program too long. max len %d bytes", (1+tx.ExtraProgramPages)*proto.MaxAppProgramLen)
 		}
 
-		if len(tx.ClearStateProgram) > proto.MaxAppProgramLen {
-			return fmt.Errorf("clear state program too long. max len %d bytes", proto.MaxAppProgramLen)
+		if len(tx.ClearStateProgram) > ((1 + tx.ExtraProgramPages) * proto.MaxAppProgramLen) {
+			return fmt.Errorf("clear state program too long. max len %d bytes", (1+tx.ExtraProgramPages)*proto.MaxAppProgramLen)
```
algorandskiy, line 351: are we allowing a single page of `ExtraProgramPages` to be count towards both clear state and approval? One can pay for a page but utilize 2x space

ledger/apply/application.go
```diff
@@ -114,6 +115,11 @@ func createApplication(ac *transactions.ApplicationCallTxnFields, balances Balan
 	totalSchema = totalSchema.AddSchema(ac.GlobalStateSchema)
 	record.TotalAppSchema = totalSchema
 
+	// Update the cached TotalExtraAppPages for this account, used
+	// when computing MinBalance
+	totalExtraPages := record.TotalExtraAppPages
+	totalExtraPages += ac.ExtraProgramPages
```
algorandskiy, line 121: these variables are not used, and `TotalExtraAppPages` is not assigned

data/transactions/transaction.go
```diff
@@ -343,12 +343,12 @@ func (tx Transaction) WellFormed(spec SpecialAddresses, proto config.ConsensusPa
 			return fmt.Errorf("tx.ForeignAssets too long, max number of foreign assets is %d", proto.MaxAppTxnForeignAssets)
 		}
 
-		if len(tx.ApprovalProgram) > proto.MaxAppProgramLen {
-			return fmt.Errorf("approval program too long. max len %d bytes", proto.MaxAppProgramLen)
+		if len(tx.ApprovalProgram) > ((1 + tx.ExtraProgramPages) * proto.MaxAppProgramLen) {
```
algorandskiy, line 346: usage of `tx.ExtraProgramPages` needs to be under a feature switch. Consider a scenario:
1. Half of nodes upgraded to `ExtraProgramPages`, other half is not.
2. Nodes receive a txn with `ExtraProgramPages` = 1 and `len(ApprovalProgram) = proto.MaxAppProgramLen * 1.5`. Then upgrade nodes will pass the txn even before the protocol upgrade, and other will break (most likely on txn decoding)

jannotti: Until the protocol upgrade, having a ExtraProgramPages > 0 will be rejected either way.  An upgraded node will be mad that ExtraProgramPages > 0 (since proto.MaxExtraAppProgramPages will be initialized to 0), while an unupgraded node will be made it contains the field at all. So I think the protocol upgrade itself acts as the feature flag.

ExtraProgramPages = 0 would be trouble, except we will reject that because it should be omitted when 0.

data/transactions/transaction.go
jannotti: With these changes, paying extra will get you an extra 1k for each.  I intend to add changes that combine them later, and then each page will get to 2k total extra.

data/transactions/transaction.go
algorandskiy: correct if moving `if tx.ExtraProgramPages > proto.MaxExtraAppProgramPages` check right above.
I assume old code will fail on decoding `ExtraProgramPages` but it is better to double check.

algorandskiy: > ExtraProgramPages = 0 would be trouble, except we will reject that because it should be omitted when 0.

@jannotti could you elaborate? 0 looks ok here.

jannotti: We usually use 'omitempty' (and certainly should here, for the extra pages parameter) and when we do that, I think it becomes impossible to sign a transaction with 0 in it for that field, because we must do some check where we reencode to canonical representation (which will differ, since the 0 field will be dropped) and check the signature against that.

shiqizng: in that case, a program of len <= proto.MaxAppProgramLen will get dropped?

jannotti: No, I was just pointing out that @algorandskiy 's concern might occur if, before the protocol upgrade:
a) Node with new release accepts a txn containing extrapages=0
b) Not upgraded node can't understand it.

But, that's not actually possible.  Txns can't *explicitly* set the extra pages field to 0. Instead, they omit it if they mean 0. So I don't think anything bad can happen.

algorandskiy: still I think we are fine:
1. old node receives a txn with omitted `ExtraProgramPages` => it decodes and proceeds as now
2. new node creates a txn without `ExtraProgramPages` => fine
3. new node receives a txn with omitted`ExtraProgramPages` => decodes as ExtraProgramPages=0 and proceeds with new code checks
4. new node receives a txn with non-zero `ExtraProgramPages` => fine, execute new code checks

jannotti: I agree

algorandskiy: I agree. So move the check `tx.ExtraProgramPages > proto.MaxExtraAppProgramPages` above and write a test checking
1) pre-upgrade txn with ExtraProgramPages=1 fails
2) post-upgrade txn with ExtraProgramPages=1 succeed

data/transactions/application.go
```diff
@@ -134,14 +134,19 @@ type ApplicationCallTxnFields struct {
...
 	// OnCompletion is equal to ClearStateOC. This program will not cause
 	// the transaction to be rejected, even if it fails. This program may
 	// read and write local and global state for this application.
-	ClearStateProgram []byte `codec:"apsu,allocbound=config.MaxAppProgramLen"`
+	ClearStateProgram []byte `codec:"apsu,allocbound=config.MaxAvailableAppProgramLen"`
+
+	// ExtraProgramPages specifies the additional app program len requested in pages.
+	// A page is MaxAppProgramLen bytes. This field enables execution of app programs
+	// larger than the default config, MaxAppProgramLen.
+	ExtraProgramPages int `codec:"apepp"`
```
jannotti, line 149: Can we make this "apep"?  All the other msgpack fields for app data are apXX.

jannotti (commented):
Look for: 
func (tx Transaction) WellFormed(spec SpecialAddresses, proto config.ConsensusParams) error {

in data/transactions/transaction.go.  We ought to complain if extra pages is set for anything besides an app create txn.

algorandskiy (commented):
still need more unit tests:
1. pre and post protocol update
2. non-zero `ExtraProgramPages` for a regular (non-create) app call
3. total pages and app deletion
4. min balance test for old and new calculations?

shiqizng:
> still need more unit tests:
> 
> 1. pre and post protocol update
> 2. non-zero `ExtraProgramPages` for a regular (non-create) app call
> 3. total pages and app deletion
> 4. min balance test for old and new calculations?

@algorandskiy  I added unit tests for 1 and 2 as you suggested. I'm not sure it's a good idea to test these as unit tests though. It seems that what I added in transactions_test.go wouldn't make much sense anymore once future consensus is promoted to current.

algorandskiy:
> @algorandskiy I added unit tests for 1 and 2 as you suggested. I'm not sure it's a good idea to test these as unit tests though. It seems that what I added in transactions_test.go wouldn't make much sense anymore once future consensus is promoted to current.

This is easy to fix. Instead of
```
curProto := config.Consensus[protocol.ConsensusCurrentVersion]
```
use
```
curProto := config.Consensus[protocol.ConsensusV27]
```

data/transactions/transaction_test.go
```diff
@@ -159,6 +259,69 @@ func TestWellFormedErrors(t *testing.T) {
...
+				ApplicationCallTxnFields: ApplicationCallTxnFields{
+					ApplicationID: 0,
+					ApplicationArgs: [][]byte{
+						[]byte("write"),
+					},
+					ExtraProgramPages: 1,
+				},
+			},
+			spec:          specialAddr,
+			proto:         curProto,
```
algorandskiy, line 280: this test will break after protocol upgrade

test/e2e-go/features/transactions/accountv2_test.go
```diff
@@ -154,7 +154,7 @@ int 1
 	// create the app
 	tx, err := client.MakeUnsignedAppCreateTx(
 		transactions.OptInOC, approvalOps.Program, clearstateOps.Program, schema, schema, nil, nil, nil, nil,
-	)
+		0)
```
algorandskiy, line 157: weird formatting here and below - move 0 to the end of args list

data/transactions/logic/eval_test.go
```diff
@@ -1458,6 +1458,17 @@ int 2
 assert
 
 
+int 1
+`
+
+const testTxnProgramTextV4 = testTxnProgramTextV3 + `
+assert
+txn AppProgramExtraPages
```
algorandskiy, line 1466: please add `gtxn` test as well

algorandskiy (commented):
Overall looks good. Please move `MaxAvailableAppProgramLen` out of consensus params as JJ suggested. See additional comments below.

data/transactions/transaction_test.go
shiqizng: should I remove this?

algorandskiy: no, just use v27 proto, not current.

algorandskiy (approved):

data/transactions/transaction.go
```diff
@@ -382,12 +385,16 @@ func (tx Transaction) WellFormed(spec SpecialAddresses, proto config.ConsensusPa
...
+		}
+
+		if uint32(len(tx.ApprovalProgram)) > ((1 + tx.ExtraProgramPages) * uint32(proto.MaxAppProgramLen)) {
+			return fmt.Errorf("approval program too long. max len %d bytes", (1+tx.ExtraProgramPages)*uint32(proto.MaxAppProgramLen))
 		}
 
-		if len(tx.ClearStateProgram) > proto.MaxAppProgramLen {
-			return fmt.Errorf("clear state program too long. max len %d bytes", proto.MaxAppProgramLen)
+		if uint32(len(tx.ClearStateProgram)) > (1+tx.ExtraProgramPages)*uint32(proto.MaxAppProgramLen) {
+			return fmt.Errorf("clear state program too long. max len %d bytes", (1+tx.ExtraProgramPages)*uint32(proto.MaxAppProgramLen))
```
jannotti, line 397: I'd like to see the formatting of this be identical to the formatting of the ApprovalProgram check.

data/transactions/transaction_test.go
```diff
@@ -121,10 +121,111 @@ func TestGoNonparticipatingWellFormed(t *testing.T) {
...
+					Fee:        basics.MicroAlgos{Raw: 1000},
+					LastValid:  105,
+					FirstValid: 100,
+				},
+				ApplicationCallTxnFields: ApplicationCallTxnFields{
+					ApplicationID: 0,
+					ApplicationArgs: [][]byte{
+						[]byte("write"),
+					},
+					ExtraProgramPages: 3,
```
jannotti, line 190: To test extra pages, don't you need to try inserting a big program?  I'm somewhat surprised this is considered well formed at all, since it's trying to create an app without any program bytes. But maybe the empty program is allowed?

test/e2e-go/cli/goal/expect/statefulTealCreateAppTest.exp
```diff
@@ -114,6 +114,17 @@ proc statefulTealTest { TEST_ALGO_DIR TEST_DATA_DIR TEAL_PROGRAM} {
     # Delete application should still succeed
     ::AlgorandGoal::AppDelete $APP_ID $WALLET_1_NAME $WALLET_1_PASSWORD $ACCOUNT_1_ADDRESS "str:hello" $TEST_PRIMARY_NODE_DIR
 
+    # App create with extra pages option should succeed
+    ::AlgorandGoal::AppCreateExPages $PRIMARY_WALLET_NAME "" $PRIMARY_ACCOUNT_ADDRESS ${TEAL_PROGS_DIR}/${TEAL_PROGRAM} "str:hello" $GLOBAL_BYTE_SLICES $LOCAL_BYTE_SLICES ${TEAL_PROGS_DIR}/clear_program_state.teal $TEST_PRIMARY_NODE_DIR 2
```
jannotti, line 118: Again, we don't want to just test that extra pages value is set, but that because it's set specifying a large program is allowed. So you need to feed it a big approval program.  (Which may require changing goal to allow such a long program when the proto allows extra-pages)

jannotti (changes requested):
I think we need to have the tests test whether big programs are actually allowed, unless I misunderstood the tests.  They appear to only test whether the extra-pages option is allowed to be set.

data/transactions/transaction_test.go
shiqizng: I believe this is done in userBalance_test.go

jannotti: That only confirms that we can _encode_ accounts with large apps in them.  So we know the max account size is still bounded by MaxEncodedAccountDataSize.  But this doesn't test that we actually _accept_ the txn. So the code where you compare the incoming program size to the value determined by extra pages is never exercised.

I'm not even sure if we've tested your allocbounds work, to ensure we can encode txns with very large programs in them.

test/e2e-go/cli/goal/expect/statefulTealCreateAppTest.exp
shiqizng: what's the simplest way to generate a large teal file? or is there one I could use?

jannotti: Since this command is expecting teal source code, you can create a file that starts "#pragma version 3" and then has as many lines of "int 1" as you'd like. Each "int 1" will contribute one more byte to the program length.

This app won't return success is executed (because the stack may overflow, and if it doesn't it still won't end properly) but that won't matter for your tests.

jannotti: I have some code in tealConsensusTest.exp that creates long teal programs.

jannotti (approved):

algorandskiy (approved):

PR #2158, opened 2021-05-18
Access values from a previous app call's scratch space

jdtzmn:
## Summary

This PR allows contract to contract composability by allowing future app call transactions to read the scratch space of previous transactions. This allows smart contracts to expose side effects for future application calls, such as a price oracle returning the exchange rate of a particular asset pair.

Changes include:
- Adding `PastSideEffects` fields to `EvalParams`.
- `gload` and `gloads` docs.

Closes algorand/go-algorand-internal#1189.

## Test Plan
- Ensure that past side effects are not `nil` after `EvalParams` construction.
- `gload` tests:
	- Simple case with two application calls.
	- Multiple application calls.
	- Test failure when accessing scratch space of current app call.
	- Test failure when accessing future app call's scratch space.
	- Test failure when accessing the scratch space of a non-app call.
	- Test failure from within logicsig / stateless
- Integration test for vFuture that tests compiling and executing `gload` from a TEAL file.


jdtzmn:
Does accessing scratch space of previous transactions expose a security risk for future application calls accessing the same scratch space? For example, could transaction 2 in a group change the scratch space bytes of transactions 1 such that transaction 3 sees a corrupted version? Uint64s shouldn't be an issue but I'm not familiar enough with Go to know if byteslices will allow malicious changes.

algorandskiy (changes requested):
I do not think `evalContext` needs to be exposed outside of logic package since it completely breaks encapsulation.
Consider alternative implementation approaches:
1) use `LedgerForLogic` interface to save/retrieve scratch space after a program successfully evaluated (if this is only needed for applications)
2) return scratch from eval and put it into EvalParams prepared for upcoming `logic.Eval` call

jdtzmn:
@algorandskiy Good point – I agree that it unnecessarily breaks encapsulation. Originally I had planned to add an `EvalSideEffects` struct to `EvalParams` but decided instead to go the more direct route of context references, since access to the evaluation context is the main requirement for achieving the composability functionality. And thank you for pointing out the `LedgerForLogic` interface, it appears to be a suitable location for exposing scratch space.

data/transactions/logic/eval_test.go
```diff
@@ -1806,6 +1807,178 @@ func testLogic(t *testing.T, program string, v uint64, ep EvalParams, problems .
...
+	}
+	proto := defaultEvalProtoWithVersion(LogicVersion)
+
+	for i, testCase := range cases {
+		t.Run(fmt.Sprintf("i=%d", i), func(t *testing.T) {
+			sources := testCase.tealSources
+			// Assemble ops
+			opsList := make([]*OpStream, len(sources))
+			for j, source := range sources {
+				ops, err := AssembleStringWithVersion(source, AssemblerMaxVersion)
```
jannotti, line 1884: Use testProg when possible.  It shows what's going on if your assembly fails, and eliminates the need to test the error.

data/transactions/logic/eval.go
```diff
@@ -1543,14 +1580,37 @@ func (cx *evalContext) txnFieldToStack(txn *transactions.Transaction, field TxnF
 		sv.Bytes = txn.FreezeAccount[:]
 	case FreezeAssetFrozen:
 		sv.Uint = boolToUint(txn.AssetFrozen)
+	case Scratch:
+		if txn.Type != protocol.ApplicationCallTx {
+			err = fmt.Errorf("can't use Scratch txn field on non-app call txn with index %d", groupIndex)
+			return
+		} else if cx.Txn.Lsig.Logic != nil {
+			err = fmt.Errorf("can't use Scratch txn field from within a LogicSig")
+			return
```
jannotti, line 1589: Simplify this down to just checking the runMode

data/transactions/logic/eval.go
```diff
@@ -1543,14 +1580,37 @@ func (cx *evalContext) txnFieldToStack(txn *transactions.Transaction, field TxnF
...
+		} else if groupIndex > cx.GroupIndex {
+			err = fmt.Errorf("can't get future Scratch from txn with index %d", groupIndex)
+			return
+		}
+		val := cx.PastSideEffects[groupIndex].GetScratchValue(uint8(arrayFieldIdx))
+		if val.Bytes != nil {
+			sv.Bytes = val.Bytes
+		} else {
+			sv.Uint = val.Uint
+		}
```
jannotti, line 1605: Perhaps this means that instead of a clone() func on StackValue, you should add an assign(StackValue).  It would not allocate a StackValue to return it, it would assign the given StackValue into the receiver, copying the bytes field.

(Is GetScratchValue currently using clone()?)

data/transactions/logic/eval.go
```diff
@@ -1543,14 +1580,37 @@ func (cx *evalContext) txnFieldToStack(txn *transactions.Transaction, field TxnF
...
 	default:
 		err = fmt.Errorf("invalid txn field %d", field)
 		return
 	}
 
 	txnField := TxnField(field)
 	txnFieldType := TxnFieldTypes[txnField]
-	if txnFieldType != sv.argType() {
+	if txnFieldType != sv.argType() && txnFieldType != StackAny {
```
jannotti, line 1613: I think we have a func for type checking that does this.

data/transactions/logic/eval.go
jdtzmn: Yes! GetScratchValue is using clone().

jdtzmn: Hmm not sure why I didn't just assign to `sv` here. I'll simplify this to `sv = cx.PastSideEffects[...]`.

data/transactions/logic/eval.go
```diff
@@ -1568,7 +1628,7 @@ func opTxn(cx *evalContext) {
 		cx.err = fmt.Errorf("invalid txn field %d", field)
 		return
 	}
-	var sv stackValue
+	var sv StackValue
 	var err error
 	sv, err = cx.txnFieldToStack(&cx.Txn.Txn, field, 0, cx.GroupIndex)
```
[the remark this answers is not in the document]
jdtzmn: I don't see why not - I'll change it.

data/transactions/logic/eval.go
```diff
@@ -49,36 +49,47 @@ const EvalMaxScratchSize = 255
...
 const MaxStringSize = 4096
 
-// stackValue is the type for the operand stack.
-// Each stackValue is either a valid []byte value or a uint64 value.
-// If (.Bytes != nil) the stackValue is a []byte value, otherwise uint64 value.
-type stackValue struct {
+// StackValue is the type for the operand stack.
+// Each StackValue is either a valid []byte value or a uint64 value.
+// If (.Bytes != nil) the StackValue is a []byte value, otherwise uint64 value.
+type StackValue struct {
```
algorandskiy, line 55: it looks like no need to name `StackValue` exported

jdtzmn: Strange, I could've sworn I reverted that rename.

jannotti: Any tools like `go vet` that could help us find things like this?  Names we export for no reason?

data/transactions/logic/eval.go
```diff
@@ -144,6 +155,22 @@ type LedgerForLogic interface {
...
 
+// EvalSideEffects contains data returned from evaluation
+type EvalSideEffects struct {
+	scratchSpace scratchSpace
+}
+
+// GetScratchValue loads and clones a stackValue
+// The value is cloned so the original bytes are protected from changes
+func (se *EvalSideEffects) GetScratchValue(stackPos uint8) stackValue {
+	return se.scratchSpace[stackPos].clone()
```
algorandskiy, line 166:
```suggestion
	return se.scratchSpace[scratchPos].clone()
```

nit: scratch is not stack

data/transactions/logic/eval.go
```diff
@@ -1568,9 +1621,7 @@ func opTxn(cx *evalContext) {
 		cx.err = fmt.Errorf("invalid txn field %d", field)
 		return
 	}
-	var sv stackValue
-	var err error
-	sv, err = cx.txnFieldToStack(&cx.Txn.Txn, field, 0, cx.GroupIndex)
+	sv, err := cx.txnFieldToStack(&cx.Txn.Txn, field, 0, cx.GroupIndex)
```
algorandskiy, line 1624: I think this should be in `gtxna` since `txn` returns self fields.

Plus, Scratch is not a real txn field, so we might want a new `gscratch` or `gload` opcode. @jannotti thoughts?

data/transactions/logic/eval_test.go
```diff
@@ -1806,6 +1806,196 @@ func testLogic(t *testing.T, program string, v uint64, ep EvalParams, problems .
...
+		tealSources []string
+		errContains string
+	}
+
+	simpleCase := scratchTestCase{
+		tealSources: []string{
+			`int 2
+store 0
+int 1`,
+			`gtxna 0 Scratch 0
```
algorandskiy, line 1822: Assuming we commit to allow only static offsets in Scratch.

data/transactions/logic/eval.go
jdtzmn: I think it's cleanup that exists across many of the txn, txna, gtxn, etc. functions, but you are correct that I only modified this one. I'll fix that.

data/transactions/logic/eval_test.go
jdtzmn: Why only static?

data/transactions/logic/eval.go
algorandskiy: to access a scratch item there are few bits of info needed: txn index in a group and scratch item position. The only one opcode allows this: `gtxna` - it is capable to read an item from an array by position from a different txn by index.

data/transactions/logic/eval_test.go
algorandskiy: I meant "constant" - because group index and scratch position are constant in `gtxna 0 Scratch 0` opcode and need to known on assembly time. This might by OK, not sure about real use cases (like passing gi and pos as app call arguments).

jdtzmn: Oh I see. The group index can be dynamic since I only modified the `txnFieldToStack` function, which I believe is used by the `gtxnsa` logic. But that's a good point that the Scratch index will have to be static.

data/transactions/logic/eval.go
jdtzmn: Yes - this change is purely cleanup and doesn't affect the implementation at all.

data/transactions/logic/eval_test.go
jannotti: Yes, that will make it available to all of txna, gtxna, gtxnsa.

I think the scratch slot is okay static for now, since we have the same limitation load load/store.

data/transactions/logic/eval.go
jannotti: I am somewhat in favor of `gload <txn> <scratch>` though iand we'd also need `gloads <scratch>` which would take the txn slot from stack. (but still take the scratch slot from immediate arg)

All in all, I think it would probably be better.  It would be easier to document (we don't have a convenient way to indicate Scratch is a stateful only field), but we have a standard way to say an opcode is stateful only.

data/transactions/logic/eval.go
```diff
@@ -1884,6 +1911,56 @@ func opStore(cx *evalContext) {
...
+		return
+	} else if scratchIdx >= 256 {
+		err = fmt.Errorf("invalid Scratch index %d", scratchIdx)
+		return
+	} else if txn := cx.TxnGroup[groupIdx].Txn; txn.Type != protocol.ApplicationCallTx {
+		err = fmt.Errorf("can't use gload on non-app call txn with index %d", groupIdx)
+		return
+	} else if cx.runModeFlags == runModeSignature {
+		err = fmt.Errorf("can't use gload from within a LogicSig")
+		return
```
jannotti, line 1926: This should be handled by setting the mode for the opcode in opcodes.go

data/transactions/logic/README.md
```diff
@@ -27,7 +27,7 @@ A program can either authorize some delegated action on a normal private key sig
...
 
-The TEAL bytecode plus the length of any Args must add up to less than 1000 bytes (consensus parameter LogicSigMaxSize). Each TEAL op has an associated cost and the program cost must total less than 20000 (consensus parameter LogicSigMaxCost). Most ops have a cost of 1, but a few slow crypto ops are much higher. Prior to v4, program costs was estimated as the static sum of all opcode costs in a program (ignoring conditionals that might skip some code).  Beginning with v4, a program's cost is tracked dynamically, while being evaluated. If the program exceeds its budget, it fails.
+The TEAL bytecode plus the length of any Args must add up to less than 1000 bytes (consensus parameter LogicSigMaxSize). Each TEAL op has an associated cost and the program cost must total less than 20000 (consensus parameter LogicSigMaxCost). Most ops have a cost of 1, but a few slow crypto ops are much higher. Prior to v4, program costs was estimated as the static sum of all opcode costs in a program (ignoring conditionals that might skip some code). Beginning with v4, a program's cost is tracked dynamically, while being evaluated. If the program exceeds its budget, it fails.
```
algorandskiy, line 30: this looks like is from another PR

ledger/eval.go
```diff
@@ -676,21 +677,24 @@ func (eval *BlockEvaluator) prepareEvalParams(txgroup []transactions.SignedTxnWi
 			continue
 		}
 
-		// Initialize group without ApplyData lazily
+		// Initialize side effects and group without ApplyData lazily
 		if groupNoAD == nil {
 			groupNoAD = make([]transactions.SignedTxn, len(txgroup))
+			pastSideEffects = make([]*logic.EvalSideEffects, len(txgroup))
```
algorandskiy, line 683:
```suggestion
			pastSideEffects = logic.MakePastSideEffects(len(txgroup))
```
no need ledger to know implementation details of PastSideEffects.

data/transactions/logic/eval.go
```diff
@@ -158,6 +185,8 @@ type EvalParams struct {
 	// GroupIndex should point to Txn within TxnGroup
 	GroupIndex int
 
+	PastSideEffects []*EvalSideEffects
```
algorandskiy, line 188: please explain a reasoning behind "slide of pointers to EvalSideEffects" instead of "slice of EvalSideEffects"

data/transactions/logic/eval_test.go
```diff
@@ -2227,6 +2241,262 @@ int 5
...
+==
+`,
+		},
+	}
+
+	multipleTxnCase := scratchTestCase{
+		tealSources: []string{
+			`byte "txn 1"
+store 0
+int 1`,
```
algorandskiy, line 2269:
```
		tealSources: []string{`
byte "txn 1"
store 0
int 1`
```

to make it more readable

data/transactions/logic/eval_test.go
```diff
@@ -2227,6 +2241,262 @@ int 5
...
+			`byte "txn 2"
+store 1
+int 1`,
+			`gload 0 0
+byte "txn 1"
+==
+gload 1 1
```
algorandskiy, line 2276: try `gload 1 2` for example to ensure txnIdx and scratchIdx are not confused

algorandskiy (commented):
Looks good! I have a couple of minor testing suggestions and a request 1) to switch from pointers to values (or explain why pointers are needed) and 2) simplify a bit initialization in ledger.

data/transactions/logic/eval.go
jdtzmn: It was unintentional, I didn't realize slices could be modified without pointers. I'll change this.

data/transactions/logic/eval.go
```diff
@@ -432,6 +463,10 @@ func eval(program []byte, cx *evalContext) (pass bool, err error) {
 	if cx.stack[0].Bytes != nil {
 		return false, errors.New("stack finished with bytes not int")
 	}
+
+	// set side effects
+	cx.PastSideEffects[cx.GroupIndex].SetScratchSpace(cx.scratch)
```
algorandskiy, line 468: I think this is a wrong place to set data.
1. `prepareEvalParams` where PastSideEffects is allocated is called only for **app call* txn.
2. this `eval()` function is called for both **logic sig** and **app call**

This means: 1) logic sig txn would cause panic since `PastSideEffects` is empty  - please add a test before fixing 2)  values from **logic sig** are not available for reading in app call txns - is it intentional @jannotti @jdtzmn ?

algorandskiy (changes requested):
The PR is getting close to be done, but there is a problem with logic sig handling - please see the comment below.

data/transactions/logic/eval.go
jdtzmn: Would setting the data in the `EvalStateful` solve this?

algorandskiy: It depends on our design intent - if we want to completely ignore **logic sig** then yes. If we want **logic sig**'s scratch to be exposed to **app call** program then no.

jannotti: I think we should completely ignore logicsig.  Two reasons a) The rule is that you can only see calls that are "earlier" in the group.  That rule does not make sense for logicsigs, as they get evaluated first.  However, the rule would now be quite weird. b) The bigger reason is: what if an app call is logsigged?  Which state do you see?  I think it's much cleaner to say that `gload` is only for accessing app state.

jannotti (approved):

data/transactions/logic/eval.go
jdtzmn: @algorandskiy I've added a check so that past side effects are only set from within an application call.

algorandskiy: why not moving it into EvalStateful?

algorandskiy (approved):
I guess it is fine now, let's ensure the tests are passing.

algorandskiy (approved):

PR #2159, opened 2021-05-19
bandwidth stats for cluster tests

brianolson:
## Summary
heapWatch.py gains the ability to get a list of hosts from `terraform-inventory` from `algonet` tool. Can have `--token` and `--admin-token` set for talking to those cluster `algod` instances. (Cluster tool is separately updated to set one token across all cluster `algod`.)

metrics_delta.py processes captured /metrics data into stats for each `algod` followed and also produces one summary line of all data processed, e.g.:
`summary: 1091.99 TPS, 26341397 tx B/s, 3490256 rx B/s`

## Test Plan

Tested by several runs of scenario1 on us-east-2 clusters and gathering stats from those experiments.


brianolson:
This is a safe chunk of work that can be committed any time. It's just dev tool python scripts for gathering stats from clusters and analyzing that data.

test/testdata/deployednettemplates/recipes/scenario1/net.json
```diff
@@ -172,6 +172,7 @@
 							"ParticipationOnly": false
 						}
 					],
+					"APIEndpoint": "{{APIEndpoint}}",
```
tsachiherman, line 175: this file is auto-generated. It's ok to merge in the changes here, but you need to track down the source. ( i.e. I think it's `node.json` and `nonPartNode.json`

test/heapwatch/heapWatch.py
```diff
@@ -64,15 +70,24 @@ def do_graceful_stop(signum, frame):
...
-    def __init__(self, path):
+    def __init__(self, path, net=None, token=None, admin_token=None):
         self.path = path
         self.nick = os.path.basename(self.path)
-        net, token, admin_token = read_algod_dir(self.path)
         self.net = net
         self.token = token
         self.admin_token = admin_token
+        if net is None:
+            net, token, admin_token = read_algod_dir(self.path)
```
algonautshant, line 90: Why is this after setting the self (net, token, admin_token)?

brianolson: good catch. fixed.

test/heapwatch/metrics_delta.py
```diff
@@ -67,74 +193,223 @@ def main():
...
+        reportpath = None
+        if args.report:
+            if args.report == '-':
+                writer = csv.writer(sys.stdout)
+            else:
+                if nick is None:
+                    reportpath = args.report
+                elif args.report.endswith('.csv'):
+                    reportpath = args.report[:-4] + nick + '.csv'
+                reportf = open(reportpath, 'wt')
```
algonautshant, line 319: This may accidentally overwrite existing files. 
Why not ask for confirmation to overwrite, append, or write only to new files?

algonautshant (dismissed):
Looks great. 
Would be useful to also report the median along with the mean. Median is more robust in case of extreme values. 
There are some commented printf/logging lines which can be removed if not needed anymore.

test/heapwatch/metrics_delta.py
brianolson: for this analysis script I definitely want it to overwrite by default, in that I frequently tweak the script, run it again, and want to see the new reports.

test/testdata/deployednettemplates/recipes/scenario1/net.json
tsachiherman: bump

brianolson: Had to add a little code to the generator, but this is covered now.

tsachiherman: I'm not sure if the code generator changes were needed. Could you verify that specifying APIEndpoint/APIEndpoint2/APIEndpoint3/APIEndpoint4 wouldn't have worked ?
( btw - even if you "mess-up", the worst case scenario would be a port collision, in which case one of the processes would still be working at the desired port.. right ? )

brianolson: The generator would have needed to change to apply APIEndpoint{N} to different nodes within a host. APIEndpoint[234] appears to only exist on a couple hand written templates. I would expect `algod` starting up and failing to open its APIEndpoint at the specified port would exit; then we'd have fewer nodes for all the colliding nodes.

tsachiherman (approved):

PR #2172, opened 2021-05-21
combine app state key/value size limits

jannotti:
## Summary

Combine the limits enforced on keys and values in Teal such that there is a 128 byte limit on a key/value pair, rather than a 64 byte limit on each.  The previous method meant app creators/users were essentially paying for ("renting") space that might be used by keys, but rarely was, and even if it was,  could serve little purpose.

### DIFF
--- a/config/consensus.go
+++ b/config/consensus.go
@@ -257,6 +257,9 @@ type ConsensusParams struct {
 	// local key/value store
 	MaxAppBytesValueLen int
 
+	// maximum sum of the lengths of the key and value of one app state entry
+	MaxAppSumKeyValueLens int
+
 	// maximum number of applications a single account can create and store
 	// AppParams for at once
 	MaxAppsCreated int
@@ -815,6 +818,7 @@ func initConsensusProtocols() {
 	v24.MaxAppProgramLen = 1024
 	v24.MaxAppKeyLen = 64
 	v24.MaxAppBytesValueLen = 64
+	v24.MaxAppSumKeyValueLens = 128
 
 	// 0.1 Algos (Same min balance cost as an Asset)
 	v24.AppFlatParamsMinBalance = 100000
@@ -903,6 +907,9 @@ func initConsensusProtocols() {
 	// but not yet released in a production protocol version.
 	vFuture := v27
 	vFuture.ApprovedUpgrades = map[protocol.ConsensusVersion]uint64{}
+
+	// Let the bytes value take more space. Key+Value is still limited to 128
+	vFuture.MaxAppBytesValueLen = 127
 
 	// FilterTimeout for period 0 should take a new optimized, configured value, need to revisit this later
 	vFuture.AgreementFilterTimeoutPeriod0 = 4 * time.Second
--- a/config/consensus.go
+++ b/config/consensus.go
@@ -909,7 +909,7 @@ func initConsensusProtocols() {
 	vFuture.ApprovedUpgrades = map[protocol.ConsensusVersion]uint64{}
 
 	// Let the bytes value take more space. Key+Value is still limited to 128
-	vFuture.MaxAppBytesValueLen = 127
+	vFuture.MaxAppBytesValueLen = 128
 
 	// FilterTimeout for period 0 should take a new optimized, configured value, need to revisit this later
 	vFuture.AgreementFilterTimeoutPeriod0 = 4 * time.Second
--- a/config/consensus.go
+++ b/config/consensus.go
@@ -818,7 +818,7 @@ func initConsensusProtocols() {
 	v24.MaxAppProgramLen = 1024
 	v24.MaxAppKeyLen = 64
 	v24.MaxAppBytesValueLen = 64
-	v24.MaxAppSumKeyValueLens = 128
+	v24.MaxAppSumKeyValueLens = 128 // Set here to have no effect until MaxAppBytesValueLen increases
 
 	// 0.1 Algos (Same min balance cost as an Asset)
 	v24.AppFlatParamsMinBalance = 100000
--- a/daemon/algod/api/server/v2/dryrun_test.go
+++ b/daemon/algod/api/server/v2/dryrun_test.go
@@ -351,6 +351,7 @@ func init() {
 	proto.MaxAppProgramCost = 700
 	proto.MaxAppKeyLen = 64
 	proto.MaxAppBytesValueLen = 64
+	proto.MaxAppSumKeyValueLens = 128
 
 	config.Consensus[dryrunProtoVersion] = proto
 }
--- a/data/basics/teal.go
+++ b/data/basics/teal.go
@@ -108,8 +108,11 @@ func (sd StateDelta) Valid(proto *config.ConsensusParams) error {
 		}
 		switch delta.Action {
 		case SetBytesAction:
-			if len(delta.Bytes) > proto.MaxAppBytesValueLen || len(key)+len(delta.Bytes) > proto.MaxAppSumKeyValueLens {
+			if len(delta.Bytes) > proto.MaxAppBytesValueLen {
 				return fmt.Errorf("value too long for key 0x%x: length was %d", key, len(delta.Bytes))
+			}
+			if sum := len(key) + len(delta.Bytes); sum > proto.MaxAppSumKeyValueLens {
+				return fmt.Errorf("key/value too long for key 0x%x: sum was %d", key, sum)
 			}
 		case SetUintAction:
 		case DeleteAction:
--- a/data/basics/teal.go
+++ b/data/basics/teal.go
@@ -112,7 +112,7 @@ func (sd StateDelta) Valid(proto *config.ConsensusParams) error {
 				return fmt.Errorf("value too long for key 0x%x: length was %d", key, len(delta.Bytes))
 			}
 			if sum := len(key) + len(delta.Bytes); sum > proto.MaxAppSumKeyValueLens {
-				return fmt.Errorf("key/value too long for key 0x%x: sum was %d", key, sum)
+				return fmt.Errorf("key/value total too long for key 0x%x: sum was %d", key, sum)
 			}
 		case SetUintAction:
 		case DeleteAction:
--- a/data/basics/teal.go
+++ b/data/basics/teal.go
@@ -108,8 +108,8 @@ func (sd StateDelta) Valid(proto *config.ConsensusParams) error {
 		}
 		switch delta.Action {
 		case SetBytesAction:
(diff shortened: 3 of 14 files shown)


config/consensus.go
```diff
@@ -904,6 +908,9 @@ func initConsensusProtocols() {
 	vFuture := v27
 	vFuture.ApprovedUpgrades = map[protocol.ConsensusVersion]uint64{}
 
+	// Let the bytes value take more space. Key+Value is still limited to 128
+	vFuture.MaxAppBytesValueLen = 127
```
jannotti, line 912: This will likely become 128, once I have unit tests that show a blank key is a legal key.

config/consensus.go
```diff
@@ -815,6 +818,7 @@ func initConsensusProtocols() {
 	v24.MaxAppProgramLen = 1024
 	v24.MaxAppKeyLen = 64
 	v24.MaxAppBytesValueLen = 64
+	v24.MaxAppSumKeyValueLens = 128
```
algorandskiy, line 821: wrong place, needs to be in vFuture

algorandskiy (commented):
Please fix `MaxAppSumKeyValueLens` initialization, otherwise looks good

config/consensus.go
jannotti: By putting it here, the code that checks the sum of the lengths does not need to be separately feature-flagged.  It can check that the sum is  <= this value, even in old txns, in old protocol versions. That is, it is just making explicit something that was always true: the sum <= 128.

algorandskiy: thanks for explaining, let me think if it can break something

algorandskiy (approved):

config/consensus.go
tsachiherman: While the code above would work correctly, it's wouldn't be the easiest to follow. I agree with @pzbitskiy in that it would be cleaner to add it to `vFuture`. The feature-flagging doesn't have to be fancy though; using
```golang
//  on version v27 we added MaxAppSumKeyValueLens to allow (key+value) < 128 bytes instead of key/value being 64 bytes each.
if sum := len(key) + len(value.Bytes);  cb.proto.MaxAppSumKeyValueLens != 0 && sum > cb.proto.MaxAppSumKeyValueLens {
```
instead of 
```golang
if sum := len(key) + len(value.Bytes); sum > cb.proto.MaxAppSumKeyValueLens {
```

would do it.

tsachiherman (changes requested):
Could you please add a unit test/e2e test that exercises the new code path (in particular - the fail cases ) and verify these?

jannotti:
> Could you please add a unit test/e2e test that exercises the new code path (in particular - the fail cases ) and verify these?

test/scripts/e2e_subs/v26/app-value-under-64.sh tests the success and failure in old proto
test/scripts/e2e_subs/app-key-value-128.sh does so under new proto

tsachiherman:
> > Could you please add a unit test/e2e test that exercises the new code path (in particular - the fail cases ) and verify these?
> 
> test/scripts/e2e_subs/v26/app-value-under-64.sh tests the success and failure in old proto
> test/scripts/e2e_subs/app-key-value-128.sh does so under new proto

I was hoping to see an explicit test that verifies the output is `value too long for key 0x%x: length was %d` and `key/value too long for key 0x%x: sum was %d` ( ideally with exact predefined values). Testing just `value too long` isn't very specific and could originate from other locations.

jannotti:
> I was hoping to see an explicit test that verifies the output is `value too long for key 0x%x: length was %d` and `key/value too long for key 0x%x: sum was %d` ( ideally with exact predefined values). Testing just `value too long` isn't very specific and could originate from other locations.

I don't like to test the complete exact strings, as I don't think that's really part of the spec.  But I have added a test so that I'm now testing each failure case (the value is too big on its own and the key + val is too big) And I've put in enough string checking to distinguish the two cases.

tsachiherman:
> > I was hoping to see an explicit test that verifies the output is `value too long for key 0x%x: length was %d` and `key/value too long for key 0x%x: sum was %d` ( ideally with exact predefined values). Testing just `value too long` isn't very specific and could originate from other locations.
> 
> I don't like to test the complete exact strings, as I don't think that's really part of the spec. But I have added a test so that I'm now testing each failure case (the value is too big on its own and the key + val is too big) And I've put in enough string checking to distinguish the two cases.

Oh no.. my intention wasn't to be spec compliant here. The reason I've asked for these testing is that other places in the code base could either now, or in the future, to produce that sub-string. For instance, a potential check for the note length might produce the error "invalid transaction, note field value too long", which would also match the one you've tested against.
If this error string companion was done on a unit test level, it would be fine, since the scope is well defined. However, when we interact with goal to "extract" an internal error string, we need to be very precise about the error testing. ( otherwise, these tests would not be as useful for us for the long run )

jannotti:
> > > I was hoping to see an explicit test that verifies the output is `value too long for key 0x%x: length was %d` and `key/value too long for key 0x%x: sum was %d` ( ideally with exact predefined values). Testing just `value too long` isn't very specific and could originate from other locations.
> > 
> > 
> > I don't like to test the complete exact strings, as I don't think that's really part of the spec. But I have added a test so that I'm now testing each failure case (the value is too big on its own and the key + val is too big) And I've put in enough string checking to distinguish the two cases.
> 
> Oh no.. my intention wasn't to be spec compliant here. The reason I've asked for these testing is that other places in the code base could either now, or in the future, to produce that sub-string. For instance, a potential check for the note length might produce the error "invalid transaction, note field value too long", which would also match the one you've tested against.
> If this error string companion was done on a unit test level, it would be fine, since the scope is well defined. However, when we interact with goal to "extract" an internal error string, we need to be very precise about the error testing. ( otherwise, these tests would not be as useful for us for the long run )

I did not think you were asking for the check for spec compliance, I brought up spec compliance to explain why I objected to a brittle test. I think it's bad to explicitly test things that are not part of the spec.  Someone reading the test would reasonably conclude that it is critical that the message does not change - after all, we're testing it, so it must matter.  Further, it leads to a code base in which nobody improves error messages, because they don't want to deal with the hassle of a slew of overly precise tests.

By testing a substring, we're providing a strong hint that the details don't really matter here, we're just testing for a failure message.  In the end, I think the right test does not check strings at all.  It tests for success, right up until the limit being considered, and tests for failure afterward, regardless of the exact message emitted.  I don't want to commit yet another fix in this vein unless we agree, however.  I'm happy to change the test to do any or all of:

a) Check the exit status, not the message
b) Check that *only* the one message is emitted
c) Check that no such message was emitted during success

But I really don't want to have a test that tests an irrelevant detail, that we chose a particular specific grammatical construction to convey the error. I don't believe that protecting us against a situation where our code simultaneously breaks, while some other code begins to emit a very similar error message is worth the negative effect of brittle tests.

tsachiherman:
> > > > I was hoping to see an explicit test that verifies the output is `value too long for key 0x%x: length was %d` and `key/value too long for key 0x%x: sum was %d` ( ideally with exact predefined values). Testing just `value too long` isn't very specific and could originate from other locations.
> > > 
> > > 
> > > I don't like to test the complete exact strings, as I don't think that's really part of the spec. But I have added a test so that I'm now testing each failure case (the value is too big on its own and the key + val is too big) And I've put in enough string checking to distinguish the two cases.
> > 
> > 
> > Oh no.. my intention wasn't to be spec compliant here. The reason I've asked for these testing is that other places in the code base could either now, or in the future, to produce that sub-string. For instance, a potential check for the note length might produce the error "invalid transaction, note field value too long", which would also match the one you've tested against.
> > If this error string companion was done on a unit test level, it would be fine, since the scope is well defined. However, when we interact with goal to "extract" an internal error string, we need to be very precise about the error testing. ( otherwise, these tests would not be as useful for us for the long run )
> 
> I did not think you were asking for the check for spec compliance, I brought up spec compliance to explain why I objected to a brittle test. I think it's bad to explicitly test things that are not part of the spec. Someone reading the test would reasonably conclude that it is critical that the message does not change - after all, we're testing it, so it must matter. Further, it leads to a code base in which nobody improves error messages, because they don't want to deal with the hassle of a slew of overly precise tests.
> 
> By testing a substring, we're providing a strong hint that the details don't really matter here, we're just testing for a failure message. In the end, I think the right test does not check strings at all. It tests for success, right up until the limit being considered, and tests for failure afterward, regardless of the exact message emitted. I don't want to commit yet another fix in this vein unless we agree, however. I'm happy to change the test to do any or all of:
> 
> a) Check the exit status, not the message
> b) Check that _only_ the one message is emitted
> c) Check that no such message was emitted during success
> 
> But I really don't want to have a test that tests an irrelevant detail, that we chose a particular specific grammatical construction to convey the error. I don't believe that protecting us against a situation where our code simultaneously breaks, while some other code begins to emit a very similar error message is worth the negative effect of brittle tests.
> 
> Essentially, I'm proposing that this is the better test:
> 
> ```
> call write global $BIG64 $BIG64
> call check global $BIG64 $BIG64
> call write global $BIG64 ${BIG64}X && exit 1
> ```
> 
> It confirms that things are fine until one extra character is added to the value, and then goal must fail.
> 
> Potentially include these, but this goes beyond what I consider an integration test should do
> 
> ```
> # These test some details of the checking, using the error message to
> # confirm which code path is being tested.  Details of strings are irrelevant
> set +o pipefail
> call write global $BIG64 ${BIG64}X 2>&1 | grep "value too long" | grep sum
> # This value so big that it fails before the sum is considered
> call write global $BIG64 ${BIG64}${BIG64}X 2>&1 | grep "value too long" | grep length
> set -o pipefail
> ```

I think that you're going too far with that, and there is no need to. The only thing that I asked was to ensure that we provide a test to confirm the newly generated errors. I agree that testing it via goal would be sub-optimal, since we loose the originating datatype, and forced to use the formatted string.

Adding a unit test for `StateDelta.Valid` and `roundCowState.SetKey` which tests for these explicit errors would be enough, as you wouldn't have exterior "noise". If you want to provide a really good testing for that, you could define custom errors and test against error class type instead of checking strings.

algorandskiy (approved):
IMO tests are ok, the only question is to be in agreement on adding new non-empty values into past protocols.

tsachiherman:
> IMO tests are ok, the only question is to be in agreement on adding new non-empty values into past protocols.

If we would start supporting different acceptable for the same block on different binaries before switching to the new protocol, that would be an issue. Is that the case ?

algorandskiy:
I attempted to model such split but it looks like impossible.

jannotti:
Requested tests added. @tsachiherman please have a look and approve or ask for more.

tsachiherman:
Will do.

tsachiherman (approved):

PR #2173, opened 2021-05-21
pool fees in a txgroup

jannotti:
## Summary

A fair amount of complexity is introduced in transaction groups when the goal is to let some entity perform an action at the expense of another.  For example, a contract account might be willing to perform an exchange, but expects the caller to compensate it to replace the fee that the contract account must pay.

This changes fee accounting to simplify these situations.  Rather than check that *each* txn in a txn group meets the min fee, the txgroup is checked as a whole to ensure the total fee exceeds n*min_fee for an n member txgroup.

Topic for discussion: This change allows an *explicit* --fee=0 or --fee=1 on a goal command to create a no, or low, fee transaction.  Previously it would have silently raised the fee to min fee. If fee is completely left off command-line, it is set to minfee.

## Test Plan

Unit and Integration tests

### DIFF
--- a/config/consensus.go
+++ b/config/consensus.go
@@ -104,7 +104,8 @@ type ConsensusParams struct {
 	MinTxnFee uint64
 
 	// EnableFeePooling specifies that the sum of the fees in a
-	// group must exceed one MinTxnFee, rather check on each Txn.
+	// group must exceed one MinTxnFee per Txn, rather than check that
+	// each Txn has a MinFee.
 	EnableFeePooling bool
 
 	// RewardUnit specifies the number of MicroAlgos corresponding to one reward
--- a/config/consensus.go
+++ b/config/consensus.go
@@ -103,6 +103,10 @@ type ConsensusParams struct {
 	// a way of making the spender subsidize the cost of storing this transaction.
 	MinTxnFee uint64
 
+	// EnableFeePooling specifies that the sum of the fees in a
+	// group must exceed one MinTxnFee, rather check on each Txn.
+	EnableFeePooling bool
+
 	// RewardUnit specifies the number of MicroAlgos corresponding to one reward
 	// unit.
 	//
@@ -926,6 +930,8 @@ func initConsensusProtocols() {
 
 	// Increase asset URL length to allow for IPFS URLs
 	vFuture.MaxAssetURLBytes = 96
+
+	vFuture.EnableFeePooling = true
 
 	Consensus[protocol.ConsensusFuture] = vFuture
 }
--- a/data/transactions/verify/txn.go
+++ b/data/transactions/verify/txn.go
@@ -108,10 +108,6 @@ func Txn(s *transactions.SignedTxn, txnIdx int, groupCtx *GroupContext) error {
 		return err
 	}
 
-	if s.Txn.Src().IsZero() {
-		return errors.New("empty address")
-	}
-
 	return stxnVerifyCore(s, txnIdx, groupCtx)
 }
 
@@ -121,13 +117,26 @@ func TxnGroup(stxs []transactions.SignedTxn, contextHdr bookkeeping.BlockHeader,
 	if err != nil {
 		return nil, err
 	}
+
+	requiredFees := uint64(0)
+	feeSum := uint64(0)
 	for i, stxn := range stxs {
 		err = Txn(&stxn, i, groupCtx)
 		if err != nil {
 			err = fmt.Errorf("transaction %+v invalid : %w", stxn, err)
 			return
 		}
+		if stxn.Txn.Type != protocol.CompactCertTx {
+			requiredFees++
+		}
+		feeSum += stxn.Txn.Fee.Raw
 	}
+	if feeSum < groupCtx.consensusParams.MinTxnFee*requiredFees {
+		err = fmt.Errorf("txgroup had %d in fees, which is less than the minimum %d * %d",
+			feeSum, requiredFees, groupCtx.consensusParams.MinTxnFee)
+		return
+	}
+
 	if cache != nil {
 		cache.Add(stxs, groupCtx)
 	}
--- a/data/transactions/verify/txn.go
+++ b/data/transactions/verify/txn.go
@@ -118,8 +118,8 @@ func TxnGroup(stxs []transactions.SignedTxn, contextHdr bookkeeping.BlockHeader,
 		return nil, err
 	}
 
-	requiredFees := uint64(0)
-	feeSum := uint64(0)
+	minFeeCount := uint64(0)
+	feesPaid := uint64(0)
 	for i, stxn := range stxs {
 		err = Txn(&stxn, i, groupCtx)
 		if err != nil {
@@ -127,13 +127,21 @@ func TxnGroup(stxs []transactions.SignedTxn, contextHdr bookkeeping.BlockHeader,
 			return
 		}
 		if stxn.Txn.Type != protocol.CompactCertTx {
-			requiredFees++
+			minFeeCount++
 		}
-		feeSum += stxn.Txn.Fee.Raw
+		feesPaid = basics.AddSaturate(feesPaid, stxn.Txn.Fee.Raw)
 	}
-	if feeSum < groupCtx.consensusParams.MinTxnFee*requiredFees {
+	feeNeeded, overflow := basics.OMul(groupCtx.consensusParams.MinTxnFee, minFeeCount)
+	if overflow {
+		err = fmt.Errorf("txgroup fee requirement overflow")
+		return
+	}
+	// feesPaid may have saturated. That's ok. Since we know
+	// feeNeeded did not overlfow, simple comparision tells us
+	// feesPaid was enough.
+	if feesPaid < feeNeeded {
 		err = fmt.Errorf("txgroup had %d in fees, which is less than the minimum %d * %d",
-			feeSum, requiredFees, groupCtx.consensusParams.MinTxnFee)
+			feesPaid, minFeeCount, groupCtx.consensusParams.MinTxnFee)
 		return
 	}
 
--- a/ledger/apply/payment_test.go
+++ b/ledger/apply/payment_test.go
@@ -221,6 +221,11 @@ func TestPaymentValidation(t *testing.T) {
 		if badFee.WellFormed(spec, tc.Proto) == nil {
 			t.Errorf("transaction with no fee %#v verified incorrectly", badFee)
 		}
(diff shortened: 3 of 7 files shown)


data/transactions/verify/txn.go
```diff
@@ -121,13 +117,26 @@ func TxnGroup(stxs []transactions.SignedTxn, contextHdr bookkeeping.BlockHeader,
...
 	for i, stxn := range stxs {
 		err = Txn(&stxn, i, groupCtx)
 		if err != nil {
 			err = fmt.Errorf("transaction %+v invalid : %w", stxn, err)
 			return
 		}
+		if stxn.Txn.Type != protocol.CompactCertTx {
+			requiredFees++
+		}
+		feeSum += stxn.Txn.Fee.Raw
```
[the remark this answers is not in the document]
zeldovich: To Pavel's point -- please avoid using the `Raw` field for precisely this reason; use the overflow-checked arithmetic operations on `MicroAlgos`.

config/consensus.go
```diff
@@ -103,6 +103,10 @@ type ConsensusParams struct {
 	// a way of making the spender subsidize the cost of storing this transaction.
 	MinTxnFee uint64
 
+	// EnableFeePooling specifies that the sum of the fees in a
+	// group must exceed one MinTxnFee, rather check on each Txn.
```
zeldovich, line 107: I think this comment is out-of-date with what you are doing: perhaps it should say "one MinTxnFee per transaction in group"?

data/transactions/verify/txn.go
```diff
@@ -108,10 +108,6 @@ func Txn(s *transactions.SignedTxn, txnIdx int, groupCtx *GroupContext) error {
 		return err
 	}
 
-	if s.Txn.Src().IsZero() {
-		return errors.New("empty address")
-	}
-
```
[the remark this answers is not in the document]
jannotti: It was being checked inside wellFormed anyway, which seems like the right place.

data/transactions/verify/txn.go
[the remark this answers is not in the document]
jannotti: Fair, I'll convert.

data/transactions/verify/txn.go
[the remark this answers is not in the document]
jannotti: It contains
```
	if tx.Sender.IsZero() {
		return fmt.Errorf("transaction cannot have zero sender")
	}
```
And Src() returns tx.Sender

data/transactions/verify/txn.go
```diff
@@ -121,13 +117,34 @@ func TxnGroup(stxs []transactions.SignedTxn, contextHdr bookkeeping.BlockHeader,
...
+	}
+	// feesPaid may have saturated. That's ok. Since we know
+	// feeNeeded did not overlfow, simple comparision tells us
+	// feesPaid was enough.
+	if feesPaid < feeNeeded {
+		err = fmt.Errorf("txgroup had %d in fees, which is less than the minimum %d * %d",
+			feesPaid, minFeeCount, groupCtx.consensusParams.MinTxnFee)
+		return
+	}
+
```
algorandskiy, line 147: The overflow fix looks good. But this entire code block needs to be under `proto.EnableFeePooling` switch, because in case of overflow a node with new code rejects the group and old non-upgraded node approves it since there is no such comparison. I understand this is not an issue on mainnet/testnet but still a possible problem on a some network with a supply larger than 2^64/10^6.

jannotti: I don't think that can happen.

It's not a matter of supply.  The only way this code overflows *and rejects* is if the total fee expected for the txgroup overflows on line 136 (not the fees paid).  That is, only if the network is expecting a fee for a single txgroup to be over 2^64.  I don't think such a txgroup can ever be paid for - rejection is inevitable.

The fees PAID can saturate without causing harm. So if the concern is someone deciding to screw around and put in a massive fee, far above the requirement, that won't cause a problem. It will saturate, and that will be higher than feeNeeded, so there's no rejection.  They'll end up paying the massive fee. (Though again, even this saturation is also impossible - there can't be more than 2^64 of supply, so how could so much be put into fees?)

If a txgroup's expected fee is > 2^64, that would be a problem, but no algorand network can have a supply high enough to pay such a fee. So if you configured a network to demand it, that txngroup is going to get rejected one way or another.

So I claim this check will never reject anything that the !EnableFeePooling accepts.

algorandskiy: you are right (again!). `feesPaid` can't overflow and it would be 2^64 in this case. Then rejection in `feesPaid < feeNeeded` will never happen.

algorandskiy (approved):
Looks good, @zeldovich please take another look as well.

jannotti:
The reason I asked for @zeldovich to have a look is because having the possibility of zero fee transactions feels strange enough that I wanted more heads on what I may have missed, it feels like an assumption that could be baked into some strange places.  For example, I need to go through and make sure we msgpack such things properly on all sdks.  Presumably we should be omitting the fee field entirely, but any sdk that does the omission of empty msgpack fields by hand will likely get it wrong (python, for example).  Anything else weird like that?

zeldovich:
I can't think of anything that would break with this change.

The encoding should be a non-issue.  There's exactly one way to canonically encode every value, including zero.  The encoding of zero happens to be done by omitting the field from the encoding.  Broken msgpack encoders could indeed be broken, but I'd be slightly surprised if canonical-msgpack encoders special-cased fields like the fee.  Even the msgpack encoder we wrote in C for the Ledger hardware wallet will do the right thing, I believe.

The overflow check looks reasonable to me too.

jannotti:
> The encoding should be a non-issue. There's exactly one way to canonically encode every value, including zero. The encoding of zero happens to be done by omitting the field from the encoding. Broken msgpack encoders could indeed be broken, but I'd be slightly surprised if canonical-msgpack encoders special-cased fields like the fee. Even the msgpack encoder we wrote in C for the Ledger hardware wallet will do the right thing, I believe.

Isn't our "emitempty" a local change?  Certainly msgpacktool does not normally eliminate zeros.
```
$ echo '{"a": 0}' | msgpacktool -e | msgpacktool -d
{
  "a": 0
}
```
In order to do that, msgpack needs the "a" to be encoded.  go-algorand can only do it because it unpacks into well-typed go structs.  In the python sdk we have to carefully strip out fields that should be omitted.  Otherwise they are encoded by standard msgpack and signatures fail. (I assume that the go-algorand side is re-encoding into what it considers canonical before checking the signature.)

zeldovich:
Right, our canonical-msgpack differs from regular msgpack.

`msgpacktool` does not produce canonical encodings; it's a debug tool.

When we talk about signing something in Algorand (or verifying the signature of something), it's almost always signing a canonical-msgpack encoding of an object, with a domain-separation prefix (`HashID`).  We should never sign non-canonical msgpack encodings, and we always compute the canonical-msgpack encoding when veerifying a signature.

jannotti:
Right, it does two things. It sorts the keys before putting the values in an OrderedDict, and makes sure it only inserts the right keys.  Here's a typical "dictify" method.  This is what gets passed to msgpack.

```
    def dictify(self):
        d = dict()
        if self.amt:
            d["amt"] = self.amt
        if self.close_remainder_to:
            d["close"] = encoding.decode_address(self.close_remainder_to)

        decoded_receiver = encoding.decode_address(self.receiver)
        if any(decoded_receiver):
            d["rcv"] = encoding.decode_address(self.receiver)

        d.update(super(PaymentTxn, self).dictify())
        od = OrderedDict(sorted(d.items()))

        return od
```

Note that amt is not added if 0, and the sorted items being put into the OrderedDict. 

Then a normal msgpack encoding is used.

```
            enc = msgpack.packb({"txn": txn.dictify()}, use_bin_type=True)
```
(use_bin_type has to do with choosing strings or bytes encoding)

Zero values are normally encoded in a msgpack map.  From their homepage:
Screen Shot 2021-05-26 at 10 34 58 AM

zeldovich:
Neat, it's convenient that `python-msgpack` already follows the same rules for integers that we require.

Indeed it looks like we should fix the SDK to encode correctly.  But the damage seems like it will be self-evident and won't dramatically break things: using this new group-fee feature through an SDK with a broken encoder will produce transactions with bad signatures.

data/transactions/verify/txn.go
```diff
@@ -121,13 +117,34 @@ func TxnGroup(stxs []transactions.SignedTxn, contextHdr bookkeeping.BlockHeader,
...
+		}
+		feesPaid = basics.AddSaturate(feesPaid, stxn.Txn.Fee.Raw)
 	}
+	feeNeeded, overflow := basics.OMul(groupCtx.consensusParams.MinTxnFee, minFeeCount)
+	if overflow {
+		err = fmt.Errorf("txgroup fee requirement overflow")
+		return
+	}
+	// feesPaid may have saturated. That's ok. Since we know
+	// feeNeeded did not overlfow, simple comparision tells us
```
jdtzmn, line 140: nit: overflow

jannotti: Don't make me face the wrath of CI again!

jdtzmn: Sorry..

ledger/apply/payment_test.go
```diff
@@ -221,6 +221,11 @@ func TestPaymentValidation(t *testing.T) {
 		if badFee.WellFormed(spec, tc.Proto) == nil {
 			t.Errorf("transaction with no fee %#v verified incorrectly", badFee)
 		}
+		badFee.Fee.Raw = 1
+		if badFee.WellFormed(spec, tc.Proto) == nil {
+			t.Errorf("transaction with low fee %#v verified incorrectly", badFee)
+		}
```
jdtzmn, line 227: I'm confused what this checks. `WellFormed` seems to confirm that individual transactions are well formed but this PR's implementation deals specifically with group transactions. Wouldn't a unit test for this PR rather test that transaction group fees are correct, as opposed to individual transactions?

jdtzmn (approved):
Had one question about the unit test, but otherwise this looks good!

ledger/apply/payment_test.go
jannotti: Good point.  I think this unit test should have existed before.  I wrote it while I was still feeling my way around, before I wrote any of this code.  It will, in fact, never be triggered after the protocol upgrade that this code requires.

But it does give me something to test against when writing the integration tests.  I wanted to confirm that a low fee transaction *was* bad, but now is not.

algorandskiy (approved):

PR #2175, opened 2021-05-21
Adding missing test for universalFetcher.go

algonautshant:
## Summary

Adding missing test for universalFetcher.go

Code coverage increased from 74%->96%
## Test Plan
Tests added.

### DIFF
--- a/catchup/fetcher_test.go
+++ b/catchup/fetcher_test.go
@@ -232,6 +232,7 @@ type testUnicastPeer struct {
 	version          string
 	responseChannels map[uint64]chan *network.Response
 	t                *testing.T
+	responseOverride *network.Response 
 }
 
 func (p *testUnicastPeer) GetAddress() string {
@@ -254,6 +255,10 @@ func (p *testUnicastPeer) Request(ctx context.Context, tag protocol.Tag, topics 
 	require.NotNil(p.t, dispather)
 	dispather.Handle(network.IncomingMessage{Tag: tag, Data: topics.MarshallTopics(), Sender: p, Net: p.gn})
 
+	if p.responseOverride != nil {
+		return p.responseOverride, nil
+	}
+	
 	// wait for the channel.
 	select {
 	case resp = <-responseChannel:
@@ -297,10 +302,14 @@ func (p *testUnicastPeer) Unicast(ctx context.Context, msg []byte, tag protocol.
 }
 
 func makeTestUnicastPeer(gn network.GossipNode, t *testing.T) network.UnicastPeer {
+	return makeTestUnicastPeerWithResponseOverride(gn, t, nil)
+}
+func makeTestUnicastPeerWithResponseOverride(gn network.GossipNode, t *testing.T, responseOverride *network.Response) network.UnicastPeer {
 	wsp := testUnicastPeer{}
 	wsp.gn = gn
 	wsp.t = t
 	wsp.version = network.ProtocolVersion
 	wsp.responseChannels = make(map[uint64]chan *network.Response)
+	wsp.responseOverride = responseOverride
 	return &wsp
 }
--- a/catchup/fetcher_test.go
+++ b/catchup/fetcher_test.go
@@ -304,6 +304,7 @@ func (p *testUnicastPeer) Unicast(ctx context.Context, msg []byte, tag protocol.
 func makeTestUnicastPeer(gn network.GossipNode, t *testing.T) network.UnicastPeer {
 	return makeTestUnicastPeerWithResponseOverride(gn, t, nil)
 }
+
 func makeTestUnicastPeerWithResponseOverride(gn network.GossipNode, t *testing.T, responseOverride *network.Response) network.UnicastPeer {
 	wsp := testUnicastPeer{}
 	wsp.gn = gn
--- a/catchup/universalFetcher.go
+++ b/catchup/universalFetcher.go
@@ -94,17 +94,17 @@ func processBlockBytes(fetchedBuf []byte, r basics.Round, debugStr string) (blk 
 	var decodedEntry rpcs.EncodedBlockCert
 	err = protocol.Decode(fetchedBuf, &decodedEntry)
 	if err != nil {
-		err = fmt.Errorf("fetchBlock(%d): cannot decode block from peer %v: %v", r, debugStr, err)
+		err = fmt.Errorf("processBlockBytes(%d): cannot decode block from peer %v: %v", r, debugStr, err)
 		return
 	}
 
 	if decodedEntry.Block.Round() != r {
-		err = fmt.Errorf("fetchBlock(%d): got wrong block from peer %v: wanted %v, got %v", r, debugStr, r, decodedEntry.Block.Round())
+		err = fmt.Errorf("processBlockBytes(%d): got wrong block from peer %v: wanted %v, got %v", r, debugStr, r, decodedEntry.Block.Round())
 		return
 	}
 
 	if decodedEntry.Certificate.Round != r {
-		err = fmt.Errorf("fetchBlock(%d): got wrong cert from peer %v: wanted %v, got %v", r, debugStr, r, decodedEntry.Certificate.Round)
+		err = fmt.Errorf("processBlockBytes(%d): got wrong cert from peer %v: wanted %v, got %v", r, debugStr, r, decodedEntry.Certificate.Round)
 		return
 	}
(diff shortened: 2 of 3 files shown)


catchup/universalFetcher_test.go
```diff
@@ -127,3 +131,136 @@ func TestUGetBlockUnsupported(t *testing.T) {
...
+
+	net.addPeer(rootURL)
+	fetcher := makeUniversalBlockFetcher(logging.TestingLog(t), net, cfg)
+
+	ls.status = http.StatusBadRequest
+	_, _, _, err := fetcher.fetchBlock(context.Background(), 1, net.GetPeers()[0])
+	require.Regexp(t,
+		"getBlockBytes error response status code 400 when requesting .* Response body",
+		err.Error())
+
```
tsachiherman, line 248: While this is probably would work, a better way would be to create a custom error object and return it.
Then, you can verify that you received the expected error type rather then trying to compare it with the formatted text.

catchup/universalFetcher_test.go
```diff
@@ -127,3 +131,136 @@ func TestUGetBlockUnsupported(t *testing.T) {
...
+}
+
+// TestGetBlockBytesHTTPErrors tests the errors reported from getblockBytes for http peer
+func TestGetBlockBytesHTTPErrors(t *testing.T) {
+
+	cfg := config.GetDefaultLocal()
+	net := &httpTestPeerSource{}
+
+	ls := &TestHTTPHandler{}
+	//rpcs.MakeBlockService(logging.Base(), blockServiceConfig, ledger, net, "test genesisID")
```
tsachiherman, line 232: remove commented out lines if not needed.

catchup/universalFetcher_test.go
```diff
@@ -127,3 +131,136 @@ func TestUGetBlockUnsupported(t *testing.T) {
...
+}
+
+func (thh *TestHTTPHandler) ServeHTTP(response http.ResponseWriter, request *http.Request) {
+	for _, c := range thh.content {
+		response.Header().Add("Content-Type", c)
+	}
+	response.WriteHeader(thh.status)
+	bytes := make([]byte, 1)
+	if thh.exceedLimit {
+		bytes = make([]byte, 5242881)
```
tsachiherman, line 219: instead of picking a magic number, use the limit and add one.

tsachiherman:
Once you're done with the Changes, please hit the "re-request review" ? Once you do, I'll take another look. Overall looks good already.

catchup/universalFetcher.go
```diff
@@ -90,21 +90,21 @@ func (uf *universalBlockFetcher) fetchBlock(ctx context.Context, round basics.Ro
 	return block, cert, downloadDuration, err
 }
 
-func processBlockBytes(fetchedBuf []byte, r basics.Round, debugStr string) (blk *bookkeeping.Block, cert *agreement.Certificate, err error) {
+func processBlockBytes(fetchedBuf []byte, r basics.Round, peerAddr string) (blk *bookkeeping.Block, cert *agreement.Certificate, err error) {
 	var decodedEntry rpcs.EncodedBlockCert
 	err = protocol.Decode(fetchedBuf, &decodedEntry)
 	if err != nil {
-		err = fmt.Errorf("fetchBlock(%d): cannot decode block from peer %v: %v", r, debugStr, err)
+		err = makeCannotDecodeBlockError(r, peerAddr, err.Error())
```
tsachiherman, line 97: That would defeat the purpose of having a customized error object. Instead, pass the err to the `makeCannotDecodeBlockError` method. The message formatting can remain the same, but you should add an Unwrap method which would return the inner error. Read https://blog.golang.org/go1.13-errors for more info.

catchup/universalFetcher.go
```diff
@@ -270,3 +270,99 @@ func (hf *HTTPFetcher) getBlockBytes(ctx context.Context, r basics.Round) (data
...
+}
+
+func makeWrongBlockFromPeerError(round, certRound basics.Round, peer string) wrongBlockFromPeerError {
+	return wrongBlockFromPeerError{
+		round:     round,
+		peer:      peer,
+		certRound: certRound}
+}
+func (wbfpe wrongBlockFromPeerError) Error() string {
+	return fmt.Sprintf("processBlockBytes: got wrong block from peer %v: wanted %v, got %v",
```
tsachiherman, line 304: nit: use `%d` for integers, or rounds instead of `%v`

catchup/universalFetcher_test.go
```diff
@@ -127,3 +132,150 @@ func TestUGetBlockUnsupported(t *testing.T) {
...
+	blkData := protocol.Encode(&blk)
+	bc := protocol.EncodeReflect(rpcs.PreEncodedBlockCert{
+		Block: blkData,
+	})
+
+	// Check for cert error
+	_, _, err := processBlockBytes(bc, 22, "test")
+	var wcfpe wrongCertFromPeerError
+	require.True(t, errors.As(err, &wcfpe))
+	err.Error()
```
tsachiherman, line 154: why ?

catchup/universalFetcher.go
```diff
@@ -270,3 +270,99 @@ func (hf *HTTPFetcher) getBlockBytes(ctx context.Context, r basics.Round) (data
 func (hf *HTTPFetcher) address() string {
 	return hf.rootURL
 }
+
+type wrongCertFromPeerError struct {
```
tsachiherman, line 274: note that the common naming conventions for errors are "ErrXXX". in your case, leave it with lower-case, but still name it `errWrongCertFromPeer` rather then `wrongCertFromPeerError`

catchup/universalFetcher.go
algonautshant: I'll add the Unwrap method. However, in this case it is not useful. msgp returns msgp.errWrapped, which does not implement Unwrap, and it is only used with Error() call, which msgp internally unwraps.

catchup/universalFetcher_test.go
algonautshant: Replaced this with a proper test to test the error types.

tsachiherman (approved):
looks great. Thanks for the changes.

catchup/fetcher_test.go
```diff
@@ -297,10 +302,14 @@ func (p *testUnicastPeer) Unicast(ctx context.Context, msg []byte, tag protocol.
 }
 
 func makeTestUnicastPeer(gn network.GossipNode, t *testing.T) network.UnicastPeer {
+	return makeTestUnicastPeerWithResponseOverride(gn, t, nil)
+}
```
algorandskiy, line 306: add an empty line

catchup/universalFetcher.go
```diff
@@ -270,3 +270,102 @@ func (hf *HTTPFetcher) getBlockBytes(ctx context.Context, r basics.Round) (data
...
+	peer      string
+	certRound basics.Round
+}
+
+func makeErrWrongCertFromPeer(round, certRound basics.Round, peer string) errWrongCertFromPeer {
+	return errWrongCertFromPeer{
+		round:     round,
+		peer:      peer,
+		certRound: certRound}
+}
```
algorandskiy, line 285: line

catchup/universalFetcher.go
```diff
@@ -90,21 +90,21 @@ func (uf *universalBlockFetcher) fetchBlock(ctx context.Context, round basics.Ro
 	return block, cert, downloadDuration, err
 }
 
-func processBlockBytes(fetchedBuf []byte, r basics.Round, debugStr string) (blk *bookkeeping.Block, cert *agreement.Certificate, err error) {
+func processBlockBytes(fetchedBuf []byte, r basics.Round, peerAddr string) (blk *bookkeeping.Block, cert *agreement.Certificate, err error) {
 	var decodedEntry rpcs.EncodedBlockCert
 	err = protocol.Decode(fetchedBuf, &decodedEntry)
 	if err != nil {
-		err = fmt.Errorf("fetchBlock(%d): cannot decode block from peer %v: %v", r, debugStr, err)
+		err = makeErrCannotDecodeBlock(r, peerAddr, err)
```
algorandskiy, line 97: it is interesting, since `error` is interface, I thought one needs to return a pointer to a struct. But attaching `Error()`  to the type, not pointer looks working, nice.

algonautshant: looks like a pointer or struct can be passed for an interface both for return type, or a function parameter.

PR #2176, opened 2021-05-21
testing: improve e2e protocol upgrade tests

tsachiherman:
## Summary

This PR addresses a bug in the protocol upgrade tests where the `AgreementFilterTimeout` was not set correctly.
In addition, this test removes `SendPaymentFromUnencryptedWallet()` and replaces it with separate `SendPaymentFromWallet()` and `GetUnencryptedWalletHandle()` calls for higher efficiency.

Last, this change *disables* the tests parallelism. Running the upgrade tests in parallel seems to be slower overall. This is still pending investigation, but can be decoupled from these changes for now.

## Test Plan

This is a test.

### DIFF
--- a/test/e2e-go/upgrades/send_receive_upgrade_test.go
+++ b/test/e2e-go/upgrades/send_receive_upgrade_test.go
@@ -82,6 +82,9 @@ func generateFastUpgradeConsensus() (fastUpgradeProtocols config.ConsensusProtoc
 		fastParams.MaxUpgradeWaitRounds = 0
 		fastParams.MaxVersionStringLen += len(consensusTestFastUpgrade(""))
 		fastParams.ApprovedUpgrades = make(map[protocol.ConsensusVersion]uint64)
+		// set the small lambda to 500 for the duration of dependent tests.
+		fastParams.AgreementFilterTimeout = 500 * time.Millisecond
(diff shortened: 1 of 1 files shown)


test/e2e-go/upgrades/send_receive_upgrade_test.go
```diff
@@ -141,23 +141,41 @@ func testAccountsCanSendMoneyAcrossUpgrade(t *testing.T, templatePath string) {
...
+			expectedPongBalance = expectedPongBalance - transactionFee - amountPongSendsPing + amountPingSendsPong
+
+			lastTxnSendRound = curStatus.LastRound
+		}
 
 		curStatus, err = pongClient.Status()
 		a.NoError(err)
 
-		time.Sleep(time.Second)
+		pongWalletHandle, err = pongClient.GetUnencryptedWalletHandle()
```
algonautshant, line 170: why this needs to be called at each iteration?

algonautshant (approved):
Looks great.

test/e2e-go/upgrades/send_receive_upgrade_test.go
tsachiherman: The handle would expire after a while.. so it need to be refreshed frequently. I could have added a error-handling for that and refresh the handle only when needed, but I figured out that this would likely to be an overkill for this scenario.

PR #2183, opened 2021-05-24
Clean up `goal wallet new` output

cce:
## Summary

When you run `goal wallet new` there is a confusing warning ("`One or more non-printable characters ...`") presented just before the backup phrase, which is displayed between some sanitized ANSI codes. This is due to security improvements in #1585 to prevent `goal` from printing control characters. This PR removes the ANSI color formatting altogether from the `infoBackupPhrase` (already done for Windows in #1942), which is one simple way to clean up this output and remove the warning message.

### Before
```
~$ goal wallet new test-wallet -d testnet.algorand
Please choose a password for wallet 'test-wallet':
Please confirm the password:
Creating wallet...
Created wallet 'test-wallet'
Your new wallet has a backup phrase that can be used for recovery.
Keeping this backup phrase safe is extremely important.
Would you like to see it now? (Y/n):
Your backup phrase is printed below.
Keep this information safe -- never share it with anyone!

One or more non-printable characters were ommited from the subsequent line:
[32mtest believe fancy alter salt bachelor section shell pear word feel wish follow crisp prepare tree maze moral tortoise east hollow wonder slam about push[0m
~$
```
### After
```
~$ goal wallet new test-wallet-2
Please choose a password for wallet 'test-wallet-2':
Please confirm the password:
Creating wallet...
Created wallet 'test-wallet-2'
Your new wallet has a backup phrase that can be used for recovery.
Keeping this backup phrase safe is extremely important.
Would you like to see it now? (Y/n):
Your backup phrase is printed below.
Keep this information safe -- never share it with anyone!

test pool repair shell kind aisle vendor elegant divorce moment install top talk clap lens foam green blossom bunker loop apology eager name abandon secret
~$
```

## Test Plan

Ensure existing integration tests work with updated expected outputs.


tsachiherman:
Printing control characters might be still a valid option. I don't think that we should disable this completely. On this particular case, I'm pretty neutral, as the control characters doesn't add much value. 
I would have changed the 
```golang
reportInfof(infoBackupPhrase, mnemonic)
```
to
```golang
fmt.Printf(infoBackupPhrase, mnemonic)
```

Note that I think that we should try and detect the console type. If it's a tty, then we should print a version without any control characters. For a console, we should try to emit control characters for better usability.

algorandskiy:
I agree with Tsachi, ANSI escape-compatible terminal users should be able to see colorful output when needed

cce:
So, something like https://pkg.go.dev/golang.org/x/term#IsTerminal ?

cce:
I grepped around looking for other colorized outputs that would be involved, but AFAICT this wallet backup phrase formatter is the only use of ANSI escape codes in goal, is that right?

tsachiherman:
> I grepped around looking for other colorized outputs that would be involved, but AFAICT this wallet backup phrase formatter is the only use of ANSI escape codes in goal, is that right?

for the `goal node status -d <data dir> -w 200` there is some usage of escape codes as well. It's used to implement a watch-like behavior.

algorandskiy (approved):
If there is no other colors, then good to go

tsachiherman:
> So, something like https://pkg.go.dev/golang.org/x/term#IsTerminal ?

exactly.

onetechnical:
I think in this case, the upside is super small, but the downstream effect is very annoying.

PR #2185, opened 2021-05-25
Improve fresh node startup time

tsachiherman:
## Summary

This PR contains three different changes that would allow the node to start up faster:
1. It disables the block 0 fix we previously deployed. This fix was intended to address catching up archival nodes, where we would calculate the block 0 hash incorrectly. The fix was long deployed, and the hash calculation is also fixed since.
2. The tracker and blocks databases are now being opened concurrently.
3. The accounts tracker database schema upgrade would now skip some of the steps that aren't required for a fresh database.

Combining all the above, the ledger startup of a new node is now about 2-3 times faster.

## Test Plan

This change has no functional enduser change. The existing tests provide sufficient coverage.


cmd/algod/main.go
```diff
@@ -90,52 +93,53 @@ func main() {
...
 
 	if absPathErr != nil {
 		fmt.Fprintf(os.Stderr, "Can't convert data directory's path to absolute, %v\n", dataDir)
-		os.Exit(1)
+		return 1
 	}
 
-	if *genesisFile == "" {
-		*genesisFile = filepath.Join(dataDir, config.GenesisJSONFile)
+	genesisFileName := *genesisFile
```
algorandskiy, line 115: mb `genesisPath` instead of `genesisFileName`

ledger/accountdb.go
```diff
@@ -521,27 +521,27 @@ func getCatchpoint(tx *sql.Tx, round basics.Round) (fileName string, catchpoint
...
 	var creatableMigrated bool
-	err := tx.QueryRow("SELECT 1 FROM pragma_table_info('assetcreators') WHERE name='ctype'").Scan(&creatableMigrated)
+	err = tx.QueryRow("SELECT 1 FROM pragma_table_info('assetcreators') WHERE name='ctype'").Scan(&creatableMigrated)
 	if err == sql.ErrNoRows {
 		// Run migration
 		for _, migrateCmd := range creatablesMigration {
 			_, err = tx.Exec(migrateCmd)
 			if err != nil {
-				return err
+				return false, err
```
algorandskiy, line 540: you can simply `return` here and below without arguments

ledger/acctupdates.go
```diff
@@ -1471,6 +1477,7 @@ func (au *accountUpdates) upgradeDatabaseSchema1(ctx context.Context, tx *sql.Tx
 		au.log.Infof("accountsInitialize found that no accounts needed to be reencoded")
 	}
 
+schemaUpdateComplete:
```
algorandskiy, line 1480: this looks like old good MSDN examples with goto to the end :)

algorandskiy (commented):
Nice cleanup, only a few minor notes

algorandskiy (approved):

PR #2187, opened 2021-05-25
tests: disable goal expect tests

tsachiherman:
## Summary

Temporarily disable expect tests, as the failures there are blocking out testing.

## Test Plan

n/a

### DIFF
--- a/test/e2e-go/cli/goal/expect/goal_expect_test.go
+++ b/test/e2e-go/cli/goal/expect/goal_expect_test.go
@@ -24,6 +24,7 @@ import (
 
 // TestGoalWithExpect Process all expect script files with suffix Test.exp within the test/e2e-go/cli/goal/expect directory
 func TestGoalWithExpect(t *testing.T) {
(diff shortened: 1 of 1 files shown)


algonautshant (approved):
LGTM

algorandskiy:
I suggest to disable _some_ expect tests that are failing: listExpiredParticipationKeyTest, catchpointCatchupTest, ...

algorandskiy:
I suggest to disable _some_ expect tests that are failing: listExpiredParticipationKeyTest, catchpointCatchupTest, ...

tsachiherman:
> I suggest to disable _some_ expect tests that are failing: listExpiredParticipationKeyTest, catchpointCatchupTest, ...

My concern is that these seems to be too many failures.. and I don't think that we did anything to cause this.
I think that when working on fixing these, we should do just what you describe:
- restore the expect tests that aren't failing frequently first
- investigate the rest of the expect test for the downstream cause of failure

But I really want to take this offline. We have too many in-queue PRs that are perfectly good PRs and should not be blocked.

PR #2190, opened 2021-05-25
Add support for RHEL Centos 8 RPM 

egieseke:
## Summary

Add packaging support for RHEL Centos 8 PRM.

## Test Plan

Testing with release pipeline.


tsachiherman:
Just wondering - why wouldn't the Centos 7 installation work on Centos 8 ?

onetechnical:
> Just wondering - why wouldn't the Centos 7 installation work on Centos 8 ?

Centos 8 has different dependencies (dnf-automatic instead of yum-cron).

onetechnical:
> > Just wondering - why wouldn't the Centos 7 installation work on Centos 8 ?
> 
> Centos 8 has different dependencies (dnf-automatic instead of yum-cron).

Oh wait, I think I see what you're saying. Can't we just have one RPM that works with both distributions?

onetechnical:
In answer to my own question, I think the answer is 'no', you can't have a single RPM that supports both distributions based on brief research.

docker/build/cicd.centos8.Dockerfile
```diff
@@ -0,0 +1,35 @@
+ARG ARCH="amd64"
+
+FROM ${ARCH}/centos:8
+ARG GOLANG_VERSION
+ARG ARCH="amd64"
+RUN yum install -y https://dl.fedoraproject.org/pub/epel/epel-release-latest-8.noarch.rpm && \
```
onetechnical, line 6: Why do we `yum install` this but `dnf install` later? Shouldn't it all be dnf?

scripts/release/common/docker/centos8.Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM centos:8
+
+WORKDIR /root
+RUN yum install -y https://dl.fedoraproject.org/pub/epel/epel-release-latest-8.noarch.rpm
```
onetechnical, line 4: Same here, should we dnf everywhere?

docker/build/cicd.centos8.Dockerfile
```diff
@@ -0,0 +1,35 @@
+ARG ARCH="amd64"
+
+FROM ${ARCH}/centos:8
```
onetechnical, line 3: I think this is fine, but I'm curious if there's a more supported stream image, e.g.:

quay.io/centos/centos:stream8

I don't actually know if that would be better or worse. ¯\_(ツ)_/¯

test/muleCI/Jenkinsfile
```diff
@@ -1,3 +1,3 @@
-@Library('go-algorand-ci') _
+@Library('go-algorand-ci@update-cicd-cachename') _
```
onetechnical, line 1: Don't forget to update this.

onetechnical (commented):
Most of this looks alright, had some comments.

Also, I'm curious how this will work with repositories. I don't know of a good way to test this.

docker/build/cicd.centos8.Dockerfile
bricerisingalgorand: oh right, i thought we were going to use that

docker/build/cicd.centos8.Dockerfile
bricerisingalgorand: it shouldn't actually make a difference, but it wouldn't hurt to switch to dnf

package.yaml
```diff
@@ -21,6 +21,17 @@ agents:
       - VERSION=$VERSION
     workDir: $HOME/projects/go-algorand
 
+  - name: rpm.centos8
+    dockerFilePath: docker/build/cicd.centos8.Dockerfile
+    image: algorand/go-algorand-ci-linux-centos8
+    version: scripts/configure_dev-deps.sh
```
bricerisingalgorand, line 27: does this file still exist? I thought tsachi was removing it

tsachiherman: I cleared it's content, but kept it around so it won't create build failures.

bricerisingalgorand:
Assuming it works I think this looks ok to me. It doesn't appear to include publishing to the rpm repo, but I think that's ok for this story

bricerisingalgorand (dismissed):

bricerisingalgorand:
I gave it a tentative approval as long as John's feedback is resolved

docker/build/cicd.centos8.Dockerfile
egieseke: Replaced use of yum with dnf.

onetechnical:
> Assuming it works I think this looks ok to me. It doesn't appear to include publishing to the rpm repo, but I think that's ok for this story

Oh, you know what, I believe we lock in on `s3://$PREFIX/algorand-$VERSION-1.x86_64.rpm` and `s3://$PREFIX/algorand-devtools-$VERSION-1.x86_64.rpm` in `scripts/release/mule/deploy/rpm/deploy.sh`, so we should be fine. I think we'll need to add a story later to add this to repo properly, once we thoroughly test.

scripts/release/common/docker/centos8.Dockerfile
egieseke: Updated to use dnf.

test/muleCI/Jenkinsfile
egieseke: Removed branch.

docker/build/cicd.centos8.Dockerfile
egieseke: Updated to pull from quay.io/centos/centos:stream8.

docker/build/cicd.centos8.Dockerfile
egieseke: updated to use dnf.

scripts/release/common/setup.sh
```diff
@@ -105,6 +105,7 @@ fi
 
 sudo usermod -a -G docker ubuntu
 sg docker "docker pull centos:7"
+sg docker "docker pull centos:8"
```
onetechnical, line 108: Might want to replace this with the stream image, but not really a blocker.

onetechnical (approved):
Provided this passes testing, I'm okay with this.

onetechnical:
This PR is still in draft mode, though.

scripts/release/common/setup.sh
egieseke: Updated to pull quay.io/centos/centos:stream8.

onetechnical (approved):
This still looks fine to me, though I'd like to see it exercised to verify it works.

algobarb:
Approved. Good work incorporating all the other suggestions!

algobarb (approved):

PR #2193, opened 2021-05-26
Add batch verification

algoidan:


tsachiherman:
assuming that this is a long standing branch, maybe mark it as draft and add "[wip]" to the title ?

algoidan:
This task won't be done as it is.
need to make changes in the batch verification algorithm first.

PR #2196, opened 2021-05-26
testing: fix bug in TestApplicationsUpgradeOverREST

tsachiherman:
## Summary

The TestApplicationsUpgradeOverREST e2e had a bug in the testing code -
if the upgrade takes place before the broadcast takes place, the broadcast result could be non-err.
This PR ensures that if we get a non-error, we have already upgraded.

## Test Plan

This is a test.


algonautshant (dismissed):
LGTM

algonautshant (approved):
LGTM

algorandskiy (approved):

PR #2197, opened 2021-05-26
sectok e2e faster

brianolson:
## Summary

run some independent commands in parallel, total test 230s -> 170s
`20210526_115456 :179 finished e2e_subs/sectok-app.sh OK in 235.471021 seconds`
down to
`20210526_122739 :179 finished e2e_subs/sectok-app.sh OK in 169.771514 seconds`

## Test Plan

This is a test, it still passes.


tsachiherman:
how does this solution handle error code generated by goal ? would it still fail the script as before ?

brianolson:
bash `wait` will now return the error code of the awaited background process.
```sh
#!/usr/bin/env bash

set -ex
set -o pipefail
export SHELLOPTS

false &
WA=$!
wait $WA
echo still going
`
above script does not reach 'still going'

PR #2198, opened 2021-05-26
testing: e2e sub assets-app speedup

brianolson:
## Summary

speedup
one 240s test -> two 120s tests that run in parallel

## Test Plan

This is a test, it should still pass.


tsachiherman:
I'm still trying to unwrap my head around how one test that was previously executed in 320s was equally broken to two now takes 240s each.

But given that 240s << 320s, I'd opt in for merge-in first and optimize later.

tsachiherman (approved):

PR #2212, opened 2021-05-27
Re-enable gofmt check during travis build

cce:
## Summary

While talking to @tsachiherman we noticed stray fmt'd files in my repo after running `make sanity`, and he suggested submitting a fix PR. I discovered this shell line generated a list of `gofiles` that is always empty, probably because the way we set up builds for branches (and vendoring) has changed since it was written.

## Test Plan

This re-enables gofmt checking in TravisCI builds.

### DIFF
--- a/catchup/fetcher_test.go
+++ b/catchup/fetcher_test.go
@@ -232,7 +232,7 @@ type testUnicastPeer struct {
 	version          string
 	responseChannels map[uint64]chan *network.Response
 	t                *testing.T
-	responseOverride *network.Response 
+	responseOverride *network.Response
 }
 
 func (p *testUnicastPeer) GetAddress() string {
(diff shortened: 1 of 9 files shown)


tsachiherman:
don't forget to update the makefile to use `gofmt -l` as well.

cce:
It's unnecessary, I just learned why, the `go fmt` command is set up to automatically do `-l -w` on named packages (or `./...`), whereas `gofmt` doesn't know about packages (just filenames or `.`). So they are complementary, sort of.

```
$ go help fmt
usage: go fmt [-n] [-x] [packages]

Fmt runs the command 'gofmt -l -w' on the packages named
by the import paths. It prints the names of the files that are modified.

For more about gofmt, see 'go doc cmd/gofmt'.
For more about specifying packages, see 'go help packages'.
```

tsachiherman (approved):
looks good.

PR #2213, opened 2021-05-27
[WIP] Use travis arm64 infrastructure.

winder:
## Summary

I noticed that travis has arm64 support, testing to see if it works.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,32 +25,38 @@ jobs:
   include:
     - stage: build_commit
       os: linux
+      arch: amd64
       script:
         - travis_retry scripts/travis/build_test.sh
 
     - stage: build_pr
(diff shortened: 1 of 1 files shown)


winder:
This sort of works, but the arm builds fail with:
```
/home/travis/.gimme/versions/go1.14.7.linux.arm64/pkg/tool/linux_arm64/link: signal: killed
```

I think the failure is while building binaries, but it's not totally clear from the Travis logs. Some brief research suggests that this could be due to [memory running out (during tests)](https://github.com/golang/go/issues/31517#issuecomment-484086859) or a [compiler bug (related to something like escape analysis)](https://github.com/golang/go/issues/31517#issuecomment-484115989).

Or maybe we have a bug.

PR #2215, opened 2021-05-27
Optimize constant assembly

jasonpaulos:
## Summary

With this PR, constants introduced by pseudo-ops (`int`, `byte`, `addr`) are now placed into constant blocks or loaded with `pushint`/`pushbytes` in the optimal way to save program space, starting with TEAL v4.

This optimization effectively reorders the `intcblock` and `bytecblock` that the assembler creates for pseudo-op constants such that the most frequently referenced constants are first (and can thereby take advantage of the space-saving `intc_X`/`bytec_X` opcodes). Additionally, any constants referenced only once are taken out of the constant block and instead loaded with `pushint` or `pushbytes` to further save space.

Other changes:
* Renamed `OpStream`'s `noIntcBlock` and `noBytecBlock` to `hasIntcBlock` and `hasBytecBlock` to clarify this flag is true only if the input code defines a constant block.
* `pushbytes` now disassembles with a comment containing a parsed representation of its contents, like `bytec`.

Closes algorand/go-algorand-internal#1191.

## Test Plan

Added the dedicated unit test `TestAssembleOptimizedConstants` in addition to updating existing tests. Note that the expected v4 compiled nonsense program now has a different prefix than the previous versions, since optimization changes the constant blocks at the start.


data/transactions/logic/assembler.go
```diff
@@ -1182,13 +1214,354 @@ const AssemblerDefaultVersion = 1
...
+		}
+
+		raw = replaceBytes(raw, ref.position, currentBytesLen, newBytes)
+		positionDelta := len(newBytes) - currentBytesLen
+
+		for i := range ops.intcRefs {
+			if ops.intcRefs[i].position > ref.position {
+				ops.intcRefs[i].position += positionDelta
+			}
+		}
```
jannotti, line 1515: I'm a little confused why you have to change the refs of ints in while optimizing bytes AND vice versa. I would think you'd only have to do the one that happens second.  Maybe you're doing it for symmetry and the refs won't actually be used again?  Could you comment a bit about this movement?

I am wondering if you'd be better off with one big rewrite, from pending into a new, longer byte buffer, rather than sliding things around for each constant.

data/transactions/logic/assembler_test.go
```diff
@@ -298,7 +298,7 @@ var compiled = map[uint64]string{
...
+	4: "042004010200b7a60c26040242420c68656c6c6f20776f726c6421208dae2087fbba51304eb02b91f656948397a7946390e8cb70fc9ea4d95f92251d047465737400320032013202320380021234292929292a0431003101310231043105310731083109310a310b310c310d310e310f3111311231133114311533000033000133000233000433000533000733000833000933000a33000b33000c33000d33000e33000f3300113300123300133300143300152d2e01022581f8acd19181cf959a1281f8acd19181cf951a81f8acd19181cf1581f8acd191810f082209240a220b230c240d250e230f23102311231223132314181b1c28171615400003290349483403350222231d4a484848482a50512a632223524100034200004322602261222b634848222862482864286548482228236628226724286828692422700048482471004848361c0037001a0031183119311b311d311e311f312024221e312131223123312431253126312731283129312a312b312c312d312e312f44782522531422542b2355220823564c4d4b0222382123391c0081e80780046a6f686e2281d00f24231f880003420001892223902291922394239593a0a1a2a3a4a5a6a7a8a9aaabacadae23af",
```
jannotti, line 301: Neat that it looks like this saved about 20 bytes in the nonsense program.

data/transactions/logic/assembler_test.go
```diff
@@ -530,13 +530,23 @@ func TestOpBytes(t *testing.T) {
...
 func TestAssembleInt(t *testing.T) {
 	t.Parallel()
+
+	expectedDefaultConsts := "012001bef5fad70c22"
+	expectedOptimizedConsts := "0181bef5fad70c"
+
 	for v := uint64(1); v <= AssemblerMaxVersion; v++ {
 		t.Run(fmt.Sprintf("v=%d", v), func(t *testing.T) {
+			expected := expectedDefaultConsts
+			if v >= backBranchEnabledVersion {
```
jannotti, line 540: If you do a different constant, don't miss this one.

data/transactions/logic/assembler_test.go
```diff
@@ -592,6 +612,178 @@ func TestAssembleBytesString(t *testing.T) {
...
+bytec_2 // "test"
+bytec_2 // "test"
+pushbytes 0xb49276bd3ec0977eab86a321c449ead802c96c0bd97c2956131511d2f11eebec // addr WSJHNPJ6YCLX5K4GUMQ4ISPK3ABMS3AL3F6CSVQTCUI5F4I65PWEMCWT3M
+bytec_1 // 0x0103
+bytec_1 // 0x0103
+bytec_1 // 0x0103
+`
+		for v := uint64(backBranchEnabledVersion); v <= AssemblerMaxVersion; v++ {
+			t.Run(fmt.Sprintf("v=%d", v), func(t *testing.T) {
+				expectedOps, err := AssembleStringWithVersion(expected, v)
```
jannotti, line 647: consider testProg here (anywhere you just need to assemble correctly).  It only saves one line (the error check), but when it fails, it reports the problem in more detail.

jannotti (approved):
Seems good. My main concern was going to be jump offsets could somehow be confused.  I think I saw that's handled right, and unit tests should give us some confidence too, but can you just go through in your own head that back jumps will not be messed up?  We probably don't have many unit tests that use them, and maybe none that have constants that are being optimized in a way that causes resize between a back jump and its target.

data/transactions/logic/assembler.go
```diff
@@ -1182,13 +1214,354 @@ const AssemblerDefaultVersion = 1
...
+
+		var currentBytesLen int
+		switch raw[ref.position] {
+		case 0x22:
+			fallthrough
+		case 0x23:
+			fallthrough
+		case 0x24:
+			fallthrough
+		case 0x25:
```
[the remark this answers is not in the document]
jasonpaulos: Was not aware of this feature, thanks!

data/transactions/logic/assembler.go
jasonpaulos: Technically this correction is only necessary in the `optimizeIntcBlock` function, since it runs first and must update the byte references. However it seemed weird to enforce and document that ints must be optimized before bytes, so I added the equivalent logic to `optimizeBytecBlock`, even though it's practically a no-op as is. I'll also add a comment for this.

jasonpaulos: > I am wondering if you'd be better off with one big rewrite, from pending into a new, shorter byte buffer, rather than sliding things around for each constant.

Perhaps that would be better for performance, though it wouldn't simplify any logic. If the performance impact of a change like this were to be significant, I can make a separate PR for it.

jasonpaulos:
@jannotti thanks for the review. I've used a separate constant to keep track of when optimization is enabled, added a test case for backwards jumps, and addressed other misc items.

data/transactions/logic/assembler.go
```diff
@@ -1182,13 +1223,359 @@ const AssemblerDefaultVersion = 1
...
+		switch raw[ref.position] {
+		case opIntc0, opIntc1, opIntc2, opIntc3:
+			currentBytesLen = 1
+		case opIntc:
+			currentBytesLen = 2
+		default:
+			return ops.errorf("Unexpected op at intReference: %d, from source line %d", raw[ref.position], ops.OffsetToLine[ref.position])
+		}
+
+		raw = replaceBytes(raw, ref.position, currentBytesLen, newBytes)
```
jannotti, line 1354: Since currentBytesLen and newBytes will so often be the same length, I wonder if your replaceBytes should relax its rule that it always produces a new slice, and instead return the existing slice with the bytes replaced?

In fact, when len(newBytes) < currentBytesLen, it can also return in place. It feels in keeping with "append" which has a similar contract.

Just bike shedding.  No need to do this.

jannotti (dismissed):

data/transactions/logic/assembler.go
jasonpaulos: Great suggestion, I'm happy to make this improvement in a follow-up PR

jasonpaulos:
Just rebased on master to fix conflict.

data/transactions/logic/assembler.go
```diff
@@ -1182,13 +1223,359 @@ const AssemblerDefaultVersion = 1
...
+				fixedOffsetsToLine[pos+positionDelta] = sourceLine
+			} else {
+				fixedOffsetsToLine[pos] = sourceLine
+			}
+		}
+		ops.OffsetToLine = fixedOffsetsToLine
+	}
+
+	ops.pending.Reset()
+	ops.pending.Write(raw)
```
cce, line 1387: tiny nit: I think it may be slightly more performant to do `ops.pending = bytes.NewBuffer(raw)` versus `Write()` which will trigger some slice-growing stuff based on `len(raw)`

cce: unless someone else is holding on to a reference to `pending` ... I don't know much about the code or how it's used

algorandskiy (commented):
Overall looks good. Please evaluate generalization of freq-based intc/bytec generation.

data/transactions/logic/assembler.go
```diff
@@ -1073,6 +1104,11 @@ func (ops *OpStream) assemble(fin io.Reader) error {
 		}
 	}
 
+	// if the version was not fixed above (because there were no opcodes), fix it now
+	if ops.Version == assemblerNoVersion {
+		ops.Version = AssemblerDefaultVersion
+	}
+
```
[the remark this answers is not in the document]
jasonpaulos: I only added this code because my optimization functions use `OpsByName[ops.Version]`, and some of the tests in `TestPragmas` broke this because they don't include any ops (meaning `ops.Version` remains -1). I don't expect this code to have any effect in actual programs.

data/transactions/logic/assembler.go
```diff
@@ -1182,13 +1223,359 @@ const AssemblerDefaultVersion = 1
...
+// optimizeIntcBlock rewrites the existing intcblock and the ops that reference
+// it to reduce code size. This is achieved by ordering the intcblock from most
+// frequently referenced constants to least frequently referenced, since the
+// first 4 constant can use the intc_X ops to save space. Additionally, any
+// ints with a reference of 1 are taken out of the intcblock and instead created
+// with the pushint op.
+//
+// This function only optimizes constants introduces by the int pseudo-op, not
+// preexisting intcblocks in the code.
+func (ops *OpStream) optimizeIntcBlock() error {
```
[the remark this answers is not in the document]
jasonpaulos: Thanks for the suggestion, I did this in the most recent commit

data/transactions/logic/assembler.go
jasonpaulos: I managed to implement this in the most recent commit

jannotti (approved):
I like the new replaceBytes.

data/transactions/logic/assembler.go
cce: can you tell I used to be a C++ programmer who was always looking for opportunities to call `.swap()` on data structures

data/transactions/logic/assembler.go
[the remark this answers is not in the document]
jasonpaulos: My latest commit removes this because it's no longer necessary.

PR #2217, opened 2021-05-28
txnsync: fix few non-fetching clients bugs

tsachiherman:
## Summary

Fix few bugs which prevented the non-fetching clients from working correctly, causing undesired load.

## Test Plan

Tested manually, and appear to be working correctly.

### DIFF
--- a/network/wsNetwork.go
+++ b/network/wsNetwork.go
@@ -422,6 +422,9 @@ func (wn *WebsocketNetwork) Address() (string, bool) {
 	parsedURL := url.URL{Scheme: wn.scheme}
 	var connected bool
 	if wn.listener == nil {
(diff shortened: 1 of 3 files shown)


network/wsNetwork.go
```diff
@@ -422,6 +422,9 @@ func (wn *WebsocketNetwork) Address() (string, bool) {
 	parsedURL := url.URL{Scheme: wn.scheme}
 	var connected bool
 	if wn.listener == nil {
+		if wn.config.NetAddress == "" {
+			parsedURL.Scheme = ""
+		}
```
tsachiherman, line 427: For NetAddress of "", we want the method to return "" rather than "http:".

brianolson (approved):

PR #2220, opened 2021-05-30
Fixing a bug in eval and TestOnSwitchToUnSupportedProtocol

algonautshant:
## Summary

TestOnSwitchToUnSupportedProtocol had multiple segments. This test is separated into multiple tests.

This PR fixes two issues:
1. Fix for a bug in eval: in the event of an error and early termination of the eval, the loading of accounts in a separate go routine was not terminated. loadAccounts will be terminated when eval returns. 
2. TestOnSwitchToUnSupportedProtocol1 had a bug in setting the blocks for protocol switch. The first block was not getting the NextProtocolSwitchOn set to 1. Despite the bug, the test was passing most of the time, and failing some of the time. 
3. TestOnSwitchToUnSupportedProtocol3 had a bug in setting the blocks for protocol switch. AddBlocks starts adding from the next round of the passed block. NextProtocolSwitchOn was getting set one round late.

## Test Plan
Fix 1 has existing tests covering it.
Fixes 2 and 3 are test fixes.
I ran the tests more than 100 times and all passed. 100 runs was enough to expose the problem.

### DIFF
--- a/catchup/service_test.go
+++ b/catchup/service_test.go
@@ -422,7 +422,6 @@ func TestServiceFetchBlocksMalformed(t *testing.T) {
 }
 
 func TestOnSwitchToUnSupportedProtocol(t *testing.T) {
-	t.Skip("This test is flacky and need to be fixed.")
 
 	// Test the interruption in the initial loop
 	// This cannot happen in practice, but is used to test the code.
@@ -535,7 +534,8 @@ func helperTestOnSwitchToUnSupportedProtocol(
 	config := defaultConfig
 	config.CatchupParallelBlocks = 2
 
-	remote, _, blk, err := buildTestLedger(t, bookkeeping.Block{}) //mRemote.blocks[0])
+	block1 := mRemote.blocks[1]
+	remote, _, blk, err := buildTestLedger(t, block1)
 	if err != nil {
 		t.Fatal(err)
 		return local, remote
--- a/catchup/service_test.go
+++ b/catchup/service_test.go
@@ -421,97 +421,96 @@ func TestServiceFetchBlocksMalformed(t *testing.T) {
 	//require.True(t, s.fetcherFactory.(*MockedFetcherFactory).fetcher.client.closed)
 }
 
-func TestOnSwitchToUnSupportedProtocol(t *testing.T) {
+// Test the interruption in the initial loop
+// This cannot happen in practice, but is used to test the code.
+func TestOnSwitchToUnSupportedProtocol1(t *testing.T) {
 
-	// Test the interruption in the initial loop
-	// This cannot happen in practice, but is used to test the code.
-	{
-		lastRoundRemote := 5
-		lastRoundLocal := 0
-		roundWithSwitchOn := 0
-		local, remote := helperTestOnSwitchToUnSupportedProtocol(t, lastRoundRemote, lastRoundLocal, roundWithSwitchOn, 0)
+	lastRoundRemote := 5
+	lastRoundLocal := 0
+	roundWithSwitchOn := 0
+	local, remote := helperTestOnSwitchToUnSupportedProtocol(t, lastRoundRemote, lastRoundLocal, roundWithSwitchOn, 0)
 
-		// Last supported round is 0, but is guaranteed
-		// to stop after 2 rounds.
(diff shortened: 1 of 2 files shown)


catchup/service_test.go
```diff
@@ -421,98 +421,96 @@ func TestServiceFetchBlocksMalformed(t *testing.T) {
 	//require.True(t, s.fetcherFactory.(*MockedFetcherFactory).fetcher.client.closed)
 }
 
-func TestOnSwitchToUnSupportedProtocol(t *testing.T) {
```
algonautshant, line 424: No changes in TestOnSwitchToUnSupportedProtocol except for dividing it into TestOnSwitchToUnSupportedProtocol1, TestOnSwitchToUnSupportedProtocol2, TestOnSwitchToUnSupportedProtocol3 and TestOnSwitchToUnSupportedProtocol4.

catchup/service_test.go
```diff
@@ -535,14 +533,16 @@ func helperTestOnSwitchToUnSupportedProtocol(
 	config := defaultConfig
 	config.CatchupParallelBlocks = 2
 
-	remote, _, blk, err := buildTestLedger(t, bookkeeping.Block{}) //mRemote.blocks[0])
+	block1 := mRemote.blocks[1]
```
brianolson, line 536: I like this the previous way because in this way it could erroneously use block1 or mRemote.blocks[1] after buildTestLedger, but it's pass by value not pass by reference. I want to avoid trying to use the unmodified blank block and thinking we're getting something else.

ledger/eval.go
```diff
@@ -1174,7 +1174,15 @@ func eval(ctx context.Context, l ledgerForEvaluator, blk bookkeeping.Block, vali
 		return ledgercore.StateDelta{}, err
 	}
 
-	paysetgroupsCh := loadAccounts(ctx, l, blk.Round()-1, paysetgroups, blk.BlockHeader.FeeSink, blk.ConsensusProtocol())
+	accountLoadingCtx, accountLoadingCancel := context.WithCancel(ctx)
+	paysetgroupsCh := loadAccounts(accountLoadingCtx, l, blk.Round()-1, paysetgroups, blk.BlockHeader.FeeSink, blk.ConsensusProtocol())
+	// ensure that before we exit from this method, the account loading is no longer active.
+	defer func() {
+		accountLoadingCancel()
+		// wait for the paysetgroupsCh to get closed.
```
winder, line 1182:
```suggestion
		// wait for the paysetgroupsCh to get closed and/or remove the remaining accounts from the channel.
```

winder (approved):
Seems fine to me once addressing Brian's comment.

ledger/eval.go
algonautshant: Will this stop waiting if the channel is emptied? I think it has to get closed.

catchup/service_test.go
algonautshant: What do you mean by?
> unmodified blank block


mRemote.blocks[1] will have the NextProtocolSwitchOn set according to the parameters passed to NextProtocolSwitchOn. That is why it is important to pick mRemote.blocks[1] instead of bookkeeping.Block{}.

ledger/eval.go
tsachiherman: The channel being closed is an indication that the go-routine is done.. removing the accounts from the channel is not really interesting here, but rather a way to detect that the channel got closed ( which ensures that no database actions would be taken place ).

winder: You're right, "and/or" was wrong. I thought it was interesting to mention that if there were unprocessed accounts (i.e. if there was an error) they would be released.

catchup/service_test.go
brianolson: Your other comment said no change in behavior, just reorg, but this line is a change in behavior of what `buildTestLedger()` gets from `Block{}` to whatever `testingenvWithUpgrade()` may have put into `mRemote.blocks[1]`. Just checking that's intentional and there's no pass-by-value pass-by-reference mixup here.

algonautshant: oh I see. sorry for the confusion I caused with that comment. 
That comment applies exclusively to the TestOnSwitchToUnSupportedProtocol function.

PR #2226, opened 2021-06-02
txnsync: recalibrate the MsgDigestSkip threshold

tsachiherman:
## Summary

Increase the threshold of the filtered message digest size, so that we still keep most of the proposal messages, but eliminate the transaction sync messages.

## Test Plan

Tested by deploying a S1 network and analyzing its MS message counts and overall bandwidth consumption.

### DIFF
--- a/network/wsNetwork.go
+++ b/network/wsNetwork.go
@@ -54,7 +54,11 @@ import (
 )
 
 const incomingThreads = 20
-const messageFilterSize = 5000 // messages greater than that size may be blocked by incoming/outgoing filter
+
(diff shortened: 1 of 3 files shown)


nicholasguoalgorand:
Could it be that it would be better to use the message type instead of using a size threshold?

tsachiherman:
> Could it be that it would be better to use the message type instead of using a size threshold?

I was contemplating about that one as well. I'm trying to get the network library to be as tag-agnostic as we can, however, it might be the case where we hit a wall.. I don't mind if you'll convert this to a PP in a subsequent PR, but please ensure that it would play nicely with previous network protocol..

PR #2228, opened 2021-06-02
codecov integration

winder:
## Summary
Integrate with codecov.io to attach coverage reports to PRs


.travis.yml
```diff
@@ -207,6 +207,9 @@ before_cache:
    esac
    docker save -o $HOME/docker_cache/images.tar $(docker images -a -q)
 
+after_success:
+  - bash <(curl -s https://codecov.io/bash)
```
tsachiherman, line 211: codecov were hacked in the past.. and the exploit was exactly this line of code - someone was able to inject a malware at this URL, forcing all the clients to execute it. luckily, the solution is quite straight-forward:
download the bash and add it to our repository. Then, invoke that bash script whenever needed.
Alternatively, you can still download the bash script, but perform a MD5 check on it to ensure it wasn't tampered.

cce: Agreed, definitely want to download and commit the bash script to the repo. I did this for years and it prevented my projects from being hacked.

winder: @cce what do you think about the checksum approach? Did you ever get into a situation where upstream changes to the bash script caused the CI system to fail?

cce: I didn't ... as I recall I checked in codecov.sh on a project in 2016 or 2017 and never had an issue with the integration, but either way would work. I noticed they are publishing SHA hashes in a Github repo since their security incident this year, and have some script examples in https://github.com/codecov/codecov-bash#%EF%B8%8F-verifying-the-bash-uploader

winder: There is also a python uploader, so `pip install codecov` is also an option.

winder: It also occurs to me that I should use `env -i` and pass the token in directly to avoid giving this script access to the environment to begin with.

.travis.yml
```diff
@@ -207,6 +207,9 @@ before_cache:
    esac
    docker save -o $HOME/docker_cache/images.tar $(docker images -a -q)
 
+after_success:
+  - scripts/travis/upload_coverage.sh
```
tsachiherman, line 211: I think it would be better to do
```bash
scripts/travis/upload_coverage.sh||true
```
so that the build won't fail in case we have a code coverage report uploading error.
( i.e. to make sure it doesn't break the build )

Makefile
```diff
@@ -228,17 +228,13 @@ $(GOPATH1)/bin/%:
 	cp -f $< $@
 
 test: build
-	go test $(GOTAGS) -race $(UNIT_TEST_SOURCES) -timeout 3600s | logfilter
+	go test $(GOTAGS) -race $(UNIT_TEST_SOURCES) -timeout 1h -coverprofile=coverage.txt -covermode=atomic
```
tsachiherman, line 231: could you store the output `coverage.txt` in a place *other* then the repo's root ? maybe ~/coverage.txt ?
( otherwise, we would also need to add this to the `.gitignore` )

scripts/travis/upload_coverage.sh
```diff
@@ -0,0 +1,22 @@
...
+
+# fail if shasum detects a problem
+set -e
+
+FILE_ARG=""
+if [[ -f "$1" ]]; then
+  FILE_ARG="-f $1"
+fi
+
+curl -fLso codecov https://codecov.io/bash;
```
tsachiherman, line 13: The `;` at the eol is abit confusing for me, but it's not incorrect.

scripts/travis/upload_coverage.sh
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env bash
+
+# Download codecov bash script, verify checksum from github, run script
+
+# fail if shasum detects a problem
+set -e
```
tsachiherman, line 6: you're using pipes here, so I'd suggest `set -o pipefail` to ensure we don't mask out errors.

scripts/travis/upload_coverage.sh
```diff
@@ -0,0 +1,22 @@
...
+FILE_ARG=""
+if [[ -f "$1" ]]; then
+  FILE_ARG="-f $1"
+fi
+
+curl -fLso codecov https://codecov.io/bash;
+VERSION=$(grep -o 'VERSION=\"[0-9\.]*\"' codecov | cut -d'"' -f2);
+for i in 1 256 512
+do
+  shasum -a $i -c --ignore-missing <(curl -s "https://raw.githubusercontent.com/codecov/codecov-bash/${VERSION}/SHA${i}SUM")
```
tsachiherman, line 17: this is quite risky, since the downloaded file name is determine by the other endpoint. break this into two calls:
```
curl -s "https://raw.githubusercontent.com/codecov/codecov-bash/${VERSION}/SHA${i}SUM" -o codecov.sha${I}
shasum -a $i -c codecov.sha${I}
```

Makefile
winder: I think gitignore is better than putting stuff in people's home directory. I also added some of the other files missing from the cover/Prof targets

winder:
After looking at the codecov bash uploader repo, I think it's best that we pin the version. Their workflow did not instill me with confidence.

.travis.yml
winder: I came around. The script is now copied to our repo.

scripts/travis/upload_coverage.sh
```diff
@@ -0,0 +1,15 @@
...
+
+# Check if there is a new version.
+curl -fLso codecov https://codecov.io/bash
+UPSTREAM_VERSION=$(grep -o 'VERSION=\"[0-9\.]*\"' codecov | cut -d'"' -f2)
+LOCAL_VERSION=$(grep -o 'VERSION=\"[0-9\.]*\"' scripts/travis/codecov | cut -d'"' -f2)
+if [[ "${UPSTREAM_VERSION}" != "${LOCAL_VERSION}" ]]; then
+  echo "WARN: version ${UPSTREAM_VERSION} of the codecov upload script is available."
+fi
+
+/usr/bin/env bash scripts/travis/codecov
```
tsachiherman, line 15: maybe use the `-f` to tell codecov exactly which file to upload ?

winder: The default patterns suit our needs for now. Maybe revisit this when we include integration test results?

tsachiherman (approved):
looks good.

PR #2229, opened 2021-06-02
Use gzip to compress txnsync msgs

nicholasguoalgorand:
## Summary

Add gzip compression to txnsync msgs when size of msgs > 1000B and network speed is slow enough that compression would save time.

## Test Plan

Wrote unit tests and benchmarks


txnsync/exchange.go
```diff
@@ -56,6 +56,7 @@ type packedTransactionGroups struct {
 	_struct struct{} `codec:",omitempty,omitemptyarray"`
 
 	Bytes []byte `codec:"g,allocbound=maxEncodedTransactionGroupBytes"`
+	Compressed bool `codec:"c"`
```
tsachiherman, line 59: Maybe replace the "Compressed" with "CompressionFormat" of type uint8 ?
This would give us higher control over the format in the future.
( leave the value 0 for non-compress, and value of 1 for gzip )

txnsync/txngroups_test.go
```diff
@@ -363,34 +363,73 @@ func TestTxnGroupEncodingLarge(t *testing.T) {
...
+	}
+
+	fmt.Println("new data: ", size)
+}
+
+func BenchmarkTxnGroupCompression(b *testing.B) {
+	txnGroups, _, _, err := txnGroupsData()
+	require.NoError(b, err)
+	var size int
+	encodedGroupsBytes, err := encodeTransactionGroups(txnGroups, false)
```
tsachiherman, line 384: The `releaseEncodedTransactionGroups` should be called once per `encodeTransactionGroups`.

txnsync/txngroups.go
```diff
@@ -25,7 +27,10 @@ import (
 	"github.com/algorand/go-algorand/protocol"
 )
 
-func encodeTransactionGroups(inTxnGroups []transactions.SignedTxGroup) ([]byte, error) {
+const compressionSpeed = 23071093.0 // bytes per second
```
tsachiherman, line 30: `compressionSpeed ` -> `estimatedGzipCompressionSpeed`
`compressionSavings` -> `estimatedGzipCompressionGains`
in addition to the above, also add proper description for each of them, as well as the name of the benchmark that is used to calculate them

txnsync/txngroups.go
```diff
@@ -60,23 +65,52 @@ func encodeTransactionGroups(inTxnGroups []transactions.SignedTxGroup) ([]byte,
...
 		}
 	}
 	stub.finishDeconstructSignedTransactions()
 
-	return stub.MarshalMsg(protocol.GetEncodingBuf()[:0]), nil
+	encoded := stub.MarshalMsg(protocol.GetEncodingBuf()[:0])
+
+	if len(encoded) > 1000 && (1.0 / compressionSpeed) < (compressionSavings / float32(dataExchangeRate)) {
+		compressedBytes, err := compressTransactionGroupsBytes(encoded)
+		return compressedBytes, true, err
```
tsachiherman, line 80: in case you've failed to compress the encoded bytes, just fall back to return the encoded bytes without any compression. ( at this point, you might want to log a warning message as well )

txnsync/txngroups.go
```diff
@@ -60,23 +65,52 @@ func encodeTransactionGroups(inTxnGroups []transactions.SignedTxGroup) ([]byte,
...
 				index++
 			}
 		}
 	}
 	stub.finishDeconstructSignedTransactions()
 
-	return stub.MarshalMsg(protocol.GetEncodingBuf()[:0]), nil
+	encoded := stub.MarshalMsg(protocol.GetEncodingBuf()[:0])
+
+	if len(encoded) > 1000 && (1.0 / compressionSpeed) < (compressionSavings / float32(dataExchangeRate)) {
```
tsachiherman, line 78: 1. move the magic number 1000 into it's own constant - `minEncodedTransactionGroupsCompressionThreshold`
2. Refactor the expression `(1.0 / compressionSpeed) < (compressionSavings / float32(dataExchangeRate))` so it would be more understandable.  I'm pretty sure that the expression is wrong because it's not taking into account the `len(encoded)`. More specifically :
projectedUncompressedSendingTime = len(encoded) / dataExchangeRate
projectedCompressedSendingTime = len(encoded) / compressionSpeed + len(encoded)* (1-compressionSavings) / dataExchangeRate

and the if statement should determine if projectedUncompressedSendingTime is more than projectedCompressedSendingTime

txnsync/txngroups.go
```diff
@@ -60,23 +65,52 @@ func encodeTransactionGroups(inTxnGroups []transactions.SignedTxGroup) ([]byte,
...
+	return encoded, false, nil
+}
+
+func compressTransactionGroupsBytes(data []byte) ([]byte, error) {
+	b := make([]byte, 0, len(data))
+	buf := bytes.NewBuffer(b)
+	zw := gzip.NewWriter(buf)
+
+	if _, err := zw.Write(data); err != nil {
+		return nil, fmt.Errorf("error gzip compressing data: %w", err)
```
tsachiherman, line 92: You need to close the writer even in case of an error.

txnsync/txngroups.go
```diff
@@ -111,6 +145,19 @@ func decodeTransactionGroups(bytes []byte, genesisID string, genesisHash crypto.
 	return txnGroups, nil
 }
 
+func decompressTransactionGroupsBytes(data []byte) (decoded []byte, err error){
+	zr, err := gzip.NewReader(bytes.NewBuffer(data))
```
tsachiherman, line 149: I *think* that the reader need to be closed as well, but I'm not sure about it.

txnsync/txngroups_test.go
```diff
@@ -363,34 +363,73 @@ func TestTxnGroupEncodingLarge(t *testing.T) {
...
+	b.ReportAllocs()
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
+		encodedGroupsBytes, err := encodeTransactionGroups(txnGroups, false)
+		require.NoError(b, err)
+		size = len(encodedGroupsBytes)
+		releaseEncodedTransactionGroups(encodedGroupsBytes)
+	}
+
+	fmt.Println("new data: ", size)
```
tsachiherman, line 377: could you add a metric with that information instead of printf ?

tsachiherman (changes requested):
looks pretty good. I've added few requests above.

txnsync/txngroups.go
nicholasguoalgorand: len(encoded) cancels out on both sides. However, I can still format it to multiply len(encoded) on both sides of the expressions to make the code understandable

tsachiherman: yes, that's true. I think that few lines that explain how you got to this formula would be a good idea.
also, it can be simplified to `compressionSavings * compressionSpeed > dataExchangeRate`, right ?

txnsync/txngroups.go
```diff
@@ -60,23 +69,61 @@ func encodeTransactionGroups(inTxnGroups []transactions.SignedTxGroup) ([]byte,
...
+	encoded := stub.MarshalMsg(protocol.GetEncodingBuf()[:0])
+
+	// check if time saved by compression: estimatedGzipCompressionGains * len(msg) / dataExchangeRate
+	// is greater than by time spent during compression: len(msg) / estimatedGzipCompressionSpeed
+	if len(encoded) > minEncodedTransactionGroupsCompressionThreshold && float32(dataExchangeRate) < (estimatedGzipCompressionGains * estimatedGzipCompressionSpeed) {
+		compressedBytes, err := compressTransactionGroupsBytes(encoded)
+		if err == nil {
+			return compressedBytes, 1, nil
+		}
+		logging.Base().Warnf("failed to compress txnsync msg: %v", err)
```
tsachiherman, line 89: make the `encodeTransactionGroups` method a `(s *syncState)` method, and use the `s.log` to log warning messages. Also, when reporting the error, make sure to report the size of the input buffer.

txnsync/txngroups.go
```diff
@@ -60,23 +69,61 @@ func encodeTransactionGroups(inTxnGroups []transactions.SignedTxGroup) ([]byte,
...
+
+	if compressionFormat == 1 {
+		data, err = decompressTransactionGroupsBytes(data)
+		if err != nil {
+			return
+		}
+	}
+
+	if compressionFormat > 1 {
+		return nil, fmt.Errorf("invalid compressionFormat, %v", compressionFormat)
```
tsachiherman, line 122: `%v` -> `%d` since `compressionFormat` is an integer.

txnsync/txngroups.go
```diff
@@ -42,7 +51,7 @@ func encodeTransactionGroups(inTxnGroups []transactions.SignedTxGroup) ([]byte,
 		if len(txGroup.Transactions) > 1 {
 			for _, txn := range txGroup.Transactions {
 				if err := stub.deconstructSignedTransactions(index, &txn); err != nil {
-					return nil, fmt.Errorf("failed to encodeTransactionGroups: %w", err)
+					return nil, 0, fmt.Errorf("failed to encodeTransactionGroups: %w", err)
```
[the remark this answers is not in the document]
nicholasguoalgorand: why should it be `msgp:ignore`?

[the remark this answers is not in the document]
tsachiherman: assuming you'll keep it as uint8 on the data structure, there is no need to create a corresponding msgp marshaler for that. If you're going to use it on the original data structure, then just ignore what I've wrote here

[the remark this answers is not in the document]
tsachiherman: I think that my point is - make sure you don't return 0/1/ or test 0/1. instead, you want to compare it with some enums values.

txnsync/txngroups_test.go
```diff
@@ -363,30 +366,72 @@ func TestTxnGroupEncodingLarge(t *testing.T) {
...
+
+	b.ReportAllocs()
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
+		compressedGroupBytes, err := compressTransactionGroupsBytes(encodedGroupsBytes)
+		require.NoError(b, err)
+		size = len(compressedGroupBytes)
+	}
+
+	b.ReportMetric(float64(size), "encodedDataBytes")
```
tsachiherman, line 401: This benchmark would need to "generate" two metrics:
1. The estimatedGzipCompressionSpeed, by `len(encodedGroupsBytes)/loop time`
2. The estimatedGzipCompressionGains, by `( len(encodedGroupsBytes)-size) / len(encodedGroupsBytes)`

These would allow us to calibrate the above parameters when we flip the compression library.

tsachiherman:
@nicholasguoalgorand , the changes looks good.
Beside the small comment above, I'll need you to rebase this one after I'll finish merging the master onto the branch. ( wasn't trivial merge.. )

txnsync/txngroups_test.go
nicholasguoalgorand: Is there a convenient way to access loop time?

tsachiherman: Not when using the go benchmarks. You can make this method into a *TestXXX* and then use the `testing.Benchmark` method to get the `BenchmarkResult` and print it out. But that's already more than really needed. Just wrap the for loop in a `loopStartTime := time.Now()... loopDuration := time.Now().Sub(loopStartTime)` and you'll have what you need.

tsachiherman:
btw - note the formatting issues on travis.

PR #2230, opened 2021-06-02
txnsync: add deflate library support

tsachiherman:
## Summary

Add deflate library; The deflate.go export the Gzip compression/decompression routines.

The core library is https://github.com/ebiggers/libdeflate, and a deflate.go wrapper was added that just makes the calls into that library.

### Performance testing

The two benchmarks BenchmarkCompression and BenchmarkDecompression provides some visibility to the performance characteristics:

BenchmarkCompression/compress/gzip-8         	     375	   3178807 ns/op    <-- using the standard go library compress/gzip
BenchmarkCompression/deflateCompression-8    	     466	   2536654 ns/op <-- using this library 

BenchmarkDecompression/compress/gzip-8       	    4995	    234960 ns/op <-- using the standard go library compress/gzip
BenchmarkDecompression/deflateCompression-8  	   10000	    104704 ns/op <-- using this library 

### Summary

Compared to the standard go library, this library have similar compression rate and uses ~20% less cpu when compressing and 66% less cpu when decompressing Gzip encoded blocks.

Note that in all fairness, the standard go library is a stream based library, whereas this one is a block-based library. So, it might not be a truly fair comparison. However, our planned usage of this library is as a block encoder, so the stream-base interface is not of a value for us.

## Test Plan

Unit test & Benchmarks included.


tsachiherman:
@nicholasguoalgorand - fyi

brianolson:
speed test vs https://golang.org/pkg/compress/flate/  ?

tsachiherman:
> speed test vs https://golang.org/pkg/compress/flate/ ?

I'll update the header.

tsachiherman:
> speed test vs https://golang.org/pkg/compress/flate/ ?

just to be clear -
The flate/gzip/zlib variation are basically all the same implementation with some different headers/packaging. That's true for all these implementation. That's also why the https://golang.org/pkg/compress/gzip is using https://golang.org/pkg/compress/flate as part of it's implementation.

util/compress/deflate_test.go
```diff
@@ -0,0 +1,119 @@
...
+		for k := 0; k < b.N; k++ {
+			outBuffer := bytes.NewBuffer(outBuffer)
+			writer := gzip.NewWriter(outBuffer)
+			writer.Write(buffer)
+			writer.Close()
+			targetLength = outBuffer.Len()
+		}
+	})
+	// figure out desired compression level.
+	compressionLevel := 1
```
nicholasguoalgorand, line 64: how will finding out the best compression level work in practice?

util/compress/deflate_test.go
```diff
@@ -0,0 +1,119 @@
...
+	require.Equal(t, compressedBuffer[:len], compressedOutput)
+
+	decompressedBuffer := make([]byte, 0, bufLen)
+	decompressedOutput, err := Decompress(compressedOutput, decompressedBuffer)
+	require.NoError(t, err)
+	require.Equal(t, decompressedOutput, buffer)
+}
+
+func BenchmarkCompression(b *testing.B) {
+	bufLen := 1024000
```
nicholasguoalgorand, line 47: do we know how the compression algorithms compare with bufLen around 10000 (around the size of 100 txns)

util/compress/deflate_test.go
tsachiherman: In practice it's always picking level=1. However, we need to repeat that test with real data in the txnsync test so we can verify it works as expected.

util/compress/deflate_test.go
tsachiherman: compression algorithms are heuristic based, so in order to get a better sense, we have to test it with some "real" data.

PR #2231, opened 2021-06-03
Extend catchpoint wait for round timeout from 10 to 60 seconds.

egieseke:
## Summary

Extend catchpoint test next round timeout from 10 to 60 seconds to avoid test failure.

## Test Plan

Tested locally.

### DIFF
--- a/test/e2e-go/features/catchup/catchpointCatchup_test.go
+++ b/test/e2e-go/features/catchup/catchpointCatchup_test.go
@@ -220,7 +220,7 @@ func TestBasicCatchpointCatchup(t *testing.T) {
 	targetRound = uint64(37)
 	log.Infof("Second node catching up to round 36")
 	for {
-		err = fixture.ClientWaitForRound(secondNodeRestClient, currentRound, 10000*time.Millisecond)
+		err = fixture.ClientWaitForRound(secondNodeRestClient, currentRound, 60*time.Second)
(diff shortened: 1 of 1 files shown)


bricerisingalgorand (approved):
This looks reasonable to me

tsachiherman:
while it's reasonable change, I would be happy to hear why this issue start failing - 
is advancing a single round takes 60 seconds ? if so, maybe we need to stop everything and fix that bug ?

egieseke:
> while it's reasonable change, I would be happy to hear why this issue start failing -
> is advancing a single round takes 60 seconds ? if so, maybe we need to stop everything and fix that bug ?

It now takes more than 10 seconds and it would be good to know why.  I will create a ticket a separate ticket to investigate.

algobarb:
Since this is to fix the nightly builds, I think it seems ok to go ahead with this as we troubleshoot why it's taking more than 10 seconds and whether 60 seconds is good.

algobarb (approved):

onetechnical:
I'm okay with trying this to see if it unblocks the nightly builds, but I want to know _how_ long it's taking and why, as to Tsachi's point, if it really is taking > 10s to advance a round, that would be very bad. (I don't think this is actually the case, there may be something else in the testing apparatus causing a problem.)

PR #2232, opened 2021-06-03
Use minimal travis container since we install go ourselves.

winder:
## Summary
Since we manage go ourselves I want to see if our build works without running the travis go setup step.

### DIFF
--- a/scripts/travis/codegen_verification.sh
+++ b/scripts/travis/codegen_verification.sh
@@ -42,7 +42,7 @@ function runGoFmt() {
 }
 
 function runGoLint() {
-    warningCount=$("$GOPATH"/bin/golint $(GO111MODULE=off go list ./... | grep -v /vendor/ | grep -v /test/e2e-go/) | wc -l | tr -d ' ')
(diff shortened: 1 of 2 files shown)


tsachiherman:
I think that there are some differences in the checkout location on travis go builds. But I would be happy to align with with a minimal travis build if that works.

scripts/travis/codegen_verification.sh
```diff
@@ -42,7 +42,7 @@ function runGoFmt() {
 }
 
 function runGoLint() {
-    warningCount=$("$GOPATH"/bin/golint $(GO111MODULE=off go list ./... | grep -v /vendor/ | grep -v /test/e2e-go/) | wc -l | tr -d ' ')
+    warningCount=$("$GOPATH"/bin/golint $(GO111MODULE=on go list ./... | grep -v /vendor/ | grep -v /test/e2e-go/) | wc -l | tr -d ' ')
```
tsachiherman, line 45: At this point, can't we just omit the `GO111MODULE=on` ?

winder: I'll give it a try, it should work since we export it that way.

tsachiherman (approved):
looks good.

Issue #2221, opened 2021-06-01
opcode `opGetByte` and `opSetByte`


amityadav0:
### Subject of the issue
Index check for the set byte should have `>=`.

In `opGetByte`
https://github.com/algorand/go-algorand/blob/23bfcd799470618e59abbc6b0247357225ffb20f/data/transactions/logic/eval.go#L2454
we are checking for equal condition because the indexing assumed here starts from `0`.

but in case of `opSetByte`
https://github.com/algorand/go-algorand/blob/23bfcd799470618e59abbc6b0247357225ffb20f/data/transactions/logic/eval.go#L2471
This one should be same as `opGetByte`

Please correct me if i am not understanding it right.

jannotti:
Thank you.  Addressing with #2236

PR #2236, opened 2021-06-03
Check setbyte length properly.

jannotti:
We were checking the against the length with > instead of >=.  That's
wrong, so we could cause a panic instead of clean error when trying to
set the byte 1 past the array length.  Fortunately, both have the same
effect: the txn fails. But we should do this properly.  Thanks to
@amityadav0 for the report in #2221.

New unit test to detect regression:
```
	testPanics(t, `global ZeroAddress; dup; concat; int 64; int 7; setbyte; int 1; return`, 3)
	testAccepts(t, `global ZeroAddress; dup; concat; int 63; int 7; setbyte; int 1; return`, 3)
```

### DIFF
--- a/data/transactions/logic/eval.go
+++ b/data/transactions/logic/eval.go
@@ -2452,7 +2452,7 @@ func opGetByte(cx *evalContext) {
 	target := cx.stack[prev]
 
 	if idx >= uint64(len(target.Bytes)) {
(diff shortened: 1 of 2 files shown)


brianolson (approved):

tsachiherman (approved):
looks good.

PR #2237, opened 2021-06-04
Modify build_release to match build_pr and remove ARM Deploy.

onetechnical:
## Summary

This pulls in the cleanups we've made to other testing into the rel/ tests. Additionally, the ARM Deploy builder has not successfully built in a while, so we'll remove it altogether.

## Test Plan

Merge to master and re-try the nightly trigger.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -94,6 +94,11 @@ jobs:
 
     - stage: build_release
       os: linux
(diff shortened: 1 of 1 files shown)


tsachiherman (approved):
Sure. As long as you have other means of compiling the arm variation, I see no issue with this change.

PR #2238, opened 2021-06-04
Fix and enable TestNewAccountCanGoOnlineAndParticipate

algonautshant:
## Summary

TestNewAccountCanGoOnlineAndParticipate was failing because the test was not waiting enough to get to the round where the newly funded account's funds will be considered for proposing purposes.

- Fixed the calculation of the round when the refunded account can propose. 
- Fixed the way the rounds were considered for deciding when the account is funded, which was prone to race conditions.
- The test was using WaitForRoundWithTimeout which may be very slow if the current round is already ahead. Instead, replaced it with ClientWaitForRound, which does not care if an individual round is delayed.
- Changed the configuration parameters so that the refunded account proposes sooner than the default config values. This is to avoid waiting to round 320. 

## Test Plan
This is a test fix.

### DIFF
--- a/test/e2e-go/features/participation/onlineOfflineParticipation_test.go
+++ b/test/e2e-go/features/participation/onlineOfflineParticipation_test.go
@@ -18,13 +18,18 @@ package participation
 
 import (
 	"path/filepath"
+	"runtime"
 	"testing"
+	"time"
 
 	"github.com/stretchr/testify/require"
 
 	"github.com/algorand/go-algorand/config"
+	"github.com/algorand/go-algorand/protocol"
 	"github.com/algorand/go-algorand/data/basics"
+	"github.com/algorand/go-algorand/test/e2e-go/globals"
 	"github.com/algorand/go-algorand/test/framework/fixtures"
+
 )
 
 func TestParticipationKeyOnlyAccountParticipatesCorrectly(t *testing.T) {
@@ -96,18 +101,28 @@ func waitForAccountToProposeBlock(a *require.Assertions, fixture *fixtures.RestC
 }
 
 func TestNewAccountCanGoOnlineAndParticipate(t *testing.T) {
-	/*if runtime.GOOS == "darwin" {
-		t.Skip()
-	}
 	if testing.Short() {
 		t.Skip()
-	}*/
-	t.Skip() // temporary disable the test since it's failing.
+	}
 
 	t.Parallel()
 	a := require.New(fixtures.SynchronizedTest(t))
 
+	// Make the seed lookback shorter, otherwise will waite for 320 rounds
+	consensus := make(config.ConsensusProtocols)
+	shortPartKeysProtocol := config.Consensus[protocol.ConsensusCurrentVersion]
+	shortPartKeysProtocol.ApprovedUpgrades = map[protocol.ConsensusVersion]uint64{}
+	shortPartKeysProtocol.SeedLookback = 2
+	shortPartKeysProtocol.SeedRefreshInterval = 8
+	if runtime.GOARCH == "amd64" {
+		// amd64 platforms are generally quite capable, so accelerate the round times to make the test run faster.
+		shortPartKeysProtocol.AgreementFilterTimeoutPeriod0 = 1 * time.Second
+		shortPartKeysProtocol.AgreementFilterTimeout = 1 * time.Second
+	}
+	consensus[protocol.ConsensusVersion("shortpartkeysprotocol")] = shortPartKeysProtocol
+	
 	var fixture fixtures.RestClientFixture
+	fixture.SetConsensus(consensus)
 	fixture.Setup(t, filepath.Join("nettemplates", "TwoNodesOneOnline.json"))
 	defer fixture.Shutdown()
 	client := fixture.LibGoalClient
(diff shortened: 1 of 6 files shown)


test/e2e-go/features/participation/onlineOfflineParticipation_test.go
```diff
@@ -96,18 +101,28 @@ func waitForAccountToProposeBlock(a *require.Assertions, fixture *fixtures.RestC
...
 	if testing.Short() {
 		t.Skip()
-	}*/
-	t.Skip() // temporary disable the test since it's failing.
+	}
 
 	t.Parallel()
 	a := require.New(fixtures.SynchronizedTest(t))
 
+	// Make the seed lookback shorter, otherwise will waite for 320 rounds
```
tsachiherman, line 111: typo : waite -> wait

test/e2e-go/features/participation/onlineOfflineParticipation_test.go
```diff
@@ -159,35 +174,45 @@ func TestNewAccountCanGoOnlineAndParticipate(t *testing.T) {
...
-	a.True(delta > 0)
+	fundedAccountRoundBefore := balanceRoundOf(basics.Round(fundedRoundBefore), params)
+	fundedAccountRoundAfter := balanceRoundOf(basics.Round(fundedRoundAfter), params)
 
 	// Need to wait for funding to take effect on selection, then we can see if we're participating
 	// Stop before the account should become eligible for selection so we can ensure it wasn't
-	err = fixture.WaitForRoundWithTimeout(fundedRound + uint64(delta) - 1)
+	err = fixture.ClientWaitForRound(fixture.AlgodClient, uint64(fundedAccountRoundBefore),
+		time.Duration(uint64(globals.MaxTimePerRound) * uint64(fundedAccountRoundBefore)))
+	err = fixture.WaitForRoundWithTimeout(uint64(fundedAccountRoundBefore))
```
tsachiherman, line 198: why fundedAccountRoundBefore and not fundedAccountRoundAfter ?
You want to get to a round at which the newly founded account can vote. For fundedAccountRoundBefore, the account won't be able to vote.

test/e2e-go/features/participation/onlineOfflineParticipation_test.go
```diff
@@ -159,35 +174,45 @@ func TestNewAccountCanGoOnlineAndParticipate(t *testing.T) {
...
 
-	// check that account starts participating after a while
-	proposalWindow := 20 // arbitrary
-	blockWasProposedByNewAccountRecently := waitForAccountToProposeBlock(a, &fixture, newAccount, proposalWindow)
+	// Now wait until the round where the funded account will be used. Give additonal rounds for stability. 
+	err = fixture.ClientWaitForRound(fixture.AlgodClient, uint64(fundedAccountRoundAfter) + 4,
+		time.Duration(uint64(globals.MaxTimePerRound) * uint64(fundedAccountRoundAfter + 4)))
+	a.NoError(err)
+	
+	blockWasProposedByNewAccountRecently := fixture.VerifyBlockProposedRange(newAccount, int(fundedAccountRoundAfter) + 4,
```
tsachiherman, line 210: this would test *all* the blocks instead just the range of fundedAccountRoundBefore -> fundedAccountRoundAfter.
while it would likely to be working correctly, it's more error prone since you're looking at blocks that you know you won't need.

test/e2e-go/features/participation/onlineOfflineParticipation_test.go
```diff
@@ -96,18 +101,28 @@ func waitForAccountToProposeBlock(a *require.Assertions, fixture *fixtures.RestC
...
+	shortPartKeysProtocol.SeedRefreshInterval = 8
+	if runtime.GOARCH == "amd64" {
+		// amd64 platforms are generally quite capable, so accelerate the round times to make the test run faster.
+		shortPartKeysProtocol.AgreementFilterTimeoutPeriod0 = 1 * time.Second
+		shortPartKeysProtocol.AgreementFilterTimeout = 1 * time.Second
+	}
+	consensus[protocol.ConsensusVersion("shortpartkeysprotocol")] = shortPartKeysProtocol
+	
 	var fixture fixtures.RestClientFixture
+	fixture.SetConsensus(consensus)
```
tsachiherman, line 125: One node network would be enough for running this test.

test/e2e-go/features/participation/onlineOfflineParticipation_test.go
```diff
@@ -159,35 +174,45 @@ func TestNewAccountCanGoOnlineAndParticipate(t *testing.T) {
...
+		int(fundedAccountRoundBefore)-1)
+	a.False(blockWasProposed, "account should not be selected until BalLookback (round %d) passes", int(fundedAccountRoundBefore))
 
-	// check that account starts participating after a while
-	proposalWindow := 20 // arbitrary
-	blockWasProposedByNewAccountRecently := waitForAccountToProposeBlock(a, &fixture, newAccount, proposalWindow)
+	// Now wait until the round where the funded account will be used. Give additonal rounds for stability. 
+	err = fixture.ClientWaitForRound(fixture.AlgodClient, uint64(fundedAccountRoundAfter) + 4,
+		time.Duration(uint64(globals.MaxTimePerRound) * uint64(fundedAccountRoundAfter + 4)))
+	a.NoError(err)
```
tsachiherman, line 208: while it's ok to wait for 4 rounds, it should be needed. Assuming the fundedAccountRoundAfter calculation is correct, the proposer for these rounds MUST be the new account, since the old account has no money.

test/e2e-go/features/participation/onlineOfflineParticipation_test.go
```diff
@@ -159,35 +174,45 @@ func TestNewAccountCanGoOnlineAndParticipate(t *testing.T) {
...
 	a.NoError(err, "client should be able to get information about new account")
 	a.Equal(basics.Online.String(), newAccountStatus.Status, "new account should be online")
+
+	nodeStatus, _ = client.Status()
+	fundedRoundBefore := nodeStatus.LastRound
+	
 	// account receives almost all of rich account's stake (minus enough to
 	// keep it over MinBalance), so it will be selected for participation
 	amountToSend := richBalance - 3*transactionFee - amountToSendInitial - minAcctBalance
 	fixture.SendMoneyAndWait(onlineRound, amountToSend, transactionFee, richAccount, newAccount)
```
tsachiherman, line 184: Since you're sending all the money, I would advise to sent it as a closeto, so that the old account would be removed from the system. This would make sure the test is more robust.

test/e2e-go/features/participation/onlineOfflineParticipation_test.go
algonautshant: This is rewritten with different logic.

test/e2e-go/features/participation/onlineOfflineParticipation_test.go
algonautshant: This is rewritten with more exact checks.

test/e2e-go/features/participation/onlineOfflineParticipation_test.go
algonautshant: Done.

test/e2e-go/features/participation/onlineOfflineParticipation_test.go
algonautshant: Done.

PR #2240, opened 2021-06-04
Fix rel nightly test by extending catchpoint round timeout

egieseke:
## Summary

Fix catchpoint unit test by extending timeout from 10 to 60 seconds.

## Test Plan

Tested locally.

### DIFF
--- a/test/e2e-go/features/catchup/catchpointCatchup_test.go
+++ b/test/e2e-go/features/catchup/catchpointCatchup_test.go
@@ -220,7 +220,7 @@ func TestBasicCatchpointCatchup(t *testing.T) {
 	targetRound = uint64(37)
 	log.Infof("Second node catching up to round 36")
 	for {
-		err = fixture.ClientWaitForRound(secondNodeRestClient, currentRound, 10000*time.Millisecond)
+		err = fixture.ClientWaitForRound(secondNodeRestClient, currentRound, 60*time.Second)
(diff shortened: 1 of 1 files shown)


algobarb (approved):
This looks good based on our conversation.

test/e2e-go/features/catchup/catchpointCatchup_test.go
```diff
@@ -200,7 +200,7 @@ func TestBasicCatchpointCatchup(t *testing.T) {
 	targetRound = uint64(1)
 	log.Infof("Second node catching up to round 1")
 	for {
-		err = fixture.ClientWaitForRound(secondNodeRestClient, currentRound, 10000*time.Millisecond)
+		err = fixture.ClientWaitForRound(secondNodeRestClient, currentRound, 60*time.Second)
```
tsachiherman, line 203: Assuming that this issue is reproducible, we should figure out why it's happening. Just masking this issue would make the platform less reliable ( i.e. this type of waits would, for example, mask off a 21 second rounds blocks ). Maybe try to conduct a binary search to find where the regression point was ?

onetechnical: +1 on this. I'm a little worried this is highlighting a bigger issue, especially if it is a consistent failure.

egieseke:
First commit that caused this

PR #2241, opened 2021-06-04
fix docker builds: skip the `make deps`

tsachiherman:
## Summary

Fix the two docker files - avoid make deps step, which is no longer needed.
The compilation of go-algorand no longer requires the installation ( and therefore the validation ) of the dependencies.

## Test Plan

Tested manually.


algonautshant (approved):
LGTM

PR #1261, opened 2020-07-17
Eric/create stateful teal expect test

egieseke:
## Summary

Create an Expect test for stateful teal applications.  Specifically to test the command `goal app create`.  

## Test Plan

Ran Expect tests.

PR #2243, opened 2021-06-04
Expose creatable IDs to TEAL code within the same group

jdtzmn:
## Summary
This PR adds `gaid` and `gaids` opcodes (similar to `gtxn` and `gtxns`), which will allow smart contracts to access asset IDs of assets/apps which were created earlier in the same transaction group.

Previously, accessing creatable IDs required subsequent app calls with asset/application IDs as app arguments.

Closes #1261

## Test Plan

- Unit test for TEAL programs
  - Opcodes
  - Test cases where failure is expected
- Integration tests
  - Apps can get other app IDs and asset IDs in the same group


data/transactions/logic/doc.go
```diff
@@ -316,6 +316,7 @@ var txnFieldDocs = map[string]string{
 	"AssetCloseTo":             "32 byte address",
 	"GroupIndex":               "Position of this transaction within an atomic transaction group. A stand-alone transaction is implicitly element 0 in a group of 1",
 	"TxID":                     "The computed ID for this transaction. 32 bytes.",
+	"CreatableID":              "The assigned creatable ID for an asset or application creation transaction, or 0 if the transaction does not create anything.",
```
algorandskiy, line 319: this is a pseudo field, seems like the first one. Need to mention that and the fact it is apps-only.
Maybe a new opcode? :)

ledger/cow.go
```diff
@@ -56,6 +56,7 @@ type roundCowState struct {
 	commitParent *roundCowState
 	proto        config.ConsensusParams
 	mods         ledgercore.StateDelta
+	groupIdx     int
```
algorandskiy, line 59: move down to trackedCreatables and document by saying it is a group index of a current transaction used in conjunction with trackedCreatables

ledger/cow.go
```diff
@@ -68,15 +69,19 @@ type roundCowState struct {
...
-		commitParent: nil,
-		proto:        config.Consensus[hdr.CurrentProtocol],
-		mods:         ledgercore.MakeStateDelta(&hdr, prevTimestamp, hint, 0),
-		sdeltas:      make(map[basics.Address]map[storagePtr]*storageDelta),
+		lookupParent:      b,
+		commitParent:      nil,
+		proto:             config.Consensus[hdr.CurrentProtocol],
+		mods:              ledgercore.MakeStateDelta(&hdr, prevTimestamp, hint, 0),
+		sdeltas:           make(map[basics.Address]map[storagePtr]*storageDelta),
+		trackedCreatables: make(map[int]basics.CreatableLocator),
```
algorandskiy, line 84: why `CreatableLocator` and not `CreatableID`?

ledger/cow.go
```diff
@@ -255,6 +279,9 @@ func (cb *roundCowState) commitToParent() {
 			}
 		}
 	}
+	for gidx, tc := range cb.trackedCreatables {
+		cb.commitParent.trackedCreatables[gidx] = tc
+	}
```
algorandskiy, line 284: this is not really needed. In eval:
```golang
cow := eval.state.child(len(txgroup))
...
for gi, txad := range txgroup {
	var txib transactions.SignedTxnInBlock
	cow.setGroupIdx(gi)
	err := eval.transaction(txad.SignedTxn, evalParams[gi], txad.ApplyData, cow, &txib)
...
}
```
So there is a single cow for entire group and tracking stuff should only exist within it and have not been propagated down to block's cow where it does not make sense.

data/transactions/logic/doc.go
jannotti: GroupIndex is kind of a pseudo field, as is TxID.  (It's a hash that is not actually included in the txn.)

data/transactions/logic/eval_test.go
```diff
@@ -1534,6 +1539,42 @@ return
...
 
+func TestTxnCreatableID(t *testing.T) {
+	t.Parallel()
+	checkCreatableIDProg := `
+gtxn 0 CreatableID
+int 0
+>
+`
+	ops, err := AssembleStringWithVersion(checkCreatableIDProg, 4)
+	require.NoError(t, err)
```
jannotti, line 1550: use testProg

ledger/cow.go
```diff
@@ -204,6 +213,10 @@ func (cb *roundCowState) put(addr basics.Address, new basics.AccountData, newCre
 	}
 }
 
+func (cb *roundCowState) trackCreatable(creatable *basics.CreatableLocator) {
```
jannotti, line 216: I think I'd rather see this take it's argument by value, since it ends up copying anyway.

data/transactions/logic/README.md
```diff
@@ -275,6 +275,8 @@ Some of these have immediate data in the byte or bytes after the opcode.
 | 53 | GlobalNumByteSlice | uint64 | Number of global state byteslices in ApplicationCall. LogicSigVersion >= 3. |
 | 54 | LocalNumUint | uint64 | Number of local state integers in ApplicationCall. LogicSigVersion >= 3. |
 | 55 | LocalNumByteSlice | uint64 | Number of local state byteslices in ApplicationCall. LogicSigVersion >= 3. |
+| 56 | AppProgramExtraPages | uint64 |  |
+| 57 | CreatableID | uint64 | The assigned creatable ID for an asset or application creation transaction, or 0 if the transaction does not create anything. LogicSigVersion >= 4. |
```
jannotti, line 279: Are you allowed to look forward to a later slot?  It's nearly possible to do so (since the id is determined by the sequence number of the txn), but I don't think it's a good idea (since I want to make it possible to create an asa in teal - soon it won't be possible to know if an createable will be made or not).  Anyway, if you can't, we should document it, as usually you can look forward with gtxn.

data/transactions/logic/TEAL_opcodes.md
```diff
@@ -425,6 +425,8 @@ Overflow is an error condition which halts execution and fails the transaction.
 | 53 | GlobalNumByteSlice | uint64 | Number of global state byteslices in ApplicationCall. LogicSigVersion >= 3. |
 | 54 | LocalNumUint | uint64 | Number of local state integers in ApplicationCall. LogicSigVersion >= 3. |
 | 55 | LocalNumByteSlice | uint64 | Number of local state byteslices in ApplicationCall. LogicSigVersion >= 3. |
+| 56 | AppProgramExtraPages | uint64 |  |
+| 57 | CreatableID | uint64 | The assigned creatable ID for an asset or application creation transaction, or 0 if the transaction does not create anything. LogicSigVersion >= 4. |
```
jannotti, line 429: Also, I think we need to use "CreatedID" to be more clear that this doesn't just tell you the id of the asa or app the txn is working on, it's the one that's created in this transaction.

ledger/cow.go
jannotti: I'm not sure I agree.  It's the standard group index, right?  It may only be used with trackedCreatables now, but it has a real meaning that makes sense on its own.

ledger/cow.go
jannotti: Yeah, I think the CreatableLocator is the wrong thing.  (I think it goes back to a model in which we thought asas and app would be named by a creator and an account local index.). I don't think the Index in this thing is the appId, it's 5 if it's the user's 6th app created. I see your tests check that it's > 0, but you need to somehow test that it's a valid appid.

You might be able to compare it to `global CurrentApplicationID` in the creating teal code. I think that value is available, even during creation.  And depending on where that comes from, maybe that's an easier way to implement this.

ledger/cow.go
algorandskiy: cow is kind of transaction-agnostic and builds up own state while processing a transaction. So until it is committed to parent is has data relevant to `groupIdx`. Parent cow is a sum of all deltas, and `groupIdx` does not make sense there (it does not contains data from txn 0 applying). Same for `trackedCreatables`, it only has valid data within a group evaluation.
A clean solution might be `txnGroupCowState` that includes `roundCowState`  (well, naming) but it looks like as overkill here. So grouping data with a limited meaning IMO must.

ledger/cow.go
algorandskiy: yes, `CreatableIndex`

ledger/cow.go
jdtzmn: I thought so as well, but then I noticed that child cows are used during the StatefulEval function:
```go
func (cb *roundCowState) StatefulEval(params logic.EvalParams, aidx basics.AppIndex, program []byte) (pass bool, evalDelta basics.EvalDelta, err error) {
	// Make a child cow to eval our program in
	calf := cb.child(1)
	params.Ledger, err = newLogicLedger(calf, aidx)
	if err != nil {
		return false, basics.EvalDelta{}, err
	}
    ...
```
However, I agree with your premise and I'll play around and see if I can get it to work without committing state to the parent.

ledger/cow.go
jdtzmn: Good idea

data/transactions/logic/README.md
jdtzmn: I'll add errors to make sure you can't if you are going to make it impossible soon anyways. I was also wondering what should happen with LogicSigs? Will this txn field work from within stateless contracts?

data/transactions/logic/TEAL_opcodes.md
jdtzmn: I agree that CreatedID is more clear, but I noticed that creatable ID is already widely used within the codebase so I was hesitant to go against the existing precedent. Let me know if you still think it's worth changing though, since it's an easy change and I'm not too tied to one or the other.

ledger/cow.go
jdtzmn: Figured I would try to make it future-proof with a low overhead cost in case we wanted to expose other creatable properties to TEAL later, but I'll change it.

data/transactions/logic/TEAL_opcodes.md
jannotti: Let me look around at the uses.  I think it's ok to make it Created if Creatable is used more generically - any id for a creatable resource.  But if we are explicitly using that phrase only around an actual id created in a given transaction, then we should stick with it.  That is, I agree that this "a" CreatableID (in some sense that's its type), but this particular one is the CreatedID of the transaction.

ledger/cow.go
algorandskiy: this child cow is only for TEAL evaluation so it does not affect `trackedCreatables`. So copying them into child as already implemented should be sufficient: group cow has `trackedCreatables` filled, they are copying to a child in `StatefulEval`, logic ledger has all data needed. Data are read only and do not need to be copied back to group cow.

data/transactions/logic/doc.go
jannotti: We've concluded @algorandskiy is right.  Although there may be other pseudo fields, this would have been the first that depended on state unknown until the txn is being put on chain.  That would have made it a stateful only field.  That's weird.  We're going with `gaid` group asset/app id. Which will take a slot, like the other g-opcodes.

jannotti: It may be that we want to clean this up by AVM 1.0, so that we have a new `txn`-like opcode that is stateful only - it's for inspecting transaction specific things that are computed are proposal time.  It would subsume this opcode, the gload opcodes, and perhaps provide access to previous txns' apply data.

ledger/cow.go
jdtzmn: Oh I see - I misunderstood. I'll update it.

data/transactions/logic/eval.go
```diff
@@ -2102,6 +2103,56 @@ func opGtxnsa(cx *evalContext) {
 	cx.stack[last] = sv
 }
 
+func opGaidImpl(cx *evalContext, groupIdx int, opName string) (sv stackValue, err error) {
+	if groupIdx >= len(cx.TxnGroup) {
+		err = fmt.Errorf("%s lookup TxnGroup[%d] but it only has %d", opName, groupIdx, len(cx.TxnGroup))
+		return
+	} else if groupIdx > cx.GroupIndex {
+		err = fmt.Errorf("%s can't get future creatable ID of txn with index %d", opName, groupIdx)
```
tsachiherman, line 2111: this message is not wrong, but could be better written:
%s can't get creatable ID of txn ahead of the current one (index %d) in the transaction group

I think that the wording of "future" is misleading, as it might refer to future transaction groups or future rounds.

data/transactions/logic/README.md
```diff
@@ -275,6 +277,7 @@ Some of these have immediate data in the byte or bytes after the opcode.
 | 53 | GlobalNumByteSlice | uint64 | Number of global state byteslices in ApplicationCall. LogicSigVersion >= 3. |
 | 54 | LocalNumUint | uint64 | Number of local state integers in ApplicationCall. LogicSigVersion >= 3. |
 | 55 | LocalNumByteSlice | uint64 | Number of local state byteslices in ApplicationCall. LogicSigVersion >= 3. |
+| 56 | AppProgramExtraPages | uint64 |  |
```
shiqizng, line 280: Missing a description for AppProgramExtraPages?

data/transactions/logic/eval.go
```diff
@@ -2102,6 +2103,56 @@ func opGtxnsa(cx *evalContext) {
...
+		return
+	} else if groupIdx == cx.GroupIndex {
+		err = fmt.Errorf("can't use %s on self, use `global CurrentApplicationID` instead", opName)
```
tsachiherman, line 2114: interesting... but why not ? i.e. I think that we should be able to do that, right ?
If we do decide to keep this - should we also add this check to the static analysis ?

data/transactions/logic/eval.go
```diff
@@ -2102,6 +2103,56 @@ func opGtxnsa(cx *evalContext) {
...
+		return
+	} else if txn := cx.TxnGroup[groupIdx].Txn; !(txn.Type == protocol.ApplicationCallTx || txn.Type == protocol.AssetConfigTx) {
+		err = fmt.Errorf("can't use %s on txn that is not an app call nor an asset config txn with index %d", opName, groupIdx)
+		return
+	}
+
+	cid, err := cx.getCreatableID(groupIdx)
```
shiqizng, line 2121: if err!=nil, do you return immediately or do you still want to create sv?

data/transactions/logic/README.md
jdtzmn: Yes, I think that's the case as this was generated. Is there some description you'd like me to add and include in this PR?

data/transactions/logic/eval.go
jdtzmn: My assumption is that an error indicates an application call failure, which causes code execution to end and the transaction/transaction group to be rejected. In other words, the `sv` would never be used if `err != nil`, so I don't see a need to set it.

data/transactions/logic/README.md
shiqizng: maybe Number of extra pages in ApplicationCall. LogicSigVersion >= 4

jdtzmn: Will do!

jannotti:
Remember this is autogenerated by opdoc.

data/transactions/logic/eval.go
jannotti: I don't think we should allow access to "self" with these kind of opcodes (`gload` and `gaid`).

We do allow it with `gtxn` even though it is "dumb", that is, it's an extra byte of code for no advantage over `txn`.  I think the reason we allow it is because it would be awfully strange to have the opcode allow access to earlier txns and later txns, but exclude self.

With these opcodes, that already can't look into higher slots, that reasoning breaks down.  We are definitely not going to let them look *past* self, so should we exclude self too?  It won't leave an unsightly hole this time, but it remains an ugly way to do something that is already allowed.

The argument for: maybe you want to loop from 0 to self, doing the same work on each transaction, including self.  My argument against: That's almost certainly an off-by-one bug that you didn't intend.  When txns use an opcode like this, they're trying to confirm that something happened in another transaction.  I can't imagine needing to do the same kind of examination on a bunch of transactions *including self*.

But I'm not sure we should add a static check.  We *can't* for `gaids` which reads the index from scratch.  And we can't at assembly time for either of them, since we don't know our index at assembly time.  The only possibility is to check the `gaid <slot>` case, at network submission time.  That will certainly fail, yes, but so would `gtxn 62`. Nonetheless, we allow it (at Check() time, and fail at Eval() time). I think we might want a comprehensive static checker, but I'm not sure it should be in algod.  It should be a tool to find bugs ahead of time, used by developers.

data/transactions/logic/README.md
jasonpaulos: Yes, ideally the doc will say something along the lines of "Number of additional pages for each of the application's approval and clear state programs. A page is 1024 bytes." Does that imply strongly enough that 1 extra page means 2048 more bytes total?

jdtzmn: I'm not sure it does – how about
> Number of additional pages for each of the application's approval and clear state programs. An ExtraProgramPages of 1 means 2048 more total bytes, or 1024 for each program.

It is very long though, so hopefully we can pare it down.

data/transactions/logic/eval.go
jdtzmn: > I would be in favor of a single >= check though, and write the error message so it simply says you can only look at previous slots in the group.

I'd reword the error message to state that `gaid` is only for looking at previous slots in the group, but I'd rather not leave the user hanging by not offering an alternative when there is an straightforward one available.

ledger/cow.go
```diff
@@ -204,6 +214,10 @@ func (cb *roundCowState) put(addr basics.Address, new basics.AccountData, newCre
 	}
 }
 
+func (cb *roundCowState) trackCreatable(creatable basics.CreatableLocator) {
```
jannotti, line 217: Just pass in the index, now that you don't need the Locator

ledger/eval.go
```diff
@@ -248,6 +252,11 @@ func (cs *roundCowState) Put(addr basics.Address, acct basics.AccountData) error
 
 func (cs *roundCowState) PutWithCreatable(addr basics.Address, acct basics.AccountData, newCreatable *basics.CreatableLocator, deletedCreatable *basics.CreatableLocator) error {
 	cs.put(addr, acct, newCreatable, deletedCreatable)
+
+	// store the creatable locator
+	if newCreatable != nil {
+		cs.trackCreatable(*newCreatable)
```
jannotti, line 258: just the index

jannotti (changes requested):
Just a tiny change and I'll approve immediately.

ledger/cow.go
algorandskiy: +1 to JJ

data/transactions/logic/README.md
```diff
@@ -275,6 +277,7 @@ Some of these have immediate data in the byte or bytes after the opcode.
 | 53 | GlobalNumByteSlice | uint64 | Number of global state byteslices in ApplicationCall. LogicSigVersion >= 3. |
 | 54 | LocalNumUint | uint64 | Number of local state integers in ApplicationCall. LogicSigVersion >= 3. |
 | 55 | LocalNumByteSlice | uint64 | Number of local state byteslices in ApplicationCall. LogicSigVersion >= 3. |
+| 56 | ExtraProgramPages | uint64 | Number of additional pages for each of the application's approval and clear state programs. An ExtraProgramPages of 1 means 2048 more total bytes, or 1024 for each program. LogicSigVersion >= 4. |
```
shiqizng, line 280: @jannotti , should it be LogicSigVersion >= 3? that's the value in langSpec file.

algorandskiy: we enable it [along with teal v4](https://github.com/algorand/go-algorand/pull/2255/files#diff-a75ef7294adbd7ae646c49ae2a518253936983c0a16594b84f20f15403767a2fR935) so this comment looks correct. 

Additionally I have that is version 4. So the lang spec is incorrect.
```{AppProgramExtraPages, StackUint64, 4}```

jannotti (approved):

shiqizng (approved):

AyAggarwal (approved):

PR #2247, opened 2021-06-06
Remove ci-deps and update GOPROXY in Dockerfiles

onetechnical:
## Summary

Recent refactoring has changed dependency installation. ci-deps did not appear to be used in a meaningful way, so it was removed from the Makefile and Dockerfiles. Additionally, downloading from go has become unstable, so updating the GOPROXY options should help.

## Test Plan

Run build pipeline against branch and verify this works.

### DIFF
--- a/docker/build/cicd.ubuntu.Dockerfile
+++ b/docker/build/cicd.ubuntu.Dockerfile
@@ -15,9 +15,9 @@ ENV GOROOT=/usr/local/go \
 RUN mkdir -p $GOPATH/src/github.com/algorand
 COPY . $GOPATH/src/github.com/algorand/go-algorand
 ENV PATH=$GOPATH/bin:$GOROOT/bin:$PATH \
-    GOPROXY=https://pkg.go.dev
+    GOPROXY=https://proxy.golang.org,https://pkg.go.dev
 WORKDIR $GOPATH/src/github.com/algorand/go-algorand
-RUN make ci-deps && make clean
+RUN make clean
 RUN rm -rf $GOPATH/src/github.com/algorand/go-algorand && \
     mkdir -p $GOPATH/src/github.com/algorand/go-algorand
 RUN echo "vm.max_map_count = 262144" >> /etc/sysctl.conf
--- a/docker/build/cicd.ubuntu.Dockerfile
+++ b/docker/build/cicd.ubuntu.Dockerfile
@@ -15,7 +15,7 @@ ENV GOROOT=/usr/local/go \
 RUN mkdir -p $GOPATH/src/github.com/algorand
 COPY . $GOPATH/src/github.com/algorand/go-algorand
 ENV PATH=$GOPATH/bin:$GOROOT/bin:$PATH \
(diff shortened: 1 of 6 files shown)


scripts/release/mule/Makefile.mule
```diff
@@ -8,7 +8,7 @@ ci-clean:
 	rm -rf tmp
 
 ci-deps:
-	scripts/configure_dev-deps.sh && \
+	scripts/buildtools/install_buildtools.sh && \
```
tsachiherman, line 11: Are you sure that ci-deps is truely needed ?

docker/build/cicd.ubuntu.Dockerfile
```diff
@@ -15,9 +15,9 @@ ENV GOROOT=/usr/local/go \
 RUN mkdir -p $GOPATH/src/github.com/algorand
 COPY . $GOPATH/src/github.com/algorand/go-algorand
 ENV PATH=$GOPATH/bin:$GOROOT/bin:$PATH \
-    GOPROXY=https://pkg.go.dev
+    GOPROXY=https://proxy.golang.org,https://pkg.go.dev
```
tsachiherman, line 18: If the goproxy needs to be updated here, wouldn't it need to be updated everywhere else as well ?

onetechnical: Yes, this is just me testing things separately. I'll shift this back to draft.

tsachiherman: No worries. Its perfectly fine to do a partial fix in one pr, and follow up in another.

docker/build/cicd.ubuntu.Dockerfile
```diff
@@ -15,9 +15,9 @@ ENV GOROOT=/usr/local/go \
 RUN mkdir -p $GOPATH/src/github.com/algorand
 COPY . $GOPATH/src/github.com/algorand/go-algorand
 ENV PATH=$GOPATH/bin:$GOROOT/bin:$PATH \
-    GOPROXY=https://pkg.go.dev
+    GOPROXY=https://proxy.golang.org,https://pkg.go.dev,https://goproxy.io,direct
 WORKDIR $GOPATH/src/github.com/algorand/go-algorand
-RUN make ci-deps && make clean
```
bricerisingalgorand, line 20:
```
./scripts/buildtools/install_buildtools.sh
```
I think we'll need to run this for the image to work, unless you've already seen that it works without it

algobarb (approved):
The GOPROXY changes look good. I'm not too familiar with the new changes we've done in ci-deps, but it looks like we're removing it, including the check_deps.sh. If that's ok, then I'm ok with these changes.

docker/build/cicd.ubuntu.Dockerfile
onetechnical: Yes, this works without it for the build. (Historically it didn't, I know.)

bricerisingalgorand (approved):

PR #2250, opened 2021-06-07
simplify CatchpointCatchupAccessor.FinishBlocks

brianolson:
## Summary

abort branch was never used. delete unused code.

## Test Plan

Still passes available unit tests and integration tests.


tsachiherman:
I think that some of the functionality here is still needed:
if we start a fast catchup, and then stop it - we want the downloaded blocks to be removed.
I do realize that it's not doing it at this time.. but it should. When that does get implemented, we should use the functionality you're deleting..

brianolson:
dead code lies dreaming, and may yet wake and resurface

PR #2251, opened 2021-06-07
Ignore msgp_gen.go files in coverage report.

winder:
## Summary

Exclude msgp_gen.go from the coverage report sent to codecov.


tsachiherman:
My long-term plan was to add auto-generated unit test to msgp so that it would always be 100% covered. I'm ok with excluding this off for now though.

winder:
There was a lot more coverage here than I expected

PR #2252, opened 2021-06-07
If the branch is rel/nightly set channel to nightly

onetechnical:
## Summary

The way that the new pipeline scripts computed channel did not account for the rel/nightly branch, which should set the channel to nightly. This was preventing rpm and deb binaries from being built. This change identifies if the TRAVIS_BRANCH env is set to rel/nightly, and if so, will set the channel to nightly.

## Test Plan

Verified with: `TRAVIS_BRANCH=rel/nightly ./scripts/release/mule/common/get_channel.sh devnet` which returned nightly. Specifying 'betanet', 'testnet', and 'mainnet' remain the same. 

Should verify with a nightly build using the new pipeline.

### DIFF
--- a/scripts/release/mule/common/get_channel.sh
+++ b/scripts/release/mule/common/get_channel.sh
@@ -8,6 +8,10 @@ then
 elif [ "$NETWORK" = mainnet ] || [ "$NETWORK" = testnet ]
 then
     echo stable
+elif [ "$TRAVIS_BRANCH" = 'rel/nightly' ]
+then
+    # The rel/nightly branch is only the nightly channel
+    echo nightly
 else
     echo dev
 fi


egieseke (approved):

algobarb (approved):
Makes sense to me

winder:
Should this use `scripts/compute_branch_channel.sh`?

onetechnical:
> Should this use `scripts/compute_branch_channel.sh`?

In this particular case, I'd like to leave the scripting because it's pulled in by the new pipeline scripting and passes network. In the long run, yes, de-duping these would be a good idea. I'm not sure if the master channel should be 'master' or 'dev' though.

PR #2253, opened 2021-06-07
Use cgo deflate compression for txnsync msgs

nicholasguoalgorand:
## Summary

Use the cgo library in `util/compress` for txnsync msg compression

## Test Plan

Rerun unit tests

### DIFF
--- a/txnsync/txngroups.go
+++ b/txnsync/txngroups.go
@@ -156,11 +156,19 @@ func decodeTransactionGroups(ptg packedTransactionGroups, genesisID string, gene
 }
 
 func decompressTransactionGroupsBytes(data []byte, lenDecompressedBytes uint64) (decoded []byte, err error) {
-	if lenDecompressedBytes > maxEncodedTransactionGroupBytes || lenDecompressedBytes < uint64(len(data)) {
+	compressionRatio := lenDecompressedBytes / uint64(len(data)) // data should have been compressed between 0 and 95%
+	if lenDecompressedBytes > maxEncodedTransactionGroupBytes || compressionRatio <= 0 || compressionRatio >= 20 {
 		return nil, fmt.Errorf("invalid lenDecompressedBytes: %d, lenCompressedBytes: %d", lenDecompressedBytes, len(data))
 	}
 	out := make([]byte, 0, lenDecompressedBytes)
-	return compress.Decompress(data, out)
+	decoded, err = compress.Decompress(data, out)
+	if err != nil {
+		return
+	}
+	if uint64(len(decoded)) != lenDecompressedBytes {
+		return nil, fmt.Errorf("lenDecompressedBytes didn't match: expected %d, actual %d", lenDecompressedBytes, len(decoded))
+	}
+	return
 }
 
 func releaseEncodedTransactionGroups(buffer []byte) {
--- a/txnsync/txngroups.go
+++ b/txnsync/txngroups.go
@@ -27,8 +27,8 @@ import (
 )
 
 // gzip performance constants measured by BenchmarkTxnGroupCompression
-const estimatedGzipCompressionSpeed = 23071093.0 // bytes per second of how fast gzip compresses data
-const estimatedGzipCompressionGains = 0.32       // fraction of data reduced by gzip on txnsync msgs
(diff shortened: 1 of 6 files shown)


txnsync/txngroups.go
```diff
@@ -157,18 +155,12 @@ func decodeTransactionGroups(data []byte, compressionFormat byte, genesisID stri
...
-	_, err = buf.ReadFrom(zr)
-	if err != nil {
-		return nil, fmt.Errorf("error gzip decompressing data: %w", err)
+func decompressTransactionGroupsBytes(data []byte, lenDecompressedBytes uint64) (decoded []byte, err error) {
+	if lenDecompressedBytes > maxEncodedTransactionGroupBytes || lenDecompressedBytes < uint64(len(data)) {
+		return nil, fmt.Errorf("invalid lenDecompressedBytes: %d, lenCompressedBytes: %d", lenDecompressedBytes, len(data))
 	}
-	return buf.Bytes(), nil
+	out := make([]byte, 0, lenDecompressedBytes)
+	return compress.Decompress(data, out)
```
tsachiherman, line 163: Add verification that the return value of the Decompress is lenDecompressedBytes bytes long.

txnsync/txngroups_test.go
```diff
@@ -298,6 +298,9 @@ func txnGroupsData() (txnGroups []transactions.SignedTxGroup, genesisID string,
 			return
 		}
 		ntx++
+		//if ntx > 3 {
+		//	break
+		//}
```
tsachiherman, line 303: remove if not needed.

txnsync/txngroups.go
```diff
@@ -157,18 +155,12 @@ func decodeTransactionGroups(data []byte, compressionFormat byte, genesisID stri
...
-	defer zr.Close()
-	if err != nil {
-		return nil, fmt.Errorf("error gzip decompressing data: %w", err)
-	}
-	var buf bytes.Buffer
-	_, err = buf.ReadFrom(zr)
-	if err != nil {
-		return nil, fmt.Errorf("error gzip decompressing data: %w", err)
+func decompressTransactionGroupsBytes(data []byte, lenDecompressedBytes uint64) (decoded []byte, err error) {
+	if lenDecompressedBytes > maxEncodedTransactionGroupBytes || lenDecompressedBytes < uint64(len(data)) {
```
tsachiherman, line 159: one more restriction : make sure that the expected decompression ratio makes sense -
i.e. ratio of 0% - 95% of compression gains might be reasonable. Anything better is unlikely.

txnsync/txngroups.go
```diff
@@ -157,18 +155,20 @@ func decodeTransactionGroups(data []byte, compressionFormat byte, genesisID stri
...
 }
 
-func decompressTransactionGroupsBytes(data []byte) (decoded []byte, err error) {
-	zr, err := gzip.NewReader(bytes.NewBuffer(data))
-	defer zr.Close()
-	if err != nil {
-		return nil, fmt.Errorf("error gzip decompressing data: %w", err)
+func decompressTransactionGroupsBytes(data []byte, lenDecompressedBytes uint64) (decoded []byte, err error) {
+	compressionRatio := lenDecompressedBytes / uint64(len(data)) // data should have been compressed between 0 and 95%
+	if lenDecompressedBytes > maxEncodedTransactionGroupBytes || compressionRatio <= 0 || compressionRatio >= 20 {
```
tsachiherman, line 160: for completeness, we also need to verify the compression ratio while we compress the data. If we get >95%, we should send it uncompressed so that it won't fail in the decompression.
( this is more of a hypothetical issue - I don't believe we'll be that lucky )

tsachiherman: please also have the magic number 20 define as a const.. but it looks good otherwise.

tsachiherman:
> the decompressed version is used when this function returns an error and the error is just logged

Do you think we could do better in term of logging the precise error ? The code is appear to be correct, but it would be cleaner if we could conduct as many of the testing ahead of time, and log a more detailed error.

I don't have any concrete expectation here, but I'd like to have these warning be a useful on their own.

tsachiherman:
I'm going to merge this in for now, but would like you to reiterate on the error testing path and try to make it more "trivial".

tsachiherman (approved):

PR #2254, opened 2021-06-08
Disable compact certs and auction tests

bricerisingalgorand:
## Summary

They need work and shouldn't block our releases since they features aren't currently in use/enabled

## Test Plan

I ran these tests locally and saw that they all got skipped

### DIFF
--- a/test/e2e-go/features/auction/auctionCancel_test.go
+++ b/test/e2e-go/features/auction/auctionCancel_test.go
@@ -27,6 +27,7 @@ import (
 )
 
 func TestStartAndCancelAuctionNoBids(t *testing.T) {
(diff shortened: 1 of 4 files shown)


tsachiherman:
I'm ok with these changes, however, i would prefer to see the auction test along with the auction code being removed from go-algorand. Could you reach out to @algoidan , as he might be looking into that ?

Disabling the compact cert test is fine regardless.

onetechnical (approved):
Idan isn't actively working on this, but may in the future. I think this is fine for now.

onetechnical:
Can you also make sure to add a ticket to re-enable and debug compact cert code?

egieseke (approved):
Looks good.

PR #2255, opened 2021-06-08
Introduce V28 consensus version

algorandskiy:
* TEAL v4
* Larger programs
* Larger app/asset lookup limits
* Longer asset URL
* Fee pooling within a group
* Keyreg txn additional checks

Removed InitialRewardsRateCalculation and PaysetCommit from vFuture since it is already in v26

Fixed some tests after exposing zero fees and strict keyreg as current consensus

### DIFF
--- a/config/consensus.go
+++ b/config/consensus.go
@@ -89,7 +89,7 @@ type ConsensusParams struct {
 	// DefaultKeyDilution specifies the granularity of top-level ephemeral
 	// keys. KeyDilution is the number of second-level keys in each batch,
 	// signed by a top-level "batch" key.  The default value can be
-	// overriden in the account state.
+	// overridden in the account state.
 	DefaultKeyDilution uint64
 
 	// MinBalance specifies the minimum balance that can appear in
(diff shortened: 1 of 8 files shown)


config/consensus.go
```diff
@@ -940,25 +960,9 @@ func initConsensusProtocols() {
...
 	vFuture.CompactCertSecKQ = 128
 
-	vFuture.EnableKeyregCoherencyCheck = true
-
-	// Enable support for larger app program size
-	vFuture.MaxExtraAppProgramPages = 3
-	vFuture.MaxAppProgramLen = 2048
-
-	// enable the InitialRewardsRateCalculation fix
-	vFuture.InitialRewardsRateCalculation = true
```
tsachiherman, line 950: the following `vFuture.PaysetCommit = PaysetCommitMerkle` can also be dropped.

onetechnical:
For #2263 do we also need:

```
	vFuture.MaxAppTxnForeignApps = 8
	vFuture.MaxAppTxnForeignAssets = 8
```

config/consensus.go
algorandskiy: done

tsachiherman (approved):
looks good to me.

jannotti (approved):
I don't know the part keys stuff, but the included things in v28 looks right to me.

PR #2259, opened 2021-06-09
1k assets: large asset holdings support in pingpong

algorandskiy:
## Summary

* New `numassethld` cli parameter
* Changed required balance calculation

## Test Plan

Tested manually
TODO: publish TPS


gmalouf:
@algorandskiy / @tsachiherman - do we want to contribute this to Hyper Flow's performance work or should we close?

tsachiherman:
I think that we want to keep the branch, but close the PR.

PR #2263, opened 2021-06-09
Allow AVM code to access a max number of foreign refs.

jannotti:
Along the lines of similar "combine" PRs, this allows AVM code to access a *total* number of "foreign" references, as opposed to limiting each one individually.

### DIFF
--- a/config/consensus.go
+++ b/config/consensus.go
@@ -262,7 +262,7 @@ type ConsensusParams struct {
 
 	// maximum number of "foreign references" (accounts, asa, app)
 	// that can be attached to a single app call.
-	MaxAppTxnReferences int
+	MaxAppTotalTxnReferences int
 
 	// maximum cost of application approval program or clear state program
 	MaxAppProgramCost int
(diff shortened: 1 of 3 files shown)


config/consensus.go
```diff
@@ -859,6 +863,10 @@ func initConsensusProtocols() {
 	// Can look up 2 assets to see asset parameters
 	v24.MaxAppTxnForeignAssets = 2
 
+	// Appears superfluous here in v24, but allows increasing the
+	// individual limits while maintaining max components in later vers.
+	v24.MaxAppTxnReferences = 8
```
algorandskiy, line 868: mb call it `MaxAppTxnTotalReferences` and comment is is set to v24.MaxAppTxnForeignAssets + v24.MaxAppTxnForeignApps + v24.MaxAppTxnAccounts.

config/consensus.go
```diff
@@ -953,6 +961,16 @@ func initConsensusProtocols() {
...
 
+	// Individual limits raised
+	vFuture.MaxAppTxnForeignApps = 8
+	vFuture.MaxAppTxnForeignAssets = 8
+	// but MaxAppTxnReferences is unchanged.
+
+	// MaxAppTxnAccounts has not been raised yet.  It is already
+	// higher (4) and there is a multiplicative effect in
+	// "reachability" between accounts and creatables, so we
+	// retain 4 x 4 as worst case.
```
algorandskiy, line 972:
```suggestion
	// retain 4 x 8 as worst case.
```
4 accounts and 8 creatables each?

shiqizng (dismissed):

shiqizng:
Looks good.

config/consensus.go
jannotti: Since the total (accounts, asa, apps) is 8, the worst case is splitting that total in half, 4 for account, 4 for asa/app.

config/consensus.go
jannotti: But in the future (vFuture!) it won't be the sum.  It's sum here, so that it has no effect on v24.  In vFuture, it's actually the same as the app or asa values.  That's to allow the teal code to freely decide which of the object types it needs more access to.  Maybe your point was to explain why it's 8 here.  I agree - will do.

And I do like your name better.  It's in keeping with how we added Total in some of the other combined PRs.

config/consensus.go
algorandskiy: ah, right, got confused

jdtzmn (approved):

PR #2265, opened 2021-06-09
testing: add unit tests for ledger BlockQueue

nicholasguoalgorand:
## Summary

Add tests to increase coverage for BlockQueue

## Test Plan

These are tests

### DIFF
--- a/ledger/blockqueue_test.go
+++ b/ledger/blockqueue_test.go
@@ -45,12 +45,13 @@ func TestPutBlockTooOld(t *testing.T) {
 
 	blk := bookkeeping.Block{}
 	var cert agreement.Certificate
-	err = l.blockQ.putBlock(blk, cert)
+	err = l.blockQ.putBlock(blk, cert) // try putBlock for a block in a previous round
 
 	expectedErr := &ledgercore.BlockInLedgerError{}
 	require.True(t, errors.As(err, expectedErr))
 }
 
+// TestGetEncodedBlockCert tests getEncodedBlockCert with valid and invalid round numbers.
 func TestGetEncodedBlockCert(t *testing.T) {
 	genesisInitState, _, _ := genesis(10)
 
@@ -60,8 +61,9 @@ func TestGetEncodedBlockCert(t *testing.T) {
 	require.NoError(t, err)
 	defer l.Close()
 
-	blk := bookkeeping.Block{BlockHeader: bookkeeping.BlockHeader{Round: 1}}
-	cert := agreement.Certificate{Round: 1}
+	blkent := randomBlock(1)
+	blk := blkent.block
+	cert := blkent.cert
 	err = l.blockQ.putBlock(blk, cert)
 	require.NoError(t, err)
 
--- a/ledger/blockqueue_test.go
+++ b/ledger/blockqueue_test.go
@@ -18,7 +18,6 @@ package ledger
 
 import (
 	"fmt"
-	"github.com/algorand/go-algorand/protocol"
 	"testing"
 
 	"github.com/stretchr/testify/require"
@@ -28,6 +27,7 @@ import (
 	"github.com/algorand/go-algorand/crypto"
 	"github.com/algorand/go-algorand/data/bookkeeping"
 	"github.com/algorand/go-algorand/logging"
+	"github.com/algorand/go-algorand/protocol"
 )
 
 func TestPutOldBlock(t *testing.T) {
--- a/ledger/blockqueue_test.go
+++ b/ledger/blockqueue_test.go
@@ -30,7 +30,7 @@ import (
 	"github.com/algorand/go-algorand/protocol"
 )
 
-func TestPutOldBlock(t *testing.T) {
+func TestPutBlockTooOld(t *testing.T) {
 	genesisInitState, _, _ := genesis(10)
 
 	dbName := fmt.Sprintf("%s.%d", t.Name(), crypto.RandUint64())
--- a/ledger/blockqueue_test.go
+++ b/ledger/blockqueue_test.go
@@ -49,6 +49,15 @@ func TestPutBlockTooOld(t *testing.T) {
 
 	expectedErr := &ledgercore.BlockInLedgerError{}
 	require.True(t, errors.As(err, expectedErr))
+
+	blkent := randomBlock(1)
+	blk = blkent.block
+	cert = blkent.cert
+	err = l.blockQ.putBlock(blk, cert) // add block for round 1 to blockQueue
+	require.NoError(t, err)
+
+	err = l.blockQ.putBlock(blk, cert) // try adding same block again (should fail)
+	require.True(t, errors.As(err, expectedErr))
 }
 
 // TestGetEncodedBlockCert tests getEncodedBlockCert with valid and invalid round numbers.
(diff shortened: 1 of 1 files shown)


ledger/blockqueue_test.go
```diff
@@ -0,0 +1,81 @@
...
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with go-algorand.  If not, see <https://www.gnu.org/licenses/>.
+
+package ledger
+
+import (
+	"fmt"
+	"github.com/algorand/go-algorand/protocol"
```
tsachiherman, line 21: Move to the group below

ledger/blockqueue_test.go
```diff
@@ -0,0 +1,81 @@
...
+	const inMem = true
+	cfg := config.GetDefaultLocal()
+	cfg.Archival = true
+	l, err := OpenLedger(logging.Base(), dbName, inMem, genesisInitState, cfg)
+	require.NoError(t, err)
+	defer l.Close()
+
+	blk := bookkeeping.Block{}
+	var cert agreement.Certificate
+	err = l.blockQ.putBlock(blk, cert)
```
tsachiherman, line 46: Please do something more meaningful, like checking that the opearion truely did what it was supposed to do..

ledger/blockqueue_test.go
```diff
@@ -0,0 +1,81 @@
...
+	cert := agreement.Certificate{Round: 1}
+	err = l.blockQ.putBlock(blk, cert)
+	require.NoError(t, err)
+
+	var blkBytes []byte
+	var certBytes []byte
+
+	blkBytes, certBytes, err = l.blockQ.getEncodedBlockCert(0)
+	require.Equal(t, protocol.Encode(&genesisInitState.Block), blkBytes)
+	require.Equal(t, protocol.Encode(&agreement.Certificate{}), certBytes)
```
tsachiherman, line 71: Use a non empty cert.. otherwise, this test would be very meaningful, wouldn't ot ?

tsachiherman (changes requested):
Tests should be more than just testing return codes.. please verify that the implementation performed the expected task.

ledger/blockqueue_test.go
nicholasguoalgorand: Sorry, the purpose of this test was only to make sure that `putBlock` would error when the block put was too old. Perhaps the name of the test should have been made more clear to convey that.

ledger/blockqueue_test.go
nicholasguoalgorand: I believe by default round 0 has an empty cert. A non empty cert is checked for round 1. See line 76 `require.Equal(t, protocol.Encode(&cert), certBytes)` which checks that the cert declared on line 62 `cert := agreement.Certificate{Round: 1}` was fetched properly.

ledger/blockqueue_test.go
tsachiherman: ah - so, in that case,  please make sure you're testing a specific error rather than checking the existence of an error.

ledger/blockqueue_test.go
```diff
@@ -0,0 +1,81 @@
...
+	require.NoError(t, err)
+
+	blkBytes, certBytes, err = l.blockQ.getEncodedBlockCert(1)
+	require.Equal(t, protocol.Encode(&blk), blkBytes)
+	require.Equal(t, protocol.Encode(&cert), certBytes)
+	require.NoError(t, err)
+
+	_, _, err = l.blockQ.getEncodedBlockCert(100) // should not be entry for this round
+	require.Error(t, err)
```
tsachiherman, line 80: test specific error code.

ledger/blockqueue_test.go
```diff
@@ -0,0 +1,81 @@
...
+	"github.com/algorand/go-algorand/protocol"
+)
+
+func TestPutBlockTooOld(t *testing.T) {
+	genesisInitState, _, _ := genesis(10)
+
+	dbName := fmt.Sprintf("%s.%d", t.Name(), crypto.RandUint64())
+	const inMem = true
+	cfg := config.GetDefaultLocal()
+	cfg.Archival = true
```
tsachiherman, line 39: I could be wrong, but I don't believe you need an archival node here.

ledger/blockqueue_test.go
```diff
@@ -0,0 +1,81 @@
...
+	l, err := OpenLedger(logging.Base(), dbName, inMem, genesisInitState, cfg)
```
tsachiherman, line 40: given that it's in-memory database, it would be enough to use the t.Name() as the dbName. I don't see any benefits of formatting a special database name unless it's going to be on-disk database.

ledger/blockqueue_test.go
```diff
@@ -0,0 +1,85 @@
...
+	"github.com/algorand/go-algorand/agreement"
+	"github.com/algorand/go-algorand/config"
+	"github.com/algorand/go-algorand/crypto"
+	"github.com/algorand/go-algorand/data/bookkeeping"
+	"github.com/algorand/go-algorand/ledger/ledgercore"
+	"github.com/algorand/go-algorand/logging"
+	"github.com/algorand/go-algorand/protocol"
+)
+
+func TestPutBlockTooOld(t *testing.T) {
```
algonautshant, line 35: Maybe this should be renamed to TestPutDuplicateBlock?

ledger/blockqueue_test.go
```diff
@@ -0,0 +1,85 @@
...
+
+	blk := bookkeeping.Block{}
+	var cert agreement.Certificate
+	err = l.blockQ.putBlock(blk, cert)
+
+	expectedErr := &ledgercore.BlockInLedgerError{}
+	require.True(t, errors.As(err, expectedErr))
+}
+
+func TestGetEncodedBlockCert(t *testing.T) {
```
algonautshant, line 54: Can you please add a description, listing what the test is doing? 
i.e. test getEncodedBlockCert with a valid and invalid round numbers.

ledger/blockqueue_test.go
```diff
@@ -0,0 +1,85 @@
...
+
+	var blkBytes []byte
+	var certBytes []byte
+
+	blkBytes, certBytes, err = l.blockQ.getEncodedBlockCert(0)
+	require.Equal(t, protocol.Encode(&genesisInitState.Block), blkBytes)
+	require.Equal(t, protocol.Encode(&agreement.Certificate{}), certBytes)
+	require.NoError(t, err)
+
+	blkBytes, certBytes, err = l.blockQ.getEncodedBlockCert(1)
```
algonautshant, line 76: This seems to be the only getEncodedBlockCert test. 
For block of round 1, except for the round number, everything seems to be 0. Maybe populate the block and cert fields and test that the round trip is valid?

ledger/blockqueue_test.go
nicholasguoalgorand: I think it should be TooOld because it checks to see if the round number is in the past

algonautshant: Sounds good. 
Would you make it such that it adds a handful of blocks, and checks several of them? 
Since round 0 is a special case, it is a good idea to diversify by checking with 0, other than 0, and the most recent one.

tsachiherman (approved):
looks good.

PR #2266, opened 2021-06-09
Temporarily disabled TestBasicCatchpointCatchup test

egieseke:
Temporarily disabled TestBasicCatchpointCatchup and reset test timeouts from 60 to 10 seconds.

## Summary

Disabled TestBasicCatchpointCatchup test until the failure is resolved to keep this from blocking the rel/nightly build. 

## Test Plan

Tested locally to ensure that the test file runs successfully.

### DIFF
--- a/test/e2e-go/features/catchup/catchpointCatchup_test.go
+++ b/test/e2e-go/features/catchup/catchpointCatchup_test.go
@@ -79,6 +79,7 @@ func (ec *nodeExitErrorCollector) Print() {
 }
 
 func TestBasicCatchpointCatchup(t *testing.T) {
+	t.Skip("Temporarily disabling since they need work and shouldn't block releases")
 	if testing.Short() {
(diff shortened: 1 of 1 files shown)


algobarb (approved):
looks ok to me.

algobarb:
We just to test it without the catchpoint test and we don't want to merge it with the test deactivated, right?

egieseke:
> We just want to test it without the catchpoint test and we don't want to merge it with the test deactivated, right?

I think that we do want to merge it so that we can unblock the rel/nightly deployment to devnet. 
Comments @yaovi-a or @onetechnical ?

onetechnical (approved):

onetechnical:
I want to be clear that we _temporarily_ want to disable this test. It's only run on rel/nightly, so we want to see if this is the only change breaking rel/nightly builds. We should revert it later.

PR #2273, opened 2021-06-10
Enable fulltest for PRs.

winder:
## Summary

Just checking what sort of impact this has for the build times.

## Test Plan

CI

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -56,7 +56,7 @@ jobs:
           packages:
             - awscli
       script:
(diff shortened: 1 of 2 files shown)


onetechnical:
I believe this will currently fail due to the catchup testing issue, but am interested to see the results.

tsachiherman:
I do want to have this - but we need to counter balance that with execution time: The goal is to complete the PR tests sufficiently fast. As long as we can achieve that, I don't mind running more tests.

PR #2276, opened 2021-06-10
Make account endpoint produce deterministic JSON

jasonpaulos:
## Summary

The JSON return value of the `/v2/accounts/{addr}` endpoint represents an account with several arrays which are populated from Go maps (such as held assets, created apps, etc.). Currently these arrays are unordered, meaning every time you query this endpoint, you'll likely receive a response with a different ordering of these arrays. This makes SDK testing more difficult than it needs to be, since different responses cannot be directly compared to each other.

This PR sorts the slices in `generated.Account` so that the `/v2/accounts/{addr}` endpoint will always produce arrays with deterministic order.

I don't believe any change to our API spec is necessary, since the object being returned still adheres to the spec, but now the implementation does not randomize the order of arrays on every request.

## Test Plan

Updated unit tests to verify order of arrays in `generated.Account`.

### DIFF
--- a/daemon/algod/api/server/v2/account.go
+++ b/daemon/algod/api/server/v2/account.go
@@ -19,6 +19,7 @@ package v2
 import (
 	"encoding/base64"
 	"errors"
+	"sort"
 
 	"github.com/algorand/go-algorand/crypto"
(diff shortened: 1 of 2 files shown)


algorandskiy (approved):
Thank you for the fix, I've hit this problem recently

PR #2278, opened 2021-06-11
Minor typo fix

fabrice102:
## Summary

Remove extraneous \` in two lines

## Test Plan

Just a change in documentation.


tsachiherman (changes requested):
The changes here are ok, but in the wrong place : 
you'll need to update these in `doc.go`, and regenerate the `TEAL_opcodes.md`.

jannotti:
These files are autogenerated.  I'll make the fix in the input files.

PR #2284, opened 2021-06-15
Add ledger coverage for onlineacct roundlru txtail

nicholasguoalgorand:
## Summary

Add tests to increase coverage for onlineacct roundlru txtail

## Test Plan

These are tests

### DIFF
--- a/ledger/onlineacct_test.go
+++ b/ledger/onlineacct_test.go
@@ -126,12 +126,13 @@ func TestOnlineTopHeap_Pop(t *testing.T) {
 		},
 	}
 
-	acct0 := h.accts[0]
-	acct1 := h.accts[1]
+	originalAccounts := h.accts
 
 	h.Pop()
 
 	require.Equal(t, 2, h.Len())
-	require.Equal(t, acct0, h.accts[0])
-	require.Equal(t, acct1, h.accts[1])
+	require.Equal(t, 3, cap(h.accts))
+	require.Equal(t, originalAccounts[0], h.accts[0])
+	require.Equal(t, originalAccounts[1], h.accts[1])
+	require.Nil(t, originalAccounts[2])
 }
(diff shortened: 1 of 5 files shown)


ledger/onlineacct_test.go
```diff
@@ -0,0 +1,137 @@
...
+
+	h.Push(acct2)
+
+	require.Equal(t, 3, h.Len())
+	require.Equal(t, acct0, h.accts[0])
+	require.Equal(t, acct1, h.accts[1])
+	require.Equal(t, acct2, h.accts[2])
+}
+
+func TestOnlineTopHeap_Pop(t *testing.T) {
```
tsachiherman, line 111: The `Pop` method has a bug.. which isn't covered by this unit test:
Resizing the slice to remove the entry is correct, however, you would also want to set that entry to nil, so that the GC could release the memory if needed.

specifically on this unit test, you want to check that -
```golang
originalAccounts := h.accts
...
require.Equal(t, 3, cap(h.accts))
require.Nil(t, originalAccounts[2])
```

ledger/roundlru.go
```diff
@@ -121,7 +121,7 @@ func (hlc *heapLRUCache) inc() {
 	}
 }
 func (hlc *heapLRUCache) reIndex() {
-	if hlc.entries.heap != nil || len(hlc.entries.heap) == 0 {
+	if hlc.entries.heap == nil || len(hlc.entries.heap) == 0 {
```
tsachiherman, line 124: that's a good find. However, I think that you can simplify it to 
```golang
if len(hlc.entries.heap) == 0 {
```
since `len(nil)` is zero.

tsachiherman (changes requested):
Looks good; The is a small GC bug that need to be addressed, but great otherwise.

tsachiherman (approved):
looks great.

PR #2285, opened 2021-06-15
Fixes to peer selector and TestBasicCatchpointCatchup

algonautshant:
## Summary
peerSelector.go: various bug fixes
- introduce peerSelectorPeer to wrap the network.Peer and add peerClass information, to be able to distinguish between peers of the same address but different classes.
- keep track of download failures to be able to exponentially increase the cost of each failure when failing more than succeeding. This is to evict the peer faster when constantly failing to download.
- initialize rankSum and rankSamples to initialRank of the class. Otherwise, the peer rank will have a very long warmup time before relfecting the correct rank.
- let resetRequestPenalty bound the rank within the class bounds. Otherwise, the penalty calculation pushes the rank out of the class bounds (bug).
- getNextPeer, rankPeer, and peerDownloadDurationToRank are local to the package, since they are using non-exported peerSelectorPeer
- getNextPeer, PeerDownloadDurationToRank and RankPeer use peerSelectorPeer instead of network.Peer
- refreshAvailablePeers distinguishes between peers with the same address but of different peer class
- findPeer returns the peer given the address and the peer class (instead of just the address)

catchpointCatchup_test.go:
- Remove comment about giving the second node all the stake, since it is not the case here.
- Use the round from the catchpoint instead of guessing the round as 36. In case the following catchpoint was obtained due to race conditions, checking for round 37 will be trivial, since it will also be obtained from the catchpoint.

catchpointService.go and service.go:
- Update the code to use peerSelectorPeer instead of network.Peer with peerSelector

peerSelector_test.go:
- Add new tests to check the peerSelector fixes in this PR
- Update the tests to use peerSelectorPeer instead of network.Peer with peerSelector
- Cleanup debugging printouts.

## Test Plan
Added tests to confirm the fixes. 
- TestClassUpperBound
- TestClassLowerBound
- TestEvictionAndUpgrade

### DIFF
--- a/catchup/peerSelector.go
+++ b/catchup/peerSelector.go
@@ -65,6 +65,13 @@ const (
 
 var errPeerSelectorNoPeerPoolsAvailable = errors.New("no peer pools available")
 
+// peerSelectorPeer is a peer wrapper, adding to it the peerClass information
+// There can be multiple peer elements with the same address but different peer classes, and it is important to distinguish between them.
+type peerSelectorPeer struct {
+	network.Peer
+	peerClass network.PeerOption
+}
+
 // peerClass defines the type of peer we want to have in a particular "class",
 // and define the network.PeerOption that would be used to retrieve that type of
 // peer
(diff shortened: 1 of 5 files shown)


catchup/peerSelector.go
```diff
@@ -259,10 +277,10 @@ func makePeerSelector(net peersRetriever, initialPeersClasses []peerClass) *peer
 	return selector
 }
 
-// GetNextPeer returns the next peer. It randomally selects a peer from a pool that has
+// getNextPeer returns the next peer. It randomally selects a peer from a pool that has
 // the lowest rank value. Given that the peers are grouped by their ranks, allow us to
 // prioritize peers based on their class and/or performance.
-func (ps *peerSelector) GetNextPeer() (peer network.Peer, err error) {
+func (ps *peerSelector) getNextPeer() (psp *peerSelectorPeer, err error) {
```
tsachiherman, line 283: I think that my preference here would to keep it with lower-case, as it's not being exported outside the scope of this package. ( we have plenty of components that do.. and it's a good marker for when we need to be more careful around interfaces )

tsachiherman (approved):

winder (commented):
The changes look good. But I wonder, is there a missing test for the case when a peer exists in multiple classes?

algonautshant:
Yes. catchup/peerSelector_test.go line 431. 
The peer b2 is in two classes.

PR #2290, opened 2021-06-15
Fix darwin-arm64 builds

grakshith:
## Summary

Fix `go-algorand` dev builds for darwin-arm64 machines (Apple's M1 CPUs).

## Test Plan

Verified that compilation succeeds for darwin-arm64 machines.


brianolson (approved):
Sadly I don't have my own M1 Mac to verify on, but this all looks plausible and it didn't break anything else.

tsachiherman (changes requested):
Excellent first PR. It clearly would require some additional work before we can claim M1 support, but this change is a great start.

I'm going to block this one until we're done with the 2.7.0 release, and would merge it in after.

PR #2295, opened 2021-06-16
Allow fee to be below minfee, if given explicitly.

jannotti:
This makes other txns accept explicitly low fees.  Sorry for the code
duplication, but I did not want to change libgoal's existing behavior
where it increases fee to minfee.


jasonpaulos (approved):
Looks good. We should probably refactor the fee handling to be more consistent at some point, but functionally I see no issue with this.

jdtzmn (approved):

PR #2300, opened 2021-06-16
txnsync: optimize selectPendingTransactions loop

tsachiherman:
## Summary

The txsync profiler shown that the selectPendingTransactions takes longer-then-desired computation time. To improve that, this PR adds the following:
1. add tracking of the lastTransactionGroupCounter per each set of request params and bloom filter. Together, these would prevent redundant scanning of the pending transactions, and would allow controlled number of iterations when scanning for previously-missed-transactions.
2. extract the loop over the bloom filters outside of the pending transactions loop, so that we need to iterate on these only once, and extract only the bloom filter indices that are relevant.
3. fix a bug in the `decodeBloomFilter` which failed to update the encoding params.

## Test Plan

- [x] Tested using the emulator
- [x] Tested using S1 deployed network
- [x] Bandwidth testing

### DIFF
--- a/txnsync/peer.go
+++ b/txnsync/peer.go
@@ -17,6 +17,7 @@
 package txnsync
 
 import (
+	"fmt"
 	"sort"
 	"time"
 
@@ -144,7 +145,7 @@ type Peer struct {
 
 	// lastTransactionSelectionGroupCounter is the last transaction group counter that we've evaluated on the selectPendingTransactions method.
 	// it used to ensure that on subsequent calls, we won't need to scan the entire pending transactions array from the begining.
-	lastTransactionSelectionGroupCounter uint64
+	lastTransactionSelectionTracker transactionGroupCounterTracker
 
 	// nextStateTimestamp indicates the next timestamp where the peer state would need to be changed.
 	// it used to allow sending partial message while retaining the "next-beta time", or, in the case of outgoing relays,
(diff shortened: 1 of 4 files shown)


txnsync/peer.go
```diff
@@ -167,6 +171,82 @@ type Peer struct {
...
+	firstGroupCounter := (*t)[i].groupCounters[0]
+	copy((*t)[i].groupCounters[0:], (*t)[i].groupCounters[1:])
+	(*t)[i].groupCounters[len((*t)[i].groupCounters)-1] = firstGroupCounter
+}
+
+// index is a helper method for the transactionGroupCounterTracker, helping to locate the index of
+// a requestParamsGroupCounterState in the array that matches the provided request params. The method
+// uses a linear search, which works best against small arrays.
+func (t *transactionGroupCounterTracker) index(offset, modulator byte) int {
+	for i := range *t {
```
brianolson, line 242: nit: this loop would be more concise by using the value half of the range iterator, `i, v := range *t`

brianolson (approved):
compared to the previous non-saturated run, NPN tx bytes are down a bit.
One language quibble. Otherwise looks good.

txnsync/peer.go
tsachiherman: yes, good idea. will do!

PR #2301, opened 2021-06-16
Creator access and app_params_get

jannotti:
Adds the ability to get the creator of apps and assets in teal.

This is a simple new field for `asset_params_get`, but it introduces a new opcode `app_params_get` to enable access for apps.  Meanwhile `app_params_get` allows access to other global parameters about an app - the schema sizes, extra pages, and the programs themselves.

This also begins work on LogicVersion=5, which should be AVM 1.0


jasonpaulos (approved):
This looks good to me

(Does an approval for a draft PR even mean anything? I guess I'll find out)

jdtzmn (approved):
Looks good to me. Thank you for adding `t.Parallel()` to many of the `evalStateful_test` tests!

jasonpaulos (dismissed):

data/transactions/logic/eval.go
```diff
@@ -1755,6 +1758,37 @@ func (cx *evalContext) assetParamsEnumToValue(params *basics.AssetParams, field
 	return
 }
 
+func (cx *evalContext) appParamsEnumToValue(params *basics.AppParams, creator basics.Address, field uint64) (sv stackValue, err error) {
+	switch AppParamsField(field) {
+	case AppApprovalProgram:
+		sv.Bytes = params.ApprovalProgram[:]
+	case AppClearStateProgram:
+		sv.Bytes = params.ApprovalProgram[:]
```
algorandskiy, line 1766: shouldn't copy? If I remember right all txn fields used to be copied when exposed to TEAL

data/transactions/logic/evalStateful_test.go
```diff
@@ -1081,7 +1072,7 @@ byte 0x414c474f
 	testApp(t, text, now, "no such app")
 
 	// create the app and check the value from ApplicationArgs[0] (protocol.PaymentTx) does not exist
-	ledger.newApp(txn.Txn.Sender, 100, makeSchemas(0, 0, 0, 1))
+	ledger.newApp(txn.Txn.Sender, 100, basics.AppParams{})
```
algorandskiy, line 1075: NumByteSlices used to be 1 and it is 0 now, why is it?

algorandskiy (commented):
Looks good, couple questions/comments below

data/transactions/logic/eval.go
jasonpaulos: Also the clear state program is returning the approval program.

jannotti: Fixed the Approval/Clear thing. We don't copy because the rule is that TEAL is not allowed to modify byteslices in place. assetParamsEnumToValue, for example, does things this way.

(You can change the sv.Bytes in an opcode, you just can't modify, say, sv.Bytes[2])

data/transactions/logic/evalStateful_test.go
jannotti: The NumByteSlices was never used to check correctness of puts in these tests.  I only added it recently to unit test min_balance.  It definitely wasn't needed here.

While it might be worth making the testLedger check such things, I'm more interested in getting rid of the testLedger, and instead construct a "real" logicLedger, using a mock Balances object.  It would be more realistic and help me test the txn creation code.

data/transactions/logic/evalStateful_test.go
```diff
@@ -1416,6 +1413,20 @@ intc_1
...
+	t.Parallel()
+	ep, ledger := makeSampleEnv()
+	ledger.newAccount(ep.Txn.Txn.Sender, 1)
+	ledger.newApp(ep.Txn.Txn.Sender, 100, basics.AppParams{})
+
+	/* app id is in ForeignApps, but does not exist */
+	source := "int 56; app_params_get AppExtraProgramPages; int 0; ==; assert; int 0; =="
+	testApp(t, source, ep)
+	/* app id is in ForeignApps, but has zero ExtraProgramPages */
+	source = "int 100; app_params_get AppExtraProgramPages; int 1; ==; assert; int 0; =="
```
algorandskiy, line 1426: could you pleases add a test for `app_params_get` of different approval and clear state programs? Just because Jason spotted it and this would prevent further copy/paste errors on these fields in future

algorandskiy (dismissed):

jannotti:
I'll do this as an integration test.  I really don't want to bother adding
more functionality to the testLedger to hold program source code.  We've
got at least three different mock ledger-like things and the combined code
for them is getting out of hand, and doesn't *really* test much.  If I can
hook the real logicLedger type up to an underlying mock cow underneath, I
think it makes sense as a unit test as well.

data/transactions/logic/evalStateful_test.go
jannotti: Ok, all params are tested in e2e subs test now. (teal-app-params.sh)

jdtzmn (approved):

algorandskiy (dismissed):
Thank you for adding these tests!

test/scripts/e2e_subs/tealprogs/app-params.teal
```diff
@@ -0,0 +1,78 @@
...
+        // Check the creator is the address provided
+        int 0
+        app_params_get AppCreator
+        assert
+        txn ApplicationArgs 0
+        ==
+        assert
+
+
+        // Check this program is version 5
```
jasonpaulos, line 15: This is very nice

jasonpaulos (approved):

algorandskiy (dismissed):

jasonpaulos (approved):

PR #2302, opened 2021-06-17
testing: fix random failure in TestAppEmptyAccountsLocal

tsachiherman:
## Summary

The test had two unrelated bugs:
1. We need to call `WaitForCommit` before `reloadLedger` to ensure the block is being written to disk before the blockQ getting reinitialized ( and loose its content ).
2. The calculation of the total rewards unit in `makeNewEmptyBlock` was wrong. I corrected it. For tests that run only one or two rounds, this might be good enough, but for long-running tests, it would start fail pretty quickly.

## Test Plan

Run using travis to confirm fix.
I was able to repro this locally by running 500 times consecutively. After the fix, I was no longer able to reproduce the failure.

### DIFF
--- a/ledger/applications_test.go
+++ b/ledger/applications_test.go
@@ -926,7 +926,7 @@ return`
 		Header:                   txHeader,
 		ApplicationCallTxnFields: appCallFields,
 	}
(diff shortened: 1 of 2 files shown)


algorandskiy (approved):
Thank you for fixing this. I wonder how's `makeNewEmptyBlock` would even work on non-empty ledgers.

PR #2308, opened 2021-06-18
Specify truncated division is used in TEAL

fabrice102:
## Summary

Specify that the truncated division is used (https://en.wikipedia.org/wiki/Modulo_operation#Variants_of_the_definition)

## Test Plan

Just a minor change in documentation.

### DIFF
--- a/data/transactions/logic/README.md
+++ b/data/transactions/logic/README.md
@@ -80,7 +80,7 @@ An application transaction must indicate the action to be taken following the ex
 
 Most operations work with only one type of argument, uint64 or bytes, and panic if the wrong type value is on the stack.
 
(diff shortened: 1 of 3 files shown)


tsachiherman:
is the "truncated division" here is to clarify that the result of the division is not rounded up to the closest integer value ? This is usually the default division mode, where-as rounding to the nearest integer would require some explicit remarks.

But I don't think that clarifying this is wrong in any way.

One thing to note is that this PR would need to include update to the spec file and/or README files as well.

fabrice102:
Yes, it's to clarify that the result of the division is truncated, which in TEAL case is just floored (since all integers are unsigned).

It's the default for most programming languages, but I think it's a good idea to have it written explicitly.

I'm always confused about which files are auto-generated and which files need to be manually updated, and of the process to run the auto-generation. But yes, specs will need to be updated afterwards.

jannotti (approved):
I've added the spec changes. (There are a couple extra changes from typos that have been fixed as well.)

Issue #2131, opened 2021-05-05
Building from source is a hodgepodge compared to standard Go apps


braderhart:
## Summary

This all started when I just wanted to get started using the CLI tools, and the instructions on the website suggested that I run a script on my computer. After following the script, it was just too convoluted for me to reliably trust that it would even be a good idea. I have built multiple Go applications from source, including those that use C libraries. 

When building go-algorand from source, it is currently expected that people run a Makefile that is chained to multiple Bash scripts within Bash scripts. This is contrary to standard Go packages that should be as simple as running `go install ./cmd/goal`. I personally think it has just been over-engineered. For releases, there are examples within the [Hugo repo](https://github.com/gohugoio/hugo/blob/master/goreleaser.yml) on ways to simplify.

Documentation often is lot more useful than putting everything into a script. Expecting users wanting to build from source to run a bunch of foreign code that is a pain to follow in order eventually run a simple `go build/install` should be concerning. As a developer approaching any project like this I have to ask myself why are things so obfuscated from standard practices, and is it really necessary?

There is also a concern about how you are implementing `libsodium`. Currently, the patched version is committed directly to the repo. This makes it difficult to see the changes from upstream, and clearly it isn't getting updates from upstream either. I saw closed tickets #20 #190 and #293. Ideally you'd use Go's standard crypto/ed25519 and so maybe there could be more work to see if you can optimize the code and submit issues/requests to the upstream project about any performance issues. This could also be highly dependent on the architecture and OS that the benchmarking was done on as well, so maybe more tests would be warranted in multiple environments.

Ideally, if there is a need to continue to use libsodium indefinitely you'd track your changes using a submodule and then create a patch file to continually test against upstream's latest release. Currently, we're just expected to trust your fork of libsodium which no on in their right mind would do.

## Scope/Requirements

- [ ] Simplify build to use standard Go tools.
- [ ] Remove excessive use of Makefile and provide instructions on building without.
- [ ] Cleanup docs and don't ask developers to blindly run an inception of scripts to setup environments. Running any script blindly is bad security practice.
- [ ] Update libsodium to upstream release and use submodule or patch for libsodium source during install.
- [ ] Revisit implementing Go's standard crypto/ed25519 package and submit track any issues upstream.
- [ ] Implement VRF in Go module to remove CGO requirement.
- [ ] Consider consolidating CLI tools into subcommands to reduce complexity.

## Urgency/Relative Priority
[...]

PR #2310, opened 2021-06-19
vrf: Add a pure Go VRF implementation

tmc:
## Summary

This proposes a pure Go implementation of the VRF (Verifiable Random Functions) primitives used in Algorand.

The goal of this change is to demonstrate the viability of a pure Go implementation of the VRF primitives in Algorand.

It goes without saying that any change to the cryptographic primitives in a project must undergo significant scrutiny. I intend this changeset to be a kicking off point for further discussion.

To most closely match the existing libsodium implementation there are a few design decisions that may differ in a final implementation (better error value propagation in particular). 

## Test Plan

Included in this PR are a number of unit tests and benchmarks that compare the output of the two implementations.

This additional [test case](https://github.com/tmc/go-algorand/blob/purego-vrf-additional-tests/crypto/vrf_go_comparisons_test.go#L11) provides wide value coverage in terms of inputs to build confidence in the accuracy of the Go implementation.

## Initial Performance Testing Results

```shell
$ benchstat results
name            time/op
ProveBytes-2    283µs ± 3%
ProveBytesGo-2  257µs ± 0%
VrfVerify-2     345µs ± 1%
VrfVerifyGo-2   333µs ± 1%
```

These results were compiled from many runs on a GCE e2-medium (2 vCPUs, 4 GB memory) which reports as: `cpu: Intel(R) Xeon(R) CPU @ 2.30GHz`

Raw benchmark results are available [here](https://gist.github.com/tmc/f1364bcee04d1192278dd2f297cf116f).

My initial interpretation of these results is that it is viable from a performance perspective to implement the core VRF algorithms in pure Go. Having this implemented directly in Go carries a number of benefits, some of which are highlighted in #2131 

I'd like feedback from the core team on this proposed change and we can map out what a reasonable pathway regarding further testing and eventual inclusion might be.

I want to thank and acknowledge @FiloSottile for guidance in using the (awesome) edwards25519 package.


crypto/vrf_go.go
```diff
@@ -0,0 +1,465 @@
...
+	h.Write(seed[:])
+	hSum := h.Sum(nil)
+	copy(sk[:], hSum[:32])
+	sk[0] &= 248
+	sk[31] &= 127
+	sk[31] |= 64
+	p := edwards25519.NewScalar()
+	skBytes := make([]byte, 64)
+	copy(skBytes, sk[:32])
+	p.SetUniformBytes(skBytes)
```
FiloSottile, line 46: You can replace lines 39-46 with a call to `SetBytesWithClamping(hSum[:32])`.

crypto/vrf_go.go
```diff
@@ -0,0 +1,465 @@
...
+ */
+func vrfVerifyAndHash(pk []byte, proof []byte, msg []byte) ([]byte, error) {
+	Y := &edwards25519.Point{}
+	// validate key
+	if _, err := Y.SetBytes(pk); err != nil {
+		return nil, err
+	}
+	isSmallOrder := (&edwards25519.Point{}).MultByCofactor(Y).Equal(edwards25519.NewIdentityPoint()) == 1
+	if isSmallOrder {
+		return nil, fmt.Errorf("expected key to have small order")
```
FiloSottile, line 81: Other way around in the error text.

```suggestion
		return nil, fmt.Errorf("public key is a small order point")
```

crypto/vrf_go.go
```diff
@@ -0,0 +1,465 @@
...
+ */
+func (sk VrfPrivkey) expand() (*edwards25519.Point, *edwards25519.Scalar, []byte, error) {
+	var tmp [64]byte
+	h := sha512.New()
+	h.Write(sk[:32])
+	hSum := h.Sum(nil)
+	copy(tmp[:], hSum[:64])
+	xScaler := edwards25519.NewScalar()
+	tmpBytes := make([]byte, 64)
+	copy(tmpBytes, tmp[:32])
```
FiloSottile, line 108: `tmpBytes` is unused.

crypto/vrf_go.go
```diff
@@ -0,0 +1,465 @@
...
+ * into the public point Y, the private saclar x, and truncated hash of the
+ * seed to be used later in nonce generation.
+ * Return 0 on success, -1 on failure decoding the public point Y.
+ */
+func (sk VrfPrivkey) expand() (*edwards25519.Point, *edwards25519.Scalar, []byte, error) {
+	var tmp [64]byte
+	h := sha512.New()
+	h.Write(sk[:32])
+	hSum := h.Sum(nil)
+	copy(tmp[:], hSum[:64])
```
FiloSottile, line 105: Can just use `hSum` instead of `tmp`.

crypto/vrf_go.go
```diff
@@ -0,0 +1,465 @@
...
+	xScaler := edwards25519.NewScalar()
+	tmpBytes := make([]byte, 64)
+	copy(tmpBytes, tmp[:32])
+	xScaler.SetBytesWithClamping(tmp[:32])
+
+	truncatedHashedSKString := tmp[32:]
+	Y := edwards25519.NewIdentityPoint()
+	Y.SetBytes(sk[32:])
```
FiloSottile, line 113: Missing error checking. `sk` might be always valid, but better be safe.

crypto/vrf_go.go
```diff
@@ -0,0 +1,465 @@
...
+ * Constant time in everything except the length of alpha.
+ */
+func pureGoVrfProve(Y *edwards25519.Point, xScalar *edwards25519.Scalar, truncHashedSk []byte, alpha []byte) (VrfProof, error) {
+	var pi VrfProof
+	H, err := vrfHashToCurveElligator225519(Y, alpha)
+	if err != nil {
+		return VrfProof{}, err
+	}
+	Gamma := edwards25519.NewIdentityPoint()
+	Gamma.ScalarMult(xScalar, H)
```
FiloSottile, line 147: You can chain these

```
	Gamma := edwards25519.NewIdentityPoint().ScalarMult(xScalar, H)
```

or even skip the initialization

```
	Gamma := new(edwards25519.Point).ScalarMult(xScalar, H)
```

crypto/vrf_go.go
```diff
@@ -0,0 +1,465 @@
...
+	rr2.Add(rr2, rr2)
+	rr2Bytes := rr2.Bytes()
+	rr2Bytes[0]++
+	rr2.SetBytes(rr2Bytes) // rr2[0]++;
+	rr2.Invert(rr2)        // fe25519_invert(rr2, rr2);
+
+	x = &field.Element{}
+
+	const curve25519A = 486662
+	curve25519AElement := (&field.Element{}).One().Mult32((&field.Element{}).One(), curve25519A)
```
FiloSottile, line 216:
```suggestion
	one := new(field.Element).One()
	curve25519AElement := new(field.Element).Mult32(one, curve25519A)
```

Also, this could be a global, so you only compute it once.

crypto/vrf_go.go
```diff
@@ -0,0 +1,465 @@
...
+	x2.Mult32(x2, curve25519A) // fe25519_mul(x2, x2, curve25519_A);
+	e.Add(x2, e)               // fe25519_add(e, x2, e);
+
+	e = chi25519(e) // chi25519(e, e);
+	s = e.Bytes()   // fe25519_tobytes(s, e);
+
+	eIsMinus1 = int(s[1] & 1) // e_is_minus_1 = s[1] & 1;
+	eIsNotMinus1 := eIsMinus1 ^ 1
+	negx = (&field.Element{}).Set(x)
+	negx.Negate(negx)                               // fe25519_neg(negx, x);
```
FiloSottile, line 237:
```suggestion
	negx := new(field.Element).Negate(x)                               // fe25519_neg(negx, x);
```

crypto/vrf_go.go
```diff
@@ -0,0 +1,465 @@
...
+	copy(str[2+(32*1):], P2.Bytes())
+	copy(str[2+(32*2):], P3.Bytes())
+	copy(str[2+(32*3):], P4.Bytes())
+	h := sha512.New()
+	h.Write(str[:])
+	sum := h.Sum(nil)
+
+	copy(result[:], sum[:16])
+	r := edwards25519.NewScalar()
+	r.SetCanonicalBytes(result)
```
FiloSottile, line 302:
```suggestion
	if _, err := r.SetCanonicalBytes(result); err != nil { panic(err) }
```

This will never happen, but I would rather see it checked anyway rather than leave the scalar at zero.

crypto/vrf_go.go
```diff
@@ -0,0 +1,465 @@
...
+	return r
+
+}
+
+func chi25519(z *field.Element) *field.Element {
```
FiloSottile, line 307: This function _might_ be `field.Element.Invert`.

FiloSottile (commented):
(not an audit, just some library usage notes)

crypto/vrf_go.go
tmc: That would be great! I'd like to eliminate as much of this file as possible.

tmc:
The failure in TestMetricSegment appears unrelated. Pushing some documentation cleanups to trigger a new build.

crypto/vrf_go.go
tmc: This does appear to be the case. Running a longer comparison test.

algorandskiy:
Thank you for the submission. Any modification of crypto primitives is considered as a high risk and requires some extra time for evaluation.

tmc:
> Thank you for the submission. Any modification of crypto primitives is considered as a high risk and requires some extra time for evaluation.

Of course. I wouldn't expect this sort of contribution to be rushed and I understand the sensitivity.

It does raise the general problem of how to safely compare implementations in this sort of context. My use/abuse of go test's benchmark in the linked file is a pretty basic way to compare the implementations.

tmc:
Rebasing and pushing.

tmc:
I rebased this to recent master and vendored a copy of the edwards25519 package.

tsachiherman:
> I rebased this to recent master and vendored a copy of the edwards25519 package.

Thanks. We're currently preparing for a major release, so I would have to keep this on hold until we're ready to complete the review and testing of this PR.

tmc:
> > I rebased this to recent master and vendored a copy of the edwards25519 package.
> 
> Thanks. We're currently preparing for a major release, so I would have to keep this on hold until we're ready to complete the review and testing of this PR.

No worries, I'll continue to rebase and expand testing of this change.

tmc:
I pushed an additional test that allows you to run a predefined number of comparisons between the C and Go implementations (vrf_go_comparisons_test.go).

Example output:

```shell
tmc@vmbp:~/go/src/github.com/algorand/go-algorand/crypto(purego-vrf) $ go test -v -run TestPureGoImplementationComparison -purego-compare
=== RUN   TestPureGoImplementationComparison
    vrf_go_comparisons_test.go:21: running with 10000 iterations
    vrf_go_comparisons_test.go:36: prepped
    vrf_go_comparisons_test.go:44: 1000/10000
    vrf_go_comparisons_test.go:44: 2000/10000
    vrf_go_comparisons_test.go:44: 3000/10000
    vrf_go_comparisons_test.go:44: 4000/10000
    vrf_go_comparisons_test.go:44: 5000/10000
    vrf_go_comparisons_test.go:44: 6000/10000
    vrf_go_comparisons_test.go:44: 7000/10000
    vrf_go_comparisons_test.go:44: 8000/10000
    vrf_go_comparisons_test.go:44: 9000/10000
--- PASS: TestPureGoImplementationComparison (8.55s)
PASS
ok  	github.com/algorand/go-algorand/crypto	9.088s
```

tmc:
Pushed a change that adds a build tag guarded comparison implementation that runs both Verify() implementations and panic if the output differs.

Example compilation.
```shell
tmc@vmbp:~/go/src/github.com/algorand/go-algorand/cmd/algod(purego-vrf) $ go install  -tags compare_purego_implementation
```

tmc:
I'm happy to report that I have run a full catchup on each netowrk (mainnet, testnet, devnet, betanet) running the two implementations side-by-side and panicing on differences.

All nodes are caught up and I've encountered no crashes!

Screen Shot 2021-09-08 at 5 59 46 PM

tmc:
I'm going to rebase this work on top of the current master branch.

tmc:
Rebasing again.

bbroder-algo:
This is a very expansive PR and impressive in its scope.  The distance between the current system and the one this is based on has grown, but I think it still has value from a reference and what-if perspective.  I am going to close this but I suspect we will be looking back to it when we need to revisit areas to accelerate the system or shake off dependencies.

tmc:
Rebased and reopened in #5364

PR #2311, opened 2021-06-20
added benchmark to lruAccounts write function

algonathan:
## Summary
lruAccounts write function benchmark

### DIFF
--- a/ledger/lruaccts_test.go
+++ b/ledger/lruaccts_test.go
@@ -17,6 +17,7 @@
 package ledger
 
 import (
+	"encoding/binary"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


ledger/lruaccts_test.go
```diff
@@ -195,3 +196,31 @@ func TestLRUAccountsOmittedPendingWrites(t *testing.T) {
...
+			addr:        basics.Address(digest),
+			round:       basics.Round(i),
+			rowid:       int64(i),
+			accountData: basics.AccountData{MicroAlgos: basics.MicroAlgos{Raw: microAlgos}},
+		}
+	}
+
+	baseAccts := make([]lruAccounts, b.N)
+	for i := 0; i < b.N; i++ {
+		baseAccts[i].init(logging.TestingLog(b), 10, 5)
```
tsachiherman, line 216: In order to get realistic timings, you'll need to fill this data structure with dummy data.
Testing the performance again an empty datastructure is less than likly to represent a typical load.

algonathan: I see. 
i've changed the coed to generate 5k accounts for the benchmark, and made certain to fill the `baseAccts` with 95k different `persistedAccountData` before running the benchmark.

ledger/lruaccts_test.go
```diff
@@ -195,3 +196,40 @@ func TestLRUAccountsOmittedPendingWrites(t *testing.T) {
...
+			baseAcct.write(account)
+		}
+	}
+}
+
+func generatePersistedAccountData(startRound, endRound int) []persistedAccountData {
+	accounts := make([]persistedAccountData, endRound-startRound)
+	for i := startRound; i < endRound; i++ {
+
+		digest := crypto.Hash([]byte{byte(i)})
```
tsachiherman, line 224: This would create many entries.. but only 256 unique accounts.
I don't believe it was your intent, right ?

tsachiherman: Creating overlapping account between the two groups is fine, of course

algonathan: You are correct, thank you for noticing!

I've decided to avoid overlapping accounts, I hope this isn't an issue.

PR #2312, opened 2021-06-21
testing: improve lruaccts benchmark

algonathan:
## Summary
lruAccounts benchmark update: memory consumption reduced


tsachiherman (changes requested):
looks good - only one comment:
In order to create better coverage of the `write` method, you could replace
```golang
binary.BigEndian.PutUint32(buffer, uint32(i-startRound))
```
instead of 
```golang
binary.BigEndian.PutUint32(buffer, uint32(i))
```

This would create some overlapping, which would exercise the "update" path as well.

tsachiherman (approved):

PR #2313, opened 2021-06-21
REST API: make extra-program-pages and apps-total-extra-pages optional

algorandskiy:
## Summary

Make extra-program-pages and apps-total-extra-pages optional in order to do not exposing them before the protocol switch.

## Test Plan

Existing tests updated

### DIFF
--- a/daemon/algod/api/server/v2/account.go
+++ b/daemon/algod/api/server/v2/account.go
@@ -125,7 +125,7 @@ func AccountDataToAccount(
 		AuthAddr:                    addrOrNil(record.AuthAddr),
 		AppsLocalState:              &appsLocalState,
 		AppsTotalSchema:             &totalAppSchema,
-		AppsTotalExtraPages:         &totalExtraPages,
+		AppsTotalExtraPages:         numOrNil(totalExtraPages),
 	}, nil
 }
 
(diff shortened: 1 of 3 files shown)


daemon/algod/api/server/v2/account.go
```diff
@@ -125,8 +125,13 @@ func AccountDataToAccount(
 		AuthAddr:                    addrOrNil(record.AuthAddr),
 		AppsLocalState:              &appsLocalState,
 		AppsTotalSchema:             &totalAppSchema,
-		AppsTotalExtraPages:         &totalExtraPages,
```
tsachiherman, line 128: you could have just called
```golang
AppsTotalExtraPages:         numOrNil(totalExtraPages),
```

algorandskiy: thanks, did not know about this helper, fixed!

tsachiherman (approved):
looks good.

jasonpaulos (approved):
Looks good

winder:
Is this really something people are asking for?

jasonpaulos:
Not really an answer, but I'm in favor of making all "optional" API fields actually optional, and this is a step in that direction.

winder:
> Not really an answer to the above question, but I'm in favor of making all "optional" API fields actually optional, and this is a step in that direction.

Yes, that's fine, but I would probably accommodate that by marking this field as required in the REST API spec.

algorandskiy:
> Yes, that's fine, but I would probably accommodate that by marking this field as required in the REST API spec.

Please follow indexer discussion. We want to have this field not set before protocol upgrade.

PR #2315, opened 2021-06-21
Restore TestPeriodicSync and TestBasicCatchpointCatchup tests.

onetechnical:
## Summary

These tests were temporarily disabled while we were troubleshooting the pipeline. As we are actively fixing these, we should re-enable these tests.

## Test Plan

N/A, these are tests, and are expected to fail. I don't think they run during a standard PR test, however.

### DIFF
--- a/catchup/service_test.go
+++ b/catchup/service_test.go
@@ -162,7 +162,6 @@ func TestServiceFetchBlocksSameRange(t *testing.T) {
 }
 
 func TestPeriodicSync(t *testing.T) {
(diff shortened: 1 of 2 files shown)


algobarb (approved):
Ok since you mentioned "As we are actively fixing these, we should re-enable these tests."

tsachiherman:
Fyi https://github.com/algorand/go-algorand/pull/2285

egieseke (approved):

PR #2316, opened 2021-06-22
Replace Sortition cpp with Go

barnjamin:
## Summary

Trying to replace another bit of CGo

## Test Plan

go test
go test -bench=.

```
ben@LAPTOP-4EV1BSD6:~/go/src/github.com/algorand/go-algorand/data/committee/sortition$ benchstat old.txt new.txt
name          old time/op  new time/op  delta
Sortition-12  1.70µs ± 0%  1.64µs ± 0%   ~     (p=1.000 n=1+1)
```


barnjamin:
Feel free to just close this, mostly curious if it'd work and how it'd change perf.

barnjamin:
needs more testing and might need to work with logprob instead

PR #2319, opened 2021-06-22
network: reduce unneeded contention around checking the peers connectivity

tsachiherman:
## Summary

Existing `messageHandlerThread` was checking that all the connected peers are properly communicating by examining their recent message timings. This implementation served us well - however, it was executed redundantly. How much redundantly ? 19 times too many every 3 minutes ( and all of them at the **exact** same time.. ).

This PR ensures that all the `messageHandlerThread` shares the same ticker for testing the `checkPeersConnectivity`. This is expected to reduce the pressure on the internal `peersLock`.

## Test Plan

Use existing unit tests.

### DIFF
--- a/network/wsNetwork.go
+++ b/network/wsNetwork.go
@@ -402,6 +402,12 @@ type WebsocketNetwork struct {
 	// that messagesOfInterestEnc does not change once it is set during
 	// network start.
 	messagesOfInterestMu deadlock.Mutex
+
+	// peersConnectivityCheckTicker is the timer for testing that all the connected peers
+	// are still transmitting or receiving information. The channel produced by this ticker
+	// is consumed by any of the messageHandlerThread(s). The ticker itself is created during
+	// Start(), and being shut down when Stop() is called.
+	peersConnectivityCheckTicker *time.Ticker
 }
 
 type broadcastRequest struct {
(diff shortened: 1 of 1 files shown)


algorandskiy (approved):

network/wsNetwork.go
```diff
@@ -1126,10 +1144,9 @@ func (wn *WebsocketNetwork) ServeHTTP(response http.ResponseWriter, request *htt
 	incomingPeers.Set(float64(wn.numIncomingPeers()), nil)
 }
 
-func (wn *WebsocketNetwork) messageHandlerThread() {
+func (wn *WebsocketNetwork) messageHandlerThread(peersConnectivityCheckCh <-chan time.Time) {
 	defer wn.wg.Done()
-	inactivityCheckTicker := time.NewTicker(connectionActivityMonitorInterval)
-	defer inactivityCheckTicker.Stop()
```
winder, line 1132: Previously each handler would be checked every 3 minutes, now a random handler is checked every 3 minutes?

Is there a downside to having inactive connections remain on the peer list?

tsachiherman: The downside is that they consume memory, and need to be cleaned out.. eventually.

PR #2322, opened 2021-06-23
Improve TestMetricSegment test reliability

tsachiherman:
## Summary

Improve TestMetricSegment test reliability by repeating the test with incrementing time delays.
This implementation would allow faster execution on faster platforms, and allow fallback for slower platforms.

## Test Plan

Tested locally as well as on travis CI.


algonautshant (approved):
Looks great.

PR #2324, opened 2021-06-23
testing: move travis_retry to skip rebuilding

winder:
## Summary

Remove the top-level retry command. This will allow build / lint failures to terminate the build with no retry, and may speed up test failure retries.

## Test Plan

CI

### DIFF
--- a/scripts/travis/travis_retry.sh
+++ b/scripts/travis/travis_retry.sh
@@ -2,6 +2,9 @@
 
 # https://github.com/travis-ci/travis-build/blob/master/lib/travis/build/bash/travis_retry.bash
 
+ANSI_RED="\033[31;1m"
+ANSI_RESET="\033[0m"
+
 travis_retry() {
   local result=0
   local count=1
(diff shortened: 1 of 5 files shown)


scripts/travis/travis_retry.sh
```diff
@@ -0,0 +1,27 @@
...
+  done
+
+  [[ "${count}" -gt 3 ]] && {
+    echo -e "\\n${ANSI_RED}The command \"${*}\" failed 3 times.${ANSI_RESET}\\n" >&2
+  }
+
+  return "${result}"
+}
+
+travis_retry $@
```
brianolson, line 27: shouldn't this be `"$@"` ?
https://tiswww.case.edu/php/chet/bash/bashref.html#Special-Parameters
`"$@" is equivalent to "$1" "$2"`

tsachiherman (changes requested):
I think that you'll need to remove the travis_wait from build_test.sh, since it would block all the output from the 3 attempts - and we don't want that. i.e. look at https://travis-ci.com/github/algorand/go-algorand/jobs/518395939

tsachiherman (approved):

PR #2329, opened 2021-06-24
Lruaccts write speedup

algonathan:
## Summary
Optimized lruAccounts write function. gains: from ~260ns to ~200ns per write op


ledger/persistedaccts_list.go
```diff
@@ -0,0 +1,127 @@
+package ledger
+
+// persistedAccountDataList represents a doubly linked list.
+// must initiate with newPersistedAccountList.
+type persistedAccountDataList struct {
+	root     persistedAccountDataListNode // sentinel list element, only &root, root.prev, and root.next are used
+	len      int                          // current list length excluding (this) sentinel element
+	freeList *persistedAccountDataList    // preallocated nodes location
```
tsachiherman, line 8: I think that having a simple `* persistedAccountDataListNode` is going to be more performant here.

ledger/persistedaccts_list.go
```diff
@@ -0,0 +1,127 @@
...
+type persistedAccountDataListNode struct {
+	// Next and previous pointers in the doubly-linked list of elements.
+	// To simplify the implementation, internally a list l is implemented
+	// as a ring, such that &l.root is both the next element of the last
+	// list element (l.Back()) and the previous element of the first list
+	// element (l.Front()).
+	next, prev *persistedAccountDataListNode
+
+	// The list to which this element belongs. (helps removing items fast)
+	list *persistedAccountDataList
```
tsachiherman, line 20: maintaining a `list *persistedAccountDataList` doesn't get you much - other than validation. 
I think you'd be better off without that extra validation - since it's not really an issue anyway for a correctly written code.

ledger/persistedaccts_list.go
```diff
@@ -0,0 +1,127 @@
...
+	l.len = 0
+	return l
+}
+
+func (l *persistedAccountDataList) allocateFreeNodes(numAllocs int) *persistedAccountDataList {
+	if l.freeList == nil {
+		return l
+	}
+	for i := 0; i < numAllocs; i++ {
+		l.freeList.pushFront(&persistedAccountData{})
```
tsachiherman, line 44: for all the entries in the freelist, it would be best to keep the Value as `nil`. ( when placed back on to the freelist, it should be reset to nil as well )

ledger/persistedaccts_list.go
```diff
@@ -0,0 +1,127 @@
+package ledger
+
+// persistedAccountDataList represents a doubly linked list.
+// must initiate with newPersistedAccountList.
+type persistedAccountDataList struct {
+	root     persistedAccountDataListNode // sentinel list element, only &root, root.prev, and root.next are used
+	len      int                          // current list length excluding (this) sentinel element
```
tsachiherman, line 7: I can't see any real good use of `len`. The only "usage" of it is to determine if we have any entries in the list, and we can do that by comparing `root.next == root.prev`

ledger/persistedaccts_list.go
algonathan: i've changed it, though i haven't notice a change in performance.

ledger/persistedaccts_list.go
algonathan: i've removed the len property. The inspection for 0 length is now: `list.root.next == &list.root`.

tsachiherman:
I think that it looks good; Could you please add the gains to the PR description ?

tsachiherman:
@algorandskiy could you review this one as well ?

ledger/persistedaccts_list_test.go
```diff
@@ -0,0 +1,208 @@
...
+	l := newPersistedAccountList()
+	e1 := l.pushFront(&persistedAccountData{addr: basics.Address{1}})
+	e2 := l.pushFront(&persistedAccountData{addr: basics.Address{2}})
+	e3 := l.pushFront(&persistedAccountData{addr: basics.Address{3}})
+	checkListPointers(t, l, []*persistedAccountDataListNode{e3, e2, e1})
+
+	l.remove(e2)
+	checkListPointers(t, l, []*persistedAccountDataListNode{e3, e1})
+	l.remove(e3)
+	checkListPointers(t, l, []*persistedAccountDataListNode{e1})
```
algorandskiy, line 46: could add a test for `move` method:
 - move to front/back at single element list
 - swap on two elements list

algorandskiy (commented):
Looks good, couple comments

algorandskiy (dismissed):
Thank you for the fixes. Please fix licence check and it is ready to go.

tsachiherman (approved):

PR #2334, opened 2021-06-28
Add initial devMode support

tsachiherman:
## Summary

Add new local private network mode - devmode. devMode allows the developer to deploy a single node network, where every transaction being sent to the node automatically generates a new block.

This feature is focused primarily around 3rd parties that want to test their solution on the Algorand platform, without waiting for the network to make progress.

## Test Plan

Add e2e test to test changes.

### DIFF
--- a/data/bookkeeping/genesis.go
+++ b/data/bookkeeping/genesis.go
@@ -71,8 +71,9 @@ type Genesis struct {
 
 	// DevMode defines whether this network operates in a developer mode or not. Developer mode networks
 	// are a single node network, that operates without the agreement service being active. In liue of the
-	// agreement service, a new block is generated each time a node receives a block. The default value for
-	// this field is "false", which makes this field empty from it's encoding, and therefore safe.
+	// agreement service, a new block is generated each time a node receives a transaction group. The
+	// default value for this field is "false", which makes this field empty from it's encoding, and
+	// therefore backward compatible.
 	DevMode bool `codec:"devmode"`
 }
 
--- a/data/bookkeeping/genesis.go
+++ b/data/bookkeeping/genesis.go
@@ -68,6 +68,12 @@ type Genesis struct {
 
 	// Arbitrary genesis comment string - will be excluded from file if empty
 	Comment string `codec:"comment"`
+
+	// DevMode defines whether this network operates in a developer mode or not. Developer mode networks
+	// are a single node network, that operates without the agreement service being active. In liue of the
+	// agreement service, a new block is generated each time a node receives a block. The default value for
+	// this field is "false", which makes this field empty from it's encoding, and therefore safe.
+	DevMode bool `codec:"devmode"`
 }
 
 // LoadGenesisFromFile attempts to load a Genesis structure from a (presumably) genesis.json file.
--- a/node/node.go
+++ b/node/node.go
@@ -474,24 +474,9 @@ func (node *AlgorandFullNode) Ledger() *data.Ledger {
 }
 
 // writeDevmodeBlock generates a new block for a devmode, and write it to the ledger.
-func (node *AlgorandFullNode) writeDevmodeBlock(lastRound basics.Round) (err error) {
-	// get the last block
-	blk, err := node.ledger.Block(lastRound)
-	if err != nil {
-		return
-	}
-
-	// modify the block blk round number, to emulate into-the-future block. The OnNewBlock doesn't
-	// really care about the rest of this data structure anyway, so it would be safe to do.
-	blk.BlockHeader.Round++
-
-	// calling OnNewBlock would trigger recomputation of the txn pool for round blk+1.
-	node.transactionPool.OnNewBlock(blk, ledgercore.StateDelta{})
-
-	// The above was already pregenerating the entire block,
-	// so there won't be any waiting on this call.
+func (node *AlgorandFullNode) writeDevmodeBlock() (err error) {
 	var vb *ledger.ValidatedBlock
-	vb, err = node.transactionPool.AssembleBlock(lastRound+1, time.Now().Add(config.ProposalAssemblyTime))
+	vb, err = node.transactionPool.AssembleDevModeBlock()
 	if err != nil || vb == nil {
 		return
 	}
@@ -503,8 +488,6 @@ func (node *AlgorandFullNode) writeDevmodeBlock(lastRound basics.Round) (err err
 
 // BroadcastSignedTxGroup broadcasts a transaction group that has already been signed.
 func (node *AlgorandFullNode) BroadcastSignedTxGroup(txgroup []transactions.SignedTxn) (err error) {
-	var lastRound basics.Round
-
 	// in developer mode, we need to take a lock, so that each new transaction group would truely
 	// render into a unique block.
 	if node.devMode {
@@ -513,13 +496,13 @@ func (node *AlgorandFullNode) BroadcastSignedTxGroup(txgroup []transactions.Sign
 			// if we added the transaction successfully to the transaction pool, then
 			// attempt to generate a block and write it to the ledger.
 			if err == nil {
-				err = node.writeDevmodeBlock(lastRound)
+				err = node.writeDevmodeBlock()
 			}
 			node.mu.Unlock()
 		}()
 	}
 
-	lastRound = node.ledger.Latest()
+	lastRound := node.ledger.Latest()
 	var b bookkeeping.BlockHeader
 	b, err = node.ledger.BlockHdr(lastRound)
 	if err != nil {
--- a/node/node.go
+++ b/node/node.go
@@ -113,7 +113,7 @@ type AlgorandFullNode struct {
 	rootDir     string
 	genesisID   string
 	genesisHash crypto.Digest
-	devMode     bool // is this node operates in a developer mode ? ( benine agreement, broadcasting transaction generates a new block )
+	devMode     bool // is this node operates in a developer mode ? ( benign agreement, broadcasting transaction generates a new block )
 
 	log logging.Logger
 
--- a/node/node.go
+++ b/node/node.go
@@ -113,6 +113,7 @@ type AlgorandFullNode struct {
 	rootDir     string
 	genesisID   string
 	genesisHash crypto.Digest
+	devMode     bool // is this node operates in a developer mode ? ( benine agreement, broadcasting transaction generates a new block )
 
 	log logging.Logger
 
@@ -164,6 +165,11 @@ func MakeFull(log logging.Logger, rootDir string, cfg config.Local, phonebookAdd
 	node.log = log.With("name", cfg.NetAddress)
 	node.genesisID = genesis.ID()
 	node.genesisHash = crypto.HashObj(genesis)
+	node.devMode = genesis.DevMode
+
+	if node.devMode {
+		cfg.DisableNetworking = true
+	}
 
 	// tie network, block fetcher, and agreement services together
 	p2pNode, err := network.NewWebsocketNetwork(node.log, node.config, phonebookAddresses, genesis.ID(), genesis.Network)
@@ -238,11 +244,16 @@ func MakeFull(log logging.Logger, rootDir string, cfg config.Local, phonebookAdd
 
 	blockValidator := blockValidatorImpl{l: node.ledger, verificationPool: node.highPriorityCryptoVerificationPool}
 	agreementLedger := makeAgreementLedger(node.ledger, node.net)
-
+	var agreementClock timers.Clock
+	if node.devMode {
+		agreementClock = timers.MakeFrozenClock()
+	} else {
+		agreementClock = timers.MakeMonotonicClock(time.Now())
+	}
 	agreementParameters := agreement.Parameters{
 		Logger:         log,
 		Accessor:       crashAccess,
-		Clock:          timers.MakeMonotonicClock(time.Now()),
+		Clock:          agreementClock,
 		Local:          node.config,
 		Network:        gossip.WrapNetwork(node.net, log),
 		Ledger:         agreementLedger,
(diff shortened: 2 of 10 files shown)


node/node.go
```diff
@@ -462,10 +473,55 @@ func (node *AlgorandFullNode) Ledger() *data.Ledger {
...
+	blk.BlockHeader.Round++
+
+	// calling OnNewBlock would trigger recomputation of the txn pool for round blk+1.
+	node.transactionPool.OnNewBlock(blk, ledgercore.StateDelta{})
+
+	// The above was already pregenerating the entire block,
+	// so there won't be any waiting on this call.
+	var vb *ledger.ValidatedBlock
+	vb, err = node.transactionPool.AssembleBlock(lastRound+1, time.Now().Add(config.ProposalAssemblyTime))
+	if err != nil || vb == nil {
```
algorandskiy, line 495: how `vb` can be nill and no error set?

algorandskiy, line 495: nit: maybe call it `NewDevModeBlock` and move to transaction pool in order to make easier to understand (well, subjective).
In this case `BroadcastSignedTxGroup` would call 3 funcs: ledger.Block(), pool.NewDevModeBlock(), ledger. AddValidatedBlock().

algorandskiy (commented):
Nice work! couple comments

data/bookkeeping/genesis.go
```diff
@@ -68,6 +68,13 @@ type Genesis struct {
 
 	// Arbitrary genesis comment string - will be excluded from file if empty
 	Comment string `codec:"comment"`
+
+	// DevMode defines whether this network operates in a developer mode or not. Developer mode networks
+	// are a single node network, that operates without the agreement service being active. In liue of the
+	// agreement service, a new block is generated each time a node receives a transaction group. The
+	// default value for this field is "false", which makes this field empty from it's encoding, and
+	// therefore backward compatible.
+	DevMode bool `codec:"devmode"`
```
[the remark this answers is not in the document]
tsachiherman: it's implied by the
```golang
```

node/node.go
tsachiherman: it cannot be. I added it here just for extra safety measure, since we're going to dereference it.

algorandskiy (dismissed):

jeapostrophe:
This is awesome. The main question I have on it is how the Indexer will deal with this. My assumption is that if there's a timer in it rather than something that just notices the changes, then something will go wrong.

tsachiherman:
The indexer is polling blocks as usual from the node. The indexer would go
to sleep for a short while if no block is available.

So, the indexer might be a bit lagging behind, exactly like it is doing
today.

brianolson:
The Indexer uses the 'status after block' API call which waits until a next block is committed. I'll guess that `algod` internals will do the right thing and still notify when the block commits and not be waiting 5 sec, and Indexer will do the right thing and index the new block when it's available. In my development environment Indexer imports local small blocks in under 50ms, so it shouldn't have any trouble keeping up.

winder:
It would be trivial to enable this in the sandbox, we just need to modify the template: https://github.com/algorand/sandbox/blob/master/images/algod/future_template.json

algorandskiy (approved):

jeapostrophe:
The indexer needs to be updated to parse the new Genesis struct ---
https://github.com/algorand/indexer/blob/ae0a9746f592f0235df045926f4fb5d6fabcd90f/types/types.go#L366 ---
```
{"level":"info","msg":"fetching genesis from algod","time":"2021-07-01T21:38:25Z"}
{"error":"error decoding genesis, json decode error [pos 783]: no matching struct field found when decoding stream map with key devmode","level":"error","msg":": could not load genesis json, error decoding genesis, json decode error [pos 783]: no matching struct field found when decoding stream map with key devmode",
"time":"2021-07-01T21:38:25Z"}
```

tsachiherman:
> The indexer needs to be updated to parse the new Genesis struct ---
> https://github.com/algorand/indexer/blob/ae0a9746f592f0235df045926f4fb5d6fabcd90f/types/types.go#L366 ---
> 
> ```
> {"level":"info","msg":"fetching genesis from algod","time":"2021-07-01T21:38:25Z"}
> {"error":"error decoding genesis, json decode error [pos 783]: no matching struct field found when decoding stream map with key devmode","level":"error","msg":": could not load genesis json, error decoding genesis, json decode error [pos 783]: no matching struct field found when decoding stream map with key devmode",
> "time":"2021-07-01T21:38:25Z"}
> ```

Correct. Will be fixed for the next indexer release.

jeapostrophe:
I think I might be stupid @tsachiherman

I believe that things are working, because the rounds are not progressing. However, this means that the default parameters say that the first valid round is 0, but the JS SDK will not accept a `0` value. (The msgpack format rejects all `0` fields and if you don't include the round, then it will say it must be a positive number.)

My guess is that the simplest and best strategy is to just make a single txn not using the JS SDK so that the rounds are 1. Is that dumb?

tsachiherman:
> I think I might be stupid @tsachiherman
> 
> I believe that things are working, because the rounds are not progressing. However, this means that the default parameters say that the first valid round is 0, but the JS SDK will not accept a `0` value. (The msgpack format rejects all `0` fields and if you don't include the round, then it will say it must be a positive number.)
> 
> My guess is that the simplest and best strategy is to just make a single txn not using the JS SDK so that the rounds are 1. Is that dumb?

It sounds like a localized issue with the JS SDK, that should be fixed there. The workaround you've proposed would work, of course - but I wouldn't feel comfortable calling it "best strategy". 

Note that the indexer would need this field primarily so that it could calculate the genesis hash correctly - and not for any other reason.

jeapostrophe:
I've integrated this into Reach and have some feedback.

Here are some logs --- https://gist.github.com/jeapostrophe/ba0d09d43d866f3bbcbfce3ac8b1e897

I can confirm that the blocks are built immediately, which is pretty cool and very noticeable when I'm only interacting with the algod node.

If you look here --- https://gist.github.com/jeapostrophe/ba0d09d43d866f3bbcbfce3ac8b1e897#file-reach-log-L509-L511 --- you can see that Reach sees an application created in round 5 and then we start polling the indexer for it --- https://gist.github.com/jeapostrophe/ba0d09d43d866f3bbcbfce3ac8b1e897#file-reach-log-L616-L640

The creation time was 2021-07-01T23:25:50.399Z and we started polling at 2021-07-01T23:25:50.439Z but we don't get the confirmation until 2021-07-01T23:26:51.096Z  --- https://gist.github.com/jeapostrophe/ba0d09d43d866f3bbcbfce3ac8b1e897#file-reach-log-L2200 --- that's about a minute.

You can see this polling in the indexer log too --- https://gist.github.com/jeapostrophe/ba0d09d43d866f3bbcbfce3ac8b1e897#file-indexer-log-L18 --- 2021-07-01 23:25:50.4369857 is the first one then the update actually happens --- https://gist.github.com/jeapostrophe/ba0d09d43d866f3bbcbfce3ac8b1e897#file-indexer-log-L78 --- at 2021-07-01T23:26:50Z and the next poll gets the answer

Comparing with the algod log, the txn shows up here --- https://gist.github.com/jeapostrophe/ba0d09d43d866f3bbcbfce3ac8b1e897#file-node-log-L158 --- at 2021-07-01 23:25:50.43117

Then it looks like the node just stalls ---
https://gist.github.com/jeapostrophe/ba0d09d43d866f3bbcbfce3ac8b1e897#file-node-log-L159-L170 --- it is doing some telemetry and catchup stuff and then there's two GETs and it works. I think the one on line 169 is put into the log out of order.

I know that in your commit, you also introduced disabling networking, but I don't believe I can use that, because I am connecting to `algod` via the JS SDK across a Docker network. When I tried it, it failed.

I did some investigating and found this line --- https://github.com/algorand/go-algorand/blob/fc52ab9b886df7a7f76530bbe68e19a4ee08ce52/cmd/algod/main.go#L175 --- and when I set `ALGOTEST`, I got the telemetry to go away, but the hanging is still there.

 It appears to me that it is from the catchup service, but there doesn't appear to be a way to turn it off; seems important :)

Can you give me some advice?

jeapostrophe:
If it useful to know, my config is https://gist.github.com/jeapostrophe/8c0be3ee2f35ce6b353d16f2023c75c4

tsachiherman:
@jeapostrophe ,

Leaving the indexer aside, could you provide some more info why you think that the node is stalled ?
In this devmode, it will make no progress, so it's expected to look as if it's "stalled".
( you can feed it with some dummy transactions to keep it moving forward )

The messages you've mentioned in the node's log file are all expected and are all benign messages : in ideal world, these wouldn't be there, but their existence isn't an indication of anything goes wrong.

As for the disabled networking - it's the disabling of the gossip networking ( the networking that responsible for propagating the votes, proposals and transactions across the network ). The SDKs are using the REST APIs entry point, which isn't considered to be "networking".

The connection to the telemetry could definitely create redundant log entries, but should not be a practical issue. It's expected to run in parallel, so it won't block the execution of the node.

Hope this would help.

jeapostrophe:
Thanks @tsachiherman --- I'm also on the Reach/Algorand Slack if you want to talk high bandwidth.

I think something is wrong because in the workload I'm running, the whole program runs from 2021-07-01T23:25:50.016Z to 2021-07-01T23:29:52.437Z, which is about 4 minutes. When I watch it run, it has about four "stalls" as it is polling the indexer for a long time. They correspond to application creation and then three `appl` txns. 

When I go back to the non-DevMode branch, such as my CI is using, I get output --- https://gist.github.com/jeapostrophe/faca7713f409a3bb4a239ba4788db81d#file-ci-log-L1318 --- where the program runs from 2021-07-01T19:47:44.549Z to 2021-07-01T19:48:55.786Z, which is about 1 minute. That, btw, is on a slow "small" instance cloud machine, not my super powerful dev machine. So, this new DevMode is actually slower than the the old version.

You'd know better than me, but I think that there is some problem with catchup. If you look at this line in the algod log --- https://gist.github.com/jeapostrophe/ba0d09d43d866f3bbcbfce3ac8b1e897#file-node-log-L169 --- you can see that the `GET /v2/status/wait-for-block-after/5` request is received at 2021-07-01 23:25:50.4120134 but it doesn't complete until 2021-07-01T23:26:50.412576Z even though this line --- https://gist.github.com/jeapostrophe/ba0d09d43d866f3bbcbfce3ac8b1e897#file-node-log-L158 --- shows that the application created at 2021-07-01T23:25:50.431464Z, which Reach sees --- https://gist.github.com/jeapostrophe/ba0d09d43d866f3bbcbfce3ac8b1e897#file-reach-log-L476 & https://gist.github.com/jeapostrophe/ba0d09d43d866f3bbcbfce3ac8b1e897#file-reach-log-L511 --- as being confirmed in round 5

--

That's good to know about the logs and networking.

jeapostrophe:
FWIW, I disabled networking and I'm getting the same hang but without the extra log messages. See -- https://gist.github.com/jeapostrophe/4b1422b283d7780385133ad4b038ac42#file-node-log-L140-L141 
```
{"file":"logger.go","function":"github.com/algorand/go-algorand/daemon/algod/api/server/lib/middlewares.(*LoggerMiddleware).handler.func1","level":"info","line":56,"msg":"192.168.0.4:57890 - - [2021-07-02 14:02:46.9561025 +0000 UTC m=+1.533271001] \"GET /v2/applications/4 HTTP/1.1\" 200 804 \"\" 235.5µs","time":"2021-07-02T14:02:46.956349Z"}
{"file":"logger.go","function":"github.com/algorand/go-algorand/daemon/algod/api/server/lib/middlewares.(*LoggerMiddleware).handler.func1","level":"info","line":56,"msg":"127.0.0.1:44518 - - [2021-07-02 14:02:46.9515135 +0000 UTC m=+1.528682101] \"GET /v2/status/wait-for-block-after/5 HTTP/1.1\" 200 567 \"Go-http-client/1.1\" 1m0.000638s","time":"2021-07-02T14:03:46.952229Z"}
```

The first message is a sign that the application is made and that Reach is reading the information about it, the second is the Indexer waiting for that block. Notice that it goes like this:
- 02:46.9515 --- (line 2, req time) Indexer runs `/v2/status/wait-for-block-after/5`
- 02:46.9561 --- (line 1) Block 5 happens and Reach immediately looks up the application information
- 03:46.9522 --- (line 2, resp time) Indexer gets its result

jeapostrophe:
I think the problem is right here: https://github.com/algorand/go-algorand/blob/8c7311cd967b5dee1432ec7bbb970d16c80f949e/daemon/algod/api/server/v2/handlers.go#L327-L328

It assumes that waiting 1 minute is enough time and for some reason `ledger.Wait` isn't beating that case

jeapostrophe:
Ah, the *real* problem is that line 328 should be `case <-ledger.Wait(basics.Round(round)):` even though the function is named "wait-for-block-after", the docs say --- https://developer.algorand.org/docs/reference/rest-apis/algod/v2/#get-v2statuswait-for-block-afterround --- "Gets the node status after waiting for the given round." meaning "/v2/status/wait-for-block/after/5" means "Gets the node status after waiting for round 5"

tsachiherman:
Agree. This looks like the culprit - and it's not related just for devmode. It would also affect the normal flow. The only difference is that in the normal flow, rounds would keep coming and so the enduser would not notice that.

jeapostrophe:
I made the change and now the program takes about 7 seconds: 14:22:59.424 to 14:23:06.052.

Boom! Awesome @tsachiherman ; thank you!

jeapostrophe:
Same problem in v1 --- https://github.com/algorand/go-algorand/blob/master/daemon/algod/api/server/v1/handlers/handlers.go#L614

jeapostrophe:
FYI, I don't plan to submit a PR, I assume that you'll take care of it :) Thanks again!

jeapostrophe:
This is the patch I'm using and testing with... it works great: https://github.com/reach-sh/reach-lang/blob/0_1_3_and_avm/scripts/devnet-algo/reach2.patch

jeapostrophe:
The `WaitForBlock` fix is still not applied in the most recent Algorand release. I created an issue for it.

PR #2335, opened 2021-06-28
WIP: Call ParamsRound() to subtract from nextRound in mainLoop init

cce:
## Summary

I recently reviewed #1896 and #1900 with @zeldovich and @yossigi while discussing changes to the agreement package. While looking at calls to `Ledger.ConsensusVersion()` and  `Ledger.ConsensusParams()`, I noticed that all but one (non-test) call to those functions use the `ParamsRound(r)` helper function to wrap the round input argument. This unwrapped call to `ConsensusVersion(r)` happens in service.go when the the mainLoop initializes a new `player` and sets `FilterTimeout` before starting the main agreement event loop.

```
ga/agreement$ git grep -E '\.Consensus(Params|Version)\(' | grep -v '_test.go:'
bundle.go:	proto, err := l.ConsensusParams(ParamsRound(b.Round))
demux.go:		proto, err := d.ledger.ConsensusVersion(ParamsRound(e.ConsensusRound()))
demux.go:	if proto, err := d.ledger.ConsensusVersion(ParamsRound(currentRound)); err != nil {
demux.go:	proto, err := l.ConsensusVersion(ParamsRound(synthetic.ConsensusRound()))
proposal.go:	cparams, err := ledger.ConsensusParams(ParamsRound(rnd))
proposal.go:	cparams, err := ledger.ConsensusParams(ParamsRound(rnd))
pseudonode.go:	cparams, err := n.ledger.ConsensusParams(ParamsRound(voteRound))
selector.go:	cparams, err := l.ConsensusParams(ParamsRound(r))
service.go:		nextVersion, err := s.Ledger.ConsensusVersion(nextRound) // XXX
vote.go:	proto, err := l.ConsensusParams(ParamsRound(rv.Round))
vote.go:	proto, err := l.ConsensusParams(ParamsRound(rv.Round))
```

My understanding is that `ParamsRound(r)` returns `r-2`, and is used as the parameter for ConsensusVersion/ConsensusParams, so that you don't have to speculate into a future or in-progress round to know if a consensus version has changed.

The recommendation during from that discussion was to try and back out the fix in #1900 in favor of calling `ParamsRound()` and then having a goto-retry loop during initialization to ensure `Ledger.ConsensusVersion()` returns successfully before setting up the player. I tried a hand at implemented this with a simple 1-second wait, feedback appreciated.

This is a corner case, and the value of `FilterTimeout` is overwritten by `player` later, but backing out the version speculation code in `ConsensusVersion()` should make the code less complex and keep it closer in line with the similar/duplicate logic in `ConsensusParams()`.

## Test Plan

Hopefully this doesn't break any tests, but a new test is likely needed based on `TestConsensusVersion` from #1900 (?)

### DIFF
--- a/agreement/service.go
+++ b/agreement/service.go
@@ -23,7 +23,6 @@ import (
 
 	"github.com/algorand/go-algorand/config"
 	"github.com/algorand/go-algorand/logging"
-	"github.com/algorand/go-algorand/protocol"
 	"github.com/algorand/go-algorand/util/db"
 	"github.com/algorand/go-algorand/util/execpool"
 	"github.com/algorand/go-algorand/util/timers"
@@ -198,15 +197,18 @@ func (s *Service) mainLoop(input <-chan externalEvent, output chan<- []action, r
 			s.log.Infof("decode (agreement): restored crash state from database (pending %v @ %+v)", a, status)
 		}
 	}
+
+checkNextRound:
 	// err will tell us if the restore/decode operations above completed successfully or not.
 	if err != nil || status.Round < s.Ledger.NextRound() {
 		// in this case, we don't have fresh and valid state
 		// pretend a new round has just started, and propose a block
 		nextRound := s.Ledger.NextRound()
-		nextVersion, err := s.Ledger.ConsensusVersion(nextRound)
+		nextVersion, err := s.Ledger.ConsensusVersion(ParamsRound(nextRound))
 		if err != nil {
-			s.log.Errorf("unable to retrieve consensus version for round %d, defaulting to binary consensus version", nextRound)
-			nextVersion = protocol.ConsensusCurrentVersion
+			s.log.Errorf("unable to retrieve consensus version for round %d, waiting for successful Ledger.ConsensusVersion", nextRound)
+			time.Sleep(1 * time.Second)
+			goto checkNextRound
 		}
 		status = player{Round: nextRound, Step: soft, Deadline: FilterTimeout(0, nextVersion)}
 		router = makeRootRouter(status)
--- a/data/basics/units.go
+++ b/data/basics/units.go
@@ -121,7 +121,7 @@ func OneTimeIDForRound(round Round, keyDilution uint64) crypto.OneTimeSignatureI
 	}
 }
 
-// SubSaturate subtracts two rounds with saturation arithmetic that does not
+// SubSaturate subtracts x rounds with saturation arithmetic that does not
 // wrap around past zero, and instead returns 0 on underflow.
 func (round Round) SubSaturate(x Round) Round {
 	if round < x {
--- a/data/basics/units.go
+++ b/data/basics/units.go
@@ -121,7 +121,7 @@ func OneTimeIDForRound(round Round, keyDilution uint64) crypto.OneTimeSignatureI
 	}
 }
 
-// SubSaturate subtracts x rounds with saturation arithmetic that does not
+// SubSaturate subtracts two rounds with saturation arithmetic that does not
 // wrap around past zero, and instead returns 0 on underflow.
 func (round Round) SubSaturate(x Round) Round {
 	if round < x {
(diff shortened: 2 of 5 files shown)


tsachiherman (commented):
Few comments -
1. Using the ParamsRound for the nextRound is probably a good idea. I would suggest writing a unit test that create a dummy upgraded protocol, and fail the consensus.
2. I don't think there is a use case where we can't figure out the consensus version for two rounds before. All nodes have the latest 1000 rounds available ( otherwise, these nodes cannot verify the latest 1000 rounds of transaction ids ).
3. This looping, in the mainLoop, as implemented, would cause more harm than good. The service need to be able to execute its main loop; otherwise, the incoming buffers would overflow, and there won't be a way to stop the service gracefully ( i.e. Shutdown ).
4. I can't see why the changes in the `ledger.go` would be an issue : it's only extend the knowledge of the protocol version beyond the present into the future. It would not negatively affect the agreement.

cce:
Re #2 and #3, then would you suggest just calling Fatalf instead? I wasn't sure if this could happen because LedgerReader hadn't caught up or something, I am not well versed in the ledger code.

tsachiherman:
> Re #2 and #3, then would you suggest just calling Fatalf instead? I wasn't sure if this could happen because LedgerReader hadn't caught up or something, I am not well versed in the ledger code.

In this case, I think that a Warn is enough - 
1. Warnings are sent to the telemetry(& log file), without a stack trace. ( and that's fine, since we don't need stack trace here; the stack is pretty obvious here ).
2. Restarting the node won't help here : the node was just restarted. ** note that generally speaking, restarting is not a good solution, since many of the nodes are installed as a service.

Note that in the worst case scenario, the only ramification of the consensusParam is the filter timeout. This would get reset the next time a new round is being retrieved by the catchup, and being reported to the agreement in the demux loop.

The LedgerReader is pretty much a subset of the Ledger. ( ** tunneled through data/Ledger )

zeldovich:
> 3. This looping, in the mainLoop, as implemented, would cause more harm than good. The service need to be able to execute its main loop; otherwise, the incoming buffers would overflow, and there won't be a way to stop the service gracefully ( i.e. Shutdown ).

If we can't fetch the most recent blocks from the ledger, there isn't much use in running the agreement service.

It might be reasonable to check for shutdown in that loop, so that it's possible to shut down the agreement service even when it enters some kind of degenerate state this way.

> 4. I can't see why the changes in the `ledger.go` would be an issue : it's only extend the knowledge of the protocol version beyond the present into the future. It would not negatively affect the agreement.

I think the benefit of cleaning up that `ledger.go` code is simplicity and consistency.  The spec for what `ConsensusParams()` and `ConsensusVersion()` return becomes very simple and consistent between the two.  The simplicity also means that, if we decide to change how protocol upgrades work, at some point, we don't need to worry about keeping the `ledger.go` copy of that logic up-to-date.

tsachiherman:
> If we can't fetch the most recent blocks from the ledger, there isn't much use in running the agreement service.
> 
> It might be reasonable to check for shutdown in that loop, so that it's possible to shut down the agreement service even when it enters some kind of degenerate state this way.

I agree with that. I think that it would be quite an overhead to do it here, though. Keep in mind that you need to do most of the stuff in the demux loop *other* then moving the clock forward. I'm not suggesting that the existing implementation is perfect - in that case, it would make redundant proposals/votes; but these would get ignored.

> I think the benefit of cleaning up that ledger.go code is simplicity and consistency. The spec for what ConsensusParams() and ConsensusVersion() return becomes very simple and consistent between the two. The simplicity also means that, if we decide to change how protocol upgrades work, at some point, we don't need to worry about keeping the ledger.go copy of that logic up-to-date.

The main difference between `ConsensusVersion()` and `ConsensusParams()` is that the ledger is required to support the protocol version in order to provide information about the parameters, where as it doesn't require to support the consensus version in order to return `ConsensusVersion()`.

I think that the logic of extracting the version given a block header is valuable. I agree that if we change the upgrade logic we would probably need to modify this as well, but that's just an argument for better encapsulation, right ? In particular, I think that it's being used ( or should be ), in the catchup service.

jannotti:
I support the simplification, I have run into places where dealing with this prediction logic is annoying.

If the FilterTimeout is replaced anyway, is it crazy to just not fetch the consensus version, and put something bogus in?

PR #2342, opened 2021-06-30
Fix docker repo update

onetechnical:
## Summary

The scripts to update the docker repository do not do a full rebuild, and the submitted Dockerfile does not change. This means that if you run the docker image/repo update, it'll re-use the old image. To fix this, we call docker build with --no-cache. Other refactoring helps simplify the code.

We add a new `--cached` flag to the `docker/releases/build_releases.sh` script to explicitly cache. Otherwise the testnet update would issue a full rebuild. We additionally automatically handle the 'latest' tagging in build_releases.sh as well.

## Test Plan

Ran the update for betanet and mainnet/testnet, which seemed to update the container appropriately.

### DIFF
--- a/docker/releases/build_releases.sh
+++ b/docker/releases/build_releases.sh
@@ -27,9 +27,13 @@ DEPLOY=true
 IMAGE_NAME=stable
 NETWORK=mainnet
 TAGNAME=latest
+CACHED=false
 
 while [ "$1" != "" ]; do
     case "$1" in
+        --cached)
+            CACHED=true
+            ;;
         --name)
             shift
             IMAGE_NAME="${1-stable}"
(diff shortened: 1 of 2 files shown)


docker/releases/build_releases.sh
```diff
@@ -53,50 +57,65 @@ while [ "$1" != "" ]; do
     shift
 done
 
-if [[ ! "$NETWORK" =~ ^mainnet$|^testnet$|^betanet$ ]]
```
onetechnical, line 56: This was confusing split logic. Combined these into a simpler case statement.

docker/releases/build_releases.sh
```diff
@@ -53,50 +57,65 @@ while [ "$1" != "" ]; do
...
-        CHANNEL=beta
-    fi
+    ;;
+esac
 
-    IMAGE_NAME="$NETWORK"
-    NETWORK="-g $NETWORK"
-fi
-
-build_image () {
```
onetechnical, line 71: We create a subroutine here, then call it immediately. Might as well just dump in the code.

scripts/release/mule/deploy/docker/docker.sh
```diff
@@ -23,24 +23,14 @@ pushd docker/releases
...
 then
-    # Build and push mainnet.
-   ./build_releases.sh
+  # Build and push mainnet.
+  ./build_releases.sh --tagname "$VERSION"
 
-    # Build and push testnet.
-   ./build_releases.sh --network testnet
-
-    if [ -z "$VERSION" ]
```
onetechnical, line 32: This logic is unnecessary, because we test for this at the very start of the script.

tsachiherman (approved):

docker/releases/build_releases.sh
egieseke: Appreciate the improved clarity.

egieseke (approved):

PR #2343, opened 2021-06-30
txsync: adjust proposed block size to network bandwidth

tsachiherman:
## Summary

This PR modifies the transaction sync to update the txsync connector about the updated average data exchange rate.
The txsync connector is updating the transaction pool of that rate, which later on being used by the evaluator as a way to extrapolate the expected proposal size that would fit into the node's available bandwidth.

## Test Plan

Tested using limited bandwidth deployed network.


txnsync/interfaces.go
```diff
@@ -71,7 +71,10 @@ type NodeConnector interface {
 	Random(uint64) uint64
 	GetPeers() []PeerInfo
 	GetPeer(interface{}) PeerInfo // get a single peer given a network peer opaque interface
-	UpdatePeers([]*Peer, []interface{})
+	// UpdatePeers call is being made to inform the node that either a link need to be established
+	// between the set of the txsyncPeers peers and the set of netPeers, or that the averageDataExchangeRate
+	// was recalculated and could potentially be updated.
+	UpdatePeers(txsyncPeers []*Peer, netPeers []interface{}, averageDataExchangeRate uint64)
```
winder, line 77: Could you mention the units of the rate somewhere?

Also, it isn't clear that this is the `averagePerPeerDataExchangeRate`

data/pools/transactionPool.go
```diff
@@ -965,10 +974,50 @@ func (pool *TransactionPool) assembleEmptyBlock(round basics.Round) (assembled *
...
 }
+
+// SetDataExchangeRate updates the data exchange rate this node is expected to have.
+func (pool *TransactionPool) SetDataExchangeRate(dataExchangeRate uint64) {
+	atomic.StoreUint64(&pool.latestMeasuredDataExchangeRate, dataExchangeRate)
+}
+
+// calculateMaxTxnBytesPerBlock computes the optimal block size for the current node, based
+// on it's effective network capabilities. This number is bound by the protocol MaxTxnBytesPerBlock.
+func (pool *TransactionPool) calculateMaxTxnBytesPerBlock(consensusVersion protocol.ConsensusVersion) int {
```
winder, line 992: Thinking out loud for a moment... This introduces a new dependency on the protocol. While you didn't add the import, previously we were only using a `TxType` constant in this file. Maybe pass `dataExchangeRate` instead and do this computation in the evaluator?

Another option would be moving the function to Eval as a helper.

For testability, you can avoid the receiver here. The only dependencies are `config.ConsensusProtocols` and `pool.latestMeasuredDataExchangeRate`, if you pass them directly there is no concern about side effects.

ledger/eval.go
```diff
@@ -382,16 +383,21 @@ type ledgerForCowBase interface {
...
 
-func startEvaluator(l ledgerForEvaluator, hdr bookkeeping.BlockHeader, paysetHint int, validate bool, generate bool) (*BlockEvaluator, error) {
+func startEvaluator(l ledgerForEvaluator, hdr bookkeeping.BlockHeader, paysetHint int, validate bool, generate bool, maxTxnBytesPerBlock int) (*BlockEvaluator, error) {
 	proto, ok := config.Consensus[hdr.CurrentProtocol]
 	if !ok {
 		return nil, protocol.Error(hdr.CurrentProtocol)
 	}
 
+	// if the caller did not provide a valid block size limit, default to the consensus params defaults.
+	if maxTxnBytesPerBlock == 0 {
```
winder, line 397:
```suggestion
	// if the caller did not provide a valid block size limit, default to the consensus params defaults.
	if maxTxnBytesPerBlock <= 0 || maxTxnBytesPerBlock > proto.MaxTxnBytesPerBlock {
```

winder (changes requested):
Looks good. One change needed and some questions/comments/nits.

ledger/eval.go
```diff
@@ -382,16 +383,21 @@ type ledgerForCowBase interface {
 // of the block that the caller is planning to evaluate. If the length of the
 // payset being evaluated is known in advance, a paysetHint >= 0 can be
 // passed, avoiding unnecessary payset slice growth.
```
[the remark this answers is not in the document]
tsachiherman: I was thinking about that. The reason I ended up with the current approach was to avoid creating "branching". I think that this new field isn't any different from the "hint", so it might be there as well.
I also considered creating the evaluator, and then calling a method on the evaluator to update this variable. Again, I think that it would be simpler from readability perspective to have this parameter on the `StartEvaluator`.

If we will end up passing too many optional arguments to the evaluator's constructor, I would create a struct for these.. but it doesn't look like we're there yet.

txnsync/interfaces.go
tsachiherman: yes. I'll rename & clarify documentation.

data/pools/transactionPool.go
tsachiherman: starting from the end:
I think that for testability, it would be very trivial as is ( I'll add few tests ).
It's definitely doable for the evaluator to "consume" that value directly off the transaction pool.. but it makes it much more linked - which isn't a good thing. Having this computation being done on either the transaction pool or the evaluator is fine - but I don't know if we need the evaluator to be aware of "network" related constructs.

In fact, I think it would have been fine moving this logic from the transaction pool into the transaction sync connector if it wasn't for the block-specific version, which might affect the block size.

ledger/eval.go
tsachiherman: sure.

txnsync/interfaces.go
```diff
@@ -71,7 +71,12 @@ type NodeConnector interface {
 	Random(uint64) uint64
 	GetPeers() []PeerInfo
 	GetPeer(interface{}) PeerInfo // get a single peer given a network peer opaque interface
-	UpdatePeers([]*Peer, []interface{})
+	// UpdatePeers call is being made to inform the node that either a link need to be established
+	// between the set of the txsyncPeers peers and the set of netPeers, or that the peersAverageDataExchangeRate
+	// was recalculated and could potentially be updated.
+	// The peersAverageDataExchangeRate passed in here is the average communication rate ( measured in bytes per second )
+	// across all the connected peers.
+	UpdatePeers(txsyncPeers []*Peer, netPeers []interface{}, peersAverageDataExchangeRate uint64)
```
winder, line 79: Sorry but this still isn't clear to me, unless I'm mistaking something.

I see `averageDataExchangeRate /= uint64(len(peersInfo))`, so it looks like this isn't the rate across all peers but the rate per peer.

tsachiherman: it's the average..

winder (approved):
LGTM

data/pools/transactionPool.go
```diff
@@ -965,10 +974,50 @@ func (pool *TransactionPool) assembleEmptyBlock(round basics.Round) (assembled *
...
+	halfMaxBlockSize := int(time.Duration(dataExchangeRate) * proto.AgreementFilterTimeoutPeriod0 / (2 * time.Second))
+
+	// if the amount of data is too high, bound it by the consensus parameters.
+	if halfMaxBlockSize > proto.MaxTxnBytesPerBlock {
+		return proto.MaxTxnBytesPerBlock
+	}
+
+	// if the amount of data is too low, use the low transaction bytes threshold.
+	if halfMaxBlockSize < minMaxTxnBytesPerBlock {
+		return proto.MaxTxnBytesPerBlock
```
nicholasguoalgorand, line 1019: Should this be `return minMaxTxnBytesPerBlock`?

data/pools/transactionPool.go
```diff
@@ -965,10 +974,50 @@ func (pool *TransactionPool) assembleEmptyBlock(round basics.Round) (assembled *
...
+
+	// get the consensus parameters for the given consensus version.
+	proto, ok := config.Consensus[consensusVersion]
+	if !ok {
+		// if we can't figure out the consensus version, just return 0.
+		return 0
+	}
+
+	// calculate the amount of data we can send in half of the agreement period.
+	halfMaxBlockSize := int(time.Duration(dataExchangeRate) * proto.AgreementFilterTimeoutPeriod0 / (2 * time.Second))
```
nicholasguoalgorand, line 1010: This logic is a tad bit confusing. Can we fix the type casting so that the units make sense? (for example the `2 * time.Seconds` is not really a value in seconds.)

data/pools/transactionPool.go
```diff
@@ -157,6 +162,10 @@ const (
 	// duration it would take to execute the GenerateBlock() function
 	generateBlockBaseDuration        = 2 * time.Millisecond
 	generateBlockTransactionDuration = 2155 * time.Nanosecond
+
+	// minMaxTxnBytesPerBlock is the minimal maximum block size that the evaluator would be asked to create, in case
```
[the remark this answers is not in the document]
nicholasguoalgorand: Specifically, I had a hard time understanding what `minMaxTxnBytesPerBlock` was

txnsync/mainloop.go
```diff
@@ -369,6 +369,9 @@ func (s *syncState) getPeers() (result []*Peer) {
 	peersInfo := s.node.GetPeers()
 	updatedNetworkPeers := []interface{}{}
 	updatedNetworkPeersSync := []*Peer{}
+
+	var averageDataExchangeRate uint64
```
nicholasguoalgorand, line 373: Just wanted to initiate a thought experiment: Is average the right metric? How would using average compare to using the maximum instead?

nicholasguoalgorand (commented):
Looks good, had a few comments / questions

data/pools/transactionPool.go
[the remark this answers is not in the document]
tsachiherman: Add example. We use the minMaxXXX in other locations in the codebase as well. I agree that it's not super-intuitive at first.

data/pools/transactionPool.go
tsachiherman: yes. Fixed.

data/pools/transactionPool.go
tsachiherman: yes; done.

txnsync/mainloop.go
tsachiherman: If we have 4 connected relays. The connection to one of them is slow. We still want to try and build the proposal so it would get on time to most of them. So, we can't take the fastest. But taking the slowest opens to an attack vector, since it gives a slow relay the ability to control the proposal size.

Taking the average seems to be a reasonable compromise. It's definitely not perfect, but it would become better once you're done with the proposal breakout.

nicholasguoalgorand (approved):
lgtm

PR #2345, opened 2021-06-30
riff on experimental Go binomial distribution

brianolson:
## Summary

Adapted from  #2316
add test to compare output of Boost CDF vs gonum

## Test Plan

Testing currently shows substantial difference in Go CDF vs Boost; keep testing until they line up.


barnjamin:
The output from the CDF call is weird, it should be continuously increasing but when the exponents get large enough it swings wildly. 

just varying n or p, it doesn't fail a test for < the previous value but moving them in opposite directions, by e14 and e-14, it fails consistently.

2021/06/30 13:54:41 Testing 7.336342e+06, 4.827838e-06
2021/06/30 13:54:41 Testing 7.336342e+07, 4.827838e-07
2021/06/30 13:54:41 Testing 7.336342e+08, 4.827838e-08
2021/06/30 13:54:41 Testing 7.336342e+09, 4.827838e-09
2021/06/30 13:54:41 Testing 7.336342e+10, 4.827838e-10
2021/06/30 13:54:41 Testing 7.336342e+11, 4.827838e-11
2021/06/30 13:54:41 Testing 7.336342e+12, 4.827838e-12
2021/06/30 13:54:41 Testing 7.336342e+13, 4.827838e-13
2021/06/30 13:54:41 2.464781e-01 < 3.201747e-01

```
package main

import (
        "log"
        "math"

        "gonum.org/v1/gonum/stat/distuv"
)

var (
        sn = 7.336342e+5
        sp = 4.827838e-5
)

func main() {
        for e := 1; e < 100; e++ {
                n := sn * math.Pow(10, float64(e))
                p := sp / math.Pow(10, float64(e))
                log.Printf("Testing %e, %e", n, p)
                distro := distuv.Binomial{N: n, P: p}
                last := 0.0
                for x := 0; x < 100; x++ {
                        v := distro.CDF(float64(x))
                        if v < last {
                                log.Printf("%e < %e", v, last)
                                return
                        }
                        last = v
                }
        }
}
```

I'm not sure if this is just instability caused by the size of the float64 of something with the algorithm being used (https://github.com/gonum/gonum/blob/89503b3627cbcf7718866503a835c10d03e4da9c/mathext/internal/cephes/incbeta.go#L25)

barnjamin:
The test you wrote has them matching well when total money is 1e12 but starts to get iffy past that

barnjamin:
Tried out https://github.com/vsivsi/bigbinomial here:

https://play.golang.org/p/HEIToSw82Jp 

and it doesn't seem to suffer from the same numerical instability.  

You want a pr for this or you want to work on it?

barnjamin:
Sorry to keep spamming this pr but there is an implementation of the `regularized incomplete beta function` in the golang perf tools here: https://github.com/golang/perf/blob/master/internal/stats/beta.go

To me the perf one is easier to read than the gonum one, tho probably not as optimized.

brianolson:
@barnjamin good one! I hacked that it in and it got perfect match for 500 tests! Now I just need to mod the test to run a billion times and let some server test that for the next week.

brianolson:
The correctness is a good first step, but bigbinomial might be too slow
```
BenchmarkBoostCdfWalk-8         	   71038	     16873 ns/op
BenchmarkBigbinomialCdfWalk-8   	      58	  37525372 ns/op
```
But bigbinomial is a clean implementation using `math/big`, transliterating it to plain float64 might be the answer... ?

barnjamin:
>But bigbinomial is a clean implementation using math/big, transliterating it to plain float64 might be the answer... ?

I think it has a lot to do with the algorithm used, gonum uses that regularized incomplete beta function method to compute the value, same as boost, so it was competative.

Also gonum was insanely wrong so probably bailing early.  I can try to implement a simpler version of the RIBF over the weekend if you'd like.

I was also trying to come up with a better way to do this, but the money supply changes over time so we cant just compute it on start and use a fat LUT, right?

barnjamin:
enabling the integer fast path in the bigfloat Pow function improves the situation but we're still an order of magnitude off at least. 
```
BenchmarkBoostCdfWalk-12                   66174             18175 ns/op
BenchmarkBigbinomialCdfWalk-12              1422            925557 ns/op
```

barnjamin:
I'm pretty sure I can just accumulate the pmf during the "walk the cdf" iteration to find the boundary instead of dealing with memoizing in some foreign struct. 

This seems to both be fast and (mostly?) correct.

I can't force push here, just update mine?


```
    sortition_test.go:109: 1 total err across 100 tests, avg=0.010000 (0.000204), max=1
--- PASS: TestSortitionExhausting (0.01s)
goos: linux
goarch: amd64
pkg: github.com/algorand/go-algorand/data/committee/sortition
BenchmarkSortition
BenchmarkSortition-12             552352              1864 ns/op
BenchmarkBoostCdfWalk
BenchmarkBoostCdfWalk-12           67935             17363 ns/op
BenchmarkCdfAccum
BenchmarkCdfAccum-12              498415              2316 ns/op
PASS
ok      github.com/algorand/go-algorand/data/committee/sortition        3.697s
```



```Go
func FindBinomialBoundary(p, ratio float64, money uint64) uint64 {
        b := distuv.Binomial{P: p, N: float64(money)}
        cdf := 0.0
        for j := uint64(0); j < money; j++ {
                //Get prob at x
                px := b.Prob(float64(j))

                // Idk?  Probably wrong
                if px == 0.0 {
                        return money
                }

                //Add discrete prob to accumate density
                cdf += px

                if ratio <= cdf {
                        return j
                }

        }
        return money
}
```

barnjamin:
Another option to maybe speed this up and get accurate answers is to use the Poisson approximation. 

It is more accurate with large n and small p, where the lambda factor is just n*p.  Since n  in our case is # microalgos, it _should_ be a decent approximator but I haven't tried it yet.

My sense is that with large float values, the more calculations we do, the more rounding errors we introduce.  So comparing Binomial pmf vs Poisson pmf, Poisson has fewer steps so _should_(?) be more accurate

https://github.com/gonum/gonum/blob/master/stat/distuv/binomial.go#L51-L57

https://github.com/gonum/gonum/blob/master/stat/distuv/poisson.go#L44-L50

barnjamin:
Using Poisson and comparing it to the bigbinomial for accuracy, it is 100% accurate in the limited tests and it is even faster than the  binomial pmf accumulation.

`BenchmarkPoissonCdfWalk-12        647182              1756 ns/op`

barnjamin:
https://github.com/algorand/go-algorand/pull/2526 submitted a non draft

gmalouf:
Closing this PR for now: It would be good to potentially revisit after

a) Opening an issue describing the issue/goals we are working towards + getting buy-in on the approach
b) Having a path towards merging this laid out.

PR #2358, opened 2021-06-30
selector pseudo-op in support of ABI

jannotti:
Adds a Teal pseudo-op `selector` that assembles as if it were the `byte` pseudo-op, but stores 4 bytes of hash.

This allows Teal, like:
```
txn ApplicationArgs 0
selector "add(uint64,uint64)uint128"
==
bnz add
```

and avoid the need to embed the actual hash in a .teal file (which would also require calculating it)

Unlike `byte`, the argument *must* be a quoted string - no base64 or hex allowed, as the argument should be a method signature.

### DIFF
--- a/data/transactions/logic/assembler.go
+++ b/data/transactions/logic/assembler.go
@@ -644,10 +644,10 @@ func assembleByte(ops *OpStream, spec *OpSpec, args []string) error {
 	return nil
 }
 
-// selector "add(uint64,uint64)uint64"
-func assembleSelector(ops *OpStream, spec *OpSpec, args []string) error {
+// method "add(uint64,uint64)uint64"
+func assembleMethod(ops *OpStream, spec *OpSpec, args []string) error {
 	if len(args) == 0 {
-		return ops.error("selector requires a literal argument")
+		return ops.error("method requires a literal argument")
 	}
 	arg := args[0]
 	if len(arg) > 1 && arg[0] == '"' && arg[len(arg)-1] == '"' {
@@ -1056,7 +1056,7 @@ var keywords = map[string]OpSpec{
 	// parse basics.Address, actually just another []byte constant
 	"addr": {0, "addr", nil, assembleAddr, nil, nil, oneBytes, 1, modeAny, opDetails{1, 2, nil, nil}},
 	// take a signature, hash it, and take first 4 bytes, actually just another []byte constant
-	"selector": {0, "selector", nil, assembleSelector, nil, nil, oneBytes, 1, modeAny, opDetails{1, 2, nil, nil}}}
+	"method": {0, "method", nil, assembleMethod, nil, nil, oneBytes, 1, modeAny, opDetails{1, 2, nil, nil}}}
 
 type lineError struct {
 	Line int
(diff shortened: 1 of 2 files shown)


jasonpaulos (commented):
This looks good, though I worry that `selector` is too similar to `select`. Maybe one of `method_selector`, `method_sel`, `m_selector`, `msel`, or similar would differentiate the two a little more?

data/transactions/logic/assembler.go
```diff
@@ -1036,7 +1055,8 @@ var keywords = map[string]OpSpec{
 	"byte": {0, "byte", nil, assembleByte, nil, nil, oneBytes, 1, modeAny, opDetails{1, 2, nil, nil}},
 	// parse basics.Address, actually just another []byte constant
 	"addr": {0, "addr", nil, assembleAddr, nil, nil, oneBytes, 1, modeAny, opDetails{1, 2, nil, nil}},
-}
+	// take a signature, hash it, and take first 4 bytes, actually just another []byte constant
+	"selector": {0, "selector", nil, assembleSelector, nil, nil, oneBytes, 1, modeAny, opDetails{1, 2, nil, nil}}}
```
algorandskiy, line 1059: version 1 ? if so why is the code below needed?
```
			if spec.Version > 1 && spec.Version > ops.Version {
				ok = false
			}
```

algorandskiy (commented):
The code looks good, not sure about the keyword name. In fact it converts human readable method signature to ABI method signature, so maybe use some close synonym of "signature" like "imprint", "autograph", "ack", "endorsement" ?

data/transactions/logic/assembler.go
jannotti: Because originally, I felt pseudo-ops must be version controlled like normal ops.  And, in the general case, they do (if they assemble down to an op that is, itself, versioned).  But this one does not need to be versioned, as it combiles to `byte` which has always existed.

I left the version checking in though, so that the next pseudo-op, which maybe needs the checking, is done right. (For example, we're about to add a pseudo op along with a new `extract` opcode)

jannotti:
> This looks good, though I worry that `selector` is too similar to `select`. Maybe one of `method_selector`, `method_sel`, `m_selector`, `msel`, or similar would differentiate the two a little more?

I wanted to stick to the nomenclature from the ABI document. I could probably be convinced to change it, but I don't think there's a ton of room for confusion, as they take different (immediate) arguments and do such different things.

AyAggarwal (approved):

data/transactions/logic/assembler.go
```diff
@@ -643,6 +644,24 @@ func assembleByte(ops *OpStream, spec *OpSpec, args []string) error {
...
+// method "add(uint64,uint64)uint64"
+func assembleMethod(ops *OpStream, spec *OpSpec, args []string) error {
+	if len(args) == 0 {
+		return ops.error("method requires a literal argument")
+	}
+	arg := args[0]
+	if len(arg) > 1 && arg[0] == '"' && arg[len(arg)-1] == '"' {
+		val, err := parseStringLiteral(arg)
+		if err != nil {
+			return ops.error(err)
```
algorandskiy, line 656: nit: convert `fmt.Errorf("no quotes")` in `parseStringLiteral` to a typed `QuotesError` type and handle it here.

algorandskiy (approved):

jasonpaulos (approved):

PR #2413, opened 2021-07-01
optimize txtail memory consumption

tsachiherman:
## Summary

The changes in this PR are as follows:
1. The unused method `Ledger. GetRoundTxIds` was removed. As a result, the `txTail. getRoundTxIds` can be removed as well. This makes the `txids` map stored in the `roundTxMembers` structure redundant.
2. In the `cow.go`, avoid adding empty leases to the `cb.mods.Txleases` map. Since we already not testing for empty leases, we can safely avoid storing them.
3. Optimize the `txTail. loadFromDisk` to generate optimal `lastValid` map sizes.
4. Optimize the `txTail. loadFromDisk` to avoid storing empty leases in the `txleases` map.

## Test Plan

- [x] Use existing travis tests to cover basic functionality
- [x] Unit test added
- [x] Full catchup test


ledger/txtail.go
```diff
@@ -71,19 +79,44 @@ func (t *txTail) loadFromDisk(l ledgerForTracker) error {
...
 		}
+
 		for _, txad := range payset {
 			tx := txad.SignedTxn
-			t.recent[old].txids[tx.ID()] = tx.Txn.LastValid
-			t.recent[old].txleases[ledgercore.Txlease{Sender: tx.Txn.Sender, Lease: tx.Txn.Lease}] = tx.Txn.LastValid
-			t.putLV(tx.Txn.LastValid, tx.ID())
+			if consensusParams.SupportTransactionLeases && (tx.Txn.Lease != [32]byte{}) {
+				t.recent[old].txleases[ledgercore.Txlease{Sender: tx.Txn.Sender, Lease: tx.Txn.Lease}] = tx.Txn.LastValid
+			}
```
algorandskiy, line 92: can't this IF cause problems on early history replaying

ledger/txtail.go
```diff
@@ -71,19 +79,44 @@ func (t *txTail) loadFromDisk(l ledgerForTracker) error {
...
 		}
 	}
 
+	// add all the entries in roundsLastValids to their corresponding map entry in t.lastValid
+	for lastValid, list := range roundsLastValids {
+		lastValueMap := make(map[transactions.Txid]struct{}, len(list))
+		for _, id := range list {
+			lastValueMap[id] = struct{}{}
+		}
+		t.lastValid[lastValid] = lastValueMap
```
algorandskiy, line 118: there is no `t.putLV` in new code, is it functionally equivalent?

ledger/txtail.go
tsachiherman: I don't believe it would cause issues. But I should definitely perform a full catchup to verify.
The case where we have an empty lease is never tested in `checkDup` below.

ledger/txtail.go
tsachiherman: yes; it's doing exactly what `t.putLV` does, but pre-allocate the maps to exact size needed. This is very important when we start a node that was under a great transactional load before : we don't want the node to start up consuming lot's of memory that isn't really needed. Instead, it allocates *exactly* the amount of needed map space. Note that as a side affect, the `roundsLastValids` here could be substantial, however, it will all be dumped when this method aborts.

tsachiherman:
Yes, I let it run through a complete mainnet catchup. No issues found.

algorandskiy (approved):

PR #2415, opened 2021-07-01
Save Logging Information in Data Directory If Provided

AlgoStephenAkiki:
## Summary

Our logic should be as follows:

When Loading:

 - We first look inside the provided data-directory.  If a config file is there, load it and return it
- Otherwise, look in the global directory.  If a config file is there, load it and return it.

When Saving:
- If a data-directory was provided then save the config file there.
- Otherwise, save the config file in the global directory

## Test Plan

Tested manually.

### DIFF
--- a/logging/log.go
+++ b/logging/log.go
@@ -391,9 +391,6 @@ func (l logger) GetTelemetrySession() string {
 }
 
 func (l logger) GetTelemetryHostName() string {
-	if !l.GetTelemetryEnabled() {
-		return ""
-	}
 	return l.loggerState.telemetry.telemetryConfig.getHostName()
 }
 
--- a/logging/log.go
+++ b/logging/log.go
@@ -391,7 +391,7 @@ func (l logger) GetTelemetrySession() string {
 }
 
 func (l logger) GetTelemetryHostName() string {
-	if l.loggerState.telemetry == nil {
+	if !l.GetTelemetryEnabled() {
 		return ""
 	}
 	return l.loggerState.telemetry.telemetryConfig.getHostName()
--- a/logging/log.go
+++ b/logging/log.go
@@ -391,6 +391,9 @@ func (l logger) GetTelemetrySession() string {
 }
 
 func (l logger) GetTelemetryHostName() string {
+	if l.loggerState.telemetry == nil {
+		return ""
+	}
 	return l.loggerState.telemetry.telemetryConfig.getHostName()
 }
 
--- a/logging/telemetry.go
+++ b/logging/telemetry.go
@@ -93,24 +93,42 @@ func makeTelemetryState(cfg TelemetryConfig, hookFactory hookFactory) (*telemetr
 func ReadTelemetryConfigOrDefault(dataDir string, genesisID string) (cfg TelemetryConfig, err error) {
 	err = nil
 	dataDirProvided := dataDir != ""
+	var configPath string
+
+	// If we have a data directory, then load the config
 	if dataDirProvided {
-		configPath := filepath.Join(dataDir, TelemetryConfigFilename)
+		configPath = filepath.Join(dataDir, TelemetryConfigFilename)
+		// Load the config, if the GUID is there then we are all set
+		// However if it isn't there then we must create it, save the file and load it.
 		cfg, err = LoadTelemetryConfig(configPath)
 	}
+
+	// We couldn't load the telemetry config for some reason
+	// If the reason is because the directory doesn't exist or we didn't provide a data directory then...
 	if (err != nil && os.IsNotExist(err)) || !dataDirProvided {
-		var configPath string
+
 		configPath, err = config.GetConfigFilePath(TelemetryConfigFilename)
 		if err != nil {
+			// In this case we don't know what to do since we couldn't
+			// create the directory.  Just create an ephemeral config.
 			cfg = createTelemetryConfig()
 			return
 		}
+
+		// Load the telemetry from the default config path
 		cfg, err = LoadTelemetryConfig(configPath)
 	}
+
+	// If there was some error loading the configuration from the config path...
 	if err != nil {
+		// Create an ephemeral config
 		cfg = createTelemetryConfig()
+
+		// If the error was that the the config wasn't there then it wasn't really an error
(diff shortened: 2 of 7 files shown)


logging/log.go
```diff
@@ -391,10 +391,7 @@ func (l logger) GetTelemetrySession() string {
 }
 
 func (l logger) GetTelemetryHostName() string {
-	if !l.GetTelemetryEnabled() {
-		return ""
-	}
-	return l.loggerState.telemetry.telemetryConfig.getHostName()
+	return l.loggerState.telemetry.telemetryConfig.GUID
```
winder, line 394: I think we still want `getHostName()`, the optional name field is also static so this would only give us better information if available and still be static.

logging/telemetry.go
```diff
@@ -93,24 +93,44 @@ func makeTelemetryState(cfg TelemetryConfig, hookFactory hookFactory) (*telemetr
...
+
+	// If there was some error loading the configuration from the config path...
 	if err != nil {
+		// Create an ephemeral config
 		cfg = createTelemetryConfig()
+
+		// If the error was that the the config wasn't there then it wasn't really an error
+		// We actually want to save the information here
 		if os.IsNotExist(err) {
+			cfg.Save(configPath)
```
winder, line 130: I think we can leave out the save, it's possible this was left out by design. It looks like `diagcfg` is the only consumer of this function and it saves the config itself after making a modification, maybe move this function into the diagcfg package?

winder (changes requested):
Looks good, just 2 small requests

logging/telemetry.go
tsachiherman: With the save, what would make this GUID static ?

winder: Diagcfg uses this function and saves the config file immediately after fetching the default. Algod/Algoh use a different `EnsureTelemetry` function which does the create/save or read.

AlgoStephenAkiki: The save makes it so that the ephemeral config created on 125 is saved and loaded the next time this function (or the related EnsureTelemetryConfigCreated) is called.

AlgoStephenAkiki: > I think we can leave out the save, it's possible this was left out by design. It looks like `diagcfg` is the only consumer of this function and it saves the config itself after making a modification, maybe move this function into the diagcfg package?

It does seem to only be called there.  I'm fine dropping it if everyone agrees.

winder (approved):

AlgoStephenAkiki:
No unit tests added but manually tested via log statements and enabling/disabling telemetry reporting.

logging/telemetry.go
```diff
@@ -134,9 +152,20 @@ func EnsureTelemetryConfig(dataDir *string, genesisID string) (TelemetryConfig,
...
 // whether EnsureTelemetryConfig had to create the config.
 func EnsureTelemetryConfigCreated(dataDir *string, genesisID string) (TelemetryConfig, bool, error) {
+	/*
+		Our logic should be as follows:
+			- We first look inside the provided data-directory.  If a config file is there, load it
+			  and return it
+			- Otherwise, look in the global directory.  If a config file is there, load it and return it.
+			- Otherwise, if a data-directory was provided then save the config file there.
+			- Otherwise, save the config file in the global directory
+
```
winder, line 162: @tsachiherman do you know how we handle the deb/rpm installations? I thought there was another case where it could be stored somewhere in the /var directory

winder (approved):

logging/telemetry.go
tsachiherman: I think that the `var` location was meant as a global location, and it's used by rpm installers.

tsachiherman (dismissed):
Please remove the deleted `.circleci/config.yml` from the PR.. you don't want to affect that..

AlgoStephenAkiki:
> Please remove the deleted `.circleci/config.yml` from the PR.. you don't want to affect that..

Resolved.

winder (approved):

logging/telemetry_test.go
```diff
@@ -335,4 +335,5 @@ func TestReadTelemetryConfigOrDefaultNoDataDir(t *testing.T) {
 	a.Equal(defaultCfgSettings.UserName, cfg.UserName)
 	a.Equal(defaultCfgSettings.Password, cfg.Password)
 	a.Equal(len(defaultCfgSettings.GUID), len(cfg.GUID))
+
```
tsachiherman, line 338: nit: avoid changing this file, as there are no meaningful changes here.

logging/telemetryConfig.go
```diff
@@ -33,6 +33,9 @@ import (
 // TelemetryConfigFilename default file name for telemetry config "logging.config"
 var TelemetryConfigFilename = "logging.config"
 
+var DefaultTelemetryUsername = "telemetry-v9"
+var DefaultTelemetryPassword = "oq%$FA1TOJ!yYeMEcJ7D688eEOE#MGCu"
```
tsachiherman, line 37: use lower-case for `DefaultTelemetryUsername` . i.e. `defaultTelemetryUsername`
and the same for DefaultTelemetryPassword, since we don't want to export these.

logging/telemetryConfig_test.go
```diff
@@ -119,3 +119,14 @@ func TestLoadTelemetryConfig(t *testing.T) {
 	require.Equal(t, createTelemetryConfig().Password, tc.Password)
 
 }
+
+func TestLoadTelemetryConfigBlankUsernamePassword(t *testing.T ) {
```
tsachiherman, line 123: you could get some extra credit if you'll add a test for saving the file as well.

logging/log.go
```diff
@@ -391,7 +391,7 @@ func (l logger) GetTelemetrySession() string {
 }
 
 func (l logger) GetTelemetryHostName() string {
-	if !l.GetTelemetryEnabled() {
+	if l.loggerState.telemetry == nil {
```
tsachiherman, line 394: isn't this change effectively rolling back to the previous implementation ?
If so, keeping  `!l.GetTelemetryEnabled()` would be cleaner.

tsachiherman (approved):

PR #2418, opened 2021-07-01
Specs and unit test to avoid forgetting in the future

jannotti:
Pure spec update.


tsachiherman (approved):
looks good to me.

data/transactions/logic/doc_test.go
```diff
@@ -77,6 +77,15 @@ func TestOpImmediateNote(t *testing.T) {
 	require.Empty(t, xd)
 }
 
+func TestAllImmediatesDocumented(t *testing.T) {
+	for _, op := range OpSpecs {
+		if len(op.Details.Immediates) > 0 {
+			xd := OpImmediateNote(op.Name)
+			require.NotEmpty(t, xd, "%s has undocumented immediates", op.Name)
```
jdtzmn, line 84: This is good enough for now, but ideally we'd have a test that confirms the two lengths are equivalent.

jdtzmn (approved):

winder:
You should put a call to the makefile in data/transactions/logic in `scripts/travis/codegen_verification.sh`, it will automatically fail the build if any changes are made.

tsachiherman:
@jannotti - The changes here are good to go as is. However, if you're planning to follow any of the suggestions here, please let me know and I will hold off with merging this one in.

jannotti:
The code gen check sounds useful. I'll do that tonight.

algorandskiy (dismissed):

data/transactions/logic/doc_test.go
jannotti: I added a check that counts braces.  It has to be skipped for a few more complicated opcodes, but it's a nice addition for most.

jannotti:
> You should update `scripts/travis/codegen_verification.sh` with a call to the makefile in data/transactions/logic, it will automatically fail the build if any changes are made.

I did this, does it seem right?
```
echo "Updating TEAL Specs"
make -C data/transactions/logic
```

I gather the idea is that this will be a no-op if the generated files are correctly checked in, so the git checkout remains clean.

winder (approved):
LGTM

PR #2430, opened 2021-07-02
allow to parse ipv6 localhost "[::]:4601"

brianolson:
## Summary

Allow parsing of ipv6 localhost -colon- port address.
"[::]:4601" was failing to parse. This and other forms pass ParseHostOrURL() now. 

## Test Plan

Added unit test for previously uncovered `ParseHostOrURL()`

### DIFF
--- a/network/wsNetwork.go
+++ b/network/wsNetwork.go
@@ -1917,12 +1917,21 @@ var HostColonPortPattern = regexp.MustCompile("^[^:]+:\\d+$")
 // ParseHostOrURL handles "host:port" or a full URL.
 // Standard library net/url.Parse chokes on "host:port".
 func ParseHostOrURL(addr string) (*url.URL, error) {
-	var parsedURL *url.URL
+	// If the entire addr is "host:port" grab that right away.
+	// Don't try url.Parse() because that will grab "host:" as if it were "scheme:"
 	if HostColonPortPattern.MatchString(addr) {
-		parsedURL = &url.URL{Scheme: "http", Host: addr}
-		return parsedURL, nil
+		return &url.URL{Scheme: "http", Host: addr}, nil
 	}
-	return url.Parse(addr)
+	parsed, err := url.Parse(addr)
+	if err == nil {
+		return parsed, nil
+	}
+	// This turns "[::]:4601" into "http://[::]:4601" which url.Parse can do
+	parsed, e2 := url.Parse("http://" + addr)
+	if e2 == nil {
+		return parsed, nil
+	}
+	return parsed, err /* return original err, not our prefix altered try */
 }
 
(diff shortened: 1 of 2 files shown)


tsachiherman (dismissed):
Three requests - 
1. It would be nice to have few errors in the `urlTestCases`; your error case handling looks like it's already ready to handle these.
2. it would be better to declare `urlTestCases` as local variable. There is no benefits of making it global.
3. Add use cases that doesn't include the optional port number, and also pick a non-uniform port number.

Beside these, the change itself looks fine.

network/wsNetwork.go
```diff
@@ -1911,18 +1911,35 @@ var errBcastInvalidArray = errors.New("invalid broadcast array")
...
+		return &url.URL{Scheme: "http", Host: addr}, nil
+	}
+	parsed, err := url.Parse(addr)
+	if err == nil {
+		if parsed.Host == "" {
+			return nil, errUrlNoHost
+		}
+		return parsed, nil
+	}
+	if strings.HasPrefix(addr, "http:") || strings.HasPrefix(addr, "https:") || strings.HasPrefix(addr, "ws:") || strings.HasPrefix(addr, "wss:") {
```
tsachiherman, line 1934: you'd better add the `//` to all these prefixes as well.

network/wsNetwork.go
```diff
@@ -1911,18 +1911,35 @@ var errBcastInvalidArray = errors.New("invalid broadcast array")
...
+		return parsed, err
```
tsachiherman, line 1935: if you're returning an error here, avoid returning `parsed` and return `nil` explicitly. ( since you know that it's nil )

network/wsNetwork_test.go
```diff
@@ -1969,3 +1970,52 @@ func BenchmarkVariableTransactionMessageBlockSizes(t *testing.B) {
...
+		"ws://localhost:WAT",
+		"wss://localhost:WAT",
+	}
+	for _, tc := range urlTestCases {
+		t.Run(tc.text, func(t *testing.T) {
+			v, err := ParseHostOrURL(tc.text)
+			if err != nil {
+				t.Errorf("unexpected error, %s", err)
+				return
+			}
```
tsachiherman, line 2006: nit: we usually use `require.NoError(t, err)` but this one is fine as well.

tsachiherman:
unit tests are good.

network/wsNetwork.go
brianolson: the prior behavior was `return url.Parse()` so I'm maintaining that

network/wsNetwork.go
brianolson: added, and more test cases.

tsachiherman:
looks like some of the travis tests are failing.

tsachiherman (approved):

PR #2444, opened 2021-07-06
Wrap up TBD items on circleCI work

bricerisingalgorand:
## Summary

I'm making this PR at the start of work so that I can continuously see that it's running the circleci jobs correctly. Below are the items I am working on:

- [x] Add mac
- [x] Right size instances
- [x] Collect coverage reports
- [x] Make sure parallelism is 1 everywhere
- [x] Maybe break out e2e subs into another job

Will do later
- Implement linter to enforce partition line in test files
- Optimize package installs


testpartitioning/testFiltering.go
```diff
@@ -0,0 +1,69 @@
...
+// published by the Free Software Foundation, either version 3 of the
+// License, or (at your option) any later version.
+//
+// go-algorand is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with go-algorand.  If not, see <https://www.gnu.org/licenses/>.
```
winder, line 31: license was added twice

testpartitioning/testFiltering.go
```diff
@@ -0,0 +1,69 @@
...
+		t.Skip()
+	}
+}
+
+func stringToUint64(str string) uint64 {
+	sum := uint64(0)
+	for _, x := range str {
+		sum += uint64(x)
+	}
+	return sum
```
winder, line 68: Sorry for this comment, this obviously isn't critical.

Technically the test name isn't going to give you a uniformly distributed partition key, so you would need to hash this:
```suggestion
	h := fnv.New64a()
	h.Write([]byte(s))
	return h.Sum64()
```

This isn't perfect either, since there are tests with the same name in different packages:
```
~$ rg "func Test"| cut -f 2 -d ' '|sort|wc -l
1360

~$ rg "func Test"| cut -f 2 -d ' '|sort|uniq|wc -l
1344
```

cce: oh interesting, for a more even distribution you could get the filename & line number of the caller too by `runtime.Caller(1)`. Also, passing the `pc` value it returns to `runtime.FuncForPC` could get you a version of the calling function name and package name, here's an example: https://play.golang.org/p/tEW91-qEdqR

winder: @cce I was thinking about something similar, but wouldn't that give the line/file of the helper function instead of the test function? I guess hashing a random value would work too (unless the partitions should be deterministic).

cce: `Caller(N)` will go N levels up the stack, so it will get the test function instead of the helper function — Caller() and FuncForPC() are sometimes used in logging/metrics/tracing helper functions to grab names like this for capturing monitoring data.

Deterministic seems good IMO, especially for re-running tests

cce: I gave a try at implementing @winder's idea along with the `runtime.Caller(1)` in https://github.com/cce/go-algorand/commit/0cf3b2d1a97fba19fb7f225c17487a3b6758f209 CC @algonautshant

scripts/travis/before_build.sh
```diff
@@ -18,7 +18,9 @@ SCRIPTPATH="$( cd "$(dirname "$0")" ; pwd -P )"
 OS=$("${SCRIPTPATH}"/../ostype.sh)
 ARCH=$("${SCRIPTPATH}"/../archtype.sh)
 
-echo "Building libsodium-fork..."
-make crypto/libs/${OS}/${ARCH}/lib/libsodium.a
+if [ ! -f crypto/libs/${OS}/${ARCH}/lib/libsodium.a ]; then
+  echo "Building libsodium-fork..."
+  make crypto/libs/${OS}/${ARCH}/lib/libsodium.a
```
winder, line 23: How would we update these binaries when we change libsodium?

scripts/travis/build.sh
```diff
@@ -44,11 +44,12 @@ OS=$("${SCRIPTPATH}/../ostype.sh")
 ARCH=$("${SCRIPTPATH}/../archtype.sh")
 
 # Get the go build version.
-GOLANG_VERSION=$(./scripts/get_golang_version.sh)
-
-curl -sL -o ~/gimme https://raw.githubusercontent.com/travis-ci/gimme/master/gimme
-chmod +x ~/gimme
-eval "$(~/gimme "${GOLANG_VERSION}")"
+if [ -z "${NO_GIMME}" ]; then
```
winder, line 47: It looks like this is only used for windows, what wasn't working? According to the docs windows should be supported.

.circleci/config.yml
```diff
@@ -1,8 +1,412 @@
...
+      circleci_home:
+        type: string
+        default: "/home/circleci"
+    steps:
+      - run:
+          name: install deps
+          shell: bash.exe
+          command: |
+            choco install -y msys2 pacman make wget --force
+            choco install -y golang --version=1.14 --force
```
winder, line 66: The current version is 1.14.7

scripts/travis/build.sh
bricerisingalgorand: I can't remember the exact error, but when I was having issues with it on circleci's windows machines. I could go ahead and retract this change since we're planning to get windows working after merging these changes

scripts/travis/before_build.sh
bricerisingalgorand: Circleci is caching the libsodum builds based on the md5 sums of all files in libsodium fork, so if it ever changes a new build will happen and the cache will get updated

```
    - run:
          name: Get libsodium md5
          command: |
            mkdir -p tmp
            find crypto/libsodium-fork -type f -exec openssl md5 "{}" + > tmp/libsodium.md5
      - save_cache:
          key: 'libsodium-fork-{{ .Environment.CIRCLE_STAGE }}-{{ checksum "tmp/libsodium.md5" }}'
          paths:
            - crypto/libs
```

algobarb (approved):
I think it'll be good to merge it in soon and start testing the config file with other PRs. There are still a few areas that we can improve on in the next iterations, but this PR is already quite big and does build the tests on Ubuntu Linux and MacOS. I tested this code out on Thursday and it also works as a good base for testing nightly. I also checked that the pipeline run passsed in CircleCI today.

Makefile
```diff
@@ -23,6 +23,7 @@ BUILDBRANCH      := $(shell ./scripts/compute_branch.sh)
 CHANNEL          ?= $(shell ./scripts/compute_branch_channel.sh $(BUILDBRANCH))
 DEFAULTNETWORK   ?= $(shell ./scripts/compute_branch_network.sh $(BUILDBRANCH))
 DEFAULT_DEADLOCK ?= $(shell ./scripts/compute_branch_deadlock_default.sh $(BUILDBRANCH))
+export GOCACHE=$(SRCPATH)/tmp/go-cache
```
tsachiherman, line 26: I'm not sure that this is even needed. but if it does, the "touch" was somehow disappeared in this refactoring.

tsachiherman (changes requested):
We should not change the `msgp_gen_test.go` files directly, as these are auto-generated files. Instead, we have a project in `algorand/msgp` that does just that. That project would need to be adjusted to generate test files with the proposed changes. Once you've carries out the changes in the msgp library, you'll need to make a "release" for that, and update the release version in the go.mod/go.sum files.

my suggestion : 
1. rollback the changes in the `msgp_gen_test.go`
2. finish with the rest of the changes here.
3. make the changes in the `algorand/msgp` and create a new release.
4. make the msgp changes in the `algorand/go-algorand` to auto-generate the msgp.

Makefile
bricerisingalgorand: Thanks I just added it back. It was a while ago, but I think the touch was needed because of some sort of issue with go accessing empty folders on virtualized disks in docker containers.

.travis.yml
```diff
@@ -9,11 +9,11 @@ stages:
...
     if: NOT (branch =~ /^rel\//) AND type != pull_request
   - name: build_pr
-    if: type = pull_request
+    if: type = pull_request AND branch !~ /circleci/
   - name: build_release
-    if: branch =~ /^rel\// AND type != pull_request
+    if: branch =~ /^rel\// AND type != pull_request AND branch !~ /circleci/
   - name: deploy
-    if: branch =~ /^rel\// AND type != pull_request
+    if: branch =~ /^rel\// AND type != pull_request AND branch !~ /circleci/
```
algobarb, line 16: Was `AND branch !~ /circleci/` added for testing or is this necessary to add to the travis.yml file?

bricerisingalgorand: Oh yea thanks, that was for testing. I'll remove that now

.circleci/config.yml
bricerisingalgorand: Thanks! I fixed that

bricerisingalgorand:
> We should not change the `msgp_gen_test.go` files directly, as these are auto-generated files. Instead, we have a project in `algorand/msgp` that does just that. That project would need to be adjusted to generate test files with the proposed changes. Once you've carries out the changes in the msgp library, you'll need to make a "release" for that, and update the release version in the go.mod/go.sum files.
> 
> my suggestion :
> 
> 1. rollback the changes in the `msgp_gen_test.go`
> 2. finish with the rest of the changes here.
> 3. make the changes in the `algorand/msgp` and create a new release.
> 4. make the msgp changes in the `algorand/go-algorand` to auto-generate the msgp.

Thanks! I made the PR to `algorand/msgp` here

.circleci/config.yml
```diff
@@ -1,8 +1,408 @@
...
+            GOLANG_VERSION=$(./scripts/get_golang_version.sh)
+            eval "$(~/gimme "${GOLANG_VERSION}")"
+            scripts/configure_dev.sh
+            scripts/buildtools/install_buildtools.sh -o "gotest.tools/gotestsum"
+            PACKAGES="$(go list ./... | grep -v /go-algorand/test/)"
+            export PACKAGE_NAMES=$(echo $PACKAGES | tr -d '\n')
+            export PARTITION_TOTAL=$CIRCLE_NODE_TOTAL
+            export PARTITION_ID=$CIRCLE_NODE_INDEX
+            export PARALLEL_FLAG="-p 1"
+            gotestsum --format pkgname --junitfile /tmp/results/<< parameters.result_subdir >>/results.xml --jsonfile /tmp/results/<< parameters.result_subdir >>/testresults.json -- --tags "sqlite_unlock_notify sqlite_omit_load_extension" -short -race -timeout 1h -coverprofile=coverage.txt -covermode=atomic -p 1 $PACKAGE_NAMES
```
winder, line 173: I think these scripts should be rolled into the CI scripts at `scripts/travis`

There are subtle differences between what you have here and the scripts.

Also, there is a lot of code duplication, for example in the current scripts gotestsum is declared once in the Makefile, now it's installed and called in many places.

.circleci/config.yml
```diff
@@ -1,8 +1,408 @@
...
+            GOLANG_VERSION=$(./scripts/get_golang_version.sh)
+            eval "$(~/gimme "${GOLANG_VERSION}")"
+            scripts/configure_dev.sh
+            scripts/buildtools/install_buildtools.sh -o "gotest.tools/gotestsum"
+            export ALGOTEST=1
+            export SHORTTEST=-short
+            export TEST_RESULTS=/tmp/results/<< parameters.result_subdir >>
+            export PARTITION_TOTAL=$CIRCLE_NODE_TOTAL
+            export PARTITION_ID=$CIRCLE_NODE_INDEX
+            test/scripts/run_integration_tests.sh
```
winder, line 225: Call `make integration` so that the `GOTESTCOMMAND` variable is initialized. Also, the go e2e tests wont be captured (missing junit option)

.circleci/config.yml
```diff
@@ -1,8 +1,408 @@
...
+            - tmp/go-cache
+
+  upload_coverage:
+    description: Collect coverage reports and upload them
+    steps:
+      - run:
+          name: Upload Coverage Reports
+          no_output_timeout: 10m
+          command: |
+            scripts/travis/upload_coverage.sh || true
```
winder, line 191: Did you try disabling the travis upload to make sure all partitioned coverage files make it to codecov?

winder:
What are the expectations for this PR? Everything looks like it should work. There's no integration with our current CI scripts and lots of copied code, so it still seems like a prototype.

daemon/algod/api/server/router_test.go
```diff
@@ -178,5 +178,6 @@ func (s *TestSuite) TestGetTransactionByID() {
 	assert.Equal(s.T(), callsBefore+1, s.calls)
 }
 func TestTestSuite(t *testing.T) {
+
```
winder, line 181: Do the tests in here need `partitiontest.PartitionTest(s.T())`?

daemon/algod/server_test.go
```diff
@@ -23,6 +23,7 @@ import (
 	"net"
 	"testing"
 
+	"github.com/algorand/go-algorand/test/partitiontest"
 	"github.com/stretchr/testify/require"
 )
```
winder, line 28:
```suggestion
	"github.com/stretchr/testify/require"
	
	"github.com/algorand/go-algorand/test/partitiontest"
)
```

daemon/kmd/config/config.go
```diff
@@ -54,6 +55,11 @@ type SQLiteWalletDriverConfig struct {
 	ScryptParams ScryptParams `json:"scrypt"`
 }
 
+// LedgerWalletDriverConfig is configuration specific to the LedgerWalletDriver
+type LedgerWalletDriverConfig struct {
+	Disable bool `json:"disable"`
```
winder, line 60: How was this disabled previously?

data/transactions/logic/assembler_test.go
```diff
@@ -23,6 +23,7 @@ import (
 	"strings"
 	"testing"
 
+	"github.com/algorand/go-algorand/test/partitiontest"
 	"github.com/stretchr/testify/require"
```
winder, line 27:
```suggestion
	"github.com/stretchr/testify/require"

	"github.com/algorand/go-algorand/test/partitiontest"
```

test/partitiontest/filtering.go
```diff
@@ -0,0 +1,55 @@
+// Copyright (C) 2019-2021 Algorand, Inc.
```
winder, line 1: Rename this to `filtering_test.go` so that it isn't included in production binaries.

scripts/travis/build.sh
winder: I don't really mind the flag, but maybe name it something like `SKIP_GO_INSTALL`

.circleci/config.yml
winder: Better would be to use update `scripts/travis/integration_test.sh`

test/scripts/e2e.sh
```diff
@@ -124,6 +126,12 @@ duration "serial client runners"
 
 deactivate
 
+fi # if $SKIP_E2E_SUBS = ""
+
+if [ "${E2E_SUBS_ONLY}" != "" ]; then
```
winder, line 131: Choose a `SKIP` or `ENABLE` pattern, not both. I think `ENABLE_E2E_SUBS`, `ENABLE_E2E_GO` and `ENABLE_E2E_EXPECT` would be easier for someone new to these scripts to understand, `SKIP_*` might be easier to integrate.

.circleci/config.yml
winder: Maybe `NO_BUILD=true` would help? I saw that flag below, not sure when it helps.

test/scripts/e2e_go_tests.sh
```diff
@@ -91,7 +91,14 @@ fi
 
 echo "PARALLEL_FLAG = ${PARALLEL_FLAG}"
 
-if [ "${#TESTPATTERNS[@]}" -eq 0 ]; then
+if [ "$USE_CIRCLECI_SPLIT" != "" ]; then
```
winder, line 94: Why use circleci tests split instead of the partitioning?

algobarb:
> What are the expectations for this PR? Everything looks like it should work. There's no integration with our current CI scripts and lots of copied code, so it still seems like a prototype.
> 
> This would be significantly easier to review if the test partitioning changes were made separately. Could add those in a separate PR so it's easier to see the scripting changes?

I think one of the expectation is to merge this PR into master so that we can start getting more branches running CircleCI. This is our last week on our extended trial with Circle CI, so we want to get an idea of what regular CircleCI usage would look like in terms of costs and regular usage. Since this PR absorbed the hackathon work, it will be good if it can be merged so that others can submit smaller PRs afterwards to improve on the script. It is a lot of changed files at once though...

.circleci/config.yml
bricerisingalgorand: oh good call, I just checked that circleci was showing up here https://codecov.io/gh/algorand/go-algorand/commit/00f33c0090e99b2f00c6ab36a6087c50cb7b9bd3/build

I'll test that too

daemon/kmd/config/config.go
bricerisingalgorand: I didn't see a way to disable them before. We only had to do this for circle because they don't allow you to scan usbs on the osx agents

test/scripts/e2e_go_tests.sh
bricerisingalgorand: I'm not actually sure what this change is, could one of you two explain? @algonautshant @cce

daemon/kmd/config/config.go
winder: If you needed this, I guess it means kmd crashes if HID is not supported. We should probably fix that instead of disabling KMD on circle. Or at least create an issue for fixing the issue and removing this workaroud.

test/scripts/e2e_go_tests.sh
cce: oh, this is from when I was evaluating the "circleci tests split" runner before Shant's partitioning stuff was ready. you can remove these lines to get rid of it .. as you noticed the `elif [ "${#TESTPATTERNS[@]}" -eq 0 ]; then` is the regular code path we have used and the partitioning stuff is now using.

winder: You would still need to keep the `gotestsum` call, or set `GOTESTCOMMAND` to include the junit stuff

daemon/kmd/config/config.go
cce: While digging into it with @bricerisingalgorand and it seems like without forking the https://github.com/karalabe/hid project and modifying it to use a new fork of the https://github.com/signal11/hidapi project it would be hard to fix the issue, the crashes are happening deep in some C code embedded in karalabe/hid. Some related tickets
https://github.com/signal11/hidapi/issues/205
https://github.com/signal11/hidapi/issues/337

In general, the ability to disable USB probing for a ledger via configuration seemed like a useful feature anyway, and we haven't seen this crash from users, so it seemed like a pretty specific corner case related to the way the CI VMs are set up.

winder: yikes, too bad. I was hoping it would be a matter of checking for this error:
```
func LedgerEnumerate() ([]hid.DeviceInfo, error) {
	if !hid.Supported() {
		return nil, fmt.Errorf("HID not supported")
	}
```

daemon/algod/api/server/router_test.go
cce: I think just this TestTestSuite function would need it, the other functions are all called from this test via the testify/suite feature

.circleci/config.yml
cce: It seems hard to maintain scripts that handle both Travis and Circle runs, especially since we are basically trying to cut up what Travis does. Re: code duplication, that is a matter of personal preference, but I will point out that if later we remove Travis automation, there will be less duplication, and a lot of this can migrate to CircleCI-specific scripts. However IMO total DRY-proofing this PR seems like a big request given the intermediate state we want to be in running Travis and Circle side-by-side.

.circleci/config.yml
cce: For some background, I was purposefully trying to avoid updating any of the Travis scripts when I worked on this, so that they could run side-by-side without disrupting the Travis system.

cce:
> There's no integration with our current CI scripts and lots of copied code ...

I would disagree with this characterization. Most of the shell snippets in the `run` sections are just setting up environment variables to call the current CI scripts — configure_dev.sh, install_buildtools.sh, run_integration_tests.sh. The only copied code besides environment setup IMO is the line where it calls gotestsum directly given a list of packages.

test/scripts/e2e_go_tests.sh
cce: the entire `if USE_CIRCLECI_SPLIT` section can be removed. This file doesn't need to be changed for this PR.

.circleci/config.yml
winder: I disagree, these scripts are already platform agnostic. We've been gradually removing the travis specific aspects for a long time. Most of the heavy lifting happens in the Makefile anyway.

For example, I think this section could be replaced with:
```
            export KMD_NOUSB=True
            export PARTITION_TOTAL=$CIRCLE_NODE_TOTAL
            export PARTITION_ID=$CIRCLE_NODE_INDEX
            scripts/travis/test.sh
```

To take it one more step you could set those 3 environment variables in `test.sh`, I don't think that part is as important.

cce: Moving all the environment things into "if CIRCLECI" sections is definitely something you could do to bring the remaining stuff into the existing scripts.. I haven't tested whether Makefile dependency resolution works with the kind of workspace-passing between jobs/VMs that Circle and Github Actions provide though, because file timestamps would change, that seems potentially tricky.

daemon/kmd/config/config.go
cce: sadly supported is just a `return true` https://github.com/karalabe/hid/blob/master/hid_enabled.go#L61-L66 based on whether you have the right platform-specific build tag enabled or not. the crash was coming from hid.Enumerate

test/scripts/e2e_go_tests.sh
cce: @bricerisingalgorand could you please apply this to your branch? https://github.com/bricerisingalgorand/go-algorand/commit/5a1b71eec5b229ba906442bafce8fcc5cbc3e330

test/scripts/e2e_go_tests.sh
```diff
@@ -91,14 +91,7 @@ fi
 
 echo "PARALLEL_FLAG = ${PARALLEL_FLAG}"
 
-if [ "$USE_CIRCLECI_SPLIT" != "" ]; then
-    PARALLEL_FLAG="-p 1"
-    PACKAGES="$(go list ./... | circleci tests split --split-by=timings --timings-type=classname)"
-    PACKAGE_NAMES=$(echo $PACKAGES | tr -d '\n')
-    echo "Testing the following packages:"
-    echo $PACKAGE_NAMES
-    gotestsum --format pkgname --junitfile $TEST_RESULTS/results.xml --jsonfile $TEST_RESULTS/testresults.json -- ${RACE_OPTION} ${PARALLEL_FLAG} -timeout 1h -v ${SHORTTEST} ${PACKAGE_NAMES}
```
winder: Unfortunately, I don't think this change is quite right. `GOTESTCOMMAND` isn't being set and it defaults to `go test`, so after this change we'll revert back to `go test`.  If we switch to calling `make integration` instead of no longer be setting `./test/scripts/run_integration_tests.sh`, then `GOTESTCOMMAND` will be set but wont include `--junitfile $TEST_RESULTS/results.xml --jsonfile $TEST_RESULTS/testresults.json`.

cce: This is some dead code and so this code path was not being exercised because USE_CIRCLECI_SPLIT is not set anywhere. This commit should have no effect on the tests we're running in CircleCI.

I'm not sure how running make targets will work after attaching a workspace saved by a different job.

algobarb (approved):

cce (approved):

bricerisingalgorand:
Could you update your review if you think it looks ok @tsachiherman?

winder (approved):
Approved pending follow-up tickets (my priorities in parens):
1) verify codecov report still working and correct (high)
2) router_test.go test suite isn't partitioned (high)
3) unify gotestsum configuration (medium)
4) consistent SKIP_E2E_SUBS / E2E_SUBS_ONLY convention - `SKIP_` or `ENABLE_`, not both. (medium)
5) `NO_GIMME` -> `SKIP_GO_INSTALLATION` (medium)
6) verify that go e2e tests are included in reporting, I think that gotestsum may not be properly initialized (medium)
7) unify script changes (medium-low)
8) fix import orders to match our convention (low)

PR #2445, opened 2021-07-06
txsync: bitmask unit tests

algonautshant:
## Summary
In this pr:
- Moved existing bitmask tests from txnsync/txngroups_test.go to txnsync/bitmask_test.go
- Refactored tests and added missing test coverage
- Introduced more exception handling in the code: check for array bounds and return errors to prevent panic

## Test Plan
Tests are added in this pr.


txnsync/bitmask.go
```diff
@@ -110,9 +119,10 @@ func (b *bitmask) trimBitmask(entries int) {
 	default:
 	}
 
-	*b = bytes.TrimRight(*b, string(0))
+	*b = bytes.TrimRight(*b, fmt.Sprintf("%c", 0))
```
tsachiherman, line 122: this would have undesired performance impact. try to use "\x00" instead.

algonautshant: Fixed.

txnsync/bitmask.go
```diff
@@ -21,29 +21,51 @@ import (
...
 }
 
-func (b *bitmask) trimBitmask(entries int) {
+// trimBitmask compresses the bitmask into one of the 4 types:
+// type 0: intput bitmask bit pos x b -> output bitmask bit pos x b
+// type 1: intput bitmask bit pos x b -> output bitmask bit pos x !b
+// type 2: stores the positions of bits where b = 1
+//         intput bitmask first b=1 pos A, second b=1 pos B, ...
+//         output bitmask byte 2,A/256,A%256,(B-A)/256,(B-A)%256,...
+// type 3: same as type 2, but stures the positons where b = 0
```
tsachiherman, line 62: typos: `same as type 2, but stures the positons where b = 0` -> `same as type 2, but stores the positions where b = 0`

txnsync/bitmask.go
```diff
@@ -21,26 +21,41 @@ import (
...
+// entryExists converts the bitmask to type 0 (if not already)
+func (b *bitmask) entryExists(index int, entries int) bool {
 	if len(*b) == 0 {
 		return false
 	}
 	if (*b)[0] != 0 {
-		b.expandBitmask(entries)
+		err := b.expandBitmask(entries)
+		if err != nil {
+			return false
```
tsachiherman, line 45: if we've detected this issue, we shouldn't keep it for ourselves. we need to return it upstream to the caller and fail the decoding of this message.

txnsync/bitmask_test.go
```diff
@@ -0,0 +1,172 @@
...
+
+func TestBitmaksTypeX(t *testing.T) {
+	b := make(bitmask, bytesNeededBitmask(80))
+	b[0] = 4
+	require.Equal(t, b.iterate(0, 0, nil), errInvalidBitmaskType)
+}
+
+func trimIterateHelper(t *testing.T, setBits []int) {
+	b := make(bitmask, bytesNeededBitmask(80))
+	entries := 80
```
nicholasguoalgorand, line 88: reverse these two lines to only have one const

txnsync/bitmask_test.go
```diff
@@ -0,0 +1,172 @@
...
+	var errTestError = errors.New("some error")
+	errorAfter := 0
+	errfunc := func(i int, index int) error {
+		if index > errorAfter {
+			return errTestError
+		}
+		return nil
+	}
+
+	require.Equal(t, errTestError, b.iterate(entries, len(setBits), errfunc))
```
nicholasguoalgorand, line 107: what is this testing?

txnsync/bitmask_test.go
```diff
@@ -0,0 +1,172 @@
...
+		entryExist[setIndex] = true
+		return nil
+	})
+
+	s := 0
+	for i := 0; i < entries; i++ {
+		exists := entryExist[i]
+		if s < len(setBits) && i == setBits[s] {
+			require.True(t, exists)
+			require.True(t, iterated[i], i)
```
nicholasguoalgorand, line 122: why are we checking for `iterated` as well as `entryExists`?

nicholasguoalgorand (commented):
Had a few nits and questions about the tests. Please take a look.

txnsync/bitmask.go
```diff
@@ -249,11 +205,12 @@ func (b *bitmask) Iterate(entries int, maxIndex int, callback func(int, int) err
 			index++
 		}
 	default:
-		return errors.New("invalid bitmask type")
+		return errInvalidBitmaskType
 	}
 	return nil
 }
 
+// bytesNeededBitmask returns the number of bytes needed to store enteris bits.
```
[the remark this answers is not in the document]
algonautshant: typo

txnsync/bitmask_test.go
algonautshant: done.

txnsync/bitmask_test.go
algonautshant: Testing the case where the callback function returns an error.

txnsync/bitmask_test.go
algonautshant: You are right to question this. These are redundant.

txnsync/bitmask.go
```diff
@@ -21,26 +21,26 @@ import (
...
-		b.expandBitmask(entries)
-	}
-	byteIndex := index/8 + 1
-	return byteIndex < len(*b) && ((*b)[byteIndex]&(1<<(index%8)) != 0)
-}
-
+// trimBitmask compresses the bitmask into one of the 4 types:
+// type 0: intput bitmask bit pos x b -> output bitmask bit pos x b
+// type 1: intput bitmask bit pos x b -> output bitmask bit pos x !b
+// type 2: stores the positions of bits where b = 1
```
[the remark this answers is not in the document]
algonautshant: The method does not check for the original bitmask type. One less check for performance. 
However, the code only uses this on type 0 bitmask.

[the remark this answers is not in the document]
nicholasguoalgorand: I was referring to the comment you wrote on the following two lines. I'm not sure if they are helpful/accurate, can we remove them?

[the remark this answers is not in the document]
algonautshant: Helpful: I believe it is helpful. Any comment which is accurate is helpful. 
Accurate: what do you think is not accurate? Which argument?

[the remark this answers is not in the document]
nicholasguoalgorand: I don't understand what `first b=1 pos A` means. I assume it means the first byte is for position A, which is inaccurate. If I'm wrong please update the comment to make the meaning more clear.

Next, saying `output bitmask byte 2,A/256,A%256,(B-A)/256,(B-A)%256` is not helpful since that simply rewrites the code in a comment and doesn't explain anything. A better comment would say something regarding using adjacent differences of indices where the bit is set to 1.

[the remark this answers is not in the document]
algonautshant: I hope this is now clearer. Updated.

txnsync/bitmask.go
```diff
@@ -21,26 +21,26 @@ import (
...
-	byteIndex := index/8 + 1
-	return byteIndex < len(*b) && ((*b)[byteIndex]&(1<<(index%8)) != 0)
-}
-
+// trimBitmask compresses the bitmask into one of the 4 types:
+// type 0: intput bitmask bit pos x b -> output bitmask bit pos x b
+// type 1: intput bitmask bit pos x b -> output bitmask bit pos x !b
+// type 2: stores the positions of bits where the bit value is 1
+//         intput bitmask first bit 1 at pos A, second bit 1 at pos B, ...
+//         output bitmask stores A, B, ...
```
nicholasguoalgorand, line 42: The output bitmask doesn not store A, B, C....
It stores A, B-A, C-B, etc

txnsync/bitmask.go
```diff
@@ -21,26 +21,26 @@ import (
...
-	}
-	if (*b)[0] != 0 {
-		b.expandBitmask(entries)
-	}
-	byteIndex := index/8 + 1
-	return byteIndex < len(*b) && ((*b)[byteIndex]&(1<<(index%8)) != 0)
-}
-
+// trimBitmask compresses the bitmask into one of the 4 types:
+// type 0: intput bitmask bit pos x b -> output bitmask bit pos x b
```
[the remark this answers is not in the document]
algonautshant: Thanks. fixed.

PR #2494, opened 2021-07-07
Fix 10 minute timeout in travis. (Revert parts of #2324)

winder:
## Summary
We started seeing the 10-minute timeout error on travis after the recent changes to where travis_retry gets called.

This means travis_wait is still needed. travis_wait and travis_retry don't play well together on the ephemeral build machine, so we basically need to rollback the entire change in #2324 

Note: while making this change I noticed that we don't use travis_retry for all of the build_test.sh / integration_test.sh entries. Not sure why but I left them as they were before #2324 

## Test Plan
CI

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,7 +28,7 @@ jobs:
     - stage: build_commit
       os: linux
       script:
-        - scripts/travis/build_test.sh
+        - travis_retry scripts/travis/build_test.sh
 
     - stage: build_pr
       os: linux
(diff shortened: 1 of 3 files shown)


tsachiherman (commented):
I think that the changes to `build_test.sh` would suffice.

winder:
@tsachiherman I was thinking the same thing so I checked the original issue to see if there was a comment about why I included that originally:

https://github.com/algorand/go-algorand/pull/2324#pullrequestreview-691272612

tsachiherman:
> @tsachiherman I was thinking the same thing so I checked the original issue to see if there was a comment about why I included that originally:
> 
> [#2324 (review)](https://github.com/algorand/go-algorand/pull/2324#pullrequestreview-691272612)

hmm. Yes, I missed that. I guess that a better solution than would be to do 
```bash
 "${SCRIPTPATH}/travis_retry.sh" ${SCRIPTPATH}/travis_wait.sh" 90 "${SCRIPTPATH}/test.sh"
```

so that we would have the retry outside of the wait.

tsachiherman (approved):

PR #2496, opened 2021-07-07
Disable 'make fulltest' in travis builds.

winder:
## Summary
The nightly release hasn't been going out because of testing issues. To unblock things, use the same test profile as PRs until we can finish fixing tests.

## Test Plan
CI

### DIFF
--- a/scripts/travis/run_tests.sh
+++ b/scripts/travis/run_tests.sh
@@ -13,8 +13,9 @@ if [ "${BUILD_TYPE}" = "integration" ]; then
     fi
     export SHORTTEST 
     "${SCRIPTPATH}/travis_retry.sh" make integration
-elif [ "${TRAVIS_EVENT_TYPE}" = "cron" ] || [[ "${TRAVIS_BRANCH}" =~ ^rel/ ]]; then
-    "${SCRIPTPATH}/travis_retry.sh" make fulltest -j2
+# TODO: re-enable fulltest
+#elif [ "${TRAVIS_EVENT_TYPE}" = "cron" ] || [[ "${TRAVIS_BRANCH}" =~ ^rel/ ]]; then
+#    "${SCRIPTPATH}/travis_retry.sh" make fulltest -j2
 else
     "${SCRIPTPATH}/travis_retry.sh" make shorttest -j2
 fi


onetechnical (approved):
I'm a little nervous that we'll forget to re-enable this, but let's give it a whirl.

winder:
> I'm a little nervous that we'll forget to re-enable this, but let's give it a whirl.

Let's wait for Tsachi to weigh in

winder:
Here's the followup ticket, I'm not sure what epic to tag it with: https://github.com/algorand/go-algorand-internal/issues/1380

tsachiherman:
I'm going to propose something that is a bit weird; why don't we flip the day and night, and run the nightly build on a PR ? That way, we can explicitly test the nightly build outcome on a PR that is easily being invoked.

As an alternative, for nightly builds, we can run it with `|| true` so that it won't fail the travis build on error ( if that's really a blocker ).

onetechnical:
> I'm going to propose something that is a bit weird; why don't we flip the day and night, and run the nightly build on a PR ? That way, we can explicitly test the nightly build outcome on a PR that is easily being invoked.
> 
> As an alternative, for nightly builds, we can run it with `|| true` so that it won't fail the travis build on error ( if that's really a blocker ).

I think that's what Will proposes with https://github.com/algorand/go-algorand-internal/issues/1380 so we can prioritize _a little_ later, but still get the nightlies to pass and update DevNet. The other way is preferred, but would make whoever next needed to pass a PR have to debug the tests. (I'm fine with that, but it does kind of make everyone have to stop and fix it.)

winder:
I'm not sure `|| true` would work here, the build would probably still end up failing because of a timeout.

tsachiherman:
> > I'm going to propose something that is a bit weird; why don't we flip the day and night, and run the nightly build on a PR ? That way, we can explicitly test the nightly build outcome on a PR that is easily being invoked.
> > As an alternative, for nightly builds, we can run it with `|| true` so that it won't fail the travis build on error ( if that's really a blocker ).
> 
> I think that's what Will proposes with [algorand/go-algorand-internal#1380](https://github.com/algorand/go-algorand-internal/issues/1380) so we can prioritize _a little_ later, but still get the nightlies to pass and update DevNet. The other way is preferred, but would make whoever next needed to pass a PR have to debug the tests. (I'm fine with that, but it does kind of make everyone have to stop and fix it.)

Oh - so it's about Devnet. I did not realize that.. 
In that case, I'm ok with this change, but we need to get back to run the nightly builds as well.
Can we modify the travis.yml file so that we will run it with both short and long tests, and have the travis.yml allow failures on the long runs ?

winder:
I'm going to close this for now so that we don't ignore the issue. Maybe we can disable individual tests that are still failing.

Issue #2366, opened 2021-07-01
Add some more metrics to track transaction pool stats


grakshith:
## Summary

Create new metric(s) to track transaction pool size of nodes.

## Scope/Requirements

Add new metrics to `transactionPool.go` and visualize it with algonet.

## Urgency/Relative Priority

Medium priority.

PR #2498, opened 2021-07-08
Add metrics to track transaction pool stats and connected peers

grakshith:
## Summary

Adds some metrics to track the expired transactions per round and the connected peers. The connected peers information wil be used to build a topology of the network.

## Test Plan

Tested on algonet that these metrics show up on Prometheus


grakshith:
Closes #2366

network/wsNetwork.go
```diff
@@ -1731,6 +1735,41 @@ func (wn *WebsocketNetwork) sendPeerConnectionsTelemetryStatus() {
...
+// function to send periodic connected peer metrics
+// the peer information is stored as the metric's labels
+func (wn *WebsocketNetwork) sendPeerConnectionsMetricStatus() {
+	ticker := time.NewTicker(10 * time.Second)
+	defer ticker.Stop()
+	for {
+		select {
+		case <-ticker.C:
+			outPeerMap, inPeerMap := wn.updateConnectedPeersLabelMaps()
+			incomingPeerAddr.Set(1, inPeerMap)
```
brianolson, line 1747: So now a busy peer with 1000 incoming connections has 1000 lines of /metrics output instead of the ~40 it has now? Are we sure we really want this?

tsachiherman: I'm not sure that we want this solution as is.. but I think that we should hold off until @grakshith has a working end-to-end solution..
Once we have it, he will be able to remove the code-portions that aren't really needed.

grakshith: The connected peers are already being sent as telemetry events. But for some reason, I couldn't see them on the ephemeral telemetry elasticsearch. I added these metrics for now, to help build the monitoring system, and as @tsachiherman said, I will remove the bits not needed once I have the complete solution. I will convert the PR to draft.

PR #2502, opened 2021-07-08
testing: avoid division by zero during TestBasicCatchpointWriter

tsachiherman:
## Summary

The `randomFullAccountData` method was dividing by `lastCreatableID` which could be zero.
The probability for that is pretty slim, but given that it was found during a travis run, we should fix it.

## Test Plan

This is a test fix.


algorandskiy (commented):
Looks good but the PR has 2045 commits, needs rebasing

algorandskiy (approved):

algonautshant (approved):
LGTM

PR #2503, opened 2021-07-08
testing: avoid running TestAgreementSynchronous10 on nightly build

tsachiherman:
## Summary

The test TestAgreementSynchronous10 is failing. We should be fixing it, but until we do that, I'm going to disable it so it won't mask other issues.

## Test Plan

Disable the test.

### DIFF
--- a/agreement/service_test.go
+++ b/agreement/service_test.go
@@ -948,6 +948,7 @@ func TestAgreementSynchronous5(t *testing.T) {
 }
 
 func TestAgreementSynchronous10(t *testing.T) {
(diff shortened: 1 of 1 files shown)


winder (dismissed):

onetechnical (approved):
Ah - thanks for the rebase. Looks fine, though I want to be sure we remember to re-enable and fix.

winder (approved):

tsachiherman:
> Ah - thanks for the rebase. Looks fine, though I want to be sure we remember to re-enable and fix.

Created a ticket to track it.

PR #2505, opened 2021-07-09
fix regex for matching charset of valid DNS hosts, add test

brianolson:
## Summary

A regex was trying to match valid DNS names but missed the '-' char. Fix.

## Test Plan

Added unit test case.

### DIFF
--- a/network/wsNetwork.go
+++ b/network/wsNetwork.go
@@ -1913,8 +1913,8 @@ var errBcastQFull = errors.New("broadcast queue full")
 
 var errURLNoHost = errors.New("could not parse a host from url")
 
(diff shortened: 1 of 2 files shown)


tsachiherman (approved):
I think that PR is good, since it addresses a concrete issue.
However, the host name in the regex is still wrong from spec standpoint : the spec allows for unicode host names, which aren't supported here.

brianolson:
'unicode in DNS' gets squashed via Punycode to fit into the old charset
`Bücher.example` -> `xn--bcher-kva.example `
https://en.wikipedia.org/wiki/Internationalized_domain_name

PR #2508, opened 2021-07-09
metrics Gauge cleanup

brianolson:
## Summary

Gauge no longer has a timeout.

Gauge that didn't change used to disappear from results; this could make a non-zero value appear zero, confusing metrics.

gaugeCommon.go split was confusing, merge back into gauge.go

## Test Plan

Refactor continue to pass existing tests.


util/metrics/gauge.go
```diff
@@ -139,26 +142,11 @@ func (cv *gaugeValues) createFormattedLabel() {
 	cv.formattedLabels = buf.String()[1:]
 }
 
-// filterExpiredMetrics scans the gauge.valuesIndices map and removing all the gauges that
-// haven't been updated in the past <maxMetricRetensionDuration> units of time.
-func (gauge *Gauge) filterExpiredMetrics() {
```
winder, line 144: Removing this would be a problem if someone decides to use the label feature. It looks like that allows for creating dynamic guages based on user-defined keys. We don't currently use the dynamic labels feature in the networking code. In theory, someone might add a gauge by session ID (i.e. ping per connection). At that point removing this function would become be issue since it would grow unbounded.

A couple of possible solutions:
1) Remove the `labels` feature (it seems to only be used by `segment.go:EnterSegment` and that function is not called anywhere).
2) Return from this function immediately when `len(valueIndices) <= 1`
3) Set `maxMetricRetensionDuration` to something very large in `config.json`

winder (changes requested):
Mostly good, I just spot one possible problem

util/metrics/gauge.go
brianolson: `MakeSegement` `EnterSegment` `LeaveSegment` never occur in the code. It's completely unused and I'm tempted to delete segment.go entirely.
`labels` is also unused, I checked all uses of Gauge.Set() and labels is always nil.

Since there's no problem with the code as it is but only as it might be, I'd like to finish this bugfix and file a new issue for further util/metrics cleanup of dead code.

winder (approved):
Approved pending a new ticket is created. In the new ticket could you mention that `StringGauge` has the same problem?

algorandskiy (approved):

PR #2509, opened 2021-07-09
testing: Fix random failure in TestWebsocketNetworkPrioLimit

tsachiherman:
## Summary

The peers array is modified when adding/removing entries from it. When that does happen, we increase the `peersChangeCounter`, so that the broadcast method would know that it's peers list need to be refreshed.
The said update was missing from `prioTracker.setPriority`, which was causing the issue.

## Test Plan

The existing test provide sufficient coverage ( as it was able to detect the issue ).

### DIFF
--- a/network/netprio.go
+++ b/network/netprio.go
@@ -18,6 +18,7 @@ package network
 
 import (
 	"container/heap"
+	"sync/atomic"
 
 	"github.com/algorand/go-algorand/data/basics"
 	"github.com/algorand/go-algorand/protocol"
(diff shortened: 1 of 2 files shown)


zeldovich (approved):
Looks good to me.

Consider removing the netprio stuff altogether.  We added it shortly before launch as a potential defense against relay DoS attacks.  But I don't think we've ever actually used it.  Furthermore, our relay topology has changed significantly since launch, and I think we are unlikely to use this DoS defense mechanism.

PR #2513, opened 2021-07-11
txnsync: use signed transaction group hash

tsachiherman:
## Summary

Use signed transaction group hash instead of the hash of the first signed transaction.
This is needed in order to synchronize transactions that haven't been verified.

This PR depends on https://github.com/algorand/go-algorand/pull/2515

## Test Plan

Test manually.


txnsync/encodedgroupstypes.go
```diff
@@ -35,7 +35,7 @@ const maxAddressBytes = maxEncodedTransactionGroupEntries * crypto.DigestSize
 var errInvalidTxType = errors.New("invalid txtype")
 
 //msgp:allocbound txnGroups maxEncodedTransactionGroupEntries
-type txnGroups []transactions.SignedTxn
+type txnGroups transactions.SignedTxnSlice
```
nicholasguoalgorand, line 38: How will the allocbound work with the above `//msgp:allocbound txnGroups maxEncodedTransactionGroupEntries`?

data/transactions/transaction.go
```diff
@@ -660,14 +660,25 @@ type SignedTxGroup struct {
...
 	EncodedLength int
 }
 
+// SignedTxnSlice is a slice of SignedTxn(s), allowing us to
+// easily define the ID() function.
+//msgp:allocbound SignedTxnSlice config.MaxTxGroupSize
+type SignedTxnSlice []SignedTxn
+
+// ID calculate the hash of the signed transaction group.
+func (s SignedTxnSlice) ID() Txid {
```
nicholasguoalgorand, line 676: Is this supposed to be the same way we use to find the txgroup hash? If not, should it be?

nicholasguoalgorand (commented):
Had a couple comments/questions

data/transactions/transaction.go
tsachiherman: no, it's *supposed* not to be. You see, when finding the regular group hash, we want to calculate the hash without all the signatures, arguments, etc. However, here, we're adding this in order to exchange the transaction groups *before* validating them. That means that we must make sure that no one can provide us with false representation of a transaction group. Hashing it provides that guarantee.

tsachiherman: If you can think of a better naming for that.. that might be a good thing to do.

txnsync/encodedgroupstypes.go
tsachiherman: When creating the msgp decoder, it makes sure that the decoded buffer has less(or equal) number of entries to the one specified here. It's equivalent to using struct tags, which cannot be used here ( since it's not a struct.. )

nicholasguoalgorand: I see that we have an alloc bound here as well as for declaring the `transactions.SignedTxnSlice`, will it just apply both bounds?

data/transactions/transaction.go
nicholasguoalgorand: Okay, this makes sense then. I'm wondering, is there some way to convey that the txns in the slice should correspond to the same group? Or is that not necessary?

txnsync/encodedgroupstypes.go
tsachiherman: If you'll look on the generated code, you'll see that the decoder would now perform the `txnGroups` decoding without any bounds checking by deferring the decoding to the `transactions.SignedTxnSlice` decoder ( that does the bounds checking )

data/transactions/transaction.go
tsachiherman: Interesting idea, but I don't think that the two are strongly correlated.

nicholasguoalgorand: I see, I had this idea since in the comment it says ```// ID calculate the hash of the signed transaction group.```

nicholasguoalgorand (approved):

brianolson (approved):

PR #2515, opened 2021-07-11
testing: improve testing of alloc bounded slices

tsachiherman:
## Summary

This change is needed in order to support codec types such as the following:
```golang
//msgp:allocbound typeA 16
type typeA []typeB
```
since we want the codec object randomizer to recognize that `typeA` has a allocbound defined, and use that bound.
( this would be used in the feature/txnsync branch, but the change doesn't really related to any of the other changes in the feature/txnsync branch and would be a good change regardless )

## Test Plan

This is a test.


nicholasguoalgorand (approved):
Looks good.

PR #2516, opened 2021-07-11
testing: make TestPeersDownloadFailed and TestHistoricData predictable

tsachiherman:
## Summary

Both the `TestPeersDownloadFailed` as well as `TestHistoricData` were randomly failing, as they were depending on a "genuine" random distribution of the random function in order to succeed. When that doesn't happen, they were failing.

This PR make sure to "bake-in" the random seed so that the tests are repeatable and a success is repeatable as well.

## Test Plan

These are testings.

### DIFF
--- a/catchup/peerSelector_test.go
+++ b/catchup/peerSelector_test.go
@@ -17,7 +17,9 @@
 package catchup
 
 import (
+	"bytes"
 	"context"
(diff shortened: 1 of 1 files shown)


algonautshant (approved):
Look great. 
Thanks for fixing this.

Issue #2347, opened 2021-06-30
ABI - AVM Support - Alternative substring


chaihoang:
- Have an alternative substring method that takes parameters start index and length as opposed to end index (consider whether these should be stacked parameters or immediates)

jannotti:
Today, we have `substring` which takes two immediate values, a start and end, and extracts those bytes.  Since immediates are only one byte (so they can only hold up to 255) and they must be hardcoded, we also have `substring3` which takes its arguments from the stack (also a start and end).

When decoding ABI encoded types, it will be very common to extract 1, 2, 4, and 8 byte substrings, and convert them integers.  With `substring3` that involves a lot of constants that take code space (one for the start and end).   That is, code that looks like:
```
txn ApplicationArgs 1
int 312
int 316
substring3
btoi
txn ApplicationArgs 1
int 356
int 360
substring3
btoi
add
```
is the most compact way to pull out two uint32s and add them.  This consumes about 27 bytes of code space, and 11 units of cost.

`getbyte` works well for extracting a single byte and converting it to an int.  We should add similar for 2, 4, 8 byte extractions . The above becomes
```
txn ApplicationArgs 1
int 312
get4bytes
txn ApplicationArgs 1
int 356
get4bytes
add
```
which consumes about 17 bytes of space, and 7 units of cost.  This plan would add `get2bytes` `get4bytes` and `get8bytes`. An alternative plan would be to only add `getbytes n`, yielding:
```
txn ApplicationArgs 1
int 312
getbytes 4
txn ApplicationArgs 1
int 356
getbytes 4
add
```
which would be two extra bytes.  This isn't so much more space, but it seems overly general, as it seems to support 5 or 9 (?!) byte integers.

Beyond this, there does not seem to be a strong need for a `substring` like opcode that takes a length (rather than end position), nor for extracting short strings (and not converting them to a uint64).  `substring` seems fine for that, even if a little math is required.  This issue is to shorten code that we expect to be *very* common.

jeapostrophe:
[...]

PR #2521, opened 2021-07-12
Add Extract opcodes 

algochoi:
## Summary

Adds `extract` opcodes that allow a substring to be extracted given a start index and a length (#2347). 

The specs for the `extract` and `extract3` opcodes are the same as the `substring` and `substring3` opcodes, but takes a length rather than an end index. This also adds `extract16bits`, `extract32bits`, `extract64bits` which extracts 2, 4, and 8 byte strings respectively and converts them into integers. 

## Test Plan

Added unit tests and tested using `go test`.


data/transactions/logic/doc.go
```diff
@@ -24,91 +24,97 @@ import (
...
+	"substring":     "pop a byte-array A. For immediate values in 0..255 S and E: extract a range of bytes from A starting at S up to but not including E, push the substring result. If E < S, or either is larger than the array length, the program fails",
+	"substring3":    "pop a byte-array A and two integers B and C. Extract a range of bytes from A starting at B up to but not including C, push the substring result. If C < B, or either is larger than the array length, the program fails",
+	"getbit":        "pop a target A (integer or byte-array), and index B. Push the Bth bit of A.",
+	"setbit":        "pop a target A, index B, and bit C. Set the Bth bit of A to C, and push the result",
+	"getbyte":       "pop a byte-array A and integer B. Extract the Bth byte of A and push it as an integer",
+	"setbyte":       "pop a byte-array A, integer B, and small integer C (between 0..255). Set the Bth byte of A to C, and push the result",
+	"extract":       "pop a byte-array A. For immediate values in 0..255 S and L: extract a range of bytes from A starting at S up to but not including S+L, push the substring result. If S or S+L is larger than the array length, the program fails",
```
jannotti, line 112: Let's do the end of string shortcut: L=0 means "take from S until the end". (only for "extract", not "extract3")

data/transactions/logic/doc.go
```diff
@@ -24,91 +24,97 @@ import (
...
+	"extract3":      "pop a byte-array A and two integers B and C. Extract a range of bytes from A starting at B up to but not including B+C, push the substring result. If B or B+C is larger than the array length, the program fails",
+	"extract16bits": "pop a byte-array A and integer B. Extract a range of bytes from A starting at B up to but not including B+2, push the substring result. If B or B+2 is larger than the array length, the program fails",
+	"extract32bits": "pop a byte-array A and integer B. Extract a range of bytes from A starting at B up to but not including B+4, push the substring result. If B or B+4 is larger than the array length, the program fails",
+	"extract64bits": "pop a byte-array A and integer B. Extract a range of bytes from A starting at B up to but not including B+8, push the substring result. If B or B+8 is larger than the array length, the program fails",
```
jannotti, line 116: For these I think we also want to convert the 2, 4, or 8 bytes to uint.  I can convinced otherwise, but it seems the intent is to make int extract easier.

data/transactions/logic/doc.go
```diff
@@ -234,7 +241,7 @@ func OpDocExtra(opName string) string {
 
 // OpGroups is groupings of ops for documentation purposes.
 var OpGroups = map[string][]string{
-	"Arithmetic":           {"sha256", "keccak256", "sha512_256", "ed25519verify", "+", "-", "/", "*", "<", ">", "<=", ">=", "&&", "||", "shl", "shr", "sqrt", "bitlen", "exp", "==", "!=", "!", "len", "itob", "btoi", "%", "|", "&", "^", "~", "mulw", "addw", "divmodw", "expw", "getbit", "setbit", "getbyte", "setbyte", "concat", "substring", "substring3"},
+	"Arithmetic":           {"sha256", "keccak256", "sha512_256", "ed25519verify", "+", "-", "/", "*", "<", ">", "<=", ">=", "&&", "||", "shl", "shr", "sqrt", "bitlen", "exp", "==", "!=", "!", "len", "itob", "btoi", "%", "|", "&", "^", "~", "mulw", "addw", "divmodw", "expw", "getbit", "setbit", "getbyte", "setbyte", "concat", "substring", "substring3", "extract", "extract3", "extract16bits", "extract32bits", "extract64bits"},
```
jannotti, line 244: Perhaps we want to break the byte array handling routines into their own group.

data/transactions/logic/eval.go
```diff
@@ -2571,6 +2571,81 @@ func opSetByte(cx *evalContext) {
...
+	return
+}
+
+func opExtract16Bits(cx *evalContext) {
+	last := len(cx.stack) - 1 // start
+	prev := last - 1          // bytes
+	start := cx.stack[last].Uint
+	cx.stack[prev].Bytes, cx.err = extract(cx.stack[prev].Bytes, int(start), 2) // extract 16 bits
+
+	cx.stack[prev].Uint = convertBytesToInt(cx.stack[prev].Bytes)
```
jannotti, line 2622: It looks like you are doing the conversion, so my earlier comment should instead be a note to change the spec to say so.

jannotti (changes requested):
This is looking nice.  extract with L=0 is the main thing I'd like to see changed.

jannotti (commented):
I forgot to add: I believe the codegen check is going to fail for you.  You should run `make` inside of `data/transactions/logic` to regenerate the two .md files there.

data/transactions/logic/doc.go
algochoi: My bad, the spec doc is wrong -- will change so that it says to push an int

data/transactions/logic/doc.go
algochoi: Sound good -- I'll add a `Byte Array Slicing` group and move the `substring` and `extract` ops there?

data/transactions/logic/eval.go
```diff
@@ -2571,6 +2571,76 @@ func opSetByte(cx *evalContext) {
...
+	length := cx.stack[last].Uint
+	if start > math.MaxInt32 || length > math.MaxInt32 {
+		cx.err = errors.New("extract range beyond length of string")
+		return
+	}
+	cx.stack[pprev].Bytes, cx.err = extract(cx.stack[pprev].Bytes, int(start), int(length))
+	cx.stack = cx.stack[:prev]
+}
+
+func convertBytesToInt(x []byte) (out uint64) {
```
jdtzmn, line 2612: I think this is the same as the `binary.BigEndian.Uint64()` function from the `encoding/binary` package?

data/transactions/logic/eval.go
```diff
@@ -2571,6 +2571,76 @@ func opSetByte(cx *evalContext) {
...
+	out = uint64(0)
+	for _, b := range x {
+		out = out << 8
+		out = out | (uint64(b) & 0x0ff)
+	}
+	return
+}
+
+func opExtractNBits(cx *evalContext, n int) {
```
jdtzmn, line 2621: nit: `opExtractNBitsImpl`

data/transactions/logic/eval.go
```diff
@@ -2571,6 +2571,76 @@ func opSetByte(cx *evalContext) {
...
+	// Shortcut: if length is 0, take bytes from start index to the end
+	l := int(length)
+	if length == 0 {
+		l = len(cx.stack[last].Bytes) - int(start)
+	}
+	cx.stack[last].Bytes, cx.err = extract(cx.stack[last].Bytes, int(start), l)
+}
+
+func opExtract3(cx *evalContext) {
+	last := len(cx.stack) - 1 // length
```
jdtzmn, line 2599: Could just be me, but using `last` as an argument index and `start` as a slicing index confused me. What do you think about using names such as `lengthIdx`, `startIdx`, and `bytesIdx`? This would also make it easier to read line 2608 without having to refer back to what `pprev` is.

jdtzmn (changes requested):
Good work on this PR! I had just a few comments.

data/transactions/logic/eval.go
```diff
@@ -2571,6 +2571,76 @@ func opSetByte(cx *evalContext) {
 	cx.stack = cx.stack[:prev]
 }
 
+func extract(x []byte, start, length int) (out []byte, err error) {
```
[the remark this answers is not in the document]
algochoi: Ah I was looking at the [substring](https://github.com/algorand/go-algorand/blob/master/data/transactions/logic/eval.go#L2415) implementations and copied that format, but I think the `opXImpl` convention looks more recent. I'll change to `opExtractImpl` and `opExtractNBitsImpl` unless anyone else has any objections.

data/transactions/logic/eval.go
```diff
@@ -2571,6 +2571,76 @@ func opSetByte(cx *evalContext) {
...
+	out = x
+	end := start + length
+	if start > len(x) || end > len(x) {
+		err = errors.New("extract range beyond length of string")
+		return
+	}
+	out = x[start:end]
+	err = nil
```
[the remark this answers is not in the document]
algochoi: I think you're right -- I'll remove this to clean it up.

data/transactions/logic/eval.go
algochoi: I think your suggestion makes more sense -- will change the indices to have `*Idx`.

data/transactions/logic/eval.go
algochoi: You're right, it looks quite similar -- I just lifted this from the [`btoi` function](https://github.com/algorand/go-algorand/blame/master/data/transactions/logic/eval.go#L1080)...

jdtzmn: Oh interesting. I'm not familiar enough with the binary syntax to know whether they are equivalent; I wonder what @jannotti thinks.

jannotti: I wondered the same, whether we should be using a library function.  But when I saw the btoi function did it this way, I didn't bother saying anything.  It would be great to investigate and see if we can use the library. Always preferable not to do things like this ourselves.  One thing to confirm - the library call need to accept "short" byte arrays. A single byte is a legal uint64 decoded "our way".

jannotti: No dice.
```
func (bigEndian) Uint64(b []byte) uint64 {
	_ = b[7] // bounds check hint to compiler; see golang.org/issue/14808
	return uint64(b[7]) | uint64(b[6])<<8 | uint64(b[5])<<16 | uint64(b[4])<<24 |
		uint64(b[3])<<32 | uint64(b[2])<<40 | uint64(b[1])<<48 | uint64(b[0])<<56
}
```

algochoi: Good catch JJ -- I'll leave it as is then.

jannotti: You could extract it into a function if you'd like.

Either way, could you add a comment explaining why we do it?

algochoi: Done!

data/transactions/logic/eval.go
```diff
@@ -2571,6 +2571,77 @@ func opSetByte(cx *evalContext) {
 	cx.stack = cx.stack[:prev]
 }
 
+func opExtractImpl(x []byte, start, length int) (out []byte, err error) {
```
AyAggarwal, line 2574: This is a nice

AyAggarwal (approved):

jdtzmn (approved):

shiqizng (approved):

data/transactions/logic/README.md
```diff
@@ -139,8 +139,6 @@ For three-argument ops, `A` is the element two below the top, `B` is the penulti
 | `getbyte` | pop a byte-array A and integer B. Extract the Bth byte of A and push it as an integer |
 | `setbyte` | pop a byte-array A, integer B, and small integer C (between 0..255). Set the Bth byte of A to C, and push the result |
 | `concat` | pop two byte-arrays A and B and join them, push the result |
-| `substring s e` | pop a byte-array A. For immediate values in 0..255 S and E: extract a range of bytes from A starting at S up to but not including E, push the substring result. If E < S, or either is larger than the array length, the program fails |
```
jannotti, line 142: Why are these gone?  I thought maybe they got re-ordered, but then wouldn't I see them being added elsewhere?

jannotti (approved):
opExtractNBits ought to be called opExtraNBytes, right? (It's the helper for the bit ops, but the n arg is in bytes)

data/transactions/logic/README.md
algochoi: I think it's because I moved the `substring` ops from the original arithmetic opgroup... I'll take a closer look at this too

algochoi:
> opExtractNBits ought to be called opExtraNBytes, right? (It's the helper for the bit ops, but the n arg is in bytes)

You're right -- I'll change this

PR #2523, opened 2021-07-12
Use golangci-lint

cce:
## Summary

This adds a golangci-lint configuration file that runs an initial minimal set of linters: golint, govet, and misspell. This makes it easier to integrate with editors and IDEs and also opens the door to using from CI.

## Test Plan

Maybe later, this could be incorporated as a CI pass and/or for use in the `make sanity` step.


tsachiherman (commented):
would the 5 min timeout be an issue in case we would need to perform any compilation as part of the execution ?

cce:
I think 5m ought to be enough for now — anecdotally, the 1m default timeout didn't work for me when running with a clean cache, so I bumped it to 5m.

tsachiherman (approved):

winder:
@cce great to see that you're looking into this! I think putting this in the CI right away would be good. Maybe for information at first and we can make it more strict after some trial period.

How did you decide which linters to enable/disable? The last time I ran this was with the default configuration, and all of the output looked semi-useful. We should add this to Indexer as well, would it make sense to have a standard config for both projects?

@tsachiherman here is the sort of output it provides (with all linters enabled): https://github.com/algorand/go-algorand-internal/issues/1329#issuecomment-867600892

tsachiherman:
> @cce great to see that you're looking into this! I think putting this in the CI right away would be good. Maybe for information at first and we can make it more strict after some trial period.
> 
> How did you decide which linters to enable/disable? The last time I ran this was with the default configuration, and all of the output looked semi-useful. We should add this to Indexer as well, would it make sense to have a standard config for both projects?
> 
> @tsachiherman here is the sort of output it provides (with all linters enabled): [algorand/go-algorand-internal#1329 (comment)](https://github.com/algorand/go-algorand-internal/issues/1329#issuecomment-867600892)

@winder - I like linters and static code testing tools. One thing that I'd like to ensure is that we don't take these as a code correctness validation tool but more as an advanced informational styling guide. As such, I hope that each of these warnings could be turned off via an in-code comment ( i.e. we don't want to create yet-another-list of false-positive entries.. )

winder:
> > @cce great to see that you're looking into this! I think putting this in the CI right away would be good. Maybe for information at first and we can make it more strict after some trial period.
> > How did you decide which linters to enable/disable? The last time I ran this was with the default configuration, and all of the output looked semi-useful. We should add this to Indexer as well, would it make sense to have a standard config for both projects?
> > @tsachiherman here is the sort of output it provides (with all linters enabled): [algorand/go-algorand-internal#1329 (comment)](https://github.com/algorand/go-algorand-internal/issues/1329#issuecomment-867600892)
> 
> @winder - I like linters and static code testing tools. One thing that I'd like to ensure is that we don't take these as a code correctness validation tool but more as an advanced informational styling guide. As such, I hope that each of these warnings could be turned off via an in-code comment ( i.e. we don't want to create yet-another-list of false-positive entries.. )

@tsachiherman unlike so many other go tools, this one appears to be highly configurable. In this PR @cce showed how we can create a blanket policy to ignore a category of errors. [You can also disable them with a comment in the code](https://golangci-lint.run/usage/false-positives/): 

Disable ALL linters:
```
//nolint
func FunctionWithIssues() {
}
```

Disable One linter:
```
//nolint:deadcode
func unusedFunctionThatWeWantToKeep() {
}
```

cce:
@winder I picked the linters we already have been using (govet, golint, misspell) that already pass. I didn't want to turn any more on without first ensuring they come out clean. You can see the rest of the default set of linters enabled by golangci-lint in the `disabled:` section and I think some of them would provide value to enable later.

Since govet and golint don't support `//nolint` comments, and golint has been deprecated and archived (but still contains a useful set of checks IMO), golangci-lint is pretty popular among Go projects and gives us more ability to iterate on our linting policy by just modifying the golangci-lint config file. It is a nice upgrade to be able to put an inline `thingThatCausesLinterToComplain() // nolint:XXX` too.

PR #2525, opened 2021-07-12
[DRAFT] Testing Full Test on CircleCI

algobarb:
## Summary
Testing for Full Test on CircleCI.


.circleci/config.yml
```diff
@@ -1,8 +1,408 @@
...
+            GOLANG_VERSION=$(./scripts/get_golang_version.sh)
+            eval "$(~/gimme "${GOLANG_VERSION}")"
+            scripts/configure_dev.sh
+            scripts/buildtools/install_buildtools.sh -o "gotest.tools/gotestsum"
+            export ALGOTEST=1
+            export SHORTTEST=
+            export TEST_RESULTS=/tmp/results/<< parameters.result_subdir >>
+            export PARTITION_TOTAL=$CIRCLE_NODE_TOTAL
+            export PARTITION_ID=$CIRCLE_NODE_INDEX
+            test/scripts/run_integration_tests.sh
```
winder, line 225: To avoid splitting the logic across multiple CI config files and make it easier to support our multiple build systems, all of these command blocks should be part of the existing CI scripts. We can rename the `scripts/travis` directory to `scripts/ci` and use conditions like:
```
if [ "$JENKINS_URL" != "" ];then

elif [ "$CI" = "true" ] && [ "$TRAVIS" = "true" ] && [ "$SHIPPABLE" != "true" ]; then

elif [ "$CI" = "true" ] && [ "$CIRCLECI" = "true" ]; then

fi
```

.circleci/config.yml
```diff
@@ -1,8 +1,408 @@
...
+            GOLANG_VERSION=$(./scripts/get_golang_version.sh)
+            eval "$(~/gimme "${GOLANG_VERSION}")"
+            scripts/configure_dev.sh
+            scripts/buildtools/install_buildtools.sh -o "gotest.tools/gotestsum"
+            PACKAGES="$(go list ./... | grep -v /go-algorand/test/)"
+            export PACKAGE_NAMES=$(echo $PACKAGES | tr -d '\n')
+            export PARTITION_TOTAL=$CIRCLE_NODE_TOTAL
+            export PARTITION_ID=$CIRCLE_NODE_INDEX
+            export PARALLEL_FLAG="-p 1"
+            gotestsum --format testname --junitfile /tmp/results/<< parameters.result_subdir >>/results.xml --jsonfile /tmp/results/<< parameters.result_subdir >>/testresults.json -- --tags "sqlite_unlock_notify sqlite_omit_load_extension" -race -timeout 1h -coverprofile=coverage.txt -covermode=atomic -p 1 $PACKAGE_NAMES
```
winder, line 173: You should find a way to add the junitfile option in the Makefile. Does this somehow trigger rebuilds when the circleci task is split across multiple workers?

In particular, the `gotestsum` command/arguments are exported in the `GOTESTCOMMAND` variable at the top of the Makefile, and that variable is used later on when running integration tests.

If I'm correct, this means that the integration tests fall back to using `go test`, and there is no report generated for the e2e tests.

algobarb:
This PR was only used as a testing Workspace.

PR #2526, opened 2021-07-12
Replace cpp sortition with Go

barnjamin:
## Summary

I think removing C++ from this repo will help with maintenance over time.  This was a small bit of C++ that should be easy to replace.  

This code updates the logic to find the appropriate boundary when evaluating the binomial distribution as well as adding a few tests and benchmarks that @brianolson wrote.

The original code calls a CDF function and compares the return value against a threshold value.  Initially I started down this path but found that the CDF is just the sum of the probabilities to that point.  This means we can just compute the PMF as we walk up the distribution and accumulate the sum of probability masses as the CDF.

Additionally I changed the distribution from a Binomial to a Poisson distribution since the approximation is close as N is large and P is small and I think N should be at least 1e6 microalgos.  I think using Poisson over Binomial to compute PMF reduces the chance of numerical computation issues since the Poisson can be computed with fewer steps.

I've run the test against the C++ code with several hundred thousand rounds so far and it appears to be giving the same answer for all of them but I'm leaving it in for now in case there are more tests you'd like to see.

As for performance, computing the PMF for each iteration and summing instead of computing the CDF already gives us an advantage and Poisson appx also improves the situation so the total performance increase is about a full order of mangitude.

```
BenchmarkBoostCdfWalk
BenchmarkBoostCdfWalk-12                   67533             17955 ns/op
BenchmarkPoissonCdfWalk
BenchmarkPoissonCdfWalk-12            635524              1769 ns/op
```

## Test Plan

Tests written by @brianolson for the comparison but there should probably be a lot more.  Happy to write them if you can tell me what you'd like to see.


barnjamin:
Quick comparison of Poisson vs Binomial as the sum of the absolute difference between them after the first 100 iterations. 

https://gist.github.com/barnjamin/98545fd23b9beb8f564e881faa0d8646

image

barnjamin:
I also tried to convert the existing C++ to the summing of binomial pmf (pdf in boost) and found that the rounding errors would accumulate to where if the ratio being compared against was high (> 0.9950) the summed cdf would never reach it. 

In practice I don't expect the ratio to ever be that high but it could be.

When it didn't get stuck on a high ratio case it was much faster at around 90ns/op

```
uint64_t sortition_binomial_cdf_walk_sum(double n, double p, double ratio, uint64_t money) {
  boost::math::binomial_distribution<double> dist(n, p);
  double cdf = 0;
  for (uint64_t j = 0; j < money; j++) {
    // Add the pdf to the cdf
    cdf += pdf(dist, j);

    if (ratio <= cdf) {
      return j;
    }
  }
  return money;
}
```

barnjamin:
:(

Issue #1015, opened 2020-05-01
REST v2 Codegeneration: Avoid Re-defining Types in /blocks and /transactions


winder:
The generated code defines `Block` and `SignedTransaction` as a generic structure specified by `map[string]interface{}`. In the spec this is just a JSON `object`. The generated type was intended to express "map of objects" and was intentionally under-specifying so that we don't need to synchronize the backend type and the specification. Rather the REST API will always return the backend representation of these objects.

I was able to get a `map[string]interface{}` by serializeing and deserializing as the `map[string]interface{}` type. There was some trouble getting the generic type working with `json/encoder`, at the time I didn't realize we also use `go-codec` for JSON, so maybe that one would work. For now there are 3 spots which redefine the types inline:
```
response := struct {
	AssetIndex *uint64 `codec:"asset-index,omitempty"`
	CloseRewards *uint64 `codec:"close-rewards,omitempty"`
	ClosingAmount *uint64 `codec:"closing-amount,omitempty"`
	ConfirmedRound *uint64 `codec:"confirmed-round,omitempty"`
	PoolError string `codec:"pool-error"`
	ReceiverRewards *uint64 `codec:"receiver-rewards,omitempty"`
	SenderRewards *uint64 `codec:"sender-rewards,omitempty"`
-->     Txn transactions.SignedTxn `codec:"txn"`
}
```

To avoid this we could explore the following:
1) expand the oapi-codegen type-mapping so that we can specify an existing type + import it in the generated code.
2) See if the serialize/deserialize/serialize trick works using `go-codec` instead of `json/encoder`

ian-algorand:
merged, closing parent ticket. Correct me if i'm wrong here, Will

jasonpaulos:
I believe this issue is still relevant.

PR #2529, opened 2021-07-13
Opcode: log

shiqizng:
This PR adds a new opcode `log`. It takes one "bytes" stack argument. It can be called up to 32 times in a program, and log up to a total of 1k bytes.


data/transactions/logic/doc.go
```diff
@@ -142,6 +142,8 @@ var opDocByName = map[string]string{
 	"b&":  "A bitwise-and B, where A and B are byte-arrays, zero-left extended to the greater of their lengths",
 	"b^":  "A bitwise-xor B, where A and B are byte-arrays, zero-left extended to the greater of their lengths",
 	"b~":  "X with all bits inverted",
+
+	"log": "write bytes to log state of the current application",
```
jannotti, line 146: It would be good to explain the 1k and 8 calls limitation in opExtraDocs.

data/transactions/logic/eval.go
```diff
@@ -731,6 +741,15 @@ func (cx *evalContext) checkStep() (int, error) {
 	if (cx.runModeFlags & spec.Modes) == 0 {
 		return 0, fmt.Errorf("%s not allowed in current mode", spec.Name)
 	}
+
+	if spec.Name == "log" {
+		cx.logCalls++
+	}
```
jannotti, line 747: By putting this in Check(), the program will only be able to have 8 log calls in it.  But we want to enforce is that only 8 log calls actually happen.  So I would not track this here, but rather in Eval().

This also means you can place the increment in opLog, instead of checking the name and incrementing here.

data/transactions/logic/eval.go
```diff
@@ -3053,3 +3072,25 @@ func opAppParamsGet(cx *evalContext) {
...
+	log := cx.stack[last]
+	cx.logSize += len(log.Bytes)
+
+	if cx.logSize > MaxLogSize {
+		cx.err = fmt.Errorf("%v is too large. Up to %v bytes are allowed", cx.logSize, MaxLogSize)
+		return
+	}
+
+	// write log to applyData
+	err := cx.Ledger.SetLog(log.toTealValue())
```
jannotti, line 3088: "AddLog" might be better than "SetLog"

data/transactions/logic/evalStateful_test.go
```diff
@@ -787,6 +804,7 @@ func testApp(t *testing.T, program string, ep EvalParams, problems ...string) ba
 		require.NoError(t, err)
 		require.Empty(t, delta.GlobalDelta)
 		require.Empty(t, delta.LocalDeltas)
+		require.Empty(t, delta.LogDelta)
```
jannotti, line 807: "Logs" is probably a better name than "LogDelta" since there is no "change" here, implied by the "Delta" naming.

ledger/appcow.go
```diff
@@ -425,6 +430,18 @@ func (cb *roundCowState) DelKey(addr basics.Address, aidx basics.AppIndex, globa
 	return nil // note: deletion cannot cause us to violate maxCount
 }
 
+// AppendLog creates a new key-value {addr, aidx} in log
+func (cb *roundCowState) AppendLog(aidx basics.AppIndex, value basics.TealValue) error {
```
jannotti, line 434: Earlier I suggested "AddLog". Given this naming, "AppendLog" could be used to match.

jannotti (commented):
I'd like to see some tests that contain multiple calls to `log`

data/transactions/logic/eval.go
```diff
@@ -3053,3 +3072,25 @@ func opAppParamsGet(cx *evalContext) {
...
 }
+
+func opLog(cx *evalContext) {
+	last := len(cx.stack) - 1
+
+	log := cx.stack[last]
+	cx.logSize += len(log.Bytes)
+
+	if cx.logSize > MaxLogSize {
+		cx.err = fmt.Errorf("%v is too large. Up to %v bytes are allowed", cx.logSize, MaxLogSize)
```
tsachiherman, line 3083: nit: try to use `%d` for integer printouts.

runvnc:
Hello, this is really great. I have been doing my own "log()" in Tealang using get, put and concat.  One thing I wonder though, why is it necessary to limit to 8 times, rather than using the existing cost system?

jannotti:
> Hello, this is really great. I have been doing my own "log()" in Tealang using get, put and concat. One thing I wonder though, why is it necessary to limit to 8 times, rather than using the existing cost system?

After discussion on Discord, we've decided that allowing 32 calls makes sense, with the 1k limit intact. For posterity - the reason for two such limits is, as with ApplicationArgs, there is additional space overhead associated with each call, so it should not be possible to log 1000 1 byte messages, as that would consume much more than 1k.  But 32 should not really make much difference.

data/transactions/logic/doc.go
```diff
@@ -225,6 +227,7 @@ var opDocExtras = map[string]string{
 	"asset_holding_get": "params: Txn.Accounts offset (or, since v4, an account address that appears in Txn.Accounts or is Txn.Sender), asset id (or, since v4, a Txn.ForeignAssets offset). Return: did_exist flag (1 if exist and 0 otherwise), value.",
 	"asset_params_get":  "params: Before v4, Txn.ForeignAssets offset. Since v4, Txn.ForeignAssets offset or an asset id that appears in Txn.ForeignAssets. Return: did_exist flag (1 if exist and 0 otherwise), value.",
 	"app_params_get":    "params: Txn.ForeignApps offset or an app id that appears in Txn.ForeignApps. Return: did_exist flag (1 if exist and 0 otherwise), value.",
+	"log":               "`log` can be called up to 8 times in a program, and log up to a total of 1k bytes.",
```
barnjamin, line 230: s/8/32

data/transactions/logic/eval_test.go
```diff
@@ -4419,3 +4419,98 @@ func TestBytesConversions(t *testing.T) {
 	testAccepts(t, "byte 0x11; byte 0x10; b+; btoi; int 0x21; ==", 4)
 	testAccepts(t, "byte 0x0011; byte 0x10; b+; btoi; int 0x21; ==", 4)
 }
+
+func TestLog(t *testing.T) {
```
jdtzmn, line 4423: This test function is very clean 👌

data/transactions/logic/assembler_test.go
```diff
@@ -294,6 +294,7 @@ int 100
 
 const v5Nonsense = `
 app_params_get AppExtraProgramPages
+log
```
jdtzmn, line 297: Shouldn't the compiled map below reflect changes made to the nonsense map?

jdtzmn (approved):

data/basics/teal.go
```diff
@@ -134,6 +134,8 @@ type EvalDelta struct {
 	// When decoding EvalDeltas, the integer key represents an offset into
 	// [txn.Sender, txn.Accounts[0], txn.Accounts[1], ...]
 	LocalDeltas map[uint64]StateDelta `codec:"ld,allocbound=config.MaxEvalDeltaAccounts"`
+
+	Log []string `codec:"lg,allocbound=-"`
```
jdtzmn, line 138: Shouldn't `allocbound` be set to `logic.MaxLogCalls`? Maybe it makes sense to move the `MaxLogCalls` constant to the `config` package?

jdtzmn (changes requested):
Looks good, just had one question.

data/basics/teal.go
shiqizng: how about leaving logic.MaxLogCalls as is but create a variable in config for decoding purposes.

shiqizng: or should MaxLogCalls and MaxLogSize be consensus params? @jannotti

shiqizng: I moved to MaxLogCalls to config. and allocbound is now set to config.MaxLogCalls.

data/transactions/logic/eval_test.go
```diff
@@ -4419,3 +4419,98 @@ func TestBytesConversions(t *testing.T) {
...
+		runMode:     runModeApplication,
+	}
+
+	failCase5 := failCase{
+		source:      `byte  "a logging message"; log; int 1`,
+		errContains: "log not allowed in current mode",
+		runMode:     runModeSignature,
+	}
+
+	failCases := []failCase{failCase0, failCase1, failCase2, failCase3, failCase4, failCase5}
```
winder, line 4498: nit: A pattern we use a lot for this is an anonymous struct:
```
failCases := []struct {
	source      string
	runMode     runMode
	errContains string
} {
    // cases here.
}
```

data/transactions/logic/eval.go
```diff
@@ -159,6 +162,8 @@ type LedgerForLogic interface {
 	DelGlobal(key string) error
 
 	GetDelta(txn *transactions.Transaction) (evalDelta basics.EvalDelta, err error)
+
+	AppendLog(value basics.TealValue) error
```
jasonpaulos, line 166: Why take the argument as a `TealValue` instead of `[]byte`?

data/transactions/logic/eval.go
```diff
@@ -3053,3 +3060,26 @@ func opAppParamsGet(cx *evalContext) {
 	cx.stack[last] = value
 	cx.stack = append(cx.stack, stackValue{Uint: exist})
 }
+
+func opLog(cx *evalContext) {
+	last := len(cx.stack) - 1
```
jasonpaulos, line 3065: Are there any checks to ensure the argument is bytes, not an integer? Relying on the assembler isn't enough, since it can't type check loading values from places like scratch space, global state, etc.

data/transactions/logic/eval_test.go
```diff
@@ -4419,3 +4419,98 @@ func TestBytesConversions(t *testing.T) {
...
+	sb := strings.Builder{}
+	ep := defaultEvalParams(&sb, &txn)
+	ep.Proto = &proto
+	ep.Ledger = ledger
+
+	source := `byte  "a logging message"; log; int 1`
+	source1 := `byte  "a logging message"; log; byte  "second logging message"; log; int 1`
+	source2 := fmt.Sprintf(`%s int 1`, strings.Repeat(`byte "a logging message"; log;`, config.MaxLogCalls))
+	source3 := `int 1; loop: byte "a logging message"; log; int 1; +; dup; int 30; <; bnz loop;`
+	sources := []string{source, source1, source2, source3}
```
jasonpaulos, line 4442: I couldn't find a test anywhere which ensures logging a 1000-byte string works successfully. That would probably be good to add to make sure there is no mistake about the boundary condition.

ledger/appcow_test.go
```diff
@@ -1299,3 +1330,24 @@ func TestCowDelKey(t *testing.T) {
...
+	aidx := basics.AppIndex(1)
+	c := getCow([]modsData{
+		{addr, basics.CreatableIndex(aidx), basics.AppCreatable},
+	})
+
+	val := strings.Repeat("val", 100)
+	tv := basics.TealValue{Type: basics.TealBytesType, Bytes: val}
+	err := c.AppendLog(aidx, tv)
+	a.Error(err)
+	a.Contains(err.Error(), "value too long")
```
jasonpaulos, line 1346: Why should this case fail? Since `val := strings.Repeat("val", 100)`, `val` should be 300 bytes long, which is well under the 1000 byte limit -- let me know if I'm missing something here.

ledger/applications_test.go
```diff
@@ -1295,3 +1299,19 @@ func testAppAccountDeltaIndicesCompatibility(t *testing.T, source string, accoun
...
+
+	addr := getRandomAddress(a)
+	aidx := basics.AppIndex(1)
+	c := newCowMock([]modsData{
+		{addr, basics.CreatableIndex(aidx), basics.AppCreatable},
+	})
+	l, err := newLogicLedger(c, aidx)
+	a.NoError(err)
+	a.NotNil(l)
+	tv := basics.TealValue{Type: basics.TealBytesType, Uint: 1}
```
jasonpaulos, line 1314: I think this is a mistake? It should probably be `basics.TealValue{Type: basics.TealBytesType, Bytes: someByteString}` -- this is another argument for why `AppendLog` might be better off taking a byte slice as an argument.

jasonpaulos (changes requested):
Mostly good, but I had some questions and suggestions about types.

data/transactions/logic/eval.go
shiqizng: I used TealValue to be consistent with existing methods like SetLocal/SetGlobal

ledger/appcow_test.go
shiqizng: this is because I had `len(value.Bytes) > cb.proto.MaxAppBytesValueLen` in AppendLog. it should be compared to MaxLogSize instead.

data/transactions/logic/eval.go
jasonpaulos: Those methods only use `TealValue` for global/local state values, which can be either integers or bytes. They actually use `string` for the keys which are bytes-only, so perhaps `AppendLog` should also use `string` to be consistent with them.

jasonpaulos (dismissed):
Looks good to me!

jasonpaulos (approved):

ledger/appcow_test.go
```diff
@@ -1299,3 +1330,21 @@ func TestCowDelKey(t *testing.T) {
 	a.Panics(func() { c.DelKey(getRandomAddress(a), aidx, false, key, 0) })
 	a.Panics(func() { c.DelKey(addr, aidx+1, false, key, 0) })
 }
+func TestCowAppendLog(t *testing.T) {
```
jdtzmn, line 1333: nit: add newline between these functions

jdtzmn (approved):
Looks good, just one formatting issue.

data/transactions/logic/eval.go
```diff
@@ -53,6 +53,9 @@ const MaxStringSize = 4096
 // MaxByteMathSize is the limit of byte strings supplied as input to byte math opcodes
 const MaxByteMathSize = 64
 
+// MaxLogSize is the limit of total log size from n log calls in a program
+const MaxLogSize = 1000
```
jannotti, line 57: Let's make this 1024 so you can make 32 logs of a 32 byte message.

data/transactions/logic/evalStateful_test.go
```diff
@@ -466,6 +468,18 @@ func (l *testLedger) GetDelta(txn *transactions.Transaction) (evalDelta basics.E
...
 
+func (l *testLedger) AppendLog(value string) error {
+	appIdx := l.appID
+	_, ok := l.applications[appIdx]
+	if !ok {
+		return fmt.Errorf("no such app")
+	}
+
+	// append logs
+	l.logs[appIdx] = append(l.logs[appIdx], value)
```
jannotti, line 479: If this is the only way a log can be written, the l.logs should probably just be []string, not a map.  Only the current app can have logs.

data/transactions/logic/eval_test.go
```diff
@@ -4487,3 +4487,99 @@ func TestBytesConversions(t *testing.T) {
...
+	for _, s := range sources {
+		ops := testProg(t, s, AssemblerMaxVersion)
+
+		err := CheckStateful(ops.Program, ep)
+		require.NoError(t, err, source)
+
+		pass, err := EvalStateful(ops.Program, ep)
+		require.NoError(t, err)
+		require.True(t, pass)
+	}
```
jannotti, line 4522: There should be a test that the log messages actually ended up in the ledger of ep.

data/transactions/logic/eval_test.go
```diff
@@ -4487,3 +4487,99 @@ func TestBytesConversions(t *testing.T) {
...
+	ep := defaultEvalParams(&sb, &txn)
+	ep.Proto = &proto
+	ep.Ledger = ledger
+
+	source := `byte  "a logging message"; log; int 1`
+	source1 := `byte  "a logging message"; log; byte  "second logging message"; log; int 1`
+	source2 := fmt.Sprintf(`%s int 1`, strings.Repeat(`byte "a logging message"; log;`, config.MaxLogCalls))
+	source3 := `int 1; loop: byte "a logging message"; log; int 1; +; dup; int 30; <; bnz loop;`
+	source4 := fmt.Sprintf(`byte "%s"; log; int 1`, strings.Repeat("a", MaxLogSize))
+	sources := []string{source, source1, source2, source3, source4}
```
jannotti, line 4511: Are the source* variables used explicitly? If not, you can declare them all at once in sources.
```
sources := []string {
  `byte "a logging message"; log; int 1`,
  ...
}
```

ledger/appcow.go
```diff
@@ -425,6 +425,18 @@ func (cb *roundCowState) DelKey(addr basics.Address, aidx basics.AppIndex, globa
 	return nil // note: deletion cannot cause us to violate maxCount
 }
 
+// AppendLog creates a new key-value {addr, aidx} in log
+func (cb *roundCowState) AppendLog(aidx basics.AppIndex, value string) error {
+	// Enforce maximum value length
+	if len(value) > logic.MaxLogSize {
+		return fmt.Errorf("value too long, length was %d", len(value))
+	}
```
jannotti, line 433: How is the maximum total length and/or number of logs being checked?

ledger/appcow.go
```diff
@@ -519,6 +532,17 @@ func (cb *roundCowState) BuildEvalDelta(aidx basics.AppIndex, txn *transactions.
 			}
 		}
 	}
+
+	// logDeltas
+	for appid, ldelta := range cb.logs {
```
jannotti, line 537: This makes me think `cb.log` should just be an array too, rather than map.  But I'm not 100% sure.  Perhaps we want to use this "wider" representation in ApplyDelta in preparation for differentiating log messages from multiple apps, once/when we allow apps to call apps.

ledger/apply/application_test.go
```diff
@@ -814,6 +814,15 @@ func TestAppCallClearState(t *testing.T) {
 	a.Equal(0, len(br.AppLocalStates))
 	a.Equal(basics.StateSchema{}, br.TotalAppSchema)
 	a.Equal(basics.EvalDelta{GlobalDelta: gd}, ad.EvalDelta)
+
+	b.ResetWrites()
+	b.pass = true
+	b.err = nil
+	logs := map[basics.AppIndex][]string{appIdx: {"a"}}
```
jannotti, line 821: What's the program actually being run here?  It should do at least two logs, so we can test that they both appear.

Also test that a program with 33 logs fails.  And one with two 513 byte logs.

jannotti (changes requested):

jannotti:
We should make sure that the algod endpoints that yield transactions actually show these logs in the ApplyData.  I think that has to be an integration test that hits the REST API.  I don't know whether new code has to be written to make that work, or it happens based on reflection on the ApplyData type.  But I guess that's why we test.

We'll also need a companion issue to ensure the indexer shows these from similar endpoints.

ledger/appcow.go
shiqizng: for now I will use an array instead of a map.

ledger/appcow.go
shiqizng: it's done in opLog(), https://github.com/shiqizng/go-algorand/blob/opcode/log/data/transactions/logic/eval.go#L3135

jannotti: Oh, right, of course. So why does this code have the MaxLogSize test at all.  Maybe that shouldn't happen at this layer.

ledger/apply/application_test.go
jannotti: I think I retract this.  As you say, these checks happen at a higher layer.

jasonpaulos:
For REST API changes, I believe the `DryrunTxnResult` and `PendingTransactionResponse` types in `daemon/algod/api/algod.oas2.json` need to be updated to include the array of logs from application calls. You'll then need to regenerate other files based on changes to this file, see `daemon/algod/api/README.md` for details.

The above steps will update the types, but you'll also need to add some code to copy the right values into these types.
* For `DryrunTxnResult`, I believe you'll need to modify `doDryrunRequest` in `daemon/algod/api/server/v2/dryrun.go` to include logged values, similar to how state deltas are added. `dryrun_test.go` is probably a good place to add a unit test for this.
* For `PendingTransactionResponse`, you'll need to modify `PendingTransactionInformation` in `daemon/algod/api/server/v2/handlers.go`. Weirdly it looks like we don't actually use `PendingTransactionResponse` (issue to fix is #1015), so you'll have to add the new field to the anonymous struct there. I'm not sure where the best place to unit test this is.

daemon/algod/api/server/v2/dryrun_test.go
```diff
@@ -332,6 +332,21 @@ done:
 
 var localStateCheckProg []byte
 
+const LogsCheckSource = `
```
jasonpaulos, line 335: I don't think this program gets used anywhere?

jasonpaulos (changes requested):

daemon/algod/api/server/v2/test/helpers.go
```diff
@@ -115,7 +115,23 @@ func (m mockNode) BroadcastSignedTxGroup(txgroup []transactions.SignedTxn) error
 }
 
 func (m mockNode) GetPendingTransaction(txID transactions.Txid) (res node.TxnWithStatus, found bool) {
-	res = node.TxnWithStatus{}
+	t := transactions.Transaction{
```
jasonpaulos, line 118: I think the modifications to this file are no longer needed since you're testing with e2e tests instead?

data/transactions/logic/evalStateful_test.go
```diff
@@ -467,6 +470,21 @@ func (l *testLedger) GetDelta(txn *transactions.Transaction) (evalDelta basics.E
 	return
```
jasonpaulos, line 470: I believe you should also add the logs the return value of the `GetDelta` function.

ledger/appcow.go
```diff
@@ -519,6 +532,10 @@ func (cb *roundCowState) BuildEvalDelta(aidx basics.AppIndex, txn *transactions.
 			}
 		}
 	}
+
+	// logs
+	evalDelta.Logs = cb.logs
```
jasonpaulos, line 537: I think the intent is to have `LogItem.ID` refer to an index into `txn.ForeignApps` for log items when they are in the `EvalDelta` struct, since that will save space compared to storing a larger app ID for every log item.

The question is what should `LogItem.ID` refer to inside of `LedgerForLogic`/`roundCowState`?

Currently it refers to the real app ID. If we want to keep it that way, we'll need to convert each log item's ID into an index into `txn.ForeignApps`. That should happen in this function, since it also converts local state delta addresses into indexes into `txn.Accounts` (the same should also happen in the `GetDelta` function from `data/transactions/logic/evalStateful_test.go`). Though if we decide to do this, I'd recommend creating two structs to represent the two different `LogItem` versions: one with app ID and one with index into `txn.ForeignApps`. Otherwise it'll be difficult to keep track what `ID` actually means.

The alternative would be to make `LogItem.ID` always be an index into `txn.ForeignApps`. Then the `AppendLog` functions could convert the app ID to the index when the `LogItem` is created.

test/e2e-go/restAPI/restClient_test.go
```diff
@@ -186,6 +187,36 @@ func waitForTransaction(t *testing.T, testClient libgoal.Client, fromAddress, tx
...
 }
 
+func waitForTransactionV2(t *testing.T, testClient libgoal.Client, fromAddress, txID string, timeout time.Duration) (tx v1.Transaction, err error) {
+	a := require.New(fixtures.SynchronizedTest(t))
+	rnd, err := testClient.Status()
+	a.NoError(err)
+	if rnd.LastRound == 0 {
+		t.Fatal("it is currently round 0 but we need to wait for a transaction that might happen this round but we'll never know if that happens because ConfirmedRound==0 is indestinguishable from not having happened")
+	}
+	timeoutTime := time.Now().Add(timeout * time.Second)
```
jasonpaulos, line 197: This should not multiply by `time.Second` since `timeout` is already a `time.Duration`

daemon/algod/api/server/v2/utils.go
```diff
@@ -266,6 +266,30 @@ func convertToDeltas(txn node.TxnWithStatus) (*[]generated.AccountStateDelta, *g
...
+func convertToLogItems(txn node.TxnWithStatus) *[]generated.LogItem {
+	var logItems *[]generated.LogItem
+	if len(txn.ApplyData.EvalDelta.Logs) > 0 {
+		l := make([]generated.LogItem, 0)
+
+		for _, v := range txn.ApplyData.EvalDelta.Logs {
+			// Resolve appid from index
+			var appid uint64
+			if v.ID == 0 {
+				appid = uint64(txn.Txn.Txn.ApplicationID)
```
jasonpaulos, line 278: This will work for most app calls, but it won't for an app creation call, since `txn.Txn.Txn.ApplicationID` will be 0.

To fix it, you could make this function take an additional `*uint64` param which you use to pass in `response.ApplicationIndex` from `daemon/algod/api/server/v2/handlers.go`. Then if `txn.Txn.Txn.ApplicationID` is 0 and the additional `*uint64` param is not nil, assign it to `appid` instead.

jasonpaulos (changes requested):

ledger/appcow.go
shiqizng: I would choose the alternative because I would prefer not to keep track of different LogItem versions.

runvnc:
Forgive me for the naive question, but is there any chance of seeing these logs if the transaction is rejected? Or maybe during a dryrun?

daemon/algod/api/server/v2/utils.go
```diff
@@ -266,6 +266,30 @@ func convertToDeltas(txn node.TxnWithStatus) (*[]generated.AccountStateDelta, *g
...
+			// Resolve appid from index
+			var appid uint64
+			if txn.Txn.Txn.ApplicationID == 0 && aidx != nil {
+				appid = *aidx
+			} else {
+				//to be added when app to app call is available
+			}
+			l = append(l, generated.LogItem{
+				Id:    appid,
+				Value: v.Message,
```
algorandskiy, line 284: base64

daemon/algod/api/server/v2/test/helpers.go
```diff
@@ -268,12 +268,12 @@ func testingenv(t testing.TB, numAccounts, numTxs int, offlineAccounts bool) (*d
 	}
```
algorandskiy, line 268: are these changed from go fmt?

daemon/algod/api/server/v2/handlers.go
```diff
@@ -531,14 +532,13 @@ func (v2 *Handlers) PendingTransactionInformation(ctx echo.Context, txid string,
...
-
 		response.AssetIndex = computeAssetIndexFromTxn(txn, v2.Node.Ledger())
 		response.ApplicationIndex = computeAppIndexFromTxn(txn, v2.Node.Ledger())
-
 		response.LocalStateDelta, response.GlobalStateDelta = convertToDeltas(txn)
+		response.Logs = convertToLogItems(txn, response.ApplicationIndex)
 	}
-
 	data, err := encode(handle, response)
+
```
algorandskiy, line 541: nl not needed since this is one logical block

ledger/appcow.go
```diff
@@ -425,6 +425,17 @@ func (cb *roundCowState) DelKey(addr basics.Address, aidx basics.AppIndex, globa
...
 }
 
+// AppendLog adds message in logs
+func (cb *roundCowState) AppendLog(idx uint64, value string) error {
+	cb.logs = append(cb.logs, basics.LogItem{ID: idx, Message: value})
+	return nil
+}
+
+// GetLogs returns logs
+func (cb *roundCowState) GetLogs() []basics.LogItem {
```
algorandskiy, line 435: is it really used somewhere except tests? I do not think a public method needed only for test usage

test/e2e-go/restAPI/restClient_test.go
```diff
@@ -905,3 +935,76 @@ func TestClientPrioritizesPendingTransactions(t *testing.T) {
...
+	tx, err := testClient.MakeUnsignedApplicationCallTx(0, nil, addresses, nil, nil, transactions.NoOpOC, approv, clst, gl, lc, 0)
+	tx.Sender = addr
+	tx.Fee = basics.MicroAlgos{Raw: minTxnFee}
+	tx.FirstValid = firstRound
+	tx.LastValid = lastRound
+	tx.GenesisHash = gh
+
+	txid, err := testClient.SignAndBroadcastTransaction(wh, nil, tx)
+	a.NoError(err)
+	_, err = waitForTransactionV2(t, testClient, someAddress, txid, 60*time.Second)
```
jasonpaulos, line 1003: If you don't use the output of `waitForTransactionV2`, this shouldn't be any different than using the v1 version of the function.

test/e2e-go/restAPI/restClient_test.go
```diff
@@ -905,3 +935,76 @@ func TestClientPrioritizesPendingTransactions(t *testing.T) {
...
+	a.NoError(err)
+	txn, err := testClient.PendingTransactionInformationV2(txid)
+	a.NoError(err)
+	a.NotNil(txn.Logs)
+	a.Equal(29, len(*txn.Logs))
```
jasonpaulos, line 1008: Could you check the contents of the log items too? Since this is an app creation transaction, the app ID in the log items should equal the generated ID from `txn.ApplicationIndex`.

It would be amazing if you could also create and submit a normal app call transaction after this to verify that the app ID is correct in that case too.

jasonpaulos (changes requested):

data/transactions/logic/eval.go
```diff
@@ -159,6 +162,9 @@ type LedgerForLogic interface {
 	DelGlobal(key string) error
 
 	GetDelta(txn *transactions.Transaction) (evalDelta basics.EvalDelta, err error)
+
+	AppendLog(txn *transactions.Transaction, value string) error
```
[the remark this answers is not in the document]
shiqizng: This is because I need to call a txn method to get the app index. https://github.com/shiqizng/go-algorand/blob/opcode/log/ledger/applications.go#L243

daemon/algod/api/server/v2/handlers.go
```diff
@@ -531,13 +532,16 @@ func (v2 *Handlers) PendingTransactionInformation(ctx echo.Context, txid string,
...
 		response.CloseRewards = &txn.ApplyData.CloseRewards.Raw
-
 		response.AssetIndex = computeAssetIndexFromTxn(txn, v2.Node.Ledger())
 		response.ApplicationIndex = computeAppIndexFromTxn(txn, v2.Node.Ledger())
-
 		response.LocalStateDelta, response.GlobalStateDelta = convertToDeltas(txn)
-	}
+		response.Logs, err = convertToLogItems(txn, response.ApplicationIndex)
+		if err != nil {
+			msg := fmt.Sprintf("%v", err)
```
jasonpaulos, line 540: `err.Error()` would also work

ledger/applications_test.go
```diff
@@ -1324,3 +1334,29 @@ func testAppAccountDeltaIndicesCompatibility(t *testing.T, source string, accoun
...
+		ApplicationID: 0,
+		Accounts:      []basics.Address{},
+	}
+	appCall := transactions.Transaction{
+		Type:                     protocol.ApplicationCallTx,
+		ApplicationCallTxnFields: appCallFields,
+	}
+
+	err = l.AppendLog(&appCall, "a")
+	a.NoError(err)
```
jasonpaulos, line 1361: Perhaps call `getLogs` to verify that "a" made it into the logs?

jasonpaulos (approved):
This looks good! I just left some insignificant comments

jasonpaulos (changes requested):
One small thing I noticed

data/transactions/logic/eval.go
```diff
@@ -3154,3 +3162,31 @@ func opAppParamsGet(cx *evalContext) {
...
 	cx.stack = append(cx.stack, stackValue{Uint: exist})
 }
+
+func opLog(cx *evalContext) {
+	last := len(cx.stack) - 1
+
+	if cx.stack[last].argType() != StackBytes {
+		cx.err = fmt.Errorf("log arg 0 wanted type []byte got uint64")
+		return
+	}
```
algorandskiy, line 3172: it is checked automatically in `step()` function but please double check before removal this branch

config/consensus.go
```diff
@@ -422,6 +422,10 @@ var MaxEvalDeltaAccounts int
 // in a StateDelta, used for decoding purposes.
 var MaxStateDeltaKeys int
 
+// MaxLogCalls is the largest number of log calls that may appear in
+// an eval delta, used for decoding purposes.
+var MaxLogCalls int
```
algorandskiy, line 427: it should be consensus param since it is checked in `opLog` and a blind change would introduce compatibility problem.

ledger/appcow.go
```diff
@@ -444,6 +444,17 @@ func (cb *roundCowState) DelKey(addr basics.Address, aidx basics.AppIndex, globa
 	return nil // note: deletion cannot cause us to violate maxCount
 }
 
+// AppendLog adds message in logs
```
algorandskiy, line 447:
```suggestion
// AppendLog adds message in logs. idx is expected to be an index in txn.ForeignApps
```

ledger/applications.go
```diff
@@ -235,3 +238,15 @@ func (al *logicLedger) DelGlobal(key string) error {
...
+	idx, err := txn.IndexByID(txn.ApplicationID)
+	if err != nil {
+		return err
+	}
+	return al.cow.AppendLog(idx, value)
+}
+
+func (al *logicLedger) GetLogs() []basics.LogItem {
+	return al.cow.getLogs()
+}
```
algorandskiy, line 252: I do not see a need for both `GetLogs ` and `getLogs` - please remove. In `logic` tests there is a mock ledger instance available with direct access to logs.

daemon/algod/api/server/v2/dryrun_test.go
```diff
@@ -1091,3 +1091,102 @@ int 1`)
 		logResponse(t, &response)
 	}
 }
+
+func TestDryrunLogs(t *testing.T) {
```
jasonpaulos, line 1095: It would be amazing if you could add a unit test to ensure logged values are still present even if the app call fails. I just tested locally and that didn't seem to be the case.

test/e2e-go/restAPI/restClient_test.go
```diff
@@ -186,6 +188,36 @@ func waitForTransaction(t *testing.T, testClient libgoal.Client, fromAddress, tx
 	}
 }
 
+func waitForTransactionV2(t *testing.T, testClient libgoal.Client, fromAddress, txID string, timeout time.Duration) (tx v1.Transaction, err error) {
```
algorandskiy, line 191: the function is not used, please remove

test/e2e-go/restAPI/restClient_test.go
```diff
@@ -905,3 +936,119 @@ func TestClientPrioritizesPendingTransactions(t *testing.T) {
...
+	a.NoError(err)
+
+	firstRound := basics.Round(params.LastRound + 1)
+	lastRound := basics.Round(params.LastRound + 1000)
+	var gh crypto.Digest
+	copy(gh[:], params.GenesisHash)
+
+	prog := `#pragma version 5
+int 1
+loop: byte "a"
```
algorandskiy, line 973: let's check different values

config/consensus.go
jannotti: I disagree - these are the constants that we use only to bound message sizes.  This constant should *never* be used anywhere else, as these constants are allowed to be conservative, that is, it should be allowed to make this 37 and nothing bad would happen. (It would become possible to send txn objects with 37 log entries, but they should be illegal based on other rules, elsewhere.)

If that's *not* the intent - if this constant is used elsewhere to control the size (for example in the teal opcode), then it should be defined there, and this constant should be set separate.  There could be an argument for making that a consensus parameter rather than defined in logic package.  I can see both sides. But we generally know that any change in the way an opcode works is a consensus upgrade, and we gate them all on a single logicversion parameter, rather than define a bunch of separate ones.

algorandskiy: I'm fine with having it in`logic` similarly to max cumulative log size. But not the way it is done now - set a as a param and used in opcode.

jannotti: So I guess have two constants, and this one should be commented like: "Never decrease this value. It should be the highest allowable number of log messages allowed in any version"

This would be really weird, but I almost want to make it 33 so nobody thinks it's the right constant to use for anything else.

daemon/algod/api/server/v2/dryrun_test.go
jasonpaulos: This seems to be a larger change, so I made a separate issue for it: #2687

daemon/algod/api/server/v2/dryrun_test.go
```diff
@@ -1091,3 +1091,102 @@ int 1`)
...
+int 30
+<
+bnz loop
+`)
+
+	require.NoError(t, err)
+	approval := ops.Program
+	ops, err = logic.AssembleString("int 1")
+	clst := ops.Program
+	ops, err = logic.AssembleString("#pragma version 5 \nint 1")
```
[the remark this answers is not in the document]
shiqizng: the second app was for checking that logs field is omitted when empty.

config/consensus.go
algorandskiy: this needs to be const, no need to initialize it `checkSetAllocBounds`.
```suggestion
const MaxLogCalls = 32
```

`checkSetAllocBounds` might have a check that the value is not greater than 32 (or 33 as JJ suggested)

daemon/algod/api/server/v2/dryrun_test.go
```diff
@@ -1091,3 +1091,109 @@ int 1`)
...
+	for i, m := range logs {
+		if i == 0 {
+			assert.Equal(t, base64.StdEncoding.EncodeToString([]byte("a")), m.Value)
+		} else if i == 1 {
+			assert.Equal(t, base64.StdEncoding.EncodeToString([]byte("b")), m.Value)
+		} else if i == 2 {
+			assert.Equal(t, base64.StdEncoding.EncodeToString([]byte("c")), m.Value)
+		}
+
+	}
```
algorandskiy, line 1191: Calculating values looks more maintainable, check below:
```suggestion
	for i, m := range logs {
		assert.Equal(t, base64.StdEncoding.EncodeToString([]byte(string(rune('a' + i)))), m.Value)
	}
```

If logs are not sorted, use sort.Slice. If logs is map (not sure) then flatten it:
```golang
	type flog struct {
		idx int,
		li LogItem
	}
	for k, v := range logs {
		flatlog = append(flatlog, flog{i, v})
	}
	sort.Slice(flatlog, func(i, j int) {return flatlog[i].idx < flatlog[j].idx})
```

data/basics/teal.go
```diff
@@ -134,6 +143,8 @@ type EvalDelta struct {
 	// When decoding EvalDeltas, the integer key represents an offset into
 	// [txn.Sender, txn.Accounts[0], txn.Accounts[1], ...]
 	LocalDeltas map[uint64]StateDelta `codec:"ld,allocbound=config.MaxEvalDeltaAccounts"`
+
+	Logs []LogItem `codec:"lg,allocbound=config.MaxLogCalls"`
```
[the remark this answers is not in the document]
algorandskiy: +1 update `TestEvalDeltaEqual` test.

data/transactions/logic/eval.go
```diff
@@ -53,6 +53,12 @@ const MaxStringSize = 4096
 // MaxByteMathSize is the limit of byte strings supplied as input to byte math opcodes
 const MaxByteMathSize = 64
 
+// MaxLogSize is the limit of total log size from n log calls in a program
+const MaxLogSize = 1024
+
+// MaxLogCalls is the limit of total log calls during a program execution
+const MaxLogCalls = 32
```
algorandskiy, line 60:
```suggestion
const MaxLogCalls = config.MaxLogCalls
```

data/transactions/logic/eval.go
```diff
@@ -3154,3 +3164,26 @@ func opAppParamsGet(cx *evalContext) {
 	cx.stack[last] = value
 	cx.stack = append(cx.stack, stackValue{Uint: exist})
 }
+
+func opLog(cx *evalContext) {
+	last := len(cx.stack) - 1
+
+	if cx.logCalls == MaxLogCalls {
+		cx.err = fmt.Errorf("too many log calls in program. up to %d is allowed", config.MaxLogCalls)
```
algorandskiy, line 3172: config.MaxLogCalls -> MaxLogCalls

data/transactions/logic/eval_test.go
```diff
@@ -4731,3 +4731,127 @@ func TestBytesConversions(t *testing.T) {
...
+			source: `byte  "a logging message"; log; int 1`,
+			loglen: 1,
+		},
+		{
+			source: `byte  "a logging message"; log; byte  "a logging message"; log; int 1`,
+			loglen: 2,
+		},
+		{
+			source: fmt.Sprintf(`%s int 1`, strings.Repeat(`byte "a logging message"; log;`, config.MaxLogCalls)),
+			loglen: config.MaxLogCalls,
```
algorandskiy, line 4763: config.MaxLogCalls -> MaxLogCalls

daemon/algod/api/server/v2/dryrun_test.go
```diff
@@ -1091,3 +1091,102 @@ int 1`)
...
 	}
 }
+
+func TestDryrunLogs(t *testing.T) {
+	t.Parallel()
+
+	ops, err := logic.AssembleString(`
+#pragma version 5
+int 1
+loop: byte "a"
```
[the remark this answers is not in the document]
algorandskiy: the idea is to use ascii char codes to obtain new char value on each loop iteration.

test/e2e-go/restAPI/restClient_test.go
algorandskiy: I still want to see logging 32 different values in the test:
```
int 0 // loop variable
loop:
dup // copy
int 97 // 0x61 'a'
+ // 97, 98, 99, ...
itob // 'a', 'b', 'c', ...
log
int 1
+
dup
int 32
<
bnz loop
```

daemon/algod/api/server/v2/dryrun_test.go
jasonpaulos: The order of logs matter, so they should not be sorted before comparison

daemon/algod/api/server/v2/dryrun_test.go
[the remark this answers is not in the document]
jasonpaulos: Actually `itob` always produces an 8 byte slice. It would be better to so something like:
```
byte "a"
int 0
dup2
getbyte // convert 0th byte of "a" to integer, 97
int 1
+ // add 1
setbyte // set 0th byte of "a" to the byte given by the integer on the stack (97 + 1, "b")
```

daemon/algod/api/server/v2/dryrun_test.go
algorandskiy: well ok, then `assert.Equal(t, base64.StdEncoding.EncodeToString([]byte(string(rune('a' + i)))), m.Value)` will work just fine

test/e2e-go/restAPI/restClient_test.go
algorandskiy: see Jason's note about `getbyte` above https://github.com/algorand/go-algorand/pull/2529#discussion_r685341889

config/consensus.go
```diff
@@ -422,6 +422,10 @@ var MaxEvalDeltaAccounts int
 // in a StateDelta, used for decoding purposes.
 var MaxStateDeltaKeys int
 
+// MaxLogCalls is the highest allowable log messages that may appear in
+// any version, used for decoding purposes. Never decrease this value.
+const MaxLogCalls = 32
```
algorandskiy, line 427:
```suggestion
// MaxLogCalls is the highest allowable log messages that may appear in
// any version, used for decoding purposes. Never decrease this value.
const MaxEncodedLogCalls = 32
```

data/transactions/logic/eval.go
```diff
@@ -53,6 +53,12 @@ const MaxStringSize = 4096
 // MaxByteMathSize is the limit of byte strings supplied as input to byte math opcodes
 const MaxByteMathSize = 64
 
+// MaxLogSize is the limit of total log size from n log calls in a program
+const MaxLogSize = 1024
+
+// MaxLogCalls is the limit of total log calls during a program execution
+const MaxLogCalls = config.MaxLogCalls
```
algorandskiy, line 60:
```suggestion
const MaxLogCalls = 32
```

data/transactions/logic/eval_test.go
```diff
@@ -4734,3 +4734,127 @@ func TestBytesConversions(t *testing.T) {
...
+		{
+			source: `byte  "a logging message"; log; int 1`,
+			loglen: 1,
+		},
+		{
+			source: `byte  "a logging message"; log; byte  "a logging message"; log; int 1`,
+			loglen: 2,
+		},
+		{
+			source: fmt.Sprintf(`%s int 1`, strings.Repeat(`byte "a logging message"; log;`, config.MaxLogCalls)),
```
algorandskiy, line 4765:
```suggestion
			source: fmt.Sprintf(`%s int 1`, strings.Repeat(`byte "a logging message"; log;`, MaxLogCalls)),
```

data/transactions/logic/eval_test.go
```diff
@@ -4734,3 +4734,127 @@ func TestBytesConversions(t *testing.T) {
...
+			errContains: fmt.Sprintf(">  %d bytes limit", MaxLogSize),
+			runMode:     runModeApplication,
+		},
+		{
+			source:      fmt.Sprintf(`byte  "%s"; log; byte  "%s"; log; byte  "%s"; log; int 1`, msg, msg, msg),
+			errContains: fmt.Sprintf(">  %d bytes limit", MaxLogSize),
+			runMode:     runModeApplication,
+		},
+		{
+			source:      fmt.Sprintf(`%s; int 1`, strings.Repeat(`byte "a"; log;`, config.MaxLogCalls+1)),
```
algorandskiy, line 4817:
```suggestion
			source:      fmt.Sprintf(`%s; int 1`, strings.Repeat(`byte "a"; log;`, MaxLogCalls+1)),
```

algorandskiy (commented):
I'll incorporate these `MaxLogCalls` fixes...

algorandskiy (approved):

data/basics/teal.go
```diff
@@ -165,6 +183,14 @@ func (ed EvalDelta) Equal(o EvalDelta) bool {
 		return false
 	}
 
+	// Logs must be equal
```
jasonpaulos, line 186: You should compare the lengths of the logs before iterating over them

data/basics/teal_test.go
```diff
@@ -221,4 +223,29 @@ func TestEvalDeltaEqual(t *testing.T) {
...
+
+	d1 = EvalDelta{
+		Logs: []LogItem{{ID: 1, Message: "val2"}},
+	}
+	a.False(d1.Equal(d2))
+
+	d1 = EvalDelta{
+		Logs: []LogItem{{ID: 0, Message: "val"}},
+	}
+	a.True(d1.Equal(d2))
```
jasonpaulos, line 250: Suggested test case: compare two `EvalDelta`s with logs `[A]` and `[A, B]` -- I believe this would currently pass because of the length issue I mentioned

data/transactions/application.go
```diff
@@ -233,3 +233,45 @@ func (ac *ApplicationCallTxnFields) IndexByAddress(target basics.Address, sender
...
 	return 0, fmt.Errorf("invalid Account reference %s", target)
 }
+
+// AppIDByIndex converts an integer index into an application id associated with the
+// transaction. Index 0 corresponds to the current app, and an index > 0
+// corresponds to an offset into txn.ForeignApps. Returns an error if the index is
+// not valid.
+func (ac *ApplicationCallTxnFields) AppIDByIndex(i uint64) (basics.AppIndex, error) {
+
+	// Index 0 always corresponds to the sender
```
jasonpaulos, line 243: nit: sender still mentioned here and in function below

ledger/applications.go
```diff
@@ -235,3 +237,11 @@ func (al *logicLedger) DelGlobal(key string) error {
 func (al *logicLedger) GetDelta(txn *transactions.Transaction) (evalDelta basics.EvalDelta, err error) {
 	return al.cow.BuildEvalDelta(al.aidx, txn)
 }
+
+func (al *logicLedger) AppendLog(txn *transactions.Transaction, value string) error {
+	idx, err := txn.IndexByAppID(txn.ApplicationID)
+	if err != nil {
+		return err
+	}
```
jasonpaulos, line 245: Just to be paranoid, could you return an error if idx != 0? It shouldn't happen, but it would be really bad if it did happen

test/e2e-go/features/transactions/application_test.go
```diff
@@ -0,0 +1,127 @@
...
+	"github.com/algorand/go-algorand/protocol"
+	"github.com/algorand/go-algorand/test/framework/fixtures"
+)
+
+func checkEqual(expected []basics.LogItem, actual []basics.LogItem) bool {
+	if len(expected) != len(actual) {
+		return false
+	}
+	for i, e := range expected {
+		if e.ID != actual[i].ID || e.Message != actual[i].Message {
```
jasonpaulos, line 39: nit: you could use the `Equals` method you made here

test/e2e-go/restAPI/restClient_test.go
```diff
@@ -905,3 +907,128 @@ func TestClientPrioritizesPendingTransactions(t *testing.T) {
...
+	a.NoError(err)
+	_, err = waitForTransaction(t, testClient, someAddress, txid, 60*time.Second)
+	a.NoError(err)
+	txn, err = testClient.PendingTransactionInformationV2(txid)
+	a.NoError(err)
+	a.NotNil(txn.Logs)
+	a.Equal(32, len(*txn.Logs))
+	for i, l := range *txn.Logs {
+		a.Equal(expectedAppID, l.Id)
+		assert.Equal(t, base64.StdEncoding.EncodeToString([]byte(string(rune('B'+i)))), l.Value)
```
jasonpaulos, line 1031: nit: is it necessary to import `assert` instead of using `a.Equal`?

jasonpaulos (changes requested):

jasonpaulos (dismissed):
`partitiontest.PartitionTest` may need to be added to tests to satisfy reviewdog (I don't know the purpose of that), but this looks good to me

jasonpaulos (approved):

jannotti (approved):

PR #2533, opened 2021-07-13
[TEST_DRAFT] Update README.md 2 with updated branch

algobarb:


algobarb:
Was testing why some builds didn't run Travis, but it seems to be an actual Travis-CI issue that is affecting some users from triggerring builds on GitHub commits.

PR #2536, opened 2021-07-13
Added Txnsync Service Unit Tests

AlgoStephenAkiki:
Added unit tests and fixed some spelling errors

### DIFF
--- a/txnsync/service_test.go
+++ b/txnsync/service_test.go
@@ -1,3 +1,19 @@
+// Copyright (C) 2019-2021 Algorand, Inc.
+// This file is part of go-algorand
+//
+// go-algorand is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as
+// published by the Free Software Foundation, either version 3 of the
+// License, or (at your option) any later version.
+//
+// go-algorand is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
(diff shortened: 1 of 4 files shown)


txnsync/service_test.go
```diff
@@ -0,0 +1,138 @@
...
+	hashDigest := crypto.Hash([]byte{0x41, 0x6b, 0x69, 0x6b, 0x69})
+
+	service := MakeTransactionSyncService(mLogger, &mNodeConnector, true, "GENID", hashDigest, cfg, mThreadPool)
+
+	a.NotNil(service)
+
+	service.Start()
+	service.Stop()
+
+	a.True(calledEventsInNodeConnector)
```
tsachiherman, line 91: I think that this test might randomally fail and could result in a data race.

txnsync/service_test.go
```diff
@@ -0,0 +1,138 @@
...
+}
+
+// TestMakeTransactionSyncService tests that an appropriate transaction sync service was made
+func TestMakeTransactionSyncService(t *testing.T) {
+
+	a := require.New(t)
+
+	mLogger := mockLogger{}
+	mNodeConnector := &mockNodeConnector{}
+	cfg := config.Local{}
```
tsachiherman, line 105: Avoid using the empty config. Instead use GetDefaultLocal

tsachiherman (changes requested):
Looks good. Few small requests.

txnsync/service_test.go
AlgoStephenAkiki: Isn't the Events() function in mainloop.go:124 called before any of the channel switch statements?  The Stop() function also waits for the goroutine to run right?

tsachiherman (approved):

PR #2537, opened 2021-07-13
test conditional slack alert

algobarb:
## Summary
Adding the conditional slack alert. 

Travis is currently unable to **only** notify on failures of a particular branch, however, it can notify on us on the build of a particular branch (both success and failures). Since we are also considering CircleCI, I think it would be best to just add the slack notification into the travis file now for the rel/nightly branch and notify the #devops-jenkins channel (where most of our jobs status activity goes) when it both succeeds/fails. The "successes" will also help us determine whether or not the test ran at all.

## Test Plan

Test the alert on master branch and check that it alerted on slack and then will change the branch to specify "rel/nightly".


algojack:
this part doesn't work to notify us only on failures? https://docs.travis-ci.com/user/notifications#changing-notification-frequency

algobarb:
> this part doesn't work to notify us only on failures? https://docs.travis-ci.com/user/notifications#changing-notification-frequency

The limitation is on the changing-notification-frequency of a particular branch. It can only be set at the "slack" level, meaning on_failures of all builds. See the "Note" in the Changing-notification-frequency section. Other people on forums have also mentioned this as a limitation and will resort to using a webhook instead.

algojack (approved):

bricerisingalgorand (approved):

algobarb:
@onetechnical could you take a look and merge it (as an authorized user) if it looks good?

PR #2539, opened 2021-07-13
use periodicSyncLogger for cleaner test runs

algonautshant:
## Summary
Use periodicSyncLogger in a test to avoid unnecessary log outputs. 

## Test Plan
This is a change to a test.


tsachiherman (approved):
Lgtm

PR #2541, opened 2021-07-14
Stack manipulation opcodes - uncover and cover

AyAggarwal:
This PR adds two opcodes.
cover:  "remove top of stack, and place it down the stack such that N elements are above it",
uncover:  "remove the value at depth N in the stack and shift above items down so the Nth deep value is on top of the stack"


AyAggarwal:
Hello All, the PR build is failing due to "scripts/travis/codegen_verification.sh". Will look into this tomorrow morning (PST)

data/transactions/logic/eval.go
```diff
@@ -1700,6 +1700,40 @@ func opDig(cx *evalContext) {
 	cx.stack = append(cx.stack, sv)
 }
 
+func opCover(cx *evalContext) {
+	depth := int(uint(cx.program[cx.pc+1]))
+	idx := len(cx.stack) - 1 - depth
+	// Need to check stack size explicitly here because checkArgs() doesn't understand dig
```
jannotti, line 1706: Cover, not dig

data/transactions/logic/eval.go
```diff
@@ -1700,6 +1700,40 @@ func opDig(cx *evalContext) {
...
+	// so we can't expect our stack to be prechecked.
+	if idx < 0 {
+		cx.err = fmt.Errorf("uncover %d with stack size = %d", depth, len(cx.stack))
+		return
+	}
+	topIndex := len(cx.stack) - 1
+	Idxsv := cx.stack[idx]
```
jannotti, line 1730: Use lower case for locals

jannotti (commented):
I think the shifts can be done clearer and faster with copy()

data/transactions/logic/eval.go
AyAggarwal: done

AyAggarwal:
updated to use copy

jannotti:
I'm realizing there is a tricky issue with these opcodes - normally our assembler and the Check() call try to keep track of the types on the stack. If you push an int, then a bytes, then use +, you will get an error. (In the presence of branches, we downgrade this error to a warning).

This can only happen because each opcode (so far) only manipulates the top of the stack. The OpSpec describes that manipulation in terms of types. But these opcodes are different. They manipulate the stack deeply, and potentially change the types of many elements (the element doesn't change, but the type at a particular location does, as they all shift).

We need to figure out how that can be made to work. It might just be that they need their own check function (like the branches have) or it might be that they can explicitly manipulate the type stack during assembly. They might need both, or even more.

First step, let's get a unit test in that shows the problem. Something like:

`int 4; byte "john"; int 5; cover 2; pop; +`

I think this will not type check, but it is correct (if I haven't messed up).

data/transactions/logic/eval.go
```diff
@@ -1700,6 +1700,36 @@ func opDig(cx *evalContext) {
 	cx.stack = append(cx.stack, sv)
 }
 
+func opCover(cx *evalContext) {
+	depth := int(uint(cx.program[cx.pc+1]))
```
jdtzmn, line 1704: Why use multiple type casts?

data/transactions/logic/eval.go
```diff
@@ -1700,6 +1700,36 @@ func opDig(cx *evalContext) {
...
+	idx := len(cx.stack) - 1 - depth
+	// Need to check stack size explicitly here because checkArgs() doesn't understand cover
+	// so we can't expect our stack to be prechecked.
+	if idx < 0 {
+		cx.err = fmt.Errorf("cover %d with stack size = %d", depth, len(cx.stack))
+		return
+	}
+	topidx := len(cx.stack) - 1
```
algochoi, line 1712: nit: `topIdx`

data/transactions/logic/eval.go
```diff
@@ -1700,6 +1700,36 @@ func opDig(cx *evalContext) {
...
+func opUncover(cx *evalContext) {
+	depth := int(uint(cx.program[cx.pc+1]))
+	idx := len(cx.stack) - 1 - depth
+	// Need to check stack size explicitly here because checkArgs() doesn't understand uncover
+	// so we can't expect our stack to be prechecked.
+	if idx < 0 {
+		cx.err = fmt.Errorf("uncover %d with stack size = %d", depth, len(cx.stack))
+		return
+	}
+	topidx := len(cx.stack) - 1
```
algochoi, line 1727: nit: `topIdx`

data/transactions/logic/eval_test.go
```diff
@@ -4134,6 +4134,22 @@ func TestDig(t *testing.T) {
 	testPanics(t, "int 3; int 2; int 1; dig 11; int 2; ==; return", 3)
 }
 
+func TestCover(t *testing.T) {
```
algochoi, line 4137: I think you *might* have to run `make sanity` on the root folder (`go-algorand`) if you haven't done so already -- I took a quick look at the Travis error and apparently there might be formatting issues.

jdtzmn, line 4137: It would be nice to see test cases here that check that other elements are shifted correctly

jdtzmn (changes requested):
Mostly good, nice work on this! I left a few comments.

data/transactions/logic/eval.go
```diff
@@ -1700,6 +1700,36 @@ func opDig(cx *evalContext) {
...
+	idxsv := cx.stack[idx]
```
algochoi, line 1728: nit: I think this should be `sv` (based on the `dig` opcode)

algochoi (commented):
Good work! Just very minor comments. 

Based on the Travis error, maybe try running `make sanity` on the root folder or `go fmt` again?

data/transactions/logic/eval_test.go
algochoi: It looks like the exact statements are: 
```
  gofmt -w /home/travis/build/[secure]/go-[secure]/data/transactions/logic/eval.go
  gofmt -w /home/travis/build/[secure]/go-[secure]/data/transactions/logic/evalStateful_test.go
```
but I think `make sanity` will do the same thing?

data/transactions/logic/eval.go
```diff
@@ -1700,6 +1700,36 @@ func opDig(cx *evalContext) {
...
+	depth := int(uint(cx.program[cx.pc+1]))
+	idx := len(cx.stack) - 1 - depth
+	// Need to check stack size explicitly here because checkArgs() doesn't understand cover
+	// so we can't expect our stack to be prechecked.
+	if idx < 0 {
+		cx.err = fmt.Errorf("cover %d with stack size = %d", depth, len(cx.stack))
+		return
+	}
+	topidx := len(cx.stack) - 1
+	topsv := cx.stack[topidx]
```
algochoi, line 1713: nit: maybe `sv` (based on the `dig` opcode)?

data/transactions/logic/eval_test.go
AyAggarwal: ok I just did that, hopefully it passes

AyAggarwal: will add these

data/transactions/logic/eval.go
AyAggarwal: copied this from dig, but I'll dig into it

jannotti:
It looks to me like all we need to do to address the type checking is to manipulate the type stack in the assemble function for cover/uncover. Currently they use asmDefault. So they'll need custom assemblers, but it should pretty much just call the default assembler, and manipulate the stack.

data/transactions/logic/eval.go
AyAggarwal: seems like it creates a type mismatch error if its not casted back to an int. without uint it becomes negative

jdtzmn: Aren't uint64 immediates required to be non-negative?

jannotti:
They are byte immediates.

Still, it seems as though this double cast is unneeded.  I recall seeing it
before, messing with it a bit, and for some reason leaving it.  I can't see
why, right now.  In C, I guess something similar would be needed, as char
is not defined to be signed or unsigned.  But byte is
definitely unsigned in Go, and the simple cast to int should preserve the
sign because there's room in int for the entire range of byte.

data/transactions/logic/assembler.go
```diff
@@ -1101,6 +1101,63 @@ func typeDig(ops *OpStream, args []string) (StackTypes, StackTypes) {
...
+	}
+	idx := len(ops.typeStack) - depth
+	if idx >= 0 {
+		sv := ops.typeStack[idx]
+		for i := idx + 1; i < len(ops.typeStack); i++ {
+			returns[i-idx-1] = ops.typeStack[i]
+		}
+		returns[len(returns)-1] = sv
+	}
+	fmt.Println(returns)
```
jdtzmn, line 1157: nit: `fmt.Println`

data/transactions/logic/assembler.go
```diff
@@ -1101,6 +1101,63 @@ func typeDig(ops *OpStream, args []string) (StackTypes, StackTypes) {
...
+	anys := make(StackTypes, depth)
+	for i := range anys {
+		anys[i] = StackAny
+	}
+	returns := make(StackTypes, depth)
+	for i := range returns {
+		returns[i] = StackAny
+	}
+	idx := len(ops.typeStack) - depth
+	if idx >= 0 {
```
jdtzmn, line 1122: Will the idx here ever be less than 0? Won't that be caught in the `opCover` function?

data/transactions/logic/assembler.go
```diff
@@ -1101,6 +1101,63 @@ func typeDig(ops *OpStream, args []string) (StackTypes, StackTypes) {
...
+		sv := ops.typeStack[len(ops.typeStack)-1]
+		for i := idx; i < len(ops.typeStack)-1; i++ {
```
jdtzmn, line 1124: Could this be shifted more cleanly using `copy()`?

data/transactions/logic/eval.go
```diff
@@ -1699,6 +1699,36 @@ func opDig(cx *evalContext) {
...
+func opCover(cx *evalContext) {
+	depth := int(cx.program[cx.pc+1])
+	idx := len(cx.stack) - 1 - depth
+	// Need to check stack size explicitly here because checkArgs() doesn't understand cover
+	// so we can't expect our stack to be prechecked.
+	if idx < 0 {
+		cx.err = fmt.Errorf("cover %d with stack size = %d", depth, len(cx.stack))
+		return
+	}
+	topIdx := len(cx.stack) - 1
```
jdtzmn, line 1711: This may read better if you move `topIdx` above `idx` like so:
```go
topIdx := len(cx.stack - 1)
idx := topIdx - depth
```

jdtzmn (changes requested):

data/transactions/logic/eval.go
```diff
@@ -1699,6 +1699,36 @@ func opDig(cx *evalContext) {
 	cx.stack = append(cx.stack, sv)
 }
 
+func opCover(cx *evalContext) {
+	depth := int(cx.program[cx.pc+1])
+	idx := len(cx.stack) - 1 - depth
+	// Need to check stack size explicitly here because checkArgs() doesn't understand cover
+	// so we can't expect our stack to be prechecked.
```
jannotti, line 1706: This may not be true any more with the changes to checkStack.  I think that now opCover _is_ sure to have a big enough stack to work with.

data/transactions/logic/eval_test.go
```diff
@@ -4204,6 +4204,25 @@ func TestDig(t *testing.T) {
...
+}
+
+func TestUncover(t *testing.T) {
+	t.Parallel()
+	testAccepts(t, "int 4; int 3; int 2; int 1; uncover 2; int 3; ==; return", 5)
+	testAccepts(t, "int 4; int 3; int 2; int 1; uncover 3; int 4; ==; return", 5)
+	testAccepts(t, "int 4; int 3; int 2; int 1; uncover 3; pop; int 1; ==; return", 5)
+	testAccepts(t, "int 4; int 3; int 2; int 1; uncover 3; pop; pop; int 2; ==; return", 5)
+	testAccepts(t, "int 1; int 3; int 2; int 1; uncover 3; pop; pop; int 2; ==; return", 5)
+	testPanics(t, obfuscate("int 4; int 3; int 2; int 1; uncover 11; int 3; ==; return"), 5)
```
jannotti, line 4223: Can you add one that panics, but just barely? In other words, confirm theres no off-by-one error by uncovering with an argument that's just one too high.

jannotti (approved):

data/transactions/logic/assembler.go
jannotti: It could, but I'm to blame for setting the precedent with `dig`. We need cleaner helper functions for building these arrays in the type* functions.  Then we can also do `select`, `dup`, `dup`.

jdtzmn: Agreed. Does it make sense to include those changes in this PR or should it be done in a separate one?

jannotti: I think we should do those separately.  I'm still mulling over the right interface, but I think I'm imagining two things, one I'm fairly sure about, and one less so.

Fairly sure: the typeCheck functions should check the appropriate depth first, and return early if the stack is short (with an appropriately sized slice to cause checkStack to complain about length).  This means the rest of the typeCheck function can assume the stack is deep enough, and it'll be easier to write the code (using copy!) that returns slices that match parts of the existing stack.

Less sure: the interface should be reworked a little so that immediates are parsed early in one place, and supplied in a structured way to the typeCheck and assemble functions.  That way it only happens once, and assemble functions are no longer expected to do that parsing or error checking.

AyAggarwal: I agree with the first one for sure, the second point is good as well - would that still mean if a program failed in typeCheck it fails during assembly?

data/transactions/logic/eval.go
AyAggarwal: good point

data/transactions/logic/eval.go
AyAggarwal: good point, i'll re visit this

data/transactions/logic/assembler.go
AyAggarwal: I believe this check occurs first, as jj said below opCover should now be sure that it has the correct size stack

AyAggarwal: This is actually because the arrays will create out of bounds error if you don't check here. It doesn't panic though

jdtzmn (approved):

data/transactions/logic/doc.go
```diff
@@ -246,7 +250,7 @@ var OpGroups = map[string][]string{
 	"Byteslice Arithmetic": {"b+", "b-", "b/", "b*", "b<", "b>", "b<=", "b>=", "b==", "b!=", "b%"},
 	"Byteslice Logic":      {"b|", "b&", "b^", "b~"},
 	"Loading Values":       {"intcblock", "intc", "intc_0", "intc_1", "intc_2", "intc_3", "pushint", "bytecblock", "bytec", "bytec_0", "bytec_1", "bytec_2", "bytec_3", "pushbytes", "bzero", "arg", "arg_0", "arg_1", "arg_2", "arg_3", "txn", "gtxn", "txna", "gtxna", "gtxns", "gtxnsa", "global", "load", "store", "gload", "gloads", "gaid", "gaids"},
-	"Flow Control":         {"err", "bnz", "bz", "b", "return", "pop", "dup", "dup2", "dig", "swap", "select", "assert", "callsub", "retsub"},
+	"Flow Control":         {"err", "bnz", "bz", "b", "return", "pop", "dup", "dup2", "dig", "cover", "uncover", "swap", "select", "assert", "callsub", "retsub"},
```
algorandskiy, line 253: just wondering how did `"pop", "dup", "dup2", "dig", "cover", "uncover"` end up to be flow control

data/transactions/logic/eval_test.go
```diff
@@ -4400,6 +4400,27 @@ func TestDig(t *testing.T) {
...
+	testAccepts(t, "int 4; int 3; int 2; int 1; cover 1; int 2; ==; return", 5)
+	testAccepts(t, "int 4; int 3; int 2; int 1; cover 2; int 2; ==; return", 5)
+	testAccepts(t, "int 4; int 3; int 2; int 1; cover 2; pop; pop; int 1; ==; return", 5)
+	testPanics(t, obfuscate("int 4; int 3; int 2; int 1; cover 11; int 2; ==; return"), 5)
+	testPanics(t, obfuscate("int 4; int 3; int 2; int 1; cover 4; int 2; ==; return"), 5)
+}
+
+func TestUncover(t *testing.T) {
+	t.Parallel()
+	testAccepts(t, "int 4; int 3; int 2; int 1; uncover 2; int 3; ==; return", 5)
```
algorandskiy, line 4415: please add `unvcover 0` test as an edge case test

data/transactions/logic/eval.go
```diff
@@ -1699,6 +1699,36 @@ func opDig(cx *evalContext) {
...
+func opUncover(cx *evalContext) {
+	depth := int(cx.program[cx.pc+1])
+	idx := len(cx.stack) - 1 - depth
+	// Need to check stack size explicitly here because checkArgs() doesn't understand uncover
+	// so we can't expect our stack to be prechecked.
+	if idx < 0 {
+		cx.err = fmt.Errorf("uncover %d with stack size = %d", depth, len(cx.stack))
+		return
+	}
+	topIdx := len(cx.stack) - 1
```
algorandskiy, line 1726: make it the same as prev function
```suggestion
	depth := int(cx.program[cx.pc+1])
	topIdx := len(cx.stack) - 1
	idx := topIdx - depth
	// Need to check stack size explicitly here because checkArgs() doesn't understand uncover
	// so we can't expect our stack to be prechecked.
	if idx < 0 {
		cx.err = fmt.Errorf("uncover %d with stack size = %d", depth, len(cx.stack))
		return
	}
```

algorandskiy (commented):
Looks good, one more test request and a minor code style fix

data/transactions/logic/doc.go
jannotti: I think it's because it used to only be `pop` and `dup` so we didn't create a category for stack manipulation.

jannotti (approved):
Looks good.

PR #2544, opened 2021-07-14
build: fix travis nightly build failure

tsachiherman:
## Summary

Travis nightly build failed due to no-output for over 10m from build process.

## Test Plan

Run on travis.

### DIFF
--- a/scripts/travis/build_test.sh
+++ b/scripts/travis/build_test.sh
@@ -15,7 +15,7 @@ SCRIPTPATH="$( cd "$(dirname "$0")" ; pwd -P )"
 
 if [ "${USER}" = "travis" ]; then
     # we're running on a travis machine
-    "${SCRIPTPATH}/build.sh" --make_debug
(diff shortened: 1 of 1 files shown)


onetechnical (approved):
Looks fine, but do we think it'll pass in 30 minutes?

tsachiherman:
> Looks fine, but do we think it'll pass in 30 minutes?

No; I don't think that we'll pass the 30 minutes. I don't think that the tests would run for 90 minutes either.

Note that this is a high-watermark threshold. I already know that in some cases it clearly takes more than 10 minutes for the go-compiler build, so I figured that for the entire compilation ( including libsodium ), I should probably pad it generously.

winder (approved):
Find by me if we need it. Did something change in the build? I don't think I removed this, so the requirement is new.

tsachiherman:
> Find by me if we need it. Did something change in the build? I don't think I removed this, so the requirement is new.

It used to retry the entire execution 3 times.. and we moved that to the test phase only.

Issue #2202, opened 2021-05-26
Prototype and benchmark assets table with WITHOUT ROWID statement


algorandskiy:
### Idea

Instead of asset groups and manual groups management, use the following layout
```
CREATE TABLE creatablestore(
    address_id INTEGER NOT NULL,
    creatable integer NOT NULL,
    ctype integer NOT NULL,
    data BLOB NOT NULL,
    PRIMARY KEY(address_id, creatable)
) WITHOUT ROWID
```

[Preliminary benchmarks](https://github.com/algorandskiy/go-algorand/tree/storagetableperformance) showed it might perform better than asset groups.

Pros:
1. ~2x less code then groups-based implementation
2. Potential smaller balance record => less memory pressure

Cons:
1. the need to store numeric address_id in `accountbase` table
2. might require more storage space comparing to groups table

### Acceptance
1. PoC
2. Run BenchmarkReadingRandomBalancesDiskLarge on PoC vs groups

algorandskiy:
Benchmark results:

1. Without rowid
```
BenchmarkReadingRandomBalancesDiskLarge/holdings=1-8         	     660	  14,543,377 ns/op
BenchmarkReadingRandomBalancesDiskLarge/holdings=512-8       	     512	  11,894,420 ns/op
BenchmarkReadingRandomBalancesDiskLarge/holdings=1000-8      	     494	  11,344,862 ns/op
BenchmarkReadingRandomBalancesDiskLarge/holdings=2000-8      	     268	   6,051,561 ns/op
BenchmarkReadingRandomBalancesDiskLarge/holdings=5000-8      	     512	  11,227,518 ns/op
BenchmarkReadingRandomBalancesDiskLarge/holdings=10000-8     	     526	  11,282,919 ns/op
BenchmarkReadingRandomBalancesDiskLarge/holdings=100000-8    	     783	  24,083,647 ns/op
```

2. Asset groups
```
BenchmarkReadingRandomBalancesDiskLarge/holdings=1_simple=true-8         	     976	  12,242,124 ns/op
BenchmarkReadingRandomBalancesDiskLarge/holdings=512_simple=true-8       	     703	   8,592,828 ns/op
BenchmarkReadingRandomBalancesDiskLarge/holdings=1000_simple=true-8      	     640	  11,163,492 ns/op
BenchmarkReadingRandomBalancesDiskLarge/holdings=2000_simple=false-8     	     878	  13,292,007 ns/op
[...]

PR #2546, opened 2021-07-14
1k+ assets: without rowid PoC

algorandskiy:
## Summary

This a PoC for `Assets` and `AssetParams` detached storage in a separate `accountcreatabledata` table. This is much simpler than the original 1k+ assets groups-based implementation (most of PersistedAccountData and reconciliation logic are not needed anymore), although it has almost the same code in cow/apply.

TODO: benchmark / figure out if it can be improved / remove leftovers from groups-based implementation.

## Test Plan

Existing ledger tests adjusted

### DIFF
--- a/ledger/accountdb.go
+++ b/ledger/accountdb.go
@@ -1675,6 +1675,10 @@ func accountsNewUpdate(qabu, qabq, qaeu, qaei, qaed *sql.Stmt, addr basics.Addre
 		}
 	}
 
+	if err != nil {
+		return updatedAccounts, err
+	}
+
 	// same logic as above but for asset params
 	if delta.old.pad.NumAssetParams() <= assetsThreshold && len(delta.new.AssetParams) <= assetsThreshold {
 		// AccountData assigned above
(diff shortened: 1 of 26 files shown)


algorandskiy:
@tolikzinovyev FYI

ledger/accountdb.go
```diff
@@ -825,10 +695,15 @@ func removeEmptyAccountData(tx *sql.Tx, queryAddresses bool) (num int64, address
...
 
-func createAccountExtTable(tx *sql.Tx) error {
-	tableCreate := `CREATE TABLE IF NOT EXISTS accountext (
-		id integer primary key,
-		data blob)`
+func updateToExtendedAccountGadgetTable(tx *sql.Tx) error {
+	tableCreate := `CREATE TABLE IF NOT EXISTS accountcreatabledata (
+		addrid integer NOT NULL,
+		creatable integer NOT NULL,
+		ctype integer NOT NULL,
```
tolikzinovyev, line 702: Performance will be better if we had multiple tables instead of a generic one containing `ctype`.

algorandskiy: proof?

tolikzinovyev: No rowid tables store the whole row data in the intermediary b-tree nodes. By increasing the size of the data, b-tree nodes will store less nodes, so the height of the tree will be bigger.

algorandskiy:
Benchmark data is at #2202

PR #2305, opened 2021-06-17
Deploy centos8 package

egieseke:
## Summary

Added support for a separate package for the centos8 stream operating system

## Test Plan

Ran pipeline to build centos 7 and centos 8 stream packages.

PR #2547, opened 2021-07-14
Replace yum-cron with dnf-automatic to support centos 8

onetechnical:
## Summary

Both CentOS 7 and CentOS 8 support dnf-automatic, so using that for cron update is preferred. This should enable using the same package on both systems.

This PR is meant to replace #2305.

## Test Plan

- Install and verify on CentOS 7 using yum and dnf
- Install and verify on CentOS 8 using dnf
- Install stable version of algorand on CentOS 7, then install this package using yum and dnf

Verify that the yum cron file is removed and the dnf cron file is added to /etc/cron.hourly. Test the command in that file.


algobarb (approved):
I haven't tested it yet, but the updates for the bugs look pretty straight-forward.

bricerisingalgorand (approved):

PR #2555, opened 2021-07-14
add handling for non-utf 8 asset names

tsachiherman:
## Summary

The asset parameter fields name, unit name and url might contain non-utf 8 printable characters. This PR modifies the
REST API so that the existing field would only be returned in case the entire string is a utf-8 printable string. To complement that, a _b64 version of the field would be provided, allowing the client to support non-printable strings.

## Test Plan

Use existing unit tests.


daemon/algod/api/server/v1/handlers/handlers.go
```diff
@@ -163,16 +164,33 @@ func participationKeysEncode(r basics.AccountData) *v1.Participation {
...
 	paramsModel := v1.AssetParams{
 		Total:         params.Total,
 		DefaultFrozen: params.DefaultFrozen,
 		Decimals:      params.Decimals,
 	}
 
-	paramsModel.UnitName = strings.TrimRight(string(params.UnitName[:]), "\x00")
-	paramsModel.AssetName = strings.TrimRight(string(params.AssetName[:]), "\x00")
-	paramsModel.URL = strings.TrimRight(string(params.URL[:]), "\x00")
+	paramsModel.UnitName = sanitizePrintableUTF8String(params.UnitName)
```
algonautshant, line 191: Is this not backwards compatible? What used to be some value will be empty now?

daemon/algod/api/server/v2/account.go
```diff
@@ -422,6 +424,23 @@ func AppParamsToApplication(creator string, appIdx basics.AppIndex, appParams *b
 	return app
 }
 
+// sanitizePrintableUTF8String checks to see if the entire string is a UTF8 printable string.
+// If this is the case, the string is returned as is. Otherwise, the empty string is returned.
+func sanitizePrintableUTF8String(in string) string {
```
algonautshant, line 429: What is the reason for not sharing this with daemon/algod/api/server/v1/handlers/handlers.go ?

daemon/algod/api/algod.oas3.yml
```diff
@@ -1027,9 +1027,14 @@
             "type": "string"
           },
           "name": {
-            "description": "\\[an\\] Name of this asset, as supplied by the creator.",
+            "description": "\\[an\\] Name of this asset, as supplied by the creator. Field presents only if the asset name is composed of printable utf-8 characters.",
```
algonautshant, line 1030: I suggest saying "Available only if ..." instead of "Field presents only if ..."

daemon/algod/api/server/v2/account.go
tsachiherman: v1 is going to be deprecated, and when that happen, the code-duplication would no longer be an issue. But I agree that in general we should avoid code duplication.

daemon/algod/api/server/v1/handlers/handlers.go
tsachiherman: it does break backward compatibility - but only for the cases of non uff-8; currently, I'm unaware of any such cases.

daemon/algod/api/server/v2/account.go
winder: You can put this in `v2/utils.go`.

daemon/algod/api/server/v2/account.go
```diff
@@ -422,6 +424,23 @@ func AppParamsToApplication(creator string, appIdx basics.AppIndex, appParams *b
...
+	// check to see if all the characters in the input string are utf-8.
+	if !utf8.ValidString(in) {
+		return ""
+	}
+	// iterate throughout all the characters in the string to see if they are all printable.
+	for _, c := range in {
+		// is this a printable character ?
+		if !unicode.IsPrint(c) {
+			return ""
+		}
```
winder, line 439: You can skip the `utf8.ValidString` call and do both checks in one iteration of the string
```suggestion
		}
		if c == utf8.RuneError {
			return ""
		}
```

daemon/algod/api/algod.oas3.yml
```diff
@@ -1040,12 +1045,22 @@
             "x-algorand-format": "uint64"
           },
           "unit-name": {
-            "description": "\\[un\\] Name of a unit of this asset, as supplied by the creator.",
+            "description": "\\[un\\] Name of a unit of this asset, as supplied by the creator. Field presents only if the name of a unit of this asset is composed of printable utf-8 characters.",
```
winder, line 1048:
```suggestion
            "description": "\\[un\\] Name of a unit of this asset, as supplied by the creator. Field present only if the asset unit name is composed of printable utf-8 characters.",
```

daemon/algod/api/algod.oas3.yml
winder: There are several places in the current descriptions with phrasing like `This is only included when`, and `This endpoint is only enabled when`

daemon/algod/api/server/v2/utils.go
```diff
@@ -263,3 +265,17 @@ func convertToDeltas(txn node.TxnWithStatus) (*[]generated.AccountStateDelta, *g
 
 	return localStateDelta, stateDeltaToStateDelta(txn.ApplyData.EvalDelta.GlobalDelta)
 }
+
+// sanitizePrintableUTF8String checks to see if the entire string is a UTF8 printable string.
+// If this is the case, the string is returned as is. Otherwise, the empty string is returned.
+func sanitizePrintableUTF8String(in string) string {
```
brianolson, line 271: I feel like "sanitize" is the wrong connotation here, I'd expect that to return the input string but without a few bad parts. `printableOrEmpty()` ? `squashIfAnyNonPrintable()` ?

winder (approved):
Updates LGTM

daemon/algod/api/server/v2/utils.go
winder: Ideally I think this would return a bool and be called "isPrintableUTF8", but then you couldn't use it inline

daemon/algod/api/server/v2/account.go
```diff
@@ -430,9 +449,12 @@ func AssetParamsToAsset(creator string, idx basics.AssetIndex, params *basics.As
 		Total:         params.Total,
 		Decimals:      uint64(params.Decimals),
 		DefaultFrozen: &frozen,
-		Name:          strOrNil(params.AssetName),
-		UnitName:      strOrNil(params.UnitName),
-		Url:           strOrNil(params.URL),
+		Name:          strOrNil(sanitizePrintableUTF8String(params.AssetName)),
+		NameB64:       strOrNil(base64.StdEncoding.EncodeToString([]byte(params.AssetName))),
```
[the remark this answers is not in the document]
tsachiherman: No - we're *storing* one copy, but returning (sometimes) both, so that clients that use b64 could always decode that and could ignore the non-b64 version.

daemon/algod/api/server/v2/utils.go
tsachiherman: I like the `strOrNil` patterns we've used below as it avoids the testing patten.
The `printableUTF8OrEmpty` suggestion makes sense.
thanks!

winder (approved):

algonautshant (approved):
Looks great.

PR #2558, opened 2021-07-15
lruaccounts-no-copy

algonathan:
… on write

## Summary
improved speed of lruaccounts write functionality by avoiding copying persistedAccountData in writes


tsachiherman:
@algonathan - I need to spend some more time on this one. The intent is obviously good here, but I'm trying to verify there are no re-use of the pointers passed to the `write` method, since the `write` method also stores the pointer internally.
( in other words - I'm trying to convince myself that the input is immutable )

algorandskiy:
@algonathan what was the perf gain?

algonathan:
@algorandskiy Hi, I've looked through some old messages I've sent Tsachi. It should be about 100ns per op. 
If I remember correctly, it should be a speedup of about 1.5~2x to the operation itself.

algonathan:
@algonoah, it awaits review, i think this PR can be ignored and closed though.

PR #2565, opened 2021-07-16
txnsync: reduce the amount of logging

tsachiherman:
## Summary

Add a new config option `EnableVerbosedTransactionSyncLogging` ( defaults to false ), to allow us to control whether we want to log incoming/outgoing messages. These messages are being sent at a high frequency ( especially on relays ), and would bloat the `node.log` file very quickly.

## Test Plan

Unit tests updated.

### DIFF
--- a/test/testdata/configs/config-v17.json
+++ b/test/testdata/configs/config-v17.json
@@ -44,6 +44,7 @@
     "EnableProfiler": false,
     "EnableRequestLogger": false,
     "EnableTopAccountsReporting": false,
+    "EnableVerbosedTransactionSyncLogging": false,
     "EndpointAddress": "127.0.0.1:0",
     "FallbackDNSResolverAddress": "",
     "ForceFetchTransactions": false,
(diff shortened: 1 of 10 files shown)


txnsync/incoming.go
```diff
@@ -250,7 +250,9 @@ func (s *syncState) evaluateIncomingMessage(message incomingMessage) {
 		// if the peer's round is more than a single round behind the local node, then we don't want to
 		// try and load the transactions. The other peer should first catch up before getting transactions.
 		if (peer.lastRound + 1) < s.round {
-			s.log.Infof("Incoming Txsync #%d late round %d", seq, peer.lastRound)
+			if s.config.EnableVerbosedTransactionSyncLogging {
```
algonautshant, line 253: It will be nice to add another method to txnsync.Logger e.g. verboseInfof, and call it here.
verboseInfof can check EnableVerbosedTransactionSyncLogging before logging.

test/testdata/configs/config-v17.json
```diff
@@ -44,6 +44,7 @@
     "EnableProfiler": false,
     "EnableRequestLogger": false,
     "EnableTopAccountsReporting": false,
+    "EnableVerbosedTransactionSyncLogging": false,
```
algonautshant, line 47: This is very useful. 
We need something similar for the agreement logs.

algonautshant (approved):
Looks great. We need to apply this technique elsewhere too, e.g. agreement. 
I have one suggestion, otherwise good to go.

PR #2566, opened 2021-07-16
txnsync: update default bloom filter to use xor32

tsachiherman:
## Summary

Change the default bloom filter to use xor32.

## Test Plan

Use the existing tests & update them.


brianolson (approved):
Good comments.
I'm not sure xor32 is the right tradeoff long term, but this is fine.

algonautshant (approved):
Looks good. 
Thanks for the extra comments.

PR #2572, opened 2021-07-16
txnsync: Add fields to txnsync msgs for proposal sending

nicholasguoalgorand:
## Summary

These changes add fields to the txnsync msgs in order to support sending proposals using the txnsync. For now, the txnsync message handler ignores all messages of this type.

### DIFF
--- a/txnsync/exchange.go
+++ b/txnsync/exchange.go
@@ -17,6 +17,7 @@
 package txnsync
 
 import (
+	"github.com/algorand/go-algorand/crypto"
 	"github.com/algorand/go-algorand/data/basics"
 )
 
@@ -34,6 +35,7 @@ type transactionBlockMessage struct {
 	UpdatedRequestParams requestParams           `codec:"p"`
 	TransactionGroups    packedTransactionGroups `codec:"g"`
 	MsgSync              timingParams            `codec:"t"`
+	ProposalData         relayedProposal         `codec:"pd"`
 }
 
 type encodedBloomFilter struct {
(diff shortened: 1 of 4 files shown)


txnsync/exchange.go
```diff
@@ -34,6 +35,7 @@ type transactionBlockMessage struct {
 	UpdatedRequestParams requestParams           `codec:"p"`
 	TransactionGroups    packedTransactionGroups `codec:"g"`
 	MsgSync              timingParams            `codec:"t"`
+	ProposalData         relayedProposal         `codec:"pd"`
```
tsachiherman, line 38: `ProposalData` -> `RelayedProposal`

txnsync/exchange.go
```diff
@@ -73,3 +75,18 @@ type timingParams struct {
...
+	proposalByte
+	proposalFilter
+)
+
+type relayedProposal struct {
+	_struct struct{} `codec:",omitempty,omitemptyarray"`
+
+	ProposalBytes   []byte        `codec:"pb,allocbound=maxEncodedTransactionGroupBytes"`
+	ProposalFilter  crypto.Digest `codec:"pf"`
+	ProposalMsgType byte          `codec:"mt"`
```
tsachiherman, line 91: add a comment describing when this would be used.
Generally speaking, when having the field of a data structure, we should avoid prefixing the data structure name from the fields. i.e. avoid `ProposalXXX`, since it's kind of weird to have this prefix on all the fields.

my suggestion is to use -
RawBytes, ExcludeProposal and Content

beside these, try to use a single character codec name as well as a different const for the byte array size limit.
we should probably pick a limit of 10K transactions for now.. so that's -
10K * 32 + sizeof(block header) ~= 350K

txnsync/exchange.go
```diff
@@ -73,3 +75,18 @@ type timingParams struct {
 	AcceptedMsgSeq      []uint64 `codec:"a,allocbound=maxAcceptedMsgSeq"`
 	NextMsgMinDelay     uint64   `codec:"m"`
 }
+
+const (
+	noProposal byte = iota
+	transactionsForProposal
+	proposalByte
+	proposalFilter
```
tsachiherman, line 83: I don't think we'll need proposalByte/proposalFilter since these are implied from the data itself.

nicholasguoalgorand: you mean we need exactly one of these right?

tsachiherman: yes, we need the `transactionsForProposal` for now.

nicholasguoalgorand: oh sorry, misunderstood

tsachiherman (approved):
looks good to me.

PR #2573, opened 2021-07-16
Use byte format for asset b64 fields

winder:
## Summary

While adding Indexer support for these, I realized that for a model object we can specify byte formatting. The HTTP responses won't change with this change, so this isn't strictly required for the release.

## Test Plan

e2e sub test


tsachiherman (approved):
looks good to me, although it's a bit confusing since I can't really see the base64 encoder in action.

winder:
There are calls like `ctx.JSON(http.StatusOK, response)` which do encoding with the echo framework for most endpoints

PR #2574, opened 2021-07-16
Enable ineffassign linter and resolve its errors

cce:
## Summary

The [ineffassign](https://github.com/gordonklaus/ineffassign) linter is in the default set of linters enabled by golangci-lint but wasn't added in #2523 because it was reporting issues.

It's also one of the least objectionable linters; ineffectual assignments are rarely intentional and can point out bugs related to assumptions that a variable assignment had some effect.

There were various fixes here but I'm not sure if all the changes to remove or fix ineffectual assignments were the right ones.

## Test Plan

Existing tests should ideally pass.

### DIFF
--- a/agreement/proposal.go
+++ b/agreement/proposal.go
@@ -201,6 +201,7 @@ func verifyNewSeed(p unauthenticatedProposal, ledger LedgerReader) error {
 		if !ok {
 			return fmt.Errorf("payload seed proof malformed (%v, %v)", prevSeed, p.SeedProof)
 		}
+		_ = vrfOut // ignoring output of Verify
 		// TODO remove the following Hash() call,
 		// redundant with the Verify() call above.
 		vrfOut, ok = p.SeedProof.Hash()
--- a/agreement/proposal.go
+++ b/agreement/proposal.go
@@ -197,14 +197,13 @@ func verifyNewSeed(p unauthenticatedProposal, ledger LedgerReader) error {
 
 	if value.OriginalPeriod == 0 {
 		verifier := proposerRecord.SelectionID
-		ok, vrfOut := verifier.Verify(p.SeedProof, prevSeed)
+		ok, _ := verifier.Verify(p.SeedProof, prevSeed) // ignoring VrfOutput returned by Verify
 		if !ok {
 			return fmt.Errorf("payload seed proof malformed (%v, %v)", prevSeed, p.SeedProof)
 		}
-		_ = vrfOut // ignoring output of Verify
 		// TODO remove the following Hash() call,
 		// redundant with the Verify() call above.
-		vrfOut, ok = p.SeedProof.Hash()
+		vrfOut, ok := p.SeedProof.Hash()
 		if !ok {
 			// If proof2hash fails on a proof we produced with VRF Prove, this indicates our VRF code has a dangerous bug.
 			// Panicking is the only safe thing to do.
--- a/catchup/service.go
+++ b/catchup/service.go
@@ -307,7 +307,8 @@ func (s *Service) fetchAndWrite(r basics.Round, prevFetchCompleteChan chan bool,
 				}
 
 				if s.cfg.CatchupVerifyTransactionSignatures() || s.cfg.CatchupVerifyApplyData() {
-					vb, err := s.ledger.Validate(s.ctx, *block, s.blockValidationPool)
+					var vb *ledger.ValidatedBlock
+					vb, err = s.ledger.Validate(s.ctx, *block, s.blockValidationPool)
 					if err != nil {
 						if s.ctx.Err() != nil {
 							// if the context expired, just exit.
--- a/cmd/tealdbg/localLedger.go
+++ b/cmd/tealdbg/localLedger.go
@@ -188,12 +188,12 @@ func getAppCreatorFromIndexer(indexerURL string, indexerToken string, app basics
 	client := &http.Client{}
 	request, err := http.NewRequest("GET", queryString, nil)
 	if err != nil {
-		return basics.Address{}, fmt.Errorf("application request error: %s", err)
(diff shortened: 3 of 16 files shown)


cmd/tealdbg/localLedger.go
```diff
@@ -187,6 +187,9 @@ func getAppCreatorFromIndexer(indexerURL string, indexerToken string, app basics
 	queryString := fmt.Sprintf("%s/v2/applications/%d", indexerURL, app)
 	client := &http.Client{}
 	request, err := http.NewRequest("GET", queryString, nil)
+	if err != nil {
+		return basics.Address{}, fmt.Errorf("application request error: %s", err)
```
tsachiherman, line 191: `%s` -> `%v`, since err is an object.

cmd/tealdbg/localLedger.go
```diff
@@ -215,6 +218,9 @@ func getBalanceFromIndexer(indexerURL string, indexerToken string, account basic
 	queryString := fmt.Sprintf("%s/v2/accounts/%s?round=%d", indexerURL, account, round)
 	client := &http.Client{}
 	request, err := http.NewRequest("GET", queryString, nil)
+	if err != nil {
+		return basics.AccountData{}, fmt.Errorf("account request error: %s", err)
```
tsachiherman, line 222: %s -> %v, since err is an object.

debug/doberman/main.go
```diff
@@ -58,6 +58,9 @@ func main() {
 
 	// write logo
 	tf, err := ioutil.TempFile("", "algorand-logo.png")
+	if err != nil {
+		panic(err)
```
tsachiherman, line 62: that's not so nice.. just print an error and os.Exit(1)

shared/pingpong/pingpong.go
```diff
@@ -153,10 +153,7 @@ func computeAccountMinBalance(client libgoal.Client, cfg PpConfig) (requiredBala
 		fmt.Printf("required min balance for app accounts: %d\n", requiredBalance)
 		return
 	}
-	var fee uint64 = 1000
-	if cfg.MinFee > fee {
-		fee = cfg.MinFee
-	}
+	var fee uint64
```
tsachiherman, line 156: while this code is technically correct, I think that we shouldn't do that as is.
Instead, we need to evaluate the bigger picture and (Maybe) eliminate the `MinFee` from the config file, as it seems redundant.

agreement/proposal.go
```diff
@@ -201,6 +201,7 @@ func verifyNewSeed(p unauthenticatedProposal, ledger LedgerReader) error {
 		if !ok {
 			return fmt.Errorf("payload seed proof malformed (%v, %v)", prevSeed, p.SeedProof)
 		}
+		_ = vrfOut // ignoring output of Verify
```
tsachiherman, line 204: I think that I'd rather seeing 
```golang
ok, _ := verifier.Verify(p.SeedProof, prevSeed)
```
instead if we're going to intentionally ignore the vrf output.

tsachiherman (changes requested):
several small requests, but looks great otherwise!

debug/doberman/main.go
cce: I was following the convention of the other `if err != nil` handling in this program.

cmd/tealdbg/localLedger.go
cce: I copied these line verbatim from a few lines down, but I think the modern thing to do is [`%w` here](https://blog.golang.org/go1.13-errors)

tsachiherman: correct. Using `%w` for errors is better.

shared/pingpong/pingpong.go
cce: Yes I left a comment here https://github.com/algorand/go-algorand/pull/1715/files#r671536585 about whether this was intentional or not ...

catchup/service.go
```diff
@@ -307,7 +307,8 @@ func (s *Service) fetchAndWrite(r basics.Round, prevFetchCompleteChan chan bool,
 				}
 
 				if s.cfg.CatchupVerifyTransactionSignatures() || s.cfg.CatchupVerifyApplyData() {
-					vb, err := s.ledger.Validate(s.ctx, *block, s.blockValidationPool)
+					var vb *ledger.ValidatedBlock
+					vb, err = s.ledger.Validate(s.ctx, *block, s.blockValidationPool)
```
cce, line 311: @tsachiherman this is the most significant issue I found — before this change, the value of `err = s.ledger.AddValidatedBlock(*vb, *cert)` on line 319/320 is ineffectual, and has no impact later on line 324/325 where it checks `if err != nil`

shared/pingpong/pingpong.go
cce: pinging @algorandskiy for his opinion since these lines here were made ineffectual by #1715

algorandskiy: `MinFee` is used in `fee()` func only for fee randomization. I think it is OK to remove it completely as a separate PR. For this one removing the branch is perfectly fine.

A note about using `SuggestedFee` here: pingpong funds accounts between iterations to ensure they have enough algos to interact, and on a high load with app/assets they some of them might not have enough algos to submit new transactions because the pool conjunction.

algorandskiy (approved):

shared/pingpong/pingpong.go
cce: oh OK, I see so MinFee is still used in pingpong.go:WorkerState.fee(). I just wanted to make sure you realized these lines 156-159 being removed above were ineffectual (1000 is unused, and fee = cfg.MinFee is always overwritten).

tsachiherman (approved):

PR #2575, opened 2021-07-17
Addressing panic in assembler

barnjamin:
## Summary
I'm probably doing something wrong in my code somewhere* but started getting no response from remote compile.  Tracked it down to one of the bytec functions that does check the slice it accesses is long enough but doesn't bail, so it panics.

Took the liberty of adding returns to the other functions in the area but I don't feel confident enough to add them elsewhere. 

*had a pushbytes with no arg

## Test Plan

I didn't

### DIFF
--- a/data/transactions/logic/assembler.go
+++ b/data/transactions/logic/assembler.go
@@ -445,11 +445,11 @@ func assembleIntC(ops *OpStream, spec *OpSpec, args []string) error {
 }
 func assembleByteC(ops *OpStream, spec *OpSpec, args []string) error {
 	if len(args) != 1 {
-		ops.error("bytec operation needs one argument")
+		return ops.error("bytec operation needs one argument")
 	}
 	constIndex, err := strconv.ParseUint(args[0], 0, 64)
 	if err != nil {
-		ops.error(err)
+		return ops.error(err)
 	}
 	ops.Bytec(uint(constIndex))
 	return nil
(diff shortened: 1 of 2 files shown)


jannotti (approved):
Looks great. We try to continue assembly in the face of errors (to show more problems), but clearly we can't continue this instruction without an argument. The return after the ParseUint _might_ not be necessary, but there's no further error reporting in this instruction possible, so better to follow the usual Go conventions and return immediately.  Thanks.

We appreciate the PR!

jannotti (changes requested):
Actually, could you insert this test in logic/assembler_test.go?
```
func TestConstantArgs(t *testing.T) {
	t.Parallel()
	for v := uint64(1); v <= AssemblerMaxVersion; v++ {
		testProg(t, "int", v, expect{1, "int needs one argument"})
		testProg(t, "intc", v, expect{1, "intc operation needs one argument"})
		testProg(t, "byte", v, expect{1, "byte operation needs byte literal argument"})
		testProg(t, "bytec", v, expect{1, "bytec operation needs one argument"})
		testProg(t, "addr", v, expect{1, "addr operation needs one argument"})
	}
	for v := uint64(3); v <= AssemblerMaxVersion; v++ {
		testProg(t, "pushint", v, expect{1, "pushint needs one argument"})
		testProg(t, "pushbytes", v, expect{1, "pushbytes needs one argument"})
	}
}
```

jannotti (approved):

PR #2578, opened 2021-07-18
Add batch verification

algoidan:


crypto/multisig.go
```diff
@@ -217,11 +217,17 @@ func MultisigAssemble(unisig []MultisigSig) (msig MultisigSig, err error) {
...
+}
 
+// MultisigBatchVerify verifies an assembled MultisigSig.
+// if the batchverifier is not nil, this function DOES NOT validate the digital signature.
+func MultisigBatchVerify(msg Hashable, addr Digest, sig MultisigSig, batchVerifier *BatchVerifier) (verified bool, err error) {
 	verified = false
 	// short circuit: if msig doesn't have subsigs or if Subsigs are empty
 	// then terminate (the upper layer should now verify the unisig)
 	if (len(sig.Subsigs) == 0 || sig.Subsigs[0] == MultisigSubsig{}) {
+		err = errors.New(errorinvalidnumberofsignature)
```
tsachiherman, line 230: it would be nice to use a preallocated error object here.

crypto/multisig.go
```diff
@@ -263,6 +269,11 @@ func MultisigVerify(msg Hashable, addr Digest, sig MultisigSig) (verified bool,
 	var verifiedCount int
 	for _, subsigi := range sig.Subsigs {
 		if (subsigi.Sig != Signature{}) {
+			if batchVerifier != nil {
```
tsachiherman, line 272: maybe move this if statement outside the `for..range` loop, so that it would be executed exactly once ?

tsachiherman (commented):
looks good. two minor comments, but neither of them would be a blocker.

crypto/cryptoerror.go
```diff
@@ -18,13 +18,16 @@ package crypto
...
-const errorinvalidversion = "Invalid version"
-const errorinvalidaddress = "Invalid address"
-const errorinvalidthreshold = "Invalid threshold"
-const errorinvalidnumberofsignature = "Invalid number of signatures"
-const errorkeynotexist = "Key does not exist"
-const errorsubsigverification = "Verification failure: subsignature"
-const errorkeysnotmatch = "Public key lists do not match"
-const errorinvalidduplicates = "Invalid duplicates"
+var (
+	errorinvalidversion           = errors.New("Invalid version")
```
tsachiherman, line 22: errors in go are usually named errXXXX or ErrXXX, depending if they are exported or not. But in either case, the variable naming convention is either camel case or pascal case.

data/transactions/verify/txn.go
```diff
@@ -100,6 +100,26 @@ func (g *GroupContext) Equal(other *GroupContext) bool {
...
+	}
+	if err := batchVerifier.Verify(); err != nil {
+		return err
+	}
+	return nil
+}
+
+// TxnBatchVerify verifies a SignedTxn having no obviously inconsistent data.
+// Block-assembly time checks of LogicSig and accounting rules may still block the txn.
+// it is the caller responsibility to call batchVerifier.verifiy()
```
tsachiherman, line 121: `batchVerifier.verifiy()` -> `batchVerifier.verify()`

tsachiherman (changes requested):
Two small comments, but looks good otherwise!

tsachiherman (approved):

Issue #2431, opened 2021-07-06
ExtraProgramPages not usable after Creation


jeapostrophe:
You can set `ExtraProgramPages` to something greater than 0 at creation time, but you can't update an application and use extra space you allocated

If you look at this code:

https://github.com/algorand/go-algorand/blob/c7b017a7d38317aa7484fd0c95a417418a4386e4/data/transactions/transaction.go#L353-L408

You can see that
- L348 - is true when you are updating
- L353 - guarantees that `tx.ExtraProgramPages` is 0
- L399 - is 0 because of above
- L406 - is going to be true (i.e. error) because of that

I believe that this is not a problem when you are not updating (i.e. when you are calling) because `tx.ApprovalProgram` and `tx.ClearProgram` are expected to be 0 since they aren't being updated.

I believe that the fix is to look up the actual application `ExtraProgramPages` when you aren't updating and use the stored value

jeapostrophe:
I'll work on a patch and get back to you about my proposed fixed, but I figured learning about the problem fast is good :)

jeapostrophe:
@shiqizng

jannotti:
> I believe that the fix is to look up the actual application `ExtraProgramPages` when you aren't updating and use the stored value

That sounds right to me.

jeapostrophe:
I spent some time working on a patch for this and I now think that `WellFormed` should not be checking the program lengths on update like this. Basically `WellFormed` says `// WellFormed checks that the transaction looks reasonable on its own (but not necessarily valid against the actual ledger). It does not check signatures.` The only way to do this check is to look at the current ledger, which is not available to this function. So instead, it should make sure the program is not more than is possible and then `updateApplication` in `application.go` should check to make sure the new programs are valid

jeapostrophe:
This is the patch I'm using and it is working well --- https://gist.github.com/94149c776903646a9d8e0120811b5e02

shiqizng:
the fix looks good to me.

PR #2580, opened 2021-07-19
Fix- extraProgramPages

shiqizng:
This PR contains patch proposed by @jeapostrophe that fixes issue [2431](https://github.com/algorand/go-algorand/issues/2431).


config/consensus.go
```diff
@@ -381,6 +381,8 @@ type ConsensusParams struct {
 	// 5. checking that in the case of going online the VoteFirst is less or equal to the LastValid+1.
 	// 6. checking that in the case of going online the VoteFirst is less or equal to the next network round.
 	EnableKeyregCoherencyCheck bool
+
+	EnableExPagesOnAppUpdate bool
```
jannotti, line 385: Since "Ex" reads to me like "no longer" (Like Ex Husband, or Ex Googler), I'd avoid this abbreviation, and just use Extra.

ledger/apply/application_test.go
```diff
@@ -962,6 +963,58 @@ func TestAppCallApplyUpdate(t *testing.T) {
...
+	b.appCreators = map[basics.AppIndex]basics.Address{appIdx: creator}
+
+	// check program len check doesn't happen in current consensus proto version
+	b.SetProto(protocol.ConsensusV28)
+	proto = b.ConsensusParams()
+	ep.Proto = &proto
+
+	b.pass = true
+	err = ApplicationCall(ac, h, &b, ad, &ep, txnCounter)
+	a.NoError(err)
```
jannotti, line 1008: Why does this work?  You are updating with a 6k program, under the old proto.  That should fail.  Am I misreading?

ledger/apply/application_test.go
```diff
@@ -962,6 +963,58 @@ func TestAppCallApplyUpdate(t *testing.T) {
...
+
+	//check program len check happens in future consensus proto version
+	b.SetProto(protocol.ConsensusFuture)
+	proto = b.ConsensusParams()
+	ep.Proto = &proto
+
+	b.pass = true
+	err = ApplicationCall(ac, h, &b, ad, &ep, txnCounter)
+	a.Contains(err.Error(), "updateApplication app programs too long")
```
jannotti, line 1017: We want to confirm that it IS possible to have a 3k approval program here.  Currently, you can not update to 3k, even though your extra pages = 1 when you created it (which is what I think this is testing).

shiqizng: this test is for validating this added check lines are actually invoked, https://github.com/algorand/go-algorand/blob/af3858cf92829aa8f6eb2ffcedd4a03c0b820c88/ledger/apply/application.go#L203-L209. 

But, yes, I need to add another test to validate that it is possible to have a 3K approval program.

ledger/apply/application_test.go
shiqizng: this works because b is a mocked Balances obj and it actually invokes this mocked StatefulEval https://github.com/algorand/go-algorand/blob/af3858cf92829aa8f6eb2ffcedd4a03c0b820c88/ledger/apply/application_test.go#L189. 

since I wanted to test updateApplication() only, b.pass is set to true at line 1006.

jannotti: ApplicationCall should still have failed, I think, even before it got to StatefulEval().  Your extra program pages is 1 in params.  But you are calling this with UpdateApplicationOC and a 6k approval program.  Why does updateApplication allow such a big approval program?

I don't think we can trust these kinds of tests for actual transaction execution very much.  We need to put them in e2e subs tests against a running network or we'll never have the level of confidence we need.

jannotti: I see the bug.  In the new code, you don't check the length at all in old versions.  But you must check the length - you must maintain the same behavior in old versions.  So the behaviour should be:

 Old version: check against a single program page of space
 New version: check against the extra pages specified in params

shiqizng: I'll add an e2e test. 
do I need to add a check for the old version in updateApplication() though? 
doesn't these checks, https://github.com/algorand/go-algorand/blob/e3e419754c198b5232927c735763b36390e47d68/data/transactions/transaction.go#L405-L414 enforce that the old version would maintain the same behavior?

jannotti: I think you are right.  I find it a bit confusing because in old code, a too big program is rejected in WellFormed, while in the new code, it's rejected in updateApplication.  When just reading updateApplication, it appears we have no check at all unless `proto.EnableExtraPagesOnAppUpdate` is true.

Could you add a comment in updateApplication explaining that WellFormed rejects all multipage program updates when `proto.EnableExtraPagesOnAppUpdate` is false? Or else simply reject them here too, even though it's redundant.

ledger/apply/application_test.go
```diff
@@ -962,6 +963,67 @@ func TestAppCallApplyUpdate(t *testing.T) {
 	a.Equal([]byte{2}, br.AppParams[appIdx].ApprovalProgram)
 	a.Equal([]byte{2}, br.AppParams[appIdx].ClearStateProgram)
 	a.Equal(basics.EvalDelta{}, ad.EvalDelta)
+
+	// check app program len
+	appr := make([]byte, 6050, 6050)
+
+	for i := range appr {
+		appr[i] = 2
+	}
```
algorandskiy, line 972: nit:
```suggestion
	appr := bytes.Repeat([]byte{2}, 6050)
```

But! it not even needed since `b.pass = true` assumes `logic.EvalStateful` not even called.
so this works and setting zero byte to version 4 not needed as well:
```suggestion
	appr := make([]byte, 6050)
```

ledger/apply/application.go
```diff
@@ -199,6 +199,16 @@ func updateApplication(ac *transactions.ApplicationCallTxnFields, balances Balan
 	// Fill in the new programs
 	record.AppParams = cloneAppParams(record.AppParams)
 	params := record.AppParams[appIdx]
+	proto := balances.ConsensusParams()
+	// when proto.EnableExtraPageOnAppUpdate is false, WellFormed rejects all updates with a multiple-page program
+	if proto.EnableExtraPagesOnAppUpdate {
+		allowed := int(1+params.ExtraProgramPages) * proto.MaxAppProgramLen
```
algorandskiy, line 205: please check if this formula `int(1+params.ExtraProgramPages) * proto.MaxAppProgramLen` is used somewhere else in **ledger/apply** package, and if so, make a function.

data/transactions/transaction_test.go
```diff
@@ -447,6 +447,37 @@ func TestWellFormedErrors(t *testing.T) {
...
+				ApplicationCallTxnFields: ApplicationCallTxnFields{
+					ApplicationID:     1,
+					ApprovalProgram:   []byte(strings.Repeat("X", 1025)),
+					ClearStateProgram: []byte(strings.Repeat("X", 1025)),
+					ExtraProgramPages: 0,
+					OnCompletion:      UpdateApplicationOC,
+				},
+			},
+			spec:          specialAddr,
+			proto:         curProto,
```
algorandskiy, line 463: this will fail after protocol switch, please use a concrete version

data/transactions/transaction_test.go
```diff
@@ -447,6 +447,37 @@ func TestWellFormedErrors(t *testing.T) {
...
+			expectedError: fmt.Errorf("app programs too long. max total len %d bytes", curProto.MaxAppProgramLen),
+		},
+		{
+			tx: Transaction{
+				Type:   protocol.ApplicationCallTx,
+				Header: okHeader,
+				ApplicationCallTxnFields: ApplicationCallTxnFields{
+					ApplicationID:     1,
+					ApprovalProgram:   []byte(strings.Repeat("X", 1025)),
+					ClearStateProgram: []byte(strings.Repeat("X", 1025)),
```
algorandskiy, line 473: shouldn't it fail because the combined length is 2050 that is more than the limit of 2048?

data/transactions/transaction_test.go
```diff
@@ -447,6 +447,37 @@ func TestWellFormedErrors(t *testing.T) {
...
+					ExtraProgramPages: 0,
+					OnCompletion:      UpdateApplicationOC,
+				},
+			},
+			spec:  specialAddr,
+			proto: futureProto,
+		},
```
algorandskiy, line 480: please add a test checking `ExtraProgramPages ` update with `OnCompletion : UpdateApplicationOC`

data/transactions/transaction.go
```diff
@@ -344,6 +344,7 @@ func (tx Transaction) WellFormed(spec SpecialAddresses, proto config.ConsensusPa
 			}
 		}
 
+		effectiveEPP := tx.ExtraProgramPages
```
algorandskiy, line 347: call it `maxExtraProgramPages` ?

ledger/apply/application_test.go
```diff
@@ -962,6 +963,67 @@ func TestAppCallApplyUpdate(t *testing.T) {
 	a.Equal([]byte{2}, br.AppParams[appIdx].ApprovalProgram)
 	a.Equal([]byte{2}, br.AppParams[appIdx].ClearStateProgram)
 	a.Equal(basics.EvalDelta{}, ad.EvalDelta)
+
```
algorandskiy, line 966: add a test case checking a current (base) behavior: update failure on `ConsensusV28` when extra pages are used in app params.

algorandskiy (changes requested):

data/transactions/transaction_test.go
shiqizng: this is because it's an app update. the check for program lens for an app update has moved to https://github.com/shiqizng/go-algorand/blob/fix/exPages/ledger/apply/application.go#L204-L210

ledger/apply/application_test.go
shiqizng: It's needed to validate this new check in updateApplication, https://github.com/shiqizng/go-algorand/blob/fix/exPages/ledger/apply/application.go#L208.

data/transactions/transaction_test.go
shiqizng: this is because it's an app update. the check for program lens for an app update has moved to https://github.com/shiqizng/go-algorand/blob/fix/exPages/ledger/apply/application.go#L204-L210

ledger/apply/application.go
shiqizng: no other occurrence is found.

data/transactions/transaction_test.go
shiqizng: ExtraProgramPages can only be set during application creation. I think that's covered by this case? https://github.com/shiqizng/go-algorand/blob/fix/exPages/data/transactions/transaction_test.go#L354-L368

data/transactions/transaction.go
shiqizng: I think effectiveEPP is better because it doesn't always equal to maxExtraProgramPages. when ApplicationID=0, effectiveEPP = tx.ExtraProgramPages

ledger/apply/application_test.go
shiqizng: I don't think this behavior has changed. we are still only allowing ExtraProgramPages to be set at app creation.

algorandskiy: the problem is this behavior was not tested, and now we are making a protocol upgrade change. So a base behavior needs to be checked to ensure the difference.

ledger/apply/application_test.go
algorandskiy: the line you pointed is a size check and not eval call (it is mocked in this unit test)

data/transactions/transaction_test.go
algorandskiy, line 463: this test will fail after the protocol upgrade, use v28 instead?

data/transactions/transaction_test.go
algorandskiy: fair

ledger/apply/application_test.go
shiqizng: I get static check error  from checkProgram if I set appr := make([]byte, 6050). `check failed on ApprovalProgram: pc=702 static cost budget of 700 exceeded`. https://github.com/shiqizng/go-algorand/blob/fix/exPages/ledger/apply/application.go#L364

ledger/apply/application_test.go
shiqizng: ok. Added this case to transaction_test

ledger/apply/application_test.go
algorandskiy: then `appr[0] = 4` is needed indeed

algorandskiy (approved):
LGTM

PR #2582, opened 2021-07-19
txnsync: avoid pseudonode execution goroutine hang

tsachiherman:
## Summary

While running a longevity tests on the transaction sync feature branch, the following issue showed up on some of the participating nodes : 
```json
{"file":"actions.go","function":"github.com/algorand/go-algorand/agreement.pseudonodeAction.do","level":"error","line":386,"msg":"pseudonode.MakeVotes call failed(attest) pseudonode input channel is full","time":"2021-07-19T13:28:35.078410Z"}
```

Digging further, it seems that both `pseudonodeVotesTask.execute` as well as `pseudonodeProposalsTask.execute` could lead to this situation ( same issue ). There are two distinct issues in the current implementation that have been addressed in this PR:
1. The call to `verifyVote` might silently fail. In that case, the caller ( i.e. the `execute` method ) would still expect to find the result in the passed-in output channel. When this does not happen, the method could be blocked indefinitely.
2. When the `execute` method tries to send the output result back to the output channel, it might fail doing so due to an issue on the "other" side. When that happens, we have had no error indication. The changes in the codebase would now allow a timeout for these messages to ensure it won't be "stuck" forever.

## Test Plan

Use existing unit tests. Run a longevity test.

### DIFF
--- a/agreement/asyncVoteVerifier.go
+++ b/agreement/asyncVoteVerifier.go
@@ -131,7 +131,7 @@ func (avv *AsyncVoteVerifier) executeEqVoteVerification(task interface{}) interf
 	}
 }
 
-func (avv *AsyncVoteVerifier) verifyVote(verctx context.Context, l LedgerReader, uv unauthenticatedVote, index int, message message, out chan<- asyncVerifyVoteResponse) {
+func (avv *AsyncVoteVerifier) verifyVote(verctx context.Context, l LedgerReader, uv unauthenticatedVote, index int, message message, out chan<- asyncVerifyVoteResponse) error {
 	select {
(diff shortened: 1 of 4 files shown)


nicholasguoalgorand:
Is there a reason why we are not trying to merge this to master but only to the txnsync branch?

tsachiherman:
> Is there a reason why we are not trying to merge this to master but only to the txnsync branch?

Yes - I want to verify the fix on this branch ( where I was able to reproduce it ), and then port it to master.

It will take few days of runtime to reproduce it, so I figured that I'll start as soon as I can on this branch.

nicholasguoalgorand (approved):

cce:
follow-up: this was split off and released in #2741

PR #2588, opened 2021-07-20
Make `AccountDeltas.accts` public

tolikzinovyev:
## Summary

Indexer needs access to all the accounts from `AccountDeltas`. Rather then iterating over all addresses and doing a hash table lookup for each, just make the data public.

### DIFF
--- a/ledger/ledgercore/statedelta.go
+++ b/ledger/ledgercore/statedelta.go
@@ -85,7 +85,7 @@ type StateDelta struct {
 // AccountDeltas stores ordered accounts and allows fast lookup by address
 type AccountDeltas struct {
 	// actual data
-	accts []basics.BalanceRecord
+	Accts []basics.BalanceRecord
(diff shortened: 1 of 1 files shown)


tsachiherman (changes requested):
Instead of exposing the `accts` directly, implement an accessor method. It would reduce the "complexity" of the needed dependencies. i.e. add a `ModifiedBalanceRecords() []basics.BalanceRecord`

tolikzinovyev:
What dependencies are you referring to? I would normally write an accessor method in other languages, but go doesn't have constant pointers/references. The user of `ModifiedBalanceRecords()` might think that he gets a copy of the array, when in fact modifying it modifies the `AccountDeltas` as well. To make it more evident, I just made `accts` public.

tolikzinovyev:
Actually, just discovered there is `AccountDeltas.GetByIdx()` function. This change is not needed. :)

PR #2590, opened 2021-07-20
txsync: incoming unit tests

algonautshant:
## Summary

This adds unit tests for incoming.go

## Test Plan

This adds unit tests.

### DIFF
--- a/txnsync/incoming_test.go
+++ b/txnsync/incoming_test.go
@@ -24,10 +24,11 @@ import (
 
 	"github.com/stretchr/testify/require"
 
+	"github.com/algorand/msgp/msgp"	
+	
 	"github.com/algorand/go-algorand/config"
 	"github.com/algorand/go-algorand/data/transactions"
 	"github.com/algorand/go-algorand/logging"
-	"github.com/algorand/msgp/msgp"
 )
 
 type incomingLogger struct {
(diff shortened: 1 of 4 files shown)


txnsync/incoming_test.go
```diff
@@ -0,0 +1,270 @@
...
+	"fmt"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/algorand/go-algorand/config"
+	"github.com/algorand/go-algorand/data/transactions"
+	"github.com/algorand/go-algorand/logging"
+	"github.com/algorand/msgp/msgp"
```
tsachiherman, line 30: nit: move this to it's own group, before the `go-algorand` packages.

txnsync/incoming_test.go
```diff
@@ -0,0 +1,270 @@
...
+		log: wrapLogger(&incLogger, &cfg)}
+
+	// expect UnmarshalMsg error
+	messageBytes[0] = 0
+	err := s.asyncIncomingMessageHandler(nil, nil, messageBytes, sequenceNumber)
+	msgpe := msgp.TypeError{}
+	require.True(t, errors.As(err, &msgpe))
+
+	// expect wrong version error
+	message = transactionBlockMessage{Version: 3}
```
tsachiherman, line 64: just for future compatibility, it would be better to use a version that we're not going to use.

tsachiherman (commented):
I think it looks great, I added two small comments.
I looked at the coverage, and it looks like the 
```golang
// if the queue is empty ( not expected, since we peek'ed into it before ), then we can't do much here.
```
statement is not being covered; that's expected, as there should be no logical way to get there. however, I think that we need to re-write the peekSequence()/pop() pair so it would avoid getting to this situation on the first place.

AlgoStephenAkiki (approved):

tsachiherman (approved):

PR #2601, opened 2021-07-20
bug: disable networking on devmode

tsachiherman:
## Summary

The current code would not pass the `DisableNetworking` flag to the network package. This could be worked around by adding the `DisableNetworking` to the `config.json` file, or with this fix.

## Test Plan

Tested manually.

### DIFF
--- a/node/node.go
+++ b/node/node.go
@@ -161,7 +161,6 @@ func MakeFull(log logging.Logger, rootDir string, cfg config.Local, phonebookAdd
 
 	node := new(AlgorandFullNode)
 	node.rootDir = rootDir
-	node.config = cfg
 	node.log = log.With("name", cfg.NetAddress)
 	node.genesisID = genesis.ID()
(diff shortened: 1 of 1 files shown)


node/node.go
```diff
@@ -170,6 +169,7 @@ func MakeFull(log logging.Logger, rootDir string, cfg config.Local, phonebookAdd
 	if node.devMode {
 		cfg.DisableNetworking = true
 	}
+	node.config = cfg
```
winder, line 172:
```suggestion
		node.config.DisableNetworking = true
	}
```

tsachiherman: I was hoping to avoid setting this field twice - instead, I think that we should try and have a single copy of the `config.Local`. Until we do this, I think that ensuring that we "just" copy it here would make the next change *slightly* easier.

winder: Oops, I didn't realize cfg was used below as well.

jdtzmn (approved):

PR #2602, opened 2021-07-20
Install with dnf if present, otherwise fallback to yum

onetechnical:
## Summary

In rpm package testing, dnf is not always installed. Check to see if it's available; otherwise fall back to yum.

## Test Plan

Tried it on current rel/beta branch and it worked.


tsachiherman (approved):
seems reasonable.

algobarb (approved):
Makes sense

PR #2605, opened 2021-07-22
Remove references to centos8 rpm packaging

egieseke:
## Summary

Remove old references to centos8 rpm packaging.  There is now a single package for both centos 7 & 8.

## Test Plan

make clean, make install, make test.  Validate PR testing completes successfully.


algobarb (approved):
Looks like it passed all the tests and removed all the lines we had an issue with when running the beta rpm deploy. Looks good to me.

onetechnical (approved):

PR #2607, opened 2021-07-22
Allow txnsync to send proposals

nicholasguoalgorand:

### DIFF
--- a/txnsync/peer.go
+++ b/txnsync/peer.go
@@ -53,6 +53,7 @@ const defaultDataExchangeRate = minDataExchangeRateThreshold
 const defaultRelayToRelayDataExchangeRate = 10 * 1024 * 1024 / 8 // 10Mbps
 const bloomFilterRetryCount = 3                                  // number of bloom filters we would try against each transaction group before skipping it.
 const maxTransactionGroupTrackers = 15                           // number of different bloom filter parameters we store before rolling over
+const maxProposalFilterCacheSize = 20                            // number of entries we would hold in the proposalFilterCache
 
 const (
 	// peerStateStartup is before the timeout for the sending the first message to the peer has reached.
@@ -169,6 +170,9 @@ type Peer struct {
 	// used by the selectPendingTransactions method, the lastSelectedTransactionsCount contains the number of entries selected on the previous iteration.
 	// this value is used to optimize the memory preallocation for the selection IDs array.
 	lastSelectedTransactionsCount int
+
+	// proposalFilterCache keeps track of the most recent proposal bytes that the peer does not want to receive
+	proposalFilterCache proposalFilterCache
 }
 
 // requestParamsGroupCounterState stores the latest group counters for a given set of request params.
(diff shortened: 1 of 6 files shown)


txnsync/emulatorNode_test.go
```diff
@@ -75,6 +77,226 @@ type emulatedNode struct {
...
+
+	ProposalBytes []byte              `codec:"b,allocbound=maxNumProposalBytes"`
+	TxGroupIds    []transactions.Txid `codec:"h,allocbound=maxNumTxGroupHashesBytes"` // TODO: make this []byte
+}
+
+const maxNumProposalBytes = 30000       // sizeof(block header)
+const maxNumTxGroupHashesBytes = 320000 // 10K * 32
+
+// MarshalMsg implements msgp.Marshaler
+func (z *proposalData) MarshalMsg(b []byte) (o []byte) {
```
nicholasguoalgorand, line 91: Is there a better way of doing this other than copying the data structure and its encoding and decoding methods

tsachiherman: yes - create a child package inside txnsync called "testing". Create the file there and have it msgp'ed. then, import this package only from the test.

txnsync/peer.go
```diff
@@ -378,6 +400,7 @@ func (p *Peer) selectPendingTransactions(pendingTransactions []pooldata.SignedTx
 		if p.recentIncomingBloomFilters[filterIdx].filter == nil {
 			continue
 		}
+		// TODO: should this use all bloom filters for proposal sending?
```
nicholasguoalgorand, line 403: should this use all bloom filters for proposal sending?

tsachiherman: The `p.recentIncomingBloomFilters` would filter out quite many of these.. but the bloom filters would help with the rest. A good starting point would be to try and use all the bloom filters. I don't know if this could be sustained, and/or what the real benefits of that. ( we should be measuring that to figure out if this is something we should use or not )

agreement/actions.go
```diff
@@ -149,12 +150,23 @@ func (a networkAction) do(ctx context.Context, s *Service) {
...
+			payload := transmittedPayload{
+				unauthenticatedProposal: msg.Proposal,
+				PriorVote:               msg.Vote,
+			}
+			data = protocol.Encode(&payload)
+			logging.Base().Info("sending proposal")
+			s.TxnSync.RelayProposal(data, txns)
 		}
-		data = protocol.Encode(&payload)
+		return
```
tsachiherman, line 169: this ignores disconnect/ignore.

agreement/demux.go
```diff
@@ -323,6 +375,17 @@ func (d *demux) next(s *Service, deadline time.Duration, fastDeadline time.Durat
 		d.UpdateEventsQueue(eventQueueTokenized[protocol.ProposalPayloadTag], 0)
 		d.monitor.inc(demuxCoserviceType)
 		d.monitor.dec(tokenizerCoserviceType)
+	case m, open := <-d.txnsyncProposals:
+		if !open {
+			logging.Base().Warnf("channel was closed")
```
tsachiherman, line 380: `logging.Base().Warnf` -> `s.log.with(logEvent).Warnf`

nicholasguoalgorand: Thanks, just letting you know that I have a bunch of logging messages that are still in the branch mostly just for testing

PR #2608, opened 2021-07-22
Use pre-existing homebrew state on CircleCI VMs

cce:
## Summary

CircleCI VMs come pre-installed with homebrew and some of the taps and packages we need already set up, so skipping running update and tap and install for those pre-existing assets reduces each job by ~6 minutes each (x 4 jobs).

## Test Plan

Should not impact tests, just make them run quicker


onetechnical (approved):
Thanks for this!

PR #2610, opened 2021-07-22
txnsync: avoid reallocating pending transaction ids slice

tsachiherman:
## Summary

The current code was rolling the `ackPendingTxids` slice by using append. This works correctly from functional standpoint, but require an additional allocation + copy. Given that this allocation appeared in the memory profiler, I'm going to defer to doing a plain copy and avoid the allocation completely.

## Test Plan

- [x] Tested using s1 network.
- [x] Tested using emulator.


nicholasguoalgorand (approved):
Looks good.

Issue #1699, opened 2020-11-14
[Investigate] REST API `/v2/blocks/{rnd}` returns invalid json


algorandskiy:
### Subject of the issue

If block contains app state change, it is incorrectly serialized to json - `EvalDelta.LocalDeltas` has non-quoted integer keys.
Seems like `protocol.EncodeJSONStrict` needs to be used.

### Your environment
* Software version: `algod -v`
```
algod -v
8590155987
2.2.90323.master [master] (commit #e8a24ee9+)
go-algorand is licensed with AGPLv3.0
```

### Steps to reproduce
Use newest **pingpong** with updated apps test (see https://github.com/algorand/go-algorand/pull/1700) to create blocks with local deltas. Then iterate over blocks and observe json decoding error.
```
goal network create -t test/testdata/nettemplates/TwoNodes50EachFuture.json -r ~/networks/two-fut -n two-fut

goal network start -r ~/networks/two-fut/
sleep 2
pingpong run -d ~/networks/two-fut/Primary --numapp 100 --appprogops 300 --numappoptin 10 --appproggk 10 --appproglk 10 --tps 1000 --rest 0  --refresh 1 --numaccounts 500 --duration 60 --minaccount 34000000
sleep 30
ROUND=$(curl -k http://$(cat ~/networks/two-fut/Primary/algod.net)/v2/status -H "X-Algo-API-Token: $(cat ~/networks/two-fut/Primary/algod.token)" | jq '.["last-round"]')
for i in $(seq 1 $ROUND); do curl -k http://$(cat ~/networks/two-fut/Primary/algod.net)/v2/blocks/$i -H "X-Algo-API-Token: $(cat ~/networks/two-fut/Primary/algod.token)" | python3 -m json.tool || break; done
goal network stop -r ~/networks/two-fut/
# goal network delete -r ~/networks/two-fut
```

### Expected behaviour
No parsing error

### Actual behaviour
```
Expecting property name enclosed in double quotes: line 11963 column 13 (char 397656)
```

nullun:
I just stumbled upon this error myself. Whilst attempting to decode a response from testnet my code failed due to invalid JSON being returned. Here's a snippet which appears to be the problem "Expected STRING got NUMBER, line 3:  **0: {**

```
...
"dt": {
          "ld": {
            0: {
              "key-one": {
                "at": 2,
                "ui": 1
              },
              "key-two": {
                "at": 2,
                "ui": 1
              }
            }
          }
        },
...
[...]

PR #2611, opened 2021-07-23
Use strict json encoder in REST API v2 endpoints

algorandskiy:
## Summary

As #1699 demonstrates, api v2 uses `JSONHandle` that incorrectly encodes maps with numeric fields. Switching to `JSONStrictHandle` fixes the issue since it was specially created for this purpose.

## Test Plan

A test added

### DIFF
--- a/daemon/algod/api/server/v2/handlers.go
+++ b/daemon/algod/api/server/v2/handlers.go
@@ -393,7 +393,7 @@ func (v2 *Handlers) TealDryrun(ctx echo.Context) error {
 
 	var dr DryrunRequest
 	var gdr generated.DryrunRequest
-	err := decode(protocol.JSONHandle, data, &gdr)
+	err := decode(protocol.JSONStrictHandle, data, &gdr)
 	if err == nil {
 		dr, err = DryrunRequestFromGenerated(&gdr)
 		if err != nil {
--- a/daemon/algod/api/server/v2/test/handlers_test.go
+++ b/daemon/algod/api/server/v2/test/handlers_test.go
@@ -122,7 +122,7 @@ func TestGetBlockJsonEncoding(t *testing.T) {
 	require.NoError(t, err)
 
 	// make an app call txn with eval delta
-	lsig := transactions.LogicSig{Logic: []byte{2, 0x20, 1, 1, 0x22}} // int 1
+	lsig := transactions.LogicSig{Logic: retOneProgram} // int 1
 	program := logic.Program(lsig.Logic)
 	lhash := crypto.HashObj(&program)
 	var sender basics.Address
--- a/daemon/algod/api/server/v2/test/handlers_test.go
+++ b/daemon/algod/api/server/v2/test/handlers_test.go
@@ -18,6 +18,7 @@ package test
 
 import (
 	"bytes"
+	"encoding/json"
 	"errors"
 	"io"
 	"net/http"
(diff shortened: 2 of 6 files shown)


algorandskiy:
This is a single-line fix but with a quite large test.

tsachiherman (commented):
I'm not opposing to use strict json formatting, since the non-strict variation doesn't create a valid json on integer keyed map; but I am a bit concerned that this change would create other issues. Two potential issues -
1. Does this change affect only block requests ? If not, it might be an issue.
2. What about the msgptool -d ? It being used to decode blocks as well. Do we need to change it as well ?

brianolson:
Based on the difference in `protocol/codec.go` where the only difference is adding `MapKeyAsString`, I feel good about using this for API return values, which I'm pretty sure is the only time `algod` uses JSON (no internal storage of JSON like Indexer).

winder (approved):
Thanks for bringing this up again. This would only affect endpoints that have a `format` option. Since msgpack is available to fetch the canonical data, I think we can ignore any slight differences that this change may cause. These are the affected endpoints:
`/v2/transactions/pending`
`/v2/transactions/pending/{txid}`
`/v2/blocks/{round}`
`/v2/accounts/{address}`

The `TealDryrun` command `protocol.JSONHandle` for reading the request, do you think that should use the strict encoder?

tsachiherman:
What would happen if we would follow the same pattern used in `msgpacktool`, and add ":int" to the key's string when encoding it as JSON ? this would allow the reverse operation to perform that symmetrically. ( similar to ":b64" )
i.e. I'm not afraid of "breaking" the JSON formatted output here, since it's already "broken" for that use case.

algorandskiy:
Thank you for reviewing the PR.

> What about the msgptool -d ? It being used to decode blocks as well. Do we need to change it as well?

It has `-strict` option to use `JSONStrictHandle` so it is all set.

> The TealDryrun command protocol.JSONHandle for reading the request, do you think that should use the strict encoder?

It does not matter (added a test to **protocol/codec_test.go**) but I changed it to be in sync with rest of the code.

Updated the PR and fixed TODO in rest-proof-endpoint test.

winder (approved):

algorandskiy:
>What would happen if we would follow the same pattern used in msgpacktool, and add ":int" to the key's string when encoding it as JSON ? this would allow the reverse operation to perform that symmetrically. ( similar to ":b64" )
i.e. I'm not afraid of "breaking" the JSON formatted output here, since it's already "broken" for that use case.

This looks overkill, although `:b64` or `:addr` is useful as encoding hints. Let's simply emit valid json here - JSONHandle and JSONStrictHandle are interchangable for encoding-decoding in Go code so there is no risk.

tsachiherman:
> > What would happen if we would follow the same pattern used in msgpacktool, and add ":int" to the key's string when encoding it as JSON ? this would allow the reverse operation to perform that symmetrically. ( similar to ":b64" )
> > i.e. I'm not afraid of "breaking" the JSON formatted output here, since it's already "broken" for that use case.
> 
> This looks overkill, although `:b64` or `:addr` is useful as encoding hints. Let's simply emit valid json here - JSONHandle and JSONStrictHandle are interchangable for encoding-decoding in Go code so there is no risk.

I agree that we need to emit valid jsons here; and for most of the endpoints, it would not be an issue. However, when it comes for *some*, it does. Two examples:
- when requesting the pending transactions, the caller might want to recalculate the transaction ids. In order to do so, the caller would need to msgp these - which would require a canonical msgp->json->msgp path. ( no maps here, but we do have few ":b64" )
- when requesting a block, the caller might want to perform a validation to ensure that the block header matches the payset. This would not work unless we have a canonical msgp->json->msgp.

by fixing it as is now, we're hosing ourselves since it would require us to break backward compatibility in the future.

algorandskiy:
The caller needs to stick to msgp in these scenarios. Or we can implement "cjson" (canonical json) format later if needed.

winder:
> The caller needs to stick to msgp in these scenarios. Or we can implement "cjson" (canonical json) format later if needed.

I like the `cjson` idea for a backwards-compatible way to add this if/when we need to support such a format.

tsachiherman:
The `cjson` sounds like a "hey, we messed up before - let's do it right this time" approach. Given that this is a new feature, I'd like to make sure we're not making any stupid mistakes here.

algorandskiy:
Adding ":b64" or ":int" suffixes to REST API json a bit bizarre for me. Our SDKs have REST API models, can load json data and then serialize them into mspg if needed. We do have models generator for Java and Go, and need to extend it for Python/JS to keep it in sync with REST automatically.

brianolson:
I think I'd like a policy that msgpack is always canonical and json is always best-effort and for convenience where API consumers find json easier than msgpack.

tsachiherman:
Would this handle correctly a non-utf8 map keys ?
i.e.
```golang
 v.Map["a\xc5z"] = "def"
```

I think that the root of the issue here is that the objects we're trying to return via the REST API were not designed to be returned using that interface. For the cases that we did design ( i.e. we == @winder ), we have no issues.

zeldovich:
Looks reasonable to me.  I believe we always use msgpack for the authoritative version of any data (blocks, transactions, account state, etc), when storing or signing things, so the question for this change is just about the clients of this REST API.  Seems good to emit valid JSON, if we can deal with existing clients (if any) that are expecting integer map keys.

I agree with Pavel that adding `:int` suffixes is a bit weird in this context.  The reason I added a `:b64` suffix in `msgpacktool` is because `msgpacktool` has no schema for the data it's transforming, and it has to guess the type when converting JSON back to msgpack.  This is not the case for REST API responses.

jasonpaulos (approved):
This looks like a good change to me. I remember users have ran into this issue before, and I believe we've told them to use the msgpack response as a workaround. This PR fixes the underlying issue, which is great.

PR #2615, opened 2021-07-23
Remove msgp 1.1.47 from go.mod/go.sum files

bricerisingalgorand:
These shouldn't be there anymore after upgrading to msgp 1.1.48


tsachiherman (approved):
looks good.

PR #2616, opened 2021-07-23
Bump arm integration jobs to large on circle

bricerisingalgorand:
## Summary

We have seem some test failures that have happened only for arm64 integration tests on PRs that don't touch the code. This PR raises the resource class to give those jobs more compute to see if that resolves the problem.

## Test Plan

The plan is to run jobs with this resource class and see if there's improvement in consistency.


algobarb (approved):
Looks good. Yay for small PRs

Issue #1426, opened 2020-08-18
Eliminate Lookup from ledgerForEvaluator and reuse LookupWithoutRewards


tsachiherman:
This interface is used in several locations ( i.e. accountUpdates ), and this is quite a trivial change.

PR #2618, opened 2021-07-23
CircleCI: parameterizing no_output_timeout and -short

algobarb:
## Summary
Parameterize no_output_timeout and -short in the general commands on the circle config file to reduce duplicate code.

I did also increase the size of integration tests to large as I saw that we had errors in mac_amd64_integration, arm64_integration and arm64_integration recently but would run after manually rerunning from failure.

## Test Plan
Tested it on a Pull request that would run the nightly and then tested the regular short tests once.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -285,13 +285,19 @@ commands:
         default: "/home/circleci"
       result_subdir:
         type: string
+      no_output_timeout:
+        type: string
+        default: 30m
+      go_test_flags:
+        type: string
+        default: "-short"
     steps:
       - attach_workspace:
           at: << parameters.circleci_home >>
       - run: mkdir -p /tmp/results/<< parameters.result_subdir >>
       - run:
           name: Run integration tests
-          no_output_timeout: 30m
+          no_output_timeout: << parameters.no_output_timeout >>
           command: |
             set -x
             export PATH=$(echo "$PATH" | sed -e 's|:<< parameters.circleci_home >>/\.go_workspace/bin||g' | sed -e 's|:/usr/local/go/bin||g')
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -385,7 +311,7 @@ commands:
             scripts/configure_dev.sh
             scripts/buildtools/install_buildtools.sh -o "gotest.tools/gotestsum"
             export ALGOTEST=1
-            export SHORTTEST=-short
+            export SHORTTEST=<< parameters.go_test_flags >>
```
algojack, line 314: since it can now be any kind of flags, shouldn't the left side change also (from SHORTTEST to TESTFLAGS or something?)

algobarb: SHORTTEST is an environment variable used in 1run_integration_tests.sh1 which I think calls `make` I think I want to avoid modifying existing code right now.

algobarb: Hmmm.. but you're right that doesn't make as much sense in this context as when it's called in the gotestsum command, so I think I should rename the parameters.go_test_flags to just parameters.short_test_flag or something. Thanks!

.circleci/config.yml
```diff
@@ -216,60 +216,20 @@ commands:
...
-            set -x
-            export PATH=$(echo "$PATH" | sed -e 's|:<< parameters.circleci_home >>/\.go_workspace/bin||g' | sed -e 's|:/usr/local/go/bin||g')
-            export KMD_NOUSB=True
-            export GOPATH="<< parameters.circleci_home >>/go"
-            export PATH="${PATH}:${GOPATH}/bin"
-            export ALGORAND_DEADLOCK=enable
-            GOLANG_VERSION=$(./scripts/get_golang_version.sh)
-            eval "$(~/gimme "${GOLANG_VERSION}")"
-            scripts/configure_dev.sh
-            scripts/buildtools/install_buildtools.sh -o "gotest.tools/gotestsum"
```
bricerisingalgorand, line 248: Was the vm setup stuff here unnecessary?

algobarb: I'm not sure, I kept what was already in the file

bricerisingalgorand: Ohh i see, the diff view just doesn't show that it's still in the command config from the duplicate command

bricerisingalgorand (approved):
LGTM

algojack (approved):

.circleci/config.yml
algobarb: Oh that's what you meant! Yeah that tripped me up when I first looked at the diff too.

.circleci/config.yml
```diff
@@ -216,60 +216,20 @@ commands:
...
-  generic_nightlytest:
-    description: Run nightly tests from build workspace, for re-use by diferent architectures
-    parameters:
-      circleci_home:
+      no_output_timeout:
         type: string
-        default: "/home/circleci"
-      result_subdir:
+        default: 30m
+      short_test_flag:
```
egieseke, line 230: I wonder if this could be made more generic to support any test flag?

egieseke (approved):
Looks good.

bricerisingalgorand (approved):
LGTM

.circleci/config.yml
algobarb: I'm not sure I want to do that anymore:
1. We're calling a bunch of other parameters inside the test itself right now, so it could be confusing to add an any_test flag because you can't add whatever parameter you want. Only parameters that haven't been specified yet
2. `-short` is the main flag we're using and it helps to repeat the same variable name used in generic_integration for consistency
3. Will mentioned that he wants us to stop calling gotestsum directly in the file eventually, so I expect further modifications on this command. (#1426 in our CircleCI epic)

PR #2619, opened 2021-07-23
fix TestVoteAggregatorBundles and check router.dispatch() output

cce:
## Summary

While working on the agreement package, I noticed the wrong messageEvent type was being used in TestVoteAggregatorBundles (voteVerified instead of a bundleVerified). This leads voteAggregator to use filterVote instead of filterBundle, and read the empty `messageEvent.Input.Vote` value (with round/period/step of 0/0/0) rather than the value of `messageEvent.Input.Bundle` that has the correct RPS and bundle of votes. This leads the events to be thrown away as from an old round instead of being processed by dispatch.

## Test Plan

This test didn't originally fail, I encountered the issue on a different branch where it was causing the test to crash — so I added an additional check to examine the output event returned by dispatch to make sure the votes were not being thrown away as too old.

### DIFF
--- a/agreement/voteAggregator_test.go
+++ b/agreement/voteAggregator_test.go
@@ -80,7 +80,9 @@ func TestVoteAggregatorVotes(t *testing.T) {
 				PlayerLastConcluding: 0,
 			}}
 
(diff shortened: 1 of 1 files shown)


tsachiherman (approved):
looks great.

PR #2621, opened 2021-07-26
E2E test: txnsync message rate increases with txn rate

grakshith:
## Summary

 An end-to-end test to confirm that the txnsync message rate increases with the txn rate.

## Test Plan

This is a test.

### DIFF
--- a/test/e2e-go/features/transactions/messageRate_test.go
+++ b/test/e2e-go/features/transactions/messageRate_test.go
@@ -18,6 +18,7 @@ package transactions
 
 import (
 	"bufio"
+	"context"
 	"encoding/json"
 	"fmt"
 	"math"
@@ -92,11 +93,23 @@ func testMessageRateChangesWithTxnRate(t *testing.T, templatePath string, txnRat
 	a.NoError(err)
 	listeningURL := strings.Split(listeningURLRaw, "//")[1]
 
-	// seek to `bytesRead` bytes when reading the log file
-	bytesRead := 0
+	// // seek to `bytesRead` bytes when reading the log file
+	// bytesRead := 0
 
 	// store message rate for each of the txn rates
 	prevMsgRate := 0.0
+
+	errChan := make(chan error)
+	resetChan := make(chan bool)
+	msgRateChan := make(chan float64)
+	ctx, stopParsing := context.WithCancel(context.Background())
+	defer stopParsing()
+
+	// parseLog continuously monitors the log for txnsync messages
+	// resetChan is used to signal it to send results on msgRate chan
+	// and reset its internal counters
+	// errChan is used to propagate errors if any
+	go parseLog(ctx, logPath, listeningURL, errChan, msgRateChan, resetChan)
 
 	for _, txnRate := range txnRates {
 		startTime := time.Now()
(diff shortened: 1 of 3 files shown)


test/e2e-go/features/transactions/messageRate_test.go
```diff
@@ -0,0 +1,179 @@
...
+
+func parseLog(logPath string, startByte int) (float64, int, error) {
+	file, err := os.Open(logPath)
+	if err != nil {
+		return 0, 0, err
+	}
+	defer file.Close()
+
+	// start reading log file from startByte
+	_, err = file.Seek(int64(startByte), 0)
```
algonautshant, line 145: I think something is going wrong here. 
Can you verify with the sequence number in  "Outgoing Txsync \#seqNum" that each time you are starting from the right line in the log?
If this is wrong and is getting tricky, you can use the sequence number to anchor where to start reading from the file.

test/e2e-go/features/transactions/messageRate_test.go
```diff
@@ -0,0 +1,179 @@
...
+	a.NoError(err)
+	var pongAccount string
+	for _, acct := range pongAccounts {
+		if acct.Address == pingAccount {
+			continue
+		}
+		// we found an account.
+		pongAccount = acct.Address
+		break
+	}
```
tsachiherman, line 87: The ping/pong accounts are not needed. You just need a single account from which you'll send payment transactions to itself.

test/e2e-go/features/transactions/messageRate_test.go
```diff
@@ -0,0 +1,179 @@
...
+	minTxnFee, minAcctBalance, err := fixture.CurrentMinFeeAndBalance()
+	a.NoError(err)
+
+	transactionFee := minTxnFee + 5
+	amountPingSendsPong := minAcctBalance * 3 / 2
+
+	// build the path for the primary node's log file
+	var logPath strings.Builder
+	logPath.WriteString(fixture.PrimaryDataDir())
+	logPath.WriteString("/node.log")
```
tsachiherman, line 100: avoid assembling paths yourself. Use `filepath.Join(fixture.PrimaryDataDir(), "node.log")` to accomplish that.

tsachiherman (changes requested):
The test itself looks reasonable - but when reviewing the operational parameters, these could definitely get improved:
1. we need to test it against much higher txnrates. exponentially increasing rates would make more sense here. i.e. [50, 300, 1500, 2000]
2. in order to implement that, you'll need to avoid using the kmd for signing transactions. instead, you'll need to sign these on your own.
3. The parseLog approach is very simple - but not very effective. it would be much better to create a background go-routine that would continuously "listen" for the new lines coming in and process them on the go.

test/e2e-go/features/transactions/messageRate_test.go
grakshith: Fixed.

test/e2e-go/features/transactions/messageRate_test.go
grakshith: Fixed.

test/e2e-go/features/transactions/messageRate_test.go
grakshith: Fixed.

txnsync/mainloop.go
```diff
@@ -227,7 +227,7 @@ func (s *syncState) onTransactionPoolChangedEvent(ent Event) {
 	newBeta := beta(ent.transactionPoolSize)
 
 	// see if the newBeta is at least 10% smaller or bigger than the current one
-	if (float32(s.lastBeta)*(1.0-betaGranularChangeThreshold)) <= float32(newBeta) || (float32(s.lastBeta)*(1.0+betaGranularChangeThreshold)) >= float32(newBeta) {
+	if (float32(s.lastBeta)*(1.0+betaGranularChangeThreshold)) <= float32(newBeta) || (float32(s.lastBeta)*(1.0-betaGranularChangeThreshold)) >= float32(newBeta) {
```
algonautshant, line 230: I think this is still wrong. Now it will never go into the if condition. 
You need to change the || to &&.

grakshith: Yes, you are right. I will add these changes in a separate PR (with a unit test) so that this PR will just contain the end to end test.

test/e2e-go/features/transactions/messageRate_test.go
```diff
@@ -68,36 +69,28 @@ func testMessageRateChangesWithTxnRate(t *testing.T, templatePath string, txnRat
...
-
-	a.NotEqual(pingAccount, pongAccount, "accounts under study should be different")
+	account := accountsList[0].Address
 
 	minTxnFee, minAcctBalance, err := fixture.CurrentMinFeeAndBalance()
 	a.NoError(err)
 
 	transactionFee := minTxnFee + 5
-	amountPingSendsPong := minAcctBalance * 3 / 2
+	amount := minAcctBalance * 3 / 2
```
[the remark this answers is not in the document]
grakshith: Fixed.

test/e2e-go/features/transactions/messageRate_test.go
```diff
@@ -0,0 +1,264 @@
...
+	if testing.Short() {
+		txnRates = []uint{50, 300}
+	}
+	testMessageRateChangesWithTxnRate(t, filepath.Join("nettemplates", "OneNodeTwoRelays.json"), txnRates)
+}
+
+func throttleTransactionRate(startTime time.Time, txnRate uint, sentSoFar uint) {
+	timeDelta := time.Since(startTime)
+	currentTps := float64(sentSoFar) / timeDelta.Seconds()
+	if currentTps > float64(txnRate) {
```
algonautshant, line 57: The test should also check here if the expected transaction rate is being achieved, and fail here if the rate is not possible on the system where the test is running. 
Otherwise, the test will fail blaming the message rate instead of the transaction rate.

test/e2e-go/features/transactions/messageRate_test.go
```diff
@@ -0,0 +1,264 @@
...
+
+		// send reset on resetChan to signal the parseLog goroutine to send the msgRate and reset its counters
+		resetChan <- true
+
+		select {
+		case err := <-errChan:
+			a.Error(err)
+		case msgRate := <-msgRateChan:
+			aErrorMessage := fmt.Sprintf("TxSync message rate not monotonic for txn rate: %d", txnRate)
+			a.GreaterOrEqual(msgRate, prevMsgRate, aErrorMessage)
```
algonautshant, line 152: Will be helpful to also print the txnRate and msgRate even if the test is passing to have an idea on these numbers.

algonautshant (changes requested):

algonautshant:
The message rate is about 6.xx for transaction rates 50, 300 and 800. 
There is significant variation between runs. 
Although the overall bias is for the message rate to increase as the transaction rate increases, the run to run variations are large enough not to get the test consistently fail.

One observation I made is that the transaction rate is starting slow, and then increasing. Perhaps, the message rate should be given a "warmup" period before recording the numbers. 
Testing for the transaction rate will make this problem more obvious. 

I am not able to achieve 1300 transaction rate.

algonautshant:
After building algod with the fix (in different PR), the message rate issue is resolved: 
5, 16, 40, 73 respectively for 50, 300, 800 and 1300 tps.
However, 1300 tps is not achieved. It is more like 1100 tps.

test/e2e-go/features/transactions/messageRate_test.go
grakshith: Fixed.

test/e2e-go/features/transactions/messageRate_test.go
grakshith: Fixed.

algonautshant (changes requested):

tsachiherman (approved):

PR #2638, opened 2021-07-26
Replace `Balances.PutWithCreatable()`

tolikzinovyev:
## Summary

Delete `Balances.PutWithCreatable()` and add new functions that notify the COW that a creatable was created or deleted. Coming up next is similar two functions for asset/application local state, needed by indexer.

### DIFF
--- a/ledger/appcow.go
+++ b/ledger/appcow.go
@@ -222,7 +222,7 @@ func errAlreadyStorage(addr basics.Address, aidx basics.AppIndex, global bool) e
 }
 
 // Allocate creates kv storage for a given {addr, aidx, global}. It is called on app creation (global) or opting in (local)
-func (cb *roundCowState) Allocate(addr basics.Address, aidx basics.AppIndex, global bool, space basics.StateSchema) error {
+func (cb *roundCowState) AllocateApp(addr basics.Address, aidx basics.AppIndex, global bool, space basics.StateSchema) error {
 	// Check that account is not already opted in
 	allocated, err := cb.allocated(addr, aidx, global)
 	if err != nil {
@@ -241,11 +241,19 @@ func (cb *roundCowState) Allocate(addr basics.Address, aidx basics.AppIndex, glo
 	lsd.action = allocAction
 	lsd.maxCounts = &space
 
+	if global {
+		cb.mods.Creatables[basics.CreatableIndex(aidx)] = ledgercore.ModifiedCreatable{
+			Ctype: basics.AppCreatable,
+			Creator: addr,
+			Created: true,
+		}
+	}
+
 	return nil
 }
 
 // Deallocate clears storage for {addr, aidx, global}. It happens on app deletion (global) or closing out (local)
-func (cb *roundCowState) Deallocate(addr basics.Address, aidx basics.AppIndex, global bool) error {
+func (cb *roundCowState) DeallocateApp(addr basics.Address, aidx basics.AppIndex, global bool) error {
 	// Check that account has allocated storage
 	allocated, err := cb.allocated(addr, aidx, global)
 	if err != nil {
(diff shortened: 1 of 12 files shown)


algorandskiy:
What it the rationale of replacing one function with another? Why not `Put(addr, ad, ctype, cidx)`?

tolikzinovyev:
With indexer needs, there will be five scenarios:

- update account data
- update account data + create creatable
- update account data + delete creatable
- update account data + opt in creatable
- update account data + opt out creatable

Rather than blow up the interface of `PutWithCreatable()` I propose we have separate functions. This is not ideal, so I'm open to ideas.

I don't understand your suggestion though. What would `Put(addr, ad, ctype, cidx)` do?

algorandskiy:
> I don't understand your suggestion though. What would Put(addr, ad, ctype, cidx) do?

Same as `PutWithCreatable` - based on ctype it checks AssetParams or AppPArams and sets created (if not there) or deleted (if there).

tolikzinovyev:
This kind of switching seems error-prone. Also what do I do with opting in/out? Another parameter global/local in `Put()`?

algorandskiy:
Currently there is no need for local/global (well, it is in my branch and I went with `Allocate` there).
I think balances interfaces does not provide right abstraction anymore and that's a root of the problem. Originally there were no assets/apps, and no "complex" states like maps of assets/apps. The first modification is in registering asset/app creation/deletion, and the second - state allocation/deallocation. You see, `balances` interface was originally operating only on AccountData without knowing details about its modifications, but now it also tracks app storage, assets and maps creation that are in fact AccountData modifications but needed to be tracked separately for underlying layers. In my branch it also tracks holdings creation.
It looks like `balances` (or `Put` method) needs to be replaced by some high-level ops similar to `Move`. I attempted to draft it but ended with way to many methods. Just sharing some thoughts.

An alternative to replacing `PutWithCreatable` can be re-purposing Allocation/Deallocate for creatables as well so all asset params, holdings, app params and local state would follow more-less similar sequence of calls.

algorandskiy:
Please consider the following alternative: use `Allocate`/`Deallocate` for registering events for account data that require additional processing rather than a simple balance record updating
```golang

Allocate(addr basics.Address, cidx basics.CreatableIndex, ctype basics.CreatableType, data CreatableOptions) error
Deallocate(addr basics.Address, cidx basics.CreatableIndex, ctype basics.CreatableType, data CreatableOptions) error
```

`basics.CreatableType` will need to include `AppLocalCreatable` and `AssetHoldingCreatable` to allow using it for holdings and app local states. `AssetHoldingCreatable` will be no-op now but could be implemented when account data will be using a separate storage for all complex data that is in maps now.

`CreatableOptions` would be an aggregate of 4 subtypes specific for asset params, holdings (not needed now), app params, and app local states (state schema field).
This looks natural since allocating global state assumes app creation. Allocating `AppLocalCreatable` would not trigger new app state registering.
Having the same API for both assets and apps would be a good step in a direction of unification of these

tolikzinovyev:
@algorandskiy, thanks for the feedback! I've made the changes we discussed today.

ledger/appcow.go
```diff
@@ -619,7 +638,7 @@ func applyStorageDelta(data basics.AccountData, aapp storagePtr, store *storageD
 			delete(owned, aapp.aidx)
 		case allocAction, remainAllocAction:
 			// note: these should always exist because they were
-			// at least preceded by a call to PutWithCreatable
+			// at least preceded by a call to Put()
```
algorandskiy, line 641:
```
at least preceded by a call to AllocateApp
```

algorandskiy (approved):
One small comment remark

algorandskiy:
This follows the direction of unifying assets and apps (both param and states), so it is a good change.

ledger/appcow.go
tolikzinovyev: This is looking up in `owned` which is a copy of `data.AppParams`. It is set in `Put()`, isn't it?

algorandskiy: This is correct, the comment most likely refers to `owned` rather than `store.action`

tolikzinovyev: so there is no problem?

algorandskiy: yes, all looks correct

tolikzinovyev:
Thanks for review, Pavel! Can somebody merge?

tsachiherman:
> Thanks for review, Pavel! Can somebody merge?

sure. I was waiting to ensure you and Pavel completed your discussion..

tolikzinovyev:
Thanks!

PR #2639, opened 2021-07-27
txnsync: fix pingpong sending invalid transactions

tsachiherman:
## Summary

The pingpong was generating transactions whose validity depends on the execution order. The changes in this PR ensures that the transactions that are being sent are valid regardless of their ordering.

This becomes important as the transaction sync shuffles the order at which transactions are propagated across the network.

## Test Plan

Test against S1 network.


shared/pingpong/pingpong.go
```diff
@@ -234,6 +236,31 @@ func fundAccounts(accounts map[string]uint64, client libgoal.Client, cfg PpConfi
 			throttleTransactionRate(startTime, cfg, totalSent)
 		}
 	}
+	// wait until all the above transactions are sent, or that we have no more transactions
+	// in our pending transaction pool.
+	for {
```
egieseke, line 241: Does this assume that all nodes have the same set of transactions in their transaction pool?

tsachiherman: No - on this branch, nodes that do no have any online stake do not receive any transactions from the network. But I get your point - I'll revise the code and send it for review later on.

PR #2645, opened 2021-07-27
codec_tester: use runtime.Caller to search for //msgp:allocbound lines

cce:
## Summary

Since #2527, when you run msgp-generated tests that call `protocol.RunEncodingTest` from a cwd that doesn't contain "go-algorand" in the name and with GOPATH not set, the test may fail due to `//msgp:allocbound` not being read from the source, leading to "msgp: length overflow" errors (test data exceeding the allocbound).

## Test Plan

Adding a new test that fails, then fixing it in a follow-up commit.


tsachiherman:
I think that you forgot to add the `codec_tester.go` that you modified locally to this PR.

tsachiherman (approved):
looks good to me. I think that the test-of-a-test is an overkill here, but I guess it helped you to verify the fix..

PR #2647, opened 2021-07-28
Add initial golangci-lint action

cce:
## Summary

Run golangci-lint from Github actions using https://github.com/golangci/golangci-lint-action

## Test Plan

Adds new check to PRs and merges.


winder:
Woah, this is really cool!

cce:
Hrmm, too bad, reviewdog has not yet implemented the ability to pass through the different golangci-lint severity levels (errors vs warnings) to Github checks yet https://github.com/reviewdog/reviewdog/issues/343

.github/workflows/reviewdog.yml
```diff
@@ -0,0 +1,80 @@
...
+          echo "$GITHUB_WORKSPACE/bin" >> $GITHUB_PATH
+          echo "$RUNNER_WORKSPACE/$(basename $GITHUB_REPOSITORY)/bin" >> $GITHUB_PATH
+      - name: Install specific golang
+        uses: actions/setup-go@v2
+        with:
+          go-version: '1.16.6'
+      - name: Create folders for golangci-lint
+        run: mkdir -p cicdtmp/golangci-lint
+      - name: Check if custom golangci-lint is already built
+        id: cache-golnagci-lint
```
cce, line 51: typo here (and below)

.github/workflows/reviewdog.yml
```diff
@@ -0,0 +1,90 @@
...
+          --out-format line-number
+          -c .golangci-warnings.yml
+          --allow-parallel-runners
+          | reviewdog
+          -f=golangci-lint
+          -name="Lint Warnings"
+          -reporter=github-check
+          -filter-mode=added
+          -fail-on-error=false
+          -level=warning
```
cce, line 90: It seems to be showing up as red failure rather than a warning?

cce (commented):
great!

.github/workflows/reviewdog.yml
algojack: red warning, but non-blocking, i wasn't able to get it to show yellow if that's what you mean

.github/workflows/reviewdog.yml
algojack: fixed ty

.golangci.yml
```diff
@@ -1,35 +1,25 @@
...
-      original-url: github.com/algorand/go-algorand/cmd/partitiontest_linter
-#  govet:
-#    check-shadowing: true
+severity:
+  default-severity: error
 
 issues:
-  # don't use default exclude rules listed in `golangci-lint run --help`
+  # use these new lint checks on code since #2574
+  new-from-rev: 07d3c1a94a54a176a74c84ae762a41c746a93d3b
```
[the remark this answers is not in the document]
algojack: just did, ty

algobarb (commented):
This is pretty cool.
So we're adding in the ReviewDog workflow which would act on github actions and it calls the tests directly, separating errors from warnings. So we moved partition-test away from `.golangci.yml` tests to `.golangci-warnings.yml` so that we only get warnings from reviewdogs instead of errors on the PRs right? What does changing `.golangci.yml` `tests` from `true` to `false` do?

.golangci-warnings.yml
```diff
@@ -0,0 +1,76 @@
...
+      path: cmd/partitiontest_linter/plugin.so
+      description: This custom linter checks files that end in '_test.go', specifically functions that start with 'Test' and have testing argument, for a line 'partitiontest.ParitionTest(<testing arg>)'
+      original-url: github.com/algorand/go-algorand/cmd/partitiontest_linter
+
+severity:
+  default-severity: warning
+
+issues:
+  # use these new lint checks on code since #2574
+  new-from-rev: eb019291beed556ec6ac1ceb4a15114ce4df0c57
```
bricerisingalgorand, line 30: What happens when we add new linters, will we just update this line to the commit we add new linters?

.golangci-warnings.yml
```diff
@@ -0,0 +1,76 @@
...
+
+  # Disable default exclude rules listed in `golangci-lint run --help` (selectively re-enable some below)
+  exclude-use-default: false
+
+  # Maximum issues count per one linter. Set to 0 to disable. Default is 50.
+  max-issues-per-linter: 0
```
bricerisingalgorand, line 36: Very helpful comment, using 0 to disable this feels very counter intuitive lol

.golangci-warnings.yml
cce: yes, or make an effort to "clean" out the repo for that specific linter like I did in https://github.com/algorand/go-algorand/pull/2574

algojack:
> This is pretty cool.
> So we're adding in the ReviewDog workflow which would act on github actions and it calls the tests directly, separating errors from warnings. So we moved partition-test away from `.golangci.yml` tests to `.golangci-warnings.yml` so that we only get warnings from reviewdogs instead of errors on the PRs right? What does changing `.golangci.yml` `tests` from `true` to `false` do?

Tests: true -> false, tells the linters not to check test files that end in "_test.go". We have it true in warnings but false in blocking errors. And even in warnings we only limit them to partitiontest linter by excluding all other linters.

.golangci-warnings.yml
algojack: yes? @cce what do you think?

algobarb (approved):
It looks good to me.

bricerisingalgorand (approved):

PR #2648, opened 2021-07-28
txnsync: correctness e2e test

algonautshant:
### Correctness test: 
This test sets up a network with 2 nodes and 2 relays. 
The two nodes send payment transactions.
For each transaction, the test checks if the relays and the nodes (including the node that originated the transaction) have the transaction in the pool (i.e. the transactionInfo.ConfirmedRound == 0).

The tests needs to check for the transactions in the pool fast enough before they get evicted from the pool to the block.
To achieve this, it sends transactions during the first half of the round period, to give the test enough time to check for the transactions. 

### Fix for a possible data race

TxHandler processDecodedArray and syncState evaluateIncomingMessage
may execute at the same time.

When this happens, evaluateIncomingMessage reads transactionGroups to
count the number if transactions there. At the same time,
processDecodedArray will be replacing verifiedTxGroup arrays with a
copy.

This leads to data race.

This change caches the number of transactions in the groups to avoid
reading transactionGroups to count that number.

## Test Plan
This is a test.

### DIFF
--- a/test/e2e-go/features/transactions/txnsync_test.go
+++ b/test/e2e-go/features/transactions/txnsync_test.go
@@ -61,12 +61,7 @@ import (
 func TestTxnSync(t *testing.T) {
 	t.Parallel()
 
-	// maxParallelChecks is the number of goroutines will simultaniously check if the
-	// transaction is received by the node.
-	// If this is too large, the system will report too many open files.
-	// If too small, the txns will be moved to the block.
-	maxParallelChecks := 1
-	numberOfSends := 1
+	numberOfSends := 2
 	targetRate := 300 // txn/sec
 	if testing.Short() {
 		numberOfSends = 100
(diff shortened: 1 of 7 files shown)


test/e2e-go/features/transactions/txnsync_test.go
```diff
@@ -0,0 +1,374 @@
...
+			fmt.Printf("waiting for pending verificaitons [%d] %d / %d\n", unprocessed, x, maxWait)
+		}
+	}
+	require.Equal(t, 0, unprocessed, "missing %d transactions", unprocessed)
+}
+
+type transactionTracker struct {
+	t                   *testing.T
+	ctx                 context.Context
+	mu                  sync.Mutex
```
tsachiherman, line 267: could you please replace this sync.Mutex with deadlock.Mutex ?

PR #2649, opened 2021-07-28
Add GroupID global to TEAL

zlangley:
## Summary

Adds the global field `GroupID` to retrieve the group of the transaction. This follows the pattern established for `GroupSize`. 

## Test Plan

Added/ran unit tests.

### DIFF
--- a/data/transactions/logic/eval_test.go
+++ b/data/transactions/logic/eval_test.go
@@ -911,7 +911,11 @@ const globalV4TestProgram = globalV3TestProgram + `
 `
 
 const globalV5TestProgram = globalV4TestProgram + `
-// No new globals in v5
+global GroupID
+int 32
+bzero
+==
+&&
 `
 
 func TestGlobal(t *testing.T) {
@@ -941,6 +945,7 @@ func TestGlobal(t *testing.T) {
 			CreatorAddress, globalV5TestProgram,
 			EvalStateful, CheckStateful,
 		},
+		6: {GroupID, globalV5TestProgram, Eval, Check},
 	}
 	ledger := makeTestLedger(nil)
 	ledger.appID = 42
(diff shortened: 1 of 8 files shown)


data/transactions/logic/eval_test.go
```diff
@@ -911,7 +911,11 @@ const globalV4TestProgram = globalV3TestProgram + `
 `
 
 const globalV5TestProgram = globalV4TestProgram + `
-// No new globals in v5
+global GroupID
+int 32
+bzero
+==
+&&
```
zlangley, line 918: Ideally we should also have a test where GroupID is not empty...

brianolson:
That definitely looks like something missing that should be in there, but I kinda feel like this should be a `txn` field (fields.go TxnField) because Group is a field on a txn in the data model (it _will_ be the same for everything in a group, so `global` is also kinda true, but I'd put it in `txn`)

zlangley:
It seems to me that `Group` being a transaction field is just an implementation detail; logically, the Group ID is a property of the group, not the transaction, no? So my instinct is that `GroupID` belongs in the same scope as the other group properties (e.g., `GroupSize`), but if you still feel strongly the other way I can change it.

jannotti:
It's absolutely debatable, since we've (mostly) used global to mean "Truly global - things about the blockchain" and txn to mean "About a single txn within the group".  For me (and zach, it seems), the decider was GroupSize, which we did put under global. It seemed right to put GroupID and GroupSize in the same place.  And yet, "grp" is indeed truly in the individual txns, so it seems like you should be able to get it from there.

Honestly, it seems weird, but I could imagine allowing it from both.

data/transactions/logic/eval_test.go
jannotti: That should be doable with some hacking around in evalStateful_test.go, which builds up some groups with two txns in them.

data/transactions/logic/eval_test.go
```diff
@@ -941,6 +945,7 @@ func TestGlobal(t *testing.T) {
 			CreatorAddress, globalV5TestProgram,
 			EvalStateful, CheckStateful,
 		},
+		6: {GroupID, globalV5TestProgram, Eval, Check},
```
algorandskiy, line 948: I'm not sure why does it work, this assumes assembler version 6 but the field defined as version 5 constant
```
{GroupID, StackBytes, modeAny, 5},
```
This means the test is never called. I'll submit a PR checking `len(tests) <= AssemblerMaxVersion`

algorandskiy (changes requested):
Requesting two changes:
1. Add non-empty group id test
2. Rebase on master - `TestGlobal` will fail because of non-existing version 6

jannotti:
Redone as #2838 to deal with conflicts.

PR #2651, opened 2021-07-28
Update systemd-setup.sh to install system.json

bricerisingalgorand:
## Summary

Right now a system.json file is not created when setting up systemd through the setup-systemd.sh script. Therefore, users aren't letting their nodes know that the algod process will be managed with systemd. This change has the setup-systemd.sh script download the systemd system.json file and lets the user know they need to copy that into any systemd managed data directory.

## Test Plan

I ran the script on a raspberry pi and saw that it worked.

### DIFF
--- a/cmd/goal/messages.go
+++ b/cmd/goal/messages.go
@@ -75,7 +75,7 @@ const (
 	errorNodeRunning                  = "Node must be stopped before writing APIToken"
 	errorNodeFailGenToken             = "Cannot generate API token: %s"
 	errorNodeCreation                 = "Error during node creation: %v"
-	errorNodeManagedBySystemd         = "This node is managed by systemd, you must run the following command to make your desired state change to your node:\n\nsystemctl %s algorand.service"
+	errorNodeManagedBySystemd         = "This node is managed by systemd, you must run the following command to make your desired state change to your node:\n\nsystemctl %s algorand@$(systemd-escape $ALGORAND_DATA)"
 	errorKill                         = "Cannot kill node: %s"
 	errorCloningNode                  = "Error cloning the node: %s"
 	infoNodeCloned                    = "Node cloned successfully to: %s"
--- a/installer/algorand@.service.template
+++ b/installer/algorand@.service.template
@@ -30,6 +30,7 @@ After=network.target
 AssertPathExists=%I
 
 [Service]
+ExecStartPre=bash -c "[[ ! -f /home/pi/node/data/system.json ]] && echo '{\"shared_server\":true,\"systemd_managed\":true}' > /home/pi/node/data/system.json || :"
 ExecStart=@@BINDIR@@/algod -d %I
 User=@@USER@@
 Group=@@GROUP@@
--- a/cmd/goal/node.go
+++ b/cmd/goal/node.go
@@ -165,7 +165,7 @@ var startCmd = &cobra.Command{
 		}
 		onDataDirs(func(dataDir string) {
 			if libgoal.AlgorandDaemonSystemdManaged(dataDir) {
-				reportErrorf(errorNodeManagedBySystemd, "start")
+				reportErrorf(errorNodeManagedBySystemd)
 			}
 
 			nc := nodecontrol.MakeNodeController(binDir, dataDir)
@@ -237,7 +237,7 @@ var stopCmd = &cobra.Command{
 		}
 		onDataDirs(func(dataDir string) {
 			if libgoal.AlgorandDaemonSystemdManaged(dataDir) {
-				reportErrorf(errorNodeManagedBySystemd, "stop")
+				reportErrorf(errorNodeManagedBySystemd)
 			}
 
 			nc := nodecontrol.MakeNodeController(binDir, dataDir)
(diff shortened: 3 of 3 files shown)


algobarb (approved):
looks good

installer/algorand@.service.template
```diff
@@ -30,6 +30,7 @@ After=network.target
 AssertPathExists=%I
 
 [Service]
+ExecStartPre=bash -c "[[ ! -f /home/pi/node/data/system.json ]] && echo '{\"shared_server\":true,\"systemd_managed\":true}' > /home/pi/node/data/system.json || :"
```
onetechnical, line 33: Hardcoded path

```suggestion
ExecStartPre=bash -c "[[ ! -f %I/system.json ]] && echo '{\"shared_server\":true,\"systemd_managed\":true}' > %I/system.json || :"
```

onetechnical (changes requested):

cmd/goal/messages.go
```diff
@@ -75,7 +75,7 @@ const (
 	errorNodeRunning                  = "Node must be stopped before writing APIToken"
 	errorNodeFailGenToken             = "Cannot generate API token: %s"
 	errorNodeCreation                 = "Error during node creation: %v"
-	errorNodeManagedBySystemd         = "This node is managed by systemd, you must run the following command to make your desired state change to your node:\n\nsystemctl %s algorand.service"
+	errorNodeManagedBySystemd         = "This node is managed by systemd, you must run the following command to make your desired state change to your node:\n\nsystemctl %s algorand@$(systemd-escape $ALGORAND_DATA)"
```
winder, line 78: nit: This command doesn't work for all shells. What do you think about redirecting them to the website instead?
```suggestion
	errorNodeManagedBySystemd         = "This node is using systemd and should be managed with systemctl. For additional information refer to https://developer.algorand.org/docs/run-a-node/setup/install/#installing-algod-as-a-systemd-service"
```

onetechnical (dismissed):
Tentative approval; Will's comment on pointing to docs is interesting, and Jack's suggestion to remove the system.json file on stopping is also intriguing, so would be open to either/both those improvements.

algojack (approved):

egieseke (approved):
Looks good.

bricerisingalgorand:
I think deleting the system.json might be a good idea, but it depends on the purpose we have in mind for it. If it’s meant to only reserve certain interactions with the node after every start that makes sense, but if it’s meant to always restrict the node in some instances it wouldn’t. We may want to consider that design decision more before we make it.

I think Will is right that we should refer to the docs instead of giving prescriptive feedback here though, so I just committed that change 👍

onetechnical (approved):
Looks good; did you run it / does it work?

bricerisingalgorand:
Yep, I made sure the file gets created, doesn't get edited when it exists and here are the goal messages

```
pi@raspberrypi:~ $ ./goal node stop
This node is using systemd and should be managed with systemctl. For additional information refer to https://developer.algorand.org/docs/run-a-node/setup/install/#installing-algod-as-a-systemd-service
pi@raspberrypi:~ $ ./goal node start
This node is using systemd and should be managed with systemctl. For additional information refer to https://developer.algorand.org/docs/run-a-node/setup/install/#installing-algod-as-a-systemd-service
pi@raspberrypi:~ $ ./goal node restart
This node is using systemd and should be managed with systemctl. For additional information refer to https://developer.algorand.org/docs/run-a-node/setup/install/#installing-algod-as-a-systemd-service
```

algobarb (approved):
Output that Brice posted looks good.

cmd/goal/messages.go
```diff
@@ -75,7 +75,7 @@ const (
 	errorNodeRunning                  = "Node must be stopped before writing APIToken"
 	errorNodeFailGenToken             = "Cannot generate API token: %s"
 	errorNodeCreation                 = "Error during node creation: %v"
-	errorNodeManagedBySystemd         = "This node is managed by systemd, you must run the following command to make your desired state change to your node:\n\nsystemctl %s algorand.service"
+	errorNodeManagedBySystemd         = "This node is using systemd and should be managed with systemctl. For additional information refer to https://developer.algorand.org/docs/run-a-node/setup/install/#installing-algod-as-a-systemd-service"
```
egieseke, line 78: Possibly add "Detected system.json file in the data directory, this node is using systemd and should be ..."

egieseke (approved):
Looks good, one comment about the message.

tsachiherman:
@bricerisingalgorand - I'm not a bit fan of providing exact documentation links inside the binary since the documentation links could change over time, forcing us to update the codebase for no good reason.
regardless of that, while changing the error string you forgot to change the corresponding expect test, which is testing against that string. ( admittedly, the expect tests are currently disabled, since they were causing random build failures )

bricerisingalgorand:
> @bricerisingalgorand - I'm not a bit fan of providing exact documentation links inside the binary since the documentation links could change over time, forcing us to update the codebase for no good reason.
> regardless of that, while changing the error string you forgot to change the corresponding expect test, which is testing against that string. ( admittedly, the expect tests are currently disabled, since they were causing random build failures )

Oh shoot, thanks I'll fix that. Also I can change the error message back to a suggested systemctl command if that's better

PR #2652, opened 2021-07-28
partition router_test.go

algobarb:
## Summary

Add partition to daemon/algod/api/server/router_test.go

## Test Plan

Test on CircleCI via PR.

### DIFF
--- a/daemon/algod/api/server/router_test.go
+++ b/daemon/algod/api/server/router_test.go
@@ -35,7 +35,7 @@ type TestSuite struct {
 }
 
 func (s *TestSuite) SetupSuite() {
-	partitiontest.PartitionTest(t)
+	partitiontest.PartitionTest(s)
 	s.e = echo.New()
 	handler := func(context lib.ReqContext, context2 echo.Context) {
 		s.calls++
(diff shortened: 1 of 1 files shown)


daemon/algod/api/server/router_test.go
```diff
@@ -178,6 +191,6 @@ func (s *TestSuite) TestGetTransactionByID() {
 	assert.Equal(s.T(), callsBefore+1, s.calls)
 }
 func TestTestSuite(t *testing.T) {
-
+	partitiontest.PartitionTest(s)
```
winder, line 194: I think you would put partition here or in each of the tests, but not both. When we call `suite.Run` it will run all of the above tests.

daemon/algod/api/server/router_test.go
```diff
@@ -166,6 +178,7 @@ func (s *TestSuite) TestGetAsset() {
 	assert.Equal(s.T(), callsBefore+1, s.calls)
 }
 func (s *TestSuite) TestGetTransactionByID() {
+	partitiontest.PartitionTest(s)
```
winder, line 181: These probably need to be `partitiontest.PartitionTest(s.T())`

algobarb: ok let me update them

egieseke (approved):
Looks good.

PR #2653, opened 2021-07-28
replace the KMD usage in pingpong with direct signing

### LINKED ISSUE #2644: faster pingpong
## Problem

pingpong may currently be limited by the speed of doing API requests to kmd and/or this API connection within machine is inefficient and causes excess CPU load in testing.

## Solution

Improve pingpong to create key pairs internally and do signing within process. ( i.e. instead of using the kmd )

Keep key pairs as msgpack blob in current directory (or location based on flag) so that accounts can be re-used on a subsequent pingpong session.

## Urgency

This will improve testing of the system beyond 1000 TPS

AlgoStephenAkiki:
## Summary

Removes signing calls to KMD by creating internal hashmap and signing transactions locally.

Resolves #2644 

## Test Plan

Tested using performance system.

### DIFF
--- a/shared/pingpong/pingpong.go
+++ b/shared/pingpong/pingpong.go
@@ -19,7 +19,6 @@ package pingpong
 import (
 	"context"
 	"fmt"
-	"github.com/algorand/go-algorand/data/transactions/logic"
 	"math"
 	"math/rand"
 	"os"
@@ -30,6 +29,7 @@ import (
 	v1 "github.com/algorand/go-algorand/daemon/algod/api/spec/v1"
 	"github.com/algorand/go-algorand/data/basics"
 	"github.com/algorand/go-algorand/data/transactions"
+	"github.com/algorand/go-algorand/data/transactions/logic"
 	"github.com/algorand/go-algorand/libgoal"
 )
 
--- a/shared/pingpong/pingpong.go
+++ b/shared/pingpong/pingpong.go
@@ -40,6 +40,8 @@ type CreatablesInfo struct {
 	OptIns      map[uint64][]string
 }
 
+// pingPongAccount represents the account state for each account in the pingpong application
+// This includes the current balance and public/private keys tied to the account
 type pingPongAccount struct {
 	balance uint64
 	sk      *crypto.SignatureSecrets
@@ -49,7 +51,6 @@ type pingPongAccount struct {
 // WorkerState object holds a running pingpong worker
 type WorkerState struct {
 	cfg PpConfig
-	//accounts map[string]uint64
 	accounts map[string]*pingPongAccount
 	cinfo    CreatablesInfo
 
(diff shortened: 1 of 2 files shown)


brianolson:
Still should have local storage for generated accounts

shared/pingpong/pingpong.go
```diff
@@ -208,12 +220,12 @@ func fundAccounts(accounts map[string]uint64, client libgoal.Client, cfg PpConfi
 	}
 
 	fmt.Printf("adjusting account balance to %d\n", minFund)
-	for addr, balance := range accounts {
+	for addr, val := range accounts {
```
tsachiherman, line 223: `val` -> `acct`

shared/pingpong/pingpong.go
```diff
@@ -39,10 +40,17 @@ type CreatablesInfo struct {
 	OptIns      map[uint64][]string
 }
 
+type pingPongAccount struct {
```
tsachiherman, line 43: add a very short line of comment explaining what this struct is used for.

shared/pingpong/pingpong.go
```diff
@@ -39,10 +40,17 @@ type CreatablesInfo struct {
...
+	sk      *crypto.SignatureSecrets
+	pk      basics.Address
+}
+
 // WorkerState object holds a running pingpong worker
 type WorkerState struct {
-	cfg      PpConfig
-	accounts map[string]uint64
+	cfg PpConfig
+	//accounts map[string]uint64
```
tsachiherman, line 52: drop commented out lines.

shared/pingpong/pingpong.go
```diff
@@ -786,32 +798,25 @@ func (pps *WorkerState) constructTxn(from, to string, fee, amt, aidx uint64, cli
...
-		stxn, err = client.SignTransactionWithWalletAndSigner(h, nil, signer, txn)
+		stxn, err = txn.Sign(accounts[signer].sk), nil
 	} else if len(cfg.Program) > 0 {
 		// If there's a program, sign it and use that in a lsig
-		psig, err = client.SignProgramWithWallet(h, nil, signer, cfg.Program)
-		if err != nil {
-			return
-		}
+
+		psig = accounts[signer].sk.Sign(logic.Program(cfg.Program))
```
tsachiherman, line 810: I could be wrong, but I think that you might want to do
```golang
progb := logic.Program(cfg.Program)
psig = accounts[signer].sk.Sign(&progb)
```

tsachiherman (changes requested):
This PR seems to fail for the following test cases:
asset_transfer_small, asset_transfer_large, stateful_teal_small, stateful_teal_medium, stateful_teal_large.
Could you manually test that locally and make sure it generates valid transactions ?

tsachiherman (approved):
looks good.

Issue #2624, opened 2021-07-26
Expose testing helper functions from logic package


fergalwalsh:
## Problem

Testing Teal code is hard. I have tried multiple methods including using sdk transactions with the sandbox and using the dryrun endpoint. These are both slow and error prone and are end-to-end integration tests instead of unit tests.

This repository contains a framework for testing Teal programs which turns out to be a very useful way of writing unit tests for both stateful apps and logicsig programs. Unfortunately the members are all private to the `logic` package so the only way I have found to use them is to copy large sections of the code. This works fine but it's a risky problem when new releases are made.

## Solution

Expose some of the functions and types from `eval_test.go` & `evalStateful_test.go` such as:
[makeTestLedger](https://github.com/algorand/go-algorand/blob/58fd804cbc8094bdce0a80b2f8ae2badc3784fb0/data/transactions/logic/evalStateful_test.go#L98)

[defaultEvalParams](https://github.com/algorand/go-algorand/blob/58fd804cbc8094bdce0a80b2f8ae2badc3784fb0/data/transactions/logic/eval_test.go#L73)

[testApp](https://github.com/algorand/go-algorand/blob/58fd804cbc8094bdce0a80b2f8ae2badc3784fb0/data/transactions/logic/evalStateful_test.go#L762)

This is not an exhaustive list, just some examples. These particular members would not necessarily need to be made public, it would be sufficient to have public wrappers for them. 

Would this be possible? Is there any reason not to make this functionality public?

chaihoang:
Can you provide more information about your particular use case?  If these were exposed, how would it assist in what you're trying to achieve?

fergalwalsh:
Here is an example of a test I have written. It tests an app call within a transaction group with a specific initial ledger state and asserts that the app program passes and that the local state deltas are as expected.

```
func TestValidatorBootstrap(t *testing.T) {
	t.Parallel()

	source := getValidatorAppSource()

	txgroup := makeBootstrapTxnGroup()
	var txn = &txgroup[1]

	sender := txgroup[0].Txn.Sender

	ep := defaultEvalParams(nil, nil)
	ep.Txn = txn
	ep.Txn.Txn.ApplicationID = appId
	ep.TxnGroup = txgroup

	ledger := makeTestLedger(
		map[basics.Address]uint64{
			sender: 1000000,
		},
	)
	ledger.newApp(txn.Txn.Sender, appId, makeSchemas(0, 0, 0, 0))

[...]

PR #2661, opened 2021-07-30
Application transactions

jannotti:
Allow AVM 1.0 programs to "execute transactions"

This is a pretty big change so I will comment on individual pieces by reviewing my own code, but the high level summary is.

1) An app now has authority over an account which is the hash of its appid.
2) The app can create transactions with three new opcodes, tx_begin, tx_field, and tx_submit.
3) If the transaction so created would need to be authorized by that app account from 1, the effects occur.

So far, only pay and axfer are allowed. acfg and afrz are under consideration, but have separate issues written.

This makes ApplyData a recursive datastructure, so that the "inner" transactions of app calls can be encoded in EvalDelta and, themselves, have ApplyData.  However, this PR does not yet expose that new inner transactions list to the REST API.  Another issue covers that need.


data/basics/userBalance.go
```diff
@@ -379,6 +380,16 @@ type AssetParams struct {
...
 
+func (app AppIndex) ToBeHashed() (protocol.HashID, []byte) {
+	buf := make([]byte, 8)
+	binary.BigEndian.PutUint64(buf, uint64(app))
+	return protocol.AppIndex, buf
+}
+
+func (app AppIndex) Address() Address {
+	return Address(crypto.HashObj(app))
+}
```
jannotti, line 391: This is the address for which the app has authority.

data/transactions/logic/README.md
```diff
@@ -303,6 +303,7 @@ Global fields are fields that are common to all the transactions in the group. I
 | 7 | LatestTimestamp | uint64 | Last confirmed block UNIX timestamp. Fails if negative. LogicSigVersion >= 2. |
 | 8 | CurrentApplicationID | uint64 | ID of current application executing. Fails if no such application is executing. LogicSigVersion >= 2. |
 | 9 | CreatorAddress | []byte | Address of the creator of the current application. Fails if no such application is executing. LogicSigVersion >= 3. |
+| 10 | CurrentApplicationAddress | []byte | Address that the current application controls. Fails if no such application is executing. LogicSigVersion >= 5. |
```
jannotti, line 306: The app address is exposed to AVM with `global CurrentApplicationAddress`

data/transactions/logic/assembler.go
```diff
@@ -1157,6 +1174,17 @@ func typeUncover(ops *OpStream, args []string) (StackTypes, StackTypes) {
...
 
+func typeTxField(ops *OpStream, args []string) (StackTypes, StackTypes) {
+	if len(args) != 1 {
+		return oneAny, nil
+	}
+	fs, ok := txnFieldSpecByName[args[0]]
+	if !ok {
+		return oneAny, nil
+	}
+	return StackTypes{fs.ftype}, nil
```
jannotti, line 1185: tx_field attempts to report typing errors at assembly time

data/transactions/logic/backwardCompat_test.go
```diff
@@ -387,7 +388,7 @@ func TestBackwardCompatGlobalFields(t *testing.T) {
 	}
 	require.Greater(t, len(fields), 1)
 
-	ledger := makeTestLedger(nil)
+	ledger := logictest.MakeLedger(nil)
```
jannotti, line 391: I introduced a new package, logictest, to hold some of helper code that was in evalStateful_test.go.

data/transactions/logic/eval.go
```diff
@@ -159,6 +173,8 @@ type LedgerForLogic interface {
 	DelGlobal(key string) error
 
 	GetDelta(txn *transactions.Transaction) (evalDelta basics.EvalDelta, err error)
+
+	Perform(txn *transactions.Transaction, spec transactions.SpecialAddresses) error
```
jannotti, line 177: LedgerForLogic needs to be able to "Perform" a transaction now.

data/transactions/logic/eval.go
```diff
@@ -272,7 +290,8 @@ type evalContext struct {
 
 	stack     []stackValue
 	callstack []int
-	program   []byte // txn.Lsig.Logic ?
+	subtxn    transactions.Transaction // place to build for tx_submit
```
jannotti, line 293: tx_begin and tx_field work together to populate subtxn. tx_submit Performs()s it.

tx_submit should also clear it.

data/transactions/logic/eval.go
```diff
@@ -3154,3 +3199,124 @@ func opAppParamsGet(cx *evalContext) {
 	cx.stack[last] = value
 	cx.stack = append(cx.stack, stackValue{Uint: exist})
 }
+
+func authorizedSender(cx *evalContext, addr basics.Address) bool {
+	appAddr, err := cx.getApplicationAddress()
+	return err == nil && bytes.Equal(addr[:], appAddr)
+	// This should also accept accounts that have been rekeyed to the appAddr
```
jannotti, line 3206: Apps will also have authority over accounts that are rekeyed to their address.  This is both more consistent, and will allow some interesting tricks to "hand over" accounts to the app which can be used to store more state (in the account's local state for the app that now controls it!)

data/transactions/logic/eval.go
```diff
@@ -3154,3 +3199,124 @@ func opAppParamsGet(cx *evalContext) {
...
+	bytes, err := cx.getApplicationAddress()
+	if err != nil {
+		cx.err = err
+		return
+	}
+	var addr basics.Address
+	copy(addr[:], bytes)
+	cx.subtxn.Header = transactions.Header{
+		Sender:     addr,
+		Fee:        basics.MicroAlgos{Raw: cx.Proto.MinTxnFee},
```
jannotti, line 3222: The app's account is charged a min fee for using tx_perform.  Since we now allow fee pooling, a better idea is to allow "leftover" fees from the transactions in the group to cover these fees first.

data/transactions/logic/evalStateful_test.go
```diff
@@ -632,14 +228,14 @@ pop
...
 		map[basics.Address]uint64{
 			txn.Txn.Sender: 1,
 		},
 	)
-	ledger.newApp(txn.Txn.Sender, 100, basics.AppParams{})
-	ledger.balances[txn.Txn.Sender].locals[100]["ALGO"] = algoValue
-	ledger.newAsset(txn.Txn.Sender, 5, params)
+	ledger.NewApp(txn.Txn.Sender, 100, basics.AppParams{})
+	ledger.NewLocal(txn.Txn.Sender, 100, "ALGO", algoValue)
+	ledger.NewAsset(txn.Txn.Sender, 5, params)
```
jannotti, line 238: Cleaner ways to set up complicated balance tables for testing.  This was, coincidentally, recently requested from an outside contributor.

ledger/applications.go
```diff
@@ -235,3 +237,66 @@ func (al *logicLedger) DelGlobal(key string) error {
...
+	}
+	var empty transactions.ApplyData
+	if !ad.Equal(empty) {
+		return fmt.Errorf("Pay caused ad change %v", ad)
+	}
+	return nil
+}
+
+func (l *logicLedger) Perform(tx *transactions.Transaction, spec transactions.SpecialAddresses) error {
+	balances, err := l.balances()
```
jannotti, line 274: This is the heart of the matter, in the real (not mock) ledger.  There's work to be done here, particularly in how the effects are communicated, by packing them into ApplyData.

data/basics/userBalance.go
```diff
@@ -379,6 +380,16 @@ type AssetParams struct {
 	Clawback Address `codec:"c"`
 }
 
+func (app AppIndex) ToBeHashed() (protocol.HashID, []byte) {
+	buf := make([]byte, 8)
```
tsachiherman, line 384: just a wild idea - the address we're going to pick is a pseudo-account public address, right ?
if so, can we pick an address that has no private key ?

data/transactions/logic/eval.go
algorandskiy: not sure I understand. If I own account A and applicationAddress is P, then I can rekey A to P once. I do not see how  would P get more space from A since they are physically different.

data/transactions/logic/fields.go
```diff
@@ -323,10 +322,13 @@ const (
 	CurrentApplicationID
 
 	// v3
-
 	// CreatorAddress [32]byte
 	CreatorAddress
 
+	// v4
```
algorandskiy, line 328: // v5

ledger/eval.go
```diff
@@ -675,7 +675,7 @@ func (eval *BlockEvaluator) TransactionGroup(txads []transactions.SignedTxnWithA
 
 // prepareEvalParams creates a logic.EvalParams for each ApplicationCall
 // transaction in the group
-func (eval *BlockEvaluator) prepareEvalParams(txgroup []transactions.SignedTxnWithAD) (res []*logic.EvalParams) {
+func (eval *BlockEvaluator) prepareEvalParams(txgroup []transactions.SignedTxnWithAD, spec *transactions.SpecialAddresses) (res []*logic.EvalParams) {
```
algorandskiy, line 678: why do you need to pass it? `BlockEvaluator` has `eval.block.BlockHeader` data. For unit testing?

data/basics/userBalance.go
```diff
@@ -379,6 +380,16 @@ type AssetParams struct {
...
+	binary.BigEndian.PutUint64(buf, uint64(app))
+	return protocol.AppIndex, buf
```
algorandskiy, line 386: this name `protocol.AppIndex` was confusing, did not get it is a prefix

ledger/applications.go
```diff
@@ -235,3 +237,66 @@ func (al *logicLedger) DelGlobal(key string) error {
...
+
+func (al *logicLedger) balances() (apply.Balances, error) {
+	balances, ok := al.cow.(apply.Balances)
+	if !ok {
+		return nil, fmt.Errorf("cannot get a Balances object from %v", al)
+	}
+	return balances, nil
+}
+
+func (al *logicLedger) pay(from basics.Address, to basics.Address, amount uint64, spec transactions.SpecialAddresses) error {
```
algorandskiy, line 249: does not look to be used

ledger/applications.go
```diff
@@ -235,3 +237,66 @@ func (al *logicLedger) DelGlobal(key string) error {
...
+	default:
+		err = fmt.Errorf("%s tx in AVM", tx.Type)
+	}
+	if err != nil {
+		return err
+	}
+	if !ad.Equal(transactions.ApplyData{}) {
+		return fmt.Errorf("Perform caused ad change %v", ad)
+	}
+	return nil
```
algorandskiy, line 300: nice!
```golang
deltas := l.cow.deltas()
```
And then... repeat the same logic that is in accountUpdates? I wonder what is going to be written in ApplyData - new addr A balance / asset amount or a change? Full values are easier.

Please also do not forget to unit test the fact balance/asset changes from a cow created at `func (cb *roundCowState) StatefulEval` are propagated into parent cow and into a ledger eventually.
In this case ApplyData will continue serve informational purpose for indexer is it does now.

algorandskiy (commented):
Nice work and looks like a clever integration into existing code. I guess a little more code for updating `ApplyData` with side effects left, and quite a bit tests in ledger.
I do not like some code duplication in `Perform` and `tx_submit` but have no better ideas on how to get rid of it.

data/transactions/logic/backwardCompat_test.go
jannotti: It is not yet, but I want to do some integration tests in which ledger code uses one of these ledgers.  That's not possible across packages if the code lives in _test.go files.  I think we have too much "distance" between our unit tests, which are very self-contained, and our e2e tests which fire up entire networks and run against algod.  I want to make short tests, in _test.go files, that do "some" integration - they test a couple/few packages together.

data/transactions/logic/eval.go
jannotti: One at a time.  After `tx_perform` you can `tx_begin` again, set some fields, and `tx_perform` again.  If/when we allow the construction of groups, this field would have to become a slice, to hold onto multiple pending txns at once.

data/transactions/logic/eval.go
jannotti: A would also have to be opted into the app.  If it is, then P can store into its local state.  This state is more like global state (because nobody can take it away).  Normally, the extra state that an app gets when an address opts in is subject to ClearState at any time.

I suppose this is mostly already true. You could opt a fresh account into an account, and then rekey it to zero, so nobody could remove the local state.  So I guess the only real new thing here is that it effectively gives the app a separate account to store algos / assets into and spend from them.

data/transactions/logic/eval.go
jannotti: I think we would allow `txfield Fee` if we want Fee settable.  My real plan for Fee is:

Add up all the fees in the group, make sure it is at least n*min_fee. (That is standard fee pooling.)
For any amount that it exceeds that minimum, keep track, and use that overage first, rather than charge a fee in these txns.

I don't know if it's worth doing that though.  I guess the nice this is that it means this stuff would be usable without prefunding the app account.

data/basics/userBalance.go
jannotti: I think we're just putting our faith in the magic of cryptographic hashes, right?  The intent is that the newly allocated address does not have a private key (or, equivalently, such a key exists, but nobody can find it).  It's the same reasoning behind logicsig's controlling an address that we certainly hope does already belong to somebody.

I wrote all that before really realizing what you meant - I'll leave it in case it's helpful to someone else.  I now realize you're asking if, upon app creation, maybe we can pick our app address, as long as the account is not in use.  I think that's insecure because I may know your public address before you fund it (this would be particularly true of a templatized logic sig that has not been instantiated yet, it would also be true if I watch the transaction pool).  I should not be able to make my app use that address, or else when you "come into existence" the app will control your funds.

As for setting the address.  I think the rule you would want is that you can only give the app authority over an account that you yourself have authority.  That is, you certainly don't want to reset an app address to a big foundation address and ask it for money! So I believe that we accomplish this by approaching it from the opposite direction.  We allow you to rekey *your* account to the app's address. And then it has authority. But, of course, you can only rekey that account because you own it.

data/transactions/logic/assembler.go
```diff
@@ -1043,6 +1043,23 @@ func assembleAppParams(ops *OpStream, spec *OpSpec, args []string) error {
 	return nil
 }
 
+func asmTxField(ops *OpStream, spec *OpSpec, args []string) error {
+	if len(args) != 1 {
+		return ops.errorf("%s expects one argument", spec.Name)
+	}
+	fs, ok := txnFieldSpecByName[args[0]]
```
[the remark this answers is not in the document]
jannotti: I suppose so.  I was thinking of it as an evaluation issue, so you could look at the opTxField for the definitive list of what's allowed in any Teal version.  It's certainly true that the allowed fields at eval() time must never change for a teal version. So at some point we'll either need to do some version checking in that code, or we'll move to a table based design which would make it easier to use the same value to control assembly and evaluation.

data/basics/userBalance.go
jannotti: That's the way these are all written.  See program.go for example.  We could change them all, I guess.  The only think making this slightly odder is that I had to turn the uint64 into a []byte.

ledger/applications.go
jannotti: True, thanks.  That was the old design.

ledger/applications.go
jannotti: This is just my stub code so that I can wrap my head around what can happen in the ad during one of these subtxn evaluations.  I hope nothing, but it will require more thought.  For example, if we allow closeTo, I guess it would write into ad.  If we send algos to a new address, I think it would update their rewards.

data/basics/userBalance.go
jasonpaulos: I think @tsachiherman is actually asking about choosing an invalid ed25519 curve point such that no private key exists that could match the address. I believe the MainNet fee sink and rewards pool are invalid points, though I'm not sure how to derive such an invalid point.

tsachiherman: Correct - what I meant was to select a ed25519 which no private key exists for. This is a post-quantum guarantee equivalent, since no crypto operation would generate that point. The x coordinates on ed25519 curve are mostly on the positive side; in fact, if you'll pick a number on the negative side that is smaller than -486662, I believe that there won't be any solution for that. ( i.e. that number would not be on the curve.. )

data/transactions/logic/eval.go
jannotti: Latest commit implements the fee pooling into the app transactions.

data/basics/userBalance.go
jannotti: I had no idea that the public key was used so directly - I sort of assumed it seeded a PRNG that was then used to pick points. I assume by "negative", you mean the 32 bytes is treated as a two-complement u256. So really we could pretty just flip the high bit?

ledger/eval.go
jannotti: I was trying to avoid allocating the little `spec` object twice.  The caller needs to make it, then pass it in here and to eval.transaction. So I make it first and use it for both.

data/basics/userBalance.go
tsachiherman: I completely agree that we should consult a cryptographer or two here. The randomness you're referring to is used when generating the private key; the public key is a pretty trivial calculation from there : 
```c
int
crypto_sign_ed25519_seed_keypair(unsigned char *pk, unsigned char *sk,
                                 const unsigned char *seed)
{
    ge25519_p3 A;

    crypto_hash_sha512(sk, seed, 32);
    sk[0] &= 248;
    sk[31] &= 127;
    sk[31] |= 64;

    ge25519_scalarmult_base(&A, sk);
    ge25519_p3_tobytes(pk, &A);

    memmove(sk, seed, 32);
    memmove(sk + 32, pk, 32);

    return 0;
}
```

Note that there is another randomness taking place when verifying a signature ( irrelevant in here ).

I do like the idea that you've had of using a hash function, as it "distributes" the account addresses as well, btw.

data/transactions/logic/backwardCompat_test.go
jannotti: Also: https://github.com/algorand/go-algorand/issues/2624

data/transactions/logic/eval_test.go
```diff
@@ -994,6 +1028,11 @@ const globalV4TestProgram = globalV3TestProgram + `
 
 const globalV5TestProgram = globalV4TestProgram + `
 // No new globals in v5
```
jasonpaulos, line 1030: nit: the comment can disappear

ledger/applications.go
```diff
@@ -85,6 +87,15 @@ func (al *logicLedger) MinBalance(addr basics.Address, proto *config.ConsensusPa
 	return record.MinBalance(proto), nil
 }
 
+func (al *logicLedger) Authorizer(addr basics.Address) basics.Address {
+	record, err := al.cow.Get(addr, false) // pending rewards unneeded
+	if err != nil || record.AuthAddr.IsZero() {
```
jasonpaulos, line 92: If `err != nil`, does that mean no account exists for `addr`, so it's ok to interpret it as having no AuthAddr?

data/transactions/logic/eval.go
```diff
@@ -3154,3 +3206,165 @@ func opAppParamsGet(cx *evalContext) {
...
+
+	fee := cx.Proto.MinTxnFee
+	if cx.OverpaidFee != nil {
+		if fee < *cx.OverpaidFee {
+			fee = 0
+		} else {
+			// Use up the overpay to shrink the fee
+			fee -= *cx.OverpaidFee
+		}
+		// We don't change OverpiadFee here, because they
```
jasonpaulos, line 3241: sp

data/transactions/logic/README.md
jannotti: I agree, will do.

data/transactions/logic/TEAL_opcodes.md
```diff
@@ -464,6 +464,7 @@ FirstValidTime causes the program to fail. The field is reserved for future use.
 | 7 | LatestTimestamp | uint64 | Last confirmed block UNIX timestamp. Fails if negative. LogicSigVersion >= 2. |
 | 8 | CurrentApplicationID | uint64 | ID of current application executing. Fails if no such application is executing. LogicSigVersion >= 2. |
 | 9 | CreatorAddress | []byte | Address of the creator of the current application. Fails if no such application is executing. LogicSigVersion >= 3. |
+| 10 | CurrentApplicationAddress | []byte | Address that the current application controls. Fails if no such application is executing. LogicSigVersion >= 5. |
```
tsachiherman, line 467: Adding a controlled address is cool. Is there any application benefit for supporting multiple application controlled accounts for a single application ?

ledger/eval.go
```diff
@@ -728,15 +728,19 @@ func (eval *BlockEvaluator) transactionGroup(txgroup []transactions.SignedTxnWit
 	cow := eval.state.child(len(txgroup))
 
 	// Prepare eval params for any ApplicationCall transactions in the group
-	evalParams := eval.prepareEvalParams(txgroup)
+	spec := transactions.SpecialAddresses{
```
tsachiherman, line 731: it would be cleaner if you'll move the `transactions.SpecialAddresses` to be a member of the `BlockEvaluator`. That way, you won't need to pass it to `eval.transaction`, and would still have it available for `eval.prepareEvalParams`. As a bonus, it would also allow you to cleanup `testTransaction` and use it there as well.

test/scripts/e2e_client_runner.py
```diff
@@ -447,13 +447,12 @@ def main():
...
         create_kmd_config_with_unsafe_scrypt(env['ALGORAND_DATA2'])
 
-
     xrun(['goal', '-v'], env=env, timeout=5)
     xrun(['goal', 'node', 'status'], env=env, timeout=5)
 
     rs = RunSet(env)
     for scriptname in args.scripts:
-        rs.start(scriptname, args.timeout-10)
+        rs.start(os.path.abspath(scriptname), args.timeout-10)
```
tsachiherman, line 455: I think that this is a good change, but I'm not sure it has to be coupled with the rest of this PR.

data/transactions/logic/eval.go
```diff
@@ -3154,3 +3206,165 @@ func opAppParamsGet(cx *evalContext) {
...
+		// Over paying - accumulate into OverpaidFee
+		overpaid := paid - cx.Proto.MinTxnFee
+		if cx.OverpaidFee == nil {
+			cx.OverpaidFee = new(uint64)
+		}
+		*cx.OverpaidFee = basics.AddSaturate(*cx.OverpaidFee, overpaid)
+	} else {
+		underpaid := cx.Proto.MinTxnFee - paid
+		// Try to pay with OverpaidFee, else fail.
+		if cx.OverpaidFee != nil && *cx.OverpaidFee > underpaid {
```
tsachiherman, line 3360: not sure if I'm reading it correctly, but I think that you meant `*cx.OverpaidFee >= underpaid`, right ?

data/transactions/logic/eval.go
```diff
@@ -3154,3 +3206,165 @@ func opAppParamsGet(cx *evalContext) {
...
+		return basics.Address{}, err
+	}
+	return addr, nil
+}
+
+func (cx *evalContext) available_asset(sv stackValue) (basics.AssetIndex, error) {
+	aid, err := sv.uint()
+	if err != nil {
+		return basics.AssetIndex(0), err
+	}
```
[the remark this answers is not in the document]
jannotti: It's the usual deal - we currently want to know ahead of time all the things you will touch. In principle we could prefetch those accounts.  Else you could transfer, say, 15 different assets between the same two accounts, forcing us to load in the 15 different asset creator balance records.

I do think I'd like to relax the requirement to specify the application address though.  It just feels like a hassle, and it's a bounded requirement. Until then, you've rightly pointed out a bug, thanks.  I'll probably fix it to check for now, and decide on relaxing later.

ledger/applications.go
jannotti: Correct.  An account needs to exist to hold a rekey address to be rekeyed.  When it doesn't exist, it's controlled the usual way. That often means it can't do anything anyway, but it could be a logicsig or, now, an app address that simply hasn't been funded yet. It *could* still "sign" a txn.

data/transactions/logic/TEAL_opcodes.md
jannotti: There is at least one - suppose you wanted to escrow algos from many accounts, and always be wiling to give them back, with the rewards they've earned in the meantime.  Since rewards are impossible to calculate in contract, you'd need to segregate your deposits, rather than keep them in one pot.

I originally planned on having a way to create any number of app accounts, for exactly this reason (and some more obscure ideas). But I think that the ability to rekey to the app addr is powerful enough.  To use such a contract, a depositor would make a new account, opt it into the contract, and then, in an atomic group:
1. make a deposit
2. call the contract to bind the new address to the depositor (by writing the deposit address in address's app local state)
3. rekey the new account to the app account

Now, when depositing or withdrawing, the caller tells the contract where his deposit account is.  So the contract can use these extra accounts that were ceded over to it because of rekeying. Which also prevents the caller from messing with them ever again.

(If you really like cleanliness, you'd have an operation to ask the contract to close out the account if you wanted to withdraw it all.)

ledger/eval.go
jannotti: Good idea, since that also means avoiding many more allocations.

data/transactions/logic/eval.go
jannotti: Correct.  I'm happy to report that left this here as a reminder to look later , but found the bug with a test before I got around to it.

ledger/eval_test.go
```diff
@@ -452,6 +455,7 @@ func testEvalAppGroup(t *testing.T, schema basics.StateSchema) (*BlockEvaluator,
 	defer l.Close()
 
 	newBlock := bookkeeping.MakeBlock(genesisInitState.Block.BlockHeader)
+	newBlock.RewardsState = genesisInitState.Block.RewardsState
```
jannotti, line 458: @tsachiherman , as part of moving SpecialAddresses into the BlackEvaluator, I needed to add this to several tests - it copies the RewardState over from genesis to the next block.  That's probably not exactly how that should work, but it was sufficient for the test (the new block needed the addresses from the old).  Let me know if you think something else should be done instead.

ledger/eval.go
jannotti: This is now in eval.specials so no need to allocate or pass.

ledger/apptxn_test.go
```diff
@@ -0,0 +1,418 @@
...
+		ApplicationID: appIndex,
+		ForeignAssets: []basics.AssetIndex{asaIndex},
+		Accounts:      []basics.Address{addrs[0], addrs[1]},
+	}
+	eval = l.nextBlock(t)
+	eval.txgroup(t, &overpay, &clawmove)
+	l.endBlock(t, eval)
+
+	amount, _ := l.asa(t, addrs[1], asaIndex)
+	require.Equal(t, amount, uint64(1000))
```
algorandskiy, line 417: let's test txn counter in a block header as well - to double check it gets incremented

data/transactions/logic/eval.go
```diff
@@ -2673,35 +2745,44 @@ func opExtractNBytes(cx *evalContext, n int) {
...
 	}
 	addr := basics.Address{}
 	copy(addr[:], account.Bytes)
 	idx, err := cx.Txn.Txn.IndexByAddress(addr, cx.Txn.Txn.Sender)
+
+	if err != nil {
+		// Application address is acceptable. index is meaningless though
+		appAddr, _ := cx.getApplicationAddress()
+		if appAddr == addr {
+			return addr, uint64(0xffffffffffffffff), nil
```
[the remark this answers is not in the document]
jasonpaulos: Or maybe `math.MaxUint64`

jasonpaulos:
I've yet to look at everything, but here are my comments so far

jasonpaulos (commented):
I've yet to look at everything, but here are my comments so far

data/transactions/logic/assembler.go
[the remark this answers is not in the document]
jannotti: I'm deferring this to a followon PR. It will certainly be done, as it will allow us to make the opcode field specs automatically if we encode when (and if) fields are available in a table-driven way.

ledger/applications.go
jannotti: This turned out a bit differently than I originally expected.  It became much easier to encode all the possible things that could happen in ApplyData by encoding the entire inner transactions, along with *their* ApplyData, rather than attempt to "flatten" the ADs into a top-level AD.

data/basics/teal.go
```diff
@@ -123,80 +123,6 @@ func (sd StateDelta) Valid(proto *config.ConsensusParams) error {
...
-func (l LogItem) Equal(o LogItem) bool {
-
-	return l.ID == o.ID && l.Message == o.Message
-
-}
-
-// EvalDelta stores StateDeltas for an application's global key/value store, as
-// well as StateDeltas for some number of accounts holding local state for that
-// application
-type EvalDelta struct {
```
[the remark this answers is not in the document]
jannotti: It's pretty transaction specific.  At any rate, I moved it to break a circular reference (it contains a transactions.SignedTransactionWithAD now) rather than out of a strong belief that it "belongs" here.

data/transactions/logic/eval.go
```diff
@@ -3183,25 +3273,205 @@ func opAppParamsGet(cx *evalContext) {
...
+	appAddr, err := cx.getApplicationAddress()
+	if err != nil {
+		return false
+	}
+	return appAddr == cx.Ledger.Authorizer(addr)
+}
+
+func opTxBegin(cx *EvalContext) {
+	if cx.subtxn != nil {
+		cx.err = errors.New("tx_begin without tx_submit")
```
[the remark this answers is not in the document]
jannotti: No nesting allowed.  We only have one place to store txn data, so tx_begin cleans out the subtxn field. (This will have to change if we want inner groups, not just inner txns, but even then, no nesting.)

data/transactions/signedtxn.go
```diff
@@ -129,3 +130,32 @@ func WrapSignedTxnsWithAD(txgroup []SignedTxn) []SignedTxnWithAD {
...
+func FeeCredit(txgroup []SignedTxn, minFee uint64) (uint64, error) {
+	minFeeCount := uint64(0)
+	feesPaid := uint64(0)
+	for _, stxn := range txgroup {
+		if stxn.Txn.Type != protocol.CompactCertTx {
+			minFeeCount++
+		}
+		feesPaid = basics.AddSaturate(feesPaid, stxn.Txn.Fee.Raw)
+	}
+	feeNeeded, overflow := basics.OMul(minFee, minFeeCount)
```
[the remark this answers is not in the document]
jannotti: I'm going to resolve this, based on our discussion of congestion fee being outside consensus.

data/transactions/logic/eval.go
```diff
@@ -3183,25 +3273,205 @@ func opAppParamsGet(cx *evalContext) {
...
+	}
+	last := len(cx.stack) - 1
+	field := TxnField(uint64(cx.program[cx.pc+1]))
+	sv := cx.stack[last]
+	var i uint64
+	switch field {
+	case Type:
+		cx.subtxn.Txn.Type = protocol.TxType(sv.Bytes)
+	case TypeEnum:
+		i, cx.err = sv.uint()
```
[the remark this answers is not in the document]
jannotti: Can't use := to assign to cx.err. But I can move the i declaration in tighter.

ledger/appcow.go
```diff
@@ -482,48 +476,52 @@ func MakeDebugBalances(l ledgerForCowBase, round basics.Round, proto protocol.Co
...
 	// If program passed, build our eval delta, and commit to state changes
 	if pass {
 		evalDelta, err = calf.BuildEvalDelta(aidx, &params.Txn.Txn)
 		if err != nil {
-			return false, basics.EvalDelta{}, err
+			return false, transactions.EvalDelta{}, err
 		}
 		calf.commitToParent()
+		evalDelta.SetLogs(cx.Logs)
+		evalDelta.InnerTxns = cx.Effects
```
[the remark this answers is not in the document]
jannotti: Agree, will change Effects to InnerTxns.  I was changing my mind a lot about this naming, but "Inner Transactions" seems to be the winner.  It's in the REST API now!

data/transactions/teal.go
```diff
@@ -0,0 +1,113 @@
...
+
+	// TODO: check inner transactions are equal
+
+	return true
+}
+
+// SetLogs taks a simple slice of log messages and creates LogItems
+// with a zero ID. LogItem should probably go away, since logs will
+// always appear with the app that produced them, a simple string list
+// would be fine.
```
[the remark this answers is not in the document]
jannotti: Yes, that's right, which means my whole push to get an app id associated with a log message was a bad idea.  I should rip it out (because now it *must* be the current app).  We do lose the inherent ability to know ordering, callers would have to emit a marker log message when calling apps to ensure they could recover the order if it matters to them. (Not that they can call apps yet, but they will eventually.)

daemon/algod/api/server/v2/utils.go
```diff
@@ -295,6 +295,41 @@ func convertToLogItems(txn node.TxnWithStatus, aidx *uint64) (*[]generated.LogIt
 	return logItems, nil
 }
 
+func convertInners(txn *node.TxnWithStatus) *[]preEncodedTxInfo {
+	inner := make([]preEncodedTxInfo, len(txn.ApplyData.EvalDelta.InnerTxns))
```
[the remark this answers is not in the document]
jannotti: Seemed weird to pass these big things around by value.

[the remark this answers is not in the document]
jannotti: I did the following test, which seems to show inner-txns does not show up when empty:
```
TXID=$(${gcmd} app optin --app-id "$APPID" --from "${SMALL}" | app-txid)
# Rest succeeds, no stray inner-txn array
[ "$(rest "/v2/transactions/pending/$TXID" | jq '.["inner-txn"]')" == null ]
[ "$(balance "$SMALL")" = 998000 ] # 1000 fee
```
jq returns "null" when it doesn't exist (I also looked at the json).
So maybe omitempty is dealing with nil or []?

data/transactions/logic/TEAL_opcodes.md
```diff
@@ -1212,3 +1214,30 @@ bitlen interprets arrays as big-endian integers, unlike setbit/getbit
...
+- Prepare a new application action
+- LogicSigVersion >= 5
+- Mode: Application
+
+## tx_field f
+
+- Opcode: 0xb2 {uint8 transaction field index}
+- Pops: *... stack*, any
+- Pushes: _None_
+- Set field F of the current application action
```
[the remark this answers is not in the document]
jannotti: I believe that's true for all.  I could not convince myself I preferred consistency over this, which avoids shouting in the immediate docs, but has the parameter very obvious when it's in prose.

ledger/eval.go
```diff
@@ -463,6 +468,10 @@ func startEvaluator(l ledgerForEvaluator, hdr bookkeeping.BlockHeader, proto con
 			eval.block.BlockHeader.GenesisHash = eval.genesisHash
 		}
 		eval.block.BlockHeader.RewardsState = eval.prevHeader.NextRewardsState(hdr.Round, proto, incentivePoolData.MicroAlgos, prevTotals.RewardUnits())
+		eval.specials = transactions.SpecialAddresses{
+			FeeSink:     eval.block.BlockHeader.FeeSink,
+			RewardsPool: eval.block.BlockHeader.RewardsPool,
+		}
```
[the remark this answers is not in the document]
jannotti: startEvaluator() can be invoked with a block header that does not have them set up, so the first setting leaves them as the zero address in that case.  Then, after calling NextRewardsState, they are set up and can be set (again, but this time properly).  I could not convince myself that it was ok to remove the initial setting, since this second setting only happens conditionally.

I'll look a little closer to see if I can confirm this makes sense, or if the startEvaluator() calls should always have a block hdr with the special addrs set.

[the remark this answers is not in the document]
jannotti: Ok, improved. We were not filling in the RewardsState when creating a "next header" (because much of it needs to be calculated). Therefore it was empty at the start. But we need the special addresses to do evaluation.  So now, `bookkeeping.MakeBlock` carries forward the special addresses, even though it can't fill the entire RewardsState.  Therefore, only the top assignment in startEvaluator() is required.

data/transactions/logic/eval.go
```diff
@@ -281,23 +304,33 @@ func (ep EvalParams) log() logging.Logger {
...
-	scratch   scratchSpace
-
-	cost     int // cost incurred so far
-	logCalls int // number of log calls so far
-	logSize  int // log size of the program so far
+
+	subtxn *transactions.SignedTxn // place to build for tx_submit
+
+	// The transactions Performed() and their effects
+	Effects []transactions.SignedTxnWithAD
```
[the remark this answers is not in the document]
jannotti: I want to keep this close to subtxn, as they go together conceptually.  I also don't want this at the top because the first thing someone thinks when they read what an EvalContext is shouldn't be "oh, it can create transactions", it should "oh, it's the stack, the pc, the program bytes, etc"

Note that `Logs` is also exported, but of low importance.  `cost/Cost` might reasonably be exported soon.  I don't think that should control their position.

data/transactions/logic/eval.go
```diff
@@ -2253,33 +2298,56 @@ func (cx *evalContext) getLatestTimestamp() (timestamp uint64, err error) {
...
 	}
 	return uint64(cx.Ledger.GetCreatableID(groupIndex)), nil
 }
 
-func (cx *evalContext) getCreatorAddress() ([]byte, error) {
+func (cx *EvalContext) getCreatorAddress() ([]byte, error) {
 	if cx.Ledger == nil {
 		return nil, fmt.Errorf("ledger not available")
 	}
-	addr := cx.Ledger.CreatorAddress()
```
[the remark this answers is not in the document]
jannotti: I agree I should remove the CreatorAddress() methods if we keep this change.  I think we should keep it, as I prefer to trim the LedgerForLogic interface, and special purpose code, when we have a more general purpose mechanism. Surely we cache things such that this is never a db lookup, since it's getting params for the currently running app.  Or don't we cache anything in roundCowState? (or perhaps below it?)

data/transactions/teal.go
```diff
@@ -0,0 +1,113 @@
...
+		return false
+	}
+	for i, l := range ed.Logs {
+		ok := l.Equal(o.Logs[i])
+		if !ok {
+			return false
+		}
+	}
+
+	// TODO: check inner transactions are equal
```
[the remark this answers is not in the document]
jannotti: done.  Weirdly.  I encoded them and check byte equality.  I can't believe how tedious field-by-field equality would be.  It's still got to be cheaper that hashing to get a txid internally though.

daemon/algod/api/algod.oas2.json
```diff
@@ -811,7 +811,10 @@
         ],
         "responses": {
           "200": {
-            "$ref": "#/responses/PendingTransactionResponse"
+            "description": "Given a transaction id of a recently submitted transaction, it returns information about it.  There are several cases when this might succeed:\n- transaction committed (committed round \u003e 0)\n- transaction still in the pool (committed round = 0, pool error = \"\")\n- transaction removed from pool due to error (committed round = 0, pool error != \"\")\n\nOr the transaction may have happened sufficiently long ago that the node no longer remembers it, and this will return an error.",
```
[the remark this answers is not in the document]
jannotti: You might have deleted some context on this one.  Should I resolve the conversation, or do you think there's something to be done here?

ledger/applications.go
```diff
@@ -234,17 +243,52 @@ func (al *logicLedger) DelGlobal(key string) error {
...
+		return ad, err
+	}
+
+	// We don't check min balances during in app txns.
+
+	// func (eval *BlockEvaluator) checkMinBalance will take care
+	// of it when the top-level txn concludes, as long as we
+	// ensure that cow there returns all changed accounts in
+	// modifiedAccounts().  It seems it must, but this note will
+	// be here until confirmed.
```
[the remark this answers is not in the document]
jannotti: confirmed and in unit test

algorandskiy (dismissed):
I guess all my remarks are resolved:
1. fee credit is calculated based on MinFee proto param in order to be deterministic. Max inner txn is set to 16.
2. min balance checks are applied to inner txns and accounts appear in cow's modifiedAccounts() that is confirmed in tests

`TxnCounter` update might be done in a separate PR

ledger/apptxn_test.go
```diff
@@ -0,0 +1,426 @@
...
+	eval = l.nextBlock(t)
+	eval.txn(t, withdraw.Noted("2"))
+	l.endBlock(t, eval)
+
+	amount, in = l.asa(t, appIndex.Address(), asaIndex)
+	require.True(t, in) // Zero left, but still opted in
+	require.Equal(t, amount, uint64(0))
+
+	eval = l.nextBlock(t)
+	eval.txn(t, withdraw.Noted("3"), "underflow on subtracting")
```
jasonpaulos, line 318: This is to test that an app can't overspend assets, correct? So why isn't the error "insufficient balance" or similar?

ledger/apptxn_test.go
```diff
@@ -0,0 +1,426 @@
...
+
+	amount, in = l.asa(t, appIndex.Address(), asaIndex)
+	require.True(t, in) // Zero left, but still opted in
+	require.Equal(t, amount, uint64(0))
+
+	close := txntest.Txn{
+		Type:            "appl",
+		ApplicationID:   appIndex,
+		Sender:          addrs[0],
+		ApplicationArgs: [][]byte{[]byte("close")},
```
jasonpaulos, line 329: It might be worth testing asset close with a nonzero balance, so you can verify the closeTo addr properly gets the assets.

test/scripts/e2e_subs/app-accounts.sh
```diff
@@ -0,0 +1,118 @@
...
+[ "$(rest "/v2/transactions/pending/$TXID" \
+        | jq '.["inner-txns"][0].txn.txn.amt')" = 20000 ]
+[ "$(rest "/v2/transactions/pending/$TXID?format=msgpack" | msgpacktool -d \
+        | jq '.["inner-txns"][0].txn.txn.type')" = '"pay"' ]
+# Now confirm it's in blocks API (this time in our internal form)
+ROUND=$(rest "/v2/transactions/pending/$TXID" | jq '.["confirmed-round"]')
+rest "/v2/blocks/$ROUND" | jq .block.txns[0].dt.itx
+
+[ "$(balance "$SMALL")" = 865000 ]   # 1 fee, 20,000 withdrawn
+[ "$(balance "$APPACCT")" = 129000 ] # 20k withdraw, fee paid by app account
```
jasonpaulos, line 90: So in this case the fee is "lost"? Meaning the local state for the deposited account still shows 130,000, but the app's balance is only 129,000 now.

This potentially exposes a larger issue: if an app wants to make sure it _never_ pays a fee when creating an inner transaction (e.g. fee credit always used), how could it do that?

daemon/algod/api/server/v2/utils.go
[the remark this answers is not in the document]
jasonpaulos: Strange, I will run this locally to see what's happening

test/scripts/e2e_subs/app-accounts.sh
jasonpaulos: I guess you could have the app assert that `txn Fee` is always greater than or equal to 2 * `global MinTxnFee`. But due to pooling fees that would be stricter than necessary in some cases (e.g. txn 1 pays a huge fee, txn 2 is an app call with 0 fee but fee credit is still positive).

The way to solve this problem in every situation might be to add a `global FeeCredit` field, which would diminish with each inner txn which draws from the excess fee credit. Though it's not perfect, since inner txn fees can't be inspected, so it would require assuming that their fee will always be `global MinTxnFee`. Then you could assert `global MinTxnFee` > `global FeeCredit` before `tx_submit`-ing.

Not sure if this needs to be addressed at this time, but worth thinking about.

jasonpaulos: Actually it might be better to ask for forgiveness instead of permission. And by that I mean use the `balance` op after `tx_submit` to verify the app account's balance is what you expect (i.e. no fee was taken, if that's the goal). If it's not what you expect, simply reject the app call.

data/transactions/logic/evalStateful_test.go
```diff
@@ -2925,54 +2469,10 @@ func TestAppLoop(t *testing.T) {
 	// Double until > 10. Should be 16
 	testApp(t, stateful+"int 1; loop: int 2; *; dup; int 10; <; bnz loop; int 16; ==", ep)
 
-	testApp(t, stateful+"int 1; loop: int 2; *; dup; int 10; <; bnz loop; int 16; ==", ep)
-
 	// Infinite loop because multiply by one instead of two
 	testApp(t, stateful+"int 1; loop:; int 1; *; dup; int 10; <; bnz loop; int 16; ==", ep, "dynamic cost")
 }
 
-func TestWriteLogs(t *testing.T) {
```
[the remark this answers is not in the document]
jannotti: None of that happens any more.  Instead of jamming the Logs down into the ledger and then yanking them back out to make the EvalDeltas, they are pulled directly out of the EvalContext to make the EvalDelta.

data/transactions/logic/evalAppTxn_test.go
```diff
@@ -0,0 +1,264 @@
...
+	ledger.NewAccount(ledger.ApplicationID().Address(), 10000) // plenty for fees
+	ledger.NewHolding(ledger.ApplicationID().Address(), 77, 3000, false)
+	testApp(t, "global CurrentApplicationAddress; int 77; asset_holding_get AssetBalance; assert; int 3000; ==;", ep)
+
+	testApp(t, "txn Sender; txn Accounts 1; int 100"+axfer, ep, "unauthorized")
+	testApp(t, "global CurrentApplicationAddress; txn Accounts 0; int 100"+axfer, ep,
+		fmt.Sprintf("Receiver (%s) not opted in", ep.Txn.Txn.Sender)) // txn.Sender (receiver of the axfer) isn't opted in
+	testApp(t, "global CurrentApplicationAddress; txn Accounts 1; int 100000"+axfer, ep,
+		"insufficient balance")
+	testApp(t, "global CurrentApplicationAddress; txn Accounts 1; int 100"+axfer+"int 1", ep)
```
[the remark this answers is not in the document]
jannotti: Done.

Second gives: Sender (F3W7PAA7ZTPJHJACQL55RBTN4QXG2ZWAEOXC6XLMPIQQPGDDIA4P6JYSKM) not opted in to 0

I suppose there's an argument for a clearer message - 0 isn't even a legal asset id.

data/transactions/logic/evalAppTxn_test.go
```diff
@@ -0,0 +1,264 @@
...
+	testApp(t, "tx_begin; int acfg; tx_field TypeEnum; tx_submit; int 1;", ep, "Invalid action type")
+	testApp(t, "tx_begin; int afrz; tx_field TypeEnum; tx_submit; int 1;", ep, "Invalid action type")
+	testApp(t, "tx_begin; int appl; tx_field TypeEnum; tx_submit; int 1;", ep, "Invalid action type")
+
+	// "insufficient balance" because app account is charged fee
+	// (defaults make these 0 pay|axfer to zero address, from app account)
+	testApp(t, "tx_begin; byte \"pay\"; tx_field Type; tx_submit; int 1;", ep, "insufficient balance")
+	testApp(t, "tx_begin; byte \"axfer\"; tx_field Type; tx_submit; int 1;", ep, "insufficient balance")
+	testApp(t, "tx_begin; int pay; tx_field TypeEnum; tx_submit; int 1;", ep, "insufficient balance")
+	testApp(t, "tx_begin; int axfer; tx_field TypeEnum; tx_submit; int 1;", ep, "insufficient balance")
```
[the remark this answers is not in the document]
jannotti: Because tx_begin notices that it can't pay the fee.

data/transactions/logic/evalAppTxn_test.go
```diff
@@ -0,0 +1,264 @@
...
+  tx_submit
+`
+
+	ep, ledger := makeSampleEnv()
+	ledger.NewApp(ep.Txn.Txn.Receiver, 888, basics.AppParams{})
+	testApp(t, "txn Sender; balance; int 0; ==;", ep)
+	testApp(t, "txn Sender; txn Accounts 1; int 100"+pay, ep, "unauthorized")
+	ledger.NewAccount(ep.Txn.Txn.Sender, 120+ep.Proto.MinTxnFee)
+	ledger.Rekey(ep.Txn.Txn.Sender, basics.AppIndex(888).Address())
+	testApp(t, "txn Sender; txn Accounts 1; int 100"+pay+"; int 1", ep)
```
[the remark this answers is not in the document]
jannotti: Agreed, I'll do it there.  (This ledger doesn't know how to close an account in the sense needed for this test.)

ledger/apptxn_test.go
jannotti: It's actually: "underflow on subtracting 101 from sender amount 100" or some such.  I suppose I could construct the entire message.

ledger/apptxn_test.go
```diff
@@ -0,0 +1,546 @@
...
+	eval = l.nextBlock(t)
+	eval.txgroup(t, &overpay, &clawmove)
+	l.endBlock(t, eval)
+
+	amount, _ := l.asa(t, addrs[1], asaIndex)
+	require.Equal(t, amount, uint64(1000))
+}
+
+// TestRekeyAction ensures an app can transact for a rekeyed account
+func TestRekeyAction(t *testing.T) {
```
jasonpaulos, line 448: This is a good test, but I don't think I properly explained that I was interested in a test which closes, reopens, and tries to transact on a rekeyed account all in one app call. I wrote up a test which does this following your examples, and it passes: https://gist.github.com/jasonpaulos/62a3724c62c07fad79741322b692d6cf

ledger/applications.go
jannotti: I'm going to spend a little time digging into this.  I share your concern. At the very least, perhaps we should return an error that explicitly communicates "NoAccount".  Then this is safe.

As I started thinking about how we handle this for top-level transactions, I realized we probably don't - if the account doesn't exist, who cares what its AuthAddr is? It can't participate in transactions, because it has nothing and can't even pay fees!

Except, no.  Nowadays, a properly authorized transaction should be able to do a clawback transaction, potentially without paying any fees, or having any algos or assets, because of fee pooling.  I have been told that this doesn't work and I've been meaning to investigate.  This is probably why - I bet we throw away a transaction that has this error.

algorandskiy (approved):

daemon/algod/api/server/v2/utils.go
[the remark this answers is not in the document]
jasonpaulos: After manually inspecting the REST endpoint, it appears this does properly omit the slice when it's empty. I'm not exactly sure why this works when pointers to empty slices returned from `AccountDataToAccount` _are not_ omitted, but that might be something to look into later.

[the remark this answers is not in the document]
jannotti: I'm willing to return nil if you want me to. Was hoping there was an answer that would keep things simple.

jasonpaulos (commented):
One tiny request, but other than this and updating the old LogItem tests we talked about, this is looking good

ledger/applications.go
jannotti: Widened interface to avoid ambiguity.

data/transactions/teal.go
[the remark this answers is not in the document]
jannotti: LogItem eliminated

jasonpaulos (approved):

data/transactions/teal.go
```diff
@@ -0,0 +1,111 @@
...
+
+	// When decoding EvalDeltas, the integer key represents an offset into
+	// [txn.Sender, txn.Accounts[0], txn.Accounts[1], ...]
+	LocalDeltas map[uint64]basics.StateDelta `codec:"ld,allocbound=config.MaxEvalDeltaAccounts"`
+
+	Logs []string `codec:"lg,allocbound=config.MaxLogCalls"`
+
+	// Intentionally, temporarily wrong - need to decide how to
+	// allocbound properly when structure is recursive.  Even a bound
+	// of 2 would allow arbitrarily large object if deep.
```
tolikzinovyev, line 42: What is the solution for this one?

jannotti: It turned out that the allocbounds are not our only protection.  We don't decode streamed msgpack messages.  We take them off the wire, with bounded size, and then decode.

tolikzinovyev: Given what Nickolai said about msgpack pre-allocating an array with a decoded integer size, I think there can an amplification attack where a few decoded bytes allocate 16 new `SignedTxnWithAD` objects.

Something like this when decoding `SignedTxnWithAD`:
```
dt itx 16 dt itx 16 dt itx 16 ...
```

PR #2662, opened 2021-07-30
Add a public `Eval()` function for indexer

tolikzinovyev:
## Summary

Add a public `Eval()` function in ledger that Indexer can use. The function accepts a custom protocol config so that Indexer can override it to always get asset close amounts, as opposed to only when the protocol supports it.

Closes https://github.com/algorand/go-algorand-internal/issues/1211.

## Test Plan

What kind of test can I write?


tsachiherman (approved):
looks good to me.

algorandskiy (approved):

PR #2664, opened 2021-07-30
txnsync: fix compression

tsachiherman:
## Summary

The `getMessageBuffer()` might return buffer that is bigger than what we need.
This doesn't present a problem on its own other than the fact that we were using that in order to detect the case where the compressed data consumes more space than the decompressed data.

For readability, I've renamed the `data` and `out` to `uncompressedData` and `compressedData` respectively.

## Test Plan

Use existing unit tests.

### DIFF
--- a/txnsync/txngroups.go
+++ b/txnsync/txngroups.go
@@ -99,28 +99,32 @@ func (s *syncState) encodeTransactionGroups(inTxnGroups []transactions.SignedTxG
 	}, nil
 }
 
-func (s *syncState) compressTransactionGroupsBytes(data []byte) ([]byte, byte) {
+func (s *syncState) compressTransactionGroupsBytes(uncompressedData []byte) ([]byte, byte) {
 	b := getMessageBuffer()
-	if cap(b) < len(data) {
+	if cap(b) < len(uncompressedData) {
 		releaseMessageBuffer(b)
-		b = make([]byte, 0, len(data))
+		b = make([]byte, 0, len(uncompressedData))
 	}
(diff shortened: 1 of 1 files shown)


txnsync/txngroups.go
```diff
@@ -99,28 +99,32 @@ func (s *syncState) encodeTransactionGroups(inTxnGroups []transactions.SignedTxG
...
 	}
-	if len(data) > len(out)*maxCompressionRatio {
-		s.log.Infof("compression exceeded compression ratio: compressed data len: %d", len(out))
+	if len(uncompressedData) > len(compressedData)*maxCompressionRatio {
+		s.log.Infof("compression exceeded compression ratio: compressed data len: %d", len(compressedData))
 		releaseMessageBuffer(b)
-		return data, compressionFormatNone
+		return uncompressedData, compressionFormatNone
+	} else if len(uncompressedData) <= len(compressedData) {
+		// compression wasn't effective and ended up spending more data.
```
nicholasguoalgorand, line 123: In theory this should never be reached. The function compress.Compress should return an error if the compression makes the compression bigger. See line 111.

tsachiherman: That would not happen if the buffer passed to `compress.Compress` is bigger than `len(uncompressedData)`

nicholasguoalgorand: Ah, I see, since we first attempt to use getMessageBuffer.

nicholasguoalgorand (approved):

PR #2668, opened 2021-07-30
Fix oapi-codegen installation.

winder:
## Summary

The REST API generate command is failing to install `oapi-codegen` because it needs an older version of `kin-openapi` than we define in go.sum. I'm attempting to install it with the buildtools script.

Also added REST server generation to the codegen test.

## Test Plan

CI


AlgoStephenAkiki (approved):

tsachiherman (approved):
Looks good

PR #2669, opened 2021-07-30
Extend global and txn field tests for all TEAL versions and protocols

algorandskiy:
## Summary

* Move tests from backwardCompat_test.go to fields_test.go
* Check every field on a version before it was introduced
* Enable txna fields checks
* Fix TestGlobal to prevent invalid version specs in tests map

## Test Plan

This is tests improvement PR


data/transactions/logic/fields_test.go
```diff
@@ -17,12 +17,241 @@
...
+		for v := uint64(0); v < field.version; v++ {
+			testLine(t, text, v, "...available in version...")
+		}
+		testLine(t, text, field.version, "")
+
+		ops := testProg(t, text, AssemblerMaxVersion)
+
+		// check on a version before the field version
+		preLogicVersion := field.version - 1
+		preConsensusVersion := logicVersionByProto[preLogicVersion]
```
tsachiherman, line 135: Since it's a unit test, get the second param from the map query and require it to be true

tsachiherman (dismissed):
Looks great, 1 small possible improvement.

data/transactions/logic/fields_test.go
```diff
@@ -17,12 +17,241 @@
...
+		proto := config.Consensus[preConsensusVersion]
+		if preLogicVersion < appsEnabledVersion {
+			require.False(t, proto.Application)
+		}
+		ep := defaultEvalParams(nil, nil)
```
jannotti, line 140: I'd rather see a loop like above, for testProg, but now you would create an `ep` using defaultEvalParamsWithVersion() to test that evaluation fails before field.version, and succeeds afterward.  I think the parsing consensus.go is a distraction - in the end your just finding the version that has the right LogicSigVersion, so you might as well just supply it directly.

data/transactions/logic/fields_test.go
algorandskiy: done

data/transactions/logic/fields_test.go
algorandskiy: I agree. I think I wanted to play with go ast parser. Fixed by generating fake params with 0-`LogicVersion`

data/transactions/logic/fields_test.go
```diff
@@ -17,12 +17,177 @@
...
+		if fs.version > 1 {
+			fields = append(fields, fs)
+		}
+	}
+	require.Greater(t, len(fields), 1)
+
+	consensusParamsByLogicVersion := make(map[uint64]config.ConsensusParams, LogicVersion)
+	for v := uint64(0); v <= LogicVersion; v++ {
+		consensusParamsByLogicVersion[v] = defaultEvalProtoWithVersion(v)
+	}
```
jannotti, line 52: No reason for the map anymore, right?  Just invoke defaultEvalProtoWIthVersion where you would do the lookup and avoid dealing with require.True(t, ok) afterward.

jannotti (approved):

data/transactions/logic/fields_test.go
algorandskiy: true, fixed

jannotti (approved):

PR #2671, opened 2021-07-30
Create New Participation Key Interface REST endpoints

### LINKED ISSUE #2593: Create New Participation Key Interface REST endpoints
Implement the interface and add stub functions.

Acceptance criteria:
* Update spec and regenerate code.

**Endpoint details from design document:**

```
POST /v2/participation/
consumes: application/msgpack
produces: application/json
```
Add a participation key to the node. The key format has not been changed and can be generated by goal or algokey.

For the first iteration, the endpoint can consume the file generated by algokey in the body of the request. This file happens to be a sqlite database file, which is an awkward format to depend on. In the future we should define our own internal format and msgpack encode it similar to all of the other data we pass around. For development purposes we can continue to pass around the sqlite database file as a deprecated and undocumented input parameter.

The return value is a simple json response containing the ParticipationID.

There is an additional constraint: make sure that the participation key file is unused. It must contain all of the original keys. Specifically: it should have keys for all rounds in the valid range. This restriction only applies to this endpoint.

```
DELETE /v2/participation/<participation-id>
produces: application/json
```
Delete a participation key by ID. The ID is a new type retrieved by the status endpoint. A simple success response JSON is returned (or maybe just a 200).
```
GET /v2/participation/
GET /v2/participation/<participation-id>
produces: application/json
```
Get information about installed participation keys. This information is similar to the goal account partkeyinfo command and includes additional information like the last time a key was used.

Returns a simple list of basic participation key information if no ID is provided. Otherwise returns more detailed information about the requested participation key.

AlgoStephenAkiki:
Resolves #2593 

Creates new participation key interface REST endpoints for participation keys.


daemon/algod/api/algod.oas2.json
```diff
@@ -462,59 +462,224 @@
...
+          "application/msgpack"
+        ],
+        "produces": [
+          "application/json"
+        ],
+        "schemes": [
+          "http"
+        ],
+        "summary": "Add a participation key to the node",
+        "operationId": "AddParticipationID",
```
winder, line 534:
```suggestion
        "operationId": "AddParticipationKey",
```

daemon/algod/api/algod.oas2.json
```diff
@@ -462,59 +462,224 @@
...
         "parameters": [
           {
-            "type": "string",
-            "description": "The `account-id` to update, or `all` to update all accounts.",
-            "name": "address",
-            "in": "path",
-            "required": true
+            "description": "The participation id to add to the node",
```
winder, line 537:
```suggestion
            "description": "The participation key to add to the node",
```

daemon/algod/api/algod.oas2.json
```diff
@@ -462,59 +462,224 @@
...
+            "$ref": "#/responses/PostParticipationResponse"
           },
-          {
-            "type": "integer",
-            "default": 1000,
-            "description": "The fee to use when submitting key registration transactions. Defaults to the suggested fee.",
-            "name": "fee",
-            "in": "query"
+          "400": {
+            "description": "Bad Request - Malformed Algorand transaction ",
```
winder, line 553:
```suggestion
            "description": "Bad Request ",
```

daemon/algod/api/algod.oas2.json
```diff
@@ -462,59 +462,224 @@
...
+          },
+          "default": {
+            "description": "Unknown Error"
           }
+        }
+
+      }
+
+    },
+    "/v2/participation/{participationid}": {
```
winder, line 584:
```suggestion
    "/v2/participation/{participation-id}": {
```

daemon/algod/api/algod.oas2.json
```diff
@@ -462,59 +462,224 @@
...
+            "description": "Unknown Error"
           }
         }
       },
       "parameters": [
         {
           "type": "string",
-          "description": "Account address where keys will be registered.",
-          "name": "address",
+          "name": "participationid",
```
winder, line 682:
```suggestion
          "name": "participation-id",
```

winder (commented):
Looks great, just a couple of nits

algorandskiy:
Shouldn't we keep existing `register-participation-keys` endpoint since it is general available?

winder:
> Shouldn't we keep existing `register-participation-keys` endpoint since it is general available?

I'm not sure. Since it was never implemented I wouldn't feel bad removing/renaming it, but technically you could consider it a breaking change.

AlgoStephenAkiki:
> > Shouldn't we keep existing `register-participation-keys` endpoint since it is general available?
> 
> I'm not sure. Since it was never implemented I wouldn't feel bad removing/renaming it, but technically you could consider it a breaking change.

I've added back the `register-partcipation-keys` endpoint.

algorandskiy:
>I'm not sure. Since it was never implemented I wouldn't feel bad removing/renaming it, but technically you could consider it a breaking change.

I did not notice it was not implemented. If so we can drop it indeed, no questions.

winder (approved):
LGTM.

We should remove `/v2/register-participation-keys/{address}` and `/v2/shutdown` but no need to do that here.

PR #2673, opened 2021-07-30
txnsync: fix beta updation condition

grakshith:
## Summary

The condition to check if the `beta` parameter should be updated was incorrect causing beta parameter to not update even with increasing transaction rate. This PR fixes the condition.

## Test Plan

Unit test added.


tsachiherman (approved):
Looks great

PR #2677, opened 2021-08-02
Participation Metrics

winder:
## Summary

Draft PR - Currently have the public interface and plumbing to call functions from the agreement code.

This PR adds a new SQLite DB to store metadata and metrics for participation keys.

## Test Plan

New unit tests TBD


agreement/abstractions.go
```diff
@@ -229,6 +229,15 @@ type KeyManager interface {
...
 	VotingKeys(votingRound, keysRound basics.Round) []account.Participation
+
+	// RecordVote sets the LastVote field for the Account.
+	RecordVote(account basics.Address, round basics.Round) error
+
+	// RecordBlockProposal sets the LastBlockProposal field for the Account.
+	RecordBlockProposal(account basics.Address, round basics.Round) error
+
+	// RecordCompactCertificate sets the LastCompactCertificate field for the Account.
+	RecordCompactCertificate(account basics.Address, round basics.Round) error
```
tsachiherman, line 240: The compact certificate is not related to the agreement.

agreement/abstractions.go
```diff
@@ -229,6 +229,15 @@ type KeyManager interface {
 	// valid for the provided votingRound, and were available at
 	// keysRound.
 	VotingKeys(votingRound, keysRound basics.Round) []account.Participation
+
+	// RecordVote sets the LastVote field for the Account.
+	RecordVote(account basics.Address, round basics.Round) error
+
+	// RecordBlockProposal sets the LastBlockProposal field for the Account.
+	RecordBlockProposal(account basics.Address, round basics.Round) error
```
tsachiherman, line 237: The vote and proposal aren't really that far apart from each other. In fact, tracking the proposal-vote is suffice to indicate a block proposals. I'd collapse these two into a single call, and add a bool indicating whether it's a proposal vote or not.

node/participation.go
```diff
@@ -0,0 +1,109 @@
...
+	return nil
+}
+
+func (db *participationDB) Register(id ParticipationID, on basics.Round) error {
+	return nil
+}
+
+func (db *participationDB) RecordVote(account basics.Address, round basics.Round) error {
+	return nil
+}
```
tsachiherman, line 101: make sure this one is non-blocking.

node/participation.go
```diff
@@ -0,0 +1,109 @@
...
+
+	// VRFSecrets
+	// OneTimeSignatureSecrets
+}
+
+// ParticipationIDNotFoundErr is used when attempting to update a set of keys which do not exist.
+var ParticipationIDNotFoundErr error
+
+func init() {
+	ParticipationIDNotFoundErr = fmt.Errorf("the participation ID was not found")
```
tsachiherman, line 51: nit:  `Errorf` -> `errors.New()`

node/participation.go
```diff
@@ -0,0 +1,109 @@
...
+
+import (
+	"fmt"
+	"github.com/algorand/go-algorand/crypto"
+	"github.com/algorand/go-algorand/data/basics"
+	"github.com/algorand/go-algorand/util/db"
+)
+
+// ParticipationID identifies a particular set of participation keys.
+type ParticipationID crypto.Digest
```
tsachiherman, line 27: I like this; you'll be able to define "shortID" method on top of that.

tsachiherman (commented):
looks like a promising start. I'd put all the new node files in `node/keymanager/...` since there will be more files around that.

agreement/abstractions.go
winder: I changed this so there's a single  `Record(basics.Address, basics.Round, <participation type>)` function.

winder: One problem with this was that I had to import the node package in the agreement code in order to get this new participation type.

To avoid that, I moved the current keymanager implementation into the data/account directory for this new type. The current partkey db code is in data/account, so this isn't unprecedented. I'm trying to find a better way, so this is may change.

tsachiherman: moving the `simpleKeyManager` to a different package makes it a non-testing code - which isn't really what we want.

I'm not sure that I followed the reasoning for that, as adding the `account.ParticipationAction` should not change any of the dependencies.

winder: These were 2 different changes.

> moving the `simpleKeyManager` to a different package makes it a non-testing code - which isn't really what we want.
I didn't realize that. So public packages that are only imported by _test.go files are excluded from the production binaries?


> I'm not sure that I followed the reasoning for that, as adding the `account.ParticipationAction` should not change any of the dependencies.
Previously it was `node.ParticipationAction` defined in `node/participation.go`, so the agreement abstraction would have had to include node.

I'll continue to think about these two items as I implement the interfaces.

agreement/abstractions.go
```diff
@@ -229,6 +229,10 @@ type KeyManager interface {
 	// valid for the provided votingRound, and were available at
 	// keysRound.
 	VotingKeys(votingRound, keysRound basics.Round) []account.Participation
+
+	// Record indicates that the given participation action has been taken.
+	// This function is also part of account.ParticipationRegistry.
+	Record(account basics.Address, round basics.Round, participationType account.ParticipationAction) error
```
tsachiherman, line 235: I'm wondering if we really need to have this `error` return variable. I can't think of anything useful we can do with it. Can you ? i.e. having the the `error` between the `participationRegistry` and the `AlgorandFullNode` makes sense, since we want the node to log all the errors. However, providing that feedback to the agreement isn't very helpful since we don't have any useful thing we can do with that.
( i.e. just ignoring it implies that the interface is wrong.. )

agreement/pseudonode.go
```diff
@@ -386,6 +386,7 @@ func (t pseudonodeVotesTask) execute(verifier *AsyncVoteVerifier, quit chan stru
 
 	var totalWeight uint64
 	for _, result := range verifiedResults {
+		t.node.keys.Record(result.v.R.Sender, result.v.R.Round, account.Vote)
```
tsachiherman, line 389: consider moving this to 
```golang
case t.out <- messageEvent{T: voteVerified, Input: r.message, Err: makeSerErr(r.err)}:
```
as it would be more CPU efficient as well as more "correct" in the case where we failed to persist.

data/account/participationRegistry.go
```diff
@@ -0,0 +1,108 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with go-algorand.  If not, see <https://www.gnu.org/licenses/>.
+
+package account
+
+import (
+	"errors"
```
tsachiherman, line 20: nit: space between the groups.

data/account/participationRegistry.go
```diff
@@ -0,0 +1,108 @@
...
+
+// ParticipationAction types
+const (
+	Vote ParticipationAction = iota
+	BlockProposal
+	CompactCertificate
+)
+
+// ParticipationIDNotFoundErr is used when attempting to update a set of keys which do not exist.
+var ParticipationIDNotFoundErr error
```
tsachiherman, line 59: nit: errors should start with ErrXXXX.
also - defining the error as 
```golang
var ParticipationIDNotFoundErr = errors.New("the participation ID was not found")
```
is perfectly fine.. no need to overkill it with init() functions.

node/node.go
```diff
@@ -269,6 +271,12 @@ func MakeFull(log logging.Logger, rootDir string, cfg config.Local, phonebookAdd
 	node.catchupService = catchup.MakeService(node.log, node.config, p2pNode, node.ledger, node.catchupBlockAuth, agreementLedger.UnmatchedPendingCertificates, node.lowPriorityCryptoVerificationPool)
 	node.txPoolSyncerService = rpcs.MakeTxSyncer(node.transactionPool, node.net, node.txHandler.SolicitedTxHandler(), time.Duration(cfg.TxSyncIntervalSeconds)*time.Second, time.Duration(cfg.TxSyncTimeoutSeconds)*time.Second, cfg.TxSyncServeResponseSize)
 
+	node.participationRegistry, err = ensureParticipationDB()
+	if err != nil {
+		log.Errorf("Cannot get participation metrics: %v", err)
```
tsachiherman, line 276: could you improve the text abit ? i.e. maybe something in the lines of `unable to initialize the participation database :  %v`

util/db/initialize_test.go
```diff
@@ -0,0 +1,219 @@
...
+		addToFoo(100),
+	}
+	err = Initialize(accessor, migrations)
+	require.NoError(t, err)
+
+	accessor.Atomic(func(ctx context.Context, tx *sql.Tx) error {
+		verifyFoo(111)(t, ctx, tx)
+		return nil
+	})
+}
```
tsachiherman, line 219: add EOL

tsachiherman (commented):
looks like promising progress.

agreement/abstractions.go
winder: I hadn't thought about it like that. I think that the `ParticipationRegistry` interface can return the error, and the agreement abstraction can remove it. In the shim function implemented by the FullNode I'll log a warning and the agreement code can ignore that case.

data/account/participationRegistry.go
```diff
@@ -0,0 +1,382 @@
...
+
+	return &participationDB{
+		store: accessor,
+	}, nil
+}
+
+// Queries
+var (
+	createKeysets = `CREATE TABLE Keysets (
+			pk INTEGER PRIMARY KEY,
```
tsachiherman, line 111: Add `NOT NULL` to the primary key definition - in sqlite it's not implied.

data/account/participationRegistry.go
```diff
@@ -0,0 +1,382 @@
...
+
+	// SELECT pk FROM Keysets WHERE participationID = ?
+	selectPK      = `SELECT pk FROM Keysets WHERE participationID = ? LIMIT 1`
+	selectLastPK  = `SELECT pk FROM Keysets ORDER BY pk DESC LIMIT 1`
+	selectRecords = `SELECT 
+			participationID, account, firstValidRound, lastValidRound, keyDilution,
+			lastVoteRound, lastBlockProposalRound, lastCompactCertificateRound,
+			effectiveFirstValidRound, effectiveLastValidRound
+		FROM Keysets, Rolling
+		WHERE Keysets.pk = Rolling.pk`
```
tsachiherman, line 144: make sure that you use a join here

data/account/participationRegistry.go
```diff
@@ -0,0 +1,382 @@
...
+			return fmt.Errorf("unable to update registered key: %w", err)
+		}
+
+		return nil
+	})
+}
+
+func (db *participationDB) Record(account basics.Address, round basics.Round, participationAction ParticipationAction) error {
+	var field string
+	switch participationAction {
```
tsachiherman, line 353: nit: use a predefined map to map the `participationAction` to database fields.

node/node.go
```diff
@@ -742,6 +751,16 @@ func (node *AlgorandFullNode) GetPendingTxnsFromPool() ([]transactions.SignedTxn
 	return bookkeeping.SignedTxnGroupsFlatten(node.transactionPool.PendingTxGroups()), nil
 }
 
+// ensureParticipationDB opens or creates a participation DB.
+func ensureParticipationDB(genesisDir string) (account.ParticipationRegistry, error) {
+	accessorFile := filepath.Join(genesisDir, config.ParticipationRegistryFilename)
+	accessor, err := db.MakeAccessor(accessorFile, false, false)
```
tsachiherman, line 757: avoid reinventing the wheel - use `db.Pair` which would get you reader/writer accessors.

data/account/participation.go
```diff
@@ -56,6 +57,24 @@ type Participation struct {
...
+	binary.Write(data, binary.LittleEndian, part.KeyDilution)
+	if part.VRF != nil {
+		data.Write(part.VRF.PK[:])
+	}
+
+	// this too?
+	//part.Write(part.Voting.SubKeyPK[:])
+
+	return ParticipationID(crypto.Hash(data.Bytes()))
+}
```
tsachiherman, line 76: The common way we calculate a hash of a structure is by msgp-ing it. Are you thinking of any particular reason we shouldn't do it here ?

winder: My understanding is that the subkeys will change over time, so this was an attempt to make sure the same keyset wouldn't produce different ParticipationIDs

data/account/participationRegistry.go
```diff
@@ -0,0 +1,385 @@
...
+		_, err = tx.Exec(setRegistered, on+maxBalLookback, record.LastValid, id[:])
+		if err != nil {
+			return fmt.Errorf("unable to update registered key: %w", err)
+		}
+
+		return nil
+	})
+}
+
+func (db *participationDB) Record(account basics.Address, round basics.Round, participationAction ParticipationAction) error {
```
tsachiherman, line 361: don't forget that you'd want to make this into a non-blocking : missing a database update because the disk access is slow, is fine. But you don't want to slow down the agreement due to slow disk performance.

winder: Do you think running this in a goroutine is sufficient? It looks like the synchronization options have a non-zero chance of DB corruption.

agreement/keyManager_test.go
```diff
@@ -0,0 +1,46 @@
+package agreement
+
+import (
+	"fmt"
```
tsachiherman, line 4: nit: add group spacing

agreement/keyManager_test.go
```diff
@@ -0,0 +1,46 @@
...
+func (m recordingKeyManager) DeleteOldKeys(r basics.Round) {
+}
+
+// Record implements KeyManager.Record.
+func (m recordingKeyManager) Record(acct basics.Address, round basics.Round, action account.ParticipationAction) {
+	if _, ok := m.recording[acct]; !ok {
+		m.recording[acct] = make(map[account.ParticipationAction]basics.Round)
+	}
+	if action == account.BlockProposal {
+		fmt.Println("it happens")
```
tsachiherman, line 43: I'd rather to avoid direct fmt.PrintX since it's not maskable. Instead, use either testlog or testing.Log
( you can pass these to the `makeRecordingKeyManager`)

winder: Oops this was for a breakpoint

data/account/participation.go
tsachiherman: While what you just wrote is correct, it's not what I meant.
I was thinking of -
```golang
serializing := Participation{
   FirstValid: part.FirstValid,
   ...
```
while omitting the changing parts ( i.e. Voting keys ).

then
```golang
   return ParticipationID(crypto.Hash(serializing.Marshal(nil)))
```

data/account/participationRegistry.go
```diff
@@ -0,0 +1,385 @@
...
+var ErrParticipationIDNotFound = errors.New("the participation ID was not found")
+
+// ErrInvalidRegisterRange is used when attempting to register a participation key on a round that is out of range.
+var ErrInvalidRegisterRange = errors.New("key would not be active within range")
+
+// ErrUnknownParticipationAction is used when record is given something other than the known actions.
+var ErrUnknownParticipationAction = errors.New("unknown participation action")
+
+// ParticipationRegistry contain all functions for interacting with the Participation Registry.
+type ParticipationRegistry interface {
```
tsachiherman, line 80: don't forget to add a Close() method, which will close the underlying database.
( and also - please make sure to do that as well in all the unit tests )

data/account/participationRegistry.go
winder: @tsachiherman I decided to keep it synchronous at this point and name the agreement abstraction `RecordAsync`. The node layer makes it non-blocking.

data/account/participationRegistry.go
```diff
@@ -0,0 +1,393 @@
...
+	"database/sql"
+	"errors"
+	"fmt"
+
+	"github.com/algorand/go-algorand/crypto"
+	"github.com/algorand/go-algorand/data/basics"
+	"github.com/algorand/go-algorand/util/db"
+)
+
+const maxBalLookback = 320
```
winder, line 30: @tsachiherman I don't want to bring in a dependency on the consensus params here, so I'd like to remove this and simply record the registration round:
`EffectiveFirst` -> `RegisteredAt`
`EffectiveLast` -> `DeregisteredAt`

Also, I think I have an off-by-one error and `EffectiveLast` == `EffectiveFirst` when replacing a key. Do you happen to know if `EffectiveLast` should be -1 or `EffectiveFirst` should be +1?

data/account/participation.go
```diff
@@ -17,10 +17,11 @@
 package account
 
 import (
+	"bytes"
 	"context"
 	"database/sql"
+	"encoding/binary"
 	"fmt"
-
```
tsachiherman, line 23: nit: leave this spacing line.

data/account/participationRegistry.go
tsachiherman: The `RegisteredAt` is very different from `EffectiveFirst`, and `EffectiveLast` is very different from `DeregisteredAt`. While I don't mind storing these, there is no "notable" gains here.
Btw - I always used the term "register/unregister".. so "deregister" sounds a bit weird. I went looking, and found this - https://english.stackexchange.com/questions/25931/unregister-vs-deregister. so, it looks like you're correct : since you're denoting the action rather then the state, it should be `Deregister` and it would make it's state `unregistered`.

As for the other question, the effective round should be real round numbers where the node can use the voting key. so, having these equal is legit.

data/account/participationRegistry.go
tsachiherman: I think that you'd rather do it here so that you can update the in-memory statistics without blocking, while allowing the persistence take longer.
i.e. Ideally, you would be able to respond to a http query without any blocking on the database.

data/account/participationRegistry.go
winder: For the second question let me clarify. When I'm overriding key1 with key2, I'm not sure which round to use for key1.EffectiveLast and key2.EffectiveFirst.

Currently I have it so `key1.EffectiveLast = key2.EffectiveFirst = currentRound + 320`, I'm confident that at least one of those is wrong.

PR #2679, opened 2021-08-02
Remove components.NodeContext, node.nodeContext and unused test implementations

cce:
## Summary

Spotted this dead code, and after discussing with @tsachiherman and @zeldovich, it should be fine to remove.

## Test Plan

Existing tests should pass.


tsachiherman:
I was under the impression that it's being used by some of the rest endpoint testing, but maybe it *was* used in the past and no longer needed.

winder (approved):
Looks like it was used by some agreement tests at one point: https://github.com/algorand/go-algorand-archive/commit/1414aa952804858bf5a9500ac1fdffccb791deb7#

Looking at `simulate.go`, I wasn't able to find where this code stopped being used... it seems to have disappeared somewhere between `go-algorand-archive` and `go-algorand`... but if deleting it doesn't break any tests, lets delete it.

PR #2680, opened 2021-08-02
Record modified local creatable states in state delta.

tolikzinovyev:
## Summary

Indexer needs the list of deleted local creatable states. This PR adds this information in ledger's state delta.

## Test Plan

Added unit tests.

### DIFF
--- a/ledger/apply/asset.go
+++ b/ledger/apply/asset.go
@@ -101,7 +101,11 @@ func AssetConfig(cc transactions.AssetConfigTxnFields, header transactions.Heade
 		}
 
 		// Tell the cow what asset we created
-		return balances.AllocateAsset(header.Sender, newidx, true)
+		err = balances.AllocateAsset(header.Sender, newidx, true)
+		if err != nil {
+			return err
+		}
+		return balances.AllocateAsset(header.Sender, newidx, false)
 	}
 
 	// Re-configuration and destroying must be done by the manager key.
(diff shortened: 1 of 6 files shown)


ledger/apply/asset.go
```diff
@@ -134,6 +138,10 @@ func AssetConfig(cc transactions.AssetConfigTxnFields, header transactions.Heade
 		if err != nil {
 			return err
 		}
+		err = balances.DeallocateAsset(creator, cc.ConfigAsset, false)
+		if err != nil {
+			return err
+		}
```
algorandskiy, line 144: double call does not look good, how about adding bit flags specifying global or local, or both.

tolikzinovyev: What's wrong with double call? Asset params and asset holdings are modified separately in account data.

ledger/ledgercore/statedelta.go
```diff
@@ -78,6 +90,11 @@ type StateDelta struct {
 	// previous block timestamp
 	PrevTimestamp int64
 
+	// Modified local creatable states. The value is true if the creatable local state
+	// is created and false if deleted. Used by indexer.
+	ModifiedAssetHoldings  map[AccountAsset]bool
+	ModifiedAppLocalStates map[AccountApp]bool
```
tsachiherman, line 96: since these two share an index, I would suggest making it into a map[CreatableIndex]CreatableType which would have a very similar cost, but would pack it all in one.

ledger/apply/asset.go
tolikzinovyev: The `Balance` interface is simpler as is.

tsachiherman: Calling twice is a clear indication that the interface is wrong.. I'd agree with Pavel that we should update the function call.

tsachiherman (approved):

PR #2685, opened 2021-08-04
testing: fix unit test TestAsyncTelemetryHook_QueueDepth

tsachiherman:
## Summary

Fix the unit test TestAsyncTelemetryHook_QueueDepth:
```golang
	// the anonymous goroutine in createAsyncHookLevels might pull an entry off the pending list before
	// writing it off to the underlying hook. when that happens, the total number of sent entries could
	// be one higher then the maxDepth.
```

## Test Plan

This is a test.


logging/telemetryhook_test.go
```diff
@@ -217,5 +217,10 @@ func TestAsyncTelemetryHook_QueueDepth(t *testing.T) {
 	close(filling)
 	hook.Close()
 
-	require.Equal(t, maxDepth, len(testHook.entries()))
+	hookEntries := len(testHook.entries())
+	require.GreaterOrEqual(t, hookEntries, maxDepth)
+	// the anonymous goroutine in createAsyncHookLevels might pull an entry off the pending list before
+	// writing it off to the underlying hook. when that happens, the total number of sent entries could
+	// be one higher then the maxDepth.
+	require.LessOrEqual(t, hookEntries, maxDepth+1)
```
winder, line 225: I think that moving `hook.ready = true` after the `Fire` call would remove the race condition.

This logging code is pretty hard to follow, but I think this is the main loop:
```
		for !exit {
			exit = !hook.waitForEventAndReady()

			hasEvents := true
			for hasEvents {
				select {
				case entry := <-hook.entries:
					hook.appendEntry(entry)
				default:
					hook.Lock()
					var entry *logrus.Entry
---->					if len(hook.pending) > 0 && hook.ready { // entry is always nil when ready = false
						entry = hook.pending[0]
						hook.pending = hook.pending[1:]
					}
					hook.Unlock()
					if entry != nil {
---->						err := hook.wrappedHook.Fire(entry) // therefore the mock will not be triggered
						if err != nil {
							Base().Warnf("Unable to write event %#v to telemetry : %v", entry, err)
						}
						hook.wg.Done()
					} else {
						hasEvents = false
					}
				}
			}
```

tsachiherman: I agree that the code is a bit spaghetti like; but I don't believe that changing `hook.ready` would help:
The variable is not being blocked upon. `waitForEventAndReady` doesn't wait for the `ready` flag, but rather just return the ready flag state.

The `filling` variable helps to ensure we block the `err := hook.wrappedHook.Fire(entry)` call, so that we can fill up the `hook.entries` channel. Then, once the channel is full, we unblock the Fire call.

tsachiherman: sorry - I misread your response. I believe that depending on the scenario we want to execute, there are more than one way to fix the test. My plan was to change the expectation, as the test current doing the following:
1. write the first entry to the hook.
2. the Fire is writing to the entries; entries is read and calling the inner Fire. The inner Fire is being block.
3. the rest of the entries are being written, leaving 10 entries in the pending array.
4. The inner Fire is getting unblocked.
5. The anonymous method finish iterating over the remaining entries.

The total number of entries is either 10 or 11, depending on whether the #2 was executed before all the entries were written.

winder: The test name is `TestAsyncTelemetryHook_QueueDepth`, so I believe moving the `ready` location forces the surplus events to be discarded according to the QueueDepth. So it seems like the right way to go.

`TestAsyncTelemetryHook_CloseDrop` also uses the ready flag, I don't think it tests "CloseDrop" at all.

The complexity of this code is pretty absurd, maybe we can revive the "Externalized Telemetry" project and replace it entirely with a better design.

winder (approved):
Stephen found a data race related to modifying the telemetry hook while entries were queued. Since there are plans to redo telemetry at some point, I don't think we should spend time investigating that race.

tsachiherman:
We shouldn't have any data races.. but I don't mind holding off with fixing the data race if it doesn't happen on production.

PR #2696, opened 2021-08-05
Renamed no_gimme env var to skip_go_installation

algojack:
## Summary

Renamed no_gimme env var to skip_go_installation

## Test Plan

Checking travis and circleci tests below


tsachiherman (commented):
looks good to me.. is there any particular reason that pushed you toward this change ?
( just curious, since I can't see any obvious reason )

winder:
This is a follow-up item from the circleci work. The variable was added due to a windows issue, I think the new name is more self-documenting.

winder (approved):
LGTM

algobarb (approved):
I wondered about the notes you added to the parent issue, but the changes here look good.

PR #2699, opened 2021-08-06
Add individual program length checks into updateApplication

algorandskiy:
## Summary

updateApplication does not enforce individual program length but looks like it should. This PR enforces that.

## Test Plan

Added tests


shiqizng (approved):

jannotti:
Isn't this enough, in WellFormed?
```
		lap := len(tx.ApprovalProgram)
		lcs := len(tx.ClearStateProgram)
		pages := int(1 + effectiveEPP)
		if lap > pages*proto.MaxAppProgramLen {
			return fmt.Errorf("approval program too long. max len %d bytes", pages*proto.MaxAppProgramLen)
		}
		if lcs > pages*proto.MaxAppProgramLen {
			return fmt.Errorf("clear state program too long. max len %d bytes", pages*proto.MaxAppProgramLen)
		}
```

jannotti (approved):

PR #2705, opened 2021-08-06
Consolidate SKIP_E2E_SUBS and E2E_SUBS_ONLY

algojack:
## Summary

Replaced SKIP_E2E_SUBS var and E2E_SUBS_ONLY var with E2E_SUBS var that has values of "ONLY" and "SKIP" as well as not set at all. We might need to add another value to this var in the future for another option.

## Test Plan

Check if tests work correctly still.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -374,7 +374,7 @@ jobs:
     resource_class: medium
     parallelism: 4
     environment:
-      E2E_SUBS: "SKIP"
+      E2E_SUBS_SCRIPTS: "SKIP"
     steps:
       - prepare_go
       - generic_integration:
@@ -387,7 +387,7 @@ jobs:
     resource_class: medium
     parallelism: 4
     environment:
-      E2E_SUBS: "SKIP"
+      E2E_SUBS_SCRIPTS: "SKIP"
     steps:
       - prepare_go
       - generic_integration:
@@ -399,7 +399,7 @@ jobs:
       image: ubuntu-2004:202104-01
     resource_class: large
     environment:
-      E2E_SUBS: "ONLY"
+      E2E_SUBS_SCRIPTS: "ONLY"
     steps:
       - prepare_go
       - generic_integration:
@@ -411,7 +411,7 @@ jobs:
       image: ubuntu-2004:202104-01
     resource_class: large
     environment:
-      E2E_SUBS: "ONLY"
+      E2E_SUBS_SCRIPTS: "ONLY"
     steps:
       - prepare_go
       - generic_integration:
@@ -459,7 +459,7 @@ jobs:
     resource_class: arm.medium
     parallelism: 4
     environment:
-      E2E_SUBS: "SKIP"
+      E2E_SUBS_SCRIPTS: "SKIP"
     steps:
       - checkout
       - prepare_go
@@ -473,7 +473,7 @@ jobs:
     resource_class: arm.medium
     parallelism: 4
     environment:
-      E2E_SUBS: "SKIP"
+      E2E_SUBS_SCRIPTS: "SKIP"
     steps:
       - checkout
       - prepare_go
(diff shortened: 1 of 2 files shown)


algobarb (dismissed):
I think this looks ok with the new change if we want to leave the option of adding another option in the future.

algojack:
Talked to Will, will be refactoring to use switch statement.

winder:
It looks like it would work, but has the same mixed skip/only pattern so in general, the design seems strange to me. There exists this undefined "other stuff" that doesn't need to be so mysterious.

I think selecting which tests to run would be easy to understand, and flexible for future subdivisions (like when we re-enable expect tests):
```
if [ $E2E_TESTS = "" || $E2E_TESTS="SCRIPTS" ]; then
    run_e2e_subs
fi
if [ $E2E_TESTS = "" || $E2E_TESTS="GO" ]; then
    run_go_tests
fi
if [ $E2E_TESTS = "" || $E2E_TESTS="EXPECT" ]; then
    run_go_tests --expect
fi
```

A switch/case could also be nice if wrapping everything in if's is too ugly

algojack:
@winder Changed to switch case and combined into one variable as we discussed. Refactoring everything in that file into it's own if statements and adding variables where it's used might be a bit outside of the scope of this ticket. So we should file another ticket if you want to refactor this file. What do you think?

winder:
This looks more complex with no benefit. My main concern was that we used both enable and skip so this has the same configuration complexity and extra code complexity. Why not use a simple if block that supports future cases with a direct configuration and simple code?

Either way I'll defer to the DevOps team. Whoever re-enables the expect tests will need to revisit this code.

algojack:
that's what i thought too, that's why i was confused when you said you wanted switch statement. I don't think we understand what you want here @winder

algojack:
putting this ticket in blocked/icebox until we have clear description and delivierables.

algojack:
@winder how about this? If this is not it, we should definitely have another call once you are back. Apologies if I misunderstood.

algobarb:
> It looks like it would work, but has the same mixed skip/only pattern so in general, the design seems strange to me. There exists this undefined "other stuff" that doesn't need to be so mysterious.
> 
> I think selecting which tests to run would be easy to understand, and flexible for future subdivisions (like when we re-enable expect tests):
> 
> ```
> if [ -z $E2E_TEST_FILTER || $E2E_TEST_FILTER="SCRIPTS" ]; then
>     run_e2e_subs
> fi
> if [ -z $E2E_TEST_FILTER || $E2E_TEST_FILTER="GO" ]; then
>     run_go_tests
> fi
> if [ -z $E2E_TEST_FILTER || $E2E_TEST_FILTER="EXPECT" ]; then
>     run_go_tests --expect
> fi
> ```
> 
> A switch/case could also be nice if wrapping everything in if's is too ugly

Why is the if statement checking for `-z $E2E_TEST_FILTER` across all of these? 
Do we expect to run all tests when `-z $E2E_TEST_FILTER`?

E2E_TEST_FILTER just seems a bit odd to me. Maybe it could check for `$E2E_TEST = "ALL"` so that it is more explicit when to run all of the tests and when it will only run only "SCRIPTS" and "GO" e2e tests?

algobarb:
> > Why is the if statement checking for `-z $E2E_TEST_FILTER` across all of these?
> > Do we expect to run all tests when `-z $E2E_TEST_FILTER`?
> 
> Yes, that's the default right now. We have places that run this without any flags.
> 
> > E2E_TEST_FILTER just seems a bit odd to me. Maybe it could check for `$E2E_TEST = "ALL"` so that it is more explicit when to run all of the tests and when it will only run only "SCRIPTS" and "GO" e2e tests?
> 
> I'm not sure that's where Will wants to go with this, so far as I understand it we want the default (no flags) to be defined to run these (vs not running)

In that case, I guess that makes sense.

algobarb (approved):

PR #2708, opened 2021-08-06
Return an error in `EncodeSignedTxn()` and `DecodeSignedTxn()` if consensus protocol is unknown.

tolikzinovyev:


node/indexer/indexer_test.go
```diff
@@ -34,6 +34,10 @@ import (
 	"github.com/algorand/go-algorand/test/partitiontest"
 )
 
+const genesisID string = "foo"
+
+var genesisHash crypto.Digest
```
tsachiherman, line 39: rename these to testGenesisID and testGenesisHash.
The testGenesisHash can be defined here as 
```golang
var testGenesisHash = crypto.Digest{0x1, 0x2, 0x3..}
```
which is abit cleaner than defining it here and initializing it in `init`

node/topAccountListener_test.go
```diff
@@ -284,14 +284,19 @@ func TestInit(t *testing.T) {
 
 func makeBlockWithTxnFor(senders []byte, receivers []byte) bookkeeping.Block {
 	var blk bookkeeping.Block
+	blk.BlockHeader.GenesisID = "foo"
+	blk.BlockHeader.GenesisHash[0] = 3
```
tsachiherman, line 288: I'm not sure why you need to make the GenesisID/GenesisHash.. but there is nothing wrong with these changes.

tsachiherman: it would generally be better if you were to pick a genesisHash that is really pseudo-random, i.e. crypto.Hash([]byte{1,2,3,4})

tsachiherman (changes requested):
looks good; one request for renaming global test variable names.

node/topAccountListener_test.go
tolikzinovyev: GenesisID and GenesisHash must the same in transactions and block header. Why is pseudo-random better?

node/indexer/indexer_test.go
tolikzinovyev: done

node/topAccountListener_test.go
tsachiherman: a true random means that the result of the test might be random.. and we don't want that.
filling up only selected bytes doesn't guarantee that the comparison was testing all the bytes.
(i.e. both me and you knows it does.. but we want to ensure we test that )

tolikzinovyev: Thanks. Done.

tsachiherman (approved):

PR #2709, opened 2021-08-06
Fix flaky ledger app tests that use commitRound directly

algorandskiy:
## Summary

The problem was in `commitSyncer` goroutine that triggered `commitRound` in additional to controlled `commitRound` calls.
Fixed by closing deferred commits channel and waiting `commitSyncer` to exit.

## Test Plan

Checked with -count 1000


tsachiherman (approved):
Looks good to me

PR #2711, opened 2021-08-07
Pool App call budget in group transactions

### LINKED ISSUE #2636: Pool computational budget for application call transactions
- Add up number of application calls in transaction group - budget should be 700 x that number
- After each evaluation, we need to know how much of the budget we used to carry it over for execution for the remaining app calls
- Update error messages

Tests:
- Test that verify is allowed in stateful Teal. Currently `ed25519verify` is only allowed in Stateless TEAL, but after this change, it should be allowed if the total pooled budget is greater than this cost. 
- Unit tests
- Protocol upgrade tests

algochoi:
## Summary

- Adds feature to allow App calls to pool their budget in a grouped transaction (e.g. in a 16 app call txn group, the entire group has access to 11200 units of pooled cost).
- Allows `ed25519verify` to be called in Application calls, given that there is enough budget in the pool
- Added `vFuture` consensus parameter

Closes #2636 

## Test Plan

- Added unit tests in `ledger/` to check that group transactions pool the App call budget (700 * number of app calls)
- Added tests to check that pooling is not allowed in V28
- Added tests to check that `ed25519verify` is allowed in App calls, given enough budget in the pool


config/consensus.go
```diff
@@ -108,6 +108,11 @@ type ConsensusParams struct {
 	// each Txn has a MinFee.
 	EnableFeePooling bool
 
+	// EnableAppFeePooling specifies that the sum of fees for application calls
+	// in a group is checked against the sum of the budget for application calls,
+	// rather than check each individual app call is within the budget.
+	EnableAppFeePooling bool
```
jannotti, line 114: Let's try to avoid the word Fee in this context.  We'll reserve "fee" for referring to charge imposed by the network, in aglos, for a transaction.  I'd prefer using "Cost" or "Budget" here.  Maybe "EnableAppCostPooling".

data/transactions/logic/eval.go
```diff
@@ -255,6 +258,9 @@ func (ep EvalParams) budget() int {
 	if ep.runModeFlags == runModeSignature {
 		return int(ep.Proto.LogicSigMaxCost)
 	}
+	if ep.PooledApplicationBudget > ep.Proto.MaxAppProgramCost {
+		return ep.PooledApplicationBudget
+	}
 	return ep.Proto.MaxAppProgramCost
```
jannotti, line 264: Not sure, but it looks like this is going to mean that a second app call is going to get 700, even if a first app call used 1399.  That is is, the first used up both budgets.

data/transactions/logic/evalStateful_test.go
```diff
@@ -592,20 +592,23 @@ pop
 !=
 `
 	type desc struct {
-		source string
-		eval   func([]byte, EvalParams) (bool, error)
-		check  func([]byte, EvalParams) error
+		source       string
+		eval         func([]byte, EvalParams) (bool, error)
+		evalStateful func([]byte, EvalParams) (int, bool, error)
+		check        func([]byte, EvalParams) error
```
jannotti, line 598: I think I'm going to argue for a shared pointer to a single cost budget counter.  Each ep would point to the same place.  It avoids leaking the cost into the call and returning interface and creating this difference.

ledger/appcow.go
```diff
@@ -485,10 +485,18 @@ func (cb *roundCowState) StatefulEval(params logic.EvalParams, aidx basics.AppIn
...
 
 	// Eval the program
-	pass, err = logic.EvalStateful(program, params)
+	txnCost := 0
+	txnCost, pass, err = logic.EvalStateful(program, params)
 	if err != nil {
 		return false, basics.EvalDelta{}, ledgercore.LogicEvalError{Err: err}
 	}
+	if params.Proto.EnableAppFeePooling {
+		cb.pooledApplicationCosts += txnCost
```
jannotti, line 494: If I'm right about the problem where every app gets at least 700, I think this might catch the problem here.  So that's good.  But I don't think good enough.  We would like to catch the overage on the exact instruction that goes over, and stop.

ledger/eval.go
```diff
@@ -682,20 +683,24 @@ func (eval *BlockEvaluator) prepareEvalParams(txgroup []transactions.SignedTxnWi
...
-			GroupIndex:      i,
-			PastSideEffects: pastSideEffects,
-			MinTealVersion:  &minTealVersion,
+			Txn:                     &groupNoAD[i],
+			Proto:                   &eval.proto,
+			TxnGroup:                groupNoAD,
+			GroupIndex:              i,
+			PastSideEffects:         pastSideEffects,
+			MinTealVersion:          &minTealVersion,
+			PooledApplicationBudget: pooledApplicationBudget,
```
jannotti, line 703: I think your life gets way, way simpler if you just make this a pointer, so you shared &pooledApplicationBudget in all eps.  Then each one decrements as it goes.  If you ever hit zero, you're done then.

Then (at least by line count) your patch gets much much smaller too, because all of the interfaces stay the same.

jannotti (changes requested):
Let's discuss the shared budget pointer idea. I think it fits more closely with how we have done some other recent features. (In fact, see my PR for how I'm handling the excess fee credit in app txn creation: Look for "FeeCredit" in https://github.com/algorand/go-algorand/pull/2661/files)

ledger/eval.go
algochoi: I think I like that suggestion -- I'll take a look at your other comments later, but I'll work on this first.

data/transactions/logic/eval.go
```diff
@@ -255,6 +258,9 @@ func (ep EvalParams) budget() int {
 	if ep.runModeFlags == runModeSignature {
 		return int(ep.Proto.LogicSigMaxCost)
 	}
+	if ep.Proto.EnableAppCostPooling && ep.PooledApplicationBudget != nil {
```
algochoi, line 261: Is the `nil` check necessary here? It fails some tests, e.g. `ledger/apply/application_test.go:TestAppCallApplyCreateOptIn()`, without it. I thought `prepareEvalParams()` was always called before an App call, and if this is the case, can we remove this `nil` check and modify those unit tests?

ledger/eval_test.go
```diff
@@ -562,6 +562,295 @@ func TestEvalAppAllocStateWithTxnGroup(t *testing.T) {
...
+
+	cases := []struct {
+		prog          string
+		expectedError string
+	}{
+		{approvalProgram1, "dynamic cost budget exceeded, executing ed25519verify: remaining budget is 700 but program cost was 1905"},
+		{approvalProgram2, "dynamic cost budget exceeded, executing keccak256: remaining budget is 700 but program cost was 783"},
+	}
+
+	for _, testCase := range cases {
```
algochoi, line 691: I think at one point we wanted to do the protocol checks in `data/transactions/logic/evalStateful_test.go`, but there is some ledger code that also needs to be tested (`prepareEvalParams`) for correct protocol, so I put this test in the ledger tests.

data/transactions/logic/eval.go
```diff
@@ -355,6 +361,10 @@ func EvalStateful(program []byte, params EvalParams) (pass bool, err error) {
 	cx.EvalParams = params
 	cx.runModeFlags = runModeApplication
 	pass, err = eval(program, &cx)
+	if pass && cx.EvalParams.Proto.EnableAppCostPooling {
+		// if eval passes, then budget is always greater than cost, so should not have underflow
+		*cx.EvalParams.PooledApplicationBudget -= cx.cost
+	}
```
jannotti, line 367: If an app doesn't pass, everything's probably going to get aborted. Nonetheless, I don't see a good reason not to decrease the pooled budget.

This also needs to be *super* careful about underflow.  I would use SubSaturate or similar.  Going "negative" here almost certainly leads to the possibility of a the budget being interpreted as a high number elsewhere.

config/consensus.go
algochoi: Agreed - changed to `EnableAppCostPooling `

ledger/eval.go
```diff
@@ -672,11 +672,16 @@ func (eval *BlockEvaluator) prepareEvalParams(txgroup []transactions.SignedTxnWi
 	var groupNoAD []transactions.SignedTxn
 	var pastSideEffects []logic.EvalSideEffects
 	var minTealVersion uint64
+	pooledApplicationBudget := 0
 	res = make([]*logic.EvalParams, len(txgroup))
 	for i, txn := range txgroup {
 		// Ignore any non-ApplicationCall transactions
 		if txn.SignedTxn.Txn.Type != protocol.ApplicationCallTx {
 			continue
+		} else if eval.proto.EnableAppCostPooling {
```
jannotti, line 681: Since you've broken early with `continue`, I'd remove the `else` and start a new `if` here.

data/transactions/logic/eval.go
```diff
@@ -355,6 +361,10 @@ func EvalStateful(program []byte, params EvalParams) (pass bool, err error) {
 	cx.EvalParams = params
 	cx.runModeFlags = runModeApplication
 	pass, err = eval(program, &cx)
+	if pass && cx.EvalParams.Proto.EnableAppCostPooling {
+		// if eval passes, then budget is always greater than cost, so should not have underflow
+		*cx.EvalParams.PooledApplicationBudget -= cx.cost
```
jannotti, line 366: We should be consistent on whether PooledApplicationBudget is sure to be set.  Earlier you checked against nil, here you do not.

algochoi: Good point -- I was wondering if we need to check it at all, but I will add the check here as well.

data/transactions/logic/opcodes.go
```diff
@@ -141,7 +141,7 @@ var OpSpecs = []OpSpec{
 	{0x02, "keccak256", opKeccak256, asmDefault, disDefault, oneBytes, oneBytes, 2, modeAny, costly(130)},
 	{0x03, "sha512_256", opSHA512_256, asmDefault, disDefault, oneBytes, oneBytes, 2, modeAny, costly(45)},
 
-	{0x04, "ed25519verify", opEd25519verify, asmDefault, disDefault, threeBytes, oneInt, 1, runModeSignature, costly(1900)},
+	{0x04, "ed25519verify", opEd25519verify, asmDefault, disDefault, threeBytes, oneInt, 1, modeAny, costly(1900)},
```
algorandskiy, line 144: this is not right, `ed25519verify` must be disallowed in TEAL v2-v4 and allowed only in v5. So instead of removal, a new entry with the same opcode but different version is needed - see `txn` as an example. Plus a test verifying it is off in v2-v4 stateless and is on in v5

data/transactions/logic/evalStateful_test.go
```diff
@@ -678,9 +678,8 @@ pop
 		})
 	}
 
-	// check ed25519verify and arg are not allowed in statefull mode
+	// check that arg is not allowed in stateful mode
 	disallowed := []string{
-		"byte 0x01\nbyte 0x01\nbyte 0x01\ned25519verify",
```
algorandskiy, line 683: this should be checked for TEAL v2-v4.

data/transactions/logic/eval.go
```diff
@@ -216,6 +216,9 @@ type EvalParams struct {
 
 	// determines eval mode: runModeSignature or runModeApplication
 	runModeFlags runMode
+
+	// Total pool of app call budget in a group transaction
+	PooledApplicationBudget *uint64
```
algorandskiy, line 221: Please add a test near `prepareEvalParams` that checks budget value for 2 app calls in a group. That would ensure if someone changes it to a value a red flag will be raised as soon as possible.

data/transactions/logic/eval.go
algorandskiy: yes, please modify the test. It constructs `var ep logic.EvalParams` directly since there is no access to `prepareEvalParams` from `apply` package

ledger/eval_test.go
algorandskiy: these tests look good but I think we should not test everything here because in ledger package we only set pooled budget for multiple app calls in one group. And the decisions are made in logic.
Let's do the following:
1.  a unit test for`prepareEvalParams` call on a group with 2 txns with app calls on old and new protos. No ledger construction/blocks needed.
2. Preserve a high-level test with block evaluator: no pooling on old proto and pooling on new proto. Programs should be simple but valid, something like `"int 1" + strings.Repeat("keccak256", 50)`.
3. Make more precise testing on `ed25519verify` in logic package - it is needed anyway, and the `EvalParams` with proto and budget can be easily set there.

algorandskiy (changes requested):
Looks good, some work needed: 1) `ed25519verify` must be only allowed in stateful TEAL v5+ 2) restructure your TEAL tests in legder: it only should check evaluation on old/new proto with as trivial programs as possible. All TEAL heavy-lifting should go into `logic` package. Especially this change does not depends on ledger interface for TEAL, only `EvalParams` preparation, so this preparation in isolation, and a big picture needed to be tested there.

data/transactions/logic/opcodes.go
algochoi: Good point -- made two versions with v2-v4 with `runModeSignature` and v5 with `modeAny`

data/transactions/logic/evalStateful_test.go
algochoi: Will add a separate test that checks `verify` is not allowed in v4 and lower, and add another check that checks if `arg` is not allowed in v5 and beyond.

ledger/eval_test.go
algochoi: I added the `prepareEvalParams` tests above like you have suggested. Just to be clear, we should probably only check for correct pooling here, and check for cost/budget in the `data/transactions/logic/evalStateful_test.go` file?

algorandskiy: It is always a good idea to check both happy case and error case. The idea is to have 1) minimal sufficient coverage 2) without knowing to much details about stuff from other packeges.
I I suggest to use the following code to generate test programs of difference cost by varying N:
```golang
prog := "byte "a\n" + strings.Repeat("keccak256\n", N) + "pop\nint 1\n";
```
So here in `ledger` the following four cases would be enough:
1. cost ~701 + cost ~100 => FAIL on v29.
2. same combo OK on vFuture
3. ~701 + ~701 => FAIL on both v29 and vFuture

ledger/eval_test.go
```diff
@@ -428,6 +432,36 @@ func TestPrepareEvalParams(t *testing.T) {
 			}
 		})
 	}
+
+	// Test cost pooling on old and new consensus protos
+	for i, testCase := range cases {
+		t.Run(fmt.Sprintf("i=%d", i), func(t *testing.T) {
```
algorandskiy, line 438: this is not must but something you might to consider: there are two almost identical loops.
If fact, it can be done as a single loop over `cases`.
1. Remove fake proto from `eval` construction above
2. Make `params` array consisting on 3 elements: fake params with `Application: true, MaxAppProgramCost: 700`, vFuture, v29.
3. Modify `evalTestCase` struct and add number of app calls in tx group (`numAppCalls`)
4. Run loops over cases, params, and calculate pooled cost as `proto.MaxAppProgramCost * numAppCalls`, and compare against a value EvalParams

algochoi: Good suggestion, will refactor this part.

ledger/eval_test.go
```diff
@@ -562,6 +577,140 @@ func TestEvalAppAllocStateWithTxnGroup(t *testing.T) {
...
+		isSuccessVFuture     bool
+		expectedErrorV29     string
+		expectedErrorVFuture string
+	}{
+		{source(5, 47), true, true,
+			"",
+			""},
+		{source(5, 48), false, true,
+			"pc=157 dynamic cost budget exceeded, executing pushint: remaining budget is 700 but program cost was 701",
+			""},
```
algorandskiy, line 689: where does `pushint` come from? it is not in the program.

algorandskiy (commented):
LGTM, one question about `pushint` in tests.

ledger/eval_test.go
algorandskiy: ah, that's optimizer replaced `int 1` by `pushint 1` - did not know it makes that.

algorandskiy (approved):

ledger/eval_test.go
```diff
@@ -385,48 +390,58 @@ func TestPrepareEvalParams(t *testing.T) {
...
+						require.NotNil(t, res[k])
+						require.NotNil(t, res[k].PastSideEffects)
+						require.Equal(t, res[k].GroupIndex, k)
+						require.Equal(t, res[k].TxnGroup, expGroupNoAD)
+						require.Equal(t, *res[k].Proto, eval.proto)
+						require.Equal(t, *res[k].Txn, testCase.group[k].SignedTxn)
+						if reflect.DeepEqual(param, config.Consensus[protocol.ConsensusV29]) {
+							require.Equal(t, *res[k].PooledApplicationBudget, uint64(eval.proto.MaxAppProgramCost))
+						} else if reflect.DeepEqual(param, config.Consensus[protocol.ConsensusFuture]) {
+							require.Equal(t, *res[k].PooledApplicationBudget, uint64(eval.proto.MaxAppProgramCost*testCase.numAppCalls))
```
jannotti, line 437: An easy bug to introduce (I think) would be if prepareEvalParams put a *different* pointer in each evalParams, for the pooled budget.  Could you test that all of them are the same pointer?

(You could add the same test for minTealVersion if you can't find it anywhere.)

ledger/eval_test.go
jannotti: It's new, @jasonpaulos added it.

PS.  These are great tests.  I will use them as examples that I want to make much simpler to write with my recent testing convenience functions.

jannotti (approved):

PR #2713, opened 2021-08-07
Try setting consensus to new version

onetechnical:

### DIFF
--- a/ledger/apply/application_test.go
+++ b/ledger/apply/application_test.go
@@ -1101,8 +1101,8 @@ func TestAppCallApplyDelete(t *testing.T) {
 	a.Equal(cbr, br)
 	a.Equal(basics.EvalDelta{}, ad.EvalDelta)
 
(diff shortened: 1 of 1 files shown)


ledger/apply/application_test.go
```diff
@@ -1101,8 +1101,8 @@ func TestAppCallApplyDelete(t *testing.T) {
 	a.Equal(cbr, br)
 	a.Equal(basics.EvalDelta{}, ad.EvalDelta)
 
-	// check calculation on ConsensusV28. TotalExtraAppPages does not change
-	b.SetProto(protocol.ConsensusV28)
+	// check calculation on ConsensusV29. TotalExtraAppPages does not change
+	b.SetProto(protocol.ConsensusV29)
```
tsachiherman, line 1105: I'd think that the consensusfuture later on is the one that need to be updated.

onetechnical: This was just a random test, this does need to be V28 because of what it's testing.

PR #2717, opened 2021-08-09
CI: use libboost-math-dev instead of libboost-all-dev

cce:
## Summary

## Test Plan

CI tests should pass as before.


tsachiherman:
take a look on the travis failure. it looks like on arm64 it's now missing `make`. ( which I guess was a dependency )

PR #2720, opened 2021-08-09
using the appIdx instead of txn app id

barnjamin:
## Summary

Passing a foreign app id to the `goal app create` cmd with dryrun tries to use ApplicationId (0) instead of the foreign app id.

### DIFF
--- a/libgoal/libgoal.go
+++ b/libgoal/libgoal.go
@@ -1035,7 +1035,7 @@ func MakeDryrunStateGenerated(client Client, txnOrStxn interface{}, other []tran
 				} else {
 					// otherwise need to fetch app state
 					var app generatedV2.Application
-					if app, err = client.ApplicationInformation(uint64(tx.ApplicationID)); err != nil {
+					if app, err = client.ApplicationInformation(uint64(appIdx)); err != nil {
 						return
 					}
 					appParams = app.Params


nullun:
Can confirm this fixed a bug I was facing when supplying a foreign app during a `goal app create` call.

Thank you @barnjamin!

algorandskiy (approved):
Thank you for the fix it looks obviously correct. Ideally is to have a test but it is optional

barnjamin:
I can write a test but I'll need a little hand holding, not sure how to set it up correctly

algorandskiy:
The test would be the following:
1. Create a `transactions.Transaction` object with type `protocol.ApplicationCallTx` with the following `ApplicationCallTxnFields` fields set:
  - ApprovalProgram
  - ForeignApps
2. call `MakeDryrunStateGenerated(cl, txn, []transactions.SignedTxn{}, protocol.ConsensusFuture)`
3. Iterate over `dr.Apps` (return value) and check a value from `ForeignApps` presents.

The only problem is in providing `Client` instance that would require a node to make it work and automatically makes it a slow integration test, and not worth it. Introducing a client interface would solve the problem but again, it is too much of coding for such a trivial fix.

barnjamin:
I appreciate the detailed explanation but sounds like not worth it?

PR #2721, opened 2021-08-09
testing: update crypto library to use partitiontest

tsachiherman:
## Summary

update crypto library to use partitiontest

## Test Plan

These are tests.


algoidan (commented):
looks good!

algoidan (approved):

PR #2722, opened 2021-08-10
txnsync: change iterate

tsachiherman:
## Summary

Change the iterate method

## Test Plan

Use existing unit tests.

### DIFF
--- a/txnsync/bitmask.go
+++ b/txnsync/bitmask.go
@@ -164,13 +164,14 @@ func (b *bitmask) iterate(entries int, maxIndex int, callback func(int, int) err
 		}
 	case 2:
 		sum := 0
-		for index := 0; index*2+2 < len(*b); index++ {
+		elementsCount := (len(*b) - 1) / 2
+		if elementsCount > maxIndex {
+			return errDataMissing
+		}
+		for index := 0; index < elementsCount; index++ {
 			sum += int((*b)[index*2+1])*256 + int((*b)[index*2+2])
 			if sum >= entries {
 				return errIndexNotFound
-			}
-			if index >= maxIndex {
-				return errDataMissing
 			}
 			if err := callback(sum, index); err != nil {
 				return err
--- a/txnsync/bitmask.go
+++ b/txnsync/bitmask.go
@@ -170,7 +170,7 @@ func (b *bitmask) iterate(entries int, maxIndex int, callback func(int, int) err
 		}
 		for index := 0; index < elementsCount; index++ {
 			sum += int((*b)[index*2+1])*256 + int((*b)[index*2+2])
-			if sum >= entries {
+			if sum > entries {
 				return errIndexNotFound
 			}
 			if err := callback(sum, index); err != nil {
--- a/txnsync/bitmask.go
+++ b/txnsync/bitmask.go
@@ -113,96 +113,132 @@ func (b *bitmask) trimBitmask(entries int) {
 }
 
 // iterate through the elements of bitmask without expanding it.
-// If more than maxIndex set bits are found, return error. For each set value, call callback.
-func (b *bitmask) iterate(entries int, maxIndex int, callback func(int, int) error) error {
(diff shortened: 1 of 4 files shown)


txnsync/bitmask.go
```diff
@@ -164,14 +164,15 @@ func (b *bitmask) iterate(entries int, maxIndex int, callback func(int, int) err
 		}
 	case 2:
 		sum := 0
-		for index := 0; index*2+2 < len(*b); index++ {
+		elementsCount := (len(*b) - 1) / 2
+		if elementsCount > maxIndex {
```
algonautshant, line 168: This can also be done for case 3, since it is the complement of case 2.

algonautshant (approved):
This is correct and nice. And the same can be done for case 3. 
This justifies checking the maxIndex inside iterate.

txnsync/bitmask.go
```diff
@@ -164,14 +164,15 @@ func (b *bitmask) iterate(entries int, maxIndex int, callback func(int, int) err
...
+			return errDataMissing
+		}
+		for index := 0; index < elementsCount; index++ {
 			sum += int((*b)[index*2+1])*256 + int((*b)[index*2+2])
-			if sum >= entries {
+			if sum > entries {
```
algonautshant, line 173: I think this is not correct. 
Look at txnsync/encodedgroupsunmarshalers.go
`entries` is `int(stub.TotalTransactionsCount)`
and the `sum` passed to `callback` gets to `i`:
`signedTxns[i].Msig.Version`

txnsync/bitmask_test.go
```diff
@@ -145,7 +147,9 @@ func trimIterateHelper(t *testing.T, setBits []int) {
 
 	// For types 2, let the sum exceed entries
 	if int((b)[0]) == 2 {
-		require.Equal(t, errIndexNotFound, b.iterate(setBits[len(setBits)-1], len(setBits)-1, iterfunc))
```
algonautshant, line 148: This should be an error according to how iterate is used in encodedgroupsunmarshalers.go

txnsync/bitmask_test.go
```diff
@@ -168,3 +172,47 @@ func trimIterateHelper(t *testing.T, setBits []int) {
...
+		var b bitmask
+		b = make([]byte, blen)
+		b[0] = byte(bitmaskType)
+		for i := 1; i < blen; i++ {
+			b[i] = rand()
+		}
+		entries := int(rand())
+		maxIndex := int(rand())
+		lastEntryIndex := -1
+		b.iterate(entries, maxIndex, func(i, j int) error {
```
algonautshant, line 200: Testing the return of this is important. 
Code is getting covered because of TestFuzzBitmask, but the covered code is not verified by the test.

txnsync/bitmask_test.go
```diff
@@ -145,7 +147,9 @@ func trimIterateHelper(t *testing.T, setBits []int) {
 
 	// For types 2, let the sum exceed entries
 	if int((b)[0]) == 2 {
-		require.Equal(t, errIndexNotFound, b.iterate(setBits[len(setBits)-1], len(setBits)-1, iterfunc))
+		require.Equal(t, errIndexNotFound, b.iterate(setBits[len(setBits)-1]-1, len(setBits), iterfunc))
+		require.Nil(t, b.iterate(setBits[len(setBits)-1], len(setBits), iterfunc))
+		require.Equal(t, errDataMissing, b.iterate(setBits[len(setBits)-1], len(setBits)-1, iterfunc))
```
algonautshant, line 152: This case is the same as the case on line 146:
`	require.Equal(t, errDataMissing, b.iterate(entries, len(setBits)-1, iterfunc))
`

txnsync/bitmask.go
nicholasguoalgorand: I agree with Shant. If sum = entries, we will be trying to access an index that is out of bounds.

PR #2730, opened 2021-08-11
Convenient and High-fidelity Transaction Processing Tests

jannotti:
This refactors some of the `ledger` code so that it's convenient
to write short tests that operate on a fairly complete ledger,
including rewards payouts.  Previously, tests used genesis(), but
genesis() build the genesis block internally "by hand" rather
than using MakeGenesisBlock, so it missed some details (like
setting up RewardsState). Presumably, this was because
MakeGenesisBlock was in the `data` package, and could not be
imported.  That is the motivation behind moving it, and some
related code, to bookkeeping (where various Genesis related code
already existed).

The txntest packaged is motivated purely by a desire for more
concise tests.  It allows for the construction of
transaction.Transaction objects concisely, and we can add all
sort of conveneince routines here that would not make sense in
the production code (turning these into SignedTxns,
SignedTxnWithADs, TransactionGroups, etc).


ledger/eval_test.go
```diff
@@ -1030,6 +1031,210 @@ func TestModifiedAssetHoldings(t *testing.T) {
...
+	l, err := OpenLedger(logging.Base(), dbName, inMem, InitState{
+		Block:       genBlock,
+		Accounts:    balances.Balances,
+		GenesisHash: genHash,
+	}, cfg)
+	require.NoError(t, err)
+	return l
+}
+
+func (ledger *Ledger) nextEval(t testing.TB) *BlockEvaluator {
```
algorandskiy, line 1100: nextEval / endBlock need some comments about intended use. Name it "newEval"?

data/ledger.go
```diff
@@ -76,83 +75,35 @@ type roundSeed struct {
 	elements [2]roundSeedPair
 }
 
-func makeGenesisBlock(proto protocol.ConsensusVersion, genesisBal GenesisBalances, genesisID string, genesisHash crypto.Digest) (bookkeeping.Block, error) {
```
algorandskiy, line 79: why moving into bookkeeping package instead of exporting from data?

data/txntest/txn.go
```diff
@@ -0,0 +1,183 @@
...
+	"github.com/algorand/go-algorand/crypto"
+	"github.com/algorand/go-algorand/crypto/compactcert"
+	"github.com/algorand/go-algorand/data/basics"
+	"github.com/algorand/go-algorand/data/transactions"
+	"github.com/algorand/go-algorand/protocol"
+)
+
+// Txn exists purely to make it easier to write a
+// transaction.Transaction in Go source.
+type Txn struct {
```
algorandskiy, line 46: Not sure I'm a big fan of one more Txn type (in addition to Transaction, Signed, SignedTxnWithAD, ...)

algorandskiy (commented):
Shouldn't spend some time after merging it and unify some tests to new functions?

algochoi (commented):
Like the change! I think it will make testing transactions much simpler

data/ledger.go
jannotti: Because data imports ledger (because it has a type data.Ledger that wraps ledger.Ledger!) .  So ledger can't import data.

data/txntest/txn.go
jannotti: The goal here is to have one concise type (you need write substructure, since all fields are at top-level) and then write conversion functions like SignedTxn(), Txn(), even TxGroup() that do all the conversions to give you those types in a handy way when writing tests.  This type should never appear outside of _test.go files.

ledger/eval_test.go
```diff
@@ -1030,6 +1031,210 @@ func TestModifiedAssetHoldings(t *testing.T) {
...
+}
+
+func (ledger *Ledger) lookup(t testing.TB, addr basics.Address) basics.AccountData {
+	rnd := ledger.Latest()
+	ad, err := ledger.Lookup(rnd, addr)
+	require.NoError(t, err)
+	return ad
+}
+
+func (eval *BlockEvaluator) txn(t testing.TB, txn *txntest.Txn) {
```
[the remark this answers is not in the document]
jannotti: It doesn't sign them.  It's a big lie.

I might not object to testTxn, except there is already testTransaction().  testTransaction() does a bunch of tests, but does not actually push it through the evaluator.

ledger/eval_test.go
jannotti: I prefer `next` because it implies the key point that this is the "next" evaluator, a new block on your test chain.  I will add commentary.

ledger/eval_test.go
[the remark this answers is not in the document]
algochoi: I see, just read the comment above about the unsigned `SignedTxn()`. Maybe something like `prepareAndTestTxn`? If not, we can leave it

data/txntest/txn.go
jannotti: As some evidence, I converted two tests, they were about 130 lines each, now about 65 each.

algorandskiy (approved):
Looks good to me. Let's someone from other teams be familiar with new test harness.

algorandskiy:
@tolikzinovyev mind checking how does `TestModifiedAssetHoldings` look with new testing api?

PR #2732, opened 2021-08-11
Make config.MaxLogCalls dependant on some consensus param

algorandskiy:
## Summary

* Commented some constants in teal evaluator and added a test
  enforcing their values and directing on how to change them if/when needed.

## Test Plan

Existing tests

### DIFF
--- a/config/consensus.go
+++ b/config/consensus.go
@@ -423,8 +423,8 @@ var MaxEvalDeltaAccounts int
 var MaxStateDeltaKeys int
 
 // MaxLogCalls is the highest allowable log messages that may appear in
-// any version, used for decoding purposes. Never decrease this value.
-const MaxLogCalls = 32
+// any version, used only for decoding purposes. Never decrease this value.
+var MaxLogCalls int
 
 // MaxLogicSigMaxSize is the largest logical signature appear in any of the supported
 // protocols, used for decoding purposes.
@@ -486,6 +486,9 @@ func checkSetAllocBounds(p ConsensusParams) {
 	checkSetMax(p.MaxExtraAppProgramPages, &MaxExtraAppProgramLen)
 	// MaxAvailableAppProgramLen is the max of supported app program size
 	MaxAvailableAppProgramLen = MaxAppProgramLen * (1 + MaxExtraAppProgramLen)
+	// There is no consensus parameter for MaxLogCalls and MaxAppProgramLen as an approximation
+	// Its value is much larger than any possible reasonable MaxLogCalls value in future
+	checkSetMax(p.MaxAppProgramLen, &MaxLogCalls)
 }
 
 // SaveConfigurableConsensus saves the configurable protocols file to the provided data directory.
--- a/data/transactions/logic/eval.go
+++ b/data/transactions/logic/eval.go
@@ -44,8 +44,8 @@ import (
 // EvalMaxVersion is the max version we can interpret and run
 const EvalMaxVersion = LogicVersion
 
-// EvalMaxScratchSize is the maximum number of scratch slots.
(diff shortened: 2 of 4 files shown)


data/basics/teal.go
```diff
@@ -23,6 +23,10 @@ import (
 	"github.com/algorand/go-algorand/config"
 )
 
+// MaxLogCalls is the highest allowable log messages that may appear in
+// any version, used only for decoding purposes.
+const MaxEncodedTealLogCalls = 32
```
shiqizng, line 28: should you note that this value should never be decreased?

algorandskiy: Right, done!

config/consensus.go
```diff
@@ -422,10 +422,6 @@ var MaxEvalDeltaAccounts int
 // in a StateDelta, used for decoding purposes.
 var MaxStateDeltaKeys int
 
-// MaxLogCalls is the highest allowable log messages that may appear in
-// any version, used for decoding purposes. Never decrease this value.
-const MaxLogCalls = 32
-
 // MaxLogicSigMaxSize is the largest logical signature appear in any of the supported
 // protocols, used for decoding purposes.
 var MaxLogicSigMaxSize int
```
jannotti, line 427: I don't want to move this one variable out when all the nearby values are of the same variety - they are not consensus parameters, rather they are conservative esitmates of the largest that some encoded message part can be.  It's not just this variable that is like that.

algorandskiy: The problem with other values they are derived from consensus param, and this one is not because `MaxLogCalls` not in consensus. In fact it should not be even called `Max...`

data/transactions/logic/eval.go
```diff
@@ -57,7 +57,7 @@ const MaxByteMathSize = 64
 const MaxLogSize = 1024
 
 // MaxLogCalls is the limit of total log calls during a program execution
-const MaxLogCalls = config.MaxLogCalls
+const MaxLogCalls = 32
```
jannotti, line 60: By not using the same variable, the MaxEncodedTealLogCalls becomes completely "invisible". A reasonable person might make it so that teal v6 allows more logs calls, but preserves compatibility for v5 and below.  They would not even know there is a place where we put alloc bounds on it.

Here's an entirely different proposal that I think is simple, correct, and the same as what we do elsewhere.  Notice how we set the bounds on deltas:
```
	// These bounds could be tighter, but since these values are just to
	// prevent DoS, setting them to be the maximum number of allowed
	// executed TEAL instructions should be fine (order of ~1000)
	checkSetMax(p.MaxAppProgramLen, &MaxStateDeltaKeys)
	checkSetMax(p.MaxAppProgramLen, &MaxEvalDeltaAccounts)
	checkSetMax(p.MaxAppProgramLen, &MaxAppProgramLen)
```

This is wildly bigger than it needs to be, but we're only trying to avoid crazy huge DoS messages.  So let's just set the bound on MaxLogCalls the same way.  Add:
`checkSetMax(p.MaxAppProgramLen, &MaxLogCalls)`

Now it's derived from consensus and doesn't need to be revisited.

Although we have *another* problem - nowadays, I suppose these bounds are not as obviously correct as they used to be.  You could loop, setting many keys, perhaps more that 1000.  We could use the MaxAppCost, rather than MaxAppLen though, since that's what would stop us. (Though it would need to be MaxGroupSize*MaxAppCost. because of pooling)

algorandskiy: Looks reasonable, done

jannotti (approved):
Seems good, glad we found middle ground.

shiqizng (approved):
look good to me

PR #2741, opened 2021-08-13
agreement: handle pseudonode enqueueing failures

tsachiherman:
## Summary

This PR addresses two separate issues in the `pseudonodeNode` implementation:
1. `pseudonodeVotesTask.execute` and `pseudonodeProposalsTask.execute ` do not handle vote verification enqueueing failures ( to the execution pool ). This could lead the the pseudonode processing go-routine being stuck.
2. The `pseudonodeVotesTask.execute` could block forever in case the output channel is not being read from. 

## Trigger
This issue was detected by the telemetry:
```
 ({num-occurances}): {telemetry error string}
 (67): pseudonode.MakeVotes call failed(attest) pseudonode input channel is full
```

## Test Plan

Few unit tests added. More to come.


agreement/asyncVoteVerifier.go
```diff
@@ -140,16 +140,18 @@ func (avv *AsyncVoteVerifier) verifyVote(verctx context.Context, l LedgerReader,
 		// if we're done while waiting for room in the requests channel, don't queue the request
 		req := asyncVerifyVoteRequest{ctx: verctx, l: l, uv: &uv, index: index, message: message, out: out}
 		avv.wg.Add(1)
-		if avv.backlogExecPool.EnqueueBacklog(avv.ctx, avv.executeVoteVerification, req, avv.execpoolOut) != nil {
+		if err := avv.backlogExecPool.EnqueueBacklog(avv.ctx, avv.executeVoteVerification, req, avv.execpoolOut); err != nil {
```
cce, line 143: Interestingly, inside of EnqueueBacklog it is also doing a select on `case <-avv.ctx.Done()` (passed in as enqueueCtx) which will cause EnqueueBacklog to return ctx.Err(). If you have two pending selects like that on the stack, is it deterministic which one will fire first?

cce: There is this comment warning "instead, enqueue so the worker will set the error value and return the cancelled vote properly." However if the waiting inside EnqueueBacklog() is interrupted by one of the Done()s it's waiting on, it will never reach the worker.

agreement/cryptoVerifier.go
```diff
@@ -205,7 +205,14 @@ func (c *poolCryptoVerifier) voteFillWorker(toBundleWait chan<- bundleFuture) {
 			}
 
 			uv := votereq.message.UnauthenticatedVote
-			c.voteVerifier.verifyVote(votereq.ctx, c.ledger, uv, votereq.TaskIndex, votereq.message, c.votes.out)
+			err := c.voteVerifier.verifyVote(votereq.ctx, c.ledger, uv, votereq.TaskIndex, votereq.message, c.votes.out)
+			if err != nil && c.votes.out != nil {
+				select {
+				case c.votes.out <- asyncVerifyVoteResponse{index: votereq.TaskIndex, err: err}:
```
cce, line 211: Looking deeper, this changes the behavior from what happens in AsyncVoteVerifier.executeVoteVerification. Today the only way verifyVote() and EnqueueBacklog() can return errors is if they were canceled, and there is code in `executeVoteVerification` to set cancelled: true in that case. Should this be filled out?

Or maybe no new response should be sent to c.votes.out? It seems like this was the old behavior — there were already cases where verifyVote and EnqueueBacklog might get select-exited, and then verifyVote would just silently drop it, and it would never show up on the cryptoVerifier.VerifiedVotes() channel in demux.

agreement/cryptoVerifier.go
```diff
@@ -205,7 +205,14 @@ func (c *poolCryptoVerifier) voteFillWorker(toBundleWait chan<- bundleFuture) {
...
+				default:
+					c.log.Infof("poolCryptoVerifier.voteFillWorker unable to write failed enqueue response to output channel")
+				}
```
cce, line 214: It seems like with this non-blocking send, and the fact that demux.go is listening on c.votes.out in a big select on lots of channels (on d.crypto.VerifiedVotes()) , you will almost always end up in this default case. Is that right? If so maybe not trying to introduce a new type of error response at all is best?

agreement/cryptoVerifier.go
cce: Running down what happens when asyncVerifyVoteResponse.cancelled be set — AFAICT all the "if e.Cancelled" code paths result in filtering the event the same way as if e.Err was set.

agreement/cryptoVerifier_test.go
```diff
@@ -385,3 +387,24 @@ func BenchmarkCryptoVerifierBundleVertification(b *testing.B) {
...
+	defer voteVerifier.Quit()
+
+	cryptoVerifier := makeCryptoVerifier(nil, nil, voteVerifier, logging.TestingLog(t))
+	defer cryptoVerifier.Quit()
+
+	cryptoVerifier.VerifyVote(context.Background(), cryptoVoteRequest{message: message{Tag: protocol.AgreementVoteTag}, Round: basics.Round(8)})
+	// read the failed response from VerifiedVotes:
+	votesout := cryptoVerifier.VerifiedVotes()
+	voteResponse := <-votesout
+	require.Equal(t, context.Canceled, voteResponse.err)
```
cce, line 409: Arguably you would want voteResponse.cancelled = true to match a context.Canceled err.

cce: (or maybe preserving current behavior of no response at all is fine, as I mentioned earlier)

agreement/pseudonode.go
```diff
@@ -438,15 +447,42 @@ func (t pseudonodeVotesTask) execute(verifier *AsyncVoteVerifier, quit chan stru
...
 		case <-quit:
 			return
 		case <-t.context.Done():
 			// we done care about the output anymore; just exit.
 			return
+		default:
+		}
+
+		if outputTimeout == nil && !timedOut {
+			outputTimeout = time.After(maxPseudonodeOutputWaitDuration)
```
cce, line 468: So this is to save the extra resources required to make a timeout channel per iteration, is that right? I wonder how often you are going to end up breaking out into the default and retry ... it seems like you might end up here most of the time, but hard to say.

agreement/asyncVoteVerifier.go
tsachiherman: The fact that we have two pending selects isn't an issue. The one in `verifyVote` would be evaluated when we get to that point, and if it's not canceled yet, it would go to the `default` statement, executing `EnqueueBacklog`, where it would also block on the same channel.

As for the comment, I think it's a bug - I think that the intent was to have a `fallthrough` here. But I would leave that to a separate PR, as it's not really related to this change. ( i.e. in out case, we're passing a TODO context, which would never expire and therefore this statement would never be executed )

agreement/cryptoVerifier.go
tsachiherman: Correct - all the `e.Cancelled` would result in ignoring this event. However, this is not needed in case the reason for exiting is due to a system shutdown, right ?
( i.e. the agreement state machine is going down; it won't be processing this event ).

On the flip side, we could set the Cancelled flag. It wouldn't be wrong, since this is the "difference" between malicious vs. non-malicious vote verification.

agreement/cryptoVerifier.go
tsachiherman: This fail case is quite rare: the `c.voteVerifier.verifyVote` would normally return right away after enqueuing the vote verification.

The user of the `cryptoVerifier` expects the crypto verifier to return a response on the `c.votes.out` channel. If the `c.voteVerifier.verifyVote` doesn't return it, the vote "disappear". The change here makes sure that the vote won't disappear without leaving any trace.

agreement/cryptoVerifier.go
tsachiherman: thinking about that further, I think that you're correct. we should retain the same behavior we have in `executeVoteVerification`.

cce: I also just noticed that some of the existing code paths that set err & cancelled also skip req.index (but they all do provide req, which you are omitting here)
```
agreement/asyncVoteVerifier.go
103:            return &asyncVerifyVoteResponse{err: req.ctx.Err(), cancelled: true, req: &req}
112:            return &asyncVerifyVoteResponse{v: v, index: req.index, message: req.message, err: err, cancelled: cancelled, req: &req}
122:            return &asyncVerifyVoteResponse{err: req.ctx.Err(), cancelled: true, req: &req}
130:            return &asyncVerifyVoteResponse{ev: ev, index: req.index, message: req.message, err: err, cancelled: cancelled, req: &req}
`

File: agreement/cryptoVerifier.go [hunk 6]
```diff
@@ -205,7 +205,14 @@ func (c *poolCryptoVerifier) voteFillWorker(toBundleWait chan<- bundleFuture) {
 			}
 
 			uv := votereq.message.UnauthenticatedVote
-			c.voteVerifier.verifyVote(votereq.ctx, c.ledger, uv, votereq.TaskIndex, votereq.message, c.votes.out)
+			err := c.voteVerifier.verifyVote(votereq.ctx, c.ledger, uv, votereq.TaskIndex, votereq.message, c.votes.out)
+			if err != nil && c.votes.out != nil {
+				select {
+				case c.votes.out <- asyncVerifyVoteResponse{index: votereq.TaskIndex, err: err, cancelled: true}:
```
cce, line 211: AFAICT all the cases where asyncVerifyVoteResponse structs are currently made also set the `req` field.

agreement/pseudonode.go
tsachiherman: It's timing-dependent, but it would allocate a timer only once.

agreement/pseudonode.go
```diff
@@ -364,13 +366,20 @@ func (t pseudonodeVotesTask) execute(verifier *AsyncVoteVerifier, quit chan stru
...
+			asyncVerifyingVotes--
+			continue
+		}
 	}
 
-	orderedResults := make([]asyncVerifyVoteResponse, len(unverifiedVotes))
-	for i := 0; i < len(unverifiedVotes); i++ {
+	for i := 0; i < asyncVerifyingVotes; i++ {
 		resp := <-results
 		orderedResults[resp.index] = resp
```
cce, line 384: resp.index is not always set by the current code — I guess this is a bug in the existing implementation that we wouldn't really have noticed?

agreement/cryptoVerifier.go
tsachiherman: yes, but it's used only by the `asyncVoteVerifier` as a way to pass down the output channel.

agreement/pseudonode.go
tsachiherman: It's not an issue, since it's needed only in case there is no error for that vote.
( this is what you meant, right ? )

tsachiherman: no, you're correct. We should fix it. It will not work correctly without that.

agreement/pseudonode.go
cce: I was just thinking, maybe it is not worth the added complexity / code duplication if you are almost always going to end up calling time.After() on every iteration. But without evidence that is true I wouldn't know

agreement/asyncVoteVerifier.go
cce: there are other callers (unauthenticatedBundle.verify and cryptoVerifier.voteFillWorker) passing a valid context into verifyVote — just not pseudonode

agreement/pseudonode.go
```diff
@@ -31,13 +33,15 @@ import (
...
+	pseudonodeVerificationBacklog   = 32
+	maxPseudonodeOutputWaitDuration = 2 * time.Second
 )
 
-var errPseudonodeBacklogFull = fmt.Errorf("pseudonode input channel is full")
-var errPseudonodeVerifierClosedChannel = fmt.Errorf("crypto verifier closed the output channel prematurely")
-var errPseudonodeNoVotes = fmt.Errorf("no valid participation keys to generate votes for given round")
-var errPseudonodeNoProposals = fmt.Errorf("no valid participation keys to generate proposals for given round")
+var errPseudonodeBacklogFullVote = "pseudonode input channel is full; unable to make vote for (%d, %d, %d)"
+var errPseudonodeBacklogFullProposal = "pseudonode input channel is full; unable to make a proposal for (%d, %d)"
```
cce, line 41:
```suggestion
var errPseudonodeBacklogFull = fmt.Errorf("pseudonode input channel is full")
```

agreement/pseudonode.go
```diff
@@ -172,7 +176,7 @@ func (n asyncPseudonode) MakeProposals(ctx context.Context, r round, p period) (
 		return proposalTask.outputChannel(), nil
 	default:
 		proposalTask.close()
-		return nil, errPseudonodeBacklogFull
+		return nil, fmt.Errorf(errPseudonodeBacklogFullProposal, r, p)
```
cce, line 179:
```suggestion
		return nil, fmt.Errorf("unable to make proposal for (%d, %d): %w", r, p, errPseudonodeBacklogFull)
```

agreement/pseudonode.go
```diff
@@ -189,7 +193,7 @@ func (n asyncPseudonode) MakeVotes(ctx context.Context, r round, p period, s ste
 		return proposalTask.outputChannel(), nil
 	default:
 		proposalTask.close()
-		return nil, errPseudonodeBacklogFull
+		return nil, fmt.Errorf(errPseudonodeBacklogFullVote, r, p, s)
```
cce, line 196:
```suggestion
		return nil, fmt.Errorf("unable to make vote for (%d, %d, %d): %w", r, p, s, errPseudonodeBacklogFull)
```

agreement/asyncVoteVerifier.go
tsachiherman: yes, I'm aware of that - that's why I did not attempted to change this code ;-)

agreement/cryptoVerifier.go
cce: Right, but if no one is waiting at the right time, the vote will disappear due to the non-blocking send, right? The only thing waiting on c.votes.out is demux, so I was just predicting that pretty often it will miss the chance to receive this anyway.

cce (changes requested):

agreement/pseudonode_test.go
```diff
@@ -452,3 +454,94 @@ func TestPseudonodeLoadingOfParticipationKeys(t *testing.T) {
...
+	var err error
+	for i := 0; i < pseudonodeVerificationBacklog*2; i++ {
+		ch, err = pb.MakeProposals(context.Background(), startRound, period(i))
+		if err != nil {
+			require.Subset(t, []int{pseudonodeVerificationBacklog, pseudonodeVerificationBacklog + 1}, []int{i})
+			break
+		}
+		channels = append(channels, ch)
+	}
+	require.Error(t, err, "MakeProposals did not returned an error when being overflowed with requests")
```
cce, line 527: in testify 1.7.0 we can do
```
	require.ErrorIs(t, err, errPseudonodeBacklogFull, "MakeProposals did not returned an error when being overflowed with requests")
```

agreement/cryptoVerifier.go
tsachiherman: it shouldn't. The output buffer is sized so it won't lose votes.

cce: oh, I see, I didn't realize out was a buffered channel. OK!

cce (approved):
LGTM. Ideally we should only see these new log messages and errors around shutdown time or if something is broken.

tsachiherman:
> LGTM. Ideally we should only see these new log messages and errors around shutdown time or if something is broken.

yes, that's what I'm hoping for.

PR #2742, opened 2021-08-13
Fix TestRekeyUpgrade test

algorandskiy:
## Summary

* The test checks rekeying before and after upgrade but nodes running
  as separate processes might upgrade earlier than the test advances
* This fix takes this into account similarly to the app upgrade tests

## Test Plan

This is a test

### DIFF
--- a/test/e2e-go/upgrades/rekey_support_test.go
+++ b/test/e2e-go/upgrades/rekey_support_test.go
@@ -68,6 +68,9 @@ func TestRekeyUpgrade(t *testing.T) {
 	round := curStatus.LastRound
 
 	// no consensus upgrade took place (yet)
+	// in fact on slow environment it might happen faster than the test advances.
+	// that's why errors from BroadcastTransaction are checked for exact specific errors
+	// rather than simply "rekeying not yet enable" or "nonempty AuthAddr"
 
 	// Ensure no rekeying happened
 	ad, err := client.AccountData(accountA)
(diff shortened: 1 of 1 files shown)


test/e2e-go/upgrades/rekey_support_test.go
```diff
@@ -79,35 +82,49 @@ func TestRekeyUpgrade(t *testing.T) {
...
+		// should be either "transaction has RekeyTo set but rekeying not yet enable" or "txn dead"
+		if !strings.Contains(err.Error(), "transaction has RekeyTo set but rekeying not yet enable") &&
+			!strings.Contains(err.Error(), "txn dead") {
+			a.NoErrorf(err, "error message should be one of :\n%s\n%s", "transaction has RekeyTo set but rekeying not yet enable", "txn dead")
+		}
+	} else {
+		// if we had no error it must mean that we've upgraded already. Verify that.
+		curStatus, err := client.Status()
+		a.NoError(err)
+		a.NotEqual(consensusTestUnupgradedProtocol, protocol.ConsensusVersion(curStatus.LastVersion))
```
tsachiherman, line 100: If this is true, the subsequent test should be skipped, and you could skip ahead to the point where we don't need to wait for the upgrade.

algorandskiy: I agree but I found keeping the test logic sequential is better for readability than deep nested logic or goto.

tsachiherman (approved):

PR #2746, opened 2021-08-16
Dryrun return cost

AyAggarwal:
This PR adds a `Cost` Parameter to DryrunResponse. It is calculated by giving the evalParams object in dryrun.go the maximum budget allowed (maxappcost * maxgroupsize) and checking the difference after stateful evaluation. LogicSig dryruns will return cost of 0. 

Added a single test with sqrt and default opcodes

### DIFF
--- a/daemon/algod/api/server/v2/dryrun.go
+++ b/daemon/algod/api/server/v2/dryrun.go
@@ -397,17 +397,21 @@ func doDryrunRequest(dr *DryrunRequest, response *generated.DryrunResponse) {
 		return
 	}
 	proto := config.Consensus[protocol.ConsensusVersion(dr.ProtocolVersion)]
-
+	proto.EnableAppCostPooling = true
 	response.Txns = make([]generated.DryrunTxnResult, len(dr.Txns))
 	for ti, stxn := range dr.Txns {
 		pse := logic.MakePastSideEffects(len(dr.Txns))
+		maxBudget := uint64(16*700)
+		evalBudget := uint64(16*700)
 		ep := logic.EvalParams{
-			Txn:             &stxn,
-			Proto:           &proto,
-			TxnGroup:        dr.Txns,
-			GroupIndex:      ti,
-			PastSideEffects: pse,
+			Txn:                     &stxn,
+			Proto:                   &proto,
+			TxnGroup:                dr.Txns,
+			GroupIndex:              ti,
+			PastSideEffects:         pse,
+			PooledApplicationBudget: &evalBudget,
 		}
+		response.Cost = 0 //LogicSig dryrun will not return cost
 		var result generated.DryrunTxnResult
 		if len(stxn.Lsig.Logic) > 0 {
 			var debug dryrunDebugReceiver
@@ -521,7 +525,7 @@ func doDryrunRequest(dr *DryrunRequest, response *generated.DryrunResponse) {
 					}
 					result.LocalDeltas = &localDeltas
 				}
-
+				response.Cost = maxBudget - evalBudget
 				var err3 error
 				result.Logs, err3 = DeltaLogToLog(delta.Logs, appIdx)
 				if err3 != nil {
--- a/daemon/algod/api/server/v2/dryrun.go
+++ b/daemon/algod/api/server/v2/dryrun.go
@@ -398,11 +398,12 @@ func doDryrunRequest(dr *DryrunRequest, response *generated.DryrunResponse) {
 	}
 	proto := config.Consensus[protocol.ConsensusVersion(dr.ProtocolVersion)]
 	proto.EnableAppCostPooling = true
+	proto.MaxTxGroupSize = 16
 	response.Txns = make([]generated.DryrunTxnResult, len(dr.Txns))
 	for ti, stxn := range dr.Txns {
 		pse := logic.MakePastSideEffects(len(dr.Txns))
-		maxBudget := uint64(16*700)
-		evalBudget := uint64(16*700)
+		maxBudget := uint64(proto.MaxAppProgramCost * proto.MaxTxGroupSize)
+		evalBudget := uint64(proto.MaxAppProgramCost * proto.MaxTxGroupSize)
 		ep := logic.EvalParams{
 			Txn:                     &stxn,
 			Proto:                   &proto,
--- a/daemon/algod/api/server/v2/dryrun_test.go
+++ b/daemon/algod/api/server/v2/dryrun_test.go
@@ -124,6 +124,8 @@ func logResponse(t *testing.T, response *generated.DryrunResponse) {
 			}
 		}
 	}
+	t.Log("App Cost:")
+	t.Log(response.Cost)
 }
 
 var dryrunProtoVersion protocol.ConsensusVersion = "dryrunTestProto"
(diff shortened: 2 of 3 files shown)


jannotti:
Seems like there's a gofmt error.

Make sure you have a test that uses more than 700 units.

I don't think cost should be returned at all in the json for logic sigs. That seems misleading. (It's possible it isn't, as we often omit 0s in responses.)

daemon/algod/api/server/v2/dryrun.go
```diff
@@ -397,17 +397,22 @@ func doDryrunRequest(dr *DryrunRequest, response *generated.DryrunResponse) {
 		return
 	}
 	proto := config.Consensus[protocol.ConsensusVersion(dr.ProtocolVersion)]
```
algorandskiy, line 399: shouldn't protocol version be set by a caller?

daemon/algod/api/server/v2/dryrun_test.go
```diff
@@ -124,6 +124,8 @@ func logResponse(t *testing.T, response *generated.DryrunResponse) {
 			}
 		}
 	}
+	t.Log("App Cost:")
+	t.Log(response.Cost)
```
algorandskiy, line 128: not needed debug output

daemon/algod/api/server/v2/dryrun.go
algorandskiy: check `func (v2 *Handlers) TealDryrun(ctx echo.Context) error`:
``` golang
	var protocolVersion protocol.ConsensusVersion
	if dr.ProtocolVersion != "" {
		var ok bool
		_, ok = config.Consensus[protocol.ConsensusVersion(dr.ProtocolVersion)]
		if !ok {
			return badRequest(ctx, nil, "unsupported protocol version", v2.Log)
		}
		protocolVersion = protocol.ConsensusVersion(dr.ProtocolVersion)
	} else {
		protocolVersion = hdr.CurrentProtocol
	}
	dr.ProtocolVersion = string(protocolVersion)
```

algorandskiy (changes requested):
Protocol modification does not look correct

daemon/algod/api/server/v2/dryrun.go
```diff
@@ -397,17 +397,22 @@ func doDryrunRequest(dr *DryrunRequest, response *generated.DryrunResponse) {
...
-
+	proto.EnableAppCostPooling = true
+	proto.MaxTxGroupSize = 16
```
jannotti, line 401: For MaxTxGroupSize, we should figure out why it's not set.  That has been set for some time, so I don't see why you'd want it set here.

EnableCostPoooling is more interesting - On the one hand, if it's not set, the user may want his 701 unit app call to fail so that they can understand why.  But another way of looking at this is that you're seting EnableCostPooling unconditionally (and "locally") so that you can report the cost properly, and the user will thank you for reporting the 701 cost. Perhaps the best behavior is to
 1. Record the original value of EnableCostPooling
 2. Set it here so that the dry run can run long.
 3. If the call passes, but cost exceeds MaxAppCost and EnableCostPooling was not originally on, make it a failure.
 4. Report the cost either way.

daemon/algod/api/server/v2/dryrun.go
```diff
@@ -397,17 +397,22 @@ func doDryrunRequest(dr *DryrunRequest, response *generated.DryrunResponse) {
...
 	response.Txns = make([]generated.DryrunTxnResult, len(dr.Txns))
 	for ti, stxn := range dr.Txns {
 		pse := logic.MakePastSideEffects(len(dr.Txns))
+		maxBudget := uint64(proto.MaxAppProgramCost * proto.MaxTxGroupSize)
+		evalBudget := uint64(proto.MaxAppProgramCost * proto.MaxTxGroupSize)
```
jannotti, line 406: I think it's a slightly clearer thing to say `evalBudget := maxBudget` rather than repeat the init expression.  They are not two variables that happens to be set the same. evalBudget is a temporary value that must begin as maxBudget.

daemon/algod/api/server/v2/dryrun_test.go
```diff
@@ -1201,3 +1203,67 @@ return
...
+	var response generated.DryrunResponse
+	doDryrunRequest(&dr, &response)
+	require.Equal(t,uint64(11),response.Cost)
+	require.NoError(t, err)
+	checkAppCallPass(t, &response)
+	if t.Failed() {
+		logResponse(t, &response)
+	}
+
+}
```
jannotti, line 1269: Can you show that Cost ends up in the actual dry run response json object of the API?

jannotti (changes requested):
Looks like we need to decide how to handle some protocol parmater things, and test that the api response is updated to include cost.

daemon/algod/api/server/v2/dryrun.go
jannotti: You could also, if we want to make the dry run have higher fidelity output, also turn success into failure if the costs exceeds a tighter bound on the actual cost that would be available.  You could calculate that, in increasingly correct ways.
1. Use the size of the submitted request.
2. Use the number of app calls in the submitted request.
3. Actually track the available budget that would be available as multiple txns run.

daemon/algod/api/server/v2/dryrun_test.go
AyAggarwal: Yes the line above "require.Equal(t, uint64(11), response.Cost)"

daemon/algod/api/server/v2/dryrun.go
AyAggarwal: ok, I changed the line in dryrun_test.go to have protocolversion = ""

daemon/algod/api/server/v2/dryrun_test.go
AyAggarwal: just kidding, adding an e2e test

daemon/algod/api/server/v2/generated/types.go
```diff
@@ -498,6 +498,7 @@ type DryrunResponse struct {
 	// Protocol version is the protocol version Dryrun was operated under.
 	ProtocolVersion string            `json:"protocol-version"`
 	Txns            []DryrunTxnResult `json:"txns"`
+	Cost            uint64            `json:"cost"`
```
jasonpaulos, line 501: This is a generated file. Instead of modifying it, you should modify the `DryrunResponse` entry in `daemon/algod/api/algod.oas2.json`, then follow the instructions to regenerate: https://github.com/algorand/go-algorand/blob/master/daemon/algod/api/README.md#updating-the-v2-rest-api

Let me know if you have trouble

daemon/algod/api/algod.oas3.yml
```diff
@@ -361,6 +365,7 @@
                 }
               },
               "required": [
+                "cost",
```
algorandskiy, line 368: Not sure it needs to be required - when program errs no need to send cost back

daemon/algod/api/server/v2/dryrun.go
```diff
@@ -397,16 +397,19 @@ func doDryrunRequest(dr *DryrunRequest, response *generated.DryrunResponse) {
 		return
 	}
 	proto := config.Consensus[protocol.ConsensusVersion(dr.ProtocolVersion)]
-
+	proto.EnableAppCostPooling = true
```
algorandskiy, line 400: why is it set here? It is up to the caller or current protocol.

daemon/algod/api/server/v2/dryrun.go
```diff
@@ -397,16 +397,19 @@ func doDryrunRequest(dr *DryrunRequest, response *generated.DryrunResponse) {
...
 	response.Txns = make([]generated.DryrunTxnResult, len(dr.Txns))
 	for ti, stxn := range dr.Txns {
 		pse := logic.MakePastSideEffects(len(dr.Txns))
+		maxBudget := uint64(proto.MaxAppProgramCost * proto.MaxTxGroupSize)
+		evalBudget := maxBudget
```
algorandskiy, line 405: Maybe I miss something but this does not look correct. `maxBudget` is  MaxAppProgramCost * (num app call txns)

daemon/algod/api/server/v2/dryrun.go
jannotti: The idea here is to use the fee pooling mechanism, regardless of whether it is turned on in the protocol, for two reasons:
a) It gives us a way to access the cost (because it will be decremented if it's active)
b) It gives us a way to overrun the allowable cost, so we can accurately report overruns (within reason), which seems useful to developers.  We ought to indicate that the execution "failed" but we can show the real cost.

daemon/algod/api/server/v2/dryrun.go
jannotti: Related to the previous idea: We're willing to be overly generous with the cost, so that the execution will first "succeed" but we could (with a bit extra code), properly turn the success to failure after the fact.  This would be bad in production (we're letting people use the CPU too much), but nice for dry-run.

That requires a bit of code that isn't there yet (I think, I haven't looked). It would determine the *real* maxCost as well, and confirm that the cost is under that.  But it would do it *after* execution, rather than setting the pooled cost exactly right from the start.

algorandskiy: then it should be heavily commented, otherwise it is not obvious at all

AyAggarwal:
did 3 things:
1. added better commenting to explain what is going on
2. preserved user's EnableAppCostPooling choice and added logic to reject if any program is above maxAppCost
3. implemented maxCost estimate by using appCost * numAppCalls, and REJECT programs that surpass this. added tests for this case

daemon/algod/api/algod.oas2.json
```diff
@@ -2509,6 +2509,10 @@
           "protocol-version": {
             "description": "Protocol version is the protocol version Dryrun was operated under.",
             "type": "string"
+          },
+          "cost": {
+            "description": "Cost is the execution call of a stateful evaluation of an app call.",
```
shiqizng, line 2514: can this description be better? this description makes it sound like cost is an execution call.

algorandskiy: updated and move to `DryrunTxnResult`

PR #2750, opened 2021-08-16
Increase max app opt ins to 50

algorandskiy:
## Summary

As a temporary relief increase max number of opt ins to 50.
Also found and fixed an error check in new ledger testing harness.

## Test Plan

* Unit tested ledger.apply and ledger.eval

## Benchmarks

Benchmark 1: create 100k accounts, 500 apps (no local data) and opt-in 10k accounts to max number of apps. Restart ledgers, measure a single block validation time of 50k pay transactions on random opted-in accounts.
| 10 opt ins | 50 opt ins | 100 opt ins |
|-----------|-----------|------------|
| 644 ms | 660 ms | 683 ms |

Benchmark 2: same as 2 but with full local data (16 keys each 64 bytes long with 64 bytes of value data)
| 10 opt ins | 50 opt ins | 100 opt ins |
|-----------|-----------|------------|
| 1.4 - 2.2 s |  9-16 s | ~20-25 s |

Benchmark 3: same as 2 but 5k txns per block
| 50 opt ins | 30 opt ins | 20 opt ins | 10 opt ins |
|-----------|-----------|------------|-----------|
| 3.2 s | 1.7 s | 1.1 s | 0.27 s|

### DIFF
--- a/ledger/eval_test.go
+++ b/ledger/eval_test.go
@@ -20,11 +20,13 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"math/rand"
 	"os"
 	"path/filepath"
 	"reflect"
 	"runtime/pprof"
 	"strings"
+	"sync"
 	"testing"
 	"time"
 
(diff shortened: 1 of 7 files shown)


ledger/eval_test.go
```diff
@@ -1231,7 +1231,7 @@ func (eval *BlockEvaluator) txn(t testing.TB, txn *txntest.Txn) {
 	stxn := txn.SignedTxn()
 	err := eval.testTransaction(stxn, eval.state.child(1))
 	require.NoError(t, err)
-	eval.Transaction(stxn, transactions.ApplyData{})
+	err = eval.Transaction(stxn, transactions.ApplyData{})
```
algorandskiy, line 1234: @jannotti FYI

jannotti: Thanks, I have that in my application-actions branch.

ledger/eval_test.go
```diff
@@ -1402,3 +1410,57 @@ func TestModifiedAppLocalStates(t *testing.T) {
...
+	appsCreated := make(map[basics.Address]int, len(addrs)-1)
+
+	acctIdx := 0
+	for i := 0; i < maxAppsOptedIn; i++ {
+		creator := addrs[acctIdx]
+		createTxn := txntest.Txn{
+			Type:              "appl",
+			Sender:            creator,
+			ApprovalProgram:   []byte{0x02, 0x20, 0x01, 0x01, 0x22},
+			ClearStateProgram: []byte{0x02, 0x20, 0x01, 0x01, 0x22},
```
jannotti, line 1441: This will get clearer after my last PR.  You'll put the actual program text in ApprovalProgram, and it'll get assembled on demand.

ledger/eval_test.go
```diff
@@ -1402,3 +1410,57 @@ func TestModifiedAppLocalStates(t *testing.T) {
...
+			Sender:        addrs[9],
+			ApplicationID: appid + basics.AppIndex(i),
+			OnCompletion:  transactions.OptInOC,
+		}
+		txnsOptIn = append(txnsOptIn, &optInTxn)
+	}
+
+	eval := l.nextBlock(t)
+	txns := append(txnsCreate, txnsOptIn...)
+	eval.txns(t, txns...)
```
jannotti, line 1463: Can this just be `eval.txns(txnsCreate..., txnsOptIn...)`

ledger/eval_test.go
```diff
@@ -1402,3 +1410,57 @@ func TestModifiedAppLocalStates(t *testing.T) {
 		assert.False(t, created)
 	}
 }
+
+// TestAppInsMinBalance checks that accounts with MaxAppsOptedIn are accepted by block evaluator
+// and do not cause any MaximumMinimumBalance problems
```
jannotti, line 1415: I guess it should also be opted into 1000 assets, to make sure MaxMinBalance is correct.

jannotti: Also, the apps being opted into must have "full" schemas to get max min balance.

algorandskiy: > Also, the apps being opted into must have "full" schemas to get max min balance.
Agree

>I guess it should also be opted into 1000 assets, to make sure MaxMinBalance is correct.
No, because 1000 assets will consume entire MinMaxBalance alone (1000 * 0.1 = 100 that is almost 100.1 of MaxMinBalance). So a user needs to decide what he wants to store - if we want to change it, it needs to be a separate PR.

jannotti: I think that MinMaxBalance value is, essentially, a bug right now.  I believe the spec says you can opt into 1000 assets and 10 apps.  That's not true.  But I think the intent is that it is.

I guess I don't care too much if it's in this PR or another.

algorandskiy: The spec only says without binding it 1000 assets or 10 apps or something. I might be missing something tho
```
A transaction must be rejected if its inclusion in a block would cause an account's minimum balance to exceed the maximum minimum balance of 100100000 microalgos.
```

ledger/eval_test.go
algorandskiy: no, the compiler complains
```
ledger/eval_test.go:1464:30: expected ')', found txnsOptIn
ledger/eval_test.go:1467:3: expected '}', found 'EOF'
```

ledger/eval_test.go
jannotti: It also separately says both of the asset and app limits:
"Accounts can hold up to $Assets_{\max}$ assets (1000 in this protocol)"
"The maximum number of applications that a single account may opt in to is 10."

I suppose the statement on MaxMinBalance supersedes these statements, in a sense, but I think the intent was to set MaxMinBalance so that it did not prevent the other limits but we forgot to bump it when we added apps.  I'm only guessing.

ledger/eval_test.go
jannotti: I guess ... must appear only once, at the end.

jannotti:
High level question - since we will have a different value for app creation limit and app optin limit, can I do BOTH? Create 10 apps (but not opt in to them) and then opt in to 50 more?

I don't know if we currently allow 10 creations and then 10 (separate) opt-ins.  I suppose we do, so I suppose this 10+50 I mentioned above will be allowed as well.

brianolson (commented):
I'm mostly here for the expanded ledger/eval_test.go and generally it LGTM

algorandskiy:
> High level question - since we will have a different value for app creation limit and app optin limit, can I do BOTH? Create 10 apps (but not opt in to them) and then opt in to 50 more?

Yes, it is still allowed: 1) checks are in different places - create app / optin app. 2) `MaximumMinimumBalance` is not exceeded.

tsachiherman:
hm. I'm a bit concerned about the timing non-symmetry between the block assembly and validation. When we create a proposal, we limit the time to a specific threshold (250ms). The *assumption* is that the block validation would take roughly the same time. However, the transactions used in the block assembly are taken from the transaction pool - which means that the corresponding accounts have already been loaded into memory.

When receiving a transaction to validate - we do not have that "luxury" of knowing that the account information was already loaded into memory. Instead, we'll need to fetch it from disk (bummer..). As long a the account's data wasn't substantial, it was perfectly fine. However, in case we have large accounts, we need to make sure that we'll be able to load all of them into memory in a reasonable time.

@pzbitskiy's test was already on the right track - but it would need to be "fine-tuned" for the worst case scenario:
we need to find how many of these "large" accounts we can process on the validator on a block size with 5K transactions.
then, we could establish that we want to complete the block verification within ( let's say ) 3 seconds. And from there, we can find what size we can increase the app without any risk to the protocol.
( or - if that's not the case, we should make a decision based on that.. )

algorandskiy:
> When receiving a transaction to validate - we do not have that "luxury" of knowing that the account information was already loaded into memory. Instead, we'll need to fetch it from disk (bummer..). 

The test already does it - it reopens the ledger so evaluation happens on empty caches. Plus all 5k txns are from random addresses out of 10k opted in accounts so they are presumably different.

Data on 5k txn per block:
| 50 opt ins | 30 opt ins | 20 opt ins | 10 opt ins |
|-----------|-----------|------------|-----------|
| 3.2 s | 1.7 s | 1.1 s | 0.27 s|

algorandskiy:
@tsachiherman @Priegle thoughts? evaluation of 5k txns on accounts with full local state in each of 50 app slots would takes 3.2s on a cold DB.

tsachiherman:
> @tsachiherman @Priegle thoughts? evaluation of 5k txns on accounts with full local state in each of 50 app slots would takes 3.2s on a cold DB.

No - that's exactly what I was looking for. It tells me that it's fine to increase the number to application to 50 as long as our block size is roughly 5k transaction. If we will make an attempt to increase it, we could see an inevitable increase in round time.

algorandskiy:
Btw, as JJ noted, we already have a similar problem with 1k assets per account.

algorandskiy:
Chris pointed out to [the code](https://github.com/algorand/go-algorand/blob/master/agreement/pseudonode.go#L268-L277)
```
// makeProposals creates a slice of block proposals for the given round and period.
func (n asyncPseudonode) makeProposals(round basics.Round, period period, accounts []account.Participation) ([]proposal, []unauthenticatedVote) {
    deadline := time.Now().Add(config.ProposalAssemblyTime)
    ve, err := n.factory.AssembleBlock(round, deadline)
    if err != nil {
```
So the assembly deadline is hard.

For a block verification case looks like agreement switches if there is not enough cert votes to a next period that lead so a longer round time.

tsachiherman:
No, that's not why it would increase the block time. Once the agreement
recieve the proposal, it starts validating it. It cannot move to the next
round without completing the validation process since it need the state
deltas ( regardless of the incoming cert votes ). If the validation starts
at r+2, and it takes 3.2s, then the new round would start at r+5.2. The
ensure digest path would be skipped, since we have the proposal.

jannotti (approved):

algorandskiy:
Rebased

jannotti (approved):

PR #2752, opened 2021-08-16
testing: added MsgBuffer unit tests

### LINKED ISSUE #2459: Txn sync: Msgbuffer unit test
Transaction sync readiness test for message buffer.

AlgoStephenAkiki:
## Summary

Added unit tests to msgbuffer in msgbuffers_test.go

Resolves #2459 

## Test Plan

These are unit tests.

### DIFF
--- a/txnsync/msgbuffers_test.go
+++ b/txnsync/msgbuffers_test.go
@@ -17,6 +17,7 @@
 package txnsync
 
 import (
+	"sync"
 	"testing"
 
 	"github.com/stretchr/testify/require"
@@ -26,6 +27,18 @@ import (
 
 // TestMessageBuffersPool tests that a buffer pool can be retrieved and has proper length/capacity properties
 func TestMessageBuffersPool(t *testing.T) {
+
+	counter := 0
+
+	oldFunc := msgBuffersPool.New
+
+	msgBuffersPool = sync.Pool{
+		New: func() interface{} {
+			counter++
+			return oldFunc()
+		},
+	}
+
 	partitiontest.PartitionTest(t)
 
 	for i := 0; i < 10; i++ {
@@ -34,8 +47,34 @@ func TestMessageBuffersPool(t *testing.T) {
 		require.GreaterOrEqual(t, cap(bytes), messageBufferDefaultInitialSize)
 
 		releaseMessageBuffer(bytes)
+		require.Equal(t, 1, counter)
 	}
 
+	collector := [][]byte{}
+
+	counter = 1
+
+	for i := 0; i < 10; i++ {
+		bytes := getMessageBuffer()
+		collector = append(collector, bytes)
+		require.Equal(t, 0, len(bytes))
+		require.GreaterOrEqual(t, cap(bytes), messageBufferDefaultInitialSize)
+
+		require.Equal(t, i+1, counter)
+	}
+
+	for _, b := range collector {
+		releaseMessageBuffer(b)
+	}
+
+	for i := 0; i < 10; i++ {
+		bytes := getMessageBuffer()
+		collector = append(collector, bytes)
+		require.Equal(t, 0, len(bytes))
+		require.GreaterOrEqual(t, cap(bytes), messageBufferDefaultInitialSize)
+
+		require.Equal(t, 10, counter)
+	}
 }
 
 // TestTxIDSlicePool tests that the transaction id pool can be retrieved and has proper length/capacity properties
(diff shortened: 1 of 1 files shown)


txnsync/msgbuffers_test.go
```diff
@@ -0,0 +1,59 @@
...
+		releaseTxIDSliceBuffer(txIDs)
+
+	}
+
+	// Test that one of the previous buffers can be reused
+	txIDs := getTxIDSliceBuffer(5)
+	require.Equal(t, 0, len(txIDs))
+	require.GreaterOrEqual(t, cap(txIDs), 5)
+	releaseTxIDSliceBuffer(txIDs)
+
```
algonautshant, line 58: nit: extra line

txnsync/msgbuffers_test.go
```diff
@@ -0,0 +1,59 @@
...
+// TestTxIDSlicePool tests that the transaction id pool can be retrieved and has proper length/capacity properties
+func TestTxIDSlicePool(t *testing.T) {
+	partitiontest.PartitionTest(t)
+
+	for i := 10; i < 100; i += 10 {
+		txIDs := getTxIDSliceBuffer(i)
+		require.Equal(t, 0, len(txIDs))
+		require.GreaterOrEqual(t, cap(txIDs), i)
+		releaseTxIDSliceBuffer(txIDs)
+
```
algonautshant, line 50: nit: extra line

txnsync/msgbuffers_test.go
```diff
@@ -0,0 +1,59 @@
...
+	partitiontest.PartitionTest(t)
+
+	for i := 0; i < 10; i++ {
+		bytes := getMessageBuffer()
+		require.Equal(t, 0, len(bytes))
+		require.GreaterOrEqual(t, cap(bytes), messageBufferDefaultInitialSize)
+
+		releaseMessageBuffer(bytes)
+	}
+
```
algonautshant, line 38: nit: extra line

tsachiherman (approved):
looks good. We could improve that further, but it might be overlapping with the sync.Pool implementation, which already has its own tests.

txnsync/msgbuffers_test.go
```diff
@@ -0,0 +1,59 @@
...
+	for i := 10; i < 100; i += 10 {
+		txIDs := getTxIDSliceBuffer(i)
+		require.Equal(t, 0, len(txIDs))
+		require.GreaterOrEqual(t, cap(txIDs), i)
+		releaseTxIDSliceBuffer(txIDs)
+
+	}
+
+	// Test that one of the previous buffers can be reused
```
algonautshant, line 53: How are you making sure that one of the previous buffers is being reused?

txnsync/msgbuffers_test.go
```diff
@@ -0,0 +1,59 @@
...
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/algorand/go-algorand/test/partitiontest"
+)
+
+// TestMessageBuffersPool tests that a buffer pool can be retrieved and has proper length/capacity properties
+func TestMessageBuffersPool(t *testing.T) {
+	partitiontest.PartitionTest(t)
+
```
algonautshant, line 30: You can overwrite the New function to examine when and how many new entries are added to the pool, and verify it is as expected.

```
	msgBuffersPool = sync.Pool{
		New: func() interface{} {
			return make([]byte, 0, messageBufferDefaultInitialSize)
		},
	}

```

txnsync/msgbuffers_test.go
```diff
@@ -0,0 +1,59 @@
...
+	for i := 0; i < 10; i++ {
+		bytes := getMessageBuffer()
+		require.Equal(t, 0, len(bytes))
+		require.GreaterOrEqual(t, cap(bytes), messageBufferDefaultInitialSize)
+
+		releaseMessageBuffer(bytes)
```
algonautshant, line 36: Try to accumulated some pointers, and check if the pool size increases. 
Then release, and make sure after releasing, the pool size no longer increases.

algonautshant (changes requested):

txnsync/msgbuffers_test.go
AlgoStephenAkiki: Added a comment in the code explaining my reasoning

txnsync/msgbuffers_test.go
```diff
@@ -0,0 +1,100 @@
...
+
+	partitiontest.PartitionTest(t)
+
+	for i := 0; i < 10; i++ {
+		bytes := getMessageBuffer()
+		require.Equal(t, 0, len(bytes))
+		require.GreaterOrEqual(t, cap(bytes), messageBufferDefaultInitialSize)
+
+		releaseMessageBuffer(bytes)
+		require.Equal(t, 1, counter)
```
tsachiherman, line 50: this statement is not guaranteed to be correct ( and, in fact, it fails on circle-ci : https://app.circleci.com/pipelines/github/algorand/go-algorand/1365/workflows/e240c3f9-79a7-4f6d-b24d-cf9d008fd1c5/jobs/9631/tests )

The memory might be returned to the pool, but would not be available right away.
another issue with this test is that it modifies the global msgBuffersPool; this modification is made without any mutex, and this could lead to data races.

instead of conducting the test in that way, I would suggest the following:
instead of using getMessageBuffer/releaseMessageBuffer, start by allocating a buffer of size 400 ( just saying ), and fill it up with some "special value". then call releaseMessageBuffer which would place the buffer into the message buffer. 
Subsequent calls should retrieve that buffer, which could be identified by its size and content. ( i.e. you'll need to resize the buffer to it's max capacity ).
last, because you can't get a guarantee that the buffer wouldn't be consumed by other concurrent running tests, you should probably run the test for a while, and verify that you get the said buffer at least once.

AlgoStephenAkiki: I've added a new version of the test that does a variation of this.

algonautshant (approved):
Looks great! 
Thanks for the updates.

PR #2760, opened 2021-08-17
Improvements for testing txgroups conveniently.

jannotti:
Also, fees, amounts are given to Txn as uint64.  Programs as
text. They are converted internally so that testing is more convenient
and clear.

### DIFF
--- a/data/txntest/txn.go
+++ b/data/txntest/txn.go
@@ -33,11 +33,15 @@
 package txntest
 
 import (
+	"fmt"
+	"strings"
+
 	"github.com/algorand/go-algorand/config"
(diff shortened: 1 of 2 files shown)


algorandskiy (commented):
I like Fee/Amount change but ApprovalProgram brings some non-needed assembling overhead when thousands of transactions are being submitted (large tests or benchmarks as examples). Maybe `ApprovalProgram` should be an `interface{}` with only two allowed types `[]byte` and `string`.

algochoi (approved):
LGTM 👍

jannotti:
This will go in as part of application-actions

PR #2762, opened 2021-08-17
Enhance typechecking after a branch

iten-alg:
This PR sets up infrastructure to make typechecking better during assembly by dividing the code into basic blocks (segments of code where the only entrance is at the beginning and the only exit is at the end) and then erroring whenever we find type errors within a basic block. This allows us to find errors that the assembler previously relegated to warnings b/c of branching instructions. 

For now, we only check within a block, but in the future this infrastructure will be used to allow us to error when we identify that any possible path into a block results in an error (this would be stronger than just typechecking within a block b/c it allows us to identify spots where for example every path into a block has a uint64 on top of the stack, but the first op in the block requires a []byte.

Tests have been added to assembler_test.go to confirm that we do now error on certain cases that used to be warnings, but there are parts of the infrastructure that aren't used yet, so I did not add tests for those.


jannotti:
I'm going to close this one for now.  The basic block stuff here is more sophisticated than the recent addition of type checking in basic blocks, so we may revisit this if we want to do analysis beyond checking only within the block.  This PR has a chance of doing checks that confirm that since all entries to a block have the same stack shape, more checks can be done. But it'll be some work to deal with the assembler changes since then.

PR #2766, opened 2021-08-18
Improve error message in update.sh when updater is corrupt.

winder:
## Summary
A user reported a strange and not very helpful error message. This was due to incomplete error handling in the bash script relating to a truncated `updater` binary.

The current workaround for this is to delete the `updater` binary and trigger `update.sh` to re-download it, but this is not obvious and I wouldn't expect any user to reach that conclusion.

This fix adds error handling in case `updater` fails and should fix this particular failure.

Here is the error message:
```
Current Version = 8590458880
Latest Version =
./update.sh: line 250: [: 8590458880: unary operator expected
New version found
Update Downloaded to /tmp/tmp.QlV71NLGFr/.tar.gz
Expanding update...
tar (child): /tmp/tmp.QlV71NLGFr/.tar.gz: Cannot open: No such file or directory
tar (child): Error is not recoverable: exiting now
tar: Child returned status 2
tar: Error is not recoverable: exiting now
*** UPDATE FAILED: Error expanding update ***
```
## Test Plan

Ran some tests against a directory, empty file, and regular file in my shell:
```
will@arendelle:~/go/src/github.com/algorand/go-algorand/cmd/updater$ [[ -f dir && -s dir ]]
will@arendelle:~/go/src/github.com/algorand/go-algorand/cmd/updater$ echo $?
1
will@arendelle:~/go/src/github.com/algorand/go-algorand/cmd/updater$ [[ -f empty && -s empty ]]
will@arendelle:~/go/src/github.com/algorand/go-algorand/cmd/updater$ echo $?
1
will@arendelle:~/go/src/github.com/algorand/go-algorand/cmd/updater$ [[ -f util.go && -s util.go ]]
will@arendelle:~/go/src/github.com/algorand/go-algorand/cmd/updater$ echo $?
0
```

### DIFF
--- a/cmd/updater/update.sh
+++ b/cmd/updater/update.sh
@@ -246,7 +246,7 @@ function check_for_update() {
     fi
 
     if [ -z ${LATEST} ]; then
-        echo "Failed to lookup latest release. This could happen if the 'updater' binary is corrupted."
+        echo "Failed to lookup latest release"
         return 1
     fi
 
--- a/cmd/updater/update.sh
+++ b/cmd/updater/update.sh
@@ -245,7 +245,12 @@ function check_for_update() {
         return 1
     fi
 
-    echo Latest Version = ${LATEST}
+    if [ -z ${LATEST} ]; then
+        echo "Failed to lookup latest release. This could happen if the 'updater' binary is corrupted."
+        return 1
+    fi
+
+    echo "Latest Version = ${LATEST}"
 
     if [ ${CURRENTVER} -ge ${LATEST} ]; then
         if [ "${UPDATETYPE}" = "install" ]; then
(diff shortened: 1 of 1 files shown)


cmd/updater/update.sh
```diff
@@ -196,8 +196,8 @@ function get_updater_url() {
 
 # check to see if the binary updater exists. if not, it will automatically the correct updater binary for the current platform
 function check_for_updater() {
-    # check if the updater binary exist.
-    if [ -f "${SCRIPTPATH}/updater" ]; then
+    # check if the updater binary exist and is not empty.
+    if [[ -s "${SCRIPTPATH}/updater" && -f "${SCRIPTPATH}/updater" ]]; then
```
tsachiherman, line 200: could you please check this line with shellcheck ?

winder: Shellcheck doesn't have anything to say about this line

tsachiherman (approved):

PR #2769, opened 2021-08-19
ParticipationRegistry Cache

winder:
## Summary

Add a cache in the ParticipationRegistry to optimize asynchronous data access and updates.

## Test Plan

Update unit tests.

### DIFF
--- a/data/account/participationRegistry.go
+++ b/data/account/participationRegistry.go
@@ -81,6 +81,9 @@ var ErrAlreadyInserted = errors.New("these participation keys are already insert
 // ErrActiveKeyNotFound is used when attempting to update an account with no active key
 var ErrActiveKeyNotFound = errors.New("no active participation key found for account")
 
+// ErrMultipleValidKeys is used when recording a result but multiple valid keys were found. This should not be possible.
+var ErrMultipleValidKeys = errors.New("multiple valid keys found while recording key usage")
+
 // ParticipationRegistry contain all functions for interacting with the Participation Registry.
 type ParticipationRegistry interface {
 	// Insert adds a record to storage and computes the ParticipationID
@@ -495,27 +498,41 @@ func (db *participationDB) Record(account basics.Address, round basics.Round, pa
 	db.mutex.Lock()
 	defer db.mutex.Unlock()
 
+	matches := make([]ParticipationID, 0)
 	// At most one id should be updated.
 	for _, record := range db.cache {
 		if record.Account == account && record.FirstValid <= round && round <= record.LastValid {
-			switch participationAction {
-			case Vote:
-				record.LastVote = round
-			case BlockProposal:
-				record.LastBlockProposal = round
-			case CompactCertificate:
-				record.LastCompactCertificate = round
-			default:
-				return ErrUnknownParticipationAction
-			}
-
-			db.dirty[record.ParticipationID] = struct{}{}
-			copy := record
-			db.cache[record.ParticipationID] = copy
-			return nil
+			matches = append(matches, record.ParticipationID)
 		}
 	}
 
+	// Good case, one key found.
+	if len(matches) == 1 {
+		record := db.cache[matches[0]]
+
+		switch participationAction {
+		case Vote:
+			record.LastVote = round
+		case BlockProposal:
+			record.LastBlockProposal = round
+		case CompactCertificate:
+			record.LastCompactCertificate = round
+		default:
+			return ErrUnknownParticipationAction
+		}
+
+		db.dirty[record.ParticipationID] = struct{}{}
+		copy := record
+		db.cache[record.ParticipationID] = copy
+		return nil
+	}
+
+	// This probably means there is a bug in the key participation registry Register implementation.
+	if len(matches) > 1 {
+		return ErrMultipleValidKeys
+	}
+
+	// This indicates the participation registry is not synchronized with agreement.
 	return ErrActiveKeyNotFound
 }
 
--- a/data/account/participationRegistry.go
+++ b/data/account/participationRegistry.go
@@ -21,6 +21,7 @@ import (
 	"database/sql"
 	"errors"
 	"fmt"
+	"github.com/algorand/go-deadlock"
 
 	"github.com/algorand/go-algorand/crypto"
 	"github.com/algorand/go-algorand/data/basics"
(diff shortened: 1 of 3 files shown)


data/account/participation.go
```diff
@@ -58,27 +57,36 @@ type Participation struct {
...
 }
 
+// participationIDData is for msgpack encoding the participation data.
+type participationIDData struct {
+	_struct     struct{} `codec:""`
+	Parent      basics.Address
+	VRF         crypto.VRFSecrets
+	FirstValid  basics.Round
+	LastValid   basics.Round
+	KeyDilution uint64
```
tsachiherman, line 67: I'd suggest you'll try to keep the same standard codec definitions we have on this object as well.

data/account/participationRegistry.go
```diff
@@ -332,61 +407,163 @@ func (db *participationDB) GetAll() (records []ParticipationRecord, err error) {
...
+	for _, record := range db.cache {
+		copy := record
+		results = append(results, copy)
+	}
+	return results, nil
+}
+
+// updateRollingFields sets all of the rolling fields according to the record object.
+func (db *participationDB) updateRollingFields(ctx context.Context, tx *sql.Tx, record ParticipationRecord) error {
+	_, err := tx.ExecContext(ctx, updateRollingFields,
```
tsachiherman, line 435: check the number of affected records. You know that it should always result in a single row update, so you should always test for that.

data/account/participationRegistry_test.go
```diff
@@ -17,28 +17,24 @@
...
+func getRegistry(t *testing.T) *participationDB {
 	rootDB, err := db.OpenPair(t.Name(), true)
 	require.NoError(t, err)
 
 	registry, err := MakeParticipationRegistry(rootDB)
 	require.NoError(t, err)
 	require.NotNil(t, registry)
 
-	return registry
+	return registry.(*participationDB)
```
tsachiherman, line 37: you shouldn't need to do that. instead, create a `makeParticipationRegistry` that would return a `* participationDB`, and use it internally. in the `MakeParticipationRegistry`, just return the external facing interface.

data/account/participation.go
winder: Pointers don't seem to be supported, which was causing the generated msgp tests to fail. Is there a standard convenient for dealing with that?

tsachiherman: let me take a look on that one locally and I'll let you know ?

tsachiherman: so, instead of having the `crypto.VRFSecrets` here, use just the `SK` field `crypto.VrfPrivkey`. The `PK` is derived from the `SK` anyway.

winder: Maybe we can chat next week. What I have here is technically working as designed, but I had to redefine the `Participation` type. `Participation` didn't work properly because the generated marshalling function didn't seem to work properly with pointer types.

tsachiherman: see - https://github.com/winder/go-algorand/pull/4

winder: Thanks, updated.

data/account/participationRegistry.go
```diff
@@ -332,61 +415,176 @@ func (db *participationDB) GetAll() (records []ParticipationRecord, err error) {
...
+	return record, nil
+}
+
+func (db *participationDB) GetAll() ([]ParticipationRecord, error) {
+	db.mutex.RLock()
+	defer db.mutex.RUnlock()
+
+	results := make([]ParticipationRecord, 0, len(db.cache))
+	for _, record := range db.cache {
+		copy := record
```
tsachiherman, line 435: This would not perform a deep copy, if this was your intent.

data/account/participationRegistry.go
```diff
@@ -332,61 +415,176 @@ func (db *participationDB) GetAll() (records []ParticipationRecord, err error) {
...
+		record.ParticipationID[:])
+	if err != nil {
+		return err
+	}
+
+	numRows, err := result.RowsAffected()
+	if err != nil {
+		return err
+	}
+	if numRows > 1 {
```
tsachiherman, line 458: What about numRows==0 ?

data/account/participationRegistry.go
```diff
@@ -332,61 +415,176 @@ func (db *participationDB) GetAll() (records []ParticipationRecord, err error) {
...
+
+func (db *participationDB) Flush() error {
+	db.mutex.Lock()
+	defer db.mutex.Unlock()
+
+	if len(db.dirty) == 0 {
+		return nil
+	}
+
+	return db.store.Wdb.Atomic(func(ctx context.Context, tx *sql.Tx) error {
```
tsachiherman, line 572: Refactor to avoid creating the atomic transaction unless you're truely hoing to flush.

tsachiherman (changes requested):
Overall looks great. Few small requests.

data/account/participationRegistry.go
winder: This is the "Flush()" function which would be called between each round. Is there a way to do this without an Atomic transaction?

tsachiherman: no, you would still need the Atomic operation, but if you'll look into what you're doing there, you'll see that in some cases, no updates would be made.
What I'm asking is to figure this out before making the Atomic call.

winder: That check is happening with an early return above:
```
	if len(db.dirty) == 0 {
		return nil
	}
```

data/account/participationRegistry.go
winder: Thanks, I'll return an error here. I was thinking that it's safe to ignore, but looking closer that isn't always true - specifically when setting that a key is registered.

data/account/participationRegistry.go
winder: It is my intent to create a copy, I don't want the caller to modify the cache.

Are you thinking that if there were pointer values in the cache this would fail? In this case, `ParticipationRecord` has no pointers, but is there a safer way to ensure a deep copy?

data/account/participationRegistry.go
tsachiherman: This thing is - if the key is missing from the database, and we have the entry in the cache, we might want to report the error, but then - also remove it from the cache to avoid repeating this error.
( since the database won't "fix" itself, unless we somehow "fix" the situation )

data/account/participationRegistry.go
tsachiherman: Yes, but I was referring to the `ErrParticipationIDNotFound` case, which should be ( in my view ), tested early.
but thinking about that - if we have 10 part keys, and one of them would generate the unexpected `ErrParticipationIDNotFound` error - we still want to flush all the rest. So, summing this up:
- the method should update all the keys it can.
- add the log to the participationDB so that we can log the error right away : we have the potential of having multiple errors during a single Flush() invocation.
- return an error only if the caller can do something useful with it.

data/account/participationRegistry.go
```diff
@@ -154,26 +185,16 @@ var (
...
-	// there should be, at most, a single record within the effective range.
-	// only the effectiveLastValid can change here.
-	clearRegistered = `UPDATE Rolling
-		 SET registeredLastRound = ?1
-		 WHERE pk IN (SELECT pk FROM Keysets WHERE account = ?2)
-		 AND registeredFirstRound < ?1
-		 AND registeredLastRound > ?1`
-	setRegistered = `UPDATE Rolling
-		 SET registeredFirstRound=?,
+	selectRecord        = selectRecords + ` AND participationID = ?`
```
tsachiherman, line 188: I'm not sure why you need this concatenation anymore.

PR #2780, opened 2021-08-20
Make catchpoint generation asynchronous

algorandskiy:
## Summary

According to https://www.sqlite.org/isolation.html if
   - DB is opened in WAL mode
   - There are two different connections

then read transaction in connection A sees unchanged DB state
even if connection B writes data in parallel

## Test Plan

Existing test suite

### DIFF
--- a/ledger/acctupdates.go
+++ b/ledger/acctupdates.go
@@ -365,7 +365,7 @@ func (au *accountUpdates) loadFromDisk(l ledgerForTracker) error {
 	}
 
 	if writingCatchpointRound != 0 && au.catchpointInterval != 0 {
-		au.generateCatchpoint(basics.Round(writingCatchpointRound), au.lastCatchpointLabel, writingCatchpointDigest, time.Duration(0))
+		au.generateCatchpoint(basics.Round(writingCatchpointRound), au.lastCatchpointLabel, writingCatchpointDigest, time.Duration(0), nil)
 	}
 
 	au.voters = &votersTracker{}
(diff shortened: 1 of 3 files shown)


util/db/dbutil.go
```diff
@@ -132,6 +136,13 @@ func makeAccessorImpl(dbfilename string, readOnly bool, inMemory bool, params []
 	return db, err
 }
 
+// Clone creates a new Accessor with the same parameters as the original one
+func (db *Accessor) Clone() (Accessor, error) {
```
tsachiherman, line 140: This won't work correctly for writer accessor, since we can have only one writer accessor.

algorandskiy: yes, I think I'll limit it to RO and WAL in params to indicate intended use

ledger/acctupdates.go
```diff
@@ -2384,8 +2384,16 @@ func (au *accountUpdates) latest() basics.Round {
 	return au.dbRound + basics.Round(len(au.deltas))
 }
 
+func (au *accountUpdates) generateCatchpointAsync(committedRound basics.Round, label string, committedRoundDigest crypto.Digest, updatingBalancesDuration time.Duration) {
+	done := make(chan struct{})
```
tsachiherman, line 2388: make this into a buffered channel of size of 1, and add a `defer close(async)` in `generateCatchpoint` to ensure that in any case the method exits, you don't get stuck leaving zombies.

tsachiherman (changes requested):
I added a small request for the `async` channel - but beside that, we could use some more testing around this to ensure it's working correctly.
I would try to change the catchpoint interval to 2 or 3 and see if it's still generating the catchpoints correctly. I don't know if the logic in commitRound is setup correctly for this setup. In particular, I'd be looking into the `au.catchpointSlowWriting` to see if it's being evaluated correctly.

algorandskiy:
I'm closing this because it does not really makes things better b/c there is a check in `committedUpTo`:
```golang
if au.IsWritingCatchpointFile() {
  ...
  return
}
```
that prevents additional calls to `commitRound`. And the logic is correct - we do not really want to be in a situation where it is a time to write a subsequent catchpoint while a previous one still in progress.

PR #2781, opened 2021-08-20
Fix goalNodeTest.exp

algorandskiy:
## Summary

Status code differs on CircleCI. Relaxed the status code value check.
Change the skipped expect tests to be defined as "t.Skip" so we can track these as such on the log file.

## Test Plan

This is a test fix.

### DIFF
--- a/test/e2e-go/cli/goal/expect/goalNodeTest.exp
+++ b/test/e2e-go/cli/goal/expect/goalNodeTest.exp
@@ -79,7 +79,7 @@ if { [catch {
     spawn goal node start -d $TEST_PRIMARY_NODE_DIR
     expect {
         timeout { close; ::AlgorandGoal::Abort "starting node exceeded timeout" }
-        -re {^Algorand node failed to start: node exited with an error code, check node\.log for more details : exit status 1} {
+        -re {^Algorand node failed to start: node exited with an error code, check node\.log for more details : exit status \d+} {
             puts "\nExpected failuire : node failed to start"
             # wait until the eof signal is received
             expect {
--- a/test/framework/fixtures/expectFixture.go
+++ b/test/framework/fixtures/expectFixture.go
@@ -18,6 +18,7 @@ package fixtures
 
 import (
 	"bytes"
+	"fmt"
 	"io/ioutil"
 	"os"
 	"os/exec"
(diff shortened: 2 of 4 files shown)


test/e2e-go/cli/goal/expect/goalNodeTest.exp
```diff
@@ -79,7 +79,7 @@ if { [catch {
     spawn goal node start -d $TEST_PRIMARY_NODE_DIR
     expect {
         timeout { close; ::AlgorandGoal::Abort "starting node exceeded timeout" }
-        -re {^Algorand node failed to start: node exited with an error code, check node\.log for more details : exit status 1} {
+        -re {^Algorand node failed to start: node exited with an error code, check node\.log for more details : exit status .*} {
```
tsachiherman, line 82: yes, that's my thinking as well - but maybe `/d*` to ensure it's digits.

test/framework/fixtures/expectFixture.go
```diff
@@ -114,7 +114,8 @@ func MakeExpectTest(t *testing.T) *ExpectFixture {
 // Run Process all expect script files with suffix Test.exp within the current directory
 func (ef *ExpectFixture) Run() {
 	disabledTest := map[string]bool{
-		"pingpongTest.exp": true,
+		"pingpongTest.exp":                    true, // broken
+		"listExpiredParticipationKeyTest.exp": true, // flaky
```
tsachiherman, line 118: I think it would be better to modify the expect test itself to return right away - so that the "disable" is on the test itself and we don't need to maintain "good test/bad test" lists.

algorandskiy: The problem with this expected does not prints out output when it is successful so there is no way to know it did not run. With the current approach tho skipped files do not appear in `go test -v` log. Plus I'll add logging here to emphasize it was skipped.

test/e2e-go/cli/goal/expect/goalNodeTest.exp
algorandskiy: done

test/framework/fixtures/expectFixture.go
tsachiherman: What if the expect would output a "Skip" output ? we're already reading the output in the go-code, so we can just call the t.Skip in that case.

tsachiherman:
@algorandskiy - it's great that you're enabling these tests.. unfortunately these still randomly fails ( even with your fixes ).
I'm going to try and re-run the failed test and try to merge this one ( since it's better than before ), but I still think that we should disable the failed tests and create a ticket to handle these. I want to make sure we're allocating enough time to ourselves for addressing these issues, rather than addressing low hanging fruits.

tsachiherman (approved):

PR #2782, opened 2021-08-20
Peer Scheduler Unit Tests

### LINKED ISSUE #2462: Txn sync: PeerScheduler unit test

AlgoStephenAkiki:
## Summary

Resolves #2462

Adds peer scheduler unit tests in peerscheduler_test.go

## Test Plan

These are tests.

### DIFF
--- a/txnsync/peerscheduler_test.go
+++ b/txnsync/peerscheduler_test.go
@@ -1,0 +1,190 @@
+// Copyright (C) 2019-2021 Algorand, Inc.
+// This file is part of go-algorand
+//
+// go-algorand is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as
+// published by the Free Software Foundation, either version 3 of the
+// License, or (at your option) any later version.
+//
+// go-algorand is distributed in the hope that it will be useful,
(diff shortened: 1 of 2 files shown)


txnsync/peerscheduler_test.go
```diff
@@ -0,0 +1,194 @@
...
+			lastSentMessageSequenceNumber: 123,
+		},
+		Peer{
+			lastSentMessageSequenceNumber: 456,
+		},
+		Peer{
+			lastSentMessageSequenceNumber: 789,
+		},
+	}
+
```
winder, line 87: nit: some comments would be helpful.

i.e.
// check that the duration is recorded
// ensure the duration has been cleared

And below
// ensure duplicate peers are removed

etc.

winder (approved):
Looks good. Just one nit about commenting the behavior. There is some subtle details to the way you constructed these tests that weren't obvious to me

brianolson (approved):

txnsync/peerscheduler_test.go
tsachiherman: I'm going to merge in this PR, but I agree with @winder's suggestion  - we should have more comments describing the purpose of the test.

We can add these comments in a separate follow-up PR.

PR #2783, opened 2021-08-20
per-client RAM measurement scripts

brianolson:
## Summary

new scripts runNodeHost.py and nodeHostTarget.py manage starting a new ec2 node with `algod` (or several) pointed at a specific relay so that we can test per-client usage on that one relay.

improve goal: don't crash if kmd isn't present when we don't need it

tweaks to other heapWatch scripts for measuring this data.

## Test Plan

This is manual testing code. I used it to run test clusters.

### DIFF
--- a/cmd/goal/commands.go
+++ b/cmd/goal/commands.go
@@ -371,17 +371,25 @@ func ensureFullClient(dataDir string) libgoal.Client {
 	return ensureGoalClient(dataDir, libgoal.FullClient)
 }
 
-func ensureGoalClient(dataDir string, clientType libgoal.ClientType) libgoal.Client {
+func getGoalClient(dataDir string, clientType libgoal.ClientType) (client libgoal.Client, err error) {
 	clientConfig := libgoal.ClientConfig{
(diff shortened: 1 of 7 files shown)


cmd/goal/commands.go
```diff
@@ -404,7 +412,10 @@ func getWalletHandleMaybePassword(dataDir string, walletName string, getPassword
 	var dup bool
 
 	accountList := makeAccountsList(dataDir)
-	kmd := ensureKmdClient(dataDir)
+	kmd, err := getGoalClient(dataDir, libgoal.KmdClient)
+	if err != nil {
+		return nil, nil, fmt.Errorf("kmd client init error: %v", err)
```
tsachiherman, line 417: Could you please use `%w` here so that we can traverse the original underlying error ?

tsachiherman (changes requested):
looks good - one small request in changing `%v` to `%w` so we can identify the original issue.

tsachiherman (approved):

PR #2785, opened 2021-08-22
testing: disable deadlock for netgoal

tsachiherman:
## Summary

When creating large networks with bootstrapped databases ( 60m accounts ), the netgoal was failing due to a false-positive deadlock detection. Given that the netgoal is a testing utility, it would be better to turn off these false-positives testing off netgoal.

## Test Plan

Tested manually.


algonautshant (approved):
Looks good.

PR #2789, opened 2021-08-23
Add txn field for stateless first valid block timestamp

### LINKED ISSUE #2352: Stateless code needs block time independent "time" checking.
Many LogicSigs would like to lock up funds for a time. Others need to "expire" (return false unconditionally) after some time.  This is complicated by the fact that stateless code must not realy on *any* state except what is in the sent transaction.

Today, LogicSigs examine their "fv" or "lv" fields, rather than as for a "current time".  They must then map a block number to the time they estimate such a block number would be added to the chain.  They do so by guessing that blocks will take ~4.5s to run, now and into the future.

With performance work that may decrease the block time, code like this needs a new strategy that is still stateless.  That is, a timestamp can't be expose to stateless code.

Strawman: Two new opcodes, `assert_before` and `assert_after`.  These opcodes examine `lv` (for `assert_before`) or `fv` (for `assert_after`) and fail the entire transaction if the transaction could take place after time T (taken from top of stack).

`assert_before` establishes a estimate for the time as of block `lv + 1`. It does this by taking the timestamp from the previous block b, and adding (lv+1-b)*(current_block_duration-safety_factor). If T exceeds that lower bound, the txn fails. safety_factor is intended to make a conservative estimate for lv+1's block time.  This is impossible in general, since block times could potentially be shortened in consensus upgrade. To be perfect, we would have to rerun logicsigs after upgrades.

`assert_after` is easier.  The check is simply that the `fv -1` block's timestamp is less than T.  If not, fail.

This all sounds more complex than one would expect.  I'd like to see something simpler that is still stateless, but I have not thought of it.

algochoi:
## Summary

This PR adds a new transaction field `FirstValidTime` in Teal V6 that retrieves the block timestamp of round `FirstValid-1` in stateless contracts. It keeps an extra block (at most `MaxTxnLife`) in the `txTail` to make sure rnd `FirstValid-1` block can be retrieved efficiently.  
* Changed catchup and `txTail` to store one extra block in memory
* Enable `FirstValidTime` field in stateless transactions

During account updates to blockdb rewinding `accountUpdatesLedgerEvaluator` simulates `TxTail` work on reading and caching timestamps, but does it lazely so applying transactions without `txn FirstValidTime` does not cause any lookups but allocates a map with (MaxTxnTail + 1) entries 8-bytes each.
This approach prevents introducing dependency between TxTail and AccountUpdates trackers and keeps `ledgerForTrackers` interface clean of lock/non-lock paired functions.

Closes #2352 

## Test Plan

* Added `txTail` test for getting block timestamps
* Added op test in `eval_test` 
* Added txTail and blockq tests into `eval_test`


data/transactions/logic/eval.go
```diff
@@ -1872,6 +1890,15 @@ func (cx *evalContext) txnFieldToStack(txn *transactions.Transaction, field TxnF
 		sv.Uint = txn.Fee.Raw
 	case FirstValid:
 		sv.Uint = uint64(txn.FirstValid)
+	case FirstValidTime:
+		// Check if the first valid time has already been set
+		if FirstValidTime == 0 {
```
algochoi, line 1895: Set the tx field here, if it has not been initialized already (i.e. `FirstValidTime` was not called already` - this field should not be set by the user anyway)

ledger/blockqueue.go
```diff
@@ -143,6 +143,11 @@ func (bq *blockQueue) syncer() {
 			bq.mu.Unlock()
 
 			minToSave := bq.l.notifyCommit(committed)
+			// Save one extra block in the queue for checking timestamps
```
algochoi, line 146: Set min to one round earlier to save the extra block

catchup/catchpointService.go
```diff
@@ -459,7 +459,7 @@ func (cs *CatchpointCatchupService) processStageBlocksDownload() (err error) {
 	}
 
 	// pick the lookback with the greater of either MaxTxnLife or MaxBalLookback
-	lookback := config.Consensus[topBlock.CurrentProtocol].MaxTxnLife
+	lookback := config.Consensus[topBlock.CurrentProtocol].MaxTxnLife + 1
```
algochoi, line 462: Set `lookback` to +1 to download an extra block during catchup

data/transactions/transaction.go
```diff
@@ -51,13 +51,14 @@ type SpecialAddresses struct {
...
-	Fee         basics.MicroAlgos `codec:"fee"`
-	FirstValid  basics.Round      `codec:"fv"`
-	LastValid   basics.Round      `codec:"lv"`
-	Note        []byte            `codec:"note,allocbound=config.MaxTxnNoteBytes"` // Uniqueness or app-level data about txn
-	GenesisID   string            `codec:"gen"`
-	GenesisHash crypto.Digest     `codec:"gh"`
+	Sender         basics.Address    `codec:"snd"`
+	Fee            basics.MicroAlgos `codec:"fee"`
+	FirstValid     basics.Round      `codec:"fv"`
+	FirstValidTime int64             `codec:"fvt"`
```
tsachiherman, line 57: Do you intend to have this field on all transactions ?

tsachiherman (changes requested):
We can't load the historical header on demand.. it would be far too slow. Look on the txtail as a "reference guide".

data/transactions/logic/eval.go
```diff
@@ -1863,6 +1864,26 @@ func (cx *evalContext) getTxID(txn *transactions.Transaction, groupIndex int) tr
 	return txid
 }
 
+func (cx *evalContext) getFirstValidTimestamp(r basics.Round) (timestamp uint64, err error) {
+	if cx.Ledger == nil {
+		err = fmt.Errorf("ledger not available")
```
tsachiherman, line 1869: Use errors.New instead, since you're not formatting anything here

data/transactions/logic/eval.go
```diff
@@ -1863,6 +1864,26 @@ func (cx *evalContext) getTxID(txn *transactions.Transaction, groupIndex int) tr
...
+		return
+	}
+	// Get the FirstValid-1 block timestamp
+	if r > 0 {
```
tsachiherman, line 1873: Use r.subsaturate

data/transactions/logic/eval.go
```diff
@@ -1863,6 +1864,26 @@ func (cx *evalContext) getTxID(txn *transactions.Transaction, groupIndex int) tr
...
+		r--
+	}
+	ts, err := cx.Ledger.GetBlockTimeStamp(r)
+	if ts < 0 {
+		err = fmt.Errorf("first valid timestamp %d < 0", ts)
+		return
+	}
+	if err != nil {
```
tsachiherman, line 1881: Test for the error first, then for the actual value.

ledger/archival_test.go
```diff
@@ -824,7 +824,8 @@ func checkTrackers(t *testing.T, wl *wrappedLedger, rnd basics.Round) (basics.Ro
 
 		// Special case: initAccounts reflects state after block 0,
 		// so it's OK to return minSave=0 but query block 1.
-		if minSave > wl.minQueriedBlock && minSave != 0 && wl.minQueriedBlock != 1 {
+		// We also store MaxTxnLife+1, so check for minQueriedBlock+1
+		if minSave > wl.minQueriedBlock+1 && minSave != 0 && wl.minQueriedBlock != 1 {
```
tsachiherman, line 828: I think you won't need this one

ledger/blockqueue.go
tsachiherman: You wont need this if the txtail would return -1  on committedupto

catchup/catchpointService.go
tsachiherman: It's ok - but please add a comment in code explaining why we do so.

tsachiherman (changes requested):
Few more comments.. overall, i think that you have now few testing places that doesn't require the -1 anymore.

Note that your change would not work on the first round.. but that's probably ok since the upgrade would take more than a single round.

data/transactions/logic/doc.go
```diff
@@ -325,7 +324,7 @@ var txnFieldDocs = map[string]string{
 	"Sender":                   "32 byte address",
 	"Fee":                      "micro-Algos",
 	"FirstValid":               "round number",
-	"FirstValidTime":           "Causes program to fail; reserved for future use",
+	"FirstValidTime":           "timestamp of FirstValid-1 block. Panics if used before Teal V5",
```
jannotti, line 327: Capitalize Timestamp. And I think we can skip the panic sentence because the autogenerated docs will mention LogicSigVersion >= 5.

data/transactions/logic/eval.go
```diff
@@ -1863,6 +1864,24 @@ func (cx *evalContext) getTxID(txn *transactions.Transaction, groupIndex int) tr
 	return txid
 }
 
+func (cx *evalContext) getFirstValidTimestamp(r basics.Round) (timestamp uint64, err error) {
+	if cx.Ledger == nil {
+		err = errors.New("ledger not available")
+		return
+	}
+	// Get the FirstValid-1 block timestamp
+	r.SubSaturate(1)
```
jannotti, line 1873: I'd rather fail if firstValid = 0.

data/transactions/logic/evalStateful_test.go
```diff
@@ -169,6 +169,10 @@ func (l *testLedger) LatestTimestamp() int64 {
 	return int64(rand.Uint32() + 1)
 }
 
+func (l *testLedger) GetBlockTimeStamp(r basics.Round) (int64, error) {
```
jannotti, line 172: Returning a random stamp seems hard to test. Maybe we should just return something like 4.5s * round + some start epoch.

That way we get monotone increases and testability.

data/txntest/txn.go
```diff
@@ -46,16 +46,17 @@ import (
...
-	Note        []byte
-	GenesisID   string
-	GenesisHash crypto.Digest
-	Group       crypto.Digest
-	Lease       [32]byte
-	RekeyTo     basics.Address
+	Sender         basics.Address
+	Fee            basics.MicroAlgos
+	FirstValid     basics.Round
+	FirstValidTime int64
```
jannotti, line 52: This still needs to go.

ledger/cow_test.go
```diff
@@ -42,6 +43,10 @@ func (ml *mockLedger) checkDup(firstValid, lastValid basics.Round, txn transacti
 	return nil
 }
 
+func (ml *mockLedger) getBlockTimeStamp(rnd basics.Round) (int64, error) {
```
jannotti, line 46: Similar testability issue.

jannotti (commented):
I think we're close. I need to look some more from computer. Reviews on phone aren't perfect.

algochoi:
Changed the test time stamps to be `rnd*5` without a particular start epoch. I also ran the benchmarks in the `ledger` and did not see any notable changes between the master branch.

data/transactions/logic/eval_test.go
```diff
@@ -1815,6 +1831,7 @@ int 2
 			}
 			ep := defaultEvalParams(nil, &txn)
 			ep.TxnGroup = makeSampleTxnGroup(txn)
+			ep.Ledger = makeTestLedger(nil)
```
jannotti, line 1834: I'm a little worried now.  If you had to add the ledger here, I don't know what's going to happen when this is used in a logicsig.  I don't know that the ledger is available there.  Do you have an e2e with a logicsig using this?

data/transactions/verify/txn.go
```diff
@@ -346,12 +353,21 @@ func logicSigBatchVerify(txn *transactions.SignedTxn, groupIndex int, groupCtx *
 	if groupIndex < 0 {
 		return errors.New("Negative groupIndex")
 	}
+
+	var ts int64
+	if groupCtx.consensusParams.EnableFirstValidTimeStamp {
```
tsachiherman, line 358: I'm not sure if this test is needed here, since the teal 5 won't be enabled anyway.

ledger/txtail.go
```diff
@@ -207,6 +207,11 @@ func (t *txTail) putLV(lastValid basics.Round, id transactions.Txid) {
 }
 
 func (t *txTail) getBlockTimeStamp(rnd basics.Round) (int64, error) {
+	// allow timestamps in a first block
+	if rnd == 0 {
+		return 0, nil
+	}
```
tsachiherman, line 213: that's fine, but consider moving this to after the 
```golang
       if r, ok := t.recent[rnd]; ok {
 		return r.timestamp, nil
 	}
```

for performance reasons. ( i.e. we would rarely get here with rnd==0, so we should test that later on. )

tsachiherman (commented):
I'm not sure if that's the best way to have this as a stateless teal, since we would end up performing the lookup regardless of whether this feature is used or not.

another approach to implement this is to generate a new map[round]timestamp for every round, and use that map as an evaluator parameter. The map would be examined only in the case the new opcode would be used.

data/transactions/verify/txn.go
algorandskiy: True, but I'm worried about any difference in behavior caused by calling `GetBlockTimeStamp` and erroring for whatever reason (retrieving of the timestamp happens for all programs)

ledger/txtail.go
algorandskiy: done

data/transactions/logic/eval.go
```diff
@@ -252,6 +251,9 @@ type EvalParams struct {
 
 	// Total pool of app call budget in a group transaction
 	PooledApplicationBudget *uint64
+
+	// A timestamp of the Txn's (FirstValid - 1) block
+	FirstValidTimestamp uint64
```
algochoi, line 256: Should timestamp be `uint64` or `int64`?

ledger/eval.go
```diff
@@ -676,7 +672,7 @@ func (eval *BlockEvaluator) TransactionGroup(txads []transactions.SignedTxnWithA
 
 // prepareEvalParams creates a logic.EvalParams for each ApplicationCall
 // transaction in the group
-func (eval *BlockEvaluator) prepareEvalParams(txgroup []transactions.SignedTxnWithAD) []*logic.EvalParams {
+func (eval *BlockEvaluator) prepareEvalParams(txgroup []transactions.SignedTxnWithAD) ([]*logic.EvalParams, error) {
```
algochoi, line 675: My initial understanding was that `prepareEvalParams` will not initialize `EvalParams` for a stateless call (it will just `continue` in L685 below). Is this still true?

algorandskiy: yes, check data/transactions/verify for logicsig flow

data/transactions/logic/eval.go
tsachiherman: All our teal code is `uint64`, so I'd think we would want to keep this as uint64. Plus, all our timestamps are always positive numbers.

data/transactions/logic/fields.go
```diff
@@ -195,7 +195,10 @@ var txnFieldSpecs = []txnFieldSpec{
 	{Sender, StackBytes, 0, 5, false},
 	{Fee, StackUint64, 0, 5, false},
 	{FirstValid, StackUint64, 0, 0, false},
-	{FirstValidTime, StackUint64, 0, 0, false},
+	// FirstValidTime was originally a dummy field in v0, and enabled in v6.
+	// Setting version to 6 will cause assembling programs with version 0-5 to fail
+	// but it does not matter since FirstValidTime is invalid runtime field on TEAL v0-v5
+	{FirstValidTime, StackUint64, 6, 6, false},
```
tsachiherman, line 201: nit: move this to the end of this slice, so that element here would be grouped by the teal logic ver.

ledger/evalIndexer.go
```diff
@@ -134,6 +134,10 @@ func (l indexerLedgerConnector) CompactCertVoters(_ basics.Round) (*VotersForRou
 	return nil, errors.New("CompactCertVoters() not implemented")
 }
 
+func (l indexerLedgerConnector) BlockTimeStamp(r basics.Round) (int64, error) {
+	return l.BlockTimeStamp(r)
```
tsachiherman, line 138: infinite recursive call

algorandskiy: Added `BlockTimeStamp` into `indexerLedgerForEval` interface. @tolikzinovyev  FYI

data/transactions/logic/fields.go
algorandskiy: This would work since there is some internal logic requires that field name values matched to positions in `txnFieldSpecs`. Added blank lines around to make it visible.

config/consensus.go
```diff
@@ -391,6 +391,9 @@ type ConsensusParams struct {
 	EnableKeyregCoherencyCheck bool
 
 	EnableExtraPagesOnAppUpdate bool
+
+	// Allow querying txn.FirstValid - 1 timestamp
+	EnableFirstValidTimeStamp bool
```
jannotti, line 396: Does this need to be separate from enabling the field in Teal 6?

data/transactions/logic/eval_test.go
```diff
@@ -1821,13 +1836,21 @@ gtxnsas Accounts
 gtxn 0 Sender
 ==
 &&
+`
+
+	gtxnTextV6 := gtxnTextV5 + `
+gtxn 0 FirstValidTime
+int 0
+>
```
jannotti, line 1844: Why not test the exact value since you're setting it explicitly in ep?

data/transactions/logic/fields.go
```diff
@@ -195,7 +195,12 @@ var txnFieldSpecs = []txnFieldSpec{
 	{Sender, StackBytes, 0, 5, false},
 	{Fee, StackUint64, 0, 5, false},
 	{FirstValid, StackUint64, 0, 0, false},
-	{FirstValidTime, StackUint64, 0, 0, false},
+
+	// FirstValidTime was originally a dummy field in , and enabled in v6.
+	// Setting version to 6 will cause assembling programs with version 0-5 to fail
+	// but it does not matter since FirstValidTime is invalid runtime field on TEAL v0-v5
+	{FirstValidTime, StackUint64, 6, 6, false},
```
jannotti, line 202: The second 6 should be 0.  We don't want this field settable by `itxn_field`

data/transactions/verify/txn.go
```diff
@@ -346,12 +353,21 @@ func logicSigBatchVerify(txn *transactions.SignedTxn, groupIndex int, groupCtx *
...
+		ts, err = groupCtx.timestamper.BlockTimeStamp(txn.Txn.FirstValid.SubSaturate(1))
+		if err != nil {
+			return err
+		}
```
jannotti, line 362: I don't think it should subsaturate, it should error if FirstValid = 0.  You are asking for the timestamp of block -1, which doesn't exist.  Perhaps it could return the timestamp in the genesis file, but that seems too cute.

ledger/acctupdates.go
```diff
@@ -888,6 +888,13 @@ type accountUpdatesLedgerEvaluator struct {
 	// prevHeader is the previous header to the current one. The usage of this is only in the context of initializeCaches where we iteratively
 	// building the ledgercore.StateDelta, which requires a peek on the "previous" header information.
 	prevHeader bookkeeping.BlockHeader
+	// blockTimeStamps contains block timestamps needed for txn evaluation while catching account db round up to block db round.
+	// in this mode au uses accountUpdatesLedgerEvaluator as a ledger, that is enough because balances are available in the underlying accountdb.
+	// unfortunately this does not work anymore after allowing smart contact accessing block header timestamps up to MaxTxnLife.
+	// block db and TxTail has required info (TxTail tracker provides timestamps in a normal evaluation mode)
+	// but calling TxTail from accountUpdatesLedgerEvaluator introduces trackers' dependency and locking problems.
+	// instead, caching timestamps in the same manner as TxTail prevents dependency and ledgerForTracker interface pollution.
+	blockTimeStamps map[basics.Round]int64
```
jannotti, line 897: I don't understand everything here perfectly, but it seems like info in txtail, like leases, is just as important for transaction processing.  So why is txtail "good enough" with respect to leases, but won't have the info we need for timestamps?

ledger/eval.go
```diff
@@ -701,6 +702,15 @@ func (eval *BlockEvaluator) prepareEvalParams(txgroup []transactions.SignedTxnWi
 			// intentionally ignoring error here, fees had to have been enough to get here
 		}
 
+		var ts int64
+		var err error
+		if eval.proto.EnableFirstValidTimeStamp {
+			ts, err = eval.l.BlockTimeStamp(txn.Txn.FirstValid.SubSaturate(1))
+			if err != nil {
```
jannotti, line 709: Same thoughts about SubSaturate.

jannotti: And another thought.  I think we don't ever want to error here. Instead, could we put a 0 in the timestamp field, and error on access?  That way, we can handle txns with FirstValid=0, but error if they ask for the timestamp.

jannotti (commented):
I basically have two thoughts.
a) Can we make it an error to access FirstValidTimestamp if FirstValid=0?  But move the error to access time, not to evalparam prep time.
b) Why do we have to handle this differently than leases in txtail?

data/transactions/logic/eval_test.go
```diff
@@ -999,6 +999,10 @@ byte 0x0706000000000000000000000000000000000000000000000000000000000000
 &&
 `
 
+const globalV6TestProgram = globalV5TestProgram + `
+// No new globals in v4
```
algochoi, line 1003: nit: v6?

data/transactions/logic/fields.go
```diff
@@ -195,7 +195,12 @@ var txnFieldSpecs = []txnFieldSpec{
 	{Sender, StackBytes, 0, 5, false},
 	{Fee, StackUint64, 0, 5, false},
 	{FirstValid, StackUint64, 0, 0, false},
-	{FirstValidTime, StackUint64, 0, 0, false},
+
+	// FirstValidTime was originally a dummy field in , and enabled in v6.
```
algochoi, line 199: I think the comment should have been:

`FirstValidTime was originally a dummy field in v0-v5, and enabled in v6.`

ledger/evalIndexer.go
```diff
@@ -45,6 +45,8 @@ type indexerLedgerForEval interface {
 	GetAssetCreator(map[basics.AssetIndex]struct{}) (map[basics.AssetIndex]FoundAddress, error)
 	GetAppCreator(map[basics.AppIndex]struct{}) (map[basics.AppIndex]FoundAddress, error)
 	Totals() (ledgercore.AccountTotals, error)
+	// BlockTimeStamp returns timestamp for a block with possible range [TxnMaxLife - 1, Latest]
```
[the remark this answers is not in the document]
algorandskiy: correct

ledger/eval.go
jannotti: And, if we're worried about using 0 as a flag value (since I suppose it *could* be the timestamp of the first block), we could error on access by explicitly checking firstValid first, and error if it's zero, not if this value is zero.

ledger/acctupdates.go
algorandskiy: these are used in different code paths: leases, txn ids and timestamps are by evaluator that has entire Ledger with all trackers. This one is only used for "catching" up balances to the same round as block DB in order to accept new transactions. It is done within account updates tracker, and leases/ids do not matter since they come from block DB that are already persisted in blockchain. But now evaluation requires block timestamps.

config/consensus.go
algorandskiy: I'm a bit worried on erring on EvalParam preparation, and we have not accessed old blocks there before. It should work but seemed to be better on a safe side. I like the idea if returning zero and erring in runtime instead of prep time

algochoi (commented):
Should we have a stronger test for getting the timestamp from ledger than setting the `ep` manually and checking for their fields?

config/consensus.go
algorandskiy: removed, not needed anymore since `BlockTimeStamp` does not err anymore

ledger/eval.go
algorandskiy: err on access in `txn FirstValidTime`

data/transactions/verify/txn.go
algorandskiy: BlockTimeStamp does not err anymore, deferring resolution to logic.Eval

algorandskiy:
All reviews fixes. Added pedantic `TestFistValidTime` test into **eval_test.go**

data/transactions/logic/fields.go
```diff
@@ -35,7 +35,7 @@ const (
 	Fee
 	// FirstValid Transaction.FirstValid
 	FirstValid
-	// FirstValidTime panic
+	// FirstValidTime Transaction.FirstValidTime
```
jannotti, line 38: This doesn't really serve to document the constant, because FirstValidTime isn't really a field of Transaction.

ledger/acctupdates_test.go
```diff
@@ -172,6 +172,14 @@ func (ml *mockLedgerForTracker) BlockHdr(rnd basics.Round) (bookkeeping.BlockHea
 	return ml.blocks[int(rnd)].block.BlockHeader, nil
 }
 
+func (ml *mockLedgerForTracker) BlockTimeStamp(rnd basics.Round) int64 {
+	if rnd > ml.Latest() {
```
jannotti, line 176: Is Latest() the block being worked on, or the last block committed?  If being worked on, then we want >=.  (I have no idea, just double checking.)

ledger/eval.go
```diff
@@ -739,9 +742,10 @@ func (eval *BlockEvaluator) prepareEvalParams(txgroup []transactions.SignedTxnWi
 			PooledApplicationBudget: &pooledApplicationBudget,
 			FeeCredit:               &credit,
 			Specials:                &eval.specials,
+			FirstValidTimestamp:     uint64(ts),
 		}
 	}
-	return res
+	return res, nil
```
jannotti, line 748: I think you've removed the need to return error, so let's go back to the old signature.

ledger/evalIndexer.go
```diff
@@ -39,6 +39,8 @@ type indexerLedgerForEval interface {
 	GetAssetCreator(map[basics.AssetIndex]struct{}) (map[basics.AssetIndex]FoundAddress, error)
 	GetAppCreator(map[basics.AppIndex]struct{}) (map[basics.AppIndex]FoundAddress, error)
 	Totals() (ledgercore.AccountTotals, error)
+	// BlockTimeStamp returns timestamp for a block up to TxnMaxLife-1 back from Latest, or 0 if not exist
```
jannotti, line 42: Comment should say "+1" since it's talking about going "back".

ledger/eval.go
algorandskiy: thanks, overlooked that,

ledger/evalIndexer.go
algorandskiy: done

ledger/acctupdates_test.go
algorandskiy: Latest() comes from ledger so it is the last committed.

data/transactions/logic/fields.go
algorandskiy: thanks, fixed

algorandskiy (commented):
Thanks for catching these, fixed.

algorandskiy:
@tsachiherman could you take a final look?

jannotti (approved):
Looks good to me now. @tsachiherman can you have one more look?

data/transactions/logic/doc.go
```diff
@@ -372,7 +371,7 @@ var txnFieldDocs = map[string]string{
 	"Sender":         "32 byte address",
 	"Fee":            "micro-Algos",
 	"FirstValid":     "round number",
-	"FirstValidTime": "Causes program to fail; reserved for future use",
+	"FirstValidTime": "Timestamp recorded on the block header at round FirstValid-1.",
```
tsachiherman, line 374: nit: remove the "." at the end of the line. I'm not particular whether we'll have the "." or not, but we should be consistant.

data/transactions/logic/eval.go
```diff
@@ -1946,6 +1949,11 @@ func (cx *EvalContext) txnFieldToStack(txn *transactions.Transaction, fs txnFiel
 		sv.Uint = txn.Fee.Raw
 	case FirstValid:
 		sv.Uint = uint64(txn.FirstValid)
+	case FirstValidTime:
+		if cx.EvalParams.FirstValidTimestamp == 0 {
+			return sv, fmt.Errorf("unable to obtain timestamp for %d", txn.FirstValid-1)
```
tsachiherman, line 1954: nit: 
- `unable to obtain timestamp for %d` -> `unable to obtain timestamp for round %d`
- instead of `txn.FirstValid`, use txn.FirstValid.Subsaturate(1)

algochoi (commented):
Thanks for working on this -- the ledger test looks quite nice 👏

ledger/acctupdates.go
tsachiherman: The above logical distinction might suggest that we should maintain these on separate trackers. But we can refactor that on a separate PR.

data/transactions/logic/doc.go
algorandskiy: Done. The code in opdoc aware of both dot/no dot endings.

data/transactions/logic/eval.go
algorandskiy: done

ledger/acctupdates.go
```diff
@@ -920,6 +927,16 @@ func (aul *accountUpdatesLedgerEvaluator) CheckDup(config.ConsensusParams, basic
 	return fmt.Errorf("accountUpdatesLedgerEvaluator: tried to check for dup during accountUpdates initialization ")
 }
 
+// BlockTimeStamp gets the block timestamp of a specific round.
+func (aul *accountUpdatesLedgerEvaluator) BlockTimeStamp(rnd basics.Round) int64 {
+	if ts, ok := aul.blockTimeStamps[rnd]; ok {
+		return ts
+	}
+	hdr, _ := aul.au.ledger.BlockHdr(rnd)
```
tsachiherman, line 935: it's ok that you don't want to test the error here... but you'll need to preempt that test and actively verify that the blocks database truly contain <first round> - 1001. This is part of the startup code, so we need to test inconsistencies between the two databases.

algorandskiy: make sense - will add a pre-check and test it

tsachiherman: this would also be a good place to potentially delete old entries, in case we're catching up more than 1000 rounds.
i.e.
```golang
delete(aul.blockTimeStamps, rnd - 1001)
```

ledger/acctupdates.go
```diff
@@ -1157,6 +1182,9 @@ func (au *accountUpdates) initializeFromDisk(l ledgerForTracker) (lastBalancesRo
 	}
 
 	au.accountsq, err = accountsDbInit(au.dbs.Rdb.Handle, au.dbs.Wdb.Handle)
+	if err != nil {
```
tsachiherman, line 1185: nice

ledger/eval.go
```diff
@@ -1256,7 +1260,7 @@ func eval(ctx context.Context, l ledgerForEvaluator, blk bookkeeping.Block, vali
 		txvalidator.ctx = validationCtx
 		txvalidator.txgroups = paysetgroups
 		txvalidator.done = make(chan error, 1)
-		go txvalidator.run()
+		go txvalidator.run(l)
```
tsachiherman, line 1263: kind of weird that it's the only argument passed as a function parameter.. but it's legit.

algorandskiy: good catch, moved to the `txvalidator` construction code just above

ledger/acctupdates.go
algorandskiy: this actually ended up to be changes in `au.committedUpTo`: when blockdb has 1005 rounds, `txTail` says `4 = 1005 - 1001` and `au` says something like `1005 - 320 - X` where X relatively small number of not committed yet offsets. But in fact block 999 might have `txn FirstValidTime` with `FirstValid=1` (or 0), and blockdb does not have such round. That means for any round R: `au.dbRound < R <= Latest` blockdb must have a block for round `R - (MaxTxnLife + 1)`. This is 320 + X rounds more than `txTail` :(

```
               au.dbRound                       Latest
                     |  X  |    Lookback (320)    |
---------------------+-----+----------------------+
```

algorandskiy: Added test and checking code

PR #2791, opened 2021-08-23
IncomingConnectionsLimit -> 800

brianolson:
## Summary

Change incoming connection limit default to 800, which might use ~4GB of RAM given our measured estimate of 5MB per peer.
Prior peer RAM overhead may have been as much as 2MB per peer (but never before actually measured), so the old limit of 10,000 might have been too optimistic.

## Test Plan

Manual testing found this guideline of 5MB per peer.


tsachiherman:
Use 'make generate' to regenerate the needed files.

PR #2794, opened 2021-08-24
bulletproof xor bloom filter against invalid and malicious inputs

brianolson:
## Summary

Tweaks to xor/bloom code based on https://github.com/algorand/go-algorand/pull/2718

## Test Plan

Continues to pass unit tests and gets ok code coverage


util/bloom/xor.go
```diff
@@ -93,28 +94,37 @@ func (xf *XorFilter) MarshalBinary() ([]byte, error) {
 // ErrBadBinary is returned when UnmarshalBinary fails
 var ErrBadBinary = errors.New("bad XorFilter binary")
 
+// TODO: make this an option to UnmarshalBinary, or a settable global, ...
+const maxFingerprints = 1000000
```
tsachiherman, line 98: I agree with the TODO. We can do that later on. Sizing also seems to be reasonable for what we need.

util/bloom/xor.go
```diff
@@ -93,28 +94,37 @@ func (xf *XorFilter) MarshalBinary() ([]byte, error) {
...
-	if dp < 0 {
+	if dp <= 0 {
 		return ErrBadBinary
 	}
 	pos += dp
 	if lenFingerprints > 0 {
+		if lenFingerprints > maxFingerprints {
+			return ErrBadBinary
+		}
+		if (lenFingerprints * 4) > uint64(len(data)-pos) {
```
tsachiherman, line 125: instead of 4, use `unsafe. SizeUint32`

brianolson: seems strange to refer to `unsafe` for a universal constant ("int32 is 4 bytes")

tsachiherman: If you can find other location that makes the same constant - that would work as well.

tsachiherman:
Could you please review the failures ?

tsachiherman (approved):
looks good.

brianolson:
the one test that failed circleci passes locally. txnsync/msgbuffers_test.go/TestTxIDSlicePool

tsachiherman:
> the one test that failed circleci passes locally. txnsync/msgbuffers_test.go/TestTxIDSlicePool

I'll fix that test.

tsachiherman:
with these changes, we will need some better handling for empty bloom filters on the txnsync; I'll make the minimal changes needed and attempt to push these in here.

PR #2798, opened 2021-08-24
Updating to the latest stable version of libsodium

MatusKysel:
## Summary

Updating to the latest stable (1.18) version of libsodium could improve overall performance of VRF as there have been changes to  the code that is used by VRF e.g. :
https://github.com/jedisct1/libsodium/commit/4967aa8f23e1b6ea268acc11f936430fbc2d2a87
https://github.com/jedisct1/libsodium/commit/88c568a03522cda6c54fd4b5c76cfbbc52a744f4

## Test Plan

as there are no changes to go code, I have just ran make test and all crypto related test passed


tsachiherman:
Thanks for this PR. Unfortunatally, we already have some effort in place that would result in eliminating the libsodium library completly, and therefore it would not make sense to update it now and remove it later on.

PR #2807, opened 2021-08-25
Add ABI encoding support

### LINKED ISSUE #2348: Add ABI encoding support to goal
## Summary
Add support for encoding and decoding ABI values, as described in https://github.com/algorandfoundation/ARCs/pull/7

## Scope
Add ability to:
* Parse and serialize ABI type strings, e.g. `(uint64, byte[], address)`
* Encode and decode ABI values, using type information

## Dependencies
This has follow-up issues:
* https://github.com/algorand/go-algorand/issues/2803
* https://github.com/algorand/go-algorand-sdk/issues/236
* https://github.com/algorand/js-algorand-sdk/issues/425
* https://github.com/algorand/py-algorand-sdk/issues/228
* https://github.com/algorand/java-algorand-sdk/issues/244

ahangsu:
## Summary

This PR provides a prototype implementation that closes #2348:
- Encoding/Decoding method for ABI Values.
- Serialization/Deserialization of ABI Types.
- Make/Get methods for ABI Types/Values.
- Basic testcases for Encoding/Decoding.

## Test Plan

For both ABI Type serialization/deserialization and ABI Value Encoding/Decoding:
- Valid test cases are provided to see if it returns expected values.
- Invalid test cases are also provided to see if it can catch error and return err.


cmd/goal/abi/abi.go
```diff
@@ -0,0 +1,1193 @@
...
+		return ufixedTypeRes, nil
+	case str == "bool":
+		return MakeBoolType(), nil
+	case str == "address":
+		return MakeAddressType(), nil
+	case str == "string":
+		return MakeStringType(), nil
+	case len(str) > 2 && str[0] == '(' && str[len(str)-1] == ')':
+		if strings.Contains(str[1:len(str)-1], " ") {
+			return Type{}, fmt.Errorf("tuple should not contain space")
```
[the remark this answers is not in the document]
ahangsu: it's redundant, can be removed

cmd/goal/abi/abi.go
```diff
@@ -0,0 +1,1193 @@
...
+		if err != nil {
+			return Type{}, err
+		}
+		return MakeStaticArrayType(arrayType, uint16(arrayLength)), nil
+	case strings.HasPrefix(str, "uint"):
+		typeSize, err := strconv.ParseUint(str[4:], 10, 16)
+		if err != nil {
+			return Type{}, fmt.Errorf("ill formed uint type: %s", str)
+		}
+		uintTypeRes, err := MakeUintType(uint16(typeSize))
```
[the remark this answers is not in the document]
ahangsu: fixed

cmd/goal/abi/abi.go
```diff
@@ -0,0 +1,1193 @@
...
+		arrayLength, err := strconv.ParseUint(arrayLengthStr, 10, 32)
+		if err != nil {
+			return Type{}, err
+		}
+		// parse the array element type
+		arrayType, err := TypeFromString(str[:len(str)-(2+len(arrayLengthStr))])
+		if err != nil {
+			return Type{}, err
+		}
+		return MakeStaticArrayType(arrayType, uint16(arrayLength)), nil
```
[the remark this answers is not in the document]
ahangsu: changed the previous `arrayLength` parsing to 16 bits, and throw error if it exceed `uint16` scope.

cmd/goal/abi/abi.go
```diff
@@ -0,0 +1,1193 @@
...
+			return Type{}, fmt.Errorf("static array ill formated: %s", str)
+		}
+		// guaranteed that the length of array is existing
+		arrayLengthStr := stringMatches[1]
+		arrayLength, err := strconv.ParseUint(arrayLengthStr, 10, 32)
+		if err != nil {
+			return Type{}, err
+		}
+		// parse the array element type
+		arrayType, err := TypeFromString(str[:len(str)-(2+len(arrayLengthStr))])
```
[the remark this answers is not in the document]
ahangsu: Can be done. Fixed.

cmd/goal/abi/abi.go
```diff
@@ -0,0 +1,1193 @@
...
+// TypeFromString de-serialize ABI type from a string following ABI encoding.
+func TypeFromString(str string) (Type, error) {
+	switch {
+	case strings.HasSuffix(str, "[]"):
+		arrayArgType, err := TypeFromString(str[:len(str)-2])
+		if err != nil {
+			return arrayArgType, err
+		}
+		return MakeDynamicArrayType(arrayArgType), nil
+	case strings.HasSuffix(str, "]") && len(str) >= 2 && unicode.IsDigit(rune(str[len(str)-2])):
```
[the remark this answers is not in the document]
ahangsu: fixed

cmd/goal/abi/abi.go
```diff
@@ -0,0 +1,1193 @@
...
+	}
+
+	return segments, nil
+}
+
+// MakeUintType makes `Uint` ABI type by taking a type size argument.
+// The range of type size is [8, 512] and type size % 8 == 0.
+func MakeUintType(typeSize uint16) (Type, error) {
+	if typeSize%8 != 0 || typeSize < 8 || typeSize > 512 {
+		return Type{}, fmt.Errorf("type uint size mod 8 = 0, range [8, 512], error typesize: %d", typeSize)
```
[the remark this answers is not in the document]
ahangsu: fixed

cmd/goal/abi/abi.go
```diff
@@ -0,0 +1,1193 @@
...
+		typeFromEnum: String,
+	}
+}
+
+// MakeTupleType makes tuple ABI type by taking an array of tuple element types as argument.
+func MakeTupleType(argumentTypes []Type) Type {
+	return Type{
+		typeFromEnum: Tuple,
+		childTypes:   argumentTypes,
+		staticLength: uint16(len(argumentTypes)),
```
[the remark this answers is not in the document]
ahangsu: nice catch, fixed.

cmd/goal/abi/abi.go
```diff
@@ -0,0 +1,1193 @@
...
+	case Ufixed:
+		ufixedValue, err := GetUfixed(v)
+		if err != nil {
+			return []byte{}, err
+		}
+		denomSize := big.NewInt(1).Exp(big.NewInt(10), big.NewInt(int64(v.valueType.typePrecision)), nil)
+		denomRat := big.NewRat(1, 1).SetFrac(denomSize, big.NewInt(1))
+		numRat := denomRat.Mul(denomRat, ufixedValue)
+		encodeVal := numRat.Num()
+		encodeBuffer := encodeVal.Bytes()
```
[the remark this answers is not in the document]
ahangsu: I don't think we are going to use `big.Rat` in this package anymore, the `ufixed` representation is done by `big.Int`, so the complex calculation can be discarded.

cmd/goal/abi/abi.go
```diff
@@ -0,0 +1,1193 @@
...
+	}
+	if index >= value.valueType.staticLength {
+		return Value{}, fmt.Errorf("static array cannot get element: index out of scope")
+	}
+	elements := value.value.([]Value)
+	return elements[index], nil
+}
+
+// GetTupleByIndex takes an index and tries to retrieve the eleemnt ABI Value.
+func GetTupleByIndex(value Value, index uint16) (Value, error) {
```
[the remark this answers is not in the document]
ahangsu: Fixed

cmd/goal/abi/abi.go
```diff
@@ -0,0 +1,1193 @@
...
+	return Type{
+		typeFromEnum: Tuple,
+		childTypes:   argumentTypes,
+		staticLength: uint16(len(argumentTypes)),
+	}
+}
+
+// Equal method decides the equality of two types: t == t0.
+func (t Type) Equal(t0 Type) bool {
+	// assume t and t0 are well-formed
```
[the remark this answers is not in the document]
ahangsu: Yea, it looks more succinct and maybe easier to debug.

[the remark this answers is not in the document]
ahangsu: Fixed

cmd/goal/abi/abi.go
```diff
@@ -0,0 +1,1193 @@
...
+	}
+	return Value{
+		valueType: MakeStaticArrayType(elemType, uint16(len(values))),
+		value:     values,
+	}, nil
+}
+
+// MakeTuple takes an array of ABI values and an array of ABI types,
+// and returns an ABI tuple value.
+func MakeTuple(values []Value, tupleType []Type) (Value, error) {
```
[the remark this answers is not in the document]
ahangsu: I think so long as we are passing non-empty array/tuple to make function, we can definitely get away with additional type argument. I specified that the input should have length within `[1, 2^16)`, or it will throw error.

cmd/goal/abi/abi.go
```diff
@@ -0,0 +1,1193 @@
...
+// MakeBool takes a boolean value and returns an ABI bool value.
+func MakeBool(value bool) Value {
+	return Value{
+		valueType: MakeBoolType(),
+		value:     value,
+	}
+}
+
+// GetUint8 tries to retreve an uint8 from an ABI Value.
+func GetUint8(value Value) (uint8, error) {
```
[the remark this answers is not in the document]
ahangsu: good one. fixed.

cmd/goal/abi/abi.go
```diff
@@ -0,0 +1,1193 @@
...
+	return Value{
+		valueType: MakeAddressType(),
+		value:     value,
+	}
+}
+
+// MakeDynamicArray takes an array of ABI value of elemType,
+// and returns an ABI dynamic length array value.
+func MakeDynamicArray(values []Value, elemType Type) (Value, error) {
+	for i := 0; i < len(values); i++ {
```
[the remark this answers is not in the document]
ahangsu: I don't think accepting 0 length array is something we want in our code, though I thought about that for a while. Now the 0 length array are eliminated in make function in this package.

cmd/goal/abi/abi.go
```diff
@@ -0,0 +1,1193 @@
...
+	}
+	return Value{
+		valueType: typeUint,
+		value:     value,
+	}, nil
+}
+
+// MakeUfixed takes a big rational number representation, a type size, and a type precision,
+// and returns an ABI Value of ABI UFixed<size>x<precision>
+func MakeUfixed(value *big.Rat, size uint16, precision uint16) (Value, error) {
```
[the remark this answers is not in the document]
ahangsu: It might be an overkill to do representation of `ufixed` with `big.Rat` in this case. Although it is straightforward to consider `ufixed` with rational representation like `p / q` (where `p, q \in Nat`), but this incurs unnecessary computational cost.

It is already switched to `big.Int` to store only the numerator of `ufixed` value, the precision information is held by its ABI type.

cmd/goal/abi/abi.go
```diff
@@ -0,0 +1,1193 @@
...
+	headLength := 0
+	for _, headTi := range heads {
+		headLength += len(headTi)
+	}
+
+	tailCurrLength := 0
+	for i := 0; i < len(heads); i++ {
+		if isDynamicIndex[i] {
+			headValue := headLength + tailCurrLength
+			binary.BigEndian.PutUint16(heads[i], uint16(headValue))
```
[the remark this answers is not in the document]
ahangsu: fixed.

cmd/goal/abi/abi.go
```diff
@@ -0,0 +1,1193 @@
...
+func MakeByteType() Type {
+	return Type{
+		typeFromEnum: Byte,
+	}
+}
+
+// MakeUFixedType makes `UFixed` ABI type by taking type size and type precision as arguments.
+// The range of type size is [8, 512] and type size % 8 == 0.
+// The range of type precision is [1, 160].
+func MakeUFixedType(typeSize uint16, typePrecision uint16) (Type, error) {
```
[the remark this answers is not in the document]
ahangsu: changed from `UFixed` to `Ufixed`

cmd/goal/abi/abi.go
```diff
@@ -0,0 +1,1193 @@
...
+		return false
+	default:
+		return false
+	}
+}
+
+// ByteLen method calculates the byte length of a static ABI type.
+func (t Type) ByteLen() (int, error) {
+	if t.IsDynamic() {
+		return -1, fmt.Errorf("dynamic type")
```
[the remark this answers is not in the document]
ahangsu: true, I think we can only take care of the case we need, and the rest are naturally eliminated in default case.

cmd/goal/abi/abi.go
```diff
@@ -0,0 +1,1193 @@
...
+	case Uint, Ufixed:
+		return int(t.typeSize / 8), nil
+	case Bool:
+		return 1, nil
+	case ArrayStatic:
+		elemByteLen, err := t.childTypes[0].ByteLen()
+		if err != nil {
+			return -1, err
+		}
+		return int(t.staticLength) * elemByteLen, nil
```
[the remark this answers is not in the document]
ahangsu: I copied some code from `Encode` function to take care of the `Bool` case. The counterexample you suggested is correct, previous version of code will overlook the boolean compression and return wrong result.

cmd/goal/abi/abi.go
```diff
@@ -0,0 +1,1193 @@
...
+
+// Type is the struct that stores information about an ABI value's type.
+type Type struct {
+	typeFromEnum BaseType
+	childTypes   []Type
+
+	// only can be applied to `uint` size <N> or `ufixed` size <N>
+	typeSize uint16
+	// only can be applied to `ufixed` precision <M>
+	typePrecision uint16
```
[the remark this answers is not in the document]
ahangsu: fixed

cmd/goal/abi/abi.go
```diff
@@ -0,0 +1,1193 @@
...
+	case String:
+		return "string"
+	case Tuple:
+		typeStrings := make([]string, len(t.childTypes))
+		for i := 0; i < len(t.childTypes); i++ {
+			typeStrings[i] = t.childTypes[i].String()
+		}
+		return "(" + strings.Join(typeStrings, ",") + ")"
+	default:
+		return "Bruh you should not be here"
```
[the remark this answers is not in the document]
ahangsu: fixed

cmd/goal/abi/abi.go
```diff
@@ -0,0 +1,1193 @@
...
+			tupleTypes[i] = ti
+		}
+		return MakeTupleType(tupleTypes), nil
+	default:
+		return Type{}, fmt.Errorf("cannot convert a string %s to an ABI type", str)
+	}
+}
+
+// segmentIndex keeps track of the start and end of a segment in a string.
+type segmentIndex struct{ left, right int }
```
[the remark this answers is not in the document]
ahangsu: fixed

cmd/goal/abi/abi.go
```diff
@@ -0,0 +1,1193 @@
...
+		return size, nil
+	default:
+		return -1, fmt.Errorf("bruh you should not be here")
+	}
+}
+
+// Value struct is the ABI Value, holding ABI Type information and the ABI value representation.
+type Value struct {
+	valueType Type
+	value     interface{}
```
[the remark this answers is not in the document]
ahangsu: I think the most proper way to do now is to make `Type` field public, and keep `value interface{}` private.

We now have `getter` functions on `Value`. I do not want other people directly operate on `value interface{}`.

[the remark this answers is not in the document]
jannotti: I'm ok with that.

Roughly speaking, I advocate for more public stuff than some people like.  I especially think simple types that just contain values (not pointers, which can lead to surprising sharing) should just have public fields most of the time.

cmd/goal/abi/abi.go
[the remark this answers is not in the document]
jannotti: I think for dynamic arrays, 0 length makes sense. A method might take a list of something, but some callers call with an empty list.

I can't find a reason for a statically sized 0 length array though.

data/abi/abi_encode_test.go
```diff
@@ -0,0 +1,1016 @@
...
+				},
+				{
+					enumIndex:  ArrayDynamic,
+					childTypes: []Type{MakeBoolType()},
+				},
+			},
+		})
+		require.NoError(t, err, "decode tuple for static/dynamic bool array should not return error")
+		require.Equal(t, tupleVal, decoded, "decoded tuple value do not match with expected")
+	})
```
ahangsu, line 689: Given the current ABI implementation, it seems hard to make *concise* test case (i.e., define a test in a few lines), since it is tedious to make an ABI value, or an ABI type (because of the `err` return).

I added more comments, trying to explain away the attempt and mechanism behind the test.

data/abi/abi_encode_test.go
```diff
@@ -0,0 +1,1016 @@
...
+		for i := 0; i < 10; i++ {
+			utf8Str := gobberish.GenerateString(length)
+			strValue := MakeString(utf8Str)
+			strEncode, err := strValue.Encode()
+			require.NoError(t, err, "string encode fail")
+			strDecode, err := Decode(strEncode, MakeStringType())
+			require.NoError(t, err, "decoding string should not return error")
+			require.Equal(t, strValue, strDecode, "encode string not match with expected")
+		}
+	}
```
ahangsu, line 458: For some of the value encoding/decoding, if the value is suitable for iteration (e.g., `address`, `uint`, `ufixed`, `byte`, `bool`), I go through iteration (or randomly take some test point in the valid space) to check if my implementation is correct.

data/abi/abi_encode_test.go
```diff
@@ -0,0 +1,1016 @@
...
+			enumIndex:    Tuple,
+			staticLength: 0,
+			childTypes:   []Type{},
+		})
+		require.NoError(t, err, "decode empty tuple should not return error")
+		require.Equal(t, emptyTuple, decoded, "empty tuple encode should not return error")
+	})
+}
+
+func TestDecodeInvalid(t *testing.T) {
```
ahangsu, line 745: Hand written some invalid test cases for decoding, by corrupting the bytes of the ABI value.

The goal is to see decoding function is robust enough to accept erroneous bytes (without crashing itself) and distinguish if the byte match with the ideal ABI type.

Assuming that all the ABI value to be encoded are well-formed, I did not provide the test `TestEncodeValid`, but based on the current implementation of `Encode`, ill-formed ABI values will be eliminated by `Get` method in ABI value.

data/abi/abi_encode_test.go
```diff
@@ -0,0 +1,1016 @@
...
+			utf8ByteLen := len([]byte(utf8Str))
+			lengthBytes := make([]byte, 2)
+			binary.BigEndian.PutUint16(lengthBytes, uint16(utf8ByteLen))
+			expected := append(lengthBytes, []byte(utf8Str)...)
+
+			strEncode, err := strValue.Encode()
+			require.NoError(t, err, "string encode fail")
+			require.Equal(t, expected, strEncode, "encode string not match with expected")
+		}
+	}
```
ahangsu, line 160: The reason I used `gobberish` library is that, in ABI spec, strings are following utf-8 encoding, which means a string element may be more than 1 byte. To test this, `gobberish` provides random utf-8 strings to check if encoding/decoding of ABI strings provide a round-trip.

data/abi/abi.go
```diff
@@ -0,0 +1,1164 @@
...
+			return false
+		}
+		for i := 0; i < len(t.childTypes); i++ {
+			if !t.childTypes[i].Equal(t0.childTypes[i]) {
+				return false
+			}
+		}
+	}
+	return true
+}
```
ahangsu, line 368: Thanks for the review, now this `Equal` tests the struct fields, which makes things more straightforward.

data/abi/abi.go
```diff
@@ -0,0 +1,1164 @@
...
+		return true
+	default:
+		for _, childT := range t.childTypes {
+			if childT.IsDynamic() {
+				return true
+			}
+		}
+		return false
+	}
+}
```
ahangsu, line 383: `IsDynamic` is simplified accordingly.

data/abi/abi.go
```diff
@@ -0,0 +1,1164 @@
...
+				i += after
+				size++
+			} else {
+				childByteSize, err := t.childTypes[i].ByteLen()
+				if err != nil {
+					return -1, err
+				}
+				size += childByteSize
+			}
+		}
```
ahangsu, line 433: Those lines are taken from `Encode`, thanks for @jasonpaulos to point out. Now the byte size calculation for input type like `(bool[2],bool[2])` should be correct.

data/abi/abi.go
```diff
@@ -0,0 +1,1164 @@
...
+	default:
+		return -1, fmt.Errorf("%s is a dynamic type", t.String())
+	}
+}
+
+// Value struct is the ABI Value, holding ABI Type information and the ABI value representation.
+type Value struct {
+	ABIType Type
+	value   interface{}
+}
```
ahangsu, line 444: I make ABI type public, but I am not sure if `value` field should be public, probably need to discuss somewhat about that.

data/abi/abi.go
```diff
@@ -0,0 +1,1164 @@
...
+	castedTupleType, err := MakeTupleType(tupleType)
+	if err != nil {
+		return Value{}, err
+	}
+
+	return Value{
+		ABIType: castedTupleType,
+		value:   values,
+	}, nil
+}
```
ahangsu, line 1026: Changed accordingly on `Make` for `static array`, `dynamic array` and `tuple`.

I think by ABI spec, `dynamic array` and `tuple` can be empty; while `static array` must have at least 1 element:
- `dynamic array` may need additional type argument, or element type cannot be inferred for empty array.
- `tuple` just need an array of value argument.

data/abi/abi.go
```diff
@@ -0,0 +1,1164 @@
...
+	case ArrayStatic, Tuple:
+		elements := v.value.([]Value)
+		if v.ABIType.staticLength <= index {
+			return Value{}, fmt.Errorf("cannot get element: index out of scope")
+		}
+		return elements[index], nil
+	default:
+		return Value{}, fmt.Errorf("cannot get value by index for non array-like type")
+	}
+}
```
ahangsu, line 1155: I merged `Get` method for tuple and arrays into `GetValueByIndex`. See if it needs to be simplified.

data/abi/abi_value.go
```diff
@@ -0,0 +1,304 @@
...
+package abi
+
+import (
+	"fmt"
+	"math/big"
+)
+
+// Value struct is the ABI Value, holding ABI Type information and the ABI value representation.
+type Value struct {
+	ABIType Type
```
algorandskiy, line 26: `ABIType` looks internal to the package, should be lowercase then

data/abi/abi_encode.go
```diff
@@ -0,0 +1,444 @@
...
+			childT[i] = MakeByteType()
+			valueArr[i] = MakeByte(strByte[i])
+		}
+	case Address:
+		addr, err := v.GetAddress()
+		if err != nil {
+			return Value{}, err
+		}
+
+		childT = make([]Type, 32)
```
algorandskiy, line 52: there was a func like `ByteLen()` - use it (or a named constant) to minimize numeric literals usage

data/abi/abi_type.go
```diff
@@ -0,0 +1,458 @@
...
+	Tuple
+)
+
+// Type is the struct that stores information about an ABI value's type.
+type Type struct {
+	enumIndex  BaseType
+	childTypes []Type
+
+	// only can be applied to `uint` size <N> or `ufixed` size <N>
+	size uint16
```
algorandskiy, line 68: rename of comment to reflect it is a bit size

data/abi/abi_encode.go
```diff
@@ -0,0 +1,444 @@
...
+func (v Value) Encode() ([]byte, error) {
+	switch v.ABIType.enumIndex {
+	case Uint:
+		bigIntValue, err := v.GetUint()
+		if err != nil {
+			return []byte{}, err
+		}
+		// NOTE: ugly work-round for golang 1.14. if upgraded to 1.15, should use fillbytes
+		bigIntBytes := bigIntValue.Bytes()
+		buffer := make([]byte, v.ABIType.size/8-uint16(len(bigIntBytes)))
```
algorandskiy, line 97: test for overflow?

data/abi/abi_encode.go
```diff
@@ -0,0 +1,444 @@
...
+	default:
+		return []byte{}, fmt.Errorf("bruh you should not be here in encoding: unknown type error")
+	}
+}
+
+// compressMultipleBool compress consecutive bool values into a byte in ABI tuple/array value.
+func compressMultipleBool(valueList []Value) (uint8, error) {
+	var res uint8 = 0
+	if len(valueList) > 8 {
+		return 0, fmt.Errorf("value list passed in should be less than length 8")
```
algorandskiy, line 156: less than -> not greater

data/abi/abi_encode.go
```diff
@@ -0,0 +1,444 @@
...
+				fmt.Errorf("ufixed%dx%d decode: expected length %d, got byte length %d",
+					valueType.size, valueType.precision, valueType.size/8, len(valueByte))
+		}
+		ufixedNumerator := big.NewInt(0).SetBytes(valueByte)
+		return MakeUfixed(ufixedNumerator, valueType.size, valueType.precision)
+	case Bool:
+		if len(valueByte) != 1 {
+			return Value{}, fmt.Errorf("boolean byte should be length 1 byte")
+		}
+		boolValue := valueByte[0] > 0
```
algorandskiy, line 280: maybe need to be stricter here

data/abi/abi_encode_test.go
```diff
@@ -0,0 +1,1016 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with go-algorand.  If not, see <https://www.gnu.org/licenses/>.
+
+package abi
+
+import (
```
algorandskiy, line 19: I think these tests are great. But could you add one more random test like this:
1. select N - number of values
2. for each N select one of BaseType types
3. for each type generate value. For type[K] select a random K.
4. run encode/decode and ensure the result matches to the initial generated sequence

data/abi/abi_encode.go
```diff
@@ -0,0 +1,444 @@
...
+	}
+
+	tailCurrLength := 0
+	for i := 0; i < len(heads); i++ {
+		if isDynamicIndex[i] {
+			headValue := headLength + tailCurrLength
+			if headValue >= (1 << 16) {
+				return []byte{}, fmt.Errorf("encoding error: byte length exceed 2^16")
+			}
+			binary.BigEndian.PutUint16(heads[i], uint16(headValue))
```
algorandskiy, line 243: could you please clarify a bit this heads[i] encoding, can't match to the spec `head(x[i]) = enc(len( head(x[1]) ... head(x[N]) tail(x[1]) ... tail(x[i-1]) ))`

data/abi/abi_type.go
```diff
@@ -0,0 +1,499 @@
...
+	case String:
+		return "string"
+	case Tuple:
+		typeStrings := make([]string, len(t.childTypes))
+		for i := 0; i < len(t.childTypes); i++ {
+			typeStrings[i] = t.childTypes[i].String()
+		}
+		return "(" + strings.Join(typeStrings, ",") + ")"
+	default:
+		panic("Bruh you should not be here")
```
algochoi, line 108: 🤔

ahangsu: We're not talking about this 🤫

data/abi/abi_encode.go
```diff
@@ -0,0 +1,468 @@
...
+
+		encoded, err := tupleEncoding(convertedTuple)
+		if err != nil {
+			return []byte{}, err
+		}
+		return append(lengthEncode, encoded...), nil
+	case Tuple:
+		return tupleEncoding(v)
+	default:
+		return []byte{}, fmt.Errorf("bruh you should not be here in encoding: unknown type error")
```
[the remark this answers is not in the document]
ahangsu: ah duh 🤦

data/abi/abi_value.go
```diff
@@ -0,0 +1,304 @@
...
+
+// MakeUint32 takes a go `uint32` and gives an ABI Value of ABI type `uint32`.
+func MakeUint32(value uint32) (Value, error) {
+	bigInt := big.NewInt(int64(value))
+	return MakeUint(bigInt, 32)
+}
+
+// MakeUint64 takes a go `uint64` and gives an ABI Value of ABI type `uint64`.
+func MakeUint64(value uint64) (Value, error) {
+	bigInt := big.NewInt(int64(0)).SetUint64(value)
```
jannotti, line 50: I find this a little more concise `new(big.Int).SetUint64(value)`

data/abi/abi_value.go
```diff
@@ -0,0 +1,304 @@
...
+}
+
+// MakeUint32 takes a go `uint32` and gives an ABI Value of ABI type `uint32`.
+func MakeUint32(value uint32) (Value, error) {
+	bigInt := big.NewInt(int64(value))
+	return MakeUint(bigInt, 32)
+}
+
+// MakeUint64 takes a go `uint64` and gives an ABI Value of ABI type `uint64`.
+func MakeUint64(value uint64) (Value, error) {
```
jannotti, line 49: I think I would prefer these did not return errors.  They are constrained by the type system so that's impossible.  So it would be nice to not have to worry about the error returns in a public function.  IMO, it would be ok to panic on error internally.  Perhaps others disagree.

data/abi/abi_value.go
```diff
@@ -0,0 +1,304 @@
...
+// MakeString takes a string and returns an ABI String type Value.
+func MakeString(value string) Value {
+	return Value{
+		ABIType: MakeStringType(),
+		value:   value,
+	}
+}
+
+// MakeByte takes a byte and returns an ABI Byte type value.
+func MakeByte(value byte) Value {
```
jannotti, line 95: Do we have a byte type and a uint8 type?

data/abi/abi_value.go
```diff
@@ -0,0 +1,304 @@
...
+	return Value{
+		ABIType: MakeAddressType(),
+		value:   value,
+	}
+}
+
+// MakeDynamicArray takes an array of ABI value (can be empty) and element type,
+// returns an ABI dynamic length array value.
+func MakeDynamicArray(values []Value, elemType Type) (Value, error) {
+	if len(values) >= (1 << 16) {
```
jannotti, line 113: Perhaps `len(values) > math.MaxUint16`

data/abi/abi_value.go
```diff
@@ -0,0 +1,304 @@
...
+		return Value{}, fmt.Errorf("dynamic array make error: pass in argument number larger than 2^16")
```
jannotti, line 114: "larger that maximum uint16"

Since the current phrasing is off by one.

data/abi/abi_value.go
```diff
@@ -0,0 +1,304 @@
...
+	return Value{
+		ABIType: MakeDynamicArrayType(elemType),
+		value:   values,
+	}, nil
+}
+
+// MakeStaticArray takes an array of ABI value and returns an ABI static length array value.
+func MakeStaticArray(values []Value) (Value, error) {
+	if len(values) >= (1 << 16) {
+		return Value{}, fmt.Errorf("static array make error: pass in argument number larger than 2^16")
```
jannotti, line 131: Same constant comment, and this error message should talk about length of array, not "argument number".

data/abi/abi_value.go
```diff
@@ -0,0 +1,304 @@
...
+	} else if len(values) == 0 {
+		return Value{}, fmt.Errorf("static array make error: 0 argument passed in")
+	}
```
jannotti, line 134: Are zero length dynamic arrays disallowed?  They seem ok to me.

data/abi/abi_value.go
```diff
@@ -0,0 +1,304 @@
...
+	return Value{
+		ABIType: MakeStaticArrayType(values[0].ABIType, uint16(len(values))),
+		value:   values,
+	}, nil
+}
+
+// MakeTuple takes an array of ABI values and returns an ABI tuple value.
+func MakeTuple(values []Value) (Value, error) {
+	if len(values) >= (1 << 16) {
+		return Value{}, fmt.Errorf("tuple make error: pass in argument number larger than 2^16")
```
jannotti, line 150: samesies

data/abi/abi_value.go
```diff
@@ -0,0 +1,304 @@
...
+	return Value{
+		ABIType: MakeBoolType(),
+		value:   value,
+	}
+}
+
+// GetUint8 tries to retreve an uint8 from an ABI Value.
+func (v Value) GetUint8() (uint8, error) {
+	if v.ABIType.abiTypeID != Uint || v.ABIType.bitSize > 8 {
+		return 0, fmt.Errorf("value type unmatch or bitSize too large")
```
jannotti, line 179: "mismatch" would be a better word.

data/abi/abi_value.go
```diff
@@ -0,0 +1,304 @@
...
+	bigIntForm, err := v.GetUint()
+	if err != nil {
+		return 0, err
+	}
+	return uint8(bigIntForm.Uint64()), nil
+}
+
+// GetUint16 tries to retrieve an uint16 from an ABI Value.
+func (v Value) GetUint16() (uint16, error) {
+	if v.ABIType.abiTypeID != Uint || v.ABIType.bitSize > 16 {
```
jannotti, line 190: maybe all these uint size checks should be a small helper, checkUint(t Type, bitSize int)

data/abi/abi_value.go
```diff
@@ -0,0 +1,304 @@
...
+// GetUint tries to retrieve an big uint from an ABI Value.
+func (v Value) GetUint() (*big.Int, error) {
+	if v.ABIType.abiTypeID != Uint {
+		return nil, fmt.Errorf("value type unmatch")
+	}
+	bigIntForm := v.value.(*big.Int)
+	sizeThreshold := big.NewInt(0).Lsh(big.NewInt(1), uint(v.ABIType.bitSize))
+	if sizeThreshold.Cmp(bigIntForm) <= 0 {
+		return nil, fmt.Errorf("value is larger than uint bitSize")
+	}
```
jannotti, line 233: Is this checking that the value's size is bigger than its own belief about its bitsize?

data/abi/abi_value.go
```diff
@@ -0,0 +1,304 @@
...
+}
+
+// GetUfixed tries to retrieve an big rational number from an ABI Value.
+func (v Value) GetUfixed() (*big.Int, error) {
+	if v.ABIType.abiTypeID != Ufixed {
+		return nil, fmt.Errorf("value type unmatch, should be ufixed")
+	}
+	bigIntForm := v.value.(*big.Int)
+	sizeThreshold := big.NewInt(0).Lsh(big.NewInt(1), uint(v.ABIType.bitSize))
+	if sizeThreshold.Cmp(bigIntForm) <= 0 {
```
jannotti, line 244: The comment says it returns a big.Rat, but I think it just returns the value, scaled up to an integer?

data/abi/abi_type.go
```diff
@@ -0,0 +1,499 @@
...
+func TypeFromString(str string) (Type, error) {
+	// if the regexps for type str matching are not initialized
+	// attempt to initialize with `initRegexps`
+	// if error return err
+	if !regexpInitialized {
+		err := initRegexps()
+		if err != nil {
+			return Type{}, err
+		}
+	}
```
jannotti, line 143: I think it's better to call this from an init() function. https://golang.org/doc/effective_go#init

data/abi/abi_type.go
```diff
@@ -0,0 +1,499 @@
...
+
+// Equal method decides the equality of two types: t == t0.
+func (t Type) Equal(t0 Type) bool {
+	if t.abiTypeID != t0.abiTypeID {
+		return false
+	} else if t.precision != t0.precision || t.bitSize != t0.bitSize {
+		return false
+	} else if t.staticLength != t0.staticLength {
+		return false
+	} else {
```
jannotti, line 392: I think these will read better without the `else`s.  Just a bunch of ifs looking like guard clauses.

data/abi/abi_value.go
ahangsu: byte type is just an alias of uint8. If it's for the sake of consistency, I can switch that to uint8.

jannotti: I suppose we want it mainly so we can have byte[32], instead of uint8[16].  I suppose that's fine. It's slightly odd that you can have two methods f(byte) and f(uint8), but I guess that's the same problem we have with an alias (we have a few, I guess).

data/abi/abi_type.go
jannotti: I think so.  They're static patterns, so it would mean there is a badly written regexp.

data/abi/abi_value.go
ahangsu: Yes, this checks is exactly checking sizeThreshold <= bigIntForm.

I want to make sure the value size is within the bitsize defined in type.

data/abi/abi_value.go
jannotti: Should it return the width as well, then?  It seems weird to "lose information" when using these Get* accessors.  I suppose the caller could read it from the value (is the width a public field?)

data/abi/abi_value.go
jasonpaulos: I agree, for `MakeUint(8|16|32|64)`, the functions should probably just panic on error, since it should not be possible for an error to happen

jasonpaulos (commented):
Thank you for breaking up the code into multiple files, this was much easier to review

data/abi/abi_value.go
ahangsu: Zero length dynamic array, or tuple are allowed

jannotti: Oh, right, we're in static constructor here.  I guess we can disallow [0]byte, etc.

data/abi/abi_value.go
jannotti: Oh, but actually you just want `big.NewInt(value)` here, right?

data/abi/abi_encode.go
```diff
@@ -258,12 +258,14 @@ func tupleEncoding(v Value) ([]byte, error) {
 	}
 
 	// concat everything as the abi encoded bytes
-	head, tail := make([]byte, 0), make([]byte, 0)
-	for i := 0; i < len(v.ABIType.childTypes); i++ {
-		head = append(head, heads[i]...)
-		tail = append(tail, tails[i]...)
+	encoded := make([]byte, 0)
```
jasonpaulos, line 261: If the total capacity is known in advance, you can add it as a 3rd argument. This will avoid unnecessary intermediate allocations.
```suggestion
	encoded := make([]byte, 0, headLength+tailCurrLength)
```

jasonpaulos (approved):
Great, this looks good to me

data/abi/abi_type.go
```diff
@@ -0,0 +1,491 @@
...
+
+// String serialize an ABI Type to a string in ABI encoding.
+func (t Type) String() string {
+	switch t.abiTypeID {
+	case Uint:
+		return "uint" + strconv.Itoa(int(t.bitSize))
+	case Byte:
+		return "byte"
+	case Ufixed:
+		return "ufixed" + strconv.Itoa(int(t.bitSize)) + "x" + strconv.Itoa(int(t.precision))
```
jannotti, line 91: No need to change, but I `fmt.Sprintf("ufixed%dx%d", t.bitSize, t.precision)` might be easier on the eyes.

data/abi/abi_type.go
```diff
@@ -0,0 +1,491 @@
...
+func init() {
+	var err error
+	// Note that we allow only decimal static array length
+	staticArrayRegexp, err = regexp.Compile(`^([a-z\d\[\](),]+)\[([1-9][\d]*)]$`)
+	if err != nil {
+		panic(err.Error())
+	}
+	ufixedRegexp, err = regexp.Compile(`^ufixed([1-9][\d]*)x([1-9][\d]*)$`)
+	if err != nil {
+		panic(err.Error())
```
jannotti, line 125: Simplify with https://pkg.go.dev/regexp#MustCompile ?  I suppose that could even make it doable outside init() at global level.

jannotti (approved):

data/abi/abi_type.go
ahangsu: Modified.

data/abi/abi_type.go
ahangsu: I removed the init function, put the regexp compile to global level.

PR #2808, opened 2021-08-26
Register keys with ParticipationRegistry.

winder:
## Summary

Call participationRegistry Register command in `OnNewBlock`.

Required adjusting how the ParticipationID is computed, it's now a hash of the keyreg fields + account address.

## Test Plan

Existing unit tests.

### DIFF
--- a/data/account/participation.go
+++ b/data/account/participation.go
@@ -61,8 +61,11 @@ type Participation struct {
 type participationIDData struct {
 	_struct struct{} `codec:",omitempty,omitemptyarray"`
 
-	transactions.KeyregTxnFields
-	Parent basics.Address `codec:"addr"`
+	Parent      basics.Address    `codec:"addr"`
+	VRFSK       crypto.VrfPrivkey `codec:"vrfsk"`
+	FirstValid  basics.Round      `codec:"fv"`
+	LastValid   basics.Round      `codec:"lv"`
+	KeyDilution uint64            `codec:"kd"`
 }
 
 // ToBeHashed implements the Hashable interface.
@@ -70,27 +73,19 @@ func (id *participationIDData) ToBeHashed() (protocol.HashID, []byte) {
 	return protocol.ParticipationKeys, protocol.Encode(id)
 }
 
-// MakeParticipationID generates the ParticipationID from an address and a key registration.
-func MakeParticipationID(addr basics.Address, fields transactions.KeyregTxnFields) ParticipationID {
-	return ParticipationID(crypto.HashObj(&participationIDData{
-		KeyregTxnFields: fields,
-		Parent:          addr,
-	}))
-}
-
 // ParticipationID computes a ParticipationID.
 func (part Participation) ParticipationID() ParticipationID {
-	// Allow a panic if Voting or VRF are not present.
-	fields := transactions.KeyregTxnFields{
-		VotePK:           part.Voting.OneTimeSignatureVerifier,
-		SelectionPK:      part.VRF.PK,
-		VoteFirst:        part.FirstValid,
-		VoteLast:         part.LastValid,
-		VoteKeyDilution:  part.KeyDilution,
-		Nonparticipation: false,
+	idData := participationIDData{
+		Parent:      part.Parent,
+		FirstValid:  part.FirstValid,
+		LastValid:   part.LastValid,
+		KeyDilution: part.KeyDilution,
+	}
+	if part.VRF != nil {
+		copy(idData.VRFSK[:], part.VRF.SK[:])
 	}
 
-	return MakeParticipationID(part.Parent, fields)
+	return ParticipationID(crypto.HashObj(&idData))
 }
 
 // PersistedParticipation encapsulates the static state of the participation
--- a/data/account/participation.go
+++ b/data/account/participation.go
@@ -61,11 +61,8 @@ type Participation struct {
 type participationIDData struct {
 	_struct struct{} `codec:",omitempty,omitemptyarray"`
 
-	Parent      basics.Address    `codec:"addr"`
-	VRFSK       crypto.VrfPrivkey `codec:"vrfsk"`
-	FirstValid  basics.Round      `codec:"fv"`
-	LastValid   basics.Round      `codec:"lv"`
-	KeyDilution uint64            `codec:"kd"`
+	transactions.KeyregTxnFields
+	Parent basics.Address `codec:"addr"`
 }
 
 // ToBeHashed implements the Hashable interface.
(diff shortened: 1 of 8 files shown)


node/node.go
```diff
@@ -869,6 +869,18 @@ func (node *AlgorandFullNode) OnNewBlock(block bookkeeping.Block, delta ledgerco
...
+		if tx.Txn.Type == protocol.KeyRegistrationTx {
+			id := account.MakeParticipationID(tx.Txn.Sender, tx.Txn.KeyregTxnFields)
+			err := node.participationRegistry.Register(id, block.BlockHeader.Round)
+			// If the key is not installed on this node, ErrParticipationIDNotFound is quickly returned.
+			if err != nil && err != account.ErrParticipationIDNotFound {
+				node.log.Error("Problem with participationRegistry.Register: %w", err)
+			}
+		}
+	}
+
```
tsachiherman, line 883: I'm not sure that this is the best way to do that. Is it possible to conceptually defer this to the moment where we actually need the information and check on the ledger ?

data/account/participation.go
```diff
@@ -61,31 +61,36 @@ type Participation struct {
 type participationIDData struct {
 	_struct struct{} `codec:",omitempty,omitemptyarray"`
 
-	Parent      basics.Address    `codec:"addr"`
-	VRFSK       crypto.VrfPrivkey `codec:"vrfsk"`
-	FirstValid  basics.Round      `codec:"fv"`
-	LastValid   basics.Round      `codec:"lv"`
-	KeyDilution uint64            `codec:"kd"`
+	transactions.KeyregTxnFields
```
tsachiherman, line 64: I think that the previous implementation was fine - the identity of the participation key is not really related to the key registration transaction fields. ( it happened to have a similar set of fields, but it's a different "creature" ).

winder: My thinking was that Idan's project is likely to change both of these, and maybe this would allow the change to happen in one place.

Interestingly, the existing ID doesn't include key dilution. I'm not sure if that's a bug, but including it here definitely gives more entropy to the ID.

node/node.go
winder: So when GetKeys is called it fetches all keys here and checks if any should be registered?

I think that could work and be less error-prone but require more account lookups. I'll investigate.

tsachiherman: I think that GetKeys would be called only on user-driven cases, whereas this would be called on each and every round.. so we should try and make it so that we push the "extra work" to be per-user case.

data/account/participation.go
tsachiherman: This ParticipationID hash need to capture the "identity" of the file.. but that's about it. I don't think we need to worry too much about Idan's upcoming changes, as there would probably mean adding one or two more fields to the `transactions.KeyregTxnFields` which we *can* add to the ParticipationID, but don't have to.

data/account/participationRegistry.go
```diff
@@ -119,10 +124,10 @@ type ParticipationRegistry interface {
 	Delete(id ParticipationID) error
 
 	// Get a participation record.
-	Get(id ParticipationID) (ParticipationRecord, error)
+	Get(id ParticipationID) ParticipationRecord
 
 	// GetAll of the participation records.
-	GetAll() ([]ParticipationRecord, error)
```
winder, line 125: Most of the changes are related to simplifying the interface call. There isn't a DB error to pass along anymore, so no need for the error value. The caller can check for a zero value / empty list if they need to know that nothing was found.

node/node.go
```diff
@@ -1161,6 +1161,12 @@ func (node *AlgorandFullNode) VotingKeys(votingRound, keysRound basics.Round) []
 		}
 		participations = append(participations, part)
 		matchingAccountsKeys[part.Address()] = true
+
+		// This is usually a no-op, but the first time it will update the DB.
+		err := node.participationRegistry.Register(part.ParticipationID(), keysRound)
+		if err != nil {
+			node.log.Error("Failed to register participation key (%s) with participation registry.", part.ParticipationID())
```
tsachiherman, line 1168: `Error` -> `Warnf`.
Errors are meant for critical functional issues with the node. The node would continue to work just fine even if this call fails.

node/node.go
```diff
@@ -1161,6 +1161,12 @@ func (node *AlgorandFullNode) VotingKeys(votingRound, keysRound basics.Round) []
 		}
 		participations = append(participations, part)
 		matchingAccountsKeys[part.Address()] = true
+
+		// This is usually a no-op, but the first time it will update the DB.
+		err := node.participationRegistry.Register(part.ParticipationID(), keysRound)
```
tsachiherman, line 1166: please benchmark the `Register` and make sure it won't take too long. ( if it takes less than 50ms, then it's probably just fine. Otherwise, we need to consider moving it to a go-routine )

data/account/participationRegistry.go
```diff
@@ -46,13 +46,18 @@ type ParticipationRecord struct {
...
+	EffectiveFirst         basics.Round
+	EffectiveLast          basics.Round
 
 	// VRFSecrets
 	// OneTimeSignatureSecrets
 }
 
+// IsZero returns true if the object contains zero values.
+func (r ParticipationRecord) IsZero() bool {
+	return r == (ParticipationRecord{})
```
tsachiherman, line 58: I'm wondering if creating a global empty object and comparing it to that one would get executed faster.

winder: I suppose it couldn't hurt. If it does help maybe we should do that for Address as well

node/node.go
winder: Looks very fast to me. When you say ms are you think microseconds or milliseconds? Either way we're still under 50 if there's only 1 account registered:
```
BenchmarkKeyRegistration1/KeyInsert_1
BenchmarkKeyRegistration1/KeyInsert_1-8    	  202903	      6154 ns/op
BenchmarkKeyRegistration1/KeyRegistered_1
BenchmarkKeyRegistration1/KeyRegistered_1-8         	   26833	     44209 ns/op
BenchmarkKeyRegistration1/NoOp_1
BenchmarkKeyRegistration1/NoOp_1-8                  	  194818	      5980 ns/op
BenchmarkKeyRegistration5
BenchmarkKeyRegistration5/KeyInsert_5
BenchmarkKeyRegistration5/KeyInsert_5-8             	   41196	     30087 ns/op
BenchmarkKeyRegistration5/KeyRegistered_5
BenchmarkKeyRegistration5/KeyRegistered_5-8         	    5782	    221874 ns/op
BenchmarkKeyRegistration5/NoOp_5
BenchmarkKeyRegistration5/NoOp_5-8                  	   39292	     29898 ns/op
BenchmarkKeyRegistration10
BenchmarkKeyRegistration10/KeyInsert_10
BenchmarkKeyRegistration10/KeyInsert_10-8           	   19551	     59042 ns/op
BenchmarkKeyRegistration10/KeyRegistered_10
BenchmarkKeyRegistration10/KeyRegistered_10-8       	    2823	    448233 ns/op
BenchmarkKeyRegistration10/NoOp_10
BenchmarkKeyRegistration10/NoOp_10-8                	   19406	     60828 ns/op
BenchmarkKeyRegistration50
BenchmarkKeyRegistration50/KeyInsert_50
BenchmarkKeyRegistration50/KeyInsert_50-8           	    3784	    394140 ns/op
BenchmarkKeyRegistration50/KeyRegistered_50
BenchmarkKeyRegistration50/KeyRegistered_50-8       	     394	   3207132 ns/op
BenchmarkKeyRegistration50/NoOp_50
BenchmarkKeyRegistration50/NoOp_50-8                	    2625	    427918 ns/op
```

tsachiherman: milliseconds. and these looks good. ( I just wanted to make sure )

winder: Do you think it's worth making this asynchronous? This design means that we'll frequently re-check for keys to register so it no longer needs to update the DB in real-time.

tsachiherman: The `VotingKeys` is called by the pseudonode once every round so that it could use the keys for voting on that round. By slowing this down, we're holding off the agreement from making the proposal message ( which is a time critical). This could be really bad if it takes too long; however, if we're taking about 50Kns, than it's probably wouldn't make a difference. Just a side note that if we take any lock in there - that lock might take considerably more than 50Kns, so we need to be careful around that.

winder: There is a lock taken for most of the record function. There is probably a clever solution with multiple mutexes that would avoid a DB access lock in the agreement path, I'll need to think about that more.

winder: I'll create a follow-up ticket for avoiding mutexes and DB operations in Record/Register.

PR #2810, opened 2021-08-26
Add ParticipationRegistry Insert/Delete usage in node.

winder:
## Summary

This PR comes after #2808

## Test Plan

CI


data/accountManager.go
```diff
@@ -31,23 +30,11 @@ import (
 	"github.com/algorand/go-algorand/protocol"
 )
 
-// A ParticipationKeyIdentity defines the parameters that makes a pariticpation key unique.
-type ParticipationKeyIdentity struct {
```
winder, line 35: This is a little different than my ParticipationID, but they both seem to do the same thing so I'm merging them.

data/account/participation.go
```diff
@@ -69,13 +69,18 @@ type participationIDData struct {
 }
 
 // ToBeHashed implements the Hashable interface.
-func (id *participationIDData) ToBeHashed() (protocol.HashID, []byte) {
+func (id *ParticipationKeyIdentity) ToBeHashed() (protocol.HashID, []byte) {
 	return protocol.ParticipationKeys, protocol.Encode(id)
 }
 
+// ToParticipationID creates a ParticipationID hash from the identity file.
+func (id ParticipationKeyIdentity) ToParticipationID() ParticipationID {
```
tsachiherman, line 77: I think that a method named `ID()` would be a better choice here, but I'm not overly particular here.

data/accountManager.go
tsachiherman: I think that it's ok - but you'll also need `VoteID:      participation.Voting.OneTimeSignatureVerifier` so that it would capture both.

node/node.go
```diff
@@ -825,6 +825,13 @@ func (node *AlgorandFullNode) loadParticipationKeys() error {
 				return fmt.Errorf("AlgorandFullNode.loadParticipationKeys: cannot load account at %v: %v", info.Name(), err)
 			}
 		} else {
+			// Tell the ParticipationRegistry about the Participation (dupes don't matter)
+			// TODO: Put the registry in the AccountManager?
```
tsachiherman, line 829:
```golang
// TODO: Put the registry in the AccountManager?
```

exactly - this entire function is going to be called only upon startup, and therefore it's much better to be moved into the AccountManager.

node/node.go
```diff
@@ -935,6 +942,15 @@ func (node *AlgorandFullNode) oldKeyDeletionThread() {
...
 		node.mu.Unlock()
+
+		// Delete expired records from participation registry.
+		for _, record := range node.participationRegistry.GetAll() {
+			if record.LastVote < hdr.Round {
+				if err := node.participationRegistry.Delete(record.ParticipationID); err != nil {
+					node.log.Warnf("Problem deleting key from participation registry: %w", err)
+				}
+			}
+		}
```
tsachiherman, line 953: looks like you need to add a `node.participationRegistry.DeleteAllBeforeRound()` and pass `hdr.Round`.

node/node.go
winder: I'll try doing this again. Your suggestion in my last PR to move `registry.Register` call from `OnNewBlock` -> `VotingKeys` should make this a lot cleaner.

PR #2811, opened 2021-08-26
Fix overriding consensus parameters in evaluator.

tolikzinovyev:
## Summary

We failed to override the consensus parameters in COW. This PR fixes it. Background: overriding of consensus parameters is used by Indexer to retrieve the asset close amount of a transaction even when the protocol doesn't support it.

## Test Plan

Add a unit test.

### DIFF
--- a/ledger/cow_test.go
+++ b/ledger/cow_test.go
@@ -26,6 +26,7 @@ import (
 	"github.com/algorand/go-algorand/data/bookkeeping"
 	"github.com/algorand/go-algorand/data/transactions"
 	"github.com/algorand/go-algorand/ledger/ledgercore"
+	"github.com/algorand/go-algorand/protocol"
 	"github.com/algorand/go-algorand/test/partitiontest"
 )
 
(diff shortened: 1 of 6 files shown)


ledger/cow_test.go
```diff
@@ -112,7 +113,7 @@ func TestCowBalance(t *testing.T) {
 	accts0 := randomAccounts(20, true)
 	ml := mockLedger{balanceMap: accts0}
 
-	c0 := makeRoundCowState(&ml, bookkeeping.BlockHeader{}, 0, 0)
+	c0 := makeRoundCowState(&ml, bookkeeping.BlockHeader{}, config.ConsensusParams{}, 0, 0)
```
tsachiherman, line 116: please avoid passing this `config.ConsensusParams{}`, even if that works right now.
Just pick the ConsensusParams for the default protocol version.

ledger/eval_test.go
```diff
@@ -856,7 +856,7 @@ func TestCowCompactCert(t *testing.T) {
 	blocks := make(map[basics.Round]bookkeeping.BlockHeader)
 	blockErr := make(map[basics.Round]error)
 	ml := mockLedger{balanceMap: accts0, blocks: blocks, blockErr: blockErr}
-	c0 := makeRoundCowState(&ml, bookkeeping.BlockHeader{}, 0, 0)
+	c0 := makeRoundCowState(&ml, bookkeeping.BlockHeader{}, config.ConsensusParams{}, 0, 0)
```
tsachiherman, line 859: same as above.

ledger/cow_test.go
tolikzinovyev: done

ledger/eval_test.go
tolikzinovyev: done

tsachiherman (approved):
Looks good to me.

PR #2812, opened 2021-08-26
Simplify `startEvaluator()`

tolikzinovyev:
## Summary

The current code creates an impression that the evaluator can be used for round 0 which is not true. This PR simplifies the code, so this is more evident.

## Test Plan

Relying on existing tests.

### DIFF
--- a/ledger/eval.go
+++ b/ledger/eval.go
@@ -41,6 +41,9 @@ import (
 // ErrNoSpace indicates insufficient space for transaction in block
 var ErrNoSpace = errors.New("block does not have space for transaction")
 
+// ErrRoundZero is self-explanatory
+var ErrRoundZero = errors.New("cannot start evaluator for round 0")
+
 // maxPaysetHint makes sure that we don't allocate too much memory up front
 // in the block evaluator, since there cannot reasonably be more than this
 // many transactions in a block.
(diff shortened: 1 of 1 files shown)


ledger/eval.go
```diff
@@ -383,20 +383,32 @@ func (l *Ledger) StartEvaluator(hdr bookkeeping.BlockHeader, paysetHint int) (*B
 }
 
 func startEvaluator(l ledgerForEvaluator, hdr bookkeeping.BlockHeader, proto config.ConsensusParams, paysetHint int, validate bool, generate bool) (*BlockEvaluator, error) {
+	if hdr.Round == 0 {
+		return nil, fmt.Errorf("cannot start evaluator for round 0")
```
tsachiherman, line 387: create global constant error using errors.New(), and use it here.

ledger/eval.go
```diff
@@ -412,35 +424,23 @@ func startEvaluator(l ledgerForEvaluator, hdr bookkeeping.BlockHeader, proto con
...
-
-		var ok bool
-		prevProto, ok = config.Consensus[eval.prevHeader.CurrentProtocol]
-		if !ok {
-			return nil, protocol.Error(eval.prevHeader.CurrentProtocol)
-		}
+	prevProto, ok := config.Consensus[prevHeader.CurrentProtocol]
+	if !ok {
+		return nil, protocol.Error(prevHeader.CurrentProtocol)
+	}
```
tsachiherman, line 432: move this section right after 
```golang
prevHeader, err := l.BlockHdr(hdr.Round - 1)
 	if err != nil {
 		return nil, fmt.Errorf(
 			"can't evaluate block %v without previous header: %v", hdr.Round, err)
 	}
```
for further clarity.

tsachiherman (changes requested):
few small requests, but looks good otherwise.

algorandskiy:
I personally do not see any value in this change.

winder (commented):
Are we sure this function is never called on round 0? It appears that previously it would be a NoOp (or a crash?), now it's an error.

tolikzinovyev:
I tried using it for Indexer for round 0, I got an error. It's probably in
```
	poolAddr := eval.prevHeader.RewardsPool
	// get the reward pool account data without any rewards
	incentivePoolData, _, err := l.LookupWithoutRewards(eval.prevHeader.Round, poolAddr)
	if err != nil {
		return nil, err
	}
```

`eval.prevHeader` is unset, so we are trying to get a balance for zero address.

tsachiherman (approved):
look good. thanks for the changes.

PR #2818, opened 2021-08-30
Fix bugs / refactor txnsync message encoding / decoding

nicholasguoalgorand:
## Summary

Address comments made by @tolikzinovyev regarding correctness and stylistic issues within the message encoding of the transaction sync.

### DIFF
--- a/txnsync/bitmask.go
+++ b/txnsync/bitmask.go
@@ -17,7 +17,6 @@
 package txnsync
 
 import (
-	"bytes"
 	"errors"
 )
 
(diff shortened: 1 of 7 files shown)


txnsync/bitmask.go
```diff
@@ -73,6 +72,9 @@ func (b *bitmask) trimBitmask(entries int) {
 	switch bitmaskType {
 	case 1:
 		(*b)[0] = 1
+		for i := entries; i < (len(*b) - 1) * 8; i++ {
+			b.setBit(i)
```
algonautshant, line 76: Why is this?

nicholasguoalgorand: Say the bitmask is for 7 entries, and we want to convey that all 7 entries are present. Then with the old logic, the bitmask would be for case 1 and be 0000001. What we want instead is 00000000, and for the byte to be omitted from being sent since it is 0.

nicholasguoalgorand: However, I think that this can be removed since we trim the bitmasks at the end by just using the `best size` anyway.

tolikzinovyev: Yes, I think this is not needed.

txnsync/bitmask.go
```diff
@@ -109,7 +111,7 @@ func (b *bitmask) trimBitmask(entries int) {
 	default:
 	}
 
-	*b = bytes.TrimRight(*b, "\x00")
+	*b = (*b)[:bestSize]
```
tolikzinovyev, line 114: nit: have this in case 0 and case 1, and delete `return` from case 2 and case 3. Also delete case `default`.

txnsync/bitmask.go
algonautshant: I think this will even be wrong. 
If the original bits is 10000000, it should be sent as 01111111. With this change, it will be sent as 00000000.
This will violate the statements in func iterate:
```
			if 255>>maxJ != v {
				// The remaining of the bits must be 1
				return errIndexNotFound
			}
```

algonautshant (approved):
Looks good. 
Thanks for the changes.

txnsync/bitmask.go
```diff
@@ -71,13 +70,16 @@ func (b *bitmask) trimBitmask(entries int) {
...
+	case 0:
+		*b = (*b)[:bestSize]
 	case 1:
 		(*b)[0] = 1
 		for i := range *b {
 			if i != 0 {
 				(*b)[i] = 255 - (*b)[i] // invert bits
 			}
 		}
+		*b = (*b)[:bestSize]
```
tsachiherman, line 82: now that you're writing it like this, it looks like we can modify the above for loop to be
```golang
for i := 1; i < bestSize; i++ {
   (*b)[i] = 255 - (*b)[i] // invert bits
}
```

but we can also wait with this for a separate PR, since this one looks good to me.

tsachiherman (approved):

PR #2819, opened 2021-08-31
Store blockProof in a different DB table to improve performance

Aharonee:
Added DB table blockProof and updated Persist & Restore.

### DIFF
--- a/data/account/partInstall.go
+++ b/data/account/partInstall.go
@@ -19,6 +19,8 @@ package account
 import (
 	"database/sql"
 	"fmt"
+	"github.com/algorand/go-algorand/crypto/merklekeystore"
+	"github.com/algorand/go-algorand/protocol"
 )
 
 // PartTableSchemaName is the name of the table in the Schema Versions table storing the table + version details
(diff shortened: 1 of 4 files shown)


data/account/partInstall.go
```diff
@@ -39,12 +39,18 @@ func partInstallDatabase(tx *sql.Tx) error {
...
 		lastValid INTEGER,
 
 		keyDilution INTEGER NOT NULL DEFAULT 0
+		);`)
+	if err != nil {
+		return err
+	}
+
+	_, err = tx.Exec(`CREATE TABLE BlockProof (
+		blockProof BLOB  --*  msgpack encoding of ParticipationAccount.BlockProof
```
tsachiherman, line 53: 1. I'd suggest you'll add an integer primary key, and use just index zero. This would give you the flexibility to extend the table content in the future, if needed.
2. Consider moving the sql-creation script into a global private string, and use it both here as well as in `if partVersion == 2`

Aharonee: Yes, I will definitely change those things :)

This is mostly a small patch to allow the team to keep developing from this branch, since the current version does not pass tests. 
The database is still subject to changes since I'm trying to improve the memory footprint (store the dilithium keys in different rows so we could lazy load them at will). These changes are only temporary and will not stay as they are.

PR #2823, opened 2021-09-01
expose nonparticipation txn field in teal

shiqizng:
## Summary
Expose the field Nonparticipation from KeyregTxnFields in teal. This is needed for checking keyreg txns' nonparticipation value in logicsigns. 

## Test Plan
Updated tests in logic


data/transactions/logic/doc.go
```diff
@@ -337,6 +337,7 @@ var txnFieldDocs = map[string]string{
 	"VoteFirst":                "The first round that the participation key is valid.",
 	"VoteLast":                 "The last round that the participation key is valid.",
 	"VoteKeyDilution":          "Dilution for the 2-level participation key",
+	"Nonparticipation":         "Flag for participation key",
```
jasonpaulos, line 340: A better description might be "Key registration participation flag"

data/transactions/logic/fields.go
```diff
@@ -232,6 +235,7 @@ var txnFieldSpecs = []txnFieldSpec{
 	{LocalNumUint, StackUint64, 3},
 	{LocalNumByteSlice, StackUint64, 3},
 	{ExtraProgramPages, StackUint64, 4},
+	{Nonparticipation, StackUint64, 0},
```
jasonpaulos, line 238: Unfortunately I think we'll need to make this first available in v5

jasonpaulos (commented):
For the most part looks good, except the new field should be enabled only for v5 and onward.

Also, we may want to make it such that LogicSigs running programs v4 and lower automatically reject transactions that contain true for the nonparticipation field, since they cannot inspect it. A similar thing was done when we introducing rekeying in TEAL v2. I'll see if I can find any relevant code.

edit: the relevant code is `ComputeMinTealVersion` in `data/transactions/logic/eval.go`. But since it's currently valid for programs to execute in the same group as nonparticipation keyreg txns, we'd probably need to introduce a new consensus param to make this change. I'm not sure if that would be worth it.

shiqizng:
> For the most part looks good, except the new field should be enabled only for v5 and onward.
> 
> Also, we may want to make it such that LogicSigs running programs v4 and lower automatically reject transactions that contain true for the nonparticipation field, since they cannot inspect it. A similar thing was done when we introducing rekeying in TEAL v2. I'll see if I can find any relevant code.
> 
> edit: the relevant code is `ComputeMinTealVersion` in `data/transactions/logic/eval.go`. But since it's currently valid for programs to execute in the same group as nonparticipation keyreg txns, we'd probably need to introduce a new consensus param to make this change. I'm not sure if that would be worth it.

I don't think it would be worth a consensus upgrade either.

data/transactions/logic/eval.go
```diff
@@ -1886,6 +1886,12 @@ func (cx *evalContext) txnFieldToStack(txn *transactions.Transaction, field TxnF
 		sv.Uint = uint64(txn.VoteLast)
 	case VoteKeyDilution:
 		sv.Uint = txn.VoteKeyDilution
+	case Nonparticipation:
+		if txn.Nonparticipation {
+			sv.Uint = 1
+		} else {
+			sv.Uint = 0
+		}
```
jannotti, line 1894:
```suggestion
		sv.Uint = boolToUint(txn.Nonparticipation)
```

data/transactions/logic/doc.go
```diff
@@ -337,6 +337,7 @@ var txnFieldDocs = map[string]string{
 	"VoteFirst":                "The first round that the participation key is valid.",
 	"VoteLast":                 "The last round that the participation key is valid.",
 	"VoteKeyDilution":          "Dilution for the 2-level participation key",
+	"Nonparticipation":         "Key registration participation flag",
```
jannotti, line 340: How about "Marks an account nonparticipating for rewards"?

That is similar to the language we use here: https://developer.algorand.org/docs/reference/transactions/

jannotti (changes requested):
Looks good. Two small recomendations.

jasonpaulos (approved):

jannotti (approved):

PR #2825, opened 2021-09-01
Changing golang version to 1.16.11 in go-algorand

algojack:
## Summary

Upgrading to 1.16 to help alleviate issues with working on different go versions, and update to a supported, more secure version.
Release notes for Go 1.15 and 1.16:
https://tip.golang.org/doc/go1.16
https://tip.golang.org/doc/go1.15

## Acceptance criteria
This story is only for go-algorand 
Indexer golang upgrade: https://github.com/algorand/indexer/issues/782

## Test Plan

Normal tests using CI


onetechnical:
Please do NOT merge prior to next major release.

onetechnical (commented):
This looks fine to me; don't merge it until after beta.

scripts/buildtools/install_buildtools.sh
```diff
@@ -70,16 +70,14 @@ function install_go_module {
...
     VERSION=$(get_go_version "$1")
 
-    # TODO: When we switch to 1.16 this should be changed to use 'go install'
-    #       instead of 'go get': https://tip.golang.org/doc/go1.16#modules
     if [ -z "$VERSION" ]; then
         echo "Unable to install requested package '$1' (${MODULE}): no version listed in ${SCRIPTPATH}/go.mod"
         exit 1
     else
-        OUTPUT=$(GO111MODULE=on go get "${MODULE}@${VERSION}" 2>&1)
+        OUTPUT=$(GO111MODULE=on go install "${MODULE}@${VERSION}" 2>&1)
```
tolikzinovyev, line 77: Is `GO111MODULE` needed with `go install`?

tsachiherman: I think that it used to be required before go 1.12. I'm not sure if its still applies.

onetechnical: I think it was auto (built in 1.13+ if go.mod was present) but as of 1.16 it defaults to on. We should probably remove setting it explicitly here.

brianolson: It defaults to on now, but it doesn't hurt to leave it on and might help developers a little behind in keeping their environment up to date. I'd be fine with deleting this in another 6 months or so. It would be nice if `make` reported a warning "Please upgrade your golang to 1.16 or greater".

algojack: good point, removed all references to GO111MODULE=on

onetechnical: To Brian's comment: I think it's okay to set our minimum version for support higher. I believe we call `scripts/check_golang_version.sh` and explicitly tell people to upgrade, if they run configure_dev.sh.

winder: It looks like GO111MODULE defaulting to on is linked to GOPATH being deprecated: https://github.com/golang/go/issues/41330

I think it would be a good idea to leave `GO111MODULE=on` until we completely remove `GOPATH`

brianolson: refinement on GOPATH from that golang issue since I was confused and had to read up on it:
$GOPATH/src is going away as a source organization strategy, replaced by modules (and their 99% invisible cache structure inside $GOPATH/pkg). Any build assumption about `go-algorand/` being at `$GOPATH/src/github.com/algorand/go-algorand/` should really really go away (I'm pretty sure it's gone from build but it lingered in integration tests for a while).

$GOPATH/bin is sticking around as the output directory for `go install`

winder: Thanks for the clarification Brian. I would guess setting 1.16 in the `go.mod` file would force people to upgrade regardless of `scripts/check_golang_version.sh`, so removing all references to `GO111MODULE=on` should be fine.

algorandskiy:
Before merging: ask at least 2-3 folks to fetch and build locally to ensure we support different local setups. For example with go1.17, `x/sys` requires an upgrade in order to run tests on MacOS without crashing.

yaovi-a:
@tsachiherman we'd like to get this into a release soon. What code changes are necessary to make sure we don't miss anything?

yaovi-a:
We also need additional integration, e2e and unit test to be executed again.

cce:
Could you run `go mod tidy` on this branch? With go 1.16 there is this new behavior and we could benefit from updating to have full checksums. I think this is why the travis windows build is failing, also.
>Build commands like go build and go test no longer modify go.mod and go.sum by default. Instead, they report an error if a module requirement or checksum needs to be added or updated (as if the -mod=readonly flag were used). Module requirements and sums may be adjusted with go mod tidy or go get.
https://golang.org/doc/go1.16

algojack:
> Could you run `go mod tidy` on this branch? With go 1.16 there is this new behavior (below) and we could benefit from updating to have full checksums. I think this is why the travis windows build is failing, also.
> 
> > Build commands like `go build` and `go test` no longer modify go.mod and go.sum by default. Instead, they report an error if a module requirement or checksum needs to be added or updated (as if the -mod=readonly flag were used). Module requirements and sums may be adjusted with go mod tidy or go get.
> > https://golang.org/doc/go1.16

do you want to run it once or make it run automatically somewhere?

cce:
No, I think it is fine to be updated as part of regular development. I had noticed some missing checksums on this branch, but they were actually added to master in #3177, so when you merged in master just now it brought in the missing checksums I noticed that go 1.16 would complain about.

winder:
> > since you opened this, Go released two more minor versions (including security and bug fixes) and now 1.16.9 is the current version.
> 
> Thanks, updated to 1.16.9

1.16.10 is out now!

algobarb (dismissed):

winder (approved):

egieseke (dismissed):
At one point, we had consolidated the golang version to one file.  Not sure why it's spread across so many files again.

yaovi-a:
Don't merge until 3.2.0 go-algorand beta is cut.

winder (approved):

faddat:
I'd like to recommend 1.17, and the specific reason for that is the clarity changes that you get in go.mod

 worries about determinism faults?

algojack:
> I'd like to recommend 1.17, and the specific reason for that is the clarity changes that you get in go.mod
> 
> worries about determinism faults?

I'm all for 1.17, but it requires more changes. Since it took us more than 3 months to get this in, we should probably open a separate ticket for 1.17.

cce (approved):

algorandskiy (approved):
Works on MacOS 10.15 so LGTM

tsachiherman (approved):

algojack:
Thank you guys!!

PR #2828, opened 2021-09-02
CodeQL PR

algojack:
## Summary

Looking at enabling CodeQL

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -13,7 +13,7 @@ name: "CodeQL"
 
 on:
   push:
-    branches: [rel/beta]
+    branches: [ master, rel/beta ]
   pull_request:
     # The branches below must be a subset of the branches above
     branches: [ master ]
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -13,10 +13,10 @@ name: "CodeQL"
 
 on:
   push:
-    branches: [ master, rel/beta ]
+    branches: [ master, rel/nightly, rel/beta, rel/stable ]
   pull_request:
     # The branches below must be a subset of the branches above
-    branches: [ master ]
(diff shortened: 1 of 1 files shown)


.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,71 @@
...
+        languages: ${{ matrix.language }}
+        # If you wish to specify custom queries, you can do so here or in a config file.
+        # By default, queries listed here will override any specified in a config file.
+        # Prefix the list here with "+" to use these queries and those in the config file.
+        # queries: ./path/to/local/query, your-org/your-repo/queries@main
+
+    # Autobuild attempts to build any compiled languages  (C/C++, C#, or Java).
+    # If this step fails, then you should remove it and run the build manually (see below)
+    - name: Autobuild
+      uses: github/codeql-action/autobuild@v1
```
onetechnical, line 57: We may need to change this, or add dependencies to install boost-math.

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,71 @@
...
+# ******** NOTE ********
+# We have attempted to detect the languages in your repository. Please check
+# the `language` matrix defined below to confirm you have the correct set of
+# supported CodeQL languages.
+#
+name: "CodeQL"
+
+on:
+  push:
+    branches: [rel/beta]
```
onetechnical, line 16: Saw warnings that we should add master here.

onetechnical (changes requested):
This is largely just parroting the feedback from CodeQL itself. Does this mean we're actually _clean_ for CodeQL review right now?

.github/workflows/codeql-analysis.yml
algojack: ah, i deleted one too many since i thought we are limited on minutes. I put it back. Thank you.

algojack:
seems we need to do more things to make it work. Closing for now.

PR #2834, opened 2021-09-03
txnsync: bloom filter refactoring

tsachiherman:
## Summary

Refactor the bloom filter implementation to have two "implementations" : `testableBloomFilter` and `bloomFilter`.

`testableBloomFilter` is used exclusively when being received from the network; it does not contain any encoded
fields, nor does it include `transactionsRange` which is used for generating new bloom filters.

 `bloomFilter` on the flip side is used only for outgoing bloom filters. It doesn't support `testing`, and it contains only the encoded bloom filter ( and not the underlying bloom filter needed for testing, for instance ).

This separation allows us to ensure we don't store bloom filter object beyond the required scope, improving the memory utilization.

## Test Plan

This PR does not introduce any new functionality. The existing unit tests were updated to use the updated constructs.

### DIFF
--- a/txnsync/bloomFilter_test.go
+++ b/txnsync/bloomFilter_test.go
@@ -36,6 +36,7 @@ import (
 func getTxnGroups(genesisHash crypto.Digest, genesisID string) []pooldata.SignedTxGroup {
 	return []pooldata.SignedTxGroup{
 		pooldata.SignedTxGroup{
+			GroupCounter:       0,
 			GroupTransactionID: transactions.Txid{1},
 			Transactions: []transactions.SignedTxn{
 				{
@@ -56,6 +57,7 @@ func getTxnGroups(genesisHash crypto.Digest, genesisID string) []pooldata.Signed
 			},
 		},
 		pooldata.SignedTxGroup{
+			GroupCounter:       1,
 			GroupTransactionID: transactions.Txid{2},
 			Transactions: []transactions.SignedTxn{
 				{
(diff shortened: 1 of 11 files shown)


txnsync/bloomFilter_test.go
```diff
@@ -290,32 +298,28 @@ func TestBloomFilterTest(t *testing.T) {
...
+		}
+		var enc encodedBloomFilter
+		enc.BloomFilterType = byte(filterType)
+		enc.BloomFilter, err = filter.MarshalBinary()
+		require.NoError(t, err)
+
+		testableBf, err := decodeBloomFilter(enc)
+		require.NoError(t, err)
+		for _, tx := range txnGroups {
+			ans := testableBf.test(tx.GroupTransactionID)
```
algonautshant, line 318: Why is the Modulator variations dropped here?
test checks the modulator value.

algonautshant (approved):
Looks great. 
A few minor comments.

txnsync/peer.go
```diff
@@ -373,7 +374,7 @@ func (p *Peer) selectPendingTransactions(pendingTransactions []pooldata.SignedTx
 		effectiveBloomFilters = append(effectiveBloomFilters, filterIdx)
 	}
 
-	//removedTxn := 0
+	// removedTxn := 0
```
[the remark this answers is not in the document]
tsachiherman: It's used for debugging. We should probably should figure if we want to keep it there or not. In this PR, I've just added a space so that the linter won't give me a warning about that.

txnsync/bloomFilter_test.go
tsachiherman: good point. I'll add that back. ( i.e. I think that we don't need this for the encoding part.. but we want to verify the testing )

algonautshant (approved):

PR #2836, opened 2021-09-03
Display app address in `goal app info` & show logs are binary in REST API

jasonpaulos:
## Summary

This is a follow up to #2661. It adds the app account address to the `goal app info` command, and it changes the log item type format in the REST API to "byte" in order to show that it's binary data. As a bonus, the encoder will automatically convert to base64 for us.

## Test Plan

Added a unit test to verify the encoder converts logs to base64.


jannotti (dismissed):
I dig the auto base64.

Linter is mad about partitiontest.

jannotti (approved):

PR #2837, opened 2021-09-03
Use byte type for group indexes

jannotti:
We had some places where we cast uint64 teal values to ints, which we
used as group indexes.  This could lead to negative references and
panics (which we caught, but still...).

This PR fixes them and goes further - it uses a byte type in many
places where a txgroup index is expected.  There's an argument against
that - as we do end up casting some slice indexes to bytes, which
could lose precision if we allow groups to have more that 256
transactions.  On the other hand, that bug seems less likely than the
kind of bug we had.

Discourse welcome.


jannotti:
Nah. There's too much going on here.  I'll limit the fix to teal opcodes.

PR #2838, opened 2021-09-04
global GroupID

jannotti:
Add GroupID as an accessible global field in AVM.

Co-authored-by: Zach Langley <zachary.langley@algorand.com>


algochoi (approved):
Looks good to me 👍

PR #2839, opened 2021-09-04
Use math.bits intrinsics to simplify addw/mulw opcodes

jannotti:
## Summary

Removes some code, but more importantly speeds up mulw considerably.
We have always charged 1 unit of cost for mulw, but it was actually
more expensive.  This fixes that.

A nice side-effect is that opPlus/opAddw and opMul/opMulw are
simplified and easier to follow.  Benchmarks confirm nothing is
slower (and mulw is ~5 times faster).

## Test Plan

All tests of opAddwImpl and opMulwImpl have been retained by
conducting the same test with actual teal, instead of directly testing
those (now non-existent) functions.

### DIFF
--- a/data/transactions/logic/eval_test.go
+++ b/data/transactions/logic/eval_test.go
@@ -505,35 +505,14 @@ func TestMulOverflow(t *testing.T) {
 	testPanics(t, "int 0x111111111; int 0x222222222; *; pop; int 1", 1)
 }
 
-func TestMulwImpl(t *testing.T) {
-	partitiontest.PartitionTest(t)
-
-	t.Parallel()
-	high, low, err := opMulwImpl(1, 2)
-	require.NoError(t, err)
-	require.Equal(t, uint64(0), high)
-	require.Equal(t, uint64(2), low)
-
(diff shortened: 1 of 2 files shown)


data/transactions/logic/eval_test.go
```diff
@@ -552,31 +531,14 @@ int 1                   // ret 1
...
-	partitiontest.PartitionTest(t)
-
-	t.Parallel()
-	carry, sum := opAddwImpl(1, 2)
-	require.Equal(t, uint64(0), carry)
-	require.Equal(t, uint64(3), sum)
-
-	carry, sum = opAddwImpl(0xFFFFFFFFFFFFFFFD, 0x45)
-	require.Equal(t, uint64(1), carry)
-	require.Equal(t, uint64(0x42), sum)
```
algorandskiy, line 565: could you please preserve this test as well?

jannotti: Yes, sorry I missed it while changing the format.

algorandskiy (approved):

PR #2846, opened 2021-09-06
ledger: refactor the ledger Totals

tsachiherman:
## Summary

The plan to change the tracker database to maintain the latest 320 rounds for the online accounts only have some other required modification; one of them is the semantics of the `Totals` method:
- At this time, it supports retrieving `AccountTotals` for any of the recent 320 rounds.
- We need the method to support the `AccountTotals` for the latest round only, as well as provide the circulating supply for latest-320 rounds ( i.e. the circulating supply is a subset of the `AccountTotals` ).

Once the database implementation is complete, the database would contain the `AccountTotals` information for the latest round only, plus the online circulation for the past 320 rounds.

In order to support that, I've broken up the interface into:
- `LatestTotals`, which would return the latest totals as well as the latest round associated with these totals.
- `OnlineTotals`, which receive a round number and return the online totals associated with that round.

This change is focused around the Ledger interface. Further changes would be required in the accounts update before the high-level goals could be achieved.

## Test Plan

Unit tests updated.


tsachiherman:
@tolikzinovyev - this change would impact the indexer due to the change in `ledgerForEvaluator`. I expect that the corresponding change in the indexer would be trivial, though.

ledger/acctupdates.go
```diff
@@ -941,6 +941,13 @@ func (au *accountUpdates) totalsImpl(rnd basics.Round) (totals ledgercore.Accoun
 	return
 }
 
+// latestTotalsImpl returns the totals of all accounts for the most recent round, as well as the round number
+func (au *accountUpdates) latestTotalsImpl() (basics.Round, ledgercore.AccountTotals, error) {
+	offset := len(au.deltas)
+	rnd := au.dbRound + basics.Round(len(au.deltas))
+	return rnd, au.roundTotals[offset], nil
```
tolikzinovyev, line 948: Might be more straight-forward to do `au.roundTotals[len(au.roundTotals) - 1]`

ledger/ledger.go
```diff
@@ -463,11 +463,24 @@ func (l *Ledger) LookupWithoutRewards(rnd basics.Round, addr basics.Address) (ba
...
 
-// Totals returns the totals of all accounts at the end of round rnd.
-func (l *Ledger) Totals(rnd basics.Round) (ledgercore.AccountTotals, error) {
+// LatestTotals returns the totals of all accounts for the most recent round, as well as the round number.
+func (l *Ledger) LatestTotals() (basics.Round, ledgercore.AccountTotals, error) {
 	l.trackerMu.RLock()
 	defer l.trackerMu.RUnlock()
-	return l.accts.Totals(rnd)
+	rnd := l.blockQ.latest()
+	totals, err := l.accts.Totals(rnd)
```
tolikzinovyev, line 471: why aren't we using `LatestTotals()`?

data/pools/transactionPool.go
```diff
@@ -894,7 +902,15 @@ func (pool *TransactionPool) assembleEmptyBlock(round basics.Round) (assembled *
 	next := bookkeeping.MakeBlock(prev)
 	blockEval, err := pool.ledger.StartEvaluator(next.BlockHeader, 0)
 	if err != nil {
-		err = fmt.Errorf("TransactionPool.assembleEmptyBlock: cannot start evaluator for %d: %v", round, err)
+		var nonSeqBlockEval ledgercore.ErrNonSequentialBlockEval
+		if errors.As(err, &nonSeqBlockEval) {
+			if nonSeqBlockEval.EvaluatorRound <= nonSeqBlockEval.LatestRound {
+				// in the case that the ledger have already moved beyond that round, just let the agreement know that
+				// we don't generate a block and it's perfectly fine.
+				return nil, ErrStaleBlockAssemblyRequest
```
tolikzinovyev, line 910: In one instance, the caller of this function `AssembleBlock()` will modify this error like this.
```
emptyBlockErr = fmt.Errorf("AssembleBlock: failed to construct empty block : %v", emptyBlockErr)
```

However, the caller of `AssembleBlock()` in `node/node.go` will check 
```
if err == pools.ErrStaleBlockAssemblyRequest {
```

We need to change `%v` to `%w` and change the equality to `errors.As()`.

tsachiherman: yes, I agree.

ledger/acctupdates.go
tsachiherman: What you wrote is correct; but it's less consistent with the rest of the code that is using the `len(au.deltas)`.
There is also the theoretical risk of having a negative value that doesn't apply here.

ledger/ledger.go
tsachiherman: the l.accts doesn't have LatestTotals().

data/pools/transactionPool.go
tolikzinovyev: right

ledger/ledger.go
tolikzinovyev: Right, you added `latestTotalsImpl()` to `accountUpdates`. We can just make it public?

tsachiherman: no, there is a need for a `au.LatestTotals`, which would call `au.latestTotalsImpl` while taking the read mutex.

data/pools/transactionPool.go
```diff
@@ -667,6 +668,13 @@ func (pool *TransactionPool) recomputeBlockEvaluator(committedTxIds map[transact
 	}
 	pool.pendingBlockEvaluator, err = pool.ledger.StartEvaluator(next.BlockHeader, hint)
 	if err != nil {
+		var nonSeqBlockEval ledgercore.ErrNonSequentialBlockEval
+		if errors.As(err, &nonSeqBlockEval) {
+			if nonSeqBlockEval.EvaluatorRound <= nonSeqBlockEval.LatestRound {
+				pool.log.Infof("TransactionPool.recomputeBlockEvaluator: skipped creating block evaluator for round %d since ledger already caught up with that round", nonSeqBlockEval.EvaluatorRound)
+				return
+			}
+		}
```
algorandskiy, line 677: this looks not related to the change, has it come from another branch?

algorandskiy (approved):
Looks OK to merge

data/pools/transactionPool.go
tsachiherman: No. The startevaluator has another special fail case.
This change add a special handling for that case.
( it's not a real error, and shouldn't be treated as such.. )

algorandskiy: Yes, I see, a new error from `StartEvaluator`, makes sense to have a special handling, thanks for pointing.

PR #2847, opened 2021-09-07
Add opcodes for dynamically indexing into Txn array fields

### LINKED ISSUE #2799: Support dynamic indexing into txn array fields
## Problem

Thanks to the `gtxns` and `gtxnsa` opcodes, TEAL programs can dynamically index into the group transaction array to access any transaction's fields.

However, it is not currently possible to dynamically index into the array fields of a specific transaction (`ApplicationArgs`, `Accounts`, etc.). This is a useful feature in general, and specifically it needs to be added in order to support foreign arrays like https://github.com/algorandfoundation/ARCs/pull/7 proposes.

## Solution

Introduce 3 new opcodes:
* `txnas` like `txna`, but it takes the array index as a stack arg. Usage: `int <array-index>; txnas ApplicationArgs`
* `gtxnas` like `gtxna`, but it takes the array index as a stack arg. Usage: `int <array-index>; gtxnas <gtxn-index> ApplicationArgs`
* `gtxnsas` like `gtxnsa`, but it takes the array index as an additional stack arg. Usage: `int <gtxn-index>; int <array-index>; gtxnsas ApplicationArgs`

Adding all 3 opcodes might be overkill (we may want to drop `gtxnas`), but I've listing them all for completeness.

## Dependencies

This would also require PyTeal support.

## Urgency

This should make it into TEAL 5.

algochoi:
## Summary

This commit adds `txnas`, `gtxnas`, and `gtxnsas` op codes that allow for array indices to be dynamically pushed into the stack instead of being supplied by immediate values. It also adds the `args` op code to dynamically specify an arg index from the stack.

TODO: Implement `stores`, `loads`

Closes #2799 

## Test Plan

* Unit tests


data/transactions/logic/README.md
```diff
@@ -221,6 +221,9 @@ Some of these have immediate data in the byte or bytes after the opcode.
 | `gloads i` | push Ith scratch space index of the Xth transaction in the current group |
 | `gaid t` | push the ID of the asset or application created in the Tth transaction of the current group |
 | `gaids` | push the ID of the asset or application created in the Xth transaction of the current group |
+| `txnas f` | pop an index A. push Ath value of the array field F of the current transaction |
+| `gtxnas t f` | pop an index A. push Ath value of the array field F from the Tth transaction in the current group |
+| `gtxnsas f` | pop an index A and an index B. push Bth value of the array field F from the Ath transaction in the current group |
```
jannotti, line 226: The convention, when an opcode has only one stack argument, is to call it `X`.  A bit weord, but let's stick to it until/unless we change it everywhere.

data/transactions/logic/eval.go
```diff
@@ -2062,6 +2062,29 @@ func opTxna(cx *evalContext) {
 	cx.stack = append(cx.stack, sv)
 }
 
+func opTxnas(cx *evalContext) {
+	last := len(cx.stack) - 1
+
+	field := TxnField(uint64(cx.program[cx.pc+1]))
```
jannotti, line 2068: Let's remove the inner cast, which doesn't do anything.  (I know we're doing this all over the place, I have a PR I'm working on the fix the existing ones.)

data/transactions/logic/eval.go
```diff
@@ -2122,6 +2145,35 @@ func opGtxna(cx *evalContext) {
 	cx.stack = append(cx.stack, sv)
 }
 
+func opGtxnas(cx *evalContext) {
+	last := len(cx.stack) - 1
+
+	gtxid := int(uint(cx.program[cx.pc+1]))
```
jannotti, line 2151: Same, about inner cast.

data/transactions/logic/eval.go
```diff
@@ -2122,6 +2145,35 @@ func opGtxna(cx *evalContext) {
...
+	if gtxid >= len(cx.TxnGroup) {
+		cx.err = fmt.Errorf("gtxnas lookup TxnGroup[%d] but it only has %d", gtxid, len(cx.TxnGroup))
+		return
+	}
+	tx := &cx.TxnGroup[gtxid].Txn
+	field := TxnField(uint64(cx.program[cx.pc+2]))
```
jannotti, line 2157: again

data/transactions/logic/eval.go
```diff
@@ -2184,6 +2236,37 @@ func opGtxnsa(cx *evalContext) {
...
 }
 
+func opGtxnsas(cx *evalContext) {
+	last := len(cx.stack) - 1
+	prev := last - 1
+
+	gtxid := int(cx.stack[prev].Uint)
+	if gtxid >= len(cx.TxnGroup) {
+		cx.err = fmt.Errorf("gtxnsa lookup TxnGroup[%d] but it only has %d", gtxid, len(cx.TxnGroup))
+		return
```
jannotti, line 2246: need to see if gtxid < 0. (Again, also, I know many existing places forget this check.)

data/transactions/logic/eval.go
```diff
@@ -2184,6 +2236,37 @@ func opGtxnsa(cx *evalContext) {
...
+	}
+	tx := &cx.TxnGroup[gtxid].Txn
+	field := TxnField(uint64(cx.program[cx.pc+1]))
```
jannotti, line 2249: inner cast

data/transactions/logic/eval_test.go
```diff
@@ -1788,20 +1796,34 @@ int 1
...
 int 0
 ==
 &&
 gtxn 1 ExtraProgramPages
 int 2
 ==
 &&
+`
+
+	gtxnText := gtxnTextV4 + `int 0
```
jannotti, line 1809: Let's reduce further churn by naming the gtxnTextV5 now, rather than wait for a later one.

data/transactions/logic/opcodes.go
```diff
@@ -298,6 +298,11 @@ var OpSpecs = []OpSpec{
 
 	// ABI support opcodes.
 	{0xb0, "log", opLog, asmDefault, disDefault, oneBytes, nil, 5, runModeApplication, opDefault},
+
+	// Dynamically indexing into LogicSigs
+	{0xc0, "txnas", opTxnas, assembleTxnas, disTxn, oneInt, oneAny, 5, modeAny, immediates("f")},
+	{0xc1, "gtxnas", opGtxnas, assembleGtxnas, disGtxn, oneInt, oneAny, 5, modeAny, immediates("t", "f")},
+	{0xc2, "gtxnsas", opGtxnsas, assembleGtxnsas, disTxn, twoInts, oneAny, 5, modeAny, immediates("f")},
```
jannotti, line 305: Is there room to put these near the other txn opcodes?

jannotti (changes requested):
We need to think about whether we should add some pseudo op support.  Currently, `txn ApplicationArgs 0` gets compiled to `txna ApplicationArgs 0`, because of the two arguments signaling that it's an array ref. `txnsa` can also be compiled from `txns` (I think).

I'm not sure we can do much here, but it is a little strange that once you move to the stack versions, the "a" becomes required.

data/transactions/logic/README.md
algochoi: Sounds good - I'll keep the `gtxnsas` description, since it pops two stack arguments then.

data/transactions/logic/eval.go
algochoi: Good point - I'll delete the new inner casts

data/transactions/logic/opcodes.go
algochoi: It's a good thought, but there are only 2 slots left there, and need to make 3 more op codes, on top of the three here. I'm open to suggestions on tucking these 6 op codes somewhere else though.

data/transactions/logic/eval_test.go
```diff
@@ -1790,20 +1796,34 @@ int 1
...
 int 0
 ==
 &&
 gtxn 1 ExtraProgramPages
 int 2
 ==
 &&
+`
+
+	gtxnTextV5 := gtxnTextV4 + `int 0
```
jasonpaulos, line 1809: This is missing `txnas`

jannotti (approved):

data/transactions/logic/eval.go
```diff
@@ -2133,6 +2163,38 @@ func opGtxna(cx *EvalContext) {
...
+	last := len(cx.stack) - 1
+
+	gtxid := int(cx.program[cx.pc+1])
+	if gtxid >= len(cx.TxnGroup) {
+		cx.err = fmt.Errorf("gtxnas lookup TxnGroup[%d] but it only has %d", gtxid, len(cx.TxnGroup))
+		return
+	} else if gtxid < 0 {
+		cx.err = fmt.Errorf("gtxnas lookup %d cannot be negative", gtxid)
+		return
+	}
```
jannotti, line 2176: I'll fix this up when I do loads, stores, but just want to point out that we don't want to complain that the number is negative.  In AVM, it isn't.  It's only negative (potentially) because this code casted a big uint64 to an int, and that can give you a negative result.

algochoi: Makes sense!

data/transactions/logic/assembler.go
```diff
@@ -980,6 +1034,27 @@ func assembleGtxnsa(ops *OpStream, spec *OpSpec, args []string) error {
 	return nil
 }
 
+func assembleGtxnsas(ops *OpStream, spec *OpSpec, args []string) error {
+	if len(args) != 1 {
+		return ops.error("gtxnsas expects one immediate argument")
```
jannotti, line 1039: Let's try to use %s and spec.Name as often as possible.

algochoi: Agreed -  I think we can potentially save ourselves time in the future by doing this.

data/transactions/logic/eval.go
jannotti: PS, I liked the way you handled this in args a lot.  There, you used uint64 mostly, and casted the len() calls instead of the stack argument.  I think I will fix this stuff up that way, instead of the way I had been contemplating.

data/transactions/logic/eval_test.go
```diff
@@ -2036,6 +2056,85 @@ global ZeroAddress
...
+	var txn transactions.SignedTxn
+	txn.Txn.Accounts = make([]basics.Address, 1)
+	txn.Txn.Accounts[0] = txn.Txn.Sender
+	txn.Txn.ApplicationArgs = make([][]byte, 1)
+	txn.Txn.ApplicationArgs[0] = []byte(protocol.PaymentTx)
+	txgroup := make([]transactions.SignedTxn, 1)
+	txgroup[0] = txn
+	ep := defaultEvalParams(nil, &txn)
+	ep.TxnGroup = txgroup
+	_, err := Eval(ops.Program, ep)
```
jasonpaulos, line 2080: I think you're just checking that the programs runs, not that it passes? It would be a stronger test to instead check that a simple use of `txnas` passes. I'd suggest something like the following, and you could adapt it for `gtxnas` and `gtxnsas` too.

```
int 0
txnas Accounts
txna Accounts 0
==
int 0
txnas ApplicationArgs
txna ApplicationArgs 0
==
&&
int 1
txnas ApplicationArgs
txna ApplicationArgs 1
==
&&
return
```

data/transactions/logic/eval_test.go
```diff
@@ -903,18 +903,11 @@ func TestArg(t *testing.T) {
...
-!=
-&&
-arg 4
-len
-int 9
-<
-&&`, v)
+			source := "arg 0; arg 1; ==; arg 2; arg 3; !=; &&; arg 4; len; int 9; <; &&;"
+			if v >= 5 {
+				source += "int 0; args; int 1; args; ==; assert;"
```
jasonpaulos, line 908: Could you also add a check that `int 2; args;` != `int 3; args;` too? These tests should protect against a bug where the opcode does not handle the index properly (e.g. ignores the stack arg and always assumes index 0 or something), and this helps prove the index is actually being used correctly.

jasonpaulos (commented):
Mostly good, I just have some test suggestions. It would be nice if some code could be reused between the similar opcodes, but that can wait until later.

data/transactions/logic/eval_test.go
algochoi: I put `txnas` in the `testTxnProgramTextV5` section I think.

data/transactions/logic/eval_test.go
algochoi: Good point, I will check these with e.g. `testAccepts(t, "int 0; txnas Accounts; txna Accounts 0; ==", 5)`

jannotti: I merged what existed, I'll add these.

jannotti: I concluded these were fine.  The following tests checked `pass`, and they were grabbing the accounts and comparing them to app args.

data/transactions/logic/eval_test.go
jannotti: added in my followup

PR #2849, opened 2021-09-07
tx_field tables

jannotti:
Drive the checking of whether tx_field can set a particular field in a table-driven way, rather than by simple switch.  This is cleaner as we add functionality in later versions, and will allow generating the spec for tx_field.


data/transactions/logic/TEAL_opcodes.md
```diff
@@ -18,7 +18,7 @@ Ops have a 'cost' of 1 unless otherwise specified.
 - SHA256 hash of value X, yields [32]byte
 - **Cost**:
    - 7 (LogicSigVersion = 1)
-   - 35 (2 <= LogicSigVersion <= 5)
+   - 35 (LogicSigVersion => 2)
```
brianolson, line 21: `>=` would be more conventional, and less confused with an arrow?

jannotti: oh, geez, you're right of course.  thanks.

algorandskiy (approved):

PR #2851, opened 2021-09-07
txnsync: Add linked list based short term lru cache

tsachiherman:
## Summary

Existing short term cache in the transaction sync was not supporting promoting updated cache entries.
In this PR, I have replaced the circulating buffer with a dual linked list to allow promoting cached entries.

## Test Plan

Extend existing unit test

### DIFF
--- a/txnsync/transactionCache.go
+++ b/txnsync/transactionCache.go
@@ -53,12 +53,19 @@ type ackPendingTxids struct {
 	timestamp time.Duration
 }
 
+//msgp:ignore shortTermCacheEntry
+type shortTermCacheEntry struct {
+	txid transactions.Txid
+	prev *shortTermCacheEntry
+	next *shortTermCacheEntry
+}
+
 //msgp:ignore shortTermTransactionCache
 type shortTermTransactionCache struct {
 	size            int
-	transactionsIDs []transactions.Txid
-	transactionsMap map[transactions.Txid]bool
-	oldest          int
+	first           *shortTermCacheEntry
+	free            *shortTermCacheEntry
+	transactionsMap map[transactions.Txid]*shortTermCacheEntry
 }
 
 //msgp:ignore longTermTransactionCache
@@ -73,8 +80,7 @@ func makeTransactionCache(shortTermSize, longTermSize, pendingAckTxids int) *tra
 	txnCache := &transactionCache{
 		shortTermCache: shortTermTransactionCache{
 			size:            shortTermSize,
-			transactionsIDs: make([]transactions.Txid, shortTermSize),
-			transactionsMap: make(map[transactions.Txid]bool, shortTermSize),
+			transactionsMap: make(map[transactions.Txid]*shortTermCacheEntry, shortTermSize),
 		},
 		ackPendingTxids: make([]ackPendingTxids, 0, pendingAckTxids),
 		longTermCache: longTermTransactionCache{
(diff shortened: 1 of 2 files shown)


txnsync/transactionCache.go
```diff
@@ -169,31 +198,100 @@ func (lru *transactionCache) acknowledge(seqs []uint64) {
 
 // add a given transaction ID to the short term cache.
 func (st *shortTermTransactionCache) add(txid transactions.Txid) {
-	if st.transactionsMap[txid] {
+	entry, exists := st.transactionsMap[txid]
+	if exists {
+		// promote
+		if entry.detach() {
```
brianolson, line 204: if `entry == st.first` we could just `st.first=entry.next` and not detach+insert it?

brianolson (approved):
LGTM. One possible rearrange

txnsync/transactionCache.go
```diff
@@ -53,12 +53,21 @@ type ackPendingTxids struct {
 	timestamp time.Duration
 }
 
+// shortTermCacheEntry is used as the data container of a single linked list item
```
algorandskiy, line 56: single linked list -> double linked list since it has next/prev

txnsync/transactionCache.go
```diff
@@ -68,13 +77,33 @@ type longTermTransactionCache struct {
...
+	if ce.next == ce.prev {
+		return false
+	}
+	ce.prev.next = ce.next
+	ce.next.prev = ce.prev
+	return true
+}
+
+// addToList add the element to the tail of the list who's head is firstListEntry.
+func (ce *shortTermCacheEntry) addToList(firstListEntry *shortTermCacheEntry) {
```
algorandskiy, line 93: firstListEntry -> head, as the most common name, so it would eliminate half of the comment :)

txnsync/transactionCache.go
```diff
@@ -53,12 +53,21 @@ type ackPendingTxids struct {
...
+}
+
 //msgp:ignore shortTermTransactionCache
 type shortTermTransactionCache struct {
-	size            int
-	transactionsIDs []transactions.Txid
-	transactionsMap map[transactions.Txid]bool
-	oldest          int
+	size            int                                        // the maximum number of elements in the short term cache
+	first           *shortTermCacheEntry                       // pointer to first element
```
algorandskiy, line 68: first -> head?
Also a comment that first/head is the oldest entry

txnsync/transactionCache.go
```diff
@@ -169,31 +198,100 @@ func (lru *transactionCache) acknowledge(seqs []uint64) {
...
+		return
+	}
+	used := st.first
+	free := st.free
+	free.prev.next = used
+	used.prev.next = free
+	lastFree := free.prev
+	free.prev = used.prev
+	used.prev = lastFree
+	st.first = nil
```
algorandskiy, line 294: could you please add a test that calls `reset` in 2 used and 3 free nodes and ensures `st.free` has all 5 nodes reachable in both directions?

txnsync/transactionCache.go
algorandskiy: sounds right but not sure if the code will become clearer.

algorandskiy (commented):
Looks good, couple comments

txnsync/transactionCache.go
tsachiherman: sure.

PR #2852, opened 2021-09-07
ECDSA secp256k1

algorandskiy:
## Summary

* `ecdsa_verify`, `ecdsa_pk_decompress`, `ecdsa_pk_recover` opcodes.
* crypto/secp256k1 sources from go-ethereum 1.10.7 committed into the repo.
* costs calculated (see below).
* so far only secp256k1 curve is supported.
* `ecdsa_verify` and `ecdsa_pk_recover` work with two-component keys (two 32 bytes long numbers) and not a single 33-bytes compressed or 65-bytes uncompressed keys.

## Test Plan

Added new tests.

## Cost calculation

Made as a ratio of `ed25519verify` cost
```
    BenchmarkEcDsa/ecdsa_verify-8              17341             67335 ns/op
    BenchmarkEcDsa/ecdsa_pk_decompress-8       42854             25714 ns/op
    BenchmarkEcDsa/ecdsa_pk_recover-8          15337             78249 ns/op
    
    BenchmarkEd25519Verifyx1-8         15244             75253 ns/op
```

## Reviewer Notes

Most of the code is `crypto/secp256k1` import from go-ethereum, please ignore.


data/transactions/logic/eval.go
```diff
@@ -2525,6 +2526,144 @@ func opEd25519verify(cx *EvalContext) {
 	cx.stack = cx.stack[:prev]
 }
 
+// opEcDsaVerify verifies ECDSA signature
+// it accepts PK in both compressed and uncompressed form because the underlying secp256k1.VerifySignature
+// decompressed it anyway. It also compressed PKs look more commonly used in Eth world rather uncompressed.
```
barnjamin, line 2531: It also _accepts_ compressed...?

algorandskiy: Thanks, fixed

tsachiherman:
Could you fork the eth secp256k1 to algorand/secp256k1 and use it from there ? ( and have the eth as upstream ).
( so that we won't be depending on the eth repo directly )

data/transactions/logic/eval.go
```diff
@@ -2525,6 +2526,145 @@ func opEd25519verify(cx *EvalContext) {
...
+	sigR := cx.stack[pprev].Bytes
+	sigS := cx.stack[fourth].Bytes
+	msg := cx.stack[fifth].Bytes
+
+	var pubkey []byte
+	// if PK y component is set then interpret it as uncompressed PK
+	if len(pkY) != 0 {
+		x := new(big.Int).SetBytes(pkX)
+		y := new(big.Int).SetBytes(pkY)
+		pubkey = secp256k1.CompressPubkey(x, y)
```
jasonpaulos, line 2563: Instead of compressing the PK and making `VerifySignature` decompress it, you can encode it as an uncompressed key. I believe that can be done with `secp256k1.S256().Marshal(x, y)`.

data/transactions/logic/eval.go
```diff
@@ -2525,6 +2526,145 @@ func opEd25519verify(cx *EvalContext) {
...
+		cx.err = fmt.Errorf("invalid recovery id: %d", recid)
+		return
+	}
+
+	signature := make([]byte, 0, len(sigR)+len(sigS)+1)
+	signature = append(signature, sigR...)
+	signature = append(signature, sigS...)
+	signature = append(signature, uint8(recid))
+
+	pk, err := secp256k1.RecoverPubkey(msg, signature)
```
jasonpaulos, line 2658: The returned `pk` is an [uncompressed public key](https://github.com/ethereum/go-ethereum/blob/d8ff53dfb8a516f47db37dbc7fd7ad18a1e8a125/crypto/secp256k1/ext.h#L34), and the description for this op says it returns a compressed public key. So you could do `x, y := secp256k1.S256().Unmarshal(pk)` to extract the x and y components to return those.

Returning an uncompressed key is useful for this function because Ethereum addresses are derived from the keccak hash of `x` concat `y`.

data/transactions/logic/doc.go
```diff
@@ -212,45 +218,48 @@ func OpImmediateNote(opName string) string {
...
-	"asset_params_get":  "params: Before v4, Txn.ForeignAssets offset. Since v4, Txn.ForeignAssets offset or an asset id that appears in Txn.ForeignAssets. Return: did_exist flag (1 if exist and 0 otherwise), value.",
-	"app_params_get":    "params: Txn.ForeignApps offset or an app id that appears in Txn.ForeignApps. Return: did_exist flag (1 if exist and 0 otherwise), value.",
-	"log":               "`log` can be called up to MaxLogCalls times in a program, and log up to a total of 1k bytes.",
+	"ed25519verify":       "The 32 byte public key is the last element on the stack, preceded by the 64 byte signature at the second-to-last element on the stack, preceded by the data which was signed at the third-to-last element on the stack.",
+	"ecdsa_verify":        "The 33 byte X-component of a public key is the last element on the stack, preceded by Y-component of a pubkey (can be empty, so the X-component then interpreted as a compressed pubkey), preceded by R and S components of a signature, preceded by the data that is fifth element on the stack.",
```
jasonpaulos, line 222: It's best not to mention byte lengths of the public keys and signatures, since different curves may have different sizes for these.

Also it is worth mentioning that the signed data [must be exactly 32 bytes](https://github.com/ethereum/go-ethereum/blob/90987db7334c1d10eb866ca550efedb66dea8a20/crypto/secp256k1/secp256.go#L127) (really a hash of the message). It may also be worth mentioning that [only signatures "in lower-S form"](https://github.com/ethereum/go-ethereum/blob/d8ff53dfb8a516f47db37dbc7fd7ad18a1e8a125/crypto/secp256k1/libsecp256k1/include/secp256k1.h#L366) will be accepted.

jasonpaulos (commented):
This is a great start. I have some comments on opcode details, but things are looking good in general.

data/transactions/logic/eval.go
```diff
@@ -2525,6 +2526,145 @@ func opEd25519verify(cx *EvalContext) {
 	cx.stack = cx.stack[:prev]
 }
 
+// opEcDsaVerify verifies ECDSA signature.
+// It accepts both compressed and uncompressed forms of PK because
```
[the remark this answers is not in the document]
algorandskiy: yes, I found while writing tests I confused uncompressed key with two components

data/transactions/logic/opcodes.go
```diff
@@ -142,6 +148,11 @@ var OpSpecs = []OpSpec{
 
 	{0x04, "ed25519verify", opEd25519verify, asmDefault, disDefault, threeBytes, oneInt, 1, runModeSignature, costly(1900)},
 	{0x04, "ed25519verify", opEd25519verify, asmDefault, disDefault, threeBytes, oneInt, 5, modeAny, costly(1900)},
+
+	{0x05, "ecdsa_verify", opEcDsaVerify, assembleEcDsa, disEcDsa, threeBytes.plus(twoBytes), oneInt, 5, modeAny, costlyImm(1900, "c")},
```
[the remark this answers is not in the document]
algorandskiy: set actual cost

data/transactions/logic/eval.go
algorandskiy: done

data/transactions/logic/eval.go
algorandskiy: returning x, y components now

data/transactions/logic/doc.go
algorandskiy: added

algorandskiy:
> Could you fork the eth secp256k1 to algorand/secp256k1 and use it from there ? ( and have the eth as upstream ).
Done

data/transactions/logic/README.md
```diff
@@ -139,6 +139,9 @@ For three-argument ops, `A` is the element two below the top, `B` is the penulti
 | `getbyte` | pop a byte-array A and integer B. Extract the Bth byte of A and push it as an integer |
 | `setbyte` | pop a byte-array A, integer B, and small integer C (between 0..255). Set the Bth byte of A to C, and push the result |
 | `concat` | pop two byte-arrays A and B and join them, push the result |
+| `ecdsa_verify c` | for (data A, signature B, C and pubkey D, E) verify the signature of the data against the pubkey => {0 or 1} |
```
jannotti, line 142: This is confusing because c is the name of the field (the immediate) and also (I think) the name you're using to describe the stack value.  I'd pick a new name for the immediate.

Also need to say what the immediate is/does.

data/transactions/logic/README.md
```diff
@@ -139,6 +139,9 @@ For three-argument ops, `A` is the element two below the top, `B` is the penulti
...
+| `ecdsa_pk_recover c` | for (data A, recovery id B, signature C, D) recover a public compressed key => [*... stack*, X, Y] |
+| `ecdsa_pk_decompress c` | decompress pubkey A into components X, Y => [*... stack*, X, Y] |
```
jannotti, line 144: Earlier comment applies to these too.

data/transactions/logic/doc.go
```diff
@@ -214,45 +220,48 @@ func OpImmediateNote(opName string) string {
...
-	"app_params_get":    "params: Txn.ForeignApps offset or an app id that appears in Txn.ForeignApps. Return: did_exist flag (1 if exist and 0 otherwise), value.",
-	"log":               "`log` can be called up to MaxLogCalls times in a program, and log up to a total of 1k bytes.",
+	"ed25519verify":       "The 32 byte public key is the last element on the stack, preceded by the 64 byte signature at the second-to-last element on the stack, preceded by the data which was signed at the third-to-last element on the stack.",
+	"ecdsa_verify":        "The 32 byte Y-component of a public key is the last element on the stack, preceded by X-component of a pubkey, preceded by S and R components of a signature, preceded by the data that is fifth element on the stack. The signed data must be 32 bytes long, and signatures in lower-S form are only accepted.",
+	"ecdsa_pk_decompress": "The 33 byte public key in a compressed form to be decompressed into X and Y (top) components.",
+	"ecdsa_pk_recover":    "S (top) and R elements of a signature, recovery id and data (bottom) are expected on the stack and used to deriver a public key in compressed format. The signed data must be 32 bytes long.",
```
jannotti, line 226: I think we should extend our convention for naming stack arguments. So for a five argument opcode, you can just refer to A,B,C,D,&E. (A is the deepest, E is top-of-stack).  I think that will simplify the text.

data/transactions/logic/doc.go
```diff
@@ -262,7 +271,7 @@ func OpDocExtra(opName string) string {
 
 // OpGroups is groupings of ops for documentation purposes.
 var OpGroups = map[string][]string{
-	"Arithmetic":            {"sha256", "keccak256", "sha512_256", "ed25519verify", "+", "-", "/", "*", "<", ">", "<=", ">=", "&&", "||", "shl", "shr", "sqrt", "bitlen", "exp", "==", "!=", "!", "len", "itob", "btoi", "%", "|", "&", "^", "~", "mulw", "addw", "divmodw", "expw", "getbit", "setbit", "getbyte", "setbyte", "concat"},
+	"Arithmetic":            {"sha256", "keccak256", "sha512_256", "ed25519verify", "+", "-", "/", "*", "<", ">", "<=", ">=", "&&", "||", "shl", "shr", "sqrt", "bitlen", "exp", "==", "!=", "!", "len", "itob", "btoi", "%", "|", "&", "^", "~", "mulw", "addw", "divmodw", "expw", "getbit", "setbit", "getbyte", "setbyte", "concat", "ecdsa_verify", "ecdsa_pk_recover", "ecdsa_pk_decompress"},
```
jannotti, line 274: Order here determines how they are presented, so I would put these near the ed25519verify opcode.

data/transactions/logic/eval.go
```diff
@@ -2526,6 +2527,164 @@ func opEd25519verify(cx *EvalContext) {
...
+	if size == len(data) {
+		return data, nil
+	}
+
+	buf := make([]byte, size)
+	copy(buf[size-len(data):], data)
+	return buf, nil
+}
+
+func opEcDsaVerify(cx *EvalContext) {
```
jannotti, line 2545: Why capital D? I don't actually know what the letters stand for but it seems like if this should be called EcDsa, that would imply the opcode should be ec_dsa.  I think I have normally seen them all run together, so I would only capitalize E here.

data/transactions/logic/fields.go
```diff
@@ -389,6 +389,41 @@ func (s gfNameSpecMap) getExtraFor(name string) (extra string) {
 	return
 }
 
+// EcDsaCurve is an enum for `ecdsa_` opcodes
+type EcDsaCurve int
```
jannotti, line 393: Same comment about the D, but it looks like this was very intentional, so if this how it's often broken up, fine by me.

jannotti (commented):
I don't feel qualified to comment on the crypto itself, but I've commented where I could. 

Thanks for adding work to document fields.  I'll will need that to finish up tx_field spec.

jasonpaulos (commented):
This is nearly there, I just have some minor suggestions.

Additionally, could you also add this test I made to verify compatibility with ethereum signatures? https://gist.github.com/jasonpaulos/47166d44c13e4efb18c7a1a09f4ac969

data/transactions/logic/eval.go
jasonpaulos: I agree capital D is a bit odd

tolikzinovyev:
Could we maybe have `crypto/secp256k1` in a separate repo?

data/transactions/logic/evalCrypto_test.go
```diff
@@ -0,0 +1,511 @@
...
+	rTampered := make([]byte, len(r))
+	copy(rTampered, pk)
+	rTampered[0] = 0
+
+	var verifyTests = []struct {
+		r    []byte
+		pass bool
+	}{
+		{r, true},
+		{rTampered, false},
```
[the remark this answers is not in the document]
algorandskiy: done

data/transactions/logic/fields.go
algorandskiy: done

algorandskiy (commented):
All reviews addressed except A, B, C,... naming in extended doc.

> Additionally, could you also add this test I made to verify compatibility with ethereum signatures? https://gist.github.com/jasonpaulos/47166d44c13e4efb18c7a1a09f4ac969

Added.

algorandskiy:
> Could we maybe have crypto/secp256k1 in a separate repo?

This is quite small and follows the approach of keeping crypto stuff under `crypto/' in go-algorand.

jasonpaulos (approved):

PR #2853, opened 2021-09-07
loads and stores opcodes that take scratch slot from stack

jannotti:
Lots of fixups for not checking unit64 to int casts as well.


algochoi (dismissed):
Overall looks good to me. Had one comment about naming. Thanks for fixing the `txnas` related parts 👍

data/transactions/logic/eval.go
```diff
@@ -2526,22 +2527,44 @@ func opEd25519verify(cx *EvalContext) {
 }
 
 func opLoad(cx *EvalContext) {
-	gindex := int(uint(cx.program[cx.pc+1]))
-	cx.stack = append(cx.stack, cx.scratch[gindex])
+	n := cx.program[cx.pc+1]
```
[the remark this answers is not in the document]
jannotti: I'm of two minds.  On the one hand, I get why a more descriptive name is generally better. On the other hand(s), 1. I don't mind brevity in very short functions, as these four are. 2. The fact that is was called gindex shows that people copy opcode implementations which, honestly is fine, especially for the boilerplate of arg grabbing. So using a generic name for the single argument seems appropriate so it is not misleading when copied.  I will yield to the next comment, either way.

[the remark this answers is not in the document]
jasonpaulos: I am leaning towards `n` since this is such a short function.

data/transactions/logic/eval.go
```diff
@@ -2526,22 +2527,44 @@ func opEd25519verify(cx *EvalContext) {
...
-	cx.scratch[gindex] = cx.stack[last]
+	cx.scratch[n] = cx.stack[last]
 	cx.stack = cx.stack[:last]
 }
 
-func opGloadImpl(cx *EvalContext, groupIdx int, scratchIdx int, opName string) (scratchValue stackValue, err error) {
-	if groupIdx >= len(cx.TxnGroup) {
+func opStores(cx *EvalContext) {
+	last := len(cx.stack) - 1
+	prev := last - 1
```
jasonpaulos, line 2553: I would prefer if we swap the order of arguments here. In opcodes which set a value (e.g. `app_global_put`), we normally make the first arg the location and the second the value. `<index>; <value>; stores` seems more intuitive given the behavior of our other opcodes.

jannotti: I don't feel incredibly strongly about it, but I'll tell you my reasoning and see if you find it convincing.

In the other cases where we have instructions with one version that takes a stack value instead of an immediate, we tacked it on to the "end", that is, top of stack. substring3, extract3, gtxnsas are the only "real" examples of this though, as they are the only ones where the non-stack version had stack arguments already, and so order matters.

What do you think?  I'm truly willing to go either way, whichever you think will give a bit more intuition to users.

jasonpaulos: From my point of view, `stores` is the general purpose version of the opcode, and `store <index>` in an optimized version that can be used when the index is a compile-time constant. So I think it's ok if the argument order of these two versions differ. I'm more interested in having `stores` make sense on its own.

We already have a similar situation with `gtxnas` and `gtxnsas`, since their order is `<array-index>; gtxnas <txn-index> <field>` vs `<txn-index>; <array-index>; gtxnsas <txn-field>`. The order of arguments isn't consistent there, but I think each of the opcodes makes sense on its own.

jannotti: ok, good enough for me.  I'll swap them.

jasonpaulos (approved):
Looks good

algochoi (approved):

PR #2854, opened 2021-09-08
Don't default to the current consensus version's fast recovery flag when ledger is catching up

cce:
## Summary

Log output like the following occurs when, from agreement's POV, the value of LedgerReader.NextRound() is ahead of the latest available value for LedgerReader.ConsensusVersion().
```
failed to read protocol version for fastTimeout event (proto ): ledger does not have entry 16070429 (latest 5090142, committed 5090142)
```

When the ledger is catching up, this may occur because demux has set up a fast recovery timeout deadline to feed a `fastTimeout` event to the agreement player, defaulting to using the current consensus protocol version's flag if the ledger does not know the current round's consensus parameters.

However, [later in player.handleFastTimeout()](https://github.com/algorand/go-algorand/blob/2b008855d6ade08e7b29d59914c3bda7becf249b/agreement/player.go#L125-L132), the timeout event's protocol version must be available, or handleFastTimeout() will not do anything (it needs an accompanying `FastRecoveryLambda` consensus parameter to issue fast votes). Instead, it will log this error, and refuse to issue fast votes without knowing the current round's consensus parameters. So this changes the default to disable setting up a fast recovery timeout channel when the current round's consensus version is not known, to prevent handleFastTimeout() from potentially being called later and logging this error.

## Test Plan

Existing tests should pass — but there may be a way to simulate this catchup behavior with a new test.


zeldovich (approved):

tsachiherman (commented):
I think that we can improve this fix, as it gives us all the right ingredients:
The error that we get is ErrNoEntry, which we can extract the Latest round from.
In case the latest round is greater than currentRound, there is no reason to write warnings of any sort.
Instead, we should just update the currentRound and try again.
( since this is a monotonic increasing value, there is no "risk" of getting stuck in a loop )
In case of other errors, I think that the proposed solution would work well.

cce:
Obviated by #3386 which gets rid of the FastPartitionRecovery consensus parameter altogether.

PR #2861, opened 2021-09-09
Increment TxnCounter with inner transactions

jannotti:
We increment the TxnCounter for inner transactions as well as outer, now.

The REST API used to assume it could determine a creatable ID based on the index of a transaction in a block and the block's TxnCounter. That is no longer true (since inner transactions could mean one payset transaction counts for more), so ApplyData now explicitly contains created IDs.  I keyed that ApplyData change off the same config parameter that turns on inner transactions.

### DIFF
--- a/daemon/algod/api/server/v1/handlers/handlers.go
+++ b/daemon/algod/api/server/v1/handlers/handlers.go
@@ -395,7 +395,7 @@ func computeCreatableIndexInPayset(tx node.TxnWithStatus, txnCounter uint64, pay
 // computeAssetIndexFromTxn returns the created asset index given a confirmed
 // transaction whose confirmation block is available in the ledger. Note that
 // 0 is an invalid asset index (they start at 1).
-func computeAssetIndexFromTxn(tx node.TxnWithStatus, l *data.Ledger) (aidx uint64) {
+func computeAssetIndexFromTxn(tx node.TxnWithStatus, l *data.Ledger) uint64 {
 	// Must have ledger
 	if l == nil {
 		return 0
@@ -412,6 +412,15 @@ func computeAssetIndexFromTxn(tx node.TxnWithStatus, l *data.Ledger) (aidx uint6
 	if tx.Txn.Txn.AssetConfigTxnFields.ConfigAsset != 0 {
 		return 0
 	}
+
+	aidx := uint64(tx.ApplyData.ConfigAsset)
+	if aidx > 0 {
+		return aidx
+	}
+	// If there is no ConfigAsset in the ApplyData, it must be a
+	// transaction before inner transactions were activated. Therefore
+	// the computeCreatableIndexInPayset function will work properly
+	// to deduce the aid. Proceed.
 
 	// Look up block where transaction was confirmed
 	blk, err := l.Block(tx.ConfirmedRound)
@@ -430,7 +439,7 @@ func computeAssetIndexFromTxn(tx node.TxnWithStatus, l *data.Ledger) (aidx uint6
 // computeAppIndexFromTxn returns the created app index given a confirmed
 // transaction whose confirmation block is available in the ledger. Note that
 // 0 is an invalid asset index (they start at 1).
-func computeAppIndexFromTxn(tx node.TxnWithStatus, l *data.Ledger) (aidx uint64) {
+func computeAppIndexFromTxn(tx node.TxnWithStatus, l *data.Ledger) uint64 {
 	// Must have ledger
 	if l == nil {
 		return 0
@@ -447,6 +456,15 @@ func computeAppIndexFromTxn(tx node.TxnWithStatus, l *data.Ledger) (aidx uint64)
 	if tx.Txn.Txn.ApplicationCallTxnFields.ApplicationID != 0 {
 		return 0
 	}
+
+	aidx := uint64(tx.ApplyData.ApplicationID)
+	if aidx > 0 {
+		return aidx
+	}
+	// If there is no ConfigAsset in the ApplyData, it must be a
+	// transaction before inner transactions were activated. Therefore
+	// the computeCreatableIndexInPayset function will work properly
+	// to deduce the aidx. Proceed.
 
 	// Look up block where transaction was confirmed
 	blk, err := l.Block(tx.ConfirmedRound)
--- a/daemon/algod/api/server/v1/handlers/handlers.go
+++ b/daemon/algod/api/server/v1/handlers/handlers.go
@@ -461,7 +461,7 @@ func computeAppIndexFromTxn(tx node.TxnWithStatus, l *data.Ledger) uint64 {
 	if aidx > 0 {
 		return aidx
 	}
-	// If there is no ConfigAsset in the ApplyData, it must be a
+	// If there is no ApplicationID in the ApplyData, it must be a
 	// transaction before inner transactions were activated. Therefore
 	// the computeCreatableIndexInPayset function will work properly
 	// to deduce the aidx. Proceed.
--- a/ledger/apptxn_test.go
+++ b/ledger/apptxn_test.go
@@ -624,3 +624,51 @@ func TestRekeyActionCloseAccount(t *testing.T) {
 	eval.txn(t, &useacct, "unauthorized")
 	l.endBlock(t, eval)
 }
+
+// TestInnerTxCount ensures that inner transactions increment the TxnCounter
+func TestInnerTxnCount(t *testing.T) {
+	partitiontest.PartitionTest(t)
+
+	genBalances, addrs, _ := newTestGenesis()
+	l := newTestLedger(t, genBalances)
+	defer l.Close()
+
+	create := txntest.Txn{
+		Type:   "appl",
+		Sender: addrs[0],
+		ApprovalProgram: main(`
+         tx_begin
+         int pay
+         tx_field TypeEnum
+         int 5000
+         tx_field Amount
+         txn Accounts 1
+         tx_field Receiver
+         tx_submit
+`),
+	}
+
+	fund := txntest.Txn{
+		Type:     "pay",
+		Sender:   addrs[0],
+		Receiver: basics.AppIndex(1).Address(),
+		Amount:   200000, // account min balance, plus fees
+	}
+
+	payout1 := txntest.Txn{
+		Type:          "appl",
+		Sender:        addrs[1],
+		ApplicationID: basics.AppIndex(1),
+		Accounts:      []basics.Address{addrs[1]}, // pay self
+	}
+
+	eval := l.nextBlock(t)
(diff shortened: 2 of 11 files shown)


ledger/cow.go
```diff
@@ -198,8 +202,9 @@ func (cb *roundCowState) trackCreatable(creatableIndex basics.CreatableIndex) {
 	cb.trackedCreatables[cb.groupIdx] = creatableIndex
 }
 
-func (cb *roundCowState) addTx(txn transactions.Transaction, txid transactions.Txid) {
+func (cb *roundCowState) addTx(txn transactions.Transaction, txid transactions.Txid, inners uint64) {
 	cb.mods.Txids[txid] = txn.LastValid
+	cb.txnCount += 1 + inners
```
winder, line 207: Currently, this code makes use of the uniqueness of a map and silently ignores duplicate transactions. With your change that is no longer the case.

It's very possible that we depend on that property.

I ran into this property recently during some Indexer testing. My synthetic transactions weren't random enough and if the block size was large enough I would create duplicate transactions. The Eval function (with validation disabled) would silently ignore them, which I only noticed because of unexpected TxnCounter values.

jannotti: I'm not sure I understand.  Do you mean for dupe detection?  Inner transactions are allowed to be duplicates.  There's no need for replay protection, because these transactions are not signed.  They ran because the parent ran (and that is protected). So they can't be replayed.

winder: Yes, this is related to dupe detection. The existing code correctly handles duplicate transactions and the new code does not.

jannotti: Meaning that the change does not detect "inner dupes", right?  That was the intent.  If you also think this changes behavior for top-level, that would be bad.

ledger/cow.go
```diff
@@ -180,7 +184,7 @@ func (cb *roundCowState) checkDup(firstValid, lastValid basics.Round, txid trans
 }
 
 func (cb *roundCowState) txnCounter() uint64 {
-	return cb.lookupParent.txnCounter() + uint64(len(cb.mods.Txids))
+	return cb.lookupParent.txnCounter() + cb.txnCount
```
algorandskiy, line 187:
```suggestion
	return cb.lookupParent.txnCounter() + uint64(len(cb.mods.Txids)) + uint64(len(cb.mods.innerTxns))
```
`cb.mods.innerTxns` can be a map of array of txn or txnids that is populated in a new method `addInnerTx` similar to `addTx`. The idea is to have actual data and not just a counter with non-obvious places to increment.

jannotti: Then we'd be maintaining a list, just so we can count it later.  It would have the same potential for error.  Wherever we are currently incrementing the count, we'd need to call `addInnerTx`.  But we don't need the list.

ledger/cow.go
jasonpaulos: I believe @winder is saying this function may be called multiple times with the same txn, and the behavior before this PR could properly handle this, since it has no side effects besides putting things into maps. If this is a property that we need to maintain (seems strange to me), then perhaps we could have a new map from txID to inner txn count instead of the monotonic `txnCount` variable.

jannotti: Yeah, we had a chat.  You are exactly right, and so is he. I intend to figure out if use that property, but I suspect not.  Will ran into it because he fed the validator synthetic blocks that had duplicates in them.  But we should never allow such things in a real block.

ledger/eval.go
```diff
@@ -905,7 +905,7 @@ func (eval *BlockEvaluator) transaction(txn transactions.SignedTxn, evalParams *
 	}
 
 	// Remember this txn
-	cow.addTx(txn.Txn, txid)
+	cow.addTx(txn.Txn, txid, uint64(len(applyData.EvalDelta.InnerTxns))) // Will need to be recursive when inners are
```
jasonpaulos, line 908: As a stronger reminder for us to change this if/when inner txns can create more inner txns, I'd recommend checking and returning an error if any inner txn has an inner txn.

jannotti: I'll actually probably write the recursive function, it'll just always bottom out fast.

ledger/cow.go
algorandskiy: the benefit is to make all calculations for nested cows work for free - one day teal can create app call txn that would spin-off another child cow for `StatefulEval` in `Perform`. Then propagating `innerTxns` in `commitParent` (or a counter for inner txns into parent's inner txns counter) would let `txnCounter()` func just work. No need to analyze inner txns from apply data in `eval.transaction` and so on.

jasonpaulos:
In regard to REST API changes, I am a proponent of creating a new field in `ApplyData` to explicitly store the created ID during an app/ASA creation txn. That way the ID is no longer implicit, so we don't have to calculate it again in our REST API (and in indexer).

tolikzinovyev:
Is there any usecase for the txn counter besides creating assets/apps? Perhaps it would be cleaner to rename it to "creatable counter" and allow for the outer transactions to create multiple assets: whenever an asset is created, assign the value from the counter and increment the counter.

jannotti:
> Is there any usecase for the txn counter besides creating assets/apps? Perhaps it would be cleaner to rename it to "creatable counter" and allow for the outer transactions to create multiple assets: whenever an asset is created, assign the value from the counter and increment the counter.

It's handy to know how many transactions we've executed, so we can talk TPS easily.  But I suppose not critical. If we had a creatable counter, we'd have to put that in the block header instead.  And we'd have to write the creatable id in the txn (which is what Jason is arguing for anyway), as it would be even harder to rederive it than what we do today.

jannotti:
> In regard to REST API changes, I am a proponent of creating a new field in `ApplyData` to explicitly store the created ID during an app/ASA creation txn. That way the ID is no longer implicit, so we don't have to calculate it again in our REST API (and in indexer).

I wanted to avoid this - it seemed like a waste of space.  But I'm coming around.  It does seem clean and obvious.  Of course, we have to calculate it again for old transactions, but if we simultaneously make this change and add inner-transactions, at least we'll never have to calculate it by counting up all the inner transactions in a block between the creation transaction and the end.

algorandskiy:
> at least we'll never have to calculate it by counting up all the inner transactions in a block between the creation transaction and the end.

Or it can be done as a ledger "static" function that called by REST and the indexer (since it already imports go-algorand)

winder:
> > In regard to REST API changes, I am a proponent of creating a new field in `ApplyData` to explicitly store the created ID during an app/ASA creation txn. That way the ID is no longer implicit, so we don't have to calculate it again in our REST API (and in indexer).
> 
> I wanted to avoid this - it seemed like a waste of space. But I'm coming around. It does seem clean and obvious. Of course, we have to calculate it again for old transactions, but if we simultaneously make this change and add inner-transactions, at least we'll never have to calculate it by counting up all the inner transactions in a block between the creation transaction and the end.

Everything about ApplyData is a waste of space by definition - the information is entirely derived by executing the transaction. To optimize on-chain storage there is no need to have it be part of `SignedTxnInBlock`, we could store it in a separate off-chain sqlite DB (i.e. exactly what we do with account data). It could even be optional.

jannotti:
I added the ids to ApplyData in my last commit.  Untested thus far, partly because it's annoying to test. We're maintaining the same interface but now without calculating the id in the rest calls.

Anyway, I devise some pokey unit tests to make sure it's working the way we'd expect tomorrow.  The test we really need is "In a block that is known to have inner transactions, does the REST API report the proper creatable ids anyway?

daemon/algod/api/server/v1/handlers/handlers.go
```diff
@@ -448,6 +457,15 @@ func computeAppIndexFromTxn(tx node.TxnWithStatus, l *data.Ledger) (aidx uint64)
 		return 0
 	}
 
+	aidx := uint64(tx.ApplyData.ApplicationID)
+	if aidx > 0 {
+		return aidx
+	}
+	// If there is no ConfigAsset in the ApplyData, it must be a
```
jasonpaulos, line 464:
```suggestion
	// If there is no ApplicationID in the ApplyData, it must be a
```

ledger/apptxn_test.go
```diff
@@ -229,11 +233,12 @@ submit:  tx_submit
 
 	eval := l.nextBlock(t)
 	eval.txns(t, &asa, &app)
-	l.endBlock(t, eval)
+	vb := l.endBlock(t, eval)
 
-	// Would be better to pull these out of block, or at least check them.
```
jasonpaulos, line 234: nice

ledger/apptxn_test.go
```diff
@@ -624,3 +631,112 @@ func TestRekeyActionCloseAccount(t *testing.T) {
...
+
+	eval := l.nextBlock(t)
+	eval.txns(t, &create, &fund)
+	vb := l.endBlock(t, eval)
+	require.Equal(t, 2, int(vb.blk.TxnCounter))
+
+	eval = l.nextBlock(t)
+	eval.txns(t, &payout1)
+	vb = l.endBlock(t, eval)
+	require.Equal(t, 4, int(vb.blk.TxnCounter))
```
jasonpaulos, line 741: Perhaps add an asset create txn after this to make sure it's assigned ID is correct?

jannotti: Yeah, good idea.

ledger/cow.go
jannotti: I think we're on the same page.  I did modify commitToParent(), so it does:
`cb.commitParent.txnCount += cb.txnCount`
That allows children to work properly, I think.

ledger/eval.go
jannotti: I think, given my last comment to Pavel, that I'm probably wrong about how this is will work with trees of inner transactions.  It won't need to be recursive, exactly, since the inners will already have their children committed to the cb that I'll inevitably need to evaluate them - I'll get their count from there, instead of assuming each = 1 as I do here.  So for now, I will change the comment to remind myself to look at it, but I will not change it to walk down a potential tree of inners.

ledger/cow.go
jannotti: I get your point now.  You want it be true, the whole time, that cb.txnCounter() is correct.  I was only incrementing the count after all inner transactions for a given top-level txn were done.  That's wrong, because I won't know the count after one is done, but I'm working on the next.  I'm changing it to increment the count after each inner transaction.

ledger/eval.go
jannotti: And now that I understand what Pavel was saying, I'll actually be removing the passing of the extra length.

jasonpaulos (approved):

jannotti:
Superceded by inner-asset-txns #2871

PR #2863, opened 2021-09-09
fix issue with checking RUN_EXPECT when listing packages

cce:
## Summary

@algoidan noticed we were not running some non-expect e2e-go tests and we spotted this issue in the test script when checking the value of RUN_EXPECT.

## Test Plan

Check to ensure non-expect e2e-go tests are run as expected.


winder (approved):
Nice catch!

algorandskiy (approved):

algorandskiy (approved):

algojack (approved):

PR #2749, opened 2021-08-16
Checking CircleCI tests run once and only once

algojack:
## Summary
Our partitioned tests might be skipping tests or running them twice. Here we want to make sure they don't do that.

## Test Plan

Testing in CircleCI on this branch before merging to master

PR #2864, opened 2021-09-09
use matrix for CircleCI config

cce:
## Summary

Based on the matrix job @algojack added in #2749, this turns the other per-platform jobs into matrix jobs too. The most significant change in this PR is switching from building in the home directory (/home/circleci on Linux and /Users/distiller on Mac) to instead build out of /opt/cibuild to make it easier for the matrix jobs to share configuration.

## Test Plan

It would be good to check that all the same tests are being run as before, perhaps by going through the build output manually or the artifacts (resulting JSON/XML files).


algobarb (approved):
Read through it and reviewed the CircleCI results. It seems good to me and it introduced me to yaml syntax for using `&ref_name` and `<<: *ref_name`. The extended use of matrices for the executors does reduce a lot of duplicate code.

I wonder how easy it will be for someone new to jump into and make changes to the config file later on and whether the YAML is enough or if we need any documentation on what we expect our CircleCI to currently do.

cce:
Yes, I also didn't know that YAML syntax either until I saw it recently in [some CircleCI docs](https://circleci.com/docs/2.0/workflows/) and then read their YAML guide on [Anchors and aliases](https://circleci.com/docs/2.0/writing-yaml/#anchors-and-aliases).

algojack:
lgtm, here's a good simple article on the << & and * https://medium.com/@kinghuang/docker-compose-anchors-aliases-extensions-a1e4105d70bd

algojack (approved):

algojack:
@cce is the GIMME change in the second file on purpose? what does it do exactly?

cce:
The original build.sh was downloading via curl the "gimme" script to ~/gimme.sh, so I changed it to check for a GIMME_INSTALL_DIR and use that, falling back to $HOME as before if it is not set. That made it possible for the matrix build not to be so dependent on the value of $HOME.

PR #2867, opened 2021-09-10
fix layer bug.

algoidan:

### DIFF
--- a/crypto/merklearray/layer.go
+++ b/crypto/merklearray/layer.go
@@ -36,19 +36,20 @@ type pair struct {
 }
 
 func (p *pair) ToBeHashed() (protocol.HashID, []byte) {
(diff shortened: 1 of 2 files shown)


crypto/merklearray/layer.go
```diff
@@ -36,18 +36,19 @@ type pair struct {
 }
 
 func (p *pair) ToBeHashed() (protocol.HashID, []byte) {
-	var buf [2 * crypto.DigestSize]byte
-	copy(buf[:crypto.DigestSize], p.l[:])
-	copy(buf[crypto.DigestSize:], p.r[:])
+	buf := make([]byte, 0, len(p.l)+len(p.r))
+	buf = append(buf, p.l...)
+	buf = append(buf, p.r...)
```
tsachiherman, line 41: the `copy` that was used above is marginally faster compared to `append`

PR #2869, opened 2021-09-10
txnsync: remove known txid from bloom filter

tsachiherman:
## Summary

On non-relays, exclude txid that were sent from the relay from the bloom filter sent to the relay.
This is expected to reduce the size of the bloom filters by 25%.

## Test Plan

Unit tests updated.

### DIFF
--- a/txnsync/bloomFilter.go
+++ b/txnsync/bloomFilter.go
@@ -137,13 +137,13 @@ func filterFactoryXor32(numEntries int, s *syncState) (filter bloom.GenericFilte
 
 var filterFactory func(int, *syncState) (filter bloom.GenericFilter, filterType bloomFilterType) = filterFactoryXor32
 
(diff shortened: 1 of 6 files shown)


txnsync/bloomFilter.go
```diff
@@ -89,7 +89,6 @@ func decodeBloomFilter(enc encodedBloomFilter) (outFilter *testableBloomFilter,
 	if err != nil {
 		return nil, err
 	}
-	outFilter.encodingParams = enc.EncodingParams
```
tsachiherman, line 92: this is redundant, as it's already being done above.

brianolson (approved):
LGTM. It has a bunch of commonality with the lookback for outbound filters, checking txid against a set before adding them to a filter. I'll rework my code to fit after this goes in.

PR #2871, opened 2021-09-11
Inner asset txns

jannotti:
acfg and afrz inner transactions

This also sorts out inner txcounting, so that an innner transaction knows the current txncount as it executes, so it can get the right id if it creates.

Need some more tests, but so far we have some teal unit tests, and ledger level integration tests.


algorandskiy (approved):
I like this one better. Probably need to add some e2e tests

ledger/apptxn_test.go
```diff
@@ -624,3 +631,264 @@ func TestRekeyActionCloseAccount(t *testing.T) {
...
+
+	eval = l.nextBlock(t)
+	// Can't create an asset if you have exactly 200,000 and need to pay fee
+	eval.txn(t, &create_asa, "balance 199000 below min 200000")
+	// fund it some more and try again
+	eval.txns(t, fund.Noted("more!"), &create_asa)
+	vb := l.endBlock(t, eval)
+
+	asaIndex := vb.blk.Payset[1].EvalDelta.InnerTxns[0].ConfigAsset
+	require.Equal(t, basics.AssetIndex(5), asaIndex)
```
jasonpaulos, line 846: Could you also pull out the params of the newly created asset and check its fields match the values set by the AVM?

Also, I suppose it's not possible to actually transfer the newly created asset or use the `asset_param_get` op on it since it can't be in the foreign assets array? I wonder if we should make an exception in this case and allow the asset to be "available" anyway.

data/transactions/logic/eval.go
```diff
@@ -3522,10 +3548,52 @@ func (cx *EvalContext) stackIntoTxnField(sv stackValue, fs txnFieldSpec, txn *tr
...
+	case ConfigAssetUnitName:
+		txn.AssetParams.UnitName, err = sv.string(cx.Proto.MaxAssetUnitNameBytes)
+	case ConfigAssetName:
+		txn.AssetParams.AssetName, err = sv.string(cx.Proto.MaxAssetNameBytes)
+	case ConfigAssetURL:
+		txn.AssetParams.URL, err = sv.string(cx.Proto.MaxAssetURLBytes)
+	case ConfigAssetMetadataHash:
+		if sv.Bytes == nil {
+			err = fmt.Errorf("ConfigAssetMetadataHash must be bytes")
+		} else if len(sv.Bytes) > 32 {
```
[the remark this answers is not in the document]
jannotti: I considered that, but I think I found that it's ok to submit, say, 6 bytes of metdatahash in a txn, and it gets copied in this way.  This is in libgoal:
```
	if len(metadataHash) > len(tx.AssetParams.MetadataHash) {
		return tx, fmt.Errorf("asset metadata hash %x too long (max %d bytes)", metadataHash, len(tx.AssetParams.MetadataHash))
	}
	copy(tx.AssetParams.MetadataHash[:], metadataHash)
```
So I wanted to maintain this interface.  Do SDKs enforce 32 bytes?

data/transactions/logic/eval.go
```diff
@@ -3522,10 +3548,52 @@ func (cx *EvalContext) stackIntoTxnField(sv stackValue, fs txnFieldSpec, txn *tr
 		txn.AssetReceiver, err = cx.availableAccount(sv)
 	case AssetCloseTo:
 		txn.AssetCloseTo, err = cx.availableAccount(sv)
-
-	// acfg likely next
-
-	// afrz seems easy but not high demand
+	// AssetConfig
+	case ConfigAsset:
+		txn.ConfigAsset, err = cx.availableAsset(sv)
```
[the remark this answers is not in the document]
jannotti: I considered this, too.  It seems ok to allow.  But I thought it would be better to teach developers that it's not necessary to do so, since it starts as 0. That saves 4 bytes or so of AVM code. I'm open to changing that opinion.

data/transactions/logic/evalAppTxn_test.go
```diff
@@ -352,3 +360,64 @@ func TestNumInner(t *testing.T) {
...
+  int 1           ; tx_field FreezeAssetFrozen
+  txn Accounts 1  ; tx_field FreezeAssetAccount
+  tx_submit
+  int 1
+`
+	testApp(t, freeze, ep, "invalid Asset reference")
+	ep.Txn.Txn.ForeignAssets = []basics.AssetIndex{basics.AssetIndex(889)}
+	testApp(t, freeze, ep, "does not hold Asset")
+	ledger.NewHolding(ep.Txn.Txn.Receiver, 889, 55, false)
+	testApp(t, freeze, ep)
```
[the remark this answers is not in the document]
jannotti: Will do.

ledger/apptxn_test.go
jannotti: WiIll do, on testing.

I was strongly considering adding a special case (as we did for the app address) that would allow the use of an asset that had been created in the same transaction (or txgroup?). But as for as I can tell, it's pretty much impossible to use, because no recipient could possibly be opted-in yet.

data/transactions/logic/evalAppTxn_test.go
```diff
@@ -352,3 +361,73 @@ func TestNumInner(t *testing.T) {
...
+
+func TestAssetFreeze(t *testing.T) {
+	create := `
+  tx_begin
+  int acfg                         ; tx_field TypeEnum
+  int 1000000                      ; tx_field ConfigAssetTotal
+  int 3                            ; tx_field ConfigAssetDecimals
+  byte "oz"                        ; tx_field ConfigAssetUnitName
+  byte "Gold"                      ; tx_field ConfigAssetName
+  byte "https:://gold.rush/"       ; tx_field ConfigAssetURL
```
barnjamin, line 399: `::` here and above

data/transactions/logic/eval.go
```diff
@@ -114,6 +114,30 @@ func (sv *stackValue) uint() (uint64, error) {
...
+	if err != nil {
+		return false, err
+	}
+	if u64 == 1 {
+		return true, nil
+	}
+	if u64 == 0 {
+		return false, nil
+	}
+	return false, fmt.Errorf("boolean is neither 1 nor 0: %d", u64)
```
tsachiherman, line 128: nit: switch statement here would be more elegant.

data/transactions/logic/evalAppTxn_test.go
```diff
@@ -21,6 +21,7 @@ import (
 	"testing"
 
 	"github.com/algorand/go-algorand/data/basics"
+	"github.com/stretchr/testify/require"
```
tsachiherman, line 24: nit: break into a separate group.

data/transactions/transaction.go
```diff
@@ -113,6 +113,11 @@ type ApplyData struct {
 	ReceiverRewards basics.MicroAlgos `codec:"rr"`
 	CloseRewards    basics.MicroAlgos `codec:"rc"`
 	EvalDelta       EvalDelta         `codec:"dt"`
+
+	// If asa or app is being created, the id used. Else 0.
+	// Names chosen to match naming the corresponding txn.
```
tsachiherman, line 118: add a comment stating that these fields available only on TEAL 5+.

ledger/eval_test.go
```diff
@@ -1441,6 +1441,21 @@ func (ledger *Ledger) asa(t testing.TB, addr basics.Address, asset basics.AssetI
...
 }
 
+// asaParams gets the asset params for a given asa index
+func (ledger *Ledger) asaParams(t testing.TB, asset basics.AssetIndex) (basics.AssetParams, error) {
+	creator, ok, err := ledger.GetCreator(basics.CreatableIndex(asset), basics.AssetCreatable)
+	if err != nil {
+		return basics.AssetParams{}, err
+	}
+	if !ok {
+		return basics.AssetParams{}, fmt.Errorf("no asset")
```
tsachiherman, line 1451: nit: `fmt.Errorf("no asset")` -> `errors.New("no asset")`

ledger/eval.go
```diff
@@ -904,6 +904,13 @@ func (eval *BlockEvaluator) transaction(txn transactions.SignedTxn, evalParams *
 		}
 	}
 
+	// We are not allowing InnerTxns to have InnerTxns yet.  Error if that happens.
+	for _, itx := range applyData.EvalDelta.InnerTxns {
+		if len(itx.ApplyData.EvalDelta.InnerTxns) > 0 {
+			return fmt.Errorf("inner transaction has inner transactions %v", itx)
+		}
+	}
```
tsachiherman, line 912: I'm not sure if this is really needed, since we don't have any way to trigger this; but I don't see how having this check would be a bad thing either.

data/transactions/logic/eval.go
```diff
@@ -3522,10 +3548,52 @@ func (cx *EvalContext) stackIntoTxnField(sv stackValue, fs txnFieldSpec, txn *tr
...
+	case ConfigAsset:
+		txn.ConfigAsset, err = cx.availableAsset(sv)
+	case ConfigAssetTotal:
+		txn.AssetParams.Total, err = sv.uint()
+	case ConfigAssetDecimals:
+		var decimals uint64
+		decimals, err = sv.uint()
+		if err == nil {
+			if decimals > uint64(cx.Proto.MaxAssetDecimals) {
+				err = fmt.Errorf("too many decimals")
```
tsachiherman, line 3561: nit: `fmt.Errorf("too many decimals")` -> `errors.New("too many decimals")`

tsachiherman (commented):
Overall looks great. I've added few minor code styling requests.

data/transactions/logic/eval.go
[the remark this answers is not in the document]
jasonpaulos: Some of the SDKs (Javascript, Python, and maybe Java) enforce that the metadata hash must be 32 byets -- in my opinion this is the right thing to do and Goal and the Go SDK are in the wrong. The reason being it's very confusing if you set the metadata hash to `0x1234`, but when you check it later you actually get `0x12340000...`. This would be especially concerning in a smart contract.

Unless we actually want to make this field dynamically sized, I strongly suggest sticking with 32-byte inputs.

data/transactions/logic/eval.go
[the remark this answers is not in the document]
jasonpaulos: Yeah that seems reasonable. I don't see a problem with that logic at all, but we may get developers asking why 0 isn't a valid input here.

ledger/apptxn_test.go
jasonpaulos: Well theoretically the app could have another account rekeyed to it, so it could potentially opt that account into the new asset and perform a transfer.

The bigger issue I see is that there's currently no way to get the ID of the created asset. If we don't support this, there's no reason to support the special case of adding newly created assets to the "available" list.

data/transactions/logic/eval.go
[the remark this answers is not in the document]
jannotti: That is a good point (that it comes back differently).  Would you say it's a bug in our transaction processing that we allow short metadatahashes? (if we do, I haven't 100% confirmed) it.  I think I would say it is, based on the same reasoning.

I'm willing to enforce 32 bytes here, I just hate being different from "native" processing.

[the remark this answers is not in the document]
jannotti: Actually, since this is libgoal, we might not allow it at the txn processing level, but libgoal is being a weirdo.  It is perhaps sending a 32 byte msgpack field, padded with 0s.  In *that* case, I'm 100% on board.

ledger/apptxn_test.go
jannotti: I'd certainly like it too.  Let me see if I can think of a decent interface that doesn't paint us into a corner.  Ideas welcome.

data/transactions/logic/eval.go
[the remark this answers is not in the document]
jasonpaulos: > It is perhaps sending a 32 byte msgpack field, padded with 0s

Yes, that's exactly what's happening. The issue is not with txn processing, but rather with txn creation.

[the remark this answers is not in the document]
jannotti: Sold!

ledger/apptxn_test.go
jannotti: Perhaps
```
itxna Logs X
itxn ConfigAsset
```
(and attendant `itxn NumLogs`, `itxnas`)

These would all return results from previous `tx_submit` execution.  And perhaps they become `itxn_begin` etc.

jasonpaulos:
Also, I did some local testing and pending transaction REST endpoint does not return the created asset ID when an inner txn creates an asset. See the implementation for `convertTxn` in `daemon/algod/api/server/v2/utils.go`.

jannotti:
> Also, I did some local testing and pending transaction REST endpoint does not return the created asset ID when an inner txn creates an asset. See the implementation for `convertTxn` in `daemon/algod/api/server/v2/utils.go`.

Oh my.  Thank you.

data/transactions/logic/eval.go
```diff
@@ -3422,8 +3447,8 @@ func opTxBegin(cx *EvalContext) {
 	fee := cx.Proto.MinTxnFee
 	if cx.FeeCredit != nil {
 		// Use credit to shrink the fee, but don't change FeeCredit
-		// here, because they might never tx_submit, or they might
-		// change the fee.  Do it in tx_submit.
+		// here, because they might never itxn_submit, or they might
+		// change the fee.  Do it in itxn_submit.
 		fee = basics.SubSaturate(fee, *cx.FeeCredit)
 	}
 	cx.subtxn.Txn.Header = transactions.Header{
```
mxmauro, line 3454: Hi, a quick question. Should we assume the inner tx belongs to the same group/lease than the parent despite not present and should be added when calculating the ID of the inner tx?

jannotti: The inner txn has no id.

mxmauro: Ahh ok, thanks for the info because, from the indexers and from AlgoExplorer point of view, we didn't know if these txs could be identified individually.

jannotti: Indexer may well put a fake id in its internal table in order to place itxns in the same table with top-level.  But you would obtain the info pertaining to an itxn only by asking for the top-level, and receive a tree back.

mxmauro: Yes, that's a good idea and will see with @winder the approach to show the same fake id.

agodnic: Using fake IDs may allow Indexer/AlgoExplorer users to search for an inner transaction easily using a regular TXID.

If we do that, we should agree on a definition for such fake txids. Maybe it could be something like `digest(tx, parent, offset)`

jannotti: I don't love the fake txid thing. I hope we'll keep it an implementation detail and avoid confusing users.

mxmauro: And what about adding an "inner transaction index" field to the subtxn. Node must check this field is not defined or zero when a TX is submitted. This plus the first/last round you already added, should do the work.

jannotti:
Closed in favor of #2883

PR #2872, opened 2021-09-11
Add ledger.GetBlockAddresses()

tolikzinovyev:
## Summary

Indexer needs to know all accounts referenced in a block to be able to read all of them in one batch. This PR creates a function that returns all such account addresses.

### DIFF
--- a/ledger/eval.go
+++ b/ledger/eval.go
@@ -1294,6 +1294,25 @@ type loadedTransactionGroup struct {
 	err error
 }
 
+// Return the maximum number of addresses referenced in any given transaction.
+func maxAddressesInTxn(proto *config.ConsensusParams) int {
+	return 7 + proto.MaxAppTxnAccounts
+}
+
+// Return a list of addresses referenced in `txn`. Addresses might repeat.
+func getTxnAddresses(txn *transactions.Transaction, proto *config.ConsensusParams) []basics.Address {
+	res := make([]basics.Address, 0, maxAddressesInTxn(proto))
+
+	res = append(
+		res, txn.Sender, txn.Receiver, txn.CloseRemainderTo, txn.AssetSender,
+		txn.AssetReceiver, txn.AssetCloseTo, txn.FreezeAccount)
+	for _, addr := range txn.ApplicationCallTxnFields.Accounts {
+		res = append(res, addr)
+	}
+
+	return res
+}
+
 // loadAccounts loads the account data for the provided transaction group list. It also loads the feeSink account and add it to the first returned transaction group.
 // The order of the transaction groups returned by the channel is identical to the one in the input array.
 func loadAccounts(ctx context.Context, l ledgerForEvaluator, rnd basics.Round, groups [][]transactions.SignedTxnWithAD, feeSinkAddr basics.Address, consensusParams config.ConsensusParams) chan loadedTransactionGroup {
@@ -1320,8 +1339,7 @@ func loadAccounts(ctx context.Context, l ledgerForEvaluator, rnd basics.Round, g
 		defer close(outChan)
 
 		accountTasks := make(map[basics.Address]*addrTask)
-		maxAddressesPerTransaction := 7 + consensusParams.MaxAppTxnAccounts
-		addressesCh := make(chan *addrTask, len(groups)*consensusParams.MaxTxGroupSize*maxAddressesPerTransaction)
+		addressesCh := make(chan *addrTask, len(groups)*consensusParams.MaxTxGroupSize*maxAddressesInTxn(&consensusParams))
 		// totalBalances counts the total number of balances over all the transaction groups
 		totalBalances := 0
 
@@ -1362,15 +1380,9 @@ func loadAccounts(ctx context.Context, l ledgerForEvaluator, rnd basics.Round, g
 			task := &groupTask{}
 			groupsReady[i] = task
 			for _, stxn := range group {
-				initAccount(stxn.Txn.Sender, task)
-				initAccount(stxn.Txn.Receiver, task)
-				initAccount(stxn.Txn.CloseRemainderTo, task)
-				initAccount(stxn.Txn.AssetSender, task)
-				initAccount(stxn.Txn.AssetReceiver, task)
-				initAccount(stxn.Txn.AssetCloseTo, task)
-				initAccount(stxn.Txn.FreezeAccount, task)
-				for _, xa := range stxn.Txn.Accounts {
-					initAccount(xa, task)
+				addresses := getTxnAddresses(&stxn.Txn, &consensusParams)
+				for _, address := range addresses {
+					initAccount(address, task)
 				}
 			}
 		}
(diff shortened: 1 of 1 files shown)


tsachiherman (changes requested):
The change looks good - but they don't perform that great. I tested your changes using
```sh
go test -v . --run XX --bench BenchmarkBlockEvaluatorRAMNoCrypto -count=10 | grep BenchmarkBlockEvaluatorRAMNoCrypto
```

and received the following performance metrics on master:
```sh
BenchmarkBlockEvaluatorRAMNoCrypto-8   	       2	 945332700 ns/op	     18907 ns/eval_validate_tx
BenchmarkBlockEvaluatorRAMNoCrypto-8   	       2	 887004810 ns/op	     17741 ns/eval_validate_tx
BenchmarkBlockEvaluatorRAMNoCrypto-8   	       2	 898731697 ns/op	     17976 ns/eval_validate_tx
BenchmarkBlockEvaluatorRAMNoCrypto-8   	       2	 905146646 ns/op	     18104 ns/eval_validate_tx
BenchmarkBlockEvaluatorRAMNoCrypto-8   	       2	 895749462 ns/op	     17916 ns/eval_validate_tx
BenchmarkBlockEvaluatorRAMNoCrypto-8   	       2	 876699177 ns/op	     17535 ns/eval_validate_tx
BenchmarkBlockEvaluatorRAMNoCrypto-8   	       2	 889700168 ns/op	     17795 ns/eval_validate_tx
BenchmarkBlockEvaluatorRAMNoCrypto-8   	       2	 888446000 ns/op	     17770 ns/eval_validate_tx
BenchmarkBlockEvaluatorRAMNoCrypto-8   	       2	 888201852 ns/op	     17765 ns/eval_validate_tx
BenchmarkBlockEvaluatorRAMNoCrypto-8   	       2	 916630812 ns/op	     18333 ns/eval_validate_tx
```

compared to the ones from your branch:
```sh
BenchmarkBlockEvaluatorRAMNoCrypto-8   	       1	1016715840 ns/op	     20335 ns/eval_validate_tx
BenchmarkBlockEvaluatorRAMNoCrypto-8   	       1	1026562100 ns/op	     20533 ns/eval_validate_tx
BenchmarkBlockEvaluatorRAMNoCrypto-8   	       2	1005568624 ns/op	     20112 ns/eval_validate_tx
BenchmarkBlockEvaluatorRAMNoCrypto-8   	       1	1008136765 ns/op	     20164 ns/eval_validate_tx
BenchmarkBlockEvaluatorRAMNoCrypto-8   	       1	1016944071 ns/op	     20340 ns/eval_validate_tx
BenchmarkBlockEvaluatorRAMNoCrypto-8   	       2	1016153946 ns/op	     20324 ns/eval_validate_tx
BenchmarkBlockEvaluatorRAMNoCrypto-8   	       2	 974119522 ns/op	     19483 ns/eval_validate_tx
BenchmarkBlockEvaluatorRAMNoCrypto-8   	       2	 957210924 ns/op	     19145 ns/eval_validate_tx
BenchmarkBlockEvaluatorRAMNoCrypto-8   	       2	 879668276 ns/op	     17594 ns/eval_validate_tx
BenchmarkBlockEvaluatorRAMNoCrypto-8   	       2	 913933502 ns/op	     18279 ns/eval_validate_tx
```

In the vast of the cases, `master` perform better compared to your branch. I would suggest that you'll keep the existing implementation of `loadAccounts` as is, and just have your own changes ( you might want to add some comments that it's a code duplication that need to be converged ).

For the purposes of the indexer, we can optimize that code on another task - but there is no reason to introduce a regression to the master.

winder:
Would it make sense to add this as a function to SignedTxnInBlock instead of as part of the evaluator?

tsachiherman:
> Would it make sense to add this as a function to SignedTxnInBlock instead of as part of the evaluator?

I would be ok if we were to add the method to the `transactions` package. But why `SignedTxnInBlock` and not `Transaction` ?

winder:
> > Would it make sense to add this as a function to SignedTxnInBlock instead of as part of the evaluator?
> 
> I would be ok if we were to add the method to the `transactions` package. But why `SignedTxnInBlock` and not `Transaction` ?

`Transaction` would be better. I was thinking we needed the `ApplyData` in order to figure out who participates in the inner transactions, but `txn.ApplicationCallTxnFields.Accounts` is available for that.

Do we always need all of these accounts available?  I suppose the TEAL evaluation needs to read accounts regardless of whether they are modified...

If all of these accounts are added to the `txn participation` table, this is going to change the result of a txns by account lookup. We may want to add this to the release notes, and possibly add a new address role for it.

tolikzinovyev:
I can't reproduce the difference in performance (perhaps because I'm using the newest version of go). I changed the code slightly. Could you run the benchmark again?

tsachiherman:
> I can't reproduce the difference in performance (perhaps because I'm using the newest version of go). I changed the code slightly. Could you run the benchmark again?

I'm using 
```ssh
go version go1.16.7 darwin/amd64
```

and running the same benchmark again:
```ssh
BenchmarkBlockEvaluatorRAMNoCrypto
BenchmarkBlockEvaluatorRAMNoCrypto-8   	       2	 937884934 ns/op	     18758 ns/eval_validate_tx
BenchmarkBlockEvaluatorRAMNoCrypto-8   	       2	1104762888 ns/op	     22097 ns/eval_validate_tx
BenchmarkBlockEvaluatorRAMNoCrypto-8   	       1	1051855114 ns/op	     21039 ns/eval_validate_tx
BenchmarkBlockEvaluatorRAMNoCrypto-8   	       2	 945192180 ns/op	     18905 ns/eval_validate_tx
BenchmarkBlockEvaluatorRAMNoCrypto-8   	       2	 964409754 ns/op	     19289 ns/eval_validate_tx
BenchmarkBlockEvaluatorRAMNoCrypto-8   	       2	 896160418 ns/op	     17924 ns/eval_validate_tx
BenchmarkBlockEvaluatorRAMNoCrypto-8   	       2	 945274962 ns/op	     18906 ns/eval_validate_tx
BenchmarkBlockEvaluatorRAMNoCrypto-8   	       2	 971673884 ns/op	     19435 ns/eval_validate_tx
BenchmarkBlockEvaluatorRAMNoCrypto-8   	       1	1020561048 ns/op	     20412 ns/eval_validate_tx
BenchmarkBlockEvaluatorRAMNoCrypto-8   	       2	 971733400 ns/op	     19436 ns/eval_validate_tx
```
is slightly better, but not there yet.

tolikzinovyev:
It might be something to do with the new optimization in go 1.17. https://golang.org/doc/go1.17#compiler

How significant is this benchmark anyway? The cost of eval should be dominated by crypto and disk.

tsachiherman:
There is no reason to have any performance loss for adding new, non-interfering functionality. I'm sure that the performance loss would have been much more noticeable if we had a benchmark for loadAccounts - but we don't.

Given that refactoring this code isn't really what you're after, just rollback the changes in `loadAccounts` and it will be good to go. Once we move to go 1.17, I'll be more than happy to re-test that.

ledger/eval.go
```diff
@@ -1294,6 +1294,21 @@ type loadedTransactionGroup struct {
...
+}
+
+// Write the list of addresses referenced in `txn` to `out`. Addresses might repeat.
+func getTxnAddresses(txn *transactions.Transaction, out *[]basics.Address) {
+	*out = (*out)[:0]
+
+	*out = append(
+		*out, txn.Sender, txn.Receiver, txn.CloseRemainderTo, txn.AssetSender,
+		txn.AssetReceiver, txn.AssetCloseTo, txn.FreezeAccount)
+	*out = append(txn.ApplicationCallTxnFields.Accounts)
```
tsachiherman, line 1309: I think that you meant 
```golang
*out = append(*out, txn.ApplicationCallTxnFields.Accounts...)
```

tolikzinovyev: thanks!

tsachiherman (approved):
looks good to me.

Issue #2870, opened 2021-09-10
TEAL assembly details


barnjamin:
## Problem

When debugging a TEAL program, given an error message from a Application Call, there is no straightforward way to map the line of TEAL back to the PC returned in the error message. 

## Solution

We could append the PC to each line during disassembly by default or given an appropriate flag.

Alternatively, during assembly, we could output a JSON file with these details and any other details about the program. Other details might include program cost or some details about Template Variables or hash of the contract.

## Dependencies

Probably the assembler https://github.com/algorand/go-algorand/blob/b8b7281d9dc4ef4ad855001f99ecc6b752b1fa36/data/transactions/logic/assembler.go#L1499

barnjamin:
This may be a reasonable place to put the [ABI method descriptions](https://github.com/algorandfoundation/ARCs/blob/6a9f91aed8068bbace872483a64cbf22d0e1c975/ARCs/arc-0004.md#method-description) as well

barnjamin:
It looks like the assembler already tracks the pc=>source line here https://github.com/algorand/go-algorand/blob/master/data/transactions/logic/assembler.go#L246

PR #2876, opened 2021-09-12
Adding logic for source line to pc mapping and template variable details

barnjamin:
## Summary

I'd like to be able to easily map source line of TEAL back to pc in compiled program, detailed here: https://github.com/algorand/go-algorand/issues/2870

Really just wanted to test out the idea, many changes likely necessary. If actually implemented, we'd probably want to flag off the assembly details and choose a better name for the output.

`goal clerk compile` now outputs a json object with details about the assembled program including a map of source line number to pc and template variable details.

In the `line_map`  array, the index is the line number and the value is the pc.  Any 0 past the first one is considered empty space in the teal source program and should not be used. This output could be used by an editor to display pc in line number gutter or just manually referenced when debugging. 

In the `template_labels` object, the keys are the template variable names and the values are a set of details about their location and type. This allows template contracts to be compiled directly with no "dummy variables".  It also makes the insertion or removal of variables in the compiled bytecode easier. 

Care will need to be taken when working with the bytecode to ensure the appropriate length in bytes is written after the opcode byte or the program counter in the evaluator will get confused.  Also if populated, the bytes template variables will throw off the `position` field by whatever length of bytes its populated with.  Accounting for these issues is functionality that'd be nice for the SDKs to provide.

### DIFF
--- a/cmd/goal/clerk.go
+++ b/cmd/goal/clerk.go
@@ -1005,6 +1005,10 @@ func disassembleFile(fname, outname string) {
 	}
 }
 
+// AssemblyDetails contains details from the source to assembly process
+// Right now it is just the map of line number to program counter or byte position in
+// the assembled program but may contain other details like ABI spec or Template Variable
+// relative positions in the future
 type AssemblyDetails struct {
 	LineMap []int `json:"line_map"`
 }
--- a/cmd/goal/clerk.go
+++ b/cmd/goal/clerk.go
@@ -921,7 +921,7 @@ func mustReadFile(fname string) []byte {
 	return contents
 }
 
-func assembleFileDetails(fname string) (program []byte, lineMap []int) {
+func assembleFileDetails(fname string) (program []byte, deets AssemblyDetails) {
 	text, err := readFile(fname)
 	if err != nil {
 		reportErrorf("%s: %s", fname, err)
@@ -942,7 +942,7 @@ func assembleFileDetails(fname string) (program []byte, lineMap []int) {
 		}
 	}
 
-	return ops.Program, ops.GetLineToOffset()
+	return ops.Program, AssemblyDetails{LineMap: ops.GetLineToOffset(), TemplateLabels: ops.TemplateLabels}
 }
 
 func assembleFile(fname string) (program []byte) {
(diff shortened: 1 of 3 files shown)


barnjamin:
https://github.com/barnjamin/rareaf/blob/price-tokens/src/contracts/python/tcv.py wrote some example python for the template variable insertion/removal

barnjamin:
closing for now, too much going on here. might try to resubmit in 2 diff prs or wait to see how ABI stuff is added and try to base off that.

cmd/goal/clerk.go
```diff
@@ -1023,11 +1060,22 @@ var compileCmd = &cobra.Command{
...
 				outblob = protocol.Encode(&ls)
 			}
+
 			if !noProgramOutput {
 				err := writeFile(outname, outblob, 0666)
 				if err != nil {
 					reportErrorf("%s: %s", outname, err)
 				}
+
+				mapname := outname + ".deets.json"
```
bbroder-algo, line 1070: :face_with_rolling_eyes:

data/transactions/logic/assembler.go
```diff
@@ -461,28 +480,74 @@ func asmPushInt(ops *OpStream, spec *OpSpec, args []string) error {
...
-	if err != nil {
-		return ops.error(err)
+
+	var (
+		val  uint64
+		err  error
+		tmpl string
+	)
+
+	if len(args[0]) > 5 && args[0][:5] == TmplPrefix {
```
bbroder-algo, line 490: why 5.  do you mean len(TmplPrefix)

data/transactions/logic/assembler.go
```diff
@@ -687,26 +752,52 @@ func assembleIntCBlock(ops *OpStream, spec *OpSpec, args []string) error {
...
+				// Would be nice to keep going, as in
+				// intcblock, but parseBinaryArgs would have
+				// to return a useful consumed value even in
+				// the face of errors.  Hard.
+				ops.error(err)
+				return nil
+			}
 		}
+
+		tvars = append(tvars, tvar)
```
bbroder-algo, line 779: do you really want to append this stale val to bvals if there's a tvar?

bbroder-algo:
where test code

PR #2877, opened 2021-09-13
Consensus V30 updates

algochoi:
## Summary

This commit makes consensus parameter changes for V30.

v30 introduces AVM 1.0 and TEAL 5, increases the app opt in limit to 50, and allows costs to be pooled in grouped stateful transactions.


config/consensus.go
```diff
@@ -1002,6 +1002,26 @@ func initConsensusProtocols() {
...
+	v30.EnableAppCostPooling = true
+	v30.MaxInnerTransactions = 16
+
+	// Allow 50 app opt ins
+	v30.MaxAppsOptedIn = 50
+
+	Consensus[protocol.ConsensusV30] = v30
+
+	// v29 can be upgraded to v30, with an update delay of 7 days ( see calculation above )
+	v29.ApprovedUpgrades[protocol.ConsensusV30] = 140000
```
algochoi, line 1023: Is this delay correct?

onetechnical: Yep! 140,000 is the default.

onetechnical (approved):
Didn't verify tests, but the rest of it looks good to me.

onetechnical (approved):

PR #2882, opened 2021-09-13
Improve peerScheduler.peerDuration performance

algonautshant:
## Summary

In `onTransactionPoolChangedEvent`, there is an action taken on every peer which is both in `s.interruptablePeers` and in `s.scheduler.peers`.
This is accomplished in the loop:
```
	for _, peer := range peers {
		peerNext := s.scheduler.peerDuration(peer)
```
where for each peer, the list of `s.scheduler.peers` is traversed when `peerDuration` is called. 

This is `len(peers) * len(s.scheduler.peers)` complexity. 
The task here is to make this operation linear instead of potentially a quadratic operation. 

## Test Plan
A test is added.

### DIFF
--- a/txnsync/peerscheduler.go
+++ b/txnsync/peerscheduler.go
@@ -17,7 +17,6 @@
 package txnsync
 
 import (
-	"fmt"
 	"container/heap"
 	"sort"
 	"time"
@@ -51,15 +50,11 @@ func (p *peerScheduler) Push(x interface{}) {
 	p.peers = append(p.peers, entry)
 	p.nextPeers[entry.peer] = append(p.nextPeers[entry.peer], len(p.peers)-1)
 
-
 	if len(p.nextPeers[entry.peer]) > 1 {
 		peerIndices := p.nextPeers[entry.peer]
 		sort.Slice(peerIndices, func(i, j int) bool {
 			return p.peers[peerIndices[i]].next < p.peers[peerIndices[j]].next
 		})
-		if p.peers[peerIndices[0]].next > p.peers[peerIndices[1]].next {
-			fmt.Println("push boom")
-		}
 	}
 
 }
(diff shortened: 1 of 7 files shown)


txnsync/peerscheduler.go
```diff
@@ -55,9 +97,32 @@ func (p *peerScheduler) Len() int {
...
+}
+
 // Swap implements heap.Interface
 func (p *peerScheduler) Swap(i, j int) {
 	p.peers[i], p.peers[j] = p.peers[j], p.peers[i]
+	if p.peers[i].peer == p.peers[j].peer {
+		indices := p.nextPeers[p.peers[i].peer]
+		sort.Slice(indices, func(i, j int) bool {
+			return p.peers[indices[i]].next < p.peers[indices[j]].next
+		})
```
tsachiherman, line 121: consider avoiding using the same variable names are above ( i.e. `i` and `j` ); pick different variable names.

txnsync/peerscheduler.go
```diff
@@ -55,9 +97,32 @@ func (p *peerScheduler) Len() int {
...
 
+func (p *peerScheduler) replaceIndices(indices []int, i, j int) {
+
+	for idx, x := range indices {
+		if x == i {
+			indices[idx] = j
+		} else if x == j {
+			indices[idx] = i
+		}
+	}
```
tsachiherman, line 108: In this loop you're doing 2 comparison per element.. which is not needed.
The indices array is already ordered, so why not to take advantage of that ?
i.e. I can think of -
1. binary search
2. sort I & j. Scan to find x <= i first, then do the same for j.

algonautshant: The problem is that the array `indices` is sorted by the `p.peers[indices[i]].next` value. So binary search will not be straightforward. 
Given this, and knowing that this array is not going to be more than several elements long, two checks should not be a problem.

tsachiherman (approved):

PR #2885, opened 2021-09-14
adding encode_uvarint and decode_uvarint opcodes

barnjamin:
Trying to add an opcode to parse out a uvarint


barnjamin:
Maybe swap return values, maybe make it more like the extract* opcodes and require an index into the byte array for start position?

barnjamin:
To encode a uvarint, the following pyteal can be used
```py
@Subroutine(TealType.bytes)
def encode_uvarint(val: TealType.uint64, b: TealType.bytes):
    buff = ScratchVar()
    return Seq(
        buff.store(b),
        Concat(
            buff.load(),
            If(
                val >= Int(128),
                encode_uvarint(val >> Int(7), Extract(Itob((val & Int(255)) | Int(128)), Int(7), Int(1))),
                Extract(Itob(val & Int(255)), Int(7), Int(1))
            ),
        )
    )
    ```

PR #2887, opened 2021-09-14
changing circleci workflow name

algojack:
## Summary

Fixing naming due to new way of displaying CircleCI workflow

## Test Plan

Check tests at the bottom of Conversation tab and on the Checks tab

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -7,7 +7,7 @@ orbs:
 
 workflows:
   version: 2
-  build_pr:
+  "CircleCI Tests":
     jobs:
       - codegen_verification
       - amd64_build


algobarb:
I think I would rather not use spaces if we can help it in the yaml file. Maybe something like `circleci-build-and-tests` workflow?

algojack:
> I think I would rather not use spaces if we can help it in the yaml file. Maybe something like `circleci-build-and-tests` workflow?

May I ask why? It specifically shows names like that in [their example here](https://circleci.com/docs/2.0/configuration-reference/#orbs-requires-version-21). And it is much more human readable than using dashes or underscores.

algobarb:
> > I think I would rather not use spaces if we can help it in the yaml file. Maybe something like `circleci-build-and-tests` workflow?
> 
> May I ask why? It specifically shows names like that in [their example here](https://circleci.com/docs/2.0/configuration-reference/#orbs-requires-version-21). And it is much more human readable than using dashes or underscores.

I think the reason to avoid spaces in the keys is so it's closer to usual YAML convention where keys can easily be referenced, if the need arises. Also, most of the examples they have don't use spaces for workflow names: https://circleci.com/docs/2.0/sample-config/

algojack:
@algobarb good point, updated to one of their examples "build_and_test" (also we already know it's CircleCI since it is wrapped in title "CircleCI Checks".

algobarb:
> @algobarb good point, updated to one of their examples "build_and_test" (also we already know it's CircleCI since it is wrapped in title "CircleCI Checks" at least in the Checks tab... let me know if you think we should bring back circleci piece)

I don't think explicitly saying "circleci" is necessary.

algojack:
this PR can be merged whenever we are ready. The `build_pr` job is not going to finish, since it was renamde to `build_and_test` by this PR. We do need to change the required jobs setting to have the new name `build_and_test` instead of `build_pr` when we merge this.

PR #2890, opened 2021-09-14
filter lookback

brianolson:
## Summary

Build the next filter without the elements that were in the previous filter.

## Test Plan

unit tests, cluster bandwidth test


tsachiherman (commented):
I think that the idea you have here is correct, but it would require some in order to implement it:
1. The "sender" side, need to track the highest group counter that it was including in a bloom filter, per `requestParams`. Then, it should filter out the entries below that when generating a new bloom filter.
2. the "receiver" side, should keep 50k worth of bloom filter data, per unique `requestParams` ( with some reasonable limits on the total number of bloom filter counts ).

You would likely need to add some more (trivial) data structure to `Peer` in order to accomplish that, but you wouldn't have any case where you would need to `decode` an existing bloom filter.
( the decoding portion won't work correctly since the bloom filters `requestParams` are being rotated. i.e. changing colors )

brianolson:
I think the receiver side can't limit to only one filter per (Modulo,Offset). There's now going to be multiple partial filters per (Modulo,Offset) since each won't include elements in the prior filters. I've been looking at code around `peer.recentIncomingBloomFilters` and I think it should change to a simple circular buffer just keeping the newest of whatever the peer has sent.

tsachiherman:
> I think the receiver side can't limit to only one filter per (Modulo,Offset). There's now going to be multiple partial filters per (Modulo,Offset) since each won't include elements in the prior filters. I've been looking at code around `peer.recentIncomingBloomFilters` and I think it should change to a simple circular buffer just keeping the newest of whatever the peer has sent.

I do like the idea of circular buffer. But the key here is that we want certain "count" of transactions being captured in these buffers - and the number of filters isn't a good reflection of that.

brianolson:
Trying to talk out the rational for buffering 'some number of txns' in the filters:
Assuming that txns globally process _pretty much_ in order (not sure how true that is, but maybe enough),
txns more than one or two full txn-pools in the past definitely won't be coming up new, so any filter match from that long ago would have to be a false positive. So, we should throw away filters more than ((max txn pool)*k) filter elements ago.
I'm tempted to add an explicit 'number of elements filtered' field in the filter protocol struct.

tsachiherman:
> Trying to talk out the rational for buffering 'some number of txns' in the filters:
> Assuming that txns globally process _pretty much_ in order (not sure how true that is, but maybe enough),
> txns more than one or two full txn-pools in the past definitely won't be coming up new, so any filter match from that long ago would have to be a false positive. So, we should throw away filters more than ((max txn pool)*k) filter elements ago.
> I'm tempted to add an explicit 'number of elements filtered' field in the filter protocol struct.

So, I think that we're both on the same page here - but looking on different use cases. The two use cases I'm think about are:
- When a node is *unsynchronized* with the rest of the network, the purpose of the txnsync is to get it's txpool synchronized. If we won't exchange sufficiently full bloom filters, we would end up sending more data in the form of transactions. ( hence, the total cost is going to be higher )
- Storing the lastest N bf might not be enough in some cases ( imagine a single large bf, following by a series of smaller ones. If the smaller ones cycles out the large one, and we ended up sending transactions that aren't needed ), while also might be an overkill in others ( i.e. imagine series of BF that each is the size of half tx pool ).

I think that total of (let's say) 8-10 BF is good, but we might want to keep *less* in case the total reflected transactions are exceed the size of the transaction pool.

tsachiherman:
I'm closing this one since your other PR ( for the same goal ) is almost complete. If you feel that you would still need this, please feel free to re-open it.

PR #2894, opened 2021-09-15
Fix dryrun crash on rewards calculation

algorandskiy:
## Summary

dryrun/tealdbg ledgers do not have block info so cannot provide correct value of `rewardLevel` used in balance calculation for `balance` opcode. Fixed by clearing `RewardBase` field in account data returned by dryrun/tealdbg ledgers.

The crash was [reported](https://forum.algorand.org/t/teal-rejection-error-panic-in-evaluation-overflowed-account-balance-when-applying-rewards/4294) at the community forum.

## Test Plan

Test added


jasonpaulos (approved):
Looks like a good fix to me

jannotti (approved):
Seems right.

PR #2897, opened 2021-09-15
Add `indexerLedgerForEval` interface

tolikzinovyev:
## Summary

Add a new ledger interface for Indexer that is significantly simpler and allows implementing batching in a more straight-forward way. As a result, the Indexer code doesn't need to know the specifics of go-algorand and doesn't need to implement its own accounts cache for batching.


ledger/eval.go
```diff
@@ -1525,8 +1525,108 @@ func (vb ValidatedBlock) WithSeed(s committee.Seed) ValidatedBlock {
...
+	Address basics.Address
+	Exists  bool
+}
+
+// A ledger interface that Indexer implements. This is a simplified version of the
+// ledgerForEvaluator interface. Certain functions that the evaluator doesn't use
+// in the trusting mode are excluded, and the present functions only request data
+// at the latest round.
+type indexerLedgerForEval interface {
+	BlockHdr(basics.Round) (bookkeeping.BlockHeader, error)
```
tsachiherman, line 1539: when do you need different block header rounds ?

tolikzinovyev: We only need to previous block header. Do you think I should simplify it?

tsachiherman: Yes; Naming it LatestBlockHdr() would be a good name... and in `indexerLedgerConnector. BlockHdr` you should test that round == blockHdr.Round(), or error. ( i.e. verify that the ledger wasn't really referring to a header which isn't the latest )

tsachiherman: btw - same applied for Totals()

tsachiherman (commented):
Do you think that we can move the indexer related implementation to its own file ?
maybe indexereval.X ?

ledger/eval.go
tolikzinovyev: done

tolikzinovyev:
> Do you think that we can move the indexer related implementation to its own file ?
> maybe indexereval.X ?

Good idea, I moved the indexer code.

tsachiherman (dismissed):

tsachiherman:
Looks good - could you please verify it passes all the auto tests ?

tsachiherman (approved):

PR #2900, opened 2021-09-15
Enable debugging itxn programs

algorandskiy:
## Summary

Set special addressed in EvalParams when evaluating from tealdbg/dryrun

## Test Plan

Tests added

### DIFF
--- a/daemon/algod/api/server/v2/dryrun.go
+++ b/daemon/algod/api/server/v2/dryrun.go
@@ -87,7 +87,7 @@ func (dr *DryrunRequest) ExpandSources() error {
 	for i, s := range dr.Sources {
 		ops, err := logic.AssembleString(s.Source)
 		if err != nil {
-			return fmt.Errorf("Dryrun Source[%d]: %v", i, err)
+			return fmt.Errorf("dryrun Source[%d]: %v", i, err)
 		}
 		switch s.FieldName {
 		case "lsig":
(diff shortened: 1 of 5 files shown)


daemon/algod/api/server/v2/dryrun.go
```diff
@@ -338,41 +338,6 @@ func (dl *dryrunLedger) GetCreatorForRound(rnd basics.Round, cidx basics.Creatab
 	return basics.Address{}, false, fmt.Errorf("unknown creatable type %d", ctype)
 }
 
-func (dl *dryrunLedger) getAppParams(addr basics.Address, aidx basics.AppIndex) (params basics.AppParams, err error) {
```
algorandskiy, line 341: removed some unused code

daemon/algod/api/server/v2/dryrun_test.go
```diff
@@ -1331,3 +1332,95 @@ func TestDryrunCost(t *testing.T) {
...
+			},
+		},
+	}
+
+	dr.ProtocolVersion = string(dryrunProtoVersion)
+
+	var response generated.DryrunResponse
+	doDryrunRequest(&dr, &response)
+	require.NoError(t, err)
+	checkAppCallPass(t, &response)
```
jasonpaulos, line 1422: It seems like we neglected to add inner transactions to the `DryrunTxnResult` API type. @jannotti was there a reason for this, or did we just forgot about it?

Ideally any created inner transactions would be returned in that response and this test could verify they are there and properly formatted.

jannotti: Definitely not on purpose.

jannotti: I am incapable of expressing how much I hate that dryrun lives in its own universe where we have to do everything all over again.

jannotti (approved):

PR #2903, opened 2021-09-17
Fix random failures TestWaitAndAddConnectionTimeLongtWindow    

algonautshant:
## Summary
TestWaitAndAddConnectionTimeLongtWindow is randomly failing when the execution of time sensitive statements is separated by more than the expected time gap.

This issue is fixed by replacing the sleep with simulated time passing by changing the recorded time, thus giving the impression that the event happened timeout time ago. 

## Test Plan
This is a test fix.


tsachiherman (changes requested):
looks good, but please make sure it passes all the automated test.

tsachiherman (approved):

PR #2919, opened 2021-09-17
GroupCounter based iterative txnsync filters

brianolson:
## Summary

Only send bloom/xor filter for txns that came in after the last filter was sent.

## Test Plan

unit test. S1 cluster bw test.


util/bloom/bloom.go
```diff
@@ -90,6 +92,10 @@ func (f *Filter) Test(x []byte) bool {
 	return true
 }
 
+func (f *Filter) NumEntries() int {
```
tsachiherman, line 95: It's the other way around: you need to know the number of entries on *incoming* bloom filters, so you never had the "Set" function called there. instead, you should deduce that number during the decoding from the bloom filter size.

brianolson: I pack the number all the way through, added a field to the exchange encoded bloom filter

tsachiherman: for the xor filter it might be ok, but not for the bloom filter, since it would break existing clients.

tsachiherman: I'm also abit worried if that's open an attack vector.

txnsync/bloomFilter.go
```diff
@@ -142,6 +146,10 @@ func (s *syncState) makeBloomFilter(encodingParams requestParams, txnGroups []po
 		// we want none.
 		return
 	}
+	minGroupCounter := uint64(0)
+	if hintPrevBloomFilter != nil {
+		minGroupCounter = hintPrevBloomFilter.containedTxnsRange.lastCounter
```
tsachiherman, line 151: The `hintPrevBloomFilter` is a previously generated bloom filter.. but it's not the last one sent to the peer.

tsachiherman: also, it would be incorrect to use the lastCounter from the latest bloom filter that *was* sent to the peer, unless the  encodingParams are identical. ( that's what I addressed on my branch.. )

txnsync/peer.go
```diff
@@ -497,7 +504,44 @@ func incomingPeersOnly(peers []*Peer) (incomingPeers []*Peer) {
...
+			p.recentIncomingBloomFilters[prev].filter = nil
+		} else {
+			activeElements += int(p.recentIncomingBloomFilters[prev].filter.elementsFiltered)
+		}
+		prev = (prev + maxIncomingBloomFilterHistory - 1) % maxIncomingBloomFilterHistory
+	}
+
+	p.recentIncomingBloomFiltersNext = (p.recentIncomingBloomFiltersNext + 1) % maxIncomingBloomFilterHistory
+
+	// TODO: limit total elements to maxIncomingFilterElements
```
tsachiherman, line 541: Is this comment still relevant ?

txnsync/peer.go
```diff
@@ -497,7 +504,44 @@ func incomingPeersOnly(peers []*Peer) (incomingPeers []*Peer) {
...
+				activeElements += int(p.recentIncomingBloomFilters[prev].filter.elementsFiltered)
+			}
+			prev--
+		}
+		return
+	}
+
+	p.recentIncomingBloomFilters[p.recentIncomingBloomFiltersNext] = bf
+	prev := (p.recentIncomingBloomFiltersNext + maxIncomingBloomFilterHistory - 1) % maxIncomingBloomFilterHistory
+	for prev != p.recentIncomingBloomFiltersNext {
```
tsachiherman, line 530: you can also stop when `filter == nil`

txnsync/peer.go
```diff
@@ -497,7 +504,44 @@ func incomingPeersOnly(peers []*Peer) (incomingPeers []*Peer) {
...
+		} else {
+			activeElements += int(p.recentIncomingBloomFilters[prev].filter.elementsFiltered)
+		}
+		prev = (prev + maxIncomingBloomFilterHistory - 1) % maxIncomingBloomFilterHistory
+	}
+
+	p.recentIncomingBloomFiltersNext = (p.recentIncomingBloomFiltersNext + 1) % maxIncomingBloomFilterHistory
+
+	// TODO: limit total elements to maxIncomingFilterElements
+}
```
tsachiherman, line 542: you would still need
```golang
// reset the counter, since we might need to re-evaluate some of the transaction group with the new bloom filter.
	p.lastTransactionSelectionTracker.roll(incomingFilter.encodingParams.Offset, incomingFilter.encodingParams.Modulator)
```

txnsync/peer.go
```diff
@@ -142,6 +144,8 @@ type Peer struct {
 	// This bloom filter could be stale if no bloom filter was included in the last message.
 	lastSentBloomFilter bloomFilter
 
+	lastSentFilterGroupCounter uint64
```
tsachiherman, line 147: not used.. but it should. See my branch..

tsachiherman (changes requested):
I think that we'll need to try to run this again after you'll address the various comments I've made. Take a look on my branch for an idea how to implement them.
The changes in peer.go looks good - but I'm abit worried about the 200 history, since it seems too large.

txnsync/peer.go
```diff
@@ -368,6 +372,9 @@ func (p *Peer) selectPendingTransactions(pendingTransactions []pooldata.SignedTx
 	var effectiveBloomFilters []int
 	effectiveBloomFilters = make([]int, 0, len(p.recentIncomingBloomFilters))
 	for filterIdx := len(p.recentIncomingBloomFilters) - 1; filterIdx >= 0; filterIdx-- {
+		if p.recentIncomingBloomFilters[filterIdx].filter == nil {
```
tsachiherman, line 375: we might want to limit the size of the `effectiveBloomFilters` so it would not be bigger than X ( not sure what the X should be.. )

txnsync/bloomFilter.go
```diff
@@ -136,7 +139,7 @@ func filterFactoryXor32(numEntries int, s *syncState) (filter bloom.GenericFilte
 
 var filterFactory func(int, *syncState) (filter bloom.GenericFilter, filterType bloomFilterType) = filterFactoryXor32
 
-func (s *syncState) makeBloomFilter(encodingParams requestParams, txnGroups []pooldata.SignedTxGroup, excludeTransactions *transactionCache, hintPrevBloomFilter *bloomFilter) (result bloomFilter) {
+func (s *syncState) makeBloomFilter(encodingParams requestParams, txnGroups []pooldata.SignedTxGroup, excludeTransactions *transactionCache, hintPrevBloomFilter *bloomFilter, minGroupCounter uint64) (result bloomFilter) {
```
tsachiherman, line 142: The changes that you've made here could be done by the caller, which would do them more efficiently : 
instead of comparing each of the groups in txnGroups to minGroupCounter, the caller should use sort.Search on the txnGroups. The entries in the txnGroups are ordered by their GroupCounter, which would allow you to perform a binary search. Then, sub-slice the txnGroups and pass it to `makeBloomFilter`

txnsync/outgoing.go
```diff
@@ -152,7 +152,7 @@ func (s *syncState) sendMessageLoop(currentTime time.Duration, deadline timers.D
 		// correct sent bloom filter, regardless of the message sending timing. If and when we
 		// generate the next message, we need to ensure that we're aware of this bloom filter, since
 		// it would affect whether we re-generate another bloom filter or not.
-		peer.updateSentBoomFilter(assembledBloomFilter)
+		peer.updateSentBoomFilter(assembledBloomFilter, msgEncoder.messageData.message.UpdatedRequestParams, s.round)
```
tsachiherman, line 155: you don't need to pass the `msgEncoder.messageData.message.UpdatedRequestParams` to the method; that information is already in `assembledBloomFilter.encoded.EncodingParams`

txnsync/peer.go
```diff
@@ -118,6 +119,8 @@ type Peer struct {
 	// array to determine if the peer already has this message.
 	recentIncomingBloomFilters []incomingBloomFilter
 
+	recentIncomingBloomFiltersNext int
```
tsachiherman, line 122: if you're not using `recentIncomingBloomFiltersNext`, remove it.

txnsync/peer.go
```diff
@@ -497,7 +506,76 @@ func incomingPeersOnly(peers []*Peer) (incomingPeers []*Peer) {
...
+	for i, ribf := range p.recentIncomingBloomFilters {
+		if ribf.round == oldestRound {
+			if ari == evictIndex {
+				p.recentIncomingBloomFilters[i] = bf
+				return
+			}
+			ari++
+		}
+	}
+	panic("p.recentIncomingBloomFilters changed out from under addIncomingBloomFilter")
```
tsachiherman, line 575: Please try to avoid using randomness here - it would cause the emulator to generate random results, which would make it completely useless. Instead, try to use some constant logic.
More specifically, I think that we want to evict first `p.recentIncomingBloomFilters` that share the same encoding parameters, and then other encoding parameters for the oldestRound.

tsachiherman, line 575: Also, please avoid panicking here. It's not really helpful.. I know that it shouldn't happen - in which case you can log a message and keep going.

txnsync/peer_test.go
```diff
@@ -646,6 +646,7 @@ func TestUpdateRequestParams(t *testing.T) {
 
 // TestAddIncomingBloomFilter tests adding an incoming bloom filter
 func TestAddIncomingBloomFilter(t *testing.T) {
+	t.Skip("TODO: behavior has changed, make test sane again")
```
tsachiherman, line 649: please update the unit test..

txnsync/sent_filter.go
```diff
@@ -0,0 +1,63 @@
...
+	round basics.Round
+}
+
+// sentFilters is the set of filter stats for one peer to another peer.
+// There should be at most one entry per (Modulator,Offset)
+type sentFilters []sentFilterStat
+
+const maxSentFilterSet = 100
+
+func (sf *sentFilters) setSentFilter(filter bloomFilter, encodingParams requestParams, round basics.Round) {
```
tsachiherman, line 39: the encodingParams are already part of the bloomFilter

txnsync/sent_filter.go
```diff
@@ -0,0 +1,63 @@
...
+			(*sf)[i].round = round
+			return
+		}
+	}
+	nsf := sentFilterStat{
+		EncodingParams: encodingParams,
+		lastCounter:    filter.containedTxnsRange.lastCounter,
+		round:          round,
+	}
+	// TODO: enforce limit
```
tsachiherman, line 52: yes, we should enforce a size limit - I think that using a limit of 4+3+2+1=10 should be a safe choice. It would also be better performing than trying to add a map here.

tsachiherman (changes requested):
I've made several comments throughout the code. overall looks great.

txnsync/sent_filter.go
```diff
@@ -0,0 +1,80 @@
...
+		lastCounter:    filter.containedTxnsRange.lastCounter,
+		round:          round,
+	}
+	*sf = append(*sf, nsf)
+	// trim oldest content if we're too long
+	for len(*sf) > maxSentFilterSet {
+		oldestRound := round
+		for _, sfs := range *sf {
+			if sfs.round < oldestRound {
+				oldestRound = sfs.round
```
tsachiherman, line 59: nit: store the index here so you could avoid the subsequent loop.

brianolson: Done (in both places I used that pattern)

txnsync/peer.go
```diff
@@ -142,6 +144,8 @@ type Peer struct {
 	// This bloom filter could be stale if no bloom filter was included in the last message.
 	lastSentBloomFilter bloomFilter
 
+	sentFilterParams sentFilters
```
tsachiherman, line 147: could you please add a sentence here describing this variable ?

txnsync/peer_test.go
```diff
@@ -644,13 +657,32 @@ func TestUpdateRequestParams(t *testing.T) {
...
+// bloom.GenericFilter
+type nopFilter struct{}
+
+func (nf *nopFilter) Set(x []byte) {}
+func (nf *nopFilter) Test(x []byte) bool {
+	return false
+}
+func (nf *nopFilter) NumEntries() int {
+	return 0
+}
```
tsachiherman, line 669: I think that this method can be dropped, right ?

tsachiherman (approved):

PR #2923, opened 2021-09-20
Support arm cross compile

bricerisingalgorand:
## Summary

This PR is to take advantage of arm64's ability to run arm32 binaries. Using this, we are able to build for arm32 on arm64 hosts by running containers with arm32 OS's on them. This is much faster than our qemu set up because it doesn't rely on emulations. The binaries produced with this process work and the arm64/arm32 build times are as fast as the intel builds, reducing our build time by about an hour.

## Test Plan

I ran our build pipeline against this branch using arm64 hosts and tested the produced artifacts using emulated cpu's to make sure they still work on their native architectures.

### DIFF
--- a/scripts/archtype.sh
+++ b/scripts/archtype.sh
@@ -1,5 +1,10 @@
 #!/usr/bin/env bash
 
+if [ ! -z "${GOHOSTARCH+x}" ]; then
+    echo "${GOHOSTARCH}"
+    exit 0
+fi
+
 ARCH=$(uname -m)
 
 if [[ "${ARCH}" = "x86_64" ]]; then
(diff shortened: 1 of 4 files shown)


tsachiherman:
did you test the resulting binaries on Raspberry PI ?

bricerisingalgorand:
> did you test the resulting binaries on Raspberry PI ?

I want to but I don't have an armv6 raspberry pi. Do you happen to have one handy?

tsachiherman:
> > did you test the resulting binaries on Raspberry PI ?
> 
> I want to but I don't have an armv6 raspberry pi. Do you happen to have one handy?

I have one in the office - could you send me the produced binary and I'll give it a try tomorrow morning ?

tsachiherman (approved):
I've tested this on raspberry pi 2, and it's working just fine.

onetechnical:
Successfully tested on raspberry pi 4 as well.

PR #2924, opened 2021-09-20
Added Participation Key Expiration Check

### LINKED ISSUE #2738: Add transaction to take an invalid account offline
## Problem

Currently, if an account marks itself online but the last round has passed, it counts towards the total online stake but the account can't vote. This is considered to be malicious behavior by the protocol and can undermine the protocol's security. 

## Solution
First step - Allow anyone to take online accounts which are effectively offline (online status, no more keys, hasn't renewed) offline using a special transaction. 
Second Step - algod should do it automatically 

## Urgency

See problem

AlgoStephenAkiki:
Resolves #2738

Adds the following:

- consensus.go: Adds the default MaxExpiredAccountsToProcess value for
  the new consensus protocol
- block.go: Adds necessary block header entries
- eval.go: Scans for expired accounts, modifies them to be offline and adds
  validation for this use case
- eval_test.go: Basic unit tests
- participationExpiration_test.go: Added e2e tests that verify that
  different consensus protocols behave differently


tsachiherman:
Please update the PR's Summary and Test Plan.

data/bookkeeping/block.go
```diff
@@ -125,6 +125,19 @@ type (
...
 		CompactCert map[protocol.CompactCertType]CompactCertState `codec:"cc,allocbound=protocol.NumCompactCertTypes"`
+
+		// ParticipationUpdates contains the information needed to mark
+		// certain accounts offline because their participation keys expired
+		ParticipationUpdates
+	}
+
+	// ParticipationUpdates represents participation account data that
+	// needs to be checked/acted on by the network
+	ParticipationUpdates struct {
```
tsachiherman, line 136: Add 
```golang
```
as the first field of the `ParticipationUpdates` struct.

ledger/eval.go
```diff
@@ -1032,6 +1032,41 @@ func (eval *BlockEvaluator) endOfBlock() error {
...
+		// see if it meets the criteria for adding it to the expired
+		// participation accounts list.
+		modifiedAccounts := eval.state.mods.Accts.ModifiedAccounts()
+		currentRound := eval.Round()
+
+		proto, ok := config.Consensus[eval.block.BlockHeader.CurrentProtocol]
+		if !ok {
+			return fmt.Errorf("unknown consensus version: %s", eval.block.BlockHeader.CurrentProtocol)
+		}
+
```
tsachiherman, line 1047: Use `eval.proto` instead of getting the protocol parameters from `config.Consensus`

ledger/eval.go
```diff
@@ -1049,6 +1084,47 @@ func (eval *BlockEvaluator) endOfBlock() error {
...
 }
 
+// modifyOfflineAccounts after all transactions and rewards are processed, modify the accounts so that their status is offline
+func (eval *BlockEvaluator) modifyOfflineAccounts() error {
+
+	proto, ok := config.Consensus[eval.block.BlockHeader.CurrentProtocol]
+	if !ok {
+		return fmt.Errorf("unknown consensus version: %s", eval.block.BlockHeader.CurrentProtocol)
+	}
+
```
tsachiherman, line 1094: same as before - use `eval.proto`

ledger/eval.go
```diff
@@ -1049,6 +1084,47 @@ func (eval *BlockEvaluator) endOfBlock() error {
...
+	// This works when the expected number of accounts is zero (i.e. it is disabled) as well
+	if lengthOfExpiredParticipationAccounts > expectedMaxNumberOfExpiredAccounts {
+		return fmt.Errorf("length of expired accounts (%d) was greater than expected (%d)",
+			lengthOfExpiredParticipationAccounts, expectedMaxNumberOfExpiredAccounts)
+	}
+
+	for _, accountAddr := range eval.block.ParticipationUpdates.ExpiredParticipationAccounts {
+		acctData, err := eval.state.lookup(accountAddr)
+		if err != nil {
+			return fmt.Errorf("modifyOfflineAccounts wrong: %v was not found", accountAddr)
```
tsachiherman, line 1108: an error would be returned when the lookup failed, but not when the lookup was looking for a non-existing account.
If you were to look a non-existing account, you would get the "empty" account data.
Therefore, returning an error here is good - but the error string isn't; it should say something like:
```golang
    return fmt.Errorf("modifyOfflineAccounts was unable to retrieve account %v : %w", accountAddr, err)
```

ledger/eval.go
```diff
@@ -1049,6 +1084,47 @@ func (eval *BlockEvaluator) endOfBlock() error {
...
+		}
+
+		// Reset the appropriate account data
+		acctData.Status = basics.Offline
+		acctData.VoteFirstValid = basics.Round(0)
+		acctData.VoteLastValid = basics.Round(0)
+		acctData.VoteKeyDilution = 0
+		acctData.VoteID = crypto.OneTimeSignatureVerifier{}
+		acctData.SelectionID = crypto.VRFVerifier{}
```
tsachiherman, line 1117: I think that it would be cleaner if you were to create a `ClearOnlineState` on `AccountData` and call it from here. In the future, we might have additional fields, and it would make it cleaner having all of them in a single place.

ledger/eval.go
```diff
@@ -1088,6 +1164,56 @@ func (eval *BlockEvaluator) finalValidation() error {
 				return fmt.Errorf("CompactCertType %d unexpected", ccType)
 			}
 		}
+
+		proto, ok := config.Consensus[eval.block.BlockHeader.CurrentProtocol]
+		if !ok {
+			return fmt.Errorf("unknown consensus version: %s", eval.block.BlockHeader.CurrentProtocol)
+		}
```
tsachiherman, line 1171: use `eval.proto`..

ledger/eval.go
```diff
@@ -1088,6 +1164,56 @@ func (eval *BlockEvaluator) finalValidation() error {
...
+		// If the length of the array is strictly greater than our max then we have an error.
+		// This works when the expected number of accounts is zero (i.e. it is disabled) as well
+		if lengthOfExpiredParticipationAccounts > expectedMaxNumberOfExpiredAccounts {
+			return fmt.Errorf("length of expired accounts (%d) was greater than expected (%d)",
+				lengthOfExpiredParticipationAccounts, expectedMaxNumberOfExpiredAccounts)
+		}
+
+		// For security reasons, we need to make sure that all addresses in the expired participation accounts
+		// are unique.  We make this map to keep track of previously seen address
+		addressSet := make(map[basics.Address]bool)
```
tsachiherman, line 1185: preallocate with size : 
initialize the map to have `eval.block.ParticipationUpdates.ExpiredParticipationAccounts` entries.

ledger/eval.go
```diff
@@ -1088,6 +1164,56 @@ func (eval *BlockEvaluator) finalValidation() error {
...
+				// We shouldn't have duplicate addresses...
+				return fmt.Errorf("duplicate address found: %v", accountAddr)
+			}
+
+			// Record that we have seen this address
+			addressSet[accountAddr] = true
+
+			acctData, err := eval.state.lookup(accountAddr)
+			if err != nil {
+				return fmt.Errorf("ExpiredParticipationAccount wrong: %v was not found", accountAddr)
```
tsachiherman, line 1201: same note as before; just one more comment - errors should always start with a lower-case.

ledger/eval.go
```diff
@@ -1119,6 +1245,11 @@ func (eval *BlockEvaluator) GenerateBlock() (*ValidatedBlock, error) {
 		return nil, err
 	}
 
+	err = eval.modifyOfflineAccounts()
```
tsachiherman, line 1248: could you move this call to `eval.finalValidation()`, just instead of "return nil" at the end ?
see https://github.com/algorand/go-algorand/pull/2922 for the rationale..

tsachiherman (changes requested):
Overall looks good - I've had few requests, some are just code-style, some are more around compatibility with other PRs.

ledger/eval.go
```diff
@@ -1088,9 +1149,54 @@ func (eval *BlockEvaluator) finalValidation() error {
...
+			}
+
+			if !pastCurrentRound {
+				return fmt.Errorf("finalValidation found %v round (%v) was not less than current round (%v)", accountAddr, acctData.VoteLastValid, currentRound)
+			}
+		}
 	}
 
-	return nil
+	return eval.modifyOfflineAccounts()
```
tsachiherman, line 1199: I think that I found a missing call to
```golang
finalValidation() error
```
from 
```golang
func Eval(l ledgerForEvaluator, blk *bookkeeping.Block, proto config.ConsensusParams) (ledgercore.StateDelta, []transactions.SignedTxnInBlock, error) {
```

The `Eval` should call `finalValidation` right after `endOfBlock`, just like `GenerateBlock`

ledger/eval.go
```diff
@@ -1088,9 +1149,54 @@ func (eval *BlockEvaluator) finalValidation() error {
...
+			isOnline := acctData.Status == basics.Online
+			// true if the accounts last valid round has passed
+			pastCurrentRound := acctData.VoteLastValid < currentRound
+
+			if !isOnline {
+				return fmt.Errorf("finalValidation found %v was not online but %v", accountAddr, acctData.Status)
+			}
+
+			if !pastCurrentRound {
+				return fmt.Errorf("finalValidation found %v round (%v) was not less than current round (%v)", accountAddr, acctData.VoteLastValid, currentRound)
```
tsachiherman, line 1194: the last two entries should be using `%d` instead of `%v`

tsachiherman (changes requested):
given that you've modified the encoding of the block header, you'll need to regenerate the msgp in order to make it persistent; use `make msgp` to regenerate all the auto-gen msgp encoders/decoders.

ledger/eval.go
```diff
@@ -1088,9 +1149,61 @@ func (eval *BlockEvaluator) finalValidation() error {
...
 	}
 
-	return eval.state.CalculateTotals()
+	err := eval.state.CalculateTotals()
+
+	if err != nil {
+		return err
+	}
+
+	return eval.modifyOfflineAccounts()
```
tsachiherman, line 1205: You'll need to swap the two : calculating the account totals has to happen only after the accounts get modified.

tsachiherman (changes requested):
The test you've added at `TestExpiredAccountGeneration` is good - but we would need more extensive unit testing. In practice, you change would affect the `GenerateBlock()`, `eval()` and `EvalForIndexer`. For each one of these, we need to make sure that the code changes produce their expected output;
as a generic guideline, each of the if-statements that you've included in your PR need to be tested for both success and failure.

ledger/eval_test.go
```diff
@@ -20,6 +20,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"github.com/algorand/go-algorand/data/transactions/verify"
```
tsachiherman, line 23: move this import to the group below..

PR #2925, opened 2021-09-20
Update telemetry credentials

bricerisingalgorand:
## Summary

Update telemetry configs so that we can stop receiving logs from very old algod instances that should not be running anymore. The plan is to deactivate the old credentials at some point after this change goes to mainnet

## Test Plan

I built a local branch with the new credentials and tested that telemetry.go isn't showing errors and logs are being reported to telemetry (stable-testnet-v1.0).

### DIFF
--- a/logging/telemetryConfig.go
+++ b/logging/telemetryConfig.go
@@ -33,8 +33,8 @@ import (
 // TelemetryConfigFilename default file name for telemetry config "logging.config"
 var TelemetryConfigFilename = "logging.config"
 
-var defaultTelemetryUsername = "telemetry-v9"
-var defaultTelemetryPassword = "oq%$FA1TOJ!yYeMEcJ7D688eEOE#MGCu"
+var defaultTelemetryUsername = "telemetry-v10"
+var defaultTelemetryPassword = "ndn9lNvmoRD46i5gMljBZg"
 
 const hostnameLength = 255
 
--- a/test/testdata/configs/logging/logging.config.example
+++ b/test/testdata/configs/logging/logging.config.example
@@ -5,6 +5,6 @@
   "MinLogLevel":        4,
   "ReportHistoryLevel": 4,
   "LogHistoryDepth":    100,
-  "UserName":           "telemetry-v9",
-  "Password":           "oq%$FA1TOJ!yYeMEcJ7D688eEOE#MGCu"
+  "UserName":           "telemetry-v10",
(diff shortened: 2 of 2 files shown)


winder (approved):
LGTM

tsachiherman:
I think that it might be more complicated than that. All the participating nodes that have the logging.config file, have the old (default) credentials written into these files. These credentials would override the defaults. So, changing the defaults is good, but we won't be able to deprecate the old credentials without loosing quite many nodes...

bricerisingalgorand:
> I think that it might be more complicated than that. All the participating nodes that have the logging.config file, have the old (default) credentials written into these files. These credentials would override the defaults. So, changing the defaults is good, but we won't be able to deprecate the old credentials without loosing quite many nodes...

Did enabling telemetry used to write credentials to the logging config? When I did it today it just wrote empty strings. Do you think maybe posting a warning long in advance would be enough before deactivating the old credentials?
```
"UserName":"","Password":"",
```

tsachiherman:
That's true if you were to run the new codebase. However, "old" participation nodes were creating a file in <home>/.algorand/logging.config, which included the credentials. That file is still being used today ( for backward compatibility ).

tsachiherman:
We already have a project for the next quarter that is intended to eliminate this issue.

bricerisingalgorand:
> We already have a project for the next quarter that is intended to eliminate this issue.

Oh to eliminate the issue of old nodes logging to telemetry?

algojack (approved):
lgtm

algobarb:
Brice mentioned, "I talked with tsachi earlier, this change actually won’t be needed. His team will be rewriting telemetry in Q4 and remove the need for elastic creds hardcoded in app"

winder:
@bricerisingalgorand we should try to merge this in before the consensus upgrade. I'm fairly certain that we don't write the default username/password to the file, so changing the defaults would work fine.

tsachiherman:
> @bricerisingalgorand we should try to merge this in before the consensus upgrade. I'm fairly certain that we don't write the default username/password to the file, so changing the defaults would work fine.
> 
> This is where it happens:
> 
> https://github.com/algorand/go-algorand/blob/9cefeaaf3b6c97d9e0959b882a544226c247461a/logging/telemetryConfig.go#L95

We don't write these today, but we used to. Existing nodes that already has these written on a file would use the old credentials.

PR #2934, opened 2021-09-21
Rotate Travis Env Variables and Remove codecov Env variable

algobarb:
## Summary
Rotated slack token in environment variable and removed the codecov token from Travis file. codecov token is removed because it's not needed for public repos.

## Test Plan

Check that tests are still passing with the new change. Unable to test the impact on slack alerts on rel/nightly, but the impact should be quite low. Also am using a branch rather than a forked repo since I am modifying travis environment variables.


algojack (approved):
lgtm

onetechnical (approved):

PR #2935, opened 2021-09-21
Allow getting versions for any package.

winder:
## Summary

Add an option to support version checking for arbitrary packages. This would allow us to deploy indexer releases at `algorand-releases/<channel>/indexer-*`

## Test Plan

Ran the command and see the correct regex:
```
~$ updater ver check -c nightly -p indexer
Checking for files matching: 'channel/nightly/indexer_nightly_linux-amd64_' in bucket algorand-releases
no updates found for channel 'nightly'
```

### DIFF
--- a/cmd/updater/versionCmd.go
+++ b/cmd/updater/versionCmd.go
@@ -33,17 +33,20 @@ var (
 	semanticOutput  bool
 )
 
+// DefaultPackageName is the package we'll use by default.
+const DefaultPackageName = "node"
+
 func init() {
 	versionCmd.AddCommand(checkCmd)
 	versionCmd.AddCommand(getCmd)
 
 	checkCmd.Flags().BoolVarP(&semanticOutput, "semantic", "s", false, "Human readable semantic version output.")
-	checkCmd.Flags().StringVarP(&packageName, "package", "p", "node", "Get version of specific package.")
+	checkCmd.Flags().StringVarP(&packageName, "package", "p", DefaultPackageName, "Get version of specific package.")
 	checkCmd.Flags().StringVarP(&versionBucket, "bucket", "b", "", "S3 bucket containing updates.")
 
 	getCmd.Flags().StringVarP(&destFile, "outputFile", "o", "", "Path for downloaded file (required).")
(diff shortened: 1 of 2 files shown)


cmd/updater/versionCmd.go
```diff
@@ -67,7 +70,7 @@ var checkCmd = &cobra.Command{
 		if err != nil {
 			exitErrorf("Error creating s3 session %s\n", err.Error())
 		} else {
-			version, _, err := s3Session.GetLatestUpdateVersion(channel)
+			version, _, err := s3Session.GetPackageVersion(channel, packageName, 0)
```
tsachiherman, line 73: looks like you could deprecate `GetLatestUpdateVersion` with this change.

cmd/updater/versionCmd.go
```diff
@@ -36,12 +37,14 @@ func init() {
...
 
-	checkCmd.Flags().StringVarP(&versionBucket, "bucket", "b", "", "S3 bucket containing updates.")
 	checkCmd.Flags().BoolVarP(&semanticOutput, "semantic", "s", false, "Human readable semantic version output.")
+	checkCmd.Flags().StringVarP(&packageName, "package", "p", "node", "Get version of specific package.")
+	checkCmd.Flags().StringVarP(&versionBucket, "bucket", "b", "", "S3 bucket containing updates.")
 
 	getCmd.Flags().StringVarP(&destFile, "outputFile", "o", "", "Path for downloaded file (required).")
-	getCmd.Flags().StringVarP(&versionBucket, "bucket", "b", "", "S3 bucket containing updates.")
 	getCmd.Flags().Uint64VarP(&specificVersion, "version", "v", 0, "Specific version to download.")
+	getCmd.Flags().StringVarP(&packageName, "package", "p", "node", "Get version of specific package.")
```
tsachiherman, line 46: replace the "node" string with a constant - this is not just a "claim" for a node package name; it's the naming conventions we use for creating package names. It should be a const string. Ideally, we would derive the same string when the package is created ( but we could do that on a separate, future, PR )

tsachiherman (changes requested):
few small requests, but looks good otherwise.

PR #2936, opened 2021-09-22
Lots of testing convenience improvements, including e2e python

jannotti:
Made it nice to write e2e test in python my creating a "goal" module
that emulates the convenience of goal, but in python so that
structured results come out for easy asserts.


brianolson (commented):
The e2e_subs/*.py option is great.  I think we'll still need *.sh options for explicitly testing `goal`, also for testing new features that aren't in the SDK yet. There could be some circular dependency problems where the feature isn't done yet in `algod` so the support definitely isn't in the sdk yet but we still want to test and having all the code in the go-algorand repo (e.g. `goal` and `algod`) will be important. Still, cool addition and I like it.

data/transactions/logic/README.md
```diff
@@ -36,11 +37,12 @@ TEAL LogicSigs run in Algorand nodes as part of testing a proposed transaction t
...
+The TEAL program has access to data from the transaction it is attached to (`txn` op), any transactions in a transaction group it is part of (`gtxn` op), and a few global values like consensus parameters (`global` op). Some "Args" may be attached to a transaction being validated by a TEAL program. Args are an array of byte strings. A common pattern would be to have the key to unlock some contract as an Arg. Args are recorded on the blockchain and publicly visible when the transaction is submitted to the network. These LogicSig Args are _not_ part of the transaction ID nor of the TxGroup hash. They also cannot be read from other TEAL programs in the group of transactions.
```
algorandskiy, line 40: mb leave something like "efficiently leaving them not signed" for better clarity?

algorandskiy (approved):

PR #2939, opened 2021-09-22
Cache creators in cow base

tolikzinovyev:
## Summary

The evaluator shouldn't ask the ledger who is the creator of an asset multiple times. This PR adds a cache for ledger's responses in cow base.

## Test Plan

Added a test. Is there a better way to write a mock test?

### DIFF
--- a/ledger/eval.go
+++ b/ledger/eval.go
@@ -54,6 +54,17 @@ const maxPaysetHint = 20000
 // transaction group.
 const asyncAccountLoadingThreadCount = 4
 
+type creatable struct {
+	cindex basics.CreatableIndex
+	ctype  basics.CreatableType
+}
+
+// FoundAddress is a wrapper for an address and a boolean.
+type FoundAddress struct {
+	Address basics.Address
+	Exists  bool
+}
+
 type roundCowBase struct {
 	l ledgerForCowBase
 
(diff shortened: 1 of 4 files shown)


tsachiherman (commented):
looks good to me.

ledger/eval.go
```diff
@@ -54,6 +54,17 @@ const maxPaysetHint = 20000
 // transaction group.
 const asyncAccountLoadingThreadCount = 4
 
+type creatable struct {
+	cindex basics.CreatableIndex
+	ctype  basics.CreatableType
+}
+
+// FoundAddress is a wrapper for an address and a boolean.
+type FoundAddress struct {
```
algorandskiy, line 63: why is it exported? The struct name sound a bit weird. Maybe `cachedCreator` or something?

algorandskiy (commented):
Looks good, only concerning on `FoundAddress` type name

ledger/eval.go
tolikzinovyev: `FoundAddress` is also used in `evalIndexer.go` where this type needs to be exposed to Indexer. I think this name is ok for caching as well. `FoundAddress` encodes an address that is found (or not found). That is exactly what we want to cache.

tsachiherman (commented):
When do we need to cache non-existent addresses ? It looks like the `Exists` it's a mutually exclusive with the `Address`.

tolikzinovyev:
Maybe caching responses for non-existent assets is silly, but it lets indexer preload such responses in a batch.

tsachiherman (approved):
I think we can do better here on both the struct naming as well on the cached content, but I can't see a (good) reason to block this one right away. We could iterate on the above in the future.

PR #2940, opened 2021-09-22
testing: fix TestLatestSigsFromThisNode

algonautshant:
## Summary
Timing issues between writing to and reading from the database cause the test to occasionally fail.
Make the test more tolerant to such timing issues by polling.

## Test Plan
This is a fix for a test.

### DIFF
--- a/compactcert/worker_test.go
+++ b/compactcert/worker_test.go
@@ -385,9 +385,17 @@ func TestLatestSigsFromThisNode(t *testing.T) {
 	// Wait for a compact cert to be formed, so we know the signer thread is caught up.
 	_ = <-s.txmsg
 
-	latestSigs, err := w.LatestSigsFromThisNode()
-	require.NoError(t, err)
-	require.Equal(t, len(latestSigs), len(keys))
+	var latestSigs map[basics.Address]basics.Round
+	var err error
+	for x := 0; x < 100; x++ {
+		latestSigs, err = w.LatestSigsFromThisNode()
+		require.NoError(t, err)
+		if len(latestSigs) == 10 {
+			break
+		}
+		time.Sleep(100 * time.Millisecond)
+	}
+	require.Equal(t, len(keys), len(latestSigs))
 	for _, k := range keys {
 		require.Equal(t, latestSigs[k.Parent], basics.Round(2*proto.CompactCertRounds))
 	}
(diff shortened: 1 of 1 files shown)


compactcert/worker_test.go
```diff
@@ -385,9 +385,17 @@ func TestLatestSigsFromThisNode(t *testing.T) {
...
 
-	latestSigs, err := w.LatestSigsFromThisNode()
-	require.NoError(t, err)
-	require.Equal(t, len(latestSigs), len(keys))
+	var latestSigs map[basics.Address]basics.Round
+	var err error
+	for x := 0; x < 100; x++ {
+		latestSigs, err = w.LatestSigsFromThisNode()
+		require.NoError(t, err)
+		if len(latestSigs) == 10 {
```
tsachiherman, line 393: replace the `10` with ` len(keys)`

compactcert/worker_test.go
```diff
@@ -385,9 +385,17 @@ func TestLatestSigsFromThisNode(t *testing.T) {
 	// Wait for a compact cert to be formed, so we know the signer thread is caught up.
 	_ = <-s.txmsg
 
-	latestSigs, err := w.LatestSigsFromThisNode()
-	require.NoError(t, err)
-	require.Equal(t, len(latestSigs), len(keys))
+	var latestSigs map[basics.Address]basics.Round
+	var err error
+	for x := 0; x < 100; x++ {
```
tsachiherman, line 390: change the loop so that x = 0 .. 10, and update the sleep time so that it would sleep 2 ^ x millisecond.

tsachiherman (changes requested):
small request, but looks good otherwise.

compactcert/worker_test.go
```diff
@@ -385,9 +385,17 @@ func TestLatestSigsFromThisNode(t *testing.T) {
...
-	require.Equal(t, len(latestSigs), len(keys))
+	var latestSigs map[basics.Address]basics.Round
+	var err error
+	for x := 0; x < 10; x++ {
+		latestSigs, err = w.LatestSigsFromThisNode()
+		require.NoError(t, err)
+		if len(latestSigs) == len(keys) {
+			break
+		}
+		time.Sleep(256 * time.Millisecond)
```
tsachiherman, line 396: I would rather seeing this as an exponential backoff, but it's functionally correct as is.

tsachiherman (approved):

PR #2941, opened 2021-09-23
ledger: update EvalForIndexer to avoid totals computation

tsachiherman:
## Summary
In the EvalForIndexer, we don't want to call finalValidation(). This would
skip the calculation of the account totals in the state delta, which is a serious
issue if it were to be used by algod, but it's perfectly fine for the indexer since
it doesn't track any totals and therefore cannot calculate the new totals.

## Test Plan

Test against the indexer.


tolikzinovyev (approved):
Thanks!

PR #2942, opened 2021-09-23
Fetch creators in a batch in `EvalForIndexer()`

tolikzinovyev:
## Summary

With this change, Indexer is able to preload creators of most assets and apps in a batch.

## Test Plan

Indexer import TPS goes up 10% when used with Amazon Aurora.

### DIFF
--- a/ledger/evalIndexer.go
+++ b/ledger/evalIndexer.go
@@ -26,6 +26,7 @@ import (
 	"github.com/algorand/go-algorand/data/bookkeeping"
 	"github.com/algorand/go-algorand/data/transactions"
 	"github.com/algorand/go-algorand/ledger/ledgercore"
+	"github.com/algorand/go-algorand/protocol"
 )
 
 // A ledger interface that Indexer implements. This is a simplified version of the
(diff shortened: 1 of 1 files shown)


tsachiherman (changes requested):
I think that the functionality here isn't incorrect - but rather misplaced. This optimization is indexer specific, and as such, should live within the indexer code base. The evelIndexer should contain the interfacing between the two, and be limited to that.

If needed, extending the `indexerLedgerForEval` does make sense here.

tolikzinovyev:
This is what we did with accounts, and I think this is the right place to preload things since knowing what to preload is highly specific to the evaluator. Also, having two caches (in cow base and indexer) storing the same data isn't pretty. How would I extend `indexerLedgerForEval`?

tsachiherman:
no, I was not suggesting having another cache in the indexer repo. I was saying the the indexer repo should provide a method that would analyze a block, and return a pre-init set of values for the the two caches. The `EvalForIndexer` should take these arrays/maps and populate the cowbase with them.

tolikzinovyev:
I think this will be convoluted. Eval should know what to preload, so why wouldn't it just do that?

tsachiherman:
If this was a shared functionality, then I would agree with you. However, since the indexer is the only consumer of this preloading, I think that it would be better for the indexer to figure out what need to be preloaded, load that, and then provide that information to the evaluator.

tolikzinovyev:
Are you proposing this interaction? Indexer's `AddBlock()` -> go-algorand's `FindWhatToPreload()` -> preload in indexer -> go-algorand's `EvalIndexer()` -> indexer writes the state delta. Seems too much going back and forth. The alternative is to have `FindWhatToPreload()` in Indexer, but then this eval-specific knowledge is far from eval.

tsachiherman:
> go-algorand's `FindWhatToPreload()`

why ? I was thinking of:
- AddBlock() < indexer >
- PreloadResources < indexer >
   - FindWhatToPreload
   - LoadResources
- EvalIndexer(..., <resources>)

tolikzinovyev:
So the second option then. I guess it's ok, it's just now the accounting logic of `FindWhatToPreload()` is not part of eval.

PR #2945, opened 2021-09-24
Add application account into a dryrun req created by goal

algorandskiy:
## Summary

* AVM 1.0 introduces application accounts
* `goal clerk` needs to consider this account when creates dryrun request object
* Added a new option `--dryrun-accounts` for adding accounts that were rekeyed to an app account (or any other accounts).

## Test Plan

Tested manually. I'll add example from Jason W into sub-e2e tests in subseq PRs (need inner txns for indexer as well)

### DIFF
--- a/libgoal/libgoal.go
+++ b/libgoal/libgoal.go
@@ -970,18 +970,18 @@ func (c *Client) Catchup(catchpointLabel string) error {
 const defaultAppIdx = 1380011588
 
 // MakeDryrunStateBytes function creates DryrunRequest data structure in serialized form according to the format
-func MakeDryrunStateBytes(client Client, txnOrStxn interface{}, other []transactions.SignedTxn, proto string, format string) (result []byte, err error) {
+func MakeDryrunStateBytes(client Client, txnOrStxn interface{}, otherTxns []transactions.SignedTxn, otherAccts []basics.Address, proto string, format string) (result []byte, err error) {
 	switch format {
 	case "json":
 		var gdr generatedV2.DryrunRequest
-		gdr, err = MakeDryrunStateGenerated(client, txnOrStxn, other, proto)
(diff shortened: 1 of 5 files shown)


libgoal/libgoal.go
```diff
@@ -1049,18 +1052,18 @@ func MakeDryrunStateGenerated(client Client, txnOrStxn interface{}, other []tran
...
 					Params: appParams,
 				})
 			}
 
-			accounts := append(tx.Accounts, tx.Sender)
 			for _, acc := range accounts {
 				var info generatedV2.Account
 				if info, err = client.AccountInformationV2(acc.String()); err != nil {
-					return
+					// ignore error and continue
```
jasonpaulos, line 1066: Should this actually `continue` so that the empty account info isn't appended to `dr.Accounts`?

libgoal/libgoal.go
```diff
@@ -1023,6 +1023,9 @@ func MakeDryrunStateGenerated(client Client, txnOrStxn interface{}, other []tran
 	for _, txn := range txns {
 		tx := txn.Txn
 		if tx.Type == protocol.ApplicationCallTx {
+			accounts := append(tx.Accounts, tx.Sender)
+			accounts = append(accounts, otherAccts...)
+
 			apps := []basics.AppIndex{tx.ApplicationID}
 			apps = append(apps, tx.ForeignApps...)
```
jasonpaulos, line 1030: Not directly relevant to this change, but should this code also collect all foreign assets and look up the account balances of their creator to add to `dr.Accounts`?

jasonpaulos (commented):
Mostly looks good, but I have some comments.

JasonWeathersby (approved):

jasonpaulos (approved):

PR #2948, opened 2021-09-24
Adding clean target to ci-build target

algojack:
While building with Brice, we noticed that `make ci-build` doesn't clean automatically, and we probably want that by default. So we decided to add it.

Are there any times we don't want clean included?


scripts/release/mule/Makefile.mule
```diff
@@ -4,7 +4,7 @@ PKG_DIR = $(SRCPATH)/tmp/node_pkgs/$(OS_TYPE)/$(ARCH)
 
 .PHONY: ci-clean ci-setup ci-build
 
-ci-clean:
+ci-clean: clean
```
algobarb, line 7: What does clean do? Is it like `cleanWs()` which cleans the Workspace?

algojack: @bricerisingalgorand a little help? I think it deletes temp files, but i'm not 100% sure.

bricerisingalgorand: https://github.com/algorand/go-algorand/blob/master/Makefile#L252

bricerisingalgorand: That tries to delete all build/generated files in the current repo

algobarb (approved):

egieseke (approved):
Good catch.

PR #2952, opened 2021-09-24
Add `resources` argument in `EvalForIndexer()`

tolikzinovyev:
## Summary

This way Indexer can preload most things from the database efficiently.


tsachiherman (changes requested):
the change looks fine - but I'd like to request adding some testing for the `saveResourcesInCowBase` method. Testing it individually or via the `EvalForIndexer` wrapper is up to you.
i.e. the goal is to prove that the roundCowBase here would truly initialize its state by the method. The standard cowRoundBase interface should be used to validate the correctness rather than direct access to the underlying maps.

tolikzinovyev:
I added a test. Is this what you meant?

PR #2954, opened 2021-09-25
Enable cost and fee pooling in tealdbg

algorandskiy:
## Summary

Enable cost and fee pooling in tealdbg

## Test Plan

Tests added

### DIFF
--- a/cmd/tealdbg/local.go
+++ b/cmd/tealdbg/local.go
@@ -524,20 +524,38 @@ func (r *LocalRunner) RunAll() error {
 
 	failed := 0
 	start := time.Now()
+	pooledApplicationBudget := uint64(0)
+	credit, _ := transactions.FeeCredit(r.txnGroup, r.proto.MinTxnFee)
+	// ignore error since fees are not important for debugging in most cases
+
+	evalParams := make([]logic.EvalParams, 0, len(r.runs))
 	for _, run := range r.runs {
+		if run.mode == modeStateful {
+			if r.proto.EnableAppCostPooling {
+				pooledApplicationBudget += uint64(r.proto.MaxAppProgramCost)
+			} else {
+				pooledApplicationBudget = uint64(r.proto.MaxAppProgramCost)
+			}
+		}
+		ep := logic.EvalParams{
+			Proto:                   &r.proto,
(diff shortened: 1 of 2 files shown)


cmd/tealdbg/local.go
```diff
@@ -524,20 +524,38 @@ func (r *LocalRunner) RunAll() error {
 
 	failed := 0
 	start := time.Now()
-	for _, run := range r.runs {
-		r.debugger.SaveProgram(run.name, run.program, run.source, run.offsetToLine, run.states)
+	pooledApplicationBudget := uint64(0)
+	credit, _ := transactions.FeeCredit(r.txnGroup, r.proto.MinTxnFee)
+	// ignore error since fees are not important for debugging in most cases
 
+	evalParams := make([]logic.EvalParams, 0, len(r.runs))
```
jannotti, line 531: Maybe make it full size from the start, and put them in at index, instead of using append.

I know there's no efficiency gain here, since you allocated len(r.runs).  But this way we always prefer the style of creating full sized array and putting things in, rather than using append.  Makes it less likely the code will be copied to a place that does the inefficient grow thing.

cmd/tealdbg/local.go
```diff
@@ -524,20 +524,38 @@ func (r *LocalRunner) RunAll() error {
...
+	for _, run := range r.runs {
+		if run.mode == modeStateful {
+			if r.proto.EnableAppCostPooling {
+				pooledApplicationBudget += uint64(r.proto.MaxAppProgramCost)
+			} else {
+				pooledApplicationBudget = uint64(r.proto.MaxAppProgramCost)
```
jannotti, line 537: Is this used at all when EnableAppCostPooling is false?

jannotti (approved):

cmd/tealdbg/local.go
algorandskiy: done

cmd/tealdbg/local.go
algorandskiy: Yes, `logic.eval` will fail the program, checked manually.

jannotti (approved):

PR #2955, opened 2021-09-26
[WIP] Add abstraction for Participation database

Aharonee:
and merklekeystore persistence mechanism


crypto/merklekeystore/persistent_keystore.go
```diff
@@ -0,0 +1,177 @@
...
+		return MigrateDB(tx)
+	})
+	if err != nil {
+		return PersistentKeystore{}, err
+	}
+
+	return PersistentKeystore{store}, nil
+}
+
+func MigrateDB(tx *sql.Tx) error {
```
algoidan, line 149: think we should start with the transaction creation

crypto/merklekeystore/persistent_keystore.go
```diff
@@ -0,0 +1,177 @@
...
+	var exists bool
+	schemaQuery := fmt.Sprintf("SELECT version FROM schema WHERE tablename = ?", keystoreTableSchemaName)
+	tx.QueryRow("SELECT EXISTS(%s)", schemaQuery).Scan(&exists)
```
algoidan, line 152: do we need to check for errors on Scan return?

crypto/merklekeystore/persistent_keystore.go
```diff
@@ -0,0 +1,177 @@
...
+	if !exists {
+		// In the future this should not return quietly, as blockproof keys will be required
+		return nil
+	}
+
+	var version int
+	row := tx.QueryRow(schemaQuery)
+	err := row.Scan(&version)
+	if err != nil {
+		// fmt.Errorf("restore (agreement): could not query raw state: %v", err)
```
algoidan, line 162: why is this in a comment?

crypto/merklekeystore/persistent_keystore.go
```diff
@@ -0,0 +1,177 @@
...
+
+	err = updateDB(tx, version)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func updateDB(tx *sql.Tx, version int) error {
```
algoidan, line 174: remove this

crypto/merklekeystore/keystore.go
```diff
@@ -135,22 +149,35 @@ func New(firstValid, lastValid, interval uint64, sigAlgoType crypto.AlgorithmTyp
...
 }
 
+func (s *Signer) Persist() error {
+	err := s.keyStore.Persist(s.signatureAlgorithms, s.FirstValid, s.lastValid, s.Interval)
+	if err != nil {
+		return err
+	}
+
+	// Let the garbage collector remove these from memory.Should we delete more explicitly?
+	s.signatureAlgorithms = nil
```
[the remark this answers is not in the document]
Aharonee: The alternative is to return (*Signer, *SignatureAlgorithm[], error) from New() and then pass the keys to Persist.

crypto/merklekeystore/keystore.go
```diff
@@ -224,40 +254,39 @@ func (s *Signer) Trim(before uint64) (*Signer, error) {
...
-		Interval:            s.Interval,
-		ArrayBase:           s.ArrayBase,
-
-		Tree: s.Tree,
-		mu:   deadlock.RWMutex{},
-	}
-
-	copy(signerCopy.SignatureAlgorithms, s.SignatureAlgorithms)
-	return &signerCopy
+//func (s *Signer) copy() *Signer {
```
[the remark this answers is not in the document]
Aharonee: Wasn't sure if it's needed so didn't remove it yet, can't think of a reason why we would need it though

crypto/merklekeystore/persistent_keystore.go
Aharonee: What do you mean? The DB table creation?

crypto/merklekeystore/persistent_keystore.go
Aharonee: I don't think so since we're already checking exists value which defaults to false.
I can add an error check just in case though

crypto/merklekeystore/persistent_keystore.go
Aharonee: This is where the actual migrations will be implemented, it is currently empty since there are no versions for this table.
This pattern is similar to the one in `partInstall.go`

crypto/merklekeystore/persistent_keystore.go
```diff
@@ -0,0 +1,162 @@
...
+	err := p.store.Atomic(func(ctx context.Context, tx *sql.Tx) error { // does it take too long?
+		err := keystoreInstallDatabase(tx) // assumes schema table already exists (created by partInstallDatabase)
+		if err != nil {
+			return err
+		}
+
+		// Should we count the number of rows affected?
+		var round uint64
+		for round = firstValid; round%interval != 0; round++ {} // Increase round to a multiplier of interval
+		for i, key := range keys {
```
algoidan, line 51: we could just use index_to_round.

crypto/merklekeystore/keystore_test.go
```diff
@@ -17,176 +17,185 @@
...
+		signer := generateTestSigner(i, i+1, 1, a)
+		defer signer.keyStore.store.Close()
 		_, err = signer.Sign(h, i)
 		a.NoError(err)
 	}
 
-	signer, err := New(0, 0, 1, crypto.Ed25519Type)
-	a.NoError(err)
-	a.Equal(0, len(signer.SignatureAlgorithms))
+	signer := generateTestSigner(0, 0, 1, a)
```
algoidan, line 47: maybe we should put this "test_single_limit" in a function. and there will be much less defers

crypto/merklekeystore/keystore_test.go
```diff
@@ -17,176 +17,185 @@
...
 
 	_, err = signer.Sign(h, 8)
 	a.Error(err)
 
 	_, err = signer.Sign(h, 9)
 	a.Error(err)
 
-	_, err = signer.Trim(10)
-	a.Error(err)
+	// should Trim return error when trying to trim above its last valid? or just drop what relevant keys he finds as it is currently?
```
algoidan, line 120: I think we should. consider this, Can you think of a normal flow in the code that we will want to remove more than what we have?

crypto/merklekeystore/keystore_test.go
```diff
@@ -17,176 +17,185 @@
...
-		a.Error(err, i)
-	}
-
-	for index, round := range indices {
-		pos, err := signer.getArrayIndex(round)
-		a.NoError(err)
-		a.Equal(uint64(index), pos)
-	}
-}
+//func TestDisposableKeyPositions(t *testing.T) {
```
algoidan, line 125: we should keep those tests.

algoidan, line 125: we should keep those tests.

crypto/merklekeystore/keystore.go
```diff
@@ -113,7 +127,7 @@ func (s *Signer) Marshal(pos uint64) ([]byte, error) {
 // New Generates a merklekeystore.Signer
 // The function allow creation of empty signers, i.e signers without any key to sign with.
 // keys can be created between [A,Z], if A == 0, keys created will be in the range (0,Z]
-func New(firstValid, lastValid, interval uint64, sigAlgoType crypto.AlgorithmType) (*Signer, error) {
+func New(firstValid, lastValid, interval uint64, sigAlgoType crypto.AlgorithmType, store db.Accessor) (*Signer, error) {
```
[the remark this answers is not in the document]
Aharonee: The assumptions are only before calling Persist actually

crypto/merklekeystore/keystore_test.go
Aharonee: Keys are no longer stored by index and offset by arrayBase so what makes this test useful?

crypto/merklekeystore/keystore_test.go
Aharonee: Which part exactly are you talking about?

crypto/merklekeystore/persistent_keystore.go
```diff
@@ -0,0 +1,168 @@
...
+		return nil
+	}
+
+	err := p.store.Atomic(func(ctx context.Context, tx *sql.Tx) error { // does it take too long?
+		err := keystoreInstallDatabase(tx) // assumes schema table already exists (created by partInstallDatabase)
+		if err != nil {
+			return err
+		}
+
+		// Should we count the number of rows affected?
```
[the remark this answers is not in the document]
Aharonee: It should return len(keys) actually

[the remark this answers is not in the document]
Aharonee: I think if it doesn't return error then necessarily each key has been accounted for (since we iterate over all of them).

crypto/merklekeystore/keystore.go
```diff
@@ -190,74 +212,28 @@ func (s *Signer) Sign(hashable crypto.Hashable, round uint64) (Signature, error)
...
-
-func (s *Signer) dropKeys(upTo int) {
-	for i := 0; i <= upTo; i++ {
-		// zero the keys.
-		s.SignatureAlgorithms[i] = crypto.SignatureAlgorithm{}
-	}
-	s.SignatureAlgorithms = s.SignatureAlgorithms[upTo+1:]
+// Restore loads Signer from given database, as well as restoring PersistenKeystore (where the actual keys are stored)
+func (s *Signer) Restore(store db.Accessor) error {
+	keystore, err := RestoreKeystore(store)
```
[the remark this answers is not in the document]
Aharonee: I'm returning it either way though

crypto/merklekeystore/keystore.go
```diff
@@ -135,22 +145,34 @@ func New(firstValid, lastValid, interval uint64, sigAlgoType crypto.AlgorithmTyp
...
 	}
+
 	s := &Signer{
-		SignatureAlgorithms: keys,
+		keyStore:            PersistentKeystore{store},
+		signatureAlgorithms: keys,
 		FirstValid:          firstValid,
-		ArrayBase:           0,
 		Interval:            interval,
 		mu:                  deadlock.RWMutex{},
```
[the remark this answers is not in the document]
Aharonee: I'm leaving it for now since I can't be sure yet, need to think about it some more. If someone Trims in the middle of a Sign operation some problems may arise, but from a quick glance at the logic I don't really see how that could happen. The first operation in both Trim and Sign is basically an atomic DB operation.

crypto/merklekeystore/keystore.go
```diff
@@ -135,22 +145,34 @@ func New(firstValid, lastValid, interval uint64, sigAlgoType crypto.AlgorithmTyp
...
 }
 
+// Persist dumps the keys into the database and deletes the reference to them in Signer
+func (s *Signer) Persist() error {
+	err := s.keyStore.Persist(s.signatureAlgorithms, s.FirstValid, s.Interval)
+	if err != nil {
+		return err
+	}
+
+	// Let the garbage collector remove these from memory.Should we delete more explicitly?
```
algoidan, line 171: don't think we should do anything more

crypto/merklekeystore/keystore.go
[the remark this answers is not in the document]
algoidan: don't think we should return the keystore struct

crypto/merklekeystore/keystore.go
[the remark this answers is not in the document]
Aharonee: But it doesn't have to be created before calling New, only before Persist

crypto/merklekeystore/keystore_test.go
Aharonee: I guess if the dilithium keys were not refreshed as quick as they should and we're past the last valid round.
Then the old keys deletion thread might just delete all keys up to that round

crypto/merklekeystore/keystore_test.go
```diff
@@ -17,176 +17,157 @@
...
-	_, err = New(1, 0, 1, crypto.Ed25519Type)
-	a.Error(err)
+	testSignerNumKeysLimits(0, 0, 1, 0)
+	testSignerNumKeysLimits(0, 1, 1, 1)
+	testSignerNumKeysLimits(2, 2, 2, 1)
+	testSignerNumKeysLimits(8, 21, 10, 2)
+	testSignerNumKeysLimits(8, 20, 10, 2)
+	testSignerNumKeysLimits(10, 21, 10, 2)
+	testSignerNumKeysLimits(10, 20, 10, 2)
+	testSignerNumKeysLimits(11, 20, 10, 1)
```
algoidan, line 60: dont we want this test as well:
	_, err = New(1, 0, 1, crypto.Ed25519Type)
	a.Error(err)

crypto/merklekeystore/keystore_test.go
```diff
@@ -17,176 +17,157 @@
...
 	s := crypto.SignatureAlgorithm{}
-	for _, key := range signer.SignatureAlgorithms {
+	for i := uint64(1); i <= 100; i++ {
+		key, err := signer.keyStore.GetKey(i)
+		a.NoError(err)
 		a.NotEqual(s, key)
 	}
 }
 
 func TestSignatureStructure(t *testing.T) {
```
algoidan, line 159: we might want to add a ED25519 test.

crypto/merklekeystore/keystore_test.go
```diff
@@ -318,105 +312,101 @@ func TestIncorrectByteSignature(t *testing.T) {
...
 
 	// create signer and delete all keys.
-	signer, err = New(1, 60, 1, crypto.Ed25519Type)
-	a.NoError(err)
+	signer = generateTestSigner(1, 60, 1, a)
+	defer signer.keyStore.store.Close()
 	_, err = signer.Trim(60)
 	a.NoError(err)
-	a.Equal(0, len(signer.SignatureAlgorithms))
+	a.Equal(0, length(signer, a))
```
algoidan, line 394: add the same with Trim 61

crypto/merklekeystore/persistentKeystore.go
```diff
@@ -0,0 +1,147 @@
...
+	}
+
+	err := p.store.Atomic(func(ctx context.Context, tx *sql.Tx) error { // does it take too long?
+		err := keystoreInstallDatabase(tx) // assumes schema table already exists (created by partInstallDatabase)
+		if err != nil {
+			return err
+		}
+
+		round := indexToRound(firstValid, interval, 0)
+		for i, key := range keys {
```
algoidan, line 49: lets check if we can lock the db for less

crypto/merklekeystore/persistentKeystore.go
```diff
@@ -0,0 +1,147 @@
...
+func keystoreInstallDatabase(tx *sql.Tx) error {
+	_, err := tx.Exec(`CREATE TABLE BlockProofKeys (
+    	id	  INTEGER PRIMARY KEY, 
+    	round INTEGER,	    --*  committed round for this key
+		key   BLOB  --*  msgpack encoding of ParticipationAccount.BlockProof.SignatureAlgorithm
+		);`)
+	if err != nil {
+		return err
+	}
+	// TODO: CREATE INDEX for round field in order to improve performance. Benchmark with and without
```
algoidan, line 113: we have this comment already

crypto/merklekeystore/persistentKeystore_test.go
```diff
@@ -0,0 +1,80 @@
...
+	s := generateTestSigner(25, 1023, 23, a)
+	k := s.keyStore
+	defer k.store.Close()
+
+	s2 := &Signer{}
+	a.NoError(s2.Restore(k.store))
+	a.Equal(countKeysInRange(25, 1023, 23), length(s2, a))
+}
+
+// Should I move this to posdivs.go?
```
algoidan, line 74: i think it is okay like this

ledger/apply/keyreg_test.go
```diff
@@ -202,8 +204,14 @@ func TestBlockProofPKKeyReg(t *testing.T) {
 }
 
 func createTestTxn(t *testing.T, src basics.Address, secretParticipation *crypto.SignatureSecrets, vrfSecrets *crypto.VRFSecrets) transactions.Transaction {
-	signer, err := merklekeystore.New(0, 0, 1, crypto.DilithiumType)
+	//signer, err := merklekeystore.New(0, 0, 1, crypto.DilithiumType)
```
algoidan, line 207: remove this

crypto/merklekeystore/persistentKeystore.go
Aharonee: where?

PR #2956, opened 2021-09-27
adding appaddr field to Application schema

barnjamin:
## Summary

Requesting the application information from the Algod API should return the address associated with the application.  

Added the address field to the top level Application schema, ran `make generate`, added address field to struct generated struct initialization function.

Is there more that needs to be done for this?
 
## Test Plan

Looking for guidance on where you'd want to see tests, didn't see any existing relevant tests


jannotti:
I don't see why it should do this. The app address is determined by the id. So if you make a request for app info, you already know the app address.

barnjamin:
>I don't see why it should do this. The app address is determined by the id. So if you make a request for app info, you already know the app address.

That's fair, but I guess I'd expect that there are obvious SDK methods to get the app addr from the app id

jannotti:
Yes, we definitely need those.  But they can be like python's
logic.address(), which just performs the hash.  I believe we have issues
(one for each sdk) to add them.

PR #2975, opened 2021-09-30
Extended logic eval error

algorandskiy:
## Summary

Return a bit more details on eval failure for app calls.

## Test Plan

Added `EvalContext.PcDetails()` method and tests.

### DIFF
--- a/data/transactions/logic/eval.go
+++ b/data/transactions/logic/eval.go
@@ -3938,3 +3938,29 @@ func opTxSubmit(cx *EvalContext) {
 	})
 	cx.subtxn = nil
 }
+
+// PcDetails return PC and disassembled instructions at PC up to 2 opcodes back
+func (cx *EvalContext) PcDetails() (pc int, dis string) {
+	const maxNumAdditionalOpcodes = 2
+	text, ds, err := disassembleInstrumented(cx.program, nil)
+	if err == nil {
+		for i := 0; i < len(ds.pcOffset); i++ {
+			if ds.pcOffset[i].PC == cx.pc {
+				start := 0
+				if i >= maxNumAdditionalOpcodes {
+					start = i - maxNumAdditionalOpcodes
+				}
+
+				startTextPos := ds.pcOffset[start].Offset
+				endTextPos := len(text)
+				if i+1 < len(ds.pcOffset) {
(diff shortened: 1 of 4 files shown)


jannotti (changes requested):
I think that the pc is set to the end or maybe even 1 past the end for errors that don't get detected until after the first pass (when jumps are resolved).  Could you put in a test to show things work ok when PcDetails is called then?

jannotti (approved):
Seems pretty clear you would just get an empty string, so that seems ok.

ledger/appcow.go
```diff
@@ -490,7 +490,9 @@ func (cb *roundCowState) StatefulEval(params logic.EvalParams, aidx basics.AppIn
 	var cx *logic.EvalContext
 	pass, cx, err = logic.EvalStatefulCx(program, params)
 	if err != nil {
-		return false, transactions.EvalDelta{}, ledgercore.LogicEvalError{Err: err}
+		pc, det := cx.PcDetails()
```
tsachiherman, line 493: This is fine as long as cx is non-nil. The default go pattern is that whenever err is non nil, all the other params are..
So, adding a nil check before working on cx wouldn't be a bad idea.

data/transactions/logic/eval.go
```diff
@@ -3938,3 +3938,29 @@ func opTxSubmit(cx *EvalContext) {
 	})
 	cx.subtxn = nil
 }
+
+// PcDetails return PC and disassembled instructions at PC up to 2 opcodes back
+func (cx *EvalContext) PcDetails() (pc int, dis string) {
+	const maxNumAdditionalOpcodes = 2
+	text, ds, err := disassembleInstrumented(cx.program, nil)
+	if err == nil {
```
tsachiherman, line 3946: Nit: maybe an early return here ?

ledger/appcow.go
algorandskiy: right, added the check.

data/transactions/logic/eval.go
algorandskiy: done

algorandskiy:
> Could you put in a test to show things work ok when PcDetails is called then?
added a test

tsachiherman (approved):
looks good.

Issue #2976, opened 2021-09-30
Update go-algorand internals to use Algod V2 API (No more V1 dependencies)


barnjamin:
## Problem

While digging into #2904 what it would take to support a remote API for `goal`, it was noted that some of the API calls currently only support the V1 API.

Since some public API providers no longer support the V1 API, in order to allow some remote endpoint for `goal` the API methods will need to be updated to allow V2 payloads.

## Solution

Add V2 support to goal/libgoal

## Urgency

Nice to have, may try to do it myself

## Acceptance Criteria

Research issue.

Create followup tickets that lists the specific work involved in fully removing the V1 API references.

algoanne:
@barnjamin I am labeling this "Team Scytale", and we will take a look through it during our grooming, but you are still more than welcome to work on it (I see you said you might). Until we put it in "In Progress", we are not yet actively working on it.

michaeldiamant:
Here's a sample `grep` to gauge v1 API usage within go-algorand:  

```
❯ grep -IlR "github.com/algorand/go-algorand/daemon/algod/api/spec/v1" .
./libgoal/transactions.go
./libgoal/libgoal.go
./cmd/algoh/deadman.go
./cmd/algoh/blockWatcher_test.go
./cmd/algoh/blockWatcher.go
./cmd/algoh/blockstats_test.go
./cmd/algoh/client.go
./cmd/algoh/blockstats.go
./cmd/algoh/mockClient.go
./cmd/goal/clerk.go
./test/framework/fixtures/restClientFixture.go
./test/e2e-go/features/stateproofs/stateproofs_test.go
./test/e2e-go/features/transactions/asset_test.go
./test/e2e-go/features/transactions/sendReceive_test.go
./test/e2e-go/restAPI/restClient_test.go
./shared/pingpong/accounts.go
./shared/pingpong/pingpong.go
./daemon/algod/api/algod.oas2.json
./daemon/algod/api/server/v1/handlers/responses.go
./daemon/algod/api/server/v1/handlers/handlers.go
./daemon/algod/api/swagger.json
./daemon/algod/api/algod.oas3.yml
./daemon/algod/api/client/restClient.go
```

algochoi:
RE: Version affinity discussed [in this comment](https://github.com/algorand/go-algorand/pull/2985#issuecomment-932912991)

[...]

PR #2985, opened 2021-10-02
starting to convert to v2 endpoints

barnjamin:
## Summary

As part of #2976 this pr starts to convert Algod API calls to the V2 API.  Haven't thoroughly checked that all endpoints are updated yet. 

So far pretty painless so I'm probably doing something wrong.

## Test Plan

tbd

### DIFF
--- a/cmd/goal/asset.go
+++ b/cmd/goal/asset.go
@@ -19,9 +19,11 @@ package main
 import (
 	"encoding/base64"
 	"fmt"
+	"math"
 
 	"github.com/spf13/cobra"
 
+	"github.com/algorand/go-algorand/daemon/algod/api/server/v2/generated"
 	"github.com/algorand/go-algorand/data/basics"
 	"github.com/algorand/go-algorand/libgoal"
 )
@@ -240,8 +242,8 @@ var createAssetCmd = &cobra.Command{
 				if err != nil {
 					reportErrorf(err.Error())
 				}
-				if txn.TransactionResults != nil && txn.TransactionResults.CreatedAssetIndex != 0 {
-					reportInfof("Created asset with asset index %d", txn.TransactionResults.CreatedAssetIndex)
+				if txn.AssetIndex != nil && *txn.AssetIndex != 0 {
+					reportInfof("Created asset with asset index %d", *txn.AssetIndex)
 				}
 			}
 		} else {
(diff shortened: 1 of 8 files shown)


cmd/goal/asset.go
```diff
@@ -576,17 +578,15 @@ var freezeAssetCmd = &cobra.Command{
...
 		return fmt.Sprintf("%d", amount)
 	}
 
 	// Otherwise, ensure there are decimals digits to the right of the decimal point
-	pow := uint64(1)
-	for i := uint32(0); i < decimals; i++ {
-		pow *= 10
-	}
+	pow := uint64(math.Pow(10, float64(decimals)))
+
```
jannotti, line 589: This makes me nervous.  `math.Pow(10, float64(decimals)` could be *slightly* less than the exact integer due to floating point representation, and then the uint64 cast will bring it down, not round.  I'd rather we keep the integral exponentiation.

nodecontrol/algodControl.go
```diff
@@ -86,7 +86,7 @@ func (nc NodeController) ServerURL() (url.URL, error) {
 	if err != nil {
 		return url.URL{}, err
 	}
-	return url.URL{Scheme: "http", Host: addr}, nil
+	return url.URL{Scheme: "https", Host: addr}, nil
```
jannotti, line 89: I don't think we can't do this.  People normally local algods over http.

Perhaps we can do something like checking if the file explicitly contains a protocol, if so, use it.  That way people who want to can put an entire `https://algoexplorer.io/v2" in `algod.net` Otherwise, it continues to default to http.

I am unsure if there is any other code that reads algod.net that would have to be similarly updated.

jannotti (commented):
I think this is looking good.  Are there many more to change?  I'd like to check in with someone who has been in the goal codebase more than me, to see if there's a reason we have stuck with v1.  Maybe @winder ?

nodecontrol/algodControl.go
barnjamin: Makes sense. Will take it out before submitting

barnjamin: Trying something else, just allowing specification of the full URI including protocol. Should still work for `localhost:8080` `127.0.0.1:8080` but should also work for `https://apiendpoint.com`

barnjamin:
From the `SuggestedParams` libgoal method I was led to the algod rest client where a number of the methods on the algod rest client have logic for version affinity but some don't 

https://github.com/algorand/go-algorand/blob/master/daemon/algod/api/client/restClient.go#L493-L505
vs
https://github.com/algorand/go-algorand/blob/master/daemon/algod/api/client/restClient.go#L441-L445

Considering adding the version switching/conversion logic to those but it may make this PR thicc if there are a number of downstream dependencies

PR #2988, opened 2021-10-03
Algod clientv2

barnjamin:
## Summary

Draft of Algod rest client changes needed for  #2985 

Adding additional methods with explicit version seems safer than switching on affinity and converting old to new.  It still feels weird given some methods do the switching and these wont. 

A more comprehensive refactor might expose the un-versioned method names with some dispatching to the correct version and do switching and conversion with errors reported for unsupported endpoints

## Test Plan

tbd


barnjamin:
@jannotti @winder looking for some feedback on this, would you prefer the version switching inside some unsuffixed methods or is this ok?

PR #2990, opened 2021-10-04
[WIP] State proof integration

algoidan:


cmd/goal/account.go
```diff
@@ -1313,6 +1316,9 @@ var partkeyInfoCmd = &cobra.Command{
 					SelectionID:     part.VRFSecrets().PK,
 					VoteKeyDilution: part.KeyDilution,
 				}
+				if certSigner := part.BlockProofSigner(); certSigner != nil {
```
algoidan, line 1319: bloockproof will be renamed to stateproof

crypto/hashes.go
```diff
@@ -0,0 +1,91 @@
...
+	"errors"
+	"hash"
+
+	"github.com/algonathan/sumhash"
+	"golang.org/x/crypto/sha3"
+)
+
+var sumhashCompressor sumhash.LookupTable
+
+func init() {
```
algoidan, line 30: This code will be moved to the sumhashlib.

crypto/sig_abstractions.go
```diff
@@ -0,0 +1,166 @@
...
+	Pack PackedVerifyingKey `codec:"pks"`
+}
+
+// ToBeHashed makes it easier to hash the VeryfyingKey struct.
+func (z *VerifyingKey) ToBeHashed() (protocol.HashID, []byte) {
+	return protocol.VerifyingKey, protocol.Encode(z)
+}
+
+// GetSigner fetches the Signer type that is stored inside this SignatureAlgorithm.
+func (z *SignatureAlgorithm) GetSigner() (Signer, error) {
```
algoidan, line 90: we going to create a change in msgpack in which there will be some callback to isValid. after this change, there won't be an invalid type.

crypto/merklekeystore/keystore.go
```diff
@@ -0,0 +1,295 @@
...
+	if pos >= uint64(len(s.SignatureAlgorithms)) {
+		return 0, errOutOfBounds
+	}
+	return pos, nil
+}
+
+// Trim shortness deletes keys that existed before a specific round,
+// will return an error for non existing keys/ out of bounds keys.
+// the output is a copy of the signer - which can be persisted.
+func (s *Signer) Trim(before uint64) (*Signer, error) {
```
shiqizng, line 215: maybe name it round instead of before would be better

crypto/sig_abstractions.go
```diff
@@ -0,0 +1,166 @@
...
+		return &p.Ed25519Singer, nil
+	default:
+		return nil, errUnknownSigner
+	}
+}
+
+var errNonExistingSignatureAlgorithmType = errors.New("signing algorithm type does not exist")
+
+// NewSigner receives a type of signing algorithm and generates keys.
+func NewSigner(t AlgorithmType) (*SignatureAlgorithm, error) {
```
algonautshant, line 144: Missing unit test for this.

ledger/eval.go
```diff
@@ -1089,7 +1089,7 @@ func (eval *BlockEvaluator) finalValidation() error {
 		if err != nil {
 			return err
 		}
-		if eval.block.CompactCert[protocol.CompactCertBasic].CompactCertVoters != expectedVoters {
+		if !eval.block.CompactCert[protocol.CompactCertBasic].CompactCertVoters.IsEqual(expectedVoters) {
```
algonautshant, line 1092: Will be nice to have a unit test for the not equal case here.

ledger/voters.go
```diff
@@ -236,20 +236,15 @@ func (tr *VotersForRound) loadTree(l ledgerForTracker, au *accountUpdates, hdr b
...
-		keyDilution := acct.VoteKeyDilution
-		if keyDilution == 0 {
-			keyDilution = tr.Proto.DefaultKeyDilution
-		}
-
 		participants[i] = compactcert.Participant{
-			PK:          acct.VoteID,
-			Weight:      money.ToUint64(),
-			KeyDilution: keyDilution,
+			PK:         acct.BlockProofID,
```
algonautshant, line 240: Unit test is missing here. 
Will be nice to have a unit test for these changes.

crypto/merklekeystore/keystore.go
```diff
@@ -0,0 +1,295 @@
...
+		return errReceivedRoundIsBeforeFirst
+	}
+
+	ephkey := CommittablePublicKey{
+		VerifyingKey: sig.VerifyingKey,
+		Round:        round,
+	}
+
+	pos := roundToIndex(firstValid, round, interval)
+	isInTree := merklearray.Verify(
```
shiqizng, line 281: please change this variable name to something like err since the return type for Verify is error.

data/account/participation.go
```diff
@@ -201,15 +225,16 @@ func (part PersistedParticipation) Persist() error {
...
 
 	err := part.Store.Atomic(func(ctx context.Context, tx *sql.Tx) error {
 		err := partInstallDatabase(tx)
 		if err != nil {
 			return fmt.Errorf("failed to install database: %w", err)
 		}
 
-		_, err = tx.Exec("INSERT INTO ParticipationAccount (parent, vrf, voting, firstValid, lastValid, keyDilution) VALUES (?, ?, ?, ?, ?, ?)",
-			part.Parent[:], rawVRF, rawVoting, part.FirstValid, part.LastValid, part.KeyDilution)
+		_, err = tx.Exec("INSERT INTO ParticipationAccount (parent, vrf, voting, blockProof, firstValid, lastValid, keyDilution) VALUES (?, ?, ?, ?, ?, ?,?)",
```
shiqizng, line 236: nit: add space after comma

libgoal/participation.go
```diff
@@ -165,6 +165,9 @@ func (c *Client) GenParticipationKeysTo(address string, firstValid, lastValid, k
 
 	// Fill the database with new participation keys
 	newPart, err := account.FillDBWithParticipationKeys(partdb, parsedAddr, firstRound, lastRound, keyDilution)
+	if err != nil {
```
shiqizng, line 168: newPart.Close() should also be called here if newPart is not nil?

crypto/compactcert/verifier.go
```diff
@@ -48,26 +47,29 @@ func (v *Verifier) Verify(c *Cert) error {
 		return fmt.Errorf("cert signed weight %d <= proven weight %d", c.SignedWeight, v.ProvenWeight)
 	}
```
shiqizng, line 49: is there a unit test for this method?

ledger/compactcert.go
```diff
@@ -132,10 +132,11 @@ func CompactCertParams(votersHdr bookkeeping.BlockHeader, hdr bookkeeping.BlockH
 	}
 
 	res = compactcert.Params{
-		Msg:          hdr,
-		ProvenWeight: provenWeight,
-		SigRound:     hdr.Round + 1,
-		SecKQ:        proto.CompactCertSecKQ,
+		Msg:               hdr,
+		ProvenWeight:      provenWeight,
+		SigRound:          hdr.Round,
```
shiqizng, line 137: Why is the sigRound param changed? Does it impact the existing implementation?

crypto/merklekeystore/persistentKeystore.go
```diff
@@ -0,0 +1,149 @@
...
+	}
+
+	err := p.store.Atomic(func(ctx context.Context, tx *sql.Tx) error {
+		err := keystoreInstallDatabase(tx) // assumes schema table already exists (created by partInstallDatabase)
+		if err != nil {
+			return err
+		}
+
+		round := indexToRound(firstValid, interval, 0)
+		for i, key := range keys {
```
jasonpaulos, line 49: nit: since this loop encodes potentially many keys, you could use `protocol.GetEncodingBuf()` to attempt to optimize memory allocation. For example:

```go
encoding := key.MarshalMsg(protocol.GetEncodingBuf())
_, err := tx.Exec("INSERT INTO BlockProofKeys (id, round, key) VALUES (?,?,?)", i, round, encoding)
protocol.PutEncodingBuf(encoding)
```

ledger/ledger_test.go
```diff
@@ -1080,6 +1084,24 @@ func testLedgerSingleTxApplyData(t *testing.T, version protocol.ConsensusVersion
...
+		root, err := account.GenerateRoot(store)
+		a.NoError(err)
+		p, err := account.FillDBWithParticipationKeys(store, root.Address(), basics.Round(frst), basics.Round(lst), config.Consensus[protocol.ConsensusCurrentVersion].DefaultKeyDilution)
+		signer := p.Participation.BlockProof
+		//signer, err := merklekeystore.New(frst, lst, 1, crypto.Ed25519Type)
+		require.NoError(t, err)
+
+		correctKeyregFields.BlockProofPK = *(signer.GetVerifier())
+	} else {
+		correctKeyregFields.BlockProofPK = merklekeystore.Verifier{}
```
jasonpaulos, line 1102: Not needed, since `correctKeyregFields.BlockProofPK` is already initialized to the zero value

crypto/merklekeystore/keystore.go
```diff
@@ -0,0 +1,264 @@
...
+	// Verifier is used to verify a merklekeystore.Signature produced by merklekeystore.Signer.
+	// it validate a merklekeystore.Signature by validate the commitment on the VerifyingKey and validate the signature with that key
+	Verifier struct {
+		_struct struct{} `codec:",omitempty,omitemptyarray"`
+
+		Root [KeyStoreRootSize]byte `codec:"r"`
+
+		// indicates that this verifier corresponds to a specific array of ephemeral keys.
+		// this is used to distinguish between an empty structure, and nothing to commit to.
+		HasValidRoot bool `codec:"vr"`
```
jasonpaulos, line 79: I believe `HasValidRoot` is meant to indicate whether the struct is the zero value or not, but it’s not used everywhere, since there are frequent instances of `== merklekeystore.Verifier{}`. If this is the case, I recommend creating an `IsZero` method on `merklekeystore.Verifier` that returns `HasValidRoot` and making this the only zero check in use.

crypto/sig_abstractions.go
```diff
@@ -0,0 +1,166 @@
...
+// Verifier interface represents any algorithm that can verify signatures for a specific signing scheme.
+type Verifier interface {
+	// Verify and VerifyBytes returns error on bad signature, and any other problem.
+	Verify(message Hashable, sig ByteSignature) error
+	VerifyBytes(message []byte, sig ByteSignature) error
+}
+
+// SignatureAlgorithm holds a Signer, and the type of algorithm the Signer conforms with.
+// to add a key - verify that PackedSignatureAlgorithm's function (getSigner) returns your key.
+type SignatureAlgorithm struct {
```
jasonpaulos, line 64: I think `SignatureAlgorithm` is not the best name, since it sounds like it represents an algorithm, not a signer + verifier for a specific algorithm. Perhaps `GenericSigningKey` would be better?

Then maybe `VerifyingKey` could be `GenericVerifyingKey`.

crypto/sig_abstractions.go
```diff
@@ -0,0 +1,166 @@
...
+		return &p.DilithiumPublicKey, nil
+	case Ed25519Type:
+		return &p.Ed25519PublicKey, nil
+	default:
+		return nil, errUnknownVerifier
+	}
+}
+
+// PackedSignatureAlgorithm helps  marshal SignatureAlgorithm
+type PackedSignatureAlgorithm struct {
```
jasonpaulos, line 121: I don't see the benefit of the structs `PackedVerifyingKey` and `PackedSignatureAlgorithm`. I think it would be simpler if `VerifyingKey` and `SignatureAlgorithm` simply had the key fields directly in their structs, since it avoids an additional layer of indirection.

algonautshant (commented):
roundToIndex and indexToRound functions seem to have serious issues.

go.sum
```diff
@@ -1,3 +1,10 @@
+<<<<<<< variant A
+>>>>>>> variant B
```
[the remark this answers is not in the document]
algonautshant: I can take the blame for this :-)

data/account/participation.go
```diff
@@ -121,6 +130,11 @@ func (part Participation) GenerateRegistrationTransaction(fee basics.MicroAlgos,
 			SelectionPK: part.VRF.PK,
 		},
 	}
+	if cert := part.BlockProofSigner(); cert != nil {
+		if cparams.EnableBlockProofKeyregCheck {
```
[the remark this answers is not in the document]
Aharonee: Nothing I guess. The BlockProof key will not be registered in the transaction since it's not yet supported on this protocol version.

cmd/algokey/part.go
```diff
@@ -72,12 +72,14 @@ var partGenerateCmd = &cobra.Command{
 		partdb, err := db.MakeErasableAccessor(partKeyfile)
 		if err != nil {
 			fmt.Fprintf(os.Stderr, "Cannot open partkey database %s: %v\n", partKeyfile, err)
+			os.Remove(partKeyfile)
```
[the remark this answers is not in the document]
Aharonee: I don't think there's much to do even if it fails and returns an error, I'll add a log that removing the file failed though

libgoal/participation.go
```diff
@@ -165,8 +165,11 @@ func (c *Client) GenParticipationKeysTo(address string, firstValid, lastValid, k
 
 	// Fill the database with new participation keys
 	newPart, err := account.FillDBWithParticipationKeys(partdb, parsedAddr, firstRound, lastRound, keyDilution)
-	part = newPart.Participation
```
algonautshant, line 168: This change is not necessary. 
There is something wrong here, but this change does not fix it. 
partdb should be closed. returning if err != nil leaves it open. 
There will be a fix in master, and when merged with it, should resolve it. Not making any changes is is advised.

data/account/participation.go
```diff
@@ -176,23 +190,39 @@ func FillDBWithParticipationKeys(store db.Accessor, address basics.Address, firs
...
+	// Generate a new VRF key, which lives in the participation keys db
 	vrf := crypto.GenerateVRFSecrets()
 
+	// TODO change this
+	compactCertRound := config.Consensus[protocol.ConsensusFuture].CompactCertRounds
+
+	// Generate a new key which signs the compact certificates
+	stateProofSecrets, err := merklekeystore.New(uint64(firstValid), uint64(lastValid), compactCertRound, crypto.DilithiumType, store)
+	if err != nil {
+		return PersistedParticipation{}, err
```
algonautshant, line 202: return is sufficient here. See line 182 above.

data/account/participation.go
```diff
@@ -176,23 +190,39 @@ func FillDBWithParticipationKeys(store db.Accessor, address basics.Address, firs
...
+			FirstValid:        firstValid,
+			LastValid:         lastValid,
+			KeyDilution:       keyDilution,
 		},
 		Store: store,
 	}
 	// Persist the Participation into the database
 	err = part.Persist()
+	if err != nil {
+		return PersistedParticipation{}, err
```
algonautshant, line 221: Ideally, return is sufficient here, as in line 182 above. 
However, it is safer not to change the behavior of what is returned in case of an error. 
Use this logic to only call stateProofSecrets.Persist in case of no error, and leave it as is otherwise. 
```
if err == nil {
    err = stateProofSecrets.Persist() 
}
```

crypto/dillithium.go
```diff
@@ -0,0 +1,93 @@
...
+type (
+	//DilithiumPublicKey is the public key
+	//msgp:allocbound DilithiumPublicKey
+	DilithiumPublicKey []byte
+	//DilithiumPrivateKey is the private key
+	//msgp:allocbound DilithiumPrivateKey
+	DilithiumPrivateKey []byte
+	//DilithiumSignature is the exported signature
+	//msgp:allocbound DilithiumSignature
+	DilithiumSignature ByteSignature
```
algonautshant, line 31: If these are not removed, they need to be bounded.

test/e2e-go/upgrades/stateproof_test.go
```diff
@@ -0,0 +1,160 @@
...
+	fixture.Setup(t, filepath.Join("nettemplates", "TwoNodesWithoutStateProofPartkeys.json"))
+	defer fixture.Shutdown()
+	lastValid := uint64(1000 * 5)
+
+	a.NoError(registerKey(&fixture, a, lastValid, protocol.ConsensusV29))
+	a.Error(registerKey(&fixture, a, lastValid+1, protocol.ConsensusFuture))
+
+	runUntilProtocolUpgrades(a, &fixture)
+
+	a.Error(registerKey(&fixture, a, lastValid+2, protocol.ConsensusV29))
```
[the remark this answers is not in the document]
algonautshant: test refactored.

test/e2e-go/upgrades/stateproof_test.go
```diff
@@ -0,0 +1,160 @@
...
+	consensus := getStateProofConcensus()
+
+	var fixture fixtures.RestClientFixture
+	fixture.SetConsensus(consensus)
+	fixture.Setup(t, filepath.Join("nettemplates", "TwoNodesWithoutStateProofPartkeys.json"))
+	defer fixture.Shutdown()
+	lastValid := uint64(1000 * 5)
+
+	a.NoError(registerKey(&fixture, a, lastValid, protocol.ConsensusV29))
+	a.Error(registerKey(&fixture, a, lastValid+1, protocol.ConsensusFuture))
```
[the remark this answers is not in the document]
algonautshant: test refactored

crypto/merklekeystore/keystore.go
```diff
@@ -0,0 +1,264 @@
...
+	Signature struct {
+		_struct              struct{} `codec:",omitempty,omitemptyarray"`
+		crypto.ByteSignature `codec:"bsig"`
+
+		Proof        Proof               `codec:"prf"`
+		VerifyingKey crypto.VerifyingKey `codec:"vkey"`
+	}
+
+	// Signer is a merkleKeyStore, contain multiple keys which can be used per round.
+	// Signer will generate all keys in the range [A,Z] that are divisible by some divisor d.
```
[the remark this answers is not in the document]
algonautshant: this explanation should be in the code comment :-)

crypto/merklekeystore/keystore.go
```diff
@@ -0,0 +1,264 @@
...
+)
+
+var errStartBiggerThanEndRound = errors.New("cannot create merkleKeyStore because end round is smaller then start round")
+var errReceivedRoundIsBeforeFirst = errors.New("round translated to be prior to first key position")
+var errOutOfBounds = errors.New("round translated to be after last key position")
+var errNonExistantKey = errors.New("key doesn't exist")
+var errDivisorIsZero = errors.New("received zero Interval")
+var errCannotVerify = errors.New("verifier isn't valid")
+
+// ToBeHashed implementation means CommittablePublicKey is crypto.Hashable.
```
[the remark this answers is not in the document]
algonautshant: thx.

crypto/merklekeystore/posdivs.go
```diff
@@ -0,0 +1,25 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with go-algorand.  If not, see <https://www.gnu.org/licenses/>.
+
+package merklekeystore
+
+func roundToIndex(firstValid, currentRound, interval uint64) uint64 {
```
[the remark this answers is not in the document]
algonautshant: this code is refactored.

crypto/merklekeystore/keystore.go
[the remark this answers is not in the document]
algonautshant: I added a comment.

crypto/merklekeystore/posdivs.go
```diff
@@ -0,0 +1,25 @@
...
+	return currentRound/interval - ((firstValid - 1) / interval) - 1
+}
+
+func indexToRound(firstValid, interval, pos uint64) uint64 {
+	return (((firstValid - 1) / interval) + 1 + pos) * interval
+}
```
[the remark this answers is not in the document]
algonautshant: fixed.

crypto/merklekeystore/keystore_test.go
```diff
@@ -0,0 +1,530 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with go-algorand.  If not, see <https://www.gnu.org/licenses/>.
+
+package merklekeystore
+
+import (
+	"context"
```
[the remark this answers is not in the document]
algonautshant: updated.

libgoal/participation.go
algonautshant: resolved.

data/account/participation.go
algonautshant: I updated/

data/account/participation.go
algonautshant: i updated

cmd/goal/account.go
```diff
@@ -1428,22 +1424,27 @@ var partkeyInfoCmd = &cobra.Command{
...
+				// PKI TODO: enable with state proof support.
+				//fmt.Printf("Last state proof round:    %s\n", strOrNA(part.LastStateProof))
+				fmt.Printf("Effective first round:     %s\n", strOrNA(part.EffectiveFirstValid))
+				fmt.Printf("Effective last round:      %s\n", strOrNA(part.EffectiveLastValid))
+				fmt.Printf("First round:               %d\n", part.Key.VoteFirstValid)
+				fmt.Printf("Last round:                %d\n", part.Key.VoteLastValid)
+				fmt.Printf("Key dilution:              %d\n", part.Key.VoteKeyDilution)
+				fmt.Printf("Selection key:             %s\n", base64.StdEncoding.EncodeToString(part.Key.SelectionParticipationKey))
+				fmt.Printf("Voting key:                %s\n", base64.StdEncoding.EncodeToString(part.Key.VoteParticipationKey))
 				if certSigner := part.StateProofSigner(); certSigner != nil {
```
winder, line 1442: The `ParticipationKeysResponse` object returned by `client.ListParticipationKeys()` isn't going to have this method, so this code probably won't compile.

The legacy function was renamed `client.ListParticipationKeyFiles()`, but we should avoid using it here. I think we need to create a followup ticket specifically to include state-proof information in the new REST endpoints. After that, we can uncomment the `PKI TODO` lines.

algonautshant (changes requested):

crypto/merklekeystore/keystore_test.go
```diff
@@ -0,0 +1,538 @@
...
+	a := require.New(t)
+	var err error
+
+	h := genHashableForTest()
+	signer := generateTestSigner(crypto.FalconType, 8, 9, 5, a)
+	a.NoError(err)
+	a.Equal(0, length(signer, a))
+
+	_, err = signer.Sign(h, 8)
+	a.Error(err)
```
algonautshant, line 121: Will be nice to check `err` that it is the expected error.

crypto/merklekeystore/keystore_test.go
```diff
@@ -0,0 +1,538 @@
...
+
+	_, err = signer.Sign(h, 9)
+	a.Error(err)
```
algonautshant, line 124: Will be nice to check `err` that it is the expected error.

crypto/merklekeystore/keystore_test.go
```diff
@@ -0,0 +1,538 @@
...
+
+	signer = generateTestSigner(crypto.FalconType, 1000, 1100, 101, a)
+	intervalRounds := make([]uint64, 0)
+	for i := uint64(1000); i <= 1100; i++ {
+		if i%101 == 0 {
+			intervalRounds = append(intervalRounds, i)
+			continue
+		}
+		_, err := signer.keyStore.GetKey(i)
+		a.Error(err, i)
```
algonautshant, line 163: Will be nice to check `err` that it is the expected error.

crypto/merklekeystore/keystore_test.go
```diff
@@ -0,0 +1,538 @@
...
+	signer = generateTestSigner(crypto.FalconType, start, end, 10, a)
+	defer signer.keyStore.store.Close()
+
+	sig, err = signer.Sign(hashable, start)
+	a.NoError(err)
+	a.NoError(signer.GetVerifier().Verify(start, start, 1, hashable, sig))
+
+	sig, err = signer.Sign(hashable, start+5)
+	a.Error(err)
+	a.Error(signer.GetVerifier().Verify(start, start+5, 1, hashable, sig))
```
algonautshant, line 250: Will be nice to check `err` that it is the expected error.

crypto/merklekeystore/keystore_test.go
```diff
@@ -0,0 +1,538 @@
...
+
+	signer = generateTestSigner(crypto.FalconType, 50, 100, 12, a)
+	defer signer.keyStore.store.Close()
+	a.Equal(length(signer, a), 4)
```
algonautshant, line 254: `Equal` first argument is the expected value, second is the checked value. The parameter order must be reversed.

crypto/merklekeystore/KeystoreBuilder_test.go
```diff
@@ -0,0 +1,94 @@
...
+)
+
+func TestBuilderSanity(t *testing.T) {
+	partitiontest.PartitionTest(t)
+	a := require.New(t)
+
+	numOfKeys := uint64(100)
+	keys, err := KeyStoreBuilder(numOfKeys, crypto.FalconType)
+	a.NoError(err)
+	a.Equal(uint64(len(keys)), numOfKeys)
```
algonautshant, line 38: `Equal` first argument is the expected value, second is the checked value. The parameter order must be reversed.

data/account/participation_test.go
```diff
@@ -149,3 +172,279 @@ func BenchmarkOldKeysDeletion(b *testing.B) {
...
+func TestKeyRegCreation(t *testing.T) {
+	partitiontest.PartitionTest(t)
+
+	a := require.New(t)
+
+	ppart := setupkeyWithNoDBS(t, a)
+
+	cur := config.Consensus[protocol.ConsensusCurrentVersion]
+	txn := ppart.Participation.GenerateRegistrationTransaction(basics.MicroAlgos{Raw: 1000}, 0, 100, [32]byte{}, cur)
+	a.Equal(txn.StateProofPK.IsEmpty(), true)
```
algonautshant, line 234: `Equal` first argument is the expected value, second is the checked value. The parameter order must be reversed.

data/account/participation_test.go
```diff
@@ -149,3 +172,279 @@ func BenchmarkOldKeysDeletion(b *testing.B) {
...
+
+	future := config.Consensus[protocol.ConsensusFuture]
+	txn = ppart.Participation.GenerateRegistrationTransaction(basics.MicroAlgos{Raw: 1000}, 0, 100, [32]byte{}, future)
+	a.Equal(txn.StateProofPK.IsEmpty(), false)
```
algonautshant, line 238: `Equal` first argument is the expected value, second is the checked value. The parameter order must be reversed.

crypto/merklekeystore/KeystoreBuilder_test.go
```diff
@@ -0,0 +1,94 @@
...
+	a.NoError(err)
+}
+
+func TestBuilderFitsToCPUs(t *testing.T) {
+	partitiontest.PartitionTest(t)
+	a := require.New(t)
+	numOfKeys := uint64(runtime.NumCPU() * 2)
+	keys, err := KeyStoreBuilder(numOfKeys, crypto.FalconType)
+	a.NoError(err)
+	a.Equal(uint64(len(keys)), numOfKeys)
```
algonautshant, line 52: `Equal` first argument is the expected value, second is the checked value. The parameter order must be reversed.

crypto/merklekeystore/KeystoreBuilder_test.go
```diff
@@ -0,0 +1,94 @@
...
+
+}
+
+func TestBuilderOneKey(t *testing.T) {
+	partitiontest.PartitionTest(t)
+	a := require.New(t)
+	numOfKeys := uint64(1)
+	keys, err := KeyStoreBuilder(numOfKeys, crypto.FalconType)
+	a.NoError(err)
+	a.Equal(uint64(len(keys)), numOfKeys)
```
algonautshant, line 62: `Equal` first argument is the expected value, second is the checked value. The parameter order must be reversed.

crypto/merklekeystore/KeystoreBuilder_test.go
```diff
@@ -0,0 +1,94 @@
...
+}
+
+func TestBuilderZeroKeys(t *testing.T) {
+	partitiontest.PartitionTest(t)
+	a := require.New(t)
+	numOfKeys := uint64(0)
+	keys, err := KeyStoreBuilder(numOfKeys, crypto.FalconType)
+	a.NoError(err)
+	a.Equal(uint64(len(keys)), numOfKeys)
```
algonautshant, line 71: `Equal` first argument is the expected value, second is the checked value. The parameter order must be reversed.

crypto/merklekeystore/keystore_test.go
```diff
@@ -0,0 +1,538 @@
...
+
+	signer := generateTestSigner(crypto.FalconType, 0, 10, 1, a)
+	store := signer.keyStore.store
+	defer store.Close()
+
+	out := protocol.Encode(signer)
+	decodeInto := &Signer{}
+	a.NoError(protocol.Decode(out, decodeInto))
+	decodeInto.keyStore.store = store // restore PersistentKeystore
+	a.Equal(signer, decodeInto)
```
algonautshant, line 382: `Equal` first argument is the expected value, second is the checked value. The parameter order must be reversed.

crypto/merklekeystore/keystore_test.go
```diff
@@ -0,0 +1,538 @@
...
+
+	verifier := signer.GetVerifier()
+	bs := protocol.Encode(verifier)
+	verifierToDecodeInto := Verifier{}
+	protocol.Decode(bs, &verifierToDecodeInto)
+	a.Equal(*verifier, verifierToDecodeInto)
```
algonautshant, line 388: `Equal` first argument is the expected value, second is the checked value. The parameter order must be reversed.

crypto/merklekeystore/keystore_test.go
```diff
@@ -0,0 +1,538 @@
...
+	_, err = signer.Trim(61) // should not return error for rounds bigger than lastValid
+	a.NoError(err)
+	a.Equal(0, length(signer, a))
+
+	signer = generateTestSigner(crypto.FalconType, 1, 60, 11, a)
+	defer signer.keyStore.store.Close()
+	_, err = signer.Trim(55)
+	a.NoError(err)
+	a.Equal(0, length(signer, a))
+
```
algonautshant, line 444: `Equal` first argument is the expected value, second is the checked value. The parameter order must be reversed.

.circleci/config.yml
```diff
@@ -284,7 +284,7 @@ commands:
         type: string
       no_output_timeout:
         type: string
-        default: 30m
```
algojack, line 287: Have this been discussed with devops?

algojack (changes requested):

.circleci/config.yml
algonautshant: This job is not merging with master anytime soon :-) 

This change is necessary for the branch to pass the tests on circleci. 
New cryptographic algorithms introduced in this PR are significantly more expensive, so many tests require more time. 

This parameter change will be part of a bigger solution that we are still working on to resolve the testing time.

algojack: sounds good, lets discuss on discord. It is ok while not merged to master.

cmd/goal/account.go
```diff
@@ -872,11 +899,16 @@ var addParticipationKeyCmd = &cobra.Command{
...
+		participationGen := func() {
+			_, _, err := client.GenParticipationKeysTo(accountAddress, roundFirstValid, roundLastValid, keyDilution, partKeyOutDir)
+			if err != nil {
+				reportErrorf(errorRequestFail, err)
+			}
+			reportInfof("Participation key generation successful")
 		}
-		fmt.Println("Participation key generation successful")
+
+		reportInfof("Please standby while generating keys. This might take few minutes...")
```
algonautshant, line 910:
```suggestion
		reportInfof("Please standby while generating keys. This might take a few minutes...")
```

.circleci/config.yml
algoidan: Reverting the changes for now. If something comes up, we will consult you. thanks

crypto/compactcert/const.go
```diff
@@ -0,0 +1,36 @@
...
+	HashSize        = crypto.SumhashDigestSize
+	SignatureScheme = crypto.FalconType
+)
+
+const (
+	// MaxReveals is a bound on allocation and on numReveals to limit log computation
+	MaxReveals = 1024
+
+	// MaxProofDigests is a bound on allocation on number of proofs
+	MaxProofDigests = 20 * MaxReveals
```
algonautshant, line 35: What is the purpose of `MaxProofDigests`? Where does "20" come from?
I think `MaxProofDigests` is not used.

crypto/compactcert/structs.go
```diff
@@ -63,23 +60,19 @@ type Reveal struct {
...
 	_struct struct{} `codec:",omitempty,omitemptyarray"`
 
-	SigCommit    crypto.Digest   `codec:"c"`
-	SignedWeight uint64          `codec:"w"`
-	SigProofs    []crypto.Digest `codec:"S,allocbound=maxProofDigests"`
-	PartProofs   []crypto.Digest `codec:"P,allocbound=maxProofDigests"`
+	SigCommit    crypto.GenericDigest `codec:"c"`
+	SignedWeight uint64               `codec:"w"`
+	SigProofs    merklearray.Proof    `codec:"S,allocbound=MaxProofDigests"`
+	PartProofs   merklearray.Proof    `codec:"P,allocbound=MaxProofDigests"`
```
algonautshant, line 70: `allocbound=MaxProofDigests` is useless, since `merklearray.Proof` is a struct and not an slice.

crypto/merklearray/merkle_test.go
```diff
@@ -555,6 +556,10 @@ func TestSizeLimitsMerkle(t *testing.T) {
 	var outTree Tree
 	err := protocol.Decode(bytes, &outTree)
 	require.Contains(t, err.Error(), "> 17 at Levels")
+
+	// Garbage collection helps prevent trashing
+	// for next tests
+	runtime.GC()
```
algonautshant, line 562: Why this is not (also) call this at the beginning of the test?

ledger/acctupdates_test.go
```diff
@@ -766,6 +767,9 @@ func TestLargeAccountCountCatchpointGeneration(t *testing.T) {
 			ml.trackers.waitAccountsWriting()
 		}
 	}
+
+	// Garbage collection helps prevent trashing for next tests
+	runtime.GC()
```
algonautshant, line 772: Why this is not (also) call this at the beginning of the test?

ledger/acctupdates_test.go
```diff
@@ -1399,6 +1403,9 @@ func TestAcctUpdatesCachesInitialization(t *testing.T) {
 	// make sure the deltas array end up containing only the most recent 320 rounds.
 	require.Equal(t, int(proto.MaxBalLookback), len(au.deltas))
 	require.Equal(t, recoveredLedgerRound-basics.Round(proto.MaxBalLookback), au.cachedDBRound)
+
+	// Garbage collection helps prevent trashing for next tests
+	runtime.GC()
```
algonautshant, line 1408: Why this is not (also) call this at the beginning of the test?

ledger/catchpointwriter_test.go
```diff
@@ -167,6 +168,9 @@ func TestCatchpointFileBalancesChunkEncoding(t *testing.T) {
 	require.NoError(t, err)
 
 	require.Equal(t, fbc, fbc2)
+	// Garbage collection helps prevent trashing
+	// for next tests
+	runtime.GC()
```
algonautshant, line 173: Why this is not (also) call this at the beginning of the test?

test/e2e-go/features/transactions/proof_test.go
```diff
@@ -126,5 +126,9 @@ func TestTxnMerkleProof(t *testing.T) {
 
 	elems[proofresp.Idx] = &element
 	err = merklearray.Verify(blk.TxnRoot.ToSlice(), elems, &proof)
-	a.NoError(err)
+	if err != nil {
+		t.Logf("blk.TxnRoot : %v \nproof path %v \ndepth: %d \nStibhash %v", blk.TxnRoot.ToSlice(), proof.Path, proof.TreeDepth, proofresp.Stibhash)
+		t.FailNow()
```
algonautshant, line 131: I prefer to still call `a.NoError(err)` here instead of `t.FailNow()`. 
`t.FailNow()` is not commonly used, and might act funny when using other objects implementing the interface.

ledger/catchpointwriter_test.go
algonautshant, line 173: Why this is not (also) call this at the beginning of the test?

crypto/compactcert/const.go
algoidan: I think you are right and we might not need it (because we have a bound on the proof). we will absolutely review this once stateproof are released. for now, this is unused

crypto/compactcert/verifier.go
```diff
@@ -48,26 +47,29 @@ func (v *Verifier) Verify(c *Cert) error {
...
 
-	// Verify all of the reveals
-	sigs := make(map[uint64]crypto.Digest)
-	parts := make(map[uint64]crypto.Digest)
+	sigs := make(map[uint64]crypto.Hashable)
+	parts := make(map[uint64]crypto.Hashable)
```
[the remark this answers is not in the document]
algoidan: you are right! but all this code is going to be rewritten in order to make it more SNARK friendly. (currently, it is not used).

crypto/hashes.go
```diff
@@ -0,0 +1,150 @@
...
+	}
+}
+
+// IsValid states whether the HashFactory is valid, and is safe to use.
+func (z *HashFactory) IsValid() error {
+	return z.HashType.IsValid()
+}
+
+// GenereicHashObj Makes it easier to sum using hash interface and Hashable interface
+func GenereicHashObj(hsh hash.Hash, h Hashable) []byte {
```
cce, line 111:
```suggestion
func GenericHashObj(hsh hash.Hash, h Hashable) []byte {
```

cmd/goal/account.go
```diff
@@ -877,11 +877,16 @@ var addParticipationKeyCmd = &cobra.Command{
...
 
-		_, _, err := client.GenParticipationKeysTo(accountAddress, roundFirstValid, roundLastValid, keyDilution, partKeyOutDir)
-		if err != nil {
-			reportErrorf(errorRequestFail, err)
+		participationGen := func() {
+			_, _, err := client.GenParticipationKeysTo(accountAddress, roundFirstValid, roundLastValid, keyDilution, partKeyOutDir)
+			if err != nil {
+				reportErrorf(errorRequestFail, err)
+			}
+			reportInfof("Participation key generation successful")
```
tsachiherman, line 885: same as above; please place the error reporting after the `util.RunFuncWithSpinningCursor`

crypto/hashes.go
```diff
@@ -0,0 +1,150 @@
...
+
+	"github.com/algorand/go-algorand/protocol"
+	"github.com/algorand/go-sumhash"
+)
+
+// HashType represents different hash functions
+type HashType uint16
+
+// IsValid verifies that the hash type is in a valid range.
+func (h HashType) IsValid() error {
```
tsachiherman, line 33: this isn't really a blocker, but having a method called "IsValid()" that returns an `error` doesn't seems right.
I would expect it to return a bool. If the method would have been called "Validate()", then it would be just fine.

crypto/merklearray/proof_test.go
```diff
@@ -0,0 +1,132 @@
...
+// You should have received a copy of the GNU Affero General Public License
+// along with go-algorand.  If not, see <https://www.gnu.org/licenses/>.
+
+package merklearray
+
+import (
+	"github.com/algorand/go-algorand/crypto"
+	"github.com/algorand/go-algorand/test/partitiontest"
+	"github.com/stretchr/testify/require"
+	"testing"
```
tsachiherman, line 23: reorder imports as follows:
```golang
import (
 	"testing"

)
```

crypto/merklearray/vectorCommitmentArray.go
```diff
@@ -0,0 +1,82 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with go-algorand.  If not, see <https://www.gnu.org/licenses/>.
+
+package merklearray
+
+import (
```
tsachiherman, line 19: reorder imports as follows:
```golang
import (
	"fmt"
 	"math/bits"

 )

```

crypto/merklearray/vectorCommitmentArray.go
```diff
@@ -0,0 +1,82 @@
...
+	"fmt"
+	"github.com/algorand/go-algorand/crypto"
+	"github.com/algorand/go-algorand/protocol"
+	"math/bits"
+)
+
+// ErrGetOutOfBound returned when trying to retrieve an element which is out of
+// the padded array bound.
+var (
+	ErrGetOutOfBound = "vectorCommitmentArray.Get(%d): out of bounds, full size %d"
```
tsachiherman, line 29: instead of defining an error string here, we should have a custom error object here, which has two fields : `requestIndex`, `length`. The Error() method should perform the formatting.
The unit test could verify this error by using `errors.Is`.

crypto/merklearray/vectorCommitmentArray_test.go
```diff
@@ -0,0 +1,204 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with go-algorand.  If not, see <https://www.gnu.org/licenses/>.
+
+package merklearray
+
+import (
```
tsachiherman, line 19: nit: import ordering:
```golang
import (
 	"testing"

 )
```

crypto/merklearray/merkle.go
```diff
@@ -17,63 +17,122 @@
...
+)
+
+// Merkle tree errors
+var (
+	ErrRootMismatch                  = errors.New("root mismatch")
+	ErrProvingZeroCommitment         = errors.New("proving in zero-length commitment")
+	ErrProofIsNil                    = errors.New("proof should not be nil")
+	ErrNonEmptyProofForEmptyElements = errors.New("non-empty proof for empty set of elements")
+	ErrUnexpectedTreeDepth           = errors.New("unexpected tree depth")
+	ErrPosOutOfBound                 = "pos %d larger than leaf count %d"
```
cce, line 46: the more idiomatic Go thing to do is to define this as an error type e.g. `errors.New("pos out of bound")` and then below wrap it with the specifics `fmt.Errorf("pos %d larger than leaf count %d: %w", idx, num, ErrPosOutOfBound)`. That way you can test it with errors.Is(err, ErrPosOutOfBound) as you are doing in your tests

crypto/merklearray/merkle_test.go
```diff
@@ -66,109 +75,998 @@ func (a TestRepeatingArray) Length() uint64 {
...
 	if pos >= a.count {
-		return crypto.Digest{}, fmt.Errorf("pos %d larger than length %d", pos, a.count)
+		return nil, fmt.Errorf("pos %d larger than length %d", pos, a.count)
 	}
 
-	return crypto.HashObj(a.item), nil
+	return a.item, nil
 }
 
+const OutOfBoundErrorString = "larger than leaf count"
```
cce, line 86: You should replace this with `errors.Is(ErrPosOutOfBound)` using the wrapping technique as per https://go.dev/blog/go1.13-errors

data/accountManager.go
```diff
@@ -56,14 +56,19 @@ func MakeAccountManager(log logging.Logger, registry account.ParticipationRegist
...
 }
 
-// Keys returns a list of Participation accounts.
-func (manager *AccountManager) Keys(rnd basics.Round) (out []account.Participation) {
+// Keys returns a list of Participation accounts, and their keys/secrets for requested round.
+func (manager *AccountManager) Keys(rnd basics.Round) (out []account.ParticipationRecordForRound) {
 	manager.mu.Lock()
 	defer manager.mu.Unlock()
 
 	for _, part := range manager.partKeys {
```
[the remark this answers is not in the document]
winder: `manager.partKeys` is the account managers' cache of the legacy old partkey files. Since you're switching to the participation registry it may be good to delete that object and switch everything over to `manager.registry.GetAll()`.

algonautshant (approved):
Looks Great!
There are some outstanding issue, but not blocking. Can be addressed later.

data/accountManager.go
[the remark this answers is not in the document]
Aharonee: Yes it does seem to be no longer relevant since the Participation data is now contained in the registry, I'll remove it completely.

tsachiherman (changes requested):
please see my comments below.

data/accountManager.go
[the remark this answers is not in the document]
Aharonee: I've noticed that `DeleteOldKeys()` is dependent on this `PersistedParticipation` cache, and removing it completely means the ephemeral voting keys can no longer be deleted from the old database. It basically means we remove most logic from `oldKeyDeletionThread` and leave only `DeleteExpired` which is implemented in the registry.
We should implement deleting old voting keys in the registry and probably remove the old database entirely, but it seems out of scope for this feature branch.
So I think we should keep the `partKeys` for now. 
Thoughts? @winder @cce

crypto/merklearray/merkle.go
```diff
@@ -26,54 +28,72 @@ import (
 // Tree is a Merkle tree, represented by layers of nodes (hashes) in the tree
 // at each height.
 type Tree struct {
+	_struct struct{} `codec:",omitempty,omitemptyarray"`
+
 	// Level 0 is the leaves.
-	levels []layer
+	Levels []Layer            `codec:"lvls,allocbound=-"`
```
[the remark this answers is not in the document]
algonautshant: unbounded

crypto/merklesignature/merkleSignatureScheme.go
```diff
@@ -0,0 +1,278 @@
...
+	SignerContext struct {
+		_struct struct{} `codec:",omitempty,omitemptyarray"`
+
+		FirstValid uint64           `codec:"rnd"`
+		Interval   uint64           `codec:"iv"`
+		Tree       merklearray.Tree `codec:"tree"`
+	}
+
+	// Verifier is used to verify a merklesignature.Signature produced by merklesignature.Secrets.
+	Verifier [MerkleSignatureSchemeRootSize]byte
```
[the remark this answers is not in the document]
algonautshant: `MerkleSignatureSchemeRootSize` is a constant, and Verifier is of a fixed size. Does it need explicit alloc bound?

gen/generate.go
```diff
@@ -267,6 +270,9 @@ func generateGenesisFiles(outDir string, protoVersion protocol.ConsensusVersion,
 				data.VoteFirstValid = part.FirstValid
 				data.VoteLastValid = part.LastValid
 				data.VoteKeyDilution = part.KeyDilution
+				if config.Consensus[protocol.ConsensusFuture].EnableStateProofKeyregCheck {
```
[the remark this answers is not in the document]
Aharonee: I have done so in https://github.com/algorand/go-algorand/pull/3579, I think we used the global value from `ConsensusFuture` here as a temporary solution instead adapting all the e2e tests that depend upon it to use and then having to revert these changes.

crypto/merklesignature/persistentMerkleSignatureScheme.go
```diff
@@ -0,0 +1,127 @@
...
+		);`)
+	if err != nil {
+		return err
+	}
+
+	_, err = tx.Exec(`CREATE UNIQUE INDEX IF NOT EXISTS roundIdx ON StateProofKeys (round);`)
+	if err != nil {
+		return err
+	}
+	_, err = tx.Exec("INSERT INTO schema (tablename, version) VALUES (?, ?)", merkleSignatureTableSchemaName, merkleSignatureSchemaVersion)
```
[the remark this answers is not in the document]
tsachiherman: I'm marking this as resolves as it's a non-blocker.

crypto/merklesignature/merkleSignatureScheme.go
```diff
@@ -0,0 +1,278 @@
...
+// GetVerifier can be used to store the commitment and verifier for this signer.
+func (s *Secrets) GetVerifier() *Verifier {
+	return s.SignerContext.GetVerifier()
+}
+
+// GetVerifier can be used to store the commitment and verifier for this signer.
+func (s *SignerContext) GetVerifier() *Verifier {
+	ver := [MerkleSignatureSchemeRootSize]byte{}
+	ss := s.Tree.Root().ToSlice()
+	copy(ver[:], ss)
```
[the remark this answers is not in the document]
Aharonee: Should we log some error and return a pointer to an empty verifier in that case?

crypto/merklesignature/merkleSignatureScheme.go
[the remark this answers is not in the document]
tsachiherman: let me check this one..

data/accountManager.go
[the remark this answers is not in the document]
winder: @Aharonee that sounds reasonable, thanks!

crypto/merklesignature/merkleSignatureScheme.go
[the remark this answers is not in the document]
tsachiherman: yes, this one is good.

gen/generate.go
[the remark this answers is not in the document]
tsachiherman: ok

crypto/compactcert/committableSignatureSlot.go
```diff
@@ -0,0 +1,88 @@
...
+
+type committableSignatureSlot struct {
+	sigCommit           sigslotCommit
+	serializedSignature []byte
+	isEmptySlot         bool
+}
+
+const (
+	// ErrIndexOutOfBound returned when an index is out of the array's bound
+	ErrIndexOutOfBound = "pos %d past end %d"
```
winder, line 35: nit: There's another version of this string here: [crypto/merklesignature/committablePublicKeys.go](https://github.com/algorand/go-algorand/pull/2990/files#diff-bad406866bae28f1deac671cd7992227ad06e9945554853e250518bc3aa1d45f)

Maybe they are unloaded in Github, but I don't see them used anywhere. Could they be private?

:upside_down_face: this could be a struct with position / length, and format the message with `Error()`. That might allow using `errors.Is` if we do check for the error elsewhere.

crypto/merklesignature/merkleSignatureScheme.go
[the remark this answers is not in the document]
tsachiherman: I think that a better solution would be to keep the code as is, but adding a verification in `merklesignature.New` that validates that the configured tree produces a hash of the desired length.
Then, add a comment in `GetVerifier()` that saying that you're using that assumption here as well.
( this might be ok, since the tree object is internal, and the logic of the code prevents the hash length from changing )

data/accountManager.go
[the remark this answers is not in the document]
cce: @winder so it's OK to be looping over `part := range manager.partKeys` and mixing it with a call to `manager.registry.GetForRound(part.ID(), rnd)`? you seemed concerned earlier that it was mixing accessing partKey cache with the ParticipationRegistry.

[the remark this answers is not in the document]
winder: We should still switch to `GetAll()`, but no need to fully remove `partKeys`

crypto/merklearray/merkle.go
```diff
@@ -17,63 +17,122 @@
...
 )
 
+const (
+	// MaxEncodedTreeDepth is the maximum tree depth (root only depth 0) for a tree which
+	// is being encoded (either by msbpack or by the fixed length encoding)
+	MaxEncodedTreeDepth = 16
+
+	// MaxNumLeavesOnEncodedTree is the maximum number of leaves allowed for a tree which
+	// is being encoded (either by msbpack or by the fixed length encoding)
+	MaxNumLeavesOnEncodedTree = 65536 // 2^MaxEncodedTreeDepth
```
winder, line 36:
```suggestion
	// MaxNumLeavesOnEncodedTree is the maximum number of leaves allowed for a tree which
	// is being encoded (either by msbpack or by the fixed length encoding)
	MaxNumLeavesOnEncodedTree = 1 << MaxEncodedTreeDepth
```

crypto/merklesignature/keysBuilder.go
```diff
@@ -0,0 +1,95 @@
...
+	"sync/atomic"
+
+	"github.com/algorand/go-algorand/crypto"
+)
+
+// KeysBuilder Responsible for generate slice of falcon keys
+func KeysBuilder(numberOfKeys uint64) ([]crypto.FalconSigner, error) {
+	numOfKeysPerRoutine, _ := calculateRanges(numberOfKeys)
+
+	var terminate int64
```
winder, line 31: nit: `context.WithCancel` could be used instead of terminate, it's probably more idiomatic than the integer / atomic ops pattern here.

data/accountManager.go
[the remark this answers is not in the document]
Aharonee: I'll use the `registry.GetAll()` wherever I can instead of `partKeys` but keep it for compatibility in `DeleteOldKeys`, until we have that fully implemented within the registry.

data/basics/ccertpart.go
```diff
@@ -17,10 +17,18 @@
 package basics
 
 import (
+	"encoding/binary"
+	"fmt"
```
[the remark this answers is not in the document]
tsachiherman: done

ledger/apply/keyreg_test.go
```diff
@@ -17,6 +17,8 @@
 package apply
 
 import (
+	"github.com/algorand/go-algorand/data/account"
+	"github.com/algorand/go-algorand/util/db"
```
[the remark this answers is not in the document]
tsachiherman: done

util/cmdUtils.go
```diff
@@ -0,0 +1,50 @@
...
+		errChan <- struct{}{}
+	}()
+
+	progressStrings := [...]string{"/", "-", "\\", "|"}
+
+	finished := false
+	i := 0
+
+	for !finished {
+		timer := time.NewTimer(100 * time.Millisecond)
```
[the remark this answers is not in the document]
tsachiherman: done

util/cmdUtils.go
```diff
@@ -0,0 +1,50 @@
...
+
+import (
+	"fmt"
+	"time"
+)
+
+// RunFuncWithSpinningCursor runs a given function in a go-routine,
+// while displaying a spinning cursor to the CLI
+func RunFuncWithSpinningCursor(asyncFunc func()) {
+	errChan := make(chan struct{}, 1)
```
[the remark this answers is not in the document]
tsachiherman: done

crypto/compactcert/committableSignatureSlot_test.go
```diff
@@ -0,0 +1,98 @@
...
+import (
+	"encoding/binary"
+	"github.com/algorand/go-algorand/crypto"
+	"github.com/algorand/go-algorand/crypto/merklearray"
+	"github.com/algorand/go-algorand/crypto/merklesignature"
+	"github.com/algorand/go-algorand/protocol"
+	"github.com/stretchr/testify/require"
+	"testing"
+
+	"github.com/algorand/go-algorand/test/partitiontest"
```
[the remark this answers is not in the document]
tsachiherman: done

crypto/merklesignature/keysBuilder.go
tsachiherman: done

crypto/merklesignature/persistentMerkleSignatureScheme.go
```diff
@@ -0,0 +1,138 @@
...
+	"database/sql"
+	"fmt"
+
+	"github.com/algorand/go-algorand/crypto"
+	"github.com/algorand/go-algorand/protocol"
+	"github.com/algorand/go-algorand/util/db"
+)
+
+const merkleSignatureSchemaVersion = 1
+const merkleSignatureTableSchemaName = "merklesignaturescheme"
```
winder, line 35: The version isn't set on the DB, this logic looks incomplete.

There's a  DB utility to automate versioning & upgrading SQLite DBs, you use `db.Initialize` and pass in functions that create/modify the DB: https://github.com/algorand/go-algorand/blob/c8c6fb7d2395e2298342824eeadba0d498cd3228/util/db/initialize.go#L34

Here is how the participation registry uses it: https://github.com/algorand/go-algorand/blob/c8c6fb7d2395e2298342824eeadba0d498cd3228/data/account/participationRegistry.go#L237

crypto/merklesignature/persistentMerkleSignatureScheme_test.go
```diff
@@ -0,0 +1,81 @@
...
+		}
+		a.Nil(key1)
+		a.Nil(key2)
+	}
+
+}
+
+func createTestDB(a *require.Assertions) *db.Accessor {
+	tmpname := uuid.NewV4().String() // could this just be a constant string instead? does it even matter?
+	store, err := db.MakeAccessor(tmpname, false, true)
```
winder, line 67: A constant is fine when the `inMemory` option is true, as it is here.

gen/generate.go
```diff
@@ -368,8 +374,9 @@ func generateGenesisFiles(outDir string, protoVersion protocol.ConsensusVersion,
 	jsonData := protocol.EncodeJSON(g)
 	err = ioutil.WriteFile(filepath.Join(outDir, config.GenesisJSONFile), append(jsonData, '\n'), 0666)
 
-	if (!verbose) && (rootKeyCreated > 0 || partKeyCreated > 0) {
+	if (verbose) && (rootKeyCreated > 0 || partKeyCreated > 0) {
```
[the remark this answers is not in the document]
Aharonee: In what way? Do you think this output should be written either way? I was under the assumption that this `verbose` flag could have been `!silent` all the same, since it is checked before writing any other output in this function.

crypto/merklesignature/persistentMerkleSignatureScheme_test.go
tsachiherman: this is only partially true : 
it's ok to have the same constant as long as there is only a single unit test running. Two unit tests that would use the same constant would end up using the same in-memory database, which is an issue.

ledger/apply/keyreg_test.go
```diff
@@ -31,6 +33,9 @@ import (
 
 var feeSink = basics.Address{0x7, 0xda, 0xcb, 0x4b, 0x6d, 0x9e, 0xd1, 0x41, 0xb1, 0x75, 0x76, 0xbd, 0x45, 0x9a, 0xe6, 0x42, 0x1d, 0x48, 0x6d, 0xa3, 0xd4, 0xef, 0x22, 0x47, 0xc4, 0x9, 0xa3, 0x96, 0xb8, 0x2e, 0xa2, 0x21}
 
+const DefaultParticipationFirstRound = 0
+const DefaultParticipationLastRound = 3000
```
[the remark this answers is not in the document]
Aharonee: It is not exportable since it is defined in a test file

crypto/merklesignature/posdivs.go
```diff
@@ -0,0 +1,55 @@
...
+	rofi := roundOfFirstIndex(firstValid, interval)
+	return (currentRound - rofi) / interval
+}
+
+func indexToRound(firstValid, interval, pos uint64) uint64 {
+	return roundOfFirstIndex(firstValid, interval) + pos*interval
+}
+
+func roundOfFirstIndex(firstValid, interval uint64) uint64 {
+	return ((firstValid + interval - 1) / interval) * interval
```
winder, line 54: nit: I think this is slightly clearer.
```suggestion
	return (firstValid % interval) + interval
```

data/account/participationRegistry.go
```diff
@@ -470,7 +517,11 @@ func (db *participationDB) insertInner(record Participation, id ParticipationID)
 		voting := record.Voting.Snapshot()
 		rawVoting = protocol.Encode(&voting)
 	}
-	// PKI TODO: Extract state proof from record.
+
+	// This contains all the state proof data except for the actual secret keys (stored in a different table)
+	if record.StateProofSecrets != nil {
+		rawStateProof = protocol.Encode(&record.StateProofSecrets.SignerContext)
```
winder, line 523: nit: You could rename `rawStateProof` to `rawStateProofVerifier` if thats more accurate.

crypto/merklesignature/posdivs.go
algoidan: Unfortunately, this is not accurate 
for example: 
firstvalid =1 lastvalid =9 and interval =2 
the old code will return 2. (which is the expected value)
the new code will return 3.

winder: You're right, I got mixed up while simplifying things. It would be something like this:
`firstValid - (firstValid % interval) + interval`

winder: they're both confusing, probably no use in changing it. Maybe a comment?

data/account/participationRegistry.go
```diff
@@ -873,17 +940,23 @@ func (db *participationDB) GetAll() []ParticipationRecord {
 	return results
 }
 
-// GetForRound fetches a record with all secrets for a particular round.
-func (db *participationDB) GetForRound(id ParticipationID, round basics.Round) (ParticipationRecordForRound, error) {
-	var result ParticipationRecordForRound
+func (db *participationDB) GetStateProofForRound(id ParticipationID, round basics.Round) (StateProofRecordForRound, error) {
```
winder, line 943: Missing doc string

data/bookkeeping/block.go
```diff
@@ -214,7 +214,7 @@ type (
 		// are a multiple of ConsensusParams.CompactCertRounds.  For blocks
 		// that are not a multiple of ConsensusParams.CompactCertRounds,
 		// this value is zero.
-		CompactCertVoters crypto.Digest `codec:"v"`
+		CompactCertVoters crypto.GenericDigest `codec:"v"`
```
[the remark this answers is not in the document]
Aharonee: `GenericDigest` is already bound: `crypto/digest.go:22`

data/account/participationRegistry.go
```diff
@@ -893,9 +966,53 @@ func (db *participationDB) GetForRound(id ParticipationID, round basics.Round) (
...
+	var rawSignerContext []byte
+	err = db.store.Rdb.Atomic(func(ctx context.Context, tx *sql.Tx) error {
+		// fetch stateproof public data
+		row := tx.QueryRow(selectStateProofData, id[:])
+		err := row.Scan(&rawSignerContext)
+		if err != nil {
+			return fmt.Errorf("error while querying stateproof data: %w", err)
+		}
+		return nil
+	})
```
winder, line 994: Is there a reason not to keep the public keyset data in the cache along with the rest of the `Keyset` data?

data/account/participationRegistry.go
winder: It looks like half this function is pretty much the same as `GetForRound`. Could it be simplified by calling that then fetching the missing state proof secret?

data/account/participationRegistry.go
Aharonee: I guess it is possible and might reduce runtime of this operation (which should happen once every 256 rounds as of now), but let's do some quick calculations to see the tradeoffs:
Every `SignerContext` object contains a `merklearray.Tree`, the tree is is bound by an upper limit of 2^16 leaves, meaning around 2^17 nodes in the entire tree.
Every node is basically a `sumhash` which is defined by `SumhashDigestSize` to be 64 bytes, so we're talking `2^17 * 64 ~= 8,388,000 bytes` which is basically 8.4 megabytes per instance. Say we have a few accounts on a participating node and it adds up to be quite large.

On the other hand, reading this from the disk every ~21 minutes (I think ?) for signing a Compact Certificate might also cause some hiccup as it could be an expensive IO operation, but at least it's not happening very often.

What do you think?

@algoidan can correct me if my calculations are off

libgoal/participation.go
```diff
@@ -155,6 +155,12 @@ func (c *Client) GenParticipationKeysTo(address string, firstValid, lastValid, k
 	if err != nil {
 		return
 	}
+	_, err = os.Stat(partKeyPath)
+	if !os.IsNotExist(err) {
+		err = fmt.Errorf("ParticipationKeys exist for the range %d to %d", firstRound, lastRound)
```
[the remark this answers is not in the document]
Aharonee: I suppose if `err == nil` then the file necessarily exists and we can return, and if `err!= nil` (whatever error that might be) we should just ignore it (or log? I think it'll be logged later if needed) and continue the function.
However, I need to consult @algoidan as I'm not sure if this takes into consideration the case where the file itself is corrupted (by the process of writing to it being terminated for example).

The reason for this change was mostly performance since generating the falcon keys takes around 8 minutes (if I recall correctly), but maybe the simplest solution is the best - just override whatever there is if it exists.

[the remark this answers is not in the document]
tsachiherman: so, that's not what I meant. I was thinking about this:
```golang
	_, err = os.Stat(partKeyPath)
	if !os.IsNotExist(err) {
		if err != nil {
			err = fmt.Errorf("participation key file '%s' cannot be accessed : %w", partKeyPath, err)
		} else {
			err = fmt.Errorf("ParticipationKeys exist for the range %d to %d", firstRound, lastRound)
		}
		return
	}
```

ledger/apply/keyreg_test.go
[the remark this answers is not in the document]
tsachiherman: not a blocker, but non-exportable fields/constants/functions should be lower case. We should use exportable notation only when we intend to do so.

crypto/merklesignature/merkleSignatureScheme.go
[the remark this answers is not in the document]
Aharonee: The `Root()` method returns a `GenericDigest` which is already defined to be of size `SumhashDigestSize`, same as `MerkleSignatureSchemeRootSize`. Is that not enough to feel confident in this copy?
I'm not sure what check we could add, since the way nodes (and the root) are stored in the tree is within slices of `SumhashDigestSize` anyway.

ledger/apply/keyreg_test.go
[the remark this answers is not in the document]
Aharonee: ok that makes sense

crypto/compactcert/committableSignatureSlot.go
Aharonee: They should be exported for the user of the function to be able to check the error type, right?
I'll definitely change it to a proper error instead of string though in both places

data/account/participationRegistry.go
```diff
@@ -797,6 +853,17 @@ func scanRecords(rows *sql.Rows) ([]ParticipationRecord, error) {
...
 
+		if len(rawStateProof) > 0 {
+			stateProof := merklesignature.Signer{}
+			err = protocol.Decode(rawStateProof, &stateProof.SignerContext)
+			if err != nil {
+				return nil, fmt.Errorf("unable to decode stateproof: %w", err)
+			}
+			var stateProofVerifer StateProofVerifier
+			copy(stateProofVerifer[:], stateProof.GetVerifier()[:])
+			record.StateProof = &stateProofVerifer
```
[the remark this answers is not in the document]
Aharonee: It is weird, as the `Keysets` table do not directly reflect the `ParticipationRecord` type.
The thing is, the `ParticipationRecord` should only contain the verifier, as it is the only part of the StateProof that's relevant to it.
However, when actually using the StateProof keys to sign, we require the rest of the data.

The way it is currently stored is such: the `stateProof BLOB` within the `Keysets` table is basically encoded `merklesignature.SignerContext`, which is the State Proof data without the keys themselves, as the keys are quite large and we prefer to store them separately for performance and security reasons.
The `FalconSigner` keys are stored within `StateProofKeys` table.

Since the `ParticipationRecord` does not reflect how data is actually stored in the database, we have to build it manually and extract only what's necessary - in case of stateproof the verifier, which is the root of the merkle tree built upon the keys.

Some database migration is inevitable either way in my opinion, since the voting keys should not be stored the way that they are, so we could use that migration to also update the way we handle the state proof data, but for now I'm not sure of a better way to do that.

What do you think?

data/account/participationRegistry.go
Aharonee: That's a good idea, `GetForRound` originally contained the logic for both but we've decided that we should only get the StateProof when request instead of every round, since there's some IO overhead and it should only be called once every 256 rounds (as well as caching being problematic as I've explained below).

data/account/participationRegistry.go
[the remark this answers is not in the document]
winder: Sounds good, I wanted to make sure it wasn't weird because there was some mistake!

A comment with a summary of this explanation might be helpful for the next person working on this code.

tsachiherman (approved):

PR #3009, opened 2021-10-06
Inner groups

jannotti:
Introduces `itxn_next` (feature flagged behind LogicVersion 6).  This allows contracts to create inner transaction _groups_, not just singleton transactions.  This is not of much value on its own, but is needed to make groups for inner app invocations.  Until then, the only difference between running two inner txns consecutively, and running two as part of a group is that fee pooling allows the second txn to pay for a shortfall from the first.


data/transactions/logic/eval.go
```diff
@@ -3676,33 +3676,58 @@ func authorizedSender(cx *EvalContext, addr basics.Address) bool {
...
+func opTxBegin(cx *EvalContext) {
+	if len(cx.subtxns) > 0 {
+		cx.err = errors.New("itxn_begin without itxn_submit")
+		return
+	}
+	cx.err = addInnerTxn(cx)
+}
+
+func opTxNext(cx *EvalContext) {
+	if len(cx.subtxns) == 0 {
```
jasonpaulos, line 3726: It would be a good idea to limit the size of `cx.subtxns` to the max atomic group size either here or in `addInnerTxn`. Otherwise a program can potentially waste memory by repeating `itxn_next` thousands of times without ever submitting.

data/transactions/logic/evalAppTxn_test.go
```diff
@@ -518,3 +518,65 @@ func TestFieldSetting(t *testing.T) {
...
+func TestInnerFeePooling(t *testing.T) {
+	ep, ledger := makeSampleEnv()
+	ledger.NewApp(ep.Txn.Txn.Receiver, 888, basics.AppParams{})
+	ledger.NewAccount(ledger.ApplicationID().Address(), 50_000)
+	pay := `
+int pay;    itxn_field TypeEnum;
+int 500;    itxn_field Amount;
+txn Sender; itxn_field Receiver;
+`
+	// Force the first fee to 2, but the second will default to 2*fee-3 = 2002-3
```
jasonpaulos, line 549: You mean force the first fee to 3?

data/transactions/logic/doc.go
```diff
@@ -169,9 +169,10 @@ var opDocByName = map[string]string{
 	"b~":  "X with all bits inverted",
 
 	"log":         "write bytes to log state of the current application",
-	"itxn_begin":  "begin preparation of a new inner transaction",
+	"itxn_begin":  "begin preparation of a new inner transaction in a new transaction group",
+	"itxn_next":   "begin preparation of a new inner transaction in the same transaction group",
```
jasonpaulos, line 173: An expanded description for this opcode would be helpful to explain when it's useful.

I.e., all inner transactions are atomic in the sense that if one fails, the entire top-level transaction fails. However, explicitly putting transactions into groups enables fee pooling and setting up the environment for inner transaction application calls.

test/scripts/e2e_subs/app-group.py
```diff
@@ -0,0 +1,78 @@
...
+assert app_id
+
+# Fund the app account
+txinfo, err = goal.pay(goal.account, goal.app_address(app_id), amt=400_000)
+assert not err, err
+
+
+txinfo, err = goal.app_call(joe, app_id, accounts=[goal.account])
+assert not err, err
+#assert goal.balance(joe) == joeb+6, goal.balance(joe)
```
jasonpaulos, line 76: No balance check?

data/transactions/logic/TEAL_opcodes.md
```diff
@@ -1318,7 +1318,7 @@ bitlen interprets arrays as big-endian integers, unlike setbit/getbit
 - Opcode: 0xb3
 - Pops: _None_
 - Pushes: _None_
-- execute the current inner transaction. Fail if 16 inner transactions have already been executed, or if the transaction itself fails.
+- execute the current inner transaction group. Fail if executing this group would exceed 16 total inner transactions, or if any transaction in the group fails.
```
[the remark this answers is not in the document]
jannotti: Yeah, I originally thought I'd be explicit about the whole thing, wrap a bunch of matched `itxn_begin` and `itxn_end` pairs with itxg_begin, itxg_submit.  But I couldn't justify all those opcodes once I started trying to write apps with them.

This seemed like the minimal change possible, and you can use itxn_submit to submit a single transaction, so I didn't want to waste an opcode on the distinction.  But it's not crazy if people think we should go that way.

jasonpaulos:
Additionally, the group ID of the inner txns is never set. I agree this isn't necessary from a technical perspective, but setting the ID for each submitted group txn would make it possible to distinguish between normal txns and group txns. This is probably useful for observers of the chain.

jannotti:
> Additionally, the group ID of the inner txns is never set. I agree this isn't necessary from a technical perspective, but setting the ID for each submitted group txn would make it possible to distinguish between normal txns and group txns. This is probably useful for observers of the chain.

I've been thinking about this a lot, and I don't have a perfect answer yet.  There are two interesting "fields" that appear in two "contexts".  The fields are TxId and GroupID.  The contexts are "as seen by AVM code" and "as seen on chain".

TxID does not appear on chain, but there is a well-defined way to determine it from on-chain data (the fields of the transaction).  GroupID does appear on chain (for top-level transactions) and it had better match the hashing scheme that defines it for the group members.  Both are accessible from AVM.

So, first for on-chain visibility, I can see some value in chain observers looking at a txn's list of inner-txns and being able to say "The first 3 were part of a group, then the next 2, and the last was a single transaction".  I don't think there's a _ton_ of such value, but it seems reasonable. That can be done in two ways. 1) Give each inner group a sequential group id. or 2) Give the group its "real" GroupID, based on the existing hashing rule.  The first is cheaper, the second seems more correct.  But notice that the second is still a little strange - you can get identical group IDs, even for adjacent groups, since there's no rule against duplicate transactions.  This would actually mean chain observers could have a hard time differentiating consecutive identical groups.

Now txid.  Since it never appears on chain, we don't have to worry too much about that.  But there have been calls for some sort of identifier for inner transactions. I have so far been against that, but we need to decide what `global TxID` does in an inner txn anyway.  If introduced, it would have to involve the parent txid, and the index of the inner transaction, so the resulting txid would be unique.  This is a departure from the current definition.  If that definition were exposed to AVM code, is there any code that might get confused because it tried to calculate the hash explicitly? Or because it intended to authenticate a transaction by looking at its txid?

So here is my current thinking.

In an inner transaction, with teal <= 5, using `global TxID` or `global GroupID` panics.  (In fact, there is some thought that calling teal <= 5 apps as inner transactions should be disallowed entirely.)

Starting from teal 6, `TxID` is defined to include the parent txid and index in the hash.  `GroupID` is defined similarly.  This makes them unique, but never exposes this new definition to old code.  Thoughts?

I've had some discussion of this with @fabrice102 and Shai.  Further input welcome.

For THIS PR, I'm not concerned yet, since itxn_next is behind Teal 6.

data/transactions/logic/TEAL_opcodes.md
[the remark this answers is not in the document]
algoanne: I personally like it the way you did it, I think the additional opcodes aren't worth the minimal increase in explicitness/clarity.

data/transactions/logic/eval.go
```diff
@@ -3676,33 +3676,63 @@ func authorizedSender(cx *EvalContext, addr basics.Address) bool {
...
+		return err
 	}
 
-	fee := cx.Proto.MinTxnFee
-	if cx.FeeCredit != nil {
-		// Use credit to shrink the fee, but don't change FeeCredit
-		// here, because they might never itxn_submit, or they might
-		// change the fee.  Do it in itxn_submit.
-		fee = basics.SubSaturate(fee, *cx.FeeCredit)
+	if len(cx.InnerTxns)+len(cx.subtxns) >= cx.Proto.MaxInnerTransactions ||
```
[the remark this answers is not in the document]
jannotti: Yes, I'll have to propagate down a sort of "tx budget" saying how many txns have been consumed already.

PR #3014, opened 2021-10-06
Allow all trackers committing into DB

algorandskiy:
## Summary

Historically only `account updates` tracker was committing into DB. The PR changes this:
1. `dbRound` management and `commitSyncer` business moved to trackerRegistry.
2.  `account updates` is used as a driver for committing tasks scheduling - by given `commttedRound` and `dbRound` it figures out offset in detlas for saving to the disk. This is basically old `au.committedUpTo` method`.
3. `commitRound` replaced by `prepareCommit` that returns commit and post commit callbacks
4. trackerRegistry has own `commitRound` that calls  trackers' `prepareCommit` and commit/post commit procs if any.
5. `account updates` still have a cached dbRound to be used in Lookup methods.

## Test Plan

It is refactoring, use existing tests

### DIFF
--- a/ledger/acctupdates.go
+++ b/ledger/acctupdates.go
@@ -1747,6 +1747,14 @@ func (au *accountUpdates) prepareCommit(dcc *deferredCommitContext) error {
 	// verify version correctness : all the entries in the au.versions[1:offset+1] should have the *same* version, and the committedUpTo should be enforcing that.
 	if au.versions[1] != au.versions[offset] {
 		au.accountsMu.RUnlock()
+
+		// in scheduleCommit, we expect that this function to update the catchpointWriting when
+		// it's on a catchpoint round and it's an archival ledger. Doing this in a deferred function
+		// here would prevent us from "forgetting" to update this variable later on.
+		// The same is repeated in commitRound on errors.
+		if dcc.isCatchpointRound && au.archivalLedger {
+			atomic.StoreInt32(&au.catchpointWriting, 0)
+		}
 		return fmt.Errorf("attempted to commit series of rounds with non-uniform consensus versions")
 	}
 	dcc.roundConsensusVersion = au.versions[1]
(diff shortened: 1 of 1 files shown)


ledger/bulletin.go
```diff
@@ -78,7 +78,7 @@ func (b *bulletin) Wait(round basics.Round) chan struct{} {
 	return signal.signal
 }
 
-func (b *bulletin) loadFromDisk(l ledgerForTracker) error {
+func (b *bulletin) loadFromDisk(l ledgerForTracker, dbRound basics.Round) error {
```
tsachiherman, line 81: dbRound -> _

ledger/tracker.go
```diff
@@ -98,19 +116,73 @@ type ledgerForTracker interface {
...
+	scheduleCommit(basics.Round)
+	waitAccountsWriting()
 }
 
 type trackerRegistry struct {
 	trackers []ledgerTracker
+	driver   *accountUpdates
+
+	// ctx is the context for the committing go-routine. It's also used as the "parent" of the catchpoint generation operation.
+	ctx context.Context
```
tsachiherman, line 130: I don't see that this context object is the parent of the catchpoint generation operation; did I miss it ?

ledger/tracker.go
```diff
@@ -98,19 +116,73 @@ type ledgerForTracker interface {
...
+
+	tr.ctx, tr.ctxCancel = context.WithCancel(context.Background())
+	tr.committedOffset = make(chan deferredCommit, 1)
+	tr.commitSyncerClosed = make(chan struct{})
+	go tr.commitSyncer(tr.committedOffset)
+
+	return
 }
 
 func (tr *trackerRegistry) register(lt ledgerTracker) {
```
tsachiherman, line 175: the `register` can be combined with the `initialize` method. In particular, you can just pass in a slice of `ledgerTracker` object pointers.

ledger/tracker.go
algorandskiy: I found out I do not need it. The `cancelFunc` is called only on `close()`, that calls tracker's `close()` that cancels catchpoint generation as well. I updated the comment

ledger/ledger.go
```diff
@@ -391,9 +399,19 @@ func (l *Ledger) notifyCommit(r basics.Round) basics.Round {
...
 }
 
+func (l *Ledger) waitAccountsWriting() {
+	l.trackers.waitAccountsWriting()
+}
+
+func (l *Ledger) scheduleCommit(rnd basics.Round) {
+	l.trackers.scheduleCommit(rnd)
+}
+
```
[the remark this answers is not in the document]
tsachiherman: I'm ok leaving it here for now - but it's not the right place for that. It looks to me as a "hint" that it doesn't really belong here. This part would "roll-forward" the database to be at the correct round ( i.e. x -> latest-320 ).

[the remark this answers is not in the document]
algorandskiy: I agree, it is part of account updates initialization, I did not figure out a better way to accessing upper level from an tracker.

tsachiherman (approved):
looks good.

PR #3016, opened 2021-10-07
Increase machine size to large and use consistent parallelism for circleci nightly testing.

egieseke:
## Summary

For CircleCI testing, increase machine size from medium to large and set parallelism to 4.

## Test Plan

Testing by running nightly builds on the branch.


algobarb (approved):
Same as what Jack said - not sure we need large machines and parallelism for all, but approving this to be included for 3.1.0 beta release.

PR #3026, opened 2021-10-08
testing: fix bug in TestRekeyUpgrade

tsachiherman:
## Summary

This test has a special handling is case the protocol upgrade completes before we can validate that without an upgrade the rekey feature won't work. However, when that does happen, AccountA is already rekeyed, and would require a different handling.

To avoid that while retaining the execution of the test, I've moved the second part to use a different set of accounts.

## Test Plan

This is a test


algonautshant (approved):
LGTM

PR #3028, opened 2021-10-09
network: add initial support for latency tracking

tsachiherman:
## Summary

This PR adds the node ( both client and server ), the ability to measure the time it takes to establish an outgoing connection ( excluding the TCP connection time ).
This duration is captured as the initial latency, which would need to get updated via a pingpong style logic.

## Test Plan

- [x] Extend existing unit tests
- [x] mainnet-model testing is needed as well to confirm correctness

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -5,22 +5,33 @@ orbs:
   go: circleci/go@1.7.0
   slack: circleci/slack@4.4.2
 
+parameters:
+  ubuntu_image:
+    type: string
+    default: "ubuntu-2004:202104-01"
+  build_dir:
+    type: string
+    default: "/opt/cibuild"
+  result_path:
+    type: string
+    default: "/tmp/build_test_results"
+
 executors:
   amd64_medium:
     machine:
-      image: ubuntu-2004:202104-01
+      image: << pipeline.parameters.ubuntu_image >>
     resource_class: medium
   amd64_large:
     machine:
-      image: ubuntu-2004:202104-01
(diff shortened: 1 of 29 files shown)


tsachiherman:
> wsPeer already has `lastPingRoundTripTime` ?

Yes.. this is only a POC. Clearly we won't need to have both when the solution is ready. My goal here was just to do the minimum so that we could measure the connection timing and determine if that aligns with the latencies we're configuring in the Mainnet-model.

config/config.go
```diff
@@ -84,7 +84,7 @@ type Local struct {
 	MaxConnectionsPerIP int `version[3]:"30"`
 
 	// 0 == disable
-	PeerPingPeriodSeconds int `version[0]:"0"`
+	PeerPingPeriodSeconds int `version[0]:"0" version[18]:"10"`
```
nicholasguoalgorand, line 87: Is this a parameter that was used in the past but is currently not being used?

network/wsPeer.go
```diff
@@ -639,7 +651,8 @@ func (wp *wsPeer) writeLoopSendMsg(msg sendMessage) disconnectReason {
 	}
 
 	// check if this message was waiting in the queue for too long. If this is the case, return "true" to indicate that we want to close the connection.
-	msgWaitDuration := time.Now().Sub(msg.enqueued)
+	now := time.Now()
+	msgWaitDuration := now.Sub(msg.enqueued)
```
nicholasguoalgorand, line 655: whats the difference in this change?

PR #3030, opened 2021-10-10
ledger: move StartEvaluator parameters into an object

tsachiherman:
## Summary

move StartEvaluator parameters into an object. Given that most of the time the parameters are using a "default" state, it reduce the number of passed parameters, improving the caller readability.

## Test Plan

Unit tests updated.

### DIFF
--- a/ledger/evalindexer.go
+++ b/ledger/evalindexer.go
@@ -45,7 +45,11 @@ type indexerLedgerForEval interface {
 // EvalForIndexerResources contains resources preloaded from the Indexer database.
 // Indexer is able to do the preloading more efficiently than the evaluator loading
 // resources one by one.
-type EvalForIndexerResources = internal.EvalForIndexerResources
+type EvalForIndexerResources struct {
+	// The map value is nil iff the account does not exist. The account data is owned here.
+	Accounts map[basics.Address]*basics.AccountData
+	Creators map[Creatable]ledgercore.FoundAddress
+}
 
 // Creatable represent a single creatable object.
 type Creatable = internal.Creatable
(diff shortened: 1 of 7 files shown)


ledger/internal/eval.go
```diff
@@ -94,15 +94,32 @@ type roundCowBase struct {
...
+	if preloadedAccounts != nil {
+		rcb.initializeAccountsCache(preloadedAccounts)
+	}
+	return rcb
+}
+
+func (x *roundCowBase) initializeAccountsCache(accts map[basics.Address]*basics.AccountData) {
+	for address, accountData := range accts {
+		if accountData == nil {
+			x.accounts[address] = basics.AccountData{}
```
algorandskiy, line 119: why can we have nil entries on init?

ledger/evalindexer.go
```diff
@@ -150,13 +154,17 @@ func EvalForIndexer(il indexerLedgerForEval, block *bookkeeping.Block, proto con
 	ilc := makeIndexerLedgerConnector(il, block.GenesisHash(), block.Round()-1)
 
 	eval, err := internal.StartEvaluator(
-		ilc, block.BlockHeader, proto, len(block.Payset), false, false, 0)
+		ilc, block.BlockHeader,
+		internal.EvaluatorOptions{
+			PaysetHint:        len(block.Payset),
```
algorandskiy, line 159: maybe having explicit Vaildate/Generate values is better/clearer

ledger/internal/eval.go
tsachiherman: The indexer would preload the touched accounts. If an account doesn't exist, it would place `nil` instead of an empty AccountData{}.

algorandskiy: can't we change this code in the indexer preloaded and get rid of `initializeAccountsCache` func from here?

tsachiherman: you mean implement the caching layer in `indexerLedgerConnector` instead ?

PR #3031, opened 2021-10-10
batch verification: add batch verification implementation

algoidan:
## Summary
In this change, we plan to integrate the ed25519  batch verification algorithm. This will speed up the process of verifying the digital signature of several transactions by roughly 2.4X.

This PR implements the algorithm according to the following paper:
https://eprint.iacr.org/2020/1244.pdf
This PR fixes a **bug** in the paper (Listing 1.2) that didn't reject the points (-0,-1) and (-0,1)

Another blog article with similar background: https://hdevalence.ca/blog/2020-10-04-its-25519am

**Changes Overview**
Changes made on the single ed255519 verification :
1. Signatures that contain small order group R value will now be accepted.
2. use the cofactor version of the equation to validate the signature. (multiple the result by the cofactor and compare to natural element)
3. add ge25519_is_canonical_vartime function (check if a given y point has a canonical or non-canonical encoding  ): 
    - optimize this function using the the quick check.
    - reject two non-canonical points by comparison (points (-0,-1) and (-0,1), this resolves a bug in the paper). 

Add the batch verification implementation from https://github.com/floodyberry/ed25519-donna to our fork of libsodium.
Changes made on the batch verification (to match the single verification):
1. reject small order A (PK)
2. reject non-canonical A (PK)
3. reject non-canonical R
4. reject non-canonical S
5. use the cofactor equation 

## Test Plan

The following tests were added to the libsodium's test suite.
1. Test against the White Paper's test vector
2. Test against all non-canonical options for R
3. Test against all non-canonical options for S


crypto/libsodium-fork/src/libsodium/crypto_sign/ed25519/ref10/batch.c
```diff
@@ -0,0 +1,318 @@
...
+int crypto_sign_ed25519_open_batch(const unsigned char **m, unsigned long long *mlen, const unsigned char **pk, const unsigned char **RS, size_t num, int *valid)
+{
+	batch_heap batch;
+	ge25519_p3  p;
+	ge25519_p3  p_mul;
+	sc25519 *r_scalars;
+	size_t i, batchsize;
+	unsigned char hram[64];
+	unsigned char batchsum[32];
+    int ret = 0;
```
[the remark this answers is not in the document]
cce: agreed, the mix of tabs and spaces makes it hard to follow. The rest of libsodium seems to use spaces (no tabs) with a indentation level of 4 spaces

crypto/libsodium-fork/src/libsodium/crypto_core/ed25519/ref10/ed25519_ref10.c
```diff
@@ -2583,3 +2692,17 @@ ge25519_from_uniform(unsigned char s[32], const unsigned char r[32])
 
     ge25519_p3_tobytes(s, &p3);
 }
+
+int
+sc25519_is_canonical_vartime(const unsigned char s[32])
+{
+    if ((s[31] & 240) == 0)
```
cce, line 2699: This is the same thing as https://github.com/jedisct1/libsodium/commit/61368716075992568a0adcfd696359e206827017#diff-4db2d523f04cf0c485674b4e3a5f4f15df23d9bc4f33a6d4b4d701402912141d right?

crypto/libsodium-fork/src/libsodium/crypto_core/ed25519/ref10/ed25519_ref10.c
```diff
@@ -50,6 +50,11 @@ load_4(const unsigned char *in)
 # include "fe_25_5/fe.h"
 #endif
 
+const ge25519_p3 ge25519_basepoint = 
+{
+#include "fe_51/base_ge25519.h"
+};
```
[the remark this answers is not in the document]
algoidan: agree, but this technique appears in this library (line 583). so I've tried to make it look the same.

crypto/libsodium-fork/src/libsodium/crypto_core/ed25519/ref10/ed25519_ref10.c
```diff
@@ -643,6 +648,53 @@ ge25519_tobytes(unsigned char *s, const ge25519_p2 *h)
...
+    fe25519 u;
+    fe25519 v;
+    fe25519 v3;
+    fe25519 vxx;
+    fe25519 m_root_check, p_root_check;
+
+    fe25519_frombytes(h->Y, s);
+    fe25519_1(h->Z);
+    fe25519_sq(u, h->Y);
+    fe25519_mul(v, u, d);
```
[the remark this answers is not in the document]
algoidan: do want us to rename it as well?

crypto/libsodium-fork/src/libsodium/crypto_core/ed25519/ref10/ed25519_ref10.c
algoidan: yeah
if think ours is clearer

crypto/libsodium-fork/src/libsodium/crypto_sign/ed25519/ref10/batch.c
```diff
@@ -0,0 +1,318 @@
...
+        ge25519_multi_scalarmult_vartime(&p, &batch, (batchsize * 2) + 1);
+
+		ge25519_mul_by_cofactor(&p_mul, &p);
+		if (ge25519_is_neutral_vartime(&p_mul) == 0 ) {
+			ret |= 2;
+
+			fallback:
+			for (i = 0; i < batchsize; i++) {
+				valid[i] = crypto_sign_ed25519_verify_detached(RS[i], m[i], mlen[i], pk[i]) ? 0 : 1;
+				ret |= (valid[i] ^ 1);
```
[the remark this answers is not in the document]
algoidan: valid[i] can only take 1 or 0 because of the " x ? y: z"  operator.
Am I missing something?

algoidan:
it looks like the batch verification interface uses unsigned long long *.
@cce  @tsachiherman, do you think we should change it to size_t?

crypto/libsodium-fork/src/libsodium/crypto_sign/ed25519/ref10/batch.c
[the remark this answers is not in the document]
cce: ohhhh I didn't see the ? 0 : 1 at the end. Thanks!

crypto/libsodium-fork/src/libsodium/crypto_core/ed25519/ref10/ed25519_ref10.c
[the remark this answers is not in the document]
cce: No, not necessary, but it was just confusing trying to figure out where `d` came from

crypto/libsodium-fork/test/default/sign.c
```diff
@@ -1057,6 +1057,337 @@ static void add_l(unsigned char * const S)
...
+static const  char non_canoical_public_key[][3][500]= 
+{
+{"9bedc267423725d473888631ebf45988bad3db83851ee85c85e241a07d148b41",
+ "0100000000000000000000000000000000000000000000000000000000000080", // (-0,1)
+"a9d55260f765261eb9b84e106f665e00b867287a761990d7135963ee0a7d59dca5bb704786be79fc476f91d3f3f89b03984d8068dcf1bb7dfc6637b45450ac04"} ,
+{"9bedc267423725d473888631ebf45988bad3db83851ee85c85e241a07d148b41",
+ "ecffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff", // (-0,-1)
+"a9d55260f765261eb9b84e106f665e00b867287a761990d7135963ee0a7d59dca5bb704786be79fc476f91d3f3f89b03984d8068dcf1bb7dfc6637b45450ac04"} ,
+{"9bedc267423725d473888631ebf45988bad3db83851ee85c85e241a07d148b41",
+ "eeffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff7f", // (0,2^252-18)
```
cce, line 1198: very small nit, 2^255-18 (and a few more below too)

crypto/libsodium-fork/test/default/sign.c
```diff
@@ -1057,6 +1057,337 @@ static void add_l(unsigned char * const S)
     sodium_add(S, l, sizeof l);
 }
 
+
+
+// Test vector according to : "Taming the many EdDSAs" white paper
```
cce, line 1062: in case you're looking for more test vectors, I found some in https://gist.github.com/hdevalence/93ed42d17ecab8e42138b213812c8cc7

crypto/libsodium-fork/src/libsodium/crypto_sign/ed25519/ref10/batch.c
```diff
@@ -0,0 +1,318 @@
...
+        for (i = 0; i < batchsize; i++)
+            if (ge25519_frombytes_negate_vartime(&batch.points[i+1], pk[i]) != 0)
+                goto fallback;
+        for (i = 0; i < batchsize; i++)
+            if (ge25519_frombytes_negate_vartime(&batch.points[batchsize+i+1], RS[i]) != 0)
+                goto fallback;
+
+        ge25519_multi_scalarmult_vartime(&p, &batch, (batchsize * 2) + 1);
+
+        ge25519_mul_by_cofactor(&p_mul, &p);
```
cce, line 291: In diffing against the libdonna implementation this is the only significant difference... how did it work before in libdonna? I thought you had to use the cofactor equation for batch verification.

crypto/libsodium-fork/src/libsodium/crypto_core/ed25519/ref10/ed25519_ref10_sc.c
```diff
@@ -0,0 +1,299 @@
+#include "private/ed25519_ref10_sc.h"
+
+#include <string.h>
+
+
+typedef unsigned __int128 uint128_t;
+#define HAVE_UINT128
```
cce, line 7: Diffed this file against libdonna modm-donna-64bit.h, all looks the same. This HAVE_UINT128 define is not needed, however, you brought over all the code that would have checked it without the ifdefs.

crypto/libsodium-fork/src/libsodium/crypto_sign/ed25519/ref10/open.c
```diff
@@ -10,6 +10,30 @@
...
+    if (ge25519_is_canonical_vartime(pk) == 0 ||
+        ge25519_has_small_order(pk) != 0) {
+        return -1;
+    }
+#endif
+    return 0;
+}
+
 int
 _crypto_sign_ed25519_verify_detached(const unsigned char *sig,
```
cce, line 38: Consider splitting out into another exported function.

crypto/libsodium-fork/src/libsodium/crypto_sign/ed25519/ref10/batch.c
```diff
@@ -0,0 +1,318 @@
...
+    sc25519 *r_scalars;
+    size_t i, batchsize;
+    unsigned char hram[64];
+    unsigned char batchsum[32];
+    int ret = 0;
+
+    for (i = 0; i < num; i++)
+        valid[i] = 1;
+
+    while (num > 3) {
```
cce, line 242: Only using batch verification when there are 4 or more signatures is kind of arbitrary, but I assume there was some calculation of the performance tradeoffs of batch vs single for small batches in the original implementation?

cce (commented):
generally LGTM, with some suggestions for changes

crypto/libsodium-fork/src/libsodium/crypto_sign/ed25519/ref10/batch.c
cce: I found some background here: https://gitlab.torproject.org/tpo/core/tor/-/issues/40078
>Batch verification is only used when the batch size is greater than 3, because this is the crossover point in performance for donna's implementation (but not for other implementations, where batches of size 1 can be just as fast as single verification).

crypto/libsodium-fork/src/libsodium/crypto_sign/ed25519/ref10/batch.c
```diff
@@ -0,0 +1,318 @@
...
+#include "private/ed25519_ref10_sc.h"
+#include "crypto_sign.h"
+
+
+#include <limits.h>
+#include <stdint.h>
+#include <string.h>
+
+
+void ed25519_randombytes_unsafe(void *p, size_t len);
```
cce, line 13: Where is this implemented? I only see test implementations — does there need to be one in non-test code that calls randombytes_buf?

algoidan: this function is implemented in the go code.
so we can use the same random source

crypto/libsodium-fork/src/libsodium/crypto_sign/ed25519/ref10/batch.c
algoidan: if I recall correctly, libdonna didn't use the cofactor version. that is one of the reasons why we need to change the implementation

crypto/libsodium-fork/src/libsodium/crypto_sign/ed25519/ref10/open.c
algoidan: how would you suggest I should split it?
could it be reused?
I guess I'm missing somthing.

crypto/libsodium-fork/src/libsodium/crypto_sign/ed25519/ref10/batch.c
algoidan: good thing we brought this up.
this code performs better even for 2 signatures.
I will reduce the number.

crypto/libsodium-fork/src/libsodium/crypto_sign/ed25519/ref10/batch.c
cce: right OK! I now understand better that this was part of the Tor issue https://gitlab.torproject.org/tpo/core/tor/-/issues/40078

crypto/libsodium-fork/src/libsodium/crypto_sign/ed25519/ref10/batch.c
cce: oh, OK. I didn't see anything from grepping for `ed25519_randombytes_unsafe` in the repo, but I guess `crypto_sign_ed25519_open_batch` is not yet wired up to Go yet

crypto/libsodium-fork/src/libsodium/crypto_sign/ed25519/ref10/batch.c
cce: cool, so even for evaluating individual multisig transactions this could help, neat!

crypto/libsodium-fork/src/libsodium/crypto_sign/ed25519/ref10/batch.c
```diff
@@ -0,0 +1,318 @@
...
+/*
+* verifies ed25519 signatures in  batch. The algorithm is based on https://github.com/floodyberry/ed25519-donna 
+* implemention. we changed the algorithm according to https://eprint.iacr.org/2020/1244.pdf .
+* the batch size is between 3 and 64 sigantures per batch. 
+* When the batch fails the function falls back to check singature one at a time.
+* the function returns 0 on success and fills and array of "valid" ints. 
+* 1 - for signature i passed verification
+* 0 - for siganture i failed verification
+*/
+int crypto_sign_ed25519_open_batch(const unsigned char **m, unsigned long long *mlen, const unsigned char **pk, const unsigned char **RS, size_t num, int *valid)
```
cce, line 228: just noticed this: should to `const unsigned long long *mlen` to match the other const input args here. also it seems to be the convention in libsodium to specify `_p` suffixes for output arg pointers like `int *valid_p`.

crypto/libsodium-fork/src/libsodium/crypto_sign/ed25519/ref10/sign_ed25519_ref10.h
```diff
@@ -15,4 +17,13 @@ int _crypto_sign_ed25519_verify_detached(const unsigned char *sig,
                                          unsigned long long   mlen,
                                          const unsigned char *pk,
                                          int prehashed);
+
+
+
+int _crypto_sign_ed25519_open_batch(const unsigned char **m, 
+                                unsigned long long *mlen,
```
cce, line 24:
```suggestion
                                const unsigned long long *mlen, 
```

crypto/libsodium-fork/src/libsodium/include/sodium/crypto_sign.h
```diff
@@ -63,12 +63,22 @@ int crypto_sign(unsigned char *sm, unsigned long long *smlen_p,
...
+            __attribute__ ((warn_unused_result)) __attribute__ ((nonnull(1, 2)));
+
 SODIUM_EXPORT
 int crypto_sign_open(unsigned char *m, unsigned long long *mlen_p,
                      const unsigned char *sm, unsigned long long smlen,
                      const unsigned char *pk)
             __attribute__ ((warn_unused_result)) __attribute__ ((nonnull(3, 5)));
 
+SODIUM_EXPORT
+int crypto_sign_ed25519_open_batch(const unsigned char **m, unsigned long long *mlen,
```
cce, line 77:
```suggestion
int crypto_sign_ed25519_open_batch(const unsigned char **m, const unsigned long long *mlen, 
```

crypto/libsodium-fork/src/libsodium/include/sodium/crypto_sign.h
```diff
@@ -63,12 +63,22 @@ int crypto_sign(unsigned char *sm, unsigned long long *smlen_p,
...
 SODIUM_EXPORT
 int crypto_sign_open(unsigned char *m, unsigned long long *mlen_p,
                      const unsigned char *sm, unsigned long long smlen,
                      const unsigned char *pk)
             __attribute__ ((warn_unused_result)) __attribute__ ((nonnull(3, 5)));
 
+SODIUM_EXPORT
+int crypto_sign_ed25519_open_batch(const unsigned char **m, unsigned long long *mlen, 
+                                const unsigned char **pk, const unsigned char **RS, 
+                                size_t num, int *valid)
```
cce, line 79:
```suggestion
                                size_t num, int *valid_p)
```

crypto/libsodium-fork/src/libsodium/crypto_sign/ed25519/ref10/sign_ed25519_ref10.h
```diff
@@ -15,4 +17,13 @@ int _crypto_sign_ed25519_verify_detached(const unsigned char *sig,
...
                                          int prehashed);
+
+
+
+int _crypto_sign_ed25519_open_batch(const unsigned char **m, 
+                                unsigned long long *mlen, 
+                                const unsigned char **pk, 
+                                const unsigned char **RS, 
+                                size_t num, 
+                                int *valid);
```
cce, line 28:
```suggestion
                                int *valid_p);
```

cce (commented):
> it looks like the batch verification interface uses unsigned long long *. @cce @tsachiherman, do you think we should change it to size_t?

It looks like they use unsigned long long for message lengths throughout most of the libsodium API (and size_t for some memory limits and some hashing APIs) so LGTM to keep mlen as unsigned long long to match the rest of the crypto_sign API.

crypto/curve25519.go
```diff
@@ -123,6 +123,17 @@ func ed25519Verify(public ed25519PublicKey, data []byte, sig ed25519Signature) b
 	return result == 0
 }
 
+func ed25519VerifyBatchVerificationCompatibleVersion(public ed25519PublicKey, data []byte, sig ed25519Signature) bool {
```
cce, line 126: I think you should merge this with the ed25519Verify function, that way you can be sure to catch everything using crypto_sign_ed25519_verify_detached at the lowest level.

cce: I implemented it — caught some more signature verifications and wired them up to the consensus param for signatures in votes and compact certs: https://github.com/algoidan/go-algorand/pull/14/files

cce:
While grepping to figure out why the new function `crypto_sign_ed25519_bv_compatible_verify_detached` was highlighted in red in my editor, I noticed there is a whole separate copy of our libsodium headers in the crypto/include/sodium directory. These are separate from the ones in crypto/libsodium-fork/src/libsodium/include/sodium and look to have been used for `#cgo CFLAGS: -I${SRCDIR}/include` in curve25519.go before #835 switched to using platform-specific include directories.

They are now out of sync with libsodium-fork — they look unnecessary though.

crypto/batchverifier.go
```diff
@@ -33,23 +58,42 @@ var (
...
+//export ed25519_randombytes_unsafe
+func ed25519_randombytes_unsafe(p unsafe.Pointer, len C.size_t) {
+	randLen := int(len)
+	rand := make([]byte, randLen)
+	RandBytes(rand)
+	for i := 0; i < randLen; i++ {
+		*(*byte)(unsafe.Pointer(uintptr(p) + uintptr(i*C.sizeof_uchar))) = rand[i]
+	}
+
+}
```
tsachiherman, line 70: would the following work ? 
```go
func ed25519_randombytes_unsafe2(p unsafe.Pointer, len C.size_t) {
	randBuf := (*[1 << 30]byte)(unsafe.Pointer(p))[:len:len]
	RandBytes(randBuf)
}
```

i.e. avoid the allocation and copy.

crypto/batchverifier.go
```diff
@@ -33,23 +58,42 @@ var (
...
+func MakeBatchVerifierWithAlgorithmDefaultSize() *BatchVerifier {
+	return MakeBatchVerifier(minBatchVerifierAlloc, true)
+}
+
 // MakeBatchVerifierDefaultSize create a BatchVerifier instance. This function pre-allocates
-// amount of free space to enqueue signatures without exapneding
-func MakeBatchVerifierDefaultSize() *BatchVerifier {
-	return MakeBatchVerifier(minBatchVerifierAlloc)
+// amount of free space to enqueue signatures without expanding
+func MakeBatchVerifierDefaultSize(shouldUseBatchVerification bool) *BatchVerifier {
```
tsachiherman, line 81: nit: align with protocol parameter naming; i.e. shouldUseBatchVerification -> enableBatchVerification

crypto/batchverifier.go
```diff
@@ -88,11 +132,67 @@ func (b *BatchVerifier) Verify() error {
...
 
+	if b.useBatchVerification {
+		var messages = make([][]byte, b.GetNumberOfEnqueuedSignatures())
+		for i, m := range b.messages {
+			messages[i] = hashRep(m)
+		}
+		if batchVerificationImpl(messages, b.publicKeys, b.signatures) {
+			return nil
+		}
+		return fmt.Errorf("erro checking batch")
```
tsachiherman, line 143: this is a "constant" error. Pre-define a global error, and fix the typo.

crypto/batchverifier.go
```diff
@@ -88,11 +132,67 @@ func (b *BatchVerifier) Verify() error {
...
 		verifier := SignatureVerifier(b.publicKeys[i])
-		if !verifier.Verify(b.messages[i], b.signatures[i]) {
+		if !verifier.Verify(b.messages[i], b.signatures[i], false) {
 			return ErrBatchVerificationFailed
 		}
 	}
 	return nil
 }
+
+// DonnaBatchVerification calls the batch verification implemention in C. it prepares the arguments
```
tsachiherman, line 158: `DonnaBatchVerification` -> `batchVerificationImpl`

crypto/curve25519.go
```diff
@@ -123,6 +123,17 @@ func ed25519Verify(public ed25519PublicKey, data []byte, sig ed25519Signature) b
...
+	// &data[0] will make Go panic if msg is zero length
+	d := (*C.uchar)(C.NULL)
+	if len(data) != 0 {
+		d = (*C.uchar)(&data[0])
+	}
+
+	result := C.crypto_sign_ed25519_bv_compatible_verify_detached((*C.uchar)(&sig[0]), d, C.ulonglong(len(data)), (*C.uchar)(&public[0]))
```
tsachiherman, line 133: could you have some type of GC testing to make sure that the GC won't collect `d` prematurely ?

crypto/libsodium-fork/src/libsodium/crypto_core/ed25519/ref10/ed25519_ref10.c
```diff
@@ -738,6 +794,101 @@ ge25519_double_scalarmult_vartime(ge25519_p2 *r, const unsigned char *a,
...
+    ge25519_p1p1_to_p3(&u, &t);
+    ge25519_p3_to_cached(&Ai[6], &u);
+
+    ge25519_add(&t, &A2, &Ai[6]);
+    ge25519_p1p1_to_p3(&u, &t);
+    ge25519_p3_to_cached(&Ai[7], &u);
+
+    ge25519_p3_0(r);
+
+    for (i = 255; i >= 0; --i) {
```
tsachiherman, line 864: scanning an array in reverse works really poorly on modern cpus; modern cpus are optimized for forward-scanning of arrays.

crypto/libsodium-fork/src/libsodium/crypto_core/ed25519/ref10/ed25519_ref10.c
```diff
@@ -982,6 +1152,32 @@ ge25519_is_on_main_subgroup(const ge25519_p3 *p)
     return fe25519_iszero(pl.X);
 }
 
+int
+ge25519_is_canonical_vartime(const unsigned char *s)
```
tsachiherman, line 1156: I'm pretty confident that this implementation could be improved by avoiding using the memcmp, and making explicit comparison. On modern 64-bit cpu, we really need only 4 comparison for each of the addresses.

crypto/libsodium-fork/src/libsodium/crypto_core/ed25519/ref10/ed25519_ref10.c
```diff
@@ -999,6 +1195,18 @@ ge25519_is_canonical(const unsigned char *s)
     return 1 - (c & d & 1);
 }
 
+
+int ge25519_is_neutral_vartime(const ge25519_p3 *p) {
```
tsachiherman, line 1199: the performance of this method could be improved; the call to `fe25519_tobytes` along with the temporary static memory writings could be completely avoided.

crypto/libsodium-fork/src/libsodium/crypto_core/ed25519/ref10/ed25519_ref10_sc.c
```diff
@@ -0,0 +1,298 @@
...
+	r[0] ^= mask & (r[0] ^ t[0]);
+	r[1] ^= mask & (r[1] ^ t[1]);
+	r[2] ^= mask & (r[2] ^ t[2]);
+	r[3] ^= mask & (r[3] ^ t[3]);
+	r[4] ^= mask & (r[4] ^ t[4]);
+}
+
+void barrett_reduce256_modm(sc25519 r, const sc25519 q1, const sc25519 r1)
+{
+    sc25519 q3, r2;
```
tsachiherman, line 90: tab ?

crypto/libsodium-fork/src/libsodium/crypto_core/ed25519/ref10/ed25519_ref10_sc.c
```diff
@@ -0,0 +1,298 @@
...
+	memcpy(work, in, len);
+	x[0] = U8TO64_LE(work +  0);
+	x[1] = U8TO64_LE(work +  8);
+	x[2] = U8TO64_LE(work + 16);
+	x[3] = U8TO64_LE(work + 24);
+	x[4] = U8TO64_LE(work + 32);
+	x[5] = U8TO64_LE(work + 40);
+	x[6] = U8TO64_LE(work + 48);
+	x[7] = U8TO64_LE(work + 56);
+
```
tsachiherman, line 158: the way it's currently implemented forces the CPU to perform the worst in either scenarios:
1. if we pass len=64, then the CPU would initialize the `work` with zeros redundantly.
2. if we pass len=16, then the CPU would compute `U8TO64_LE` redundantly.

crypto/libsodium-fork/src/libsodium/crypto_core/ed25519/ref10/ed25519_ref10_sc.c
```diff
@@ -0,0 +1,298 @@
...
+	return 1;
+}
+
+/* is a == 1 */
+int isone256_modm_batch(const sc25519 a) {
+	size_t i;
+	for (i = 0; i < 5; i++)
+		if (a[i] != ((i) ? 0 : 1))
+			return 0;
+	return 1;
```
tsachiherman, line 285: I want to hope the compiler would know how to properly optimize this part, but I'm not sure if it would.

crypto/libsodium-fork/src/libsodium/crypto_sign/ed25519/ref10/batch.c
```diff
@@ -0,0 +1,318 @@
...
+static void
+heap_insert_next(batch_heap *heap) {
+    size_t node = heap->size, parent;
+    heap_index_t *pheap = heap->heap;
+    sc25519 *scalars = heap->scalars;
+
+    /* insert at the bottom */
+    pheap[node] = (heap_index_t)node;
+
+    /* sift node up to its sorted spot */
```
tsachiherman, line 54: typo: sift -> shift

crypto/libsodium-fork/src/libsodium/crypto_sign/ed25519/ref10/batch.c
```diff
@@ -0,0 +1,318 @@
...
+    childr = 2;
+    while ((childr < heap->size)) {
+        node = lt256_modm_batch(scalars[pheap[childl]], scalars[pheap[childr]], limbsize) ? childr : childl;
+        heap_swap(pheap, parent, node);
+        parent = node;
+        childl = (parent * 2) + 1;
+        childr = childl + 1;
+    }
+
+    /* sift root back up to its sorted spot */
```
tsachiherman, line 84: sift -> shift

crypto/libsodium-fork/src/libsodium/crypto_sign/ed25519/ref10/batch.c
```diff
@@ -0,0 +1,318 @@
...
+    /* start with the full limb size */
+    size_t limbsize = SC25519_LIMB_SIZE - 1;
+
+    /* whether the heap has been extended to include the 128 bit scalars */
+    int extended = 0;
+
+    /* grab an odd number of scalars to build the heap, unknown limb sizes */
+    heap_build(heap, ((count + 1) / 2) | 1);
+
+    for (;;) {
```
tsachiherman, line 188: `for (;;)` -> `while(true)`

crypto/libsodium-fork/src/libsodium/crypto_sign/ed25519/ref10/batch.c
```diff
@@ -0,0 +1,318 @@
...
+
+    for (i = 0; i < num; i++)
+        valid_p[i] = 1;
+
+    while (num >= 2) {
+        batchsize = (num > MAX_BATCH_SIZE) ? MAX_BATCH_SIZE : num;
+
+        /* validate the public key and signature */
+        for (i=0; i < batchsize; i++){
+            if (validate_ed25519_pk_and_sig(RS[i],pk[i]) != 0)
```
tsachiherman, line 247: add space after `RS[i],`

crypto/libsodium-fork/src/libsodium/crypto_sign/ed25519/ref10/batch.c
```diff
@@ -0,0 +1,318 @@
...
+    int ret = 0;
+
+    for (i = 0; i < num; i++)
+        valid_p[i] = 1;
+
+    while (num >= 2) {
+        batchsize = (num > MAX_BATCH_SIZE) ? MAX_BATCH_SIZE : num;
+
+        /* validate the public key and signature */
+        for (i=0; i < batchsize; i++){
```
tsachiherman, line 246: add space after `)`

crypto/libsodium-fork/src/libsodium/crypto_sign/ed25519/ref10/batch.c
```diff
@@ -0,0 +1,318 @@
...
+            if (validate_ed25519_pk_and_sig(RS[i],pk[i]) != 0)
+                goto fallback;
+        }
+
+        /* generate r (scalars[batchsize+1]..scalars[2*batchsize] */
+        ed25519_randombytes_unsafe (batch.r, batchsize * 16);
```
tsachiherman, line 252: this isn't incorrect, however, I would suggest that you'll pre-computer the random bytes on a single call outside the `while` loop, and use the results of that call here.

crypto/libsodium-fork/src/libsodium/crypto_sign/ed25519/ref10/batch.c
```diff
@@ -0,0 +1,318 @@
...
+            expand256_modm(r_scalars[i], batch.r[i], 16);
+
+        /* compute scalars[0] = ((r1s1 + r2s2 + ...)) */
+        for (i = 0; i < batchsize; i++) {
+            expand256_modm(batch.scalars[i], RS[i] + 32, 32);
+            mul256_modm(batch.scalars[i], batch.scalars[i], r_scalars[i]);
+        }
+        for (i = 1; i < batchsize; i++)
+            add256_modm(batch.scalars[0], batch.scalars[0], batch.scalars[i]);
+
```
tsachiherman, line 266: I haven't spent enough time on this loop - but, would it be possible to merge this loops to take advantage of cache locality ?

crypto/libsodium-fork/src/libsodium/crypto_sign/ed25519/ref10/batch.c
```diff
@@ -0,0 +1,318 @@
...
+        /* compute scalars[1]..scalars[batchsize] as r[i]*H(R[i],A[i],m[i]) */
+        for (i = 0; i < batchsize; i++) {
+            crypto_hash_sha512_state hs;   
+            crypto_hash_sha512_init(&hs);         
+            crypto_hash_sha512_update(&hs, RS[i], 32);
+            crypto_hash_sha512_update(&hs, pk[i], 32);
+            crypto_hash_sha512_update(&hs, m[i], mlen[i]);
+            crypto_hash_sha512_final(&hs, hram);
+
```
tsachiherman, line 275: I don't remember this was part of the donna library.. is that really the most efficient algorithm to do that ?

crypto/libsodium-fork/src/libsodium/include/sodium/private/ed25519_ref10_sc.h
```diff
@@ -0,0 +1,33 @@
...
+void mul256_modm(sc25519 r, const sc25519 x, const sc25519 y);
+
+void sub256_modm_batch(sc25519 out, const sc25519 a, const sc25519 b, size_t limbsize);
+int lt256_modm_batch(const sc25519 a, const sc25519 b, size_t limbsize);
+int lte256_modm_batch(const sc25519 a, const sc25519 b, size_t limbsize);
+int iszero256_modm_batch(const sc25519 a);
+int isone256_modm_batch(const sc25519 a);
+int isatmost128bits256_modm_batch(const sc25519 a);
+
+#endif
```
tsachiherman, line 33: add EOL

tsachiherman (commented):
Overall looks good. I've added comments about various potential issues I've came across.

crypto/libsodium-fork/src/libsodium/crypto_sign/ed25519/ref10/batch.c
cce: Found it in https://github.com/floodyberry/ed25519-donna/blob/master/ed25519-donna-batchverify.h#L235

crypto/libsodium-fork/src/libsodium/crypto_sign/ed25519/ref10/batch.c
algoidan: true is not native in C. so we need to define macros for that
if think while(1) is good enough + add some comment.
what do you think?

crypto/libsodium-fork/src/libsodium/crypto_core/ed25519/ref10/ed25519_ref10.c
algoidan: this function is being called only once in each batch. so we won't gain a lot from trying to optimize it.

crypto/libsodium-fork/src/libsodium/crypto_core/ed25519/ref10/ed25519_ref10.c
algoidan: I'm not sure we should spend time trying to optimize this function.
Mostly, because this function is part of the single verification (not the batch verification). 
In addition, those arrays are being built (in forward-scanning ) right before this line. so there is a good chance it is already in cache. If you still think we should investigate that, I would do that.

crypto/libsodium-fork/src/libsodium/crypto_core/ed25519/ref10/ed25519_ref10_sc.c
```diff
@@ -269,20 +324,24 @@ int lte256_modm_batch(const sc25519 a, const sc25519 b, size_t limbsize) {
...
-	size_t i;
-	for (i = 0; i < 5; i++)
-		if (a[i])
-			return 0;
-	return 1;
+	sc25519_element_t result = a[0] | 0;
+	result |= a[1] | 0;
+	result |= a[2] | 0;
+	result |= a[3] | 0;
+	result |= a[4] | 0;
```
tsachiherman, line 331: I'm abit confused here about the "| 0". `x | 0 == x`, right ? and while I'm sure the compiler would omit that, I can't find any reason to write it. anything that I missed ?

tsachiherman (dismissed):
I had one small comment, but I think that it's ready to be merged in.

crypto/libsodium-fork/src/libsodium/crypto_core/ed25519/ref10/ed25519_ref10_sc.c
algoidan: I don't think you are missing anything. thanks for the catch

cce:
Would it be addressed with the UINT64_C macro?

tsachiherman:
> Would it be addressed with the UINT64_C macro?

could be.. x86 is using little endian, arm is using big endian.
I suggest we test in both to make sure we're good.

cce:
ARM is little endian too ... we would need to use qemu or something to run on big endian

cce:
@algoidan you also have the U8TO64_LE macro you introduced elsewhere that converts from 8 unsigned chars to uint64

tsachiherman (approved):
looks good to me.

tsachiherman (approved):

Issue #2878, opened 2021-09-13
Partkey Interface: Avoid mutex and DB operation in Register/Record


winder:
The ParticipationRegistry attempts to DB access for functions called on the consensus goroutine. It does this using a cache, and a read/write mutex. Because of the mutex, there are cases where a long-running operation on another goroutine could block consensus.

Acceptance Criteria:
* use a buffered channel to pass Register/Record events to a separate goroutine.
* do not write registrations to the DB right away, use the cache and write them along with the record events. This is safe because we attempt to re-register keys at the start of every round.
* Use a different mutex for `Get` and `Flush`

yaovi-a:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Algorand-Main-Workspace-5ea71fda6d554d53dda36f16/issues/algorand/go-algorand/2878?planning-poker) with ZenHub @winder @tsachizehub @nicholasguoalgorand @algonautshant @algobolson @AlgoStephenAkiki @tolikzinovyev

PR #3034, opened 2021-10-11
Partkey unblocked

brianolson:
## Summary

for #2878


brianolson:
I think things are updated now. See the changes in tests because a few things there demanded that things be applied immediately that now aren't guaranteed until .Flush().
.Flush() blocks and waits until all IO ops are done, then returns any accumulated error from asynch io.

data/account/participationRegistry.go
```diff
@@ -691,42 +767,14 @@ func (db *participationDB) Flush() error {
 	db.mutex.Lock()
 	defer db.mutex.Unlock()
```
winder, line 778: Is the mutex still required? Seems like this would cause a deadlock when the writeThread attempts to Lock

brianolson: the flushDone condition variable temporarily releases its lock (`&db.mutex`) inside `.Wait()`; that's the receiving end of `db.flushDone.Broadcast()` in `writeThread()`

data/account/participationRegistry.go
```diff
@@ -282,15 +309,55 @@ func (db *participationDB) initializeCache() error {
...
+		var wr partDBWriteRecord
+		var chanOk bool
+		if needsFlush {
+			select {
+			case wr, chanOk = <-db.writeQueue:
+				if !chanOk {
+					return // chan closed
+				}
+			default:
+				// noting available, flush completed
```
winder, line 326: *nothing

data/account/participationRegistry.go
winder: Ahh, that's pretty cool. I found a couple of cases that I thought were race conditions but they seem to all be covered by the pattern. (multiple calls to flush, the window between sending the flush event/calling wait)

One potential issue I see... 

If an error is generated with multiple Flush() calls on the queue, we guarantee that only the first Flush returns the error. This might happen if we use Flush in the REST API and are inserting keys back-to-back.

I think we can solve that by counting the number of pending flushes and only clearing `asyncErr` when changing the pending account from 0 -> 1.

winder (commented):
Still looking good

data/account/participationRegistry.go
```diff
@@ -282,15 +309,55 @@ func (db *participationDB) initializeCache() error {
...
-	if _, ok := db.cache[id]; ok {
-		return id, ErrAlreadyInserted
+func (db *participationDB) writeThread() {
+	needsFlush := false
+	var err error
+	var lastErr error
+	for {
+		var wr partDBWriteRecord
+		var chanOk bool
+		if needsFlush {
```
winder, line 319: I think this deserves a comment. It looks like the code can be simplified but there's a lot going on and all of it is important.
```
// The flush command locks the mutex, then sends a flush event to unblock the queue read.
// When everything is flushed, the default is blocked until flush calls flushDone.Wait.
// Finally flushDone.Broadcast notifies all calls to Flush that the queue is now empty.
```

PR #3039, opened 2021-10-12
handle is left open in importRootKeys

algonautshant:
## Summary

The database handle is not closed in this code path.
In normal situations, this passes unnoticed, since it will be closed once the test terminates. However, when running the test multiple times, the OS will complain about too many open files. 

## Test Plan
This is a test infrastructure fix.


tsachiherman (approved):
Looks good. Note that is a similar leak in `messageRate_test.go:fetchSecretKey`.

PR #3046, opened 2021-10-12
ledger: Implement LookupAgreement

algorandskiy:
## Summary

1. Add  a new `OnlineAccountData` data type
2. Implement `LookupAgreement`

## Test Plan

Fixed existed tests: added `LookupAgreement` to mocked ledgers

### DIFF
--- a/data/basics/userBalance.go
+++ b/data/basics/userBalance.go
@@ -96,7 +96,8 @@ func UnmarshalStatus(value string) (s Status, err error) {
 	return
 }
 
-// OnlineAccountData contains just the account data used by agreement.
+// OnlineAccountData contains the voting information for a single account.
+//msgp:ignore OnlineAccountData
 type OnlineAccountData struct {
 	MicroAlgosWithRewards MicroAlgos
 
@@ -492,9 +493,14 @@ func (u AccountData) MinBalance(proto *config.ConsensusParams) (res MicroAlgos) 
 }
 
 // OnlineAccountData returns subset of AccountData as OnlineAccountData data structure.
-// Account is expected to be Online otherwise its stake is cleared out
+// Account is expected to be Online otherwise its is cleared out
 func (u AccountData) OnlineAccountData() OnlineAccountData {
-	result := OnlineAccountData{
+	if u.Status != Online {
+		// if the account is not Online and agreement requests it for some reason, clear it out
+		return OnlineAccountData{}
+	}
+
+	return OnlineAccountData{
 		MicroAlgosWithRewards: u.MicroAlgos,
 
 		VoteID:          u.VoteID,
@@ -503,12 +509,6 @@ func (u AccountData) OnlineAccountData() OnlineAccountData {
 		VoteLastValid:   u.VoteLastValid,
 		VoteKeyDilution: u.VoteKeyDilution,
 	}
-
-	// if the account is not Online and agreement requests it for some reason, set its voting stake to zero
-	if u.Status != Online {
-		result.MicroAlgosWithRewards = MicroAlgos{}
-	}
-	return result
 }
 
 // VotingStake returns the amount of MicroAlgos associated with the user's account
--- a/data/basics/userBalance_test.go
+++ b/data/basics/userBalance_test.go
@@ -267,9 +267,7 @@ func TestOnlineAccountData(t *testing.T) {
 	ad.Status = Offline
 
 	oad := ad.OnlineAccountData()
-	require.Empty(t, oad.MicroAlgosWithRewards)
-	require.Equal(t, ad.VoteID, oad.VoteID)
-	require.Equal(t, ad.SelectionID, oad.SelectionID)
+	require.Empty(t, oad)
 
 	ad.Status = Online
 	oad = ad.OnlineAccountData()


data/basics/userBalance.go
```diff
@@ -96,15 +96,29 @@ func UnmarshalStatus(value string) (s Status, err error) {
 	return
 }
 
+// AgreementAccountData contains just the account data used by agreement.
+type AgreementAccountData struct {
```
tsachiherman, line 100: AgreementAccountData -> OnlineAccountData

data/basics/userBalance.go
```diff
@@ -96,15 +96,29 @@ func UnmarshalStatus(value string) (s Status, err error) {
...
+	_struct struct{} `codec:",omitempty,omitemptyarray"`
+
+	Status     Status     `codec:"onl"`
```
tsachiherman, line 103: move `Status` to `AccountData`

data/basics/userBalance.go
```diff
@@ -96,15 +96,29 @@ func UnmarshalStatus(value string) (s Status, err error) {
...
 // AccountData contains the data associated with a given address.
 //
 // This includes the account balance, cryptographic public keys,
 // consensus delegation status, asset data, and application data.
 type AccountData struct {
 	_struct struct{} `codec:",omitempty,omitemptyarray"`
 
-	Status     Status     `codec:"onl"`
-	MicroAlgos MicroAlgos `codec:"algo"`
+	AgreementAccountData
```
cce, line 121: Does an anonymous embedding of a struct like this impact the msgp encoding? That would break catchpoints ... I didn't check originally. If so (or just to be safe, maybe it's best) we could have AgreementAccountData and AccountData be two completely separate types, and then have a copy constructor convert an AccountData to AgreementAccountData in LookupAgreement. Maybe the embedding was too cute and possibly disruptive (changing all the places where tests declare AccountData structs etc)

data/datatest/impls.go
```diff
@@ -106,6 +106,12 @@ func (i ledgerImpl) Lookup(r basics.Round, addr basics.Address) (basics.AccountD
 	return i.l.Lookup(r, addr)
 }
 
+// Lookup implements Ledger.LookupAgreement.
+func (i ledgerImpl) LookupAgreement(r basics.Round, addr basics.Address) (basics.AgreementAccountData, error) {
+	a, err := i.l.Lookup(r, addr)
+	return a.AgreementAccountData, err
```
cce, line 112: for example this could be `return makeAgreementAccountData(a)` instead

data/basics/userBalance.go
algorandskiy: Done. It was used in `VotingStake` to return zero if not Online. I moved this logic into `OnlineAccountData()` helper method. Agreement in theory might receive a malformed vote from the network, so this safety check looks important.

ledger/ledger.go
```diff
@@ -462,6 +462,20 @@ func (l *Ledger) Lookup(rnd basics.Round, addr basics.Address) (basics.AccountDa
...
+// LookupAgreement returns account data used by agreement.
+func (l *Ledger) LookupAgreement(rnd basics.Round, addr basics.Address) (basics.OnlineAccountData, error) {
+	l.trackerMu.RLock()
+	defer l.trackerMu.RUnlock()
+
+	// Intentionally apply (pending) rewards up to rnd.
+	data, err := l.accts.LookupWithRewards(rnd, addr)
+	if err != nil {
+		return basics.OnlineAccountData{}, err
+	}
```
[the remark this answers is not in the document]
tsachiherman: I'm hoping that in the context of this PR, we'd better get off to the account update level - and stop there.
That would ensure that when we make the next changes, these won't "leak" outside the account updates.

daemon/algod/api/server/v2/handlers.go
```diff
@@ -715,7 +715,7 @@ func (v2 *Handlers) GetAssetByID(ctx echo.Context, assetID uint64) error {
 	}
 
 	lastRound := ledger.Latest()
-	record, err := ledger.Lookup(lastRound, creator)
+	record, _, err := ledger.LookupWithoutRewards(lastRound, creator)
```
tsachiherman, line 718: It's ok to call the `ledger.LookupWithoutRewards`, but then you'll need to add the rewards.

data/basics/userBalance.go
```diff
@@ -96,6 +96,18 @@ func UnmarshalStatus(value string) (s Status, err error) {
 	return
 }
 
+// OnlineAccountData contains just the account data used by agreement.
```
tsachiherman, line 99: nit: update comment:
```
OnlineAccountData contains the voting information for a single account.
```

data/basics/userBalance.go
```diff
@@ -479,20 +491,36 @@ func (u AccountData) MinBalance(proto *config.ConsensusParams) (res MicroAlgos)
 	return res
 }
 
-// VotingStake returns the amount of MicroAlgos associated with the user's account
-// for the purpose of participating in the Algorand protocol.  It assumes the
-// caller has already updated rewards appropriately using WithUpdatedRewards().
-func (u AccountData) VotingStake() MicroAlgos {
+// OnlineAccountData returns subset of AccountData as OnlineAccountData data structure.
+// Account is expected to be Online otherwise its stake is cleared out
+func (u AccountData) OnlineAccountData() OnlineAccountData {
```
tsachiherman, line 496: The code here suggest that you could have a situation where you have a non-online account which has any of the VoteID/SelectionID/VoteFirstValid/VoteLastValid/VoteKeyDilution non-zero, which isn't correct.
If Status is online, then all these fields are valid. Otherwise, they are all empty.

data/basics/userBalance.go
```diff
@@ -96,6 +96,18 @@ func UnmarshalStatus(value string) (s Status, err error) {
 	return
 }
 
+// OnlineAccountData contains just the account data used by agreement.
+type OnlineAccountData struct {
```
tsachiherman, line 100: Add
//msgp:ignore OnlineAccountData

data/basics/userBalance.go
algorandskiy: @cce opinion?

ledger/ledger.go
```diff
@@ -462,6 +462,20 @@ func (l *Ledger) Lookup(rnd basics.Round, addr basics.Address) (basics.AccountDa
...
+
+	return data.OnlineAccountData(), nil
```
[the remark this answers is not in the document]
algorandskiy: why so? `l.accts.LookupWithRewards` applies rewards there. And looks like it is the only place with `roundTotals` to calculate rewards level.

tsachiherman (approved):
looks good to me, but I'd like @cce to take a look as well.

PR #3049, opened 2021-10-13
testing: fix failure in TestAccountInformationV2

tsachiherman:
## Summary

The test had a conceptual buggy pattern. It was using 
```golang
round, err = client.CurrentRound()
...
_, err = client.BroadcastTransaction(signedTxn)
...
resp, err = client.GetPendingTransactions(2)
```
which is doomed to fail if the test process goes to sleep for a minute between the `client.CurrentRound()` call and the `client.BroadcastTransaction` call. The trivial solution is to reorder the calls so that 
```golang
_, err = client.BroadcastTransaction(signedTxn)
...
round, err = client.CurrentRound()
...
resp, err = client.GetPendingTransactions(2)
```

## Test Plan

This is a test.

### DIFF
--- a/test/e2e-go/features/transactions/accountv2_test.go
+++ b/test/e2e-go/features/transactions/accountv2_test.go
@@ -105,11 +105,11 @@ func TestAccountInformationV2(t *testing.T) {
 
 	fee := uint64(1000)
 
-	round, err := client.CurrentRound()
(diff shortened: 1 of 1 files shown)


algorandskiy (approved):

tsachiherman:
@algorandskiy thanks.

PR #3052, opened 2021-10-13
Move secrets to partkey registry

winder:
## Summary

Start moving partkeys away from the old file format.


data/account/participationRegistry.go
```diff
@@ -80,6 +97,8 @@ func (r ParticipationRecord) Duplicate() ParticipationRecord {
 		LastCompactCertificate: r.LastCompactCertificate,
 		EffectiveFirst:         r.EffectiveFirst,
 		EffectiveLast:          r.EffectiveLast,
+		VRF:                    &vrf,
```
brianolson, line 100: we should only assign this if the source was not nil

data/account/participationRegistry.go
```diff
@@ -524,8 +561,19 @@ func (db *participationDB) Insert(record Participation) (id ParticipationID, err
...
+	}
+
+	var voting crypto.OneTimeSignatureSecrets
+	if record.Voting != nil {
+		voting = record.Voting.Snapshot()
+	}
+
 	// update cache.
-	// TODO: simplify to re-initializing with initializeCache()?
 	db.cache[id] = ParticipationRecord{
```
brianolson, line 577: this looks an awful lot like .Duplicate() ?
(and also things should only be assigned if the source was not nil)

data/account/participationRegistry.go
```diff
@@ -577,26 +627,46 @@ func scanRecords(rows *sql.Rows) ([]ParticipationRecord, error) {
...
 		if err != nil {
 			return nil, err
 		}
 
-		copy(record.ParticipationID[:], participationBlob)
-		copy(record.Account[:], accountBlob)
+		copy(record.ParticipationID[:], rawParticipation)
+		copy(record.Account[:], rawAccount)
+
+		record.VRF = &crypto.VRFSecrets{}
```
brianolson, line 655: move this inside the if{} so that record.VRF is nil if there's no rawVRF?

node/node.go
```diff
@@ -1189,6 +1189,12 @@ func (node *AlgorandFullNode) VotingKeys(votingRound, keysRound basics.Round) []
 		}
 		participations = append(participations, part)
 		matchingAccountsKeys[part.Address()] = true
+
+		// Make sure the key is registered.
+		err := node.accountManager.Registry().Register(part.ID(), keysRound)
+		if err != nil {
+			node.log.Warnf("Failed to register participation key (%s) with participation registry: %s\n", part.ID(), err.Error())
```
tsachiherman, line 1196: nit: convert the second format specifier to `%v` and use just `err`.

data/account/participationRegistry.go
winder: Yes, I didn't see a good way to reuse it.

data/account/participationRegistry.go
winder: Fixed

PR #3055, opened 2021-10-13
Refactoring ABI encoding feature

### LINKED ISSUE #3027: Reformat ABI implementation, remove redundant Value struct
## Summary

Remove redundant `Value` structs in `data/abi` and start encoding/decoding from type schemes.

See reference design (and implementation) in https://github.com/algorand/py-algorand-sdk/pull/238.

## Scope

- Remove `abi_value.go`.
- Reformat `abi_encode.go` and `abi_encode_test.go` to support (and check) the encoding/decoding from type schemes for golang's primitive value.

ahangsu:
## Summary

- Removed `data/abi/abi_value.go`
- Encode/Decode method take golang values, the behavior should be based on ABI type schemes

closes #3027 

## Test Plan

- keep all the current testing methods
- modify all the previous testing involved encoding/decoding methods to accept only golang values

## Minor Question

We will have to use reflection for inferring if `interface{}` typed value can be type-asserted to be `[]interface{}`.

We cannot directly do `a.([]interface{})`, that's why reflection comes in to achieve `interface{} -> []interface{}`.

Any thoughts on possible work-arounds? The perf seems not penalized...

### DIFF
--- a/data/abi/abi_encode.go
+++ b/data/abi/abi_encode.go
@@ -135,36 +135,36 @@ func (t Type) Encode(value interface{}) ([]byte, error) {
 
 // encodeInt encodes int-alike golang values to bytes, following ABI encoding rules
 func encodeInt(intValue interface{}, bitSize uint16) ([]byte, error) {
-	var bigInt big.Int
+	var bigInt *big.Int
 
 	switch intValue := intValue.(type) {
 	case int8:
-		bigInt = *new(big.Int).SetInt64(int64(intValue))
+		bigInt = big.NewInt(int64(intValue))
 	case uint8:
-		bigInt = *new(big.Int).SetUint64(uint64(intValue))
+		bigInt = new(big.Int).SetUint64(uint64(intValue))
 	case int16:
-		bigInt = *new(big.Int).SetInt64(int64(intValue))
+		bigInt = big.NewInt(int64(intValue))
 	case uint16:
-		bigInt = *new(big.Int).SetUint64(uint64(intValue))
+		bigInt = new(big.Int).SetUint64(uint64(intValue))
 	case int32:
-		bigInt = *new(big.Int).SetInt64(int64(intValue))
+		bigInt = big.NewInt(int64(intValue))
 	case uint32:
-		bigInt = *new(big.Int).SetUint64(uint64(intValue))
+		bigInt = new(big.Int).SetUint64(uint64(intValue))
 	case int64:
-		bigInt = *new(big.Int).SetInt64(intValue)
+		bigInt = big.NewInt(intValue)
 	case uint64:
-		bigInt = *new(big.Int).SetUint64(intValue)
+		bigInt = new(big.Int).SetUint64(intValue)
 	case uint:
-		bigInt = *new(big.Int).SetUint64(uint64(intValue))
+		bigInt = new(big.Int).SetUint64(uint64(intValue))
 	case int:
-		bigInt = *new(big.Int).SetInt64(int64(intValue))
+		bigInt = big.NewInt(int64(intValue))
 	case *big.Int:
-		bigInt = *new(big.Int).Set(intValue)
+		bigInt = intValue
 	default:
 		return nil, fmt.Errorf("cannot infer go type for uint encode")
 	}
 
-	if bigInt.Cmp(big.NewInt(0)) < 0 {
+	if bigInt.Sign() < 0 {
 		return nil, fmt.Errorf("passed in numeric value should be non negative")
 	}
 
@@ -348,7 +348,7 @@ func (t Type) Decode(encoded []byte) (interface{}, error) {
 		} else if encoded[0] == 0x80 {
 			return true, nil
 		}
-		return nil, fmt.Errorf("sinble boolean encoded byte should be of form 0x80 or 0x00")
+		return nil, fmt.Errorf("single boolean encoded byte should be of form 0x80 or 0x00")
 	case Byte:
 		if len(encoded) != 1 {
 			return nil, fmt.Errorf("byte should be length 1")
--- a/data/abi/abi_encode_test.go
+++ b/data/abi/abi_encode_test.go
@@ -469,8 +469,6 @@ func TestDecodeValid(t *testing.T) {
 			uint64(5), uint64(6),
 			uint64(7), uint64(8),
 		}
-		//expected := []interface{}{1, 2, 3, 4, 5, 6, 7, 8}
-		//arrayEncoded, err := staticUintArrT.Encode(expected)
 		arrayEncoded, err := staticUintArrT.Encode(inputUint)
 		require.NoError(t, err, "uint64 static array encode should not return error")
 		actual, err := staticUintArrT.Decode(arrayEncoded)


data/abi/abi_encode.go
```diff
@@ -20,214 +20,249 @@ import (
...
-func compressMultipleBool(valueList []Value) (uint8, error) {
-	var res uint8 = 0
-	if len(valueList) > 8 {
-		return 0, fmt.Errorf("value list passed in should be no greater than length 8")
+// encodeInt encodes int-alike golang values to bytes, following ABI encoding rules
+func encodeInt(intValue interface{}, bitSize uint16) ([]byte, error) {
+	var bigInt big.Int
+
+	switch intValue := intValue.(type) {
+	case int8:
```
jasonpaulos, line 141: Can you add a case for `byte` too?

data/abi/abi_encode_test.go
```diff
@@ -1004,200 +795,179 @@ func TestDecodeInvalid(t *testing.T) {
...
-			tupleElem := (*testValuePool)[tupleTypeIndex][tupleElemRangeIndex]
-			tupleValList[tupleElemIndex] = tupleElem
-		}
-		tupleVal, err := MakeTuple(tupleValList)
-		require.NoError(t, err, "make tuple should not return error")
-		(*testValuePool)[Tuple] = append((*testValuePool)[Tuple], tupleVal)
+			tupleElem := (*pool)[tupleTypeIndex][tupleElemRangeIndex]
+			testUnits[index] = tupleElem
+		}
+		// TODO
```
jasonpaulos, line 947: What is this for?

data/abi/abi_encode.go
```diff
@@ -268,203 +303,168 @@ func tupleEncoding(v Value) ([]byte, error) {
...
+
+// Decode is an ABI type method to decode bytes to go values from ABI encoding rules
+func (t Type) Decode(encoded []byte) (interface{}, error) {
+	switch t.abiTypeID {
+	case Uint, Ufixed:
+		if len(encoded) != int(t.bitSize)/8 {
+			return nil,
+				fmt.Errorf("uint/ufixed decode: expected byte length %d, but got byte length %d", t.bitSize/8, len(encoded))
+		}
+		return new(big.Int).SetBytes(encoded), nil
```
jasonpaulos, line 333: I wonder if we should return the types `uint8`, `uint16`, `uint32`, `uint64` if the ABI type precision can fit into them. Something like:

| ABI precision  | Go type |
| ------------- | ------------- |
| 8  | `uint8` |
| 16  | `uint16` |
| 24 - 32 | `uint32` |
| 40 - 64 | `uint64` |
| > 64 | `*big.Int` |

data/abi/abi_encode_test.go
ahangsu: 😅 forgot to remove the comments

data/abi/abi_encode.go
ahangsu: It is doable, and I think this is a very nice improvement, but the testing will have more scripts to support different types. Can't do it today, leave it tomorrow...

data/abi/abi_encode.go
```diff
@@ -20,214 +20,249 @@ import (
...
 }
 
-// tupleEncoding encodes an ABI value of tuple type into an ABI encoded byte string.
-func tupleEncoding(v Value) ([]byte, error) {
-	if v.ABIType.abiTypeID != Tuple {
-		return []byte{}, fmt.Errorf("type not supported in tupleEncoding")
+// inferToSlice infers an interface element to a slice of interface{}, returns error if it cannot infer successfully
+func inferToSlice(value interface{}) ([]interface{}, error) {
+	reflectVal := reflect.ValueOf(value)
+	if reflectVal.Kind() != reflect.Slice {
```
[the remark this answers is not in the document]
ahangsu: Maybe I understood it wrong, but I think if you are passing a slice, then the slice is automatically casted into an interface...?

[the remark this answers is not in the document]
jasonpaulos: I mean it would be nice for a go array like `[10]bool` to be a valid input as well. This type has kind `reflect.Array`, but other than that, it's pretty much the same as a slice for our purposes.

data/abi/abi_encode.go
```diff
@@ -322,15 +314,31 @@ func compressBools(boolSlice []interface{}) (uint8, error) {
...
+	}
+	switch bitSize / 8 {
+	case 1:
+		return encoded[0], nil
+	case 2:
+		return uint16(new(big.Int).SetBytes(encoded).Uint64()), nil
+	case 3, 4:
+		return uint32(new(big.Int).SetBytes(encoded).Uint64()), nil
+	case 5, 6, 7, 8:
+		return new(big.Int).SetBytes(encoded).Uint64(), nil
```
jasonpaulos, line 331: nit: `binary.BigEndian.Uint16/Uint32/Uint64(encoded)` might be better than converting to bigint first

ahangsu: This is also a ugly workaround, since the `binary.BigEdian` methods enforce the byte length to be 2/4/8.

if you are attempting 65537: `[1, 0, 1]`, then you will have array index out of scope error...

data/abi/abi_encode_test.go
```diff
@@ -34,6 +35,9 @@ func TestEncodeValid(t *testing.T) {
 	// randomly pick 1000 valid uint values and check if encoded value match with expected
 	for intSize := 8; intSize <= 512; intSize += 8 {
 		upperLimit := big.NewInt(0).Lsh(big.NewInt(1), uint(intSize))
+		uintType, err := MakeUintType(uint16(intSize))
```
jannotti, line 38: Why does MakeUintType take a uint16?  I'd imagine you'd be better off with int.  You'll still need bounds checking, but it will avoid lots of casts when calling it.

data/abi/abi_encode_test.go
```diff
@@ -72,14 +72,14 @@ func TestEncodeValid(t *testing.T) {
 			big.NewInt(1).Neg(big.NewInt(1)),
 		)
 		for precision := 1; precision <= 160; precision++ {
+			typeUfixed, err := MakeUfixedType(uint16(size), uint16(precision))
```
jannotti, line 75: Same here, why uint16?

data/abi/abi_encode_test.go
```diff
@@ -100,6 +98,7 @@ func TestEncodeValid(t *testing.T) {
 	// encoding test for address, since address is 32 byte, it can be considered as 256 bit uint
 	// randomly generate 1000 uint256 and make address values, check if encoded value match with expected
 	upperLimit := big.NewInt(0).Lsh(big.NewInt(1), 256)
+	typeAddress := MakeAddressType()
```
jannotti, line 101: Perhaps the address type should/could be a constant, rather than function?  There's no utility to have multiple instances, right?

data/abi/abi_encode_test.go
```diff
@@ -108,19 +107,15 @@ func TestEncodeValid(t *testing.T) {
...
-
-		addressValue := MakeAddress(addrBytes)
-		addrBytesActual, err := addressValue.Encode()
+		addrBytesActual, err := typeAddress.Encode(addrBytesExpected)
 		require.NoError(t, err, "address encode fail")
 		require.Equal(t, addrBytesExpected, addrBytesActual, "encode addr not match with expected")
 	}
 
 	// encoding test for bool values
+	boolType := MakeBoolType()
```
jannotti, line 116: same

data/abi/abi_encode_test.go
```diff
@@ -323,50 +282,70 @@ func TestEncodeValid(t *testing.T) {
...
-		require.NoError(t, err, "make empty tuple should not return error")
 		expected := make([]byte, 0)
-		actual, err := emptyTuple.Encode()
+		actual, err := tupleType.Encode([]interface{}{})
 		require.NoError(t, err, "encode empty tuple should not return error")
 		require.Equal(t, expected, actual, "empty tuple encode should not return error")
 	})
 }
 
+func castBigIntToInterface(val *big.Int, bitSize uint16) (interface{}, error) {
```
jannotti, line 311: I'm not seeing every place where this is used, but it seems dangerous.  IT's inferring the type based one the size of its arguments?  So I can't get a val.Uint64() from a small number.   If that's a good idea in some contexts, I'd at least prefer a name like "castBigIntToNarrowest()" or similar.

I suppose it's not so dangerous since it's only in testing code.

ahangsu:
My thoughts is that, probably hide most of the `Make...Type` functions, and directly construct ABI-types from `TypeOf`.
We might want to export some of the basic type constants like: `ByteType`, `StringType`, `AddressType`, `BoolType`.

data/abi/abi_encode_test.go
ahangsu: I discarded the function now, don't think it would be a nice idea to use it...

data/abi/abi_encode_test.go
```diff
@@ -308,41 +303,35 @@ func TestEncodeValid(t *testing.T) {
...
+			} else if intSize == 8 {
+				expected = uint8(randBig.Uint64())
+			} else {
+				expected = randBig
+			}
+
+			encodedUint, err := uintType.Encode(expected)
 			require.NoError(t, err, "uint encode fail")
-			expected, err := castBigIntToInterface(expectedBig, uint16(intSize))
+			//expected, err := castBigIntToNarrowestPrimitive(expected, intSize)
```
jasonpaulos, line 334: nit: if not needed, this should be removed

data/abi/abi_type.go
```diff
@@ -113,15 +113,15 @@ func (t Type) String() string {
 var staticArrayRegexp = regexp.MustCompile(`^([a-z\d\[\](),]+)\[([1-9][\d]*)]$`)
 var ufixedRegexp = regexp.MustCompile(`^ufixed([1-9][\d]*)x([1-9][\d]*)$`)
 
-// TypeFromString de-serialize ABI type from a string following ABI encoding.
-func TypeFromString(str string) (Type, error) {
+// TypeOf de-serialize ABI type from a string following ABI encoding.
```
jasonpaulos, line 116: nit: suggested docstring improvement/simplification
```suggestion
// TypeOf parses an ABI type string.
// For example: `TypeOf("(uint64,byte[])")`
```

jasonpaulos (approved):
Overall looks good, just 2 nits

data/abi/abi_encode.go
```diff
@@ -20,214 +20,241 @@ import (
...
-			return 0, fmt.Errorf("bool type not matching in compressMultipleBool")
-		}
-		boolVal, err := valueList[i].GetBool()
-		if err != nil {
-			return 0, err
-		}
-		if boolVal {
-			res |= 1 << uint(7-i)
+
+	if bigInt.Cmp(big.NewInt(0)) < 0 {
```
jannotti, line 167: Prefer bigInt.Sign()?
https://pkg.go.dev/math/big#Int.Sign

PR #3080, opened 2021-10-16
Fix missing TX registration

tsachiherman:
## Summary

When a 2.1 (i.e. no txsync) client connects to a 3.0 relay (i.e. txsync), the relay needs to request the client to keep sending it a TX messages - otherwise, these transactions would not get propagated.

The 2.1 and 3.0 above are network protocol versions, not a algod release version.

## Test Plan

e2e test added.


brianolson (approved):
Good catch. I think I see how my prior mixed-network test was incomplete.

PR #3085, opened 2021-10-18
ledger: create catchpoint tracker

tsachiherman:
## Summary

Extract catchpoint logic out of the accounts update tracker.

## Test Plan

Refactor existing unit tests.


ledger/acctupdates.go
```diff
@@ -640,63 +526,42 @@ func (au *accountUpdates) committedUpTo(committedRound basics.Round) (retRound b
...
-		au.log.Panicf("produceCommittingTask: block %d too far in the future, lookback %d, dbRound %d (cached %d), deltas %d", committedRound, lookback, dbRound, au.cachedDBRound, len(au.deltas))
+		au.log.Panicf("produceCommittingTask: block %d too far in the future, lookback %d, dbRound %d (cached %d), deltas %d", committedRound, dcc.lookback, dbRound, au.cachedDBRound, len(au.deltas))
 	}
 
-	hasIntermediateCatchpoint := false
-	hasMultipleIntermediateCatchpoint = false
 	// check if there was a catchpoint between au.dbRound+lookback and newBase+lookback
 	if au.catchpointInterval > 0 {
-		nextCatchpointRound := ((uint64(dbRound+lookback) + au.catchpointInterval) / au.catchpointInterval) * au.catchpointInterval
+		nextCatchpointRound := ((uint64(dbRound+dcc.lookback) + au.catchpointInterval) / au.catchpointInterval) * au.catchpointInterval
```
cce, line 553: This logic has been copied to catchpointTracker.produceCommittingTask but is still needed here, could they share?

ledger/catchpointtracker.go
```diff
@@ -0,0 +1,908 @@
...
+	// catchpointWriting help to synchronize the catchpoint file writing. When this atomic variable is 0, no writing is going on.
+	// Any non-zero value indicates a catchpoint being written.
+	catchpointWriting int32
+
+	// The Trie tracking the current account balances. Always matches the balances that were
+	// written to the database.
+	balancesTrie *merkletrie.Trie
+
+	// accountsMu is the synchronization mutex for accessing the various non-static variables.
+	accountsMu deadlock.RWMutex
```
cce, line 112: Maybe give this a different name like `catchpointsMu` or something so it doesn't get confused with `accountUpdates.accountsMu`?

ledger/catchpointtracker.go
```diff
@@ -0,0 +1,908 @@
...
+			return err0
+		}
+		return nil
+	})
+
+	if err != nil {
+		return err
+	}
+
+	ct.accountsq, err = accountsInitDbQueries(ct.dbs.Rdb.Handle, ct.dbs.Wdb.Handle)
```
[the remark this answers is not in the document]
tsachiherman: I think that once we'll have the database encapsulation, it would hide all these - so, I'd rather wait with these until we're ready.

ledger/acctupdates.go
```diff
@@ -716,27 +581,15 @@ func (au *accountUpdates) produceCommittingTask(committedRound basics.Round, dbR
 	//   so that all the instances of the catchpoint would contain exactly the same data )
 	flushTime := time.Now()
 	if !flushTime.After(au.lastFlushTime.Add(balancesFlushInterval)) && !isCatchpointRound && pendingDeltas < pendingDeltasFlushThreshold {
```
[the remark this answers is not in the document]
tsachiherman: moved to the tracker

ledger/catchpointtracker.go
```diff
@@ -0,0 +1,908 @@
...
+
+// newBlock informs the tracker of a new block from round
+// rnd and a given ledgercore.StateDelta as produced by BlockEvaluator.
+func (ct *catchpointTracker) newBlock(blk bookkeeping.Block, delta ledgercore.StateDelta) {
+	ct.catchpointsMu.Lock()
+	defer ct.catchpointsMu.Unlock()
+	ct.roundDigest = append(ct.roundDigest, blk.Digest())
+}
+
+// committedUpTo informs the tracker that the database has
```
algorandskiy, line 228: probably this large comment should live in interface definition and not in each tracker since it is generic. Plus, it does not mention `lookback`

ledger/catchpointtracker.go
```diff
@@ -0,0 +1,908 @@
...
+	balancesTrie *merkletrie.Trie
+
+	// catchpointsMu is the synchronization mutex for accessing the various non-static variables.
+	catchpointsMu deadlock.RWMutex
+
+	// roundDigest stores the digest of the block for every round starting with dbRound and every round after it.
+	roundDigest []crypto.Digest
+}
+
+// initialize initializes the accountUpdates structure
```
algorandskiy, line 118: accountUpdates -> catchpointTracker

algorandskiy (commented):
I like the separation of the logic.
In the same time I'm a bit concerned with mixing commit data (offset, base) with context data into `deferredCommitContext`. I propose to keep them separate: let trackers contribute into commit data by modifying (offset, base) as needed but only setting/getting data from the context in order to have kind of "these data are set by au and used by ct"

tsachiherman:
> I like the separation of the logic.
> In the same time I'm a bit concerned with mixing commit data (offset, base) with context data into `deferredCommitContext`. I propose to keep them separate: let trackers contribute into commit data by modifying (offset, base) as needed but only setting/getting data from the context in order to have kind of "these data are set by au and used by ct"

I've broken a subset of `deferredCommitContext` into `deferredCommitRange`, so that each tracker could individually, during `produceCommittingTask` to create a offset/base. Once all the trackers are doing that, this data structure is available as part of the `deferredCommitContext`, allowing all the trackers to review it.

algorandskiy (approved):
Looks good. The only thing worries me is `deferredCommitRange` is modified by all trackers and there are no restrictions on offset/base change. Maybe some kind "protocol" on modifications is needed otherwise there is a chance some tracker can set the offset way too off for some other tracker. It looks like the range `deferredCommitRange` wants to be a type with some encapsulated behavior with max/min offset values for this round.

tsachiherman:
> Looks good. The only thing worries me is `deferredCommitRange` is modified by all trackers and there are no restrictions on offset/base change. Maybe some kind "protocol" on modifications is needed otherwise there is a chance some tracker can set the offset way too off for some other tracker. It looks like the range `deferredCommitRange` wants to be a type with some encapsulated behavior with max/min offset values for this round.

I didn't want to be too restrictive on what can be done with the `deferredCommitRange`, as it feels like over engineering. Generally, we want only to narrow down the range as we move along, but given that we have only two parties that really care about that, I couldn't really justify that..

PR #3087, opened 2021-10-18
Make bytes and pushbytes check their immediate better (and similarly)

jannotti:
`byte "hi" "there"` did not report an error, but should.
`pushbytes` was not consistent with `bytes` in reporting errors, but should be.


jasonpaulos (approved):
Looks good

Issue #2803, opened 2021-08-25
Use ABI encoding support in goal


jasonpaulos:
## Summary
Add support in goal for sending app calls with encoded ABI arguments, and for viewing ABI encoded return values from app calls.

## Scope
Specifically:
* Add support for passing and encoding ABI method selectors and values into `goal app` calls, similar to the existing `--app-arg` support for encoding strings, integers, and other values ([see the implementation](https://github.com/algorand/go-algorand/blob/734e727e677c2c574410d096962d3c86830f1899/cmd/goal/application.go#L198-L236) for a complete set).
  * Proposal: support a new encoding type for ABI values: `abi:<abi-type-string>:<abi-value-as-json>`
    * `<abi-type-string>` would be the serialized ABI type string, e.g. `(bool,byte[],uint64)`
    * `<abi-value-as-json>` would be a JSON string that represents an ABI value. Together with the type string, an ABI value can be properly formed. The mapping from schema type to acceptable JSON values would be:
      * ABI uint/byte: accepts JSON integers, throws error if number is outside acceptable range.
      * ABI ufixed: accepts JOSN decimal numbers, throws error if number is outside acceptable range. For example, a `ufixed64x2` should accept a decimal with up to 2 digits after the decimal, e.g. `1.99`.
      * ABI bool: accepts JSON boolean values
      * ABI string: accepts a JSON string, or a JSON array containing the raw bytes this string should contain. This allows non UTF-8 characters to be encoded
      * ABI address: accepts a JSON string containing a base32 Algorand address with checksum, throws error if address is invalid
      * ABI tuple: accepts a JSON array
      * ABI static/dynamic arrays: accepts a JSON array, or **if this is an array of bytes**, additionally accept a JSON base64 encoded string, since this is the de facto format for embedding binary data in JSON
    * For example, the JSON  string `[true, [0, 1, 2], 17]` is a valid encoding of the type `(bool,byte[],uint64)`. An equivalent encoding using base64 for the contained byte array would be `[true, "AAEC", 17]`. Both should be accepted.
    * The reason behind using JSON for this is because it's already a well-known format by developers, and we won't have to implement a complex parser just to make this work.
  * Complete example: we should be able to pass `--app-arg "abi:(bool,byte[],uint64):[true, [0, 1, 2], 17]"` as a flag to goal app calls
* Add a way to view the return value of an app call. We are using `log` for this, but this may require an ABI spec change to explicitly specify the format of the return value.

[...]

PR #3088, opened 2021-10-18
Add ABI support to goal

ahangsu:
## Summary

- [x] Implement JSON unmarshal to golang values on ABI encoding
- [x] Implement a `method` sub-command to accept method signature and application args
- [x] Add a way to view the return value of an app call.

Addressing #2803 

## Test Plan

- [x] Test unmarshal JSON values from command line to ABI encoding


cmd/goal/application.go
```diff
@@ -995,3 +1023,100 @@ var infoAppCmd = &cobra.Command{
...
+		}
+
+		// copy-paste code from callAppCmd
+		dataDir := ensureSingleDataDir()
+		client := ensureFullClient(dataDir)
+
+		// Parse transaction parameters
+		_, appAccounts, foreignApps, foreignAssets := getAppInputs()
+
+		tx, err := client.MakeUnsignedAppNoOpTx(appIdx, applicationArgs, appAccounts, foreignApps, foreignAssets)
```
jasonpaulos, line 1057: Instead of always assuming OnCompletion is no-op, it would be nice if you could use `createOnCompletion` to set the OnCompletion value (and probably rename that variable to be more general).

jasonpaulos (commented):
Looks mostly good, but could you also add an e2e test for the `goal app method` subcommand and the `--app-arg "abi:type:value"` flag? A good example to follow would be `test/scripts/e2e_subs/e2e-app-cross-round.sh`

cmd/goal/application.go
ahangsu: I am not quite sure if the current code change is the correct practice, but I tried to retrieve the `onCompletion` from commandline, and instantiate a transaction (unsigned) from `MakeUnsignedApplicationCallTx`.
It seems some `onCompletion` option is not applicable here?

cmd/goal/application.go
```diff
@@ -79,9 +84,10 @@ func init() {
 	appCmd.AddCommand(clearAppCmd)
 	appCmd.AddCommand(readStateAppCmd)
 	appCmd.AddCommand(infoAppCmd)
+	appCmd.AddCommand(methodAppCmd)
 
 	appCmd.PersistentFlags().StringVarP(&walletName, "wallet", "w", "", "Set the wallet to be used for the selected operation")
-	appCmd.PersistentFlags().StringSliceVar(&appArgs, "app-arg", nil, "Args to encode for application transactions (all will be encoded to a byte slice). For ints, use the form 'int:1234'. For raw bytes, use the form 'b64:A=='. For printable strings, use the form 'str:hello'. For addresses, use the form 'addr:XYZ...'.")
+	appCmd.PersistentFlags().StringArrayVar(&appArgs, "app-arg", nil, "Args to encode for application transactions (all will be encoded to a byte slice). For ints, use the form 'int:1234'. For raw bytes, use the form 'b64:A=='. For printable strings, use the form 'str:hello'. For addresses, use the form 'addr:XYZ...'.")
```
ahangsu, line 90: Consider following input example: `--app-arg "abi:(uint64,uint64):[1,1]"`, the comma in the string will incur `StringSliceVar` to break down the string by commas. This contradicts our expected behavior.

A workaround is to use `StringArrayVar`, since each `--app-arg` only allows at most one argument input at a time, so comma should be allowed in input string.

cmd/goal/application.go
```diff
@@ -995,3 +1023,123 @@ var infoAppCmd = &cobra.Command{
...
+
+var methodAppCmd = &cobra.Command{
+	Use:     "method",
+	Short:   "Invoke a method",
+	Long:    `Invoke a method in an App (stateful contract) with an application call transaction`,
+	Args:    validateNoPosArgsFn,
+	PreRunE: validateNoPosArgsFn,
+	Run: func(cmd *cobra.Command, args []string) {
+		var applicationArgs [][]byte
+
```
ahangsu, line 1035: For method implementation, I don't think it is nice to let `--arg` and `--app-arg` be mixed together, so I additionally added a `methodArgs` string slice to store the inputs, let me know if you go for/against that opinion.

cmd/goal/application.go
```diff
@@ -266,6 +294,22 @@ func processAppInputFile() (args [][]byte, accounts []string, foreignApps []uint
...
+	for i := 0; i < len(appArgs); i++ {
+		if !strings.Contains(appArgs[i], "abi:") && len(appArgs[i]) > 0 {
+			splitByComma := strings.Split(appArgs[i], ",")
+			newAppArgs = append(newAppArgs, splitByComma...)
+		} else if strings.Contains(appArgs[i], "abi:") {
+			// TODO consider case when abi packed value comma with other packed value
+			newAppArgs = append(newAppArgs, appArgs[i])
+		}
+	}
+
```
ahangsu, line 309: Since I changed the `StringSliceVar` to`StringArrayVar`, we need to process the `appArgs` beforehand by:
- Filter out empty strings.
- Split up concatenated values of form like `str:string1,integer:1337`.
- Allowing abi values to be added to `appArgs`.

One thing I am not sure is that, do we want to process strings containing `abi:` (e.g., split up the string into slices)? Since the parsing will be quite complex by my estimation.

cmd/goal/application.go
```diff
@@ -995,3 +1044,141 @@ var infoAppCmd = &cobra.Command{
...
+			retType, err := abi.TypeOf(retTypeStr)
+			if err != nil {
+				reportErrorf("cannot cast %s to abi type: %v", retTypeStr, err)
+			}
+			decoded, err := retType.Decode(abiEncodedRet)
+			if err != nil {
+				reportErrorf("cannot decode return value %v: %v", abiEncodedRet, err)
+			}
+			fmt.Println(decoded)
```
ahangsu, line 1181: Find the logged entry with `hash("return")` 4 bytes prefix, and parse with rules specified by ABI return type.

cmd/goal/application.go
```diff
@@ -275,6 +327,9 @@ func getAppInputs() (args [][]byte, accounts []string, foreignApps []uint64, for
 	}
 
 	var encodedArgs []appCallArg
+
+	splitAppArgsByComma()
+
```
jannotti, line 332: Going along with my previous comment - this looks like magic.  If the function takes and returns the list, you can show what's going on.

cmd/goal/application.go
```diff
@@ -995,3 +1050,152 @@ var infoAppCmd = &cobra.Command{
...
+		if len(foreignApps) > 0 {
+			reportErrorf("in goal app method: --foreign-app is not supported")
+		}
+		if len(foreignAssets) > 0 {
+			reportErrorf("in goal app method: --foreign-asset is not supported")
+		}
+
+		// copy-paste code from callAppCmd
+		dataDir := ensureSingleDataDir()
+		client := ensureFullClient(dataDir)
```
jannotti, line 1094: There is a lot of copied code here.  Can you make a function and call it from both?

cmd/goal/application.go
```diff
@@ -266,6 +306,18 @@ func processAppInputFile() (args [][]byte, accounts []string, foreignApps []uint
 	return parseAppInputs(inputs)
 }
 
+func splitAppArgsByComma() {
```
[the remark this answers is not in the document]
ahangsu: I will write a docstring for that, but I will anyways explain here first.

The name is not correct, but it was first made out of purpose of splitting input string like form `str:a,str:b`, since the parsing of `appArgs` uses `StringSliceVar` and cut the string into slices with comma.

I switched it to `StringArrayVar` to support parsing of `abi:...` string, or it will cut the string into pieces. But unfortunately `StringArrayVar` will allow empty string parsed into `appArgs`.

In this way we have to first filter out the empty string with a not good magic way. If we do not do so, then it will contradict with some behavior tests...

What is your thoughts on that?

cmd/goal/application.go
ahangsu: I assume what you meant is compact the chunk of shared code into an internal function, and shared with both method, is that correct?

cmd/goal/application.go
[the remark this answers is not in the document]
jannotti: It's ok to filter out the blanks, it just needs to be named appropriately.  And it should take a slice and return a slice, not just modify globals.

cmd/goal/application.go
jannotti: Yes, unless there are many small differences.

ahangsu: I realized that it is initially copied from callAppCmd, but there are still a lot of difference here and there, does not necessarily share the structure.

I managed to compact some chunk of code, but let me know if anywhere still need to be optimized.

cmd/goal/application.go
```diff
@@ -995,3 +1050,152 @@ var infoAppCmd = &cobra.Command{
...
+			reportErrorf("in goal app method: --arg and --app-arg are mutually exclusive, do not use --app-arg")
+		}
+		if len(appAccounts) > 0 {
+			reportErrorf("in goal app method: --app-account is not supported")
+		}
+		if len(foreignApps) > 0 {
+			reportErrorf("in goal app method: --foreign-app is not supported")
+		}
+		if len(foreignAssets) > 0 {
+			reportErrorf("in goal app method: --foreign-asset is not supported")
```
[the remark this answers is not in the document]
ahangsu: I see, so it might be okay to unconstrain `foreignApps`, `foreignAssets` and `appAccounts` for now?
I suppose `appArgs` should be mutual exclusive with `--arg` in `goal app method`, so I will still disable that.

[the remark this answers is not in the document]
jasonpaulos: Assuming there was support for the special types `account`/`application`/`asset`, does it make sense to also support the `--foreign-*` flags? I think not, for the same reason we don't allow you to also specify additional raw byte arguments with `--app-arg`.

And I believe we should support the special types `account`/`application`/`asset` soon in a follow up PR. So while it's regrettable that foreign properties won't be settable immediately, I'd rather not allow the `--foreign-*` flags temporarily only to remove them (or worse, forget to remove them) later.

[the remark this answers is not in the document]
jasonpaulos: Oh I see you've already removed them @ahangsu. That's fine, I don't care enough about this to ask you to add them back. We can just address it in the follow up PR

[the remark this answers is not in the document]
jannotti: These feel different to me.  The ABI describes exactly how ApplicationArgs are to be used. So it does not make sense to allow people to set them using --app-arg, and maybe mess up what the ABI expects.  But the ABI does not specify how your foreign arrays will be packed. It merely says that IF you use a special type as an argument, it'll go (somewhere!) in the appropriate foreign-array, and the index will be encoded.

It's perfectly ok to let the caller put extra resources into the foreign arrays.  And I think it will be useful as people adopt the ABI to greater or lesser extents. I think I can even imagine that users will *need* this, even after ABI is complete and totally adopted, but I need to think more.  At any rate, the main point is that the ABI doesn't say how the foreign arrays will be packed, but does say how the application args will be packed.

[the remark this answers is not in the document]
jannotti: I suppose this also means we might be clever about encoding the foreign arrays.  If the same account is used as two different arguments, it can be put in once, and both arguments should refer to it. (I don't care a lot if we do this.)

data/abi/abi_type.go
```diff
@@ -333,8 +333,8 @@ func makeDynamicArrayType(argumentType Type) Type {
 	}
 }
 
-// makeTupleType makes tuple ABI type by taking an array of tuple element types as argument.
-func makeTupleType(argumentTypes []Type) (Type, error) {
+// MakeTupleType makes tuple ABI type by taking an array of tuple element types as argument.
+func MakeTupleType(argumentTypes []Type) (Type, error) {
 	if len(argumentTypes) >= math.MaxUint16 {
 		return Type{}, fmt.Errorf("tuple type child type number larger than maximum uint16 error")
 	}
```
ahangsu, line 340: I kinda want to get this exported. It seems @algoidurovic is working on https://github.com/algorand/go-algorand-sdk/pull/258/ and he needs this stuff?

test/scripts/e2e_subs/e2e-app-abi-arg.sh
```diff
@@ -0,0 +1,37 @@
...
+gcmd="goal -w ${WALLET}"
+
+GLOBAL_INTS=2
+ACCOUNT=$(${gcmd} account list|awk '{ print $3 }')
+
+printf '#pragma version 2\nint 1' > "${TEMPDIR}/simple.teal"
+PROGRAM=($(${gcmd} clerk compile "${TEMPDIR}/simple.teal"))
+APPID=$(${gcmd} app create --creator ${ACCOUNT} --approval-prog ${DIR}/tealprogs/app-abi-arg.teal --clear-prog ${TEMPDIR}/simple.teal --global-byteslices 0 --global-ints ${GLOBAL_INTS} --local-byteslices 0 --local-ints 0 | grep Created | awk '{ print $6 }')
+
+# Should succeed to opt in with first arg hello
```
jasonpaulos, line 24: nit: comment is not right

test/scripts/e2e_subs/e2e-app-abi-add.sh
```diff
@@ -0,0 +1,39 @@
...
+gcmd="goal -w ${WALLET}"
+
+GLOBAL_INTS=2
+ACCOUNT=$(${gcmd} account list|awk '{ print $3 }')
+
+printf '#pragma version 2\nint 1' > "${TEMPDIR}/simple.teal"
+PROGRAM=($(${gcmd} clerk compile "${TEMPDIR}/simple.teal"))
+APPID=$(${gcmd} app create --creator ${ACCOUNT} --approval-prog ${DIR}/tealprogs/app-abi-add-example.teal --clear-prog ${TEMPDIR}/simple.teal --global-byteslices 0 --global-ints ${GLOBAL_INTS} --local-byteslices 0 --local-ints 0 | grep Created | awk '{ print $6 }')
+
+# Should succeed to opt in with first arg hello
```
jasonpaulos, line 24: Same here

jasonpaulos (dismissed):
Looking good, just a nit

data/abi/abi_type.go
jasonpaulos: Right, this makes sense

test/scripts/e2e_subs/e2e-app-abi-arg.sh
ahangsu: ugh, I copied these from `e2e-app-cross-round.sh`, going to change soon.

cmd/goal/application.go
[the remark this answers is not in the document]
jasonpaulos: Good point about the ABI not specifying exactly where the foreign arguments will show up. In light of that, allowing additional foreign params sounds fine to me.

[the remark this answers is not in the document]
ahangsu: Just asking, so do we want to allow --app-arg here anyways? Or it is just fine for now?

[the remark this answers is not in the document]
jasonpaulos: I vote that it's fine for now

jasonpaulos (approved):

cmd/goal/application.go
```diff
@@ -266,6 +304,20 @@ func processAppInputFile() (args [][]byte, accounts []string, foreignApps []uint
 	return parseAppInputs(inputs)
 }
 
+// filterEmptyStrings filters out empty string parsed in by StringArrayVar
+// this function is added to support abi argument parsing
+// since parsing of `appArg` diverted from `StringSliceVar` to `StringArrayVar`
+func filterEmptyStrings(strSlice []string) []string {
+	var newStrSlice []string
+
+	for i := 0; i < len(strSlice); i++ {
```
jannotti, line 313: why not use a simpler `range` loop?

data/abi/abi_json.go
```diff
@@ -0,0 +1,254 @@
...
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"github.com/algorand/go-algorand/data/basics"
+	"math/big"
+)
+
+func castBigIntToNearestPrimitive(num *big.Int, bitSize uint16) (interface{}, error) {
+	if num.Cmp(new(big.Int).Lsh(big.NewInt(1), uint(bitSize))) >= 0 {
```
jannotti, line 28: I think you can just ask if num.BitLen() is too big.
https://pkg.go.dev/math/big#Int.BitLen

data/abi/abi_json_test.go
```diff
@@ -0,0 +1,118 @@
...
+			expected: []interface{}{},
+		},
+		{
+			input:    "[]",
+			typeStr:  "()",
+			expected: []interface{}{},
+		},
+		{
+			input:    "[65, 66, 67]",
+			typeStr:  "string",
```
jannotti, line 76: Check "[]" and typeStr "string". I guess that should give a blank string?

ahangsu: yes, it is still a blank string

cmd/goal/application.go
```diff
@@ -17,17 +17,21 @@
 package main
 
 import (
+	"bytes"
+	"crypto/sha512"
 	"encoding/base32"
 	"encoding/base64"
 	"encoding/binary"
 	"fmt"
+	"github.com/algorand/go-algorand/libgoal"
```
tzaffi, line 26: Should this import be grouped with the other `algorand` imports below?

jannotti (approved):

cmd/goal/application.go
ahangsu: yes

cmd/goal/application.go
```diff
@@ -995,3 +1047,140 @@ var infoAppCmd = &cobra.Command{
 		}
 	},
 }
+
+var methodAppCmd = &cobra.Command{
+	Use:   "method",
+	Short: "Invoke a method",
+	Long:  `Invoke a method in an App (stateful contract) with an application call transaction`,
```
tzaffi, line 1054: I'm not proposing any changes here, but view this as a learning opportunity for me about Algorand's conventions. Cleary the convention in the file is for this field to use raw string literals. I guess that makes sense, because in theory the `Long` description could span several lines. However, none of the descriptions in this file seem to utilize any of the string literal's special features. Any thoughts? You can also totally ignore.

ahangsu: I don't have specific thoughts or idea about it, just following the previous conventions. I just noticed the nearest editing for adding method/updating description is 16 months ago... and I suppose they might want to do something with string literal.

PR #3089, opened 2021-10-18
catchup: avoid requesting blocks that aren't needed by the ledger

tsachiherman:
## Summary

avoid requesting blocks that aren't needed by the ledger.

This isn't done for the sake of optimization - it's done so we won't have misleading warning messages in the log file due to frequently canceled contexts.

## Test Plan

tested manually.

### DIFF
--- a/catchup/service.go
+++ b/catchup/service.go
@@ -18,6 +18,7 @@ package catchup
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"sync"
 	"sync/atomic"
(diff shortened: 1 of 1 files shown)


catchup/service.go
```diff
@@ -218,6 +230,10 @@ func (s *Service) fetchAndWrite(r basics.Round, prevFetchCompleteChan chan bool,
 		block, cert, blockDownloadDuration, err := s.innerFetch(r, peer)
 
 		if err != nil {
+			if err == errLedgerAlreadyHasBlock {
```
algonautshant, line 233: This error is reported only when the ledger already has the block by the time innferFetch first statements are executed. 
However, when the innerFetch returns an error because the ledger already has the block, but the ledger had the block after the first innerFetch statements were executed, the error will not be reported as errLedgerAlreadyHasBlock.

Is this the expected behavior?

catchup/service.go
```diff
@@ -165,7 +177,7 @@ func (s *Service) innerFetch(r basics.Round, peer network.Peer) (blk *bookkeepin
 	go func() {
 		select {
 		case <-stopWaitingForLedgerRound:
-		case <-s.ledger.Wait(r):
+		case <-ledgerWaitCh:
```
algonautshant, line 180: Ideally, don't we want errLedgerAlreadyHasBlock returned when ledgerWaitCh fires?

tsachiherman: Sounds like a good idea. I'll make it so.

PR #3096, opened 2021-10-19
Nguo/measurelatency

nicholasguoalgorand:


tsachiherman:
could you please rebase this PR ?

tsachiherman (changes requested):
overall, I think that it looks good. Two small comments -
1. The current code is calling `s.node.GetPeerLatency` on every incoming message. That's not horrible wrong, but it would make more sense to cache the latency at the `Peer` object, and refresh it only on new round and/or new peer.
2. You've added quite many debug messages; while there is nothing wrong with that, I don't believe that there is a need to keep these on production code. Specifically for the transaction sync, it has its own logger - and the should be the way to log messages if we find that those are needed. ( I'm not sure if we really would get much value from the log messages added in this PR once we'll be over with these optimizations. ).

txnsync/incoming.go
```diff
@@ -120,7 +121,7 @@ func (s *syncState) asyncIncomingMessageHandler(networkPeer interface{}, peer *P
 		}
 	}()
 
-	incomingMessage := incomingMessage{networkPeer: networkPeer, sequenceNumber: sequenceNumber, encodedSize: len(message), peer: peer}
+	incomingMessage := incomingMessage{networkPeer: networkPeer, sequenceNumber: sequenceNumber, encodedSize: len(message), peer: peer, timeReceived: s.clock.Since()}
```
nicholasguoalgorand, line 124: So it turns out that using `s.clock.Since()` might be problematic since it causes a race condition. Additionally, the clock gets zeroed out when a new round starts. Is there a way around this?

tsachiherman: yes, but it's not super-trivial:
look at `func (tsnc *transactionSyncNodeConnector) Handle(raw network.IncomingMessage)`.
you'll see that `raw.Received` has the timestamp at which the message was received by the network socket.

If you'll change the `txnsync.IncomingMessageHandler` to include this field, you will be able to use it to determine absolute time. For the emulator, make sure that in case `Received` is zero, you have delay of zero.

nicholasguoalgorand: Does this mean for all timestamps made for the bandwidth calculations will need to be changed to use `time.Now()` instead of the WallClock? I haven't found a way to convert times from `time.Now()` to `s.clock.Since()`

tsachiherman (approved):
looks good.

PR #3097, opened 2021-10-19
crypto interfaces api fix

algonathan:
## Summary
Signer, Verifier, HashFactory had issues with unmarshaling and invalid types, as a result, there was a need to verify them in each usage. 
the following PR attempts to remove the need for validation, by doing it automatically with msgp directives.

### DIFF
--- a/netdeploy/network.go
+++ b/netdeploy/network.go
@@ -82,7 +82,8 @@ func CreateNetworkFromTemplate(name, rootDir, templateFile, binDir string, impor
 	}
 	template.Consensus = consensus
 
-	if isTest {
+	isTestFlag := isTest || (os.Getenv("ALGOTEST") != "")
+	if isTestFlag {
 		// Generate participation keys for a shorter period
 		template.Genesis.LastPartKeyRound = 3000
 	}


go.sum
```diff
@@ -15,6 +13,8 @@ github.com/algorand/graphtrace v0.0.0-20201117160756-e524ed1a6f64 h1:yvKeJdS/mvL
 github.com/algorand/graphtrace v0.0.0-20201117160756-e524ed1a6f64/go.mod h1:qFtQmC+kmsfnLfS9j3xgKtzsWyozemL5ek1R4dWZa5c=
 github.com/algorand/msgp v1.1.48 h1:5P+gVmTnk0m37r+rA3ZsFZW219ZqmCLulW5f8Z+3nx8=
```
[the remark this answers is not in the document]
algonathan: I'm really not sure, I''ve tried remaking the go.sum file too.

protocol/codec_tester.go
```diff
@@ -378,6 +378,13 @@ func RunEncodingTest(t *testing.T, template msgpMarshalUnmarshal) {
 			t.Skip()
 			return
 		}
+		if err == nil {
+			continue
+		}
+
+		if strings.Contains(err.Error(), ErrorInvalidObject.Error()) {
```
tsachiherman, line 385: can we use `errors.Is` or `errors.As` here ?

crypto/sig_abstractions.go
```diff
@@ -127,14 +144,14 @@ type PackedSignatureAlgorithm struct {
...
 	switch t {
 	case DilithiumType:
-		return &p.DilithiumSigner, nil
+		return &p.DilithiumSigner
 	case Ed25519Type:
-		return &p.Ed25519Singer, nil
+		return &p.Ed25519Singer
 	default:
-		return nil, errUnknownSigner
+		panic(errUnknownSigner)
```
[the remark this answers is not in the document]
tsachiherman: The problem with panics is that it exposes your code to an attack, and you need to actively add a "recover" method to work around that. This is a risky, since it's not obvious to the caller that the underlying code might have that condition. On the flip side, returning an error is the safer way, since the caller knows that it should test all the returned error codes.

compactcert/builder.go
```diff
@@ -18,10 +18,10 @@ package compactcert
 
 import (
 	"context"
-	"database/sql"
 	"encoding/binary"
 	"fmt"
 
+	"database/sql"
```
tsachiherman, line 24: nit: move to the previous group.

crypto/hashes.go
```diff
@@ -90,3 +109,32 @@ func HashBytes(hash hash.Hash, m []byte) []byte {
 	outhash := hash.Sum(nil)
 	return outhash
 }
+
+// InvalidHash is used to identify errors on the factory.
+// this function will return nil slice
+type InvalidHash struct {
```
tsachiherman, line 115: make `InvalidHash` into a small case `invalidHash` since there is no point where you would export this object.

crypto/sig_abstractions.go
```diff
@@ -102,18 +120,19 @@ type PackedVerifyingKey struct {
 
 	DilithiumPublicKey DilithiumVerifier `codec:"dpk"`
 	Ed25519PublicKey   Ed25519PublicKey  `codec:"edpk"`
+	InvalidVerifier    InvalidVerifier   `codec:"InvV"`
```
tsachiherman, line 123: no need to serialize invalid keys.

protocol/codec.go
```diff
@@ -26,6 +26,9 @@ import (
 	"github.com/algorand/msgp/msgp"
 )
 
+// ErrorInvalidObject is used to state that an object decoding has failed because it's invalid.
+var ErrorInvalidObject = fmt.Errorf("unmarshalled object is invalid")
```
tsachiherman, line 30: Errors are typically start with "Err" and not "Error".

tsachiherman: Also, since you're not really formatting anything in here, it should be "errors.New(...)"

PR #3098, opened 2021-10-19
testing: modify TestAccountInformationV2 to wait for transaction

tsachiherman:
## Summary

This change wait for a given broadcasted transaction to be included instead of expecting it to be included after a predetermined number of rounds. This would increase the testing platform flexibility in case we're testing on slow machines.

## Test Plan

This is a test.


test/e2e-go/features/transactions/accountv2_test.go
```diff
@@ -285,16 +287,23 @@ int 1
...
+		// Ensure the txn committed
+		resp, err = client.GetPendingTransactions(2)
+		a.NoError(err)
+		if resp.TotalTxns == 1 {
+			a.Equal(resp.TruncatedTxns.Transactions[0].TxID, txid)
+			continue
+		}
+		a.Equal(uint64(0), resp.TotalTxns)
+		break
+	}
```
algonautshant, line 306: checkEvalDelta on line 318 will still impose certain expectations on the round / test timing.

PR #3101, opened 2021-10-19
Add unit test to avoid new `string` fields

### LINKED ISSUE #3093: Avoid using `string` fields for binary data
## Summary
Recently the field `Logs []string` was added to `EvalDelta`. This field is meant to hold arbitrary binary data, but because it was declared with `[]string`, when encoding into JSON the data is interpreted as UTF-8. If instead it were declared as `[][]byte`, our encoder would base64 encode the contents when encoding into JSON. For this reason, it's necessary to use `[]byte` over `string` for binary data in all cases.

## Scope
- Add a unit test which uses reflection and recursion to examine all fields referenced from `bookkeeping.Block`. If the field is a string and we don't have a specific exception for it, error. This will prevent us from accidentally adding string fields going forward.
- For existing string fields, create exceptions so the test passes with the current structs. ~~However, investigate whether converting these fields to byte slices has any effect on their msgpack encoding -- if not, then convert them to byte slices to eliminate all strings from the structs.~~
  - Msgpack uses different byte prefixes to distinguish between strings and binary data, so we cannot migrate the existing `string` instances to `[]byte` without changing the canonical encoding of these structures.

jasonpaulos:
## Summary

Add a unit test which fails if any new `string` types are referenced from `bookkeeping.Block`.

Closes #3093. See that issue for more context.

## Test Plan

This is a test.


data/bookkeeping/encoding_test.go
```diff
@@ -66,3 +69,145 @@ func TestBlockWithTxnEncoding(t *testing.T) {
...
+}
+
+func TestBlockFields(t *testing.T) {
+	partitiontest.PartitionTest(t)
+
+	blockType := reflect.TypeOf(Block{})
+
+	// These exceptions are for pre-existing usages of string. Only add to this list if you really need to use string.
+	exceptions := []typePath{
+		typePath{}.addField("BlockHeader").addField("GenesisID"),
```
tsachiherman, line 172: would be nicer to have this as `Payset\[]SignedTxnWithAD\SignedTxn\Txn\AssetConfigTxnFields`, but given that this is a deterministic list, it's quite de-motivating to make the extra effort.

tsachiherman (dismissed):
looks great.

jasonpaulos:
Found the issue, the seen set was being too aggressive, meaning some paths weren't properly checked. Fixed now.

tolikzinovyev:
Would be nice to do the same for `AccountData`!

jannotti (approved):

PR #3104, opened 2021-10-20
Added HashType to ProofResponse REST endpoint

Aharonee:


crypto/hashes.go
```diff
@@ -65,6 +66,29 @@ type HashFactory struct {
...
+		return "sha512_256"
+	case Sumhash:
+		return "sumhash"
+	default:
+		return ""
+	}
+}
+
+// InitHashType decodes a string into the HashType enum
+func InitHashType(s string) (HashType, error) {
```
jasonpaulos, line 81: `UnmarshalHashType` would be a more consistent name for this.

The `Status` type from `data/basics/userBalance.go` already has `String` and `UnmarshalStatus` functions.

jasonpaulos (commented):
Looks good, just one comment

PR #3112, opened 2021-10-20
Disabling re-building of binaries in circleci during upload stage

algojack:


scripts/travis/deploy_packages.sh
```diff
@@ -24,7 +24,9 @@ then
     exit 1
 fi
 
-scripts/travis/build.sh
+if [ -z "${NO_BUILD}" ]; then
```
algobarb, line 27: this seems like a weird way to handle a boolean. -z is for checking if the length of string is zero. maybe it could be changed to if `"${NO_BUILD}"=true` or `"${NO_BUILD}"="true"` instead?

algobarb: I think it's because like you said NO_BUILD isn't set otherwise. Wouldn't it be better to separate the checks, if -z "${NO_BUILD}; then { NO_BUILD = "false"} and then do a "${NO_BUILD}"!=true check. Otherwise, setting NO_BUILD = "false" would have weird outcomes and it's harder to read this script to tell if we're checking if NO_BUILD is "true" or "false".

algobarb: The alternative is to use `if [ -z "${NO_BUILD}" || "${NO_BUILD}" != true ]; then` 
Isn't the commit you're referencing checking for whether if NO_BUILD is true, don't we only want to build it if it's not true?

algobarb: yeah exactly. The [commit used before](https://github.com/algorand/go-algorand/commit/413864a45405ec406b5a21e0b20b848f0442c161) was `if [ "${NO_BUILD}" = "true" ]; then` but it should have been `!=`

PR #3114, opened 2021-10-20
Expose `StateProofPK` in TEAL

jasonpaulos:
## Summary

Expose the new `StateProofPK` field from keyreg transactions to TEAL.

Note that I have not exposed compact cert transactions to TEAL because they cannot be part of atomic transaction groups, so AVM code should never see them.

## Test Plan

Unit tests added.


jannotti (approved):
This looks good. But it does make me notice this big thing we're adding to Transaction.  We copy those structs in many places.  Now we're adding (at least 64) bytes to each one.  Does StateProofPK have anything besides Root in it?  Maybe it should be a slice instead of an embedded array (more like Note?). But that's not an issue for this PR.

PR #3119, opened 2021-10-21
tealdbg: hide unexposed fields

barnjamin:
## Summary

In tealdbg, the txn obj has a number of fields that are only available after the transaction has been evaluated in ApplyData. At the moment we show "Unable to obtain effects from top-level transactions" from https://github.com/algorand/go-algorand/blob/master/data/transactions/logic/eval.go#L1939 . This should be hidden for now.

image


jannotti (commented):
Can you use fs.effects as the test, instead of the specific field names?

barnjamin:
you mean here? https://github.com/algorand/go-algorand/blob/master/data/transactions/logic/eval.go#L1887

the `txnFieldSpecByField` array is not exported so I think I'd have to push the test for these fields over there. 

Just skip appending to `fieldDesc` if I get an error back?

PR #3120, opened 2021-10-21
[WIP] Contract to contract calls.

jannotti:
Allow inner transactions to be ApplicationCalls.

This means that ApplyDatas may be recursive beyond one level, which will have Indexer implications. Let's discuss @winder @algorandskiy 

EvalDeltas are now maintained by the AVM while executing, rather than building them from the amassed changes in the cow/ledger. This is allows each inner transaction to contain only its own changes in its EvalDelta.

Many precautions, checks, and details need to be added.
- [x] Do not allow v5 and below as inner apps
- [x] Do not allow re-entrant calls
- [ ] Reinstate limits on the total number of inner txns under a top-level
- [x] `gitxn` to inspect the last group, rather than just the last txn with `itxn`

Some decisions remain
- [x] What are the semantics of `txn TxID` and `global GroupID` in inner app calls?
- [ ] Add a mechanism to directly opt-in/out/clear accounts (since re-entrancy ban precludes self call to do so)?
- [ ] Add mechanism to "self-destruct" an app?


data/transactions/logic/assembler.go
```diff
@@ -1216,9 +1225,12 @@ func asmTxField(ops *OpStream, spec *OpSpec, args []string) error {
 	if !ok {
 		return ops.errorf("txn unknown field: %#v", args[0])
 	}
-	_, ok = txnaFieldSpecByField[fs.field]
-	if ok {
-		return ops.errorf("found array field %#v in %s op", args[0], spec.Name)
+	// itxn_field appends to the field if it's an array
+	if spec.Name != "itxn_field" {
```
algorandskiy, line 1229: is it a new behavior?

data/transactions/logic/backwardCompat_test.go
```diff
@@ -399,37 +396,24 @@ func TestBackwardCompatGlobalFields(t *testing.T) {
 
 		ops := testProg(t, text, AssemblerMaxVersion)
 
-		proto := config.Consensus[protocol.ConsensusV23]
```
algorandskiy, line 402: why did it go? I'd like to do not modify these backward compat tests if possible

data/transactions/logic/eval.go
```diff
@@ -115,6 +115,16 @@ func (sv *stackValue) uint() (uint64, error) {
 	return sv.Uint, nil
 }
 
+func (sv *stackValue) uintMaxed(max uint64) (uint64, error) {
+	if sv.Bytes != nil {
+		return 0, errors.New("not a uint64")
+	}
+	if sv.Uint > max {
+		return 0, errors.New("too large")
```
algorandskiy, line 123: add values to the message for better troubleshooting

data/transactions/logic/eval.go
```diff
@@ -158,7 +168,7 @@ func stackValueFromTealValue(tv *basics.TealValue) (sv stackValue, err error) {
 // versions of TEAL. If one of the transactions in a group will execute a TEAL
 // program whose version predates a given field, that field must not be set
 // anywhere in the transaction group, or the group will be rejected.
-func ComputeMinTealVersion(group []transactions.SignedTxn) uint64 {
```
algorandskiy, line 161: update the comment

data/transactions/logic/eval.go
```diff
@@ -350,12 +434,17 @@ type EvalContext struct {
 	scratch scratchSpace
 
 	subtxns []transactions.SignedTxn // place to build for itxn_submit
-	// Previous transactions Performed() and their effects
-	InnerTxns []transactions.SignedTxnWithAD
-
-	cost    int // cost incurred so far
-	Logs    []string
```
algorandskiy, line 357: where did it go?

data/transactions/logic/eval.go
```diff
@@ -350,12 +434,17 @@ type EvalContext struct {
...
+	logSize int                      // total log size so far
+
+	// Summary of changes.  We used to compute this from the ledger after a full
+	// eval.  But now apps can call apps.  When they do, all of the changes
+	// accumulate into the parent's ledger, but EvalDelta should only have the
+	// changes from *this* call. (The changes caused by children are deeper
+	// inside - in the EvalDeltas of the InnerTxns inside this EvalDelta) Nice
+	// bonus - by keep the running changes, the debugger can be changed to
+	// display them as the app runs.
+	EvalDelta transactions.EvalDelta
```
algorandskiy, line 447: FYI, TEAL v2 original™ had it - all state deltas were computed here :)

data/transactions/logic/eval.go
```diff
@@ -2347,53 +2393,108 @@ func opGtxnsas(cx *EvalContext) {
...
+			return inners[i+1:]
+		}
+	}
+	// All have the same (non-zero) group. Return all
+	return inners
+}
+
+func opGitxn(cx *EvalContext) {
+	lastInnerGroup := cx.getLastInnerGroup()
+	gtxid := cx.program[cx.pc+1]
```
algorandskiy, line 2461: gtxid -> gtxIdx ? since the opcode expects group index, not group id

data/transactions/logic/eval.go
```diff
@@ -3373,7 +3474,12 @@ func opAppLocalPut(cx *EvalContext) {
 
 	addr, accountIdx, err := cx.accountReference(cx.stack[pprev])
 	if err == nil {
-		err = cx.Ledger.SetLocal(addr, key, sv.toTealValue(), accountIdx)
+		tv := sv.toTealValue()
+		if _, ok := cx.EvalDelta.LocalDeltas[accountIdx]; !ok {
+			cx.EvalDelta.LocalDeltas[accountIdx] = basics.StateDelta{}
+		}
+		cx.EvalDelta.LocalDeltas[accountIdx][key] = tv.ToValueDelta()
+		err = cx.Ledger.SetLocal(addr, key, tv, accountIdx)
```
algorandskiy, line 3482: do we need to test matching cx.EvalDelta to Ledger's deltas? Maybe modify `SetLocal` and Co to return  deltas in order to have such logic in a single place?

data/transactions/logic/eval.go
```diff
@@ -3976,6 +4195,15 @@ func opTxSubmit(cx *EvalContext) {
 		*cx.FeeCredit = basics.AddSaturate(*cx.FeeCredit, overpay)
 	}
 
+	cx.appStack = append(cx.appStack, cx.Ledger.ApplicationID()) // push app
+
+	// All subtxns must have zero'd GroupID now, since it can't be set
```
algorandskiy, line 4200: why is the comment? We do not check/enforce this

ledger/acctupdates.go
```diff
@@ -879,6 +879,11 @@ func (aul *accountUpdatesLedgerEvaluator) GenesisHash() crypto.Digest {
 	return aul.au.ledger.GenesisHash()
 }
 
+// GenesisProto returns the genesis consensus params
+func (aul *accountUpdatesLedgerEvaluator) GenesisProto() config.ConsensusParams {
+	return aul.au.ledger.GenesisProto()
```
algorandskiy, line 884: where is it used?

ledger/internal/eval.go
```diff
@@ -952,12 +913,14 @@ func (eval *BlockEvaluator) transaction(txn transactions.SignedTxn, evalParams *
 		}
 	}
 
-	// We are not allowing InnerTxns to have InnerTxns yet.  Error if that happens.
-	for _, itx := range applyData.EvalDelta.InnerTxns {
-		if len(itx.ApplyData.EvalDelta.InnerTxns) > 0 {
-			return fmt.Errorf("inner transaction has inner transactions %v", itx)
+	/*
```
algorandskiy, line 916: remove?

data/transactions/logic/assembler.go
jannotti: Yes, because previously no array fields were allowed (they are all in ApplicationCalls).  I did just remove the if test though, since this asm function is only used for itxn_field, it should simply allow all fields.

The versioning mechanism on fields prevents this from changing behavior on old code.  v5 is always going to reject the array fields.

data/transactions/logic/backwardCompat_test.go
jannotti: Because the goal was achieved by giving a version argument to `defaultEvalParamsWithVersion`

data/transactions/logic/eval.go
jannotti: I will do that.  Though I'm not sure these errors ever propagate to the user - they get swallowed by a generic logic error message I think.  But let's make them nice and maybe we'll expose them better later.

data/transactions/logic/eval.go
jannotti: cost remains, but InnerTxns and Logs are now kept in the EvalDelta that we are updating as we go anyway.

data/transactions/logic/eval.go
jannotti: I don't think we can. The reason we're keeping track of them here is that the ledger is accumulating too many updates - it contains updates from inner calls as well.  I suppose we could try to sort it out - reapply them from all the inners and see if it comes out the same. But that feels like a lot of work - interpret all changes twice - to see if they match.

data/transactions/logic/eval.go
jannotti: It is enforced because there is no GroupId field for `itxn_field`.  So it is not settable in an inner txn.

ledger/acctupdates.go
jannotti: I added this while trying to merge recent ledger changes. I has aded GenesisProto to the new Ledger interface.  But after discussion with @tsachiherman , that interface may be gone now.  I'll look around some more and clean up.

ledger/internal/eval.go
jannotti: yes, thanks.

ledger/internal/eval_test.go
```diff
@@ -22,7 +22,6 @@ import (
 	"errors"
 	"fmt"
 	"math/rand"
-	"reflect"
```
tzaffi, line 25: nice!

jannotti: Don't get too excited, I just moved the single test that used reflection into another package. But it still uses reflection.

I don't worry too much about using reflection in tests, do you?  Makes for easy deep equality checks.

winder: I'm pretty sure testify uses DeepEqual internally.

jannotti:
Closing in favor of a feature branch: feature/contract-to-contract

PR #3132, opened 2021-10-22
ledger: do final validation in `endOfBlock()`

tolikzinovyev:
## Summary

1. It doesn't make sense to calculate new totals in `finalValidation()`
2. `endOfBlock()` already does some validation

Moving the content of `finalValidation()` to `endOfBlock()` makes things more reasonable.

### DIFF
--- a/ledger/internal/eval.go
+++ b/ledger/internal/eval.go
@@ -1053,6 +1053,8 @@ func (eval *BlockEvaluator) TestingTxnCounter() uint64 {
 
 // Call "endOfBlock" after all the block's rewards and transactions are processed.
 func (eval *BlockEvaluator) endOfBlock() error {
+	eval.state.mods.OptimizeAllocatedMemory(eval.proto)
+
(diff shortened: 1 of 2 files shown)


tsachiherman (changes requested):
This change looks pretty good. It does change the semantics of `eval.finalValidation` during `Eval`, as it would now be performing the signature validation in parallel to the `eval.finalValidation`. That being said - I can't find a reason this would cause an issue.

Separately, the `eval.state.mods.OptimizeAllocatedMemory` call was intended to be called once we have the state delta changes ready. At their new location, this is not the case, since the `eval.resetExpiredOnlineAccountsParticipationKeys` could modify their content.

To resolve that, just move the `eval.state.mods.OptimizeAllocatedMemory` right before the `if eval.validate` predicate.

tolikzinovyev:
done

tsachiherman (approved):

PR #3141, opened 2021-10-26
Accounts endpoint update

algonathan:


test/e2e-go/restAPI/restClient_test.go
```diff
@@ -1051,3 +1051,160 @@ return
...
+			FirstValid:  firstRound,
+			LastValid:   lastRound,
+			GenesisHash: gh,
+		},
+		KeyregTxnFields: transactions.KeyregTxnFields{
+			VotePK:      votePK,
+			SelectionPK: selPK,
+			VoteFirst:   firstRound,
+			StateProofPK: merklekeystore.Verifier{
+				Root: [merklekeystore.KeyStoreRootSize]byte{1, 2, 3, 4},
```
jasonpaulos, line 1110: It would be slightly better to use a randomly generated 64 byte string for this, like the other participation keys, so we can be sure all bytes beyond the first 4 are being handled properly.

jasonpaulos (commented):
This looks pretty good, just one suggestion

PR #3164, opened 2021-10-29
Participation Key Interface

winder:
## Summary

* new `ParticipationRegistry` interface designed to manage keys and collect usage metrics.
* new REST endpoints for interacting with keys.
* improved `goal account partkeyinfo` and `goal account listpartkeys` formatting.

## Test Plan

New unit and integration tests.


cmd/goal/account.go
```diff
@@ -1065,37 +1076,53 @@ var listParticipationKeysCmd = &cobra.Command{
...
-		for fn := range parts {
-			filenames = append(filenames, fn)
-		}
-		sort.Strings(filenames)
-
-		rowFormat := "%-10s\t%-80s\t%-60s\t%12s\t%12s\t%12s\n"
-		fmt.Printf(rowFormat, "Registered", "Filename", "Parent address", "First round", "Last round", "First key")
-		for _, fn := range filenames {
+		// Squeezed this into 77 characters.
+		rowFormat := "%-10s  %-11s  %-15s  %10s  %11s  %10s\n"
```
jannotti, line 1080: This is something that might change with new part keys.  I think they will be longer.

cmd/goal/account.go
```diff
@@ -1065,37 +1076,53 @@ var listParticipationKeysCmd = &cobra.Command{
...
+						onlineInfoStr = "yes"
+					} else {
+						onlineInfoStr = "no"
+					}
+				*/
+
+				// it's okay to proceed without algod info
+				lastUsed := maxRound(0, part.LastStateProof)
+				lastUsed = maxRound(lastUsed, part.LastBlockProposal)
+				lastUsed = maxRound(lastUsed, part.LastStateProof)
```
jannotti, line 1113: Why does this fold in part.LastStateProof twice?

winder: Thanks, I fixed that yesterday but must have clobbered it with a bad merge. The first one should be `LastVote`.

agreement/abstractions.go
```diff
@@ -225,6 +225,10 @@ type KeyManager interface {
 	// valid for the provided votingRound, and were available at
 	// keysRound.
 	VotingKeys(votingRound, keysRound basics.Round) []account.Participation
+
+	// RecordAsync indicates that the given participation action has been taken.
+	// The operation needs to be asynchronous to avoid impacting agreement.
+	RecordAsync(account basics.Address, round basics.Round, participationType account.ParticipationAction)
```
tsachiherman, line 231: Could you rename this method to `RecordAsync` -> `RecordVote` ?

I think that the notation is "Async" should not be placed here, but rather on the implementation side.

agreement/pseudonode_test.go
```diff
@@ -222,7 +222,12 @@ func TestPseudonode(t *testing.T) {
 		}
 		messageEvent, typeOk := ev.(messageEvent)
 		assert.True(t, true, typeOk)
+		// Verify votes are recorded - everyone is voting and proposing blocks.
+		keyManager.mutex.Lock()
+		assert.Equal(t, startRound, keyManager.recording[messageEvent.Input.Vote.R.Sender][account.Vote])
+		assert.Equal(t, startRound, keyManager.recording[messageEvent.Input.Vote.R.Sender][account.BlockProposal])
 		events[messageEvent.t()] = append(events[messageEvent.t()], messageEvent)
+		keyManager.mutex.Unlock()
```
tsachiherman, line 230: Could you create a key manager method 
```golang
ValidateVoteRound(t, voteRound, act, account.ParticipationAction)
```
that would take the lock, conduct the test and return ?
i.e. given that this key manager is a testing key manager, it's perfectly fine to add testing method to it.

daemon/algod/api/server/v2/handlers.go
```diff
@@ -67,6 +68,144 @@ type NodeInterface interface {
...
+	GetParticipationKey(account.ParticipationID) (account.ParticipationRecord, error)
+	RemoveParticipationKey(account.ParticipationID) error
+}
+
+func roundToPtrOrNil(value basics.Round) *uint64 {
+	if value == 0 {
+		return nil
+	}
+	result := uint64(value)
+	return &result
```
tsachiherman, line 82: A participation key can have it's starting round at round 0.

daemon/algod/api/server/v2/handlers.go
```diff
@@ -67,6 +68,144 @@ type NodeInterface interface {
...
+	// These are pointers but should always be present.
+	if record.Voting != nil {
+		participationKey.Key.VoteParticipationKey = record.Voting.OneTimeSignatureVerifier[:]
+	}
+	if record.VRF != nil {
+		participationKey.Key.SelectionParticipationKey = record.VRF.PK[:]
+	}
+
+	// Optional fields.
+	participationKey.EffectiveFirstValid = roundToPtrOrNil(record.EffectiveFirst)
```
tsachiherman, line 105: A value of zero here could be a valid round number.

tsachiherman: ohh - I see, you added a special case later on. This is fine - but please add a note here.

I think that I would rather seeing a more direct approach where each variable is set exactly one time; but in this case, just adding a note could be enough.

daemon/algod/api/server/v2/handlers.go
```diff
@@ -67,6 +68,144 @@ type NodeInterface interface {
...
+	if err != nil {
+		return badRequest(ctx, err, err.Error(), v2.Log)
+	}
+
+	err = v2.Node.RemoveParticipationKey(decodedParticipationID)
+
+	if err != nil {
+
+		if errors.Is(err, account.ErrParticipationIDNotFound) {
+			return ctx.JSON(http.StatusOK, generated.ErrorResponse{Message: "participation id not found"})
```
tsachiherman, line 181: This doesn't sounds like a "StatusOK" issue to me.. shouldn't it be `http.StatusNotFound` ?

daemon/algod/api/server/v2/handlers.go
```diff
@@ -67,6 +68,144 @@ type NodeInterface interface {
...
+		}
+
+		return badRequest(ctx, err, err.Error(), v2.Log)
```
tsachiherman, line 184: `badRequest` -> `internalError`

daemon/algod/api/server/v2/handlers.go
```diff
@@ -67,6 +68,144 @@ type NodeInterface interface {
...
+	decodedParticipationID, err := account.ParticipationIDFromString(participationID)
+
+	if err != nil {
+		return badRequest(ctx, err, err.Error(), v2.Log)
+	}
+
+	participationRecord, err := v2.Node.GetParticipationKey(decodedParticipationID)
+
+	if err != nil {
+		return badRequest(ctx, err, err.Error(), v2.Log)
```
tsachiherman, line 203: as before, you should check `account.ErrParticipationIDNotFound` and issue `http.StatusNotFound` and/or `internalError`

data/account/participationRegistry.go
```diff
@@ -0,0 +1,947 @@
...
+	return (crypto.Digest(pid)).IsZero()
+}
+
+// String prints a b32 version of this ID.
+func (pid ParticipationID) String() string {
+	return base32.StdEncoding.WithPadding(base32.NoPadding).EncodeToString(pid[:])
+}
+
+// ParticipationIDFromString takes a string and returns a ParticipationID object
+func ParticipationIDFromString(str string) (d ParticipationID, err error) {
```
tsachizehub, line 52: nit: maybe rename this one to "ParseParticipationID" ? it already has the correct signature.

data/account/participationRegistry.go
```diff
@@ -0,0 +1,947 @@
...
+	decoded, err := base32.StdEncoding.WithPadding(base32.NoPadding).DecodeString(str)
+	if err != nil {
+		return d, err
+	}
+	if len(decoded) != len(d) {
+		msg := fmt.Sprintf(`Attempted to decode a string which was not a participation id: "%v"`, str)
+		return d, errors.New(msg)
```
tsachizehub, line 59: 1. use `fmt.Errorf` here instead of `fmt.Sprintf` + `errors.New`.
2. `%v` -> `%s`

data/account/participationRegistry.go
```diff
@@ -0,0 +1,947 @@
...
+	LastBlockProposal basics.Round
+	LastStateProof    basics.Round
+	EffectiveFirst    basics.Round
+	EffectiveLast     basics.Round
+
+	VRF    *crypto.VRFSecrets
+	Voting *crypto.OneTimeSignatureSecrets
+}
+
+var zeroParticipationRecord = ParticipationRecord{}
```
tsachizehub, line 84: This looks like it should have been a const, right ?

data/account/participationRegistry.go
```diff
@@ -0,0 +1,947 @@
...
+
+// MakeParticipationRegistry creates a db.Accessor backed ParticipationRegistry.
+func MakeParticipationRegistry(accessor db.Pair, log logging.Logger) (ParticipationRegistry, error) {
+	return makeParticipationRegistry(accessor, log)
+}
+
+// makeParticipationRegistry creates a db.Accessor backed ParticipationRegistry.
+func makeParticipationRegistry(accessor db.Pair, log logging.Logger) (*participationDB, error) {
+	if log == nil {
+		return nil, fmt.Errorf("invalid logger provided")
```
tsachizehub, line 193: convert this `fmt.Errorf` to `errors.New`, since you're not formatting anything in here.

data/account/participationRegistry.go
```diff
@@ -0,0 +1,947 @@
...
+		log:        log,
+		store:      accessor,
+		writeQueue: make(chan partDBWriteRecord, 10),
+	}
+	registry.flushDone = sync.NewCond(&registry.mutex)
+	go registry.writeThread()
+
+	err = registry.initializeCache()
+	if err != nil {
+		return nil, fmt.Errorf("unable to initialize participation registry cache: %w", err)
```
tsachizehub, line 216: is there a missing `accessor.Close()` here ?

data/account/participationRegistry.go
```diff
@@ -0,0 +1,947 @@
...
+	}
+
+	return registry, nil
+}
+
+// Queries
+var (
```
tsachizehub, line 223: aren't these constants ?

data/account/participationRegistry.go
```diff
@@ -0,0 +1,947 @@
...
+	// For performance reasons, state proofs are in a separate table.
+	_, err = tx.Exec(createStateProof)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// participationDB is a private implementation of ParticipationRegistry.
```
tsachizehub, line 304: maybe rephrase : 
```golang
participationDB provides a concrete implementation of the ParticipationRegistry interface.
```

data/account/participationRegistry.go
```diff
@@ -0,0 +1,947 @@
...
+
+	flush bool
+}
+
+func (db *participationDB) initializeCache() error {
+	db.mutex.Lock()
+	defer db.mutex.Unlock()
+
+	db.cache = make(map[ParticipationID]ParticipationRecord)
+	db.dirty = make(map[ParticipationID]struct{})
```
tsachizehub, line 344: as a defensive coding, I'd like to suggest that you'll defer the assignments of db.cache/db.dirty until after the for-loop. In case of an error, there is no reason for these to be populated.

daemon/algod/api/server/v2/handlers.go
winder: I'll move that check to here so that we only assign the value once.

data/account/participationRegistry.go
```diff
@@ -0,0 +1,947 @@
...
+			rawVRF)
+		if err != nil {
+			return fmt.Errorf("unable to insert keyset: %w", err)
+		}
+		rows, err := result.RowsAffected()
+		if err != nil {
+			return fmt.Errorf("unable to insert keyset: %w", err)
+		}
+		if rows != 1 {
+			return fmt.Errorf("unexpected number of rows")
```
tsachizehub, line 439: it's good that you've tested the number of rows - but in case there is an inconsistency, make sure to add the number of actual rows to the error so that we can figure out what went wrong.

data/account/participationRegistry.go
```diff
@@ -0,0 +1,947 @@
...
+		result, err = tx.Exec(insertRollingQuery, pk, rawVoting)
+		if err != nil {
+			return fmt.Errorf("unable insert rolling: %w", err)
+		}
+		rows, err = result.RowsAffected()
+		if err != nil {
+			return fmt.Errorf("unable to insert keyset: %w", err)
+		}
+		if rows != 1 {
+			return fmt.Errorf("unexpected number of rows")
```
tsachizehub, line 456: same as above.

data/account/participationRegistry.go
```diff
@@ -0,0 +1,947 @@
...
+		// Fetch primary key
+		var pk int
+		row := tx.QueryRow(selectPK, id[:])
+		err := row.Scan(&pk)
+		if err != nil {
+			return fmt.Errorf("unable to scan pk: %w", err)
+		}
+
+		// Delete rows
+		_, err = tx.Exec(deleteKeysets, pk)
```
tsachizehub, line 508: verify that 1 row was deleted.

daemon/algod/api/server/v2/handlers.go
winder: Stephen and I had the same discussion and this is what we decided on. I don't think there's a good single answer.

On one hand: I requested a delete state change. If there is nothing to delete there was no state change so return an error.

On the other: I want to make sure that the specified object does not exist in the system. The fact it doesn't exist at all is irrelevant, the final state is still what was requested.

data/account/participationRegistry.go
```diff
@@ -0,0 +1,947 @@
...
+	err := db.asyncErr
+	db.flushesPending--
+	// this allows all waiting Flush()es to see the error that happened
+	if db.flushesPending == 0 {
+		db.asyncErr = nil
+	}
+	return err
+}
+
+func (db *participationDB) Close() {
```
tsachizehub, line 940: We need to add a blocking here until the flushing thread has aborted. Just closing the `writeQueue` isn't enough, as the thread could be still busy doing writing.

node/node.go
```diff
@@ -743,6 +752,16 @@ func (node *AlgorandFullNode) GetPendingTxnsFromPool() ([]transactions.SignedTxn
...
+// ensureParticipationDB opens or creates a participation DB.
+func ensureParticipationDB(genesisDir string, log logging.Logger) (account.ParticipationRegistry, error) {
+	accessorFile := filepath.Join(genesisDir, config.ParticipationRegistryFilename)
+	accessor, err := db.OpenPair(accessorFile, false)
+	if err != nil {
+		return nil, err
+	}
+	return account.MakeParticipationRegistry(accessor, log)
+}
+
```
tsachizehub, line 764: Ideally, this logic would also be part of the key manager/participation registry. But it doesn't have to be part of this PR.

node/node.go
```diff
@@ -761,6 +780,143 @@ func (node *AlgorandFullNode) checkForParticipationKeys() {
...
+	if err != nil {
+		os.Remove(tempFile)
+		return "", err
+	}
+
+	return tempFile, nil
+}
+
+// InstallParticipationKey Given a participation key binary stream install the participation key
+func (node *AlgorandFullNode) InstallParticipationKey(partKeyBinary *[]byte) (account.ParticipationID, error) {
```
tsachizehub, line 864: 1. streaming SQLlite files is a hack that we shouldn't do. Instead, stream a msgp of the `account.Participation`.
2. The method should receive `partKeyBinary []byte`; no need for the pointer.

data/account/participationRegistry.go
winder: Unfortunately, no const structs are allowed. This is a (maybe) optimization we talked about earlier for the `IsZero` condition, do you think I should remove this object and have an empty one declared in the function?

data/account/participationRegistry.go
```diff
@@ -0,0 +1,947 @@
...
+	db.mutex.Lock()
+	defer db.mutex.Unlock()
+
+	db.flushesPending++
+	db.writeQueue <- partDBWriteRecord{
+		flush: true,
+	}
+
+	// See writeThread() for the Broadcast() we're Wait()ing on.
+	db.flushDone.Wait()
```
tsachizehub, line 930: this is a "problematic" implementation, as it would get stuck if the flushing thread is not working.

data/account/participationRegistry.go
```diff
@@ -0,0 +1,947 @@
...
+}
+
+func (db *participationDB) writeThread() {
+	needsFlush := false
+	var err error
+	var lastErr error
+	for {
+		var wr partDBWriteRecord
+		var chanOk bool
+		// A call to .Flush() waits until the flash command passes the command queue, and also that the command queue is empty.
```
tsachizehub, line 369: `flash` -> `flush`

data/account/participationRegistry.go
winder: I think it's clearer this way, I don't think we should optimize an error case for the garbage collector at the expense of readability. Maybe if we expected errors to be hit frequently, but this would cause the system to stop anyway so I don't think there's any benefit.

data/account/participationRegistry.go
winder: thanks, it looks like some of these error messages are also ambiguous so I'll make sure they're all unique as well.

node/node.go
winder: Maybe... maybe not. Passing in the accessor is nice since our db framework allows for in-memory databases, so testing is very easy. I could see hiding the participation registry in the account manager though.

node/node.go
winder: don't worry, there are tickets for it:
https://app.zenhub.com/workspace/o/algorand/go-algorand/issues/3107
https://app.zenhub.com/workspace/o/algorand/go-algorand/issues/3109
https://app.zenhub.com/workspace/o/algorand/go-algorand/issues/3106

agreement/pseudonode.go
```diff
@@ -441,6 +441,7 @@ func (t pseudonodeVotesTask) execute(verifier *AsyncVoteVerifier, quit chan stru
 	for _, r := range verifiedResults {
 		select {
 		case t.out <- messageEvent{T: voteVerified, Input: r.message, Err: makeSerErr(r.err)}:
+			t.node.keys.RecordAsync(r.v.R.Sender, r.v.R.Round, account.Vote)
```
cce, line 444: Talked to Will about this in person, but I was curious why you are recording at the moment you put a vote messages on the channel, rather than adding a new for-loop over `verifiedResults` to record all the votes like the telemetry/logging reporter is doing just above. Is it try and capture a time closer to when the votes got sent to the network?

winder: I think either would be fine here is the previous discussion: https://github.com/algorand/go-algorand/pull/2677#discussion_r681774709

tsachiherman: I think it's better here since it's being done after the `t.persistStateDone` ( i.e. write to the crash db is complete ).
Ultimately, we could place it when we write to the network package, but that's likely to be an overkill.

daemon/algod/api/server/v2/handlers.go
tsachiherman: I think the answer is in the REST definitions; look in https://www.restapitutorial.com/lessons/httpmethods.html : 
for the DELETE command, the response should be : 
200 (OK). 
404 (Not Found), if ID not found or invalid.

data/account/participationRegistry.go
tsachiherman: no, I think this one is fine.

data/account/participationRegistry.go
tsachiherman: I don't think about this one as an optimization. I was thinking of having the function making no changes in the case of an error. ( i.e. so that the state hasn't changed )

node/node.go
tsachiherman: I wasn't suggesting that the key manager/participation registry would always be working with the read database accessor. I was just saying that the loading portion would all be in one place. It's perfectly fine to have key manager to allow the caller to modify/select a different accessor.

daemon/algod/api/server/v2/handlers.go
winder: Interesting, this page goes into more detail further down:
```
There is a caveat about DELETE idempotence, however. Calling DELETE on a resource a second time will often return a 404 (NOT FOUND) since it was already removed and therefore is no longer findable. This, by some opinions, makes DELETE operations no longer idempotent, however, the end-state of the resource is the same. Returning a 404 is acceptable and communicates accurately the status of the call.
```

I see a lot of debate on the subject but the majority seem to side with the 404 approach, so I'll change it.

data/account/participationRegistry.go
winder: Ah, got it, yes that makes more sense to me. So I'll create a temporary variable and update everything at the end. Otherwise, on an error, the cache has an inconsistent state.

data/account/participationRegistry.go
winder: Fixing with https://github.com/algorand/go-algorand/pull/3191

data/account/participationRegistry.go
winder: Fixing with https://github.com/algorand/go-algorand/pull/3191

cmd/goal/account.go
winder: Yes, this is a "summary" view, so I took some liberties and did things like truncate hashes/addresses

tsachiherman (approved):

PR #3165, opened 2021-10-29
make proposal assembly time configurable

tsachiherman:
## Summary

This PR is doing the following:
1. Moves the proposal deadline computation outside of the agreement package and onto the node package. The agreement wasn't doing anything with the deadline anyway.
2. Make the proposal deadline be configured using a configuration parameter rather then a global hard-coded value in the config package.

## Test Plan

Existing unit test provides sufficient coverage for this change.


cce (approved):
LGTM! I was trying to think of any use cases where agreement would want to provide a deadline to AssembleBlock, though we're not doing this today.

When agreement enters a new round, the "assemble" pseudonodeAction and rezeroAction are queued up consecutively, so agreement's call to time.Now() before calling AssembleBlock and then its call to time.Now() in Clock.Zero() are roughly the same moment. In that scenario (new round, period 0), having transactionPool pick a deadline on its own is roughly equivalent to having agreement do it, because agreement doesn't have any more information how soon a block is needed (when the next vote is coming) than transactionPool does.

In the scenario where it's not period 0, however, and we are using different values of AgreementFilterTimeoutPeriod0 and AgreementFilterTimeout, you could imagine a scenario where agreement might want to tell the pool it has a little more time to put a block together, perhaps to relieve whatever resource issue was causing difficulty in period 0. If we ever do that, we could provide a hint to the pool using either a deadline or some other argument later.

tsachiherman:
> LGTM! I was trying to think of any use cases where agreement would have information it would want to use to set a deadline for AssembleBlock, though we're not doing this today.
> 
> When agreement enters a new round, the "assemble" pseudonodeAction and rezeroAction are queued up consecutively, so agreement's call to time.Now() before calling AssembleBlock, and then rezeroAction's call to time.Now() in Clock.Zero() are roughly the same moment. In that scenario (new round, period 0), having transactionPool pick a deadline on its own is roughly equivalent to having agreement do it, and agreement doesn't have any more information how soon a block is needed (when the next vote is coming) than transactionPool does.
> 
> In the scenario where it's not period 0, however, and we are using different values of AgreementFilterTimeoutPeriod0 and AgreementFilterTimeout, you could imagine a scenario where agreement might want to tell the pool it has a little more time to put a block together, perhaps to relieve whatever resource issue was causing difficulty in period 0. If we ever do that, we could provide a hint to the pool using either a deadline or some other argument later.

Yes, I agree with that. The agreement could pass a hint that would adjust the timing for the assembly.

PR #3168, opened 2021-10-29
txnsync: Use dynamic maxEncodedTransactionGroups

nicholasguoalgorand:
## Summary

Change constants in the txnsync around message size limits into variables based on maxTxPoolSize. This will allow for support of larger message sizes when dealing with larger load.

## Test Plan

Reran existing tests

### DIFF
--- a/txnsync/incoming.go
+++ b/txnsync/incoming.go
@@ -59,7 +59,7 @@ func (s *syncState) asyncIncomingMessageHandler(networkPeer interface{}, peer *P
 	_, err = incomingMessage.message.UnmarshalMsg(message)
 	if err != nil {
 		// if we received a message that we cannot parse, disconnect.
-		s.log.Infof("received unparsable transaction sync message from peer. disconnecting from peer: %v, bytes: %v", err, len(message))
+		s.log.Infof("received unparsable transaction sync message from peer. disconnecting from peer: %v, bytes: %d", err, len(message))
 		s.incomingMessagesQ.erase(peer, networkPeer)
 		return err
 	}
--- a/txnsync/incoming.go
+++ b/txnsync/incoming.go
@@ -59,7 +59,7 @@ func (s *syncState) asyncIncomingMessageHandler(networkPeer interface{}, peer *P
 	_, err = incomingMessage.message.UnmarshalMsg(message)
 	if err != nil {
 		// if we received a message that we cannot parse, disconnect.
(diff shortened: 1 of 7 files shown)


txnsync/incoming.go
```diff
@@ -59,7 +59,7 @@ func (s *syncState) asyncIncomingMessageHandler(networkPeer interface{}, peer *P
 	_, err = incomingMessage.message.UnmarshalMsg(message)
 	if err != nil {
 		// if we received a message that we cannot parse, disconnect.
-		s.log.Infof("received unparsable transaction sync message from peer. disconnecting from peer.")
+		s.log.Infof("received unparsable transaction sync message from peer. disconnecting from peer: %v, bytes: %v", err, len(message))
```
tsachiherman, line 62: nit: change the second formatter to `%d`

txnsync/service.go
```diff
@@ -51,6 +51,13 @@ func MakeTransactionSyncService(log logging.Logger, conn NodeConnector, isRelay
 	}
 	s.state.service = s
 	s.state.xorBuilder.MaxIterations = 10
+	if cfg.TxPoolSize > maxEncodedTransactionGroups {
```
tsachiherman, line 54: This looks correct, but think it would be cleaner if you'll move all that logic into a single method which would be called by both `init()` as well as from here.

tsachiherman: init methods are called when the package is loaded. In this case, instead of initializing the global variables where they are declared, call
```golang
func init() {
    setTransactionSyncVariables(config.GetDefaultLocal())
}
...
func setTransactionSyncVariables(cfg config.Local) {
        if cfg.TxPoolSize < maxEncodedTransactionGroups {
              return
        }
 	maxEncodedTransactionGroups = cfg.TxPoolSize
 	maxEncodedTransactionGroupEntries = cfg.TxPoolSize
 	maxBitmaskSize = (maxEncodedTransactionGroupEntries+7)/8 + 1
 	maxSignatureBytes = maxEncodedTransactionGroupEntries * len(crypto.Signature{})
 	maxAddressBytes = maxEncodedTransactionGroupEntries * crypto.DigestSize

 	maxBloomFilterSize = cfg.TxPoolSize * 5 // 32 bit xor uses slightly more than 4 bytes/element.
 	maxEncodedTransactionGroupBytes = cfg.TxPoolSize * 10000 // assume each transaction takes 10KB, as a worst-case-scenario for bounding purposes only.
 }
```

tsachiherman (approved):

PR #3171, opened 2021-11-01
Add MT missing bounds

algonautshant:
## Summary

Some structures are not properly bounded for the msgp.
This PR is adding bounds to those.


crypto/merklearray/merkle.go
```diff
@@ -31,15 +31,15 @@ type Tree struct {
...
+	Levels []Layer            `codec:"lvls,allocbound=9999"`
 	Hash   crypto.HashFactory `codec:"hsh"`
 }
 
 // Proof contains the merkle path, along with the hash factory that should be used.
 type Proof struct {
 	_struct struct{} `codec:",omitempty,omitemptyarray"`
 
-	Path        []crypto.GenericDigest `codec:"pth,allocbound=-"`
+	Path        []crypto.GenericDigest `codec:"pth,allocbound=9999"`
```
tsachiherman, line 42: Please add a constant for this one. The actual number should be much smaller than 9999. Most likely 256 is a good limit.

crypto/merklearray/merkle.go
```diff
@@ -31,15 +31,15 @@ type Tree struct {
 	_struct struct{} `codec:",omitempty,omitemptyarray"`
 
 	// Level 0 is the leaves.
-	Levels []Layer            `codec:"lvls,allocbound=-"`
+	Levels []Layer            `codec:"lvls,allocbound=9999"`
```
tsachiherman, line 34: Please add a constant for this one. The actual number should be much smaller than 9999. Most likely 256 is a good limit.

crypto/merklearray/merkle.go
algonautshant: @algoidan is looking into this.

crypto/onetimesig.go
```diff
@@ -157,12 +157,12 @@ type OneTimeSignatureSecretsPersistent struct {
 	// The odd `codec:` name is for backwards compatibility with previous
 	// stored keys where we failed to give any explicit `codec:` name.
 	FirstBatch uint64            `codec:"First"`
-	Batches    []ephemeralSubkey `codec:"Sub,allocbound=-"`
+	Batches    []ephemeralSubkey `codec:"Sub,allocbound=8888"`
```
algonautshant, line 160: @tsachiherman do you think these need to be bounded? Any idea what will be a good bound?

crypto/onetimesig.go
```diff
@@ -157,12 +157,12 @@ type OneTimeSignatureSecretsPersistent struct {
...
 
 	// FirstOffset denotes the first offset whose subkey appears in Offsets.
 	// These subkeys correspond to batch FirstBatch-1.
 	FirstOffset uint64            `codec:"firstoff"`
-	Offsets     []ephemeralSubkey `codec:"offkeys,allocbound=-"`
+	Offsets     []ephemeralSubkey `codec:"offkeys,allocbound=8888"`
```
algonautshant, line 165: @tsachiherman do you think these need to be bounded? Any idea what will be a good bound?

crypto/sig_abstractions.go
```diff
@@ -25,7 +25,7 @@ import (
 
 type (
 	//ByteSignature using unspecified bound.
-	//msgp:allocbound ByteSignature
+	//msgp:allocbound ByteSignature 9999
```
algonautshant, line 28: @tsachiherman do you think these need to be bounded? Any idea what will be a good bound?

crypto/onetimesig.go
tsachiherman: yes, I think that it should be bounded. regardless of what is the bounding value, I would suggest you'll start off by using a constant here ( i.e. maxBatches ), and discuss the numeric value separately.

crypto/sig_abstractions.go
algoidan: for that value, we should use the max signature size in all the Algorithmtypes we have.

crypto/merklearray/merkle.go
algoidan: it looks like that if we want a proof for all the leaves in the tree , this might contain all the hashes in the tree
since we bound the number of leaves to 2^16, we should bound this to 2^17

algonautshant: Tsachi thinks this should be in the order of the hight of the tree.

crypto/merklearray/merkle.go
```diff
@@ -178,7 +186,7 @@ func (tree *Tree) Prove(idxs []uint64) (*Proof, error) {
 		}
 	}
 
-	// Confirm that we got the same root hash
+	// Confirm that we got the same root hash // this comment in not correct ???
```
algonautshant, line 189: is this really checking the root hash?

crypto/onetimesig.go
algonautshant: I will move this bound to a separate task/PR

crypto/sig_abstractions.go
algonautshant: I will move this bound to a separate task/PR

crypto/onetimesig.go
algonautshant: I will move this bound to a separate task/PR

crypto/sig_abstractions.go
algonautshant: moved to https://github.com/algorand/go-algorand/pull/3196

crypto/onetimesig.go
algonautshant: moved to https://github.com/algorand/go-algorand/pull/3196

crypto/onetimesig.go
algonautshant: moved to https://github.com/algorand/go-algorand/pull/3196

crypto/merklearray/merkle.go
algonautshant: The final number I came out with is 2^15

PR #3172, opened 2021-11-01
Avoid printing genesis path in error message.

winder:
## Summary

Small formatting change to prevent leaking personal information into node.log.

### DIFF
--- a/node/node.go
+++ b/node/node.go
@@ -786,7 +786,7 @@ func (node *AlgorandFullNode) loadParticipationKeys() error {
 	genesisDir := filepath.Join(node.rootDir, node.genesisID)
 	files, err := ioutil.ReadDir(genesisDir)
 	if err != nil {
-		return fmt.Errorf("AlgorandFullNode.loadPartitipationKeys: could not read genesis directory: %v", err)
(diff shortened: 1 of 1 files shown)


node/node.go
```diff
@@ -786,7 +786,7 @@ func (node *AlgorandFullNode) loadParticipationKeys() error {
 	genesisDir := filepath.Join(node.rootDir, node.genesisID)
 	files, err := ioutil.ReadDir(genesisDir)
 	if err != nil {
-		return fmt.Errorf("AlgorandFullNode.loadPartitipationKeys: could not read directory %v: %v", genesisDir, err)
+		return fmt.Errorf("AlgorandFullNode.loadPartitipationKeys: could not read genesis directory: %v", err)
```
tsachiherman, line 789: I'm wondering if you had any particular thoughts around this one ?

I think that printing out the explicit genesis directory could help to resolve permission issues that related to symbolic links. On these, the actual directory might have the permission, but when accessing via the symbolic link, you might fail to do that.

That being said, if the directory is missing, then I agree that we could use a simpler message without much detail.

PR #3173, opened 2021-11-01
`roundCowBase`: use `lookup()` instead of going to ledger directly

tolikzinovyev:
## Summary

Going to ledger directly prevents caching of accounts.

I'm looking into poor indexer import performance on testnet at rounds > 17M, and I noticed that one account is looked up a lot of times in same round.

### DIFF
--- a/ledger/internal/eval.go
+++ b/ledger/internal/eval.go
@@ -162,7 +162,7 @@ func (x *roundCowBase) blockHdr(r basics.Round) (bookkeeping.BlockHeader, error)
 }
 
 func (x *roundCowBase) allocated(addr basics.Address, aidx basics.AppIndex, global bool) (bool, error) {
-	acct, _, err := x.l.LookupWithoutRewards(x.rnd, addr)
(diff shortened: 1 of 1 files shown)


tsachiherman:
We can't do that. This is the one place that actually does go from the cow to the ledger. If ( from any reason ), we don't have the account preloaded, we need to have the ability to fetch it dynamically.

One common use case is the transaction pool. When a new transaction is received by the node, it attempts to add it to the current block evaluator. In turn, the block evaluator would call the ledger to fetch all the required resources.

tolikzinovyev:
I'm not following. `lookup()` will go to the ledger if the account is not cached and will cache it afterwards.

tsachiherman:
> I'm not following. `lookup()` will go to the ledger if the account is not cached and will cache it afterwards.

sorry - my bad; I think that this change is fine.

tsachiherman (approved):

tsachiherman (approved):

PR #3174, opened 2021-11-02
Closing nil db on error

algonautshant:
When account.FillDBWithParticipationKeys fails and returns an error, e.g. if the key already exists, newPart will be invalid and should not be closed.
When Close is called, it will panic.


algonautshant:
This does not seem to be the case here. newPart is valid, and needs to be closed.

PR #3175, opened 2021-11-02
Fix panic on closing nil db 

algonautshant:
## Summary

FillDBWithParticipationKeys will return empty PersistedParticipation when lastValid is less than firstValid.
When this happens, newPart will not have partdb in newPart.Store, instead, will have an empty object.
Calling Close on the empty object will call close on nil Accessor pointer and panic.

This change avoids using the returned object with non-nil error, and properly closes the db with the valid object.
## Test Plan

Added test to verify the proper handling of the different errors returned from FillDBWithParticipationKeys and handled by GenParticipationKeysTo.


tsachiherman (changes requested):
looks great, but please fix the review dog comments..

algorandskiy (approved):
LGTM

tsachiherman (approved):

PR #3177, opened 2021-11-02
txnsync: rollback

tsachiherman:
## Summary

This PR rolls back the txsync, retaining only the config file changes


cce (approved):

cce:
One thing I noticed, it looks like the new version of network/latencyTracker.go from #3028 was removed in d2bbf648563b8fac089964f4ff534164fadae8f1 but the old versions of network/ping.go, network/ping_test.go, network/wsNetwork.go, and network/wsPeer.go with the old ping code are gone now too, because #3028 was not explicitly reverted on this branch.

PR #3148, opened 2021-10-26
Update participation key goal output

winder:
## Summary

Use the participation registry for displaying participation key information.

PR #3178, opened 2021-11-02
Add new participation key e2e test.

winder:
## Summary

Note: This is a draft because it depends on/includes changes in #3148 which have not been merged yet.

First tests using dev-mode and the new participation key rest API's to write more focused tests.

## Test Plan

These are tests.


node/node.go
```diff
@@ -1330,7 +1330,7 @@ func (node *AlgorandFullNode) VotingKeys(votingRound, keysRound basics.Round) []
 		matchingAccountsKeys[part.Address()] = true
 
 		// Make sure the key is registered.
-		err := node.accountManager.Registry().Register(part.ID(), keysRound)
+		err := node.accountManager.Registry().Register(part.ID(), votingRound)
```
winder, line 1333: Also found a bug, this was causing the wrong "EffectiveRound" variables to be set in the registry.

PR #3179, opened 2021-11-02
Integrate falcon scheme

algoidan:


crypto/sig_abstractions_test.go
```diff
@@ -33,17 +33,20 @@ func TestInvalidSinger(t *testing.T) {
...
-	sigAlgo, err = NewSigner(DilithiumType)
+	sigAlgo, err = NewSigner(FalconType)
 	a.NoError(err)
 	sigAlgo.Type = MaxAlgorithmType
 
 	dummyMsg = make([]byte, 6)
-	a.Equal(ByteSignature{}, sigAlgo.GetSigner().SignBytes(dummyMsg))
+
+	_, err = sigAlgo.GetSigner().SignBytes(dummyMsg)
+	a.Error(err)
```
algonautshant, line 49: Does it make sense to check the error here?

crypto/sig_abstractions.go
```diff
@@ -105,20 +107,20 @@ func (z *GenericVerifyingKey) IsValid() error {
...
 	switch z.Type {
-	case DilithiumType:
-		return &z.DilithiumSigner
+	case FalconType:
+		return &z.FalconSigner
 	case Ed25519Type:
 		return &z.Ed25519Singer
 	default:
-		return &invalidSinger{}
+		return NewInvalidSinger()
```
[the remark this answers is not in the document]
algoidan: we wanted to avoid returning an error while invoking the "get" function. getting the invalidSigner is very unlikely because the check is performed in the MSGPACK code.

crypto/sig_abstractions_test.go
algoidan: this might be tricky because the signer is an abstraction and the error might be specific to the scheme.

crypto/merklekeystore/keystoreBuilder.go
```diff
@@ -0,0 +1,63 @@
...
+		// we add this section to the last worker.
+		if endIdx+numberOfWorkers > numberOfKeys {
+			endIdx = numberOfKeys
+		}
+
+		wg.Add(1)
+		go func(starIdx, endIdx uint64) {
+			for k := starIdx; k < endIdx; k++ {
+				sigAlgo, err := crypto.NewSigner(sigAlgoType)
+				if err != nil {
```
algonautshant, line 53: The error handling is missing.  Need a channel to pass the error messages to the main function.

crypto/merklekeystore/keystoreBuilder.go
```diff
@@ -0,0 +1,63 @@
...
+
+// KeyStoreBuilder Responsible for generate slice of keys in a specific AlgorithmType.
+// this function trys to optimize this process by using Goroutines.
+func KeyStoreBuilder(numberOfKeys uint64, sigAlgoType crypto.AlgorithmType) ([]crypto.GenericSigningKey, error) {
+	keys := make([]crypto.GenericSigningKey, numberOfKeys)
+
+	var numberOfWorkers uint64
+	numOfCores := uint64(runtime.NumCPU() * 2)
+
+	if numberOfKeys > numOfCores {
```
algonautshant, line 33: It is not clear to me how `numberOfWorkers` variable is being used. 
I think it is mixing up to concepts: number of workers and number of keys per worker.

crypto/merklekeystore/keystoreBuilder.go
```diff
@@ -0,0 +1,63 @@
...
+		numberOfWorkers = numberOfKeys / numOfCores
+	} else {
+		numberOfWorkers = numberOfKeys
```
algonautshant, line 36: In this case, the multiple cores are not utilized. It runs all the keys sequentially on a single core.

algonautshant (changes requested):

crypto/falconWrapper.go
```diff
@@ -0,0 +1,89 @@
...
+}
+
+// Sign receives a message and generates a signature over that message.
+func (d *FalconSigner) Sign(message Hashable) (ByteSignature, error) {
+	hs := Hash(HashRep(message))
+	return d.SignBytes(hs[:])
+}
+
+// SignBytes receives bytes and signs over them.
+func (d *FalconSigner) SignBytes(data []byte) (ByteSignature, error) {
```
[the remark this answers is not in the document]
algoidan: SignBytes is part of the Signer interface. so it would not compile

crypto/falconWrapper.go
```diff
@@ -0,0 +1,89 @@
...
+}
+
+// Verify follows dilithium algorithm to verify a signature.
+func (d *FalconVerifier) Verify(message Hashable, sig ByteSignature) error {
+	hs := Hash(HashRep(message))
+	return d.VerifyBytes(hs[:], sig)
+}
+
+// VerifyBytes follows dilithium algorithm to verify a signature.
+func (d *FalconVerifier) VerifyBytes(data []byte, sig ByteSignature) error {
```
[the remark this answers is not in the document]
algoidan: VerifyBytes is part of the verifier interface. so it would not compile

crypto/falconWrapper.go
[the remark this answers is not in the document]
algonautshant: Why does it have to be part of the Signer interface? 
I am not seeing why SignBytes and VerifyBytes need to be part of the Signer interface. 
What happens if they are removed from the Signer interface?

crypto/merklekeystore/keystoreBuilder.go
```diff
@@ -0,0 +1,63 @@
...
+	}
+
+	var wg sync.WaitGroup
+	var endIdx uint64
+	for i := uint64(0); i < numberOfKeys; i = endIdx {
+		endIdx = i + numberOfWorkers
+		// in case the next iteration will pass the number of keys.
+		// we add this section to the last worker.
+		if endIdx+numberOfWorkers > numberOfKeys {
```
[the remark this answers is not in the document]
algonautshant: Okay. I am convinced this is correct. 
Please update the comment to something like this: 

> If this is the last worker, give it the rest of the keys.

crypto/merklekeystore/keystoreBuilder.go
```diff
@@ -27,37 +27,44 @@ import (
...
 				keys[k] = *sigAlgo
 			}
-			wg.Done()
-		}(i, endIdx)
+		}(i, endIdx, errors)
 	}
 	wg.Wait()
+	select {
+	case err := <-errors:
+		return []crypto.GenericSigningKey{}, err
```
algonautshant, line 66: Unfortunately, the user will have to wait the full time to get an error message in the end.
When an error is detected, the process should immediately returns for best user experience.

Makefile
```diff
@@ -290,7 +290,7 @@ gen/%/genesis.dump: gen/%/genesis.json
 	./scripts/dump_genesis.sh $< > $@
 
 gen/%/genesis.json: gen/%.json gen/generate.go buildsrc
-	$(GOPATH1)/bin/genesis -q -n $(shell basename $(shell dirname $@)) -c $< -d $(subst .json,,$<)
+	$(GOPATH1)/bin/genesis -q -n $(shell basename $(shell dirname $@))  -f 0 -l 1500 -c $< -d $(subst .json,,$<)
```
tsachiherman, line 293: I'm wondering why you chose to pass these as parameters instead of changing the parameters in the files.

cmd/goal/account.go
```diff
@@ -854,6 +855,32 @@ func changeAccountOnlineStatus(acct string, part *algodAcct.Participation, goOnl
...
+		errChan <- struct{}{}
+	}()
+
+	progressStrings := [...]string{"/", "-", "\\", "|"}
+
+	finished := false
+	i := 0
+
+	for !finished {
+		timer := time.NewTimer(time.Duration(100000000))
```
tsachiherman, line 871: That's a really big number.. `100 * time.Millisecond` is much more readable.

cmd/goal/account.go
```diff
@@ -869,11 +896,16 @@ var addParticipationKeyCmd = &cobra.Command{
...
 
-		_, _, err := client.GenParticipationKeysTo(accountAddress, roundFirstValid, roundLastValid, keyDilution, partKeyOutDir)
-		if err != nil {
-			reportErrorf(errorRequestFail, err)
+		participationGen := func() {
+			_, _, err := client.GenParticipationKeysTo(accountAddress, roundFirstValid, roundLastValid, keyDilution, partKeyOutDir)
+			if err != nil {
+				reportErrorf(errorRequestFail, err)
+			}
+			fmt.Println("Participation key generation successful")
```
tsachiherman, line 904: try to follow the goal utility conventions of defining output strings in messages.go, and printing these using reportInfof() instead of writing directly to stdout.

crypto/falconWrapper.go
```diff
@@ -0,0 +1,89 @@
...
+	FSecretKey [cfalcon.PrivateKeySize]byte
+	// FalconSeed represents the seed which is being used to generate Falcon keys
+	FalconSeed [FalconSeedSize]byte
+)
+
+// FalconSigner is the implementation of Signer for the Falcon signature scheme.
+type FalconSigner struct {
+	_struct struct{} `codec:",omitempty,omitemptyarray"`
+
+	PublicKey FPublicKey `codec:"pk"`
```
tsachiherman, line 41: add allocbound

crypto/merklekeystore/KeystoreBuilder_test.go
```diff
@@ -0,0 +1,93 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with go-algorand.  If not, see <https://www.gnu.org/licenses/>.
+
+package merklekeystore
+
+import (
+	"github.com/algorand/go-algorand/util/db"
```
tsachiherman, line 20: nit: move to the group below

crypto/merklekeystore/KeystoreBuilder_test.go
```diff
@@ -0,0 +1,93 @@
...
+	"runtime"
+	"testing"
+
+	"github.com/algorand/go-algorand/crypto"
+	"github.com/algorand/go-algorand/test/partitiontest"
+	uuid "github.com/satori/go.uuid"
```
tsachiherman, line 26: nit: mov to it;s own group

crypto/merklekeystore/keystoreBuilder.go
```diff
@@ -0,0 +1,70 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with go-algorand.  If not, see <https://www.gnu.org/licenses/>.
+
+package merklekeystore
+
+import (
+	"github.com/algorand/go-algorand/crypto"
```
tsachiherman, line 20: nit: move to it's own group

crypto/falconWrapper.go
algoidan: FPublicKey is a fixed-length slice. do we need to bound this even if it is fixed?

mrcali213 (approved):

Makefile
algoidan: we will resolve it in a different PR

PR #3181, opened 2021-11-03
Limit the longest validity period allowed for key registration to pre…

Aharonee:
…vent merklekeystore tree from being too big


crypto/merklekeystore/keystore.go
```diff
@@ -18,6 +18,8 @@ package merklekeystore
 
 import (
 	"errors"
+	"fmt"
+	"github.com/algorand/go-algorand/config"
```
tsachiherman, line 22: move to the group below.

crypto/merklekeystore/keystore.go
```diff
@@ -79,6 +81,11 @@ type (
 	}
 )
 
+// MaxValidPeriod defines the longest validity period allowed for key registration,
+// to prevent the merklekeystore tree depth being larger than 16
+var MaxValidPeriod = (1<<16)*config.Consensus[protocol.ConsensusFuture].CompactCertRounds - 1
```
tsachiherman, line 86: you want to calculate this dynamically using an init function.

crypto/merklekeystore/keystore.go
```diff
@@ -79,6 +81,11 @@ type (
...
+
+var errValidityPeriodTooLong = fmt.Errorf("the validity period for merkleKeyStore is too large: the limit is %v", MaxValidPeriod)
```
tsachiherman, line 88: `%v` -> `%d`

crypto/merklekeystore/keystore_test.go
```diff
@@ -21,6 +21,7 @@ import (
 	"crypto/rand"
 	"database/sql"
 	"fmt"
+	"github.com/algorand/go-algorand/config"
```
tsachiherman, line 24: move to the algorand group below.

crypto/merklekeystore/keystore.go
Aharonee: What is the advantage of initializing it in init()? Is it just a convention?

tsachiherman: The calculation you've used above is wrong - it assumes that the value in `protocol.ConsensusFuture` is the highest. This might not hold true in the future. In `config/consensus.go`, we have a `checkSetAllocBounds` which is called via an init function, which calculate the various max parameters by checking *all* the consensus params.

algonautshant: @tsachiherman is this correct?

init() function is called when the package is initialized. Example of it is the one in config/consensus.go
As I understand, Tsachi is suggesting to have an init function, which loops through all the `ConsensusProtocols`, and picks the largest value of `CompactCertRounds`.
e.g.
```
func init() {
        compactCertRounds = 0
        for _, p := range config.Consensus {
                if p. CompactCertRounds > compactCertRounds
                       compactCertRounds = p. CompactCertRounds
                }
        }
        MaxValidPeriod = (1<<16)*compactCertRounds - 1
}
```

tsachiherman: Given that this is a consensus protocol max value, why don't we "port" this one into `consensus.go` ?
We already have all the infrastructure there.

Aharonee: I'm not sure why we would need the max of all `CompactCertRounds` though, this is not the behavior I would expect when generating the StateProof keys;
The number of keys generated (and thus the depth of the merkle tree that will be built) is derived from `validityPeriod/CompactCertRounds`, so in the case where we lower `CompactCertRounds` in future consensus version the number of keys will increase accordingly - and will be bigger than the 2^16 limit we require. 
This limit defines the longest validity period that can be requested, so for a smaller `CompactCertRounds` I would require a smaller value.

tsachiherman: I could be wrong, but I believe that we use it during the decoding of the msgp. We want to create a backward-compatible limit. This number doesn't have to be "accurate", but it should provide a reasonable "protection" against potential malicious msgp-content.

crypto/merklekeystore/keystore.go
algonautshant, line 22: add an empty line between "fmt" and "github.com/algorand/go-algorand/config"

algonautshant (changes requested):
Can you please merge/rebase from feature/dilithium-scheme-integration?

crypto/merklekeystore/keystore_test.go
```diff
@@ -26,6 +26,7 @@ import (
 	uuid "github.com/satori/go.uuid"
 	"github.com/stretchr/testify/require"
 
+	"github.com/algorand/go-algorand/config"
```
algonautshant, line 29: Line 61: use `defer` for `signer.keyStore.store.Close()`

Line 117: signer is missing `signer.keyStore.store.Close()`.

Line 156: signer is missing `signer.keyStore.store.Close()`.

algonautshant (changes requested):
Looks great. 
Just some minor changes are needed.

crypto/merklekeystore/keystore.go
```diff
@@ -107,12 +106,24 @@ func (k *keysArray) Marshal(pos uint64) ([]byte, error) {
 // New Generates a merklekeystore.Signer
 // The function allow creation of empty signers, i.e signers without any key to sign with.
 // keys can be created between [A,Z], if A == 0, keys created will be in the range (0,Z]
-func New(firstValid, lastValid, interval uint64, sigAlgoType crypto.AlgorithmType, store db.Accessor) (*Signer, error) {
+func New(firstValid uint64, lastValid uint64, sigAlgoType crypto.AlgorithmType, store db.Accessor) (*Signer, error) {
```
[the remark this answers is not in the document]
Aharonee: Yes it is used for testing purposes. It seemed to me like bad practice to edit a global variable for a specific test. What if there are multiple tests running in parallel? Wouldn't we prefer having the config parameter immutable?

crypto/merklekeystore/keystore.go
```diff
@@ -80,9 +82,6 @@ type (
 	}
 )
 
-var errStartBiggerThanEndRound = errors.New("cannot create merkleKeyStore because end round is smaller then start round")
```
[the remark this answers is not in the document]
Aharonee: It is used only once so I don't really see the value of binding a variable to it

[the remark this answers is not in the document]
Aharonee: Yeah you're right it was unecessary

config/consensus_test.go
```diff
@@ -58,3 +58,14 @@ func TestConsensusUpgradeWindow(t *testing.T) {
 		}
 	}
 }
+
+func TestConsensusCompactCertParams(t *testing.T) {
+	partitiontest.PartitionTest(t)
+
+	for _, params := range Consensus {
+		if params.CompactCertRounds != 0 {
+			require.Equal(t, uint64(1<<16), (params.MaxKeyregValidPeriod+1)/params.CompactCertRounds,
```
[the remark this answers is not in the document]
Aharonee: I can use merklearray.MaxTreeDepth, but what about the consensus parameters (MaxKeyregValidPeriod for example), should it use that as well? If not there might be some inconsistency in the future.

crypto/merklekeystore/keystore_test.go
```diff
@@ -566,7 +566,7 @@ func generateTestSignerAux(a *require.Assertions) (uint64, uint64, *Signer) {
 
 func generateTestSigner(t crypto.AlgorithmType, firstValid uint64, lastValid uint64, interval uint64, a *require.Assertions) *Signer {
 	store := initTestDB(a)
-	signer, err := new(firstValid, lastValid, interval, t, *store)
+	signer, err := New(firstValid, lastValid, t, *store)
```
algonautshant, line 569: `interval` should be dropped from `generateTestSigner` parameter list.

crypto/merklekeystore/keystore.go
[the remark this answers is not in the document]
algonautshant: check out the changes here. 
It is fixing the fallout from this:
https://github.com/Aharonee/go-algorand/pull/1

crypto/merklekeystore/keystore_test.go
algonautshant: fixed in https://github.com/Aharonee/go-algorand/pull/1

algonautshant (changes requested):
Merge https://github.com/Aharonee/go-algorand/pull/1 if you agree with it. 
Then should be good to go.

algonautshant:
- Some casting to uint64 (e.g. for firstValid/lastValid) are not needed.
- I think firstValid/lastValid parameters in New and other interfaces should be basics.Round. Casting form basics.Round to uint64 should not be done.

algonautshant (changes requested):

data/account/participation.go
```diff
@@ -217,6 +217,23 @@ func (part PersistedParticipation) PersistNewParent() error {
 	})
 }
 
+func generateKeystore(firstValid, lastValid uint64, store db.Accessor, protoParams config.ConsensusParams) (*merklekeystore.Signer, error) {
```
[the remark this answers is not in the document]
Aharonee: There seems to be a cyclical import preventing us from using `basics.Round` in `crypto`:
basics imports crypto which in turn can't import basics.

data/transactions/logic/evalAppTxn_test.go
```diff
@@ -609,10 +609,29 @@ func TestKeyReg(t *testing.T) {
 `
 		ep, ledger := makeSampleEnv()
 		ep.Proto.EnableStateProofKeyregCheck = true
+		ep.Proto.MaxKeyregValidPeriod = (1<<16)*128 - 1 // 2^16 StateProof keys times CompactCertRounds (interval)
```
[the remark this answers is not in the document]
Aharonee: Yes that would definitely be better, we need a single source of truth for all these place to inherit from

data/account/participation.go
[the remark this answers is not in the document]
algonautshant: You are correct :-) Luckily the round is the only dependency.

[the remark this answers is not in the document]
algoidan: I would like to suggest something else (not by much)
why don't we pass the  MaxKeyregValidPeriod and the CompactCertRounds as params to the merkle key store? this way the merkle key store will not access consensus params and it won't be unbounded

data/transactions/transaction.go
```diff
@@ -565,6 +566,10 @@ func (tx Transaction) stateProofPKWellFormed(proto config.ConsensusParams) error
 		return nil
 	}
 
+	if proto.MaxKeyregValidPeriod != 0 && uint64(tx.VoteLast-tx.VoteFirst) > proto.MaxKeyregValidPeriod {
```
algoidan, line 569: could you please use SubSaturate instead of "-"

crypto/merklekeystore/keystore.go
algonautshant: This variable is moved to be a local non-exported variable inside a function.

config/consensus_test.go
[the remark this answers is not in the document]
algonautshant: Let's postpone this parameter issue to a later PR.

crypto/compactcert/builder_test.go
```diff
@@ -117,11 +121,11 @@ func TestBuildVerify(t *testing.T) {
 		ProvenWeight:      uint64(totalWeight / 2),
 		SigRound:          currentRound,
 		SecKQ:             128,
```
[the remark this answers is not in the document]
algonautshant: Let's postpone this #3257

data/transactions/logic/evalAppTxn_test.go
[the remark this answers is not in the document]
algonautshant: #3257

config/consensus.go
```diff
@@ -1062,6 +1068,9 @@ func initConsensusProtocols() {
 	// stat proof key registration
 	vFuture.EnableStateProofKeyregCheck = true
 
+	// Maximum validity period for key registration, to prevent generating too many StateProof keys
+	vFuture.MaxKeyregValidPeriod = 128*(1<<16) - 1
```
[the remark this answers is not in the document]
algonautshant: #3257

algonautshant (approved):

PR #3184, opened 2021-11-03
FillDBWithParticipationKeys return object on error

algonautshant:
## Summary
enhance comment for Signer struct
fix ordering of imports in keystore_test
FillDBWithParticipationKeys return on error changes


data/account/participation.go
```diff
@@ -217,12 +217,9 @@ func FillDBWithParticipationKeys(store db.Accessor, address basics.Address, firs
 	}
 	// Persist the Participation into the database
 	err = part.Persist()
-	if err != nil {
-		return PersistedParticipation{}, err
+	if err == nil {
+		err = stateProofSecrets.Persist() // must be called after part.Persist()
```
algoidan, line 221: If Persist() function is not written well, you will return a PersistedParticipation struct which might be filled with wrong data. I think, In order to protect ourselves, we should return empty struct.

for the err value- this change is great.

algonautshant: The expectations should be that if error is returned, there are no guarantees on anything else returned (this was something I fixed on master).

When you say "protect ourselves", what are the risks? Maybe, we should make sure Persist is safe when it returns an error.

I have no problem with returning PersistedParticipation{} in case of an error. However, if there are risks with returning part, we need to understand that,

data/account/participation.go
```diff
@@ -217,12 +217,9 @@ func FillDBWithParticipationKeys(store db.Accessor, address basics.Address, firs
...
 	}
-
-	err = stateProofSecrets.Persist() // must be called after part.Persist() !
-
 	return part, err
```
algonautshant, line 223: notice here, without my change, in the case of stateProofSecrets.Persist returning an error, it is still returning part with err not nil. 
If we want to return PersistedParticipation{} in case of an error, we net to check err in two places. I think this will unnecessarily complicate the code.

algoidan (approved):
looks good!

tsachiherman:
small request - when you're creating a pr, could you please try and give the PR a more informal title ?

PR #3185, opened 2021-11-04
CI: set -coverpkg when calling go test in nightly builds

cce:
## Summary

By default, Go test with -coverprofile enabled only measures the coverage of files in the same package as the tests. Providing -coverpkg provides go test with additional names of packages for which you'd like it to measure coverage. Today we don't see the true coverage of source files called from tests in other packages.

## Test Plan

This should not impact test correctness, but slows down tests due to the extra line counting required to measure coverage across more packages.

Due to the performance impact I updated the PR to only do "full coverage" reporting for nightly builds and am using the codecov flags feature to have these nightly builds be separately categorized by Codecov.

https://docs.codecov.com/docs/flags#hide-builds-eg-nightly-builds


cce:
Well, the tests pass now, and coverage reports more accurately reflect what is being tested, but collecting atomic coverage reports made the {amd64,arm64,mac_amd64}_test jobs get ~50% slower, from ~22 minutes to ~34 minutes. So without more parallelism that would make the test jobs the new slowest jobs, previously they were taking roughly about the same amount of time as the expect tests (~25 minutes).

cce:
Can revisit later

Makefile
```diff
@@ -111,7 +112,7 @@ check_shell:
 sanity: vet fix lint fmt
 
 cover:
```
algorandskiy, line 114: a feature request (not sure how hard it is)
``` sh
make cover [package=package_name]
```
Behavior: if package is set (for example, `package=ledger`) then COVERPKG_PACKAGES would be ledger + all subpackages.

Benefit:
1. partial run for only interested packages
2. faster run - can be done on dev machine

scripts/travis/upload_coverage.sh
```diff
@@ -12,4 +12,4 @@ if [[ "${UPSTREAM_VERSION}" != "${LOCAL_VERSION}" ]]; then
   echo "WARN: version ${UPSTREAM_VERSION} of the codecov upload script is available."
 fi
 
-/usr/bin/env bash scripts/travis/codecov
+/usr/bin/env bash scripts/travis/codecov "$@"
```
winder, line 15: I thought this file was being replaced with a github action, is that still in progress @algojack

algojack: different codecov was added in https://github.com/algorand/go-algorand/pull/3997/files but the old one wasn't removed yet. Are we ok to remove the old one or is there something missing still?

cce: oh this PR just modifies the .codecov.yml file to add a "nightly" tag to the coverage reports generated nightly, because nightly tests will get fuller coverage than regular PRs.

algojack: i guess @cce pointed out that while it was added, the upload action was not added in circleci. We still need that.

cce: oh I found it — https://github.com/algorand/go-algorand/issues/3962 OK I will take a look at the indexer change and put it in here

cce:
Switching to the orb doesn't work, because the codecov uploader does not support Linux with ARM
image
https://app.circleci.com/pipelines/github/algorand/go-algorand/7388/workflows/e8b6d40c-cae7-4a72-9547-8e2ae8cd5c01/jobs/131023

scripts/travis/upload_coverage.sh
cce: Switching to the orb from the bash uploader didn't work, because the codecov uploader does not support Linux with ARM. So I think we have to stick with the codecov script we have checked in until they fix that.
image
https://app.circleci.com/pipelines/github/algorand/go-algorand/7388/workflows/e8b6d40c-cae7-4a72-9547-8e2ae8cd5c01/jobs/131023

Makefile
cce: OK — I added support to do `make cover PACKAGE=ledger`

util/metrics/serviceCommon.go
```diff
@@ -30,6 +30,7 @@ type ServiceConfig struct {
 	NodeExporterListenAddress string
 	Labels                    map[string]string
 	NodeExporterPath          string
+	registry                  *Registry // if nil, will use DefaultRegistry(). Used for testing
```
algorandskiy, line 33: do we need this at all?

algorandskiy (commented):
LGTM, please fix reviewdog

util/metrics/serviceCommon.go
cce: When you build with -coverpkg it makes one big test binary across all packages (instead of one test binary per package), and it was causing tests to fail because different packages were touching the same global default registry... that's why this PR has metrics changes to allow tests to specify a registry.

algorandskiy: I understand why metrics were changed. but why this service change? can't it still use default registry? Probably I missed where it gets initialized?

cce: the tests in util/metrics start up a MetricService and hit the HTTP service after bumping some counters, asserting the prometheus output, this makes it possible for them to not use the default registry

algorandskiy: ah, in TestMetricCounter I see

algorandskiy (approved):

jannotti (commented):
If I'm honest, I have no idea what any of the metrics changes do.

I also didn't follow what actually _happens_ after this change. Nightly coverage is calculated, but where does it show up?  It seems like CI coverage is only useful if it's attached to the code that caused the changes.  Where would I go to see the nightly coverage, and who will go look at it repeatedly to see how it changes?

cce:
> If I'm honest, I have no idea what any of the metrics changes do.

I will update the PR description — they were touching global counter state and now that multiple packages are being compiled together, it was causing test assertions to fail that assumed that counter state was empty. 
 
> I also didn't follow what actually _happens_ after this change. Nightly coverage is calculated, but where does it show up? It seems like CI coverage is only useful if it's attached to the code that caused the changes. Where would I go to see the nightly coverage, and who will go look at it repeatedly to see how it changes?

After this change you would go to the codecov website to see the nightly coverage — I need to get this merged so that we can start to see it show up as a base for comparison — and then could trigger coverage diffs on PRs by triggering a special Github Actions workflow to run it against your PR.

jannotti:
> > If I'm honest, I have no idea what any of the metrics changes do.
> 
> I will update the PR description — they were touching global counter state and now that multiple packages are being compiled together, it was causing test assertions to fail that assumed that counter state was empty.

You mentioned that in stand-up and I didn't have a chance to understand why, seemed like compiling together is not a problem. But reading comments above, it seems the issue is `init` blocks?  So there's actually code being run in those other packages, outside of tests, that causes the problem?
 
> > I also didn't follow what actually _happens_ after this change. Nightly coverage is calculated, but where does it show up? It seems like CI coverage is only useful if it's attached to the code that caused the changes. Where would I go to see the nightly coverage, and who will go look at it repeatedly to see how it changes?
> 
> After this change you would go to the codecov website to see the nightly coverage — I need to get this merged so that we can start to see it show up as a base for comparison — and then could trigger coverage diffs on PRs by triggering a special Github Actions workflow to run it against your PR.

okay!

algorandskiy (approved):

PR #3190, opened 2021-11-05
Report assembly error strings for compile API endpoint 

### LINKED ISSUE #3082: Assembly Error Message
## Problem

When using the API to compile a teal contract, if there is a syntax error or something else the error returned is a generic "%d error(s)" with no hints on what it might be.

The assembler does have some list of the errors but summarizes them with a count here:
https://github.com/algorand/go-algorand/blob/master/data/transactions/logic/assembler.go#L1676-L1682

Is this done for a specific reason or could we return some string with all the error messages or at least the first N?

algochoi:
## Summary

This PR adds the assembly error strings so developers can see which errors were present in the program when calling the API. Prior to this change, the API only reported the number of errors in the program.

Closes #3082 .

## Test Plan

Tested on sandbox.
Example program:
``` 
#pragma version 7
int v
```

Example response:
```
algosdk.error.AlgodHTTPError: 1: unsupported version: 7
2: strconv.ParseUint: parsing "v": invalid syntax
```

### DIFF
--- a/data/transactions/logic/assembler.go
+++ b/data/transactions/logic/assembler.go
@@ -2130,9 +2130,6 @@ func (ops *OpStream) ReportProblems(fname string, writer io.Writer) {
 			fmt.Fprintf(writer, "%s: %s\n", fname, w)
 		}
 	}
-	if fname == "" {
-		fmt.Fprintf(writer, "%d errors", len(ops.Errors))
-	}
 }
 
 // AssembleString takes an entire program in a string and assembles it to bytecode using AssemblerDefaultVersion
--- a/data/transactions/logic/assembler.go
+++ b/data/transactions/logic/assembler.go
@@ -27,7 +27,6 @@ import (
 	"errors"
 	"fmt"
 	"io"
-	"os"
 	"sort"
 	"strconv"
 	"strings"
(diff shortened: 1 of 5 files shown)


data/transactions/logic/assembler.go
```diff
@@ -2125,6 +2125,25 @@ func (ops *OpStream) ReportProblems(fname string) {
 	}
 }
 
+// ReportProblemsString issues accumulated warnings and outputs an error string.
```
jannotti, line 2128: Is this repeating the same work as in `goal`? Can you call this there?

jannotti: fmt.Fprintf takes an io.Writer, so you should be able to refactor ReportProblems to take an io.Writer too, so that goal calls it with stderr and rest calls with a string.Builder.

jannotti: Take a "prefix" argument too, so you can get the fname in the goal output.

algochoi: Good point -- Should `fname` be left blank for the API call?

jannotti: Yeah, I think so.

data/transactions/logic/assembler.go
```diff
@@ -2109,19 +2108,30 @@ func (ops *OpStream) warnf(format string, a ...interface{}) error {
...
 		}
-		fmt.Fprintf(os.Stderr, "%s: %s\n", fname, w)
+		if fname == "" {
+			fmt.Fprintf(writer, "%s\n", w)
+		} else {
+			fmt.Fprintf(writer, "%s: %s\n", fname, w)
+		}
+	}
+	if fname == "" {
+		fmt.Fprintf(writer, "%d errors", len(ops.Errors))
```
jannotti, line 2134: I'd be ok with us not ever reporting the error count from in here.  It seems strange that we decide whether to do it based on if the fname is blank.  It's ok with me if the REST API just never says it.

algochoi: I agree, it looks funny in the code -- will delete!

jannotti (approved):

PR #3191, opened 2021-11-05
PKI PR Feedback.

winder:
## Summary

* misc nits, clarifications, typos, organization
* Close() now blocks (with timeout) until write queue completes.
* Flush() now has a timeout.
* REST API returns 404 for no-op deletes and missing keys


agreement/keyManager_test.go
```diff
@@ -53,11 +56,18 @@ func (m *recordingKeyManager) DeleteOldKeys(r basics.Round) {
...
 	if _, ok := m.recording[acct]; !ok {
 		m.recording[acct] = make(map[account.ParticipationAction]basics.Round)
 	}
 	m.recording[acct][action] = round
 }
+
+func (m *recordingKeyManager) ValidateVoteRound(t *testing.T, address basics.Address, round basics.Round) {
+	m.mutex.Lock()
+	assert.Equal(t, round, m.recording[address][account.Vote])
+	assert.Equal(t, round, m.recording[address][account.BlockProposal])
```
tsachiherman, line 71: fyi : be aware that usually using `require` is better than `assert` : `assert` would trigger a failure once the test is over, while `require` would fail right away. One important place where `require` is better is that we can test expressions that would cause the test to fail due to the wrong reason later on. ( i.e. testing for nil value in return )

data/account/participationRegistry.go
```diff
@@ -176,7 +178,7 @@ type ParticipationRegistry interface {
 	Record(account basics.Address, round basics.Round, participationType ParticipationAction) error
 
 	// Flush ensures that all changes have been written to the underlying data store.
-	Flush() error
+	Flush(duration time.Duration) error
```
tsachiherman, line 181: duration -> timeout

data/account/participationRegistry.go
```diff
@@ -312,11 +317,14 @@ type participationDB struct {
...
-	writeQueue chan partDBWriteRecord
-	asyncErr   error
+	writeQueue     chan partDBWriteRecord
+	asyncErr       error
+	writeQueueDone chan struct{}
 
 	flushDone      *sync.Cond
 	flushesPending int
+
+	Timeout time.Duration
```
tsachiherman, line 327: `Timeout` -> `flushTimeout` ( it doesn't look like there is a reason to make it public, right ? )

data/account/participationRegistry.go
```diff
@@ -366,11 +374,12 @@ func (db *participationDB) writeThread() {
 	for {
 		var wr partDBWriteRecord
 		var chanOk bool
-		// A call to .Flush() waits until the flash command passes the command queue, and also that the command queue is empty.
+		// A call to .Flush() waits until the flush command passes the command queue, and also that the command queue is empty.
 		if needsFlush {
 			select {
 			case wr, chanOk = <-db.writeQueue:
 				if !chanOk {
+					close(db.writeQueueDone)
```
tsachiherman, line 382: it's correct to place it here, but would be cleaner to use a
```golang
defer close(db.writeQueueDone)
```
at the beginning of the method; that way, you make sure that whenever the method exists the channel is being closed.

data/account/participationRegistry.go
```diff
@@ -917,31 +932,52 @@ func (db *participationDB) Record(account basics.Address, round basics.Round, pa
...
+	go func() {
+		db.mutex.Lock()
+		defer db.mutex.Unlock()
 
-	db.flushesPending++
-	db.writeQueue <- partDBWriteRecord{
-		flush: true,
-	}
+		db.flushesPending++
+		db.writeQueue <- partDBWriteRecord{
```
tsachiherman, line 943: the writing of `db.writeQueue <- partDBWriteRecord` which is done under mutex lock, could deadlock with the flush thread trying to take the lock (`db.mutex.Lock()`). In fact, if you'll try to Flush very quickly many time, it's very likely to happen.

data/account/participationRegistry.go
winder: It's not used, but I was thinking it might make sense to have it be configurable. I don't mind making it private.

data/account/participationRegistry.go
winder: Great catch, I think there were a few opportunities for this to happen with other functions as well.

I reproduced one of the deadlocks and added a separate mutex for the specific variable used by the flush thread.

tsachiherman:
I'd like to suggest a slightly different approach to the `participationDB.Flush` implementation that, I believe, would make it much simpler and more robust; the pseudocode goes like that:
```golfing
func (db *participationDB) Flush(timeout time.Duration) error {
        resultChannel := make(chan error, 1)
        timeoutCh := time.After(timeout)
	select {
        case db.writeQueue <- partDBWriteRecord{
			flush: true,
                        writeError: resultChannel,
		}
	case <-db.writeQueue:
		return err
	case <-timeoutCh:
		return fmt.Errorf("timeout while flushing changes, check results manually")
	}

	select {
        case err := <- resultChannel:
               // todo : log error here if we want to.
               return err
	case <-db.writeQueue:
		return err
	case <-timeoutCh:
		return fmt.Errorf("timeout while flushing changes, check results manually")
	}
}
```

This model doesn't really require any `Cond` variable, and given that the output channel is allocated per Flush call, there is no risk of the channel being full.

node/node.go
```diff
@@ -900,7 +903,10 @@ func (node *AlgorandFullNode) InstallParticipationKey(partKeyBinary *[]byte) (ac
 	// Tell the AccountManager about the Participation (dupes don't matter) so we ignore the return value
 	_ = node.accountManager.AddParticipation(partkey)
 
-	err = node.accountManager.Registry().Flush()
+	// PKI TODO: pick a better timeout, this is just something short. This could also be removed if we change
+	// POST /v2/participation and DELETE /v2/participation to return "202 OK Accepted" instead of waiting and getting
+	// the error message.
+	err = node.accountManager.Registry().Flush(500 * time.Millisecond)
```
tsachiherman, line 909: it would be nice to have the `500 * time.Millisecond` as a constant, but we don't need this to be done in this PR.

tsachiherman (approved):

PR #3195, opened 2021-11-08
ledger refactoring: Initial support for EnableAccountDataResourceSeparation

tsachiherman:
## Summary

In this PR, I've added a consensus parameter `EnableAccountDataResourceSeparation` and changed the catchpoint tracker so that it would not produce any catchpoints or catchpoint labels before we hit a round with `EnableAccountDataResourceSeparation`. That would help us making sure that we're all synchronized when producing catchpoints.

## Test Plan

Use existing unit test.


algorandskiy (dismissed):
The change is reasonable but please rename `ExtendApplicationStorage` to `ReenableCatchpointsStorageMigration` or similar.

tsachiherman:
> The change is reasonable but please rename ExtendApplicationStorage to ReenableCatchpointsStorageMigration or similar.

While it's true that right now it's only controlling the catchpoints, I think that this consensus parameter would expand beyond that. Furthermore, we're going to have few more of these, which is why I'd like to be cautious around picking names that could be misleading.

How do you feel about `EnableAccountDataResourceSeparation` or `EnableExplicitAccountResourceTracking` ?

PR #3196, opened 2021-11-09
Adding missing msgp bounds

algonautshant:
## Summary

The following slices are not bounded for msgp:
- OneTimeSignatureSecretsPersistent Batches, Offsets
- ByteSignature

OneTimeSignatureSecretsPersistent Batches, Offsets 
These are not bounded because the bound is keyDilution, which is a settable parameter. Moreover, the decoding of these structures does not happen for a message coming from an untrusted remote source. 

ByteSignature
The bound set is the larger of the two signature schemes used: Falcon signature size. 

## Test Plan
The existing tests are sufficient to cover these changes (e.g. TestWorkerPartialSigs).


tsachiherman (approved):
looks good to me.

PR #3199, opened 2021-11-09
Change test output format, turn up timeout from 30m to 60m for dilithium branch

cce:
## Summary

The dilithium branch has added tests that sometimes hits the no_output_default 30m timeout for test jobs. This lets those jobs complete so we can better understand their performance impacts. Otherwise we don't get much information to work with.

It also adds a change to gotestsum to print the name of the test before running it, so if your test does not finish, you will see its name rather than wait for it to complete (and be timed out before you see it).

## Test Plan

No change to tests, just more patience waiting for them to finish.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -284,7 +284,7 @@ commands:
         type: string
       no_output_timeout:
         type: string
-        default: 30m
+        default: 60m
       short_test_flag:
         type: string
         default: ""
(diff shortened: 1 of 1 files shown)


winder (approved):

algonautshant (approved):
This worked with my job, which was failing consistently many many times.

PR #3200, opened 2021-11-09
ledger refactoring: perform database schema migration

tsachiherman:
## Summary

This PR adds the migration to the new schema designed at https://docs.google.com/document/d/1E9sQvWfu9uOYvai4VjWIyUDY9Mo8bhnBX92qXh1y54o/edit#, 
by adding a new database upgrade. 
The encoded AccountData is being re-encoded into the new schema, and the old data is dropped.

It also include the long-waiting PR : https://github.com/algorand/go-algorand/pull/1974

The migration implemented in this PR is currently hard-gated. We should remove that gating only once the rest of the logic would be capable of working with the new schema.

## Test Plan

Unit tests updated/added.


ledger/accountdb.go
```diff
@@ -633,6 +644,406 @@ func accountsAddNormalizedBalance(tx *sql.Tx, proto config.ConsensusParams) erro
...
+			return err
+		}
+	}
+	return nil
+}
+
+type baseOnlineAccountData struct {
+	_struct struct{} `codec:",omitempty,omitemptyarray"`
+
+	VoteID          crypto.OneTimeSignatureVerifier `codec:"A"`
```
algorandskiy, line 670: I understand it does not matter but... maybe "v", "s", "f", "l", "d" for a little bit more semantics?

tsachiherman: this is an embedded struct, right ? we need to make sure that it's not overlapping with the codec mapping.
I was thinking that just picking the alphabet would make it easier to avoid "searching for a gap" in the future.

jannotti:
Will we not get an error at codegen time if we mess up and have an embedded
struct with field that has the same codec name as a field in its container?

tsachiherman:
I would hope so - but haven't tested that myself.

ledger/accountdb.go
```diff
@@ -633,6 +644,406 @@ func accountsAddNormalizedBalance(tx *sql.Tx, proto config.ConsensusParams) erro
...
+	GlobalState                   basics.TealKeyValue `codec:"s"`
+	LocalStateSchemaNumUint       uint64              `codec:"t"`
+	LocalStateSchemaNumByteSlice  uint64              `codec:"u"`
+	GlobalStateSchemaNumUint      uint64              `codec:"v"`
+	GlobalStateSchemaNumByteSlice uint64              `codec:"w"`
+	ExtraProgramPages             uint32              `codec:"x"`
+
+	// ResourceFlags helps to identify which portions of this structure should be used; in particular, it
+	// helps to provide a marker - i.e. whether the account was, for instance, opted-in for the asset compared
+	// to just being the owner of the asset. A comparison against the emptry structure doesn't work here -
```
algorandskiy, line 775: emptry -> empty.

Btw, basically only AppLocalState is a problem when local schema is empty

ledger/accountdb.go
```diff
@@ -633,6 +644,406 @@ func accountsAddNormalizedBalance(tx *sql.Tx, proto config.ConsensusParams) erro
...
+			NumByteSlice: ba.TotalAppSchemaNumByteSlice,
+		},
+		TotalExtraAppPages: ba.TotalExtraAppPages,
+	}
+}
+
+type resourceFlags uint8
+
+const (
+	resourceFlagsHolding    = 0 //nolint:deadcode
```
algorandskiy, line 733: is it some kind of optimization like most of the entries would have holdings?

algorandskiy (commented):
Looks good, couple comments

ledger/accountdb.go
```diff
@@ -633,6 +644,406 @@ func accountsAddNormalizedBalance(tx *sql.Tx, proto config.ConsensusParams) erro
...
+	_struct struct{} `codec:",omitempty,omitemptyarray"`
+
+	Status                     basics.Status     `codec:"a"`
+	MicroAlgos                 basics.MicroAlgos `codec:"b"`
+	RewardsBase                uint64            `codec:"c"`
+	RewardedMicroAlgos         basics.MicroAlgos `codec:"d"`
+	AuthAddr                   basics.Address    `codec:"e"`
+	TotalAppSchemaNumUint      uint64            `codec:"f"`
+	TotalAppSchemaNumByteSlice uint64            `codec:"g"`
+	TotalExtraAppPages         uint32            `codec:"h"`
```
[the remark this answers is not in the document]
tsachiherman: correct.

ledger/accountdb.go
tsachiherman: exactly. Most of the entries would be for "holding". very few will be for "owning".

ledger/accountdb.go
tsachiherman: it's also an issue for assets. We can create an empty asset where all the params are at their default value. At that point, we won't be able to tell if this account is the owner or not.

algorandskiy:
LGTM, @cce please take a look

ledger/accountdb.go
cce: yeah, very similar to protobuf, which just uses numbers

ledger/accountdb.go
cce: Maybe simpler (and more extensible if we want to cut the resources storage differently some day) to have four flags? 0x1 AssetHoldingExists, 0x2 AssetParamsExists, 0x4 AppParamsExist, 0x8 AppLocalStateExists? It seems a little bug-prone to have implicit meaning to empty/zero values and not distinguish between different app vs asset use cases with names like "holding" and "ownership" that apply to both. I guess the argument is you can combine the with rtype to deduce the real flag, and it saves a few bytes per record to default to holding vs owning.

cce: I guess my main discomfort is, the zero value of this struct implies the existence of a zero-valued basics.AssetHolding{} or zero-valued basics.AppLocalState (for UpdateRound zero). Maybe that's OK as long as handled safety and hidden down here in the DB.

tsachiherman: Yes, my intent is to wrap it nicely so that it won't be externally exposed. But before doing that, I would need to know what the "exported" data structures are going to be ( i.e. it's not going to be AccountData ). 
I will defer this to a future PR, as this unblocks me from moving forward.

ledger/accountdb.go
tsachiherman: btw - the reason I use the `baseOnlineAccountData` here is because it's going to be moving to a different table later on. I figured that it would be easier to have it already defined separately.

ledger/accountdb.go
```diff
@@ -633,6 +644,414 @@ func accountsAddNormalizedBalance(tx *sql.Tx, proto config.ConsensusParams) erro
...
+}
+
+func (rd *resourcesData) SetAssetParams(ap basics.AssetParams, haveHoldings bool) {
+	rd.Total = ap.Total
+	rd.Decimals = ap.Decimals
+	rd.DefaultFrozen = ap.DefaultFrozen
+	rd.UnitName = ap.UnitName
+	rd.AssetName = ap.AssetName
+	rd.URL = ap.URL
+	copy(rd.MetadataHash[:], ap.MetadataHash[:])
```
cce, line 796: why copy this one but not the byte arrays inside addresses like Manager, Reserve, etc, or the slices like ApprovalProgram?

tsachiherman: I wasn't sure if it would copy byte arrays - but apparently the answer is "yes", so I'll change that over.

PR #3204, opened 2021-11-10
goal abi method outfile checking

barnjamin:
Check if the `outFilename` argument is passed and write to the path specified as other commands

### DIFF
--- a/cmd/goal/application.go
+++ b/cmd/goal/application.go
@@ -1123,6 +1123,21 @@ var methodAppCmd = &cobra.Command{
 			reportErrorf(errorSigningTX, err)
 		}
 
+		if outFilename != "" {
+			if dumpForDryrun {
+				err = writeDryrunReqToFile(client, tx, outFilename)
+			} else {
+				// Write transaction to file
+				err = writeTxnToFile(client, sign, dataDir, walletName, tx, outFilename)
+			}
+
+			if err != nil {
+				reportErrorf(err.Error())
+			}
+
+			return
+		}
+
 		txid, err := client.BroadcastTransaction(signedTxn)
 		if err != nil {
 			reportErrorf(errorBroadcastingTX, err)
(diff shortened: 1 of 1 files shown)


ahangsu (dismissed):
LGTM

cmd/goal/application.go
```diff
@@ -176,12 +176,10 @@ func init() {
 
 	infoAppCmd.MarkFlagRequired("app-id")
 
-	methodAppCmd.MarkFlagRequired("method")      // nolint:errcheck // follow previous required flag format
-	methodAppCmd.MarkFlagRequired("app-id")      // nolint:errcheck
-	methodAppCmd.MarkFlagRequired("from")        // nolint:errcheck
-	methodAppCmd.Flags().MarkHidden("app-arg")   // nolint:errcheck
-	methodAppCmd.Flags().MarkHidden("app-input") // nolint:errcheck
-	methodAppCmd.Flags().MarkHidden("i")         // nolint:errcheck
```
jannotti, line 184: What are these and why did they go away?

ahangsu: Suppose we want to support foreignAssets/foreignApp/etc., we can pass that in cmd line and I suppose they can also be passed from a file, right?
This was left from previous #3088 about if keeping argument for foreignApp etc., and I think the option should be available and kept

jasonpaulos: The only things that went away are:
```go
methodAppCmd.Flags().MarkHidden("app-input") // nolint:errcheck
methodAppCmd.Flags().MarkHidden("i")         // nolint:errcheck
```
They are both flags which specify a JSON input file which can contain foreign references (and app args, but we already handle these properly). Since we want to allow specifying additional foreign references, I think it makes sense to keep this flag unhidden.

cmd/goal/application.go
```diff
@@ -1123,6 +1121,21 @@ var methodAppCmd = &cobra.Command{
 			reportErrorf(errorSigningTX, err)
 		}
 
+		if outFilename != "" {
```
jasonpaulos, line 1124: I think this should be moved right before `// Broadcast`, so the transaction does not always get signed. This matches the behavior of the other subcommands.

jasonpaulos (commented):
Mostly good, I just have one suggestion

jasonpaulos (approved):

PR #3208, opened 2021-11-12
Clarify get_ex opcode docs

barnjamin:
The `app_{local|global}_get_ex` opcode will return 0 for exists flag if the app or key don't exist.


jannotti:
Be sure to run `make` in data/transactions/logic and check in the shed changes that result.

jannotti (approved):
Thanks.

PR #3209, opened 2021-11-14
ledger refactoring: add support for merkle trie regeneration

tsachiherman:
## Summary

This PR updates the merkle trie regeneration to use the new schema.

## Test Plan

Updated unit tests.


ledger/accountdb.go
```diff
@@ -125,7 +126,7 @@ var accountsResetExprs = []string{
 // details about the content of each of the versions can be found in the upgrade functions upgradeDatabaseSchemaXXXX
 // and their descriptions.
 // TODO - this should be bumped to 6 if we want to enable the new database schema upgrade.
-var accountDBVersion = int32(5)
+var accountDBVersion = int32(6)
```
cce, line 129: This basically breaks all the tests until we get to them — WDYT of writing new tests or modifying tests that might be working with current v6 support using: 
```
func TestWorkingV6Thing(t *testing.T) {
  accountDBVersion = int32(6)
  t.Cleanup( func() { accountDBVersion = int32(5) } )
  ...
}

File: ledger/accountdb.go [hunk 2]
```diff
@@ -1936,7 +1947,12 @@ func (iterator *orderedAccountsIter) Next(ctx context.Context) (acct []accountAd
 	}
 	if iterator.step == oaiStepQueryAccounts {
 		// iterate over the existing accounts
-		iterator.rows, err = iterator.tx.QueryContext(ctx, "SELECT address, data FROM accountbase")
+		iterator.accountBaseRows, err = iterator.tx.QueryContext(ctx, "SELECT rowid, address, data FROM accountbase ORDER BY rowid")
```
cce, line 1950: without an ORDER BY, did it used to implicitly order by rowid? just curious

ledger/accountdb.go
tsachiherman: I think that we should avoid merging this one until the supporting code is done. It would make it much easier.
( I'm going to branch off this one + Pavel's changes and make the corresponding account updates changes. I don't think it would be that complicated )
The change you've proposed above would create a data race, btw.

ledger/accountdb.go
tsachiherman: I'm not sure that it would always be the same. especially if use have updates.
there are cases, however, where it will be : if the table is append-only, than the order will remain the same.

ledger/accountdb.go
```diff
@@ -1968,29 +1985,77 @@ func (iterator *orderedAccountsIter) Next(ctx context.Context) (acct []accountAd
...
 			if err != nil {
 				iterator.Close(ctx)
 				return
 			}
 
+			resourcesEntriesCount := accountData.TotalAppParams + accountData.TotalAppLocalStates + accountData.TotalAssetParams + accountData.TotalAssets
+			for ; resourcesEntriesCount > 0; resourcesEntriesCount-- {
+				if !iterator.resourcesRows.Next() {
+					iterator.Close(ctx)
+					err = errors.New("missing entries in resource table")
```
cce, line 2005: It would be nice for errors.New() to only occur in a var declaration and wrapped e.g. `fmt.Errorf("missing entry for %s in resource table: %w", addr.String(), errMismatchedAccountResources)` or similar. That would also provide the benefit of identifying the offending address with the wrong resources.

That reminds me that I was hoping to enable the go-err113 linter which would complain "do not define dynamic errors, use wrapped static errors instead" for cases like this.

tsachiherman:
The content of this PR was already merged in via schemaupdate4, so I'm closing this one.

PR #3214, opened 2021-11-15
ledger refactoring: fix catchpoint tracker

tsachiherman:
## Summary

When implementing the catchpoint tracker, I missed the re-initilization location for some of the local variables.
This would generate incorrect catchpoint labels after a node completes a fast-catchup.

https://github.com/algorand/go-algorand/pull/3085

## Test Plan

- [x] Add unit tests
- [x] Perform manual testing

### DIFF
--- a/ledger/catchpointtracker.go
+++ b/ledger/catchpointtracker.go
@@ -297,7 +297,7 @@ func (ct *catchpointTracker) prepareCommit(dcc *deferredCommitContext) error {
 	ct.catchpointsMu.RLock()
 	defer ct.catchpointsMu.RUnlock()
 	if dcc.isCatchpointRound {
-		dcc.committedRoundDigest = ct.roundDigest[dcc.offset+uint64(dcc.lookback)-1]
+		dcc.committedRoundDigest = ct.roundDigest[dcc.offset+uint64(dcc.lookback)]
 	}
 	return nil
 }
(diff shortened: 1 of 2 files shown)


cce:
Huh, so should I have been seeing this more than i did?

tsachiherman:
> Huh, so should I have been seeing mismatched block digests show up more than I did?

I depends on the precise scenario. A node that was using the fast catchup would produce an incorrect catchpoint label from that point and on. This would "fix itself" if the node happen to reset ( although, the historical generated catchpoint labels would remain incorrect ).

algorandskiy (approved):

ledger/catchpointtracker.go
```diff
@@ -159,6 +159,12 @@ func (ct *catchpointTracker) loadFromDisk(l ledgerForTracker, lastBalancesRound
 	ct.log = l.trackerLog()
 	ct.dbs = l.trackerDB()
 
+	ct.roundDigest = nil
+	ct.catchpointWriting = 0
+	// keep these channel closed if we're not generating catchpoint
+	ct.catchpointSlowWriting = make(chan struct{}, 1)
+	close(ct.catchpointSlowWriting)
```
cce, line 166: Hrmm ... maybe renaming the loadFromDisk tracker interface method to "reloadFromDisk" or "resetFromDisk" or something to make it clear for future authors that a tracker could have this method called multiple times during its lifetime?

cce (approved):

ledger/catchpointtracker.go
tsachiherman: I'm not opposing to that. I think that the existing method description in the interface definition already define that.

PR #3218, opened 2021-11-17
AVM: opcode for ed25519 verify of arbitrary messages.

jannotti:
Brings ed25519 verify to feature parity with newly added ecdsa opcodes.  Allows verification of any message, rather than requiring "program" and hash(code) prefix. 

Same sort of testing as ed25519verify


tsachiherman:
@jannotti could you hold off with this PR until after the batch verification get merged in ? There are some verification semantics that would be better off be uniform.

jannotti:
> @jannotti could you hold off with this PR until after the batch verification get merged in ? There are some verification semantics that would be better off be uniform.

Sure, it's not time critical (belongs in AVM 1.1 and is gated by TEAL 6 in vFuture).  But how can batch verification impact this?  This is only for signature checking inside of AVM code. As long as there exists a crypto.SignatureVerifier.VerifyBytes type function (for verifying a single signature), this code will be unchanged.

pzbitskiy (approved):
The code looks good. Not sure about `_bare` naming, and the use case (remembering `ed25519verify` has such specific behavior because of avoiding some kind of misuses).

jannotti:
Now that we have the `EnableBatchVerification` consensus parameter, would it be reasonable to put this into the upcoming release?  The call to `func (v SignatureVerifier) VerifyBytes` would be supplied the BatchVerify flag. Since it would release along with BatchVerification, there would be no backward compatibility issues.

ryanRfox:
Please consider an alternate to "bare" which may be more descriptive, such as "bytes" or "data".

ryanRfox:
@algorandskiy regarding _misuse_ of this opcode, how and where do we inform developers of these risks and proper mitigation? OpCode description and/or Dev Portal docs?

jasonpaulos (approved):
Code changes look good.

My only suggestion is that it would be nice to add a note in `opDocExtras` in either/both of the ed25519 opcode descriptions to describe how they differ, but that is not essential.

PR #3222, opened 2021-11-17
CI: use libboost-math-dev instead of libboost-all-dev

cce:
## Summary

## Test Plan

Builds should pass as before. Now that we are no longer using Travis for Linux builds, the side effect of libboost-all-dev installing make and other missing build tools on Travis encountered in #2717 is no longer a concern.


cce:
I broke this PR, it should be #3223 instead

PR #3228, opened 2021-11-18
Fix OnCompletion flag for `goal app method`

jasonpaulos:
## Summary

Fix the `--on-completion` flag for `goal app method`, introduced in #3088.

## Test Plan

Added test to `test/scripts/e2e_subs/e2e-app-abi-add.sh`

### DIFF
--- a/cmd/goal/application.go
+++ b/cmd/goal/application.go
@@ -53,7 +53,7 @@ var (
 
 	extraPages uint32
 
-	createOnCompletion string
+	onCompletion string
 
 	localSchemaUints      uint64
 	localSchemaByteSlices uint64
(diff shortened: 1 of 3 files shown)


ahangsu:
Looks pretty legit. I think it should be good to go once test (and related e2e test) passed.

cmd/goal/application.go
```diff
@@ -120,6 +120,7 @@ func init() {
 
 	methodAppCmd.Flags().StringVar(&method, "method", "", "Method to be called")
 	methodAppCmd.Flags().StringArrayVar(&methodArgs, "arg", nil, "Args to pass in for calling a method")
+	methodAppCmd.Flags().StringVar(&onCompletion, "on-completion", "NoOp", "OnCompletion action for application transaction")
```
jasonpaulos, line 123: This is the important part that was missing. Everything else is just a name change from `createOnCompletion` to `onCompletion`

ahangsu: Yes, I think this is what cobra should do to have a default `NoOp` for onCompletion in method.

jannotti (approved):
This looks good.  If I'm understanding properly, we _only_ supported on-completion for app create before?  Why not for normal app calls?

That's certainly a little outside what we're doing with the ABI work right now, but would it be reasonable to add the flag to that as well?

jasonpaulos:
The `--on-completion` flag was previously used only for `goal app create` because every OnCompletion value was baked into the subcommand. E.g. `goal app call` implies NoOp, `goal app optin` implies OptIn, etc.

jannotti:
> The `--on-completion` flag was previously used only for `goal app create` because every OnCompletion value was baked into the other subcommands. E.g. `goal app call` implies NoOp, `goal app optin` implies OptIn, etc.
> 
> Maybe there is value in supporting it for `goal app call`, but really it's just as easy to use the subcommand name as the OnCompletion value.

I see, yeah, that's true, as long as `goal app optin` supports all the usual arg and foreign array building flags.

PR #3229, opened 2021-11-18
ledger refactoring: generate catchpoints

tsachiherman:
## Summary

This PR updates the catchpoint generation to use the new schema.

## Test Plan

Testing for the content of the generated catchpoint files would be done in future PRs.


ledger/accountdb.go
```diff
@@ -192,24 +195,73 @@ const (
...
+		for _, resource := range resources {
+			normalizedAccountBalances[i].resources[resource.aidx] = resource.resourcesData
+		}
+		normalizedAccountBalances[i].encodedAccountData = balance.AccountData
+
+	}
+	return
+}
+
+// prepareNormalizedBalancesV6 converts an array of encodedBalanceRecordV5 into an equal size array of normalizedAccountBalances.
```
algorandskiy, line 232: encodedBalanceRecordV5 -> encodedBalanceRecordV6

ledger/accountdb.go
```diff
@@ -192,24 +195,73 @@ const (
...
+		for cidx, res := range balance.Resources {
+			var resData resourcesData
+			err = protocol.Decode(res, &resData)
+			if err != nil {
+				return nil, err
+			}
+			ctype := basics.AssetCreatable
+			if resData.IsEmptyApp() {
+				ctype = basics.AppCreatable
+			}
```
algorandskiy, line 260: why does `resData.IsEmptyApp()` trigger ctype = basics.AppCreatable ? I guess the resolution must be done on `ResourceFlags`.

ledger/accountdb.go
```diff
@@ -786,6 +878,57 @@ type resourcesData struct {
...
+		rd.Clawback.IsZero()
+}
+
+func (rd *resourcesData) IsAsset() bool {
+	if (rd.ResourceFlags & resourceFlagsEmptyAsset) == resourceFlagsEmptyAsset {
+		return true
+	}
+	return !rd.IsEmptyAsset()
+}
 func (rd *resourcesData) SetAssetParams(ap basics.AssetParams, haveHoldings bool) {
```
algorandskiy, line 932: btw, I think this is error prone. It is probably better setting `rd.ResourceFlags |= resourceFlagsNotHolding` unconditionally and clear this flag in SetAssetHolding

tsachiherman: I agree. I'll need to think how to re-implement this so that the *persistence* would be taking advantage of the zero for holding, while the normal operation wouldn't ( i.e. maybe two separate variables ?)

ledger/accountdb.go
```diff
@@ -786,6 +878,57 @@ type resourcesData struct {
...
+		rd.Decimals == 0 &&
+		!rd.DefaultFrozen &&
+		rd.UnitName == "" &&
+		rd.AssetName == "" &&
+		rd.URL == "" &&
+		rd.MetadataHash == [32]byte{} ||
+		rd.Manager.IsZero() ||
+		rd.Reserve.IsZero() ||
+		rd.Freeze.IsZero() ||
+		rd.Clawback.IsZero()
```
algorandskiy, line 923: should be `&&`. Otherwise a record with only holding set returns True. I'll fix with a PR with a test

ledger/accountdb.go
tsachiherman: exactly. We don't store the asset/app differentiation anywhere else, which is why we need to use the ResourceFlags to figure this out.

ledger/accountdb.go
```diff
@@ -1343,58 +1443,63 @@ func (rd *resourcesData) GetAppParams() basics.AppParams {
 	}
 }
 
-func accountDataResources(
-	ctx context.Context,
-	accountData *basics.AccountData, rowid int64,
-	cb func(ctx context.Context, rowid int64, cidx basics.CreatableIndex, ctype basics.CreatableType, rd *resourcesData) error,
```
algorandskiy, line 1349: I need this call-backed approach in catchpoint dump to reuse this function. so why removing?

ledger/accountdb.go
```diff
@@ -1496,19 +1601,16 @@ func performResourceTableMigration(ctx context.Context, tx *sql.Tx, log func(pro
 		if rowsAffected != 1 {
 			return fmt.Errorf("number of affected rows is not 1 - %d", rowsAffected)
 		}
-		rowid, err = insertRes.LastInsertId()
+		rowID, err = insertRes.LastInsertId()
 		if err != nil {
 			return err
 		}
-
-		cb := func(ctx context.Context, rowID int64, cidx basics.CreatableIndex, ctype basics.CreatableType, rd *resourcesData) error {
```
algorandskiy, line 1504: old and new code are equialiet

ledger/accountdb.go
```diff
@@ -1496,19 +1601,16 @@ func performResourceTableMigration(ctx context.Context, tx *sql.Tx, log func(pro
 		if rowsAffected != 1 {
 			return fmt.Errorf("number of affected rows is not 1 - %d", rowsAffected)
 		}
-		rowid, err = insertRes.LastInsertId()
+		rowID, err = insertRes.LastInsertId()
```
algorandskiy, line 1604: changed rowID to rowid because the latter is used everywhere else

ledger/catchpointwriter.go
```diff
@@ -38,8 +38,11 @@ const (
 	// note that the last chunk would typically be less than this number.
 	BalancesPerCatchpointFileChunk = 512
 
-	// catchpointFileVersion is the catchpoint file version
-	catchpointFileVersion = uint64(0200)
+	// catchpointFileVersionV5 is the catchpoint file version that was used when the database schema was V0-V5.
+	catchpointFileVersionV5 = uint64(0200)
+
+	// catchpointFileVersionV6 is the catchpoint file version that is matching database schema V6
+	catchpointFileVersionV6 = uint64(0201)
```
algorandskiy, line 45: please make exported, needed by catchpointdump

algorandskiy (commented):
Please preserve callback-based interface for accountDataResources. Other than that looks good to me.

ledger/accountdb.go
tsachiherman: one of the linters want the ID to be written with capital letters..

ledger/catchpointwriter.go
tsachiherman: np

ledger/accountdb.go
tsachiherman: rolled back the change.

ledger/accountdb.go
```diff
@@ -1397,6 +1507,65 @@ func accountDataResources(
...
 }
 
+//msgp:ignore resourcesRow
+type resourcesRow struct {
+	aidx  basics.CreatableIndex
+	rtype basics.CreatableType
+	resourcesData
+}
+
+func accountDataResources2(accountData *basics.AccountData) (rows []resourcesRow) {
```
algorandskiy, line 1517: looks like not used

algorandskiy (approved):

PR #3233, opened 2021-11-19
Support transaction arguments for `goal app method`

### LINKED ISSUE #3226: Support unsigned transactions as ABI arguments in `goal app method`
## Summary
This is a follow up to #3088, which added the `goal app method` subcommand for invoking ABI methods.

In addition to standard ABI type arguments, the subcommand should support the special transaction argument types `txn`, `pay`, `keyreg`, `acfg`, `axfer`, `afrz`, and `appl`, [as described in ARC-4](https://github.com/algorandfoundation/ARCs/blob/a98c1940d13194b476b543fe6e16b8798c5b5e00/ARCs/arc-0004.md#transaction-types).

This is related to #3205, which tracks adding foreign type support (i.e. `account`, `asset`, `application` types) to `goal app method`.

## Scope
Extend the `goal app method` subcommand to support these new method argument types. At a high level it should:

1. Recognize when an argument is a transaction type or a standard ABI type. The transaction argument types are only valid as top-level arguments, so direct string comparison should be sufficient.
2. Load all transaction arguments. The syntax should probably be `--arg path/to/unsigned/txn/file.msgp`, so the subcommand needs load the transaction from a file and decode it.
3. Encode all non-transaction arguments as if the transaction arguments were not present.
4. If transaction arguments are present, create an atomic transaction group and place the transactions arguments before the final app call transaction. Calculate and assign a group ID to all transactions.
5. Open question: how can the transaction arguments be signed? Perhaps if the private key of every sender is known to the wallet, they can be signed automatically and submitted, and otherwise only an unsigned (or partially signed) group can be output from this command?

jasonpaulos:
## Summary

Add the ability to pass unsigned transactions as arguments to `goal app method`.

This is done by passing the name of a file containing the unsigned transaction as an argument to the method. For example:

```bash
# create an unsigned payment trasaction
goal clerk send --from $SENDER --to $RECEIVER --amount 1000000 -o "txn-arg.tx"

# create and execute a method call using the unsigned payment transaction as an argument
goal app method --method "payment(pay,uint64)bool" --arg "txn-arg.tx" --arg 1000000 --app-id $APPID --from $SENDER
```

Closes #3226

## Test Plan

e2e test added


jannotti (approved):
I like it, worth taking as is.  But how do you make your teal e2e subs?  The labels look like maybe from PyTeal?  Maybe we should translate them directly in the test? I think goal.py could do it, then this would become a (simpler?) .py test.

barnjamin (approved):

jannotti (approved):

jasonpaulos:
> I like it, worth taking as is. But how do you make your teal e2e subs? The labels look like maybe from PyTeal? Maybe we should translate them directly in the test? I think goal.py could do it, then this would become a (simpler?) .py test.

Yes the TEAL code was generated from PyTeal. Using goal.py would have been nice since the PyTeal program could be compiled at runtime, but the purpose of this test is to test `goal`, and as I understand it goal.py does not actually invoke `goal`?

jannotti:
True, the only benefit would be in having the test program itself be more
readable.  You would still need to invoke goal so it's certainly not worth
it to add that kind of complexity for this test.

PR #3234, opened 2021-11-20
apply.balances part2: cow to acct update changes

algorandskiy:
## Summary

Part 2 of `apply.balances` and `cow` changes. This PR introduces partial deltas (`NewAccountDelta`) and some glue to make it work with existing old schema. Most of this glue code must be removed when a new schema is enabled.

## Test Plan

- [x] Existing tests
- [x] Add `Total*` fields validation test

### DIFF
--- a/ledger/accountdb_test.go
+++ b/ledger/accountdb_test.go
@@ -618,6 +618,7 @@ func BenchmarkWritingRandomBalancesDisk(b *testing.B) {
 
 		// read all the accounts to obtain the addrs.
 		rows, err := tx.Query("SELECT rowid, address FROM accountbase")
+		require.NoError(b, err)
 		defer rows.Close()
 		for rows.Next() {
 			var addrbuf []byte
@@ -660,7 +661,7 @@ func BenchmarkWritingRandomBalancesDisk(b *testing.B) {
 			}
 
 		}
-		b.ReportMetric(float64(int(time.Now().Sub(startTime))/b.N), "ns/acct_update")
+		b.ReportMetric(float64(int(time.Since(startTime))/b.N), "ns/acct_update")
 	})
 
 	b.Run("ByRowID", func(b *testing.B) {
@@ -686,7 +687,7 @@ func BenchmarkWritingRandomBalancesDisk(b *testing.B) {
 				}
 			}
 		}
-		b.ReportMetric(float64(int(time.Now().Sub(startTime))/b.N), "ns/acct_update")
+		b.ReportMetric(float64(int(time.Since(startTime))/b.N), "ns/acct_update")
 
 	})
 
@@ -859,7 +860,7 @@ func benchmarkWriteCatchpointStagingBalancesSub(b *testing.B, ascendingOrder boo
 			}
 			balances.Balances[i] = randomAccount
 		}
-		balanceLoopDuration := time.Now().Sub(balancesLoopStart)
+		balanceLoopDuration := time.Since(balancesLoopStart)
 		last64KAccountCreationTime += balanceLoopDuration
 		accountsGenerationDuration += balanceLoopDuration
 
@@ -875,13 +876,13 @@ func benchmarkWriteCatchpointStagingBalancesSub(b *testing.B, ascendingOrder boo
 		accountsLoaded += chunkSize
 	}
 	if !last64KStart.IsZero() {
-		last64KDuration := time.Now().Sub(last64KStart) - last64KAccountCreationTime
+		last64KDuration := time.Since(last64KStart) - last64KAccountCreationTime
 		fmt.Printf("%-82s%-7d (last 64k) %-6d ns/account       %d accounts/sec\n", b.Name(), last64KSize, (last64KDuration / time.Duration(last64KSize)).Nanoseconds(), int(float64(last64KSize)/float64(last64KDuration.Seconds())))
 	}
 	stats, err := l.trackerDBs.Wdb.Vacuum(context.Background())
 	require.NoError(b, err)
 	fmt.Printf("%-82sdb fragmentation   %.1f%%\n", b.Name(), float32(stats.PagesBefore-stats.PagesAfter)*100/float32(stats.PagesBefore))
-	b.ReportMetric(float64(b.N)/float64((time.Now().Sub(accountsWritingStarted)-accountsGenerationDuration).Seconds()), "accounts/sec")
+	b.ReportMetric(float64(b.N)/float64((time.Since(accountsWritingStarted)-accountsGenerationDuration).Seconds()), "accounts/sec")
 }
 
 func BenchmarkWriteCatchpointStagingBalances(b *testing.B) {
--- a/ledger/accountdb_test.go
+++ b/ledger/accountdb_test.go
@@ -24,7 +24,6 @@ import (
 	"fmt"
 	"math/rand"
 	"os"
-	"reflect"
 	"sort"
 	"strings"
 	"testing"
@@ -62,9 +61,6 @@ func checkAccounts(t *testing.T, tx *sql.Tx, rnd basics.Round, accts map[basics.
 		pad, err := aq.lookup(addr)
 		require.NoError(t, err)
 		d := pad.accountData
-		if !reflect.DeepEqual(d, data) {
-			fmt.Println("not equal", addr.String())
-		}
 		require.Equal(t, d, data)
 
 		switch d.Status {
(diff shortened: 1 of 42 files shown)


algorandskiy:
Rebased on top of the feature branch

ledger/accountdb_test.go
```diff
@@ -59,8 +60,11 @@ func checkAccounts(t *testing.T, tx *sql.Tx, rnd basics.Round, accts map[basics.
 
 	for addr, data := range accts {
 		pad, err := aq.lookup(addr)
-		d := pad.accountData
 		require.NoError(t, err)
+		d := pad.accountData
+		if !reflect.DeepEqual(d, data) {
```
tsachiherman, line 65: I'm surprised you couldn't have used require.Equal here.

ledger/accountdb_test.go
```diff
@@ -59,8 +60,11 @@ func checkAccounts(t *testing.T, tx *sql.Tx, rnd basics.Round, accts map[basics.
...
+			fmt.Println("not equal", addr.String())
```
tsachiherman, line 66: what's the added value of this compared to the `require.Equal` statement below ?

ledger/internal/appcow.go
```diff
@@ -641,91 +641,80 @@ func (lsd *storageDelta) applyChild(child *storageDelta) {
...
 		case allocAction, remainAllocAction:
 			// note: these should always exist because they were
 			// at least preceded by a call to Put (opting in),
 			// or the account has opted in before and local states are pre-allocated
-			states, ok := owned[aapp.aidx]
-			if !ok {
-				return basics.AccountData{}, fmt.Errorf("could not find existing states for %v", aapp.aidx)
+			states, exist, err := cb.lookupAppLocalState(addr, aapp.aidx)
+			if err != nil {
+				return fmt.Errorf("fetching storage (global=%v) failed for (%s, %d): %s", aapp.global, addr.String(), aapp.aidx, err.Error())
```
tsachiherman, line 693: use `%w` as the last error argument.

ledger/internal/cow.go
```diff
@@ -118,21 +122,24 @@ func (cb *roundCowState) deltas() ledgercore.StateDelta {
...
-		var delta basics.AccountData
-		var exist bool
-		if delta, exist = cb.mods.Accts.Get(addr); !exist {
-			ad, err := cb.lookup(addr)
+		for aapp, storeDelta := range smap {
+			// app params and app local states might be accessed without touching base account record
+			// from TEAL by writing/deleting KV store
+			// so store the base account to deltas here
+			// TODO: remove after the schema switch
+			acct, err := cb.lookup(addr)
```
tsachiherman, line 130: this method need to be re-written so that instead of panic, we would return an error. This doesn't need to be done in the context of this PR.

ledger/ledgercore/accountdata.go
```diff
@@ -0,0 +1,180 @@
...
+	u.VoteLastValid = basics.Round(0)
+	u.VoteKeyDilution = 0
+	u.VoteID = crypto.OneTimeSignatureVerifier{}
+	u.SelectionID = crypto.VRFVerifier{}
+}
+
+// MinBalance computes the minimum balance requirements for an account based on
+// some consensus parameters. MinBalance should correspond roughly to how much
+// storage the account is allowed to store on disk.
+func (u AccountData) MinBalance(proto *config.ConsensusParams) (res basics.MicroAlgos) {
```
tsachiherman, line 140: please extract the logic to be implemented in userBalance.go, and have it called from here.
i.e. pass all the relevant parameters on the stack. The compiler is going to inline these.

tsachiherman (changes requested):
Few minor requests, but looks great otherwise.
It would be nice to have all the reviewdog style changes in a separate PR.

ledger/ledgercore/accountdata.go
algorandskiy: done

ledger/ledgercore/statedelta.go
```diff
@@ -133,82 +155,298 @@ func MakeStateDelta(hdr *bookkeeping.BlockHeader, prevTimestamp int64, hint int,
...
+func (ad NewAccountDeltas) ModifiedAccounts() []basics.Address {
 	result := make([]basics.Address, len(ad.accts))
 	for i := 0; i < len(ad.accts); i++ {
 		result[i] = ad.accts[i].Addr
 	}
+
+	// consistency check: ensure all addresses in params/holdings/states are also in base account
+	for aapp := range ad.appParams {
+		if _, ok := ad.acctsCache[aapp.Address]; !ok {
+			panic(fmt.Sprintf("account app param delta: addr %s not in base account", aapp.Address))
```
cce, line 214: Are these checks & panics going to be temporary, or intended to stay in forever?

ledger/ledgercore/statedelta.go
```diff
@@ -230,3 +468,320 @@ func (sd *StateDelta) OptimizeAllocatedMemory(proto config.ConsensusParams) {
...
+	result := basics.AccountData{}
+	acct := ad.accts[idx].AccountData
+	AssignAccountData(&result, acct)
+
+	if len(ad.appParams) > 0 {
+		result.AppParams = make(map[basics.AppIndex]basics.AppParams)
+		for aapp, val := range ad.appParams {
+			if aapp.Address == addr {
+				if val == nil {
+					delete(result.AppParams, aapp.App)
```
cce, line 488: if you're building up result.AppParams from empty here, won't you not need to delete the app ID?

ledger/ledgercore/statedelta.go
algorandskiy: they were thought to be temporary for updating consistency checks: when we modify any resource we must update/decrement counters so `accts/acctsCache` have such address. There are also similar checks in cow so I think these could be safely rmoved

ledger/ledgercore/statedelta.go
algorandskiy: Good catch, this is leftover from the initial prototype implementation. This is used only in tests now, I'll consider removing this func completely.

algorandskiy: done

tsachiherman (approved):

ledger/ledgercore/statedelta.go
```diff
@@ -103,23 +108,40 @@ type StateDelta struct {
...
 	// with empty `AccountData`.
-	accts []basics.BalanceRecord
+	accts []newBalanceRecord
 	// cache for addr to deltas index resolution
 	acctsCache map[basics.Address]int
+
+	appParams      map[AccountApp]*basics.AppParams
+	appLocalStates map[AccountApp]*basics.AppLocalState
+	assetParams    map[AccountAsset]*basics.AssetParams
+	assets         map[AccountAsset]*basics.AssetHolding
```
cce, line 136: you use pointers here but `cachedAppParams.exists` struct/bool wrapped objects for roundCowBase, is there a difference?

ledger/ledgercore/statedelta.go
```diff
@@ -133,82 +155,298 @@ func MakeStateDelta(hdr *bookkeeping.BlockHeader, prevTimestamp int64, hint int,
...
+
+// GetAssetHolding returns asset holding delta value
+func (ad NewAccountDeltas) GetAssetHolding(addr basics.Address, aidx basics.AssetIndex) (*basics.AssetHolding, bool) {
+	holding, ok := ad.assets[AccountAsset{addr, aidx}]
+	return holding, ok
+}
+
 // ModifiedAccounts returns list of addresses of modified accounts
-func (ad *AccountDeltas) ModifiedAccounts() []basics.Address {
+func (ad NewAccountDeltas) ModifiedAccounts() []basics.Address {
```
cce, line 205: it looks like you converted all the functions that used to be pointer receivers on AccountDeltas to be value receivers on NewAccountDeltas, is that intentional?

ledger/ledgercore/statedelta.go
algorandskiy: yes, nil is an indication a resource was deleted, the cache you refer to is for base data from the underlying acct updates and DB. I store boolean field is just not to query acct updates/db again for non-existing data.

ledger/ledgercore/statedelta.go
cce: oh, only some of them. OK

algorandskiy: I changed pointer to value for read-only functions

ledger/ledgercore/statedelta.go
cce: aha, right, got it

PR #3238, opened 2021-11-22
testing: fixes to rest-participation-key e2e test

algonautshant:
## Summary
- Test to make sure RES has the right input before counting line numbers for result size.
- Rest RES to empty so that the same output is not recycled in case of an error.
- exit 1 in case of an error
- Reduce LAST_ROUND from 1200000 to 120
- "Get List of Keys" before getting NUM_IDS_3 otherwise it will recycle old RES value.

## Test Plan
This is a fix to a test.

### DIFF
--- a/test/scripts/e2e_subs/rest-participation-key.sh
+++ b/test/scripts/e2e_subs/rest-participation-key.sh
@@ -21,34 +21,21 @@ algokey part generate --first ${FIRST_ROUND} --last ${LAST_ROUND} --keyfile ${NA
 
 popd || exit 1
 
-call_and_verify "Get List of Keys" "/v2/participation" 200 'address'
+call_and_verify "Get List of Keys" "/v2/participation" 200 'address' "address" "effective-first-valid"
 
 # Find out how many keys there are installed so far
-if [[ $RES != *"address"*"effective-first-valid"* ]]; then
-    echo "Unexpected output: " $RES
-    exit 1
-fi
 NUM_IDS_1=$(echo "$RES" | python3 -c 'import json,sys;o=json.load(sys.stdin);print(len(o))')
 RES=""
 
-call_post_and_verify "Install a basic participation key" "/v2/participation" 200 ${NAME_OF_TEMP_PARTKEY} 'partId'
+call_post_and_verify "Install a basic participation key" "/v2/participation" 200 ${NAME_OF_TEMP_PARTKEY} 'partId' "partId"
 
 # Get the returned participation id from the RESULT (aka $RES) variable
-if [[ $RES != *"partId"* ]]; then
-    echo "Unexpected output: " $RES
-    exit 1
-fi
 INSTALLED_ID=$(echo "$RES" | python3 -c 'import json,sys;o=json.load(sys.stdin);print(o["partId"])')
 RES=""
 
 # Should contain the installed id
-call_and_verify "Get List of Keys" "/v2/participation" 200 'address' "${INSTALLED_ID}"
+call_and_verify "Get List of Keys" "/v2/participation" 200 'address' "${INSTALLED_ID}" "address" "effective-first-valid"
 
-# Get list of keys
-if [[ $RES != *"address"*"effective-first-valid"* ]]; then
-    echo "Unexpected output: " $RES
-    exit 1
-fi
 NUM_IDS_2=$(echo "$RES" | python3 -c 'import json,sys;o=json.load(sys.stdin);print(len(o))')
 RES=""
 
@@ -66,11 +53,8 @@ call_delete_and_verify "Delete the specific ID" "/v2/participation/${INSTALLED_I
 call_delete_and_verify "Delete the specific ID" "/v2/participation/${INSTALLED_ID}" 404 true 'participation id not found'
 
 # Get list of keys
-call_and_verify "Get List of Keys" "/v2/participation" 200 'address'
-if [[ $RES != *"address"*"effective-first-valid"* ]]; then
-    echo "Unexpected output: " $RES
-    exit 1
-fi
+call_and_verify "Get List of Keys" "/v2/participation" 200 'address' "address" "effective-first-valid"
+
 NUM_IDS_3=$(echo "$RES" | python3 -c 'import json,sys;o=json.load(sys.stdin);print(len(o))')
 RES=""
 
--- a/test/scripts/e2e_subs/rest-participation-key.sh
+++ b/test/scripts/e2e_subs/rest-participation-key.sh
@@ -21,20 +21,20 @@ algokey part generate --first ${FIRST_ROUND} --last ${LAST_ROUND} --keyfile ${NA
 
 popd || exit 1
 
-call_and_verify "Get List of Keys" "/v2/participation" 200 'address' "address" "effective-first-valid"
+call_and_verify "Get List of Keys" "/v2/participation" 200 'address' 'effective-first-valid'
 
 # Find out how many keys there are installed so far
 NUM_IDS_1=$(echo "$RES" | python3 -c 'import json,sys;o=json.load(sys.stdin);print(len(o))')
 RES=""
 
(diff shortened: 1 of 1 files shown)


test/scripts/e2e_subs/rest-participation-key.sh
```diff
@@ -9,33 +9,48 @@ date "+$0 start %Y%m%d_%H%M%S"
...
 
 algokey part generate --first ${FIRST_ROUND} --last ${LAST_ROUND} --keyfile ${NAME_OF_TEMP_PARTKEY} --parent ${ACCOUNT}
 
-popd || exit
+popd || exit 1
 
 call_and_verify "Get List of Keys" "/v2/participation" 200 'address'
 
 # Find out how many keys there are installed so far
+if [[ $RES != *"address"*"effective-first-valid"* ]]; then
```
winder, line 27: `call_and_verify` accepts a vararg for patterns to compare. So you should be able to add `effective-first-valid` there instead of having this check (and some of the checks below).

test/scripts/e2e_subs/rest-participation-key.sh
```diff
@@ -9,33 +9,48 @@ date "+$0 start %Y%m%d_%H%M%S"
...
 
 # Get list of keys
+if [[ $RES != *"address"*"effective-first-valid"* ]]; then
+    echo "Unexpected output: " $RES
+    exit 1
+fi
 NUM_IDS_2=$(echo "$RES" | python3 -c 'import json,sys;o=json.load(sys.stdin);print(len(o))')
+RES=""
 
 if [[ $((NUM_IDS_1 + 1)) -ne $NUM_IDS_2 ]]; then
```
winder, line 55: I think the way this test is using `NUM_IDS_*` is problematic. If someone added another test that installed a key, this one could become flaky.

We can remove all references to `NUM_IDS_*` and the assertions like this one that uses them. The other assertions check the for the IDs directly, so there's no need to have these catch-all checks.

i.e. after the key is installed, we query that the new key is there -- so no need to check that the number of keys is `NUM_IDS_1 + 1`.

winder (changes requested):
Thanks, Shant, these changes look good and should fix the problems. But I think we can take it one more step.

test/scripts/e2e_subs/rest-participation-key.sh
algonautshant: Can you please make a code change suggestion?

test/scripts/e2e_subs/rest-participation-key.sh
```diff
@@ -9,33 +9,48 @@ date "+$0 start %Y%m%d_%H%M%S"
...
+LAST_ROUND=120
 
 NAME_OF_TEMP_PARTKEY="tmp.${FIRST_ROUND}.${LAST_ROUND}.partkey"
 
 algokey part generate --first ${FIRST_ROUND} --last ${LAST_ROUND} --keyfile ${NAME_OF_TEMP_PARTKEY} --parent ${ACCOUNT}
 
-popd || exit
+popd || exit 1
 
 call_and_verify "Get List of Keys" "/v2/participation" 200 'address'
```
winder, line 24:
```suggestion
call_and_verify "Get List of Keys" "/v2/participation" 200 'address' 'effective-first-valid'
```

test/scripts/e2e_subs/rest-participation-key.sh
```diff
@@ -9,33 +9,48 @@ date "+$0 start %Y%m%d_%H%M%S"
...
 
 # Find out how many keys there are installed so far
+if [[ $RES != *"address"*"effective-first-valid"* ]]; then
+    echo "Unexpected output: " $RES
+    exit 1
+fi
```
winder, line 30:
```suggestion
```

test/scripts/e2e_subs/rest-participation-key.sh
```diff
@@ -9,33 +9,48 @@ date "+$0 start %Y%m%d_%H%M%S"
...
 NUM_IDS_1=$(echo "$RES" | python3 -c 'import json,sys;o=json.load(sys.stdin);print(len(o))')
+RES=""
 
 call_post_and_verify "Install a basic participation key" "/v2/participation" 200 ${NAME_OF_TEMP_PARTKEY} 'partId'
 
 # Get the returned participation id from the RESULT (aka $RES) variable
+if [[ $RES != *"partId"* ]]; then
+    echo "Unexpected output: " $RES
+    exit 1
+fi
```
winder, line 40:
```suggestion
```

test/scripts/e2e_subs/rest-participation-key.sh
```diff
@@ -9,33 +9,48 @@ date "+$0 start %Y%m%d_%H%M%S"
...
 INSTALLED_ID=$(echo "$RES" | python3 -c 'import json,sys;o=json.load(sys.stdin);print(o["partId"])')
+RES=""
 
 # Should contain the installed id
 call_and_verify "Get List of Keys" "/v2/participation" 200 'address' "${INSTALLED_ID}"
```
winder, line 45:
```suggestion
call_and_verify "Get List of Keys" "/v2/participation" 200 'address' "${INSTALLED_ID}" 'effective-first-valid'
```

test/scripts/e2e_subs/rest-participation-key.sh
```diff
@@ -9,33 +9,48 @@ date "+$0 start %Y%m%d_%H%M%S"
...
+popd || exit 1
 
 call_and_verify "Get List of Keys" "/v2/participation" 200 'address'
 
 # Find out how many keys there are installed so far
+if [[ $RES != *"address"*"effective-first-valid"* ]]; then
+    echo "Unexpected output: " $RES
+    exit 1
+fi
 NUM_IDS_1=$(echo "$RES" | python3 -c 'import json,sys;o=json.load(sys.stdin);print(len(o))')
```
winder, line 31:
```suggestion
```

test/scripts/e2e_subs/rest-participation-key.sh
```diff
@@ -9,33 +9,48 @@ date "+$0 start %Y%m%d_%H%M%S"
...
 # Get list of keys
+if [[ $RES != *"address"*"effective-first-valid"* ]]; then
+    echo "Unexpected output: " $RES
+    exit 1
+fi
 NUM_IDS_2=$(echo "$RES" | python3 -c 'import json,sys;o=json.load(sys.stdin);print(len(o))')
+RES=""
 
 if [[ $((NUM_IDS_1 + 1)) -ne $NUM_IDS_2 ]]; then
   printf "\n\nFailed test.  New number of IDs (%s) is not one more than old ID count(%s)\n\n" "${NUM_IDS_2}" "${NUM_IDS_1}"
```
winder, line 56: Also remove 2 lines below this section (github UI doesn't allow me to suggest that)
```suggestion
```

test/scripts/e2e_subs/rest-participation-key.sh
```diff
@@ -51,7 +66,13 @@ call_delete_and_verify "Delete the specific ID" "/v2/participation/${INSTALLED_I
...
+call_and_verify "Get List of Keys" "/v2/participation" 200 'address'
+if [[ $RES != *"address"*"effective-first-valid"* ]]; then
+    echo "Unexpected output: " $RES
+    exit 1
+fi
 NUM_IDS_3=$(echo "$RES" | python3 -c 'import json,sys;o=json.load(sys.stdin);print(len(o))')
+RES=""
 
 if [[ "$NUM_IDS_3" -ne "$NUM_IDS_1" ]]; then
   printf "\n\nFailed test.  New number of IDs (%s) is not equal to original ID count (%s)\n\n" "${NUM_IDS_3}" "${NUM_IDS_1}"
```
winder, line 78: Also remove 2 lines following this section
```suggestion
```

winder (approved):

PR #3239, opened 2021-11-22
Global.LatestTimestamp is NOT available in LogicSig contracts

Kiser360:
Any contract attempting to use Global.LatestTimestamp opCode will receive the following error from the network node with a 400 status:

```
global[7] not allowed in current mode
```

Looks like this will be resolved in the future, but for now I don't believe devs should be left thinking they can use this opcode.
https://github.com/algorand/go-algorand/pull/2789

Thank you @barnjamin for pointing me to the correct repo

### DIFF
--- a/data/transactions/logic/doc.go
+++ b/data/transactions/logic/doc.go
@@ -454,7 +454,7 @@ var globalFieldDocs = map[string]string{
 	"GroupSize":                 "Number of transactions in this atomic transaction group. At least 1",
 	"LogicSigVersion":           "Maximum supported TEAL version",
 	"Round":                     "Current round number",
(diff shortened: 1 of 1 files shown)


barnjamin:
Thanks!

jannotti:
This seems like a good change. We must run `make` in data/transactions/logic to regenerate docs though.

algoanne:
hey @Kiser360 , can you run `make` in data/transactions/logic to regenerate docs as JJ mentions above? Also pull in latest from master

jannotti:
I'm going to close this as I'm addressing the issue for all globals in the upcoming spec for 1.1. There are several other app-only globals, so we'll put that info in the table.

PR #3242, opened 2021-11-22
Disable parallelism for e2e-go tests

cce:
## Summary

This sets `-p 1` for the e2e-go tests, intended to make them more deterministic when running on a VM with relatively constrained resources. Since each e2e-go test might spin up a few nodes, it seems like it would help to avoid resource contention.

## Test Plan

Tests should run as before. Desired effect can be verified by looking at the test output where the value of PARALLEL_FLAG is printed out before tests are run.


winder (approved):
This seems fine to me. It might not be necessary though. The problematic tests were using `t.Parallel()` when they shouldn't have.

PR #3247, opened 2021-11-23
Gloadss

jannotti:
Standard opcode story, same tests as `gloads`


jannotti:
I expected the PR to only show the changes since after the globals were merged.  I'll re-PR.

PR #3248, opened 2021-11-24
Gloadss

jannotti:
Usual opcode updates to code, docs, tests.  Tests are the same as for gloads


data/transactions/logic/eval.go
```diff
@@ -2977,6 +2977,30 @@ func opGloads(cx *EvalContext) {
...
+	gi := cx.stack[prev].Uint
+	if gi >= uint64(len(cx.TxnGroup)) {
+		cx.err = fmt.Errorf("gloads lookup TxnGroup[%d] but it only has %d", gi, len(cx.TxnGroup))
+		return
+	}
+	scratchIdx := cx.stack[last].Uint
+	if scratchIdx >= 256 {
+		cx.err = fmt.Errorf("gloadss scratch index >= 256 (%d)", scratchIdx)
+		return
+	}
```
jasonpaulos, line 2993: It seems `opGloadImpl` does these checks as well, though with slightly different types? There may be a benefit to unifying these checks at some point, since that seems odd to me.

data/transactions/logic/opcodes.go
```diff
@@ -221,7 +221,7 @@ var OpSpecs = []OpSpec{
 	// Group scratch space access
 	{0x3a, "gload", opGload, asmDefault, disDefault, nil, oneAny, 4, runModeApplication, immediates("t", "i")},
 	{0x3b, "gloads", opGloads, asmDefault, disDefault, oneInt, oneAny, 4, runModeApplication, immediates("i")},
-	// Access creatable IDs
+	// Access creatable IDs (consider deprecating, as txn CreatedAssetID, CreatedApplicationID should be enough
```
jasonpaulos, line 224: Do we have plans to make scratch slots accessible with transaction field types as well? They are pretty similar to logs, so it would make sense to me.

jasonpaulos (approved):
A good improvement, I just had some nonconsequential questions.

data/transactions/logic/eval.go
ahangsu: ^ it looks alike with inputs from different places (and different types I suppose), would be nice to unify the checks too

jannotti: I think that, with some care, we could eliminate some repeat code.  The "rule" right now is that any time a type is being narrowed before sending it to the Impl function, it is tested ahead of time.  But, since unnarrowed types are *not* tested ahead of time, the Impl repeats the test.  We could probably use wider types for the Impl() function, and only do the tests in there.

ahangsu (approved):

PR #3250, opened 2021-11-24
Add format hint so that client generators produce long instead of int.

FrankSzendzielarz:
## Summary

Allow client generators to produce long instead of int for uint64 types

## Test Plan

Generated NSwag client locally

### DIFF
--- a/daemon/algod/api/algod.oas2.json
+++ b/daemon/algod/api/algod.oas2.json
@@ -1525,8 +1525,9 @@
       "properties": {
         "amount": {
           "description": "\\[a\\] number of units held.",
-           "type": "integer",
-           "x-algorand-format": "uint64"
+          "type": "integer",
+          "format": "int64",
+          "x-algorand-format": "uint64"
         },
         "asset-id": {
           "description": "Asset ID of the holding.",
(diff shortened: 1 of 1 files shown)


winder:
Could you give some more information about the problem this is solving?

In the places specifying uint64, we specifically need the type to be unsigned. So a long is insufficient. For this reason we've built a custom generator that understands the x-algorand-format type. We've also extended oapi-codegen for server side code generation.

FrankSzendzielarz:
Thanks, Will. I have taken on the task of updating the community .NET SDK, which seems to have been originally ported from the Java SDK. I have been using the Java SDK therefore to understand the AVM 1.0 updates. Some of those updates include changes to the models/api for the web client code.

With the standard tooling I have, based on NSwag, the output is a single file, so simply searching for generated 'long' and replacing it with ulong makes things simple for me, and for Java clients long is ambiguous about signed/unsigned anyway.

But yes, you are right, in hindsight this PR probably doesn't make much sense for broader audiences, especially if you have a custom code generator. Please could you tell me more about it? Does it generate for the different supported SDKs (python, java, go...). If so I could look to extend it for the .NET generation. I am on the builders Discord channel too by the way - I think barnji pinged you yesterday, if you're the same Will.

winder:
Closing this PR, after further discussions with Frank the issue is a bit larger than this small patch and we'd like to find a way to collaborate and address the problem head-on.

PR #3251, opened 2021-11-27
Adding libtool to ubuntu deps

barnjamin:
## Summary

The sandbox is not building with dev config using master branch https://github.com/algorand/sandbox/issues/85, complains about libtool not being installed

Guessing from this change https://github.com/algorand/go-algorand/pull/3223 

Adding libtool to UBUNTU_DEPS in install scripts

## Test Plan

Set config in sandbox to my branch
`sandbox up dev`
It built

### DIFF
--- a/scripts/install_linux_deps.sh
+++ b/scripts/install_linux_deps.sh
@@ -5,7 +5,7 @@ set -e
 DISTRIB=$ID
 
 ARCH_DEPS="boost boost-libs expect jq autoconf shellcheck sqlite python-virtualenv"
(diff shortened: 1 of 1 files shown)


winder (approved):

cce (approved):
sorry about that, I guess libtool is a dependency of libboost-all-dev that is already installed on CircleCI but not in a fresh docker container.

tsachiherman (commented):
Is that tool is required by the go-algorand , or the sandbox scriptings ? If it's due to the latter, then we should modify the sandbox.

PR #3252, opened 2021-11-29
Make dev-mode tests less flaky.

winder:
## Summary

Fix a couple flaws in the new go-e2e tests built ontop of DevMode:
* Shutdown the fixture when finished.
* Don't run in parallel.
* Longer delays / better algorithms to wait for data flushing to complete.
* Check for "out of order" keys.

## Test Plan

N/A, this is a test.


tsachiherman (approved):
The changes looks good, but please be aware that with the new keys schemas, we'll need to lower tge 6m/3m to lower values.

tsachiherman (approved):

PR #3261, opened 2021-11-30
ParticipationRegistry - StateProof loading methods

winder:
## Summary

Add ParticipationRegistry methods for setting and retrieving state proof keys. Since they aren't in master yet there is a `type StateProofKey []byte` stub which will need to be updated later.
 
## Test Plan

New unit tests.

### DIFF
--- a/data/account/participationRegistry.go
+++ b/data/account/participationRegistry.go
@@ -177,7 +177,7 @@ type ParticipationRegistry interface {
 	Insert(record Participation) (ParticipationID, error)
 
 	// AppendKeys appends state proof keys to an existing Participation record. Keys can only be appended
-	// once, an error
+	// once, an error will occur when the data is flushed when inserting a duplicate key.
 	AppendKeys(id ParticipationID, keys map[uint64]StateProofKey) error
 
 	// Delete removes a record from storage.
@@ -414,8 +414,7 @@ func (db *participationDB) writeThread() {
 		} else if !wr.insertID.IsZero() {
 			if wr.insert != (Participation{}) {
 				err = db.insertInner(wr.insert, wr.insertID)
-			}
-			if len(wr.keys) != 0 {
+			} else if len(wr.keys) != 0 {
 				err = db.appendKeysInner(wr.insertID, wr.keys)
 			}
 		} else if !wr.delete.IsZero() {
@@ -699,7 +698,7 @@ func (db *participationDB) AppendKeys(id ParticipationID, keys map[uint64]StateP
 		return ErrParticipationIDNotFound
 	}
 
-	keyCopy := make(map[uint64]StateProofKey)
+	keyCopy := make(map[uint64]StateProofKey, len(keys))
 	for k, v := range keys {
 		keyCopy[k] = v // PKI TODO: Deep copy?
 	}
--- a/data/account/participationRegistry.go
+++ b/data/account/participationRegistry.go
@@ -157,6 +157,11 @@ type ParticipationRegistry interface {
 	// Insert adds a record to storage and computes the ParticipationID
 	Insert(record Participation) (ParticipationID, error)
 
+	// PKI TODO: use a real type instead of []byte
+	// AppendKeys appends state proof keys to an existing Participation record. Keys can only be appended
+	// once, an error
+	AppendKeys(id ParticipationID, keys map[uint64][]byte) error
+
 	// Delete removes a record from storage.
 	Delete(id ParticipationID) error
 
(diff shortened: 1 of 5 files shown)


data/account/participationRegistryBench_test.go
```diff
@@ -0,0 +1,80 @@
+// Copyright (C) 2019-2021 Algorand, Inc.
```
winder, line 1: This isn't new, I just extracted the benchmarks into their own file.

go.mod
```diff
@@ -42,7 +42,7 @@ require (
 	github.com/sirupsen/logrus v1.4.2
 	github.com/spf13/cobra v0.0.3
 	github.com/spf13/pflag v1.0.5 // indirect
-	github.com/stretchr/testify v1.6.1
+	github.com/stretchr/testify v1.7.0
```
winder, line 45: Upgrading testify for the `ErrorIs` helper.

data/account/participationRegistry.go
```diff
@@ -82,6 +82,22 @@ type ParticipationRecord struct {
 	Voting *crypto.OneTimeSignatureSecrets
 }
 
+// StateProofKey is a placeholder for the real state proof key type.
+// PKI TODO: Replace this with a real object.
+type StateProofKey []byte
+
+// ParticipationRecordForRound adds in the per-round state proof key.
+type ParticipationRecordForRound struct {
```
tsachiherman, line 90: The struct name is abit awkward. How do you feel about RoundParticipationSigner ?

data/account/participationRegistry.go
```diff
@@ -152,11 +168,18 @@ var ErrMultipleKeysForID = errors.New("multiple valid keys found for the same pa
...
+// ErrSecretNotFound is used when attempting to lookup secrets for a particular round.
+var ErrSecretNotFound = errors.New("the participation ID did not have secrets for the requested round")
+
 // ParticipationRegistry contain all functions for interacting with the Participation Registry.
 type ParticipationRegistry interface {
 	// Insert adds a record to storage and computes the ParticipationID
 	Insert(record Participation) (ParticipationID, error)
 
+	// AppendKeys appends state proof keys to an existing Participation record. Keys can only be appended
+	// once, an error
```
tsachiherman, line 180: Incomplete sentence ?

data/account/participationRegistry.go
```diff
@@ -169,6 +192,9 @@ type ParticipationRegistry interface {
 	// GetAll of the participation records.
 	GetAll() []ParticipationRecord
 
+	// GetWithSecrets fetches a record with all secrets for a particular round.
+	GetWithSecrets(id ParticipationID, round basics.Round) (ParticipationRecordForRound, error)
```
tsachiherman, line 196: Maybe GetRoundSigningKeys ?

data/account/participationRegistry.go
```diff
@@ -380,7 +412,12 @@ func (db *participationDB) writeThread() {
 		if len(wr.registerUpdated) != 0 {
 			err = db.registerInner(wr.registerUpdated)
 		} else if !wr.insertID.IsZero() {
-			err = db.insertInner(wr.insert, wr.insertID)
+			if wr.insert != (Participation{}) {
+				err = db.insertInner(wr.insert, wr.insertID)
+			}
+			if len(wr.keys) != 0 {
```
tsachiherman, line 418: You need to add "&& err==nil"

data/account/participationRegistry.go
```diff
@@ -619,6 +691,30 @@ func (db *participationDB) Insert(record Participation) (id ParticipationID, err
...
 
+func (db *participationDB) AppendKeys(id ParticipationID, keys map[uint64]StateProofKey) error {
+	db.mutex.Lock()
+	defer db.mutex.Unlock()
+
+	if _, ok := db.cache[id]; !ok {
+		return ErrParticipationIDNotFound
+	}
+
+	keyCopy := make(map[uint64]StateProofKey)
```
tsachiherman, line 702: Preallocate map to len(keys)

go.sum
```diff
@@ -142,6 +142,8 @@ github.com/stretchr/testify v1.5.1 h1:nOGnQDM7FYENwehXlg/kFVnos3rEvtKTjRvOWSzb6H
 github.com/stretchr/testify v1.5.1/go.mod h1:5W2xD1RspED5o8YsWQXVCued0rvSQ+mT+I5cxcmMvtA=
 github.com/stretchr/testify v1.6.1 h1:hDPOHmpOpP40lSULcqw7IrRb/u7w6RpDC9399XyoNd0=
 github.com/stretchr/testify v1.6.1/go.mod h1:6Fq8oRcR53rry900zMqJjRRixrwX3KX962/h/Wwjteg=
+github.com/stretchr/testify v1.7.0 h1:nwc3DEeHmmLAfoZucVR881uASk0Mfjw8xYJ99tb5CcY=
```
tsachiherman, line 145: Please remove old testify hadhes.

tsachiherman (dismissed):
Few small requests, looks good otherwise.

data/account/participationRegistry.go
winder: I'm not attached to the name, but this has SigningKeys and the StateProof key. `GetForRound` could be good if we stick with the `ParticipationForRound` struct name. `ForRound` also suggests the extra parameter which seems like a pattern in a lot of go code.

data/account/participationRegistry.go
winder: Thanks, I didn't realize that this code technically supported inserting a new participation key while adding state proof keys.

The intention is that those would be done separately though, so I'll change this to "else if".

go.sum
winder: Sadly, we seem to have dependencies on each of these testify versions. I couldn't remove any of them.

AlgoStephenAkiki (approved):

data/account/participationRegistry.go
algoidan: because this struct contains more than just a signer, I wouldn't use the word signer in it.
do we plan to use this struct as a return value from the `keymanager` interface? or should we rather abstract the use of the registry and return a new one?

winder: The registry return values are deep copies of the data, so it's safe to use these objects globally. The current interfaces are sufficient for master, so any `keymanager` changes can be designed for state proofs.

tsachiherman (approved):

PR #3277, opened 2021-12-06
WIP - Not Ready for Review Draft: E2e test python abi

tzaffi:
# Very Much a Draft

## Summary 
* Enable Python E2E Testing of ABI Methods
* Make it possible to run interactive remote debugger for E2E tests written in Python

## Test Plan

This is all pure test code. Test by running locally.


tzaffi:
@algorandskiy and @ahangsu - this still has a lot of moving pieces so not truly ready for review. But the basic approach will be similar so am glad to accept any early feedback if you are not too busy.

test/scripts/e2e_subs/tealprogs/abi-demo.json
```diff
@@ -0,0 +1,216 @@
+{
+    "name": "demo-abi",
+    "appId": null,
```
ahangsu, line 3: nit: seems ok to remove `"appId": null,"` this line?

test/scripts/e2e_subs/tealprogs/app-base64_decode.json
```diff
@@ -0,0 +1,69 @@
+{
+    "name": "app-base64_decode",
+    "appId": null,
```
ahangsu, line 3: same here, okay to remove?

test/scripts/e2e_subs/tealprogs/abi-demo.json
tzaffi: Will need to figure out what to do with this when the Python SDK gets updated to reflect the final ARC-04

test/README.md
```diff
@@ -46,9 +46,67 @@ optional arguments:
...
 ```
 
-Tests in the `e2e_subs/serial` directory are executed serially instead of in parallel. This should only be used when absolutely necessary.
+Once these virtual environment var's are in place, you can keep using that shell to run specific E2E tests with a commands such as:
+
+```bash
+python3 test/scripts/e2e_client_runner.py $(pwd)/test/scripts/e2e_subs/your_e2e_script.py
+```
+
+### Interaction through a Remote Debugger in a Python E2E Test Script
```
tzaffi, line 71: @tsachiherman - Any thoughts?

tsachiherman: so, I think that adding these readme instructions is great - but it would be even better if we could avoid having the code snippet inside the README and having it implemented into the actual code ( /w an optional argument ).
Then, you'll be able to add a unit test to verify that it's working as designed.

PR #3278, opened 2021-12-06
Add Custom Scenario for Performance Testing

algobarb:
## Summary

Add Custom Scenario for Performance Testing.
Add README on how to run custom scenario and modify create_and_deploy_recipe.sh to accept a network template that will generate a new recipe.

## Test Plan

1. Test that a recipe gets generated by passing in the name of the network file. Algonet spun up using this recipe
2. Testing if the script still works as expected on scenario1.


algojack (commented):
lgtm. One question: is it all supposed to be in testdata folder? seems testdata implies it should be data (configs, etc) not templates, scripts, etc.

algobarb:
> lgtm. One question: is it all supposed to be in testdata folder? seems testdata implies it should be data (configs, etc) not templates, scripts, etc.

Not really sure. There does seem to be a lot in this folder for some reason and reorganization and cleaning up later on is probably a good idea. I mostly modified two existing scripts there and added a recipe where all the other recipes are (where Jenkins looks for when running the pipeline already).

bricerisingalgorand (approved):

algojack (approved):

egieseke (approved):
Looks good.

PR #3280, opened 2021-12-06
Falcon key gen time: set LastPartKeyRound 3K

algonautshant:
- remove `test` param from `CreateNetworkFromTemplate`
- Do not override `template.Genesis.LastPartKeyRound` if in test mode or if `ALGOTEST` env variable is set
- update test templates to explicitly set the `LastPartKeyRound` to 3k instead of relying on 3M default value. 

- Flag to cap last round of participation key generation, and use the flag for testing.
- Enhance notification messages to notify the user the keys are generated. 
- Reduce LastPartKeyRound for scenarios 1-3.


winder (approved):
LGTM. Is LastPartKeyRound in the template new?

PR #3282, opened 2021-12-07
TestEcdsa: fix flaky "tampering" of public key

cce:
## Summary

This test (TestEcdsa) tests the ecdsa_pk_decompress opcode and intentionally "tampers" with the public key by setting the first byte to zero. Occasionally this test is failing, likely because the first byte was already zero. (The test failures are for the cases where failure is expected, `pass=false`)

## Test Plan

Existing test should pass, occasional flakiness should go away.

### DIFF
--- a/data/transactions/logic/evalCrypto_test.go
+++ b/data/transactions/logic/evalCrypto_test.go
@@ -212,9 +212,9 @@ byte 0x%s
 &&`
 	pkTampered1 := make([]byte, len(pk))
 	copy(pkTampered1, pk)
-	pkTampered1[0] += byte(1) // intentional overflow
-	pkTampered2 := make([]byte, len(pk))
-	copy(pkTampered2, pk[1:])
+	pkTampered1[0] = 0                     // first byte is a prefix of either 0x02 or 0x03
+	pkTampered2 := make([]byte, len(pk)-1) // must be 33 bytes length
+	copy(pkTampered2, pk)
 
 	var decompressTests = []struct {
 		key  []byte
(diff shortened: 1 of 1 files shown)


jasonpaulos (dismissed):
Nice use of overflow, looks good to me

cce:
Thanks, that reminds me I was going to leave a comment saying "intentional overflow" — I will do that

algorandskiy (approved):

algorandskiy:
Btw, there are also couple other tests that tamper input in a similar way

jasonpaulos (dismissed):

tsachiherman (dismissed):

cce:
Ehh, I add a print statement to like the one below to identify whether it is pkTampered1 or pkTampered2 that was producing the most errors on master, and it is actually pkTampered2 ... so this doesn't fix that.

```
    pkTampered1 := make([]byte, len(pk))
    copy(pkTampered1, pk)
    pkTampered1[0] += 0
    pkTampered2 := make([]byte, len(pk))
    copy(pkTampered2, pk[1:])

    var decompressTests = []struct {
        key  []byte
        pass bool
    }{
        {pk, true},
        {pkTampered1, false},
        {pkTampered2, false},
    }
    for i, test := range decompressTests {
        t.Run(fmt.Sprintf("decompress/pass=%v", test.pass), func(t *testing.T) {
            t.Log("decompressTests i", i)
```

cce:
OK, this time I read up on the 33-byte ecdsa compressed public key format and made the tampering always work.

data/transactions/logic/evalCrypto_test.go
```diff
@@ -212,9 +212,9 @@ byte 0x%s
 &&`
 	pkTampered1 := make([]byte, len(pk))
 	copy(pkTampered1, pk)
-	pkTampered1[0] = 0
-	pkTampered2 := make([]byte, len(pk))
-	copy(pkTampered2, pk[1:])
+	pkTampered1[0] = 0                     // first byte is a prefix of either 0x02 or 0x03
+	pkTampered2 := make([]byte, len(pk)-1) // must be 33 bytes length
+	copy(pkTampered2, pk)
```
jasonpaulos, line 217: nit: `pkTampered2 := pk[:len(pk)-1]` would achieve the same thing

jasonpaulos (approved):
This makes sense to me

PR #3286, opened 2021-12-07
Separate Txn and PartKey validity for `goal account renewpartkey`

### LINKED ISSUE #3283: `goal account renewpartkey` does not separate transaction validity from partkey validity
### Subject of the issue

Reported by mstuder on https://forum.algorand.org/t/goal-account-renewpartkey-cannot-construct-transaction-txn-validity-period-is-greater-than-protocol-max-txn-lifetime-1000/5333?u=fabrice

`goal account renewpartkey` uses `roundLastValid` both for transaction validity and partkey validity.
This means that partkeys renewed this way can only last at most 1000 rounds.

This can be seen in the arguments of:
https://github.com/algorand/go-algorand/blob/387f6a62fb29d1a02968112601f7192710ab713c/cmd/goal/account.go#L959

### Your environment

```
$ ./sandbox goal version -v
Version: [v1 v2]
GenesisID: sandnet-v1
Build: 3.1.0.nightly [relbeta3.1.0] (commit #83740b45+)

$ ./sandbox goal node status
Last committed block: 135
Time since last block: 3.5s
Sync Time: 0.0s
Last consensus protocol: future
Next consensus protocol: future
Round for next consensus protocol: 136
Next consensus protocol supported: true
Last Catchpoint:
Genesis ID: sandnet-v1
Genesis hash: SICdQ+nhcQiu13tqol+neGty9wdlrc0OA6Dlepn1ylE=
```

### Steps to reproduce

1. Start sandbox `./sandbox up`
2. List the accounts `./sandbox account list`. Assume that account list contains `VF5SXUXEST4NUMQYDV5KSNOWEESQEYIWY4PAVL4K6QAFD5EBT3LIK6RDJE`
3. Renew the above key:
```
./sandbox goal account renewpartkey --address VF5SXUXEST4NUMQYDV5KSNOWEESQEYIWY4PAVL4K6QAFD5EBT3LIK6RDJE --roundLastValid 10000000
```

### Expected behaviour

It should work.

### Actual behaviour

It fails:

```
$ ./sandbox goal account renewpartkey --address VF5SXUXEST4NUMQYDV5KSNOWEESQEYIWY4PAVL4K6QAFD5EBT3LIK6RDJE --roundLastValid 10000000
  Generated participation key for VF5SXUXEST4NUMQYDV5KSNOWEESQEYIWY4PAVL4K6QAFD5EBT3LIK6RDJE (Valid 80 - 10000000)
  Error registering keys - deleting newly-generated key file: /opt/data/sandnet-v1/VF5SXUXEST4NUMQYDV5KSNOWEESQEYIWY4PAVL4K6QAFD5EBT3LIK6RDJE.80.10000000.partkey
cannot construct transaction: txn validity period ( 80 to 10000000 ) is greater than protocol max txn lifetime 1000
```

qkniep:
## Solution

Always use currentRound+proto.MaxTxnLife as last valid round for the transaction when renewing instead of using the partkey validity period. The partkey validity period can therefore now be more than proto.MaxTxnLife rounds long.

This fixes #3283

Should the txn validity period also be configurable via command line options instead? I am not sure about proper naming then, to not have users confuse txn and key validity periods.

### DIFF
--- a/cmd/goal/account.go
+++ b/cmd/goal/account.go
@@ -935,6 +935,7 @@ var renewParticipationKeyCmd = &cobra.Command{
 		if roundLastValid <= (currentRound + proto.MaxTxnLife) {
 			reportErrorf(errLastRoundInvalid, currentRound)
 		}
+		txRoundLastValid := currentRound + proto.MaxTxnLife
 
 		// Make sure we don't already have a partkey valid for (or after) specified roundLastValid
 		parts, err := client.ListParticipationKeyFiles()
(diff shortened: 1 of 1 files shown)


winder (approved):
These changes look correct to me, thanks for the fix!

I don't think a new command line flag is required here. This subcommand appears to intentionally simplify the process. For advanced usage with more control we have `algokey` / `goal account installpartkey` / `goal account changeonlinestatus`.

PR #3287, opened 2021-12-08
Minimum Account Balance in Algod

tzaffi:
## Summary - Implementing **Minimum Balance** for `algod REST` endpoint and `goal` command line

In particular
* `/v2/accounts/{{ACCOUNT}}` has a new `min-balance` key/val
* `goal account info` returns the **Minimum Balance**
* Python SDK's `account_info` function returns `min-balance` as well

### The new  `/v2/accounts/{{ACCOUNT}}` response looks like:

```json
{
  "address": "X2MV6PXPUBPVARJO3LMTXMPN5A7IHEJTYOT6BHQ2UD3IRWRUFORDLYS4JE",
  "amount": 0,
  "amount-without-pending-rewards": 0,
  "apps-local-state": [],
  "apps-total-schema": {
    "num-byte-slice": 0,
    "num-uint": 0
  },
  "assets": [],
  "created-apps": [],
  "created-assets": [],
  "min-balance": 100000,
  "pending-rewards": 0,
  "reward-base": 0,
  "rewards": 0,
  "round": 25,
  "status": "Offline"
}
```

### The new  `goal account info` response looks like:
```bash
$ goal -w WALLET account info -a ADDRESS
Created Assets:
	<none>
Held Assets:
	<none>
Created Apps:
	ID 21, global state used 0/0 uints, 0/0 byte slices
	ID 58, global state used 0/0 uints, 0/0 byte slices
Opted In Apps:
	<none>

... etc ...
Minimum Balance:  300000 microAlgos
```

### Automated End to End Tests:

1. For `algod` and `pysdk`: [min_balance.py](./test/scripts/e2e_subs/min_balance.py) - inspired by [these testnest scenarios](https://docs.google.com/document/d/1s7_Zh0jzgokEDg-7AYJRzSRS_jMD1thKKAYOCZykgmc/edit) (but not precisely the same)
2. For `goal` [min_balance.sh](./test/scripts/e2e_subs/min_balance.sh) - a much simpler sanity check

### Followups
We need to support `min-balance` in the following SDK's:
* go-algorand-sdk
* js-algorand-sdk
* java-algorand-sdk
* Indexer should also return `min-balance`


tsachiherman:
I'm not opposing to adding the min balance to the rest api endpoint, since it's a (almost) free operation. However, I'm not a big fan of the python testing - and not because of the usage of python itself. It's the circular dependency between the go-algorand and the python sdk repo that I'd like to see resolved.

daemon/algod/api/server/v2/handlers.go
```diff
@@ -238,6 +238,14 @@ func (v2 *Handlers) AccountInformation(ctx echo.Context, address string, params
 	if err != nil {
 		return internalError(ctx, err, errFailedLookingUpLedger, v2.Log)
 	}
+	block, _, err := myLedger.BlockCert(basics.Round(lastRound))
```
tzaffi, line 241: My first time using the ledger. I'm not sure this is the ideal way of doing so to retrieve `minBalance`

test/scripts/e2e_subs/goal/atomic_abi.py
```diff
@@ -0,0 +1,259 @@
+import json
```
tzaffi, line 1: WIP. Follow-up #3277 will clean this class up and make it easier to use.

test/scripts/e2e_subs/goal/goal.py
```diff
@@ -112,6 +112,18 @@ def open_kmd(self, algodata, kmd_address=None):
             kmd_address = "http://" + net
         return algosdk.kmd.KMDClient(kmd_token, kmd_address)
 
+    def get_endpoint_info(self) -> dict:
```
tzaffi, line 115: @jannotti - I needed this in order to test the temporary REST endpoints while pausing my debugger.

test/scripts/e2e_subs/min_balance.py
```diff
@@ -0,0 +1,203 @@
+#!/usr/bin/env python
```
tzaffi, line 1: Inspired by **Min Balance Test Accounts** [doc](https://docs.google.com/document/d/1s7_Zh0jzgokEDg-7AYJRzSRS_jMD1thKKAYOCZykgmc/edit)

test/scripts/e2e_subs/min_balance.sh
```diff
@@ -0,0 +1,39 @@
...
+
+echo "Minimum Balance --> $MINBAL"
+
+EXPECTED="100000"
+if [[ ${MINBAL} != ${EXPECTED} ]]; then
+    date '+min_balance FAIL goal account info should return expected Minimum Ballance %Y%m%d_%H%M%S'
+    false
+fi
+
+# see ./min_balance.py for more complicated scenarios
```
tzaffi, line 39: ... yep ^^^^

cmd/goal/account.go
```diff
@@ -693,6 +693,8 @@ func printAccountInfo(client libgoal.Client, address string, account generatedV2
 		fmt.Fprintf(report, "\tID %d, local state used %d/%d uints, %d/%d byte slices\n", localState.Id, usedInts, allocatedInts, usedBytes, allocatedBytes)
 	}
 
+	fmt.Fprintf(report, "Minimum Balance:\t%v microAlgos\n", account.MinBalance)
```
tzaffi, line 696: Adding a line to `goal account info`'s output

daemon/algod/api/algod.oas2.json
```diff
@@ -1365,6 +1366,10 @@
           "description": "\\[algo\\] total number of MicroAlgos in the account",
           "type": "integer"
         },
+        "min-balance": {
+          "description": "\\[algo\\] total number of MicroAlgos needed to allow transacting from account",
```
tzaffi, line 1370: Suggestions for better wording greatly appreciated

test/scripts/e2e.sh
```diff
@@ -123,7 +123,11 @@ if [ -z "$E2E_TEST_FILTER" ] || [ "$E2E_TEST_FILTER" == "SCRIPTS" ]; then
     python3 -m venv "${TEMPDIR}/ve"
     . "${TEMPDIR}/ve/bin/activate"
     "${TEMPDIR}/ve/bin/pip3" install --upgrade pip
-    "${TEMPDIR}/ve/bin/pip3" install --upgrade py-algorand-sdk cryptography
+    "${TEMPDIR}/ve/bin/pip3" install --upgrade cryptograpy
+    # get the latest tip of our python SDK main branch as opposed to what's on pypi:
```
tzaffi, line 127: I'm arguing that for testing in `go-algorand`, it's better to rely on the tip of our github than our latest official PyPi release. But honestly, in order to pull off my new `AtomicABI` Python test class, I needed this.

daemon/algod/api/server/v2/handlers.go
tsachiherman: it doesn't look that you really need the block. You only need the block header. try using `myLedger.BlockHdr`

test/scripts/e2e.sh
```diff
@@ -123,7 +123,11 @@ if [ -z "$E2E_TEST_FILTER" ] || [ "$E2E_TEST_FILTER" == "SCRIPTS" ]; then
...
+    "${TEMPDIR}/ve/bin/pip3" install --upgrade git+git://github.com/algorand/py-algorand-sdk@master
+    # enable remote debugging:
```
tzaffi, line 129: Doesn't affect Circle tests. But of course, user beware. Make sure not to turn on the debugger when sending to Circle (...there might be a clever way to debug remotely to circle, but I don't recommend that)

daemon/algod/api/algod.oas2.json
tsachiherman: The min-balance isn't just for transacting. I'd phrase it as :
```
total number of MicroAlgos required for maintaining the account
```

test/scripts/e2e_subs/min_balance.py
```diff
@@ -0,0 +1,184 @@
...
+    return goal.algod.account_info(account)["min-balance"]
+
+
+def get_teal_min_balance(abi, account):
+    # can't execute an abi object twice so must clone it before each execution:
+    abi = abi.clone(caller_acct=account)
+    sender_pymt = txn.PaymentTxn(account, abi.get_suggested_params(), account, 10_000)
+    sender_tx_sig = abi.get_txn_with_signer(sender_pymt)
+
+    # TODO: abi.sender_min_balance(sender_tx_sig)
```
tzaffi, line 55: Yep, `AtomicABI` is still _alpha_. Hopefully, next week I can officially announce it (#3277 )

test/scripts/e2e_subs/min_balance.py
```diff
@@ -0,0 +1,184 @@
...
+    return abi.execute_singleton("sender_min_balance", method_args=[sender_tx_sig])
+
+
+def assert_min_balance(abi_or_goal, account, expected_min_balance, skip_abi=False):
```
tzaffi, line 59: Query the `min_balance` in two ways, and compare the results and also to what we think they should be. The two ways are:
1. Via **algod REST** / **Python SDK**
2. Via **AtomicABI** to query a TEAL app that uses the `min_balance` op code

test/scripts/e2e_subs/tealprogs/abi-min_balance.teal
```diff
@@ -0,0 +1,110 @@
+#pragma version 6
+    // TODO - some explanation
```
tzaffi, line 2: This TEAL ain't perfect, but it serves the purposes of this PR.

daemon/algod/api/algod.oas2.json
tzaffi: 👍

daemon/algod/api/server/v2/handlers.go
tzaffi: 👍

test/scripts/e2e_subs/tealprogs/abi-min_balance.json
```diff
@@ -0,0 +1,69 @@
+{
```
tzaffi, line 1: `AtomicABI` needs the ABI contract in order to dynamically create the caller metnods.

tzaffi:
> I'm not opposing to adding the min balance to the rest api endpoint, since it's a (almost) free operation. However, I'm not a big fan of the python testing - and not because of the usage of python itself. It's the circular dependency between the go-algorand and the python sdk repo that I'd like to see resolved.

You bring up a good point. I'm not introducing this dependency, it predates this PR. But I am certainly reinforcing it. I've also included a very simple pure shell script test for the new functionality. 

I might have stepped into a bit of a minefield here. My opinion is that the ability to write deeper tests more quickly trumps the circular dependency issue. If we ever run into the situation that PySDK makes a change that breaks our tests, we can always pin the pre-existing tests to a particular git commit and make them pass. That's a bit cumbersome, but not very hard to do.

daemon/algod/api/server/v2/handlers.go
```diff
@@ -270,7 +278,9 @@ func (v2 *Handlers) AccountInformation(ctx echo.Context, address string, params
 		}
 	}
 
-	account, err := AccountDataToAccount(address, &record, assetsCreators, lastRound, amountWithoutPendingRewards)
+	minBalance := record.MinBalance(&consensus)
```
algorandskiy, line 281: no need to calculate minBalance here. record of the basics.AccountData is passed to `AccountDataToAccount` so it can call `MinBalance` there ? I'd better pass consensus params there since it also needed for some other calculated fields if we decide to add them later.

test/scripts/e2e_subs/goal/atomic_abi.py
algorandskiy: Is this test harness really needed for this min-balance in REST PR?

tzaffi:
@algorandskiy - regarding your comments:
* I will investigate your and @tsachiherman suggestion about how to better acquire min balance
* Is `AtomicABI.py` _really needed_? In principle no. But to run the test in the way I wanted - compare the result of the modified `algod` with the pre-existing way we would have needed to get it from the outside (via a TEAL program), this is useful. Also, I am planning on officially introducing `AtomicABI.py` in #3277. Admittedly this is a topic of controversy. Let's continue discussing here and see if a begrudging consensus is reached.

daemon/algod/api/server/v2/handlers.go
tzaffi: Will try and incorporate your suggestions. Stay tuned.

daemon/algod/api/algod.oas2.json
winder: The square brackets aren't needed for this field, the value in the brackets is the msgpack encoded name.
```suggestion
          "description": "MicroAlgo balance required by the account, the requirement starts at 1000 and grows based on the asset and application usage",
```

test/scripts/e2e_subs/tealprogs/abi-min_balance.teal
algonautshant, line 2: please please please add the explanation ...

daemon/algod/api/algod.oas3.yml
```diff
@@ -726,6 +726,10 @@
             },
             "type": "array"
           },
+          "min-balance": {
+            "description": "\\[algo\\] total number of MicroAlgos needed to allow transacting from account",
```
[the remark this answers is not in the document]
tzaffi: this should be automatic

daemon/algod/api/server/v2/generated/private/types.go
```diff
@@ -49,6 +49,9 @@ type Account struct {
 	// Note: the raw account uses `map[int] -> Asset` for this type.
 	CreatedAssets *[]Asset `json:"created-assets,omitempty"`
 
+	// \[algo\] total number of MicroAlgos needed to allow transacting from account
```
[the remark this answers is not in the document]
tzaffi: automatic

daemon/algod/api/server/v2/generated/types.go
```diff
@@ -49,6 +49,9 @@ type Account struct {
 	// Note: the raw account uses `map[int] -> Asset` for this type.
 	CreatedAssets *[]Asset `json:"created-assets,omitempty"`
 
+	// \[algo\] total number of MicroAlgos needed to allow transacting from account
```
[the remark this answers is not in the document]
tzaffi: go generated...

test/scripts/e2e_subs/tealprogs/abi-min_balance.teal
tzaffi: 👍

test/e2e-go/cli/goal/expect/README.md
```diff
@@ -77,6 +77,15 @@ go test -v test/e2e-go/cli/goal/expect/goal_expect_test.go
 
 TESTFILTER=foobar[0-9]Test TESTDIR=baz go test -v test/e2e-go/cli/goal/expect/goal_expect_test.go
 ```
+In particular, if you want to run `goalAccountInfoTest.exp` you'll need to create a temporary directory such as `deletemeTemp` with
```
tzaffi, line 80: Adding more explicit instructions for running `*.exp` tests

test/scripts/e2e.sh
tzaffi: @jasonpaulos  @tsachiherman - pinning the python SDK version

test/scripts/e2e_subs/tealprogs/abi-min_balance.teal
tzaffi: Brief explanation added

daemon/algod/api/server/v2/handlers.go
```diff
@@ -239,6 +239,16 @@ func (v2 *Handlers) AccountInformation(ctx echo.Context, address string, params
...
 	}
 
+	latestBlkHdr, err := myLedger.BlockHdr(lastRound)
+	if err != nil {
+		return internalError(ctx, err, errFailedRetrievingLatestBlockHeaderStatus, v2.Log)
+	}
+
+	consensus, ok := config.Consensus[latestBlkHdr.CurrentProtocol]
+	if !ok {
+		return notFound(ctx, errors.New(errInternalFailure), "could not retrieve consensus information for current protocol", v2.Log)
```
algorandskiy, line 249: it is `internalError` rather than `notFound` since itis result does not depends on client's input

test/scripts/e2e.sh
algorandskiy: Not sure about pinning its version, the latest and greatest looks reasonable default. Otherwise do we potentially need to update it with each pysdk release?

test/scripts/e2e_subs/min_balance.py
```diff
@@ -0,0 +1,185 @@
...
+
+if INTERACTIVE:
+    initialize_debugger()
+
+
+def get_pysdk_min_balance(goal, account):
+    return goal.algod.account_info(account)["min-balance"]
+
+
+def get_teal_min_balance(abi, account):
```
algorandskiy, line 48: not sure I understand why do we need all this python abi + teal machinery to test min balance field in REST

algorandskiy (commented):
`MinBalance` piece in TEST looks good except `notFound` error on missed consensus.
`goal account` looks good.

> Python SDK's account_info function returns min-balance as well

I'm not against having ABI helpers for e2e tests. I only do not understand why do we test sdk in go-algorand. I think it should be another way around

daemon/algod/api/server/v2/handlers.go
tzaffi: 👍

tzaffi:
> `MinBalance` piece in TEST looks good except `notFound` error on missed consensus. `goal account` looks good.
> 
> > Python SDK's account_info function returns min-balance as well
> 
> I'm not against having ABI helpers for e2e tests. I only do not understand why do we test sdk in go-algorand. I think it should be another way around

I did not intend to test the Python SDK in this PR. However, that comes for free because I'm using the Python Goal wrapper which internally calls the Python SDK. So I am pointing this fact out so that no extra work needs to be done for the Python SDK.

You make a good point that we shouldn't be testing SDK's in go-algorand. So there will need to be appropriate tests added for **min-balance** in `algorand-sdk-testing` and each of the SDK's will need to pass those tests. The fact that the Python SDK should be good to go, doesn't excuse it from passing those tests.

I will be addressing your point regarding the error type shortly.

algorandskiy (dismissed):
The `MinBalance` piece looks good.
python sdk pinning and testing harness - defer to others.

daemon/algod/api/algod.oas2.json
```diff
@@ -1365,6 +1366,10 @@
           "description": "\\[algo\\] total number of MicroAlgos in the account",
           "type": "integer"
         },
+        "min-balance": {
+          "description": "MicroAlgo balance required by the account.\n\nThe requirement starts at 100,000 and grows based on asset and application usage.",
```
tsachiherman, line 1370: I think it would be better to drop off the numeric constants from the description. These values are consensus protocol dependent and can be modified in the future.

daemon/algod/api/server/v2/account_test.go
```diff
@@ -102,7 +102,7 @@ func TestAccount(t *testing.T) {
 	b := a.WithUpdatedRewards(proto, 100)
 
 	addr := basics.Address{}.String()
-	conv, err := AccountDataToAccount(addr, &b, map[basics.AssetIndex]string{}, round, a.MicroAlgos)
+	conv, err := AccountDataToAccount(addr, &b, map[basics.AssetIndex]string{}, round, &config.ConsensusParams{MinBalance: 100000}, a.MicroAlgos)
```
tsachiherman, line 105: The minimum balance depends of a multiple fields from `config.ConsensusParams`. It would be wrong to initialize just a single field ( `MinBalance` ), since the function `record.MinBalance` might change over time.
I would suggest you'll grab a specific consensus protocol and use it. i.e. the current protocol sounds like a safe choice here.

daemon/algod/api/server/v2/account_test.go
```diff
@@ -196,7 +196,7 @@ func TestAccount(t *testing.T) {
 		// convert the same account a few more times to make sure we always
 		// produce the same generated.Account
 		for i := 0; i < 10; i++ {
-			anotherConv, err := AccountDataToAccount(addr, &b, map[basics.AssetIndex]string{}, round, a.MicroAlgos)
+			anotherConv, err := AccountDataToAccount(addr, &b, map[basics.AssetIndex]string{}, round, &config.ConsensusParams{MinBalance: 100000}, a.MicroAlgos)
```
tsachiherman, line 199: same as above.

test/scripts/e2e_subs/min_balance.py
tsachiherman: I agree with @algorandskiy about that one. This python testing should go to the python repository.
Testing the REST endpoint of the algod can easily be done using either expect or one of the existing e2e subs.
( you've already implemented min_balance.sh, which is great )

test/scripts/e2e_subs/tealprogs/abi-min_balance.json
```diff
@@ -0,0 +1,30 @@
...
+                {
+                    "type": "byte[]"
+                }
+            ],
+            "returns": {
+                "type": "byte[]"
+            }
+        }
+    ]
+}
```
tsachiherman, line 30: nit: add EOL

test/scripts/e2e_subs/tealprogs/abi-min_balance.json
```diff
@@ -0,0 +1,30 @@
+{
+    "name": "opCodes",
+    "appId": null,
```
ahangsu, line 3: nit:  😅 we are changing this field `"appId"` to `network` in the new ABI spec: https://github.com/algorandfoundation/ARCs/pull/53, it seems totally optional to have this field or not, might be okay to remove that

test/scripts/e2e_subs/tealprogs/abi-min_balance.json
```diff
@@ -0,0 +1,30 @@
...
+            "desc": "The complement (aka b~) of arg0 viewed as a bitstring",
+            "args": [
+                {
+                    "type": "byte[]"
+                }
+            ],
+            "returns": {
+                "type": "byte[]"
+            }
+        }
```
ahangsu, line 28: nit: seems not essential in this test, can remove in contract file and teal code

daemon/algod/api/server/v2/account_test.go
tzaffi: 👍

daemon/algod/api/algod.oas2.json
tzaffi: 👍

daemon/algod/api/server/v2/account_test.go
tzaffi: 👍

test/scripts/e2e_subs/tealprogs/abi-min_balance.json
tzaffi: 👍

test/scripts/e2e_subs/tealprogs/abi-min_balance.json
tzaffi: 👍

test/scripts/e2e_subs/tealprogs/abi-min_balance.json
tzaffi: 👍

daemon/algod/api/server/v2/account_test.go
tzaffi: 👍

test/scripts/e2e_subs/tealprogs/abi-min_balance.json
tzaffi: 👍

test/scripts/e2e_subs/goal/goal.py
jannotti: I don't quite get it.  The info was already available as `goal.algod.algod_address` now it's reformatted so you can say `val = goal.get_endpoint_info(); val["algod"]["url"]` ?

test/scripts/e2e_subs/min_balance.py
jannotti: I'm in favor of using python instead of an unholy combination or shell/goal/jq/curl/grep etc, so tests are clearer.  But I think the ABI stuff is overboard here. A very, very simple teal app that returns min_balance could be instantiated and called using just the sdk basics that were already here.

As a general rule, if we just use python as a more convenient way to write good tests, and stay away from testing the most recent features (ABI in this case) we will get better tests, but we can avoid difficulties with pinning the version.  Just use old tried and true SDK functions.

test/scripts/e2e_subs/min_balance.py
tzaffi: @algonautshant and @tsachiherman  - please have a look again. I've implemented the decision points from yesterday's meeting:

1. Remove dependence on Python SDK's ABI interaction implementation, and in particular, removed the `AtomicABI` test class
2. Keeping the python tests, but refactored to avoid ABI
3. Pinned the Python SDK and added a comment of explanation

algorandskiy (commented):
LGTM, please implement @tsachiherman 's suggestions

test/scripts/e2e_subs/goal/goal.py
tzaffi: 👍

tzaffi: Just making it a bit more straightforward to get all the info, but I'll revert.

test/scripts/e2e_subs/min_balance.py
tzaffi: 👍

test/scripts/e2e_subs/min_balance.py
```diff
@@ -0,0 +1,241 @@
...
+OPTIN_MIN_BALANCE = 100_000
+# app schemas:
+APP_KV_MIN_BALANCE = 25_000
+APP_INTS_MIN_BALANCE = 3_500
+APP_BYTES_MIN_BALANCE = 25_000
+
+# per userBalance.go::MinBalance() as of Dec 2021:
+EXTRA_PAGE_MIN_BALANCE = APP_MIN_BALANCE
+
+TEAL = f"""#pragma version 6
```
tzaffi, line 44: @jannotti - simpler TEAL program not using ABI

daemon/algod/api/server/v2/account_test.go
```diff
@@ -102,7 +102,7 @@ func TestAccount(t *testing.T) {
 	b := a.WithUpdatedRewards(proto, 100)
 
 	addr := basics.Address{}.String()
-	conv, err := AccountDataToAccount(addr, &b, map[basics.AssetIndex]string{}, round, a.MicroAlgos)
+	conv, err := AccountDataToAccount(addr, &b, map[basics.AssetIndex]string{}, round, &proto, a.MicroAlgos)
```
tzaffi, line 105: @tsachiherman : `proto` is **ConsensusFuture**. That seems like a reasonable option.

algorandskiy (dismissed):

test/e2e-go/cli/goal/expect/README.md
tsachiherman: interesting.. I was always using the following syntax:
```sh
cd test/e2e-go/cli/goal/expect
go test -v . --run TestGoalWithExpect/goalCmdFlagsTest.exp
```

tsachiherman (dismissed):

barnjamin (dismissed):

daemon/algod/api/server/v2/handlers.go
```diff
@@ -239,6 +239,16 @@ func (v2 *Handlers) AccountInformation(ctx echo.Context, address string, params
...
 
+	latestBlkHdr, err := myLedger.BlockHdr(lastRound)
+	if err != nil {
+		return internalError(ctx, err, errFailedRetrievingLatestBlockHeaderStatus, v2.Log)
+	}
+
+	consensus, ok := config.Consensus[latestBlkHdr.CurrentProtocol]
+	if !ok {
+		return internalError(ctx, errors.New(errInternalFailure), "could not retrieve consensus information for current protocol", v2.Log)
+	}
```
jannotti, line 250: Nicer to use `func (l *Ledger) ConsensusParams(r basics.Round) (config.ConsensusParams, error)`

test/scripts/e2e_subs/min_balance.py
jannotti: I'd use version 5 here.  We're probably going to release this before teal 6 is available.  Nothing will break, since this is only a test, but it seems weird to test it with a v6 teal script even though it's a feature that comes out before v6.

daemon/algod/api/server/v2/handlers.go
tzaffi: 👍

test/scripts/e2e_subs/min_balance.py
```diff
@@ -0,0 +1,238 @@
...
+flo = goal.new_account()
+
+print(f"Joe & Flo: {joe}, {flo}")
+
+txinfo, err = goal.pay(goal.account, joe, amt=50_000_000, send=True)
+txinfo, err = goal.pay(goal.account, flo, amt=100_000_000, send=True)
+
+
+# starting out, should be at global min
+expected_goal_mb = CONSENSUS_MIN_BALANCE + APP_MIN_BALANCE
```
tzaffi, line 143:
```suggestion
expected_goal_mb = CONSENSUS_MIN_BALANCE + APP_MIN_BALANCE

# starting out, should be at global min
```

test/scripts/e2e_subs/min_balance.py
```diff
@@ -0,0 +1,238 @@
...
+
+joe = goal.new_account()
+flo = goal.new_account()
+
+print(f"Joe & Flo: {joe}, {flo}")
+
+txinfo, err = goal.pay(goal.account, joe, amt=50_000_000, send=True)
+txinfo, err = goal.pay(goal.account, flo, amt=100_000_000, send=True)
+
+
```
tzaffi, line 141:
```suggestion
```

jannotti (approved):
Looks good.  I'll wait for the CI checks to finish, then merge.

test/e2e-go/cli/goal/expect/README.md
tzaffi: ohhh... I only saw this now. Will try to get this into the README next PR.

PR #3291, opened 2021-12-08
Compatibility mode for partkeyinfo.

winder:
## Summary

Compatibility for `partkeyinfo` was also needed by some users. In addition to the different format, the old command also allows printing key information when the node is not running

Workarounds:
1) use an older `goal` binary.
2) use `algokey part info --keyfile <file>`

## Test Plan

Tested manually:
```
~$ goal account partkeyinfo -d /tmp/private_network/Node/
Dumping participation key info from /tmp/private_network/Node/...

Participation ID:          CPLHRU3WEY3PE7XTPPSIE7BGJYWAIFPS7DL3HZNC4OKQRQ5YAYUA
Parent address:            DGS6VNX2BRMKGKVAS2LTREMYG33TOCYPFLPCQ3DUTJULQU6P6S7KJCDNTU
Last vote round:           1
Last block proposal round: 2
Effective first round:     1
Effective last round:      3000000
First round:               0
Last round:                3000000
Key dilution:              10000
Selection key:             5QRrTgzSUTqqym43QVsBus1/AOwGR5zE+I7FGwA14vQ=
Voting key:                PK0NMyZ4BKSjPQ9JuT7dQBLdTpjLQv2txuDYDKhkuqs=

~$ goal account partkeyinfo -d /tmp/private_network/Node/ -c
Dumping participation key info from /tmp/private_network/Node/...
------------------------------------------------------------------
File: Wallet2.0.3000000.partkey
{
  "acct": "DGS6VNX2BRMKGKVAS2LTREMYG33TOCYPFLPCQ3DUTJULQU6P6S7KJCDNTU",
  "last": 3000000,
  "sel": "5QRrTgzSUTqqym43QVsBus1/AOwGR5zE+I7FGwA14vQ=",
  "vote": "PK0NMyZ4BKSjPQ9JuT7dQBLdTpjLQv2txuDYDKhkuqs=",
  "voteKD": 10000
}
```

### DIFF
--- a/cmd/goal/account.go
+++ b/cmd/goal/account.go
@@ -1368,23 +1368,21 @@ func strOrNA(value *uint64) string {
 	return uintToStr(*value)
 }
 
-// DEPRECATED
-type partkeyInfo struct {
-	_struct         struct{}                        `codec:",omitempty,omitemptyarray"`
-	Address         string                          `codec:"acct"`
-	FirstValid      basics.Round                    `codec:"first"`
-	LastValid       basics.Round                    `codec:"last"`
-	VoteID          crypto.OneTimeSignatureVerifier `codec:"vote"`
-	SelectionID     crypto.VRFVerifier              `codec:"sel"`
-	VoteKeyDilution uint64                          `codec:"voteKD"`
-}
-
-
 // legacyPartkeyInfoCommand prints key information in the same
 // format as earlier versions of goal. Some users are using this information
 // in scripts and need some extra time to migrate to alternatives.
 // DEPRECATED
 func legacyPartkeyInfoCommand() {
+	type partkeyInfo struct {
+		_struct         struct{}                        `codec:",omitempty,omitemptyarray"`
+		Address         string                          `codec:"acct"`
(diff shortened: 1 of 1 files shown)


cmd/goal/account.go
```diff
@@ -1364,12 +1368,59 @@ func strOrNA(value *uint64) string {
 	return uintToStr(*value)
 }
 
+// DEPRECATED
+type partkeyInfo struct {
```
tsachiherman, line 1372: given that this data structure is used only within `legacyPartkeyInfoCommand`, I would place it within the function as anonymous struct. https://go.dev/play/p/rI92HaX4_SZ

winder: Thanks, I was thinking of trying to reduce the scope but wasn't sure if it was too clever. The legacy function body is copy/pasted, and I want to preserve that. What do you think about defining the type inside the function?

tsachiherman: that's perfectly fine.

tsachiherman (approved):

PR #3292, opened 2021-12-08
ledger refactoring: Ordered resource deltas - part 1

algorandskiy:
## Summary
MergeInMatchingAccounts requires addresses per resource but storing addresses there makes the code much slower.
Lets see if MergeInMatchingAccounts can be eliminated altogether on the schema switch.

## Test Plan

Added new test for triggering missed base account data in StateDelta for some application resources modification.

### DIFF
--- a/ledger/ledgercore/statedelta.go
+++ b/ledger/ledgercore/statedelta.go
@@ -122,6 +122,30 @@ type newBalanceRecord struct {
 	AccountData
 }
 
+type appParamsRecord struct {
+	aidx   basics.AppIndex
+	addr   basics.Address
+	params *basics.AppParams
+}
+
+type assetParamsRecord struct {
+	aidx   basics.AssetIndex
+	addr   basics.Address
+	params *basics.AssetParams
+}
+
+type appLocalStateRecord struct {
+	aidx  basics.AppIndex
+	addr  basics.Address
+	state *basics.AppLocalState
+}
+
+type assetHoldingRecord struct {
+	aidx    basics.AssetIndex
+	addr    basics.Address
+	holding *basics.AssetHolding
+}
+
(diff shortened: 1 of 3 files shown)


ledger/ledgercore/totals_test.go
```diff
@@ -206,7 +206,7 @@ var uniqueAccountTotals = []AccountTotals{
 func TestAccountTotalsMarshalMsgUnique(t *testing.T) {
 	partitiontest.PartitionTest(t)
 
-	uniqueAt := make(map[crypto.Digest]bool, 0)
+	uniqueAt := make(map[crypto.Digest]bool)
```
tsachiherman, line 209: If you're already chamgimg this is, why not to size the map to len(uniqueAccountTotals) ?

ledger/ledgercore/statedelta.go
```diff
@@ -234,23 +285,28 @@ func (ad NewAccountDeltas) ModifiedAccounts() []basics.Address {
 }
 
 // MergeAccounts applies other accounts into this StateDelta accounts
+// Used in tests only
```
tsachiherman, line 288: If it's used for testing only within the same package, consider moving it to a _test.go file.
That would also exclude it from code coverage.

tsachiherman (changes requested):
Looks good - two tiny requests. I suppose it will render bigger changes to my PR before merging.

ledger/ledgercore/totals_test.go
algorandskiy: done

ledger/ledgercore/statedelta.go
algorandskiy: This is a mistake, it actually used in cow.go. Removed

tsachiherman (approved):

PR #3298, opened 2021-12-09
Add a hash prefix for ARCs-related hashes

fabrice102:
## Summary

This is to allow ARCs (github.com/algorandfoundation/ARCs)
to have their own hash prefix without risk of collision.

## Test Plan

It is purely informational. There is no real code change.


tsachiherman (approved):
Looks good to me

PR #3299, opened 2021-12-10
Fix: stop the completed catchup service

algonautshant:
The catchup service stops when it is complete, i.e. it has reached up to the round which is being agreed on.

The catchup service knows it is complete and should stop, when it finds that a block is in the ledger before it adds it.
In other words, apart from the catchup, only the agreement adds blocks to the ledger. And when the agreement adds a block to the ledger before the catchup, it means the agreement is ahead, and the catchup is complete.

When `fetchAndWrite` detects the block is already in the ledger, it returns. The return value of `false` stops the catchup syncing. 

In previous releases, `fetchAndWrite` was only checking if the block is already in the ledger after attempting to fetch it. 
Since it fails to fetch a block not yet agreed on, the fetch fails after multiple attempts, and `fetchAndWrite` returns `false` ending the catchup.

A recent change made this process more efficient by first checking if the block is in the ledger before/during the fetch.
However, once the block was found in the ledger, `fetchAndWrite` returned true instead of false (consistent with already existing logic since forever, which was also wrong). This caused the catchup to continue syncing after catchup was complete.
This change fixes the return value from true to false.

### DIFF
--- a/catchup/service.go
+++ b/catchup/service.go
@@ -241,8 +241,10 @@ func (s *Service) fetchAndWrite(r basics.Round, prevFetchCompleteChan chan bool,
 
 		if err != nil {
 			if err == errLedgerAlreadyHasBlock {
-				// ledger already has the block, no need to request this block from anyone.
-				return true
+				// ledger already has the block, no need to request this block.
+				// only the agreement could have added this block into the ledger, catchup is complete
+				s.log.Infof("fetchAndWrite(%d): the block is already in the ledger. The catchup is complete", r)
+				return false
 			}
 			s.log.Debugf("fetchAndWrite(%v): Could not fetch: %v (attempt %d)", r, err, i)
 			peerSelector.rankPeer(psp, peerRankDownloadFailed)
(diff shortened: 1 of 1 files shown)


tsachiherman (approved):
looks good. thanks for finding & fixing this on such a short notice.

barnjamin:
To help make this more clear the comment here should probably enumerate the conditions for when this returns false, namely "if the block is already in the ledger"
https://github.com/algonautshant/go-algorand/blob/59e3d84660047d13a0fb77b105fa92d7f1037081/catchup/service.go#L198

Alternatively, should we consider returning an error instead of a bool since the only caller is the pipelineCallback and it can handle error checking logic for signaling if sync has finished?

tsachiherman:
> To help make this more clear the comment here should probably enumerate the conditions for when this returns false, namely "if the block is already in the ledger" https://github.com/algonautshant/go-algorand/blob/59e3d84660047d13a0fb77b105fa92d7f1037081/catchup/service.go#L198
> 
> Alternatively, should we consider returning an error instead of a bool since the only caller is the pipelineCallback and it can handle error checking logic for signaling if sync has finished?

Enumerating the exit cases on the method descriptions sounds like a good idea. Do you want to PR that?

PR #3303, opened 2021-12-10
blocks service: avoid serving blocks past service shutdown

tsachiherman:
## Summary

The block service was attempting to serve block via the http handler even after it has been stopped.
This lead to undesired downstream failures in the ledger, which was shutdown as well.

## Test Plan

unit test added.

### DIFF
--- a/rpcs/blockService.go
+++ b/rpcs/blockService.go
@@ -19,6 +19,7 @@ package rpcs
 import (
 	"context"
 	"encoding/binary"
+	"errors"
 	"net/http"
 	"path"
 	"strconv"
@@ -28,6 +29,8 @@ import (
 	"github.com/gorilla/mux"
 
 	"github.com/algorand/go-codec/codec"
+
+	"github.com/algorand/go-deadlock"
 
 	"github.com/algorand/go-algorand/agreement"
 	"github.com/algorand/go-algorand/config"
@@ -61,6 +64,8 @@ const (
 	BlockAndCertValue  = "blockAndCert"    // block+cert request data (as the value of requestDataTypeKey)
 )
 
+var errBlockServiceClosed = errors.New("block service is shutting down")
+
 // BlockService represents the Block RPC API
 type BlockService struct {
 	ledger                  *data.Ledger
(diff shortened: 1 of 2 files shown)


rpcs/blockService.go
```diff
@@ -356,6 +366,22 @@ func (bs *BlockService) getRandomArchiver() (endpointAddress string) {
...
+	defer bs.mu.Unlock()
+	select {
+	case _, ok := <-bs.stop:
+		if !ok {
+			// service is closed.
+			return nil, errBlockServiceClosed
+		}
+	default:
+	}
+	return RawBlockBytes(bs.ledger, round)
```
algorandskiy, line 382: wonder why is `RawBlockBytes` a part of block service since it has nothing to do with bs... but this is a separate topic

rpcs/blockService_test.go
```diff
@@ -170,6 +173,58 @@ func TestRedirectFallbackArchiver(t *testing.T) {
...
+	request = request.WithContext(requestCtx)
+	network.SetUserAgentHeader(request.Header)
+
+	requestDone := make(chan struct{})
+	go func() {
+		defer close(requestDone)
+		client.Do(request)
+	}()
+
+	bs1.Stop()
```
algorandskiy, line 224: maybe stop first and then start a goroutine?

algorandskiy (commented):
Looks good, one question for the test

rpcs/blockService_test.go
tsachiherman: The intent here is to create the "contention" so that the request would be made before/while/after the `Stop`. i.e. this is where the bug was, so I was trying to reproduce it. Moving the `Stop` before the go routine eliminate some of that randomness..

algorandskiy (approved):

rpcs/blockService.go
```diff
@@ -133,12 +141,14 @@ func (bs *BlockService) Start() {
...
 func (bs *BlockService) Stop() {
+	bs.mu.Lock()
 	close(bs.stop)
+	bs.mu.Unlock()
 	bs.closeWaitGroup.Wait()
 }
 
 // ServerHTTP returns blocks
-// Either /v{version}/block/{round} or ?b={round}&v={version}
+// Either /v{version}/{genesisID}/block/{round} or ?b={round}&v={version}
```
algonautshant, line 151: Shouldn't genesisID also be in ?b={round}&v={version}&g={genesisID} ?

PR #3306, opened 2021-12-12
Test for catchup stop on completion

algonautshant:
Adding a test for the fix in #3299

## Test Plan

This is a test

### DIFF
--- a/test/e2e-go/features/catchup/basicCatchup_test.go
+++ b/test/e2e-go/features/catchup/basicCatchup_test.go
@@ -17,6 +17,7 @@
 package catchup
 
 import (
+	"fmt"
 	"os"
 	"path/filepath"
 	"testing"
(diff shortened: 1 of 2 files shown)


barnjamin (approved):

test/e2e-go/features/catchup/basicCatchup_test.go
```diff
@@ -331,3 +331,92 @@ func shutdownClonedNode(nodeDataDir string, f *fixtures.RestClientFixture, t *te
...
+	a.NoError(err)
+
+	// Start the third node to catchup.
+	startTime := time.Now()
+	cloneClient, err := fixture.StartNode(cloneDataDir)
+	a.NoError(err)
+	defer shutdownClonedNode(cloneDataDir, &fixture, t)
+
+	// Wait for it to catchup
+	err = fixture.LibGoalFixture.ClientWaitForRoundWithTimeout(cloneClient, waitForRound)
```
tsachiherman, line 386: this isn't accurate enough; you want to wait until the round number from both nodes is the same.
i.e. while you're waiting for the node to catchup, the network is keep making progress. you want to make sure that the round number aligns.

test/e2e-go/features/catchup/basicCatchup_test.go
```diff
@@ -331,3 +331,92 @@ func shutdownClonedNode(nodeDataDir string, f *fixtures.RestClientFixture, t *te
...
+	//   will be at least our estimated catchupTime (since it keeps on growing if catchup has not stopped).
+	time.Sleep(catchupTime)
+
+	// Prevent false negative
+	// The network may have made some progress since waitForRound, it could be that the
+	// third node is still catching up even after getting to waitForRound.
+	// Moreover, it takes some time to transition from the catchup to agreement.
+	// Give it some more time and check again..
+	pass := false
+	for x := 0; x < 100; x++ {
```
tsachiherman, line 410: I'd suggest trying to figure out here when the agreement service kicks in. Once it kicks in, the `curStatus.CatchupTime` would go down. this is your signal to exit the test.

algonautshant: I do not know how I can figure out when the agreement service kicks in. Even so, there may be some time after the agreement service kicks in, and the catchup stops. 

But once the catchup stops, curStatus.CatchupTime is less than the catchup time, and that is considered the signal to exit the test.

tsachiherman (approved):

PR #3310, opened 2021-12-13
Exclude potentially long catchpoint generation from trackerRegistry lock

algorandskiy:
## Summary

* commitRound is designed to be async op triggered by commitSyncer but tracker ops must be synchronized.
* catchpoint tracker's postCommit might take too long but in fact does not need to be in sync with other trackers

## Test Plan

Use existing tests


algorandskiy:
Closed in pref of #3311

PR #3313, opened 2021-12-14
Add help and fish mode to e2e interactive mode.

winder:
## Summary

Minor improvements to e2e.sh interactive mode:
* add to -h output
* do not run start stop test in interactive mode
* support fish shell

## Test Plan

Manual testing:
```
~$ ./e2e.sh -i
...
lots of output removed
...
********** READY **********


The test environment is now set. You can now run tests in another terminal.

Configure the environment:

set -g VIRTUAL_ENV "/home/will/go/src/github.com/algorand/go-algorand/tmp/out/e2e/130013-1639576513257/ve"
set -g PATH "$VIRTUAL_ENV/bin:$PATH"

python3 "/home/will/go/src/github.com/algorand/go-algorand/test/scripts"/e2e_client_runner.py  "/home/will/go/src/github.com/algorand/go-algorand/test/scripts"/e2e_subs/SCRIPT_FILE_NAME

Press enter to shut down the test environment...
```

### DIFF
--- a/test/scripts/e2e.sh
+++ b/test/scripts/e2e.sh
@@ -138,7 +138,7 @@ if [ -z "$E2E_TEST_FILTER" ] || [ "$E2E_TEST_FILTER" == "SCRIPTS" ]; then
         fi
 
         echo ""
-        echo "python3 \"$SCRIPT_PATH\"/e2e_client_runner.py ${RUN_KMD_WITH_UNSAFE_SCRYPT} \"$SCRIPT_PATH\"/e2e_subs/SCRIPT_FILE_NAME"
+        echo "python3 \"$SCRIPT_PATH/e2e_client_runner.py\" ${RUN_KMD_WITH_UNSAFE_SCRYPT} \"$SCRIPT_PATH/e2e_subs/SCRIPT_FILE_NAME\""
         echo ""
         echo "Press enter to shut down the test environment..."
         read a
--- a/test/scripts/e2e.sh
+++ b/test/scripts/e2e.sh
@@ -22,6 +22,7 @@ Requires:
 Options:
     -c        Channel of build you are building binaries with this script
     -n        Run tests without building binaries (Binaries are expected in PATH)
+    -i        Start an interactive session for running e2e subs.
 "
 NO_BUILD=false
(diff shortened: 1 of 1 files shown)


test/scripts/e2e.sh
```diff
@@ -117,23 +118,27 @@ export GOPATH=$(go env GOPATH)
...
+            echo "set -g VIRTUAL_ENV \"${TEMPDIR}/ve\""
+            echo "set -g PATH \"\$VIRTUAL_ENV/bin:\$PATH\""
+        else
+            echo "export VIRTUAL_ENV=\"${TEMPDIR}/ve\""
+            echo "export PATH=\"\$VIRTUAL_ENV/bin:\$PATH\""
+        fi
+
         echo ""
-        echo "${TEMPDIR}/ve/bin/python3" test/scripts/e2e_client_runner.py ${RUN_KMD_WITH_UNSAFE_SCRYPT} "$SRCROOT"/test/scripts/e2e_subs/SCRIPT_FILE_NAME
+        echo "python3 \"$SCRIPT_PATH\"/e2e_client_runner.py ${RUN_KMD_WITH_UNSAFE_SCRYPT} \"$SCRIPT_PATH\"/e2e_subs/SCRIPT_FILE_NAME"
```
algonautshant, line 141: I am concerned about using python3 form the user environment instead of using python3 from the ve.
I had issues with that.
Is there a downside of using ve/bin/python3 instead of $PATH/python3?

winder: We definitely want to use python3 from the virtual environment. The environment variable should make sure we get the correct version: `$VIRTUAL_ENV/bin:$PATH`

algonautshant: However, when you open a different terminal per the instructions above, the path will not be set there. And therefore, the standard system python3 will be executed, and not the ve/bin/python3.
Python ve once activated, has a funny way of setting the path.

algonautshant: Never mind. What you have is correct. I missed the path part in the guildlines.

algonautshant (approved):
Looks great! 
Thanks for the improvements.

algonautshant (approved):

PR #3318, opened 2021-12-15
txsync: use fixed proposal sizes

nicholasguoalgorand:
## Summary

Revert back to using fixed proposal sizes instead of dynamic proposal sizes for now.

## Test Plan

Reran existing tests


tsachiherman (changes requested):
Instead of passing 0 to `StartEvaluator`, could you add a config parameter 
```golang
EnableDynamicProposalSizeLimit bool `version[20]:"false"`
```

and use that to return 0 from `calculateMaxTxnBytesPerBlock` ?

tsachiherman (approved):
Looks great, thanks for the changes.

tsachiherman (approved):

PR #3321, opened 2021-12-16
instrument the code

algonautshant:


algonautshant:
This is fixed in https://github.com/algorand/go-algorand/pull/3363

PR #3322, opened 2021-12-16
adding pc to line mapping

barnjamin:
## Summary

Addressing https://github.com/algorand/go-algorand/issues/2870

Adding an Interface SourceMap that is implemented by the AssemblyMap struct in the logic package and the session struct in the tealdbg package.

Added a separate flag for compile to produce a JSON file containing the line to pc mapping.

## Test Plan

Added a couple small tests, will add more for less trivial programs and also compare with the ouput from the tealdbg session


barnjamin:
For @algorandskiy, is this appx what you were looking for regarding testing between the two source/pc mapping implementations we have? 

More test programs will need to be added but wanted to get feedback on what is there now

cmd/tealdbg/debugger.go
```diff
@@ -58,9 +59,10 @@ type Control interface {
 	RemoveBreakpoint(line int) error
 	SetBreakpointsActive(active bool)
 
-	GetSourceMap() ([]byte, error)
 	GetSource() (string, []byte)
 	GetStates(s *logic.DebugState) AppState
+
+	logic.SourceMapper
```
algorandskiy, line 65: I like having Mapper interface. But not sure if mixing in a Mapper interface into a Control interface a good idea since looks like no benefits compared to older approach with the `GetSourceMap` member

cmd/tealdbg/util.go
```diff
@@ -118,11 +118,11 @@ func IntToVLQ(v int, buf *bytes.Buffer) {
 }
 
 // MakeSourceMapLine creates source map mapping's line entry
-func MakeSourceMapLine(tcol, sindex, sline, scol int) string {
```
algorandskiy, line 121: this intentionally has a generic interface in order to be used when/if needed

cmd/tealdbg/cdtSession.go
```diff
@@ -66,7 +66,7 @@ func makeCdtSession(uuid string, debugger Control, ch chan Notification) *cdtSes
 }
 
 func (s *cdtSession) sourceMapHandler(w http.ResponseWriter, r *http.Request) {
-	sm, err := s.debugger.GetSourceMap()
+	sm, err := GetSourceMap(s.debugger)
```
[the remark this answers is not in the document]
barnjamin: I supposed it could be but I made the original method take the interface instead of having the session receiver. I _thought_ that was what you had suggested  but may have misunderstood

cmd/tealdbg/util.go
barnjamin: I didn't see anyplace else this was used and the caller was doing some weird stuff to get it in the right format. Where else _might_ this eb used?

cmd/tealdbg/cdtSession.go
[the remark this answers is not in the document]
algorandskiy: I do not remember details, sorry. I probably thought about having a common code generating source maps from different underlying data. But here `session` object has all the data and own code for maps creation.

PR #3323, opened 2021-12-16
ledger refactoring: implement resources writing

algorandskiy:
## Summary

Basic/initial draft implementation of resources writing to the disk

TODO:
1. Consider removing `rtype` and introducing rowid into `persistedResourcesData` 
2. Consider modifying data structures to minimize lookups like for addr to addrid resolution.

## Test Plan

TODO:
1. Fix existing tests
2. Extend tests to check resources

### DIFF
--- a/ledger/accountdb.go
+++ b/ledger/accountdb.go
@@ -213,6 +213,15 @@ type compactResourcesDeltas struct {
 	cache map[accountCreatable]int
 	// misses holds indices of addresses for which old portion of delta needs to be loaded from disk
 	misses []accountCreatable
+
+	// knownAddresses holds address to addrid (accountsbase rowid) mapping in order to resolve
+	// addresses from into addrids when writing resourcesData blob into DB.
+	// It is filled in two stages:
+	// 1) using baseAccounts cache of known accounts
+	// 2) using a set of addresses from account compact deltas after loading old values from DB
+	knownAddresses map[basics.Address]int64
+	// missingAddresses contains still unresolved addresses
+	missingAddresses map[basics.Address]struct{}
 }
 
 type accountDelta struct {
@@ -350,7 +359,7 @@ func prepareNormalizedBalancesV6(bals []encodedBalanceRecordV6, proto config.Con
 // makeCompactResourceDeltas takes an array of NewAccountDeltas ( one array entry per round ), and compacts the resource portions of the arrays into a single
 // data structure that contains all the resources deltas changes. While doing that, the function eliminate any intermediate resources changes.
 // It counts the number of changes each account get modified across the round range by specifying it in the nAcctDeltas field of the resourcesDeltas.
-func makeCompactResourceDeltas(accountDeltas []ledgercore.NewAccountDeltas, baseResources lruResources) (outResourcesDeltas compactResourcesDeltas) {
+func makeCompactResourceDeltas(accountDeltas []ledgercore.NewAccountDeltas, baseAccounts lruAccounts, baseResources lruResources) (outResourcesDeltas compactResourcesDeltas) {
 	if len(accountDeltas) == 0 {
 		return
 	}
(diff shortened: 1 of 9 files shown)


ledger/acctupdates.go
```diff
@@ -1380,6 +1379,29 @@ func (au *accountUpdates) commitRound(ctx context.Context, tx *sql.Tx, dcc *defe
...
+
+	au.accountsMu.RLock()
+	for addr := range addressesNeeded {
+		if _, ok := addressesResolved[addr]; !ok {
+			if pad, has := au.baseAccounts.read(addr); has {
+				addressesResolved[pad.addr] = pad.rowid
+			}
+		}
+	}
+	au.accountsMu.RUnlock()
```
tsachiherman, line 1398: this portion is correct, but need to be moved to the `prepareCommit`, since the lock is already taken there. The compactResourcesDeltas should also contain the resolved addresses so that you won't need to pass it over the dcc.

ledger/accountdb.go
```diff
@@ -486,17 +508,106 @@ func makeCompactResourceDeltas(accountDeltas []ledgercore.NewAccountDeltas, base
...
 					outResourcesDeltas.insert(appParams.Addr, basics.CreatableIndex(appParams.Aidx), newEntry) // insert instead of upsert economizes one map lookup
 				} else {
 					outResourcesDeltas.insertMissing(appParams.Addr, basics.CreatableIndex(appParams.Aidx), newEntry)
+					missingAddresses[appParams.Addr] = struct{}{}
 				}
 			}
 		}
 	}
+
+	outResourcesDeltas.knownAddresses = make(map[basics.Address]int64, len(missingAddresses))
```
tsachiherman, line 527: I'm not sure if having the `knownAddresses` / `missingAddresses` is the best route here; what would happen if we were to try and store the `addrid` on each of the entries in `newEntry.oldResource`, and leave it with zero if we don't know that from the cache ? i.e. in `resourcesLoadOld` we would compare it against zero and attempt to load it as needed. Instead of having the `missingAddresses`, I would add the address to the `resourcesDeltas`

ledger/accountdb.go
```diff
@@ -1896,7 +2017,11 @@ func (qs *accountsDbQueries) lookup(addr basics.Address) (data persistedAccountD
 			data.addr = addr
 			if len(buf) > 0 && rowid.Valid {
 				data.rowid = rowid.Int64
-				return protocol.Decode(buf, &data.accountData)
+				err = protocol.Decode(buf, &data.accountData)
+				if err != nil {
+					panic("")
+				}
```
tsachiherman, line 2023: I'm sure it was a good thing for testing.. but it can go away now, right ?

ledger/accountdb.go
```diff
@@ -2193,6 +2332,111 @@ func accountsNewRound(tx *sql.Tx, updates compactAccountDeltas, creatables map[b
...
+			// new entry, data.oldResource does not have addrid
+			addrid = newAddressesRowIDs[addr]
+			if addrid == 0 {
+				err = fmt.Errorf("cannot resolve address %s (%d), aidx %d, data %v", addr.String(), addrid, aidx, data.newResource)
+				return
+			}
+		}
+		if data.oldResource.data.IsEmpty() {
+			// IsEmpty means we don't have a previous value. Note, can't use oldResource.data.MsgIsZero
+			// because of possibility of empty asset holdings or app local staste after opting in
```
tsachiherman, line 2368: typo : `staste` -> `state`

ledger/accountdb.go
```diff
@@ -2193,6 +2332,111 @@ func accountsNewRound(tx *sql.Tx, updates compactAccountDeltas, creatables map[b
...
+			if data.newResource.MsgIsZero() {
```
tsachiherman, line 2369: Shouldn't we use `data.newResource.IsEmpty()` here as well ?

tsachiherman (commented):
Overall, I think that it looks great. One thing that I would change is to have the address / addrid stored on the resourcesDeltas, since it would also be useful for the subsequent catchpoint tracker changes ( you can take a peek on the other draft PR )

ledger/accountdb.go
algorandskiy: good idea, done

tsachiherman (approved):

PR #3326, opened 2021-12-17
Limit number of simultaneous REST connections

tolikzinovyev:
## Summary

This PR limits the number of simultaneous REST connections we process to prevent the exhaustion of resources and ultimately a crash.

Two limits are introduced: soft and hard. When the soft limit is exceeded, new connections are returned the 429 Too Many Requests http code. When the hard limit is exceeded, new connections are accepted and immediately closed.

Partially resolves https://github.com/algorand/go-algorand-internal/issues/1814.

## Test Plan

Added unit tests.


tsachiherman (commented):
The HTTP connections are typically HTTP 1.1 connections. That makes them long-lasting connections, which won't quickly disconnect.

Limiting the number of concurrent connections to the number of CPU cores could be very problematic here. If the connections were CPU bounds, it would be just fine. But that's rarely the case. Some of the connection might be waiting for a block to arrive before returning.

I'd suggest moving the number of concurrent connection to be a config.json option, and supporting "-1" for disabling the connection limits.

tsachiherman (changes requested):
We need a unit test to confirm the code behaves as designed.
In this case, a unit test seems to be a good fit.

winder:
How are the waiting connections queued by the operating system? It looks like the goroutine would be started and block in aquire while trying to write to the sem channel. This might work, and will prevent the SQLite db from being overloaded, but ideally we would reject requests with an error after the limit is reached.

tsachiherman:
> How are the waiting connections queued by the operating system? It looks like the goroutine would be started and block in aquire while trying to write to the sem channel. This might work, and will prevent the SQLite db from being overloaded, but ideally we would reject requests with an error after the limit is reached.

The incoming connection here would reach the listening port (8080), and would be blocked there until Accept is being called. The other party would not receive a TCP ACK until we Accept the connection. When accepting the connection, a new local port would be allocated ( along with a file descriptor, naturally ), and at that point, we would be able to block the given connection. Since this is a HTTP connection, we could attempt to gracefully return "Server is busy" - but it's abit more tricky here, since we can't "write" on the accepting go-routine, since we don't want to block there.

algorandskiy:
I agree this would work but
 - num cpu does not look as a good limit
 - I'd rather see a limiter that allows X concurrent requests and drops other with HTTP 429 rather accepting all and waiting in SYN SENT state.

Maybe consider using Echo middleware and handle limits on HTTP level rather TCP?

tolikzinovyev:
> The HTTP connections are typically HTTP 1.1 connections. That makes them long-lasting connections, which won't quickly disconnect.
> 
> Limiting the number of concurrent connections to the number of CPU cores could be very problematic here. If the connections were CPU bounds, it would be just fine. But that's rarely the case. Some of the connection might be waiting for a block to arrive before returning.
> 
> I'd suggest moving the number of concurrent connection to be a config.json option, and supporting "-1" for disabling the connection limits.

Makes sense!

tolikzinovyev:
> > How are the waiting connections queued by the operating system? It looks like the goroutine would be started and block in aquire while trying to write to the sem channel. This might work, and will prevent the SQLite db from being overloaded, but ideally we would reject requests with an error after the limit is reached.
> 
> The incoming connection here would reach the listening port (8080), and would be blocked there until Accept is being called. The other party would not receive a TCP ACK until we Accept the connection. When accepting the connection, a new local port would be allocated ( along with a file descriptor, naturally ), and at that point, we would be able to block the given connection. Since this is a HTTP connection, we could attempt to gracefully return "Server is busy" - but it's abit more tricky here, since we can't "write" on the accepting go-routine, since we don't want to block there.

I think what I saw yesterday is that the OS establishes connections even when algod doesn't call `accept()` (netstat showed many ESTABLISHED connections). I can check again if that's important.

tolikzinovyev:
> I agree this would work but
> 
>     * num cpu does not look as a good limit
> 
>     * I'd rather see a limiter that allows X concurrent requests and drops other with HTTP 429 rather accepting all and waiting in SYN SENT state.
> 
> 
> Maybe consider using [Echo middleware](https://echo.labstack.com/middleware/rate-limiter/) and handle limits on HTTP level rather TCP?

I think this echo middleware limits the number of connections per second which is not what we want. In any case, go's http server creates a new goroutine for each accepted connections, so we must limit the number of connections in front of the http server.

algorandskiy:
> go's http server creates a new goroutine for each accepted connections, so we must limit the number of connections in front of the http server.

This is OK since they will be closed almost immediately with 429. I guess we'll run out socket descriptors faster than number of goroutines becomes a problem. That's why Echo simply defers new connection handing to the standard http.Server that creates new goroutine per request (note no one spawns a thread per request in production-ready http servers nowadays).

winder:
> I think this echo middleware limits the number of connections per second which is not what we want. In any case, go's http server creates a new goroutine for each accepted connections, so we must limit the number of connections in front of the http server.

That was my interpretation as well. The "burst" functionality was kind of neat, but also not relevant to the problem.


> > go's http server creates a new goroutine for each accepted connections, so we must limit the number of connections in front of the http server.
> 
> This is OK since they will be closed almost immediately with 429. I guess we'll run out socket descriptors faster than number of goroutines becomes a problem. That's why Echo simply defers new connection handing to the standard http.Server that creates new goroutine per request (note no one spawns a thread per request in production-ready http servers nowadays).

+1, I think that returning errors when the limit is reached should be ok.

tolikzinovyev:
> > go's http server creates a new goroutine for each accepted connections, so we must limit the number of connections in front of the http server.
> 
> This is OK since they will be closed almost immediately with 429. I guess we'll run out socket descriptors faster than number of goroutines becomes a problem. That's why Echo simply defers new connection handing to the standard http.Server that creates new goroutine per request (note no one spawns a thread per request in production-ready http servers nowadays).

You are right that the number of goroutines is limited by the number of sockets. On linux its ~65000. With 4kb of memory per goroutine, that's about 260MB which is fine. I'm ok with it.

tolikzinovyev:
> > > go's http server creates a new goroutine for each accepted connections, so we must limit the number of connections in front of the http server.
> > 
> > 
> > This is OK since they will be closed almost immediately with 429. I guess we'll run out socket descriptors faster than number of goroutines becomes a problem. That's why Echo simply defers new connection handing to the standard http.Server that creates new goroutine per request (note no one spawns a thread per request in production-ready http servers nowadays).
> 
> You are right that the number of goroutines is limited by the number of sockets. On linux its ~65000. With 4kb of memory per goroutine, that's about 260MB which is fine. I'm ok with it.

On the other hand, not limiting the number of connections means other applications will not be able to make connections. That's pretty concerning.

cce:
Agreed with comments here that a number in the thousands or 10s of thousands is a much better default than runtime.NumCPU(), or none at all. But 10K seems like a fine default too

tolikzinovyev:
> > go's http server creates a new goroutine for each accepted connections, so we must limit the number of connections in front of the http server.
> 
> This is OK since they will be closed almost immediately with 429. I guess we'll run out socket descriptors faster than number of goroutines becomes a problem. That's why Echo simply defers new connection handing to the standard http.Server that creates new goroutine per request (note no one spawns a thread per request in production-ready http servers nowadays).

I looked into your suggestion again. The number of file descriptors is basically unlimited. On my laptop `sysctl fs.file-max` returns a number > 10^9. So we do need a limit for the number of connections, at least to prevent memory exhaustion.

algorandskiy:
This is kernel max for the entire system, not a single process, from https://www.kernel.org/doc/Documentation/sysctl/fs.txt
```
file-max & file-nr:

The value in file-max denotes the maximum number of file-
handles that the Linux kernel will allocate. When you get lots
of error messages about running out of file handles, you might
want to increase this limit.
```

tolikzinovyev:
> 



> This is kernel max for the entire system, not a single process, from https://www.kernel.org/doc/Documentation/sysctl/fs.txt
> 
> ```
> file-max & file-nr:
> 
> The value in file-max denotes the maximum number of file-
> handles that the Linux kernel will allocate. When you get lots
> of error messages about running out of file handles, you might
> want to increase this limit.
> ```

Right, there is also a per-process limit controlled by `ulimit -n`. On ubuntu and arch it defaults to 1024. Do you know if other linux/unix distros, osx and windows also have small limits?

daemon/algod/server.go
```diff
@@ -84,6 +88,33 @@ func (s *Server) Initialize(cfg config.Local, phonebookAddresses []string, genes
 	s.log.SetLevel(logging.Level(cfg.BaseLoggerDebugLevel))
 	setupDeadlockLogger()
 
+	// Check some config parameters.
+	if cfg.RestConnectionsSoftLimit > cfg.RestConnectionsHardLimit {
+		return fmt.Errorf(
```
tolikzinovyev, line 93: Tsachi: log a warning, don't crash

daemon/algod/internal/limitlistener/limitListener.go
```diff
@@ -0,0 +1,73 @@
+// TODO: add license header
```
tolikzinovyev, line 1: Tsachi: move file to `network/`

tolikzinovyev:
Ready for review!

config/localTemplate.go
```diff
@@ -74,7 +74,7 @@ type Local struct {
 	CadaverSizeTarget uint64 `version[0]:"1073741824"`
 
 	// IncomingConnectionsLimit specifies the max number of long-lived incoming
-	// connections.  0 means no connections allowed.  -1 is unbounded.
+	// connections. 0 means no connections allowed. Must be non-negative.
```
winder, line 77: It seems strange to say `Must be non-negative` and also have the version 0 default be -1. I understand that you're supposed to define all values in config.json, but what happens if you were to omit this? Would it default to "800"?

network/limitlistener/rejectingLimitListener.go
```diff
@@ -0,0 +1,80 @@
...
+
+func (l *rejectingLimitListener) release() {
+	<-l.sem
+}
+
+func (l *rejectingLimitListener) Accept() (net.Conn, error) {
+	for {
+		select {
+		case <-l.done:
+			return nil, errors.New("Accept() limit listener is closed")
```
winder, line 44: Why do you need the for loop, couldn't you return this error after calling `c.Close()`?

network/limitlistener/rejectingLimitListener.go
```diff
@@ -0,0 +1,80 @@
+// Copyright 2013 The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// RejectingLimitListener is a modification of LimitListener in
```
winder, line 5: Would it work to implement this by wrapping `LimitListener` instead of forking it?

```
type rejectingLimitListener struct {
	listener LimitListener
}

func (l *rejectingLimitListener) Accept() (net.Conn, error) {
	select {
	case conn, err := <-l.listener.Accept()
		return conn, err
	default:
		// If the connection is limited, close it immediately instead of waiting.
		err = c.Close()
		if err != nil {
			return nil, fmt.Errorf("Accept() close err: %w", err)
		}
		return nil, errors.New("Accept() limit listener is closed")
	}
}

func (l *rejectingLimitListener) Close() error {
	return l.listener.Close()
}
```

network/limitlistener/rejectingLimitListener_test.go
```diff
@@ -0,0 +1,167 @@
...
+func TestRejectingLimitListener(t *testing.T) {
+	partitiontest.PartitionTest(t)
+
+	const max = 5
+	attempts := (maxOpenFiles() - max) / 2
+	if attempts > 256 { // maximum length of accept queue is 128 by default
+		attempts = 256
+	}
+
+	l, err := net.Listen("tcp", "127.0.0.1:0")
```
winder, line 35: This looks like it should work, but you could probably implement it with some sort of `nonClosingListener`, similar to your `errorListener` to make it a little clearer and avoid the `time.Sleep`.

network/limitlistener/rejectingLimitListener_test.go
```diff
@@ -0,0 +1,167 @@
...
+	"time"
+
+	"github.com/algorand/go-algorand/network/limitlistener"
+	"github.com/algorand/go-algorand/test/partitiontest"
+)
+
+const defaultMaxOpenFiles = 256
+const timeout = 5 * time.Second
+
+func TestRejectingLimitListener(t *testing.T) {
```
winder, line 26:
```suggestion
func TestRejectingLimitListenerSoftLimit(t *testing.T) {
```

network/limitlistener/rejectingLimitListener_test.go
```diff
@@ -0,0 +1,167 @@
...
+	net.Listener
+}
+
+func (errorListener) Accept() (net.Conn, error) {
+	return nil, errFake
+}
+
+var errFake = errors.New("fake error from errorListener")
+
+func TestRejectingLimitListenerError(t *testing.T) {
```
winder, line 89: I think thats what this error means?
```suggestion
func TestListenerError(t *testing.T) {
```

winder (commented):
The functionality looks good, just had a few nits and suggestions for you to consider.

config/localTemplate.go
tolikzinovyev: Omit `version[0]:"-1"`? Then upgrades from version 0 will not work. I think, this is nothing to worry about.

network/limitlistener/rejectingLimitListener.go
tolikzinovyev: I think it wouldn't work. `LimitListener` will not give new connections above the limit, so I will not be able to close them.

network/limitlistener/rejectingLimitListener.go
tolikzinovyev: We need the for loop to close all new connections if we are above the threshold.

network/limitlistener/rejectingLimitListener_test.go
tolikzinovyev: why soft limit?

network/limitlistener/rejectingLimitListener_test.go
tolikzinovyev: Probably. I copied the tests too and decided not to rewrite them.

network/limitlistener/rejectingLimitListener_test.go
tolikzinovyev: not following?

network/limitlistener/rejectingLimitListener.go
algorandskiy: please elaborate. One `Accept` per upcoming connection, if `l.sem` is full then default branch closes it immediately, and there is no other connections until next iteration in the loop for the same client. What is the point?

daemon/algod/server.go
```diff
@@ -84,6 +88,34 @@ func (s *Server) Initialize(cfg config.Local, phonebookAddresses []string, genes
...
 
+	// Check some config parameters.
+	if cfg.RestConnectionsSoftLimit > cfg.RestConnectionsHardLimit {
+		s.log.Warnf(
+			"RestConnectionsSoftLimit %d exceeds RestConnectionsHardLimit %d",
+			cfg.RestConnectionsSoftLimit, cfg.RestConnectionsHardLimit)
+		cfg.RestConnectionsSoftLimit = cfg.RestConnectionsHardLimit
+	}
+	if cfg.IncomingConnectionsLimit < 0 {
+		return fmt.Errorf(
```
algorandskiy, line 99: this will break algod startup for people who have `IncomingConnectionsLimit=-2` for whatever reason. This value looks allowed before the proposed change.

config/localTemplate.go
tsachiherman: I think that upgrades from version 0 would still work. It's only an issue if someone actively specifying -1, which is unlikely.

daemon/algod/api/server/lib/middlewares/connectionLimiter.go
```diff
@@ -14,10 +14,36 @@
...
+			case sem <- struct{}{}:
+				defer func() {
+					// If we fail to read from `sem`, just continue.
+					select {
+					case <-sem:
+					default:
+					}
+				}()
+				err := next(ctx)
+				return err
```
tsachiherman, line 43: style:
```golang
return next(ctx)
```

daemon/algod/server.go
```diff
@@ -84,6 +88,34 @@ func (s *Server) Initialize(cfg config.Local, phonebookAddresses []string, genes
...
+		cfg.ReservedFDs,
+		ot.Add(uint64(cfg.IncomingConnectionsLimit), cfg.RestConnectionsHardLimit))
+	if ot.Overflowed {
+		return errors.New(
+			"Initialize() overflowed when adding up ReservedFDs, IncomingConnectionsLimit " +
+				"RestConnectionsHardLimit; decrease them")
+	}
+	err = util.SetFdSoftLimit(fdRequired)
+	if err != nil {
+		return fmt.Errorf("Initialize() err: %w", err)
```
tsachiherman, line 116: could you please be more specific in this error ?
i.e. Server.Initialize() failed when calling util.SetFdSoftLimit(%d) with the following error : %w

network/limitlistener/rejectingLimitListener.go
```diff
@@ -0,0 +1,80 @@
...
+				return nil, fmt.Errorf("Accept() accept err: %w", err)
+			}
+			select {
+			case l.sem <- struct{}{}:
+				return &rejectingLimitListenerConn{Conn: c, release: l.release}, nil
+			default:
+				// Close connection immediately.
+				err = c.Close()
+				if err != nil {
+					return nil, fmt.Errorf("Accept() close err: %w", err)
```
tsachiherman, line 57: while I agree that we want to log this case, I think that we shouldn't return an error here.
returning an error here would be an attack vector : if someone could cause the Close here to return an error, it would exit from the Listening goroutine, and would no longer receive new connections.

network/limitlistener/rejectingLimitListener_test.go
```diff
@@ -0,0 +1,167 @@
...
+	defer l.Close()
+	l = limitlistener.RejectingLimitListener(l, max)
+
+	var open int32
+	go http.Serve(l, http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if n := atomic.AddInt32(&open, 1); n > max {
+			t.Errorf("%d open connections, want <= %d", n, max)
+		}
+		defer atomic.AddInt32(&open, -1)
+		time.Sleep(500 * time.Millisecond)
```
tsachiherman, line 48: using these constants guarantee that your test would randomly fail on CI system, where the machine would randomly be super sluggish.
Instead of waiting for a given timeout, use a channel to "lock" this go-routine at this state. That should give you the ability to control the scheduling of each goroutine.

network/limitlistener/rejectingLimitListener_test.go
```diff
@@ -0,0 +1,167 @@
...
+		c, err := ln.Accept()
+		if err == nil {
+			c.Close()
+			t.Errorf("Unexpected successful Accept()")
+		}
+		close(acceptDone)
+	}()
+
+	// Wait a tiny bit to ensure the Accept() is blocking.
+	time.Sleep(10 * time.Millisecond)
```
tsachiherman, line 159: As before, don't make any assumptions regarding test timing. Use channels for that.

network/limitlistener/rejectingLimitListener_test.go
tolikzinovyev: I changed the test, but unfortunately, there is no way to completely avoid timing assumptions.

network/limitlistener/rejectingLimitListener_test.go
tolikzinovyev: done

network/limitlistener/rejectingLimitListener.go
tsachiherman: yes; the alternative is to catch this error at the caller, and then to treat it differently. But I'm not sure that it's an easier option.

tolikzinovyev: done

daemon/algod/server.go
tsachiherman: I think that it's reasonable to "break" a configuration that should not have existed to start with.

tsachiherman (dismissed):

tsachiherman:
@winder & @algorandskiy could you please confirm that all your requests around this PR have been met by approving it ?

algorandskiy:
I have not got (yet) clarifications on why do we need that `for{}` loop in `Accept()`

network/limitlistener/rejectingLimitListener.go
winder: Why does it need to give a connection?

The `LimitListener` would be stuck inside this select so calling done would unblock the limit listener:
```
func (l *limitListener) acquire() bool {
	select {
	case <-l.done:
		return false
	case l.sem <- struct{}{}:
		return true
	}
}
```

I guess that you're saying we need to `Accept` each connection before closing it? That's already handled by the LimitListener, when acquiring returns false all connections are accepted then closed:
```
func (l *limitListener) Accept() (net.Conn, error) {
	if !l.acquire() {
		// If the semaphore isn't acquired because the listener was closed, expect
		// that this call to accept won't block, but immediately return an error.
		// If it instead returns a spurious connection (due to a bug in the
		// Listener, such as https://golang.org/issue/50216), we immediately close
		// it and try again. Some buggy Listener implementations (like the one in
		// the aforementioned issue) seem to assume that Accept will be called to
		// completion, and may otherwise fail to clean up the client end of pending
		// connections.
		for {
			c, err := l.Listener.Accept()
			if err != nil {
				return nil, err
			}
			c.Close()
		}
	}
```

network/limitlistener/rejectingLimitListener_test.go
winder: Aren't the errors your checking for in this test the SoftLimit? These renames are just suggestions, if there is a better name than "Test*" and "Test*Error" that would be fine as well.

network/limitlistener/rejectingLimitListener_test.go
winder: Trying to clarify the test names, you check for errors in all of them. In this case, you're testing the inner listener has an error.

If you have a different better name that would be fine, this is just a suggestion.

network/limitlistener/rejectingLimitListener.go
tolikzinovyev: Are you proposing to call `limitListener.Close()`? That would close the underlying listener, we don't want that.

network/limitlistener/rejectingLimitListener_test.go
tolikzinovyev: There is no concept of soft and hard limit within this test, just the limit. Renamed it to`TestRejectingLimitListenerBasic`, not sure it makes a big difference.

network/limitlistener/rejectingLimitListener_test.go
tolikzinovyev: renamed it to `TestRejectingLimitListenerBaseListenerError`

tolikzinovyev:
> I have not got (yet) clarifications on why do we need that `for{}` loop in `Accept()`

I'm not sure what kind of clarification you are looking for. If you know a better way to implement this, please let me know.

tsachiherman:
> This looked weird for me because [http.Serve](https://cs.opensource.google/go/go/+/refs/tags/go1.17.4:src/net/http/server.go;l=3001;drc=refs%2Ftags%2Fgo1.17.4) already has a for-accept loop. But I see why the second for-accept loop is needed here as well.

I think that @tolikzinovyev is good; the [http.Serve](https://cs.opensource.google/go/go/+/refs/tags/go1.17.4:src/net/http/server.go;l=3001;drc=refs%2Ftags%2Fgo1.17.4)  implementation is not well suited for a high throughput server. ( i.e. sleeping for 5ms is quite a long time when you want to support thousands of short-lived connections per second ).

network/limitlistener/rejectingLimitListener.go
winder: @tolikzinovyev thanks, I was missing that point. Sounds like forking the limit listener is the way to go.

winder (approved):
Thanks for the changes/explanation. This PR looks good!

algorandskiy (approved):

network/limitlistener/rejectingLimitListener.go
winder: I think @algorandskiy was suggesting the remaining connections would be flushed by the server implicitly because it constantly checks for new connections. Therefore the loop to terminate them here is not required.

PR #3327, opened 2021-12-17
ledger refactoring: update merkle trie on resource updates

tsachiherman:
## Summary

Changes in this PR:
1. rename `resourcesDeltas` -> `resourceDelta`.
2. add `address` to `resourceDelta`, which is needed by the catchpoint tracker. This also allow us to eliminate the `addresses` slice from the `compactResourcesDeltas`.
3. add support for `UpdateRound` for both accounts and resources.
4. ensure that we always initialize `resourceDelta.oldResource.aidx`, so that this field is always available.
5. better error handling in `resourcesLoadOld`
6. writing/deleting resources hash to merkle trie.

## Test Plan

Use & update existing tests.


ledger/accountdb.go
```diff
@@ -551,22 +588,25 @@ func (a *compactResourcesDeltas) resourcesLoadOld(tx *sql.Tx, knownAddresses map
...
+	for _, missIdx := range a.misses {
+		delta := a.deltas[missIdx]
+		addr := delta.address
 		if delta.oldResource.addrid != 0 {
 			addrid = delta.oldResource.addrid
 		} else if addrid, ok = knownAddresses[addr]; !ok {
 			err = addrRowidStmt.QueryRow(addr[:]).Scan(&addrid)
 			if err != nil {
-				if err == sql.ErrNoRows {
-					err = fmt.Errorf("base account not found while processing resource for addr=%s, aidx=%d", addr.String(), idx)
```
algorandskiy, line 564: does the default error include all info to identify what is missing?

ledger/accountdb_test.go
```diff
@@ -954,6 +954,15 @@ func TestCompactAccountDeltas(t *testing.T) {
 	a.Equal(sample2, data)
 }
 
+// upsertOld updates existing or inserts a new partial entry with only old field filled
```
algorandskiy, line 957: you can do the same with accounts' `upsertOld`

ledger/accountdb.go
tsachiherman: ok, added.

ledger/accountdb_test.go
tsachiherman: sure.. I'll do it on a separate PR ( it goes beyond the scope of this intended change.. )

algorandskiy (approved):

PR #3330, opened 2021-12-17
Ledger refactor: Lookup all resources for address

cce:
## Summary

This attempts to hook up Lookup() by address to retrieve all resources from deltas, caches, and SQL.

## Test Plan

Needs tests, or to check existing tests


ledger/acctupdates.go
```diff
@@ -137,7 +137,7 @@ type accountUpdates struct {
 
 	// resources stored the most recent resource state for every
 	// address&resource that appears in deltas.
-	resources map[accountCreatable]modifiedResource
+	resources map[basics.Address]map[basics.CreatableIndex]modifiedResource
```
tsachiherman, line 140: I don't think that the maintenance cost of having this double-maps is worth it. For the lookupAll, I think that iterating over all the resources is good enough, unless we can demonstrate that it's not the case.
( i.e. iterating over 100K map entries shouldn't take that long for a REST API call; given that the next upgrade would be to reduct the number of in-memory rounds from 320, I can't see why this would be needed ).

cce: I wasn't sure how much performance this would actually save, but I know we have users who make a lot of calls to algod's AccountInformation REST handler, so I was thinking of them. I'd argue that until we reduce the number of in-memory rounds from 320 it might be worth benchmarking the performance impact, just to double-check this wouldn't result in a significant degradation for these use cases. It was a pretty small LOC delta to add the extra check to see if you need create the extra map when adding a new entry, for what (possibly) could avoid a big degradation, but I'm just guessing without numbers..

tsachiherman: You're thinking about the AccountInformation REST endpoint as a common used endpoint. While it is true that this is a common use for this endpoint, the `algod` is general is not meant to be part of a scaling solution. If this optimization was affecting just the REST endpoint, I would be all for that. Unfortunately, it is not. It will definitely create a higher pressure on the GC for the "normal" operational case.
I was also contemplating if I should have proposed to have two maps instead. That approach has the same drawbacks as the previous one, which is why I dismissed it.

tsachiherman: Please correct me if I'm wrong, but keeping the storage as is would still allow us to get all the resources for a specific index using a single for-loop, right ?
i.e.
```golang
for res, mod := range resources {
    if res.addr != addr {
       continue
    }
    ... // do whatever
}
```
vs.
```golang
for idx, mod := range resources[addr] {
    ... // do whatever
}
```

tsachiherman: You would find that it's very rare to have many resources per address. Most of our addresses have only one asset associated with them.

cce: You could get them all with a single for-loop, I was just thinking of all the addresses+assets you might have to skip looking for the one address you're interested in. Having a map per address just to hold one asset does seem like a waste, I suppose you could optimize for a single-asset-per-address by having `map[basics.Address]struct{single modifiedResource; multiple map[basics.CreatableIndex]modifiedResource}` where the map is nil if there is only one resource held.

ledger/acctupdates.go
```diff
@@ -1115,6 +1136,149 @@ func (au *accountUpdates) lookupResource(rnd basics.Round, addr basics.Address,
...
+	// addToFound adds the resource to the current list of resources
+	addToFound := func(cidx basics.CreatableIndex, round basics.Round, data ledgercore.AccountResource) {
+		fr, ok := foundData[cidx]
+		if !ok { // first time seeing this cidx
+			foundCount++
+			foundData[cidx] = foundResource{round: round, res: data}
+		}
+		// is this newer than current "found" rnd for this resource? (XXX is this not possible?)
+		if round > fr.round {
+			panic("logic error in lookupAllResources") // remove after testing
```
tsachiherman, line 1170: should this be removed ?

ledger/accountdb.go
```diff
@@ -2002,26 +2002,28 @@ func (qs *accountsDbQueries) lookupAllResources(addr basics.Address) (data []per
...
-		var aidx basics.CreatableIndex
-		var rtype basics.CreatableType
 		data = nil
 		var buf []byte
 		for rows.Next() {
 			err := rows.Scan(&addrid, &dbRound, &aidx, &rtype, &buf)
 			if err != nil {
 				return err
 			}
+			if !addrid.Valid || !aidx.Valid || !rtype.Valid {
```
algorandskiy, line 2014: how do we know it is the first entry?

cce: oh right, well I just copied that logic from accountsDbQueries.listCreatables. Since addrid/aidx/rtype are NOT NULL columns you would really only ever get this in the empty case, but I see your point..

ledger/acctupdates.go
```diff
@@ -878,14 +897,80 @@ func (au *accountUpdates) lookupWithRewards(rnd basics.Round, addr basics.Addres
...
+				return false
+			}
+
+			// in cases where acct is only a holder of assets/apps, or is just a creator of assets/apps,
+			// we can know how many resources rows to look for
+			needToFind := uint64(0)
+			if ad.TotalAssetParams == 0 { // not a creator of assets
+				needToFind += uint64(ad.TotalAssets) // look for N asset holdings
+			} else if ad.TotalAssets == 0 { // not a holder of assets
+				needToFind += uint64(ad.TotalAssetParams) // look for N asset params
```
algorandskiy, line 945: Being asset creator assumes to have asset holding as well. That's why in many places only `ad.Assets` are checked (so a combination of ad.TotalAssetParams != 0 and ad.TotalAssets == 0 is not possible.

If user A created 5 assets and opted in into 2 other, so it has
TotalAssetParams = 5, TotalAssets = 7. In this case we need to load 7 rows (5 rows has both params and holdings and 2 holdings), but `needToFind` would be 0.
This means we can just set `needToFind = TotalAssets`.

ledger/acctupdates.go
```diff
@@ -878,14 +897,80 @@ func (au *accountUpdates) lookupWithRewards(rnd basics.Round, addr basics.Addres
...
+
+	// possibly avoid a trip to the DB for more resources, if we can use totals info
+	checkDone := func() bool {
+		if foundAccount { // found AccountData
+			// no resources
+			if ad.TotalAssetParams == 0 && ad.TotalAppParams == 0 &&
+				ad.TotalAssets == 0 && ad.TotalAppLocalStates == 0 {
+				return true
+			}
+			// not possible to know how many resources rows to look for: totals conceal possibly overlapping assets/apps
```
algorandskiy, line 933: this is true only for apps - please see the comment below

daemon/algod/api/server/v1/handlers/handlers.go
```diff
@@ -1320,8 +1319,7 @@ func AssetInformation(ctx lib.ReqContext, context echo.Context) {
 		return
 	}
 
-	lastRound := ledger.Latest()
-	record, err := ledger.Lookup(lastRound, creator)
+	record, _, err := ledger.LookupLatest(creator)
```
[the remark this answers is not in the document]
cce: Oh, OK, like one that just fetches a single resource by address, ctype, aidx? I didn't notice that the endpoint here is only interested in the creator AssetParams[aidx].. that would be a lot more straightforward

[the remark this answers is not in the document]
cce: fixed

ledger/accountdb.go
cce: You fixed it below.

ledger/ledgercore/statedelta.go
```diff
@@ -792,6 +810,9 @@ func (ad NewAccountDeltas) ApplyToBasicsAccountData(addr basics.Address, prev ba
 		for aidx, params := range prev.Assets {
 			result.Assets[aidx] = params
 		}
+		// if result.Assets == nil {
```
algorandskiy, line 813: I thought I removed it

ledger/internal/cow_creatables.go
```diff
@@ -74,41 +74,21 @@ func (cs *roundCowState) GetAssetParams(addr basics.Address, aidx basics.AssetIn
...
-	if _, ok := cs.mods.NewAccts.GetData(addr); !ok {
-		return fmt.Errorf("PutAssetHolding: %s not found in deltas for %d", addr.String(), aidx)
-	}
-
 	cs.mods.NewAccts.UpsertAssetHolding(addr, aidx, &data)
 	return nil
 }
 
 func (cs *roundCowState) PutAssetParams(addr basics.Address, aidx basics.AssetIndex, data basics.AssetParams) error {
-	// TODO: remove after the schema switch
```
algorandskiy, line 107: note `DeleteAssetParams` below still have such check. This is because delete is real deleting and requires `Total*s` update,  and `PutAssetParams` is for both create and update. Maybe it makes sense to separate them

ledger/ledgercore/statedelta.go
```diff
@@ -263,26 +263,34 @@ func (ad NewAccountDeltas) ModifiedAccounts() []basics.Address {
 		result[i] = ad.accts[i].Addr
 	}
 
-	// consistency check: ensure all addresses in params/holdings/states are also in base account
-	// TODO: remove after the schema switch
-	for aapp := range ad.appParamsCache {
-		if _, ok := ad.acctsCache[aapp.Address]; !ok {
-			panic(fmt.Sprintf("account app param delta: addr %s not in base account", aapp.Address))
+	// consistency check: ensure all addresses for deleted params/holdings/states are also in base accounts
```
algorandskiy, line 266: we might want to delete these loops to get some perf but it would require splitting `Put[Asset|App][Params|State|Holding]` into Create and Update methods to have such check in Create methods.

algorandskiy (approved):

PR #3335, opened 2021-12-20
tealdbg: increase intermediate reading/writing buffers

### LINKED ISSUE #3100: tealdbg: large program breaks cdt session
When using tealdbg if the disassembled teal is over a given size and the source is requested the CDT session will be disconnected 

image

I believe this is related to the buffer sizes set here: 

https://github.com/algorand/go-algorand/blob/master/cmd/tealdbg/server.go#L34-L35

To test I overwrote those values with 60k and started the sandbox with the fork and re-ran the tealdbg command. With these changes the TEAL loaded successfully.

barnjamin:
## Summary

Some large teal source files cause the tealdbg/cdt session to choke.  Upping the buffer size to allow for larger source files.

closes #3100 

## Test Plan

Run tealdbg with a large source teal file, ensure the source file makes it to cdt without choking.

### DIFF
--- a/cmd/tealdbg/server.go
+++ b/cmd/tealdbg/server.go
@@ -31,8 +31,8 @@ import (
 )
 
 var upgrader = websocket.Upgrader{
-	ReadBufferSize:  20480,
-	WriteBufferSize: 20480,
+	ReadBufferSize:  20480 * 4,
+	WriteBufferSize: 20480 * 4,
 	CheckOrigin: func(r *http.Request) bool {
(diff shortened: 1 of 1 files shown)


tsachiherman (changes requested):
The change makes sense; please define constants for these parameter, with their absolute values, and write in the description the rationale.

tsachiherman (dismissed):
Looks good, thanks for the changes.

tsachiherman:
Could you please fix the reviewdog issues ?

algorandskiy (approved):

PR #3339, opened 2021-12-21
add config.DeadlockDetectionThreshold

cce:
## Summary

This allows for the deadlock detection threshold to be set by configuration.

## Test Plan

Existing tests should pass.


config/localTemplate.go
```diff
@@ -192,6 +192,9 @@ type Local struct {
 	// negative (-1) to disable, positive (1) to enable, 0 for default.
 	DeadlockDetection int `version[1]:"0"`
 
+	// The threshold used for deadlock detection, in seconds.
+	DeadlockDetectionThreshold int `version[1]:"30"`
```
tsachiherman, line 196: `version[1]` -> `version[20]`
The config file is always being upgraded in-memory to the latest version, and therefore, we should just "append" instead of re-writing the history. Making this `version[1]` would also break many of the `config` package testing.

tsachiherman (approved):
Looks good.

PR #3340, opened 2021-12-21
add access to resources created in the current group

jannotti:
This lets AVM "see" assets and apps that were previously created in the txgroup.


data/transactions/logic/blackbox_test.go
```diff
@@ -80,7 +80,7 @@ func TestNewAppEvalParams(t *testing.T) {
 	for i, param := range params {
 		for j, testCase := range cases {
 			t.Run(fmt.Sprintf("i=%d,j=%d", i, j), func(t *testing.T) {
-				ep := logic.NewAppEvalParams(testCase.group, &param, nil, 0)
+				ep := logic.NewAppEvalParams(testCase.group, &param, nil)
```
jannotti, line 83: Removes use of an extra parameter unneeded by this new way to track the created resources.

data/transactions/logic/eval.go
```diff
@@ -211,7 +211,6 @@ type LedgerForLogic interface {
 	AssetParams(aidx basics.AssetIndex) (basics.AssetParams, basics.Address, error)
 	AppParams(aidx basics.AppIndex) (basics.AppParams, basics.Address, error)
 	OptedIn(addr basics.Address, appIdx basics.AppIndex) (bool, error)
-	GetCreatableID(groupIdx int) basics.CreatableIndex
```
jannotti, line 214: We no longer track creatables in the ledger, since ledgers are shared for inner transactions, but "racked creatables" used only the transaction index for tracking.

data/transactions/logic/eval.go
```diff
@@ -2611,27 +2639,28 @@ func (cx *EvalContext) getLatestTimestamp() (timestamp uint64, err error) {
...
 
-func (cx *EvalContext) getApplicationAddress() basics.Address {
-	// Initialize appAddrCache if necessary
-	if cx.appAddrCache == nil {
-		cx.appAddrCache = make(map[basics.AppIndex]basics.Address)
-	}
-
-	appAddr, ok := cx.appAddrCache[cx.appID]
+// getApplicationAddress memoizes app.Address() across a tx group's evaluation
+func (cx *EvalContext) getApplicationAddress(app basics.AppIndex) basics.Address {
```
jannotti, line 2643: Widened this method to cache app addresses for any given app id, so that the check for use of a created app id will be fast when repeated.

data/transactions/logic/eval.go
```diff
@@ -2611,27 +2639,28 @@ func (cx *EvalContext) getLatestTimestamp() (timestamp uint64, err error) {
...
-		err = fmt.Errorf("ledger not available")
-		return
+	if aid := cx.TxnGroup[groupIndex].ApplyData.ConfigAsset; aid != 0 {
+		return uint64(aid), nil
 	}
-	return uint64(cx.Ledger.GetCreatableID(groupIndex)), nil
+	if aid := cx.TxnGroup[groupIndex].ApplyData.ApplicationID; aid != 0 {
+		return uint64(aid), nil
+	}
+	return 0, fmt.Errorf("Index %d did not create anything", groupIndex)
```
jannotti, line 2663: Re-implemented `gaid` in terms of the ApplyDatas that we have in the EvalParams (that are kept up to date with RecordAD).

data/transactions/logic/eval.go
```diff
@@ -3302,17 +3335,40 @@ func (cx *EvalContext) accountReference(account stackValue) (basics.Address, uin
 	}
 	idx, err := cx.Txn.Txn.IndexByAddress(addr, cx.Txn.Txn.Sender)
 
+	invalidIndex := uint64(len(cx.Txn.Txn.Accounts) + 1)
+	// Allow an address for an app that was created in group
+	if err != nil && cx.version >= createdResourcesVersion {
```
jannotti, line 3340: We keep cx.created updated regardless of version, so when accessing it, we check whether we are at a version that allows in group access.

data/transactions/logic/eval.go
```diff
@@ -3533,7 +3589,7 @@ func opAppLocalPut(cx *EvalContext) {
 		return
 	}
 
-	addr, accountIdx, err := cx.accountReference(cx.stack[pprev])
+	addr, accountIdx, err := cx.mutableAccountReference(cx.stack[pprev])
```
jannotti, line 3592: New method introduced so that we do not allow AVM code that will modify an account to use the addresses created in the group.  It's not that we *want* to prevent modification, but we have no way to encode it. So we must disallow.

data/transactions/logic/eval.go
```diff
@@ -4340,6 +4426,9 @@ func opTxSubmit(cx *EvalContext) {
 			cx.err = err
 			return
 		}
+		// This is mostly a no-op, because Perform does its work "in-place", but
+		// RecordAD has some further responsibilities.
```
jannotti, line 4430: creatables must go into cx.created

ledger/apply/application.go
```diff
@@ -342,11 +342,7 @@ func ApplicationCall(ac transactions.ApplicationCallTxnFields, header transactio
 		if err != nil {
 			return
 		}
-		// No separate config for activating storage in AD because
-		// inner transactions can't be turned on without this change.
-		if balances.ConsensusParams().MaxInnerTransactions > 0 {
-			ad.ApplicationID = appIdx
-		}
+		ad.ApplicationID = appIdx
```
jannotti, line 345: Now, we always fill the AD with this useful fact (which is then recorded by using RecordAD) and we then remove it later based on protocol version, so it does not change old block evaluations.  This is more inline with how rewards AD fields were handled, and makes it easier to reason about code that wants to use this information.

ledger/apply/asset.go
```diff
@@ -100,12 +100,7 @@ func AssetConfig(cc transactions.AssetConfigTxnFields, header transactions.Heade
 			return err
 		}
 
-		// Record the index used. No separate config for activating
-		// storage in AD because inner transactions can't be turned on
-		// without this change.
-		if balances.ConsensusParams().MaxInnerTransactions > 0 {
-			ad.ConfigAsset = newidx
-		}
+		ad.ConfigAsset = newidx
```
jannotti, line 103: same.

ledger/internal/eval.go
```diff
@@ -968,15 +963,25 @@ func (eval *BlockEvaluator) applyTransaction(tx transactions.Transaction, balanc
...
 	}
-	if evalParams != nil {
-		evalParams.TxnGroup[gi].ApplyData = ad
+
+	// No separate config for activating these AD fields because inner
+	// transactions require their presence, so the consensus update to add
+	// inners also stores these IDs.
+	if params.MaxInnerTransactions == 0 {
+		ad.ApplicationID = 0
+		ad.ConfigAsset = 0
```
jannotti, line 984: This is the "use the AD with all its nice fields, and only remove them just before writing" strategy described earlier. Note similarity to rewards handling.

jannotti (commented):
Lots of explanations added.

data/transactions/logic/eval.go
```diff
@@ -272,22 +278,31 @@ type EvalParams struct {
...
+	// a foreign array. They remain empty until createdResourcesVersion.
+	created *resources
+
+	// Caching these here means the hashes can be shared across the TxnGroup
+	// (and inners, because the cache is shared with the inner EvalParams)
+	appAddrCache map[basics.AppIndex]basics.Address
+
+	// Cache the txid hashing, but do *not* share this into inner EvalParams, as
+	// the key is just the index in the txgroup.
+	txidCache map[int]transactions.Txid
```
jannotti, line 305: Bad dog! Review dog is wrong. This field is used, but it is only used through embedding.  I'll change that code to use `cx.EvalParams.txidCache` which should convince it.

PR #3341, opened 2021-12-21
Add Version KV to telemetry events

cce:
## Summary

This piggybacks off the places where "session" is added to telemetry events, to additionally add a "v" key containing the version string.

## Test Plan

Examine log output from tests.


logging/telemetry.go
```diff
@@ -264,6 +266,7 @@ func (t *telemetryState) logTelemetry(l logger, message string, details interfac
 	entry := l.entry.WithFields(Fields{
 		"session":      l.GetTelemetrySession(),
 		"instanceName": l.GetInstanceName(),
+		"v":            l.GetTelemetryVersion(),
```
winder, line 269: Could you use a full string for these?
```suggestion
		"version":            l.GetTelemetryVersion(),
```

logging/telemetryFilteredHook.go
```diff
@@ -82,6 +82,13 @@ func (hook *telemetryFilteredHook) Fire(entry *logrus.Entry) error {
 	} else {
 		newEntry = entry.WithField("session", hook.sessionGUID)
 	}
+
+	// Also add version field, if not already present.
+	if _, has := entry.Data["v"]; has {
+		newEntry = entry
+	} else {
+		newEntry = entry.WithField("v", hook.telemetryConfig.Version)
+	}
```
winder, line 91:
```suggestion
	// Also add version field, if not already present.
	if _, has := entry.Data["v"]; has {
		newEntry = newEntry
	} else {
		newEntry = newEntry.WithField("v", hook.telemetryConfig.Version)
	}
```

winder (changes requested):
Looks good, just noticed one small bug

logging/telemetry.go
cce: The concern was that it would use up more bandwidth, so squish it down to the minimum.

tsachiherman (approved):

PR #3343, opened 2021-12-21
txsync: compress msgOrderingHeap

nicholasguoalgorand:
## Summary

Whenever the msgOrderingHeap gets full, we would like to compress the messages inside of it instead of disconnecting from the corresponding peer.

## Test Plan

Wrote additional unit tests.

### DIFF
--- a/txnsync/msgorderingheap.go
+++ b/txnsync/msgorderingheap.go
@@ -70,7 +70,12 @@ func (p *messageOrderingHeap) enqueue(msg incomingMessage) error {
 	p.mu.Lock()
 	defer p.mu.Unlock()
 	if len(p.messages) >= messageOrderingHeapLimit {
-		return errHeapReachedCapacity
+		// try compressing the msgorderingheap first
+		p.compress()
+		if len(p.messages) >= messageOrderingHeapLimit {
+			// return an error if still can't enqueue
+			return errHeapReachedCapacity
+		}
 	}
 	heap.Push(p, messageHeapItem(msg))
 	return nil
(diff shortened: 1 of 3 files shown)


txnsync/msgorderingheap.go
```diff
@@ -98,3 +103,30 @@ func (p *messageOrderingHeap) pop() (msg incomingMessage, err error) {
 	entry := heap.Pop(p).(messageHeapItem)
 	return incomingMessage(entry), nil
 }
+
+func (p *messageOrderingHeap) compress() {
```
tsachiherman, line 107: maybe rename "compress" -> "compact" ? The "compress" keyword implies a reversible packing action, which isn't the case here.

txnsync/msgorderingheap.go
```diff
@@ -98,3 +103,30 @@ func (p *messageOrderingHeap) pop() (msg incomingMessage, err error) {
...
+	}
+	compressedEntry := heap.Pop(p).(messageHeapItem)
+	expectedSeqNum := compressedEntry.sequenceNumber + 1
+	for len(p.messages) != 0 {
+		nextEntry := heap.Pop(p).(messageHeapItem)
+		// compress only consecutive messages
+		if nextEntry.sequenceNumber != expectedSeqNum {
+			heap.Push(p, nextEntry)
+			break
+		}
```
tsachiherman, line 119: instead of popping and pushing back, use `popSequence`.

tsachiherman (changes requested):
I think that the changes here are a good start, but we would require more changes for this to work ( including some testing that shows that this works.. )

Specifically, in `incoming.go`, each message that being placed on the `incomingMessages` should contain it's "starting" seq number, as well as the "next message seq number". i.e. in `asyncIncomingMessageHandler`, you should initialize `incomingMessage.NextSequenceNumber` with `sequenceNumber+1`. When compacting the heap, you should retain the sequenceNumber for the first entry, and update the NextSequenceNumber for the last entry.
Last, you'll need to use that in `evaluateIncomingMessage`.

nicholasguoalgorand:
> ```go
> nextEntry.nextSequenceNumber = compressedEntry.nextSequenceNumber
> ```

Are you sure? This will keep the nextSequenceNumber of the oldest message.

tsachiherman (approved):

PR #3346, opened 2021-12-22
Fix flaky test in randomized ABI encoding test

ahangsu:
## Summary

Previously @cce noticed that there's a flaky test in ABI encoding.

> Chris Erway: just encountered a PR of mine failing due to what looks like a new flaky test related to encoding randomly generated values? https://app.circleci.com/pipelines/github/algorand/go-algorand/3937/workflows/7bd0914e-57f8-4d7e-9fa9-026f74c831bf/jobs/58162
> 

The reason for the flaky test is that we randomly generate variables to test ABI encoding, and it turns out that the generated ABI-encoding bytes is larger than `uint16` limit. The encoder would just throw an error complaining `encode length exceeds uint16 maximum`.

The solution is to scale down the parameters for generating random variables, to decrease the generated bytes length, not to reach the `2^16` length limit.

We gathered the constants in test scripts for better readability.

## Test Plan

We should expect that the modification eliminate the flaky test chance...

### DIFF
--- a/data/abi/abi_encode_test.go
+++ b/data/abi/abi_encode_test.go
@@ -27,17 +27,35 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+const (
+	UintStepLength             = 8
+	UintBegin                  = 8
+	UintEnd                    = 512
+	UintRandomTestPoints       = 1000
+	UintTestCaseCount          = 200
+	UfixedPrecision            = 160
+	UfixedRandomTestPoints     = 20
+	TupleMaxLength             = 10
+	ByteTestCaseCount          = 1 << 8
+	BoolTestCaseCount          = 2
+	AddressTestCaseCount       = 300
+	StringTestCaseCount        = 100
+	StringTestCaseSpecLenCount = 5
+	TakeNum                    = 10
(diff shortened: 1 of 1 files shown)


jannotti (commented):
This seems like the right fix.  Could you also put in come notes about how you came up with appropriate constants? For example, if you use something that is B bytes lony X times in a loop that runs Y times, perhaps you need B*X*Y to be less that 1<<16.

In fact, you could put a couple simple require.Less() calls in on the constants to document the requirements.

tzaffi:
For future reference, putting a link the the @cce's [original slack message](https://algorand-internal.slack.com/archives/C020LLY8VBM/p1640112787497300).

tzaffi (commented):
Just one suggestions.

tzaffi (dismissed):
LGTM

jannotti:
> ```go
> require.NoError(t, err, "failure to encode value", fmt.Sprintf("failing test object:\n%#v\n", testObj))
> ```

Using Go's require, you can omit the fmt.Sprintf, and just use
```
require.NoError(t, err, "failure to encode %#v", testObj)
```

jannotti (approved):

PR #3348, opened 2021-12-22
ledger refactoring: addrid lookup fix

algorandskiy:
Storing `resourceData` in `baseResources` cache for non-existing resources breaks addr to addrid resolution in makeCompactResourcesDelta. Improved the code by skipping empty addrids. This was discovered by TestAppAccountDataStorage.

Fixed some additional tests

### DIFF
--- a/ledger/accountdb.go
+++ b/ledger/accountdb.go
@@ -1614,7 +1614,7 @@ func accountDataResources(
 				rd.SetAppParams(ap, true)
 				delete(accountData.AppParams, aidx)
 			}
-			err := cb(ctx, rowid, basics.CreatableIndex(aidx), basics.AssetCreatable, &rd)
+			err := cb(ctx, rowid, basics.CreatableIndex(aidx), basics.AppCreatable, &rd)
 			if err != nil {
 				return err
 			}
@@ -1622,7 +1622,7 @@ func accountDataResources(
 		for aidx, aparams := range accountData.AppParams {
 			var rd resourcesData
 			rd.SetAppParams(aparams, false)
-			err := cb(ctx, rowid, basics.CreatableIndex(aidx), basics.AssetCreatable, &rd)
+			err := cb(ctx, rowid, basics.CreatableIndex(aidx), basics.AppCreatable, &rd)
 			if err != nil {
 				return err
 			}
--- a/ledger/accountdb.go
+++ b/ledger/accountdb.go
@@ -404,9 +404,11 @@ func makeCompactResourceDeltas(accountDeltas []ledgercore.NewAccountDeltas, base
 					// meaning empty asset hodling (or app local state)
 					newEntry.newResource.ClearAssetHolding()
 				}
-				if baseResourceData, has := baseResources.read(assetHold.Addr, basics.CreatableIndex(assetHold.Aidx)); has {
+				baseResourceData, has := baseResources.read(assetHold.Addr, basics.CreatableIndex(assetHold.Aidx))
+				existingAcctCacheEntry := has && baseResourceData.addrid != 0
+				if existingAcctCacheEntry {
 					newEntry.oldResource = baseResourceData
-					outResourcesDeltas.insert(newEntry) // insert instead of upsert economizes one map lookup
+					outResourcesDeltas.insert(newEntry)
 				} else {
 					if pad, has := baseAccounts.read(assetHold.Addr); has {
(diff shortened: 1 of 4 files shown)


ledger/accountdb.go
```diff
@@ -2848,44 +2848,65 @@ func makeOrderedAccountsIter(tx *sql.Tx, accountCount int) *orderedAccountsIter
...
-			err = errors.New("resource table entries mismatches accountbase table entries")
-			return
+		if pr.addrid != 0 {
+			// some accounts may not have resources, consider the following case:
+			// acct 1 and 3 has resources, account 2 does not
+			// in this case addrid = 3 after processing resources from 1, but acctRowid = 2
+			// and we need to skip accounts without resources
+			if pr.addrid > acctRowid {
+				err = callback(addr, 0, 0, &resourcesData{}, nil)
+				return pendingRow{}, err
```
pzbitskiy, line 2876:
```suggestion
				return pd, err
```

ledger/acctupdates.go
```diff
@@ -1197,8 +1212,13 @@ func (au *accountUpdates) lookupWithoutRewards(rnd basics.Round, addr basics.Add
 		// against the database.
 		persistedData, err = au.accountsq.lookup(addr)
 		if persistedData.round == currentDbRound {
-			au.baseAccounts.writePending(persistedData)
-			return persistedData.accountData.GetLedgerCoreAccountData(), rnd, err
+			if persistedData.rowid != 0 {
```
tsachiherman, line 1215: this change means that it will make a two disk lookup for this account rather than just one, right ?
( i.e. there will be another lookup in loadOld... )

I wasn't able to figure out from this PR why this change is needed. Would you mind sharing your insights ?

algorandskiy: ah, I see, is `baseAccounts` intended to cache non-existing accounts? `makeCompactResourceDeltas` relies on them to resolve address into rowid. So lookup op on non-existing yet account X (resource Y) for a round N puts empty records into `baseAccounts`/`baseResources` as pending, then on round N+1 they gets flushed and `makeCompactResourceDeltas` uses them to find addrid. If account X (resource Y) was created in N+1, this would cause getting zero rowid/addrid from `baseAccounts`/`baseResources` since the resolution happens before updating base cashes with actual values with results from `newAccountsRound`.
So there are two solutions: either do not cache missing records (as I did since it looked logical for me) or update `makeCompactResourceDeltas` with zero checks.

tsachiherman: looking on `makeCompactResourceDeltas`, the method would only use the `baseAccounts` when populating the `oldResource` for the case where the entry is missing.
The `resourcesLoadOld` will be invoked later on, and would fail making the `knownAddresses`, then, will default into making the disk lookup `addrRowidStmt` needlessly.
( so far, no failure would occur. just a wasted `addrRowidStmt` query ).
Last, in `accountsNewRound`, it *attempts to* find the new address for the resource in `newAddressesRowIDs`. If it fails here, it would need to perform the disk lookup ( i.e. the `addrRowidStmt` ).

All that being said, I suspect that you'll find that the true issue is in the way we prepare the random entries we put in the database. In particular, creating a resource for an empty account could have this outcome.

algorandskiy: I basically agree there is no problem with `baseAccounts` in `makeCompactResourceDeltas` but with `baseResources` for the reason stated above. If there is incomplete data in `baseResources` then it is not added to missing ones and no call to `resourcesLoadOld`. 
I'm against having additional disc lookup in `accountsNewRound`. All data needs to be resolved in advance - that's the point of having `resourcesLoadOld`

tsachiherman: I think that it makes a perfect sense loading all in advance at `resourcesLoadOld`; however, we would need to make a differentiation at that point between accounts that are known to be empty ( and therefore, should trigger a failure ) and accounts that aren't known yet. The `baseAccounts` cache has a limited size, and it being pruned often, hence, we shouldn't rely on its content unless we have the `accountsMu` lock taken.
what about storing in the `insertMissing` also a separate bool to indicate if the entry was known to be empty or whether it's missing ? that would allow us to make the correct call when `resourcesLoadOld` gets invoked.

algorandskiy: removed

tsachiherman (approved):
this PR doesn't handle the use case of avoiding reloading resource entries which we already know aren't there. But we can add this to a subsequent PR.

PR #3349, opened 2021-12-24
opcode json_ref

shiqizng:
## Summary
This PR adds a new opcode json_ref that takes in a json string and a key name, and yields the value of that key in the json string. Has one immediate argument that can be either Uint or String for the opcode.


data/transactions/logic/evalStateful_test.go
```diff
@@ -2393,6 +2393,7 @@ func TestReturnTypes(t *testing.T) {
 		// This next one is a cop out.  Can't use itxna Logs until we have inner appl
 		"itxna":         "itxn_begin; int pay; itxn_field TypeEnum; itxn_submit; itxn NumLogs",
 		"base64_decode": `pushbytes "YWJjMTIzIT8kKiYoKSctPUB+"; base64_decode StdEncoding; pushbytes "abc123!?$*&()'-=@~"; ==; pushbytes "YWJjMTIzIT8kKiYoKSctPUB-"; base64_decode URLEncoding; pushbytes "abc123!?$*&()'-=@~"; ==; &&; assert`,
+		"json_ref":      "json_ref JSONInt",
```
tzaffi, line 2396: I'm surprised the test is passing, since I would expect that some json and key bytes need to be pushed to the stack before calling `json_ref`. I wonder if the test is broken.

tzaffi (commented):
This is an ambitious approach to make it easy for the AVM writer to dig into deeply nested JSON. It does go beyond the original scope of the `json_ref` issue, but I think it's worth the extra effort. On the other hand, more edge cases need to be considered (either handled or explicitly forbidden) and I try to point some of these out.

data/transactions/logic/eval.go
```diff
@@ -4060,3 +4061,62 @@ func opBase64Decode(cx *EvalContext) {
...
+		return
+	}
+	//get value from json
+	for i := 0; i < len(fields); i++ {
+		key := fields[i]
+		switch resp.(type) {
+		case map[string]interface{}:
+			if _, ok := resp.(map[string]interface{})[key]; ok {
+				resp = resp.(map[string]interface{})[key]
+			} else {
```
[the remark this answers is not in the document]
tzaffi: > Are you planning on handling the case of type `[]interface{}` (json array) ? E.g. in [this playground](https://go.dev/play/p/ho3Mxb8M8kI). That would be nice, but it also causes complications because the natural thing to do is to handle array indices numbered parts of the path. In that case, a path such as "key1.4.key2" could be parsed as `["key1", "4", "key2"]` or as `["key1", 4, "key2"]` and one would need to handle both cases (or declare one of them illegal).

Another issue which is probably more fundamental is what if a key itself contains ".". It would be impossible to read its value under the current approach. I have to take some flack here for providing the Python glom example on slack a couple of weeks back.

data/transactions/logic/TEAL_opcodes.md
```diff
@@ -865,6 +865,16 @@ When A is a uint64, index 0 is the least significant bit. Setting bit 3 to 1 on
...
 
+## json_ref r
+
+- Opcode: 0x5d {return type}
+- Pops: *... stack*, {[]byte A}, {[]byte B}
+- Pushes: any
+- return key B's value from json text A
+- LogicSigVersion >= 6
+
+specify the return type with an immediate arg either as JSONInt or JSONString. When B is a nested key of json object A, A.B returns the value of B. Providing a json with keys containing . will return incorrect value. See <a href="https://datatracker.ietf.org/doc/html/rfc7159.html">RFC 7159</a> for additional json specifications
```
jannotti, line 876: How confident are we that the Go parsing follows this spec precisely?  There are a lot of little details about trailing commas, unicode escapes, etc that different implementations do different.  If Go's parsing folows it exactly, that's great, because we can say less here, but we need test cases for ever distinction we can think of and if any depart, say so. (Or if the spec is unclear, say so either wy)

data/transactions/logic/assembler_test.go
```diff
@@ -345,6 +345,7 @@ itxna Logs 3
 const v6Nonsense = v5Nonsense + `
 itxn_next
 base64_decode URLEncoding
+json_ref JSONInt
```
jannotti, line 348: I think we need to call the constant JSONUInt.  Because it needs to fail if the value is negative, and making that clearer would be good.  Maybe even JSONUInt64, to be clear about the limits.

data/transactions/logic/doc.go
```diff
@@ -98,15 +98,16 @@ var opDocByName = map[string]string{
...
+	"global":   "push value from globals to stack",
+	"load":     "copy a value from scratch space to the stack. All scratch spaces are 0 at program start.",
+	"store":    "pop value X. store X to the Ith scratch space",
+	"loads":    "copy a value from the Xth scratch space to the stack.  All scratch spaces are 0 at program start.",
+	"stores":   "pop indexes A and B. store B to the Ath scratch space",
+	"gload":    "push Ith scratch space index of the Tth transaction in the current group",
+	"gloads":   "push Ith scratch space index of the Xth transaction in the current group",
+	"gaid":     "push the ID of the asset or application created in the Tth transaction of the current group",
+	"gaids":    "push the ID of the asset or application created in the Xth transaction of the current group",
+	"json_ref": "return key B's value from json text A",
```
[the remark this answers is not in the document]
jannotti: I agree about ordering.  I think the json string should be deeper than the key.

data/transactions/logic/doc.go
```diff
@@ -287,6 +289,7 @@ var opDocExtras = map[string]string{
...
 	"base64_decode":       "Decodes X using the base64 encoding E. Specify the encoding with an immediate arg either as URL and Filename Safe (`URLEncoding`) or Standard (`StdEncoding`). See <a href=\"https://rfc-editor.org/rfc/rfc4648.html#section-4\">RFC 4648</a> (sections 4 and 5). It is assumed that the encoding ends with the exact number of `=` padding characters as required by the RFC. When padding occurs, any unused pad bits in the encoding must be set to zero or the decoding will fail. The special cases of `\\n` and `\\r` are allowed but completely ignored. An error will result when attempting to decode a string with a character that is not in the encoding alphabet or not one of `=`, `\\r`, or `\\n`.",
+	"json_ref":            "specify the return type with an immediate arg either as JSONInt or JSONString. When B is a nested key of json object A, A.B returns the value of B. Providing a json with keys containing . will return incorrect value. See <a href=\"https://datatracker.ietf.org/doc/html/rfc7159.html\">RFC 7159</a> for additional json specifications",
```
jannotti, line 292: Since we are _really_ returning bytes and not characters (and reading bytes as characters), we need to be clear about any encoding that's going on.  What does JSON say about encodings? Maybe it only allows low-ascii?  That'd be nice.  Otherwise we need to talking more about how we convert the bytes to characters (utf-8?) and then how we convert the string extracted to bytes.  Again, utf-8?

data/transactions/logic/doc.go
```diff
@@ -302,7 +305,7 @@ var OpGroups = map[string][]string{
 	"Byte Array Manipulation": {"substring", "substring3", "extract", "extract3", "extract_uint16", "extract_uint32", "extract_uint64", "base64_decode"},
 	"Byte Array Arithmetic":   {"b+", "b-", "b/", "b*", "b<", "b>", "b<=", "b>=", "b==", "b!=", "b%"},
 	"Byte Array Logic":        {"b|", "b&", "b^", "b~"},
-	"Loading Values":          {"intcblock", "intc", "intc_0", "intc_1", "intc_2", "intc_3", "pushint", "bytecblock", "bytec", "bytec_0", "bytec_1", "bytec_2", "bytec_3", "pushbytes", "bzero", "arg", "arg_0", "arg_1", "arg_2", "arg_3", "args", "txn", "gtxn", "txna", "txnas", "gtxna", "gtxnas", "gtxns", "gtxnsa", "gtxnsas", "global", "load", "loads", "store", "stores", "gload", "gloads", "gaid", "gaids"},
+	"Loading Values":          {"intcblock", "intc", "intc_0", "intc_1", "intc_2", "intc_3", "pushint", "bytecblock", "bytec", "bytec_0", "bytec_1", "bytec_2", "bytec_3", "pushbytes", "bzero", "arg", "arg_0", "arg_1", "arg_2", "arg_3", "args", "txn", "gtxn", "txna", "txnas", "gtxna", "gtxnas", "gtxns", "gtxnsa", "gtxnsas", "global", "load", "loads", "store", "stores", "gload", "gloads", "gaid", "gaids", "json_ref"},
```
jannotti, line 308: I think I'd put it in byte array manipulation, because it's conceptually so similar to substring.  But I could be convinced this is kind of "loading".

data/transactions/logic/eval.go
```diff
@@ -4060,3 +4061,62 @@ func opBase64Decode(cx *EvalContext) {
 	}
 	cx.stack[last].Bytes, cx.err = base64Decode(cx.stack[last].Bytes, encoding)
 }
+
+func parseJSON(jsonText []byte) (interface{}, error) {
+	if !json.Valid(jsonText) {
+		return nil, fmt.Errorf("invalid json text %s", jsonText)
+	}
+	var parsed interface{}
+	err := json.Unmarshal(jsonText, &parsed)
```
jannotti, line 4070: What do we gain with the first test?  Does Unmarshal not return a proper error if the json is invalid?

data/transactions/logic/eval.go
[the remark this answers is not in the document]
jannotti: Questions like this make me wonder if we want to support the "." path thing.  If we don't, we probably need to support a return type of "JSONObject" which return a marshalled json string, so json_ref can be used again to dig into it.  But if go this we, it really bugs me that we just can't possibly decode json with "." in the key.  I've definitely seen this (though I think it's gross).

data/transactions/logic/eval.go
```diff
@@ -4060,3 +4061,62 @@ func opBase64Decode(cx *EvalContext) {
...
+	}
+	expectedType := jsonRefTypeNames[cx.program[cx.pc+1]]
+	var val stackValue
+	switch resp.(type) {
+	case float64:
+		if expectedType != "JSONInt" {
+			cx.err = fmt.Errorf("value retrieved from JSON text is type int but expected %s", expectedType)
+			return
+		}
+		val.Uint = uint64(resp.(float64))
```
jannotti, line 4110: I don't think we can parse this way.  Very large integers will get converted to a float, and then during the cast to uint64, they will become a different integer. (also, this allows a negative number, which then gets converted to positive by the cast)

data/transactions/logic/eval_test.go
```diff
@@ -5117,3 +5117,93 @@ func TestOpBase64Decode(t *testing.T) {
...
+		},
+		{
+			source: `byte  "{\"key0\": 1,\"key1\": \"algo\",\"key2\":{\"key3\": \"teal\", \"key3\": [1,2,3]} }"; byte "key1"; json_ref JSONString; byte "algo"; ==`,
+		},
+		{
+			source: `byte  "{\"key0\": 1,\"key1\": \"algo\",\"key2\":{\"key3\": \"teal\", \"key4\": [1,2,3]} }"; byte "key2.key3"; json_ref JSONString; byte "teal"; ==`,
+		},
+		{
+			source: `byte  "{\"rawId\": \"responseId\",\"id\": \"0\",\"response\": {\"attestationObject\": \"based64url_encoded_buffer\",\"clientDataJSON\":  \"based64url_encoded_client_data\"},\"getClientExtensionResults\": {},\"type\": \"public-key\"}"; byte "response.clientDataJSON"; json_ref JSONString; byte "based64url_encoded_client_data"; ==`,
+		},
```
jannotti, line 5150: We need to test negative numbers, and floats, which should both fail. (If you try to read a value and it's -5, or 3.2, it should not silently coerce it to some integer). We also need to test very larger integers, close to 2^64.  In particular, we need to have tests that ensure an integer that can not be represented well by a float gets treated properly without conversion into and out of floating point.

I imagine we have to ask Go to parse using big.Int to represent the outcome.

data/transactions/logic/jsonspec_test.go
```diff
@@ -0,0 +1,336 @@
...
+	"testing"
+
+	"github.com/algorand/go-algorand/test/partitiontest"
+	"github.com/stretchr/testify/require"
+	"golang.org/x/text/encoding/unicode"
+	"golang.org/x/text/encoding/unicode/utf32"
+)
+
+// As of go1.17.5, json implements encoding and decoding of JSON as defined in RFC 7159. https://pkg.go.dev/encoding/json
+type utf8String string
```
jannotti, line 33: This concerns me.  Did Go *change* it's implementation and therefore before 1.17.5, we would get different behavior?  If so, we must either detect and fix, or demand compilation with 1.17.5 or greater, or else we'll get the potential for disagreement about results.

data/transactions/logic/jsonspec_test.go
```diff
@@ -0,0 +1,336 @@
...
+	require.True(t, json.Valid([]byte(text)))
+	require.Equal(t, uint64(9007199254740992), uint64(msg.(map[string]interface{})["key0"].(float64)))
+
+	//2<<52+1
+	text = `{"key0":9007199254740993}`
+	msg, err = parseJSON([]byte(text))
+	require.Nil(t, err)
+	require.True(t, json.Valid([]byte(text)))
+	require.NotEqual(t, uint64(9007199254740993), uint64(msg.(map[string]interface{})["key0"].(float64)))
+	require.Equal(t, uint64(9007199254740992), uint64(msg.(map[string]interface{})["key0"].(float64)))
```
jannotti, line 157: This is what we must avoid. The JSON _spec_ does not say that big integers get changed.  That's a common implementation issue, as many parsers convert the integers to floats.  I hope we can get Go not to do it, else we'll need a different parsing library.

(These are great tests, by the way, some of my earlier comments said we needed a lot of tests that I hadn't seen down here yet.)

data/transactions/logic/jsonspec_test.go
```diff
@@ -0,0 +1,336 @@
...
+}
+
+func TestParseExp(t *testing.T) {
+	partitiontest.PartitionTest(t)
+	text := `{"key0": 1.2E+}`
+	_, err := parseJSON([]byte(text))
+	require.NotNil(t, err)
```
jannotti, line 164: require.Error and require.NoError seem clearer here.

data/transactions/logic/jsonspec_test.go
```diff
@@ -0,0 +1,336 @@
...
+	text = `{1: 1}`
+	_, err = parseJSON([]byte(text))
+	require.NotNil(t, err)
+
+}
+
+func TestParseFileEncoding(t *testing.T) {
+	partitiontest.PartitionTest(t)
+	//utf-8
```
jannotti, line 229: Can you describe these test somewhat?  It seems like you might be addresses some of my concerns about encodings, but I don't know that I'm following everything.

jannotti (commented):
The biggest design issue to settle is whether we want to allow "paths" like x.y for keys.

data/transactions/logic/eval.go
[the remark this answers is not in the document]
tzaffi: If we provide a `split()` opcode, then parsing json with the original approach would be almost as simple as the "." functionality but not cause the headaches...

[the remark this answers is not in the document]
jannotti: Since AVM has no arrays, split itself is tricky.  Put answers on the stack? With a length? Or maybe a terminating ""?  If the later, what does `pushbytes ".."; split;` do?

data/transactions/logic/eval.go
jasonpaulos: If you use a JSON decoder with the `UseNumber()` option, you can avoid converting parsed numbers into float64. Instead, they will get decoded as a `json.Number`, which is really just the raw string holding the number as text. For example:

```go
decoder := json.NewDecoder(bytes.NewReader(jsonText))
decoder.UseNumber()

var parsed interface{}
err := decoder.Decode(&parsed)
```

From there you could call [`Int64`](https://pkg.go.dev/encoding/json#Number.Int64) to convert to an int64, or create a new `big.Int` and call [`UnmarshalText`](https://pkg.go.dev/math/big#Int.UnmarshalText) on it to convert directly to a big int. We would probably want the `big.Int` approach so that the entire uint64 range can be properly decoded.

data/transactions/logic/eval_test.go
jasonpaulos: Can be addressed by this comment: https://github.com/algorand/go-algorand/pull/3349/files#r776080968

data/transactions/logic/eval.go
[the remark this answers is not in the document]
jasonpaulos: I think completeness is more important than convenience for this opcode, so I'm included to say the key should **NOT** be interpreted as a path because of the "." issue.

Sure, in most cases fields won't have a "." in them, and perhaps we could provide another opcode to "dig" into a JSON object with a path for convenience (though I don't think it's worth it at this time), but since this opcode version has shortcomings, it should not be our only JSON parsing opcode.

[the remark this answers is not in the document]
jannotti: I think I feel the same way, but it does mean we need to add another possible extraction, an entire JSONObject, which would be serialized. And we'll have to define that serialization - it needs to have a canonical form beyond "JSON" - we'll have to specify whitespace, key order, escape usage (I think there are multiple ways to escape the same json string).

Unless we can determine that in the first use case, there is no need to dig into an object in the main object.  Sadly, I think that _is_ a requirement, but we can check and skip this for now if it isn't.

data/transactions/logic/doc.go
jannotti: I did some reading and it seems to assume the bytes going into unmashalling are utf-8.  That's good, and well defined.  And I think it's fine for use to encode to bytes as utf-8 as well.  The "BOM" needs some attention though - it seems like it is allowed but ignored in some cases?

data/transactions/logic/eval.go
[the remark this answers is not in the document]
jasonpaulos: We'll have to check for edge cases, but Go's `json.RawMessage` seems like a good way to return a subobject "Exactly as it appears in the original json".

Here's an example I made: https://go.dev/play/p/_IXxH7zx1XD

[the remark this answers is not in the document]
jannotti: Well that looks great!

data/transactions/logic/jsonspec_test.go
shiqizng: I'll update this comment. Go json started using RFC 7159 in v1.10. I think we're fine since we are requiring build to use v1.16 or later.

data/transactions/logic/doc.go
shiqizng: I'll group it with substring.

data/transactions/logic/TEAL_opcodes.md
shiqizng: will it be better to list out the specs in more details based on the test results from jsonspec_test.go?

data/transactions/logic/jsonspec_test.go
jannotti: If we can find where they talk about this change, it might be a good source of ideas for tests.  They might describe small changes between the old and new implementation.

data/transactions/logic/eval.go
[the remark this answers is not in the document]
shiqizng: the plan is to push rawmessage to stack and call json_ref again to dig the value, correct?

data/transactions/logic/jsonspec_test.go
shiqizng: https://datatracker.ietf.org/doc/html/rfc7159.html#section-1.3, it doesn't list behavior changes in this update.

data/transactions/logic/eval.go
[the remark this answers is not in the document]
jasonpaulos: @shiqizng that's correct. `json_ref` should let you extract one level at a time, and we will need `JSONObject` to extract an embedded object.

@jannotti is there a requirement to support JSON array access as well?

data/transactions/logic/doc.go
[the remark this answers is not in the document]
tzaffi: I should point out that  misread the code. It is already following the same conventions as `substring3` etc.

data/transactions/logic/jsonspec_test.go
shiqizng: added a description.

data/transactions/logic/jsonspec_test.go
shiqizng: changed to using require.Error

data/transactions/logic/TEAL_opcodes.md
```diff
@@ -865,6 +865,16 @@ When A is a uint64, index 0 is the least significant bit. Setting bit 3 to 1 on
...
 
+## json_ref r
+
+- Opcode: 0x5d {return type}
+- Pops: *... stack*, {[]byte A}, {[]byte B}
+- Pushes: any
+- return key B's value from a [valid](jsonspec.md) utf-8 encoded json text A
+- LogicSigVersion >= 6
+
+specify the return type with an immediate arg either as JSONUint64 or JSONString or JSONOjbect.
```
jannotti, line 876: JSONOjbect -> JSONObject (of course, change is in the .go code)

data/transactions/logic/eval.go
```diff
@@ -4060,3 +4062,126 @@ func opBase64Decode(cx *EvalContext) {
...
+			return
+		}
+		// val is not negative
+		if val.String()[0] == '-' {
+			cx.err = fmt.Errorf("JSON value should be a uint64")
+			return
+		}
+		// val > max uint64
+		maxVal := strconv.FormatUint(math.MaxUint64, 10)
+		if val.String() > maxVal {
```
jannotti, line 4161: Isn't this doing a string compare?  That won't be right, I don't think.

data/transactions/logic/eval.go
```diff
@@ -4060,3 +4062,126 @@ func opBase64Decode(cx *EvalContext) {
...
+			cx.err = fmt.Errorf("JSON value range within uint64 range")
+			return
+		}
+		// convert to uint64
+		var bigInt big.Int
+		bigInt.UnmarshalText([]byte(val.String()))
+		stval.Uint = bigInt.Uint64()
```
jannotti, line 4168: Perhaps it's better to do the conversion this way (and detect failures - I assume that's possible?) rather than try to check the string before conversion.  For example I think "1e200" is a valid json number (and it's a float, sort of), but it's bigger that uint64, despite the string check above.

`strconv.ParseUint(val.String(), 10, 64)` might be all we need.

Though if we do that, we need to declare that 1e3 is _not_ acceptable.  Though I'm thinking maybe it should be allowed?  Do people think of that as an int or not?

jasonpaulos: +1 to using `strconv.ParseUint`, since `big.Int`'s `UnmarshalText` also parses hex and binary

data/transactions/logic/eval.go
```diff
@@ -4060,3 +4062,126 @@ func opBase64Decode(cx *EvalContext) {
 	}
 	cx.stack[last].Bytes, cx.err = base64Decode(cx.stack[last].Bytes, encoding)
 }
+func hasDuplicateKeys(dec *json.Decoder) (bool, error) {
```
jasonpaulos, line 4065: Suggestion: make this function take a byte slice instead, since you can't reuse the same decoder after this. Also, it will make unit tests easier.

data/transactions/logic/eval.go
```diff
@@ -4060,3 +4062,126 @@ func opBase64Decode(cx *EvalContext) {
...
+	//get value from json
+	var stval stackValue
+	resp, ok := resp.(map[string]interface{})[key]
+	if !ok {
+		cx.err = fmt.Errorf("key %s not found in JSON text", key)
+		return
+	}
+	expectedType := jsonRefTypeNames[cx.program[cx.pc+1]]
+	switch expectedType {
+	case "JSONString":
```
jasonpaulos, line 4136: nit: compare `cx.program[cx.pc+1]` directly to the constants `JSONString`, `JSONUint64`, and `JSONObject` instead

data/transactions/logic/eval.go
```diff
@@ -4060,3 +4062,126 @@ func opBase64Decode(cx *EvalContext) {
...
+		bigInt.UnmarshalText([]byte(val.String()))
+		stval.Uint = bigInt.Uint64()
+	case "JSONObject":
+		_, ok := resp.(map[string]interface{})
+		if !ok {
+			cx.err = fmt.Errorf("got type %T but expected JSON object", resp)
+			return
+		}
+		var rawJSONText map[string]json.RawMessage
+		err = json.Unmarshal(cx.stack[last].Bytes, &rawJSONText)
```
jasonpaulos, line 4176: Instead of unmarshalling a JSON object again, I suggest you make the following changes to your approach:

1. Since this opcode only takes JSON objects as input (not arrays or primitive values), in your `parseJSON` function, don't decode to an `interface{}`, instead decode to a `map[string]json.RawMessage` (and now there's no need to call `decoder.UseNumber()`)

2. Then call `hasDuplicateKeys`, since at this point you are certain that the input is a JSON object.

3. For `JSONString` and `JSONUint64`, get the appropriate `json.RawMessage` value from the decoded map, then do this:
```go
var value string // change to uint64 for JSONUint64
err := json.Unmarshal(json_raw_message_casted_to_bytes_goes_here, &value)
```
I _think_ when you `json.Unmarshal` to a uint64, it uses `strconv.ParseUint` under the hood, so you should get the same result with less code. If not, or if you want to be more explicit, you could decode to a `json.Number` instead.

4. For `JSONObject`, no need to unmarshal anything else, just return the `json.RawMessage` value directly.

As an added bonus, this should be more efficient, since the original object is a `map[string]json.RawMessage`, meaning the other keys that aren't being referenced won't have their values parsed into in memory objects.

data/transactions/logic/eval_test.go
```diff
@@ -5117,3 +5119,267 @@ func TestOpBase64Decode(t *testing.T) {
...
+			json_ref JSONObject;
+			byte "{\"key40\": 10, \"key40\": \"10\"}"
+			==
+			`,
+		},
+		{
+			source: `byte  "{\"rawId\": \"responseId\",\"id\": \"0\",\"response\": {\"attestationObject\": \"based64url_encoded_buffer\",\"clientDataJSON\":  \" based64url_encoded_client_data\"},\"getClientExtensionResults\": {},\"type\": \"public-key\"}"; 
+			byte "response"; 
+			json_ref JSONObject; 
+			byte "{\"attestationObject\": \"based64url_encoded_buffer\",\"clientDataJSON\":  \" based64url_encoded_client_data\"}" // object as it appeared in input
```
jasonpaulos, line 5259: I'd also suggest testing that a unicode escape sequence remains in the JSON string. E.g. use `\u0061` instead of `a` somewhere. I used this site to find that: https://dencode.com/en/string/unicode-escape

You could also check that unicode escape sequences in the keys get handled properly, like if `\u0072\u0065\u0073\u0070\u006f\u006e\u0073\u0065` was used instead of `response` in the JSON source.

data/transactions/logic/eval.go
shiqizng: > What do we gain with the first test? Does Unmarshal not return a proper error if the json is invalid?
json.Valid returns false for an input like {"key0": "algo"}/*comment*/, but unmarshal doesn't return an error.

data/transactions/logic/README.md
```diff
@@ -166,6 +166,7 @@ various sizes.
 | `extract_uint32` | pop a byte-array A and integer B. Extract a range of bytes from A starting at B up to but not including B+4, convert bytes as big endian and push the uint64 result. If B+4 is larger than the array length, the program fails |
 | `extract_uint64` | pop a byte-array A and integer B. Extract a range of bytes from A starting at B up to but not including B+8, convert bytes as big endian and push the uint64 result. If B+8 is larger than the array length, the program fails |
 | `base64_decode e` | decode X which was base64-encoded using _encoding_ E. Fail if X is not base64 encoded with encoding E |
+| `json_ref r` | return key B's value from a [valid](jsonspec.md) utf-8 encoded json text A |
```
tzaffi, line 169: I recommend also mentioning the file of type `R`

data/transactions/logic/doc.go
```diff
@@ -98,15 +98,16 @@ var opDocByName = map[string]string{
...
+	"global":   "push value from globals to stack",
+	"load":     "copy a value from scratch space to the stack. All scratch spaces are 0 at program start.",
+	"store":    "pop value X. store X to the Ith scratch space",
+	"loads":    "copy a value from the Xth scratch space to the stack.  All scratch spaces are 0 at program start.",
+	"stores":   "pop indexes A and B. store B to the Ath scratch space",
+	"gload":    "push Ith scratch space index of the Tth transaction in the current group",
+	"gloads":   "push Ith scratch space index of the Xth transaction in the current group",
+	"gaid":     "push the ID of the asset or application created in the Tth transaction of the current group",
+	"gaids":    "push the ID of the asset or application created in the Xth transaction of the current group",
+	"json_ref": "return key B's value from a [valid](jsonspec.md) utf-8 encoded json text A",
```
tzaffi, line 110: I recommend also mentioning the field of type R

data/transactions/logic/assembler_test.go
```diff
@@ -2303,3 +2304,9 @@ func TestTxTypes(t *testing.T) {
 	testProg(t, "itxn_begin; byte 0x87123376; itxn_field Amount", 5, expect{3, "...wanted type uint64 got []byte"})
 	testProg(t, "itxn_begin; int 1; itxn_field Amount", 5)
 }
+
+func TestJsonRefAsm(t *testing.T) {
+	testProg(t, `byte  "{\"key0\": 1}"; byte \"key0\"; json_ref JSONUint64;`, 5, expect{3, "json_ref opcode was introduced in TEAL v6"})
```
tzaffi, line 2309: nice!

data/transactions/logic/eval.go
```diff
@@ -4060,3 +4061,109 @@ func opBase64Decode(cx *EvalContext) {
 	}
 	cx.stack[last].Bytes, cx.err = base64Decode(cx.stack[last].Bytes, encoding)
 }
+func hasDuplicateKeys(jsonText []byte) (bool, error) {
+	dec := json.NewDecoder(bytes.NewReader(jsonText))
+	keys := make(map[string]bool)
+	var value interface{}
+	_, err := dec.Token() // {
```
tzaffi, line 4068:
```suggestion
	_, err := dec.Token()
```

data/transactions/logic/jsonspec.md
```diff
@@ -0,0 +1,110 @@
+## JSON Spec
```
tzaffi, line 1: Great idea!

data/transactions/logic/jsonspec.md
```diff
@@ -0,0 +1,110 @@
...
+#### Invalid JSON text
+```json
+\uFEFF{"key0": 1}
+```
+```json
+{"key0": "\uFF"}
+```
+### Object
+#### duplicate key
+Duplicate key at top level is treated as error and ignored when it is in nested object.
```
tzaffi, line 21:
```suggestion
Duplicate keys at the top level result in an error; However, duplicate keys nested at a lower level are ignored. 
```

data/transactions/logic/jsonspec.md
```diff
@@ -0,0 +1,110 @@
...
+{"key0": "\uFF"}
+```
+### Object
+#### duplicate key
+Duplicate key at top level is treated as error and ignored when it is in nested object. 
+#### Invalid JSON text
+```json
+{"key0": 1,"key0": 2}
+```
+#### Valid JSON text
```
tzaffi, line 26: I hesitate to call this "valid" even if it technically is for the purposes of the opcode. Maybe instead `Acceptable JSON text` ...

data/transactions/logic/jsonspec.md
```diff
@@ -0,0 +1,110 @@
...
+```json
+{"key0": 1,"key1": {"key2":2,"key2":"10"}}
+```
+### Numbers
+#### Range 
+- Only integers between 0 and 2^64-1 are accepted
+- All other values are treated as error.
```
tzaffi, line 33:
```suggestion
- All other values result in an error.
```

data/transactions/logic/jsonspec.md
```diff
@@ -0,0 +1,110 @@
...
+
+#### Special Values
+- `null`, `true`, `false` are the only accepted special values. 
+- other spcial values such as `NaN`,`+Inf`,`-Inf` are not accepted
+
+#### Exponential Notation
+Exponential notation not accepted
```
tzaffi, line 40:
```suggestion
Exponential notation is not accepted
```

data/transactions/logic/jsonspec.md
```diff
@@ -0,0 +1,110 @@
...
+#### Invalid JSON text
+```json
+{"key": 1.2E-6}
+```
+```json
+{"key": 0.2E+8}
+```
+##### Hex values
+Hex values not accepted
```
tzaffi, line 49:
```suggestion
Hex values are not accepted
```

data/transactions/logic/jsonspec.md
```diff
@@ -0,0 +1,110 @@
...
+
+#### Invalid JSON text
+```json
+{"key0": 0x1}
+```
+```json
+{"key0": 0xFF}
+```
+### Trailing Commas
+Trailing commas not accepted.
```
tzaffi, line 59:
```suggestion
Trailing commas are not accepted.
```

data/transactions/logic/jsonspec.md
```diff
@@ -0,0 +1,110 @@
...
+#### Invalid JSON text
+```json
+{"key": 4160,,,}
+```
+```json
+{"key": "algo",,,}
+```
+### Comment
+Comment block not accepted.
```
tzaffi, line 68:
```suggestion
Comment blocks are not accepted.
```

data/transactions/logic/jsonspec.md
```diff
@@ -0,0 +1,110 @@
...
+```
+```json
+{"key0": "algo"}/*comment*/
+```
+```json
+{"key0": [1,/*comment*/,3]}
+```
+### White Spaces
+- space, tab(`\t`), new line(`\n`) and carriage return(`\r`) are allowed
+- form feed(`\f`) not allowed
```
tzaffi, line 81:
```suggestion
- form feed(`\f`) is not allowed
```

data/transactions/logic/jsonspec.md
```diff
@@ -0,0 +1,110 @@
+## JSON Spec
+
+A valid JSON text must follow the grammar defined in <a href="https://www.rfc-editor.org/rfc/rfc7159.html">RFC7159</a>
+
+Additional specifications used by **json_ref** that are extensions to RFC7159 grammar are listed below.
```
tzaffi, line 5:
```suggestion
Additional specifications used by **json_ref** that are extensions to the RFC7159 grammar are listed below.
```

data/transactions/logic/jsonspec.md
```diff
@@ -0,0 +1,110 @@
...
+
+### File Encoding
+- Only utf-8 encoded are accepted.
+- Byte order mark(BOM),"\uFEFF", is not allowed at the beginning of a JSON text,
```
tzaffi, line 9:
```suggestion
- The byte order mark (BOM), "\uFEFF", is not allowed at the beginning of a JSON text,
```

data/transactions/logic/jsonspec.md
```diff
@@ -0,0 +1,110 @@
...
+{"key0": [1,/*comment*/,3]}
+```
+### White Spaces
+- space, tab(`\t`), new line(`\n`) and carriage return(`\r`) are allowed
+- form feed(`\f`) not allowed
+
+### Escaped Characters
+
+- control chars (U+0000 - U+001F) must be escaped
+- surrogate pair is accepted
```
tzaffi, line 86:
```suggestion
- surrogate pairs are accepted  
```

tzaffi (commented):
Looks pretty solid to me whith many great test cases! I just had some cosmetic suggestions around wording.

data/transactions/logic/eval.go
```diff
@@ -4118,67 +4122,45 @@ func opJSONRef(cx *EvalContext) {
...
-		}
-		// val > max uint64
-		maxVal := strconv.FormatUint(math.MaxUint64, 10)
-		if val.String() > maxVal {
-			cx.err = fmt.Errorf("JSON value range within uint64 range")
+	case JSONString:
+		var value string // change to uint64 for JSONUint64
+		err := json.Unmarshal(parsed[key], &value)
+		if err != nil {
+			cx.err = fmt.Errorf("%v", err)
```
jasonpaulos, line 4144: nit: no need for `// change to uint64 for JSONUint64` comment

And I think `cx.err = err` is equivalent to `cx.err = fmt.Errorf("%v", err)`

data/transactions/logic/eval_test.go
```diff
@@ -5259,6 +5255,13 @@ func TestOpJSONRef(t *testing.T) {
 			byte "{\"attestationObject\": \"based64url_encoded_buffer\",\"clientDataJSON\":  \" based64url_encoded_client_data\"}" // object as it appeared in input
 			==`,
 		},
+		{
+			source: `byte  "{\"rawId\": \"responseId\",\"id\": \"0\",\"response\": {\"attestationObject\": \"based64url_encoded_buffer\",\"clientD/u0061taJSON\":  \" based64url_encoded_client_data\"},\"getClientExtensionResults\": {},\"type\": \"public-key\"}";
```
jasonpaulos, line 5259:
```suggestion
			source: `byte  "{\"rawId\": \"responseId\",\"id\": \"0\",\"response\": {\"attestationObject\": \"based64url_encoded_buffer\",\"clientD\\u0061taJSON\":  \" based64url_encoded_client_data\"},\"getClientExtensionResults\": {},\"type\": \"public-key\"}"; 
```
Isn't the slash supposed to be the other way?

data/transactions/logic/jsonspec.md
```diff
@@ -0,0 +1,109 @@
...
+### File Encoding
+- Only utf-8 encoded are accepted
+- The byte order mark (BOM), "\uFEFF", is not allowed at the beginning of a JSON text
+- Raw non-unicode characters not accepted
+#### Invalid JSON text
+```json
+\uFEFF{"key0": 1}
+```
+```json
+{"key0": "\uFF"}
```
jasonpaulos, line 16: Very minor, but I think there needs to be 4 characters after every `\u` for it to be valid JSON. E.g. `\u00FF` would work instead

```
> JSON.parse('{"key0": "\uFF"}')
JSON.parse('{"key0": "\uFF"}')
                      ^^^^^^

Uncaught SyntaxError: Invalid Unicode escape sequence
```

shiqizng: right. that is an example of invalid json text.

jasonpaulos: Oh, my mistake

data/transactions/logic/assembler.go
```diff
@@ -1266,6 +1266,28 @@ func assembleBase64Decode(ops *OpStream, spec *OpSpec, args []string) error {
...
+		return ops.errorf("%s expects one argument", spec.Name)
+	}
+
+	jsonSpec, ok := jsonRefSpecByName[args[0]]
+	if !ok {
+		return ops.errorf("%s unsupported JSON value type: %#v", spec.Name, args[0])
+	}
+	if jsonSpec.version > ops.Version {
+		//nolint:errcheck // we continue to maintain typestack
+		ops.errorf("%s %s available in version %d. Missed #pragma version?", spec.Name, args[0], jsonSpec.version)
```
algorandskiy, line 1280: return here and remove nolint:errcheck

data/transactions/logic/eval.go
```diff
@@ -4060,3 +4061,109 @@ func opBase64Decode(cx *EvalContext) {
...
 	cx.stack[last].Bytes, cx.err = base64Decode(cx.stack[last].Bytes, encoding)
 }
+func hasDuplicateKeys(jsonText []byte) (bool, error) {
+	dec := json.NewDecoder(bytes.NewReader(jsonText))
+	keys := make(map[string]bool)
+	var value interface{}
+	_, err := dec.Token()
+	if err != nil {
+		return false, err
+	}
```
algorandskiy, line 4071: why is it not inside the `for dec.More {}` loop below?

data/transactions/logic/eval.go
```diff
@@ -4060,3 +4061,109 @@ func opBase64Decode(cx *EvalContext) {
...
+	for dec.More() {
+		// Get JSON key
+		key, err := dec.Token()
+		if err != nil {
+			return false, err
+		}
+		// end of json
+		if key == '}' {
```
algorandskiy, line 4079: does it mean no nested objects?

data/transactions/logic/eval.go
```diff
@@ -4060,3 +4061,109 @@ func opBase64Decode(cx *EvalContext) {
...
+	}
+	// decode text
+	decoder := json.NewDecoder(bytes.NewReader(jsonText))
+	var parsed map[string]json.RawMessage
+	err := decoder.Decode(&parsed)
+	if err != nil {
+		return nil, err
+	}
+	// check for duplicate keys
+	hasDuplicates, err := hasDuplicateKeys(jsonText)
```
algorandskiy, line 4108: we kind of decoding json twice, `decoder.Decode(&parsed)` and inside `hasDuplicateKeys`, not good

data/transactions/logic/jsonspec.md
```diff
@@ -0,0 +1,95 @@
...
+
+Additional specifications used by **json_ref** that are extensions to the RFC7159 grammar are listed below.
+
+### File Encoding
+- Only utf-8 encoded are accepted
+- The byte order mark (BOM), "\uFEFF", is not allowed at the beginning of a JSON text
+- Raw non-unicode characters not accepted
+#### Invalid JSON text
+```json
+\uFEFF{"key0": 1}
```
algorandskiy, line 13: I think BOM is not needed here

algorandskiy (commented):
I like the PR in general, but also few notes.

data/transactions/logic/eval.go
shiqizng: right. it's only checking duplicate keys at top level. json_ref only retrieve value one level at a time.

data/transactions/logic/eval.go
shiqizng: Unfortunately, I don't see a way to check for duplicate keys without parsing it.

data/transactions/logic/jsonspec.md
shiqizng: It's an example of invalid json text.

data/transactions/logic/eval.go
algorandskiy: Two options: 1) Can't `parsed` map be checked? 2) use the `dec.More` + `dec.Token` for parsing and counting.

shiqizng: I see. I think option 1 should work.

shiqizng: actually, I went with option 2 instead.

data/transactions/logic/eval.go
```diff
@@ -4060,3 +4061,102 @@ func opBase64Decode(cx *EvalContext) {
...
+	case JSONUint64:
+		var value uint64
+		err := json.Unmarshal(parsed[key], &value)
+		if err != nil {
+			cx.err = err
+			return
+		}
+		stval.Uint = value
+	case JSONObject:
+		var value map[string]interface{}
```
jasonpaulos, line 4150: nit: you can avoid parsing all child keys by decoding to a `map[string]json.RawMessage`

data/transactions/logic/eval_test.go
```diff
@@ -5117,3 +5117,291 @@ func TestOpBase64Decode(t *testing.T) {
...
+		require.NoError(t, err)
+		require.True(t, pass)
+	}
+
+	failedCases := []struct {
+		source string
+		error  string
+	}{
+		{
+			source: `byte  "{\"key0\": 1 }"; byte "key0"; json_ref JSONString;`,
```
jasonpaulos, line 5299: I don't think you have failing test cases for the following yet?
* Calling `json_ref` on a JSON array, e.g. `byte "[1,2,3]"; byte "key"; json_ref JSONUint64`
* Calling `json_ref` on a JSON integer, e.g. `byte "2"; byte "key"; json_ref JSONUint64`
* Calling `json_ref` on a JSON null, e.g. `byte "null"; byte "key"; json_ref JSONUint64`
* Calling `json_ref` on a JSON string, e.g. `byte "\"string\""; byte "key"; json_ref JSONString`

I'm pretty sure they'll fail, but since `hasDuplicateKeys` does some interesting things with the tokenizer, I'm worried it might panic instead of failing gracefully

shiqizng: good point. tests added.

jasonpaulos (approved):

algorandskiy:
Why not merged?

algoanne:
@algorandskiy this opcode (along with the base 54 decode opcode, and the additional ecdsa curve) are being added to support FIDO2 authentication. But, how that is achieved isn't yet 100% figured out, so we are putting these three pieces of work aside until it is - we're giving them to a team who will be working on the FIDO2 integration so they can use them, and once we have a complete solution we'll merge to master.

PR #3350, opened 2021-12-27
Integrate stateproof with the updated account manager

Aharonee:


agreement/keyManager_test.go
```diff
@@ -17,6 +17,7 @@
 package agreement
 
 import (
+	"github.com/algorand/go-algorand/crypto/merklekeystore"
```
algoidan, line 20: nits: reorder imports

data/account/participationRegistry.go
```diff
@@ -863,29 +894,62 @@ func (db *participationDB) GetAll() []ParticipationRecord {
 	return results
 }
 
+// TODO: improve performance somehow but not querying the DB if stateproof key should not
+// exist for this round
+// TODO: make sure it returns a nil stateproof if none exist for this round
 // GetForRound fetches a record with all secrets for a particular round.
```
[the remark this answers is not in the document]
algoidan: + lets validate this function with tests.
1- a test that request a round in which we don't have stateproof key
2- a test that request keys for an address that doesn't have stateproof key (didn't reregiest)

node/node.go
```diff
@@ -970,12 +970,41 @@ func (node *AlgorandFullNode) loadParticipationKeys() error {
...
+			// make sure that all stateproof data (with are not the keys per round)
+			// are being store to the registry in that point
 			added := node.accountManager.AddParticipation(part)
 			if added {
 				node.log.Infof("Loaded participation keys from storage: %s %s", part.Address(), info.Name())
 			} else {
 				part.Close()
+				continue
 			}
+			// create a const value Number of records in one fetch
```
algoidan, line 982: this is a pretty long function. could we split it?

node/node.go
```diff
@@ -970,12 +970,41 @@ func (node *AlgorandFullNode) loadParticipationKeys() error {
...
+			// insert into the registry X rounds of keys via account manager
+			// all in for loop until no more keys in db
+			partID := part.ID()
+			numKeys := part.Participation.StateProofSecrets.CountKeys(part.Store)
+			for i := uint64(0); i < uint64(numKeys); i++ {
```
algoidan, line 987: what if this loop breaks?(ie node crashes)
can we recover from it?
do we have a test?

algoidan (commented):
Overall looks very good!
there are some code cleanup needed and tests

data/account/participationRegistry.go
```diff
@@ -82,22 +83,42 @@ type ParticipationRecord struct {
...
-type StateProofKey []byte
+// StateProofKey defined the type used for the compact certificate signing key
+type StateProofKey crypto.GenericSigningKey
 
 // ParticipationRecordForRound adds in the per-round state proof key.
 type ParticipationRecordForRound struct {
 	ParticipationRecord
 
-	StateProof StateProofKey
+	StateProof *merklekeystore.SignerInRound
```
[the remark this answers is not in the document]
Aharonee: But it contains more than the secret keys. Maybe StateProofSecrets?

crypto/merklekeystore/keystore.go
```diff
@@ -57,7 +58,25 @@ type (
 		// in which they will be dumped into database and disposed of.
 		// non-exported fields to prevent msgpack marshalling
 		signatureAlgorithms []crypto.GenericSigningKey
-		keyStore            PersistentKeystore
+		//keyStore            PersistentKeystore
```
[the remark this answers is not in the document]
Aharonee: They're not ephemeral though, so maybe just keys?

node/node.go
Aharonee: Maybe I should just return error instead since I'm counting keys before starting to iterate, so the reason for this `break` not longer exists.

node/node.go
Aharonee: Inner function that I could test separately maybe

data/account/participationRegistry.go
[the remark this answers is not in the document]
algoidan: yeap
go for it

data/account/participationRegistry.go
```diff
@@ -863,29 +894,62 @@ func (db *participationDB) GetAll() []ParticipationRecord {
 	return results
 }
 
+// TODO: improve performance somehow but not querying the DB if stateproof key should not
+// exist for this round
+// TODO: make sure it returns a nil stateproof if none exist for this round
 // GetForRound fetches a record with all secrets for a particular round.
 func (db *participationDB) GetForRound(id ParticipationID, round basics.Round) (ParticipationRecordForRound, error) {
```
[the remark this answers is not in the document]
Aharonee: Why?

data/account/participationRegistry.go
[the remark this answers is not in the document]
Aharonee: Added a unit test

crypto/merklekeystore/PersistentKeystore.go
```diff
@@ -68,87 +83,57 @@ func (p *PersistentKeystore) Persist(keys []crypto.GenericSigningKey, firstValid
...
+		row := tx.QueryRow("SELECT COUNT(*) FROM StateProofKeys")
+		err := row.Scan(&count)
+		if err != nil {
+			return fmt.Errorf("failed to count rows in table StateProofKeys : %w", err)
+		}
+		return nil
 	})
 	if err != nil {
-		return PersistentKeystore{}, err
+		return -1
```
algoidan, line 136: this is tricky.
it looks like the caller here convert this number into uint64.
we will have a problem with that case.

Aharonee: Maybe just return 0 instead and log error? In a typical flow of using this function the number of keys will be used for iteration, so 0 would skip the for-loop altogether which should be fine.

crypto/merklekeystore/PersistentKeystore.go
```diff
@@ -68,87 +83,56 @@ func (p *PersistentKeystore) Persist(keys []crypto.GenericSigningKey, firstValid
...
+		row := tx.QueryRow("SELECT COUNT(*) FROM StateProofKeys")
+		err := row.Scan(&count)
+		if err != nil {
+			return fmt.Errorf("failed to count rows in table StateProofKeys : %w", err)
+		}
+		return nil
 	})
 	if err != nil {
-		return PersistentKeystore{}, err
+		return 0
```
algoidan, line 135: I think we should return an error here and stop the migration process in the caller.

crypto/merklekeystore/PersistentKeystore_test.go
```diff
@@ -0,0 +1,70 @@
...
+	a.NoError(keystore.Persist(*store))
+
+	key, rnd, err := keystore.FetchKey(interval*1, *store)
+	a.Equal(keystore.GetKey(rnd), key)
+
+	key, rnd, err = keystore.FetchKey(interval*2, *store)
+	a.Equal(keystore.GetKey(rnd), key)
+
+	key, rnd, err = keystore.FetchKey(interval*5, *store)
+	a.Equal(keystore.GetKey(rnd), key)
```
algoidan, line 50: try to fetch non-existing key
+ add some tests for `count`

crypto/merklekeystore/keystore.go
```diff
@@ -46,18 +47,35 @@ type (
...
+	Signer struct {
+		SigningKey *crypto.GenericSigningKey
+
+		// The round for which this SigningKey is related to
+		Round uint64
+
+		SignerContext
+	}
+
+	// SignerContext contains all the public immutable data related to merklekeystore.Keystore
```
algoidan, line 76: not all the data is public + can we make it private struct?

crypto/merklekeystore/keystore_test.go
```diff
@@ -353,229 +324,90 @@ func TestIncorrectByteSignature(t *testing.T) {
...
-	for i := uint64(50); i <= 60; i++ {
-		sig, err := signer.Sign(genHashableForTest(), i)
-		if i%11 != 0 {
-			a.Error(err)
-			continue
-		}
-		a.NoError(signer.GetVerifier().Verify(1, i, 11, genHashableForTest(), sig))
-	}
-}
+// TODO: test marshaling SignerContext instead of Keystore
```
algoidan, line 341: lets do it :)

data/account/participationRegistry.go
```diff
@@ -82,20 +83,40 @@ type ParticipationRecord struct {
...
 
 // IsZero returns true if the object contains zero values.
 func (r ParticipationRecordForRound) IsZero() bool {
-	return r.StateProof == nil && r.ParticipationRecord.IsZero()
+	return r.StateProofSecrets == nil && r.ParticipationRecord.IsZero()
+}
+
+// VotingSigner returns the voting secrets associated with this Participation account,
+// together with the KeyDilution value.
+func (r *ParticipationRecordForRound) VotingSigner() crypto.OneTimeSigner {
```
algoidan, line 103: can we remove `participation.VotingSigner`?

data/account/participationRegistry.go
```diff
@@ -82,20 +83,40 @@ type ParticipationRecord struct {
...
+	return crypto.OneTimeSigner{
+		OneTimeSignatureSecrets: r.Voting,
+		OptionalKeyDilution:     r.KeyDilution,
+	}
+}
+
+// OverlapsInterval returns true if the partkey is valid at all within the range of rounds (inclusive)
+func (r *ParticipationRecordForRound) OverlapsInterval(first, last basics.Round) bool {
```
algoidan, line 111: this is function is not used, does it make sense ?

data/account/participationRegistry.go
```diff
@@ -460,7 +486,11 @@ func (db *participationDB) insertInner(record Participation, id ParticipationID)
 		voting := record.Voting.Snapshot()
 		rawVoting = protocol.Encode(&voting)
 	}
-	// PKI TODO: Extract state proof from record.
+
+	// This does not contain secrets! only the public immutable data
```
algoidan, line 490: not just public data.

node/node.go
```diff
@@ -970,11 +970,38 @@ func (node *AlgorandFullNode) loadParticipationKeys() error {
...
+			partID := part.ID()
+			numKeys := part.Participation.StateProofSecrets.CountKeys(part.Store)
+			keys := make(map[uint64]account.StateProofKey, numKeys)
+			for i := uint64(0); i < uint64(numKeys); i++ {
+				// TODO: should we add a method to FetchAllKeys instead?
+				key, round, err := part.Participation.StateProofSecrets.FetchKey(i, part.Store)
+				if err != nil {
+					return err
+				}
+				if key == nil {
```
algoidan, line 994: why do we need it, we know the exact number of keys

data/account/participationRegistry.go
Aharonee: It is used by:
`asyncPseudonode.makeVotes
asyncPseudonode.makeProposals
AlgorandFullNode.MakePrioResponse`,
as I replaced the `Participation` type used there.

data/account/participationRegistry.go
```diff
@@ -82,22 +83,42 @@ type ParticipationRecord struct {
 	Voting *crypto.OneTimeSignatureSecrets
 }
 
-// StateProofKey is a placeholder for the real state proof key type.
-// PKI TODO: Replace this with a real object.
-type StateProofKey []byte
+// StateProofKey defined the type used for the compact certificate signing key
+type StateProofKey crypto.GenericSigningKey
```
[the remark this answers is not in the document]
Aharonee: If tomorrow we choose to replace the GenericSigningKey type with some other (non generic for example, which seems reasonable), we'll have to search&replace all over the code base.

data/account/participationRegistry.go
algoidan: you are right
so remove  `Participation.VotingSigner`

PR #3352, opened 2021-12-27
Update `goal app method` handling of args and return values

jasonpaulos:
## Summary

This PR changes how `goal app method` handles args and return values in order to address the latest ARC-4 changes and a bug fix. Specifically:
1. Only compresses trailing arguments into a tuple if the number of arguments exceeds 15, per https://github.com/algorandfoundation/ARCs/pull/53
2. Fix a bug when compressing trailing arguments into a tuple, described here: https://github.com/algorand/go-algorand-sdk/pull/273#discussion_r773478473
3. Only inspects the final logged value to find a method's return value, per https://github.com/algorandfoundation/ARCs/pull/57

## Test Plan

Unit test added, and added an e2e test for the `empty()void` method call.

### DIFF
--- a/cmd/goal/application.go
+++ b/cmd/goal/application.go
@@ -1369,36 +1369,29 @@ var methodAppCmd = &cobra.Command{
 				return
 			}
 
-			// specify the return hash prefix
-			hashRet := sha512.Sum512_256([]byte("return"))
-			hashRetPrefix := hashRet[:4]
+			// the 4-byte prefix for logged return values
+			var abiReturnHash = []byte{0x15, 0x1f, 0x7c, 0x75}
 
-			var abiEncodedRet []byte
-			foundRet := false
-			if resp.Logs != nil {
-				for i := len(*resp.Logs) - 1; i >= 0; i-- {
-					retLog := (*resp.Logs)[i]
-					if bytes.HasPrefix(retLog, hashRetPrefix) {
-						abiEncodedRet = retLog[4:]
-						foundRet = true
-						break
-					}
-				}
+			if resp.Logs == nil || len(*resp.Logs) == 0 {
+				reportErrorf("method %s succeed but did not log a return value", method)
 			}
 
-			if !foundRet {
-				reportErrorf("cannot find return log for abi type %s", retTypeStr)
+			lastLog := (*resp.Logs)[len(*resp.Logs)-1]
+			if !bytes.HasPrefix(lastLog, abiReturnHash) {
+				reportErrorf("method %s succeed but did not log a return value", method)
 			}
 
-			decoded, err := retType.Decode(abiEncodedRet)
+			rawReturnValue := lastLog[len(abiReturnHash):]
+			decoded, err := retType.Decode(rawReturnValue)
 			if err != nil {
-				reportErrorf("cannot decode return value %v: %v", abiEncodedRet, err)
+				reportErrorf("method %s succeed but its return value could not be decoded. The error is: %s", method, err)
 			}
 
 			decodedJSON, err := retType.MarshalToJSON(decoded)
 			if err != nil {
-				reportErrorf("cannot marshal returned bytes %v to JSON: %v", decoded, err)
+				reportErrorf("method %s succeed but its return value could not be converted to JSON. The error is: %s", method, err)
 			}
+
 			fmt.Printf("method %s succeeded with output: %s\n", method, string(decodedJSON))
 		}
 	},
--- a/cmd/goal/application.go
+++ b/cmd/goal/application.go
@@ -22,6 +22,7 @@ import (
 	"encoding/base32"
 	"encoding/base64"
 	"encoding/binary"
+	"encoding/hex"
 	"fmt"
 	"os"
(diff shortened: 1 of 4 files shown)


data/abi/abi_encode.go
```diff
@@ -495,16 +503,31 @@ func ParseArgJSONtoByteSlice(argTypes []string, jsonArgs []string, applicationAr
...
+			tupleValues[i] = []byte(jsonArg)
+		}
+
+		remainingJSON, err := json.Marshal(tupleValues)
 		if err != nil {
 			return err
 		}
-		abiTypes = append(abiTypes[:14], compactedType)
 
-		remainingJSON := "[" + strings.Join(jsonArgs[14:], ",") + "]"
```
jasonpaulos, line 506: This string manipulation code isn't wrong, but it seems safer to call `json.Marshal` on a `[]json.RawMessage`, similar to other code in `data/abi/abi_json.go`

cmd/goal/application.go
```diff
@@ -1369,36 +1369,29 @@ var methodAppCmd = &cobra.Command{
...
-				for i := len(*resp.Logs) - 1; i >= 0; i-- {
-					retLog := (*resp.Logs)[i]
-					if bytes.HasPrefix(retLog, hashRetPrefix) {
-						abiEncodedRet = retLog[4:]
-						foundRet = true
-						break
-					}
-				}
+			// the 4-byte prefix for logged return values
+			var abiReturnHash = []byte{0x15, 0x1f, 0x7c, 0x75}
```
jannotti, line 1373: Why the change from calculating it?

cmd/goal/application.go
```diff
@@ -1369,36 +1369,29 @@ var methodAppCmd = &cobra.Command{
...
+			if !bytes.HasPrefix(lastLog, abiReturnHash) {
+				reportErrorf("method %s succeed but did not log a return value", method)
 			}
 
-			decoded, err := retType.Decode(abiEncodedRet)
+			rawReturnValue := lastLog[len(abiReturnHash):]
+			decoded, err := retType.Decode(rawReturnValue)
 			if err != nil {
-				reportErrorf("cannot decode return value %v: %v", abiEncodedRet, err)
+				reportErrorf("method %s succeed but its return value could not be decoded. The error is: %s", method, err)
```
jannotti, line 1387: Do you know if the error contains the bytes it code not decode?  If not, maybe that's worth printing, with hex.EncodeToString

data/abi/abi_encode.go
```diff
@@ -478,6 +479,13 @@ func decodeTuple(encoded []byte, childT []Type) ([]interface{}, error) {
 	return values, nil
 }
 
+// maxAppArgs is the maximum number of arguments for an application call transaction
```
jannotti, line 482: Let's put a comment in explaining that we have this as a constant here in order to obey ARC4, else one would expect to get it from consensus parameters. But (I suppose) if we do update the arg max to, say, 32, ARC4 apps would still want to encode this way, despite the change.

jannotti (approved):
I don't feel terribly strongly about any of my recommendations, so marking Approved now.   Let me know if you want to make any updates, then I'll merge.

cmd/goal/application.go
ahangsu: I guess we decided that the return format is first 4 byte of hash of "return", and we wrote the 4 bytes down in each SDK implementations, so we don't have to recalculate it again.

I think it is okay to define the pre-calculated prefix, anything against that?

data/abi/abi_encode.go
```diff
@@ -495,16 +503,31 @@ func ParseArgJSONtoByteSlice(argTypes []string, jsonArgs []string, applicationAr
 		return fmt.Errorf("input argument number %d != method argument number %d", len(jsonArgs), len(abiTypes))
 	}
 
-	// change the input args to be 1 - 14 + 15 (compacting everything together)
-	if len(jsonArgs) > 14 {
-		compactedType, err := MakeTupleType(abiTypes[14:])
+	// Up to 16 app arguments can be passed to app call. First is reserved for method selector,
+	// and the rest are for method call arguments. But if more than 15 method call arguments
+	// are present, then the 14th+ are placed in a tuple in the last app argument slot
```
ahangsu, line 508:
```suggestion
	// are present, then the method arguments after the 14th are placed in a tuple in the last app argument slot
```

ahangsu: I don't feel strong here, given that @jannotti previously suggested that we might want to bump the number of max-argument-num from 16 to 32, then all the comments should also be updated later...

cmd/goal/application.go
jasonpaulos: The SDKs all hardcode this value, so I didn't see a need to calculate it fresh every time.

ahangsu (approved):
LGTM

cmd/goal/application.go
jasonpaulos: Just added a comment explaining where this values comes from

ahangsu (approved):

PR #3353, opened 2021-12-28
Support app creation in `goal app method`

### LINKED ISSUE #3332: Support app creation in `goal app method`
## Summary

When `goal app method` was originally added we decided to not support app creation transactions because it was undecided how the ABI would work with them. Now, https://github.com/algorandfoundation/ARCs/pull/57 has made it clear ABI method calls should work on creation as well.

## Scope

Allow `goal app method` to be able to create apps. To do this, we need to:
* Somehow indicate an app creation. Initial idea: this could be done with specifying 0 for the app ID, or with an explicit `--create` flag.
* Accept the approval program, clear program, schema counts, and extra pages value for the creation transaction
* Somehow return the created app ID in addition to the method return value

jasonpaulos:
## Summary

Supports app creation in `goal app method`. To create an app, the `--create` flag must be passed instead of the `--app-id` flag. Additionally flags for schema sizes, extra pages, and programs should be passed. After the transaction is confirmed, `goal` will print out the created app ID in addition to the ABI return value.

Here's an example:

```bash
$ goal app method --method "create(uint64)uint64" --arg "1234" --create --approval-prog $APPROVALPROG --clear-prog $CLEARPROG --global-byteslices 0 --global-ints 0 --local-byteslices 1 --local-ints 0 --extra-pages 0 --from WF3PXLVCDHILATGW6P7BOLC7TJE7TOLZM2BQCS24KKX2B4FGMG5Y2NDHWA
Attempting to create app (approval size 500, hash EZD2SJS7UJI45CGG4GEVMCTPZCJ7GQNDUG74GU74C6WKOD2TEFAA; clear size 5, hash YOE6C22GHCTKAN3HU4SE5PGIPN5UKXAJTXCQUPJ3KKF5HOAH646A)
Issued 1 transaction(s):
Issued transaction from account WF3PXLVCDHILATGW6P7BOLC7TJE7TOLZM2BQCS24KKX2B4FGMG5Y2NDHWA, txid 2LOUCI2XOG6XDC6JZWPTVPWHPTF53JSABIEZDBBZGRWII4VM752Q (fee 1000)
Transaction 2LOUCI2XOG6XDC6JZWPTVPWHPTF53JSABIEZDBBZGRWII4VM752Q still pending as of round 2
Transaction 2LOUCI2XOG6XDC6JZWPTVPWHPTF53JSABIEZDBBZGRWII4VM752Q still pending as of round 3
Transaction 2LOUCI2XOG6XDC6JZWPTVPWHPTF53JSABIEZDBBZGRWII4VM752Q committed in round 4
Created app with app index 2
method create(uint64)uint64 succeeded with output: 2468
```

Closes #3332 

## Test Plan

Added e2e test

### DIFF
--- a/cmd/goal/application.go
+++ b/cmd/goal/application.go
@@ -1394,7 +1394,7 @@ var methodAppCmd = &cobra.Command{
 		var txid string
 		for _, stxn := range signedTxnGroup {
 			txid = stxn.Txn.ID().String()
-			reportInfof("\tIssued transaction from account %s, txid %s (fee %d)", stxn.Txn.Sender, txid, stxn.Txn.Fee.Raw)
+			reportInfof("Issued transaction from account %s, txid %s (fee %d)", stxn.Txn.Sender, txid, stxn.Txn.Fee.Raw)
 		}
 
 		if !noWaitAfterSend {
--- a/cmd/goal/application.go
+++ b/cmd/goal/application.go
@@ -46,9 +46,9 @@ var (
 	approvalProgFile string
 	clearProgFile    string
 
-	method         string
-	methodArgs     []string
-	methodCreation bool
+	method           string
+	methodArgs       []string
+	methodCreatesApp bool
 
 	approvalProgRawFile string
 	clearProgRawFile    string
@@ -123,7 +123,7 @@ func init() {
 	methodAppCmd.Flags().StringVar(&method, "method", "", "Method to be called")
 	methodAppCmd.Flags().StringArrayVar(&methodArgs, "arg", nil, "Args to pass in for calling a method")
 	methodAppCmd.Flags().StringVar(&onCompletion, "on-completion", "NoOp", "OnCompletion action for application transaction")
-	methodAppCmd.Flags().BoolVar(&methodCreation, "create", false, "Create an application in this method call")
+	methodAppCmd.Flags().BoolVar(&methodCreatesApp, "create", false, "Create an application in this method call")
 	methodAppCmd.Flags().Uint64Var(&globalSchemaUints, "global-ints", 0, "Maximum number of integer values that may be stored in the global key/value store. Immutable, only valid when passed with --create.")
 	methodAppCmd.Flags().Uint64Var(&globalSchemaByteSlices, "global-byteslices", 0, "Maximum number of byte slices that may be stored in the global key/value store. Immutable, only valid when passed with --create.")
 	methodAppCmd.Flags().Uint64Var(&localSchemaUints, "local-ints", 0, "Maximum number of integer values that may be stored in local (per-account) key/value stores for this app. Immutable, only valid when passed with --create.")
@@ -1194,7 +1194,7 @@ var methodAppCmd = &cobra.Command{
 
 		onCompletionEnum := mustParseOnCompletion(onCompletion)
 
-		if methodCreation {
+		if methodCreatesApp {
 			if appIdx != 0 {
 				reportErrorf("--app-id and --create are mutually exclusive, only provide one")
 			}
@@ -1218,7 +1218,7 @@ var methodAppCmd = &cobra.Command{
 		}
 
 		var approvalProg, clearProg []byte
-		if methodCreation || onCompletionEnum == transactions.UpdateApplicationOC {
+		if methodCreatesApp || onCompletionEnum == transactions.UpdateApplicationOC {
 			approvalProg, clearProg = mustParseProgArgs()
 		}
 
@@ -1382,7 +1382,7 @@ var methodAppCmd = &cobra.Command{
 		}
 
 		// Report tx details to user
-		if methodCreation {
+		if methodCreatesApp {
 			reportInfof("Attempting to create app (approval size %d, hash %v; clear size %d, hash %v)", len(approvalProg), crypto.HashObj(logic.Program(approvalProg)), len(clearProg), crypto.HashObj(logic.Program(clearProg)))
 		} else if onCompletionEnum == transactions.UpdateApplicationOC {
 			reportInfof("Attempting to update app (approval size %d, hash %v; clear size %d, hash %v)", len(approvalProg), crypto.HashObj(logic.Program(approvalProg)), len(clearProg), crypto.HashObj(logic.Program(clearProg)))
@@ -1408,7 +1408,7 @@ var methodAppCmd = &cobra.Command{
 				reportErrorf(err.Error())
 			}
 
-			if methodCreation && resp.ApplicationIndex != nil && *resp.ApplicationIndex != 0 {
+			if methodCreatesApp && resp.ApplicationIndex != nil && *resp.ApplicationIndex != 0 {
 				reportInfof("Created app with app index %d", *resp.ApplicationIndex)
 			}
 
--- a/cmd/goal/application.go
+++ b/cmd/goal/application.go
@@ -46,8 +46,9 @@ var (
 	approvalProgFile string
 	clearProgFile    string
 
-	method     string
-	methodArgs []string
+	method         string
+	methodArgs     []string
+	methodCreation bool
 
 	approvalProgRawFile string
 	clearProgRawFile    string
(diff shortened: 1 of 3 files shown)


test/scripts/e2e_subs/tealprogs/app-abi-method-example.teal
```diff
@@ -1,68 +1,76 @@
-// generated from https://gist.github.com/jasonpaulos/99e4f8a75f2fc2ec9b8073c064530359
+// generated from https://gist.github.com/jasonpaulos/99e4f8a75f2fc2ec9b8073c064530359/b4d37519ccc67383042b6c0fb8b7b26a2e538738
```
jasonpaulos, line 1: The PyTeal diff is here: https://gist.github.com/jasonpaulos/99e4f8a75f2fc2ec9b8073c064530359/revisions#diff-7f2d296a873a322392d91f69eebfc6993a857884a7e55cef9af800a408ef378f

cmd/goal/application.go
```diff
@@ -1365,8 +1408,12 @@ var methodAppCmd = &cobra.Command{
 				reportErrorf(err.Error())
 			}
 
+			if methodCreation && resp.ApplicationIndex != nil && *resp.ApplicationIndex != 0 {
+				reportInfof("Created app with app index %d", *resp.ApplicationIndex)
+			}
```
ahangsu, line 1413: minor question: do we need to concern the existence of `*resp.ApplicationIndex` on the condition of:
- `client.PendingTransactionInformationV2` successfully responded with no error
- creating an application from app-call transaction

If not, it seems okay to lift the check of null-ptr?

test/scripts/e2e_subs/e2e-app-abi-method.sh
```diff
@@ -16,9 +16,18 @@ gcmd="goal -w ${WALLET}"
 
 ACCOUNT=$(${gcmd} account list|awk '{ print $3 }')
 
-printf '#pragma version 2\nint 1' > "${TEMPDIR}/simple.teal"
-PROGRAM=($(${gcmd} clerk compile "${TEMPDIR}/simple.teal"))
-APPID=$(${gcmd} app create --creator ${ACCOUNT} --approval-prog ${DIR}/tealprogs/app-abi-method-example.teal --clear-prog ${TEMPDIR}/simple.teal --global-byteslices 0 --global-ints 0 --local-byteslices 1 --local-ints 0 | grep Created | awk '{ print $6 }')
+printf '#pragma version 2\nint 1' > "${TEMPDIR}/simple-v2.teal"
+printf '#pragma version 3\nint 1' > "${TEMPDIR}/simple-v3.teal"
```
ahangsu, line 20: minor question: why we want 2 versions of approval-prog, but it seems to me that the behavior is the same?

jasonpaulos: `simple-v2.teal` is the clear state program used for creation, and then we update the app to use `simple-v3.teal` later on.

cmd/goal/application.go
jasonpaulos: I'd rather have the specific checks in an if statement than assume the pointer will always be valid, since you never know what could change in the future.

For what it's worth, the existing create app command has a similar check too: https://github.com/algorand/go-algorand/blob/2e09ebfca1ca46231e43cd62c48a1025ce7c0a11/cmd/goal/application.go#L490-L492

test/scripts/e2e_subs/e2e-app-abi-method.sh
ahangsu: That makes sense to me.

test/scripts/e2e_subs/e2e-app-abi-method.sh
```diff
@@ -16,9 +16,18 @@ gcmd="goal -w ${WALLET}"
...
+
+# Create
+RES=$(${gcmd} app method --method "create(uint64)uint64" --arg "1234" --create --approval-prog ${DIR}/tealprogs/app-abi-method-example.teal --clear-prog ${TEMPDIR}/simple-v2.teal --global-byteslices 0 --global-ints 0 --local-byteslices 1 --local-ints 0 --extra-pages 0 --from $ACCOUNT 2>&1 || true)
+EXPECTED="method create(uint64)uint64 succeeded with output: 2468"
+if [[ $RES != *"${EXPECTED}"* ]]; then
+    date '+app-abi-method-test FAIL the method call to optIn(string)string should not fail %Y%m%d_%H%M%S'
```
ahangsu, line 26: the error message seems off to me: `optIn(string)string`?

test/scripts/e2e_subs/e2e-app-abi-method.sh
```diff
@@ -86,6 +95,14 @@ if [[ $RES != *"${EXPECTED}"* ]]; then
     false
 fi
 
+# Update
+RES=$(${gcmd} app method --method "update()void" --on-completion updateapplication --approval-prog ${DIR}/tealprogs/app-abi-method-example.teal --clear-prog ${TEMPDIR}/simple-v3.teal --app-id $APPID --from $ACCOUNT 2>&1 || true)
+EXPECTED="method update()void succeeded"
+if [[ $RES != *"${EXPECTED}"* ]]; then
+    date '+app-abi-method-test FAIL the method call to optIn(string)string should not fail %Y%m%d_%H%M%S'
```
ahangsu, line 102: same here for the error message.

test/scripts/e2e_subs/e2e-app-abi-method.sh
jasonpaulos: Yes, I'll fix it

ahangsu (commented):
Look good to me, waiting on @jannotti to review

cmd/goal/application.go
```diff
@@ -46,8 +46,9 @@ var (
 	approvalProgFile string
 	clearProgFile    string
 
-	method     string
-	methodArgs []string
+	method         string
+	methodArgs     []string
+	methodCreation bool
```
jannotti, line 51: This should really be "appCreation" or "contractCreation", right?

cmd/goal/application.go
```diff
@@ -1172,17 +1178,47 @@ var methodAppCmd = &cobra.Command{
...
+			NumUint:      localSchemaUints,
+			NumByteSlice: localSchemaByteSlices,
+		}
+
+		globalSchema := basics.StateSchema{
+			NumUint:      globalSchemaUints,
+			NumByteSlice: globalSchemaByteSlices,
 		}
 
 		onCompletionEnum := mustParseOnCompletion(onCompletion)
```
jannotti, line 1195: This is obviously a decision that was made before, this PR doesn't change it.  But do you think we should really be requiring on-complete to be set always?  Rather than default to no-op?

jasonpaulos: Actually you don't need to explicitly pass it in, it will default to "NoOp": https://github.com/algorand/go-algorand/blob/1642a3214a12cd49843d62b2e18a1d3e516e7316/cmd/goal/application.go#L125

The e2e test for creation does not pass in `--on-completion`.

cmd/goal/application.go
jasonpaulos: Well I meant it to mean creation for the method subcommand

ahangsu: `methodAppCreation`? (it's always hard to find a good name)

cmd/goal/application.go
jannotti: oh, yeah. I guess I wish there was no "must" on that method name then.  While it's true, it seems like it's trying to convey something important that never comes up in practice because of the default. But I don't care _that_ much.

cmd/goal/application.go
jannotti: My objection is that when you read any code that uses it, it reads wrong.  It'll say, essentially
```
if methodCreation {
 // create the app
}
```
As with the other comment, whatever you want to go with is fine.  Let me know when to merge.

jasonpaulos: Good points, I'll rename to `methodCreatesApp`

cmd/goal/application.go
jasonpaulos: I agree it's a bit misleading, but I'll leave it for now since there are multiple other helper methods in the goal code that use "must" to convey they will either succeed or panic.

jannotti (approved):

Issue #2989, opened 2021-10-03
Add "address" to "ApplicationParams" of algod API schema


ryanRfox:
## Problem

The algod API for `/v2/applications/{application-id}` does not return the address for the application-id. This field is returned by `goal app info` (although it's named [account](https://github.com/algorand/go-algorand/blob/master/cmd/goal/application.go#L978)).

## Solution

- Add "address" to [#/components/schemas/ApplicationParams](https://github.com/algorand/go-algorand/blob/master/daemon/algod/api/algod.oas3.yml#L780)

## Dependencies

None

## Urgency

Higher than _nice to have_; this feels like an oversight. Workaround within the SDKs: calculate the address based on the appID. Sample code in Python:
```
def getAppAddress(appID: int) -> str:
    toHash = b"appID" + appID.to_bytes(8, "big")
    return encoding.encode_address(encoding.checksum(toHash))
```

jannotti:
This should appear as a local function to call, similar to getting logicsig
address from code, not from api.

tzaffi:
I agree with @ryanRfox . Maintaining parity with [goal functionality](https://github.com/tzaffi/go-algorand/blob/master/cmd/goal/application.go#L1036) is nice to have. Though each API can calculate the hash independently, it seems like an application's address is so fundamental that it should be provided by `algod`.

jannotti:
I disagree. a) you don't need that code. The SDKs all already have the function. b) goal isn't the thing that the api should try to emulate. goal is _using_ the api to get the app id and then calculating the address with an internal function. This makes it convenient for use by humans and shell scripts, neither of which can calculate hashes well. But all api consumers can.

winder:
I believe the APIs are now consistent because the creator was removed from the goal API.

PR #3359, opened 2022-01-02
WIP: Feature/algod app info app account

tzaffi:
## Summary

Want to expose an application's account address [just as we do in goal](https://github.com/tzaffi/go-algorand/blob/master/cmd/goal/application.go#L1036)

Though it's possible to calculate the account address as [the hash of the appid](https://developer.algorand.org/docs/get-details/dapps/smart-contracts/apps/#using-a-smart-contract-as-an-escrow), it seems like this should be part of a basic account info response.

## TODO: Test Plan


jannotti:
Why is this enormous? What is it?

tzaffi:
Closing this for now as the conversation in #2989 shows that this isn't an obvious "slam dunk". But definitely this is worth further discussion.

PR #3363, opened 2022-01-04
Use gc to avoid trashing.

algonautshant:
TestArchivalFromNonArchival was sporadically failing, particularly on the feature/dilithium-scheme-integration branch. 
The most likely reason for the slowdown is narrowed down to system trashing. 

This fix seems to fix the issue by invoking GC before the test to have maximum amount of memory available, and after the test to minimize the memory impact on subsequent tests.

This fix is also necessary to confirm the memory hypothesis. 

The memory consumption is high because of the existing ledger design, which is skewed in this test by creating a large number of asset creation transactions. This behavior is changing with the upcoming ledger refactoring project. 

Resolves https://github.com/algorand/go-algorand/issues/3334

### DIFF
--- a/ledger/archival_test.go
+++ b/ledger/archival_test.go
@@ -730,7 +730,11 @@ func TestArchivalFromNonArchival(t *testing.T) {
 	require.NoError(t, err)
 	blk := genesisInitState.Block
 
-	const maxBlocks = 1500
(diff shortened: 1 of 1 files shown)


algoidan (approved):
looks good to me!

algorandskiy (approved):

PR #3367, opened 2022-01-04
Add totals checks into acct updates tests

algorandskiy:
## Summary

After #2922 there is some leftover unused code for totals calculations. Turned this code into actual asserts.

## Test Plan

This is tests update


tsachiherman (approved):
I think that making these changes to the codebase is fine.

PR #3369, opened 2022-01-05
Allow setting manager, reserve, freeze, and clawback at goal asset create

### LINKED ISSUE #3361: goal asset create - support manager, reserve, freeze, and clawback
## Problem

When creating an asset, the transaction structure supports specifying the manager, reserve, freeze, and clawback accounts. The `goal asset create` command doesn't allow us to specify those fields, however.

## Solution

Add flags like `--manager` etc. to the create command.

## Dependencies

No

## Urgency

Not urgent, but nice to have.

fionnachan:
## Summary

Allow setting manager, reserve, freeze, and clawback at goal asset create

## Test Plan

Did not test it.
I built the binaries, ran `${GOPATH}/bin/goal node start -d ~/.algorand`, used goal to create a new wallet, but I don't know how to fund the wallet to continue :(

Closes #3361

### DIFF
--- a/cmd/goal/asset.go
+++ b/cmd/goal/asset.go
@@ -37,8 +37,13 @@ var (
 	assetURL                string
 	assetName               string
 	assetManager            string
+	assetReserve            string
 	assetClawback           string
 	assetFreezer            string
+	assetNoManager          bool
+	assetNoReserve          bool
+	assetNoFreezer          bool
+	assetNoClawback         bool
 
 	assetNewManager  string
 	assetNewReserve  string
@@ -64,6 +69,14 @@ func init() {
 	createAssetCmd.Flags().StringVar(&assetName, "name", "", "Name for the entire asset")
 	createAssetCmd.Flags().StringVar(&assetURL, "asseturl", "", "URL where user can access more information about the asset (max 32 bytes)")
 	createAssetCmd.Flags().StringVar(&assetMetadataHashBase64, "assetmetadatab64", "", "base-64 encoded 32-byte commitment to asset metadata")
+	createAssetCmd.Flags().StringVar(&assetManager, "manager", "", "Manager account that can issue transactions to re-configure or destroy the asset")
+	createAssetCmd.Flags().StringVar(&assetReserve, "reserve", "", "Reserve account that non-minted assets will reside in")
+	createAssetCmd.Flags().StringVar(&assetFreezer, "freezer", "", "Freezer account that can freeze or unfreeze the asset holdings for a specific account")
+	createAssetCmd.Flags().StringVar(&assetClawback, "clawback", "", "Clawback account that is allowed to transfer assets from and to any asset holder")
+	createAssetCmd.Flags().BoolVar(&assetNoManager, "no-manager", false, "Explicitly declare the lack of manager")
+	createAssetCmd.Flags().BoolVar(&assetNoReserve, "no-reserve", false, "Explicitly declare the lack of reserve")
+	createAssetCmd.Flags().BoolVar(&assetNoFreezer, "no-freezer", false, "Explicitly declare the lack of freezer")
+	createAssetCmd.Flags().BoolVar(&assetNoClawback, "no-clawback", false, "Explicitly declare the lack of clawback")
 	createAssetCmd.MarkFlagRequired("total")
 	createAssetCmd.MarkFlagRequired("creator")
 
@@ -189,6 +202,46 @@ var createAssetCmd = &cobra.Command{
 		client := ensureFullClient(dataDir)
 		accountList := makeAccountsList(dataDir)
 		creator := accountList.getAddressByName(assetCreator)
+		manager := creator
+		reserve := creator
+		freezer := creator
+		clawback := creator
+
+		if cmd.Flags().Changed("manager") {
+			assetManager = accountList.getAddressByName(assetManager)
+			manager = assetManager
+		}
+
+		if cmd.Flags().Changed("no-manager") {
+			manager = ""
+		}
+
+		if cmd.Flags().Changed("reserve") {
+			assetReserve = accountList.getAddressByName(assetReserve)
+			reserve = assetReserve
+		}
+
+		if cmd.Flags().Changed("no-reserve") {
+			reserve = ""
+		}
+
+		if cmd.Flags().Changed("freezer") {
+			assetFreezer = accountList.getAddressByName(assetFreezer)
+			freezer = assetFreezer
+		}
+
+		if cmd.Flags().Changed("no-freezer") {
+			freezer = ""
+		}
+
+		if cmd.Flags().Changed("clawback") {
+			assetClawback = accountList.getAddressByName(assetClawback)
+			clawback = assetClawback
+		}
+
+		if cmd.Flags().Changed("no-clawback") {
+			clawback = ""
+		}
 
 		var err error
 		var assetMetadataHash []byte
@@ -199,7 +252,7 @@ var createAssetCmd = &cobra.Command{
 			}
 		}
 
-		tx, err := client.MakeUnsignedAssetCreateTx(assetTotal, assetFrozen, creator, creator, creator, creator, assetUnitName, assetName, assetURL, assetMetadataHash, assetDecimals)
+		tx, err := client.MakeUnsignedAssetCreateTx(assetTotal, assetFrozen, manager, reserve, freezer, clawback, assetUnitName, assetName, assetURL, assetMetadataHash, assetDecimals)
 		if err != nil {
 			reportErrorf("Cannot construct transaction: %s", err)
 		}
--- a/cmd/goal/asset.go
+++ b/cmd/goal/asset.go
@@ -41,7 +41,6 @@ var (
 	assetClawback           string
 	assetFreezer            string
 	assetNoManager          bool
-	assetNoReserve          bool
 	assetNoFreezer          bool
 	assetNoClawback         bool
 
--- a/cmd/goal/asset.go
+++ b/cmd/goal/asset.go
@@ -198,19 +198,19 @@ var createAssetCmd = &cobra.Command{
 	Run: func(cmd *cobra.Command, _ []string) {
 		checkTxValidityPeriodCmdFlags(cmd)
 
-		if cmd.Flags().Changed("manager") && cmd.Flags().Changed("no-manager") {
+		if assetManager != "" && assetNoManager {
 			reportErrorf("The [--manager] flag and the [--no-manager] flag are mutually exclusive, do not provide both flags.")
 		}
 
-		if cmd.Flags().Changed("reserve") && cmd.Flags().Changed("no-reserve") {
+		if assetReserve != "" && assetNoReserve {
 			reportErrorf("The [--reserve] flag and the [--no-reserve] flag are mutually exclusive, do not provide both flags.")
 		}
 
-		if cmd.Flags().Changed("freezer") && cmd.Flags().Changed("no-freezer") {
+		if assetFreezer != "" && assetNoFreezer {
 			reportErrorf("The [--freezer] flag and the [--no-freezer] flag are mutually exclusive, do not provide both flags.")
 		}
 
-		if cmd.Flags().Changed("clawback") && cmd.Flags().Changed("no-clawback") {
+		if assetClawback != "" && assetNoClawback {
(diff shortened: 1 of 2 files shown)


jannotti:
This looks good by eye, we just need some simple tests.  You could add them to `test/scripts/e2e_subs/asset-misc.sh`

You could do another `asset create`, similar to line 24, and then confirm that the various addresses were set appropriately with some grep lines similar to line 26.

To run just that test, you should be able to do `./test/scripts/e2e_client_runner.py ./test/scripts/e2e_subs/asset-misc.sh` from the top of your repo.  That will take care of setting up a temporary test network, funding a wallet, and then invoking that script so that it can do the things inside.

fionnachan:
Added a test for `--no-manager --no-freezer --no-clawback` and another test for `--creator "${ACCOUNT}" --manager "${ACCOUNTB}"`. Let me know if you want tests for `--freezer` / `--clawback` separately.

I removed the `--no-reserve` flag because even when the reserve is set to empty explicitly, the reserve will be the creator anyway.

test/scripts/e2e_subs/asset-misc.sh
```diff
@@ -47,4 +47,44 @@ else
     exit 1
 fi
 
+# create asset with no manager, no freezer, and no clawback
+${gcmd} asset create --creator "${ACCOUNT}" --no-manager --no-freezer --no-clawback --name "${ASSET_NAME}" --unitname iamisc --total 1000000000000 --asseturl "${ASSET_URL}"
+
+IMMUTABLE_ASSET_ID=$(${gcmd} asset info --creator $ACCOUNT --unitname iamisc|grep 'Asset ID'|awk '{ print $3 }')
```
jannotti, line 53: It doesn't look like you use this, so I'd remove it.

test/scripts/e2e_subs/asset-misc.sh
```diff
@@ -47,4 +47,44 @@ else
...
+    echo ok
+else
+    date '+asset-misc immutable asset info error %Y%m%d_%H%M%S'
+    exit 1
+fi
+
+# create asset with a manager that is different from the creator
+${gcmd} asset create --creator "${ACCOUNT}" --manager "${ACCOUNTB}" --name "${ASSET_NAME}" --unitname dma --total 1000000000000 --asseturl "${ASSET_URL}"
+
+DIFF_MANAGER_ASSET_ID=$(${gcmd} asset info --creator $ACCOUNT --unitname dma|grep 'Asset ID'|awk '{ print $3 }')
```
jannotti, line 73: DIFF_MANAGER_ID also appears unused.

test/scripts/e2e_subs/asset-misc.sh
```diff
@@ -47,4 +47,44 @@ else
...
+    && [ "$FREEZE_ADDRESS" = "" ] \
+    && [ "$CLAWBACK_ADDRESS" = "" ]; then
+    echo ok
+else
+    date '+asset-misc immutable asset info error %Y%m%d_%H%M%S'
+    exit 1
+fi
+
+# create asset with a manager that is different from the creator
+${gcmd} asset create --creator "${ACCOUNT}" --manager "${ACCOUNTB}" --name "${ASSET_NAME}" --unitname dma --total 1000000000000 --asseturl "${ASSET_URL}"
```
jannotti, line 71: I'd prefer that this sets all of the different address types, not just manager, and that it use a different address for each.  That way we are testing against simple cut and paste bugs where, for example, using `--freezer X` actually sets the reserve address.

test/scripts/e2e_subs/asset-misc.sh
```diff
@@ -47,4 +47,44 @@ else
...
+
+DMA_MANAGER_ADDRESS=$(${gcmd} asset info --assetid ${DIFF_MANAGER_ASSET_ID} |grep 'Manager address'|awk '{ print $3 }')
+DMA_RESERVE_ADDRESS=$(${gcmd} asset info --assetid ${DIFF_MANAGER_ASSET_ID} |grep 'Reserve address'|awk '{ print $3 }')
+DMA_FREEZE_ADDRESS=$(${gcmd} asset info --assetid ${DIFF_MANAGER_ASSET_ID} |grep 'Freeze address'|awk '{ print $3 }')
+DMA_CLAWBACK_ADDRESS=$(${gcmd} asset info --assetid ${DIFF_MANAGER_ASSET_ID} |grep 'Clawback address'|awk '{ print $3 }')
+
+if [ "$DMA_MANAGER_ADDRESS" = "$ACCOUNTB" ] \
+    && [ "$DMA_RESERVE_ADDRESS" = "$ACCOUNT" ] \
+    && [ "$DMA_FREEZE_ADDRESS" = "$ACCOUNT" ] \
+    && [ "$DMA_CLAWBACK_ADDRESS" = "$ACCOUNT" ]; then
```
jannotti, line 83: This is where I'd like to see different equality checks for each address.

jannotti:
> I removed the `--no-reserve` flag because even when the reserve is set to empty explicitly, the reserve will be the creator anyway.

This is weird.  I don't think this should happen, I can't find any code in goal or algod that would do this on purpose, so I'd like to figure out if there's a bug here.  What exactly did you see?

jannotti (commented):
Once we figure out this reserve address thing, it looks good!

cmd/goal/asset.go
```diff
@@ -189,6 +202,46 @@ var createAssetCmd = &cobra.Command{
...
+		reserve := creator
+		freezer := creator
+		clawback := creator
+
+		if cmd.Flags().Changed("manager") {
+			assetManager = accountList.getAddressByName(assetManager)
+			manager = assetManager
+		}
+
+		if cmd.Flags().Changed("no-manager") {
```
jannotti, line 213: I suppose it would be nice to have a check to see if both switches, "manager " and "no-manager", are used, and if so, error out.  (Same for others too.)

cmd/goal/asset.go
```diff
@@ -189,6 +202,46 @@ var createAssetCmd = &cobra.Command{
...
+			manager = ""
+		}
+
+		if cmd.Flags().Changed("reserve") {
+			assetReserve = accountList.getAddressByName(assetReserve)
+			reserve = assetReserve
+		}
+
+		if cmd.Flags().Changed("no-reserve") {
```
jannotti: A slightly simpler way to use boolean flags is to set the default to false, and then you don't need to use `Changed()`, you can just use the variable directly.  See for example, how `assetFrozen` is used.

test/scripts/e2e_subs/asset-misc.sh
fionnachan: It was used for `goal asset info` to get the addresses. Is there a better way to get them? :/

jannotti:
Oh, sorry, I missed that. Of course you need it.

fionnachan:
I am seeing this with `--no-reserve`, which I didn't expect at all either before writing the test....
```
+ goal -w AF4A62B9687CF6BBAB8427229BE8F2D9 asset create --creator UOTRSRYEAHIPBIAOQA6HW3OI6FCRLXABNPPANGZ7WJJN2HTTTKKJAH3OII --no-manager --no-reserve --no-freezer --no-clawback --name 'Birlot : décollage vs. ࠶🦪' --unitname iamisc --total 1000000000000 --asseturl /ipfs/QmXoypizjW3WknFiJnKLwHCnL72vedxjQkDDP1mXWo6uco/wiki/Verifiable_random_function.html
Issued transaction from account UOTRSRYEAHIPBIAOQA6HW3OI6FCRLXABNPPANGZ7WJJN2HTTTKKJAH3OII, txid GCM66CSDDU36KB224LVIC7WU5LWNTYXLLSVZO7UYTULJQ6DIIVUQ (fee 1000)
Transaction GCM66CSDDU36KB224LVIC7WU5LWNTYXLLSVZO7UYTULJQ6DIIVUQ still pending as of round 20
Transaction GCM66CSDDU36KB224LVIC7WU5LWNTYXLLSVZO7UYTULJQ6DIIVUQ still pending as of round 21
Transaction GCM66CSDDU36KB224LVIC7WU5LWNTYXLLSVZO7UYTULJQ6DIIVUQ committed in round 22
Created asset with asset index 11
++ goal -w AF4A62B9687CF6BBAB8427229BE8F2D9 asset info --creator UOTRSRYEAHIPBIAOQA6HW3OI6FCRLXABNPPANGZ7WJJN2HTTTKKJAH3OII --unitname iamisc
++ grep 'Asset ID'
++ awk '{ print $3 }'
+ IMMUTABLE_ASSET_ID=11
++ goal -w AF4A62B9687CF6BBAB8427229BE8F2D9 asset info --assetid 11
++ grep 'Manager address'
++ awk '{ print $3 }'
+ MANAGER_ADDRESS=
++ goal -w AF4A62B9687CF6BBAB8427229BE8F2D9 asset info --assetid 11
++ grep 'Reserve address'
++ awk '{ print $3 }'
+ RESERVE_ADDRESS=UOTRSRYEAHIPBIAOQA6HW3OI6FCRLXABNPPANGZ7WJJN2HTTTKKJAH3OII
++ goal -w AF4A62B9687CF6BBAB8427229BE8F2D9 asset info --assetid 11
++ grep 'Freeze address'
++ awk '{ print $3 }'
+ FREEZE_ADDRESS=
++ goal -w AF4A62B9687CF6BBAB8427229BE8F2D9 asset info --assetid 11
++ grep 'Clawback address'
++ awk '{ print $3 }'
+ CLAWBACK_ADDRESS=
+ '[' '' = '' ']'
+ '[' UOTRSRYEAHIPBIAOQA6HW3OI6FCRLXABNPPANGZ7WJJN2HTTTKKJAH3OII = '' ']'
+ date '+asset-misc immutable asset info error %Y%m%d_%H%M%S'
asset-misc immutable asset info error 20220106_174132
+ exit 1
20220106_174132 :460 ERRORS after 93.727515 seconds: '/workspaces/go-algorand/test/scripts/e2e_subs/asset-misc.sh failed'
20220106_174132 :463 statuses-json: [{"script": "/workspaces/go-algorand/test/scripts/e2e_subs/asset-misc.sh", "ok": false, "seconds": 88.35015201568604}]
20220106_174132 :329 stop network in /tmp/tmp4nbvmh2a/net
20220106_174133 :338 stop kmd in /tmp/tmp4nbvmh2a/net/Node
20220106_174133 :351 delete network in /tmp/tmp4nbvmh2a/net
```

fionnachan:
oh look....
https://github.com/algorand/go-algorand/blob/52a1a2cc9c20750d492f6b99dd17ae7091593ff2/cmd/goal/asset.go#L634-L638

jannotti:
> oh look....
> 
> https://github.com/algorand/go-algorand/blob/52a1a2cc9c20750d492f6b99dd17ae7091593ff2/cmd/goal/asset.go#L611-L615
> 
> 
> https://github.com/algorand/go-algorand/blob/52a1a2cc9c20750d492f6b99dd17ae7091593ff2/cmd/goal/asset.go#L634-L638

Ok, good find.  This is only what `goal` reports and doesn't describe the actual value on-chain.  To me, that means you should still support `--no-reserve` which sets the value on chain to the zero value, but it is very hard to test with `goal` since `goal` lies about the value.

jannotti:
Oh, for testing the --no-reserve, you can actually check if the goal output contains `Empty. Defaulting to creator`

cmd/goal/asset.go
fionnachan: Done. For the flag checks, I used `cmd.Flags().Changed("manager") && cmd.Flags().Changed("no-manager")` instead of `cmd.Flags().Changed("manager") && assetNoManager` because it looks nicer with two flags on the same line.... 😬

jannotti: You can also use the variable for the manager flag directly, so you could do something as simple as `assetManager != "" && assetNoManager` without any use of `cmd.Flags().Changed` because they both have defaults set, so you would report the error if they both differ from their default.

I prefer to use the variables directly and only use `Changed` if I need to distinguish between when the switch gets its default value because it was not given on the command line, vs being given on the command-line, but being the same as the default.  Here, the distinction does not matter.

jannotti (dismissed):
This looks good to me, pending CI checks.  Thanks again!

algojack:
merging this to a temp branch to kick off CI correctly. Thanks again for your contribution.

jannotti:
This was merged as #3398 because we had a little CI hiccup.  Thanks again.

fionnachan:
You're welcome~

PR #3374, opened 2022-01-05
ledger refactoring: catchpoint fixes

algorandskiy:
## Summary

Fixes to `processAllResources` and `processAllBaseAccountRecords` related to a case when an account does not have resource records and a pending row needs to be carried over next account or a batch reading iteration.

Use `processAllBaseAccountRecords` in `encodedAccountsBatchIter.Next()`. This allows us to the same code for all scenarios where we read entire base account and resources tables.

### DIFF
--- a/ledger/accountdb.go
+++ b/ledger/accountdb.go
@@ -2820,44 +2820,65 @@ func makeOrderedAccountsIter(tx *sql.Tx, accountCount int) *orderedAccountsIter 
 	}
 }
 
+type pendingRow struct {
+	addrid int64
+	aidx   basics.CreatableIndex
+	rtype  basics.CreatableType
+	buf    []byte
+}
+
 func processAllResources(
 	resRows *sql.Rows,
-	addr basics.Address, accountData *baseAccountData, acctRowid int64,
+	addr basics.Address, accountData *baseAccountData, acctRowid int64, pr pendingRow,
 	callback func(addr basics.Address, creatableIdx basics.CreatableIndex, creatableType basics.CreatableType, resData *resourcesData, encodedResourceData []byte) error,
-) (err error) {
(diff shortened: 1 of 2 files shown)


ledger/accountdb.go
```diff
@@ -1735,8 +1734,11 @@ func performResourceTableMigration(ctx context.Context, tx *sql.Tx, log func(pro
 			return err
 		}
 		insertResourceCallback := func(ctx context.Context, rowID int64, cidx basics.CreatableIndex, ctype basics.CreatableType, rd *resourcesData) error {
-			encodedData := protocol.Encode(rd)
-			_, err = insertResources.ExecContext(ctx, rowID, cidx, ctype, encodedData)
+			var err error
```
algorandskiy, line 1737: this handles a case when an account does not have any resources. The callback is called anyway to let correct accounting in other use cases.

ledger/accountdb.go
```diff
@@ -2706,52 +2678,77 @@ func (iterator *encodedAccountsBatchIter) Next(ctx context.Context, tx *sql.Tx,
...
-				return
-			}
-			if nextRes.addrid != rowid {
-				break
-			}
+	var totalAppParams, totalAppLocalStates, totalAssetParams, totalAssets uint32
+	// emptyCount := 0
+	resCb := func(addr basics.Address, cidx basics.CreatableIndex, ctype basics.CreatableType, resData *resourcesData, encodedResourceData []byte) error {
+		emptyBaseAcct := baseAcct.TotalAppParams == 0 && baseAcct.TotalAppLocalStates == 0 && baseAcct.TotalAssetParams == 0 && baseAcct.TotalAssets == 0
+		if !emptyBaseAcct && resData != nil {
```
algorandskiy, line 2703: `emptyBaseAcct` means a current or prev account (relative to what was just read from resources table) does not have any resources.
`resData` is null when no resource at this call.

If I recall it right this condition handles a case when a very first account has 0 resources and the second has some.

tsachiherman (approved):

PR #3375, opened 2022-01-05
Removing C/crypto dependencies from `data/abi` package

### LINKED ISSUE #3355: Remove C dependencies from `data/abi` package
## Summary

The `data/abi` package was meant to be easily installed and used from the Go SDK, but the commit https://github.com/algorand/go-algorand/commit/d3bbe62caa11a2608c318919122457d4512045c5 introduced a dependency from `data/abi` to `data/basics`, which in turn references crypto packages with C dependencies. This means the Go SDK cannot easily update go-algorand past that commit.

We could refactor the `data/abi` package to avoid referencing `data/basics`.

## Priority

Needs to be done for the Go SDK to update past the commit https://github.com/algorand/go-algorand/commit/d3bbe62caa11a2608c318919122457d4512045c5.

ahangsu:
## Summary

We want to remove C/crypto dependencies from `data/abi` package, because the previous commit https://github.com/algorand/go-algorand/commit/d3bbe62caa11a2608c318919122457d4512045c5 introduced dependencies to crypto package with C dependencies (`SHA512/256` needed). This led to go-algorand-sdk cannot  easily update go-algorand past that commit.

The current solution is: "rewrite" the checksum function for address in `data/abi` package, which removes dependencies to `data/basics` and `algorand/crypto`, and sets dependency to `crypto` (golang).

Refactor the `data/abi` package to avoid referencing `data/basics`, closes #3355.

### DIFF
--- a/data/abi/abi_encode.go
+++ b/data/abi/abi_encode.go
@@ -201,7 +201,7 @@ func inferToSlice(value interface{}) ([]interface{}, error) {
 
 // encodeTuple encodes slice-of-interface of golang values to bytes, following ABI encoding rules
 func encodeTuple(value interface{}, childT []Type) ([]byte, error) {
-	if len(childT) >= (1 << 16) {
+	if len(childT) >= abiEncodingLengthLimit {
 		return nil, fmt.Errorf("abi child type number exceeds uint16 maximum")
 	}
 	values, err := inferToSlice(value)
@@ -277,7 +277,7 @@ func encodeTuple(value interface{}, childT []Type) ([]byte, error) {
 		if isDynamicIndex[i] {
 			// calculate where the index of dynamic value encoding byte start
 			headValue := headLength + tailCurrLength
-			if headValue >= (1 << 16) {
+			if headValue >= abiEncodingLengthLimit {
 				return nil, fmt.Errorf("cannot encode abi tuple: encode length exceeds uint16 maximum")
 			}
 			binary.BigEndian.PutUint16(heads[i], uint16(headValue))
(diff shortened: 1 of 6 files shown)


cce:
Theoretically, it would be possible to move crypto.Digest, crypto.Hash, and basics.Address into packages that didn't import libsodium crypto...

jasonpaulos:
This looks correct, but I don't love that it reimplements address encoding & decoding.

Refactoring `crypto.Digest`, `crypto.Hash`, and `basics.Address` into new packages that don't have C deps like @cce suggested seems like the best long term solution, but in the spirit of fixing our Go SDK import problem quickly, I _think_ I'm ok with this.

Though I do have two recommendations:
* Add a comment somewhere in `data/basics/address.go` that mentions there's another implementation of this in the abi package due to dependency issues (probably link to this PR too).
* Add a unit test for `addressCheckSum` that ensures it produces the same output as `basics.Address`. Maybe generate a large number of random addresses and compare the output of both functions.

cce:
Another idea, from @winder, is to wrap libsodium in a driver interface to make the existing crypto package importable even if libsodium doesn't exist. So calls to methods that use libsodium would return "Not implemented" errors or use an alternate Go implementation, if go-algorand was imported without it.

That way you wouldn't have to have a huge refactor everywhere we use crypto.Hash, crypto.Digest, etc, which don't rely on libsodium anyway.

jannotti (approved):
I think I'm ok with this too.  We reimplement these things in every SDK.  It seems extra weird to do it in Go, since there is already one implementation in Go. But I guess I don't really mind very much.

jasonpaulos (approved):
Looks good

PR #3376, opened 2022-01-06
Add test to exercise lookup corner cases

cce:
## Summary

This test attempts to cover the case when an accountUpdates.lookupX method can't find the requested address, falls through looking at deltas and the LRU accounts cache, then hits the database — only to discover that the round stored in the database (committed in `accountUpdates.commitRound`) is out of sync with `accountUpdates.cachedDBRound` (updated a little bit later in `accountUpdates.postCommit`).

In this case, the lookup method waits and tries again, iterating the `for { }` it is in. We did not have coverage for this code path before.

## Test Plan

Adds new test.


ledger/acctupdates_test.go
```diff
@@ -1700,3 +1700,145 @@ func TestConsecutiveVersion(t *testing.T) {
 		protocol.ConsensusV21,
 	}
 }
+
+func TestAcctUpdatesLookupRetry(t *testing.T) {
```
algorandskiy, line 1704: Please add a comment explaining this non-trivial test

ledger/acctupdates_test.go
```diff
@@ -1700,3 +1700,145 @@ func TestConsecutiveVersion(t *testing.T) {
...
+		checkAcctUpdates(t, au, 0, i, accts, rewardsLevels, proto)
+	}
+
+	flushRound := func(i basics.Round) {
+		// Clear the timer to ensure a flush
+		ml.trackers.lastFlushTime = time.Time{}
+
+		ml.trackers.committedUpTo(basics.Round(proto.MaxBalLookback) + i)
+		ml.trackers.waitAccountsWriting()
+		//checkAcctUpdates(t, au, i, basics.Round(proto.MaxBalLookback+14), accts, rewardsLevels, proto)
```
algorandskiy, line 1788: remove

ledger/acctupdates_test.go
```diff
@@ -1700,3 +1700,145 @@ func TestConsecutiveVersion(t *testing.T) {
...
+		time.Sleep(200 * time.Millisecond)
+		stallingTracker.postCommitReleaseLock <- struct{}{}
+	}()
+
+	// issue a LookupWithoutRewards while persistedData.round != au.cachedDBRound
+	d, validThrough, err := au.LookupWithoutRewards(rnd, addr)
+	require.NoError(t, err)
+	require.Equal(t, d, data)
+	require.GreaterOrEqualf(t, uint64(validThrough), uint64(rnd), "validThrough: %v rnd :%v", validThrough, rnd)
+	t.Log("validThrough", validThrough)
```
algorandskiy, line 1839: remove

algorandskiy (commented):
Looks good, couple minor notes

algorandskiy (approved):

tsachiherman (approved):
I'm a bit worried about random failures, since the Sleep(200ms) might create a non-consistent behavior on slow machines. But I think that this is a good starting point.

PR #3380, opened 2022-01-06
Use rejecting limit listener in `WebsocketNetwork`

tolikzinovyev:
## Summary

Replace the standard limit listener with the new rejecting limit listener in `WebsocketNetwork`. This will let the dialing node know that connection is impossible faster.

## Test Plan

Probably not necessary.


tsachiherman (approved):
looks good.

Issue #2394, opened 2021-07-01
Flaky test: TestAsyncTelemetryHook_QueueDepth


yaovi-a:
## Description
Travis Test Errors Reference Spreadsheet for details on cases where this test has failed: https://docs.google.com/spreadsheets/d/1apjaNvZodnTSIdZaQlpMaqf0nzbKgSYdIWHqvmINHS8/edit#gid=0

### Example Failure
Time: Mon 7/19/2021
Build: https://app.travis-ci.com/algorand/go-algorand/builds/233054279?utm_source=slack&utm_medium=notification
Travis Test: External ARM64 Build, MacOS AMD64 Build
Package: logging
Test: TestAsyncTelemetryHook_QueueDepth 

```
=== FAIL: logging TestAsyncTelemetryHook_QueueDepth (0.03s)
telemetryhook_test.go:220:
Error Trace:	telemetryhook_test.go:220
Error:      	Not equal:
expected: 10
actual  : 11
Test:       	TestAsyncTelemetryHook_QueueDepth
DONE 5471 tests, 23 skipped, 1 failure in 3558.591s
Makefile:241: recipe for target 'fulltest' failed
make: *** [fulltest] Error 1
The command "/home/ubuntu/go/src/github.com/[secure]/go-[secure]/scripts/travis/test.sh" failed. Retrying, 2 of 3.
```

yaovi-a:
This test was removed with #2299

algobarb:
I'm not sure, but I don't think this test was properly removed given the Travis FAIL on a false positive External ARM64 Build last Monday.

yaovi-a:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Algorand-Main-Workspace-5ea71fda6d554d53dda36f16/issues/algorand/go-algorand/2394?planning-poker) with ZenHub @algobolson @algonautshant @AlgoStephenAkiki @nicholasguoalgorand @tolikzinovyev @tsachizehub @winder

PR #3382, opened 2022-01-06
Fix Flaky Async Telemetry Test

AlgoStephenAkiki:
Resolves #2934 

https://github.com/algorand/go-algorand/issues/2394

Fix the unit test TestAsyncTelemetryHook_QueueDepth:

	// the anonymous goroutine in createAsyncHookLevels might pull an entry off the pending list before
	// writing it off to the underlying hook. when that happens, the total number of sent entries could
	// be one higher then the maxDepth.

### DIFF
--- a/logging/telemetryhook_test.go
+++ b/logging/telemetryhook_test.go
@@ -225,5 +225,10 @@ func TestAsyncTelemetryHook_QueueDepth(t *testing.T) {
 	close(filling)
 	hook.Close()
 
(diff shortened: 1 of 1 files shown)


algorandskiy:
Alternative implementation #2685

winder:
Yes, Stephen was trying out a couple of versions of this fix. Looks like Tsachi's original approach is the way to go. I'll close this PR and we can merge in #2685

PR #3385, opened 2022-01-06
Give more time for e2e_basic_start_stop's verify_none_running

cce:
## Summary

This test has been failing more often in recent nightly jobs. Giving it more time to wait for algod to shut down should help.

```
######################################################################
  e2e_basic_start_stop
######################################################################
Verifying a generic node will start using goal
Verifying we can stop it using goal
The node was successfully stopped.
Verifying a generic node will start directly
Verifying that the datadir algod lock works correctly
Logging to:  /opt/cibuild/project/tmp/out/e2e/13191-1641172161879/data/node.log
Deadlock detection is set to: enabled (Default state is 'enable')
Initializing the Algorand node... Success!
⇨ http server started on 127.0.0.1:8080
Node running and accepting RPC requests over HTTP on port 127.0.0.1:8080. Press Ctrl-C to exit
algod not expected to be running but it is
subcommand returncode 1: ['./e2e_basic_start_stop.sh']
Cleaning up temp dir.
```

## Test Plan

Modifies existing test.


algobarb:
The CircleCi tests still seem to be failing?

algobarb (approved):
Don't know if it was the last commit that made it pass, but I don't think this will hurt to try it on our flaky tests

tsachiherman:
I think that we have a bug in algod which prevents it from shutting down when requested. While this PR is not wrong on it's own, I think that figuring out the slowness during the shutdown would be a good thing to accomplish. ( which might also resolve this issue ).

algojack (approved):

PR #3393, opened 2022-01-07
Feature Networks pipeline related changes

algojack:
Added support for not having certain files in signing script


algobarb (commented):
have we tested that these changes work as expected where it's used outside of multiverse? So then we can check that expected behavior hasn't changed elsewhere.

algojack:
> have we tested that these changes work as expected where it's used outside of multiverse? So then we can check that expected behavior hasn't changed elsewhere.

i'm not sure where/how to test that. Could you connect with me to go over that?

bricerisingalgorand:
> have we tested that these changes work as expected where it's used outside of multiverse? So then we can check that expected behavior hasn't changed elsewhere.

I ran the steps in our jenkins build pipeline and made sure that it produced binaries with the correct genesis files and signatures with this change for mainnet

PR #3397, opened 2022-01-07
Feature/contract to contract -> master

jannotti:


cmd/opdoc/opdoc.go
```diff
@@ -203,6 +247,8 @@ func opToMarkdown(out io.Writer, op *logic.OpSpec) (err error) {
 		assetParamsFieldsMarkdown(out)
 	} else if op.Name == "app_params_get" {
 		appParamsFieldsMarkdown(out)
+	} else if op.Name == "acct_params_get" {
```
tsachiherman, line 250: code style:
switch-case style branching would be a more elegant solution here

cmd/opdoc/opdoc.go
```diff
@@ -361,25 +389,29 @@ func main() {
...
-	fieldTableMarkdown(assetparams, logic.AssetParamsFieldNames, logic.AssetParamsFieldTypes, logic.AssetParamsFieldDocs())
+	fieldSpecsMarkdown(assetparams, logic.AssetParamsFieldNames, logic.AssetParamsFieldSpecByName)
 	assetparams.Close()
 
 	appparams, _ := os.Create("app_params_fields.md")
-	fieldTableMarkdown(appparams, logic.AppParamsFieldNames, logic.AppParamsFieldTypes, logic.AppParamsFieldDocs())
+	fieldSpecsMarkdown(appparams, logic.AppParamsFieldNames, logic.AppParamsFieldSpecByName)
 	appparams.Close()
 
+	acctparams, _ := os.Create("acct_params_fields.md")
```
tsachiherman, line 411: Please check returned error value and add proper error handling.

cmd/opdoc/opdoc.go
jannotti: Agree.  I'll change it.

cmd/opdoc/opdoc.go
jannotti: Sure.

cmd/tealdbg/local.go
```diff
@@ -573,44 +539,19 @@ func (r *LocalRunner) Run() (bool, error) {
...
 	run := r.runs[0]
-	ep := evalParams[0]
+	// Workaround for Go's nil/empty interfaces nil check after nil assignment, i.e.
+	// r.debugger = nil
+	// ep.Debugger = r.debugger
+	// if ep.Debugger != nil // FALSE
+	if r.debugger != nil {
+		r.debugger.SaveProgram(run.name, run.program, run.source, run.offsetToLine, run.states)
+		ep.Debugger = r.debugger
+	}
```
tsachiherman, line 554: this is correct, but it looks like it needs to be applied every time you assign a debugger.
In that case, how do you feel about adding a method `EvalParams.AttachDebugger` that would accomplish that ?
( i.e. testing nil interface as well as nil debugger interface pointer )

config/consensus.go
```diff
@@ -440,8 +445,8 @@ var MaxStateDeltaKeys int
 // any version, used only for decoding purposes. Never decrease this value.
 var MaxLogCalls int
 
-// MaxInnerTransactions is the maximum number of inner transactions that may be created in an app call.
-var MaxInnerTransactions int
+// MaxInnerTransactionsPerDelta is the maximum number of inner transactions in one EvalDelta
```
tsachiherman, line 448: The change here is obvious - but it would be causing a potential issue on the "window" between the time a new binary is deployed and the time the new consensus protocol takes place.
To work around that, I suggest adding a manual testing, which be consensus protocol dependent where we would enforce the old limit on older protocol. This would complement the high-bound we set here for the msgpack decoding.

tsachiherman: Separately, I think that it's great that you renamed the variable so that it would force compile-time validation.

daemon/algod/api/server/v2/dryrun.go
```diff
@@ -299,9 +296,9 @@ func (dl *dryrunLedger) LookupWithoutRewards(rnd basics.Round, addr basics.Addre
 			}
 		}
 	}
-	if !any {
-		return basics.AccountData{}, 0, fmt.Errorf("no account for addr %s", addr.String())
-	}
+	// Returns a 0 account for account that wasn't supplied.  This is new as of
```
tsachiherman, line 299: I don't remember what was the rationale here, but I do like that we're aligning the interface functionality. @pzbitskiy do you remember the details around this one ?

data/transactions/logic/assembler.go
```diff
@@ -786,60 +778,73 @@ func assembleSubstring(ops *OpStream, spec *OpSpec, args []string) error {
...
-func assembleTxn(ops *OpStream, spec *OpSpec, args []string) error {
-	if len(args) != 1 {
-		return ops.error("txn expects one argument")
-	}
-	fs, ok := txnFieldSpecByName[args[0]]
+func txnFieldImm(name string, expectArray bool, ops *OpStream) (*txnFieldSpec, error) {
+	fs, ok := TxnFieldSpecByName[name]
 	if !ok {
-		return ops.errorf("txn unknown field: %#v", args[0])
+		return nil, fmt.Errorf("unknown field: %#v", name)
```
tsachiherman, line 784: `name` is a string. is there any particular reason to avoid using `%s` here ?

tsachiherman, line 784: `name` is a string. is there any particular reason to avoid using `%s` here ?

data/transactions/logic/assembler.go
```diff
@@ -786,60 +778,73 @@ func assembleSubstring(ops *OpStream, spec *OpSpec, args []string) error {
...
+	if expectArray != fs.array {
+		if expectArray {
+			return nil, fmt.Errorf("found scalar field %#v while expecting array", name)
+		}
+		return nil, fmt.Errorf("found array field %#v while expecting scalar", name)
 	}
 	if fs.version > ops.Version {
-		return ops.errorf("field %#v available in version %d. Missed #pragma version?", args[0], fs.version)
+		return nil,
+			fmt.Errorf("field %#v available in version %d. Missed #pragma version?", name, fs.version)
```
tsachiherman, line 794: I think that this message could be improved - we can definitely tell if the user had a pragma or not, and if so, what version he selected. That being said, it shouldn't be a blocker, and we can improve this later on.

data/transactions/logic/assembler.go
jannotti: That's an easy way to get it quoted, which I think makes a nicer error message.

data/transactions/logic/assembler.go
```diff
@@ -786,60 +778,73 @@ func assembleSubstring(ops *OpStream, spec *OpSpec, args []string) error {
...
 	ops.pending.WriteByte(spec.Opcode)
 	ops.pending.WriteByte(uint8(fs.field))
 	ops.returns(fs.ftype)
 	return nil
 }
 
-// assembleTxn2 delegates to assembleTxn or assembleTxna depending on number of operands
-func assembleTxn2(ops *OpStream, spec *OpSpec, args []string) error {
+// asmTxn2 delegates to asmTxn or asmTxna depending on number of operands
+func asmTxn2(ops *OpStream, spec *OpSpec, args []string) error {
```
tsachiherman, line 825: code style:
switch-case style looks like a more natural choice here.

data/transactions/logic/assembler.go
jannotti: To quote the name in the error message

data/transactions/logic/eval.go
```diff
@@ -158,8 +173,10 @@ func stackValueFromTealValue(tv *basics.TealValue) (sv stackValue, err error) {
 // newly-introduced transaction fields from breaking assumptions made by older
 // versions of TEAL. If one of the transactions in a group will execute a TEAL
 // program whose version predates a given field, that field must not be set
-// anywhere in the transaction group, or the group will be rejected.
-func ComputeMinTealVersion(group []transactions.SignedTxn) uint64 {
+// anywhere in the transaction group, or the group will be rejected. In
+// addition, inner app calls must not call teal from before inner app calls were
+// introduced.
+func ComputeMinTealVersion(group []transactions.SignedTxnWithAD, inner bool) uint64 {
```
tsachiherman, line 179: while this is probably ok for the first revision, should we require that all teal invocation on inner transactions should share the same teal version as the caller to ensure that any two calls to the same program would result in the same outputs ? I honestly can't think of an attack vector here, but it would clearly be more "simplified". ( on the flip side, it would require contract authors to revise their contracts ).

data/transactions/logic/eval.go
```diff
@@ -264,19 +259,121 @@ type EvalParams struct {
...
+			apps++
+		}
+	}
+
+	minTealVersion := ComputeMinTealVersion(txgroup, false)
+
+	var pooledApplicationBudget *int
+	var pooledAllowedInners *int
+
+	credit, _ := transactions.FeeCredit(txgroup, proto.MinTxnFee)
```
tsachiherman, line 315: please add a unit test for `FeeCredit` as well as test the returned error code.

config/consensus.go
jannotti: I think it's always safe if a later protocol increases the limit in a way that is only used by that new protocol version, as this does.  During the upgrade window, before the new protocol actually goes into effect and some installations have new code and some have old, they will disagree about the allocbound. But no node will produce transactions that _use_ the new allocbound (because the protocol has not been upgraded on the new nodes).

I'm not sure where we would put the manual testing code you suggest. It seems like it belongs in the unmarshalling code, but that's autogenerated.

data/transactions/logic/eval.go
jannotti: I don't think we need to enforce that.  The attack I was concerned was only concerned about the boundary between allowing and disallowing inner txns.  I was concerned that there might exist old code that used poor sources of randomness, believing that only the block proposer could take advantage of it by deciding whether to put it in the block based on outcome.  Inner transactions change that game, because a txn can trying invoking the app that uses poor randomness and, if it does not like the outcome, abort after the inner call, cancelling it.

data/transactions/logic/eval.go
```diff
@@ -685,9 +792,30 @@ func boolToUint(x bool) uint64 {
 	return 0
 }
 
-// MaxStackDepth should move to consensus params
+// MaxStackDepth should not change unless gated by a teal version change / consensus upgrade.
```
tsachiherman, line 795: you basically saying that it's need to be a consensus parameter..

data/transactions/logic/eval.go
```diff
@@ -685,9 +792,30 @@ func boolToUint(x bool) uint64 {
...
 const MaxStackDepth = 1000
 
+func (cx *EvalContext) remainingBudget() int {
```
tsachiherman, line 798: might be safer to make 0 the lower bound of this method, but I don't see any logical flaw here.

data/transactions/logic/eval.go
```diff
@@ -1385,6 +1519,19 @@ func opBytesMul(cx *EvalContext) {
...
 
+func opBytesSqrt(cx *EvalContext) {
+	last := len(cx.stack) - 1
+
+	if len(cx.stack[last].Bytes) > MaxByteMathSize {
+		cx.err = errors.New("math attempted on large byte-array")
+		return
+	}
+
+	val := new(big.Int).SetBytes(cx.stack[last].Bytes)
```
tsachiherman, line 1530: Setting bytes from the stack could result in a negative value that would panic when Sort is called.

data/transactions/logic/eval.go
jannotti: Not exactly.  There are many things that we gate off the  teal version. It would, in fact, be _wrong_ to make it a consensus parameter. Old programs must continue to see the value associated with their version, not a new value that appears in a new consensus parameter.  I think the old comment was misleading and implied we could change the value for all programs, new and old, by making it a consensus parameter.

data/transactions/logic/eval.go
jannotti: SetBytes always treats the bytes as an unsigned integer.
> SetBytes interprets buf as the bytes of a big-endian unsigned integer, sets z to that value, and returns z.

data/transactions/logic/assembler.go
jannotti: sure.

data/transactions/logic/eval.go
jannotti: Added a unit test.

By the time we're here,  TxnGroupBatchVerify from verify/txn.go has run, which is essentially the same code, so it should not be possible for this to fail.  And the method has no way to exit with an error.

A couple options:
1. I could actually call FeeCredit from TxnGroupBatchVerify, so that it is more clear that this call will not fail.
2. I could return nil to the calller if FeeCredit fails here.
3. I could do both.

I like number 1 somewhat, though it means running through the txns twice. (I can't imagine that's a real problem, it's a tiny loop.) I don't like number 2 much as it just kicks the can down the road, and all callers ought to check for nil.

I don't know if that feels like enough for you - In general, what do we think is the right thing to do when code is running over a verified and/or "well formed" transactions in such a way that it knows certain problems can not occur?

data/transactions/logic/eval.go
```diff
@@ -2089,337 +2224,366 @@ func (cx *EvalContext) txnFieldToStack(txn *transactions.Transaction, fs txnFiel
...
+	srcGroup      txnSource = iota
+	srcInner                = iota
+	srcInnerGroup           = iota
+)
+
+// opTxnImpl implements all of the txn variants.  Each form of txn opcode should
+// be able to get its work done with one call here, after collecting the args in
+// the most straightforward way possible. They ought to do no error checking, so
+// that it is all collected here.
+func (cx *EvalContext) opTxnImpl(gi uint64, src txnSource, field TxnField, ai uint64, expectArray bool) (sv stackValue, err error) {
```
tsachiherman, line 2284: variable type inconsistency : the `gi` variable would probably be of an `int` type here.
while I think that would be better to have it as `uint64` everywhere, it looks like this is the only exception - so, let's make this align with the rest of the codebase ?

data/transactions/logic/eval.go
```diff
@@ -2879,6 +3038,30 @@ func opGloads(cx *EvalContext) {
 	cx.stack[last] = scratchValue
 }
 
+func opGloadss(cx *EvalContext) {
+	last := len(cx.stack) - 1
+	prev := last - 1
+
+	gi := cx.stack[prev].Uint
+	if gi >= uint64(len(cx.TxnGroup)) {
+		cx.err = fmt.Errorf("gloads lookup TxnGroup[%d] but it only has %d", gi, len(cx.TxnGroup))
```
tsachiherman, line 3047: wrong error string : `gloads` -> `gloadss`

data/transactions/logic/eval.go
jannotti: This was intentional, as described in the comment above. opTxnImpl takes "wide" types so that it can be called safely from the various txn opcodes, some of which have bytes for gi, some have uint64 (since they got the index off the stack).  This is the one place ion the code where we confirm that these indexes fit into ints (actually, we confirm much more - that they are small enough for the txn group). And so from here on, they can be coerced into `int`.

The same is true of `ai`

data/transactions/logic/eval.go
```diff
@@ -2089,337 +2224,366 @@ func (cx *EvalContext) txnFieldToStack(txn *transactions.Transaction, fs txnFiel
...
 		return
-	} else if txn := cx.TxnGroup[groupIdx].Txn; !(txn.Type == protocol.ApplicationCallTx || txn.Type == protocol.AssetConfigTx) {
-		err = fmt.Errorf("can't use %s on txn that is not an app call nor an asset config txn with index %d", opName, groupIdx)
+	} else if txn := cx.TxnGroup[gi].Txn; !(txn.Type == protocol.ApplicationCallTx || txn.Type == protocol.AssetConfigTx) {
+		err = fmt.Errorf("can't use %s on txn that is not an app call nor an asset config txn with index %d", opName, gi)
 		return
 	}
 
-	cid, err := cx.getCreatableID(groupIdx)
+	cid, err := cx.getCreatableID(gi)
```
tsachiherman, line 2584: `getCreatableID` is being called only from here, and in this case, the error object is consistently being ignored.
Could you restructure this code so that only a single error would be generated and returned ?

data/transactions/logic/eval.go
```diff
@@ -2442,8 +2606,14 @@ func opGaid(cx *EvalContext) {
...
-	groupIdx := cx.stack[last].Uint
-	sv, err := opGaidImpl(cx, groupIdx, "gaids")
+
+	gi := cx.stack[last].Uint
+
+	if gi >= uint64(len(cx.TxnGroup)) {
+		cx.err = fmt.Errorf("gaids lookup TxnGroup[%d] but it only has %d", gi, len(cx.TxnGroup))
+		return
+	}
+	sv, err := opGaidImpl(cx, int(gi), "gaids")
```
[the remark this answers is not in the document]
jannotti: Not quite.  This relates to why I changed the opTxnImpl function that takes wide types.  Notice that `opGaidImpl` takes an `int` gi.  So, before we even call it, we must at least check that the uint64 gi we have here fits in a u32, or else it might wrap around when cast.  It might even wrap around to a nice small positive number, and not cause an error in the impl.  So we check it first.

This complexity is what I was trying to fix by the way I redid opTxnImpl, so it can accept the wider types, and do all the checking there.  I could change gaid opcodes to work in a similar way.

data/transactions/logic/eval.go
jannotti: The error is not ignored, it gets returned implicitly at the end of opGaidImpl. But I take your point - I can make this clearer. I'll probably eliminate `getCreatableID()`

data/transactions/logic/eval.go
```diff
@@ -2549,15 +2699,29 @@ func (cx *EvalContext) globalFieldToValue(fs globalFieldSpec) (sv stackValue, er
 	case LatestTimestamp:
 		sv.Uint, err = cx.getLatestTimestamp()
 	case CurrentApplicationID:
-		sv.Uint, err = cx.getApplicationID()
+		sv.Uint = uint64(cx.appID)
 	case CurrentApplicationAddress:
-		var addr basics.Address
-		addr, err = cx.getApplicationAddress()
+		addr := cx.getApplicationAddress(cx.appID)
```
tsachiherman, line 2704: should an error be generated if cx.appID is zero ?
( i.e. I can't think of any execution path that it *could* happen ), which is why I'm hesitant about this one.

data/transactions/logic/eval.go
```diff
@@ -2549,15 +2699,29 @@ func (cx *EvalContext) globalFieldToValue(fs globalFieldSpec) (sv stackValue, er
...
 	case CreatorAddress:
 		sv.Bytes, err = cx.getCreatorAddress()
 	case GroupID:
-		sv.Bytes = cx.getGroupID()
+		sv.Bytes = cx.Txn.Txn.Group[:]
+	case OpcodeBudget:
+		sv.Uint = uint64(cx.remainingBudget())
+	case CallerApplicationID:
+		if cx.caller != nil {
+			sv.Uint = uint64(cx.caller.appID)
```
tsachiherman, line 2714: interesting.. allowing to create private functions ;-)

data/transactions/logic/eval.go
jannotti: opGlobal prevents accessing an "app mode" global, which CurrentApplicationAddress is, unless it was run in app mode. And in app mode, the appID is set at the start of execution. I'll add a test that ensures that is true even for inners - that if you create an app with an inner transaction, when that app initializes, `cx.appID` is set properly.

data/transactions/logic/eval.go
```diff
@@ -3648,10 +3908,66 @@ func opAppParamsGet(cx *EvalContext) {
...
+			return
+		}
+		value.Uint = mbal.Raw
+	case AcctAuthAddr:
+		auth, err := cx.Ledger.Authorizer(addr)
+		if err != nil {
+			cx.err = err
+			return
+		}
+		if auth == addr {
```
tsachiherman, line 3957: that's.. unexpected. If the `AuthAddr` on the account data happen to be the same as the public key address, why would you use the empty address ?

jannotti: The `Authorizer()` method's job is to report the correct authorizer for the account, regardless of whether it is the original key, or new key that has been set by rekey.  But the job of this field is to report the AuthAddr - they key that an account has been rekeyed to if it it has been rekeyed at all.  So, when the Authorizer() reports the original key, there is no AuthAddr, and the field returns the zero address.

Note that this is not ambiguous.  You can not set the AuthAddr of an account to the original public key.  If you try, it simply gets unset.
```
// Rekey updates tx.Sender's AuthAddr to tx.RekeyTo, if provided
func Rekey(balances Balances, tx *transactions.Transaction) error {
	if (tx.RekeyTo != basics.Address{}) {
		acct, err := balances.Get(tx.Sender, false)
		if err != nil {
			return err
		}
		// Special case: rekeying to the account's actual address just sets acct.AuthAddr to 0
		// This saves 32 bytes in your balance record if you want to go back to using your original key
		if tx.RekeyTo == tx.Sender {
			acct.AuthAddr = basics.Address{}
		} else {
			acct.AuthAddr = tx.RekeyTo
		}

		err = balances.Put(tx.Sender, acct)
		if err != nil {
			return err
		}
	}
	return nil
}
```

data/transactions/logic/eval.go
jannotti: I was mostly worried about applications that do not wish to be invoked as inner transactions (like the bad random number generator app).  If such an application wanted to move to v6, but still not be invokable as an inner, they could check this and fail.

But I suppose an app could also refuse to run unless called by specific apps.

algorandskiy (commented):
Some comments, part 1

data/transactions/logic/eval.go
```diff
@@ -3648,10 +3908,66 @@ func opAppParamsGet(cx *EvalContext) {
...
+	}
+	cx.stack[last] = value
+	cx.stack = append(cx.stack, stackValue{Uint: exist})
+}
+
 func opLog(cx *EvalContext) {
 	last := len(cx.stack) - 1
 
-	if len(cx.Logs) == MaxLogCalls {
+	if len(cx.Txn.EvalDelta.Logs) == MaxLogCalls {
```
tsachiherman, line 3970: I'm pretty sure there is no other way to generate the log entries.. but it would generally be safer to use `>=` rather than `==` in these cases.

data/transactions/logic/eval.go
jannotti: My guess, in fact, is that did use FeeCredit in the verification code at some point, and maybe I mismerged master at some point and reverted to doing the credit "by hand".  Let me know if you think that would solve this problem. It should be clearer that the txgroup passes the same criteria twice (and I can include a comment as such).

data/transactions/logic/eval.go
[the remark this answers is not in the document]
jannotti: I've change opGaidImpl to work the same way as opTxnImpl.  It takes uint64 and all the checking is now there in one place.

data/transactions/logic/eval.go
jannotti: That's what I did. getCreatableID is gone, so the error handling is clearer now.

cmd/tealdbg/localLedger_test.go
```diff
@@ -109,22 +110,15 @@ int 2
...
-		GroupIndex:      0,
-		PastSideEffects: pse,
-	}
-	pass, delta, err := ba.StatefulEval(ep, appIdx, program)
+	ep := logic.NewEvalParams([]transactions.SignedTxnWithAD{{SignedTxn: txn}}, &proto, &transactions.SpecialAddresses{})
+	pass, delta, err := ba.StatefulEval(0, ep, appIdx, program)
 	a.NoError(err)
 	a.True(pass)
-	a.Equal(1, len(delta.GlobalDelta))
+	a.Len(delta.GlobalDelta, 1)
```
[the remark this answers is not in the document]
jannotti: It's especially nice because it shows the bad collection if it's wrong, which can help debug.

data/transactions/logic/eval.go
```diff
@@ -3372,11 +3579,28 @@ func opAppLocalPut(cx *EvalContext) {
...
+		cx.err = err
+		return
+	}
+
+	tv := sv.toTealValue()
+	if !ok || tv != etv {
+		if _, ok := cx.Txn.EvalDelta.LocalDeltas[accountIdx]; !ok {
+			cx.Txn.EvalDelta.LocalDeltas[accountIdx] = basics.StateDelta{}
+		}
+		cx.Txn.EvalDelta.LocalDeltas[accountIdx][key] = tv.ToValueDelta()
```
tsachiherman, line 3601: Shouldn't the size of `LocalDeltas` be capped to `config.MaxEvalDeltaAccounts` ?

data/transactions/logic/README.md
```diff
@@ -86,35 +217,39 @@ An application transaction must indicate the action to be taken following the ex
...
+| Opcode | Description |
+| - | -- |
+| `sha256` | SHA256 hash of value A, yields [32]byte |
+| `keccak256` | Keccak256 hash of value A, yields [32]byte |
+| `sha512_256` | SHA512_256 hash of value A, yields [32]byte |
 | `ed25519verify` | for (data A, signature B, pubkey C) verify the signature of ("ProgData" \|\| program_hash \|\| data) against the pubkey => {0 or 1} |
 | `ecdsa_verify v` | for (data A, signature B, C and pubkey D, E) verify the signature of the data against the pubkey => {0 or 1} |
-| `ecdsa_pk_recover v` | for (data A, recovery id B, signature C, D) recover a public key => [*... stack*, X, Y] |
-| `ecdsa_pk_decompress v` | decompress pubkey A into components X, Y => [*... stack*, X, Y] |
+| `ecdsa_pk_recover v` | for (data A, recovery id B, signature C, D) recover a public key |
```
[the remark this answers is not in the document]
jannotti: I've made it so that stack changes are automatically documented so I removed the extra notation in some of the documentation strings.  For example, ecdsa_recover now says this in it's header:
- Stack: ..., A: []byte, B: uint64, C: []byte, D: []byte => ..., X: []byte, Y: []byte

I think that covers what you intended. Is that right?

data/transactions/logic/fields.go
```diff
@@ -500,32 +578,44 @@ const (
...
-// AssetHoldingFieldTypes is StackUint64 StackBytes in parallel with AssetHoldingFieldNames
-var AssetHoldingFieldTypes []StackType
-
 type assetHoldingFieldSpec struct {
 	field   AssetHoldingField
 	ftype   StackType
 	version uint64
 }
 
+func (fs *assetHoldingFieldSpec) Type() StackType {
```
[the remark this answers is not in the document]
jannotti: I would really like to figure out a way to shrink all the field stuff down.  Because each field is a different type, Go makes it quite hard to parameterize over them. But in many cases, calling code really only wants a map[string]uint64 and a reverse map map[uint64]string.

data/transactions/logic/export_test.go
```diff
@@ -0,0 +1,44 @@
...
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with go-algorand.  If not, see <https://www.gnu.org/licenses/>.
+
+package logic
+
+// Export for testing only.  See
+// https://medium.com/@robiplus/golang-trick-export-for-test-aa16cbd7b8cd for a
+// nice explanation.
```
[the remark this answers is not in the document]
jannotti:
```
// Export for testing only.  See
// https://medium.com/@robiplus/golang-trick-export-for-test-aa16cbd7b8cd for a
// nice explanation. tl;dr: Since some of our testing is in logic_test package,
// we export some extra things to make testing easier there. But we do it in a
// _test.go file, so they are only exported during testing.
```

data/transactions/logic/assembler.go
```diff
@@ -1076,58 +1016,55 @@ func asmItxnOnly(ops *OpStream, spec *OpSpec, args []string) error {
...
-		return ops.errorf("%s %#v available in version %d. Missed #pragma version?", spec.Name, args[0], fs.version)
-	}
-	arrayFieldIdx, err := strconv.ParseUint(args[1], 0, 64)
-	if err != nil {
-		return ops.error(err)
-	}
-	if arrayFieldIdx > 255 {
-		return ops.errorf("%s array index beyond 255: %d", spec.Name, arrayFieldIdx)
+		//nolint:errcheck // we continue to maintain typestack
+		ops.errorf("%s %s available in version %d. Missed #pragma version?", spec.Name, args[0], fs.version)
```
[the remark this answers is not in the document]
jannotti: I don't follow. Are you suggested that this is your preferred way to turn off the linter?  Why no do it explicitly with the nolint directive?

data/transactions/logic/eval.go
```diff
@@ -4038,7 +4526,7 @@ func (cx *EvalContext) PcDetails() (pc int, dis string) {
 
 func base64Decode(encoded []byte, encoding *base64.Encoding) ([]byte, error) {
 	decoded := make([]byte, encoding.DecodedLen(len(encoded)))
-	n, err := encoding.Strict().Decode(decoded, encoded)
+	n, err := encoding.Decode(decoded, encoded)
```
tsachiherman, line 4529: this would break compatibility if it wasn't for `fidoVersion`

data/transactions/logic/backwardCompat_test.go
```diff
@@ -280,98 +277,71 @@ func TestBackwardCompatTEALv1(t *testing.T) {
...
 
 	// Costs for v2 should be higher because of hash opcode cost changes
-	ep2 := defaultEvalParamsWithVersion(&sb, &txn, 2)
-	ep2.TxnGroup = txgroup
+	ep2, tx, _ := makeSampleEnvWithVersion(2)
 	ep2.Proto.LogicSigMaxCost = 2307
-	err = Check(opsV2.Program, ep2)
-	require.Error(t, err)
-	require.Contains(t, err.Error(), "static cost")
+	//	ep2.TxnGroup[0].Lsig.Logic = opsV2.Program
```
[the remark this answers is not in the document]
jannotti: It doesn't need to be pout there ahead, because testLogicBytes puts it in before running.  I suppose I should remove the comment since it's just confusing.

data/transactions/logic/eval.go
tsachiherman: The change here from `SignedTxn` to `SignedTxnWithAD` isn't really harmful on it's own, but it would require that every single transaction ( even a non-app txn ), would now need to have it's data structure copied.

data/transactions/logic/backwardCompat_test.go
```diff
@@ -280,98 +277,71 @@ func TestBackwardCompatTEALv1(t *testing.T) {
...
-	txn.Lsig.Logic = program
-	txn.Lsig.Args = [][]byte{data[:], sig[:], pk[:], txn.Txn.Sender[:], txn.Txn.Note}
+	ep.TxnGroup[0].Lsig.Logic = program
+	ep.TxnGroup[0].Lsig.Args = [][]byte{data[:], sig[:], pk[:], tx.Sender[:], tx.Note}
 
 	// Cost remains the same, because v0 does not get dynamic treatment
 	ep.Proto.LogicSigMaxCost = 2139
-	err = Check(program, ep)
-	require.Error(t, err)
+	ep.MinTealVersion = new(uint64) // Was higher because sample txn has a rekey
```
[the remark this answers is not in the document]
jannotti: The evaluator calculates a MinTealVersion is none is supplied upon first access.  I wanted to remove that because I found it surprising, but that would require that we change a lot of callers for LogicSig evaluation that assume they can set up an EvalParams by hand (which leaves MinTealVersion = nil) and go ahead.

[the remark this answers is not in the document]
jannotti: Or are you asking why the old test did not need to have the mintealversion changed?  Because back on line 289, the old code manipulated the tx RekeyTo field by hand (resetting it).

data/transactions/logic/backwardCompat_test.go
```diff
@@ -399,35 +368,22 @@ func TestBackwardCompatGlobalFields(t *testing.T) {
...
-		ep := defaultEvalParams(nil, nil)
-		ep.Proto = &proto
-		ep.Ledger = ledger
-
-		// check failure with version check
-		_, err := Eval(ops.Program, ep)
-		require.Error(t, err)
-		require.Contains(t, err.Error(), "greater than protocol supported version")
-		_, err = Eval(ops.Program, ep)
+		ep, _, _ := makeSampleEnvWithVersion(1)
```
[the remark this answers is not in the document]
jannotti: The thing we are checking here is whether the version handling is done properly. So we set the allowed logic version to a known value, and confirm that anything higher is rejected.

If we want to separately test that protocol.ConsensusV23 has a particular version, that seems ok, but why not test that every single consensus version is exactly what it always has been?  It certainly doesn't seem like a test for the logic package.

data/transactions/logic/backwardCompat_test.go
```diff
@@ -475,44 +424,35 @@ func TestBackwardCompatTxnFields(t *testing.T) {
...
 				require.NoError(t, err)
 			}
 
-			proto := config.Consensus[protocol.ConsensusV23]
-			require.False(t, proto.Application)
-			ep := defaultEvalParams(nil, nil)
-			ep.Proto = &proto
-			ep.Ledger = ledger
-			ep.TxnGroup = txgroup
+			ep, tx, _ := makeSampleEnvWithVersion(1)
```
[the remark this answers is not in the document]
jannotti: I can't see why we would test that config.Consensus[protocl.ConsensusV23] has a specific logicsig version when what we're trying to see is if fields are properly rejected by having the wrong version.  You might as well just `require.Equal(t, config.Consensus[protocol.ConsensusV23].LogicVersion, 4)`

data/transactions/logic/eval.go
jannotti: sure, I'll change.

data/transactions/logic/eval.go
jannotti: That's an interesting question, and something I'm still grappling with for allocbounds.  I don't think that our code explicitly enforces many of the allocbounds.  Instead, the allocbounds are calculated so that they can not be exceeded (they are usually the max across _any_ protocol version, or something even more conservative).

I do think we have a problem if we ever get that wrong (as we've been talking about the last few days). But I don't know if should just sprinkle checks around.

I suppose I would feel safer if I wrote some sort of `func (ad ApplyData) checkAllocBounds()` that recursively checked them, and invoked it at the end of evaluation.  Does that sound reasonable?

data/transactions/logic/eval.go
jannotti: I convinced myself that Strict() was the default, and unneeded, and none of our test complained.  However, I'm no longer so sure, so I will try to create tests that care about this.

data/transactions/logic/eval.go
jannotti: Stateful logic processing needs to take a copy no matter what because it holds onto the copies and stores ApplyData there as it evaluates. You are right that we could avoid a copy in logicsig processing if we are careful enough.  Let me know if I should look into it.  That seems quite fast compared to costs normally associated with signature checking though.

data/transactions/logic/assembler.go
tsachiherman: The reason I'm trying to avoid using `#v` when not needing it is that is would call the `String()` method, which, in turn, might have some unexpected outcome. We had a bug where a similar case was leading to infinite recursive call ( i.e. the String() method was using format, which ended up with recursive call )

jannotti: %s also called the String() method if it exists.

data/transactions/logic/eval.go
tsachiherman: I don't consider this to be a blocker, but I have no idea what the performance implication would be. I'm pretty confident that having another method `ComputeMinTealVersionForInnerTransactions` would completely eliminate these concerns. Note that this method could be "cheaper" : there is no need to check for applications/rekeying support since this call could not be invoked before the application support anyway.

data/transactions/logic/eval.go
tsachiherman: Oh I see what you did here. Admittedly, I think that it's even worse. We shouldn't make the evaluator depend on the semantics of setting the AuthAddr field on the AccountData. What we really want is to have access to the AccountData.AuthAddr here, right ? I think that adding this to LedgerForLogic would work out nicely here. The COW should already have that cached as part of the basics.AccountData.

data/transactions/logic/eval.go
tsachiherman: The high level design was to write the code so that it wouldn't have the ability to generate unrealistic sizes of  arrays/maps. Then, the msgp allocbound were set to align with these.

I don't believe that we have too many locations that have these issue. I think that it would be less error prone to perform that inline rather than "export" that process. ( i.e. keep the limits next to the variable )

data/transactions/logic/eval.go
```diff
@@ -3372,11 +3579,28 @@ func opAppLocalPut(cx *EvalContext) {
...
-	if err == nil {
-		err = cx.Ledger.SetLocal(addr, key, sv.toTealValue(), accountIdx)
+	addr, accountIdx, err := cx.mutableAccountReference(cx.stack[pprev])
+	if err != nil {
+		cx.err = err
+		return
 	}
 
+	// if writing the same value, do nothing, matching ledger behavior with
+	// previous BuildEvalDelta mechanism
```
algorandskiy, line 3589: I do not see where do we compare old and new in ledger:
SetKey:
```
	vdelta, ok := lsd.kvCow[key]
	if !ok {
		vdelta = valueDelta{old: oldValue, oldExists: oldOk}
	}
	vdelta.new = value
	vdelta.newExists = true
	lsd.kvCow[key] = vdelta
```
In `applyStorageDeltas` we write no matter what
```
states.KeyValue[k] = v.new
```

data/transactions/logic/eval.go
```diff
@@ -3648,10 +3908,66 @@ func opAppParamsGet(cx *EvalContext) {
...
 
+func opAcctParamsGet(cx *EvalContext) {
+	last := len(cx.stack) - 1 // acct
+
+	if cx.Ledger == nil {
+		cx.err = fmt.Errorf("ledger not available")
+		return
+	}
+
+	addr, _, err := cx.accountReference(cx.stack[last])
```
algorandskiy, line 3919: mutableAccountReference? as the call to ledger.Balance below assumes this is preexisting account, and mutableAccountReference returns non-error only for existing addresses

data/transactions/logic/eval.go
```diff
@@ -3911,10 +4262,96 @@ func (cx *EvalContext) stackIntoTxnField(sv stackValue, fs txnFieldSpec, txn *tr
 	case FreezeAssetFrozen:
 		txn.AssetFrozen, err = sv.bool()
 
-	// appl needs to wait. Can't call AVM from AVM.
-
+	// ApplicationCall
+	case ApplicationID:
+		txn.ApplicationID, err = cx.availableApp(sv)
```
algorandskiy, line 4267: this might return self and we do not allow calling self, not really matter but might fail earlier here

algorandskiy (commented):
part 2...

data/transactions/logic/eval.go
jannotti: Yes, it would certainly be worse if I had introduced `Authorizer` for this purpose. Why would I combine the public key with the authaddr, just to figure that out and tear them apart again later? Perhaps there's more going on.

Authorizer() was introduced for authorization in v5. When checking if an inner transaction is allowed, the Authorizer() of the intended Sender is compared to the app address. If they are the same, the transaction is authorized. This keeps the mechanics of how authaddr overrides the public key out of logic evaluation.

Next, and separately, in v6 there was a desire to expose the AuthAddr to teal code. I could have introduced yet another method to LogicForLedger, but it seems crazy to have two methods intended to deal with getting info about the rekey address. So, I left the existing Authorizer() method alone (so that authentication code could remain less complicated) and put complexity in the "safe" place - exposing the authaddr to teal access.

So to summarize. Rekey info is needed in two places in evaluation - authorization and field exposure. Since authorization came first and is intuitively more important to get right, I left the code alone and put the complexity in field access.

If I switch to exposing the authaddr directly in LedgerForLogic, instead of offerring Authorizer() as a method, I have to repeat the code to override the private key with the authaddr in the logic evaluator's authorization code. I can't see the win there.

data/transactions/logic/eval.go
```diff
@@ -194,65 +216,38 @@ type LedgerForLogic interface {
...
 	AppParams(aidx basics.AppIndex) (basics.AppParams, basics.Address, error)
-	ApplicationID() basics.AppIndex
 	OptedIn(addr basics.Address, appIdx basics.AppIndex) (bool, error)
-	GetCreatableID(groupIdx int) basics.CreatableIndex
 
 	GetLocal(addr basics.Address, appIdx basics.AppIndex, key string, accountIdx uint64) (value basics.TealValue, exists bool, err error)
-	SetLocal(addr basics.Address, key string, value basics.TealValue, accountIdx uint64) error
-	DelLocal(addr basics.Address, key string, accountIdx uint64) error
+	SetLocal(addr basics.Address, appIdx basics.AppIndex, key string, value basics.TealValue, accountIdx uint64) error
+	DelLocal(addr basics.Address, appIdx basics.AppIndex, key string, accountIdx uint64) error
```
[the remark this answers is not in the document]
jannotti: Right, because I can't be sure I can reproduce all the ledger's tricky old code to maintain compatibility with some past changes in EvalDeltas. So the ledger still need have enough info to build the eval deltas for old protocol versions.

data/transactions/logic/eval.go
```diff
@@ -264,19 +259,121 @@ type EvalParams struct {
 	// MinTealVersion is nil, we will compute it ourselves
 	MinTealVersion *uint64
```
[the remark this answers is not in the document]
jannotti: It's because existing code uses the nil as a flag to determine the min version internally.  Code that evaluates logicsigs makes an EvalParams by hand (which leaves the value nil) and then inside the evaluator, if it sees a nil, it fills it in.

I started to change it, but we invoke it that way from regular logicsig processing, dry run, goal, debugger... it just seemed like too much to change.

data/transactions/logic/eval.go
```diff
@@ -425,41 +547,43 @@ func (pe PanicError) Error() string {
...
-
-	// update pooled budget
-	if cx.Proto.EnableAppCostPooling && cx.PooledApplicationBudget != nil {
-		// if eval passes, then budget is always greater than cost, so should not have underflow
-		*cx.PooledApplicationBudget = basics.SubSaturate(*cx.PooledApplicationBudget, uint64(cx.cost))
-	}
 	// update side effects
-	cx.PastSideEffects[cx.GroupIndex].setScratchSpace(cx.scratch)
+	cx.pastScratch[cx.GroupIndex] = &scratchSpace{}
+	*cx.pastScratch[cx.GroupIndex] = cx.scratch
```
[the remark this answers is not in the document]
jannotti: pastScratch holds all the scratch values from previous executions.  This stashes the current cx.scratch in there for future txn use.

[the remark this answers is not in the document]
jannotti: Oh, sorry, you mean why not store the pointer.  I was tempted.  I thought it might have bad garbage collection effects if I held onto a pointer into the middle of the cx, which is what that would be.  In many GCs, that would mean the entire cx is held onto.  And it also felt nice and clean to copy the values.  But I could be convinced to change.

data/transactions/logic/eval.go
```diff
@@ -1884,62 +2023,58 @@ func TxnFieldToTealValue(txn *transactions.Transaction, groupIndex int, field Tx
...
 
 	// Hashes are expensive, so we cache computed TxIDs
-	txid, ok := cx.txidCache[groupIndex]
+	txid, ok := cx.EvalParams.txidCache[groupIndex]
 	if !ok {
-		txid = txn.ID()
-		cx.txidCache[groupIndex] = txid
+		if cx.caller != nil {
+			innerOffset := len(cx.caller.Txn.EvalDelta.InnerTxns)
+			txid = txn.InnerID(cx.caller.Txn.ID(), innerOffset+groupIndex)
```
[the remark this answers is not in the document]
jannotti: `InnerID` hashes in the parent txid.

data/transactions/logic/eval.go
jannotti: mutable only means that the the account being looked up must have an index in the static array (or be the sender so we can use 0).  The idea is that we have added some rules to allow access to more accounts, but we have no way to write a LocalDelta unless we have an index. So we can't allow access to those extra accounts for writes.  That means we only need mutableAccount() when doing a put or delete.

data/transactions/logic/eval.go
jannotti: In order to make the spec clear (we are trying to document whether things fail when fields are set, or when submit is called). Rather than have a lot of special cases, we say that we only fail when a field is set if it's the wrong type, or if the "thing" (account, app, asa) isn't "available" (in the arrays or accessible by a new rule).

Everything else fails at itxn_submit time.

data/transactions/logic/eval.go
jannotti: appcow_test.go demonstrates that when a keys old and new value are the same, BuildEvalDelta does not spit out a stateDelta. (Lines 473, for example).

Maybe comes from valueDelta.serialize:
```
	if vd.oldExists && vd.old == vd.new {
		return
	}
```

data/transactions/logic/eval.go
```diff
@@ -264,19 +259,121 @@ type EvalParams struct {
...
+	for _, tx := range txgroup {
+		if tx.Txn.Type == protocol.ApplicationCallTx {
+			apps++
+		}
+	}
+
+	minTealVersion := ComputeMinTealVersion(txgroup, false)
+
+	var pooledApplicationBudget *int
+	var pooledAllowedInners *int
```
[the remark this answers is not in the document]
jannotti: Sure, I suppose that saves a couple allocations.

cmd/tealdbg/local.go
jannotti: Although this looks like "my" code in the diff, it's actually just moving some existing code to deal with the way EvalParams are shared across executions in a group.  So I'm going to assume Pavel is right here, and no other problem areas exist.

data/transactions/logic/eval.go
jannotti: Before I do this, could @tsachiherman comment on whether he thinks it's acceptable?  I would use `FeeCredit` in TxnBatchVerify, and then I would point that out here, as the explanation for why I need not check the error return.

data/transactions/logic/eval.go
jannotti: Added - so now we have more assurance that cx.appID is always set (non-zero) in app evaluation

data/transactions/logic/eval.go
jannotti: Test added.  We'll debate and decide whether we want Strict() or not before we adjust fidoVersion.

data/transactions/logic/eval.go
```diff
@@ -194,65 +216,38 @@ type LedgerForLogic interface {
...
+	Perform(gi int, ep *EvalParams) error
+	Counter() uint64
 }
 
-// setScratchSpace stores the scratch space
-func (se *EvalSideEffects) setScratchSpace(scratch scratchSpace) {
-	se.scratchSpace = scratch
+// resources contains a list of apps and assets. It's used to track the apps and
+// assets created by a txgroup, for "free" access.
+type resources struct {
```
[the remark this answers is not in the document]
jannotti: I would still have to keep them separate, since I need to only lookup assets when looking for assets and apps when looking for assets.

I think your main point is about lookup speed? I expect these lists to be quite short.  These are the assets or apps that were created earlier in this txgroup. I think I'd prefer an occasional O(n) lookup in a short list than to make two maps each time.  I guess I could be convinced if you feel strongly though.

data/transactions/logic/eval.go
```diff
@@ -1884,62 +2023,58 @@ func TxnFieldToTealValue(txn *transactions.Transaction, groupIndex int, field Tx
 	if groupIndex < 0 {
 		return basics.TealValue{}, fmt.Errorf("negative groupIndex %d", groupIndex)
 	}
-	cx := EvalContext{EvalParams: EvalParams{GroupIndex: uint64(groupIndex)}}
+	cx := EvalContext{
+		GroupIndex: groupIndex,
+		Txn:        &transactions.SignedTxnWithAD{SignedTxn: transactions.SignedTxn{Txn: *txn}},
+	}
```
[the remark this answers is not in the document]
jannotti: I think you're right. Since it only cares about those values for "effects" fields, I think we're ok putting in a blank cx until the debugger wants those values.

data/transactions/logic/eval.go
tsachiherman: I think that the implementation you've made for `Authorizer` is fine, and there is no need to change it. It's just the new `AuthAddr` field that is the issue here. When retrieving this field, we should avoid any logic being applied and just provide it as is. Having the extra logic would mean that the correctness of the teal logic would be tied to the implementation of the rekey apply logic; that's a design risk that could be easily avoided.

jannotti: Ok, I can add AccountData() as a LedgerForLogic methods, and remove Balance() and MinBalance(). Then get all of that info from the AccountData.

One begins to wonder if the LedgerForLogic interface is doing much any more, as we've had to expose so much for c2c and this.

data/transactions/logic/assembler.go
jannotti: I agree, but I'm going to defer this unless you disagree.

algorandskiy (commented):
I like the change and exceed testing. It might be worth exploring a case of self opting-in and setting own local state when signed with a logic sig.
Another point is to check the old cow ApplyData calculation and the new ApplyData in logic produce roughly the same results (exclude nil vs empty maps and so on).
I'll make another readout of logic.eval and revisit previous questions.

Btw, I found `divw` and `bsqrt` are not relevant to this PR but looks like it is too much work to extract it.

data/transactions/logic/evalAppTxn_test.go
```diff
@@ -564,19 +632,1094 @@ txn Sender; itxn_field Receiver;
...
+
+}
+
+// TestInnerCallDepth ensures that inner calls are limited in depth
+func TestInnerCallDepth(t *testing.T) {
+	t.Parallel()
+
+	ep, tx, ledger := MakeSampleEnv()
+	// Allow a lot to make the test viable
+	ep.Proto.MaxAppTxnForeignApps = 50
```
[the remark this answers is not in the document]
jannotti: Before we decided to do the limit on maxAppCallDepth, this code was written to go much deeper. So it seems good to leave the test in that makes it so that it _could_ go deep, and ensure that it can't.

data/transactions/logic/evalStateful_test.go
```diff
@@ -2531,3 +2518,54 @@ func TestAppAddress(t *testing.T) {
...
+
+func TestSelfMutate(t *testing.T) {
+	ep, _, ledger := makeSampleEnv()
+
+	/* In order to test the added protection of mutableAccountReference, we're
+	   going to set up a ledger in which an app account is opted into
+	   itself. That was impossible before v6, and indeed we did not have the
+	   extra mutable reference check then. */
+	ledger.NewLocals(basics.AppIndex(888).Address(), 888)
+	ledger.NewLocal(basics.AppIndex(888).Address(), 888, "hey",
```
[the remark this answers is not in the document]
jannotti: If the address were in the ForeignApps array, it could be modified.  This is the problem were EvalDeltas can only be written if we can use an index to name the account being modified.  So unless the account is supplied in the foreign array, it can't be allowed to be written.

ledger/apply/asset.go
```diff
@@ -100,12 +100,7 @@ func AssetConfig(cc transactions.AssetConfigTxnFields, header transactions.Heade
 			return err
 		}
 
-		// Record the index used. No separate config for activating
-		// storage in AD because inner transactions can't be turned on
-		// without this change.
-		if balances.ConsensusParams().MaxInnerTransactions > 0 {
-			ad.ConfigAsset = newidx
-		}
+		ad.ConfigAsset = newidx
```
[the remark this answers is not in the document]
jannotti: It's useful for the value to be in the ApplyData, where it can be accessed by later transactions that use `itxn CreatedConfigID`. And, in fact, I changed `gaid` so that it uses this mechanism to get a created id as well. (Rather than track created ids in two different ways.). `gaid` exists even before we had inner transactions, so this must be stored for its use regardless of whether MaxInnerTransactions is turned on.

However, the ApplyData that we write in old protocols does not include this. It is cleared out the same way we clear out rewards info before the AD was supposed to have it.

data/transactions/logic/eval_test.go
```diff
@@ -981,7 +925,7 @@ int 0
 >
 &&
 global CurrentApplicationID
-int 42
+int 888
```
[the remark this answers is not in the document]
jannotti: True, but it was not the answer to "What is my current application ID when run with testApp?"

data/transactions/logic/eval.go
tsachiherman: The fact that you need to do it from two places is a good indication that we haven't built the execution pipeline optimally here. I don't think it should be a blocker, but we should prepare for refactor this in a better way.
From practical perspective :
1. add a comment above the said call explaining the situation
2. Make TxnGroupBatchVerify call `transactions.FeeCredit`.
3. Create a ticket for scheduling that work

data/transactions/logic/eval.go
jannotti: In that regard, this is true here.  The allocbound is huge (2048) but the number of different accounts that can be the accountIdx here is limited to 5. (The number of accounts in the static array, plus one for sender.)

data/transactions/logic/evalAppTxn_test.go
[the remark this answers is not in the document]
jannotti: If we don't allow a bunch of ForeignApps, we can't construct a txn that will even get that deep. We have to call at least 8 different apps to check the depth limit, and the default ForeignApps limit is smaller.

data/transactions/logic/assembler.go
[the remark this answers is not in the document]
algorandskiy: I did not realize `ops.errorf` is supposed to be used without returning the error, I checked the code and it looks like both variants are used frequently.

data/transactions/logic/backwardCompat_test.go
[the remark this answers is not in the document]
algorandskiy: ConsensusV23 is pre-app version where teal programs might have v0 and v1, and the test checks if pre-app TEAL programs fail with with rekey set and etc. The new code makes a fake proto with TEAL v1 only, and if fact does not check real scenario.

data/transactions/logic/eval.go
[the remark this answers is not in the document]
algorandskiy: I think you are right. Please then add a comment :)

data/transactions/logic/eval.go
```diff
@@ -685,9 +792,30 @@ func boolToUint(x bool) uint64 {
...
+
+func (cx *EvalContext) remainingInners() int {
+	if cx.Proto.EnableInnerTransactionPooling && cx.pooledAllowedInners != nil {
+		return *cx.pooledAllowedInners
+	}
+	// Before EnableInnerTransactionPooling, MaxInnerTransactions was the amount
+	// allowed in a single txn. No consensus version should enable inner app
+	// calls without turning on EnableInnerTransactionPoolin, else inner calls
+	// could keep branching with "width" MaxInnerTransactions
+	return cx.Proto.MaxInnerTransactions - len(cx.Txn.EvalDelta.InnerTxns)
```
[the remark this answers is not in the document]
algorandskiy: I mean `remainingInners` are always used as `len(cx.subtxns) > cx.remainingInners()` where `len` is supposed to return a non-negative number, returning a non-negative from `remainingInners` does not leave chances for integer overflow

algorandskiy:
Please consider adding the following test:
1. App A calls B
2. B opt-ins into self and sets some local state
3. check B.Address() account for existing non-empty local state.

jannotti:
> Please consider adding the following test:
> 
> 1. App A calls B
> 2. B opt-ins into self and sets some local state
> 3. check B.Address() account for existing non-empty local state.

An app cannot opt its account into itself, because that would require re-entrancy. (To optin into yourself, you'd need to issue an app call to yourself with OnCompletion=OptIn)

Is the main idea to see if, when A calls B, if B makes changes, those changes are visible to A?

algorandskiy:
Looks like we also missing global/local state reads in scenarios like these:
1. A calls B, B sets B's state
2. A calls C, C reads B's state
3. A reads B's state

From the code I see it should work but I do not see tests checking this (no add_global_get opcode is on only in old tests).

data/transactions/logic/eval.go
jannotti: This turned out to have a similar problem as we were discussing with ComputeMinTealVersion. Some code would like to work on the SignedTxns, and some on SignedTxnWithADs.  So I'm going to trim down the FeeCredit function to be local (`feeCredit`) have it operate on the WithADS, and not return an error (it will document that it depends upon the group being validated earlier.

data/transactions/logic/eval.go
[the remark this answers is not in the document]
jannotti: len returns an int, so the comparison is signed.  This seems safe. I'm reluctant to switch to a sub saturate form, because then remainingInners() has a surprising interface. Someone might think that could make an inner call, and then check if remainingInner() < 0.  That is a sort of reasonable expectation.  But if it returns 0 at the lowest, it would be wrong.

tsachiherman (approved):

PR #3399, opened 2022-01-08
Add `SqliteReadConcurrency` parameter

tolikzinovyev:
## Summary

Add `SqliteReadConcurrency` parameter to local config and take it into account when calculating the number of available file descriptors required. `SqliteReadConcurrency` is then used to limit number of read "connections" to the block and tracker databases.

Each database "connection" will create 3 file descriptors for opening the database: .sqlite, sqlite-shm and .sqlite-wal. The current FD budgeting doesn't account for this.

## Test Plan

None.

### DIFF
--- a/config/localTemplate.go
+++ b/config/localTemplate.go
@@ -430,6 +430,9 @@ type Local struct {
 	// The http server does not accept new connections as long we have this many
 	// (hard limit) connections already.
 	RestConnectionsHardLimit uint64 `version[20]:"2048"`
+
+	// How many times tracker and block sqlite dbs can be open for reading simultaneously.
+	SqliteReadConcurrency uint64 `version[20]:"128"`
 }
 
 // DNSBootstrapArray returns an array of one or more DNS Bootstrap identifiers
--- a/daemon/algod/server.go
+++ b/daemon/algod/server.go
@@ -103,13 +103,20 @@ func (s *Server) Initialize(cfg config.Local, phonebookAddresses []string, genes
 
 	// Set large enough soft file descriptors limit.
 	var ot basics.OverflowTracker
-	fdRequired := ot.Add(
-		cfg.ReservedFDs,
-		ot.Add(uint64(cfg.IncomingConnectionsLimit), cfg.RestConnectionsHardLimit))
+	fdRequired :=
+		ot.Add(cfg.ReservedFDs, uint64(cfg.IncomingConnectionsLimit))
+	fdRequired = ot.Add(fdRequired, cfg.RestConnectionsHardLimit)
+	{
+		// Each of block and tracker dbs requires 3 file descriptors: .sqlite,
+		// .sqlite-shm and .sqlite-wal.
+		sqliteFds := ot.Mul(cfg.SqliteReadConcurrency, 6)
+		fdRequired = ot.Add(fdRequired, sqliteFds)
+	}
 	if ot.Overflowed {
 		return errors.New(
-			"Initialize() overflowed when adding up ReservedFDs, IncomingConnectionsLimit " +
-				"RestConnectionsHardLimit; decrease them")
+			"Initialize() overflowed when calculating ReservedFDs " +
+				"IncomingConnectionsLimit + RestConnectionsHardLimit + " +
+				"SqliteReadConcurrency*6; decrease them")
 	}
 	err = util.SetFdSoftLimit(fdRequired)
 	if err != nil {
--- a/config/local_defaults.go
+++ b/config/local_defaults.go
@@ -104,6 +104,7 @@ var defaultLocal = Local{
 	RestReadTimeoutSeconds:                     15,
 	RestWriteTimeoutSeconds:                    120,
 	RunHosted:                                  false,
+	SqliteReadConcurrency:                      128,
 	SuggestedFeeBlockHistory:                   3,
 	SuggestedFeeSlidingWindowSize:              50,
 	TLSCertFile:                                "",
--- a/installer/config.json.example
+++ b/installer/config.json.example
@@ -83,6 +83,7 @@
     "RestReadTimeoutSeconds": 15,
     "RestWriteTimeoutSeconds": 120,
     "RunHosted": false,
+    "SqliteReadConcurrency": 128,
     "SuggestedFeeBlockHistory": 3,
     "SuggestedFeeSlidingWindowSize": 50,
     "TLSCertFile": "",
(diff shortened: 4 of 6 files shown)


daemon/algod/server.go
```diff
@@ -103,13 +103,20 @@ func (s *Server) Initialize(cfg config.Local, phonebookAddresses []string, genes
...
-		ot.Add(uint64(cfg.IncomingConnectionsLimit), cfg.RestConnectionsHardLimit))
+	fdRequired :=
+		ot.Add(cfg.ReservedFDs, uint64(cfg.IncomingConnectionsLimit))
+	fdRequired = ot.Add(fdRequired, cfg.RestConnectionsHardLimit)
+	{
+		// Each of block and tracker dbs requires 3 file descriptors: .sqlite,
+		// .sqlite-shm and .sqlite-wal.
+		sqliteFds := ot.Mul(cfg.SqliteReadConcurrency, 6)
+		fdRequired = ot.Add(fdRequired, sqliteFds)
+	}
```
algorandskiy, line 114: why is in a block?

config/localTemplate.go
```diff
@@ -430,6 +430,9 @@ type Local struct {
 	// The http server does not accept new connections as long we have this many
 	// (hard limit) connections already.
 	RestConnectionsHardLimit uint64 `version[20]:"2048"`
+
+	// How many times tracker and block sqlite dbs can be open for reading simultaneously.
+	SqliteReadConcurrency uint64 `version[20]:"128"`
```
algorandskiy, line 435: where does 128 come from?

daemon/algod/server.go
tolikzinovyev: to hide `sqliteFds`

tsachiherman (changes requested):
As implemented, I think that it's not a great implementation. It could slow down the internal processing due to a REST endpoint call.
I would suggest creating a separate connection that would be used by the REST endpoints interface. However, before going into such an endeavor, I'd like to see some e2e test that shows that this is truly a viable issue.

tolikzinovyev:
> As implemented, I think that it's not a great implementation. It could slow down the internal processing due to a REST endpoint call. I would suggest creating a separate connection that would be used by the REST endpoints interface. However, before going into such an endeavor, I'd like to see some e2e test that shows that this is truly a viable issue.

Can't REST endpoints slow down internal processing without this change? I don't really see the difference.
What kind of e2e test?

tsachiherman:
> > As implemented, I think that it's not a great implementation. It could slow down the internal processing due to a REST endpoint call. I would suggest creating a separate connection that would be used by the REST endpoints interface. However, before going into such an endeavor, I'd like to see some e2e test that shows that this is truly a viable issue.
> 
> Can't REST endpoints slow down internal processing without this change? I don't really see the difference. What kind of e2e test?

Right now, the number of sqlite readers isn't limited. So in theory, it would slow down the internal performance only once it exceeded the underlying hardware capabilities. With your change, you're capping it at the driver level. That could lead to starvation of a node's internal operation, which I'd rather avoid.

tsachiherman:
> > As implemented, I think that it's not a great implementation. It could slow down the internal processing due to a REST endpoint call. I would suggest creating a separate connection that would be used by the REST endpoints interface. However, before going into such an endeavor, I'd like to see some e2e test that shows that this is truly a viable issue.
> 
> Can't REST endpoints slow down internal processing without this change? I don't really see the difference. What kind of e2e test?

The e2e test that I'm thinking of would create a network, saturate it with some kind of load, and be able to cause the node to crash due to high number of concurrent database queries.

tolikzinovyev:
> > > As implemented, I think that it's not a great implementation. It could slow down the internal processing due to a REST endpoint call. I would suggest creating a separate connection that would be used by the REST endpoints interface. However, before going into such an endeavor, I'd like to see some e2e test that shows that this is truly a viable issue.
> > 
> > 
> > Can't REST endpoints slow down internal processing without this change? I don't really see the difference. What kind of e2e test?
> 
> Right now, the number of sqlite readers isn't limited. So in theory, it would slow down the internal performance only once it exceeded the underlying hardware capabilities. With your change, you're capping it at the driver level. That could lead to starvation of a node's internal operation, which I'd rather avoid.

If it's only a matter of number 128, we can increase it. However, I don't think a bigger number would make a difference.

tolikzinovyev:
> > > As implemented, I think that it's not a great implementation. It could slow down the internal processing due to a REST endpoint call. I would suggest creating a separate connection that would be used by the REST endpoints interface. However, before going into such an endeavor, I'd like to see some e2e test that shows that this is truly a viable issue.
> > 
> > 
> > Can't REST endpoints slow down internal processing without this change? I don't really see the difference. What kind of e2e test?
> 
> The e2e test that I'm thinking of would create a network, saturate it with some kind of load, and be able to cause the node to crash due to high number of concurrent database queries.

I wasn't able to cause a crash, probably because ssds are too fast or some sqlite caching was involved. In any case, the way I see it, with this change there is a proof that crash won't happen due to FD exhaustion, without this change I don't have such proof.

tsachiherman:
> > > > As implemented, I think that it's not a great implementation. It could slow down the internal processing due to a REST endpoint call. I would suggest creating a separate connection that would be used by the REST endpoints interface. However, before going into such an endeavor, I'd like to see some e2e test that shows that this is truly a viable issue.
> > > 
> > > 
> > > Can't REST endpoints slow down internal processing without this change? I don't really see the difference. What kind of e2e test?
> > 
> > 
> > Right now, the number of sqlite readers isn't limited. So in theory, it would slow down the internal performance only once it exceeded the underlying hardware capabilities. With your change, you're capping it at the driver level. That could lead to starvation of a node's internal operation, which I'd rather avoid.
> 
> If it's only a matter of number 128, we can increase it. However, I don't think a bigger number would make a difference.

No, the 128 number is good. It's the fact the these 128 database operations would be pulled out of the same pool as the internal operations.

tolikzinovyev:
> > > > > As implemented, I think that it's not a great implementation. It could slow down the internal processing due to a REST endpoint call. I would suggest creating a separate connection that would be used by the REST endpoints interface. However, before going into such an endeavor, I'd like to see some e2e test that shows that this is truly a viable issue.
> > > > 
> > > > 
> > > > Can't REST endpoints slow down internal processing without this change? I don't really see the difference. What kind of e2e test?
> > > 
> > > 
> > > Right now, the number of sqlite readers isn't limited. So in theory, it would slow down the internal performance only once it exceeded the underlying hardware capabilities. With your change, you're capping it at the driver level. That could lead to starvation of a node's internal operation, which I'd rather avoid.
> > 
> > 
> > If it's only a matter of number 128, we can increase it. However, I don't think a bigger number would make a difference.
> 
> No, the 128 number is good. It's the fact the these 128 database operations would be pulled out of the same pool as the internal operations.

There is a similar pool in the operating system with similar contention characteristics. If there is no contention in algod's pool, there will be contention in OS's pool.

PR #3336, opened 2021-12-20
testing: add testing for rewards rate calculation

AlgoStephenAkiki:
Resolves #3284 

Adds additional unit tests for the initial block calculation

PR #3403, opened 2022-01-10
ledger: fix `NextRewardsState()`

tolikzinovyev:
## Summary

A modification of #3336. Added a new test where the rewards pool overspends and proposed a fix in `NextRewardsState()` requiring a consensus upgrade.

## Test Plan

This is mostly tests.


data/bookkeeping/block_test.go
```diff
@@ -461,3 +462,72 @@ func TestInitialRewardsRateCalculation(t *testing.T) {
...
+		curRewardsState = nextRewardState
+	}
+}
+
+func TestNextRewardsRateWithFix(t *testing.T) {
+	partitiontest.PartitionTest(t)
+
+	params, ok := config.Consensus[protocol.ConsensusCurrentVersion]
+	require.True(t, ok)
+	params.RewardsCalculationFix = true
```
tsachiherman, line 496: The test is "almost there". You'll need to confirm expected failure without the fix, and expected succeess with the fix.

tsachiherman: It would also be a good idea to add expected failure after the fix, providing invalid input.

tolikzinovyev: Done (1). Why test on invalid input?

tsachiherman: testing it against invalid input ensures that the method does behave in the way we expect it to. I realize that there could be multitude of invalid inputs, and ideally, we need to analyze the codebase to ensure all the code-paths are taken and validated individually.

data/bookkeeping/block_test.go
```diff
@@ -248,20 +251,30 @@ func TestTime(t *testing.T) {
...
 	prev.RewardsLevel = 1
 	prev.RewardsRate = 10
 
 	rewardUnits := uint64(10)
-	state := prev.NextRewardsState(prev.Round()+1, proto, basics.MicroAlgos{}, rewardUnits)
+	state := prev.NextRewardsState(prev.Round()+1, proto, basics.MicroAlgos{}, rewardUnits, log)
 	require.Equal(t, uint64(2), state.RewardsLevel)
 	require.Equal(t, uint64(0), state.RewardsResidue)
+
+	assert.Zero(t, buf.Len())
```
tsachiherman, line 268: in unit tests, you almost always want to use `require` and not `assert`. If the expression fails, there is no point in keep running the test. ( in fact, it could lead into some other test failures )

tsachiherman (changes requested):
For testing the expected empty log output, your change is fine.
However, you did not test the expected output in case of an error.

tolikzinovyev:
> For testing the expected empty log output, your change is fine. However, you did not test the expected output in case of an error.

The last test does plus some long running tests check that log output is non-empty. What other errors are you referring to?

tolikzinovyev:
There are some other overflow checks in the code, but they really shouldn't happen...

tsachiherman:
> There are some other overflow checks in the code, but they really shouldn't happen...

Yes, I agree that it shouldn't happen under normal running conditions - but you're implementing a unit test; in this case, just explicitly check the output for the error string.

tolikzinovyev:
> > There are some other overflow checks in the code, but they really shouldn't happen...
> 
> Yes, I agree that it shouldn't happen under normal running conditions - but you're implementing a unit test; in this case, just explicitly check the output for the error string.

Do you want me to cause those overflows to happen and check the logs? I'm not following.

tsachiherman:
Yes, exactly. Pass artificial parameters that would reproduce the problem,
and confirm the implementation.

tsachiherman (approved):

PR #3405, opened 2022-01-11
Make TestAcctUpdates faster

algorandskiy:
* TestAcctUpdates used to compare acct updates content every round
  but this is slow due to amount of data needed to be compared.
* This fix suggest making comparison at first, end and approx every
  10 iterations.

### DIFF
--- a/ledger/acctupdates_test.go
+++ b/ledger/acctupdates_test.go
@@ -493,7 +493,9 @@ func TestAcctUpdates(t *testing.T) {
 	lastCreatableID := crypto.RandUint64() % 512
 	knownCreatables := make(map[basics.CreatableIndex]bool)
 
-	for i := basics.Round(10); i < basics.Round(proto.MaxBalLookback+15); i++ {
+	start := basics.Round(10)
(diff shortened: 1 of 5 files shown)


ledger/acctupdates_test.go
```diff
@@ -526,7 +528,11 @@ func TestAcctUpdates(t *testing.T) {
 		accts = append(accts, newAccts)
 		rewardsLevels = append(rewardsLevels, rewardLevel)
 
-		checkAcctUpdates(t, au, 0, i, accts, rewardsLevels, proto)
+		// checkAcctUpdates is kind of slow because of amount of data it needs to compare
+		// instead, compare at start, end in between approx 10 rounds
+		if i == start || i == end-1 || crypto.RandUint64()%10 == 0 {
```
tsachiherman, line 533: could you avoid using a real random number here and make it a pseudo random ? just so it will be consistent ?

algorandskiy: I intentionally made it truly random to cover different intervals across multiple runs

tsachiherman (approved):

PR #3406, opened 2022-01-12
ledger refactoring: track per-resource in NewAccountDeltas

cce:
## Summary

This changes NewAccountDeltas by merging the AppParams/AppLocalState and AssetParams/AssetHolding accounting from tracking 4 types to 2 types (more similar to DB storage than abstraction provided by balances interface). The apply.Balances interface implementation (roundCowBase/roundCowState) is responsible for converting from 4 to 2.


ledger/accountdb.go
```diff
@@ -516,22 +446,25 @@ func makeCompactResourceDeltas(accountDeltas []ledgercore.NewAccountDeltas, base
...
+					address:     res.Addr,
 					newResource: makeResourcesData(deltaRound * updateRoundMultiplier),
 				}
-				if appParams.Params != nil {
-					newEntry.newResource.SetAppParams(*appParams.Params, false)
+				if !res.State.Deleted && res.State.State != nil {
+					newEntry.newResource.SetAppLocalState(*res.State.State)
+				}
+				if !res.Params.Deleted && res.Params.Params != nil {
+					newEntry.newResource.SetAppParams(*res.Params.Params, false)
```
algorandskiy, line 456: false -> IsHolding

ledger/internal/appcow.go
```diff
@@ -670,32 +677,47 @@ func applyStorageDelta(cb *roundCowState, addr basics.Address, aapp storagePtr,
...
 			}
-			cb.mods.NewAccts.UpsertAppLocalState(addr, aapp.aidx, nil)
+			// fetch AppParams to store along with deleted AppLocalState
+			var params ledgercore.AppParamsDelta
+			if ap, exist, err := cb.lookupAppParams(addr, aapp.aidx); err != nil {
+				return fmt.Errorf("fetching storage (global=%v) failed for (%s, %d) AppLocalState: %w", aapp.global, addr.String(), aapp.aidx, err)
+			} else if exist {
+				params = ap
+			}
+			cb.mods.NewAccts.UpsertAppResource(addr, aapp.aidx, params, ledgercore.AppLocalStateDelta{Deleted: true})
```
algorandskiy, line 720: maybe we need to test dealloc local state and then dealloc app params in two subsequent tnxs

ledger/internal/cow_creatables.go
```diff
@@ -58,38 +59,106 @@ func (cs *roundCowState) CountAssetParams(addr basics.Address) (int, error) {
 }
 
 func (cs *roundCowState) GetAppParams(addr basics.Address, aidx basics.AppIndex) (ret basics.AppParams, ok bool, err error) {
-	return cs.lookupAppParams(addr, aidx)
+	d, ok, err := cs.lookupAppParams(addr, aidx)
+	if d.Params != nil {
+		ret = *d.Params
+	}
+	if d.Deleted { // XXX return ok=false for deleted?
```
algorandskiy, line 66: check the old implementation:
```
func (cb *roundCowState) lookupAppParams(addr basics.Address, aidx basics.AppIndex) (basics.AppParams, bool, error) {
	params, ok := cb.mods.NewAccts.GetAppParams(addr, aidx)
	if ok {
		if params == nil {
			return basics.AppParams{}, false, nil
		}
		return *params, ok, nil
	}
```
=> ok = false for deleted

cce: oh, that's weird, so lookupAppParams is now going to return ok=true for deleted, but the caller here has to change it to ok=false for the implementation of apply.Balances.

cce: so I have to audit all the other uses of roundCowState.lookupX? hrmm.

algorandskiy: maybe lookupAppParams can do the analysis as before?

ledger/ledgercore/statedelta.go
```diff
@@ -113,32 +113,44 @@ type NewBalanceRecord struct {
...
+	Deleted bool
 }
 
-// AssetHoldingRecord represents asset holding in deltas
-type AssetHoldingRecord struct {
+// AppResourceRecord represents AppParams and AppLocalState in deltas
+type AppResourceRecord struct {
+	Aidx   basics.AppIndex
+	Addr   basics.Address
+	Params AppParamsDelta
```
tsachiherman, line 144: maybe `Params` -> `GlobalParams`, and `State` -> `LocalState` ?

ledger/ledgercore/statedelta.go
```diff
@@ -190,10 +195,8 @@ func MakeNewAccountDeltas(hint int) NewAccountDeltas {
 		accts:      make([]NewBalanceRecord, 0, hint*2),
 		acctsCache: make(map[basics.Address]int, hint*2),
 
-		appParamsCache:      make(map[AccountApp]int),
-		appLocalStatesCache: make(map[AccountApp]int),
-		assetParamsCache:    make(map[AccountAsset]int),
-		assetHoldingsCache:  make(map[AccountAsset]int),
+		appResourcesCache:   make(map[AccountApp]int),
+		assetResourcesCache: make(map[AccountAsset]int),
```
tsachiherman, line 199: maybe initialize the `assetResourcesCache`/`appResourcesCache` with the size of the `hint` ?
most of the performance cost is when you get to the point where you start copying data.

ledger/ledgercore/statedelta.go
```diff
@@ -563,10 +496,10 @@ func (ad NewAccountDeltas) GetBasicsAccountData(addr basics.Address) (basics.Acc
 
 // ToModifiedCreatables creates map of ModifiedCreatable
 func (ad NewAccountDeltas) ToModifiedCreatables(seen map[basics.CreatableIndex]struct{}) map[basics.CreatableIndex]ModifiedCreatable {
-	result := make(map[basics.CreatableIndex]ModifiedCreatable, len(ad.appParams)+len(ad.assetParams))
-	for aapp, idx := range ad.appParamsCache {
-		rec := ad.appParams[idx]
-		if rec.Params == nil {
+	result := make(map[basics.CreatableIndex]ModifiedCreatable, len(ad.appResources)+len(ad.assetResources))
```
tsachiherman, line 499: please add a comment on this method that it's only being used in tests.

ledger/ledgercore/statedelta.go
cce: OK!

ledger/ledgercore/statedelta.go
```diff
@@ -207,41 +210,41 @@ func (ad NewAccountDeltas) GetData(addr basics.Address) (AccountData, bool) {
 }
 
 // GetAppParams returns app params delta value
-func (ad NewAccountDeltas) GetAppParams(addr basics.Address, aidx basics.AppIndex) (*basics.AppParams, bool) {
-	idx, ok := ad.appParamsCache[AccountApp{addr, aidx}]
-	var result *basics.AppParams
+func (ad NewAccountDeltas) GetAppParams(addr basics.Address, aidx basics.AppIndex) (AppParamsDelta, bool) {
+	idx, ok := ad.appResourcesCache[AccountApp{addr, aidx}]
```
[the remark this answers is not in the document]
cce: with latest changes from @algorandskiy I can get it down to
```
    if idx, ok := ad.appResourcesCache[AccountApp{addr, aidx}];	ok {
	result := ad.appResources[idx].Params
        return result, result.Deleted || result.Params != nil
    }
    return AppParamsDelta{}, false
}```

ledger/ledgercore/statedelta.go
cce: done! additionally, if we wanted to get really paranoid, we could put a init() function is _test.go code to set an env variable or flag that needs to be true to use functions like these.

ledger/ledgercore/statedelta.go
cce: I can change State to LocalState but I was aiming for using only a "words after App/Asset" naming convention like Holding, Params, etc.

algorandskiy (approved):

PR #3411, opened 2022-01-13
add context to (most) agreement logged writes

tsachiherman:
## Summary

Current agreement code only writes a `context : agreement` to a subset of the logged messages.
This change extends the said entry, which would make it easier to pre-process logs entries by their corresponding component. The change in this PR is focused on:
1. make sure that the "root" agreement logger always injects the `context : agreement` argument.
2. change the various locations in the agreement code to use the root agreement logger instead of referring to the application-global instance (`logging.Base()`).

## Test Plan

Test the output logs manually.


nicholasguoalgorand (approved):
Looks good.

PR #3417, opened 2022-01-14
txnsync: Use 100 txns/msg maximum for txnsync messages

nicholasguoalgorand:
## Summary

Use fixed message sizes for transaction sync to enable release.

## Test Plan

Ran performance tests with various message sizes and scenarios and chose the most reasonable size.


tsachiherman (changes requested):
The changes looks resonable, but please make sure that all the pr tests would pass.

tsachiherman (approved):

PR #3418, opened 2022-01-14
Feature/contract to contract

algoidurovic:
## Summary

Added integration tests for C2C:

- Test that basic and complex/indirect reentry isn't allowed
- Verify that the limit on inner transactions in a group is enforced
- Verify that failure on inner app call causes an abort for the entire transaction group
- Verify only the correct resources can be accessed. External resources explicitly passed in foreign object arrays are accessible along with apps and assets created previously in the tx group execution. Note: created accounts are not accessible at the moment.
- Verify v6 teal contracts cannot invoke contracts with previous versions through inner app calls
- Stress test: Create tx group execution with max fan out of inner txns and max call stack depth. Because a deep call stack maintains previous frames, memory usage may be a concern. Attempt to use the maximum allowed memory per stack frame.


ledger/internal/apptxn_test.go
```diff
@@ -1500,3 +1506,936 @@ func TestGtxnEffects(t *testing.T) {
...
+		ForeignApps:   []basics.AppIndex{index1, index0},
+	}
+	eval = nextBlock(t, l, true, nil)
+	txn(t, l, eval, &call1, "attempt to re-enter")
+	endBlock(t, l, eval)
+}
+
+// This tests a valid form of reentry (which may not be the correct word here).
+// When A calls B then returns to A then A calls C which calls B, the execution
+// should not produce an error because B doesn't occur in the call stack twice.
```
jannotti, line 1608: Nice idea.

ledger/internal/apptxn_test.go
```diff
@@ -1500,3 +1506,936 @@ func TestGtxnEffects(t *testing.T) {
...
+	for i := 1; i < 16; i++ {
+		callTxGroup[i] = txntest.Txn{
+			Type:          "appl",
+			Sender:        addrs[0],
+			ApplicationID: index1,
+			ForeignApps:   []basics.AppIndex{},
+		}
+	}
+	eval = nextBlock(t, l, true, nil)
+	err := txgroup(t, l, eval, &callTxGroup[0], &callTxGroup[1], &callTxGroup[2], &callTxGroup[3], &callTxGroup[4], &callTxGroup[5], &callTxGroup[6], &callTxGroup[7], &callTxGroup[8], &callTxGroup[9], &callTxGroup[10], &callTxGroup[11], &callTxGroup[12], &callTxGroup[13], &callTxGroup[14], &callTxGroup[15])
```
jannotti, line 2014: Since we decided to allow 256 calls even from a singleton app call (not a 16 app group) we should adjust this test.  I'll let you know when I have that merged in.

jannotti (approved):

PR #3424, opened 2022-01-16
avoid generating log error on EnsureValidatedBlock / EnsureBlock

algonautshant:
In EnsureBlock,, do not log as error message if the error is `ledgercore.ErrNonSequentialBlockEval` and the block round is in the past (i.e. already in the ledger). 

Added a test to verify the behavior. 

Resolves https://github.com/algorand/go-algorand-internal/issues/1831


tsachiherman (changes requested):
Thinking about this over the weekend, I think it would be better to "convert" the `ledgercore.ErrNonSequentialBlockEval` to `ledgercore.BlockInLedgerError` in `ledger.AddBlock` when err != nil and EvaluatorRound <= errNSBE.LatestRound.

The handling for both cases is the same and it would simplify the client code.

tsachiherman (changes requested):
in addition to the above, please also make sure to use distinguishable log messages, so that we could trace these back if/when needed.

tsachiherman (approved):
looks good.

algorandskiy (approved):

PR #3425, opened 2022-01-16
Fix stateproof in goal

algoidan:


data/account/participationRegistry.go
```diff
@@ -79,13 +79,17 @@ type ParticipationRecord struct {
...
+	StateProofPK *StateProofPublicKey
+	VRF          *crypto.VRFSecrets
+	Voting       *crypto.OneTimeSignatureSecrets
 }
 
 // StateProofKey defined the type used for the compact certificate signing key
 type StateProofKey crypto.GenericSigningKey
 
+// StateProofPublicKey defined the type used for the stateproofs public key
+type StateProofPublicKey merklekeystore.Verifier
```
Aharonee, line 91: Maybe StateProofVerifier instead and change StateProofKey -> StateProofSigner ?

data/account/participationRegistry.go
```diff
@@ -137,9 +141,17 @@ func (r ParticipationRecord) Duplicate() ParticipationRecord {
 		LastStateProof:    r.LastStateProof,
 		EffectiveFirst:    r.EffectiveFirst,
 		EffectiveLast:     r.EffectiveLast,
+		StateProofPK:      nil,
 		VRF:               &vrf,
 		Voting:            &voting,
 	}
+	if r.StateProofPK != nil {
```
Aharonee, line 148: Move this logic to line 132 and refactor to stay consistent with the way `VRF` and `Voting` are copied.

data/account/participationRegistry.go
```diff
@@ -704,11 +716,16 @@ func (db *participationDB) Insert(record Participation) (id ParticipationID, err
 		Voting:            voting,
 		VRF:               vrf,
 	}
+	if record.StateProofSecrets != nil {
+		var stateProofPk StateProofPublicKey
+		copy(stateProofPk[:], record.StateProofSecrets.GetVerifier()[:])
```
Aharonee, line 721: Same as above, I think you should stay consistent with how the voting the vrf are copied, and move this code to line 703.

data/account/participationRegistry.go
```diff
@@ -806,6 +825,17 @@ func scanRecords(rows *sql.Rows) ([]ParticipationRecord, error) {
...
 		}
 
+		if len(rawStateProof) > 0 {
+			stateProof := merklekeystore.Signer{}
+			err = protocol.Decode(rawStateProof, &stateProof)
+			if err != nil {
+				return nil, fmt.Errorf("unable to decode stateproof: %w", err)
+			}
+			var stateProofPk StateProofPublicKey
+			copy(stateProofPk[:], stateProof.GetVerifier()[:])
```
Aharonee, line 835: So when you `Insert` the participation into the registry you store the `SignerContext` in its entirety, but in the cache only the verifier is stored, and when retrieved from the DB only the verifier is extracted. Doesn't that seem a bit inconsistent and confusing?
The only place the entire `SignerContext` is used it when calling `GetForRound()`.
I think maybe we should completely separate the StateProof data from the Participation data, even store it in a different table `(ParticipationID, StateProofData)`. But we can talk about it for a different PR I guess.

data/account/participationRegistryBench_test.go
```diff
@@ -18,6 +18,7 @@ package account
 
 import (
 	"fmt"
+	"github.com/stretchr/testify/require"
```
Aharonee, line 21: reorder imports

agreement/keyManager_test.go
```diff
@@ -46,7 +45,6 @@ func (m *recordingKeyManager) VotingKeys(votingRound, _ basics.Round) []account.
 	var km []account.ParticipationRecordForRound
 	for _, acc := range m.keys {
 		if acc.OverlapsInterval(votingRound, votingRound) {
-			var signerInRound merklekeystore.Signer
 			if acc.StateProofSecrets != nil {
 				acc.StateProofSecrets.GetSigner(uint64(votingRound))
 			}
```
Aharonee, line 50: This statement doesn't seem to do much

data/account/participationRegistry.go
```diff
@@ -126,7 +135,15 @@ func (r ParticipationRecord) Duplicate() ParticipationRecord {
 	if r.Voting != nil {
 		voting = r.Voting.Snapshot()
 	}
-	return ParticipationRecord{
+
+	var stateProof *StateProofVerifier
+	var statproofVerifer StateProofVerifier
```
Aharonee, line 140: Why do you need these 2 variables instead of 1?

data/account/participationRegistry.go
```diff
@@ -689,6 +712,13 @@ func (db *participationDB) Insert(record Participation) (id ParticipationID, err
 		*voting = record.Voting.Snapshot()
 	}
 
+	var stateProofVeriferPtr *StateProofVerifier
+	var statproofVerifier StateProofVerifier
```
Aharonee, line 716: Again, 2 vars

PR #3438, opened 2022-01-19
non archival node catchpoint writing

tsachiherman:
## Summary

This PR makes the following changes:
1. It renames the `archivalLedger` in account updates tracker & catchpoint tracker to `enableGeneratingCatchpointFiles`, which better describe how this variable is being used.
2. It allows the end user to specify `CatchpointTracking:2` on non-archival nodes, and allow non-archival nodes to generate catchpoint files.
3. Adds support for `CatchpointTracking:99`, which is an internal debug option for enabling catchpoint writing on non-archival nodes even *before* the `accountDataResourceSeparationRound` has reached. ( i.e. before the consensus upgrade ).
4. Fix missing re-encoding when transcoding V5 database catchpoints onto a V6 database.

## Test Plan
Use the above features in order to test generated catchpoint files.


ledger/catchupaccessor.go
```diff
@@ -800,7 +806,7 @@ func (c *CatchpointCatchupAccessorImpl) finishBalances(ctx context.Context) (err
 	start := time.Now()
 	ledgerCatchpointFinishBalsCount.Inc(nil)
 	err = wdb.Atomic(func(ctx context.Context, tx *sql.Tx) (err error) {
-		var balancesRound uint64
+		var balancesRound, hashRound uint64
```
algorandskiy, line 809: not sure I understand why balancesRound and hashRound can be different, looks like we were updating them simultaneously.

ledger/acctupdates.go
```diff
@@ -259,7 +263,15 @@ func (r resourcesUpdates) getForAddress(addr basics.Address) (ret []modifiedReso
...
-	au.archivalLedger = cfg.Archival
+	switch cfg.CatchpointTracking {
+	case forceCatchpointFileGenerationTrackingMode:
+		// this CatchpointTracking is used for debugging of the catchpoint generation.
+		fallthrough
+	case 2:
+		au.enableGeneratingCatchpointFiles = true
+	default:
+		au.enableGeneratingCatchpointFiles = cfg.Archival
+	}
```
algorandskiy, line 274: can't this code moved to tracker.go and called from here and from ct ?

algorandskiy (commented):
Looks good, couple suggestions

ledger/catchupaccessor.go
tsachiherman: In some cases, we want to reset the hashRound by setting it to zero.

algorandskiy (approved):

PR #3452, opened 2022-01-19
update arm32 builds from v6 to v7

algobarb:
## Summary

The Dockerfile we previously used for arm32 builds has conflicts with the time64 dependency introduced in alpine3.3.0. Updating to arm32v7 will allows us to build the binary on ubuntu instead.

## Test Plan

Will need to run the CI pipeline after the merge.


tsachiherman:
please make sure the resulting binary is still compatible with all the raspberry models we used to target for..

algobarb:
> please make sure the resulting binary is still compatible with all the raspberry models we used to target for..

Raspberry PI 1 may no longer be compatible, but Raspberry pi 2 uses ARMv7 so binary should continue to work for Raspberry Pi 2+.

yaovi-a:
Can someone even run node reliability on Raspberry Pi 1 anymore?

tsachiherman:
> Can someone even run node reliability on Raspberry Pi 1 anymore?

The short answer is yes. But I'd argue that the question is wrong. The correct question is "should we have Raspberry Pi 1 as one of our target platforms ?", and that question should be answered from a product-value proposition and not from a technical perspective.

algobarb:
> > Can someone even run node reliability on Raspberry Pi 1 anymore?
> 
> The short answer is yes. But I'd argue that the question is wrong. The correct question is "should we have Raspberry Pi 1 as one of our target platforms ?", and that question should be answered from a product-value proposition and not from a technical perspective.

Technically, also possibly no. We don't test the binary on old Raspberry Pis for our releases (John has tested them on raspberry pis in some of our previous releases) and our minimum recommended node requirements exceed the capabilities of older Raspberry Pi's models.

onetechnical:
> Technically, also possibly no. We don't test the binary on old Raspberry Pis for our releases (John has tested them on raspberry pis in some of our previous releases) and our minimum recommended node requirements exceed the capabilities of older Raspberry Pi models.

I also use a Raspberry Pi 4, which is am armv7, not armv6.

algojack (approved):

onetechnical (approved):
This is still in discussion for stable, but approved for beta.

PR #3453, opened 2022-01-19
Fix a potential problem of committing non-uniform consensus versions

algorandskiy:
## Summary

If accountdb accumulates a large backlog, it is possible catchpoint tracker would attempt to commit a wider range than account updates tracker expects.

## Test Plan

Added a test simulating said behavior.


tsachiherman (approved):
The change looks good to me.

PR #3459, opened 2022-01-20
Goal: adding auth addr to signed txn if signer is passed

barnjamin:
## Summary

When trying to sign a transaction file with goal, if the sender is a normal address but the signer is a logic program, the call fails with an error message like:

```
send.txn: txn[0] error LogicNot signed and not a Logic-only account
```

The error message comes from https://github.com/algorand/go-algorand/blob/33b87c432065d3be0ce1fdad682604f416676133/data/transactions/verify/txn.go#L314

It notes that a valid case is when Auth is set to the hash of the program but we aren't setting the auth on the transaction even if the -S flag is passed

## Test Plan

Made a txn where sender is a regular account and signed it with 

```
goal clerk sign -i tmp.txn -o tmp.txn -S BJATCHES5YJZJ7JITYMVLSSIQAVAWBQRVGPQUDT5AZ2QSLDSXWWM46THOY -p tmp.teal
```


cmd/goal/clerk.go
```diff
@@ -787,10 +787,17 @@ var signCmd = &cobra.Command{
 
 		for _, group := range groupsOrder {
 			txnGroup := []transactions.SignedTxn{}
-			for _, txn := range txnGroups[group] {
+			for idx, txn := range txnGroups[group] {
```
tsachiherman, line 790: doesn't `client.SignTransactionWithWalletAndSigner` already does that ? I think that this would only be needed in the case of `lsig.Logic != nil`.

barnjamin: Sorry, do you mean move this if case into the lsig != nil check?

tsachiherman: I think so. But I could be wrong.
I think that the correct way to approach this is to have this tested.
We have some expect tests for the goal commands. I think that extending those to cover this use case wound't be a bad idea.

barnjamin: Tweaked the logic but I'm struggling to find where the tests mentioned live

tsachiherman: The tests lives in /go-algorand/test/e2e-go/cli/goal/expect

PR #3464, opened 2022-01-21
ledger refactoring: closeout tests

algorandskiy:


algorandskiy:
Impossible to recreate with a real DB

PR #3478, opened 2022-01-21
pingpong: improve transaction scheduling

tolikzinovyev:
## Summary

Currently the transaction scheduling logic in pingpong is lots of complicated sleep calls. Some sleeps were definitely extra and were hurting tps; this PR deprecates them. The whole scheduling logic is replaced with a much simpler one. Let timestamp sequence `{t}_{i >= 0} = {currenttime + 1 / tps * i}`. Then transaction `i` is sent at time `t_i` with best effort.

In my tests `./pingpong run --tps 115 --rest 0 --refresh 1800 --numaccounts 500` produced 440 txns per block. Now `./pingpong run --tps 115 --refresh 1800 --numaccounts 500` produces 475 txns per block. (`--rest` was deprecated)

Related to https://github.com/algorand/go-algorand-internal/issues/1833.

## Test Plan

I ran it.

### DIFF
--- a/shared/pingpong/pingpong.go
+++ b/shared/pingpong/pingpong.go
@@ -269,8 +269,7 @@ func schedule(tps uint64, nextSendTime *time.Time) {
 		time.Sleep(dur)
 	}
 
-	*nextSendTime = nextSendTime.Add(
-		time.Duration((1.0 / float64(tps)) * float64(time.Second)))
+	*nextSendTime = nextSendTime.Add(time.Second / time.Duration(tps))
 }
 
 func (pps *WorkerState) fundAccounts(accounts map[string]*pingPongAccount, client libgoal.Client, cfg PpConfig) error {
--- a/shared/pingpong/pingpong.go
+++ b/shared/pingpong/pingpong.go
@@ -262,6 +262,17 @@ func computeAccountMinBalance(client libgoal.Client, cfg PpConfig) (fundingRequi
 	return
 }
 
+// Wait for `*nextSendTime` and update it afterwards.
+func schedule(tps uint64, nextSendTime *time.Time) {
+	dur := time.Until(*nextSendTime)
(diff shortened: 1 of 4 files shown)


shared/pingpong/pingpong.go
```diff
@@ -249,6 +249,17 @@ func computeAccountMinBalance(client libgoal.Client, cfg PpConfig) (fundingRequi
 	return
 }
 
+// Wait for `*nextSendTime` and update it afterwards.
+func schedule(tps uint64, nextSendTime *time.Time) {
+	dur := time.Until(*nextSendTime)
+	if dur > 0 {
```
manotoor, line 255: should there be a check on this that dur is not a large number

shared/pingpong/pingpong.go
```diff
@@ -249,6 +249,17 @@ func computeAccountMinBalance(client libgoal.Client, cfg PpConfig) (fundingRequi
...
+		time.Sleep(dur)
+	}
+
+	*nextSendTime = nextSendTime.Add(
+		time.Duration((1.0 / float64(tps)) * float64(time.Second)))
```
brianolson, line 260:
```suggestion
		time.Duration(float64(time.Second) / float64(tps)))
```

brianolson: or better yet, at startup, calculate `period := Second/tps` then later `nextSendTime.Add(period)`

brianolson (commented):
the `schedule()` abstraction looks good and cleans up a bunch of code. One tweak suggestion.

shared/pingpong/pingpong.go
tolikzinovyev: done

brianolson (approved):

algorandskiy (approved):

PR #3484, opened 2022-01-23
ledger refactoring: fix lookupResources

tsachiherman:
## Summary

The lookupResources had a small bug: it wasn't returning the correct persistedResourcesData in case the requested account did not had any holding of the requested asset. That, in turn got stored into the resources cache, triggering
issues down the road.

## Test Plan

### DIFF
--- a/ledger/accountdb.go
+++ b/ledger/accountdb.go
@@ -1978,7 +1978,6 @@ func (qs *accountsDbQueries) lookupResources(addr basics.Address, aidx basics.Cr
 
 		// this should never happen; it indicates that we don't have a current round in the acctrounds table.
 		if err == sql.ErrNoRows {
-			data.data = makeResourcesData(0)
 			// Return the zero value of data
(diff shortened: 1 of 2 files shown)


ledger/accountdb.go
```diff
@@ -1971,12 +1971,14 @@ func (qs *accountsDbQueries) lookupResources(addr basics.Address, aidx basics.Cr
...
 				return protocol.Decode(buf, &data.data)
 			}
+			data.data = makeResourcesData(0)
 			// we don't have that account, just return the database round.
 			return nil
 		}
 
 		// this should never happen; it indicates that we don't have a current round in the acctrounds table.
 		if err == sql.ErrNoRows {
+			data.data = makeResourcesData(0)
```
algorandskiy, line 1981: this does not look needed as we return an error and something empty

tsachiherman: agree

algorandskiy (approved):

PR #3495, opened 2022-01-24
[WIP] Secp256r1 support for ECDSA verify and decompress opcodes

algoidurovic:
ECDSA opcodes only support the Secp256k1 curve. The goal of this PR is to extend the opcodes to support the Secp256r1 curve as well. Note: the PkRecover opcode is not covered here.

Testing included correctness and benchmark tests. Benchmark tests indicate a significant slowdown when the Secp256r1 curve is used, meaning the current opcode cost may not be viable.

Added some helper test functions.


jasonpaulos:
From [doc.go](https://github.com/algoidurovic/go-algorand/blob/93bed2f925c43643434f646d623e6bc5c89aceb4/data/transactions/logic/doc.go#L245) for `ecdsa_verify`:

> The signed data must be 32 bytes long, and signatures in lower-S form are only accepted

I suspect these constraints might not be true for the secp256r1 curve. And if that's the case we should document it (or decide to enforce them).

michaeldiamant:
Closing because:
* The PR is merged into https://github.com/algorand/go-algorand/pull/3514.
* New issue exists to refine secp256r1 curve cost:  https://github.com/algorand/go-algorand/issues/3776.

PR #3496, opened 2022-01-25
ledger refactoring: fix incorrect schema during catchpoint catchup

tsachiherman:
## Summary

During the catchpoint catchup, we create temporary tables.
These tables becomes the production table at the end of the catchup.

The current generated table `accountbase` was not using the correct schema.

## Test Plan

tested manually.


algorandskiy (approved):

Successahead (approved):

Successahead:
Approve

PR #3498, opened 2022-01-25
Revert `fillBytes` method to `bigIntToBytes` for lower golang version

ahangsu:
# Summary

We are probably not going to higher golang version (1.15+) at this point, while `fillBytes` from `big.Int` is introduced in 1.15.
Changing all the `fillBytes` in `ABI` to self-made `bigIntToBytes` in favor of lower go version.

### DIFF
--- a/data/abi/abi_encode.go
+++ b/data/abi/abi_encode.go
@@ -26,7 +26,7 @@ import (
 )
 
 // bigIntToBytes casts non-negative big integer to byte slice with specific byte length
-// NOTE: THIS IS A WORKAROUND FOR `fillBytes` METHOD BEFOER GOLANG 1.15+
+// NOTE: THIS IS A WORKAROUND FOR `fillBytes` METHOD BEFORE GOLANG 1.15+
 //       SHOULD BE REMOVED AFTER WE MOVE TO HIGHER VERSION
 func bigIntToBytes(x *big.Int, byteLen uint) ([]byte, error) {
 	if x.Cmp(big.NewInt(0)) < 0 {
--- a/data/abi/abi_encode.go
+++ b/data/abi/abi_encode.go
@@ -26,7 +26,7 @@ import (
 )
 
 // bigIntToBytes casts non-negative big integer to byte slice with specific byte length
-// NOTE: THIS IS A WORKAROUND FOR `fillBytes` METHOD BEFORE GOLANG 1.15+
+// DEPRECATED: THIS IS A WORKAROUND FOR `fillBytes` METHOD BEFORE GOLANG 1.15+
 //       SHOULD BE REMOVED AFTER WE MOVE TO HIGHER VERSION
 func bigIntToBytes(x *big.Int, byteLen uint) ([]byte, error) {
 	if x.Cmp(big.NewInt(0)) < 0 {
--- a/data/abi/abi_encode.go
+++ b/data/abi/abi_encode.go
@@ -38,9 +38,7 @@ func bigIntToBytes(x *big.Int, byteLen uint) ([]byte, error) {
 
 	buffer := make([]byte, byteLen)
 	intBytes := x.Bytes()
-	for i := 0; i < len(intBytes); i++ {
-		buffer[len(buffer)-1-i] = intBytes[len(intBytes)-1-i]
-	}
+	copy(buffer[int(byteLen)-len(intBytes):], intBytes)
 	return buffer, nil
 }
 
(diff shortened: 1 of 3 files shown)


jannotti (dismissed):
I might have preferred a fill-in method with the exact same signature and semantics as FillBytes, so the change would be even smaller, but I'mm totally fine with this.

Let's merge after checks pass.

data/abi/abi_encode.go
```diff
@@ -25,6 +25,25 @@ import (
...
+	}
+	if uint(x.BitLen()) > byteLen*8 {
+		return nil, fmt.Errorf("ABI: big Int To Bytes error: integer byte length > given byte length")
+	}
+
+	buffer := make([]byte, byteLen)
+	intBytes := x.Bytes()
+	for i := 0; i < len(intBytes); i++ {
+		buffer[len(buffer)-1-i] = intBytes[len(intBytes)-1-i]
+	}
```
tsachiherman, line 43: I could be reading this wrong, but it looks to me like you're trying to do
```golang
copy(buffer[byteLen-len(intBytes):], intBytes)
```

jannotti, line 43: Could this have been `copy(buffer, x.Bytes())`?

ahangsu: yes exactly, I am somewhat rusty on golang now, will follow JJ's comments and change this line accordingly.

jannotti:
> I might have preferred a fill-in method with the exact same signature and semantics as FillBytes, so the change would be even smaller, but I'mm totally fine with this.
> 
> Let's merge after checks pass.

For what it's worth, emulating FillBytes exactly would be annoying, as it is willing to take a buffer of any size that's big enough.  Our use case only needs to work on exactly the right size buffer.

data/abi/abi_encode.go
ahangsu: yes, please see Tsachi's comment above.

data/abi/abi_json_test.go
```diff
@@ -31,14 +31,18 @@ func TestRandomAddressEquality(t *testing.T) {
...
 		randomAddrInt, err := rand.Int(rand.Reader, upperLimit)
 		require.NoError(t, err, "cryptographic random int init fail")
 
-		randomAddrInt.FillBytes(addrBasics[:])
-		randomAddrInt.FillBytes(addrABI)
+		expected, err := bigIntToBytes(randomAddrInt, uint(addressByteSize))
+		require.NoError(t, err, "big int to byte conversion error")
+		for i := 0; i < addressByteSize; i++ {
+			addrBasics[i] = expected[i]
+			addrABI[i] = expected[i]
```
jasonpaulos, line 44: `copy` could be used here as well instead of this loop

jasonpaulos (approved):
Looks correct. I agree with the earlier comments about `copy` and I left another, but they are minor

ahangsu:
I tried to emulate a method like FillBytes before, but we would need to introduce new type alias for big Int (in favor of defining new methods), and a bunch of type casting to use new method, and I am reluctant to create a larger delta.

The current fix should be sufficient for the purpose of casting big.Int to specific length of byte slice, please let me know your thoughts.

jasonpaulos (approved):

data/abi/abi_encode.go
```diff
@@ -25,6 +25,23 @@ import (
 	"strings"
 )
 
+// bigIntToBytes casts non-negative big integer to byte slice with specific byte length
+// NOTE: THIS IS A WORKAROUND FOR `fillBytes` METHOD BEFORE GOLANG 1.15+
+//       SHOULD BE REMOVED AFTER WE MOVE TO HIGHER VERSION
```
winder, line 30: nit: putting deprecated in the comment here triggers some IDEs (like goland) to cross out the method.
```suggestion
// bigIntToBytes casts non-negative big integer to byte slice with specific byte length
// DEPRECATED: THIS IS A WORKAROUND FOR `fillBytes` METHOD BEFORE GOLANG 1.15+
//       SHOULD BE REMOVED AFTER WE MOVE TO HIGHER VERSION
```

data/abi/abi_encode_test.go
```diff
@@ -1162,7 +1165,7 @@ func TestParseArgJSONtoByteSlice(t *testing.T) {
 
 	for i, test := range tests {
 		t.Run(fmt.Sprintf("index=%d", i), func(t *testing.T) {
-			applicationArgs := [][]byte{}
+			applicationArgs := make([][]byte, 0)
```
jannotti, line 1168: Need not put a 0 here.

jannotti (approved):

PR #3507, opened 2022-01-25
ledger refactoring: remove index and type from `ledgercore.AccountResource`

tolikzinovyev:
## Summary

The ledger for evaluator interface's `LookupResource()` returns creatable index and type inside `ledgercore.AccountResource` even though the caller knows what they are. This PR removes these redundant fields.

## Test Plan

This is just refactoring.

### DIFF
--- a/ledger/acctupdates.go
+++ b/ledger/acctupdates.go
@@ -244,16 +244,16 @@ func (r resourcesUpdates) get(ac accountCreatable) (m modifiedResource, ok bool)
 	return
 }
 
-func (r resourcesUpdates) getForAddress(addr basics.Address) (ret []modifiedResource) {
-	if len(r) == 0 {
-		return
-	}
-	for ac, mres := range r {
-		if ac.address == addr {
-			ret = append(ret, mres)
+func (r resourcesUpdates) getForAddress(addr basics.Address) map[basics.CreatableIndex]modifiedResource {
+	res := make(map[basics.CreatableIndex]modifiedResource)
+
+	for k, v := range r {
+		if k.address == addr {
+			res[k.index] = v
 		}
 	}
-	return
+
+	return res
 }
 
 // initialize initializes the accountUpdates structure
@@ -817,8 +817,6 @@ func (au *accountUpdates) newBlockImpl(blk bookkeeping.Block, delta ledgercore.S
 		mres, _ := au.resources.get(key)
 		mres.resource.AssetHolding = res.Holding.Holding
 		mres.resource.AssetParams = res.Params.Params
-		mres.resource.CreatableIndex = basics.CreatableIndex(res.Aidx)
-		mres.resource.CreatableType = basics.AssetCreatable
 		mres.ndeltas++
 		au.resources.set(key, mres)
 	}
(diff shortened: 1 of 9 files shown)


ledger/acctupdates.go
```diff
@@ -114,6 +114,9 @@ type accountCreatable struct {
 }
 
 type modifiedResource struct {
+	// TODO: delete
```
tsachiherman, line 117: This "todo" is right. There is more work here around removing this index from here and making sure getForAddress returns the index from the map's key.

tolikzinovyev: done

tsachiherman (dismissed):
I think that this change is good - but I'd like @cce  and @algorandskiy take another look as well.

ledger/ledgercore/accountresource.go
```diff
@@ -22,44 +22,37 @@ import (
...
-	switch r.CreatableType {
-	case basics.AssetCreatable:
-		if r.AssetParams != nil {
-			if ad.AssetParams == nil {
-				ad.AssetParams = make(map[basics.AssetIndex]basics.AssetParams)
-			}
-			ad.AssetParams[basics.AssetIndex(r.CreatableIndex)] = *r.AssetParams
+// AssignAccountResourceToAccountData assignes the Asset/App params/holdings contained
+// in the AccountResource to the given basics.AccountData, creating maps if necessary.
+func AssignAccountResourceToAccountData(cindex basics.CreatableIndex, resource AccountResource, ad *basics.AccountData) {
```
algorandskiy, line 33: maybe we move AccountResource to `basics` package and and add 
```
func (ad * AccountData) AssignResource(cidx CreatableIndex, resource AccountResource)
```
there?

ledger/acctupdates.go
```diff
@@ -985,8 +981,8 @@ func (au *accountUpdates) lookupLatest(addr basics.Address) (data basics.Account
 		}
 
 		// check for resources modified in the past rounds, in the deltas
-		for _, mr := range au.resources.getForAddress(addr) {
-			if err := addResource(mr.resource.CreatableIndex, rnd, mr.resource); err != nil {
+		for cindex, mr := range au.resources.getForAddress(addr) {
```
algorandskiy, line 984: please name it `cidx` for consistency, `cindex` looks like came from `evalindexer.go`. Naming same things by same names reduces cognitive load while reading the code.

tolikzinovyev: done

ledger/ledgercore/accountresource.go
tolikzinovyev: @tsachiherman, @cce are you ok with it? I don't know what your plans for `basics.AccountData` are.

algorandskiy: We can do it in another PR since it can be large.

tsachiherman: I don't think that we can move this to `basics` due to the `AccountResource`, but I'm wondering about the choice of having this as independent method instead of a member function:
```golang
func (r *AccountResource) AssignAccountData(ad *basics.AccountData, cidx basics.CreatableIndex)
```

cce: I had assumed some of these types will need to move into basics, AccountResource seems pretty harmless to move in there..

tsachiherman: I'm not sure if `AccountResource` is a good candidate for an object that lives outside the ledger package. The fact that we've been able to keep it in ledgercore is a good indication that it probably *shouldn't* be exposed. The *content* of that structure might need to be exported, through.

tsachiherman (approved):

PR #3513, opened 2022-01-26
ledger refactoring: add unique index on address to catchpointbalances

cce:
This index was in the migration schema but not the catchpoint restore schema.


tsachiherman (approved):
Looks good to me!

PR #3514, opened 2022-01-26
AVM: FIDO2 OpCodes

tzaffi:
# Experimental FIDO2 OpCodes

This is a branch containing OpCodes which are needed for FIDO2 authentication:

* `base64_decode`
* `json_ref`
* `ecdsa_*` opcodes for new curve`Secp256r1`

The usage patterns are not set in stone and these op codes may change before being released.
## Note
This PR is NOT intended to cause any changes in the current protocol or even, really, in vFuture. These opcodes need to be gated so that the new opcodes only get turned on by testers who want it (we will maintain a small feature branch with the changes to turn them on).

## TESTING

- [x] [Commit](https://github.com/algorand/go-algorand/pull/3514/commits/b38c152cc510ecaf9c893ecb9330abd94189913a) `b38c152c` with `fidoVersion = 7` passes all tests
- [x] [Commit](https://github.com/algorand/go-algorand/pull/3514/commits/d566072b7be11fe270c1ec3d96ab0064f27ddf36) `d566072b` with `fidoVersion = LogicVersion + 1` passes all tests


data/transactions/logic/opcodes.go
```diff
@@ -23,6 +23,11 @@ import (
 // LogicVersion defines default assembler and max eval versions
 const LogicVersion = 6
 
+// TODO: remove the following temporary constant when FIDO2 strategy finalized
+
+// LogicVersionInConsideration defines experimental assembler versions that are under consideration
+const LogicVersionInConsideration = 42
```
jannotti, line 29: I think I'd prefer this to be `LogicVersion+1`.

Perhaps the naming should be even more specific `const fidoVersion = LogicVersion + 1`

Then all three opcodes will be grouped, and it's easy to have a branch that turns them all on.  This might be a ncie way forward with other things, like exploring `splice` opcodes too.

data/transactions/logic/eval_test.go
```diff
@@ -5112,6 +5112,7 @@ func b64TestDecodeEval(tb testing.TB, args []b64DecodeTestArgs) {
 }
 
 func TestOpBase64Decode(t *testing.T) {
+	t.Skip("TODO: remove this skip when FIDO2 is accepted")
```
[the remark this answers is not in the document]
tzaffi: Will work to remove skip and provide more fine grained version-based expectations.

data/transactions/logic/eval_test.go
```diff
@@ -5117,3 +5129,399 @@ func TestOpBase64Decode(t *testing.T) {
...
+		source             string
+		previousVersErrors []expect
+	}{
+		{
+			source: `byte  "{\"key0\": 0,\"key1\": \"algo\",\"key2\":{\"key3\": \"teal\", \"key4\":3}, \"key5\": 18446744073709551615 }"; 
+			byte "key0"; 
+			json_ref JSONUint64; 
+			int 0; 
+			==`,
+			previousVersErrors: []expect{{5, "unknown opcode: json_ref"}, {9, "== arg 0 wanted type uint64 got []byte"}},
```
tzaffi, line 5194: @shiqizng - I ended adding expected errors for when attempting to run on old TEAL versions

cmd/opdoc/opdoc.go
```diff
@@ -35,7 +35,10 @@ func opGroupMarkdownTable(names []string, out io.Writer) {
 	opSpecs := logic.OpsByName[logic.LogicVersion]
 	// TODO: sort by logic.OpSpecs[].Opcode
 	for _, opname := range names {
-		spec := opSpecs[opname]
+		spec, ok := opSpecs[opname]
```
tzaffi, line 38: Skip future ops

data/transactions/logic/assembler_test.go
```diff
@@ -344,25 +344,43 @@ itxna Logs 3
...
+const v7CompiledNonsense = v6CompiledNonsense + "5c005d01"
+
 var nonsense = map[uint64]string{
 	1: v1Nonsense,
 	2: v2Nonsense,
 	3: v3Nonsense,
 	4: v4Nonsense,
 	5: v5Nonsense,
 	6: v6Nonsense,
+
```
tzaffi, line 369: to pass test with `LogicVersion == fidoVersion == 7` unocmment the following:

data/transactions/logic/opcodes.go
```diff
@@ -21,7 +21,10 @@ import (
 )
 
 // LogicVersion defines default assembler and max eval versions
-const LogicVersion = 6
+const LogicVersion = 7
```
tzaffi, line 24: Should I bump up the version like I'm doing here? I'm assuming that this is going out _after_ v. 6 and that new code should therefore have a higher version.

data/transactions/logic/opcodes.go
```diff
@@ -21,7 +21,10 @@ import (
...
+
+// TODO: remove the following temporary constant when FIDO2 strategy finalized
+const fidoVersion = LogicVersion
```
tzaffi, line 27: Maybe `fidoversion` should be +1 ?

data/transactions/logic/evalCrypto_test.go
```diff
@@ -339,6 +341,128 @@ ecdsa_verify Secp256k1`, hex.EncodeToString(r), hex.EncodeToString(s), hex.Encod
...
+	byteLen := (curve.Params().BitSize + 7) / 8
+	compressed := make([]byte, 1+byteLen)
+	compressed[0] = byte(y.Bit(0)) | 2
+	bitIntFillBytes(x, compressed[1:])
+	return compressed
+}
+
+func TestEcdsaWithSecp256r1(t *testing.T) {
+	if LogicVersion < fidoVersion {
+		return
```
tzaffi, line 358: No assertions being made about panicking for previous logic versions. Do we still need?

data/transactions/logic/eval.go
```diff
@@ -2664,7 +2718,7 @@ func opEcdsaVerify(cx *EvalContext) {
 		return
 	}
 
-	if fs.field != Secp256k1 {
+	if fs.field != Secp256k1 && fs.field != Secp256r1 {
```
tzaffi, line 2721: I think we should have a version assertion here.

tzaffi:
I'm planning to merge this _after_ `feature/contract-to-contract` has been merged. I expect some small changes to arise because of this (mostly comments).

data/transactions/logic/TEAL_opcodes.md
```diff
@@ -55,9 +55,10 @@ The 32 byte public key is the last element on the stack, preceded by the 64 byte
 
 `ECDSA` Curves:
 
-| Index | Name | Notes |
-| - | ------ | --------- |
-| 0 | Secp256k1 | secp256k1 curve |
+| Index | Name | In | Notes |
+| - | ------ | - | --------- |
+| 0 | Secp256k1 |      | secp256k1 curve |
+| 1 | Secp256r1 | v8  | secp256r1 curve |
```
tzaffi, line 61: Do we want this?

data/transactions/logic/evalCrypto_test.go
```diff
@@ -323,34 +325,157 @@ ecdsa_verify Secp256k1`, hex.EncodeToString(r), hex.EncodeToString(s), hex.Encod
...
+// TODO: replace with elliptic.MarshalCompressed when updating to go 1.15+
+func marshalCompressed(curve elliptic.Curve, x, y *big.Int) []byte {
+	byteLen := (curve.Params().BitSize + 7) / 8
+	compressed := make([]byte, 1+byteLen)
+	compressed[0] = byte(y.Bit(0)) | 2
+	bitIntFillBytes(x, compressed[1:])
+	return compressed
 }
 
+// broken due to missing testAcceptsWithField
```
tzaffi, line 340: TODO

data/transactions/logic/eval_test.go
```diff
@@ -4679,3 +4677,400 @@ By Herman Melville`, "",
...
+		},
+	}
+	for _, s := range noDuplicates {
+		hasDuplicates, _, err := hasDuplicateKeys(s.text)
+		require.Nil(t, err)
+		require.False(t, hasDuplicates)
+	}
+}
+
+// Currently broken as depends on defaultEvalProtoWithVersion
```
tzaffi, line 4717: TODO

data/transactions/logic/eval_test.go
```diff
@@ -3910,15 +3917,28 @@ func obfuscate(program string) string {
 
 type evalTester func(pass bool, err error) bool
 
-func testEvaluation(t *testing.T, program string, introduced uint64, tester evalTester) error {
+func testEvaluation(t *testing.T, program string, introduced uint64, tester evalTester, xtras ...uint64) error {
```
tzaffi, line 3920: I needed to add `xtras` to re-gain the ECDSA tests which relied on `testEvaluationWithField()`

data/transactions/logic/fields.go
```diff
@@ -568,6 +573,47 @@ func (s base64EncodingSpecMap) getExtraFor(name string) (extra string) {
...
+type jsonRefSpec struct {
+	field   JSONRefType
+	ftype   StackType
+	version uint64
+}
+
+var jsonRefSpecs = []jsonRefSpec{
+	{JSONString, StackBytes, 6},
+	{JSONUint64, StackUint64, 6},
+	{JSONObject, StackBytes, 6},
```
tzaffi, line 601:
```suggestion
	{JSONString, StackBytes, fidoVersion},
	{JSONUint64, StackUint64, fidoVersion},
	{JSONObject, StackBytes, fidoVersion},
```

data/transactions/logic/opcodes.go
```diff
@@ -58,8 +58,8 @@ const txnEffectsVersion = 6
 // the Foreign arrays.
 const createdResourcesVersion = 6
 
-// "Future" opcodes
-const fidoVersion = LogicVersion + 1 // base64, json, secp256r1
+// "Future" opcodes (may end up pushed back past current LogicVersion)
```
tzaffi, line 61:
```suggestion
// "Future" opcodes - experimental and may get pushed back further or removed
```

data/transactions/logic/doc.go
```diff
@@ -101,16 +101,17 @@ var opDocByName = map[string]string{
...
+	"load":     "Ith scratch space value. All scratch spaces are 0 at program start.",
+	"store":    "store A to the Ith scratch space",
+	"loads":    "Ath scratch space value.  All scratch spaces are 0 at program start.",
+	"stores":   "store B to the Ath scratch space",
+	"gload":    "Ith scratch space value of the Tth transaction in the current group",
+	"gloads":   "Ith scratch space value of the Ath transaction in the current group",
+	"gloadss":  "Bth scratch space value of the Ath transaction in the current group",
+	"gaid":     "ID of the asset or application created in the Tth transaction of the current group",
+	"gaids":    "ID of the asset or application created in the Ath transaction of the current group",
+	"json_ref": "return key B's value from a [valid](jsonspec.md) utf-8 encoded json object A",
```
jannotti, line 114: Can you put a blank line before "json_ref" and reformat?  It's logically quite different, and that way the diff will be a lot smaller, since the strings above won't get pushed right.

data/transactions/logic/eval.go
jannotti: I can't see a reason for this entire `if` clause (here, or in the other opEcdsa functions).  Doesn't the previous `!ok` check handle it?

data/transactions/logic/eval_test.go
jannotti: I think I recall why it went away.  I think I made the error messages more uniform, so that the version test would be the  same for both. I don't think there's a great reason for one to be "... was introduced .." and the other to be "...available in version...".

jannotti (commented):
I think this is good.  Note to further reviewers: This PR is NOT intended to cause any changes in the current protocol.  These opcodes needs to be gated so that the new opcodes only get turned on by testers who want it (we will maintain a small feature branch with the changes to turn them on).

With that in mind, I think fidoVersion should be set to LogicSigVersion+1 before merging. (And set to LogicSigVersion only in the small feature branch.)

config/consensus.go
```diff
@@ -1104,6 +1104,8 @@ func initConsensusProtocols() {
 	// Remove limits on maximum number of apps a single account can opt into
 	v32.MaxAppsOptedIn = 0
 
+	v32.LogicSigVersion = 7
+
```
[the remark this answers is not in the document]
tzaffi: 👍

data/transactions/logic/doc.go
tzaffi: 👍

data/transactions/logic/eval.go
jannotti: We should ensure that we have a unit test that tests behavior when the field is wrong.  This is somewhat hard to write: You have to compile the TEAL, then edit the byte to be wrong, since the assembler isn't going to let you create the bytecode with an invalid field number.  We have some tests like that though.

If it is caught by normal evaluation (it is unless I'm reading wrong), I'd really not like to see us checking the same thing that was checked two lines earlier.

data/transactions/logic/opcodes.go
```diff
@@ -58,7 +58,7 @@ const txnEffectsVersion = 6
 // the Foreign arrays.
 const createdResourcesVersion = 6
 
-// "Future" opcodes
+// "Future" opcodes - experimental and may get pushed back further or removed
```
tzaffi, line 61:
```suggestion
// Experimental:
```

data/transactions/logic/opcodes.go
```diff
@@ -58,8 +58,8 @@ const txnEffectsVersion = 6
 // the Foreign arrays.
 const createdResourcesVersion = 6
 
-// "Future" opcodes
-const fidoVersion = LogicVersion + 1 // base64, json, secp256r1
+// Experimental:
```
tzaffi, line 61:
```suggestion
// experimental:
```

data/transactions/logic/opcodes.go
```diff
@@ -58,8 +58,8 @@ const txnEffectsVersion = 6
 // the Foreign arrays.
 const createdResourcesVersion = 6
 
-// "Future" opcodes
-const fidoVersion = LogicVersion + 1 // base64, json, secp256r1
+// experimental:
```
tzaffi, line 61:
```suggestion
// experimental-
```

jannotti (approved):

PR #3523, opened 2022-01-27
Add logs in catchpoint service

tolikzinovyev:
## Summary

Add logs in catchpoint service measuring the time it takes to
- write catchpoint balances to disk
- download the catchpoint + write balances to disk
- build the merkle trie

This PR also simplifies `util.WatchdogStreamReader`: `Reset()` now doesn't return an error. This allows us to delete redundant code on the user side. Also, exiting when `Reset()` returns an error in `catchup/ledgerFetcher.go` is not good because `tarReader` might still contain buffered data. The implementation of `tarReader` doesn't, but nothing in its specification says it doesn't.

## Test Plan

I ran fast catchup and verified that logs are printed.

### DIFF
--- a/catchup/catchpointService.go
+++ b/catchup/catchpointService.go
@@ -287,10 +287,14 @@ func (cs *CatchpointCatchupService) processStageLedgerDownload() (err error) {
 			return cs.abort(err)
 		}
 		peer := psp.Peer
+		start := time.Now()
 		err = ledgerFetcher.downloadLedger(cs.ctx, peer, round)
 		if err == nil {
+			cs.log.Infof("ledger downloaded in %d seconds", time.Since(start) / time.Second)
+			start = time.Now()
 			err = cs.ledgerAccessor.BuildMerkleTrie(cs.ctx, cs.updateVerifiedAccounts)
 			if err == nil {
+				cs.log.Infof("built merkle trie in %d seconds", time.Since(start) / time.Second)
 				break
 			}
 			// failed to build the merkle trie for the above catchpoint file.
--- a/catchup/catchpointService.go
+++ b/catchup/catchpointService.go
@@ -290,11 +290,11 @@ func (cs *CatchpointCatchupService) processStageLedgerDownload() (err error) {
 		start := time.Now()
 		err = ledgerFetcher.downloadLedger(cs.ctx, peer, round)
 		if err == nil {
-			cs.log.Infof("ledger downloaded in %d seconds", time.Since(start) / time.Second)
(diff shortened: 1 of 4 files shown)


tsachiherman (changes requested):
Three requests:
1. I couldn't understand if you have found a concrete issue with the watchdogReader or not. Unless you found an issue, I'd rather keeping the existing code as is. Any code removal ( like this ), need to have some unit tests to proves that it would not become a bug. I don't want to ask you to provide any such validations, since it's not within the scope of your task.
2. I don't think that the time-measurements in processStageLedgerDownload are the optimal place to make these.
      In particular, I would suggest the following:
      - in getPeerLedger, use a deferred method to test the returned err variable if its zero, log the elapsed *download* time.
      - in CatchpointCatchupAccessorImpl.ProgressStagingBalances, you can add duration counters to CatchpointCatchupAccessorProgress and log the time taken. You might want to create a breakdown of the various goroutines in processStagingBalances.
      - in CatchpointCatchupAccessorImpl.BuildMerkleTrie, add time measurement, and log it.
3. I'd rather avoid merging this PR to master at this time. It's ok that you're testing it against master, but we should just merge to the feature branch. Later on, the feature branch will get merged into master.

tolikzinovyev:
> Three requests:
> 
>     1. I couldn't understand if you have found a concrete issue with the watchdogReader or not. Unless you found an issue, I'd rather keeping the existing code as is. Any code removal ( like this ), need to have some unit tests to proves that it would not become a bug. I don't want to ask you to provide any such validations, since it's not within the scope of your task.
> 
>     2. I don't think that the time-measurements in processStageLedgerDownload are the optimal place to make these.
>        In particular, I would suggest the following:
>        
>        * in getPeerLedger, use a deferred method to test the returned err variable if its zero, log the elapsed _download_ time.
>        * in CatchpointCatchupAccessorImpl.ProgressStagingBalances, you can add duration counters to CatchpointCatchupAccessorProgress and log the time taken. You might want to create a breakdown of the various goroutines in processStagingBalances.
>        * in CatchpointCatchupAccessorImpl.BuildMerkleTrie, add time measurement, and log it.
> 
>     3. I'd rather avoid merging this PR to master at this time. It's ok that you're testing it against master, but we should just merge to the feature branch. Later on, the feature branch will get merged into master.

1. Well, if `"archive/tar"` changes its implementation, it will become a concrete issue.
2. What is wrong with the current approach? In my opinion, it is simpler than what you are proposing.
3. Sure.

tolikzinovyev:
Closing in favor of https://github.com/algorand/go-algorand/pull/3531.

PR #3526, opened 2022-01-28
Remove rtype from resources table

algorandskiy:
It appears rtype is not really used and it is safe to remove.
DB-level code deducts resource type from `data.ResourceFlags`


ledger/accountdb.go
```diff
@@ -1971,13 +1959,22 @@ func (qs *accountsDbQueries) lookupResources(addr basics.Address, aidx basics.Cr
...
+				err = protocol.Decode(buf, &data.data)
+				if err != nil {
+					return err
+				}
+				if ctype == basics.AssetCreatable && !data.data.IsAsset() {
+					return fmt.Errorf("lookupResources asked for an asset but got %v", data.data)
+				}
+				if ctype == basics.AppCreatable && !data.data.IsApp() {
+					return fmt.Errorf("lookupResources asked for an app but got %v", data.data)
+				}
```
tsachiherman, line 1976: I think that this could be pushed upstream, but I'm ok with that here for now.

tsachiherman (approved):

PR #3531, opened 2022-01-28
ledger refactoring: measure and log performance during fast catchup

tolikzinovyev:
## Summary

Add logs in catchpoint service measuring the time it takes to
- write catchpoint balances to disk
- write catchpoint creatables to disk
- write catchpoint hashes to disk
- write all catchpoint data to disk
- download the catchpoint + write all catchpoint data to disk
- build the merkle trie

## Test Plan

I ran fast catchup and verified that logs are printed.

### DIFF
--- a/catchup/ledgerFetcher.go
+++ b/catchup/ledgerFetcher.go
@@ -139,10 +139,25 @@ func (lf *ledgerFetcher) getPeerLedger(ctx context.Context, peer network.HTTPPee
 	defer watchdogReader.Close()
 	tarReader := tar.NewReader(watchdogReader)
 	var downloadProgress ledger.CatchpointCatchupAccessorProgress
+	var writeDuration time.Duration
+
+	printLogsFunc := func() {
+		lf.log.Infof(
+			"writing balances to disk took %d seconds, "+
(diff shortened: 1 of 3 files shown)


catchup/ledgerFetcher.go
```diff
@@ -139,10 +139,25 @@ func (lf *ledgerFetcher) getPeerLedger(ctx context.Context, peer network.HTTPPee
 	defer watchdogReader.Close()
 	tarReader := tar.NewReader(watchdogReader)
 	var downloadProgress ledger.CatchpointCatchupAccessorProgress
+	var writeDuration time.Duration
+
+	printLogsFunc := func() {
+		lf.log.Infof(
```
tsachiherman, line 145: This is fine - but please write these stats only on success. You can determine that by checking the err variable.

tsachiherman (commented):
Looks good; only a small request

catchup/ledgerFetcher.go
tolikzinovyev: It is written only on success.

tsachiherman (approved):

PR #3533, opened 2022-01-28
ledger refactoring: get rid of LookupLatest from /v1/assets

cce:
## Summary

In #3330 the /v1/assets endpoint was modified to use LookupLatest just to fetch a single AssetParams object for a creator address, when it should just use LookupResource. This eliminates the unnecessary call to LookupLatest.

## Test Plan

Existing tests should pass.


algorandskiy (approved):

tsachiherman (approved):
Looks good

PR #3535, opened 2022-01-30
fix install bug

algoidan:


node/node.go
```diff
@@ -1002,30 +1002,14 @@ func (node *AlgorandFullNode) loadParticipationKeys() error {
...
-			return err
-		}
-
-		keys[round] = account.StateProofSigner(*key)
-	}
-	err = node.accountManager.Registry().AppendKeys(partID, keys)
-	// kinda useless since AppendKeys returns nil exclusively and appends asynchronously
-	if err != nil {
-		return err
+	keys := part.StateProofSecrets.GetAllKeys()
```
Aharonee, line 1006: i'm not sure it needs to be a different function anymore, since this can only be used here and only if part was created by `RestoreParticipationWithSecrets`

crypto/merklesignature/persistentMerkleSignatureScheme.go
```diff
@@ -88,46 +94,34 @@ func (s *Secrets) Persist(store db.Accessor) error {
...
+func (s *Secrets) RestoreAllSecrets(store db.Accessor) error {
+	var keys []crypto.GenericSigningKey
 
 	err := store.Atomic(func(ctx context.Context, tx *sql.Tx) error {
-		row := tx.QueryRow("SELECT key,round FROM StateProofKeys WHERE id = ?", id)
-		err := row.Scan(&keyB, &round)
+		rows, err := tx.Query("SELECT key FROM StateProofKeys")
 		if err != nil {
-			return fmt.Errorf("failed to select stateProof key for round %d : %w", round, err)
+			return fmt.Errorf("%s - %w", errSelectKeysError, err)
```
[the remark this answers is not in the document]
algoidan: it can check `string.Constains` - which is not perfect but better in my opnion. We've done in the siganturescehme

crypto/merklesignature/merkleSignatureScheme.go
```diff
@@ -78,6 +78,14 @@ type (
 	// Verifier is used to verify a merklesignature.Signature produced by merklesignature.Secrets.
 	// It validates a merklesignature.Signature by validating the commitment on the GenericVerifyingKey and validating the signature with that key
 	Verifier [MerkleSignatureSchemeRootSize]byte
+
+	//KeyRound represents an ephemeral signing key with it's corresponding round
+	KeyRound struct {
+		_struct struct{} `codec:",omitempty,omitemptyarray"`
+
+		Round               uint64                    `codec:"rnd"`
+		EphemeralSigningKey *crypto.GenericSigningKey `codec:"key"`
```
Aharonee, line 87: Could have just used `Key` instead, it seems a bit verbose considering what it's used for

PR #3542, opened 2022-01-31
ledger refactor: changes to REST API to support more resources

cce:
Add extra calls for querying assets and applications


winder (commented):
Seems like a fine start. But if we're providing per-account versions of these functions, we should probably provide per-account versions of the asset-param/app-global endpoints for consistency.

Likewise, if we have endpoints for per-account asset-holding and application-local, should we also have generic versions of those endpoints?

My preference is to start with the general-purpose option and add the more specific forms later if we still want them. If we're confident those won't be needed, then the proposal in this PR seems fine.

daemon/algod/api/algod.oas2.json
```diff
@@ -213,6 +229,160 @@
         }
       ]
     },
+    "/v2/accounts/{address}/assets/{asset-id}": {
```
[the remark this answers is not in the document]
cce: Here, the return type "ApplicationAssetResponse" provides both or either of the AssetHolding and AssetParams objects associated with this address — so it is general purpose in that sense.

daemon/algod/api/algod.oas2.json
```diff
@@ -164,6 +164,13 @@
             "in": "path",
             "required": true
           },
+          {
+            "name": "exclude-creatable-data",
```
cce, line 168: I don't like this name

cce: updated as per discussion yesterday

daemon/algod/api/server/v2/handlers.go
```diff
@@ -259,15 +260,24 @@ func (v2 *Handlers) AccountInformation(ctx echo.Context, address string, params
...
+	record, lastRound, amountWithoutPendingRewards, err := myLedger.LookupLatest(addr)
 	if err != nil {
 		return internalError(ctx, err, errFailedLookingUpLedger, v2.Log)
 	}
 
-	consensus, err := myLedger.ConsensusParams(lastRound)
-	if err != nil {
-		return internalError(ctx, err, fmt.Sprintf("could not retrieve consensus information for last round (%d)", lastRound), v2.Log)
+	// check against configured total limit on assets/apps
+	if maxResults := v2.Node.Config().MaxAccountsAPIResults; maxResults != 0 {
```
tsachiherman, line 275: the idea here is correct, but you'd want to do this without retrieving all the assets from the ledger.
i.e. this would need to be broken into 
- get base account data : then, figure out if the account has too many resources. if so, fail.
- otherwise, go get the resources and return them.

daemon/algod/api/server/v2/handlers.go
```diff
@@ -310,6 +319,191 @@ func (v2 *Handlers) AccountInformation(ctx echo.Context, address string, params
...
+	if record.AssetHolding != nil {
+		// AccountInformation REST API endpoint and generated.AssetHolding provides creator address with AssetHolding
+		creatorAddr, ok, err := ledger.GetCreator(basics.CreatableIndex(assetID), basics.AssetCreatable)
+		var creator string
+		if err == nil && ok {
+			creator = creatorAddr.String()
+		} else {
+			// Asset may have been deleted, so we can no
+			// longer fetch the creator
+			creator = ""
```
tsachiherman, line 437: if the asset was deleted, then all the holding of that asset would have been deleted as well.
I'd suggest you'll respond with InternalServerError in this case.

daemon/algod/api/server/v2/handlers.go
```diff
@@ -450,6 +649,7 @@ func (v2 *Handlers) GetStatus(ctx echo.Context) error {
 		CatchpointVerifiedAccounts:  &stat.CatchpointCatchupVerifiedAccounts,
 		CatchpointTotalBlocks:       &stat.CatchpointCatchupTotalBlocks,
 		CatchpointAcquiredBlocks:    &stat.CatchpointCatchupAcquiredBlocks,
+		MaxAccountsApiResults:       maxAcctResults,
```
tsachiherman, line 652: It doesn't feel right that we return that in status.. It's more of a configuration, since it'a always a static value.
Another point is that I think it'd better to "hide" it anyway : 
- a client that would be able to support using it, would also be able to process the response error code.
- a client that wouldn't be able to support it, wouldn't get much from this value being reported.

ledger/ledger.go
```diff
@@ -460,17 +460,36 @@ func (l *Ledger) ListApplications(maxAppIdx basics.AppIndex, maxResults uint64)
...
+}
+
+// LookupAccount uses the accounts tracker to return the account state (without
+// resources) for a given address, for a given round. The returned account values
+// reflect the changes of all blocks up to and including the returned round number.
+func (l *Ledger) LookupAccount(round basics.Round, addr basics.Address) (ledgercore.AccountData, basics.Round, basics.MicroAlgos, error) {
+	l.trackerMu.RLock()
+	defer l.trackerMu.RUnlock()
+
+	data, rnd, rewardsProto, rewardsLevel, err := l.accts.lookupWithoutRewards(round, addr, true /* take lock */)
```
tsachiherman, line 483: adding the rewardsProto, rewardsLevel to the return value of `lookupWithoutRewards` could negatively impact performance : the rewardsProto is a pretty large object.

two ideas:
1. pass back the version ( which is an immutable string ), and perform the config.Consensus lookup.
2. create a separate method in accts for LookupAccount.

daemon/algod/api/server/v2/handlers.go
cce: Yes, I was thinking of modifying LookupLatest to either take a limit parameter, or call LookupWithoutRewards first to check the limits, and then call LookupLatest again  if it's below the limit. I was thinking the second.

daemon/algod/api/server/v2/handlers.go
cce: I copied this behavior from the current implementation of the `/v2/accounts/X` endpoint handler in https://github.com/algorand/go-algorand/blob/master/daemon/algod/api/server/v2/handlers.go#L288-L303 where it loops through all AssetHoldings to fill out a "creator" string for the response. I would love to get rid of that field though..

I believe it is possible to be a holder of a deleted asset however. You will still need to submit a transaction to close it out.

ledger/ledger.go
cce: OK I will have it pass back the version, that's a lot smaller..

daemon/algod/api/server/v2/handlers.go
tsachiherman: given that this is a new entry point, we're not committed to support the same fields. I honestly can't tell if this is even something that would be of interest to a client.

I think that you can't be a holder to a deleted asset, but you could have opt-in for that asset.
( i.e. when deleting an asset, it checks that you hold all the instances of that asset )

daemon/algod/api/server/v2/handlers.go
tsachiherman: Yes, the second is good.

daemon/algod/api/server/v2/handlers.go
cce: Right, you will still show up as opted in to hold a 0 balance of a deleted asset. There are lots of accounts in this situation and I think we would want to still show this as a holding for them.

daemon/algod/api/server/v2/handlers.go
cce: updated

daemon/algod/api/server/v2/handlers.go
cce: Otherwise they won't know what to do so they can opt out and lower their min balance. For example if you bought 1000 NFTs, then got rid of them, and then the creator deleted the asset, you would not even be able to see why your min balance was so high now.

daemon/algod/api/server/v2/handlers.go
cce: Removed

daemon/algod/api/server/v2/handlers.go
tsachiherman: If the asset was destroyed, then they can't have non-zero asset amount.
But it's true that they would need to close-out to the creator address in order to opt-out and reduce their min balance.

cce: I removed the "creator" string from the AssetHolding response type, which I think resolves this

tsachiherman (approved):

Successahead (commented):
Good

PR #3544, opened 2022-02-01
AVM: SHA3-256 hash function opcode

Matt-Ryley:
- Implemented the SHA3-256 hash function required for interoperability of Algorand with the ICON blockchain.
- Added the OPCode 0x96 for this functionality.
- Updated documentation and unit tests to include the new SHA3-256 hashing function.


data/transactions/logic/opcodes.go
```diff
@@ -294,6 +294,9 @@ var OpSpecs = []OpSpec{
 	{0x94, "exp", opExp, asmDefault, disDefault, twoInts, oneInt, 4, modeAny, opDefault},
 	{0x95, "expw", opExpw, asmDefault, disDefault, twoInts, twoInts, 4, modeAny, costly(10)},
 
+	//More Hash Functions
+	{0x96, "sha3_256", opSHA3_256, asmDefault, disDefault, oneBytes, oneBytes, 6, modeAny, costly(130)},
```
jannotti, line 298: How did you derive the cost?

Matt-Ryley: I wasn't sure how to derive the cost. The cost of the SHA3-256 should be roughly the same as the Keccak256 as they are very similar algorithms, except the padding in the SHA3-256 adds up to 4 additional bits. Im open to any suggestions of how to derive this cost more accurately?

jannotti: There is some benchmarking code in eval_test.go.  But hold off for a couple days. We're in the middle of a merge that will change it slightly.  But once that merge happens, you'll find
`func BenchmarkBase64Decode(b *testing.B)` that was used to compare base64 decode to the costs of various hashes.  By adding sha3_256 in there, we should be able to get a feel for relative costs.

(The function may already be in master, if you want to look now.)

Matt-Ryley: Thanks for the information! Which merge should I be looking out for?

jannotti: feature/contract-to-contract going into master

Matt-Ryley: Thanks for the heads up @algoanne! Ill get on this as soon as possible

Matt-Ryley: To me they are in the same ball park as one another and thus should have the same cost, as the eqv for other hashes like
BenchmarkBase64Decode/sha256_small-8             	 1260788	      1003 ns/op
BenchmarkBase64Decode/sha512_256_small-8         	 1695235	       715.2 ns/op

Are very different

jannotti: Yes, it sounds like it should have the same cost as keccak.  I think we can take this as soon as CI passes.  I'll be making a pass over it once that happens as I have changed the opcode interface a bit.  But I think your work is done! Thanks for the contribution.

jannotti:
Sorry, it looks like one thing you'll have to do to make CI happy is run `make` in the `data/transactions/logic` directory.  That will regenerate a couple spec files from your docstrings.  That's the `codegen_verification` step that's failing.

Matt-Ryley:
@jannotti after upstream merge the arm64 test is now failing in one of the tests for the ledger. As I haven't modified any code in that directory I am confused to why this test fails. Do you know of any reason? My local machine is amd64 architecture and those tests run successfully in the CI pipeline and also on my local machine, so I can't really debug to find the cause of the error.

jannotti:
I'm not sure, but I'll be looking at it more closely today.

I used a new PR, so I would be able to tweak it (to handle merge conflicts).
https://github.com/algorand/go-algorand/pull/3762

I believe I merged your commits in properly so you will retain credit.
Thanks again.

jannotti:
Closed in favor of #3762

PR #3547, opened 2022-02-01
Support logs and Inner transactions in tealdbg

algochoi:
## Summary

This PR adds logs as an array and one layer of inner txns as an object in the local scope of `tealdbg`. The logs are represented as a `[]string` and inner txns as `[]transactions.SignedTxnWithAD` in `AppState`. Currently this PR is based on the `feature/contract-to-contract` branch to support nested inner txns.

### Short explanation
* User wants to look at the logs/itxns, `makeLocalScope()` is called and the logs/itxns array is created with their respective object IDs. The object IDs map to `makeLogsState()` and `makeInnerTxnsState()`. 
* When logs and itxns are encoded, the group ID is appended at the end of the object ID, e.g. `innerTxnObjID_id_01_02_03` denotes `s.innerTxn[1][2][3]` by supplying an array of group IDs. For a single inner transaction, it uses the `innerTxnObjID_id` prefix and for an array of nested inner transactions, it is encoded as `innerTxnObjID_nested`. 
* When the ID is decoded, traverse to the inner transaction. If the ID is a single inner transaction, then call `makeInnerTxnImpl()`, build the transaction object and encode the log obj and array of nested inner transactions ID. If the ID is an array of inner transactions, then create a slice of inner transaction objects and encode each inner transaction object by appending the group ID. 

## Test Plan

Visually checked output and updated test case. There is a script for generating a test transaction and debugging context [here](https://github.com/algochoi/algorand-onboard/tree/master/tealdbg-tutorial).
Screen Shot 2022-02-03 at 10 44 58 PM

For any interested observers, there is an example test case with nested itxns you can run on the browser in [this repo](https://github.com/algochoi/algorand-onboard/tree/nested-itxns/tealdbg-tutorial). You can do something like:
* checkout this branch then `make install`
* run `tealdbg` from your GOPATH bin with the example dryrun `msgp`, e.g. `./tealdbg debug path-to-test-repo/itxn_call2.teal -d path-to-test-repo/generated-data/dryrun_txn.msgp --group-index 5`


jasonpaulos (commented):
I tested locally and this looks pretty good! I just have some questions/comments.

1. This PR does not implement "step into" for c2c calls. Is there a separate issue for that?
2. Inner transactions have all of the "normal" transaction fields displayed, but they don't have any "effect" fields displayed besides logs. Ideally you would be able to access `CreatedAssetID` and `CreatedApplicationID` as well.
3. I noticed that the inner transactions and logs arrays in transactions are hidden if they are 0. Personally I'd prefer them shown as 0-length arrays similar to other txn array fields, but this is a minor thing.
4. Also, not sure if this is appropriate to address here, but the global field `OpcodeBudget` does not update after every op like I would expect. I think it's because the debugger assumes global fields never change, which up until now was true.

algochoi:
> I tested locally and this looks pretty good! I just have some questions/comments.
> 1. This PR does not implement "step into" for c2c calls. Is there a separate issue for that?

I think this issue addresses the step into: https://github.com/algorand/go-algorand-internal/issues/1789

> 2. Inner transactions have all of the "normal" transaction fields displayed, but they don't have any "effect" fields displayed besides logs. Ideally you would be able to access `CreatedAssetID` and `CreatedApplicationID` as well.

Good point - I can try to address this in this PR 👍 

> 3. I noticed that the inner transactions and logs arrays in transactions are hidden if they are 0. Personally I'd prefer them shown as 0-length arrays similar to other txn array fields, but this is a minor thing.

I'll play around with this and see what works better - I initially moved it out because the sidebar looked a bit congested, but I think that's a fair point.

> 4. Also, not sure if this is appropriate to address here, but the global field `OpcodeBudget` does not update after every op like I would expect. I think it's because the debugger assumes global fields never change, which up until now was true.

True, I'll play around with this as well - I will address this in this PR or create an issue to address this.

cmd/tealdbg/cdtState.go
```diff
@@ -755,9 +893,27 @@ func makeTxnImpl(txn *transactions.Transaction, groupIndex int, preview bool) (d
 	return
 }
 
+func makeInnerTxnImpl(txn *transactions.SignedTxnWithAD, groupIndexes []int, preview bool) (desc []cdt.RuntimePropertyDescriptor) {
+	groupIndex := groupIndexes[len(groupIndexes)-1]
+	desc = makeTxnImpl(&txn.Txn, groupIndex, true, preview)
+
+	logs := makeArray("logs", len(txn.EvalDelta.Logs), encodeLogsID(groupIndexes))
+	innerTxns := makeArray("innerTxns", len(txn.EvalDelta.InnerTxns), encodeNestedInnerTxnID(groupIndexes))
+	createdApplicationID := fieldDesc{Name: "CreatedApplicationID", Value: strconv.Itoa(int(txn.ApplicationID)), Type: "number"}
```
jasonpaulos, line 902: nit: it's unlikley to matter for these fields, but most other fields use the `bigint` type instead of `number` in case they exceed JavaScript's `Number.MAX_SAFE_INTEGER` (2^53 - 1).

In this same vein, you could do `strconv.FormatUint(uint64(txn.ApplicationID), 10)` instead of `strconv.Itoa(int(txn.ApplicationID))` to allow 32+ bits for these fields.

jasonpaulos (commented):
The budget changes look great!

I have one last suggestion that should not matter practically, but it's probably a good idea

cmd/tealdbg/cdtState.go
algochoi: Good point, will do

cmd/tealdbg/cdtState.go
```diff
@@ -704,21 +839,25 @@ func makeGlobals(s *cdtState, preview bool) (desc []cdt.RuntimePropertyDescripto
 	fields := prepareGlobals(s.globals)
 	desc = make([]cdt.RuntimePropertyDescriptor, len(fields))
 	for i, field := range fields {
+		// We need to dynamically override this field with the proper value each step.
+		if field.Name == "OpcodeBudget" {
```
algorandskiy, line 843: I think `state.Update()` should be updating globals value. This is only "renderer" and should not contains any mapping/fields updating logic.

cmd/tealdbg/local.go
```diff
@@ -158,10 +158,24 @@ type evalResult struct {
...
-	schemas basics.StateSchemas
-	global  map[basics.AppIndex]basics.TealKeyValue
-	locals  map[basics.Address]map[basics.AppIndex]basics.TealKeyValue
+	appIdx       basics.AppIndex
+	schemas      basics.StateSchemas
+	global       map[basics.AppIndex]basics.TealKeyValue
+	locals       map[basics.Address]map[basics.AppIndex]basics.TealKeyValue
+	logs         []string
+	innerTxns    []transactions.SignedTxnWithAD
+	opcodeBudget int
```
algorandskiy, line 167: this should be in part of `cdtStateUpdate` and not `AppState`
Use it directly at
```
				appState := s.debugger.GetStates(&dbgState)
				state.Update(cdtStateUpdate{
					dbgState.Stack, dbgState.Scratch,
					dbgState.PC, dbgState.Line, dbgState.Error,
					dbgState. opcodeBudget,  <----
					appState,
				})
```

jasonpaulos (approved):

cmd/tealdbg/cdtState.go
algochoi: I changed the `globals[opcodeBudgetFieldIdx]` instead :+1:

cmd/tealdbg/cdtState.go
```diff
@@ -74,6 +76,8 @@ const (
 	addressHint
 )
 
+const opcodeBudgetFieldIdx = 12
```
algorandskiy, line 79:
```suggestion
const opcodeBudgetFieldIdx = logic.OpcodeBudget
```

algorandskiy (commented):
One last remark

cmd/tealdbg/cdtState.go
algochoi: Good point, replaced with named constant

algorandskiy (dismissed):

cmd/tealdbg/cdtSession.go
```diff
@@ -169,7 +169,7 @@ func (s *cdtSession) websocketHandler(w http.ResponseWriter, r *http.Request) {
 		state.Update(cdtStateUpdate{
 			dbgState.Stack, dbgState.Scratch,
 			0, 0, "",
-			s.debugger.GetStates(nil),
+			0, s.debugger.GetStates(nil),
```
jannotti, line 172: Is the starting opcode budget not available yet for some reason?

algochoi: Good catch, we can just set the starting budget here... I'll push another commit

jannotti (approved):
Approved (Pavel also had before the last commit removed it.)

PR #3549, opened 2022-02-02
ledger refactoring: avoid passing ctype into account updates LookupResources

tsachiherman:
## Summary

This PR removes the `ctype` parameter from the account updates LookupResources.
Instead of doing this validation at the accounts update level, the validation was moved to the Ledger.LookupResources.

## Test Plan

Unit tests updated.


ledger/accountdb.go
```diff
@@ -1976,12 +1976,6 @@ func (qs *accountsDbQueries) lookupResources(addr basics.Address, aidx basics.Cr
 				if err != nil {
 					return err
 				}
-				if ctype == basics.AssetCreatable && !data.data.IsAsset() {
```
algorandskiy, line 1979: what is the benefit of moving it two layers higher?

tsachiherman: The lower layer doesn't really do anything useful with the given parameter, so I'd rather it to be opaque. The exact same type of validation can be safely performed at a higher level without loosing any of the benefits.
Note that we cannot take it higher than that, since the indexer does need to have this separation.

cce: My concern is that this locks us into the current schema, if we ever want to have a different schema or DB supporting a different way of looking up resources later. For example, if ctype were part of the lookup key. Also, much less likely, if there are ever new "creatables" that have exclusive ID space, we would be similarly locked in.

tsachiherman: I don't think that this one necessarily locks is in any way. If we would need to, we can always undo these changes. At this point in time, I'm unaware of any concrete new resource type that we would need to add - and as such, I'm trying to align the code with the concrete expected behavior. Extending it to have additional resource type might or might not happen in the future, but without any concrete plans around that, I'm quite hesitant if it would be wise to retain unnecessary infrastructure.

cce: IMO it just seems like it would be nicer to hide the specifics of the DB storage schema from the interface presented for lookup. It is kind of an under-the-hood optimization that different resource types are stored together for the SQL query, but if we had a different backend that wasn't the accountDBQueries object we might want the ctype.

tsachiherman: I'm ok with closing this PR for now - but I think that we should have more lean interface on the account updates. Some of that would be enforced by the upcoming 320 round changes, but it would be eventually up to us to make sure we don't perform any redundant operations.

PR #3553, opened 2022-02-02
V31 consensus upgrade

algoidan:
## Summary

Create an update path that would enable the following features:
1. Batch Verification (https://github.com/algorand/go-algorand/pull/3031)
2. State Proof Keys (https://github.com/algorand/go-algorand/pull/2990)
3. TEAL 6 (https://github.com/algorand/go-algorand/pull/3397)
4. Expired Participation Keys (https://github.com/algorand/go-algorand/pull/2924)
5. Fix rewards calculation bug (https://github.com/algorand/go-algorand/pull/3403)

## Test Plan

Unit tests added.


tsachiherman (changes requested):
Four more requests here:
1. Don't forget to update ConsensusCurrentVersion to ConsensusV30
2. Add v30.ApprovedUpgrades[protocol.ConsensusV31] = 140000
3. Add v31.RewardsCalculationFix to true ( and remove from vFuture )
4. Add v31.MaxProposedExpiredOnlineAccounts to 32 ( and remove from vFuture )

config/consensus.go
```diff
@@ -1040,9 +1040,15 @@ func initConsensusProtocols() {
 	// v29 can be upgraded to v30, with an update delay of 7 days ( see calculation above )
 	v29.ApprovedUpgrades[protocol.ConsensusV30] = 140000
 
+	v31 := v30
+
```
algorandskiy, line 1044:
```suggestion
	v31 := v30
	v31.ApprovedUpgrades = map[protocol.ConsensusVersion]uint64{}
```

ledger/apply/keyreg_test.go
```diff
@@ -188,21 +188,21 @@ func TestStateProofPKKeyReg(t *testing.T) {
...
 	tx := createTestTxn(t, src, secretParticipation, vrfSecrets)
-	mockBal := makeMockBalances(protocol.ConsensusCurrentVersion)
+	mockBal := makeMockBalances(protocol.ConsensusV30)
 	err := Keyreg(tx.KeyregTxnFields, tx.Header, mockBal, transactions.SpecialAddresses{FeeSink: feeSink}, nil, basics.Round(0))
 	require.NoError(t, err)
 
 	acct, err := mockBal.Get(tx.Src(), false)
 	require.NoError(t, err)
-	require.Equal(t, acct.StateProofID.IsEmpty(), true)
+	require.Equal(t, true, acct.StateProofID.IsEmpty())
```
jannotti, line 197: There is `require.True()` if you'd like.

ledger/apply/keyreg_test.go
```diff
@@ -188,21 +188,21 @@ func TestStateProofPKKeyReg(t *testing.T) {
...
 
-	mockBal = makeMockBalances(protocol.ConsensusFuture)
+	mockBal = makeMockBalances(protocol.ConsensusCurrentVersion)
 	err = Keyreg(tx.KeyregTxnFields, tx.Header, mockBal, transactions.SpecialAddresses{FeeSink: feeSink}, nil, basics.Round(0))
 	require.NoError(t, err)
 
 	acct, err = mockBal.Get(tx.Src(), false)
 	require.NoError(t, err)
-	require.Equal(t, acct.StateProofID.IsEmpty(), false)
+	require.Equal(t, false, acct.StateProofID.IsEmpty())
```
jannotti, line 205: And `require.False()`

test/e2e-go/features/participation/participationExpiration_test.go
```diff
@@ -76,7 +76,7 @@ func testExpirationAccounts(t *testing.T, fixture *fixtures.RestClientFixture, f
 		_, currentRound := fixture.GetBalanceAndRound(richAccount)
 		// account adds part key
 		partKeyFirstValid := uint64(0)
-		partKeyValidityPeriod := uint64(150)
+		partKeyValidityPeriod := uint64(150) // TODO: maybe increase?
```
jannotti, line 79: What are the units?  I would have guessed rounds, but that would be very short.

jannotti (approved):
Looks ok from AVM 1.1 perspective.  My other comments are very minor.

config/consensus.go
```diff
@@ -1057,9 +1057,31 @@ func initConsensusProtocols() {
...
+	// Enable TEAL 6 / AVM 1.1
+	v31.LogicSigVersion = 6
+	v31.EnableInnerTransactionPooling = true
+	v31.IsolateClearState = true
+
+	// stat proof key registration
+	v31.EnableStateProofKeyregCheck = true
+
+	// Maximum validity period for key registration, to prevent generating too many StateProof keys
+	v31.MaxKeyregValidPeriod = 256*(1<<16) - 1
```
gmalouf, line 1075: I was about to ask some questions about the numbers going into this formula, seems like intent is to address this in a follow on issue: #3257 cc @algonautshant

test/e2e-go/features/participation/participationExpiration_test.go
Aharonee: Yes they are rounds and it is very short, maybe we should put 1000-3000 there instead, as no State Proof keys will be generated for such a short interval.

Aharonee (changes requested):
I've added some minor fixes here: https://github.com/algoidan/go-algorand/pull/15

test/e2e-go/restAPI/restClient_test.go
```diff
@@ -590,6 +594,7 @@ func TestAccountParticipationInfo(t *testing.T) {
 	a.Equal(uint64(firstRound), account.Participation.VoteFirst, "API must print correct first participation round")
 	a.Equal(uint64(lastRound), account.Participation.VoteLast, "API must print correct last participation round")
 	a.Equal(dilution, account.Participation.VoteKeyDilution, "API must print correct key dilution")
+	// TODO: should we update the v1 API to support state proof? Currently it does not return this field.
```
gmalouf, line 597: @tsachiherman did we already address this?

gmalouf (commented):
Non-blocking FYI, didn't update a few references to ConsensusFuture -  quick find in IDE will turn some up.

`participationRegistry_test.go`
`accountdb_test.go`

I'm opening an issue for the general case - we should have a single approach to referencing the ConsensusParams protocol inclusive of current+future that does not have to change when we bump the version.

Aharonee:
> Non-blocking FYI, didn't update a few references to ConsensusFuture - quick find in IDE will turn some up.
> 
> `participationRegistry_test.go` `accountdb_test.go`
> 
> I'm opening an issue for the general case - we should have a single approach to referencing the ConsensusParams protocol inclusive of current+future that does not have to change when we bump the version.

@gmalouf If you're referring to data/account/participationRegistry_test.go:45 then it should remain `ConsensusFuture` still, since it references a consensus parameter which will only be enabled when we enable the Compact Certificates.

In ledger/accountdb_test.go:293 the `ConsensusFuture` can change in my opinion and the test won't break. I'll push a fix into https://github.com/algoidan/go-algorand/pull/15.

tsachiherman:
> > Non-blocking FYI, didn't update a few references to ConsensusFuture - quick find in IDE will turn some up.
> > `participationRegistry_test.go` `accountdb_test.go`
> > I'm opening an issue for the general case - we should have a single approach to referencing the ConsensusParams protocol inclusive of current+future that does not have to change when we bump the version.
> 
> @gmalouf If you're referring to data/account/participationRegistry_test.go:45 then it should remain `ConsensusFuture` still, since it references a consensus parameter which will only be enabled when we enable the Compact Certificates.
> 
> In ledger/accountdb_test.go:293 the `ConsensusFuture` can change in my opinion and the test won't break. I'll push a fix into [algoidan#15](https://github.com/algoidan/go-algorand/pull/15).

@Aharonee I agree with @gmalouf here; the proper way to test this before the protocol is created is to take an existing protocol version, and "add" the needed flags to its parameters in order to test using a custom protocol. I know that this might mean slightly more work, but it's the "cleaner" way to accomplish that. It also means that there won't be any changes when we apply new protocol version.

Aharonee (approved):

config/consensus.go
algonautshant: The assumption is that the source of the truth is here: `v31.MaxKeyregValidPeriod = 256*(1<<16) - 1`
Anywhere else should be referencing this parameter. 

The issue in #3257 is to make sure this variable is used for testing (instead of hardcoding the same value elsewhere), and fix the package dependency issue.

This is not a blocking issue.

algonautshant: `MaxKeyregValidPeriod` combines 256 and 16, which are based on two different limits. 
for example: 
merklearray/merkle.go:	MaxEncodedTreeDepth = 16

gmalouf:
Right, I was suggesting a reference to constants articulating what those
limits source from would be ideal.

tsachiherman (approved):

PR #3554, opened 2022-02-02
bug fix : tx sync was incorrectly validating transactions

tsachiherman:
## Summary

The TestSyncFromClient was randomly failing. This is a real issue in production code, and the fix it pretty trivial.

## Test Plan

Use existing tests to verify test.

### DIFF
--- a/rpcs/txSyncer.go
+++ b/rpcs/txSyncer.go
@@ -127,6 +127,7 @@ func (syncer *TxSyncer) syncFromClient(client TxSyncClient) error {
 		return fmt.Errorf("TxSyncer.Sync: peer '%v' error '%v'", client.Address(), err)
 	}
 
+	var pendingTxidMap map[transactions.Txid]struct{}
 	// test to see if all the transaction that we've received honor the bloom filter constraints
 	// that we've requested.
 	for _, txgroup := range txgroups {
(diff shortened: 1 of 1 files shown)


rpcs/txSyncer.go
```diff
@@ -127,18 +127,31 @@ func (syncer *TxSyncer) syncFromClient(client TxSyncClient) error {
...
+				// need to perform explicit check. This is not too bad since we're doing this check only on the fail
+				// cases.
+				if pendingTxidMap == nil {
+					// construct and initialize it.
+					pendingTxidMap = make(map[transactions.Txid]struct{}, len(pending))
+					for _, txid := range pending {
+						pendingTxidMap[txid] = struct{}{}
+					}
+				}
+				if _, has := pendingTxidMap[txID]; !has {
```
algonautshant, line 148: Shouldn't this be if `has` instead of `!has`?

algonautshant (dismissed):
Looks great. 
Just one issue.

rpcs/txSyncer.go
tsachiherman: yes. I believe that you're correct.

algonautshant (approved):

PR #3556, opened 2022-02-03
Inner clearstate

jannotti:
Prevents inner txns and bad opcode budget use in clear state programs.

### DIFF
--- a/config/consensus.go
+++ b/config/consensus.go
@@ -290,6 +290,9 @@ type ConsensusParams struct {
 	// should the number of inner transactions be pooled across group?
 	EnableInnerTransactionPooling bool
 
+	// provide greater isolation for clear state programs
+	IsolateClearState bool
+
 	// maximum number of applications a single account can create and store
 	// AppParams for at once
 	MaxAppsCreated int
@@ -1060,6 +1063,7 @@ func initConsensusProtocols() {
 	// Enable TEAL 6 / AVM 1.1
 	vFuture.LogicSigVersion = 6
 	vFuture.EnableInnerTransactionPooling = true
+	vFuture.IsolateClearState = true
 
 	vFuture.MaxProposedExpiredOnlineAccounts = 32
 
--- a/data/transactions/logic/eval.go
+++ b/data/transactions/logic/eval.go
@@ -561,7 +561,11 @@ func EvalContract(program []byte, gi int, aid basics.AppIndex, params *EvalParam
 	}
 	pass, err := eval(program, &cx)
 
-	// update side effects
+	// update side effects. It is tempting, and maybe even a good idea, to store
+	// the pointer to cx.scratch instead.  Since we don't modify them again,
+	// it's probably safe. However it may have poor GC characteristics (because
+	// we'd be storing a pointer into a much larger structre, the cx), and
+	// copying seems nice and clean.
 	cx.pastScratch[cx.GroupIndex] = &scratchSpace{}
 	*cx.pastScratch[cx.GroupIndex] = cx.scratch
 
--- a/data/transactions/logic/eval.go
+++ b/data/transactions/logic/eval.go
@@ -339,10 +339,8 @@ func NewEvalParams(txgroup []transactions.SignedTxnWithAD, proto *config.Consens
 }
 
 // NewInnerEvalParams creates an EvalParams to be used while evaluating an inner group txgroup
-func NewInnerEvalParams(txg []transactions.SignedTxn, caller *EvalContext) *EvalParams {
-	txgroup := transactions.WrapSignedTxnsWithAD(txg)
-
-	minTealVersion := ComputeMinTealVersion(txgroup, true)
+func NewInnerEvalParams(txg []transactions.SignedTxnWithAD, caller *EvalContext) *EvalParams {
+	minTealVersion := ComputeMinTealVersion(txg, true)
 	// Can't happen currently, since innerAppsEnabledVersion > than any minimum
 	// imposed otherwise.  But is correct to check, in case of future restriction.
 	if minTealVersion < *caller.MinTealVersion {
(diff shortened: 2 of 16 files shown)


data/transactions/logic/eval.go
```diff
@@ -559,9 +572,20 @@ func EvalContract(program []byte, gi int, aid basics.AppIndex, params *EvalParam
...
+		}
+	}
+
 	pass, err := eval(program, &cx)
 
-	// update side effects
+	// update side effects. It is tempting, and maybe even a good idea, to store
+	// the pointer to cx.scratch instead.  Since we don't modify them again,
+	// it's probably safe. However it may have poor GC characteristics (because
+	// we'd be storing a pointer into a much larger structre, the cx), and
```
michaeldiamant, line 587: @jannotti Minor - Looks like a typo.

```suggestion
	// we'd be storing a pointer into a much larger structure, the cx), and
```

data/transactions/logic/eval.go
```diff
@@ -4043,6 +4074,10 @@ func opTxBegin(cx *EvalContext) {
 		cx.err = errors.New("itxn_begin without itxn_submit")
 		return
 	}
+	if cx.Proto.IsolateClearState && cx.Txn.Txn.OnCompletion == transactions.ClearStateOC {
+		cx.err = errors.New("clear state prorgams can not issue inner transactions")
```
michaeldiamant, line 4078: @jannotti Minor - Looks like a typo in the error message.

```suggestion
		cx.err = errors.New("clear state programs can not issue inner transactions")
```

data/transactions/logic/eval.go
```diff
@@ -547,6 +547,19 @@ func (pe PanicError) Error() string {
...
+// reject the transaction.  Normally, an error in evaluation would not cause a
+// ClearState txn to fail. However, callers fail a txn for ClearStateBudgetError
+// because the transaction has not provided enough budget to let ClearState do
+// its job.
+type ClearStateBudgetError struct {
+	offered int
+}
+
+func (e ClearStateBudgetError) Error() string {
+	return fmt.Sprintf("ClearState execution with only %d", e.offered)
```
tzaffi, line 560: nit: can the errar message refer to `PooledApplicationBudget` or something like that?

ledger/internal/apptxn_test.go
```diff
@@ -2439,3 +2611,195 @@ func BenchmarkMaximumCallStackDepth(b *testing.B) {
...
+ itxn_field ApplicationID
+itxn_submit
+`),
+		ForeignApps: []basics.AppIndex{innerId},
+	}
+
+	eval = nextBlock(t, l, true, nil)
+	txn(t, l, eval, &outer)
+	vb = endBlock(t, l, eval)
+	outerId := vb.Block().Payset[0].ApplicationID
```
jasonpaulos, line 2658: What's the purpose of the outer app in this test? It doesn't seem used in a relevant way

jasonpaulos (commented):
Looks correct, I just have questions/comments on tests.

Additionally, I don't think there's any tests for the changes in clear state budget? One suggestion would be a test where an app calls an inner clear state that has an infinite loop and we verify the `global OpcodeBudget` only decreases by ~700.

jannotti:
> Looks correct, I just have questions/comments on tests.
> 
> Additionally, I don't think there's any tests for the changes in clear state budget? One suggestion would be a test where an app calls an inner clear state that has an infinite loop and we verify the `global OpcodeBudget` only decreases by ~700. Another test I'd like to see is one that makes sure `ClearStateBudgetError` actually fails the txn.

Correct, I still need both tests. Going into csp with less than 700 should NOT clear, but should fail. A csp that tries to take 701 should fail, but not kill caller, and should clear state.

ledger/internal/apptxn_test.go
jannotti: The intent is just to use outer as an app that opts its account into inner, then clears state with an inner, to make sure inner clears work. I'm on phone now, so I'll confirm that's actually what's going on when I get back to computer.

jannotti: I got it wrong, I was doing the doing test of a normal account opting into inner.

jannotti: No it tests inner clear state.  I'll get the budget stuff a bit later.

data/transactions/logic/eval.go
jannotti: Changed:
`return fmt.Sprintf("Attempted ClearState execution with low OpcodeBudget %d", e.offered)`
That should call to mind the `global OpcodeBudget` which is the problem.

config/consensus.go
```diff
@@ -290,6 +290,9 @@ type ConsensusParams struct {
 	// should the number of inner transactions be pooled across group?
 	EnableInnerTransactionPooling bool
 
+	// provide greater isolation for clear state programs
+	IsolateClearState bool
```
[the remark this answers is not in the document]
jannotti: IsolateClearState is actually go to change the semantics of *previous* TEAL versions too - just barely.  We've concluded this is a good change, as the previous semantics allow a malicious clear state to consume some extra resources. Still, it's not related to AVM version, but rather consensus, so it will go into effect if/when this consensus version is accepted by the network, even for old code (but not for old blocks, of course).

data/transactions/logic/eval.go
```diff
@@ -799,6 +825,16 @@ func (cx *EvalContext) remainingBudget() int {
 	if cx.runModeFlags == runModeSignature {
 		return int(cx.Proto.LogicSigMaxCost) - cx.cost
 	}
+
+	// restrict clear state programs from using more than standard unpooled budget
+	// cx.Txn is not set during check()
```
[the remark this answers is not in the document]
jannotti: Unit tests!  It's nil during `check()`, never during `eval()`. It would be annoying to fill it, since the public interface `CheckProgram()` does not take the transaction index.  It would probably be possible to add it, but I didn't want to change calling code.

config/consensus.go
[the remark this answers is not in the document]
michaeldiamant: @jannotti Got it - makes sense.  Thanks for the explanation.  Resolving.

jasonpaulos (commented):
Changes look good to me

data/transactions/logic/eval.go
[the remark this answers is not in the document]
michaeldiamant: @jannotti Thanks for the reply, I think I'm following better now.  Resolving.

PR #3558, opened 2022-02-03
Removed hard-coded constant in `sortition.go`.

Olshansk:
## Summary

Remove a hardcoded constant in sortition.go which was
used as the denominator in determining the selection ratio.

This clarifies what the maximum possible output size is of
the output VRF based on the SHA algorithm used to generate it.

## Purpose

This is simply a code quality change to make it easier to read and
understand without any focus on new functionality or performance.

It is unlikely that the VRF SHA length used by Algorand will changed,
so the generalization to a dynamic length based on the length of the
Crypto digest is only for the reader.

The use of an `init` function to initialize the `maxFloat` once rather than
anytime `sortition` runs is a minor optimization, but likely to be neglibile
in the grand scheme of things.

## Test Plan

Due to the usage of C src code, a simple `go test -v ./data/committee/sortition/`
cannot be executed without the proper configurations. However, circle CI
should be able to execute the unit tests automatically and verify the changes.

Recommendation: Running the recommended make targets to test the changes
did not automatically due to missing dependencies, so potentially a Dockerfile
could help resolve this for future contributors.

### DIFF
--- a/data/committee/sortition/sortition.go
+++ b/data/committee/sortition/sortition.go
@@ -36,7 +36,7 @@ func Select(money uint64, totalMoney uint64, expectedSize float64, vrfOutput cry
 	binomialP := expectedSize / float64(totalMoney)
 
 	precision := uint(8 * (len(vrfOutput) + 1))
-	maxFloatString := fmt.Sprintf("0x%s", strings.Repeat("f", crypto.DigestSize*2+1))
+	maxFloatString := fmt.Sprintf("0x%s", strings.Repeat("f", crypto.DigestSize*2))
 	max, b, err := big.ParseFloat(maxFloatString, 0, precision, big.ToNearestEven)
 	if b != 16 || err != nil {
 		panic("failed to parse big float constant in sortition")
--- a/data/committee/sortition/sortition.go
+++ b/data/committee/sortition/sortition.go
@@ -30,13 +30,14 @@ import (
 	"github.com/algorand/go-algorand/crypto"
 )
 
+var maxFloatString string = fmt.Sprintf("0x%s", strings.Repeat("f", crypto.DigestSize*2))
+
 // Select runs the sortition function and returns the number of time the key was selected
 func Select(money uint64, totalMoney uint64, expectedSize float64, vrfOutput crypto.Digest) uint64 {
 	binomialN := float64(money)
 	binomialP := expectedSize / float64(totalMoney)
 
 	precision := uint(8 * (len(vrfOutput) + 1))
-	maxFloatString := fmt.Sprintf("0x%s", strings.Repeat("f", crypto.DigestSize*2))
 	max, b, err := big.ParseFloat(maxFloatString, 0, precision, big.ToNearestEven)
 	if b != 16 || err != nil {
 		panic("failed to parse big float constant in sortition")
--- a/data/committee/sortition/sortition.go
+++ b/data/committee/sortition/sortition.go
@@ -22,6 +22,7 @@ package sortition
 // #include <stdlib.h>
 // #include "sortition.h"
 import "C"
+
 import (
 	"fmt"
 	"math/big"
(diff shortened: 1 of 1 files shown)


data/committee/sortition/sortition.go
```diff
@@ -33,15 +35,16 @@ func Select(money uint64, totalMoney uint64, expectedSize float64, vrfOutput cry
 	binomialN := float64(money)
 	binomialP := expectedSize / float64(totalMoney)
 
-	t := &big.Int{}
-	t.SetBytes(vrfOutput[:])
-
 	precision := uint(8 * (len(vrfOutput) + 1))
-	max, b, err := big.ParseFloat("0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff", 0, precision, big.ToNearestEven)
+	maxFloatString := fmt.Sprintf("0x%s", strings.Repeat("f", crypto.DigestSize*2+1))
```
jannotti, line 39: This can't be right, can it?  Surely an odd number of hex digits is not what you want.

Olshansk: You are correct. This was something I added while manually testing. Thanks!

jannotti: I'd prefer to see the `fmt.Sprintf` done once, rather than each time, as well.

Olshansk: Added a local var since you can't assign the output of formatting to a constant.

tsachiherman:
@Olshansk - can I suggest a small improvement for your PR ?
I've implemented these proposals in https://github.com/algorand/go-algorand/pull/3559.

tsachiherman:
@Olshansk - you'll need to sign the [CLAassistant](https://github.com/CLAassistant) before the proposed code changes could be merged in.

data/committee/sortition/sortition.go
jannotti: For that matter, perhaps you could move `max` out of the function as well, so it's not reparsed each time. I suspect the PR would be more palatable if it had a performance improvement, rather than only making the constant explicit. (On the explicitness front, why not `strings.Repeat("ff", crypto.DIgestSize)`?)

Olshansk: Done!

jannotti (approved):
This looks good to me.

tsachiherman (approved):
Thanks for the changes.

Olshansk:
@tsachiherman @jannotti I was wondering if either of you could help me understand the logic behind `sortition_binomial_cdf_walk`.

I've looked at the paper and read through some online articles ([example](https://community.algorand.org/blog/the-intuition-behind-algorand-cryptographic-sortition)) but am still struggling to understand this:

In short, I understand that we use VRF to compute a proof and a hash, which is normalized to a value in [0, 1) using the hashlen (i.e. the `ratio`). The binomial distribution is built using the total voting power, and the number of expected leaders. We then compute the CDF for each value `j` less than the specific node's voting power and compare it boundary of the CDF.

```
    if (ratio <= boundary) {
      return j;
    }
```

However, I'm still struggling to understand the significance of `j` in the check above. Are we saying that a single node has the opportunity to vote with at most `money` tokens (i.e. their voting power), but based on the random value of the VRF hash, it is going to be some value less than that based on the CDF?

jannotti:
It's not something I'm super familiar with, but I think you've got the idea.  Suppose I have 1M algos. I don't "win" the VRF lottery will all of them, or lose with all of them.  Some fraction of them win and some fraction loses, based on the outcome of the VRF.

Olshansk:
Yea, the high level idea makes sense and even most of the intricacies except for that one… Maybe it's something we could figure out together?

I've looked at section 5 of the [whitepaper](https://algorandcom.cdn.prismic.io/algorandcom%2Fa26acb80-b80c-46ff-a1ab-a8121f74f3a3_p51-gilad.pdf) and my understanding is:

Assume / let the following:
- Each user as a `Vk` (Verification Key) and a `Sk` (Secret Key)
- Vk of each user has been broadcasted to all other users in the network
- Seed `S` is determined and publically known (but only computed AFTER `Vk` has been broadcasted)
- Every user `i` has a voting power `Wi` where `W` is the voting power of the whole network

`p, hash = VRFsk(seed)`
- `p` is the proof that can be verified with `VRFvk(seed)`
- `hash` is uniformly distributed and is of size `hashlen`

The binomial distribution is set to:
- Number of trials is the individual nodes `voting power` (each unit is an individual attempt to be selected to protect against Sybil attacks)
- The probably of each trial is `number of expected candidates in committee / total voting power` (each unit is an individual sub-user)

After re-reading it, I actually think the core of the “magic” is here:

Screen Shot 2022-02-07 at 10 41 25 PM

1. Hash / 2^hashlen normalizes the uniformly distributed VRF output
2. Using the binomial distribution, we can find out if EXACTLY j sub-users are selected.

The maximum value for j is is the maximum number of their sub-users that may be selected. However, if exactly j sub-users are not selected (based on the normalized VRF output), it doesn’t mean that a value less than j might not be selected. E.g. if the total voting power is 1000 ALGO, and a single node has 100, they should still have an opportunity to be a potential committee member if only 50 of their ALGO is elected as their voting round for that view change.

My understanding is that this is the reason we’re using a CDF of the binomial distribution and accepting any value of j smaller than their individual voting power as the value to be a candidate.

PR #3559, opened 2022-02-03
Tsachi/minor sortition fix2

tsachiherman:


tsachiherman:
included in https://github.com/algorand/go-algorand/pull/3558

PR #3562, opened 2022-02-04
WIP: ledger refactoring: test 6M assets/apps per account instrumented for investigation

algonautshant:
ledger refactoring: test 6M assets/apps per account
resolves https://github.com/algorand/go-algorand-internal/issues/1857

Additional logging messages are added to investigate 2 outstanding problems:
1. account information returned with 0 assets or apps when there are 1000 of them (without reporting error)
2. cow cash miss


test/e2e-go/features/accountPerf/fourMillion_test.go
```diff
@@ -0,0 +1,355 @@
...
+		if txn == nil {
+			break
+		}
+		walletHandle, err := client.GetUnencryptedWalletHandle()
+		require.NoError(t, err)
+
+		stxn, err := client.SignTransactionWithWallet(walletHandle, nil, *txn)
+		if err != nil {
+			fmt.Printf("Error signing: %v\n", err)
+		}
```
tsachiherman, line 71: There are two issues here that would kill the performance:
1. Since you're using a devmode network, please try to use transaction groups of size 16. It would make each block contain these 16 transactions.
2. avoid using the KMD. It won't be sufficiently performant.
3. when sending transaction, use the BroadcastTransactionGroup.

algonautshant: Done!

data/basics/userBalance.go
```diff
@@ -56,7 +56,7 @@ const (
 	// EncodedMaxAppLocalStates is the decoder limit for number of opted-in apps in a single account.
 	// It is verified in TestEncodedAccountAllocationBounds to align with
 	// config.Consensus[protocol.ConsensusCurrentVersion].MaxppsOptedIn
-	EncodedMaxAppLocalStates = 64
+	EncodedMaxAppLocalStates = 6000001
```
algonautshant, line 59: `		ad, err := client.AccountData(nacc.pk.String()) `
will return an error if this bound is not increased.
There are similar parameters here which are not tested for unlimited apps/accounts.

test/e2e-go/features/accountPerf/sixMillion_test.go
```diff
@@ -0,0 +1,1322 @@
...
+	expectedCount uint64,
+	address string,
+	context string) (info generated.Account, err error) {
+
+	for x := 0; x < 50; x++ { // retry only 50 times
+		info, err = client.AccountInformationV2(address, true)
+		if err == nil {
+			if expectedCount > 0 && int(expectedCount) != len(*info.CreatedApps) {
+				fmt.Printf("Missing appsPerAccount: %s got: %d expected: %d\n", address, len(*info.CreatedApps), expectedCount)
+				fmt.Printf("%+v\n\n", info)
```
cce, line 149:
```suggestion
				fmt.Printf("%s\n\n", spew.Sdump(info))
```

PR #3571, opened 2022-02-04
Add comment to help explain txntest.Txn rationale

michaeldiamant:
## Summary

Adds inline comment linking to the PR that created `txntest.Txn`.  For developers new to _go-algorand_, the originating PR provides context to help understand _how_ `txntest.Txn` simplifies life.

## Test Plan
N/A - Only changes comments.

### DIFF
--- a/data/txntest/txn.go
+++ b/data/txntest/txn.go
@@ -47,6 +47,7 @@ import (
 
 // Txn exists purely to make it easier to write a
 // transaction.Transaction in Go source.
+// To better understand Txn rationale, see https://github.com/algorand/go-algorand/pull/2730.
 type Txn struct {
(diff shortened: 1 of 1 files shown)


jannotti (approved):
ok

algorandskiy (changes requested):
I personally do not think it is a good practice to refer to another PR. If more explanation needed add a summary comment right in the code.

michaeldiamant:
> I personally do not think it is a good practice to refer to another PR. If more explanation needed add a summary comment right in the code.

@algorandskiy I'm on-board with the suggestion.  I replaced the PR with https://github.com/algorand/go-algorand/pull/3574 to address the feedback.

PR #3575, opened 2022-02-04
Moving windows build to Github Actions

algojack:
Moved it to build-test-windows here on Github.
Artifact gathering and upload are not currently working, since they seem to be cached files and we are not sure how useful those are. But if we can determine what exactly we need and for what purpose, we can add them back in.

Slack notifications for failed nightly builds added using slack webhook to #dev-alerts channel.
Travis integration removed.

Example slack message:
image


onetechnical (dismissed):
The python version is interesting, but otherwise looks alright to me.

.github/workflows/build.yml
```diff
@@ -0,0 +1,43 @@
...
+      #   uses: actions/upload-artifact@v2
+      #   with:
+      #     if-no-files-found: error
+      #     path: $(git ls-files -o | grep -v crypto/libs | grep -v crypto/copies | grep -v crypto/libsodium-fork | grep -v crypto/lib/ | grep -v ^gen/ | grep -v swagger.json.validated | tr "\n" ":")
+      - name: Slack Notification
+        env:
+          SLACK_WEBHOOK: ${{ secrets.SLACK_WEBHOOK }}
+        run: |
+          curl -X POST --data-urlencode "payload={\"text\": \"Nightly windows build test on Github failed. ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }} \"}" $SLACK_WEBHOOK
+        if: ${{ failure() && contains(github.base_ref, 'rel/nightly') }}
```
algobarb, line 43: Can we get a message for `rel/*` here? Then, we can get failure messages for rel/beta merge too.

algojack: Added

algojack:
> The python version is interesting, but otherwise looks alright to me.

Removed, got it from circleci attempt to implement this, but seems we don't need it.

algobarb (approved):

algojohnlee (approved):

PR #3577, opened 2022-02-05
adding another field for lsig disassembly

### LINKED ISSUE #3563: DryrunTransactionResult `disassembly` field ambiguous
## Problem

When debugging a set of transactions using the Dryrun REST endpoint, if an application call is signed by logic two programs are evaluated but there is only space for one disassembled program:

https://github.com/algorand/go-algorand/blob/0d65a0f93f214087569939ad2cea0c0f1b15d8d8/daemon/algod/api/server/v2/generated/types.go#L310-L311

## Solution

Add another field to allow both disassembled programs to be returned.

barnjamin:
## Summary

Adding separate field for disassembled lsig program.

closes #3563 

## Test Plan

None yet


jannotti:
Would like to get at least one more opinion on this. Current return value certainly seems wrong.  It has one place for disassembly, and if there's an app it simply overwrites the logicsig disassembly.  I would consider this PR a win compared to that.  It adds a new place to tuck the lsig disassembly so that the potential app disassembly overwrites only the first copy of it (which is still returned for backward compatibility).

I wonder if we shouldn't add yet another place for disassembly - a place just for app disassembly. Then, in a latter version we could remove the existing location which is "shared" by logicsigs and apps.

Of course, I don't really see why we're returning disassembly at all.  Seems an orthogonal issue to dry-run.  But we are, so let's do it well.

Maybe @jasonpaulos @winder @JasonWeathersby would have feelings about such an API change.

winder:
> Would like to get at least one more opinion on this. Current return value certainly seems wrong.  It has one place for disassembly, and if there's an app it simply overwrites the logicsig disassembly.  I would consider this PR a win compared to that.  It adds a new place to tuck the lsig disassembly so that the potential app disassembly overwrites only the first copy of it (which is still returned for backward compatibility).
> 
> I wonder if we shouldn't add yet another place for disassembly - a place just for app disassembly. Then, in a latter version we could remove the existing location which is "shared" by logicsigs and apps.
> 
> Of course, I don't really see why we're returning disassembly at all.  Seems an orthogonal issue to dry-run.  But we are, so let's do it well.
> 
> Maybe @jasonpaulos @winder @JasonWeathersby would have feelings about such an API change.

The overloaded disassembly field does seem strange. If the field really does have dubious value, a query param to select what (and whether) you'd like to disassemble something would work instead of adding another field.

I'm not familiar with the functionality to have any real opinion about this. Seems fine to add of it's useful, but it does require updating the SDKs.

jasonpaulos (approved):
I agree that I'm not sure how useful the disassembly field for this endpoint is in general, but the fact that the lsig disassembly gets overridden is a clear bug, so this makes sense as an incremental improvement.

daemon/algod/api/server/v2/dryrun.go
```diff
@@ -400,7 +400,8 @@ func doDryrunRequest(dr *DryrunRequest, response *generated.DryrunResponse) {
 			ep.Debugger = &debug
 			pass, err := logic.Eval(stxn.Lsig.Logic, ep)
 			var messages []string
-			result.Disassembly = debug.lines
+			result.Disassembly = debug.lines          // Keep backwards compat
+			result.LogicSigDisassembly = &debug.lines // Also add to Lsig specific
```
jannotti, line 404: Why does these have different types?  They look the same in oas3.yml and oas2.json

jannotti: Since dry-run requests have at least one program (lsig or app) the disassembly will always get populated. But lsig-disassembly would not.

Though I think dry run requests can have some normal txns in them (just a simple pay), right? That would not populate disassembly at all.  You you get DryrunTxnResponse for such?

barnjamin: yea, I get an object with `{disassembly:""}`

jannotti (approved):
Looks good.  Let's recall that we may wish to add an app specific disassembly field, and then deprecate the old shared field. Perhaps we would do this at once in a breaking v3 update.

PR #3579, opened 2022-02-06
Update consensus v31

Aharonee:
## Summary

New consensus version containing state proof parameters for the new participation key registration.


tsachiherman:
@Aharonee , I'd rather having this targeting master directly, since we'd like to make a *single* consensus upgrade for all the changes in master. ( i.e. each change takes a week before being applied; there is no point in waiting for few weeks for these features.. )

tsachiherman:
I believe that we won't need this one anymore, right ?

PR #3586, opened 2022-02-07
Document rationale for disallowing clear state program inner transactions

michaeldiamant:
## Summary
Adds inline comment explaining how the team decided to drop support for inner transactions in a clear state program.

* Relates to https://github.com/algorand/go-algorand/pull/3556/.
* Consider the PR optional.  As a new team member, I felt the comment benefits future readers' comprehension.  If folks feel the explanation is obvious, then I welcome closing the PR.  

## Test Plan
N/A - Only changes comments.

### DIFF
--- a/data/transactions/logic/eval.go
+++ b/data/transactions/logic/eval.go
@@ -4062,6 +4062,13 @@ func opTxBegin(cx *EvalContext) {
 		cx.err = errors.New("itxn_begin without itxn_submit")
 		return
 	}
+
+	// Product decision - Prevent clear state programs (CSPs) from issuing inner
+	// transactions to minimize complexity.
+	// * If CSPs support inner transactions, downstream program errors may
+	//   complicate CSP guarantees for opting out accounts from apps.
+	// * Since the team felt CSP inner transactions are a narrow(er) use case,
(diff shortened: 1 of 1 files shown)


data/transactions/logic/eval.go
```diff
@@ -4062,6 +4062,13 @@ func opTxBegin(cx *EvalContext) {
 		cx.err = errors.New("itxn_begin without itxn_submit")
 		return
 	}
+
+	// Product decision - Prevent clear state programs (CSPs) from issuing inner
+	// transactions to minimize complexity.
+	// * If CSPs support inner transactions, downstream program errors may
+	//   complicate CSP guarantees for opting out accounts from apps.
```
jannotti, line 4069: I don't think that's real reason.  It has more to do with the inability to properly account for fees paid for transactions that never end up committing (because the CSP failed, so its effects must rollback).  Some of the possible solutions to _that_ problem have have complications that would make it difficult for accounts to opt out of an account.

All in all, I think explaining the rationale would take a page of code and maybe isn't justified.

michaeldiamant: Had an out-of-band discussion and opting to defer on the PR for now.  Resolving.

PR #3588, opened 2022-02-07
devmode: avoid returning participation keys

tsachiherman:
## Summary

When running a devmode node, avoid returning any keys from `VotingKeys`. By avoiding returning keys, it would prevent the agreement from generating proposals and votes.

## Test Plan

Tested manually.


algorandskiy (approved):

tsachiherman:
I'm going to hold off this one until we release master; I'd like to reduce the amount of changes going into the upcoming release.

PR #3590, opened 2022-02-08
Add catchpointdump into tools package

algorandskiy:
## Summary

Having `catchpointdump` in tools looks as not a bad idea in general.
In particular I need a linux build of `catchpointdump` on my Mac machine in order
to plug it into algonet linux network. The only way to make such a cross build is calling
`deploy_linux_version.sh` that builds and upload packages in a docker container.

Modified `create_and_deploy_recipe.sh` to support multicomponent `GOPATH`

## Test Plan

Tested manually and it works as expected:
1. catchpointdump is in tools package
2. Multicomponent GOPATH supported well
3. Content of tmp/go-cache is not sent into docker

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -29,6 +29,7 @@ cmd/updater/updater
 tmp/dev_pkg
 tmp/out
 tmp/node_pkgs
+tmp/go-cache
 
 # Ignore vim backup and swap files
 *~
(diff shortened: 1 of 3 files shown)


tsachiherman:
please test this and let me know if it works as expected.

tsachiherman (approved):

PR #3592, opened 2022-02-08
move txn template filler above fee setting

### LINKED ISSUE #3591: Goal asset send does not respect fee set to 0
### Subject of the issue

When trying to create a group transaction where the fee on some transactions are set to 0 goal does not respect the 0 setting

```
ben@Algo-Guidarelli:~$ sb goal asset send --assetid 1 -f P75HT5DGUZ3TXI2TAESMJ2O7SMIC3WYFFQP5XARFTMAINI6OXOHANVMCOM -t P75HT5DGUZ3TXI2TAESMJ2O7SMIC3WYFFQP5XARFTMAINI6OXOHANVMCOM -a 100 -o tmp.txn --fee 0
ben@Algo-Guidarelli:~$ sb goal clerk inspect tmp.txn
tmp.txn[0]
{
  "txn": {
    "aamt": 100,
    "arcv": "P75HT5DGUZ3TXI2TAESMJ2O7SMIC3WYFFQP5XARFTMAINI6OXOHANVMCOM",
    "fee": 1000,
    "fv": 1,
    "gh": "XNWO3Gc3JDDO1Uj0pab/aHGs0RZAJAnwVwrnC4fWuxM=",
    "lv": 1001,
    "note": "MNg6BVaswnM=",
    "snd": "P75HT5DGUZ3TXI2TAESMJ2O7SMIC3WYFFQP5XARFTMAINI6OXOHANVMCOM",
    "type": "axfer",
    "xaid": 1
  }
}
```

### Your environment
sandbox with feature/contract-to-contract

barnjamin:
## Summary

closes #3591 

## Test Plan

```sh
goal asset send --assetid 1 -f P75HT5DGUZ3TXI2TAESMJ2O7SMIC3WYFFQP5XARFTMAINI6OXOHANVMCOM -t P75HT5DGUZ3TXI2TAESMJ2O7SMIC3WYFFQP5XARFTMAINI6OXOHANVMCOM -a 100 -o tmp.txn --fee 0

goal clerk inspect tmp.txn # should have no fee set
```

### DIFF
--- a/cmd/goal/asset.go
+++ b/cmd/goal/asset.go
@@ -530,21 +530,23 @@ var sendAssetCmd = &cobra.Command{
 			reportErrorf("Cannot construct transaction: %s", err)
 		}
 
+		tx.Note = parseNoteField(cmd)
+		tx.Lease = parseLease(cmd)
+
 		firstValid, lastValid, err = client.ComputeValidityRounds(firstValid, lastValid, numValidRounds)
 		if err != nil {
 			reportErrorf("Cannot determine last valid round: %s", err)
 		}
+
 		tx, err = client.FillUnsignedTxTemplate(sender, firstValid, lastValid, fee, tx)
 		if err != nil {
 			reportErrorf("Cannot construct transaction: %s", err)
 		}
+
 		explicitFee := cmd.Flags().Changed("fee")
 		if explicitFee {
 			tx.Fee = basics.MicroAlgos{Raw: fee}
 		}
-
-		tx.Note = parseNoteField(cmd)
-		tx.Lease = parseLease(cmd)
 
 		if outFilename == "" {
 			wh, pw := ensureWalletHandleMaybePassword(dataDir, walletName, true)
--- a/cmd/goal/asset.go
+++ b/cmd/goal/asset.go
@@ -529,13 +529,6 @@ var sendAssetCmd = &cobra.Command{
 		if err != nil {
 			reportErrorf("Cannot construct transaction: %s", err)
 		}
-		explicitFee := cmd.Flags().Changed("fee")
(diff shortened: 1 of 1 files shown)


cmd/goal/asset.go
```diff
@@ -545,6 +538,13 @@ var sendAssetCmd = &cobra.Command{
 		if err != nil {
 			reportErrorf("Cannot construct transaction: %s", err)
 		}
+		explicitFee := cmd.Flags().Changed("fee")
```
algochoi, line 541: Is this change okay for single transactions? I see that in `client.FillUnsignedTxTemplate(sender, firstValid, lastValid, fee, tx)`, there is an explicit check for [`if fee == 0`](https://github.com/algorand/go-algorand/blob/941c8204591b166e68bb205d04ab0eb0b7b318cb/libgoal/transactions.go#L402) to set the fee to the minimum Tx Fee. It looks like this change might override that setting for both single and group transactions.

algochoi (commented):
Just had a question about setting min fees

algochoi (dismissed):

jannotti:
Could I convince you to return the note and lease processing back to their old spot?  It is correct in either location, but all the other commands do those lines early.  I'd rather keep the commands as similar to one another as possible (increasing the likelihood that some enthusiastic potential contributor abstracts the commonality and puts all that work in one place)

jannotti (approved):
Thanks for catching. I'm pretty sure that was my bug.

ahangsu (approved):
sorry for the late response, I overlooked the `FillUnsignedTxTemplate` and I was wondering what could be wrong. Now I am convinced.

PR #3605, opened 2022-02-09
Update goal and libgoal

nicholasguoalgorand:
## Summary

Update goal and libgoal to support changes to the AccountInformation Rest API. Refactor goal commands to use the AccountInformationV2 if possible and pull only the information required to complete.

## Test Plan

Modified existing expect and e2e tests.


daemon/algod/api/client/restClient.go
```diff
@@ -402,8 +407,12 @@ func (client RestClient) AccountInformation(address string) (response v1.Account
 }
 
 // AccountInformationV2 gets the AccountData associated with the passed address
-func (client RestClient) AccountInformationV2(address string) (response generatedV2.Account, err error) {
-	err = client.get(&response, fmt.Sprintf("/v2/accounts/%s", address), nil)
+func (client RestClient) AccountInformationV2(address string, includeAppAssets bool) (response generatedV2.Account, err error) {
+	if includeAppAssets {
```
cce, line 411:
```suggestion
func (client RestClient) AccountInformationV2(address string, includeCreatables bool) (response generatedV2.Account, err error) {
	if includeCreatables {
```
tiny nit, wdyt

cmd/goal/interact.go
```diff
@@ -694,19 +693,18 @@ var appQueryCmd = &cobra.Command{
 			}
 
 			// Get creator information
-			ad, err := client.AccountData(app.Params.Creator)
+			resp, err := client.AccountApplicationInformation(app.Params.Creator, appIdx)
```
cce, line 696: The AccountApplicationInformation and AccountAssetInformation endpoints support msgpack too (returns a ledgercore.AccountResource object), so adding a method to hit the "raw" format like `libgoal.Client:AccountData` does would result in fewer changes to implement type conversions between the JSON generated types and the basics types. Doing this would make us feel a little safer that there were no unexpected differences marshaling and unmarshaling through JSON on its way to goal.

cmd/goal/account.go
```diff
@@ -611,31 +613,35 @@ func printAccountInfo(client libgoal.Client, address string, account generatedV2
...
 	}
 	for _, assetHolding := range heldAssets {
-		assetParams, err := client.AssetInformationV2(assetHolding.AssetId)
-		if err != nil {
-			hasError = true
-			fmt.Fprintf(errorReport, "Error: Unable to retrieve asset information for asset %d referred to by account %s: %v\n", assetHolding.AssetId, address, err)
-			fmt.Fprintf(report, "\tID %d, error\n", assetHolding.AssetId)
-		}
+		if onlyShowAssetIds {
+			fmt.Fprintf(report, "\tID %d\n", assetHolding.AssetId)
```
tsachiherman, line 617: code style: use `continue` here so that you won't need to deep branching.

cmd/goal/interact.go
```diff
@@ -21,6 +21,7 @@ import (
 	"encoding/base64"
 	"encoding/json"
 	"fmt"
+	"github.com/algorand/go-algorand/daemon/algod/api/server/v2/generated"
```
tsachiherman, line 24: move to the bottom group

daemon/algod/api/client/restClient.go
tsachiherman: I agree with @cce here, and maybe consider writing this as:
```golang
func (client RestClient) AccountInformationV2(address string, includeCreatables bool) (response generatedV2.Account, err error) {
        var infoParams accountInformationParams
 	if includeCreatables {
              infoParams = accountInformationParams{Include: "all"}
        } else {
              infoParams = accountInformationParams{Include: "none"}
        }
        err = client.get(&response, fmt.Sprintf("/v2/accounts/%s", address), infoParams)
        return
}
```

cmd/goal/interact.go
nicholasguoalgorand: I just realized that the raw format still doesn't work since the `AccountResource` type holds a bunch of pointers, and simply encoding with messagepack only encodes the pointers. Any suggestions on how to make the raw AccountApplicationInformation api work properly?

ledger/ledgercore/accountresource.go
```diff
@@ -28,6 +28,34 @@ type AccountResource struct {
 	AppParams     *basics.AppParams
 }
 
+// AccountResourceModel used to encode AccountResource
+type AccountResourceModel struct {
```
tsachiherman, line 32: this need to be "transplanted" into `/go-algorand/daemon/algod/api/spec/v2/model.go`, since it's a UI model only.

libgoal/libgoal.go
```diff
@@ -19,6 +19,7 @@ package libgoal
 import (
 	"encoding/json"
 	"fmt"
+	"github.com/algorand/go-algorand/ledger/ledgercore"
```
tsachiherman, line 22: move to the bottom list.

test/scripts/e2e_subs/assets-app-b.sh
```diff
@@ -40,7 +40,7 @@ XFER4=11
 APP_CREATED_STR='Created app with app index'
 ERR_APP_CL_STR='only clearing out is supported for applications that do not exist'
 ERR_APP_NE_STR='application does not exist'
-ERR_APP_OI_STR1='has not opted in to application'
+ERR_APP_OI_STR1='HTTP 404 Not Found: {"message":"account application info not found"}'
```
tsachiherman, line 43: this error should not get modified : the goal need to be smart enough to identify this use case and emit a backward-compatible message.

nicholasguoalgorand: The Rest API isn't smart enough to identify this use case. It only tells you whether the account resource exists or not and does not differentiate between the app not existing or the user not opting in. Is it worth adding an extra api call to tell whether the app exists or the user hasn't opted in?

tsachiherman: no - this isn't a REST API issue. The error return by libgoal should contain the inner error message rather than the HTTP error. ( i.e. strip the outer HTTP error )
Separately, I think that "has not opted in to application" is definitely more informative than "account application info not found", since it provides a piece of information that the latter does not contain.

nicholasguoalgorand: "has not opted in to application" is not an accurate error message anymore because it might not be the case

tsachiherman (approved):
looks good to me.

PR #3608, opened 2022-02-09
fix minor regression in txnMerkleToRaw

tsachiherman:
## Summary

My previous PR was causing a minor regression. Thanks to @jannotti comment, I think that I was able to fix the regression, and added a proper testing for that.

## Test Plan

New benchmark result ( no inner method anymore )
```
goos: darwin
goarch: amd64
pkg: github.com/algorand/go-algorand/data/bookkeeping
cpu: Intel(R) Core(TM) i7-1068NG7 CPU @ 2.30GHz
BenchmarkTxnMerkleToRaw/copy-8         	1000000000	         1.129 ns/op	       0 B/op	       0 allocs/op
BenchmarkTxnMerkleToRaw/append-8       	243079730	         4.886 ns/op	       0 B/op	       0 allocs/op
PASS
ok  	github.com/algorand/go-algorand/data/bookkeeping	3.183s
```

### DIFF
--- a/data/bookkeeping/txn_merkle.go
+++ b/data/bookkeeping/txn_merkle.go
@@ -70,7 +70,7 @@ type txnMerkleElem struct {
 	stib transactions.SignedTxnInBlock
 }
 
-func txnMerkleToRaw(txid []byte, stib []byte) (buf []byte) {
+func txnMerkleToRaw(txid [crypto.DigestSize]byte, stib [crypto.DigestSize]byte) (buf []byte) {
 	buf = make([]byte, 2*crypto.DigestSize)
 	copy(buf[:], txid[:])
 	copy(buf[crypto.DigestSize:], stib[:])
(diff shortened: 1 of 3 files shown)


jannotti (approved):
This looks good.  Am I understanding properly that the big win came from not needing to build slices, and instead passing the digests as arrays?  Maybe that then allowed the copy loop to be unrolled because the size was hardcoded?

tsachiherman:
> This looks good. Am I understanding properly that the big win came from not needing to build slices, and instead passing the digests as arrays? Maybe that then allowed the copy loop to be unrolled because the size was hardcoded?

The fact that the size is fixed ( hence no slice ), makes it simpler for the copy, since there is no risk of "overlapping". 
Passing large arrays isn't great, since it's usually mean that we have a copy of the array; however, in the case of this method, the compiler can just inline the function and therefore eliminate the argument-copying.

As for the wins in avoiding building a slice : that's also true. The size of a slice is 24 bytes. The size of the digest is 32 bytes. so, the outcome might have been different for bigger slices. For 32 byte slices, passing it as an array seems to perform better.

Aharonee (approved):

algorandskiy (approved):

PR #3609, opened 2022-02-09
add unit test to verify merkle array root size 

tsachiherman:
## Summary

Instead of performing a check in `SignerContext.GetVerifier`, I've added a unit test to verify that the output of the merklearray root has the same length as the Verifier object.

## Test Plan

Unit test.

### DIFF
--- a/libgoal/participation.go
+++ b/libgoal/participation.go
@@ -157,7 +157,11 @@ func (c *Client) GenParticipationKeysTo(address string, firstValid, lastValid, k
 	}
 	_, err = os.Stat(partKeyPath)
 	if !os.IsNotExist(err) {
-		err = fmt.Errorf("ParticipationKeys exist for the range %d to %d", firstRound, lastRound)
+		if err != nil {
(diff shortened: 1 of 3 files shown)


libgoal/participation.go
```diff
@@ -157,7 +157,11 @@ func (c *Client) GenParticipationKeysTo(address string, firstValid, lastValid, k
 	}
 	_, err = os.Stat(partKeyPath)
 	if !os.IsNotExist(err) {
-		err = fmt.Errorf("ParticipationKeys exist for the range %d to %d", firstRound, lastRound)
+		if err != nil {
```
Aharonee, line 160: If find this flow a bit confusing to read, how about rewriting it as:

```
if err == nil {
	err = fmt.Errorf("ParticipationKeys exist for the range %d to %d", firstRound, lastRound)
	return
} else if !os.IsNotExist(err) {
	err = fmt.Errorf("participation key file '%s' cannot be accessed : %w", partKeyPath, err)
	return
}
```

tsachiherman: done.

PR #3612, opened 2022-02-11
AVM: Improve error handling and execution time

jannotti:
This reworks the opcode function interface to return errors directly instead of setting cx.err which was error-prone.

It also makes a couple small optimizations.

### DIFF
--- a/data/transactions/logic/assembler.go
+++ b/data/transactions/logic/assembler.go
@@ -425,7 +425,7 @@ func assembleInt(ops *OpStream, spec *OpSpec, args []string) error {
 }
 
 // Explicit invocation of const lookup and push
-func assembleIntC(ops *OpStream, spec *OpSpec, args []string) error {
+func asmIntC(ops *OpStream, spec *OpSpec, args []string) error {
 	if len(args) != 1 {
 		return ops.error("intc operation needs one argument")
 	}
@@ -436,7 +436,7 @@ func assembleIntC(ops *OpStream, spec *OpSpec, args []string) error {
 	ops.Intc(uint(constIndex))
 	return nil
 }
-func assembleByteC(ops *OpStream, spec *OpSpec, args []string) error {
+func asmByteC(ops *OpStream, spec *OpSpec, args []string) error {
 	if len(args) != 1 {
 		return ops.error("bytec operation needs one argument")
 	}
@@ -661,7 +661,7 @@ func assembleMethod(ops *OpStream, spec *OpSpec, args []string) error {
 	return ops.error("Unable to parse method signature")
 }
 
-func assembleIntCBlock(ops *OpStream, spec *OpSpec, args []string) error {
+func asmIntCBlock(ops *OpStream, spec *OpSpec, args []string) error {
 	ops.pending.WriteByte(spec.Opcode)
 	var scratch [binary.MaxVarintLen64]byte
 	l := binary.PutUvarint(scratch[:], uint64(len(args)))
@@ -681,7 +681,7 @@ func assembleIntCBlock(ops *OpStream, spec *OpSpec, args []string) error {
 	return nil
 }
 
-func assembleByteCBlock(ops *OpStream, spec *OpSpec, args []string) error {
+func asmByteCBlock(ops *OpStream, spec *OpSpec, args []string) error {
 	ops.pending.WriteByte(spec.Opcode)
 	bvals := make([][]byte, 0, len(args))
 	rest := args
@@ -726,7 +726,7 @@ func assembleAddr(ops *OpStream, spec *OpSpec, args []string) error {
 	return nil
 }
 
-func assembleArg(ops *OpStream, spec *OpSpec, args []string) error {
+func asmArg(ops *OpStream, spec *OpSpec, args []string) error {
 	if len(args) != 1 {
 		return ops.error("arg operation needs one argument")
 	}
@@ -751,7 +751,7 @@ func assembleArg(ops *OpStream, spec *OpSpec, args []string) error {
 	return asmDefault(ops, &altSpec, args)
 }
 
-func assembleBranch(ops *OpStream, spec *OpSpec, args []string) error {
+func asmBranch(ops *OpStream, spec *OpSpec, args []string) error {
 	if len(args) != 1 {
 		return ops.error("branch operation needs label argument")
 	}
@@ -764,7 +764,7 @@ func assembleBranch(ops *OpStream, spec *OpSpec, args []string) error {
 	return nil
 }
 
-func assembleSubstring(ops *OpStream, spec *OpSpec, args []string) error {
+func asmSubstring(ops *OpStream, spec *OpSpec, args []string) error {
 	err := asmDefault(ops, spec, args)
 	if err != nil {
 		return err
@@ -1040,7 +1040,7 @@ func asmGitxn(ops *OpStream, spec *OpSpec, args []string) error {
 	return ops.errorf("%s expects two or three arguments", spec.Name)
 }
 
-func assembleGlobal(ops *OpStream, spec *OpSpec, args []string) error {
+func asmGlobal(ops *OpStream, spec *OpSpec, args []string) error {
 	if len(args) != 1 {
 		return ops.errorf("%s expects one argument", spec.Name)
 	}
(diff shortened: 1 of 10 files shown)


data/transactions/logic/eval.go
```diff
@@ -1237,65 +1221,68 @@ func opDivw(cx *EvalContext) {
...
 	last := len(cx.stack) - 1
 	prev := last - 1
 	cond := cx.stack[prev].Uint < cx.stack[last].Uint
 	cx.stack[prev].Uint = boolToUint(cond)
 	cx.stack = cx.stack[:last]
+	return nil
 }
 
-func opGt(cx *EvalContext) {
+// opSwap, opLt, and opNot always succeed (return nil). So error checking elided in Gt,Le,Ge
```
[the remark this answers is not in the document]
jannotti: The opcode errors, but the opFunc does not.  Type checking is done before calling the opFunc, in the main interpreter loop. (Unless, like opEq, the type checking is bespoke.)

data/transactions/logic/eval.go
```diff
@@ -2654,51 +2668,47 @@ func opGaidImpl(cx *EvalContext, giw uint64, opName string) (sv stackValue, err
...
-		cx.err = err
-		return
+		return err
 	}
 
 	cx.stack[last] = sv
+	return nil
 }
 
 func (cx *EvalContext) getRound() (rnd uint64, err error) {
```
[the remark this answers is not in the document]
jannotti: Agree on removing them, and wonder if there is agreement in what I've found - using the named return values in combination with naked "return" is way harder to understand at a glance. It also somehow tempts the author (as here) to "flow through" more .  Just assign to the `err` and continue.  Precious braincells are them wasted trying to make sure that's correct.  I've moved to returning explicitly almost always.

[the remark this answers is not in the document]
jasonpaulos: That makes a lot of sense. With named return values, I believe there's also a risk of variable shadowing if you do `err := ...` inside of a block in the function. Go's compiler seems to be good at detecting when this is problematic, but not in all cases, so it takes me a little more mental effort to review cases where I worry this might be happening.

data/transactions/logic/eval.go
```diff
@@ -501,7 +489,7 @@ type EvalContext struct {
 	programHashCached crypto.Digest
 
 	// Stores state & disassembly for the optional debugger
-	debugState DebugState
+	debugState *DebugState
```
michaeldiamant: @jannotti For my understanding:  How did you determine it makes sense to `debugState` from struct field value to pointer?

I'm asking because the value vs pointer distinction isn't something I've had to think about in some time.  So, I'm trying to improve my heuristics for appropriate usage.

From what I can tell in the PR, it looks like `debugState` usage relies on modifying the `EvalContext` instance.  So, I'm _guessing_ the change makes the intended usage pattern clearer to the reader.

data/transactions/logic/eval.go
```diff
@@ -74,54 +74,43 @@ type stackValue struct {
 	Bytes []byte
 }
 
-func (sv *stackValue) argType() StackType {
+func (sv stackValue) argType() StackType {
```
michaeldiamant: @jannotti Similar to my question about `debugState`, can you help me understand how you decided to change `stackValue` interface method definitions from pointer to value receivers?

From my review, I'm _guessing_ the change tries to signal that `stackValue` is considered immutable-ish.

data/transactions/logic/eval.go
michaeldiamant: @jannotti Thanks for the explanation in https://github.com/algorand/go-algorand/pull/3612#issuecomment-1040390583.  Makes sense - Resolving.

data/transactions/logic/eval.go
```diff
@@ -3615,153 +3569,128 @@ func opAppGlobalGetEx(cx *EvalContext) {
...
-func opAppLocalPut(cx *EvalContext) {
+func opAppLocalPut(cx *EvalContext) error {
 	last := len(cx.stack) - 1 // value
 	prev := last - 1          // state key
 	pprev := prev - 1         // account
 
 	sv := cx.stack[last]
 	key := string(cx.stack[prev].Bytes)
 
-	if cx.Ledger == nil {
```
algorandskiy: I understand this is a repetitive check... is it prereq now?

algorandskiy (commented):
Looks good, couple suggestions/questions

data/transactions/logic/debugger.go
```diff
@@ -191,14 +191,14 @@ func valueDeltaToValueDelta(vd *basics.ValueDelta) basics.ValueDelta {
 	}
 }
 
-func (cx *EvalContext) refreshDebugState() *DebugState {
-	ds := &cx.debugState
+func (cx *EvalContext) refreshDebugState(err error) *DebugState {
```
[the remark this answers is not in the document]
jannotti: good idea

data/transactions/logic/eval.go
jannotti: Yes, `EvalContract` does the check when first entering stateful evaluation, so it seems pointless to keep checking it in the stateful ops.

data/transactions/logic/assembler.go
```diff
@@ -1925,6 +1925,12 @@ func (ops *OpStream) optimizeConstants(refs []constReference, constBlock []inter
 
 		// update all indexes into ops.pending that have been shifted by the above line
 
+		// This is a huge optimization for long repetitive programs. Takes
+		// BenchmarkUintMath from 160sec to 19s.
+		if positionDelta == 0 {
```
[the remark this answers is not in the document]
jannotti: Sadly, it's just the assembler that's faster, not execution.

data/transactions/logic/eval.go
jannotti: I think I missed the first because of it being in a weird place.  I'll fix and relocate.

stackValueFromTealValue could be changed but I didn't notice it.  My bias is to use values for read-only if the value in question is, say, 32 bytes or smaller.  That would include basics.TealValue, so I suppose I'd prefer a value argument here.

It would be nice to have a well-written benchmark to check my intuition on that 32 byte size.

jannotti: Changed both

data/transactions/logic/evalStateful_test.go
```diff
@@ -2322,22 +2317,22 @@ func TestReturnTypes(t *testing.T) {
 	ledger.NewLocal(tx.Receiver, 1, string(key), algoValue)
 	ledger.NewAccount(appAddr(1), 1000000)
 
+	// We try to form a snippet that will test every opcode, by sandwiching it
```
michaeldiamant, line 2320: @jannotti While improving my comprehension about the test, I opted to provide an optional refactoring:  https://github.com/jannotti/go-algorand/pull/4.  Can you take a look?

If you'd prefer to keep as is, feel welcomed to close the PR.  While I feel the PR _is_ an improvement, I don't feel strongly.

jannotti: merged

michaeldiamant: @jannotti Thanks for bringing in the PR.

data/transactions/logic/eval.go
michaeldiamant: @jannotti Thanks for the change in https://github.com/algorand/go-algorand/pull/3612/commits/51a882c2f40d9acf13565e5713a71c064b814b18.  And got it RE:  heuristics for considering value vs pointer arguments/receivers.

data/transactions/logic/assembler.go
```diff
@@ -425,7 +425,7 @@ func assembleInt(ops *OpStream, spec *OpSpec, args []string) error {
 }
 
 // Explicit invocation of const lookup and push
-func assembleIntC(ops *OpStream, spec *OpSpec, args []string) error {
+func asmIntC(ops *OpStream, spec *OpSpec, args []string) error {
```
michaeldiamant, line 428: @jannotti Minor:  Is there a by-design reason why https://github.com/algorand/go-algorand/pull/3612/commits/85db0ac73fe00d0c48a728a08209547e90220dca _omits_ renaming functions used in `keywords` from `assembler.go` (e.g. `assembleInt`)?

I might be missing context.  Judging from the commit diff and name, I'm assuming the intent is to universally change naming within AVM.

jannotti: I _only_ renamed all of the functions that perform assembly (and the name of their type).

michaeldiamant (approved):
@jannotti I'm leaving approval with the following notes:
* I most closely reviewed renames and changes to `data/transactions/logic/evalStateful_test.go`.
* I scanned other changes, but didn't scrutinize _all_ of them.  Since I'm new to the codebase, I leave it at your discretion if you believe a 2nd approval is warranted prior to merging.

jasonpaulos (approved):
Changes look good to me

PR #3614, opened 2022-02-11
Goal: Warn on invalid method signature assembly

jasonpaulos:
## Summary

With this PR, the assembler will now warn if an invalid method signature is used as an argument to the `method` pseudo-op. This is just a warning so the program will still compile, but it should help users be aware of mistakes they might have made.

No urgency to this PR.

## Example

test.teal:
```
#pragma version 6
method "abc"
pop
method "abc(uint64)"
pop
method "abc(uint64)void"
pop
method "abc(blah)void"
pop
int 1
return
```

```bash
$ goal clerk compile test.teal        
warning: 2: Invalid ARC-4 ABI method signature for method op: No parenthesis in method signature: "abc"
warning: 4: Invalid ARC-4 ABI method signature for method op: Error parsing return type: cannot convert the string "" to an ABI type
warning: 8: Invalid ARC-4 ABI method signature for method op: Error parsing argument type at index 0: cannot convert the string "blah" to an ABI type
3 warnings
test.teal: 43KDA5RKR22UG7M53XNVD7N7CYTS6KI7WRAAZL6MNGVUZIOHLKGLO4VYQ4
```

## Test Plan

Unit test added.

### DIFF
--- a/cmd/goal/clerk.go
+++ b/cmd/goal/clerk.go
@@ -469,7 +469,7 @@ var sendCmd = &cobra.Command{
 
 			// Append the signer since it's a rekey txn
 			if basics.Address(addr) == stx.Txn.Sender {
-				reportWarnln(rekeySenderTargetSameError)
+				reportWarnln("Warning:" + rekeySenderTargetSameError)
 			}
 			stx.AuthAddr = basics.Address(addr)
 		}
@@ -557,7 +557,7 @@ var rawsendCmd = &cobra.Command{
 				for _, txn := range txgroup {
 					txnErrors[txn.ID()] = err.Error()
 				}
-				reportWarnf(errorBroadcastingTX, err)
+				reportWarnf("Warning: "+errorBroadcastingTX, err)
 				continue
 			}
 
@@ -584,7 +584,7 @@ var rawsendCmd = &cobra.Command{
 				txn, err := client.PendingTransactionInformation(txidStr)
 				if err != nil {
 					txnErrors[txid] = err.Error()
-					reportWarnf(errorRequestFail, err)
+					reportWarnf("Warning: "+errorRequestFail, err)
 					continue
 				}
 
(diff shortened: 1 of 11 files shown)


cmd/goal/clerk.go
```diff
@@ -948,6 +948,17 @@ func assembleFile(fname string) (program []byte) {
...
+	if len(ops.Warnings) != 0 {
+		for _, warning := range ops.Warnings {
+			reportInfoln(warning.Error())
+		}
+		plural := "s"
+		if len(ops.Warnings) == 1 {
+			plural = ""
+		}
+		reportInfof("%d warning%s", len(ops.Warnings), plural)
+	}
```
jasonpaulos, line 960: I thought `goal` already reported warnings, but I was wrong, so I added this code

ahangsu (dismissed):
LGTM

jannotti (commented):
Looks good. Let's do something similar for "event" (after spending a little time defining it), and then a bare `selector` pseudo-op. Once we have those, I think this would becomes an error rather than warning.

Why did you have to suppress printing, rather than separating assembly output from earnings by using stderr?

jasonpaulos:
Originally I just used the same convention for warnings as other places in goal, which is to print to stdout. But I agree this seems odd, so I now changed it so that all warnings go to stderr for goal commands.

This is a little more far reaching than I intended, so I hope it doesn't disrupt anything else.

ahangsu:
Hey folks, I was just reviewing this PR again and was wondering about what JJ previously mentioned:

> Let's do something similar for "event" (after spending a little time defining it), and then a bare selector pseudo-op.

What is "event" referring to, does that mean another op/pseudo-op?

ahangsu:
Uh I see, thanks for explaining! I was also reading: https://medium.com/mycrypto/understanding-event-logs-on-the-ethereum-blockchain-f4ae7ba50378, seems matching what you just mentioned.

I suppose this "event" sounds like a "structured log", which allows for some structured fields are filled by ABI encodings, and it might worth it to check the validity of those fields, just like this PR.

jannotti (approved):
Seems good. Merging.

PR #3618, opened 2022-02-12
ledger-refactoring: testing network template

algorandskiy:


ledger/acctupdates.go
```diff
@@ -1312,6 +1312,7 @@ func (au *accountUpdates) lookupWithoutRewards(rnd basics.Round, addr basics.Add
 					return d, rnd, rewardsVersion, rewardsLevel, nil
 				}
 			}
+			au.log.Errorf("addr %s not in deltas (%d). offset=%d, rnd=%d, macct=%v", addr.String(), len(au.deltas), offset, rnd, macct)
```
tsachiherman, line 1315: This isn't an error. It would happen all the time..
Imagine the following:
Account is updated on round 1000.
Than, you're trying to get the balance for the account at round 999.
The account will be in 'accounts' but will be missing from the deltas since you'll be
Scanning the entries 680..999.

tsachiherman: In this case, you might find the account in the base cache.

algorandskiy: fair

tsachiherman (approved):

PR #3619, opened 2022-02-13
add a fallback for missing state proof keys in participation file

Aharonee:
## Summary
Restoring participation has failed in the nightly build, since it was trying to query for StateProofKeys table which doesn't exist yet.

Made sure to call the migration in the correct places to fix this issue.

## Test Plan

Unit tests added.

### DIFF
--- a/crypto/merklesignature/persistentMerkleSignatureScheme_test.go
+++ b/crypto/merklesignature/persistentMerkleSignatureScheme_test.go
@@ -48,7 +48,7 @@ func TestSecretsDatabaseUpgrade(t *testing.T) {
 	a.NoError(err)
 
 	err = store.Atomic(func(ctx context.Context, tx *sql.Tx) error {
-		err := InstallDatabase(tx) // assumes schema table already exists (created by partInstallDatabase)
+		err := InstallStateProofTable(tx) // assumes schema table already exists (created by partInstallDatabase)
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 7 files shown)


crypto/merklesignature/persistentMerkleSignatureScheme_test.go
```diff
@@ -114,3 +117,19 @@ func createTestDB(a *require.Assertions) *db.Accessor {
...
 	return &store
 }
+
+func getStateProofTableSchemaVersions(db db.Accessor) (int, error) {
+	var version int
+	err := db.Atomic(func(ctx context.Context, tx *sql.Tx) (err error) {
+		row := tx.QueryRow("SELECT version FROM schema where tablename = ?", merkleSignatureTableSchemaName)
+
+		err = row.Scan(&version)
+		if err != nil {
```
tsachiherman, line 127: if err is `sql.ErrNoRows` than you'd better return 0, since it just means that the database was never upgraded.

data/account/participation.go
```diff
@@ -48,7 +48,7 @@ type Participation struct {
 
 	VRF    *crypto.VRFSecrets
 	Voting *crypto.OneTimeSignatureSecrets
-	// StateProofSecrets is used to sign compact certificates. might be nil
+	// StateProofSecrets is used to sign compact certificates. MIGHT BE NIL!
```
tsachiherman, line 51: nothing in this line is wrong; but anytime we have a pointer variable, we should assume that it could be nil, unless we explicitly verify that.

tsachiherman (dismissed):

crypto/merklesignature/persistentMerkleSignatureScheme_test.go
tsachiherman: just because it better aligns with the database versioning : 
no version means version 0; any other version means a specific version.
In your case, returning version 0 here would fail in 
```golang
a.Equal(merkleSignatureSchemaVersion, version)
```

tsachiherman (dismissed):

tsachiherman (approved):

PR #3621, opened 2022-02-14
Participation key tests

algonathan:
## Summary
I've encountered an issue with the participation keys expiration test, which might cause flakiness. 

## Test Plan
I've managed to reproduce the same error as shown in 
https://app.circleci.com/pipelines/github/algorand/go-algorand/5013/workflows/8c1f10b2-c827-4c0a-81a3-cfe18ec0c794/jobs/78980
By halting the secondary prior to the request of sclient for the round the rich client has.

### DIFF
--- a/test/e2e-go/features/participation/participationExpiration_test.go
+++ b/test/e2e-go/features/participation/participationExpiration_test.go
@@ -145,6 +145,10 @@ func testExpirationAccounts(t *testing.T, fixture *fixtures.RestClientFixture, f
 
 	_, richAccountLatestRound = fixture.GetBalanceAndRound(richAccount)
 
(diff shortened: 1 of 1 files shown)


tsachiherman (commented):
The changes here aren't bad, and probably would fix the failure; however, I do have few comments :
1. The test contains several calls to `GetBalanceAndRound`; most of them ignores the balance completely. Why don't we replace these with GetStatus() instead.
2. The added `sClient.WaitForRound(richAccountLatestRound)` makes sense, however, I would suggest replacing it with `sClient.WaitForRound(lastValidRound+1)` which aligns with the round number that the primary node was waited for.

tsachiherman (approved):
looks good to me. thanks for the changes.

PR #3623, opened 2022-02-14
ledger: fix ledger.Eval code comments

algorandskiy:
* Minor linter fixes in internal package as well


tsachiherman (approved):
looks good.

PR #3624, opened 2022-02-14
testing: improve TestAccountInformationV2

tsachiherman:
## Summary

The TestAccountInformationV2 test was making the assumption that a transaction would get included into a block with 5 rounds. This assumption might not hold true.

In this PR, we're waiting for the transaction to be included, and then using the PendingTransactionInformation to determine on which round it was included.

## Test Plan

This is a test.


gmalouf (approved):

algonautshant (approved):
Look correct!

PR #3631, opened 2022-02-15
testing: fix bug in TestAssetGroupCreateSendDestroy

tsachiherman:
## Summary

Fix a bug in TestAssetGroupCreateSendDestroy. The test uses two nodes, sending transaction to one and checking the state on the other. Before checking, the test need to verify that both nodes successfully reached the same round number.

## Test Plan

This is a test.

### DIFF
--- a/test/e2e-go/features/transactions/asset_test.go
+++ b/test/e2e-go/features/transactions/asset_test.go
@@ -632,9 +632,18 @@ func TestAssetGroupCreateSendDestroy(t *testing.T) {
 	err = client0.BroadcastTransactionGroup(stxns)
 	a.NoError(err)
 
-	_, curRound := fixture.GetBalanceAndRound(account0)
-	confirmed := fixture.WaitForAllTxnsToConfirm(curRound+5, txids)
+	status0, err := fixture.LibGoalClient.Status()
+	a.NoError(err)
+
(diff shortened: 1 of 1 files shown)


test/e2e-go/features/transactions/asset_test.go
```diff
@@ -632,10 +632,19 @@ func TestAssetGroupCreateSendDestroy(t *testing.T) {
...
+	a.NoError(err)
+
+	confirmed := fixture.WaitForAllTxnsToConfirm(status0.LastRound+5, txids)
 	a.True(confirmed)
 
+	status0, err = client0.Status()
+	a.NoError(err)
+
+	// wait for client1 to reach the same round as client0
+	_, err = client1.WaitForRound(status0.LastRound)
```
algonautshant, line 645: So, I understand that the idea here is that both nodes agree on the block with all the transactions, `client0` adds the block in its ledger, returns `confirmed` on line 638.

However, it could be possible that `client1` has not yet added the block and advanced the round, and the `client1.AssetInformation(assetID1)` fails. 

So, by `client1.WaitForRound(status0.LastRound)` you are making sure that `client1` has also added the block before calling `client1.AssetInformation(assetID1)`.

Is this right?

algonautshant (approved):
Looks great. 
Just one comment to confirm my understanding.

tsachiherman:
Exactly

PR #3632, opened 2022-02-15
API: Fix prove bug on API

algoidan:
## Summary

This PR fixes a bug on algod's API. When a tree contains a missing child (not a full tree), the api handler omits this from the proof response and leads to a root mismatch 

## Test Plan
Add unit tests as well as convert the e2e to test this edge case.

### DIFF
--- a/test/e2e-go/features/transactions/proof_test.go
+++ b/test/e2e-go/features/transactions/proof_test.go
@@ -76,7 +76,8 @@ func TestTxnMerkleProof(t *testing.T) {
 
 	// Transfer some money to acct0, as well as other random accounts to
 	// fill up the Merkle tree with more than one element.
-	for i := 0; i < 10; i++ {
+	// we do not want to have a full tree in order the catch an empty element edge case
+	for i := 0; i < 4; i++ {
 		accti, err := client.GenerateAddress(walletHandle)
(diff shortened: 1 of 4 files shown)


test/e2e-go/features/transactions/proof_test.go
```diff
@@ -120,6 +121,19 @@ func TestTxnMerkleProof(t *testing.T) {
...
+	s, _, err := blk.DecodeSignedTxn(blk.Payset[proofresp.Idx])
+	a.NoError(err)
+
+	if s.ID().String() != txid.String() {
+		t.Log(txid.String())
+		for i := 0; i < len(blk.Payset); i++ {
+			s, _, _ := blk.DecodeSignedTxn(blk.Payset[i])
+			t.Log(s.Txn.ID())
+		}
+		t.FailNow()
```
tsachiherman, line 134: I think that the intent here is correct - but you should log more useful information that would help you to debug the issue further.
In particular, I'd suggest dumping the entire `confirmedTx` ( which includes the block number ), as well as the entire block.

algoidan: thanks for the advice

tsachiherman (approved):

Issue #1686, opened 2020-11-05
Improved Network Create Command


winder:
## Summary
`goal node create --network <network> --destination ./` is incomplete and can be improved.

The ingredients to configure a node for a specific network are:
1. A compatible release
2. The correct genesis.json file
3. The correct DNS Bootstrap URL (used to find a relay)

The `--network` currently ONLY attempts to properly set the genesis.json file, and even that piece has problems. The main problem is that there is not a consistent way for the `--network` option to find the genesis file, it searches 3 locations and the files are not always in those locations.

## Open Question

Is there an opportunity for consolidation? The `genesis.json` files are shipped in many ways:
* `make install` places genesis files in the `$GOPATH/bin/genesisfiles` directory
* The `update.sh` script installs them relative to the binaries
* The release tarballs include genesis files in the `<archive>/genesis/` directory
* The deb/rpm installers include genesis files in `/var/lib/algorand`
* The sandbox includes and manages them internally

Who is actually using `goal network create`? I see it in a few places and ONLY to install a private network based on a template.

## Scope/Requirements

There are two options

1) Make the "template" field required. This is now a non-issue because `goal network create` wouldn't need to lookup the genesis.json file.

2) If we decide we can't remove the install-a-real-network mode we could modify `goal node create` with the following:
* Bake the genesis files into the binary. mainnet/testnet/betanet/devnet are all committed to the go-algorand repository and serve as the source of truth, so they can be included in the goal command.
* Alongside the baked in genesis files include the DNS Bootstrap ID configuration.
* The network option should ONLY allow using the baked in configurations.

## Urgency/Relative Priority

This command is extremely confusing to developers and customers.
The problem of finding genesis files should be something we have an actual design for, currently the design is organic and keeps being redesigned when someone runs into the problem.

kehiy:
@winder is this issue done?

PR #3639, opened 2022-02-16
goal node create improvements

algorandskiy:
## Summary

1. `--full-config` options allows saving config file with all options, not only non-default.
2. Embed genesis.json for mainnet, testnet, betanet, devnet

## Test Plan

Tested manually


jasonpaulos (dismissed):
Changes make sense to me

cmd/goal/node.go
```diff
@@ -595,6 +597,10 @@ var createCmd = &cobra.Command{
 		secondChoicePath := filepath.Join("var", "lib", "algorand", "genesis", newNodeNetwork, "genesis.json")
 		thirdChoicePath := filepath.Join(exePath, "genesisfiles", "genesis", newNodeNetwork, "genesis.json")
 		paths := []string{firstChoicePath, secondChoicePath, thirdChoicePath}
+		if gopath := os.Getenv("GOPATH"); len(gopath) > 0 {
+			fourthChoicePath := filepath.Join(gopath, "src", "github.com", "algorand", "go-algorand", "installer", "genesis", newNodeNetwork, "genesis.json")
+			paths = append(paths, fourthChoicePath)
+		}
```
tsachiherman, line 603: I'm not a big fan of this approach. The goal should come with pre-baked knowledge of the 3 supported networks, plus the ability to specify a custom genesis file. There shouldn't be any "search" done.

algorandskiy: I kind of agree, the GOPATH approach not ideal but at least it makes `goal node create` working on developers machines why do not use updater/packages.

algorandskiy:
@tsachiherman followed your suggestion and embedded genesis files.

cmd/goal/bundle_genesis_json.sh
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env bash
+
+THISDIR=$(dirname $0)
+
+cat <<EOM | gofmt > $THISDIR/bundledGenesisInject.go
+// Code generated by bundle_genesis_json.sh, along with langspec.json. DO NOT EDIT.
```
jasonpaulos, line 6: I don't think langspec.json is relevant here

cmd/goal/bundledGenesisInject.go
```diff
@@ -0,0 +1,2793 @@
+// Code generated by bundle_genesis_json.sh, along with langspec.json. DO NOT EDIT.
```
winder, line 1: so close but so far: https://pkg.go.dev/embed

cmd/goal/node.go
```diff
@@ -92,6 +96,7 @@ func init() {
 	createCmd.Flags().BoolVarP(&newNodeIndexer, "indexer", "i", localDefaults.IsIndexerActive, "Configure the new node to enable the indexer feature (implies --archival)")
 	createCmd.Flags().StringVar(&newNodeRelay, "relay", localDefaults.NetAddress, "Configure as a relay with specified listening address (NetAddress)")
 	createCmd.Flags().StringVar(&listenIP, "api", "", "REST API Endpoint")
+	createCmd.Flags().BoolVar(&newNodeFullConfig, "full-config", false, "Store full config file")
```
winder, line 99: Should this default to true?

winder (approved):
LGTM

cmd/goal/bundledGenesisInject.go
algorandskiy: it is only for 1.16+ folks, not for us

cmd/goal/node.go
algorandskiy: I preserve the old behavior by default, mb some folks like it

cmd/goal/bundle_genesis_json.sh
algorandskiy: fixed

jasonpaulos (approved):

algorandskiy:
Closes #1686

tsachiherman (approved):
looks good.

PR #3640, opened 2022-02-16
ledger refactoring: rename NewAcct to Acct in StateDelta

algorandskiy:
## Summary

NewAccountDeltas and NewAccts were temporary names while old and new implementations co-existed.
No need to keep new naming anymore.

## Test Plan

Use existing tests


tsachiherman (approved):
looks great.

PR #3642, opened 2022-02-16
ledger refactoring: do not store block header in StateDelta

algorandskiy:
## Note for reviewers

This PR includes changes from #3640, so please consider merging #3640 first to get cleaner delta

## Summary

Instead of storing a full header in `StateDelta` store only required `round` and `rewardsLevel` fields.

## Test Plan

Existing tests.


tsachiherman:
@algorandskiy I believe that we should keep the block header in the state deltas, as we will need it when implementing the firstvalidroundtimestamp change in the ledger. We might be able to get rid of it afterward.

algorandskiy:
Ok, lets postpone this for a while then

algorandskiy:
I checked #2789 there is no mention of state deltas or cow.

PR #3636, opened 2022-02-15
ledger refactor: merge latest into feature branch

cce:
## Summary

Pulls in latest changes from master to the feature/unlimited-assets branch.

## Test Plan

Existing tests should pass.

PR #3643, opened 2022-02-16
add StateProofID to AccountData.ClearOnlineState()

cce:
## Summary

Clear out AccountData.StateProofID when ClearOnlineState() is called, by resetExpiredOnlineAccountsParticipationKeys (as part of #2924)

## Test Plan

Updated TestExpiredAccountGeneration to generate random online account data, so that it fails without this change.

### DIFF
--- a/ledger/internal/eval_test.go
+++ b/ledger/internal/eval_test.go
@@ -760,6 +760,8 @@ func TestEvalFunctionForExpiredAccounts(t *testing.T) {
 		}
 		tmp := genesisInitState.Accounts[addr]
 		tmp.Status = basics.Online
+		crypto.RandBytes(tmp.StateProofID[:])
+		crypto.RandBytes(tmp.SelectionID[:])
 		genesisInitState.Accounts[addr] = tmp
 	}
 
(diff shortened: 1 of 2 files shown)


tsachiherman (approved):
looks good.

tsachiherman (dismissed):
while I do agree with this change, we can't implement it like this.

tsachiherman:
I think that this should be implemented with a future flag. Then, we'll be able to merge it to master.

pzbitskiy:
I think we need a hotfix into betanet

algoidan:
I would suggest we move those fields into a new struct and clean the entire struct at once.
clearing one field at a time will lead to errors (as we have seen)

algorandskiy (approved):

cce:
@algoidan we've already moved the fields into a VotingData{} struct on the feature/unlimited-assets branch, and I just made the change you suggested to clear the entire struct here: https://github.com/algorand/go-algorand/pull/3636/commits/a5e867ffc4de1a70ad49d343ae8b58683ac4754b

ledger/internal/eval_test.go
```diff
@@ -759,6 +760,8 @@ func TestEvalFunctionForExpiredAccounts(t *testing.T) {
 		}
 		tmp := genesisInitState.Accounts[addr]
 		tmp.Status = basics.Online
+		crypto.RandBytes(tmp.StateProofID[:])
+		crypto.RandBytes(tmp.SelectionID[:])
```
algoidan, line 764: should we also randomize VoteID?

tsachiherman: I'm perfectly fine with randomizing the VoteID as well, but let's do that after the release ?
( i.e. since it's just a better testing )

algoidan: Gotcha !

algoidan (approved):

tsachiherman (approved):

PR #3645, opened 2022-02-17
CICD: Adding PR Type label checker

algojack:
We want to make PR Type labels required on PRs. This is a github action that will check for us if a correct label has been added"

Tested on this PR. Feel free to test it here too by adding a label or removing it.


.github/workflows/label-checker.yml
```diff
@@ -0,0 +1,18 @@
...
+  pull_request:
+    types: [opened, synchronize, reopened, labeled, unlabeled, edited]
+jobs:
+  check_label:
+    runs-on: ubuntu-latest
+    name: Check PR Type label is set
+    steps:
+      - name: "Failed to find proper PR Type label. Please add one of the following: 'New Feature', 'Enhancement', or 'Bug-Fix'"
+        if: |
+          !contains(github.event.pull_request.labels.*.name, 'New Feature') &&
```
algobarb, line 12: Shouldn't this be "or" statements?

algojack: if does NOT have blah and does NOT have blah and does NOT have blah. (Making sure all 3 labels are missing.

algobarb: I had to think about it a bit, but I think you're right since we're negating contains for the "when to fail"

algobarb:
I'm good with this if we add the label and category to this PR per specs.

yaovi-a:
Same comment as Barbara. But would we just check that there's some text with a colon before the PR change?

egieseke (commented):
Should it check for only one type label?
Can we add a regex to check the PR title?

algobarb (approved):
LGTM

PR #3652, opened 2022-02-17
Algod : merge unlimited assets to master

tsachiherman:
## Summary

This PR adds support of for unlimited assets.

Done:
- [x] Update this title
- [x] Complete & Merge the implementation of the accounts prefetched during the validation (https://github.com/algorand/go-algorand/pull/3666)
- [x] Create & Merge a spec update ( https://github.com/algorandfoundation/specs/pull/62#issuecomment-1047961971 )
- [x] Create & Merge a consensus upgrade (https://github.com/algorand/go-algorand/pull/3674)

## Test Plan

Extend, update and add new unit tests.

### DIFF
--- a/daemon/algod/api/server/v2/handlers.go
+++ b/daemon/algod/api/server/v2/handlers.go
@@ -480,7 +480,7 @@ func (v2 *Handlers) AccountApplicationInformation(ctx echo.Context, address stri
 	ledger := v2.Node.LedgerForAPI()
 
 	lastRound := ledger.Latest()
-	record, err := ledger.LookupResource(lastRound, addr, basics.CreatableIndex(applicationID), basics.AssetCreatable)
+	record, err := ledger.LookupResource(lastRound, addr, basics.CreatableIndex(applicationID), basics.AppCreatable)
 	if err != nil {
 		return internalError(ctx, err, errFailedLookingUpLedger, v2.Log)
 	}


daemon/algod/api/algod.oas2.json
```diff
@@ -173,7 +184,161 @@
...
+          ],
+          "type": "string",
+          "name": "format",
+          "in": "query"
+        }
+      ]
+    },
+    "/v2/accounts/{address}/applications/{application-id}": {
+      "get": {
+        "description": "Given a specific account public key, this call returns the account's application local state and application params (if they exist) about a given application ID",
```
jannotti, line 306: Perhaps this should be reworded to be clear that it would contain global state, not just "params" which I would have taken to mean the schemas.

data/basics/userBalance_test.go
```diff
@@ -190,8 +194,11 @@ func TestEncodedAccountDataSize(t *testing.T) {
 		}
 		maxLocalState[maxKey] = maxValue
 	}
-
-	for appCreatorApps := 0; appCreatorApps < currentConsensusParams.MaxAppsCreated; appCreatorApps++ {
+	maxAppsCreate := currentConsensusParams.MaxAppsCreated
+	if maxAppsCreate == 0 {
+		maxAppsCreate = config.Consensus[protocol.ConsensusV30].MaxAppsCreated
```
jannotti, line 199: It's unclear to me what this test actually does in a world with unlimited assets.  On the one hand, setting maxAppsCreate to the v30 level means we're lying - the return maximum size is wildly wrong (there is no limit to the size of marshaling basics.AccountData) On the other hand, I don't know why the test is checking the size here. Presumably we will never marshal one of these things again?

daemon/algod/api/spec/v2/model.go
```diff
@@ -0,0 +1,51 @@
...
+// Package v2 defines models exposed by algod rest api
+package v2
+
+import (
+	"github.com/algorand/go-algorand/data/basics"
+	"github.com/algorand/go-algorand/ledger/ledgercore"
+)
+
+// AccountResourceModel used to encode AccountResource
+type AccountResourceModel struct {
```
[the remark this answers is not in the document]
tsachiherman: I think that we *should* have a separate object for the model; that would be a better long-term approach if we want to retain backward compatibility. That being said, if you feel that we need a "trigger" to verify that these contains the same fields, it would be fine to create a unit test that uses reflection to verify that.

data/basics/userBalance_test.go
tsachiherman: This unit test is still there as a "backing" for the `MaxEncodedAccountDataSize` constant. This constant is currently used in the catchpoint files, such as `encodedBalanceRecordV5`. In the future, we're going to stop using this constant, as we plan to refactor the catchpoint file structure. But we still need to support the old files, and so this test need to stay "as-is".

jannotti: That makes sense. If it's true that when proto.MaxAppsCreated == 0 that the size of the encoded account does not grow at all (I think that's true), then I'd rather see this unit test leave maxAppsCreate == 0 when the proto value == 0, rather than set it to an old value from v30.  It's easier to understand that creating apps does not affect the size, and therefore isn't done in the test, than it is to understand why the v30 value is used.

Perhaps the test should be that the encoded size does not grow at all when apps are created if proto.MaxAppsCreated == 0.

cmd/tealdbg/localLedger.go
```diff
@@ -285,11 +285,36 @@ func (l *localLedger) CheckDup(config.ConsensusParams, basics.Round, basics.Roun
 	return nil
 }
 
-func (l *localLedger) LookupWithoutRewards(rnd basics.Round, addr basics.Address) (basics.AccountData, basics.Round, error) {
+func (l *localLedger) LookupResource(rnd basics.Round, addr basics.Address, aidx basics.CreatableIndex, ctype basics.CreatableType) (ledgercore.AccountResource, error) {
```
jannotti, line 288: I notice that nearly every "top-level" caller of LookupResource, knows the type it wants. I think it would make all these callers clearer if the Ledger interface offered `LookupAsset` and `LookupApp` instead, which would not need the extra argument, and would be typed more accurately (they would take the specific index type instead of basics.CreateableIndex.) They would also return a more specific type than `ledgercore.AccountResource`.

Are there callers that need to call LookupResource and they don't know what they want?

tsachiherman: I do like the idea, but I think that we should hold off with that type of changes due to their cascading consequences. Changing LookupResource to LookupAsset+LookupApp suggest that the return value would also need to be modified to be class-specific. This would have trailing effects that I'm not sure if we really want to explore at this time.
Let's create a PR for that which we could implement once this feature is shipped ?

daemon/algod/api/server/v2/account.go
```diff
@@ -128,10 +126,14 @@ func AccountDataToAccount(
...
 		Participation:               apiParticipation,
 		CreatedAssets:               &createdAssets,
+		TotalCreatedAssets:          uint64(len(createdAssets)),
 		CreatedApps:                 &createdApps,
+		TotalCreatedApps:            uint64(len(createdApps)),
 		Assets:                      &assets,
+		TotalAssets:                 uint64(len(assets)),
 		AuthAddr:                    addrOrNil(record.AuthAddr),
 		AppsLocalState:              &appsLocalState,
+		TotalAppsLocalState:         uint64(len(appsLocalState)),
```
jannotti, line 136: Why are we adding the Total* forms? I think it might be to let the caller calculate minimum balance requirements? If so, that seems unneeded since we give them that info. Further, I don't think it's enough, unless AppsTotalSchema included global schemas (for apps created by this account) _and_ local schemas (for apps this account has opted into) but I think maybe it's just the globals?

daemon/algod/api/server/v2/test/handlers_test.go
```diff
@@ -195,7 +196,7 @@ func TestGetBlockJsonEncoding(t *testing.T) {
 	blk.TxnRoot, err = blk.PaysetCommit()
 	require.NoError(t, err)
 
-	err = l.AddBlock(blk, agreement.Certificate{})
+	err = l.(*data.Ledger).AddBlock(blk, agreement.Certificate{})
```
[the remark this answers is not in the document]
tsachiherman: The API shouldn't require that.. it's only for testing ( and maybe it should be mocked out differently )

daemon/algod/api/server/v2/account.go
tsachiherman: The caller can use this information to figure out how many resources are associated with this object.

daemon/algod/api/server/v2/test/handlers_test.go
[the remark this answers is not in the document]
jannotti: Yes, sorry, I missed `_test` in the file name.  I'm fine with this sort of thing in tests, no need for more mocks.

daemon/algod/api/server/v2/account.go
jannotti: Maybe `TotalAssets` -> `TotalAssetsOptedIn` and `TotalLocalStates` -> `TotalAppsOptedIn`?

ledger/accountdb.go
```diff
@@ -633,43 +1055,741 @@ func accountsAddNormalizedBalance(tx *sql.Tx, proto config.ConsensusParams) erro
...
+	}
+}
+
+func accountDataResources(
+	ctx context.Context,
+	accountData *basics.AccountData, rowid int64,
+	outputResourceCb func(ctx context.Context, rowid int64, cidx basics.CreatableIndex, rd *resourcesData) error,
+) error {
+	// does this account have any assets ?
+	if len(accountData.Assets) > 0 || len(accountData.AssetParams) > 0 {
```
jannotti: Why check these lengths, since you're about to range over them, skipping them if len == 0 anyway?

(Same question below for Apps)

Makefile
```diff
@@ -80,7 +80,7 @@ GOLDFLAGS := $(GOLDFLAGS_BASE) \
 UNIT_TEST_SOURCES := $(sort $(shell GOPATH=$(GOPATH) && GO111MODULE=off && go list ./... | grep -v /go-algorand/test/ ))
 ALGOD_API_PACKAGES := $(sort $(shell GOPATH=$(GOPATH) && GO111MODULE=off && cd daemon/algod/api; go list ./... ))
 
-MSGP_GENERATE	:= ./protocol ./protocol/test ./crypto ./crypto/merklearray ./crypto/merklesignature ./crypto/compactcert ./data/basics ./data/transactions ./data/committee ./data/bookkeeping ./data/hashable ./agreement ./rpcs ./node ./ledger ./ledger/ledgercore ./compactcert ./data/account
+MSGP_GENERATE	:= ./protocol ./protocol/test ./crypto ./crypto/merklearray ./crypto/merklesignature ./crypto/compactcert ./data/basics ./data/transactions ./data/committee ./data/bookkeeping ./data/hashable ./agreement ./rpcs ./node ./ledger ./ledger/ledgercore ./compactcert ./data/account ./daemon/algod/api/spec/v2
```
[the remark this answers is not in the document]
cce: There is some goal code that handles TEAL app local/global KVs that was relying on the msgpack'd KV types in REST API for LookupAccount /v2/accounts/{address} ... my thinking was to minimize potential changes and not have to worry about introducing an issue by converting to/from JSON in those goal commands, the LookupResource should also return a msgpack type. But that is a pretty narrow use case, maybe it is not a big deal?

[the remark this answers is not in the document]
cce: I also wasn't sure if there are SDK users that currently rely on handling msgpack versions of app/asset resource types.

ledger/accountdb.go
```diff
@@ -860,7 +2071,8 @@ func (qs *accountsDbQueries) lookup(addr basics.Address) (data persistedAccountD
 			data.addr = addr
 			if len(buf) > 0 && rowid.Valid {
 				data.rowid = rowid.Int64
-				return protocol.Decode(buf, &data.accountData)
+				err = protocol.Decode(buf, &data.accountData)
```
jannotti: It seems like either this change is unneeded (because directly returning the error will cause it to be propagated to the outer `err` variable by `db.Retry`, or if this change IS needed (because that doesn't happen), then a similar change is needed on line 2084.

My guess is that returning the error directly is ok. But I'm pretty sure they should be done the same way.

ledger/accountdb.go
```diff
@@ -303,54 +701,34 @@ func (a *compactAccountDeltas) len() int {
...
-
-// upsert updates existing or inserts a new entry
-func (a *compactAccountDeltas) upsert(addr basics.Address, delta accountDelta) {
-	if idx, exist := a.cache[addr]; exist { // nil map lookup is OK
-		a.deltas[idx] = delta
-		return
-	}
-	a.insert(addr, delta)
+func (a *compactAccountDeltas) getByIdx(i int) accountDelta {
+	return a.deltas[i]
```
jannotti: Now that this function (and `compactResourceDeltas.getByIdx`) are so simple, I think calling code would be clearer if it just said `a.deltas[i]` and these functions removed.  By using this function, which doesn't have the word "delta" in it, the meaning is obscured.

ledger/accountdb.go
```diff
@@ -1075,72 +2290,376 @@ func accountsPutTotals(tx *sql.Tx, totals ledgercore.AccountTotals, catchpointSt
...
-						err = fmt.Errorf("failed to delete accountbase row for account %v, rowid %d", addr, data.old.rowid)
-					}
-				}
+				// this case was already handled in the first loop.
+				continue
 			} else {
-				normBalance := data.new.NormalizedOnlineBalance(proto)
-				result, err = updateStmt.Exec(normBalance, protocol.Encode(&data.new), data.old.rowid)
+				if !data.newResource.IsApp() && !data.newResource.IsAsset() {
+					err = fmt.Errorf("unknown creatable for addr %s (%d), aidx %d, data %v", addr.String(), addrid, aidx, data.newResource)
```
jannotti: no need to invoke `addr.String()`, %s will do it.

daemon/algod/api/server/v2/handlers.go
```diff
@@ -56,9 +58,25 @@ type Handlers struct {
 	Shutdown <-chan struct{}
 }
 
+// LedgerForAPI describes the Ledger methods used by the v2 API.
+type LedgerForAPI interface {
```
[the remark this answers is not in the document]
cce: EncodedBlockCert is a transitive dependency because v2 API calls rpcs.RawBlockBytes and passes it a ledger that needs to support EncodedBlockCert.. agreed that pkgname.LedgerForPkgName is kind of redundant, I would also be happy to name the interfaces v2.Ledger and BlockService.Ledger in keeping with https://github.com/golang/go/wiki/CodeReviewComments#interfaces

daemon/algod/api/spec/v2/model.go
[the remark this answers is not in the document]
cce: I was trying to avoid more response types, but it sounds like it is better to have more response types.

ledger/apply/application.go
```diff
@@ -220,21 +228,30 @@ func optInApplication(balances Balances, sender basics.Address, appIdx basics.Ap
 		return err
 	}
 
+	if !balances.ConsensusParams().Application {
+		return fmt.Errorf("cannot opt in app %d for %s: applications not supported", appIdx, sender.String())
+	}
+
```
jannotti, line 234: We seem awfully deep to be checking this. Surely we rejected the appl transaction if the current proto doesn't support it? (Also, no need to call .String())

ledger/apply/application_test.go
```diff
@@ -768,18 +839,22 @@ func TestAppCallClearState(t *testing.T) {
 	b.pass = true
 	// check app not exist and not opted in
 	b.balances[sender] = basics.AccountData{}
-	err := ApplicationCall(ac, h, &b, ad, 0, &ep, txnCounter)
+	err := ApplicationCall(ac, h, b, ad, 0, &ep, txnCounter)
 	a.Error(err)
 	a.Contains(err.Error(), "is not currently opted in to app")
 	a.Equal(0, b.put)
+	a.Equal(0, b.putAppLocalState)
+	a.Equal(0, b.deleteAppLocalState)
```
jannotti, line 847: I certainly don't care whether you do this or not, but I just learned `a.Zero(b.put)` exists, if you're interested.

ledger/apply/asset.go
```diff
@@ -308,39 +337,51 @@ func AssetTransfer(ct transactions.AssetTransferTxnFields, header transactions.H
...
 			return fmt.Errorf("cannot close asset ID in allocating account")
 		}
 
 		// Fetch our asset holding, which should exist since we're
 		// closing it out
-		sndHolding, ok := snd.Assets[ct.XferAsset]
+		sndHolding, ok, err := balances.GetAssetHolding(source, ct.XferAsset)
+		if err != nil {
+			return err
+		}
```
jannotti, line 369: It seems like line 356 _ought_ to make line _366_ fast. But I don't think it does. From what I see, looking up a params does not put the holding info into a cache. Am I missing it?

ledger/apply/asset.go
```diff
@@ -308,39 +337,51 @@ func AssetTransfer(ct transactions.AssetTransferTxnFields, header transactions.H
...
 		if !ok {
 			return fmt.Errorf("asset %v not present in account %v", ct.XferAsset, source)
 		}
 
-		// Fetch the destination balance record to check if we are
+		// Fetch the destination asset params to check if we are
 		// closing out to the creator
-		dst, err := balances.Get(ct.AssetCloseTo, false)
+		dstAssetParamsExist, err := balances.HasAssetParams(ct.AssetCloseTo, ct.XferAsset)
```
jannotti, line 376: This makes me think balances.HasAssetParams() should have an optimization for asking about the zero address.  That will be the most common case by far, and the answer is always False.

ledger/internal/cow.go
```diff
@@ -38,7 +38,12 @@ import (
 //                  ||     ||
 
 type roundCowParent interface {
-	lookup(basics.Address) (basics.AccountData, error)
+	lookup(basics.Address) (ledgercore.AccountData, error)
+	lookupAppParams(addr basics.Address, aidx basics.AppIndex, fromCache bool) (ledgercore.AppParamsDelta, bool, error)
+	lookupAssetParams(addr basics.Address, aidx basics.AssetIndex, fromCache bool) (ledgercore.AssetParamsDelta, bool, error)
+	lookupAppLocalState(addr basics.Address, aidx basics.AppIndex, fromCache bool) (ledgercore.AppLocalStateDelta, bool, error)
+	lookupAssetHolding(addr basics.Address, aidx basics.AssetIndex, fromCache bool) (ledgercore.AssetHoldingDelta, bool, error)
+
```
jannotti, line 46: Two things about `fromCache`. 1) `cacheOnly` would be a clearer name, since I might imagine that fromCache=false means "Do no look in cache". But it seems to mean that it's ok to look beyond the cache. 2) This would be a good place for a comment explaining the intuition for why you might use this flag.  I have one idea, but it's only a guess.

I'm not sure my idea is right, because I now realize this means "Don't look in my parent". But the parent of a roundCow might be another roundCowState, which is in memory and fast, right?  It doesn't mean what I thought it meant - don't go to disk.

ledger/internal/eval.go
```diff
@@ -41,13 +41,17 @@ import (
...
+	LookupWithoutRewards(basics.Round, basics.Address) (ledgercore.AccountData, basics.Round, error)
+	LookupResource(basics.Round, basics.Address, basics.CreatableIndex, basics.CreatableType) (ledgercore.AccountResource, error)
 	GetCreatorForRound(basics.Round, basics.CreatableIndex, basics.CreatableType) (basics.Address, bool, error)
 }
 
 // ErrRoundZero is self-explanatory
 var ErrRoundZero = errors.New("cannot start evaluator for round 0")
 
+// ErrNotInCowCache is returned when a lookup method requests a cached value, but it can't be found
+var ErrNotInCowCache = errors.New("can't find object in roundCowBase")
```
jannotti, line 53: Isn't the "roundCowBase" the parent interface? So that's the place that wasn't looked into at all, if only the cache was examined.

ledger/internal/eval.go
```diff
@@ -1217,7 +1428,7 @@ func (eval *BlockEvaluator) resetExpiredOnlineAccountsParticipationKeys() error
 		acctData.ClearOnlineState()
```
algoidan, line 1428: looks like we have two `ClearOnlineState` function one for `basic.AccountData` and the other for `ledgercore.AccountData` and the first is not referenced  if I understand correctly

ledger/accountdb.go
```diff
@@ -360,8 +738,16 @@ func writeCatchpointStagingBalances(ctx context.Context, tx *sql.Tx, bals []norm
...
+	insertRscStmt, err = tx.PrepareContext(ctx, "INSERT INTO catchpointresources(addrid, aidx, data) VALUES(?, ?, ?)")
+	if err != nil {
+		return err
+	}
+
+	var result sql.Result
+	var rowID int64
 	for _, balance := range bals {
-		result, err := insertAcctStmt.ExecContext(ctx, balance.address[:], balance.normalizedBalance, balance.encodedAccountData)
+		result, err = insertAcctStmt.ExecContext(ctx, balance.address[:], balance.normalizedBalance, balance.encodedAccountData)
```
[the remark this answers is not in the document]
tsachiherman: The trigger for that was the variable shadowing, which we try to avoid. In this case, it's not obvious that the `err` variable here is the *same* one as defined above.
Second motivation is to avoid unneeded variable cleanup on each iteration of the loop. These variable are always being written to, and therefore doesn't require to be re-initialized ( admittedly, the compiler might be able to optimize for that.. in some cases ).

ledger/apply/application_test.go
tsachiherman: sure. The same applies to many other of our tests files. I'd rather avoid making a mass-change here, since it's not really related to this PR. But I've fixed all the cases on this file.

ledger/apply/asset.go
tsachiherman: hmm. let's trace it:
the call to `balances.HasAssetParams` calls `roundCowState.HasAssetParams`, which calls `roundCowState.lookupAssetParams`, which calls `roundCowBase.lookupAssetParams` which calls `roundCowBase.updateAssetResourceCache`.

did I miss something ?

ledger/internal/eval.go
tsachiherman: yes - my bad, a merge mistake..

ledger/apply/asset.go
jannotti: Yes, I think I figured it out eventually.  I was expecting a cache at each cow level, but it's only in the base.  And that seems like a good idea now that I understand it.

ledger/apply/application.go
tsachiherman: agree. I'll remove this check both from `createApplication` as well as from `optInApplication`.

tsachiherman: updated.

daemon/algod/api/server/v2/account.go
tsachiherman: @cce making this change would require to change the oas file. Could you take a look to figure out its implications ?

ledger/accountdb.go
tsachiherman: this change allow you to place a breakpoint on the `return err` so you could observe the generated error. looking on your comment, I was worried about shadowed variables, but that was never the case here.

ledger/accountdb.go
```diff
@@ -633,43 +1055,741 @@ func accountsAddNormalizedBalance(tx *sql.Tx, proto config.ConsensusParams) erro
...
+		if err = rows.Err(); err != nil {
+			return 0, nil, err
+		}
+	}
+
+	result, err := tx.Exec("DELETE from accountbase where length(data) = 1 and data = x'80'")
+	if err != nil {
+		return 0, nil, err
+	}
+	num, err = result.RowsAffected()
```
algoidan: if we ignore the affected result, why not skip this check?

ledger/accountdb.go
```diff
@@ -682,7 +1802,7 @@ func removeEmptyAccountData(tx *sql.Tx, queryAddresses bool) (num int64, address
 // the full AccountData because we need to store a large number of these
 // in memory (say, 1M), and storing that many AccountData could easily
 // cause us to run out of memory.
-func accountDataToOnline(address basics.Address, ad *basics.AccountData, proto config.ConsensusParams) *ledgercore.OnlineAccount {
+func accountDataToOnline(address basics.Address, ad ledgercore.AccountData, proto config.ConsensusParams) *ledgercore.OnlineAccount {
```
algoidan: why did we discard the pointer?

ledger/accountdb.go
tsachiherman: that's probably a good point, but I'd rather leaving it as is for now. This code is part of the upgradeDatabaseSchema4, and it wasn't causing any issue I'm aware of. If we were to change this, I would want to have some testing around this change - which I'd like to avoid for now..

ledger/accountdb.go
tsachiherman: In this case the pointer is meaningless, since this method is going to get inlined, and therefore the stack representation isn't important anymore.

algoidan: Can we force the compiler to inline it? I'm sure that there is a way to do it
if we can't force it, I think we should pass this parameter as a pointer so in case it is not inlined we will not copy the structure 

BTW
I've tried to use `go tool nm ~/go/bin/algod | grep accountDataToOnline` and it looks like the compiler did create a function for this (I think it means that it is not inlined)

algorandskiy: maybe because it returns a pointer? I've seen an article about a vscode extension for golang that parses the extended compiler output and shows gc some memory allocation and inlining decision made by the compiler, but can't find the article :(

algoidan: that's is why I think it is a bit risky to count on the compiler to optimize it.
looks like it didn't do that 
we can actually use `go tool objdump -s github.com/algorand/go-algorand/ledger.accountDataToOnline ~/go/bin/algod
`
and see the actual implementation of the function in assmbly

daemon/algod/api/server/v2/account.go
cce: OK from discussion just now with @yaovi-a and @tsachiherman it is more clear to call them TotalAssetsOptedIn and TotalAppsOptedIn, plus similar names in the REST API fields. will update ASAP @tolikzinovyev @nicholasguoalgorand

jannotti:
I don't think that's a great way to be sure either, though it may be true
in this case. For example, there could be other calls to the function that
are not inlined.  If we care enough that we want to be sure, I think we
have to parse -gcflags=-m

But stepping up a level, `ledgercore.AccountData` is the new, small,
struct, right?  The old basics.AccountData was big. I think it's
appropriate to pass small (maybe 32-64 bytes?) by value - I have been
surprised to see that was faster than using pointers in some of my
benchmarks.

algoidan:
Maybe I miss something. why not just use a pointer.
then the inline discussion is not relevant

daemon/algod/api/server/v2/account.go
cce: Updated in https://github.com/algorand/go-algorand/pull/3676

ledger/accountdb.go
tsachiherman: ok, merged changes.

ledger/internal/eval_blackbox_test.go
```diff
@@ -211,16 +211,16 @@ func TestBlockEvaluator(t *testing.T) {
...
 	l.AddValidatedBlock(*validatedBlock, agreement.Certificate{})
 
-	bal0new, err := l.Lookup(newBlock.Round(), addrs[0])
+	bal0new, err := l.LookupAgreement(newBlock.Round(), addrs[0])
 	require.NoError(t, err)
-	bal1new, err := l.Lookup(newBlock.Round(), addrs[1])
+	bal1new, err := l.LookupAgreement(newBlock.Round(), addrs[1])
 	require.NoError(t, err)
-	bal2new, err := l.Lookup(newBlock.Round(), addrs[2])
+	bal2new, err := l.LookupAgreement(newBlock.Round(), addrs[2])
```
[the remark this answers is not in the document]
tsachiherman: fixed.

ledger/apply/asset.go
tsachiherman: since this is a non-issue, I'm going to resolve this one. Please let me know if I missed anything.

ledger/internal/eval.go
tsachiherman: `roundCowBase` is the last level being tested before hitting the actual ledger. and the code seems to be correct here. i.e. I'm looking at roundCowBase.lookupAssetHolding; maybe it's not the case in other places ?
I'm not opposing to change this into `can't find object in cow cache`, if that make it any better though.

jannotti: I think it makes the error message make sense, even if propagated up through roundCowState lookups. For example:
```
func (cb *roundCowState) lookupAssetParams(addr basics.Address, aidx basics.AssetIndex, fromCache bool) (ledgercore.AssetParamsDelta, bool, error) {
	params, ok := cb.mods.Accts.GetAssetParams(addr, aidx)
	if ok {
		return params, ok, nil
	}

	return cb.lookupParent.lookupAssetParams(addr, aidx, fromCache)
}
```
As written, this function says it was unable to find the asset in "roundCowBase" but the reality is that it was not able to find it in the roundCowState OR the roundCowBase.

ledger/internal/cow.go
cce: updated with:
```
    // lookup retrieves data about an address, eventually querying the ledger if the address was not found in cache.
    lookup(basics.Address) (ledgercore.AccountData, error)

    // lookupAppParams, lookupAssetParams, lookupAppLocalState, and lookupAssetHolding retrieve data for a given address and ID.
    // If cacheOnly is set, the ledger DB will not be queried, and only the cache will be consulted.
    // This is used when we know a given value is already in cache (from a previous query for that same address and ID),
    // and would rather have an error returned if that assumption is wrong, rather than hit the ledger.
    lookupAppParams(addr basics.Address, aidx basics.AppIndex, cacheOnly bool) (ledgercore.AppParamsDelta, bool, error)
    lookupAssetParams(addr basics.Address, aidx basics.AssetIndex, cacheOnly bool) (ledgercore.AssetParamsDelta, bool, error)
    lookupAppLocalState(addr basics.Address, aidx basics.AppIndex, cacheOnly bool) (ledgercore.AppLocalStateDelta, bool, error)
    lookupAssetHolding(addr basics.Address, aidx basics.AssetIndex, cacheOnly bool) (ledgercore.AssetHoldingDelta, bool, error)
```

ledger/internal/eval.go
tsachiherman: oh - I see what you mean. You're saying that the error itself is good as long it is was limited to `roundCowBase`. however, once it's propagated upstream, the scope of that error is getting wider, and it won't align with the original error message anymore.

Given that the gains of detecting the error message, and modifying it is pretty much non-existing, I'll follow your suggestion and make this change with some extra documentation for that use case.

cce: yes I guess roundCowBase is the end of the line just before it hits the DB, but a cached result could have been found in a cache in one of the "mods" caches in roundCowState or the cache in roundCowBase.. updating..

ledger/apply/asset.go
```diff
@@ -308,39 +337,51 @@ func AssetTransfer(ct transactions.AssetTransferTxnFields, header transactions.H
...
-		// figure out how much of the asset to move.
-		snd, err := balances.Get(source, false)
+		record, err := balances.Get(source, false)
 		if err != nil {
 			return err
 		}
 
+		if record.TotalAssets == 0 {
+			return fmt.Errorf("cannot close asset holding on account %s : account is not opted in asset %d ", source.String(), ct.XferAsset)
+		}
```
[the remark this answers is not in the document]
cce: this check saves a potential DB query against the resources table later, for the rare close transaction where an address has no assets.

daemon/algod/api/spec/v2/model.go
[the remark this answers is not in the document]
cce: Addressing in https://github.com/algorand/go-algorand/pull/3679

daemon/algod/api/algod.oas2.json
cce: Updated in https://github.com/algorand/go-algorand/pull/3679

daemon/algod/api/server/v2/test/handlers_resources_test.go
```diff
@@ -0,0 +1,181 @@
...
+	err := handlers.AccountInformation(ctx, addr.String(), params)
+	require.NoError(t, err)
+	require.Equal(t, expectedCode, rec.Code)
+
+	var ret map[string]interface{}
+	err = json.Unmarshal(rec.Body.Bytes(), &ret)
+	require.NoError(t, err)
+
+	// totals should be present in both 200 and 400
+	require.Equal(t, float64(acctSize), ret["total-apps-opted-in"].(float64)+ret["total-assets-opted-in"].(float64)+ret["total-created-apps"].(float64)+ret["total-created-assets"].(float64))
```
jasonpaulos, line 152: If float comparison proves flaky, you could unmarshal the JSON into a temporary struct with these 4 fields as uint64s

config/localTemplate.go
```diff
@@ -430,6 +430,10 @@ type Local struct {
 	// The http server does not accept new connections as long we have this many
 	// (hard limit) connections already.
 	RestConnectionsHardLimit uint64 `version[20]:"2048"`
+
+	// MaxAccountsAPIResults limits the total number of assets and applications per account that will be
+	// provided by the REST API before returning a 413 Payload Too Large. Set zero for no limit.
+	MaxAccountsAPIResults uint64 `version[21]:"1000000"`
```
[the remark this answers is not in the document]
cce: Oops that is out of date. Should be 400, will update

ledger/tracker.go
```diff
@@ -517,7 +525,7 @@ func (tr *trackerRegistry) initializeTrackerCaches(l ledgerForTracker) (err erro
 	if lastBalancesRound < lastestBlockRound {
 		accLedgerEval.prevHeader, err = l.BlockHdr(lastBalancesRound)
 		if err != nil {
-			return err
+			return fmt.Errorf("unalbe to load block header %d : %w", lastBalancesRound, err)
```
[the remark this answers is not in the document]
cce: fixed

ledger/persistedresources_list.go
```diff
@@ -0,0 +1,143 @@
...
+	l := new(persistedResourcesDataList)
+	l.root.next = &l.root
+	l.root.prev = &l.root
+	// used as a helper but does not store value
+	l.freeList = new(persistedResourcesDataListNode)
+
+	return l
+}
+
+func (l *persistedResourcesDataList) inserNodeToFreeList(otherNode *persistedResourcesDataListNode) {
```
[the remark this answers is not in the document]
cce: updated

daemon/algod/api/server/v2/test/handlers_resources_test.go
cce: With such a low number it doesn't seem to be flaky, but there is require.InDelta too for float-precision-tolerant comparisons

ledger/ledgercore/statedelta.go
```diff
@@ -90,112 +95,298 @@ type StateDelta struct {
...
+	Aidx    basics.AssetIndex
+	Addr    basics.Address
+	Params  AssetParamsDelta
+	Holding AssetHoldingDelta
+}
+
 // AccountDeltas stores ordered accounts and allows fast lookup by address
 type AccountDeltas struct {
 	// Actual data. If an account is deleted, `accts` contains a balance record
 	// with empty `AccountData`.
```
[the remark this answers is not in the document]
tsachiherman: I'll clarify that.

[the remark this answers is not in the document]
tsachiherman: ok, updated. Please re-open this if you'd like feel that better clarification is needed.

ledger/ledgercore/accountdata.go
```diff
@@ -0,0 +1,171 @@
...
+type VotingData struct {
+	VoteID       crypto.OneTimeSignatureVerifier
+	SelectionID  crypto.VRFVerifier
+	StateProofID merklesignature.Verifier
+
+	VoteFirstValid  basics.Round
+	VoteLastValid   basics.Round
+	VoteKeyDilution uint64
+
+	// MicroAlgosWithReward basics.MicroAlgos
```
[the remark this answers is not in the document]
tsachiherman: Yes - the sibsequent ledger refactoring project would touch exactly that point, so I'd rather leaving it there for one more montg..

cmd/tealdbg/localLedger.go
tsachiherman: created a follow up issue to work on this issue.

ledger/accountdb.go
tsachiherman: ok, removed.

data/basics/userBalance.go
```diff
@@ -437,23 +437,35 @@ func PendingRewards(ot *OverflowTracker, proto config.ConsensusParams, microAlgo
 	return MicroAlgos{Raw: ot.Mul(rewardsUnits, rewardsDelta)}
 }
 
-// WithUpdatedRewards returns an updated number of algos in an AccountData
+// WithUpdatedRewards returns an updated number of algos, total rewards and new rewards base
 // to reflect rewards up to some rewards level.
-func (u AccountData) WithUpdatedRewards(proto config.ConsensusParams, rewardsLevel uint64) AccountData {
-	if u.Status != NotParticipating {
+func WithUpdatedRewards(
```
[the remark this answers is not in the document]
tsachiherman: I agree that the naming choice for this method is no longer correct, and it should be updated. I think that we can come up with something better the `UpdateRewards`, since the method doesn't really do any updating. But I'm going to defer that to a subsequent PR ( just because it's not really breaking anything as is ).

daemon/algod/api/algod.oas2.json
```diff
@@ -1600,10 +1784,6 @@
           "type": "integer",
           "x-go-name": "AssetID"
         },
-        "creator": {
-          "description": "Address that created this asset. This is the address where the parameters for this asset can be found, and also the address where unwanted asset units can be sent in the worst case.",
```
algonautshant, line 1604: Will this impact sdk backwards compatibility? Please comment.

jasonpaulos: As is, yes it will. I'd like to suggest we manually retain this field in the SDK models so they don't have an explicitly breaking change. The field can just be empty if algod doesn't return it.

cce: IMO it would be preferable to introduce a breaking change than provide an empty value. That way you would find out  that you need to pull the creator from /v2/assets/{asset-id} instead. However I don't have a lot of experience with how we deprecate functionality in the SDKs or warn users of issues like this, whether a compiler error or compiler warning is more typical.

ledger/accountdb.go
```diff
@@ -633,43 +1055,739 @@ func accountsAddNormalizedBalance(tx *sql.Tx, proto config.ConsensusParams) erro
...
+	// UpdateRound is the round that modified this account data last. Since we want all the nodes to have the exact same
+	// value for this field, we'll be setting the value of this field to zero *before* the EnableAccountDataResourceSeparation
+	// consensus parameter is being set. Once the above consensus takes place, this field would be populated with the
+	// correct round number.
+	UpdateRound uint64 `codec:"z"`
+}
+
+// IsEmpty return true if any of the fields other then the UpdateRound are non-zero.
+func (ba *baseAccountData) IsEmpty() bool {
+	return ba.Status == 0 &&
```
algoidan, line 1116: why not use `return *ba == baseAccountData{}`

ledger/accountdb.go
```diff
@@ -1075,72 +2288,376 @@ func accountsPutTotals(tx *sql.Tx, totals ledgercore.AccountTotals, catchpointSt
...
+	close()
+}
+
+type accountsSQLWriter struct {
+	insertCreatableIdxStmt, deleteCreatableIdxStmt             *sql.Stmt
+	deleteByRowIDStmt, insertStmt, updateStmt                  *sql.Stmt
+	deleteResourceStmt, insertResourceStmt, updateResourceStmt *sql.Stmt
+}
+
+func (w *accountsSQLWriter) close() {
```
algoidan, line 2312: why not use the same pattern fro closing the statements as `func (qs *accountsDbQueries) close()`

daemon/algod/api/algod.oas2.json
cce: Related: https://github.com/algorand/indexer/pull/900#discussion_r815128273

cmd/goal/account.go
```diff
@@ -519,19 +521,19 @@ var infoCmd = &cobra.Command{
 	Run: func(cmd *cobra.Command, args []string) {
 		dataDir := ensureSingleDataDir()
 		client := ensureAlgodClient(dataDir)
-		response, err := client.AccountInformationV2(accountAddress)
+		response, err := client.AccountInformationV2(accountAddress, true)
 		if err != nil {
 			reportErrorf(errorRequestFail, err)
```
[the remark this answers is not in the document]
cce: I just gave a try running the command against mainnet using a address I spotted on algoexplorer's latest transactions that had some assets:
```
$ goal account info -d mainnet-2022-02-25.node/ -a XXX
Error processing command: HTTP 400 Bad Request: {"max-results":10,"message":"Result limit exceeded","total-apps-opted-in":15,"total-assets-opted-in":38,"total-created-apps":0,"total-created-assets":15}
```

ledger/accountdb.go
algoidan: it could also be used for the set and the get.
but as long as we don't "miss" something. it is good

PR #3653, opened 2022-02-17
TealDbg: Support for StepOver and refactoring object IDs

### LINKED ISSUE #2289: Update tealdbg for TEAL 4
## Summary
TEAL v4 now allows looping, function calls, and has some additional opcodes. We need to make sure the debugger is able to understand these new changes.

## Scope/Requirements
Test that the debugger can handle the new feature from TEAL 4. I suspect the following may require debugger changes:
* Changes to foreign app/asset/account indexing
* Function calls & backward branches

If any changes are required, make them.

## Urgency/Relative Priority
Required to ensure TEAL 4 programs can be debugged properly.

algochoi:
## Summary

This PR contains:
* Support for stepping over subroutines when there is a `callsub` op. Dynamically inspects the callstack to make sure depth is equal to what it entered as. Also added a channel so that the updates can be synchronized within the debugger control (else updates may be received at any time and the state may not be updated before the depth check happens).
* Refactors object IDs in a separate file. We may want to map out the actual addresses to objects in the future.
* Adds a unit test for `RunAll` in tealdbg to ensure proper group txn execution.
* Include and expose the call stack in CDT. 

This should also close #2289 

## Test Plan

* Unit tests in `local_test.go` and `debugger_test.go`
* Ran tealdbg on contracts with subroutines. Example [dryrun here](https://github.com/algochoi/algorand-onboard/tree/tealdbg-tests/tealdbg-tutorial/generated-data#:~:text=dryrun_subroutine.msgp), and can be run with `./tealdbg debug -d path-to-file/generated-data/dryrun_subroutine.msgp`

`gload` seems to work in tealdbg if the entire txn group is supplied. e.g. through dryrun: `tealdbg debug -d dryrun_grp.msgp`.


algorandskiy (commented):
Looks good, couple comments

cmd/tealdbg/debugger.go
```diff
@@ -187,6 +188,28 @@ func (s *session) Step() {
...
+	currentLine := s.line.Load()
+	// Get the first TEAL opcode in the line
+	currentOp := strings.Fields(s.lines[currentLine])[0]
+
+	func() {
+		s.mu.Lock()
+		defer s.mu.Unlock()
+		// Step over a function call (callsub op).
+		if currentOp == "callsub" {
+			err := s.setBreakpoint(currentLine + 1)
```
jasonpaulos, line 201: Is this condition enough to support step over for recursive subroutine calls? I suspect not, since if you only wait for the next invocation of `currentLine + 1`, that will happen before the recursive callsub returns -- you'll be at the right place in the program code, but too deep in the subroutine call stack.

algorandskiy: Jason is right. The implementation also misses callstack feature in CDT. Maintaining internal callstack would allow 1) display the callstack in CDT 2) correctly process breakpoint only at required call depth.

cmd/tealdbg/local_test.go
```diff
@@ -1452,3 +1452,404 @@ func TestGroupTxnIdx(t *testing.T) {
 	a.NoError(err)
 	a.True(pass)
 }
+
+func TestRunAllGloads(t *testing.T) {
+
+	partitiontest.PartitionTest(t)
+	a := require.New(t)
+
+	// Transaction group with 5 transactions
```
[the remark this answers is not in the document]
algochoi: Good suggestion - done

cmd/tealdbg/debugger.go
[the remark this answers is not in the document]
algochoi: I think the only time it would error would be if `callsub` was at the end of the program and `setBreakpoint()` will try to set a breakpoint past the program. In that case, it feels normal to not log something and just continue, but maybe I'm missing a case.

cmd/tealdbg/local.go
```diff
@@ -535,13 +535,14 @@ func (r *LocalRunner) RunAll() error {
...
 
 	ep := logic.NewEvalParams(txngroup, &r.proto, &transactions.SpecialAddresses{})
-	ep.Debugger = r.debugger
 
 	var last error
 	for i := range r.runs {
 		run := &r.runs[i]
-		r.debugger.SaveProgram(run.name, run.program, run.source, run.offsetToLine, run.states)
-
+		if r.debugger != nil {
```
algochoi, line 542: Changes related to C2C - `RunAll()` doesn't cooperate without this check.

data/transactions/logic/debugger.go
```diff
@@ -192,10 +200,32 @@ func valueDeltaToValueDelta(vd *basics.ValueDelta) basics.ValueDelta {
...
 }
 
+// parseCallStack initializes an array of CallFrame objects from the raw
+// callstack.
+func (d *DebugState) parseCallstack(callstack []int) []CallFrame {
+	callFrames := make([]CallFrame, 0)
+	lines := strings.Split(d.Disassembly, "\n")
+	for _, pc := range callstack {
+		// The callsub is pc - 3 from the callstack pc
+		callsubLineNum := d.PCToLine(pc - 3)
```
algochoi, line 210: I did a bit of hardcoding here (`callsub` saves `pc+3` to the callstack in `eval.go`) and added a "soft" check to see if the line is a `callsub` later on so we can try to parse the label.

data/transactions/logic/debugger.go
```diff
@@ -192,10 +200,32 @@ func valueDeltaToValueDelta(vd *basics.ValueDelta) basics.ValueDelta {
...
+		callSubLine := strings.Fields(lines[callsubLineNum])
+		label := ""
+		if callSubLine[0] == "callsub" {
```
michaeldiamant, line 213: @algochoi As I understand it, to reach the highlighted line, a program must successfully assemble.  So, _in principle_, the highlighted parsing always succeeds.  

Nevertheless, I'm wondering:  Is there a way to avoid the error case or otherwise handle the error?

I performed some review before asking though I suspect I don't fully understand the space I'm asking about.  Part of the reason for asking is to further my understanding.

From what I saw, it seems avoiding the error case isn't possible without serious effort.  If true, I'm wondering if there's conventions for signaling a failed invariant to the end user.

algochoi: > So, in principle, the highlighted parsing always succeeds.

I think so too - it should always succeed unless we change the `callsub` semantics somehow. Currently `callsub` just pushes `pc+3` to the stack so I just hardcoded `pc-3` to mean the `callsub` line. If this change and this line isn't updated things might get scary afterwards, but I'm guessing existing opcode changes would be almost never done or done very carefully. 

> Nevertheless, I'm wondering: Is there a way to avoid the error case or otherwise handle the error?

If the current line is not a `callsub` then I just leave the `label` blank. I don't think it's a huge deal because the only impact is that the label name won't show in the Call stack section of CDT. Labels/Call frames didn't show names before this PR anyway, so I don't think it violates anything backwards. I added the if statement just in case because I didn't want to access `callSubLine[1]` until I was sure I was on a line that had at least two "words" or pc on it.

michaeldiamant: @algochoi Thanks for the discussion here. I think it addresses my question and I'm good to resolve.

cmd/tealdbg/local_test.go
```diff
@@ -1459,3 +1460,307 @@ func TestGroupTxnIdx(t *testing.T) {
...
+	ops, err = logic.AssembleString(sourceB)
+	a.NoError(err)
+	progB := base64.StdEncoding.EncodeToString(ops.Program)
+
+	// Transaction group with 5 transactions
+	// 1. Payment txn to app A
+	// 2. Payment txn to app B
+	// 3. App call to app A
+	// 4. App call to app B with gload on app A scratch slot
+	ddrBlob := `{
```
michaeldiamant, line 1527: @algochoi Beginner's question:  How is `ddrBlob` generated?

* I'm asking for my edification and also to consider how a future maintainer will make edits.
* I realize the highlighted content is `dryrun` output.  Though it's not entirely clear to me what initial state was provided.

michaeldiamant, line 1527: @algochoi Another edification question:  What is the significance of exercising `gload` in the PR's context?

* Perhaps mistakenly, I feel like there was a verbal discussion suggesting `gload` + debugger didn't work as intended.  I no longer recall the details.  Given my vague memory, I tried to understand what `gload`-specific changes the PR makes.
* I ran `TestRunAllGloads` against `master` and discovered the test passes once implementing the PR's debugger nil checks.  Which is to say, there's no change specific to `gload`.
* I imagine there's good reason for the test.  But, it's not apparent to me, so I'm asking the question.

algochoi: I dumped the dryrun output (as a JSON instead of msgp) like: [this](https://github.com/algochoi/algorand-onboard/blob/a80085f336f8d79606374a961e69d5268b360257/tealdbg-tutorial/tealdbg_deploy.py#L215). 

I created the apps beforehand using a `ApplicationCreateTxn`, and then created a group transaction with 4 txns annotated by the comment above.

algochoi: Currently, `gload` works if you:

1. Supply a dryrun of the entire group transaction
2. Debug every single transaction in the group transaction in sequence, i.e. call `RunAll()` (notably the previous transactions that `gload` refers to needs to be "debugged" so its state and scratch vars are initialized). So putting a `gload` in the `RunAll()` to execute all the transactions in the group seemed interesting to me.

I think `gaid` still has issues because created apps and assets are not initialized properly without using some actual ledger right now.

data/transactions/logic/debugger.go
algochoi: I added a `TestParseCallstack` so that test will fail and notify us if the `callsub` semantics have changed - hopefully some peace of mind for both of us.

cmd/tealdbg/local.go
michaeldiamant: While the highlighted issue isn't central to the PR, I'm trying to use it as an opportunity to further my understanding.  The question that arises in my mind is:  Why `Run` doesn't delegate to `RunAll`?

It might be easier to discuss live.  I'm assuming I'm missing basic intuition.  

Details:
* If `Run` invokes `RunAll`, then the check for optional debugger (among other logic) lives in one place without duplication.
* As far as I can tell, `Run` is intended to be a test facility only.  It seems to me its visibility can be made private to more clearly signal its intended usage.  
* I began local changes to delegate `Run` to `RunAll` and paused because I think I'm missing basic intuition.  Hoping some folks can help me fill in the intuition:  In what cases does it make sense to generate `len(runs) > 1`, but only test the first index (e.g. `TestDebugCostPooling`)?

algorandskiy: If I recall right `Run` used to be a simplified version of `RunAll` used in tests only.

cmd/tealdbg/debugger.go
algochoi: After more thinking and adding the call stack, I think it makes sense to check the call depth for `step over`. I think `resume` should do the same thing by checking 1) previous breakpoints 2) set the earliest possible breakpoint at a point that is closest to the top of the call stack. I'm thinking `step out` should just set a breakpoint where the call frame returns.

cmd/tealdbg/local.go
michaeldiamant: @algochoi @algorandskiy I opened https://github.com/algorand/go-algorand/pull/3805 to eliminate duplicate code paths.  Can you take a look?

We can resolve the thread once #3805 is either merged into this PR or #3805 is closed.

cmd/tealdbg/debugger.go
```diff
@@ -477,9 +535,12 @@ func (d *Debugger) Update(state *logic.DebugState) error {
 
 	// copy state to prevent a data race in this the go-routine and upcoming updates to the state
 	go func(localState logic.DebugState) {
+		// Copy callstack information
+		s.setCallStack(state.CallStack)
```
algorandskiy, line 539: I think `setCallStack` should be within the "updated" notification handing, no need to set it right here since it is not used right away

cmd/tealdbg/debugger.go
```diff
@@ -477,9 +535,12 @@ func (d *Debugger) Update(state *logic.DebugState) error {
...
 		// Check if we are triggered and acknowledge asynchronously
-		if cfg.BreakAtLine != noBreak {
-			if cfg.BreakAtLine == stepBreak || breakpointLine(localState.Line) == cfg.BreakAtLine {
+		if !cfg.NoBreak {
+			if cfg.StepBreak || (localState.Line == cfg.BreakAtLine &&
+				(!cfg.StepOutOver || cfg.CallDepth == len(state.CallStack))) {
```
algorandskiy, line 543: it is time to refactor `debugConfig ` (cfg) and make an object with methods like `isBreak(state)`, `setStepBreak()`, `setBreakAt(line)`

cmd/tealdbg/debugger.go
```diff
@@ -117,6 +113,9 @@ type session struct {
 	// notifications from eval
 	notifications chan Notification
 
+	// channel to check if state has been updated after the breakpoint
+	updateChannel chan bool
```
algorandskiy, line 117: not used

algorandskiy (commented):
The approach looks good. Please encapsulate/refactor `debugConfig` into an object with couple accessor functions - managing state of 5 bool vars outside of it is tricky.

cmd/tealdbg/debugger.go
algochoi: ty

cmd/tealdbg/debugger.go
algochoi: Agreed, currently also refactoring the `BreakAtLine` into a set so `Resume` can also break during backward jumps.

cmd/tealdbg/debugger.go
```diff
@@ -181,7 +184,54 @@ func (s *session) Step() {
...
+
+	func() {
+		s.mu.Lock()
+		defer s.mu.Unlock()
+		// Step over a function call (callsub op).
+		// Set a breakpoint at the next line and resume until we reach our
+		// desired call stack height.
+		if currentOp == "callsub" {
+			err := s.setBreakpoint(s.line.Load() + 1)
+			if err != nil {
```
michaeldiamant, line 205: @algochoi When an error happens, I'm wondering if it's appropriate to modify `s.debugConfig` _at all_?

* The PR guards against configuring `StepBreak` while allowing other state modification (e.g. `StepOutOver`).
* While I think I don't yet fully understand the implications of a partial state change, it feels error prone to me.  And I'm imagining a partial state update breaks `Debugger.Update` heuristics.  
* For the moment, I'm leaving aside consideration for if/how errors ought to be propagated to the client.  Though that question is in my mind too.

cmd/tealdbg/debugger_test.go
```diff
@@ -185,3 +191,91 @@ func TestSession(t *testing.T) {
...
+// Tests control functions for stepping over subroutines and checks
+// that call stack is inspected correctly.
+func TestCallStackControl(t *testing.T) {
+	partitiontest.PartitionTest(t)
+	s := createSessionFromSource(t, "#pragma version %d\nlab1:\nint 1\ncallsub lab1\ndup\n+\n")
+	s.callStack = []logic.CallFrame{
+		{FrameLine: 2, LabelName: "lab1"},
+	}
+	// Check for step over on a callsub line
+	s.line.Store(3)
```
michaeldiamant, line 204: @algochoi For my understanding:  Had you considered writing the test _with_ a `Debugger`?

* I'm in the middle of understanding the added tests.  I'm gauging if the question's been considered because I _suspect_ the choice impacts a number of other decisions. 
* From what I've understood so far, my primary concern is:  The test modifies `Session` state without using publicly available facilities.  I'm concerned a test is liable to sequence steps in a way that's incompatible with public APIs.

cmd/tealdbg/debugger_test.go
```diff
@@ -185,3 +191,91 @@ func TestSession(t *testing.T) {
...
+	s.StepOver()
+	<-done
+
+	require.Equal(t, true, s.debugConfig.StepBreak)
+	require.Equal(t, false, s.debugConfig.StepOutOver)
+
+	require.Equal(t, 2, ackCount)
+	require.Equal(t, initialStackDepth, len(s.callStack))
+
+	// Check that step out when call stack depth is 1 sets breakpoint to the
```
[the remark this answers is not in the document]
algochoi: `s.line` actually never changes in any of these tests (`s.line` is the state where the evaluator has currently paused in the program). We create a fake "ackFunc" instead of actually updating the line and callstack (which is called by the evaluator in the `eval.go` package). 

For control like `stepOver`, we actually inspect the current line to make sure it has a `callsub` to determine how we will set the breakpoint, but the other control functions do not, so the `s.line` value won't matter in the unit tests. Let me know if this answers your question.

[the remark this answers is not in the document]
michaeldiamant: @algochoi Thanks - That addresses the question.  Though it opens other questions in my mind.  😅 

I'll resolve here and hold for a reply on https://github.com/algorand/go-algorand/pull/3653#discussion_r834577692 while continuing review.  I think the linked thread will  cover some of what I have in mind.

cmd/tealdbg/local_test.go
michaeldiamant: @algochoi Thanks for the context here - makes sense and I assume we'll treat `gaid` out-of-band to the PR.  Resolving.

cmd/tealdbg/debugger_test.go
algochoi: I guess the main reason I wrote these tests was because it was compatible to the existing unit tests.

Most of the debugger control functions are issued by `cdtSession` through the `Control` interface:

https://github.com/algorand/go-algorand/blob/1d8bd1bd79178bc1c76cd1afb41da9d2e639a806/cmd/tealdbg/cdtSession.go#L39 

I can see your concern, and I think these tests mostly test public APIs (except for setting `s.line` and `s.callstack`, which we have to manually set for these tests due to no access to the evaluator), but if you see something specific wrong, I can try to address it.

Another concern you might be having is that we're not testing end-to-end behavior (we don't actually get real-time state from the evaluator). That's a fair point, and we could write a more end-to-end testing framework for tealdbg in the future. I fear that the diff for that might make this PR much bigger than we want.

cmd/tealdbg/debugger.go
```diff
@@ -181,7 +184,54 @@ func (s *session) Step() {
...
-		s.debugConfig = debugConfig{BreakAtLine: stepBreak}
+		s.debugConfig = debugConfig{StepBreak: true}
+	}()
+
+	s.resume()
+}
+
+func (s *session) StepOver() {
+	// Get the first TEAL opcode in the line
+	currentOp := strings.Fields(s.lines[s.line.Load()])[0]
```
michaeldiamant, line 195: @algochoi Checking to see if you're aware of the rationale:  How is it safe to access `s.line` outside of `s.mu`?

* I understand the value uses atomics to provide _safe_ reads/writes.
* The concern is:  state changes for `line` are liable to happen in a way that's out-of-sync with the rest of `session` state.
  * It seems to me `session` state is intended to be acted upon transactionally.  For example, I presume that's why `Control` implementations here wrap invocations in an anonymous function.
* Provided the understanding is accurate _and_ if no other explanation exists, I suggest moving all `s.line` reads/writes within `s.mu` control.
  * I think additional attention is needed in `Debugger.Update` because `d.getSession` uses the mutex.  And from a quick read, it seems `s.mu` is _not_ re-entrant.

cmd/tealdbg/debugger.go
algochoi: An error would happen in `setBreakpoint()` if we try to set a breakpoint past the program. In that case, it sounds okay to me if we `StepBreak` instead. `Update` would be called after execution is resumed by calling `resume()` - it needs the message from the `acknowledge` channel to continue and sends a `notification` back to the evaluator to continue execution.

I don't think this particular error is notable for the client (but maybe I'm not thinking deep enough). Worst case, we are on the last line of the program and the user wanted to `StepOver`, thus triggering the error in `setBreakpoint`. In that case, setting the breakpoint properly is not critical to the user anyway. 

Let me know if you think I'm missing something.

algochoi: So I think this is not possible in the CDT debugger, and the error should never trigger in any of the control functions, but a meaningful place where `setBreakpoint()` would produce an error would be in `Debugger.setBreakpointByUrl`. If some front-end were to set a bogus breakpoint, then we should return the error. I don't think it would be possible to set a bogus breakpoint in any of the `Control` functions because we have tight control of the execution and line numbers. Unfortunately we still have to consume the error or our linter will yell at us.

cmd/tealdbg/debugger.go
algochoi: I agree moving `s.line` into the lock is safer - pushed the changes.

Might be a pedantic comment - I don't *think* `s.line` would change unless `Update` is unblocked, and under current circumstances, I can't think of how `Update` would be unblocked in the middle of `StepOver` - I think `Update` would only resume  some time after `s.resume()` sends the `ack` to the channel. Of course, this doesn't change the fact that it's still safer to access that state inside a lock, so I agree with you.

michaeldiamant: @algochoi Thanks for the change.  However, I am thinking more broadly.  What are your thoughts on the note below?

I can also discuss live if the note is too involved.

As a motivating example:  Consider a scenario where the user issues a command (e.g. step) after `s.line.Store(state.Line)` in `Debugger.Update`, but _before_ updating the remaining state in `go func(localState logic.DebugState)`.

* Such an example breaks the implied transactional update model.  In the example, `s.line` points to an updated line, but other state hasn't been updated yet.
* And even if we can convince ourselves that the scenario works today, I don't understand why we'd take the chance.  It feels to me like it'd be more robust to ensure mutual exclusion for _all_ state changes (or otherwise change the model).  
* Extending the thinking - I don't understand why `s.line` receives special treatment via atomics.  If mutual exclusion is enforced for _all_ state changes, then there ought to be _no_ visibility concerns.

I realize some (or possibly all) of these concerns predate the PR.  I'm at least trying to gauge if we agree on the problem scope.  Because _if_ we agree, it feels like there's a fundamental issue here. 
 And we can then assess when/where to take action.

cmd/tealdbg/debugger_test.go
michaeldiamant: @algochoi Thanks for the response - it addresses some of my concerns.  I tried to outline what I'm thinking below.  

I think there's a lot happening here.  I leave the note to illustrate my thinking.  I'll schedule some time to discuss because I think a verbal will move faster.  We can summarize outcomes afterwards.

* I think these tests ought to step through a problem via `Control` using `Debugger`.  `Debugger` contains the update logic (`Update`) and a reference to `session` (implements `Control`). 
* From what I can tell, the proposed tests do _not_ have a CDT dependency.  I don't think I'm proposing an _end-to-end_ test.  
* Regarding correctness, I think manually changing state (e.g. line numbers) is too likely to produce unrepresentable states.  
  * For example, I think https://github.com/algorand/go-algorand/pull/3653#discussion_r834600725 illustrates a broken state.  Getting to line 4 with a non-empty call stack _should not_ be possible.
  * If my understanding is accurate, then I'm left feeling like I can't rely on these tests.  
* I understand the request extends the PR. 
  * I am open to finding ways to merge what's open and use successive PRs to keep scope tighter.
  * My principal concern is agreeing on the approach for correctness.

cmd/tealdbg/debugger.go
michaeldiamant: @algochoi I appreciate the detailed response here.  You've convinced me that the PR's behavior is correct.  Thanks for your patience - Resolving.

cmd/tealdbg/debugger_test.go
algochoi: RE:  off line discussion

I think I am beginning to understand your concerns and I'm working on a stronger test for recursion - we can loop back after I push some changes I am working on.

cmd/tealdbg/debugger.go
algorandskiy: First reason is `line` is really get used by debugger and part of its state. The `state` is not really needed and the debugger here is a "persistence" place to have a an initial state and apply changes as the evaluation goes.
But I agree, synchronous update would not hurt.

cmd/tealdbg/debugger_test.go
algochoi: I pushed some changes - they are actually mostly minor and might not address your initial problem: I'll explain my reasoning below. 

I'm getting more convinced that it will turn into an end-to-end test if we really want to test recursive calls or any other runtime conditions without calling the actual evaluator. My initial thoughts on this unit test is that it should not depend on a particular implementation on the node's evaluator. Since the evaluator is responsible for updating the callstack and the line numbers, then I think it's okay to set these ourselves in the unit tests. I think the current debugger is not supposed to be exposed to how the TEAL program is supposed to run anyway. 

I guess you could propose to write our own evaluator (probably a subset of it; writing the whole thing would take a while 😅 ) in the debugger to simulate recursive calls and alike, but I still think it couples our debugger with our current implementation of the evaluator. 

Let me know if this makes sense and if you have any other concerns.

cmd/tealdbg/debugger.go
```diff
@@ -89,19 +91,54 @@ type programMeta struct {
...
-// breakpointLine is a source line number with a couple special values:
-// -1 do not break
-//  0 break at next instruction
-//  N break at line N
-type breakpointLine int
+// debugConfig contains information about control execution and breakpoints.
+type debugConfig struct {
+	NoBreak     bool `json:"nobreak"`
+	StepBreak   bool `json:"stepbreak"`
+	StepOutOver bool `json:"stepover"`
```
algorandskiy, line 98: nit: It looks like not all the combinations are valid (for example NoBreak=true and StepBreak=true). This is an indication these might be flags and not boolean variables

cmd/tealdbg/debugger.go
```diff
@@ -89,19 +91,54 @@ type programMeta struct {
...
+
+func (dc *debugConfig) setStepOutOver(callDepth int) {
+	dc.StepOutOver = true
+	dc.CallDepth = callDepth
+}
+
+// setActiveBreak does not check if the line is a valid value, so it should
+// be called inside the setBreakpoint() in session.
+func (dc *debugConfig) setActiveBreak(line int) {
+	dc.ActiveBreak[line] = struct{}{}
```
algorandskiy, line 126: Andrew, I might have already asked but can't find the answer - what is the point of having `ActiveBreak` map? It is only set in `setActiveBreak`. And everywhere we call `setBreackpoint` the contol state is reset `s.debugConfig = makeDebugConfig()` leaving only a single entry in `ActiveBreak`.

algochoi: It's only used in the `Resume()` to set multiple active breakpoints (in case we jump backwards), but you're right that in everywhere else there is only 1 active break.

cmd/tealdbg/debugger_test.go
michaeldiamant: @algochoi Gotcha - A few points:
* I'll need more time to consider these thoughts.  I'll do some homework and circle back.
* I opened https://github.com/algochoi/go-algorand/pull/2 to make the test cases independent.  Can you take a look?  
  * Since the PR is opened against your fork, I think you'll need to approve the workflow run:  https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks.   
  * When I began reviewing again, I struggled to understand the initial state for each test case.  Splitting the cases ensures each scenario acts in isolation.  
  * If you think it's an improvement, I'd like to merge as soon as is reasonably possible.  The PR's changes are liable to result in conflicts.

jasonpaulos:
I've manually tested on the following recursive TEAL program, and step in and step over are working great! My only complaint is that step out is frequently buggy and takes me to the end of the program, though I've not yet debugged why this happens.

```
#pragma version 6
int 5
callsub factorial
return

factorial:
dup
int 1
<=
bz recursive_step
pop
int 1
retsub
recursive_step:
dup
int 1
-
callsub factorial
*
retsub
```

cmd/tealdbg/debugger.go
michaeldiamant: @algochoi @algorandskiy Thanks for the discussion here to help me understand the space.  I think we're good to resolve.  

I opened https://github.com/algorand/go-algorand/issues/3850 to track the concern.  We can follow up out-of-band to the PR.

cmd/tealdbg/debugger.go
```diff
@@ -181,25 +221,70 @@ func (s *session) Step() {
...
+		if len(s.callStack) == 0 {
+			s.debugConfig.setNoBreak()
+		} else {
+			callFrame := s.callStack[len(s.callStack)-1]
+			err := s.setBreakpoint(callFrame.FrameLine + 1)
+			if err != nil {
+				s.debugConfig.setStepBreak()
+			}
+			s.debugConfig.StepOutOver = true
+			s.debugConfig.CallDepth = len(s.callStack)
```
michaeldiamant, line 270: @algochoi The _suggestion_ provides a partial answer to the _step out_ issue @jasonpaulos points out in https://github.com/algorand/go-algorand/pull/3653#issuecomment-1083645448.

* Through debugging, it seems clear to me that the PR's `debugConfig.CallDepth` is _not_ compatible with `isBreak`.  During execution, the `callDepth` provided to `isBreak` is `len(s.callStack) - 1`.  Consequently, the intended breakpoint does _not_ trigger.
* The _suggestion_ improves the behavior, but:
  * I think there's at least 1 more problem.
  * I did _not_ run unit tests.  I tested manually as described below.  I recommend running the tests first.  
    * If the tests _don't_ break, then I think the tests are broken or a test is missing.
    * You may also wish to refactor to eliminate duplicate `s.callStack` index management to minimize chance of future divergence.
* I debugged manually using the program @jasonpaulos provided in the linked comment.  Here's the dryrun output I generated by using your tutorial: 
  * [dryrun_txn.msgp.txt](https://github.com/algorand/go-algorand/files/8385599/dryrun_txn.msgp.txt)
  * Github allows limited file extensions for upload.  Remove `.txt` extension.
* Problematic scenario - I discovered it and ran out of time to assess:
  * Open CDT and set breakpoint at line 7 (`dup`).
  * Resume execution.  Confirm CDT jumps to line 7 with call stack size = 2.
  * Step out.  Confirm CDT jumps to line 5 (`return`) with call stack size = 1 (`main`).
  * Resume execution.  Confirm CDT jumps to line 21.
    * I think reaching line 21 is a _problem_.  There's no opcode here.  I think execution _should_ complete.
    * Since I didn't debug, it's unclear to me if it's related to the _suggestion_.  
  * Resume execution.  Confirm CDT completes execution as intended.


```suggestion
			s.debugConfig.CallDepth = len(s.callStack) - 1
```

cmd/tealdbg/local_test.go
michaeldiamant: @algochoi Thanks for the explanation and for a quick verbal.  I'm good to resolve.

For completeness -  I have lingering concerns about the difficulty in maintaining tests that require manual setup to reconstruct test input.  Though it seems the convention is applied elsewhere, so let's run with it for now.

cmd/tealdbg/debugger_test.go
michaeldiamant: @algochoi I apologize for the long response time.  I'm good to resolve the thread.  Leaving details below if interested.

* Thanks for merging in https://github.com/algochoi/go-algorand/pull/2.
* I agree attempting to use the _real_ evaluator moves from unit to integration test territory. 
  * I attempted to write such a test locally and see there's a number of moving parts.  Given the desire to get something merged, I agree with the unit test focus.  
  * However, issues like https://github.com/algorand/go-algorand/pull/3653#issuecomment-1083645448 make me feel we may revisit the topic in future iterations.

cmd/tealdbg/debugger.go
algochoi: Thanks for the suggestion - I agree it should be `len(s.callStack) - 1`. Without the `-1`, it tries to break too early in the callstack and is most apparent when you try to `StepOut` inside the deepest point of the subroutine. 

Also changed the `pauseOnCompleted` to only be set at the most outer subroutine. This should retain previous functionality before subroutines were introduced.

michaeldiamant (approved):
@algochoi Thanks for your patience working through multiple iterations to enrich tealdbg.  ☕

michaeldiamant (approved):
Re-approving after minor change dismissed my prior approval.  Still looks good to me.

cmd/tealdbg/debugger.go
```diff
@@ -209,21 +293,30 @@ func (s *session) Resume() {
...
 	s.breakpoints[line] = breakpoint{set: true, active: true}
-	s.debugConfig = debugConfig{BreakAtLine: breakpointLine(line)}
+	s.debugConfig.setActiveBreak(line)
 	return nil
 }
 
 func (s *session) SetBreakpoint(line int) error {
 	s.mu.Lock()
 	defer s.mu.Unlock()
+	// Reset all existing flags and breakpoints and set a new bp.
```
algorandskiy, line 309: why moved the resetting `s.debugConfig` from `SetBreakpoint` to `setBreakpoint`?

data/transactions/logic/debugger_test.go
```diff
@@ -174,3 +174,27 @@ func TestValueDeltaToValueDelta(t *testing.T) {
 	require.Equal(t, base64.StdEncoding.EncodeToString([]byte(vDelta.Bytes)), ans.Bytes)
 	require.Equal(t, vDelta.Uint, ans.Uint)
 }
+
+func TestParseCallstack(t *testing.T) {
```
algorandskiy, line 178: add a real test for with two nested callstack received from the evaluator via debugger.Update.

cmd/tealdbg/debugger.go
algochoi: Good catch - I will move this back into `setBreakpoint`

algochoi: Actually, for `Resume` there can be multiple active breakpoints (code execution can jump backward and break on a previous bp) set so I moved this outside instead - let me know if that makes sense.

data/transactions/logic/debugger_test.go
algochoi: Good point - added unit test below (real test checks `line+1` due to `pragma`)

algorandskiy (approved):

PR #3655, opened 2022-02-18
AVM: Add bn256

bo-abstrlabs:
## Summary

Ethereum has a few precompiled contracts that do bn256 add, scalar multiply and pairing:

https://github.com/ethereum/go-ethereum/blob/d6f49bf764c48e9415b168e701b6dc26566a675c/core/vm/contracts.go#L89-L91

Their usage in a solidity contract to do zksnark proof verification:
https://github.com/ConsenSys/gnark/blob/cc10c32452124bbf2f1115b3d6ac9a05710dc3ab/internal/backend/bn254/groth16/solidity.go#L76
https://github.com/ConsenSys/gnark/blob/cc10c32452124bbf2f1115b3d6ac9a05710dc3ab/internal/backend/bn254/groth16/solidity.go#L98
https://github.com/ConsenSys/gnark/blob/cc10c32452124bbf2f1115b3d6ac9a05710dc3ab/internal/backend/bn254/groth16/solidity.go#L141

Our need team need to do similar zksnark verification in TEAL, however, without these crypto primitives it's prohibitively expensive.

## Test Plan

Existing make test. 
TODO:
- [x] add unit test to cover these functions in eval_test.go
- [x] figure out more accurate opcost. Currently, I estimated opcost by borrowing ethereum's gas cost and do a simple conversion. See below for details

## Current Opcost Estimation
Ethereum gas cost is calculated in:

https://github.com/ethereum/go-ethereum/blob/master/core/vm/contracts.go

required constants are defined in this file: https://github.com/ethereum/go-ethereum/blob/master/params/protocol_params.go

Result:
sha256 of 128bytes cost
```
sha256 128bytes gas = uint64(len(input)+31)/32*params.Sha256PerWordGas + params.Sha256BaseGas = 12*(128+31)/32+60 = 120
```
in ethereum, that is 35 opcost in algorand. Assume the cost scale in the same manner. We can get:

```
bn256_add in ethereum: 150
according gas in algorand: 45

bn256_scalar_mul in ethereum: 6000
according gas in algorand: 1800

bn256_pairing in ethereum: 
params.Bn256PairingBaseGasIstanbul + uint64(len(input)/192)*params.Bn256PairingPerPointGasIstanbul
45000 + 34000 = 79000
according gas in algorand: 23042
```

Noticed that bn256_pairing is even more expensive than 16*700, that make it impossible to execute even with group txn, AFAIK. Question: I'd like to know the exact way your team measure opcost and doing that for these ops.


jannotti:
Can you comment on the concerns raised here: https://tools.ietf.org/id/draft-yonezawa-pairing-friendly-curves-00.html

Particularly that BN256 ought not be used in new applications that expects 128 but security?

So you require compatibility with the curves used in Ethereum, or would BLS curves be a better choice?

Separately, since these are so computationally expensive, they would probably only be usable in logicsigs. If the inputs are computed off-chain, that will be a minor inconvenience, but doable. If a contract is constructing the inputs, you would have a bigger problem. Can you comment on your use case?

bo-abstrlabs:
> Particularly that BN256 ought not be used in new applications that expects 128 bit security?
Do you require compatibility with the curves used in Ethereum, or would BLS curves be a better choice?


Good point. Our use case is build a new verifier contract on algorand. So we are open to choose BLS curves. Bn256 may bring value to other developer that trying to keep ethereum compatible with their existing contract, but that doesn't include our team.

However when read through ietf draft, they mentioned:

> We have to note that, according to [[7]](https://tools.ietf.org/id/draft-yonezawa-pairing-friendly-curves-00.html#MSS17), the bit length of p for BLS12 to achieve 128 bits of security is calculated as 384 bits and more, which BLS12-381 does not satisfy. Although the computational time is conservatively estimated by 2^110 when exTNFS is applied with index calculus, there is no currently published efficient method for such computational time. They state that BLS12-381 achieves 127-bit security level evaluated by the computational cost of Pollard's rho.

So ethereum provided BLS-381 curve also doesn't give 128 bit security. Another concern is cost. Looking at: https://github.com/ethereum/go-ethereum/blob/d6f49bf764c48e9415b168e701b6dc26566a675c/params/protocol_params.go#L151, we can see BLS-381 is more expensive than BN256 for add, scalar mult and pairing. Therefore, I think BN256 is a more practical choice for DApps that doesn't require a 128 bit security.

> Separately, since these are so computationally expensive, they would probably only be usable in logicsigs. If the inputs are computed off-chain, that will be a minor inconvenience, but doable. If a contract is constructing the inputs, you would have a bigger problem. Can you comment on your use case?

My use case is a zksnark verifier contract (The CallApp would invoke the verifyProof method as this solidity one: https://github.com/ConsenSys/gnark/blob/cc10c32452124bbf2f1115b3d6ac9a05710dc3ab/internal/backend/bn254/groth16/solidity.go#L187). The verifyProof calls `bn256_add`, `bn256_scalar_mul` and `bn256_pairing`. The argument pass to `bn256_pairing` is calculated with `bn256_add` and `bn256_scalar_mul`, so it's unfortunately "contract is contructing the input". 

Also, I find this benchmark which shows that gnark's bn254 pairing is 20x faster than the bn256 used in go-ethereum: https://hackmd.io/@gnark/eccbench#Pairing. If it is true, it will fit in group txns. Can you show me how is existing opcost measured/estimated, so I can get a more accurate result and bench with different bn implementations? Thank you!

jannotti:
> Also, I find this benchmark which shows that gnark's bn254 pairing is 20x faster than the bn256 used in go-ethereum: https://hackmd.io/@gnark/eccbench#Pairing. If it is true, it will fit in group txns. Can you show me how is existing opcost measured/estimated, so I can get a more accurate result and bench with different bn implementations? Is it measured by `func benchmarkOperation`? Thank you!

Look in data/transactions/logic/eval_test.go for various Benchmark* functions.  They can be a bit hard to understand (they don't just spit out a cost) but they they can be used to get comparative costs of various opcodes. `BenchmarkBase64Encode` might be the easiest to modify, and it contains various other opcodes that make for easier comparisons.

These benchmarks make huge programs that run the operation over and over until we get runtime estimate. The thing that makes it hard to compare is that each opcode typically needs some extra work done before or after it runs to "clean up", before it can be run again.  So if it takes 3 extra opcodes (usually simple ones) to set up the opcode you intend to measure, you'll have to manually remove that cost from what the benchmark reports.

bo-abstrlabs:
@jannotti I run some of existing benches follow your instruction, and add bn256 bench in the similar way. This is the result:
```
BenchmarkHash/sha256-small-16         	 3244508	       471.7 ns/op	         0 waste/op	      38 B/op	       1 allocs/op
BenchmarkHash/sha256-med-16           	  927352	      1104 ns/op	        60.00 waste/op	     230 B/op	       3 allocs/op
BenchmarkHash/sha256-big-16           	  363922	      3482 ns/op	       120.0 waste/op	     999 B/op	       5 allocs/op
BenchmarkHash/keccak256-small-16      	  579124	      2234 ns/op	         0 waste/op	     968 B/op	       4 allocs/op
BenchmarkHash/keccak256-med-16        	  881739	      2588 ns/op	        60.00 waste/op	    1159 B/op	       6 allocs/op
BenchmarkHash/keccak256-big-16        	  184374	      6167 ns/op	       120.0 waste/op	    1943 B/op	       8 allocs/op
BenchmarkHash/sha512_256-small-16     	 2368228	       703.4 ns/op	         0 waste/op	      38 B/op	       1 allocs/op
BenchmarkHash/sha512_256-med-16       	 1000000	      1224 ns/op	        60.00 waste/op	     231 B/op	       3 allocs/op
BenchmarkHash/sha512_256-big-16       	  467223	      2347 ns/op	       120.0 waste/op	    1000 B/op	       5 allocs/op
BenchmarkEcdsa/ecdsa_verify-16         	   13210	     82421 ns/op	   17070 B/op	      73 allocs/op
BenchmarkEcdsa/ecdsa_pk_decompress-16  	   52358	     31851 ns/op	   16363 B/op	      71 allocs/op
BenchmarkEcdsa/ecdsa_pk_recover-16     	   13226	    107761 ns/op	   18215 B/op	     100 allocs/op

with crypto/bn256:
BenchmarkBn256/bn256_add-16                10000             28463 ns/op           21808 B/op        151 allocs/op
BenchmarkBn256/bn256_scalar_mul-16         10000           1586964 ns/op          294654 B/op       3661 allocs/op
BenchmarkBn256/bn256_pairing-16: too expensive
```

I noticed that crypto/bn256 lib is very inefficient that even scalar multiply is over expensive. So I replaced with the known best library of the same curve, gnark/bn254: 
```
with gnark/bn254:
BenchmarkBn256/bn256_add-16                10000             11440 ns/op           15866 B/op         46 allocs/op
BenchmarkBn256/bn256_scalar_mul-16         10000             65103 ns/op           16522 B/op         62 allocs/op
BenchmarkBn256/bn256_pairing-16            10000            571072 ns/op           17868 B/op         71 allocs/op
```

With this result the most expensive one, pairing, would be 10599 (If consider BenchmarkEcdsa/ecdsa_pk_recover-16=107761 ns/op=2000 opcost). It's slightly less than 700*16=11200, so I guess it'll be available in the maximum grouped transaction?

I update the implementation and the benchmark that produced the above gnark/bn254 results. Then updates the cost that scales the ecdsa_pk_recover's cost. The benchmark also automatically serves as a logic test of three new opcodes. So I mark the PR as ready for review.

bo-abstrlabs:
I found there're some places to improve the cost:
- don't use gnark's unmarshal, which does a point decompression that is expensive.
- gnark's SetBigInt also transform g1/g2 to montgometry form, which is also expensive. If raw bytes are pre-transformed, it can be cheaper too.

Also, the pairing could only pair one g1 and one g2. In Ethereum, variable length of bytes contains an array of g1 and an array of g2. However in current AVM one bytes parameter cannot be more than 128 bytes, which can only hold a single g2. AFAIK, **AVM opcode cannot take variable length argument** ? So I'd need to implement pair1, pair2, ..., pair10 to pair N element of g1 and g2, respectively. Is there other suggested approach on this? Thanks!

jannotti:
> However in current AVM one bytes parameter cannot be more than 128 bytes, which can only hold a single g2. AFAIK, **AVM opcode cannot take variable length argument** ? So I'd need to implement pair1, pair2, ..., pair10 to pair N element of g1 and g2, respectively. Is there other suggested approach on this? Thanks!

I think you have two questions here?  For the first, the AVM byte length limit on a single value is 4k - so that's much bigger than 128 bytes.  There are some shorter limits for doing math with the values, using opcodes like `b+`, but I'm not sure that matters to you.

opcodes that deal with a wariable number of elements on the stack are unusual, but `dig`, `cover`, and `uncover` have that property to an extent.  Though they always affect the total height of the stack in a constant way.

bo-abstrlabs:
Hi @jannotti Thanks for correction! I incorrectly assumed `[]byte` max 128bytes means maximum single argument size to be 128 bytes. In fact that's 4096bytes (you said 4k is correct). 4kb is sufficient for many G1 (64bytes) and G2 (128bytes), so I don't need a variable length argument! 

And good point on `dig` `cover` and `uncover`, next time when we there's a need variable length argument we'll make use of them.

Additional question on cost: for `sha256` I was assume the cost=35 is measured by the largest possible input. **That is, cost to sha256 a 4kb input. Is this correct?**

jannotti:
No, the hash cost is for a typical case. Algorand may move to a data
dependent cost model in order to more accurate charge opcode costs for
short or long hashes.

bo-abstrlabs:
> No, the hash cost is for a typical case. Algorand may move to a data
dependent cost model in order to more accurate charge opcode costs for
short or long hashes.

Makes sense. NEAR's runtime use `base cost + n * byte cost` for sha256 etc. What is that typical case that results in `sha256 cost=35`? If it's not the largest (most expensive) input, transactions can be undercharged.

bo-abstrlabs:
Before:
```
BenchmarkBn256/bn256_add-16                10000             11440 ns/op           15866 B/op         46 allocs/op
BenchmarkBn256/bn256_scalar_mul-16         10000             65103 ns/op           16522 B/op         62 allocs/op
BenchmarkBn256/bn256_pairing-16            10000            571072 ns/op           17868 B/op         71 allocs/op
```
With updated no point compression:
```
BenchmarkBn256/bn256_add-16                10000              8688 ns/op           16691 B/op         59 allocs/op
BenchmarkBn256/bn256_scalar_mul-16         10000             48063 ns/op           22827 B/op         74 allocs/op
BenchmarkBn256/bn256_pairing-16            10000            370644 ns/op           24854 B/op         87 allocs/op
```
I updated cost accordingly. 

Set montgometry form instead of regular isn't possible with current gnark APIs. Given:
1) cost is already looks okay, 
2) regular form is valid integers in AVM that can be processed with arithmetic operators while montgometry form would be a black box
3) Ethereum precompiled contracts also use regular form as input

I think use the regular form is OK. So I'm not going to fork gnark to add "SetMontgometryRawBytes"

bo-abstrlabs:
My previous comment about bench is wrong. I made an mistake in deserialization. After fixing it in [2e571cc](https://github.com/algorand/go-algorand/pull/3655/commits/2e571ccf94e6f38f1d87b75140cd28e9dbc9a291), it turns out that the marshal is not obviously cheaper than gnark's unmarshal. 

Also, further investigate shows that unmarshal is trivial for bn256_scalar_mul and bn256_pairing. It's only worth optimizing for bn256_add. **And generally, bn256 add and bn256 scalar mul, although expensive, is still acceptable. But bn256_pairing is too expensive to be possible:**

Above cost of bn256_pairing:
```
BenchmarkBn256/bn256_pairing-16            10000            571072 ns/op           17868 B/op         71 allocs/op
```
Barely fit into 11200 cost upperbound. However it is count for pairing one G1 point with one G2 point. Our more practical use case: verify a zksnark proof, requires pairing 4 G1 points with 4 G2 points. And it costs about twice. This is the one I have to optimize

data/transactions/logic/eval.go
```diff
@@ -1260,6 +1262,121 @@ func opDivw(cx *EvalContext) {
...
+		return
+	}
+
+	cx.stack = cx.stack[:last]
+	if ok {
+		cx.stack[prev].Bytes = []byte{1}
+	} else {
+		cx.stack[prev].Bytes = []byte{0}
+	}
+}
```
jannotti, line 1378: The usual convention is to use a Uint to convey a boolean value, rather than a bytes.  See, for example, opEcdsaVerify.  You probably want:
```
cx.stack[prev].Uint = boolToUint(ok)
cx.stack[prev].Bytes = nil
```

data/transactions/logic/opcodes.go
```diff
@@ -319,6 +319,9 @@ var OpSpecs = []OpSpec{
 	{0x96, "bsqrt", opBytesSqrt, asmDefault, disDefault, oneBytes, oneBytes, 6, modeAny, costly(40)},
 	{0x97, "divw", opDivw, asmDefault, disDefault, twoInts.plus(oneInt), oneInt, 6, modeAny, opDefault},
 	{0x98, "sha3_256", opSHA3_256, asmDefault, disDefault, oneBytes, oneBytes, 7, modeAny, costly(130)},
+	{0x99, "bn256_add", opBn256Add, asmDefault, disDefault, twoBytes, oneBytes, 6, modeAny, costly(166)},
+	{0x9a, "bn256_scalar_mul", opBn256ScalarMul, asmDefault, disDefault, twoBytes, oneBytes, 6, modeAny, costly(892)},
+	{0x9b, "bn256_pairing", opBn256Pairing, asmDefault, disDefault, twoBytes, oneBytes, 6, modeAny, costly(6879)},
```
jannotti, line 324: Can you comment on how you chose the set of opcodes you feel are complete enough to prove snarks.  In particular, it looks like Ethereum is adding 9 pre-compiles in support of BLS12-381. https://eips.ethereum.org/EIPS/eip-2537

jannotti: Though I see that they only added the three you chose for support of bn256. So I suppose there are good reasons, and the curves are typically used differently.

bo-abstrlabs: The major reason is cost. bls-381 is more expensive than bn256. And even bn256 barely fit in AVM's 11200 (stateful 16 grouped tx) / 20000 (stateless) cost limit. 

I made mistake in unmarshal, which makes 6879 above wrong. Correct one is about 8000-9000 for pairing one pair. In verifying SNARK it's pairing 4 pairs, which costs about 20000.

**Is there plans to allow group more than 16 txns?** The 11200 cost limit is too strict for bn256 pairing operations. Actually, go-ethereum has selected a less efficient library and got a much more expensive cost.Our bn256 pairing is about 6x (!) faster than go-ethereum, but it's still doesn't fit 11200. This indicates two things:
1) it's unlikely to make non-trivial optimization on the timing over the library that we used for bn256 pairing (gnark's bn254);
2) Ethereum has a much higher gas limit than AVM, if compare the cost by doing same operations
So to support operations as expensive as pairing, I think we do need a higher bar. And theoratically, I think it doesn't hurt to allow expensive operations by paying more tx fee. More specifically, I know a tx can call 256 inner txns to have a totally 179.2K cost limit, so is it possible to group 256 txns to have 179.2K cost limit? :thinking: 

(!): 6x is from both this: https://hackmd.io/@gnark/eccbench and my local benchmark of go-algorand and go-ethereum git-blame of cost parameter, which refer to [eip-1108](https://eips.ethereum.org/EIPS/eip-1108.) and this [bench](https://gist.github.com/zac-williamson/838410a3da179d47d31b25b586c15e53). 
More detail result:

go-ethereum:
```
bn256 pairing: 
3037 microseconds, 1 pairing
14663 microseconds, 10 pairings
= 1745 microseconds + 1292 microseconds per pairing
= 45000 + k*35000 eth gas
when k = 4, it's about 185000 eth gas, 61x 116 ~ 7000 microseconds

In go-algorand
pairing k=1 about 460 microseconds
pairing k=4 about 1000 microseconds
ecrecover on my machine: about 100 microseconds, 2000 cost
```

jannotti: I think you're answering a different question. While opcode cost matters, I'm asking a different question.

Why is it that you (and Ethereum, so you're good company!) have only added three new opcodes to provide support for bn256, but nine new opcodes have been proposed to support bn12-381? I'm trying to ensure we get the basic primitives right, and I'm surprised that two different curves have a different sent of operations.

ahangsu:
Hey, just a (somewhat) relevant question: I noticed that we are using `bn254` from `gnark`, while we name this is `bn256`.
I noticed from this link: https://hackmd.io/@liangcc/bls-solidity#alt_bn128-bn256-bn254-why-so-many-names-and-so-confusing that people treat them as the same thing.

Just want a verification that `bn254` is the same curve as `bn256`, while 254 bit prime order fits into 256bit (32byte) field?

data/transactions/logic/opcodes.go
bo-abstrlabs: Sorry didn't address your question. **My understanding is 1) 3-primitive in bn256 is sufficient for SNARK verification; 2) BLS curves also have a signature verification use case, which need additional primitives.** Evidence below.

1. Three bn256 primitives are sufficient to verify SNARKs generated from a bn256 curve, [solidity example](https://github.com/ConsenSys/gnark/blob/cc10c32452124bbf2f1115b3d6ac9a05710dc3ab/internal/backend/bn254/groth16/solidity.go) and [ported pyteal](https://github.com/AbstrLabs/algorand_zkp_verifier/blob/verifier-contract-v1/assets/zkp_verifier.py). For same reason we also only added three bn256 primitives in NEAR.

2. Bn256 were added to Ethereum earlier than bls12-381 in [eip-196](https://eips.ethereum.org/EIPs/EIPS/eip-196) and [eip-197](https://eips.ethereum.org/EIPs/EIPS/eip-197). The purpose as mentioned in eip is "perform zkSNARK verification". Bls12-381 were introduced in [eip-2537](https://eips.ethereum.org/EIPS/eip-2537), the purpose are "BLS signature verification and perform SNARKs verifications".

bo-abstrlabs:
> Just want to make sure that `bn254` is the same curve as `bn256`, 

Yes you're looking at the right documentation! We manually checked the field definition of two libraries and can confirm they are the same curve. We had also indirectly compare the same operation with same operands of two libs and they result in same result.

> while 254 bit prime order fits into 256bit (32byte) space?

True. bn254 library marshal field to 32 byte

ahangsu (commented):
Just took a quick pass on the implementation, and want to point out one minor thing about the build:
I think we are building this repo over golang 1.16, while this branch is building on golang 1.14, which cause something in `gnark-crypto` error in compiling. You probably want to this branch to be updated to (somewhat) latest to have a build pass...

go.mod
```diff
@@ -13,6 +13,7 @@ require (
 	github.com/algorand/websocket v1.4.4
 	github.com/aws/aws-sdk-go v1.16.5
 	github.com/chrismcguire/gobberish v0.0.0-20150821175641-1d8adb509a0e
+	github.com/consensys/gnark-crypto v0.6.1 // indirect
```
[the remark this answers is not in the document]
bo-abstrlabs: Sounds good!

data/transactions/logic/eval.go
```diff
@@ -1260,6 +1262,125 @@ func opDivw(cx *EvalContext) {
 	cx.stack[pprev].Uint = quo
 }
 
+func bytesToBN254Field(b []byte) (ret fp.Element) {
+	// var x big.Int
+	// x.SetBytes(b[:32])
+	// ret.SetBigInt(&x)
+	ret.SetBytesRaw(b)
```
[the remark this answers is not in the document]
bo-abstrlabs: Good catch! I was experimenting faster options in serializing, and SetRawBytes were a method I patched to drop montgometry conversion, but it has no speedup, so we'll keep use existing SetBytes API

bo-abstrlabs:
@ahangsu Thanks for reviewing. I addressed your and John's comments. Clean up the code and benchmark. Remove empty lines to be consistent with other opcodes. And most importantly, update cost according to latest result:
```
go-algorand (add-bn256) go test -benchmem -run='^$' -bench '^(BenchmarkBn256)$' github.com/algorand/go-algorand/data/transactions/logic
goos: linux
goarch: amd64
pkg: github.com/algorand/go-algorand/data/transactions/logic
cpu: Intel(R) Core(TM) i7-10875H CPU @ 2.30GHz
BenchmarkBn256/bn256_add-16               265399              3777 ns/op                 1.000 extra/op      103 B/op          2 allocs/op
BenchmarkBn256/bn256_scalar_mul-16         22376             52262 ns/op           16799 B/op         67 allocs/op
BenchmarkBn256/bn256_pairing-16             2528            468702 ns/op           17717 B/op         60 allocs/op
PASS
ok      github.com/algorand/go-algorand/data/transactions/logic 5.647s
```

The cost is scaled according to ns/op of ecrecover:
```
BenchmarkEcdsa/ecdsa_pk_recover-16     	   13226	    107761 ns/op	   18215 B/op	     100 allocs/op
```
which has a op cost 2000.

ahangsu:
I think there are still some issues in testing (and plausibly in implementation), from my observation in https://app.circleci.com/pipelines/github/algorand/go-algorand/5945/workflows/e338a2a3-f859-413f-a92c-4058f5cbbc07/jobs/100027
- This test https://github.com/algorand/go-algorand/blob/master/data/transactions/logic/assembler_test.go#L413 failed for not applying all three `bn256*` opcodes in compilable nonsense code https://github.com/algorand/go-algorand/blob/master/data/transactions/logic/assembler_test.go#L344, and you might also want to modify https://github.com/algorand/go-algorand/blob/master/data/transactions/logic/assembler_test.go#L379 accordingly.
- This test https://github.com/algorand/go-algorand/blob/master/data/transactions/logic/doc_test.go#L28 failed for not adding docs (i.e., a short description) for all three `bn256*` opcodes in https://github.com/algorand/go-algorand/blob/master/data/transactions/logic/doc.go#L24 
- This test https://github.com/algorand/go-algorand/blob/master/data/transactions/logic/doc_test.go#L62 failed for not adding all three opcodes in https://github.com/algorand/go-algorand/blob/master/data/transactions/logic/doc.go#L336, which I think should fall in `"Arithmetic"` section
- This test https://github.com/algorand/go-algorand/blob/master/data/transactions/logic/evalStateful_test.go#L2269 failed for putting a 7 byte array on stack, while your implementation expecting a 32 byte thing.
  While it is unsure how you would like to handle it, there might be a few steps we can take:
  - probably put three opcodes in `skipCmd`, since they are expecting exactly 32 bytes on stack
  - probably check the length of argument byte to be exactly 32 bytes, or alternatively prepend 0 to the arguments (assuming they are big endian).

My apologies for these long comments (don't have prev experience writing opcodes tho), but we cannot merge unless we make CI happy 😅

bo-abstrlabs:
@ahangsu Thanks for explaining every failure! Your CI setup is quite reasonable and conprehensive. Sorry for not examine CI tests before - most of my previous commits have a pending CI due to the CLA which made me forget about CI :sweat_smile: 

I did these tests, which probably only prove the correctness with valid input:
- benches as in the PR
- `make install`, run a local node. Deploy a teal with all three `bn256_*` instructions and obtained a correct result: https://github.com/AbstrLabs/algorand_zkp_verifier/pull/1/files (the contract is in pyteal, so also added the new opcodes in pyteal)

I'll address CI failures, again thanks for your nice comments!

data/transactions/logic/opcodes.go
bo-abstrlabs: updated two broken links

bo-abstrlabs:
@ahangsu Only have this CI check left, which seems need your approval to proceed:
image


For this failure:
> This test https://github.com/algorand/go-algorand/blob/master/data/transactions/logic/evalStateful_test.go#L2269 failed for putting a 7 byte array on stack, while your implementation expecting a 32 byte thing.

it was giving a lot of panics:
```
...
    --- FAIL: TestReturnTypes/mode=Signature,opcode=bn256_scalar_mul (0.00s)
        /home/bo/workspace/go-algorand/data/transactions/logic/evalStateful_test.go:2468: 
            	Error Trace:	evalStateful_test.go:2468
            	Error:      	Received unexpected error:
            	            	panic in TEAL Eval: runtime error: slice bounds out of range [:32] with capacity 7
            	            	goroutine 159 [running]:
            	            	github.com/algorand/go-algorand/data/transactions/logic.eval.func1()
            	            		/home/bo/workspace/go-algorand/data/transactions/logic/eval.go:653 +0x8c
            	            	panic({0xdd4b60, 0xc00011eb88})
            	            		/usr/local/go/src/runtime/panic.go:1047 +0x266
```
 I add length checks to make every failure returns human readable error messages, like other opcodes did. And the test looks like this:
```
    --- FAIL: TestReturnTypes/mode=Signature,opcode=bn256_scalar_mul (0.00s)
        /home/bo/workspace/go-algorand/data/transactions/logic/evalStateful_test.go:2468: 
            	Error Trace:	evalStateful_test.go:2468
            	Error:      	Received unexpected error:
            	            	expect G1 in 64 bytes
            	Test:       	TestReturnTypes/mode=Signature,opcode=bn256_scalar_mul
            	Messages:   	bn256_scalar_mul: expect G1 in 64 bytes
```

And now it is safe to apply your advice:

> probably put three opcodes in skipCmd, since they are expecting exactly 32 bytes on stack

ahangsu:
implementation seems largely good to me, it seems `bN254G2ToBytes` can be removed (review-dog's complaining), since all the addition/multiplication are happening over G1, and G2 is only needed for pairing check, am I correct about that?

Another thing to verify: the byte-representations of curve elements over stack are big-endian encodings, and we rely on the `bn254` to internally convert them into Montgometry format, right?

bo-abstrlabs:
> and G2 is only needed for pairing check, am I correct about that?

Yes

> Another thing to verify: the byte-representations of curve elements over stack are big-endian encodings, and we rely on the bn254 to internally convert them into Montgometry format, right?

Correct. User provides the int value original form, big-endian bytes. bn254 store it in Montgometry format internally.

bo-abstrlabs:
@ahangsu Could you take another look? Thanks

bo-abstrlabs:
Locally the failed test is passing:
```
1|go-algorand (add-bn256) TESTFILTER=catchpointCatchupTest go test -v test/e2e-go/cli/goal/expect/goal_expect_test.go
=== RUN   TestGoalWithExpect
=== RUN   TestGoalWithExpect/catchpointCatchupTest.exp
    fixture.go:171: algoDir: /tmp/tmp854728164/expect/catchpointCatchupTest/algod
        testDataDir:/home/bo/workspace/go-algorand/test/testdata
--- PASS: TestGoalWithExpect (74.78s)
    --- PASS: TestGoalWithExpect/catchpointCatchupTest.exp (74.78s)
PASS
ok      command-line-arguments  74.796s
```

I check the catchpointCatchupTest.exp and it looks like flaky on CI unrelated to this change

bo-abstrlabs:
@ahangsu @jannotti Is there anything else we need do to merge this work? Our grant work is blocked by this feature. Thanks!

jannotti:
> @ahangsu @jannotti Is there anything else we need do to merge this work? Our grant work is blocked by this feature. Thanks!

Thanks for being patient. Changes like this can only be made with a consensus upgrade, and especially for cryptographic primitives, a lot of care about the libraries used.  We are now looking to have gnark-crypto audited, and want to have that done before adding these opcodes to general release (testnet and mainnet).  Can you meet your grant responsibilities in a devnet?  The simplest way forward would be to adjust this PR to add the opcodes in `pairingVersion` which would be (for now) set to 7 which makes it only available in vFuture networks.  This is how we have added things like fido opcodes, and randomness support, until they are finalized.  If that works for you, I can take it from here and get it merged.

bo-abstrlabs:
Thanks @jannotti for clarification!

> Can you meet your grant responsibilities in a devnet? The simplest way forward would be to adjust this PR to add the opcodes in pairingVersion which would be (for now) set to 7 which makes it only available in vFuture networks. This is how we have added things like fido opcodes, and randomness support, until they are finalized. If that works for you, I can take it from here and get it merged.

Good point! I set the opcode to be version 7: https://github.com/algorand/go-algorand/pull/3655/files#diff-ab0f8735d77364f4410c8b3a7af228ef2f318ea0dc6a6516524d2ede859037a0R538. And we've done a demo with a local algod of hacked consensus version (not in PR):
```
--- a/protocol/consensus.go
+++ b/protocol/consensus.go
@@ -188,7 +188,7 @@ const ConsensusFuture = ConsensusVersion(
 
 // ConsensusCurrentVersion is the latest version and should be used
 // when a specific version is not provided.
-const ConsensusCurrentVersion = ConsensusV32
+const ConsensusCurrentVersion = ConsensusFuture

```

Our next milestone will require consensus upgrade on testnet. I totally understand the extra care on making consensus upgrade and it would be a great first step to get this merged!

go.sum
```diff
@@ -1,5 +1,62 @@
...
+cloud.google.com/go/bigquery v1.7.0/go.mod h1://okPTzCYNXSlb24MZs83e2Do+h+VXtc4gLoIoXIAPc=
+cloud.google.com/go/bigquery v1.8.0/go.mod h1:J5hqkt3O0uAFnINi6JXValWIb1v0goeZM77hZzJN/fQ=
+cloud.google.com/go/datastore v1.0.0/go.mod h1:LXYbyblFSglQ5pkeyhO+Qmw7ukd3C+pD7TKLgZqpHYE=
+cloud.google.com/go/datastore v1.1.0/go.mod h1:umbIZjpQpHh4hmRpGhH4tLFup+FVzqBi1b3c64qFpCk=
+cloud.google.com/go/firestore v1.6.1/go.mod h1:asNXNOzBdyVQmEU+ggO8UPodTkEVFW5Qx+rwHnAz+EY=
+cloud.google.com/go/pubsub v1.0.1/go.mod h1:R0Gpsv3s54REJCy4fxDixWD93lHJMoZTyQ2kNxGRt3I=
+cloud.google.com/go/pubsub v1.1.0/go.mod h1:EwwdRX2sKPjnvnqCa270oGRyludottCI76h+R3AArQw=
+cloud.google.com/go/pubsub v1.2.0/go.mod h1:jhfEVHT8odbXTkndysNHCcx0awwzvfOlguIAii9o8iA=
+cloud.google.com/go/pubsub v1.3.1/go.mod h1:i+ucay31+CNRpDW4Lu78I4xXG+O1r/MAHgjpRVR+TSU=
+cloud.google.com/go/storage v1.0.0/go.mod h1:IhtSnM/ZTZV8YYJWCY8RULGVqBDmpoyjwiyrjsg+URw=
```
jannotti, line 43: DO you know where all this cloud.google.com stuff is coming from?

jannotti:
I've created https://github.com/algorand/go-algorand/pull/4013 from a branch I made in order to make some organizational changes.  When it passes checks and I have a little review here, I'll merge.

go.sum
bo-abstrlabs: Sorry not very sure about it. Looks like you figured out a clean go.sum in #4013

bo-abstrlabs:
Superseded by #4013

